PR #1184, opened 2015-03-05
remove unused and deprecated _adDupT function

MartinNowak:


yebblies:
Auto-merge toggled on

PR #1185, opened 2015-03-05
tweak vdparser benchmark to work with 2.066

rainers:
dmd 2.066 did not find any files under windows.

### DIFF
--- a/benchmark/gcbench/vdparser.d
+++ b/benchmark/gcbench/vdparser.d
@@ -28,7 +28,16 @@ void main(string[] argv)
     Project prj = new Project;
     foreach(i; 0..nIter)
     {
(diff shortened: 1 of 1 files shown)


yebblies:
Why is `globMatch` giving different results to `dirEntries` here

rainers:
Hmmm, it seems I have tested this with dmd 2.066.0. It doesn't list any file with
```
import std.stdio, std.file, std.path;

void main()
{
	foreach(string name; dirEntries(buildPath("gcbench", "vdparser.extra"), "*.d", SpanMode.depth))
		writeln(name);
}
```
when compiled with `dmd -O -inline -release -allinst test.d` and crashes for Win64. That seems a compiler bug that's not in dmd 2.066.1.

PR #1186, opened 2015-03-05
Issue 14247 - string within demangled symbol name should be made escape

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=14247


MartinNowak:
Auto-merge toggled on

PR #1187, opened 2015-03-06
make __setArrayAllocLength really pure

MartinNowak:
- use compare-and-swap instead of synchronized
  as the latter isn't really pure

- using cas instead of a global mutex
  also heavily reduces contention

### DIFF
--- a/src/rt/lifetime.d
+++ b/src/rt/lifetime.d
@@ -269,6 +269,8 @@ private class ArrayAllocLengthLock
   */
 bool __setArrayAllocLength(ref BlkInfo info, size_t newlength, bool isshared, const TypeInfo tinext, size_t oldlength = ~0) pure nothrow
 {
+    import core.atomic;
+
     size_t typeInfoSize = structTypeInfoSize(tinext);
 
     if(info.size <= 256)
@@ -281,15 +283,7 @@ bool __setArrayAllocLength(ref BlkInfo info, size_t newlength, bool isshared, co
         {
             if(isshared)
             {
-                try synchronized(typeid(ArrayAllocLengthLock))
-                {
-                    if(*length == cast(ubyte)oldlength)
-                        *length = cast(ubyte)newlength;
-                    else
-                        return false;
-                }
-                catch (Throwable t)
-                    assert(0, "Failed to synchronize.");
+                return cas(cast(shared)length, cast(ubyte)oldlength, cast(ubyte)newlength);
             }
             else
             {
@@ -301,7 +295,7 @@ bool __setArrayAllocLength(ref BlkInfo info, size_t newlength, bool isshared, co
         }
         else
         {
-            // setting the initial length, no lock needed
+            // setting the initial length, no cas needed
             *length = cast(ubyte)newlength;
         }
         if (typeInfoSize)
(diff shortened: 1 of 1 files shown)


src/rt/lifetime.d
```diff
@@ -281,15 +283,7 @@ bool __setArrayAllocLength(ref BlkInfo info, size_t newlength, bool isshared, co
...
-                try synchronized(typeid(ArrayAllocLengthLock))
-                {
-                    if(*length == cast(ubyte)oldlength)
-                        *length = cast(ubyte)newlength;
-                    else
-                        return false;
-                }
-                catch (Throwable t)
-                    assert(0, "Failed to synchronize.");
+                return cas(cast(shared)length, cast(ubyte)oldlength, cast(ubyte)newlength);
```
schveiguy: I'm curious how this even compiles. I thought pure functions could not use shared data?

MartinNowak: It's weakly pure, only changes stuff that's passed into the function and doesn't have a side-effect.
But for weakly pure functions any reference might alias global or shared data.
And also `shared` is a type qualifier not a storage class like `static`.

```d
void foo(ref shared(int) a) pure { a = 2; } // foo is weakly pure
shared int a;
void bar() { foo(a); } // bar is impure
```

MartinNowak: But locking a mutex really has a side effect and therefor isn't pure.

schveiguy: Well, I suppose that makes sense, but I don't understand the point of it. The entire point of weak pure functions is to be called from strong-pure functions. Clearly, a strong-pure function cannot pass in shared data.

I don't agree that locking a mutex is any different than touching shared data. But accessing the global typeid should not work (is there a bug report for that?)

MartinNowak: [Issue 14251 - synchronized (mtx) doesn't check attributes (pure, const)](http://issues.dlang.org/show_bug.cgi?id=14251)

MartinNowak: > I don't agree that locking a mutex is any different than touching shared data.

Once again shared is a type qualifier while pure ensures that you don't access global state, those are distinct concepts even though they often overlap.

This is perfectly pure.
```d
void foo() pure
{
    auto ary = new shared(int[])(10);
    ary.length = 20;
}
```

A mutex can change global state in the kernel which is a side-effect and thus not pure.

schveiguy: > This is perfectly pure.

And perfectly useless. May as well use `int[]` instead of `shared(int[])`.

Effectively all memory allocation uses a global mutex, so if that isn't pure, how does `new anything` work in a pure function? I know that memory allocation isn't actually pure, but shouldn't we apply the same rules to a mutex?

MartinNowak: Memory allocation has no real side effect. Before you allocate there is a GC with infinite memory, afterwards it's still a GC with infinite memory. When you lock a mutex you change it's state. Now one could argue, that many mutex implementations, e.g. a spinlock, only change the state of the mutex itself, but many mutex implementation will change some additional global variables, therefor synchronized can't be pure.

Also this change removes a severe contention bottleneck, so let's go on.

schveiguy: Well, synchronized shouldn't leave a mutex locked after the pure function is done. But as far as this change goes, it has benefits besides purity. I would be concerned with making all synchronized calls impure, but that isn't this change.

schveiguy:
Auto-merge toggled on

src/rt/lifetime.d
MartinNowak: We're trying to introduce lowering for `synchronized` in 2.068 so it can use more restrictive attributes, depending on the mutex implementation.

CyberShadow:
Could someone look at this please? Master is broken for me:
https://issues.dlang.org/show_bug.cgi?id=14291

MartinNowak:
See #1191.

PR #1188, opened 2015-03-06
roll back the synchronized nothrow change

MartinNowak:
- we're having trouble to change vibe.d's
  Mutex implementation to nothrow

- added correctly attributed internal functions to Mutex
  and Condition so that we don't have to roll back changes
  in core.thread and std.concurrency relying on nothrow


MartinNowak:
requires https://github.com/D-Programming-Language/phobos/pull/3042
both pull requests need to be merged together

MartinNowak:
More info on the vibe.d issue here https://github.com/rejectedsoftware/vibe.d/pull/972#issuecomment-77639611.

PR #1190, opened 2015-03-12
Make TickDuration "to" a freestanding, non-member function

mrkline:
This gives it the same semantics as all "normal" `to` functions in std.conv, allowing you to call it as `to!Duration(tickDur)` or in the current manner via UFCS. Fewer edge cases is good.

This also seems to fix conversions from `TickDuration` to `Duration`. At time of writing, the following fails to compile with the druntime master branch but compiles with this branch:

```d
import std.stdio;
import std.conv : to;
import std.datetime : Clock, Duration;

void main()
{
	write("System has been up for ");
	writeln(Clock.currSystemTick.to!Duration.total!"seconds", " seconds.");
}

```

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -1669,6 +1669,48 @@ unittest
 }
 
 /++
+    Converts a $(D TickDuration) to the given units as either an integral
+    value or a floating point value.
+
+    Params:
+        units = The units to convert to. Accepts $(D "seconds") and smaller
+                only.
+        T     = The type to convert to (either an integral type or a
+                floating point type).
+
+        td    = The TickDuration to convert
+  +/
+T to(string units, T, D)(D td) @safe pure nothrow @nogc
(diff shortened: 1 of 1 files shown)


MartinNowak:
TickDuration is about to be replaced by MonoTime.

MartinNowak:
Auto-merge toggled on

MartinNowak:
There are complaints about overload issues, see the reference above.

mrkline:
Hm. Perhaps I'm misunderstanding template rules, but how could this overshadow `to!string` when it takes a different number of template arguments and seems to have constraints that would preclude `string` from being the first template argument?

burner:
there is also a stray "this" inside

CyberShadow:
https://issues.dlang.org/show_bug.cgi?id=14530

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=15179

PR #1192, opened 2015-03-18
fix issue 14303 rt.util.container.array.Array has invalid unittest

braddr:
replace invalid test with something similar but valid.


MartinNowak:
Thanks for finding this. The problem here is that Array isn't @safe and might delete the existing array on insert. It would be possible to detect and handle this case in the insert.
Related to https://github.com/D-Programming-Language/druntime/pull/1143

PR #1193, opened 2015-03-18
OSX Events

etcimon:
The file is a copy of FreeBSD events.d but with minor changes in the enum identifiers for the EVTFILT_* names.


MartinNowak:
Nice, thanks Etienne.

PR #1194, opened 2015-03-20
GC: improve performance by inlining findBin

rainers:
This simplifies `ctfeBins` and inlines `findBin`, avoiding a gratuitious check. The original method was not inlined due to the static variable declared inside the `findBin`.

The benchmarks indicate an improvement by up to a few percent.


MartinNowak:
Auto-merge toggled on

PR #1196, opened 2015-03-22
core.sys.posix.stdio: Change dynamic to static array.

redstar:
Otherwise the unit test will not compile because of @nogc.

It's also a bug in 2.067.0-rc1.


MartinNowak:
Why didn't we notice, isn't that for tested?

redstar:
That is guess, too. The interface-only files are not compiled, only copied. Based on this they seem to be excluded from the unit tests, too.
(I only found this error because LDC was complaining.)

PR #1197, opened 2015-03-24
GC debug fixes

CyberShadow:


CyberShadow:
Not sure what's wrong with the rt.lifetime unittests. Ping @rainers @orvid

src/gc/gc.d
```diff
@@ -17,6 +17,10 @@ module gc.gc;
 
 /************** Debugging ***************************/
 
+// Note: Some of these need to be set globally
+// (for the entire Druntime) in the Makefile
+// instead of uncommenting the lines here.
+
```
rainers: AFAICT it's only SENTINEL that needs to be set globally to disable some unittests.

src/gc/gc.d
```diff
@@ -699,9 +703,9 @@ struct GC
 
 
     /**
-     * Attempt to in-place enlarge the memory block pointed to by p by at least
-     * minbytes beyond its current capacity, up to a maximum of maxsize.  This
-     * does not attempt to move the memory block (like realloc() does).
+     * Attempt to in-place enlarge the memory block pointed to by p to at least
+     * minsize bytes, up to a maximum of maxsize.
+     * This does not attempt to move the memory block (like realloc() does).
```
rainers: I don't like the current semantics too much, but extend is actually adding memory in the range minsize to maxsize. So the changed comment is incorrect.

src/gc/gc.d
```diff
@@ -3260,3 +3264,32 @@ else
...
+	import core.memory;
+	auto p = cast(ubyte*)GC.malloc(1);
+	assert(p[-1] == 0xF4);
+	assert(p[ 1] == 0xF5);
+
+	p[1] = 0;
+	bool thrown;
+	try
+		GC.free(p);
+	catch (Error e)
```
rainers: Catching an error thrown by the GC is problematic in a unittest, because it can leave the GC in a bad state, i.e. functions are nothrow so the stack isn't unwound, and the GC lock is not released.

rainers:
> Not sure what's wrong with the rt.lifetime unittests.

I don't see something obvious. Do you have more info on where the tests are failing?

CyberShadow:
The ones I disabled in 03ab3d65774c6009d2c8c893c43482cc90e7752d.

MartinNowak:
- We should restructure those debug tools as wrappers instead of polluting the GC code.
- The code in rt.lifetime is too specifically written for our current GC, maybe we need to move some functionality inside the GC.
- It's possible to use valgrind with gcstub to debug memory corruptions.
- Don't know why debug STOMP doesn't work with struct finalizers.

CyberShadow:
> It's possible to use valgrind with gcstub to debug memory corruptions.

Is there more information about this? I'm currently debugging memory corruption and was thinking about trying Valgrind.

However gcstub won't help find use-after-free bugs in cases where the GC incorrectly collected something.

> Don't know why debug STOMP doesn't work with struct finalizers.

The problem is with SENTINEL, not MEMSTOMP, so probably an alignment issue somewhere.

CyberShadow:
> I'm currently debugging memory corruption and was thinking about trying Valgrind.

I've got something working here:

https://github.com/CyberShadow/druntime/compare/pull-20150323-233811-gc-debug...valgrind

MartinNowak:
Pretty interesting.

MartinNowak:
> However gcstub won't help find use-after-free bugs in cases where the GC incorrectly collected something.

Works, because gcstub never frees, that's a problem on it's own though.

CyberShadow:
No, that's not the problem I'm talking about.

Say you allocated an object X, and stored the only pointer somewhere the GC can't see it. The GC collects it and allocates a new object at the same address. The code that then attempts to refer to X causes memory corruption.

This is a nontrivial problem even with Valgrind integration. What I'd need to do is make the GC collect objects (and stomp on freed memory), but instead of reusing the memory, keep allocating new memory, so that dangling pointers cause Valgrind errors. Any advice on the easiest way to do this?

MartinNowak:
Just comment out the cleaning of the free bits.

MartinNowak:
Somewhere in sweep.
This https://github.com/D-Programming-Language/druntime/blob/95737c0d947fa9096de8e8123440fbc9ba0b9153/src/gc/gc.d#L2196 and that https://github.com/D-Programming-Language/druntime/blob/95737c0d947fa9096de8e8123440fbc9ba0b9153/src/gc/gc.d#L2270.

rainers:
> Not sure what's wrong with the rt.lifetime unittests.

The runFinalizer functions pass the wrong size to the `rt_hasFinalizerInSegment` and `rt_finalizeFromGC` functions. It should be something like this:
```
            debug(SENTINEL)
                immutable qsize = *sentinel_size(p);
            else
                immutable qsize = size;
            if(!rt_hasFinalizerInSegment(p, qsize, attr, segment))
                continue;

            rt_finalizeFromGC(p, qsize, attr);
```
(replace p with q for the small object pool).

src/gc/gc.d
```diff
@@ -23,6 +23,8 @@ module gc.gc;
 //debug = LOGGING;              // log allocations / frees
 //debug = MEMSTOMP;             // stomp on memory
 //debug = SENTINEL;             // add underrun/overrrun protection
+                                // note: this needs to be enabled globally in the makefiles
+                                // (-debug=SENTINEL) instead of uncommented here.
```
rainers: I think you should restrict the comment to building the unittests, e.g. ".. when building druntime unittests".

src/gc/gc.d
rainers: You could used version(unittest) to change the behaviour of the invariant, so it could just count errors while executing this unittest. Not sure how important that is here, though. Maybe just add a comment why the test isn't executed.

MartinNowak:
Please fixup the finalizer if you find time @CyberShadow.

CyberShadow:
Sorry, forgot about this pull. Will do.

PR #1198, opened 2015-03-24
When calling assumeSafeAppend, cache the block info for subsequent calls

schveiguy:
Library code may conservatively call assumeSafeAppend on an array without knowing if the block will be appended to. However, if the block is not cached in the block info cache, it would not add it to the cache. This meant that code that continually calls assumeSafeAppend is constantly looking up the block info from the heap. This fixes that situation.

### DIFF
--- a/src/rt/lifetime.d
+++ b/src/rt/lifetime.d
@@ -637,7 +637,8 @@ extern(C) void _d_arrayshrinkfit(const TypeInfo ti, void[] arr) /+nothrow+/
     auto tinext = unqualify(ti.next);
     auto size = tinext.tsize;                  // array element size
     auto cursize = arr.length * size;
-    auto bic = __getBlkInfo(arr.ptr);
+    auto isshared = typeid(ti) is typeid(TypeInfo_Shared);
+    auto bic = !isshared ? __getBlkInfo(arr.ptr) : null;
     auto info = bic ? *bic : GC.query(arr.ptr);
     if(info.base && (info.attr & BlkAttr.APPENDABLE))
     {
@@ -660,6 +661,10 @@ extern(C) void _d_arrayshrinkfit(const TypeInfo ti, void[] arr) /+nothrow+/
         // Since it is not shared, we also know it won't throw (no lock).
         if (!__setArrayAllocLength(info, newsize, false, tinext))
             onInvalidMemoryOperationError();
+
+        // cache the block if not already done.
+        if(!isshared && !bic)
+            __insertBlkInfoCache(info, bic);
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/rt/lifetime.d
```diff
@@ -637,7 +637,8 @@ extern(C) void _d_arrayshrinkfit(const TypeInfo ti, void[] arr) /+nothrow+/
     auto tinext = unqualify(ti.next);
     auto size = tinext.tsize;                  // array element size
     auto cursize = arr.length * size;
-    auto bic = __getBlkInfo(arr.ptr);
+    auto isshared = typeid(ti) is typeid(TypeInfo_Shared);
+    auto bic = !isshared ? __getBlkInfo(arr.ptr) : null;
```
andralex: auto bic = isshared ? null : __getBlkInfo(arr.ptr);

src/rt/lifetime.d
```diff
@@ -660,6 +661,10 @@ extern(C) void _d_arrayshrinkfit(const TypeInfo ti, void[] arr) /+nothrow+/
         // Since it is not shared, we also know it won't throw (no lock).
         if (!__setArrayAllocLength(info, newsize, false, tinext))
             onInvalidMemoryOperationError();
+
+        // cache the block if not already done.
+        if(!isshared && !bic)
```
andralex: space after `if`

schveiguy:
Updated with suggestions.

MartinNowak:
Cache on read, makes sense.

MartinNowak:
Auto-merge toggled on

schveiguy:
I actually considered last night, why put the cache only for array appending, why not just have a general `GC.query` cache? I doubt there are many occasions to use `GC.query`, and this would kind of move it into the right place (right now, there is this dance during collection where the GC tells the runtime about destroyed blocks so it can update the block info cache). Thoughts @MartinNowak?

schveiguy:
Hm... but the cache depends on knowing that the block is unshared. Maybe that wouldn't work in the general case.

MartinNowak:
We will add thread local caches for the GC soon, and one part of the will be to cache the info, making a separate cache unnecessary.
But yes, it would be a good first step to move the cache to GC.query.

schveiguy:
Yeah, a GC that knows which blocks are shared or not seems like a prerequisite. Looking forward to that!

MartinNowak:
That's just an optimization to avoid synchronization, but the GC has to deal with that anyhow.
As I already discussed with you we can't afford to synchronize attributes, so appending to an array from two threads might perform superfluous reallocs.

PR #1199, opened 2015-03-24
gc.gc: Fix out-of-bounds pagetable access

CyberShadow:
The out-of-bounds access occurs 3 lines below:

`pagetable[i + p]`

We never check that `i + p < npages`.

This patch also acts as a small optimization (don't look at the last `n-1` pages because it's impossible to find a free chunk of at least `n` pages from those positions).

Found using Vagrind: https://github.com/D-Programming-Language/druntime/pull/1197#issuecomment-85671211

### DIFF
--- a/src/gc/gc.d
+++ b/src/gc/gc.d
@@ -2944,7 +2944,7 @@ struct LargeObjectPool
         {
             assert(pagetable[i] == B_FREE);
             size_t p = 1;
-            while (p < n && pagetable[i + p] == B_FREE)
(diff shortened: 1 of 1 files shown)


rainers:
IIRC the check `i < npages` was delibirately chosen to not get wrong results for `largest`. I guess the fix should go into the while loop accessing `pagetable[i+p]`.

CyberShadow:
Oh, OK.

MartinNowak:
Auto-merge toggled on

PR #1200, opened 2015-03-27
fix Issue 14350 - Unit test failures are not displayed in Windows GUI pr...

CyberShadow:
...ograms

https://issues.dlang.org/show_bug.cgi?id=14350

### DIFF
--- a/src/core/runtime.d
+++ b/src/core/runtime.d
@@ -42,7 +42,7 @@ private
     extern (C) TraceHandler rt_getTraceHandler();
 
     alias void delegate( Throwable ) ExceptionHandler;
-    extern (C) void printThrowable(Throwable t);
+    extern (C) void _d_print_throwable(Throwable t);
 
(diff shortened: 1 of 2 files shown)


CyberShadow:
BTW, shouldn't Druntime `extern(C)` functions have a `_d_` prefix OSLT? We're polluting the C namespace.

MartinNowak:
Yes, can you please change it to _d_print_throwable?

CyberShadow:
Done, though this is one of many.

Also I hope this won't break code in case people are calling this function manually from their WinMain functions OSLT.

MartinNowak:
Used to be `private void printThrowable(Throwable t)` (even `extern(D)`), pretty hard to call that one.

MartinNowak:
Auto-merge toggled on

CyberShadow:
Oh, doh. Right.

PR #1201, opened 2015-03-29
Solaris: Change array declaration from C to D style.

redstar:


9rnsr:
Auto-merge toggled on

PR #1203, opened 2015-03-31
Mark core.memory.GC.addRange/removeRange as nogc

rsw0x:
NOTE: this is my first druntime PR, please be sure to check it extra carefully and feel free to give me tips.

This addresses [issue 14171](https://issues.dlang.org/show_bug.cgi?id=14171). Phobos utilities that use the non-allocating GC functions in areas that should be nogc(see: std.container.array, refcounted, etc,) this should make them able to infer nogc.

I'm completely unsure of how to handle throwing errors in nogc, what I did is probably completely incorrect(? I know it shouldn't be done for exceptions - are errors an exception(heh) to this?) so I could use some advice.

### DIFF
--- a/src/core/sync/mutex.d
+++ b/src/core/sync/mutex.d
@@ -134,7 +134,7 @@ class Mutex :
     }
 
     // undocumented function for internal use
-    final void lock_nothrow() nothrow @trusted
+    final void lock_nothrow() nothrow @trusted @nogc
     {
         version( Windows )
         {
@@ -144,7 +144,11 @@ class Mutex :
         {
             int rc = pthread_mutex_lock( &m_hndl );
             if( rc )
-                throw new SyncError( "Unable to lock mutex" );
+            {
+                SyncError syncErr = cast(SyncError) cast(void*) typeid(SyncError).init;
+                syncErr.msg = "Unable to lock mutex.";
+                throw syncErr;
+            }
         }
     }
 
@@ -161,7 +165,7 @@ class Mutex :
     }
 
     // undocumented function for internal use
-    final void unlock_nothrow() nothrow @trusted
+    final void unlock_nothrow() nothrow @trusted @nogc
     {
         version( Windows )
         {
@@ -171,7 +175,11 @@ class Mutex :
         {
             int rc = pthread_mutex_unlock( &m_hndl );
             if( rc )
-                throw new SyncError( "Unable to unlock mutex" );
+            {
+                SyncError syncErr = cast(SyncError) cast(void*) typeid(SyncError).init;
+                syncErr.msg = "Unable to unlock mutex.";
+                throw syncErr;
+            }
         }
     }
 
--- a/src/gc/gc.d
+++ b/src/gc/gc.d
@@ -258,12 +258,12 @@ const uint GCVERSION = 1;       // increment every time we change interface
 // This just makes Mutex final to de-virtualize member function calls.
 final class GCMutex : Mutex
 {
-    final override void lock() nothrow @trusted
+    final override void lock() nothrow @trusted @nogc
     {
         super.lock_nothrow();
     }
 
-    final override void unlock() nothrow @trusted
+    final override void unlock() nothrow @trusted @nogc
     {
         super.unlock_nothrow();
     }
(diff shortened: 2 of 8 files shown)


src/core/sync/mutex.d
```diff
@@ -170,8 +171,9 @@ class Mutex :
         else version( Posix )
         {
             int rc = pthread_mutex_unlock( &m_hndl );
+            static immutable syncEx = new SyncError("Unable to unlock mutex");
             if( rc )
-                throw new SyncError( "Unable to unlock mutex" );
+                throw syncEx;
```
JakobOvrum: The exception mechanism doesn't integrate at all with `const`, `immutable` and `shared`, so this is no good.

`onOutOfMemoryError` does it like this:
```d
throw cast(OutOfMemoryError) cast(void*) typeid(OutOfMemoryError).init;
```
But I'm guessing `TypeInfo` isn't TLS either, so it's really almost as bad. Statically allocated exceptions might be an acceptable hack for `Error` exceptions, but it's not a long-term solution unless we change the chaining mechanism.

See [threads such as this one](http://forum.dlang.org/post/stlslhjndgugecvmbowd@forum.dlang.org) for more information.

rsw0x: Yeah, that was the other "solution" to this that I found - but both seem to be giant hacks so I thought I'd just submit a PR and see what people better than I am in D think should be done here.

Thanks for the link, I'll look it over.

MartinNowak:
Go with the typeid hack for now, we'll figure out how to fix it later.
https://issues.dlang.org/show_bug.cgi?id=14119

src/core/sync/mutex.d
```diff
@@ -144,7 +144,7 @@ class Mutex :
         {
             int rc = pthread_mutex_lock( &m_hndl );
             if( rc )
-                throw new SyncError( "Unable to lock mutex" );
+                throw cast(SyncError) cast(void*) typeid(SyncError).init;
```
rsw0x: Did you want me to keep the error message in or not?(I was unsure if this was UB due to the init hack.)

MartinNowak: Yes, would be nice to keep the message.

rsw0x:
Is that fine as is, or does __ctor need called?(would have to mark quite a bit more as nogc if so, I believe.)

src/core/sync/mutex.d
```diff
@@ -128,13 +128,13 @@ class Mutex :
      * Throws:
      *  SyncError on error.
      */
-    @trusted void lock()
+    @trusted @nogc void lock()
```
MartinNowak: Please don't change that function, it's part of the public API, and inherited classes might not implement `@nogc`.

src/core/sync/mutex.d
```diff
@@ -155,13 +159,13 @@ class Mutex :
      * Throws:
      *  SyncError on error.
      */
-    @trusted void unlock()
+    @trusted @nogc void unlock()
```
MartinNowak: Same here.

src/gc/gc.d
```diff
@@ -1128,7 +1128,7 @@ struct GC
     /**
      * add range to scan for roots
      */
-    void addRange(void *p, size_t sz, const TypeInfo ti = null) nothrow
+    @nogc void addRange(void *p, size_t sz, const TypeInfo ti = null) nothrow
```
MartinNowak: Attributes on right side of the function please.

src/rt/util/container/treap.d
```diff
@@ -32,7 +32,7 @@ nothrow:
         rand48.defaultSeed();
     }
 
-    void insert(E element)
+    @nogc void insert(E element)
```
MartinNowak: Maybe you can mark the whole module/struct @nogc?

src/rt/util/random.d
```diff
@@ -24,19 +24,19 @@ nothrow:
         popFront();
     }
 
-    auto opCall()
+    @nogc auto opCall()
```
MartinNowak: Same here, just put `@nogc` at the top of the struct, like the nothrow.

MartinNowak:
Rest LGTM.

src/rt/util/container/treap.d
rsw0x: the @nogc in treap propagates very far down the GC due to the opApply delegates and I'm not very familiar with it at all.

MartinNowak:
Auto-merge toggled on

MartinNowak:
Thanks a lot rsw0x.

PR #1204, opened 2015-04-02
fmemopen, open_memstream unavail on OSX

smolt:
The memstream family of POSIX functions are not available as of MacOSX 10.10.

Something like this is needed to link all unittests on OS X (and iOS).  Not sure of the the best approach to disabling these POSIX functions because other OS might be involved, but wanted to start the conversation with this pull request.

### DIFF
--- a/src/core/sys/posix/stdio.d
+++ b/src/core/sys/posix/stdio.d
@@ -199,10 +199,12 @@ FILE*  popen(in char*, in char*);
 
 // memstream functions are conforming to POSIX.1-2008.  These functions are
 // not specified in POSIX.1-2001 and are not widely available on other
-// systems.  For now list systems without, but may make more sense to instead
-// specify systems that do have.
-version( OSX ) {}
-else
+// systems.
+version( linux )                             // as of glibc 1.0x
+    version = HaveMemstream;
+else version( FreeBSD )                      // as of FreeBSD 9.2
+    version = HaveMemstream;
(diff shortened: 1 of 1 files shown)


smolt:
@redstar suggested that @MartinNowak take a look.

MartinNowak:
We usually take a whitelist approach, only enable something when we're sure it exists.

smolt:
Of the TARGETS that define Posix in mars.c, man pages say they exist in OpenBSD, FreeBSD, and of course linux glibc since the beginning of time.  For the BSDes, it was added:
- FreeBSD 9.2 2013-09-30
- OpenBSD 5.4 2013-11-01
I'll revise to these and see what you think.  I can squash if that helps to make a better final commit.

MartinNowak:
Auto-merge toggled on

MartinNowak:
Looks fine, thanks.

braddr:
Those functions must not be used anywhere in druntime or phobos, or freebsd 8.x supports them.  The freebsd auto testers are all 8.x based.

MartinNowak:
Thanks Brad, why do the tests currently pass though?

MartinNowak:
Auto-merge toggled off

smolt:
The unittest doesn't appear to be run for core.sys.posix.stdio.  It does get compiled and linked in LDC's druntime-test-runner, that is how it showed up as unresolved symbols on OS X during 2.067 merge into LDC.

PR #1205, opened 2015-04-03
Issue 14401 - typeid(shared X).init is empty for class types

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=14401

`TypeInfo_Class` needs to override `init()` method for the `TypeInfo_Const` and its derived TypeInfo classes.

Now all `typeid(X).init` for any type `X` is typed as `const(void)[]`.
This is a small but necessary breaking change. If a class instance contains reference type fields, the instance image must be typed as `const(void)[]` rather than `byte[n]`.

### DIFF
--- a/src/object.di
+++ b/src/object.di
@@ -183,7 +183,7 @@ class TypeInfo_Class : TypeInfo
     @property auto info() @safe nothrow pure const { return this; }
     @property auto typeinfo() @safe nothrow pure const { return this; }
 
-    byte[]      init;   // class static initializer
+    byte[]      m_init; // class static initializer
     string      name;   // class name
     void*[]     vtbl;   // virtual function pointer table
     Interface[] interfaces;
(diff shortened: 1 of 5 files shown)


src/object.di
```diff
@@ -183,7 +183,7 @@ class TypeInfo_Class : TypeInfo
     @property auto info() @safe nothrow pure const { return this; }
     @property auto typeinfo() @safe nothrow pure const { return this; }
 
-    byte[]      init;   // class static initializer
```
andralex: urgh... we really should have a compiler error for this stuff

andralex:
Auto-merge toggled on

andralex:
@9rnsr thanks! Could you please take a look at the failures, they seem related.

schveiguy:
compiler errors should be fixed with the merge of https://github.com/D-Programming-Language/phobos/issues/3143

andralex:
@schveiguy thanks

PR #1206, opened 2015-04-05
add win32coff targets

rainers:
This adds targets `druntime32mscoff` and `unittest32mscoff` to win64.mak.

The 32-bit compiler is derived from the setting for the 64-bit compiler, but can be overwritten on the make command line with `CC32=`

The usual conditions for `VCDIR` and `SDKDIR` apply, ~~though the default is improved by using `$(ProgramFiles)`~~. As for Win64 since the `-conf=` setting, it is necessary to set `LINKCMD`and `LIB` in the environment to link the unittests.

### DIFF
--- a/win64.mak
+++ b/win64.mak
@@ -15,6 +15,8 @@ CP=cp
 DOCDIR=doc
 IMPDIR=import
 
+MAKE=make
+
 DFLAGS=-m$(MODEL) -conf= -O -release -dip25 -inline -w -Isrc -Iimport
 UDFLAGS=-m$(MODEL) -conf= -O -release -dip25 -w -Isrc -Iimport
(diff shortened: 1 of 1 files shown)


rainers:
The build fails because the tester tries to patch the makefile instead of passing changed variables as arguments.

MartinNowak:
Guess you'll have to make a pull for the d-tester as well.
https://github.com/braddr/d-tester/tree/master/client/src

rainers:
I have removed the rather unrelated changes to VCDIR and SDKDIR.

MartinNowak:
Seems like the patch still doesn't apply.

rainers:
The addition of the `MAKE=make` line was still in the context of the build server patch.

MartinNowak:
Auto-merge toggled on

PR #1207, opened 2015-04-06
Add TypeInfo classes for cent/ucent.

redstar:
The types are defined in the spec but not yet implemented. Add the
required TypeInfo classes for supporting compilers. The classes are
guarded to prevent compile errors on non-supporting compilers.

See https://github.com/D-Programming-Language/dmd/pull/4565 for
the name mangling.


MartinNowak:
Auto-merge toggled on

MartinNowak:
If we decide to go with a different mangling this can still be changed.

PR #1208, opened 2015-04-07
fix Issue 12891: add atomicFetchAdd to core.atomic

jadbox:
This is the placeholder pull request for the work on improving performance around fetchAdd atomic operations. This has a new implementation for += and -= atomicOp and provides a new method atomicFetchAdd. I'd debate if we require atomicFetchSub for this feature, but I'm open to thoughts.

Ideally, I need help for someone to benchmark this PR to see if it does indeed improve performance on x86 and x86_x64.

https://issues.dlang.org/show_bug.cgi?id=12891

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -239,14 +239,6 @@ else version( AsmX86_32 )
         // +    -   *   /   %   ^^  &
         // |    ^   <<  >>  >>> ~   in
         // ==   !=  <   <=  >   >=
-        static if( op == "+=" && __traits(isIntegral, T) && is(T==V1) ) {
-            return atomicFetchAdd!(T)(val, mod);
-        }
-        else
-        static if( op == "-=" && __traits(isIntegral, T) && is(T==V1) ) {
-            return atomicFetchAdd!(T)(val, -mod);
-        }
-        else
         static if( op == "+"  || op == "-"  || op == "*"  || op == "/"   ||
                    op == "%"  || op == "^^" || op == "&"  || op == "|"   ||
                    op == "^"  || op == "<<" || op == ">>" || op == ">>>" ||
@@ -262,6 +254,14 @@ else version( AsmX86_32 )
         //
         // +=   -=  *=  /=  %=  ^^= &=
         // |=   ^=  <<= >>= >>>=    ~=
+        static if( op == "+=" && __traits(isIntegral, T) && is(T==V1) ) {
+            return atomicFetchAdd!(T)(val, mod);
+        }
+        else
+        static if( op == "-=" && __traits(isIntegral, T) && is(T==V1) ) {
+            return atomicFetchAdd!(T)(val, -mod);
+        }
+        else
         static if( op == "+=" || op == "-="  || op == "*="  || op == "/=" ||
                    op == "%=" || op == "^^=" || op == "&="  || op == "|=" ||
                    op == "^=" || op == "<<=" || op == ">>=" || op == ">>>=" ) // skip "~="
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -180,7 +180,7 @@ else version( AsmX86_32 )
         static if (T.sizeof == 1) asm pure nothrow @nogc { lock; xadd[EDX], AL; }
         else static if (T.sizeof == 2) asm pure nothrow @nogc { lock; xadd[EDX], AX; }
         else static if (T.sizeof == 4) asm pure nothrow @nogc { lock; xadd[EDX], EAX; }
-        
+            
         asm pure nothrow @nogc 
         { 
             mov mod, EAX; 
@@ -1518,20 +1518,20 @@ version( unittest )
         shared ubyte u8 = 1;
         shared ushort u16 = 2;
         shared uint u32 = 3;
-        shared ulong u64 = 4;
         shared byte i8 = 5;
         shared short i16 = 6;
         shared int i32 = 7;
-        shared long i64 = 8;
      
         assert(atomicOp!"+="(u8, 8) == 9);
         assert(atomicOp!"+="(u16, 8) == 10);
         assert(atomicOp!"+="(u32, 8) == 11);
-        assert(atomicOp!"+="(u64, 8) == 12);
         assert(atomicOp!"+="(i8, 8) == 13);
         assert(atomicOp!"+="(i16, 8) == 14);
         assert(atomicOp!"+="(i32, 8) == 15);
         version( AsmX86_64 ) {
+            shared ulong u64 = 4;
+            shared long i64 = 8;
+            assert(atomicOp!"+="(u64, 8) == 12);
             assert(atomicOp!"+="(i64, 8) == 16);
         }
     }
@@ -1541,20 +1541,20 @@ version( unittest )
         shared ubyte u8 = 1;
         shared ushort u16 = 2;
         shared uint u32 = 3;
-        shared ulong u64 = 4;
         shared byte i8 = 5;
         shared short i16 = 6;
         shared int i32 = 7;
-        shared long i64 = 8;
 
         assert(atomicOp!"-="(u8, 1) == 0);
         assert(atomicOp!"-="(u16, 1) == 1);
         assert(atomicOp!"-="(u32, 1) == 2);
-        assert(atomicOp!"-="(u64, 1) == 3);
         assert(atomicOp!"-="(i8, 1) == 4);
         assert(atomicOp!"-="(i16, 1) == 5);
         assert(atomicOp!"-="(i32, 1) == 6);
         version( AsmX86_64 ) {
+            shared ulong u64 = 4;
+            shared long i64 = 8;
+            assert(atomicOp!"-="(u64, 1) == 3);
             assert(atomicOp!"-="(i64, 1) == 7);
         }
     }
@@ -1564,20 +1564,20 @@ version( unittest )
         shared ubyte u8 = 1;
         shared ushort u16 = 2;
         shared uint u32 = 3;
-        shared ulong u64 = 4;
         shared byte i8 = 5;
         shared short i16 = 6;
         shared int i32 = 7;
-        shared long i64 = 8;
-
+        
         assert(atomicFetchAdd(u8, 8) == 9);
         assert(atomicFetchAdd(u16, 8) == 10);
         assert(atomicFetchAdd(u32, -1) == 2);
-        assert(atomicFetchAdd(u64, -1) == 3);
         assert(atomicFetchAdd(i8, -1) == 4);
         assert(atomicFetchAdd(i16, -1) == 5);
         assert(atomicFetchAdd(i32, -1) == 6);
         version( AsmX86_64 ) {
+            shared ulong u64 = 4;
+            shared long i64 = 8;
+            assert(atomicFetchAdd(u64, -1) == 3);
             assert(atomicFetchAdd(i64, -1) == 7);
         }
     }
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -159,7 +159,7 @@ else version( AsmX86_32 )
 {
     // Uses specialized asm for fast fetch and add operations
     private HeadUnshared!(T) atomicFetchAdd(T)( ref shared T val, size_t mod ) pure nothrow @nogc
-        if( __traits(isIntegral, T) )
+        if( __traits(isIntegral, T) && T.sizeof <= 4)
     in
     {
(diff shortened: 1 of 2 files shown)


MartinNowak:
Yes, there should also be a `fetchSub` because it matches a different intrinsic.
https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#index-g_t_005f_005fatomic_005fadd_005ffetch-3535
It might be a better overall design to not introduce new function, but to only change the internal implementation of atomicOp.

src/core/atomic.d
```diff
@@ -190,6 +254,14 @@ else version( AsmX86_32 )
         //
         // +=   -=  *=  /=  %=  ^^= &=
         // |=   ^=  <<= >>= >>>=    ~=
+        static if( op == "+=" && __traits(isIntegral, T) && is(T==V1) ) {
```
MartinNowak: Try `pragma(msg, T, " == ", V1, " = ", is(T == V1), " (op: ", op, ")");` just on the function entry below `body`.

The `is(T == V1)` constraint will miss half of the cases where this optimization is applicable, e.g. `atomicOp!"+="(u8, 1)`.

Note that you can also do this, which is defined in terms of `int` promoted add with subsequent coercion to `ubyte`.
```d
    ubyte u8;
    int val = 1;
    u8 += val;
```

I think we can implement this like the following.
```d
size_t tmp = mod; // convert all operands to size_t

asm
{
    mov RAX, tmp;
    mov RDX, val;
}

static if (T.sizeof == 1) asm { lock; xadd[RDX], AL; }
else static if (T.sizeof == 2) asm { lock; xadd[RDX], AX; }
else static if (T.sizeof == 4) asm { lock; xadd[RDX], EAX; }
else static if (T.sizeof == 8) asm { lock; xadd[RDX], RAX; }

asm { mov tmp, RAX; }

return cast(T)(tmp + mod);
```

src/core/atomic.d
```diff
@@ -56,6 +56,70 @@ version( AsmX86 )
...
+    }
+    body
+    {
+        HeadUnshared!(T) oval = void;
+        static if (T.sizeof == 1)
+        {
+            asm pure nothrow @nogc
+            {
+                mov AL, mod;
+                mov RDX, val;
```
MartinNowak: Those `RDX` registers are for `X86_64` only, you need to copy&paste the code and use `EDX` for 32-bit `X86`.

src/core/atomic.d
jadbox: size_t tmp = mod; // convert all operands to size_t
^ Does this incur any performance hit to copy into tmp? Would the below work and possibly be faster:
atomicFetchAdd(T)( ref shared T val, **size_t** mod )

src/core/atomic.d
jadbox: Thanks, it's been almost a decade since I've written asm. =]

src/core/atomic.d
jadbox: What does this code do:
asm { mov tmp, RAX; }
return cast(T)(tmp + mod);

I also didn't understand why this:
return oval + mod;

Why is any addition going on here (and why with mod)?

jadbox: Btw, the pragma wrote this in the console for the unittest:

ulong == ulong = true (op: +=)
ulongulong ==  == ulong = true (op: ulong-=)
 = true (op: +=)
ulong == ulong = true (op: -=)
int == int = true (op: +=)
int == int = true (op: -=)
ulong == ulong = true (op: +=)
ulong == ulong = true (op: -=)

jadbox:
>> It might be a better overall design to not introduce new function, but to only change the internal implementation of atomicOp.

Agreed. I've made atomicFetchAdd private. I think separating this logic into another function (at least internally) aids code readability here.

jadbox:
@MartinNowak  Btw, please read my replies on your previous code comments. (GitHub hides them once a new commit is made)

src/core/atomic.d
MartinNowak: > Why is any addition going on here (and why added with mod.. isn't xadd doing this job)?

XADD returns you the value before the addition, so you have to add it again to return the total result.

MartinNowak:
Well it turned out to be a bit more involved, but thanks for driving this.

src/core/atomic.d
```diff
@@ -190,6 +222,14 @@ else version( AsmX86_32 )
         //
         // +=   -=  *=  /=  %=  ^^= &=
         // |=   ^=  <<= >>= >>>=    ~=
+        static if( op == "+=" && __traits(isIntegral, T) ) {
```
MartinNowak: Looks like you need a `&& T.sizeof <= 4` for the 32 bit case.

src/core/atomic.d
```diff
@@ -157,6 +157,38 @@ version( CoreDdoc )
 }
 else version( AsmX86_32 )
 {
+    // Uses specialized asm for fast fetch and add operations
+    private HeadUnshared!(T) atomicFetchAdd(T)( ref shared T val, size_t mod ) pure nothrow @nogc
+        if( __traits(isIntegral, T) )
```
MartinNowak: Also add `T.sizeof <= 4` as template constraint.

jadbox:
I couldn't find any good info online on how xaddl and xadd differs. Is there a reason to use one over the other? 

Btw, I'm thinking that for +=1 and -=1 atomicOp calls (using a static if check) will instead use atomic INC asm calls. This way, you transparently are getting the best performance possible.

In atomicFetchAdd, I'm thinking something like this:
static if(mod==1) {
            asm pure nothrow @nogc
            {
                mov RDX, val;
                lock; inc[RDX];
            }
            return ????
else { ...

jadbox:
>> XADD returns you the value before the addition, so you have to add it again to return the total result.
I'm still confused.. why do we need to "add it again"? Doesn't this mean that this atomic action is then taking two addition ops to complete.. and hence be slow? (asm noob speaking)

MartinNowak:
> Btw, I'm thinking that for +=1 and -=1 atomicOp calls (using a static if check) will instead use atomic INC asm calls. This way, you transparently are getting the best performance possible.

No need to, `inc` won't be faster than `xadd`, and you'd need to make the 1 a template parameter to check this.

> I'm still confused.. why do we need to "add it again"?

GCC has 2 builtins, [atomic_add_fetch](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#index-g_t_005f_005fatomic_005fadd_005ffetch-3535) and [atomic_fetch_add](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#index-g_t_005f_005fatomic_005ffetch_005fadd-3541). The former one returns the value __after__ adding the op, the latter returns the value __before__ adding the op, this is also the behavior of `XADD`, so implementing `+=` requires us to add the op again.

> oesn't this mean that this atomic action is then taking two addition ops to complete.. and hence be slow? (asm noob speaking)

Nope, the second add simply adds 2 registers, this is literally free (your CPU can perform 3 independent adds per cycle). Doing `LOCK XADD` on a memory location synchronizes the memory bus and costs 20-50 cycles latency.

src/core/atomic.d
```diff
@@ -157,6 +157,38 @@ version( CoreDdoc )
...
+        if( __traits(isIntegral, T) && T.sizeof <= 4)
+    in
+    {
+        // NOTE: 32 bit x86 systems support 8 byte CAS, which only requires
+        //       4 byte alignment, so use size_t as the align type here.
+        static if( T.sizeof > size_t.sizeof )
+            assert( atomicValueIsProperlyAligned!(size_t)( cast(size_t) &val ) );
+        else
+            assert( atomicValueIsProperlyAligned!(T)( cast(size_t) &val ) );
+    }
```
MartinNowak: We don't need this in contract here, it is only needed for 8-byte CAS on x86-32.

src/core/atomic.d
```diff
@@ -190,6 +222,14 @@ else version( AsmX86_32 )
         //
         // +=   -=  *=  /=  %=  ^^= &=
         // |=   ^=  <<= >>= >>>=    ~=
+        static if( op == "+=" && __traits(isIntegral, T) && T.sizeof <= 4) {
+            return atomicFetchAdd!(T)(val, mod);
+        }
+        else
+        static if( op == "-=" && __traits(isIntegral, T) && T.sizeof <= 4) {
+            return atomicFetchAdd!(T)(val, -mod);
```
MartinNowak: How about adding a private `atomicSubFetch` that simply calls `atomicFetchAdd!(T)(val, -mod)`?
That way the primitive is encapsulated and the GDC/LDC people can easily map it to their intrinsics.

src/core/atomic.d
```diff
@@ -157,6 +157,38 @@ version( CoreDdoc )
 }
 else version( AsmX86_32 )
 {
+    // Uses specialized asm for fast fetch and add operations
+    private HeadUnshared!(T) atomicFetchAdd(T)( ref shared T val, size_t mod ) pure nothrow @nogc
```
MartinNowak: Apparently (see https://github.com/D-Programming-Language/druntime/pull/1208#issuecomment-91189879), we should rename this to `atomicAddFetch`.

MartinNowak: Even better, we keep this as `atomicFetchAdd`, and do the post adding in `atomicOp`.
This is because C11 and C++ only standardize `[atomic_fetch_add](http://en.cppreference.com/w/cpp/atomic/atomic_fetch_add)`.
So this function should simply `return cast(T)tmp;` and atomicOp should instead `return cast(T)(atomicFetchAdd!(T)(val, mod) + mod);`.

MartinNowak:
Motivating benchmark.
```d
import core.atomic, std.algorithm, std.datetime, std.stdio;

void main()
{
    auto dur = Duration.max;
    foreach (i; 0 .. 20)
    {
        auto sw = StopWatch(AutoStart.yes);
        shared uint rc;
        foreach (_; 0 .. 1_000_000)
        {
            atomicOp!"+="(rc, 1);
            atomicOp!"-="(rc, 1);
        }
        dur = min(dur, cast(Duration)sw.peek);
    }
    writeln(dur);
}
```
Running this with your pull gives me `18 ms` instead of `26 ms`, the old code was 44% slower/new code is 30% faster.

src/core/atomic.d
jadbox: Removed contract

jadbox:
>> implementing += requires us to add the op again.
My powers of deduction are failing me. Why doesn't += return the result of an AddFetch (without the second addition)...as the += expression means "return me the value of myself after the add".

>> Even better, we keep this as atomicFetchAdd, and do the post adding in atomicOp
Done in last commit!

Also added atomicFetchSub.

MartinNowak:
Auto-merge toggled on

MartinNowak:
Thanks.

There exists no Intel instruction that atomically adds a number to a variable and returns the result.
You only have xadd which adds a number and returns the value of the variable __before__ the adding.
So in order to get the result __after__ the add, as needed by `+=`, we have to add the number to the result of atomicFetchAdd.
Hope that answers your question.

denis-sh:
@jadbox, please follow D stadard library code style and squash such small fixing commit lists to not pollute the log.

jadbox:
@denis-sh How should I do that exactly? Just perform a rebase squash and push?

denis-sh:
@jadbox, yes. You need to prepare your local branch (e.g. rebase squash) and if its ready update remote one (push with force). *github* allows non fast-forward updates for pull requests.

jadbox:
@denis-sh It looks like it shouldn't pollute master as it was simply merged in (single commit point). How would compressing the history of the branch that was merge commit'ed over be cleaner?

denis-sh:
@jadbox, "compressed" branch view compound from only merge commits isn't enough as merged branches often contain few logically distinct changes divided into commits so when one looks through the log he inspects all merged branch commits and such sequences of non-sense fixups waste vertical space and doesn't look good. I'm not even telling about [*github*'s commits](https://github.com/D-Programming-Language/druntime/commits/master) page.

MartinNowak:
Sure, the commits don't convey any information, but there is no real need to worry about such small stuff.

MartinNowak:
If you want to do a little more on this @jadbox, you might consider supporting pointer arithmetic as well, i.e. `shared int* ptr; atomicOp!"+="(ptr, 2)`.

MartinNowak:
Rats, this is missing documentation.

MartinNowak:
My mistake, those functions are simply used internally by atomicOp.

PR #1209, opened 2015-04-07
Add cent/ucent to core.demangle.

redstar:


MartinNowak:
Auto-merge toggled on

PR #1212, opened 2015-04-12
fix Issue 14423 - struct destructors not finalized for AA values 

rainers:
This builds an appropriate TypeInfo_Struct for aaA.Entry at runtime whenever a new Impl is created.
I've removed the old unused _aaGetX as it doesn't provide enough type info, and disabled the future yet unused _aaGetZ for the same reason.

### DIFF
--- a/src/object.di
+++ b/src/object.di
@@ -206,7 +206,7 @@ class TypeInfo_Class : TypeInfo
     void*       deallocator;
     OffsetTypeInfo[] m_offTi;
     void*       defaultConstructor;
-    immutable(void)*    m_rtInfo;     // data for precise GC
+    immutable(void)*    m_RTInfo;     // data for precise GC
 
     static const(TypeInfo_Class) find(in char[] classname);
     Object create() const;
@@ -234,11 +234,16 @@ class TypeInfo_Struct : TypeInfo
     enum StructFlags : uint
     {
         hasPointers = 0x1,
+        isDynamicType = 0x2, // built at runtime, needs type info in xdtor
     }
     StructFlags m_flags;
   }
-    void function(void*)                    xdtor;
-    void function(void*)                    xpostblit;
+    union
+    {
+        void function(void*)                xdtor;
+        void function(void*, const TypeInfo_Struct ti) xdtorti;
+    }
+  void function(void*)                    xpostblit;
 
     uint m_align;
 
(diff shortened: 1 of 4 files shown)


rainers:
Filre caused by a bad test in the dmd test suite: https://github.com/D-Programming-Language/dmd/pull/4585

MartinNowak:
thx

MartinNowak:
Auto-merge toggled on

rainers:
I have no idea what's going on in the failing test, but here are some observations:
- it passes on Win32, but also fails if std.regex.tests is built as a single test
- the test is compiled on Win64, but never linked into the executable (unittest8.obj not listed in UNITTEST_OBJS)
- The only AA with struct destructor in key or value I could find is `Trie[CodepointSet]`, with `CodepointSet = InversionList!GcPolicy`, `InversionList` contains a `CowArray!SP`, and `CowArray` has a destructor.

Maybe CowArray does not expect to be called from the GC?

rainers:
A few more observations:
- the test passes if I comment the code in the destructor of `std.uni.CowArray`
- reference counting in a destructor is dangerous, because the GC might call it from a different thread
- the failing unittest is run single-threaded, so that's not the issue here
- the dtor might change garbage collected in the same cycle, but that should only interfere with using debug=MEMSTOMP in the GC.

schveiguy:
I also noticed, the failing test passes in release mode, only fails in debug mode:

```
****** FAIL debug32 std.regex.internal.tests
...
0.464s PASS release32 std.regex.internal.tests
```

rainers:
> I also noticed, the failing test passes in release mode, only fails in debug mode:

That's probably because the failure is caused by an assert that is not compiled into the release build. I guess that's also why Win32 passes.

MartinNowak:
It's a bit ugly because the traceback is missing.
https://github.com/D-Programming-Language/phobos/blob/3d0881749f1c3ae24a2978fb34b3f8485ad5ff79/std/regex/internal/tests.d#L368

rainers:
As far as I have investigated, the CodepointSet memoized in std.regex.internel.ir.wordCharacter() contains a released CowArray after a GC collection due to bad reference counting. 

My current guess is this PR uncovers an unrelated bug with respect to postblits that didn't show up so far because the CowArrays just leaked.

schveiguy:
@rainers that makes a lot of sense.

MartinNowak:
Might also be https://issues.dlang.org/show_bug.cgi?id=14443.

rainers:
I finally figured that there is a missing postblit when adding key values with `this(this)` to an AA.

I also added comments to `aaKeys` and `aaValues` about a possible fix for the missing postblits there, but it is blocked by these being `pure nothrow`.

MartinNowak:
Let's rather not add it to the next point release though, as it might cause trouble for some people.

PR #1213, opened 2015-04-12
Fix FP_ILOGB0 and FP_ILOGBNAN constants for FreeBSD

JohanEngelen:
The constants are wrong for FreeBSD, and we see issues with it in https://github.com/D-Programming-Language/phobos/pull/3161

The FreeBSD choice of constants is also C99 compliant.


rainers:
LGTM. There might be other platforms that chose these values, too. According to my search results, bionic might be also affected.

rainers:
Auto-merge toggled on

PR #1214, opened 2015-04-13
better AA benchmarks

MartinNowak:
- cover different key types (int, obj, ptr)
- fix string to be AA limited, instead of benchmarking text splitting

### DIFF
--- a/benchmark/aabench/int.d
+++ b/benchmark/aabench/int.d
@@ -1,0 +1,21 @@
+/**
+ * Benchmark int hashing.
+ *
+ * Copyright: Copyright Martin Nowak 2011 - 2015.
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:    Martin Nowak
+ */
+import std.random;
+
+void main(string[] args)
+{
+    auto rnd = Xorshift32(33);
+
+    int[int] aa;
+    foreach (_; 0 .. 10)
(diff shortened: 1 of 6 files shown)


benchmark/aabench/int.d
```diff
@@ -0,0 +1,20 @@
...
+void main(string[] args)
+{
+    auto rnd = Xorshift32(33);
+
+    int[int] aa;
+    foreach (_; 0 .. 10)
+        foreach (__; 0 .. 100_000)
+            ++aa[uniform(0, 100_000, rnd)];
+
+    assert(aa.length == 99994);
```
rainers: You should also use `if(aa.length != 99994) assert(0);` here as the benchmrks are usually not build in debug mode. Using `enforce` instead of `assert`avoids the boilerplate.

benchmark/aabench/ptr.d
```diff
@@ -0,0 +1,24 @@
...
+    int[int*] aa;
+    auto keys = new int*[](32768);
+    foreach (ref k; keys)
+        k = new int;
+
+    foreach (_; 0 .. 10)
+        foreach (__; 0 .. 100_000)
+            ++aa[keys[uniform(0, keys.length, rnd)]];
+
+    assert(aa.length == keys.length);
```
rainers: ditto

MartinNowak:
Updated, I also optimized the other benchmarks.

rainers:
LGTM

PR #1069, opened 2014-12-16
Clean up unnecessary version(Posix) and version(OS) and try to more narrowly scope catch-all else blocks

joakim-noah:
As mentioned on the newsgroup, this pull starts cleaning up unnecessarily separated version(OS) blocks, so that porters aren't given unneeded work.  Also, inside modules that are only available to Posix, `version(Posix)` checks are removed and catch-all uses of `else version(Posix)` are expanded to name each OS instead, as Sean said is the way it should be done.

I'm putting this pull up before it's done so that I can get comment on any of the proposed cleanups early on, before I spend time tracking more of these down.

PR #1216, opened 2015-04-16
Collapse all redundant version(OS) blocks

joakim-noah:
This was done in #1069 but proved controversial, so I've spun it off into this PR so as not to block the rest of the changes in that PR.

If reviewers can provide a reason why any particular version block might be needed in the future, say because some future OS we'd like to support varies, I'll add a comment indicating that instead and leave that block in.  Otherwise, I'd rather clean these up and not have this unnecessary boilerplate.

### DIFF
--- a/src/core/sys/posix/arpa/inet.d
+++ b/src/core/sys/posix/arpa/inet.d
@@ -199,23 +199,4 @@ NOTE: The following must must be defined in core.sys.posix.arpa.inet to break
 INET6_ADDRSTRLEN // from core.sys.posix.netinet.in_
 */
 
-version( linux )
-{
-    enum INET6_ADDRSTRLEN = 46;
-}
-else version( OSX )
-{
-    enum INET6_ADDRSTRLEN = 46;
-}
-else version( FreeBSD )
-{
-    enum INET6_ADDRSTRLEN = 46;
-}
-else version( Solaris )
-{
-    enum INET6_ADDRSTRLEN = 46;
-}
-else version( Android )
-{
-    enum INET6_ADDRSTRLEN = 46;
-}
+enum INET6_ADDRSTRLEN = 46;
--- a/src/core/sys/posix/grp.d
+++ b/src/core/sys/posix/grp.d
@@ -38,59 +38,12 @@ group* getgrnam(in char*);
 group* getgrgid(gid_t);
 */
 
-version( linux )
+struct group
 {
-    struct group
-    {
-        char*   gr_name;
-        char*   gr_passwd;
-        gid_t   gr_gid;
-        char**  gr_mem;
-    }
-}
-else version( OSX )
-{
-    struct group
-    {
-        char*   gr_name;
-        char*   gr_passwd;
-        gid_t   gr_gid;
-        char**  gr_mem;
-    }
-}
-else version( FreeBSD )
-{
-    struct group
-    {
-        char*   gr_name;
-        char*   gr_passwd;
-        gid_t   gr_gid;
-        char**  gr_mem;
-    }
-}
-else version( Solaris )
-{
-    struct group
-    {
-        char*   gr_name;
-        char*   gr_passwd;
-        gid_t   gr_gid;
-        char**  gr_mem;
-    }
-}
-else version( Android )
-{
-    struct group
-    {
-        char*   gr_name;
-        char*   gr_passwd;
-        gid_t   gr_gid;
-        char**  gr_mem;
-    }
-}
-else
-{
-    static assert(false, "Unsupported platform");
+    char*   gr_name;
+    char*   gr_passwd;
+    gid_t   gr_gid;
+    char**  gr_mem;
 }
 
 group* getgrnam(in char*);
--- a/src/core/sys/posix/netinet/in_.d
+++ b/src/core/sys/posix/netinet/in_.d
@@ -972,23 +972,4 @@ else version( Android )
 IPPROTO_RAW
 */
 
-version( linux )
-{
-    enum uint IPPROTO_RAW = 255;
-}
-else version( OSX )
-{
-    enum uint IPPROTO_RAW = 255;
-}
-else version( FreeBSD )
-{
-    enum uint IPPROTO_RAW = 255;
-}
-else version( Solaris )
-{
-    enum uint IPPROTO_RAW = 255;
-}
-else version( Android )
-{
-    enum uint IPPROTO_RAW = 255;
-}
+enum uint IPPROTO_RAW = 255;
--- a/src/core/sys/posix/netinet/tcp.d
+++ b/src/core/sys/posix/netinet/tcp.d
@@ -26,23 +26,4 @@ extern (C):
 TCP_NODELAY
 */
 
-version( linux )
-{
-    enum TCP_NODELAY = 1;
-}
-else version( OSX )
-{
-    enum TCP_NODELAY = 1;
-}
-else version( FreeBSD )
-{
-    enum TCP_NODELAY = 1;
-}
-else version( Solaris )
-{
-    enum TCP_NODELAY = 1;
-}
-else version( Android )
-{
-    enum TCP_NODELAY = 1;
-}
+enum TCP_NODELAY = 1;
--- a/src/core/sys/posix/pthread.d
+++ b/src/core/sys/posix/pthread.d
@@ -993,52 +993,9 @@ int pthread_rwlockattr_getpshared(in pthread_rwlockattr_t*, int*);
 int pthread_rwlockattr_setpshared(pthread_rwlockattr_t*, int);
 */
 
-version (linux)
-{
-    int pthread_condattr_getpshared(in pthread_condattr_t*, int*);
-    int pthread_condattr_setpshared(pthread_condattr_t*, int);
-    int pthread_mutexattr_getpshared(in pthread_mutexattr_t*, int*);
-    int pthread_mutexattr_setpshared(pthread_mutexattr_t*, int);
-    int pthread_rwlockattr_getpshared(in pthread_rwlockattr_t*, int*);
-    int pthread_rwlockattr_setpshared(pthread_rwlockattr_t*, int);
-}
-else version( FreeBSD )
-{
-    int pthread_condattr_getpshared(in pthread_condattr_t*, int*);
-    int pthread_condattr_setpshared(pthread_condattr_t*, int);
-    int pthread_mutexattr_getpshared(in pthread_mutexattr_t*, int*);
-    int pthread_mutexattr_setpshared(pthread_mutexattr_t*, int);
-    int pthread_rwlockattr_getpshared(in pthread_rwlockattr_t*, int*);
-    int pthread_rwlockattr_setpshared(pthread_rwlockattr_t*, int);
-}
-else version( OSX )
-{
-    int pthread_condattr_getpshared(in pthread_condattr_t*, int*);
-    int pthread_condattr_setpshared(pthread_condattr_t*, int);
-    int pthread_mutexattr_getpshared(in pthread_mutexattr_t*, int*);
-    int pthread_mutexattr_setpshared(pthread_mutexattr_t*, int);
-    int pthread_rwlockattr_getpshared(in pthread_rwlockattr_t*, int*);
-    int pthread_rwlockattr_setpshared(pthread_rwlockattr_t*, int);
-}
-else version (Solaris)
-{
-    int pthread_condattr_getpshared(in pthread_condattr_t*, int*);
-    int pthread_condattr_setpshared(pthread_condattr_t*, int);
-    int pthread_mutexattr_getpshared(in pthread_mutexattr_t*, int*);
-    int pthread_mutexattr_setpshared(pthread_mutexattr_t*, int);
-    int pthread_rwlockattr_getpshared(in pthread_rwlockattr_t*, int*);
-    int pthread_rwlockattr_setpshared(pthread_rwlockattr_t*, int);
-}
-else version (Android)
-{
-    int pthread_condattr_getpshared(pthread_condattr_t*, int*);
-    int pthread_condattr_setpshared(pthread_condattr_t*, int);
-    int pthread_mutexattr_getpshared(pthread_mutexattr_t*, int*);
-    int pthread_mutexattr_setpshared(pthread_mutexattr_t*, int);
-    int pthread_rwlockattr_getpshared(pthread_rwlockattr_t*, int*);
-    int pthread_rwlockattr_setpshared(pthread_rwlockattr_t*, int);
-}
-else
-{
-    static assert(false, "Unsupported platform");
-}
+int pthread_condattr_getpshared(in pthread_condattr_t*, int*);
+int pthread_condattr_setpshared(pthread_condattr_t*, int);
+int pthread_mutexattr_getpshared(in pthread_mutexattr_t*, int*);
+int pthread_mutexattr_setpshared(pthread_mutexattr_t*, int);
+int pthread_rwlockattr_getpshared(in pthread_rwlockattr_t*, int*);
+int pthread_rwlockattr_setpshared(pthread_rwlockattr_t*, int);
--- a/src/core/sys/posix/sched.d
+++ b/src/core/sys/posix/sched.d
@@ -135,30 +135,7 @@ version( Posix )
 int sched_yield();
 */
 
-version( linux )
-{
-    int sched_yield();
-}
-else version( OSX )
-{
-    int sched_yield();
-}
-else version( FreeBSD )
-{
-    int sched_yield();
-}
-else version (Solaris)
-{
-    int sched_yield();
-}
-else version (Android)
-{
-    int sched_yield();
-}
-else
-{
-    static assert(false, "Unsupported platform");
-}
+int sched_yield();
 
 //
 // Scheduling (TPS)
--- a/src/core/sys/posix/semaphore.d
+++ b/src/core/sys/posix/semaphore.d
@@ -128,27 +128,4 @@ version( Posix )
 int sem_timedwait(sem_t*, in timespec*);
 */
 
-version( linux )
-{
-    int sem_timedwait(sem_t*, in timespec*);
-}
-else version( OSX )
-{
-    int sem_timedwait(sem_t*, in timespec*);
-}
-else version( FreeBSD )
-{
-    int sem_timedwait(sem_t*, in timespec*);
-}
-else version (Solaris)
-{
-    int sem_timedwait(sem_t*, in timespec*);
-}
-else version( Android )
-{
-    int sem_timedwait(sem_t*, in timespec*);
-}
-else
-{
-    static assert(false, "Unsupported platform");
-}
+int sem_timedwait(sem_t*, in timespec*);
--- a/src/core/sys/posix/signal.d
+++ b/src/core/sys/posix/signal.d
@@ -2019,32 +2019,5 @@ int pthread_kill(pthread_t, int);
 int pthread_sigmask(int, in sigset_t*, sigset_t*);
 */
 
-version( linux )
-{
-    int pthread_kill(pthread_t, int);
-    int pthread_sigmask(int, in sigset_t*, sigset_t*);
-}
-else version( OSX )
-{
-    int pthread_kill(pthread_t, int);
-    int pthread_sigmask(int, in sigset_t*, sigset_t*);
-}
-else version( FreeBSD )
-{
-    int pthread_kill(pthread_t, int);
-    int pthread_sigmask(int, in sigset_t*, sigset_t*);
-}
-else version (Solaris)
-{
-    int pthread_kill(pthread_t, int);
-    int pthread_sigmask(int, in sigset_t*, sigset_t*);
-}
-else version( Android )
-{
-    int pthread_kill(pthread_t, int);
-    int pthread_sigmask(int, in sigset_t*, sigset_t*);
-}
-else
-{
-    static assert(false, "Unsupported platform");
-}
+int pthread_kill(pthread_t, int);
+int pthread_sigmask(int, in sigset_t*, sigset_t*);
--- a/src/core/sys/posix/stdlib.d
+++ b/src/core/sys/posix/stdlib.d
@@ -104,41 +104,10 @@ int setenv(in char*, in char*, int);
 int unsetenv(in char*);
 */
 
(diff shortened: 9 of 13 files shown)


DmitryOlshansky:
I guess if you can show that all of this is specified in POSIX standard it should be enough to avoid versioning because it's all in `sys.posix`.

src/core/sys/posix/arpa/inet.d
```diff
@@ -199,23 +199,4 @@ NOTE: The following must must be defined in core.sys.posix.arpa.inet to break
...
-}
-else version( Solaris )
-{
-    enum INET6_ADDRSTRLEN = 46;
-}
-else version( Android )
-{
-    enum INET6_ADDRSTRLEN = 46;
-}
+enum INET6_ADDRSTRLEN = 46;
```
joakim-noah: [Defined to be 46 in the Posix standard](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/netinet_in.h.html).

src/core/sys/posix/grp.d
```diff
@@ -38,59 +38,12 @@ group* getgrnam(in char*);
...
-}
-else
+struct group
 {
-    static assert(false, "Unsupported platform");
+    char*   gr_name;
+    char*   gr_passwd;
+    gid_t   gr_gid;
+    char**  gr_mem;
 }
```
joakim-noah: [All but gr_passwd defined in POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/grp.h.html).

src/core/sys/posix/netinet/in_.d
```diff
@@ -972,23 +972,4 @@ else version( Android )
...
-}
-else version( Solaris )
-{
-    enum uint IPPROTO_RAW = 255;
-}
-else version( Android )
-{
-    enum uint IPPROTO_RAW = 255;
-}
+enum uint IPPROTO_RAW = 255;
```
joakim-noah: [Must be defined in POSIX standard, but actual value is not specified](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/netinet_in.h.html).

src/core/sys/posix/netinet/tcp.d
```diff
@@ -26,23 +26,4 @@ extern (C):
...
-}
-else version( Solaris )
-{
-    enum TCP_NODELAY = 1;
-}
-else version( Android )
-{
-    enum TCP_NODELAY = 1;
-}
+enum TCP_NODELAY = 1;
```
joakim-noah: [Must be defined in POSIX standard, but value is not specified](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/netinet_tcp.h.html).

src/core/sys/posix/pthread.d
```diff
@@ -993,52 +993,9 @@ int pthread_rwlockattr_getpshared(in pthread_rwlockattr_t*, int*);
...
-else
-{
-    static assert(false, "Unsupported platform");
-}
+int pthread_condattr_getpshared(in pthread_condattr_t*, int*);
+int pthread_condattr_setpshared(pthread_condattr_t*, int);
+int pthread_mutexattr_getpshared(in pthread_mutexattr_t*, int*);
+int pthread_mutexattr_setpshared(pthread_mutexattr_t*, int);
+int pthread_rwlockattr_getpshared(in pthread_rwlockattr_t*, int*);
+int pthread_rwlockattr_setpshared(pthread_rwlockattr_t*, int);
```
joakim-noah: [All](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_condattr_getpshared.html) [three](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_mutexattr_getpshared.html) [pairs](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_rwlockattr_getpshared.html) of functions are defined this way in the POSIX standard.

src/core/sys/posix/sched.d
```diff
@@ -135,30 +135,7 @@ version( Posix )
...
-}
-else version (Android)
-{
-    int sched_yield();
-}
-else
-{
-    static assert(false, "Unsupported platform");
-}
+int sched_yield();
```
joakim-noah: [Defined this way in POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/functions/sched_yield.html).

src/core/sys/posix/semaphore.d
```diff
@@ -128,27 +128,4 @@ version( Posix )
...
-}
-else version( Android )
-{
-    int sem_timedwait(sem_t*, in timespec*);
-}
-else
-{
-    static assert(false, "Unsupported platform");
-}
+int sem_timedwait(sem_t*, in timespec*);
```
joakim-noah: [Defined this way in POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/functions/sem_timedwait.html).

src/core/sys/posix/signal.d
```diff
@@ -2019,32 +2019,5 @@ int pthread_kill(pthread_t, int);
...
-{
-    int pthread_kill(pthread_t, int);
-    int pthread_sigmask(int, in sigset_t*, sigset_t*);
-}
-else
-{
-    static assert(false, "Unsupported platform");
-}
+int pthread_kill(pthread_t, int);
+int pthread_sigmask(int, in sigset_t*, sigset_t*);
```
joakim-noah: [Both](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_kill.html) [are](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_sigmask.html) defined this way in POSIX standard.

src/core/sys/posix/stdlib.d
```diff
@@ -104,41 +104,10 @@ int setenv(in char*, in char*, int);
...
-else version( Solaris )
-{
-    int setenv(in char*, in char*, int);
-    int unsetenv(in char*);
+int setenv(in char*, in char*, int);
+int unsetenv(in char*);
 
-    void* valloc(size_t); // LEGACY non-standard
-}
+void* valloc(size_t); // LEGACY non-standard
```
joakim-noah: [setenv](http://pubs.opengroup.org/onlinepubs/9699919799/functions/setenv.html) and [unsetenv](http://pubs.opengroup.org/onlinepubs/9699919799/functions/unsetenv.html) are defined this way in the POSIX standard, while valloc is considered legacy, as the comment notes.

src/core/sys/posix/sys/mman.d
```diff
@@ -98,45 +98,10 @@ PROT_EXEC
...
-    enum PROT_EXEC = 0x04;
-}
-else
-{
-    static assert(false, "Unsupported platform");
-}
+enum PROT_NONE  = 0x0;
+enum PROT_READ  = 0x1;
+enum PROT_WRITE = 0x2;
+enum PROT_EXEC  = 0x4;
```
joakim-noah: [Must be defined in the POSIX standard, but values are not specified](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_mman.h.html).

src/core/sys/posix/sys/mman.d
```diff
@@ -455,35 +420,8 @@ int mlock(in void*, size_t);
...
-{
-    int mlock(in void*, size_t);
-    int munlock(in void*, size_t);
-}
-else
-{
-    static assert(false, "Unsupported platform");
-}
+int mlock(in void*, size_t);
+int munlock(in void*, size_t);
```
joakim-noah: [Defined this way in the POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/functions/mlock.html).

src/core/sys/posix/time.d
```diff
@@ -326,45 +326,10 @@ tm*   gmtime_r(in time_t*, tm*);
...
-    tm* localtime_r(in time_t*, tm*);
-}
-else
-{
-    static assert(false, "Unsupported platform");
-}
+char* asctime_r(in tm*, char*);
+char* ctime_r(in time_t*, char*);
+tm*   gmtime_r(in time_t*, tm*);
+tm*   localtime_r(in time_t*, tm*);
```
joakim-noah: [All](http://pubs.opengroup.org/onlinepubs/9699919799/functions/asctime.html) [are](http://pubs.opengroup.org/onlinepubs/9699919799/functions/ctime.html) [defined](http://pubs.opengroup.org/onlinepubs/9699919799/functions/gmtime.html) [this way](http://pubs.opengroup.org/onlinepubs/9699919799/functions/localtime.html) in the POSIX standard.

src/core/sys/posix/unistd.d
```diff
@@ -1086,26 +1086,7 @@ else version( Solaris )
...
-}
-else version( Android )
-{
-    int fsync(int) @trusted;
-}
-else version( Solaris )
-{
-    int fsync(int) @trusted;
-}
+int fsync(int) @trusted;
```
joakim-noah: [Defined this way in the POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fsync.html).

src/core/sys/posix/utime.d
```diff
@@ -35,53 +35,10 @@ struct utimbuf
...
-    struct utimbuf
-    {
-        time_t  actime;
-        time_t  modtime;
-    }
-
-    int utime(in char*, in utimbuf*);
+    time_t  actime;
+    time_t  modtime;
 }
```
joakim-noah: [Defined this way in the POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/utime.h.html).

src/core/sys/posix/utime.d
```diff
@@ -35,53 +35,10 @@ struct utimbuf
...
-{
-    struct utimbuf
-    {
-        time_t  actime;
-        time_t  modtime;
-    }
 
-    int utime(in char*, in utimbuf*);
-}
+int utime(in char*, in utimbuf*);
```
joakim-noah: [Defined this way in the POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/functions/utime.html).

DmitryOlshansky:
@joakim-noah thanks, for going through this trouble. Now I think just left out these that are not 100% pinned by the standard (such as TCP_NODELAY _value_) and this should be good to go.

joakim-noah:
Which did you have in mind?  Why keep those if they do not vary on any of the currently supported platforms?

DmitryOlshansky:
> Which did you have in mind? Why keep those if they do not vary on any of the currently supported platforms?

They might once somebody comes to implement a new one. If it's not defined the person porting D to a new OS would easily recognize potential place to add his definition. Highly unlikely that it would be a different value but just staying on the safe side here.

WalterBright:
The separate version blocks is deliberately done this way. This is because:

1. Having a mix of version'd and common blocks makes for tangled code that is harder to figure out what is what.
2. When it comes time to port to a new system, the `else static assert(0);` at the end of the version blocks signals to the porter that he has to find out what the correct declaration for the new OS is. With common blocks, the porter may simply accept the default - and this can cause VERY OBSCURE and weird bugs. Yes, this has happened to us.
3. There is no reason to conserve file space for these. Declarations that happen by chance to be the same are NOT boilerplate and are NOT dirty.
4. By having a version block for each OS, it gives some comfort that the porter actually did check the declarations rather than copy/pasta them, or worse, simply assume that the default is common across all systems.

I strongly oppose this change.

yebblies:
> By having a version block for each OS, it gives some comfort that the porter actually did check the declarations rather than copy/pasta them, or worse, simply assume that the default is common across all systems.

Except that if the porter sees 5 identical version blocks, they're just going to copy paste them.

Wouldn't it be nice if you could just do this!
```d
version( linux || OSX || FreeBSD || Solaris || Android )
{
    enum INET6_ADDRSTRLEN = 46;
}
else
{
    static assert(0);
}
```

DmitryOlshansky:
@ WalterBright  Except that 90% of this is FIXED for all POSIX by the STANDARD that is not going to change, so this is just copy-paste junk we accumulated.

WalterBright:
@DmitryOlshansky that's all fine except when a particular posix .h file is slightly different, and nobody checked. This happened to druntime for pthreads.h.

WalterBright:
> Wouldn't it be nice if you could just do this!

Yeah, and then someone added a declaration to it for Linux that turned out to be different on OSX, but of course he didn't have an OSX machine and didn't check.

I'm fed up with bugs like that in druntime. All those convenient merged version blocks turned out to cost a lot of extra time in the long run.

WalterBright:
I'd like to remind everyone that there is currently no tool to check .d header files against their .h counterparts. None. Not at all. The only way to do it is to manually check. This checking must be done for each platform, line by line. It's much harder to do when some declarations are merged and some are not. I speak from experience.

No declarations should go into druntime without being checked. There's been a lot of trouble in the past with merged and unchecked declarations.

joakim-noah:
>Having a mix of version'd and common blocks makes for tangled code that is harder to figure out what is what.

This is the status quo, I see no way around it.

>When it comes time to port to a new system, the else static assert(0); at the end of the version blocks signals to the porter that he has to find out what the correct declaration for the new OS is. With common blocks, the porter may simply accept the default - and this can cause VERY OBSCURE and weird bugs. Yes, this has happened to us.

6 of the 14 blocks collapsed here have no `static assert`, so very likely they'd have simply been missed.  I know that common blocks are not going to be checked, as I didn't check them for my Android port, only after something went amiss.  I don't think that's as bad as you claim.

>There is no reason to conserve file space for these. Declarations that happen by chance to be the same are NOT boilerplate and are NOT dirty.

It's not a matter of file space, it's a matter of a waste of time for porters, at least for the blocks with a `static assert`, the rest will probably just be ignored.  These declarations are not the same "by chance," they are defined this way by the POSIX standard.  The "chance" would be if some OS was different, in which case they can put their definitions in their own OS directory, as linux and OS X do for some definitions.

>By having a version block for each OS, it gives some comfort that the porter actually did check the declarations rather than copy/pasta them, or worse, simply assume that the default is common across all systems.

Sure, but where do you draw the line?  Why are these definitions randomly chosen to separate, just in case they may vary on some future unspecified platform, and not the hundreds of remaining common definitions in druntime?

For example, [I noted earlier that `signal` is defined differently on Android](https://github.com/D-Programming-Language/druntime/pull/1069#discussion_r27002034), so I'll need to modify the common definition in `core.stdc.signal` to get it to work on Android.  This wasn't some hair-pulling experience, I simply tried to use `signal` in some D code and the linker told me the symbol wasn't available.  I realize it may not always be so easy, but that is no reason to arbitrarily separate some declarations and not others.

>I'd like to remind everyone that there is currently no tool to check .d header files against their .h counterparts. None. Not at all. The only way to do it is to manually check. This checking must be done for each platform, line by line. It's much harder to do when some declarations are merged and some are not. I speak from experience.

This will not be true someday soon, I'm surprised it isn't already.  I used Dstep to check a lot of these, it can certainly be automated for function declarations.  The missing piece is that libclang, which Dstep is built on, doesn't provide a C API for their preprocessor, which will hopefully be remedied soon.

>No declarations should go into druntime without being checked. There's been a lot of trouble in the past with merged and unchecked declarations.

That is a lofty goal, but druntime has so many declarations that are currently not separated by OS.  How do you pick and choose which go in common blocks and which should be speculatively separated, just in case they vary on some future unspecified platform?  That is the key question here, as none of these blocks vary on any of the currently supported platforms.

WalterBright:
> as I didn't check them for my Android port

This is exactly the problem. If proper version declarations are used, the worst is the Android port is borked. With them all merged, then sloppy ports put the other platforms at risk. Nobody wants to go and recheck all the OSX declarations, for example.

> I realize it may not always be so easy, but that is no reason to arbitrarily separate some declarations and not others.

These can be VERY HARD to track down bugs to. C declarations, for example, do not have types encoded in the name. So having a type that is off can lead to erratic behavior. Having a constant with a different value can behave weirdly. Having a struct with a different layout is awful. We have NO WAY to mechanically check the accuracy of this stuff. It is boring, tedious, manual labor. The version scheme helps compartmentalize things so if you're working on OSX declarations you're not inadvertently mucking up other platforms.

> but druntime has so many declarations that are currently not separated by OS.

That's because it gets worked on by dozens of people and is in a constant state of flux; I don't see everything before it goes in. There are often attempts to undo the versioning, I only happened to catch this one by chance. This debate happens regularly, because the version scheme used is unusual, and peoples' first reaction is it is something that should be "cleaned up".

joakim-noah:
>The version scheme helps compartmentalize things so if you're working on OSX declarations you're not inadvertently mucking up other platforms.

Nobody is suggesting using common declarations where platforms like OS X differ.  The question is why are we arbitrarily separating some declarations that do not differ on any of the supported platforms and are defined by the POSIX standard to be a certain way, which is why they are currently the same across OSs and not random "chance?"

>There are often attempts to undo the versioning, I only happened to catch this one by chance.

I like the version scheme, that is not the problem.  The problem is when to use it.  You have not answered the key question above:

"druntime has so many declarations that are currently not separated by OS. How do you pick and choose which go in common blocks and which should be speculatively separated, just in case they vary on some future unspecified platform?"

Insisting that these few stay separated while hundreds of others aren't seems arbitrary, given that none of these vary on any of the currently supported POSIX platforms.

WalterBright:
I'll repeat - operating system APIs should all be separated into version blocks. The fact that this is not consistent across druntime is because lots of people have contributed to it and not followed this pattern.

DmitryOlshansky:
> I'll repeat - operating system APIs

Then start with deleting sys.posix. 
Seriously POSIX is a standard for a reason and this file is dedicated to it. No point in denying hard facts about that standard pretending that they can change.

schveiguy:
I agree with @WalterBright on most of these things. The POSIX standard does not specify layouts or implementations. And C macros can make this much more difficult than you think. pthread.h is a good example of a Posix standard API that varies wildly on implementation.

However, there are definitely some that should be common.

2 examples:

`INET6_ADDRSTRLEN` should be common, it's a constant. No chance for odd definitions.
`struct group` should NOT be common.

Even though `struct group` is defined by the posix standard, it says:
```
The <grp.h> header shall declare the group structure, which shall include the following members:
```
*shall* include, not *shall only* include. The presence here of `gr_passwd` as identified by @joakim-noah is evidence that we cannot be certain of exact layouts here.

I'd say anywhere you can point to the standard defining a specific layout or specific value some constant should be, we can simply define that struct/constant as expected. Anywhere it's left underspecified, we should have versions, even if most of them are identical.

I'll note that I strongly also agree with @yebblies about using version boolean or-ing. Copy-pasted code is copy-pasted code, whether it's done by the user or compiler. Making the user use his CTRL-C and CTRL-V keys isn't going to stop the same issue from happening. If I'm trying to get a port up and running, I pick the most similar version, copy paste it, see if it breaks, and if not, I'm done. I realize this isn't the *right* way to do it, but there's no possible way to force the *right* way on developers. I speak from experience when I did the array append update, but didn't have a Mac to test with.

joakim-noah:
>I'll repeat - operating system APIs should all be separated into version blocks. The fact that this is not consistent across druntime is because lots of people have contributed to it and not followed this pattern.

@WalterBright, how committed are you to this?  If I instead submit a PR that separates all common blocks in `core.sys.posix` by OS and adds static asserts everywhere, would you merge it?  For example, if I separated out [these 8 function declarations](https://github.com/D-Programming-Language/druntime/blob/master/src/core/sys/posix/pthread.d#L264) and [this block of 38 function declarations](https://github.com/D-Programming-Language/druntime/blob/master/src/core/sys/posix/pthread.d#L429) from `core.sys.posix.pthread` by OS, so that they were copy-pasted many more times, would that be a change worth merging?

Of course, I wouldn't actually check all those declarations on every OS, which is your real goal from all this separation by OS, but I'd add a comment saying that certain OS's weren't checked, while I could check a couple, like `linux` and `FreeBSD`.  This could certainly be an improvement for those two OS's, while it wouldn't be any worse than the status quo for the OS's that aren't checked, plus the comment might help get them checked by someone eventually.

WalterBright:
> I'll note that I strongly also agree with @yebblies about using version boolean or-ing. Copy-pasted code is copy-pasted code, whether it's done by the user or compiler. Making the user use his CTRL-C and CTRL-V keys isn't going to stop the same issue from happening. If I'm trying to get a port up and running, I pick the most similar version, copy paste it, see if it breaks, and if not, I'm done. I realize this isn't the right way to do it, but there's no possible way to force the right way on developers. I speak from experience when I did the array append update, but didn't have a Mac to test with.

I'll repeat.

1. By doing copy-pasta without checking, the porter is specifically taking action he knows is wrong. This is quite different from having a default version that is overlooked.

2. The copy-pasta damage will be limited to the ported to platform. It will not inadvertently and unwittingly break the other platforms.

3. Another one: Frankly, people tend to get confused by blocks of Boolean versions, especially when they get nested. Someone working on ABC operating system is totally focused on ABC, and having DEF stuff mixed in is a distraction and just asking for typos and inadvertent modifications creeping in.

I'll pull the experience card here. I have been doing ports, and working with ports, for 35 years. I have a LOT of experience with what goes wrong. Do these things I allege will happen actually happen? Hell yes. It's common and an ongoing nuisance. It has happened repeatedly in druntime, too, by well meaning and competent professionals who do not believe me. Declarations separated out by OS version blocks have fared a LOT better in the stability department.

I could let this go and allow people to learn the hard way again and again. But I don't think we have the resources to waste on that.

DmitryOlshansky:
> The copy-pasta damage will be limited to the ported to platform. It will not inadvertently and unwittingly break the other platforms.

I can buy that. And when you're speaking from experience it's I can trust. 

In any case I wouldn't go against your position. I hate it, but you have good reasons to hold it.

I guess with that we are closing it, @joakim-noah my appologies for the wasted work.

WalterBright:
thank you, @DmitryOlshansky

joakim-noah:
@WalterBright, [I think you missed my question above](https://github.com/D-Programming-Language/druntime/pull/1216#issuecomment-121997818).

schuetzm:
Maybe a compromise could be helpful? While every platform could potentially have its own definitions, most don't (at least for the function signatures; structs may be different). So, could there be a (package-private) module with common POSIX definitions, the relevant parts of which would however have to be publicly imported by the OS specific blocks? This should reduce C&P, while still requiring an explicit import statement for each platform, reducing the risk of mistakes...

PR #1218, opened 2015-04-17
remove doc examples from bitop

burner:
and replace with unittest examples


MartinNowak:
Auto-merge toggled on

burner:
thanks

PR #1219, opened 2015-04-17
debug auto-tester failure

MartinNowak:


yebblies:
Could it be some kind of race condition?

MartinNowak:
Not sure, seems to work now. There is a similar test issue for phobos https://issues.dlang.org/show_bug.cgi?id=14241 where the message actually comes from the test itself.

braddr:
For what it's worth, win-farm-1 is somewhat different from the other two windows boxes in the tester fleet:
    win-farm-1: windows 7, builds both win_32 and win_64
    ip-0A0001C7: windows serve 2008, builds just win_32
    WIN-F30QP0PU8L1: windows server 2012, builds just win_64
Each runs different versions of visual studio, though I'd have to go login to each to look at what exact versions they are.

The important part, hosever, is that the master branch is working and the builder doesn't do anything any differently for different branches, so it's gotta be something different about the code under test.  I don't know what, but something is.

When I get some time to rearrange hardware, the intent is for the two old windows boxes to go away and be replaced by at least one more win-farm-# machine running windows 7 and compiling both win_32 and win_64 just like win_farm_1 is doing.

PR #1220, opened 2015-04-17
TickDuration to!() unittest

burner:
the to! template didn't seam to have a unittest

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -1710,6 +1710,51 @@ T to(string units, T, D)(D td) @safe pure nothrow @nogc
         static assert(0, "Incorrect template constraint.");
 }
 
+///
+unittest
+{
+    auto t = TickDuration.from!"seconds"(1000);
+
+    long tl = to!("seconds",long)(t);
+    assert(tl == 1000);
+
+    double td = to!("seconds",double)(t);
+    assert(_abs(td - 1000) < 0.001);
+}
+
+unittest
+{
+    auto t1v = 1000;
+    auto t2v = 333;
+    auto t1 = TickDuration.from!"seconds"(t1v);
+    auto t2 = TickDuration.from!"seconds"(t2v);
+
+    foreach(F; _TypeTuple!(int,uint,long,ulong,float,double,real))
+    {
+        F t1f = to!("seconds",F)(t1);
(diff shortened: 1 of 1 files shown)


src/core/time.d
```diff
@@ -1710,6 +1710,43 @@ T to(string units, T, D)(D td) @safe pure nothrow @nogc
         static assert(0, "Incorrect template constraint.");
 }
 
+///
+unittest
+{
+    auto t = TickDuration.from!"seconds"(1000);
+    long tl = to!("seconds",long)(t);
+    double td = to!("seconds",double)(t);
```
MartinNowak: Add assertions here?

src/core/time.d
```diff
@@ -1710,6 +1710,43 @@ T to(string units, T, D)(D td) @safe pure nothrow @nogc
...
+    auto t1v = 1000;
+    auto t2v = 333;
+    auto t1 = TickDuration.from!"seconds"(t1v);
+    auto t2 = TickDuration.from!"seconds"(t2v);
+
+    foreach(F; _TypeTuple!(int,uint,long,ulong,float,double,real))
+    {
+        F t1f = to!("seconds",F)(t1);
+        F t2f = to!("seconds",F)(t2);
+        auto t12 = t1 / t2v;
```
MartinNowak: How about testing subtraction instead of division, as it doesn't make too much sense to divide 2 `TickDurations`, because a `TickDuration` is rather a point in time then a duration, which led us to the MonoTime redesign.

MartinNowak:
Auto-merge toggled on

src/core/time.d
```diff
@@ -4242,6 +4287,11 @@ long _abs(long val) @safe pure nothrow @nogc
     return val >= 0 ? val : -val;
 }
 
+double _abs(double val) @safe pure nothrow @nogc
```
MartinNowak: Those `_abs` functions should really be `private`.

burner: I will fix that tomorrow

MartinNowak: thx

burner: https://github.com/D-Programming-Language/druntime/pull/1224

PR #1221, opened 2015-04-17
Mark core.stdc.stdlib malloc functions as pure

WalterBright:
Since memory allocation via the GC is pure, it makes sense to make the C ones pure as well. Otherwise, it is very difficult to replace use of the GC with malloc/free and keep things pure.


JakobOvrum:
See #1183

MartinNowak:
Please add any arguments to #1183.

PR #1222, opened 2015-04-18
Copy object.di from object_.d

rainers:
Updating object.di and object_.d is a maintenance liability. This PR removes non-public code from object._d and adds a bit from object.di in order to generate the latter automatically.

Comparing object.di to the generated header reveals:
- aliases for string,wstring,dstring not converted correctly, needs https://github.com/D-Programming-Language/dmd/pull/4594
- overriding methods often not declared in classes declared in object.di. This can cause unexpected results when deriving from this class and calling `super.method()`.
- different types for `TpeInfo_Class.defaultConstructor`
- different return types for `TypeInfo_Struct.xtoHash`
- missing fields in classes `MemberInfo_field` and `MemberInfo_function`, can have bad effects when deriving from it or using `__traits(classInstanceSize)`.
- different next/base field declarations for `TypeInfo_Const`. See also https://github.com/D-Programming-Language/druntime/pull/1182
- header generation moves function attributes before the return type.

Unfortunately, neither the generated object_.di nor the original object_.d can be used as a replacement for object.di yet:
- object_.d does not work because it has (now local) imports to the rt package which are not supposed to be visible in general.
- object_.di does not contain all the source code necessary for CTFE. Adding `auto` can work in some places, but not for the `Exception` and `Error` constructors.
- in addition, the missing source in object_.di can disable some inlining opportunities, especially for the AA templates.

So we need better control over the omission of the code in the generated headers. Adding inline hints might be a possible solution, though somewhat abusive in the case of CTFE.

### DIFF
--- a/src/object_.d
+++ b/src/object_.d
@@ -30,8 +30,6 @@ private
     debug(PRINTF) import core.stdc.stdio;
 
     extern (C) Object _d_newclass(const TypeInfo_Class ci);
-    extern (C) void _d_arrayshrinkfit(const TypeInfo ti, void[] arr) nothrow;
-    extern (C) size_t _d_arraysetcapacity(const TypeInfo ti, size_t newcapacity, void *arrptr) pure nothrow;
     extern (C) void rt_finalize(void *data, bool det=true);
 }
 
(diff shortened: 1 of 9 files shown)


MartinNowak:
I also wanted to look at that, though I thought about ditching `object.di` in favor of always using `object.d`. Any reason against that?

src/object_.d
```diff
@@ -1317,6 +1325,201 @@ class MemberInfo_function : MemberInfo
...
+
+            auto p = cast(immutable char*)addrOf(MIname);
+            return p[0 .. strlen(p)];
+        }
+        // return null;
+    }
+
+    static int opApply(scope int delegate(ModuleInfo*) dg)
+    {
+        import rt.minfo;
```
MartinNowak: How to avoid inlining this?

rainers:
> I also wanted to look at that, though I thought about ditching object.di in favor of always using object.d. Any reason against that?

The problem is that the import references into the `rt` package are still analyzed, so you get errors in a non-druntime build. We would have to wrap these into extern(C) function or use `externDFunc`.

rainers:
BTW: I think that `ModuleInfo` should not be in object.d, it's pretty internal to the runtime. The compiler expects it to be defined, though, but never uses the definition.

MartinNowak:
> The problem is that the import references into the rt package are still analyzed, so you get errors in a non-druntime build. We would have to wrap these into extern(C) function or use externDFunc.

Then let's do that, the main reasons to import `rt` modules instead of using `extern (C)`, are also achieved using `externDFunc`.

MartinNowak:
> BTW: I think that ModuleInfo should not be in object.d, it's pretty internal to the runtime.

You need to have the interface though, we could split it into an internal and public type though.

rainers:
> Then let's do that, the main reasons to import rt modules instead of using extern (C), are also achieved using externDFunc.

Ok, will look into it. This will generate a dependency on core.internal.traits and core.demangle, though. With local imports that should be restricted to compiling with -inline.

rainers:
> You need to have the interface though, we could split it into an internal and public type though.

I was rather thinking of moving it to another core module to import it when you actually need it.

MartinNowak:
As long as `foreach (m; ModuleInfo)` keeps on working.

rainers:
> Ok, will look into it.

Done. I had to modify rt.util.string a bit to not call templates through externDFunc.

rainers:
> As long as foreach (m; ModuleInfo) keeps on working.

That would need a new import then. That change should be in another PR anyway.

rainers:
See https://github.com/D-Programming-Language/dmd/pull/4597 for the testtypeid.d failure. I've no idea how to deal with the changes to the error outputs in the fail_compilation tests. Should I just create a failing PR to be merged at the same time?

rainers:
FYI: Even though the size of object.di triples when being copied, I don't see a difference in compilation speed (slowness) when compiling a simple hello world like program using writeln. Both take slightly below 0.6 sec.

MartinNowak:
> I don't see a difference in compilation speed (slowness) when compiling a simple hello world

I'll double check that. Your baseline shouldn't be hello world, which imports std.stdio, but an empty d file.

MartinNowak:
> ../../druntime/import/object.di(1666): Error: uncaught CTFE exception object.Exception("abc"c)

The file and line number doesn't make sense, seems like it shows the location of the Exception constructor.

MartinNowak:
Have you tried to directly use object.d?

rainers:
> Your baseline shouldn't be hello world, which imports std.stdio, but an empty d file.

I thought there would be a little difference that I could weigh to a standard use case. But there was none.

> The file and line number doesn't make sense, seems like it shows the location of the Exception constructor.

There is a difference between object.di and object._d: object.di hides the implementation of the constructor of Throwable, so there is one entry less in the stack trace. I have no idea how this interacts with CTFE, though.

> Have you tried to directly use object.d?

I don't think there is a difference whether a .di or a .d file is used.

MartinNowak:
> I don't think there is a difference whether a .di or a .d file is used.

Exactly, so we can just copy the .d file, and spare the header-gen troubles.

> There is a difference between object.di and object._d: object.di hides the implementation of the constructor of Throwable, so there is one entry less in the stack trace. I have no idea how this interacts with CTFE, though.

Still doesn't make sense, in `throw new Expr(arguments)` the stacktrace should start with `throw` not the constructor of the exception. But it seems like the CTFE code was programmed like that.

MartinNowak:
[Issue 14465 – CTFE exception stacktrace shows location of Exception constructor](https://issues.dlang.org/show_bug.cgi?id=14465)
Anyhow, in the mean-time you might want to open a dmd pull that changes the expected test output.

rainers:
> Issue 14465 – CTFE exception stacktrace shows location of Exception constructor
https://github.com/D-Programming-Language/dmd/pull/4599

> Anyhow, in the mean-time you might want to open a dmd pull that changes the expected test output.
https://github.com/D-Programming-Language/dmd/pull/4600

rainers:
> Exactly, so we can just copy the .d file, and spare the header-gen troubles.

That's what the makefiles do in this PR now, although it copies to object.di to avoid troubles with stale files in peoples import folder.

MartinNowak:
> That's what the makefiles do in this PR now, although it copies to object.di to avoid troubles with stale files in peoples import folder.

There is a semantic connotation in .di headers, if it's the plain source, we should name it .d.
You can temporarily add a `rm -f import/object.di` to the makefiles.

MartinNowak:
Compile times for an empty .d file increase from 3.5ms to 6ms, compiling (not linking) a printf hello world goes from 7ms to 10ms. The increase comes from lexing overhead and is neglectable.
Though I do wonder what we could do to optimize the header, e.g. using -Hf or stripping out unittests.
There seems to be room for a libdparse based header gen.

redstar:
I have not yet tried this PR with ldc but I really like it. The differences between `object_.d` and `object.di` caused some trouble.

rainers:
> Compile times for an empty .d file increase from 3.5ms to 6ms, compiling (not linking) a printf hello world goes from 7ms to 10ms. The increase comes from lexing overhead and is neglectable.

I guess my benchmark batch uses the standard windows timer and has a resolution of 10ms only, so I did not notice the difference.

> Though I do wonder what we could do to optimize the header, e.g. using -Hf or stripping out unittests.
> There seems to be room for a libdparse based header gen.

Maybe some customization options of header generation to adopt to more use cases:
- keep all implementations (but still remove unittests)
- keep all implementations but those of private functions.
- expand version/debug conditionals

rainers:
Failures caused by compiler generated identifiers appearing in error messages. See https://github.com/D-Programming-Language/dmd/pull/4624 for making the tests agnostic to these numbers.

MartinNowak:
Auto-merge toggled on

MartinNowak:
Thanks for this @rainers, it was on my todo list for too long already.

andralex:
oh this is so awesome - I hated all that duplication

andralex:
Wait, posix.mak looks changed a lot without actual visible change - is this a line termination issue? cc @rainers @MartinNowak

MartinNowak:
Yikes! Forgot about that, @rainers can you check whether you converted to CRLF or so.

rainers:
> Yikes! Forgot about that, @rainers can you check whether you converted to CRLF or so.

Arrgh, this creeps in somehow from time to time. I suspect it's caused by a bad .gitattributes in one of my branches, but I don't know how that could infect posix.mak.

https://github.com/D-Programming-Language/druntime/pull/1258 reverts the line endings to LF only. I guess there is no way to avoid polluting the history of posix.mak, isn't there?

rainers:
I guess we should remove object.di from the source folder now? There is also https://github.com/D-Programming-Language/phobos/pull/3263 with a necessary change in phobos. I'm not sure why this isn't caught in any unittests.

MartinNowak:
> I guess we should remove object.di from the source folder now?

Yep we should, I'd still like to copy object.d instead of renaming it to .di.
For a certain period we might add an `rm -f import/object.di` to avoid using a stall header file instead of `import/object.d`.

andralex:
yah remove thx

WalterBright:
> Arrgh, this creeps in somehow from time to time.

Use the tool:
https://github.com/D-Programming-Language/tools/blob/master/tolf.d
 as part of your checkin script, and no more problems.

rainers:
> Yep we should, I'd still like to copy object.d instead of renaming it to .di.
> For a certain period we might add an rm -f import/object.di to avoid using a stall header file instead of import/object.d.

https://github.com/D-Programming-Language/druntime/pull/1260

PR #1223, opened 2015-04-18
Mark exception class constructors as @nogc

w0rp:
The exception class constructors do not themselves allocate any memory in the garbage collector, so they can be marked as @nogc. This enables the exception constructors to be used with custom allocators in the future in @nogc code.

I admit that this change was so trivial I just did it directly through GitHub.

### DIFF
--- a/src/object.di
+++ b/src/object.di
@@ -357,8 +357,8 @@ class Throwable : Object
     TraceInfo   info;
     Throwable   next;
 
-    @safe pure nothrow this(string msg, Throwable next = null);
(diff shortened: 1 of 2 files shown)


w0rp:
Ah, it looks like this fails on some tests with undefined references for strange reasons. I have no idea why, I'll leave that to others with more druntime experience to point out why.

rainers:
You'll have to add the change to object.di aswell.

w0rp:
@rainers Thanks! I was just talking to someone in IRC and he noticed the same thing. I've updated that now.

MartinNowak:
Auto-merge toggled on

PR #1224, opened 2015-04-19
_abs should be private

burner:


MartinNowak:
Auto-merge toggled on

burner:
thanks

jmdavis:
Well, this doesn't hurt, I guess, other than to increase confusion, but it already _was_ private. There's a `private:` label on line 3985, and everything at the bottom of the file is private.

PR #1225, opened 2015-04-19
fix Issue 14464 - coverage merge doesn't work

MartinNowak:
- fix parsing of existing .lst file
- use append semantics (create or merge with existing .lst file)
- enable merge by default
- some code cleanup (reuse buffers, log10, foreach, min/max)

[Issue 14464 – coverage merge doesn't work](https://issues.dlang.org/show_bug.cgi?id=14464)

### DIFF
--- a/src/core/runtime.d
+++ b/src/core/runtime.d
@@ -314,11 +314,36 @@ struct Runtime
 
 
 private:
+
     // NOTE: This field will only ever be set in a static ctor and should
     //       never occur within any but the main thread, so it is safe to
     //       make it __gshared.
     __gshared ModuleUnitTester sm_moduleUnitTester = null;
 }
+
+/**
+ * Set source file path for coverage reports.
+ *
+ * Params:
+ *  path = The new path name.
+ */
+extern (C) void dmd_coverSourcePath(string path);
+
+/**
+ * Set output path for coverage reports.
+ *
+ * Params:
+ *  path = The new path name.
+ */
+extern (C) void dmd_coverDestPath(string path);
+
+/**
+ * Enable merging of coverage reports with existing data.
+ *
+ * Params:
+ *  merge = enable/disable coverage merge mode
+ */
+extern (C) void dmd_coverSetMerge(bool flag);
 
 ///////////////////////////////////////////////////////////////////////////////
 // Overridable Callbacks
--- a/posix.mak
+++ b/posix.mak
@@ -193,7 +193,7 @@ $(DRUNTIME): $(OBJS) $(SRCS)
 UT_MODULES:=$(patsubst src/%.d,$(OBJDIR)/%,$(SRCS))
 HAS_ADDITIONAL_TESTS:=$(shell test -d test && echo 1)
 ifeq ($(HAS_ADDITIONAL_TESTS),1)
-	ADDITIONAL_TESTS:=test/init_fini test/exceptions
+	ADDITIONAL_TESTS:=test/init_fini test/exceptions test/coverage
 	ADDITIONAL_TESTS+=$(if $(SHARED),test/shared,)
 endif
 
--- a/src/rt/cover.d
+++ b/src/rt/cover.d
@@ -59,7 +59,7 @@ private
         Cover[] gdata;
         string  srcpath;
         string  dstpath;
-        bool    merge = true;
+        bool    merge;
     }
 }
 
--- a/src/rt/cover.d
+++ b/src/rt/cover.d
@@ -20,6 +20,7 @@ private
         import core.sys.posix.fcntl;
         import core.sys.posix.unistd;
     }
+    import core.stdc.config : c_long;
     import core.stdc.stdio;
     import core.stdc.stdlib;
     import rt.util.utf;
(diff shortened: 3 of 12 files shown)


MartinNowak:
Added a commit to perform race free updates of coverage files.

MartinNowak:
Race free merging of coverage files finally works on all platforms.

WalterBright:
Looks great, but I'm a bit perturbed by the automatic merging. My usage is along the lines of:
```
dmd std/file -main -unittest -cov
file
look at std-file.lst
```
and your change would make it:
```
dmd std/file -main -unittest -cov
rm std-file.lst
file
look at std-file.lst
```
which is annoying.

MartinNowak:
Let me quickly move the coverage settings to core.runtime, then it's less annoying to turn merging on.

MartinNowak:
Done.
It's still a bit surprising to have merging off by default, because other tools like gcov, gprof, and even dmd's -prof do merge by default.

MartinNowak:
Ping @WalterBright

MartinNowak:
3 weeks already, anyone?

src/core/runtime.d
```diff
@@ -313,13 +313,54 @@ struct Runtime
     }
 
 
+    /**
+     * Set source file path for coverage reports.
+     *
+     * Params:
+     *  path = The new path name.
+     */
+    static void coverSourcePath(string path)
```
WalterBright: What's the point of these trivial wrappers?

src/core/stdc/stdio.d
```diff
@@ -922,6 +922,8 @@ else version( CRuntime_DigitalMars )
     pure int  feof(FILE* stream)     { return stream._flag&_IOEOF;                       }
     ///
     pure int  ferror(FILE* stream)   { return stream._flag&_IOERR;                       }
+    ///
```
WalterBright: Redoing core headers like this should be in a separate PR. Can you please pull the stdio.d and windows.d out into separate PRs, and let me know so I can pull them for you?

src/core/runtime.d
MartinNowak: Documenting the functionality.
Some extern C functions buried in the runtime don't qualify as API.

WalterBright: I don't understand this justification at all.

src/core/stdc/stdio.d
MartinNowak: Here you go #1281.

But seriously why does this matter? It seems like an unnecessary appeal to correctness.
Considerably lowering the delay of reviews/pull requests should be a top priority, additional obstacles don't help with that.
Also from past experience of splitting work into separate but dependent pull requests, I'm ending up with a much higher review latency. Merging the shared library stuff took almost a year.

And reviewers don't easily see that a pull is blocking a huge chain of work, e.g. I can wait for ages until someone feels obliged to look at https://github.com/D-Programming-Language/dmd/pull/4650, but we're currently 3 people working on RefCounted/Unique, so it's kind of urgent b/c in a month we'll work on something else.

WalterBright:
> But seriously why does this matter?

'encapsulation' of changes, the same idea of encapsulation of functionality in functions, classes, etc.

I understand about delays, but having changes being encapsulated in more digestible and stand-alone chunks makes them more likely to get reviewed/pulled, not less.

src/core/runtime.d
MartinNowak: ???
Nobody knows that coverage generation can be configured.
If there are function that allow to change the behavior, we need to make them accessible and document them.

MartinNowak: Are you're wondering why I didn't just document the declarations for dmd_coverSourcePath?
Well b/c they have arcane names that don't fit into the rest of core.runtime's API.
I don't really like the namespacing `Runtime` struct either, but mixing it worse.
http://dlang.org/phobos/core_runtime.html#.Runtime

MartinNowak:
Anything left @WalterBright?

src/core/runtime.d
WalterBright: Yes, I was wondering why. Thanks for the explanation. But I don't think it justifies creating yet more names. I really, really do not want Phobos/Druntime to become a bloated swamp of same-only-different names. We've been through this discussion of "arcane" names before, and the pros and cons of endless renamings. Let's stop doing it.

andralex:
@MartinNowak please do not add do-nothing one-liners that forward from Annas to Caiaphas. There are a few odd names in druntime, but no need to add yet another layer on top of those. Let's leave them be. Getting the right recommendations fostered for the larger set of contributors must start with the leadership observing them. Thanks.

ibuclaw:
Exposing this in ``core.runtime`` is odd for me, as I don't use dmd-style coverage at all.  And from my cursory look around, gcov doesn't offer any such equivalents to be set at runtime either.

WalterBright:
1. We are NOT renaming functions any more. I've given the reasons why, at length, in a dozen threads. Please accept this and let's move on.
2. Defining such wrappers of module Y functions in module X means that linking in something from X will also pull in Y.
3. To make the `extern (C)` declaration of a Y function available from X, add this to X:
```
extern (C) T theYfunction();
```
The trailing ; means it is a declaration, not a definition. A wrapper is not required.

MartinNowak:
What's all the discussion about? There are 3 internal functions that need to be documented and made available for usage and it just makes sense to do that in the same form as the rest of the core.runtime API.

MartinNowak:
> Exposing this in core.runtime is odd for me, as I don't use dmd-style coverage at all. And from my cursory look around, gcov doesn't offer any such equivalents to be set at runtime either.

So any suggestion on how to deal with this @ibuclaw? Apparently we need the same settings for LDC. Maybe add assertions or better make them noops?

MartinNowak:
> We are NOT renaming functions any more. I've given the reasons why, at length, in a dozen threads. Please accept this and let's move on.

Those functions arent't currently and I still explicitly did __not__ rename them, not sure why you're making that argument.

> Defining such wrappers of module Y functions in module X means that linking in something from X will also pull in Y.

That's not true for our static libraries (we only link used functions) and dynamic libraries (they contain all code).

It is simply an API decision, how to name this functions. As the GDC/LDC discussion shows `dmd_coverSetMerge` is not adequate.

ibuclaw:
@MartinNowak maybe ``__traits(compiles, dmd_coverSetMerge)`` - I'm not one with answers on how to deal with internal APIs, just pointing out what breaks code convergence. ;-)

WalterBright:
Please, take the wrappers out.

MartinNowak:
Done

WalterBright:
Thanks!

WalterBright:
Auto-merge toggled on

PR #1127, opened 2015-01-25
refactor GC.Pool 

rainers:
- move getBits,setBits,etc, runFinalizers, getInfo into Pool
- split Pool into LargeObjectPool and SmallObjectPool

This yields a small perfomance improvement:
```
Test         Win64 master     this PR       Win32   master   this PR
R bulk             2.105 s,   2.074 s               0.765 s, 0.754 s
R resize           0.875 s,   0.868 s               0.757 s, 0.756 s
R string           0.023 s,   0.023 s               0.024 s, 0.024 s
R conalloc         0.703 s,   0.677 s               0.061 s, 0.063 s
R conappend        0.011 s,   0.011 s               0.012 s, 0.012 s
R concpu           0.835 s,   0.821 s               0.084 s, 0.085 s
R conmsg           0.583 s,   0.557 s               0.619 s, 0.597 s
R dlist            1.609 s,   1.388 s               1.448 s, 1.416 s
R huge_single      0.009 s,   0.009 s               0.010 s, 0.010 s
R rand_large       0.768 s,   0.633 s               3.084 s, 2.655 s
R rand_small       0.534 s,   0.535 s               0.616 s, 0.630 s
R slist            1.544 s,   1.374 s               1.382 s, 1.366 s
R testgc3          1.574 s,   1.401 s               1.087 s, 1.010 s
R tree1            0.610 s,   0.620 s               0.696 s, 0.706 s
R tree2            0.836 s,   0.805 s               0.814 s, 0.817 s
R words            0.808 s,   0.801 s               0.812 s, 0.818 s
```

PR #1227, opened 2015-04-20
Fix issue 14467 - arr.capacity sometimes erroneously returns 0

schveiguy:
(Note, reopened from https://github.com/D-Programming-Language/druntime/pull/1226)

In small object pool, the bits retrieved when using `Pool.getInfo` should be for the block, not the specific pointer. This was changed in https://github.com/D-Programming-Language/druntime/pull/1127 accidentally, I believe. This PR restores the original meaning.

Also, I noticed there is no other unit tests in this file. I don't know if I put the unit test in the right place.

Bugzilla: https://issues.dlang.org/show_bug.cgi?id=14467

### DIFF
--- a/src/gc/gc.d
+++ b/src/gc/gc.d
@@ -3155,6 +3155,7 @@ struct SmallObjectPool
 
         info.base = cast(void*)((cast(size_t)p) & notbinsize[bin]);
         info.size = binsize[bin];
+        offset = info.base - baseAddr;
         info.attr = getBits(cast(size_t)(offset >> shiftBy));
 
         return info;
@@ -3240,6 +3241,14 @@ struct SmallObjectPool
         (cast(List *)p).pool = &base;
         return first;
     }
+}
+
+unittest // bugzilla 14467
+{
+    int[] arr = new int[10];
+    assert(arr.capacity);
+    arr = arr[$..$];
+    assert(arr.capacity);
 }
 
 /* ============================ SENTINEL =============================== */


schveiguy:
Hang on don't merge this yet. I think the size is not properly fetched either.

schveiguy:
Nevermind, it's good. Deprecated the flaky win64 test.

MartinNowak:
> Deprecated the flaky win64 test.

So this one failed on `win-farm-1` @braddr.
You happend to know this failure from master/other pulls @schveiguy?

schveiguy:
> You happend to know this failure from master/other pulls @schveiguy?

yes, I've seen the issue in other PRs. I have no idea what it is. I'm somewhat unable to do any work to try and fix it, because I don't have a non-express Visual Studio installation. I assumed if I deprecated the test, it would possibly pass the next time ;)

Is there a reason the win 64 unit tests are so sparse in what gets logged? Even a successful test is pretty quiet.

MartinNowak:
> Is there a reason the win 64 unit tests are so sparse in what gets logged?

DM make

braddr:
Sorry, no idea what's up with it.  The host is a win7 home edition + visual studio 2013 community edition.  Nothing particularly special about it.

rainers:
> This was changed in #1127 accidentally, I believe. This PR restores the original meaning.

Yes, it wasn't meant to break things. Thanks for fixing it.

PR #1228, opened 2015-04-20
workaround issue 14468 - overload mismatch

MartinNowak:
[Issue 14468 – [Reg 2.068.0-devel] overload mismatch for template instance with typesafe variadic parameter](https://issues.dlang.org/show_bug.cgi?id=14468)


schveiguy:
I really don't like doing this. Is there a reason we can't wait for the regression to be fixed?

rainers:
I also just added a `0` as an initializer in my local copy. It's a bit annoying to have to deal with the modification when switching branches, but let's hope for a regression fix soon.

schveiguy:
> but let's hope for a regression fix soon.

Ask and ye shall receive: https://github.com/D-Programming-Language/dmd/pull/4603

PR #1232, opened 2015-04-26
Fix 14215 - Unnecessary imports of core.stdc.stddef for size_t and ptrdiff_t

joakim-noah:
[Ketmar is right](https://issues.dlang.org/show_bug.cgi?id=14215), `size_t` and `ptrdiff_t` were moved to `object.di` at least five years ago, so all these imports and references to them are wrong.  The four deleted public imports might still be inadvertently used to import the remaining `wchar_t` alias, but since that only seems likely for `core.sys.posix.sys.types`, I only left that one public import in.

### DIFF
--- a/src/core/stdc/stdint.d
+++ b/src/core/stdc/stdint.d
@@ -14,7 +14,7 @@
 
 module core.stdc.stdint;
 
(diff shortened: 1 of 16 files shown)


MartinNowak:
Auto-merge toggled on

PR #1233, opened 2015-04-27
fix Issue 14511 - Profiler does not work with multithreaded programs

WalterBright:
https://issues.dlang.org/show_bug.cgi?id=14511

This is a refactoring to modernize the code a bit and enhancement to handle multithreaded programs. The idea, suggested by Andrei, is to use thread local data for profiling data in each thread, then on thread termination merge the thread local data into the global profiling data.

The code could still be improved by using a hash table instead of a binary tree, and getting rid of the linked lists, but we'll save that for another day.

### DIFF
--- a/src/rt/trace.d
+++ b/src/rt/trace.d
@@ -1,7 +1,7 @@
 /**
  * Contains support code for code profiling.
  *
- * Copyright: Copyright Digital Mars 1995 - 2012.
+ * Copyright: Copyright Digital Mars 1995 - 2015.
  * License: Distributed under the
  *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
  *    (See accompanying file LICENSE)
@@ -9,8 +9,6 @@
  * Source: $(DRUNTIMESRC src/rt/_trace.d)
  */
 
-/*          Copyright Digital Mars 1995 - 2011.
- */
 module rt.trace;
 
 private
@@ -50,12 +48,12 @@ struct Symbol
         SymPair* Sfanout;       // list of called functions
         timer_t totaltime;      // aggregate time
         timer_t functime;       // time excluding subfunction calls
-        ubyte Sflags;
+        ubyte Sflags;           // SFxxxx
         uint recursion;         // call recursion level
-        char[] Sident;          // name of symbol
+        const(char)[] Sident;   // name of symbol
 }
 
-const ubyte SFvisited = 1;      // visited
+enum ubyte SFvisited = 1;      // visited
 
 
 //////////////////////////////////
@@ -70,23 +68,21 @@ struct Stack
     timer_t subtime;            // time used by all subfunctions
 }
 
-__gshared                       // doesn't work with multithreaded code anyway
-{
-    Symbol* root;               // root of symbol table
+Symbol* root;               // root of symbol table
+bool trace_inited;
 
-    Stack* stack_freelist;
-    Stack* trace_tos;           // top of stack
-    int trace_inited;           // !=0 if initialized
+Stack* stack_freelist;
+Stack* trace_tos;           // top of stack
+
+__gshared
+{
+    Symbol* groot;              // merged symbol table
+    int gtrace_inited;          // !=0 if initialized
+
     timer_t trace_ohd;
 
-    Symbol** psymbols;
-    uint nsymbols;              // number of symbols
-
     string trace_logfilename = "trace.log";
-    FILE* fplog;
-
     string trace_deffilename = "trace.def";
-    FILE* fpdef;
 }
 
 ////////////////////////////////////////
@@ -118,13 +114,13 @@ int trace_setdeffilename(string name)
 ////////////////////////////////////////
 // Output optimal function link order.
 
-static void trace_order(Symbol *s)
+private void trace_order(FILE* fpdef, Symbol *s)
 {
     while (s)
     {
-        trace_place(s,0);
+        trace_place(fpdef, s, 0);
         if (s.Sl)
-            trace_order(s.Sl);
+            trace_order(fpdef, s.Sl);
         s = s.Sr;
     }
 }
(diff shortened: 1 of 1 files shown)


MartinNowak:
Segfaults on OSX.

WalterBright:
Yes, I'm looking into it. Was too tired last night :-)

WalterBright:
This pull requires https://github.com/D-Programming-Language/dmd/pull/4620

src/rt/trace.d
```diff
@@ -70,23 +68,24 @@ struct Stack
...
-    timer_t trace_ohd;
+    bool trace_inited;
+}
 
-    Symbol** psymbols;
-    uint nsymbols;              // number of symbols
+__gshared
+{
+    Symbol* groot;              // merged symbol table
+    int gtrace_inited;          // !=0 if initialized
```
andralex: `bool`? or are you worried about word tearing? Also: would be nice if the order were the same in the TLS and global.

src/rt/trace.d
```diff
@@ -132,8 +131,9 @@ static void trace_order(Symbol *s)
 //////////////////////////////////////////////
 //
 
-static Stack* stack_malloc()
-{   Stack *s;
+private Stack* stack_push()
+{
+    Stack *s;
```
andralex: y u no use `?:`

src/rt/trace.d
```diff
@@ -141,13 +141,15 @@ static Stack* stack_malloc()
     }
     else
         s = cast(Stack *)trace_malloc(Stack.sizeof);
```
andralex: so this code has no {} but the previous one has (compounded with the horrific `{` followed by the stuff). So use one or the other but not all three in one.

src/rt/trace.d
```diff
@@ -70,23 +68,24 @@ struct Stack
     timer_t subtime;            // time used by all subfunctions
 }
 
-__gshared                       // doesn't work with multithreaded code anyway
+static
```
JakobOvrum: `static` doesn't mean anything at module scope

src/rt/trace.d
```diff
@@ -170,12 +170,12 @@ static int sympair_cmp(in void* e1, in void* e2)
...
-static void trace_place(Symbol *s, uint count)
-{   SymPair* sp;
-    SymPair** base;
-
+private void trace_place(FILE* fpdef, Symbol* s, uint count)
+{
     if (!(s.Sflags & SFvisited))
-    {   size_t num;
+    {
+        SymPair* sp;
```
andralex: move declaration to point of initialization

src/rt/trace.d
```diff
@@ -170,12 +170,12 @@ static int sympair_cmp(in void* e1, in void* e2)
...
+        size_t num;
```
andralex: move declaration to point of initialization

src/rt/trace.d
```diff
@@ -170,12 +170,12 @@ static int sympair_cmp(in void* e1, in void* e2)
...
         uint u;
```
andralex: move declaration to point of initialization

src/rt/trace.d
```diff
@@ -192,7 +192,7 @@ static void trace_place(Symbol *s, uint count)
             return;
 
         // Allocate and fill array
-        base = cast(SymPair**)trace_malloc(SymPair.sizeof * num);
+        auto base = cast(SymPair**)trace_malloc(SymPair.sizeof * num);
```
andralex: noice

src/rt/trace.d
```diff
@@ -209,14 +209,12 @@ static void trace_place(Symbol *s, uint count)
         for (u = 0; u < num; u++)
         {
             if (base[u].count >= count)
-            {   uint u2;
-                uint c2;
-
-                u2 = (u + 1 < num) ? u + 1 : u;
-                c2 = base[u2].count;
+            {
+                auto u2 = (u + 1 < num) ? u + 1 : u;
```
andralex: auto u2 = (u <= num) ? u + 1 : u;

src/rt/trace.d
```diff
@@ -303,26 +285,24 @@ static int symbol_cmp(in void* e1, in void* e2)
...
-    for (u = 0; u < nsymbols; u++)
+    foreach (s; psymbols)
     {
-        Symbol* s = psymbols[u];
         timer_t tl,tr;
         timer_t fl,fr;
         timer_t pl,pr;
         timer_t percall;
-        char[8192] buf;
+        char[8192] buf = void;
```
andralex: yah

src/rt/trace.d
```diff
@@ -348,31 +328,34 @@ static void trace_times(Symbol* root)
 ///////////////////////////////////
 // Initialize.
 
-static void trace_init()
+private void trace_init()
 {
-    if (!trace_inited)
+    synchronized
```
andralex: so AFAIK this `synchronized` only protects `gtrace_inited` - a comment would be nice

src/rt/trace.d
```diff
@@ -380,68 +363,160 @@ static void trace_init()
...
-        trace_inited = 2;
+    // And free the thread local stack's memory
+    while (stack_freelist)
+    {
+        auto n = stack_freelist.prev;
+        stack_free(stack_freelist);
+        stack_freelist = n;
+    }
+
+    synchronized
```
andralex: ... whereas this one protects `groot` and a few other globals, again it would be nice if a comment specified all.

What's important here is that the sets of globals controlled by each unparameterized `synchronized` statement are disjoint.

src/rt/trace.d
```diff
@@ -380,68 +363,160 @@ static void trace_init()
...
+            trace_times(fplog, psymbols);
             fclose(fplog);
+
+            trace_free(psymbols.ptr);
+            psymbols = null;
         }
 
         // Output function link order
-        fpdef = fopen(trace_deffilename.ptr,"w");
+        FILE* fpdef = fopen(trace_deffilename.ptr,"w");
```
andralex: if file can't be open, write message to `stderr` - this is very nice for users

andralex:
Nice, aside for the OSX failure I also added a few nits that may deserve a look at.

src/rt/trace.d
WalterBright: not the same

src/rt/trace.d
WalterBright: won't work here

src/rt/trace.d
WalterBright: > bool ? or are you worried about word tearing?

gtrace_inited can take on the values 0, 1 or 2

src/rt/trace.d
andralex: urgh what was I thinking

andralex:
Cool. We only need a variable `hodor` now to complement `groot`.

andralex:
Auto-merge toggled on

andralex:
I toggled but you need to fix failures.

PR #1234, opened 2015-04-27
TickDuration UFCS "to" was broken for float

burner:
https://github.com/D-Programming-Language/druntime/pull/1190 was not tested throughout, it has a missing "this".

TickDuration t;
auto tf = to!("nsecs",float)(t);

Update:
found this while making the plotting for my benchmark module better

Update2:
best PR number ever :-)

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -1889,14 +1889,13 @@ T to(string units, T, D)(D td) @safe pure nothrow @nogc
         units == "msecs" ||
         units == "usecs" ||
         units == "hnsecs" ||
-        units == "nsecs") &&
-       ((__traits(isIntegral, T) && T.sizeof >= 4) || __traits(isFloating, T)))
+        units == "nsecs"))
 {
     static if(__traits(isIntegral, T) && T.sizeof >= 4)
     {
         enum unitsPerSec = convert!("seconds", units)(1);
 
-        return cast(T)(td.length / (TickDuration.ticksPerSec / cast(real)unitsPerSec));
+        return cast(T) (td.length / (TickDuration.ticksPerSec / cast(real) unitsPerSec));
     }
     else static if(__traits(isFloating, T))
     {
@@ -1906,7 +1905,8 @@ T to(string units, T, D)(D td) @safe pure nothrow @nogc
         {
             enum unitsPerSec = convert!("seconds", units)(1);
 
-            return this.to!("seconds", T)() * unitsPerSec;
+            return cast(T) (td.length /
+                (TickDuration.ticksPerSec / cast(real) unitsPerSec));
         }
     }
     else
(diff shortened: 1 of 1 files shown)


src/core/time.d
```diff
@@ -1703,7 +1702,7 @@ T to(string units, T, D)(D td) @safe pure nothrow @nogc
         {
             enum unitsPerSec = convert!("seconds", units)(1);
 
-            return this.to!("seconds", T)() * unitsPerSec;
+            return (td.length / cast(T)TickDuration.ticksPerSec) * unitsPerSec;
```
andralex: space after `cast(T)`

src/core/time.d
```diff
@@ -1724,35 +1723,68 @@ unittest
...
 
-        static if(is(F : float))
-        {
-            assert((t1f - cast(F)t1v) < 0.0001);
-            assert((t2f - cast(F)t2v) < 0.0001);
-            assert(t3f - (cast(F)t1v) / (cast(F)t2v) < 0.0001);
-            assert(t4f - (cast(F)(t1v - t2v)) < 0.0001);
-        }
-        else
+        foreach(F; _TypeTuple!(int,uint,long,ulong,float,double,real))
```
andralex: space after `foreach` etc

src/core/time.d
```diff
@@ -4329,6 +4361,48 @@ string numToString(long value) @safe pure nothrow
...
+unittest
+{
+    string a = numToString(1);
+    assert(a == "1");
+
+    a = numToString(-1);
+    assert(a == "-1");
+}
+
+/++ A six digit double to string converstion function
```
andralex: typo

src/core/time.d
```diff
@@ -4329,6 +4361,48 @@ string numToString(long value) @safe pure nothrow
...
++/
+string numToString(double value) @safe pure nothrow
+{
+    immutable negative = value < 0;
+    string result = (negative ? "-" : "") ~ numToString(cast(long)value);
+    result ~= '.';
+    result ~= numToString(cast(long)(_abs((value - cast(long)value) * 1000000)));
```
andralex: this is quite inefficient, needs careful optimization

andralex:
this needs a fair amount of work

src/core/time.d
schveiguy: This is only used in unit tests. I'm thinking this needs to go into a `version(unittest)` block. core.time shouldn't be in the business of providing string conversion functions (and take off the ddoc for it)

`numToString(long)` is a necessary evil (not sure that should be publicly documented either, maybe it actually should go into core.internal) because of the requirements for purity during time printing, but this one isn't used in the actual implementation.

jmdavis: It shouldn't have `/++` on it - just `/+` - but it _is_ used in the actual code, not just the unit tests. Both `Duration` and `MonoTime` use it in their `toString` functions, since the proper Phobos functions aren't available.

schveiguy: They are converting doubles to strings? How did they do it before Robert added this function?

jmdavis: Wait. You're right. I read over the code too quickly. The `numString(long)` overload is required` and is used in `Duration` and `MonoTime`. `numString(double)` is new.

jmdavis: The fact that `TickDuration` uses floating points at all is one of the reasons that we're in the process of replacing it with `MonoTime`.

schveiguy: I see no harm in converting to/from floating points. I have found it very useful in fact in real code.

burner:
@jmdavis how can I use monotime with the StopWatch. I found this bug because I was working with the StopWatch.

@andralex indeed this needs works

src/core/time.d
burner: I will remove the /++ , but this function is in the private section of the file

jmdavis:
> how can I use monotime with the StopWatch. I found this bug because I was working with the StopWatch.

You can't right now, because that would be a breaking change. The benchmarking code in std.datetime would mostly work as-is with `MonoTime` and `Duration` instead of `TickDuration` except for that fact that it's primarily return types that change (so they can't be overloaded). So, those functions are going to need to be replaced with versions using `MonoTime` and `Duration` in another module like std.benchmark, at which point, the versions in std.datetime, which use `TickDuration` can be deprecated along with `TickDuration` itself, and it should then be easier to deal with the benchmarking functions, since the confusion that is `TickDuration` won't be needed. Those new functions are on my TODO list, but I haven't managed to find much time for D in the past several months. I think that that will be changing soon, but my primary focus at the moment will be dconf, or I won't have my talk ready in time, so changes like that will almost certainly have to come after. So until I get to it (or someone else does), we unfortunately have to make do with `TickDuration` when dealing with the benchmarking functions. But we did finally get `MonoTime` in the last release, so the necessary building blocks are now there.

src/core/time.d
jmdavis: It shouldn't end up in the documentation if it's private (and it probably won't if it's `version(unittest)`), but using `/++` indicates that it's intended to be in the documentation (since that would be the only reason to use two pluses instead of one), whereas `/+` doesn't.

burner: it is version(unittest) now

burner:
```
shared static ~this() groot = (nil)
timed out after 3600 seconds, step failed
```

not sure what is the cause of this, but the core.time tests pass

mrkline:
Thanks for really digging into this - my original PR was extremely simple and just hoisted `to` out of `Duration` so that it matched the rest of the standard library's semantics. I noticed some of this seemed a bit off, but didn't want to hoist it out _and_ do a rewrite all at once.

src/core/time.d
```diff
@@ -1927,35 +1927,68 @@ unittest
...
-        static if(is(F : float))
-        {
-            assert((t1f - cast(F)t1v) < 0.0001);
-            assert((t2f - cast(F)t2v) < 0.0001);
-            assert(t3f - (cast(F)t1v) / (cast(F)t2v) < 0.0001);
-            assert(t4f - (cast(F)(t1v - t2v)) < 0.0001);
-        }
-        else
+        foreach (F; _TypeTuple!(int,uint,long,ulong,float,double,real))
         {
```
MartinNowak: It's usually better to turn those foreach into a function literal to mitigate the expensive O(N^2) optimizations of the backend.

MartinNowak:
Auto-merge toggled on

burner:
thanks

PR #1235, opened 2015-04-28
Revert "Make .destroy() work with attribute inference"

JakobOvrum:
Reverts D-Programming-Language/druntime#1181

`__traits(getMember, ...)` is not allowed to get private (or otherwise inaccessible) fields, but primitives like `destroy` must recursively destruct *all* fields to be correct.

I don't know how to fix this.


JakobOvrum:
A number of high-impact `@safe`-related PRs are stalled because of this, so suggestions for how we can fix it would be highly appreciated.

schveiguy:
Crap, well at least it didn't get into the release!

What happens if you destroy a struct with private members? Does it just fail to compile? Can we just revert to old behavior if that is the case?

rsw0x:
>A number of high-impact @safe-related PRs are stalled because of this, so suggestions for how we can fix it would be highly appreciated.

this undoes a lot of @nogc inference in phobos that was enabled with this patch too AFAIK.

schveiguy:
As far as the code is concerned, this looks correct as a revert.

schveiguy:
What is the test code that proves your statement? I can't get it to fail, even with private members.

andralex:
`tupleof` enumerates all state, private or not. Should be useful here.

JakobOvrum:
I'll pursue the `tupleof` approach ASAP. I think I did try `tupleof` at first but ran into some issues with extracting the field name, which has always been hacky with it. Probably not a big deal, in the worst case I can simply use both.

JakobOvrum:
Alright, `_destructRecurse` uses `tupleof` and is fine, it is `_postblitRecurse` that is the problem. I think it can be fixed by changing it to this:
```
private string _genFieldPostblit(S)()
{
    import core.internal.traits : hasElaborateCopyConstructor;

    string code;
    foreach(i, FieldType; S.tupleof)
    {
        static if(hasElaborateCopyConstructor!S)
        {
            code ~= `
                _postblitRecurse(s.tupleof[` ~ to!string(i) ~ `]);
                scope(failure) _destructRecurse(s.tupleof[` ~ to!string(i) ~ `]);
            `;
        }
    }
    return code;
}

// Public and explicitly undocumented
void _postblitRecurse(S)(ref S s)
    if (is(S == struct))
{
    mixin(_genFieldPostblit!S());

    static if (__traits(hasMember, S, "__postblit"))
        s.__postblit();
}
```
However, this solution needs to convert an integer to a string for the mixin code. Does druntime have any facilities for this? It's just for CTFE, so it doesn't need to be performant.

An alternative would be to forego the convenience of `scope(failure)` and replicate its functionality with try-catch statements. Hopefully that's possible without string mixins and recursion.

JakobOvrum:
Filed a fix at #1243

PR #1236, opened 2015-04-28
use foreach to decode in rt/adi.d

WalterBright:
Use D core features.

### DIFF
--- a/src/rt/adi.d
+++ b/src/rt/adi.d
@@ -1,16 +1,14 @@
 /**
  * Implementation of dynamic array property support routines.
  *
- * Copyright: Copyright Digital Mars 2000 - 2010.
- * License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Copyright: Copyright Digital Mars 2000 - 2015.
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
  * Authors:   Walter Bright
+ * Source: $(DRUNTIMESRC src/rt/_adi.d)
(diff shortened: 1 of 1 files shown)


schveiguy:
In my previous tests of foreach vs. hand-written or calling decode, foreach was a lot slower because it used non-inlined code. Has that changed?

andralex:
Whichever the case, it's good to dogfood this way because we create pressure for better implementation.

andralex:
Auto-merge toggled on

WalterBright:
I agree. Removing the workarounds is the first step to improving things. foreach over strings needs to be replaced with a range-based implementation.

WalterBright:
See also https://github.com/D-Programming-Language/druntime/pull/1237 for a similar change.

schveiguy:
Wait, this is in druntime. Why is calling druntime more dogfooding than calling druntime? There's also no way autodecoding or encoding string ranges are getting in here, is there?

schveiguy:
Auto-merge toggled off

schveiguy:
Re-toggle if you disagree, I just don't think it's the right move here. If it were Phobos, that's different.

andralex:
foreach with autodecoding has been in D for a long time

andralex:
Auto-merge toggled on

schveiguy:
Sure, the argument was "it should be done with auto-decoding ranges anyway" which probably aren't coming to druntime. So that's what I was responding to.

None of the arguments here make much sense, but this entire file seems riddled with inefficiencies and dangerous code (sorting an array of code points?), so I'm not going to argue any more about it.

WalterBright:
> the argument was "it should be done with auto-decoding ranges anyway" which probably aren't coming to druntime.

But they are!

PR #1237, opened 2015-04-29
use foreach to encode in rt/cover.d

WalterBright:


schveiguy:
See my comments on #1236, I don't agree with this one either.

andralex:
eh, in this case the function call seems a tad cleaner :)

WalterBright:
> in this case the function call seems a tad cleaner

It doesn't matter because efficiency matters naught in this piece of code. What matters is when foreach is fixed, all this stuff gets fixed, too.

schveiguy:
Is there a way we can fix `foreach(dchar; ...)` before making all these changes? I'm a bit confused at the proactiveness here. These changes are easy to do later, it's not complex. I'm concerned there will be a false sense of improvement, since you will not be comparing direct calls to decode/encode with foreach, but rather the old crappy performance of foreach to the new performance (which very well may outperform encode/decode or be on-par).

I have a hard time believing someone else saying "It doesn't matter that this kills performance now, it will be fixed later when druntime is rewritten to use auto-decoding char ranges" would pass muster...

andralex:
By "cleaner" I didn't mean "faster", just that I find the new code forced and more difficult to grok the intent of. The previous code was one rather straightforward function call. To me it seems a questionable application of foreach. I'd prolly let it be if it were in existing code, but am opposed to committing this as it's a net negative.

WalterBright:
> I have a hard time believing someone else saying "It doesn't matter that this kills performance now, it will be fixed later when druntime is rewritten to use auto-decoding char ranges" would pass muster...

Please don't invent quotes for me. I meant in THIS particular piece of code, you'll never see any performance issue.

@andralex the only users of rt.util.utf are rt.adi, rt.cover, and the foreach code. rt.util.utf is not exactly 'clean' code. rt.adi's usage is now removed, that leaves rt.cover. By eliminating rt.cover's use of it, it leaves me free to re-engineer how foreach on text is done.

Note that rt.util.utf is very duplicative of what std.utf does. It's copy-pasta code, and dealing with that is far more important than a couple lines in rt.cover.

schveiguy:
I'm not trying to invent quotes for you. I'm interpreting what you said. My fake quote of another person is how your statements sounded to me.

But in any case, if this is necessary in order to fix foreach, then that's fine, I get that now.

schveiguy:
Note that it would be helpful to fully explain your purpose up front so we can review with the right context.

MartinNowak:
> Note that it would be helpful to fully explain your purpose up front so we can review with the right context.

Seems like he wants to improve foreach with auto decoding and replace all existing transcoding code in druntime.
The scope of this pull is too small, you should have made one that removes rt.util.utf.

MartinNowak:
Auto-merge toggled on

PR #1239, opened 2015-04-29
Refactor: rt.aApply

WalterBright:
This is a straightforward refactor to modernize the style a bit, just because I was annoyed when reading it.

### DIFF
--- a/src/rt/aApply.d
+++ b/src/rt/aApply.d
@@ -6,16 +6,11 @@
  * Copyright: Copyright Digital Mars 2004 - 2010.
  * License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
  * Authors:   Walter Bright
- */
-
-/*          Copyright Digital Mars 2004 - 2010.
- * Distributed under the Boost Software License, Version 1.0.
- *    (See accompanying file LICENSE or copy at
- *          http://www.boost.org/LICENSE_1_0.txt)
+ * Source: $(DRUNTIMESRC src/rt/_aApply.d)
  */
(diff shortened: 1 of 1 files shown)


WalterBright:
20 fewer lines of code, too!

src/rt/aApply.d
```diff
@@ -24,19 +19,18 @@ private import rt.util.utf;
...
-
-        d = aa[i];
+    for (size_t i = 0; i < len; )
+    {
+        dchar d = aa[i];
         if (d & 0x80)
             d = decode(aa, i);
         else
-            i++;
+            ++i;
```
yebblies: Is this really necessary?

MartinNowak: You mean optimization wise? It shouldn't nowadays, but it's a good practice, because post-increment __can__ incur extra cost for UDTs.

MartinNowak:
Auto-merge toggled on

src/rt/aApply.d
yebblies: Yes I know, but this is not a UDT and this style is not used consistently in Walter's code.  I vaguely recall Walter arguing against its use before, although I could be wrong.

MartinNowak: It's not worth arguing such details IMO.
If he changed his mind it simply won't be consistent with what he did before.

yebblies: If this was new code then I'd agree, but he deliberately changed the style of the code, and I would like to know why.

yebblies: It wouldn't be worth blocking the PR over etc

PR #1240, opened 2015-04-29
fix Issue 14519 - [Enh] foreach on strings should return replacementDcha...

WalterBright:
...r rather than throwing

https://issues.dlang.org/show_bug.cgi?id=14519

First step towards having foreach on strings use ranges rather than opApply.

### DIFF
--- a/src/rt/util/utf.d
+++ b/src/rt/util/utf.d
@@ -27,8 +27,16 @@
  */
 module rt.util.utf;
 
-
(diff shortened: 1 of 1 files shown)


CyberShadow:
Please no, see discussion in Bugzilla

MartinNowak:
There isn't going to be any argument to justify silent data corruption, see #1279 for an alternative.

PR #1241, opened 2015-04-29
remove debugging printf

WalterBright:


yebblies:
So that's what that is.

yebblies:
Auto-merge toggled on

PR #1242, opened 2015-04-29
[Refactor] Don't use loop in `rt.aaA.nextpow2`.

denis-sh:
Also add unittest.


denis-sh:
@MartinNowak, is `nextpow2` really supposed to work with zero or should it reject such case?

schveiguy:
Auto-merge toggled on

schveiguy:
I think for the time it takes to check for zero, it's not worth allowing potentially undefined behavior

MartinNowak:
Thanks.
This is not performance relevant anyhow.

PR #1181, opened 2015-02-26
Make .destroy() work with attribute inference

JakobOvrum:
Discussed in [this thread](http://forum.dlang.org/thread/krccldftrbbczahasvbi@forum.dlang.org).

With this patch, the `@safe`ty etc. of object.destroy depends on the destructor of the struct.

Also adds `postblitRecurse` to enable explicit copying with attribute inference, to be used in `emplace`.

Attribute inference for destruction/postblit of fixed-length arrays with struct elements seems to be broken even for implicit/in-built destruction/postblit, so that needs to be fixed in the compiler before it's tackled here.

This impacts attribute inference in a couple of key Phobos building blocks, such as `emplace` and the `Array` container. Grepping for `malloc` tends to be a good way to find code that uses explicit destruction.

I plan to file one or more dependent Phobos PRs to leverage this.

PR #1243, opened 2015-04-30
Fix _postblitRecurse for structs with private fields

JakobOvrum:
Fixup for #1181


JakobOvrum:
Ping @monarchdodra, @MartinNowak. Would be nice to work out the dependency chain for https://github.com/D-Programming-Language/phobos/pull/3031 .

monarchdodra:
Yeah, LGTM. Let's get this merged quickly. Unless I'm mistaken, 1181 basically broke any code that uses explicit destruction on structs with private fields?

JakobOvrum:
Destruction works fine as it uses tupleof, it is only postblit that is broken.

src/core/internal/convert.d
```diff
@@ -633,3 +633,41 @@ const(ubyte)[] toUbyte(T)(ref T val) if (is(T == struct) || is(T == union))
         return (cast(const(ubyte)*)&val)[0 .. T.sizeof];
     }
 }
+
+char[] itoa(ulong n) pure nothrow @safe
```
monarchdodra: This seems to be a duplicate of something in demangle.d:
https://github.com/D-Programming-Language/druntime/blob/5afecf8d7aa21f3269298767ec8d2c753103e6ba/src/core/demangle.d#L1706

It's implementation seems to be more efficient: It looks like you are writting the digits lsd to msd, and then reversing the digits? Instead, you can write the digits msd to lsd, and dup `[20 - length .. length]`.

Bonus points for you though because you seem to handle the case where n == 0, whereas the demangle one does not seem to?

I get that returning a `char[]` means purity can make it a string, but it also means that you will always allocate for things such as "0". I guess it is moot here though.

Both functions are used differently, but would it be possible to consolidate the implementations somewhat?

monarchdodra:
> Destruction works fine as it uses tupleof, it is only postblit that is broken.

Ah. Ok. I was wondering why we were not observing widespread destruction :)

src/core/internal/convert.d
JakobOvrum: It is only used in CTFE, so I didn't see the GC-centric interface as an issue.

If core.demangle and object were to use the same implementation, would core.internal.convert still be a good place for it?

monarchdodra: > It is only used in CTFE, so I didn't see the GC-centric interface as an issue.

Agreed. I was just pointing out a more general design issue, where an interface that returns mutable can't exploit immutability to return aliased results. As a random thought outside of this PR.

> If core.demangle and object were to use the same implementation, would core.internal.convert still be a good place for it?

I have am not very savvy wrt how druntime is organized. @MartinNowak should be able to advise.

If the interfaces don't allow consolidation, or it is two difficult, then I'm fine with it. I *would* at the very least prefer that they have similar implementations.

JakobOvrum:
`itoa` now has a better implementation. I'm not sure if touching `core.demangle` is a good idea here; it doesn't always need the result but just the length of the result, which makes the interface awkward for general use.

JakobOvrum:
Ping @schveiguy, @MartinNowak. Got any time to look at this?

src/object.di
```diff
@@ -632,17 +632,17 @@ private void _destructRecurse(E, size_t n)(ref E[n] arr)
...
-    foreach(fieldName; FieldNameTuple!S)
+    foreach(i, FieldType; typeof(S.init.tupleof))
     {
-        static if(hasElaborateCopyConstructor!(typeof(__traits(getMember, S.init, fieldName))))
+        static if(hasElaborateCopyConstructor!FieldType)
         {
             code ~= `
-                _postblitRecurse(s.` ~ fieldName ~ `);
-                scope(failure) _destructRecurse(s. ` ~ fieldName ~ `);
+                _postblitRecurse(s.tupleof[` ~ itoa(i) ~ `]);
```
MartinNowak: Doesn't `i.stringof` work?

src/object.di
```diff
@@ -632,17 +632,17 @@ private void _destructRecurse(E, size_t n)(ref E[n] arr)
...
 private string _genFieldPostblit(S)()
 {
+    import core.internal.convert : itoa;
     import core.internal.traits : hasElaborateCopyConstructor;
-    import core.internal.traits : FieldNameTuple;
 
     string code;
-    foreach(fieldName; FieldNameTuple!S)
+    foreach(i, FieldType; typeof(S.init.tupleof))
     {
```
MartinNowak: Why not simply move that loop to `_postblitRecurse` and use `foreach (ref field; s.tupleof)`?
That would make the mixin superfluous.

MartinNowak:
You need to update the pull because we removed object.di.

src/object.di
JakobOvrum: I tried that initially, but the static foreach introduces a new scope for `scope(failure)`, so it doesn't behave correctly.

JakobOvrum:
> You need to update the pull because we removed object.di.

`object.di` is still in the repository. If it's not going to be kept up to date it needs to be removed.

JakobOvrum:
Now only modifies `object_.d`.

src/object.di
JakobOvrum: It appends a `u` for unsigned but apart from that it appears to work. If it can be relied on maybe we should use that instead.

JakobOvrum:
Removed `itoa` in favour of `i.stringof`.

MartinNowak:
Auto-merge toggled on

PR #1244, opened 2015-04-30
Fix Issue 12289 - incorrect core.stdc.stdio.fpos_t alias and mbstate_t for glibc

joakim-noah:
I noticed `fpos_t` was defined wrongly in `core.stdc.stdio` a while back, finally fixing it.  I couldn't find where [`__darwin_off_t` was defined in Apple's libc headers for `OSX`](http://www.opensource.apple.com/source/Libc/Libc-1044.1.2/include/stdio.h), probably because the definition is automatically generated, so @jacob-carlborg or someone else will have to fill me in on that one.  [I read the headers for Solaris](http://src.illumos.org/source/xref/illumos-gate/usr/src/ucbhead/stdio.h#78), @redstar can tell me if I got it wrong.  The others were checked with a test C file run through the native preprocessor, so they should be okay.

I had to fix `mbstate_t` for glibc, didn't bother checking that one on the other OS's.  I also deleted the comments about removing the shared file handle `_iobuf` eventually, as it's used more than I thought in `std.stdio`.

### DIFF
--- a/src/core/stdc/stdio.d
+++ b/src/core/stdc/stdio.d
@@ -25,10 +25,6 @@ private
   {
     import core.sys.posix.sys.types;
   }
-  else version (Android)
-  {
-    import core.sys.posix.sys.types: off_t;
-  }
 }
 
 extern (C):
@@ -237,7 +233,7 @@ enum
 version( CRuntime_DigitalMars )
 {
     ///
-    alias int fpos_t; //check this
+    alias c_long fpos_t;
 
     ///
     struct _iobuf
@@ -258,7 +254,7 @@ version( CRuntime_DigitalMars )
 else version( CRuntime_Microsoft )
 {
     ///
-    alias int fpos_t; //check this
+    alias long fpos_t;
 
     ///
     struct _iobuf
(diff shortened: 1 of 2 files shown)


joakim-noah:
Hmm, it appears I've hit some strange `static if` bug, perhaps similar to [this long-standing forward referencing one](https://issues.dlang.org/show_bug.cgi?id=3743).  If I comment out the `static if` check around `off_t` in `core.sys.posix.sys.types`, it has no problem finding `off_t`, compiling and passing the unit tests on linux/x86.  Anyone run into this before?

MartinNowak:
Yeah there are some subtle forward reference issues when something is wrapped in static if.

joakim-noah:
Looked into this a bit more, it appears that the issue crops up in the semantic pass for `core.thread`, `core.time`, and `test_runner.d`, not the semantic pass for `core.stdc.stdio` itself, as the error message implies.  That could very well be the forward referencing issue with `static if` somehow cropping up when `core.stdc.stdio` is imported in the wrong order somewhere.

I've worked around it by simply using `long` instead of `off_t` for now.  It can be updated later when the `static if` issue is fixed in dmd.  I've added this info to that `static if` bug report.

jacob-carlborg:
`__darwin_off_t` is defined in "sys/_types.h" as: `typedef __int64_t	__darwin_off_t`. `__int64_t` in "i386/_types.h" as `typedef long long		__int64_t`. I just used "grep" in the system include directory.

joakim-noah:
Thanks, updated with that.  I don't have an OS X install to grep.

joakim-noah:
Ping.  I think this is ready for merging.

PR #1245, opened 2015-04-30
check whether TLS is ready to be used in the profiler

MartinNowak:
- the emulated TLS on OSX is initialized by druntime,
  so we can't profile until that is done


WalterBright:
Thanks for pointing out how to do this, I appreciate it and will likely need it in the future. But I don't really want to add 21 lines of code just so C main() can be pointlessly profiled. C main() is the usual entry point of a D program, and so of course the runtime isn't initialized yet, even if the TLS works.

In fact, the TLS, while existing on non-OSX platforms, is not initialized when C main() is called. (The TLS constructors have not been run.) It is only default initialized. Trying to make the profiler run on an un-initialized runtime is just beyond the scope of the profiler.

MartinNowak:
Fair enough, but other entry points than C main are possible and profiling won't work for them.

PR #1247, opened 2015-05-01
Simplify the check if cent/ucent is supported.

redstar:
The same check is used in Phobos in module std.traits.


yebblies:
Auto-merge toggled on

PR #1248, opened 2015-05-02
fix Issue 14476 - core.thread unit tests failing on FreeBSD 9+

MartinNowak:
- the newly created thread fails because of a stack overflow
  in a syscall to sbrk triggered by pthread_attr_get_np/calloc

- [Issue 14476 – core.thread unit tests failing on FreeBSD 9+](https://issues.dlang.org/show_bug.cgi?id=14476)

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -5214,6 +5214,7 @@ version (FreeBSD) unittest
 
 unittest
 {
-    auto thr = new Thread(function{}, 10).start();
+    // use >PAGESIZE to avoid stack overflow (e.g. in an syscall)
+    auto thr = new Thread(function{}, 4096 + 1).start();
     thr.join();
 }


jmdavis:
Wait, a stack size of 10 causes a stack overflow, but 5006 doesn't? Clearly, I'm going to have to dig into this deeper to have anything intelligent to say about it.

MartinNowak:
> Wait, a stack size of 10 causes a stack overflow, but 5006 doesn't? Clearly, I'm going to have to dig into this deeper to have anything intelligent to say about it.

They are round up to the next page size, means 4KiB vs. 8KiB.

MartinNowak:
Ping

schveiguy:
The change LGTM. I have no way of testing that it works, and we can't trust the auto-tester here.

> Wait, a stack size of 10 causes a stack overflow, but 5006 doesn't?

Maybe I'm not understanding here, but this makes sense to me:

10 < 5006, therefore 5006 bytes for stack means more room and less chance of overflow.

> They are round up to the next page size, means 4KiB vs. 8KiB.

Can you put a comment in to document what's happening and why that number is chosen? Would it be incorrect to just put the actual size you are looking for as the parameter, or are we testing that feature with this unit test?

joakim-noah:
I tried this fix out on 9.1 i386 and it turns out that this test is not the problem after all on that system.  It appears that the tests hang in that VM because of the FreeBSD-only unit test that was meant to fix some threading issues, but doesn't work on 9.1 i386 for some reason.

I mistook this test as the problem because when I commented this test out, the tests passed right afterwards when I ran them once, but it turns out they'll pass around 10% of the time randomly.  After rerunning the tests 10 or so times, I realize that this test is not the issue on 9.1 i386.

MartinNowak:
Updated with a comment.

I tested the bug on FBSD-10 X86_64 and the stack overflow was really the problem here.

joakim-noah:
Works fine on FreeBSD 9.1 i386 either way, LGTM.

PR #1249, opened 2015-05-02
silence deprecation warning in unittest for deprecated HiddenFuncError

MartinNowak:


schveiguy:
Auto-merge toggled on

PR #1250, opened 2015-05-03
add rt/profilegc.d

WalterBright:
Adds memory profiler implementation.

### DIFF
--- a/src/rt/profilegc.d
+++ b/src/rt/profilegc.d
@@ -1,0 +1,150 @@
+/*
+ * Data collection and report generation for
+ *   -profile=gc
+ * switch
+ *
+ * Copyright: Copyright Digital Mars 2015 - 2015.
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors:   Andrei Alexandrescu and Walter Bright
+ * Source: $(DRUNTIMESRC src/rt/_profilegc.d)
+ */
+
+module rt.profilegc;
+
+private:
+
+import core.stdc.stdio;
+import core.stdc.stdlib;
+import core.stdc.string;
+
+import core.exception : onOutOfMemoryError;
+
+char[] buffer;
+size_t[string] newCounts;
+
+__gshared
+{
+    size_t[string] globalNewCounts;
+    string logfilename = "profilegc.log";
+}
+
+/****
+ * Set file name for output.
+ * A file name of "" means write results to stdout.
+ * Params:
+ *      name = file name
+ */
+
+extern (C) void profilegc_setlogfilename(string name)
+{
+    logfilename = name;
+}
+
+
+
+public void accumulate(string file, uint line, string funcname, string type, size_t sz)
+{
+    char[3 * line.sizeof + 1] buf;
+    auto buflen = snprintf(buf.ptr, buf.length, "%u", line);
+
+    auto length = type.length + 1 + funcname.length + 1 + file.length + 1 + buflen;
+    if (length > buffer.length)
+    {
+        // Enlarge buffer[] so it is big enough
+        auto p = cast(char*)realloc(buffer.ptr, length);
+        if (!p)
+            onOutOfMemoryError();
+        buffer = p[0 .. length];
+    }
+
+    // "type funcname file:line"
+    buffer[0 .. type.length] = type[];
+    buffer[type.length] = ' ';
+    buffer[type.length + 1 ..
+           type.length + 1 + funcname.length] = funcname[];
+    buffer[type.length + 1 + funcname.length] = ' ';
+    buffer[type.length + 1 + funcname.length + 1 ..
+           type.length + 1 + funcname.length + 1 + file.length] = file[];
+    buffer[type.length + 1 + funcname.length + 1 + file.length] = ':';
+    buffer[type.length + 1 + funcname.length + 1 + file.length + 1 ..
+           type.length + 1 + funcname.length + 1 + file.length + 1 + buflen] = buf[0 .. buflen];
+
+    if (auto pcount = cast(string)buffer[0 .. length] in newCounts)
+        *pcount += sz;                          // existing entry
+    else
+        newCounts[buffer[0..length].idup] = sz; // new entry
+}
+
+// Merge thread local newCounts into globalNewCounts
+static ~this()
+{
+    if (newCounts.length)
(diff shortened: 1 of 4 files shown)


src/rt/profilegc.d
```diff
@@ -0,0 +1,151 @@
...
+    {
+        // Enlarge buffer[] so it is big enough
+        auto p = cast(char*)realloc(buffer.ptr, length);
+        if (!p)
+            onOutOfMemoryError();
+        buffer = p[0 .. length];
+    }
+
+    // "type funcname file:line"
+    memcpy(buffer.ptr, type.ptr, type.length);
```
andralex: can't you use slice operations here? `buffer[0 .. type.length] = type[];` etc

src/rt/profilegc.d
```diff
@@ -0,0 +1,151 @@
...
+    buffer[type.length] = ' ';
+    memcpy(buffer.ptr + type.length + 1, funcname.ptr, funcname.length);
+    buffer[type.length + 1 + funcname.length] = ' ';
+    memcpy(buffer.ptr + type.length + 1 + funcname.length + 1, file.ptr, file.length);
+    buffer[type.length + 1 + funcname.length + 1 + file.length] = ':';
+    memcpy(buffer.ptr + type.length + 1 + funcname.length + 1 + file.length + 1, buf.ptr, buflen);
+
+    auto pcount = cast(string)buffer[0 .. length] in newCounts;
+    if (pcount)
```
andralex: if (auto pcount = cast(string)buffer[0 .. length] in newCounts)

src/rt/profilegc.d
```diff
@@ -0,0 +1,151 @@
...
+        *pcount += sz;                          // existing entry
+    else
+        newCounts[buffer[0..length].idup] = sz; // new entry
+
+    //string name = cast(string)(type ~ " " ~ funcname ~ " " ~ file ~ ":" ~ buf[0 .. buflen]);
+    //newCounts[name] += sz;
```
andralex: remove commented code

src/rt/profilegc.d
```diff
@@ -0,0 +1,151 @@
...
+            {
+                // Merge
+                foreach (name, count; newCounts)
+                {
+                    globalNewCounts[name] += count;
+                }
+            }
+            else
+                // Assign
+                globalNewCounts = newCounts;
```
andralex: nice touch

src/rt/profilegc.d
```diff
@@ -0,0 +1,151 @@
...
+
+// Write report to stderr
+shared static ~this()
+{
+    static struct Result
+    {
+        string name;
+        size_t count;
+
+        // qsort() comparator to sort by count field
```
andralex: urgh, can't we move this to Phobos and take advantage of it?

src/rt/tracegc.d
```diff
@@ -13,7 +13,11 @@
 
 module rt.tracegc;
 
-import core.stdc.stdio;
+//version = print;
+
+import rt.profilegc;
+
+version (print) import core.stdc.stdio;
```
andralex: version tracegc?

src/rt/tracegc.d
```diff
@@ -323,51 +401,83 @@ extern (C) void[] _d_arrayappendwd(ref byte[] x, dchar c);
...
+            file.length, file.ptr,
+            line,
+            funcname.length, funcname.ptr
+            );
+    }
+    size_t n;
+    if (c <= 0xFFFF)
+        n = 1;
+    else
+        n = 2;
```
andralex: 5:1 compression: `size_t n = 1 + (c <= 0xFFFF);`

andralex:
got a few nits but basically this is where it's at

src/rt/profilegc.d
WalterBright: I thought about it, but it belongs in druntime.

yebblies: The hooks definitely belong in druntime, but it could do something like define an `interface` which phobos (or a third party) could implement.

rainers:
Wouldn't it be simpler to capture a part of the stack trace in gc.proxy? It helps in places where the traced call location is too generic to be useful (e.g. in some template), but you need to crawl up the stack to the actual user function.
That's also what the more sophisticated memory profilers do and does not need compiler support.

andralex:
@rainers yah a small stack trace (2-5 levels) would be more informative, but wouldn't that be pig slow? Anyhow this is a nice idea to do this work without changing the compiler (no flag no nothing) assuming the speed isn't killed. @WalterBright thoughts?

andralex:
Oh and btw: in my experience the caller site should be good enough to work with, but definitely distinguishing a couple more caller levels would improve things.

src/rt/profilegc.d
WalterBright: The reason that tracegc.d and profilegc.d are in separate files is so one can define one's own accumulate() function and thereby override the one in druntime.

WalterBright:
@rainiers Yes, this is not a sophisticated memory profiler. But it is as simple as it gets.

1. getting file/line/function info using debug info and stack walking is always complex and fragile code, and often has to be completely different implementations for each platform.
2. the emphasis on this feature is utter user simplicity - throw a compiler switch and get a report, in the same vein as -cov, -unittest, -D and -profile.
3. since it's a compiler switch, it is only turned on for specific selected modules rather than the whole program.
4. I put this together in a short period of time, and it should be pretty robust.
5. I've used memory profilers like this for decades, and these simple things turn out to be incredibly useful. Sophisticated ones are only a marginal improvement.

If someone wants to do a more sophisticated one, great, but it's unlikely to be me.

BTW, the -profile switch collects caller info. By looking at both -profile and -profilegc reports, you can combine the information.

andralex:
worksforme

andralex:
Auto-merge toggled on

rainers:
@WalterBright Thanks for explanation. IMO there are too many runtime callbacks already that should be library defined (e.g. associative and dynamic arrays), and this adds a lot more. Granted, the new ones don't add too much complexity and hopefully can be changed in lock step with the non-profiling ones.

> BTW, the -profile switch collects caller info. By looking at both -profile and -profilegc reports, you can combine the information.

I guess too much information is lost by the current reports to combine them sensibly, but the -profile option could be used to build the call stack without having to analyze debug information.

joakim-noah:
This pull litters a file `profilegc.log` any time the unit tests are run.  We need the auto-tester to check for that too, that no PR leaves files behind after the unit tests are run.

PR #1252, opened 2015-05-04
Allow a fiber in HOLD state to be reset bypassing scope(exit) etc.

Dzugaru:
Discussion: http://forum.dlang.org/thread/bltdzwskligncpgwdmxc@forum.dlang.org

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -4013,14 +4013,19 @@ class Fiber
 
     /**
      * Resets this fiber so that it may be re-used, optionally with a
-     * new function/delegate.  This routine may only be called for
+     * new function/delegate.  This routine should only be called for
      * fibers that have terminated, as doing otherwise could result in
      * scope-dependent functionality that is not executed.
      * Stack-based classes, for example, may not be cleaned up
-     * properly if a fiber is reset before it has terminated.
+     * properly if a fiber is reset before it has terminated. Calling it
+     * for a fiber in state HOLD also won't cleanup the old stack, 
+     * so it may leak resources which will lead to stack overflow if reset
+     * is called too many times for such a fiber. It will 
+     * properly reset the stack though, so the fiber should behave like 
+     * a new one.
      *
      * In:
-     *  This fiber must be in state TERM.
+     *  This fiber must be in state TERM or HOLD.
      */
     final void reset() nothrow
     in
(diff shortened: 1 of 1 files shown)


src/core/thread.d
```diff
@@ -4013,14 +4013,19 @@ class Fiber
...
      * fibers that have terminated, as doing otherwise could result in
      * scope-dependent functionality that is not executed.
      * Stack-based classes, for example, may not be cleaned up
-     * properly if a fiber is reset before it has terminated.
+     * properly if a fiber is reset before it has terminated. Calling it
+     * for a fiber in state HOLD also won't cleanup the old stack, 
+     * so it may leak resources which will lead to stack overflow if reset
+     * is called too many times for such a fiber. It will 
+     * properly reset the stack though, so the fiber should behave like 
+     * a new one.
```
MartinNowak: That's actually much more confusing than before.

One should only reset a fiber in term state, because resetting it in hold state won't cleanup the stack, e.g. scope(exit) and struct destructors aren't called.

MartinNowak:
Looking at the implementation again, there is indeed a bug.
The `assert( m_ctxt.tstack == m_ctxt.bstack );` in reset will only work for terminated fibers not for hold ones. We'd need to explicitly set `m_ctxt.tstack = m_ctxt.bstack` for those.

Dzugaru:
Now I get it, reset without parameters works for HOLD state iff the stack is clean. Reset with parameters (new function/delegate) works with any state cause there is no check there and leads to stack overflow.

> We'd need to explicitly set m_ctxt.tstack = m_ctxt.bstack for those.

That would be awesome, one could cope with scope(exit) and struct destructors not executing but having a way to completely reset HOLD fiber is very nice.

MartinNowak:
> We'd need to explicitly set m_ctxt.tstack = m_ctxt.bstack for those.

Can you change the pull to just do that and add a test case.
```d
unittest
{
    // test unsafe reset in hold state
    auto fib = new Fiber({ubyte[2048] buf = void; Fiber.yield();}, 4096);
    foreach (_; 0 .. 10)
    {
        fib.call();
        assert(fib.state == Fiber.State.HOLD);
        fib.reset();
    }
}
```

Dzugaru:
> Can you change the pull to just do that and add a test case.

Did that, rebuilt everything with Digger on win32 and that test case works.

MartinNowak:
Auto-merge toggled on

MartinNowak:
Thx

PR #1253, opened 2015-05-04
Add missing definitions to core.sys.posix.sys.msg

redstar:
MIPS definition was incomplete and MIPS64 and PPC64 missing.


MartinNowak:
Auto-merge toggled on

PR #1254, opened 2015-05-05
fix aabench/stomper for 32-bit

MartinNowak:


WalterBright:
Auto-merge toggled on

PR #1255, opened 2015-05-05
Seperate core.sys.signal architecture support into separate files

ibuclaw:
This implements https://github.com/D-Programming-Language/druntime/pull/744  but using a more complex module.

I'm still not convinced this is the right way to go about it...

@WalterBright feel free to review.


MartinNowak:
So is this really an improvement?
It's already a huge PITA to keep the different [sections.d](https://github.com/D-Programming-Language/druntime/blob/master/src/rt/sections.d) implementations aligned, and I ended up remerging the more complicated [sections_elf_shared](https://github.com/D-Programming-Language/druntime/blob/master/src/rt/sections_elf_shared.d) for FreeBSD and Linux.

ibuclaw:
> So is this really an improvement?

I'm not sold on this method, but people also weren't sold on my proposal either.  What I am certain about is that this is something that can't be continually ignored.

> It's already a huge PITA to keep the different sections.d implementations aligned,

Would it be a bigger PITA if they were all bunched together in one file?

I envisioned that we'd hit some sort of brick wall on the burden of adding ports and general maintenance hell a couple of years back.  Maybe it's not as serpentine as I made it out, but a couple of questions I'd pose are:

1. When porting to a new architecture, how easy was it to find and correct problems at runtime when crossing the D <-> C boundaries?

We are in a better position now than what we were two years back, but there are still places where relaxed version conditions may allow things to fall through.

2. Would the port to Android x86/ARM have been easier if platforms/targets were split up?

This is something I can leave you to ponder over.

MartinNowak:
> Would it be a bigger PITA if they were all bunched together in one file?

It's ugly, but it doesn't hurt much, except for lexing/parsing overhead maybe, which hasn't been a huge problem so far.

> We are in a better position now than what we were two years back, but there are still places where relaxed version conditions may allow things to fall through.

This is something that can be fixed without changing the file layout.

> Would the port to Android x86/ARM have been easier if platforms/targets were split up?

Apart from a few unterminated version conditions, I'd say no.

Geod24:

PR #1257, opened 2015-05-05
lazily init _ticksPerSecond

MartinNowak:
- remove shared static this()
- construction of MonoTime is now impure
  (obtaining the current time is impure anyhow)
- any construction will init _ticksPerSecond
- use pureTicksPerSecond and assert a previous init
  to maintain pure for MonoTime ops
- make ticksToNSecs/nsecsToTicks impure because the
  might have to obtain the clock resolution

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -2163,6 +2163,12 @@ struct MonoTimeImpl(ClockType clockType)
 
 @safe:
 
+    this(long ticks) inout
+    {
+        ticksPerSecond; // init ticks per second
+        _ticks = ticks;
+    }
+
     version(Windows)
     {
         static if(clockType != ClockType.coarse &&
@@ -2242,12 +2248,6 @@ struct MonoTimeImpl(ClockType clockType)
       +/
     static @property MonoTimeImpl currTime() @trusted nothrow @nogc
     {
-        if(ticksPerSecond == 0)
-        {
-            assert(0, "MonoTimeImpl!(ClockType." ~ __traits(allMembers, ClockType)[_clockIdx] ~
-                      ") failed to get the frequency of the system's monotonic clock.");
-        }
-
         version(Windows)
         {
             long ticks;
@@ -2266,14 +2266,15 @@ struct MonoTimeImpl(ClockType clockType)
             if(clock_gettime(clockArg, &ts) != 0)
                 assert(0, "Call to clock_gettime failed.");
 
-            return MonoTimeImpl(convClockFreq(ts.tv_sec * 1_000_000_000L + ts.tv_nsec,
-                                              1_000_000_000L,
-                                              ticksPerSecond));
+            MonoTimeImpl res = void;
+            res._ticks = convClockFreq(ts.tv_sec * 1_000_000_000L + ts.tv_nsec,
+                                       1_000_000_000L, ticksPerSecond);
+            return res;
         }
     }
 
 
-    static @property pure nothrow @nogc
+    static @property nothrow @nogc
     {
     /++
         A $(D MonoTime) of $(D 0) ticks. It's provided to be consistent with
@@ -2406,7 +2407,7 @@ assert(before + timeElapsed == after).
         if(op == "-")
     {
         immutable diff = _ticks - rhs._ticks;
-        return Duration(convClockFreq(diff , ticksPerSecond, hnsecsPer!"seconds"));
+        return Duration(convClockFreq(diff, pureTicksPerSecond, hnsecsPer!"seconds"));
     }
 
     unittest
@@ -2461,8 +2462,10 @@ assert(before + timeElapsed == after).
     MonoTimeImpl opBinary(string op)(Duration rhs) const pure nothrow @nogc
         if(op == "+" || op == "-")
     {
-        immutable rhsConverted = convClockFreq(rhs._hnsecs, hnsecsPer!"seconds", ticksPerSecond);
-        mixin("return MonoTimeImpl(_ticks " ~ op ~ " rhsConverted);");
+        immutable rhsConverted = convClockFreq(rhs._hnsecs, hnsecsPer!"seconds", pureTicksPerSecond);
+        MonoTimeImpl res;
+        res._ticks = mixin("_ticks " ~ op ~ " rhsConverted");
+        return res;
     }
 
     unittest
@@ -2507,7 +2510,7 @@ assert(before + timeElapsed == after).
     ref MonoTimeImpl opOpAssign(string op)(Duration rhs) pure nothrow @nogc
         if(op == "+" || op == "-")
     {
-        immutable rhsConverted = convClockFreq(rhs._hnsecs, hnsecsPer!"seconds", ticksPerSecond);
+        immutable rhsConverted = convClockFreq(rhs._hnsecs, hnsecsPer!"seconds", pureTicksPerSecond);
         mixin("_ticks " ~ op ~ "= rhsConverted;");
         return this;
     }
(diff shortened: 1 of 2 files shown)


MartinNowak:
As discussed here https://github.com/D-Programming-Language/druntime/pull/990#discussion_r29628877.

MartinNowak:
Purity hardly matters for MonoTime because getting the current time is impure and it's the only use-case for MonoTime.

jmdavis:
These changes trigger https://issues.dlang.org/show_bug.cgi?id=14517 . Also, isn't this less efficient than the current solution, since it has to load `_ticksPerSecond` rather than just using it? It _is_ shorter though, which is nice. Regardless, it's not going to work if it's triggering https://issues.dlang.org/show_bug.cgi?id=14517.

An alternative solution would be to use `version(CoreDdoc)` to version out the stuff that's causing problems. Worst case, all of `MonoTimeImpl` could be versioned with `CoreDdoc`.

src/core/time.d
```diff
@@ -2428,8 +2428,10 @@ assert(before + timeElapsed == after).
     MonoTimeImpl opBinary(string op)(Duration rhs) const pure nothrow @nogc
         if(op == "+" || op == "-")
     {
-        immutable rhsConverted = convClockFreq(rhs._hnsecs, hnsecsPer!"seconds", ticksPerSecond);
-        mixin("return MonoTimeImpl(_ticks " ~ op ~ " rhsConverted);");
+        immutable rhsConverted = convClockFreq(rhs._hnsecs, hnsecsPer!"seconds", pureTicksPerSecond);
+        MonoTimeImpl res;
```
jmdavis: You might as well make this `= void;` since you assign to it right after.

src/core/time.d
```diff
@@ -2627,103 +2632,82 @@ assert(before + timeElapsed == after).
...
-        long ticksPerSecond;
-        if(QueryPerformanceFrequency(&ticksPerSecond) != 0)
+        version(Windows)
         {
-            foreach(i, typeStr; __traits(allMembers, ClockType))
-                _ticksPerSecond[i] = ticksPerSecond;
+            long ticksPerSecond = void;
+            if (QueryPerformanceFrequency(&ticksPerSecond) != 0)
+                return ticksPerSecond;
+            else
```
jmdavis: Now that this code is returning rather than just assigning, this `else` is unnecessary.

src/core/time.d
```diff
@@ -2627,103 +2632,82 @@ assert(before + timeElapsed == after).
...
-        if(mach_timebase_info(&info) == 0)
+        else version(OSX)
         {
-            long ticksPerSecond = 1_000_000_000L * info.numer / info.denom;
-            foreach(i, typeStr; __traits(allMembers, ClockType))
-                _ticksPerSecond[i] = ticksPerSecond;
+            mach_timebase_info_data_t info = void;
+            if (mach_timebase_info(&info) == 0)
+                return 1_000_000_000L * info.numer / info.denom;
+            else
```
jmdavis: This `else` is also unnecessary.

src/core/time.d
```diff
@@ -2627,103 +2632,82 @@ assert(before + timeElapsed == after).
...
-    else version(Posix)
-    {
-        timespec ts;
-        foreach(i, typeStr; __traits(allMembers, ClockType))
+        else version(Posix)
         {
-            mixin("enum clockType = ClockType." ~ typeStr ~ ";");
-            static if(clockType != ClockType.second)
+            clockid_t clkid = void;
+            version (linux)
```
jmdavis: This whole section with `clkid` and the switch statements is unnecessary. `clkid` isn't even used, and because this code is now inside of `MonoTimeImpl` instead of outside of it, it can just use the `clockArg` which is declared at the beginning of `MonoTimeImpl` - and that's what it's currently doing, making this whole section with `clkid` redundant.

src/core/time.d
```diff
@@ -2627,103 +2632,82 @@ assert(before + timeElapsed == after).
...
+            timespec ts;
+            if (clock_getres(clockArg, &ts) == 0)
+                // For some reason, on some systems, clock_getres returns
+                // a resolution which is clearly wrong (it's a millisecond
+                // or worse, but the time is updated much more frequently
+                // than that). In such cases, we'll just use nanosecond
+                // resolution.
+                return ts.tv_nsec >= 1000 ? 1_000_000_000L
+                    : 1_000_000_000L / ts.tv_nsec;
+            else
```
jmdavis: Unnecessary `else`, and the `if` really should have braces with this comment inside of it. As it stands, the code is quite a ways from the `if` condition, making this harder to read without braces.

jmdavis:
This PR solves the problem with far fewer changes https://github.com/D-Programming-Language/druntime/pull/1259 and it won't affect the non-documentation build at all, so it won't trigger that compiler bug.

MartinNowak:
> These changes trigger https://issues.dlang.org/show_bug.cgi?id=14517.

This pull removes the static constructors, so it shouldn't trigger the issue. So have to see what's causing the issue on OSX.

> Also, isn't this less efficient than the current solution, since it has to load _ticksPerSecond rather than just using it?

That's completely irrelevant here. A raw atomic load (no synchronization required) is really cheap, especially compared to clock_gettime. It's not an intrinsic on DMD yet (but on GDC and LDC), so there is a small call overhead, still it's cheap and only required for construction, pureTicksPerSecond just normally loads the value because it is used after the value is already initialized.

jmdavis:
> That's completely irrelevant here. A raw atomic load (no synchronization required) is really cheap, especially compared to clock_gettime.

I don't see how it's irrelevant. With the current solution, once the static constructor has been run, `_ticksPerSecond` has been initialized and can just be returned, whereas with this, it has to check whether it's been initialized every time it fetches it. So, while an atomic load may be fast, it's still extra work that has to be done every time that a `MonoTimeImpl` is constructed.

Aside from fixing the documentation build problem (which has now been fixed with another pull), why is what's in this PR better than what we have now?

MartinNowak:
> Aside from fixing the documentation build problem (which has now been fixed with another pull), why is what's in this PR better than what we have now?

Because it avoids a static constructor that runs for everyone using core.time, makes MonoTime usable without init order issues in the GC, and allows us to get rid of a bunch of metaprogramming mixins.

> I don't see how it's irrelevant. With the current solution, once the static constructor has been run, _ticksPerSecond has been initialized and can just be returned, whereas with this, it has to check whether it's been initialized every time it fetches it.

It's irrelevant, because you won't be able to measure a difference.

MartinNowak:
Another reason against static constructors is that they tend to bloat statically linked binaries.

schveiguy:
I'm against lazy construction because of the performance cost (it is not zero) and the lack of purity (yes, you cannot obtain a MonoTime purely, but at that point, it becomes POD, and can be used purely). I contend that the binary bloat is not significant.

Really, the point of static constructors is to avoid having lazy construction (i.e. singleton). The cost of this construction is so miniscule.

The GC problem of relying on static ctors is solved.

And finally, if the static construction in templates issue is fixed, we can revert this back to the "simple" mechanism Jonathan originally used.

MartinNowak:
> And finally, if the static construction in templates issue is fixed, we can revert this back to the "simple" mechanism Jonathan originally used.

My main motivation for this is using MonoTime within druntime, it's a very valuable tool and the lack of it makes it harder to benchmark something.

> I'm against lazy construction because of the performance cost (it is not zero) and the lack of purity

There is a small cost because atomicLoad isn't an intrinsic.

> but at that point, it becomes POD, and can be used purely). I contend that the binary bloat is not significant.

I used a `pureTicksPerSecond` function that just loads the value, so only construction is impure. I understand that this looks a bit hacky, but it's better than to initialize all _ticksPerSeconds for all clock types using a function with string mixins. We could also get rid of that clockIdx.

schveiguy:
> My main motivation for this is using MonoTime within druntime, it's a very valuable tool and the lack of it makes it harder to benchmark something.

An option is to have druntime specifically run the init instead of using static ctor. I can't believe I didn't think of that before :)

You'll have to cast away immutability, but that should be fine here.

does that solve all the problems?

schveiguy:
I'll make a PR for that, hopefully it solves this.

MartinNowak:
OK, thanks

MartinNowak:
You might want to look at #1276 first to avoid merge conflicts.

schveiguy:
Not sure if I can avoid merge conflicts there... Almost done with the PR, so we will see, I'll try a merge to see if it fails.

schveiguy:
Took a closer look, I'm definitely going to conflict, as I factored out the static if code for translating clock type to clock arg, it's done in 2 places, and the code is identical. I'll update mine to use the new form you used.

schveiguy:
I'm including all the updates you made to that PR except for the unit test reduction. I think that should be a separate refactoring, since @jmdavis loves his unit tests ;)

It will probably still conflict, but be much less of an update for you.

#1278

PR #1258, opened 2015-05-06
posix.mak: convert CRLF back to LF

rainers:
https://github.com/D-Programming-Language/druntime/pull/1222 inadvertedly translated line endings in posix.mak. This reverts this change.


andralex:
thx!

WalterBright:
Use the tool:
https://github.com/D-Programming-Language/tools/blob/master/tolf.d
as part of your checking switch, and no more problems.

PR #1259, opened 2015-05-08
Fix bug# 14546 - Make the druntime documentation buildable again.

jmdavis:
This makes it so that ClockType isn't actually used in core.time in the
documentation build (since if it is, it triggers static assertions,
since none of the systems support all of the ClockTypes, but they all
have to be defined for the documentation build in order to show up in
the docs).

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -2648,7 +2648,14 @@ private immutable long[__traits(allMembers, ClockType).length] _ticksPerSecond;
 
 @trusted shared static this()
 {
-    version(Windows)
+    // If we try to do anything with ClockType in the documentation build, it'll
+    // trigger the static assertions related to ClockType, since the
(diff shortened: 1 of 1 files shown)


src/core/time.d
```diff
@@ -2648,7 +2648,14 @@ private immutable long[__traits(allMembers, ClockType).length] _ticksPerSecond;
 
 @trusted shared static this()
 {
-    version(Windows)
+    // If we try and doing anything with ClockType in the documentation build,
```
CyberShadow: > try and doing

CyberShadow:
Thanks.

src/core/time.d
jmdavis: Ouch. Will fix.

jmdavis: Fixed.

MartinNowak:
Auto-merge toggled on

PR #1261, opened 2015-05-10
fix Issue 14565 - dmd -profile produces garbled output for long-runni…

WalterBright:
…ng CPU-intensive processes

https://issues.dlang.org/show_bug.cgi?id=14565

Increased the counter from 32 bits to 64 bits.

### DIFF
--- a/src/rt/trace.d
+++ b/src/rt/trace.d
@@ -35,7 +35,7 @@ struct SymPair
 {
     SymPair* next;
     Symbol* sym;        // function that is called
-    uint count;         // number of times sym is called
+    ulong count;        // number of times sym is called
 }
 
 /////////////////////////////////////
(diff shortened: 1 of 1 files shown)


src/rt/trace.d
```diff
@@ -35,7 +35,7 @@ struct SymPair
 {
     SymPair* next;
     Symbol* sym;        // function that is called
-    uint count;         // number of times sym is called
+    long count;         // number of times sym is called
```
yebblies: Should be `ulong`?

WalterBright: yes. Oops!

src/rt/trace.d
```diff
@@ -163,18 +163,23 @@ private int sympair_cmp(in void* e1, in void* e2)
     auto psp1 = cast(SymPair**)e1;
     auto psp2 = cast(SymPair**)e2;
 
-    return (*psp2).count - (*psp1).count;
+    long cmp = (*psp2).count - (*psp1).count;
```
yebblies: Doesn't this pattern produce the wrong results in some cases?  Isn't it better to just use `if (a < b) ...` and let the compiler optimize it?

MartinNowak:
Auto-merge toggled on

PR #1263, opened 2015-05-11
Fix core.demangle tests

ibuclaw:
When making some changes to my own demangler in GDB.  I noticed these started failing.

I compiled some tests to closely match what the demangled symbol is, and I'm pretty sure these are all wrong.  Or, if they were correct at one stage, this is no longer the case in DMD.

The culprit here is the nested ``FZv``.  DMD only emits the type for the symbol, and not the type of any of it's parents.


ibuclaw:
This symbol starting failing too:
```
_D8serenity9persister6Sqlite70__T15SqlitePersisterTS8serenity9persister6Sqlite11__unittest6FZv4TestZ15SqlitePersister12__T7opIndexZ7opIndexMFmZS8serenity9persister6Sqlite11__unittest6FZv4Test
```
But this will take longer to write a proper test for...

MartinNowak:
Auto-merge toggled on

PR #1266, opened 2015-05-12
try to debug auto-tester deadlock in core.sync.semaphore

MartinNowak:


MartinNowak:
Problem found, see #1274.

PR #1010, opened 2014-11-04
Separate linux kernel and glibc APIs as much as possible

joakim-noah:
This pull uses the new `CRuntime_Glibc` and `CRuntime_Bionic` predefined global versions from [dmd pulls #4111](https://github.com/D-Programming-Language/dmd/pull/4111) and [#3643](https://github.com/D-Programming-Language/dmd/pull/3643) to start separating out linux kernel APIs from glibc and bionic APIs.  I used dstep and a modified Dscanner to help find type and struct declarations, but had to check enums again by hand.  dmd pull #4111 will have to be merged before this one.

PR #764, opened 2014-04-14
WIP: Better Android support

mathstuf:
Currently a work-in-progress, but feedback is still nice. The current "fix the world" commit will be broken into smaller, logical chunks (first, hoisting the `Android` checks, then adding ARM and MIPS support, then fixing any bugs).

Basically, all `version(Android)` checks need to be hoisted above `version(linux)` checks since GDC's `Android` implies `linux`. Also adding support for non-x86 platforms.

Maybe adding a `Linux` or `GNULinux` version would be better which is only defined if `linux && !Android`?

PR #1267, opened 2015-05-13
Add Android/ARM support

joakim-noah:
When I first added Android support, it was based on the x86 headers of Android API 9, with any definitions coming from architecture-dependent headers placed inside a `version(X86)` block, just in case.  For this pull, I checked each of those blocks for the three 32-bit platforms from Android API 9, `X86`, `ARM`, and `MIPS32`.  If they're all the same, the versioning was removed.  If any arch varied, the definition was put inside its own `version(ARM)` block.

I did not check the new 64-bit platforms added with the latest Android API 21.  This pull overlaps and obsoletes some of #764, which has not been updated in a while.

I need to add one last bit to `core.stdc.errno`, but cannot do so till a change to that module in my previous #1010 is merged first.  Once that's merged, I can add the last necessary bit for this pull to be done.

### DIFF
--- a/src/core/stdc/fenv.d
+++ b/src/core/stdc/fenv.d
@@ -192,6 +192,11 @@ else version( Android )
 
         alias ushort fexcept_t;
     }
+    else version(ARM)
+    {
+        alias uint fenv_t;
+        alias uint fexcept_t;
+    }
     else
     {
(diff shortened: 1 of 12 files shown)


MartinNowak:
Thanks a lot.

PR #1270, opened 2015-05-14
Tweak new AA implementation to get similar performance as the old one on Win32

rainers:
This PR tries to restore performance of the AA after changing it to "state-of-the-art" dense hash map.

Results of AA related benchmarks on my mobile i7:
```
Win32:
                   oldAA     newAA     newAA_tweaked
R bulk             0.293 s   0.328 s   0.299 s
R int              0.106 s   0.117 s   0.115 s
R obj              0.129 s   0.128 s   0.127 s
R ptr              0.110 s   0.112 s   0.106 s
R resize           0.393 s   0.420 s   0.401 s
R stomper          1.026 s   1.007 s   1.013 s
R string           0.149 s   0.142 s   0.142 s
R testgc3          1.825 s   2.201 s   1.919 s
```

Win64 benefits from these optimizations aswell:
```
Win64:
                   oldAA     newAA     newAA_tweaked
R bulk             0.444 s   0.388 s   0.365 s
R int              0.139 s   0.116 s   0.113 s
R obj              0.164 s   0.144 s   0.142 s
R ptr              0.124 s   0.117 s   0.112 s
R resize           0.457 s   0.419 s   0.395 s
R stomper          0.410 s   0.389 s   0.386 s
R string           0.115 s   0.103 s   0.099 s
R testgc3          2.632 s   2.512 s   2.420 s
```

testgc3 gets slightly faster than the old implementation without the new `mix` function.

### DIFF
--- a/mak/MANIFEST
+++ b/mak/MANIFEST
@@ -243,6 +243,7 @@ MANIFEST=\
 	\
 	src\rt\util\array.d \
 	src\rt\util\hash.d \
+	src\rt\util\mem.d \
 	src\rt\util\random.d \
 	src\rt\util\string.d \
 	src\rt\util\typeinfo.d \
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -129,6 +129,7 @@ SRCS=\
 	\
 	src\rt\util\array.d \
 	src\rt\util\hash.d \
+	src\rt\util\mem.d \
 	src\rt\util\random.d \
 	src\rt\util\string.d \
 	src\rt\util\typeinfo.d \
--- a/src/rt/aaA.d
+++ b/src/rt/aaA.d
@@ -55,7 +55,7 @@ private:
         buckets = allocBuckets(sz);
         firstUsed = cast(uint) buckets.length;
         entryTI = fakeEntryTI(ti.key, ti.value);
-        valoff = talign(keysz, ti.value.talign);
+        valoff = cast(uint) talign(keysz, ti.value.talign);
 
         import rt.lifetime : hasPostblit, unqualify;
 
@@ -72,7 +72,7 @@ private:
     uint firstUsed;
     immutable uint keysz;
     immutable uint valsz;
-    uint valoff;
+    immutable uint valoff;
     Flags flags;
 
     enum Flags : ubyte
@@ -174,10 +174,9 @@ private pure nothrow @nogc:
         return hash == HASH_DELETED;
     }
 
-    // not returning bool to avoid gratuitious codegen
-    @property int filled() const
+    @property bool filled() const
     {
-        return hash & HASH_FILLED_MARK;
+        return cast(ptrdiff_t)hash < 0;
     }
 }
 
--- a/src/rt/aaA.d
+++ b/src/rt/aaA.d
@@ -197,6 +197,7 @@ Bucket[] allocBuckets(size_t dim) @trusted pure nothrow
 private void* allocEntry(in Impl* aa, in void* pkey)
 {
     import rt.lifetime : _d_newitemU;
+    import rt.util.mem;
 
     immutable akeysz = aa.keysz + aa.keypad;
     void* res = void;
@@ -208,9 +209,8 @@ private void* allocEntry(in Impl* aa, in void* pkey)
         res = GC.malloc(akeysz + aa.valsz, flags);
     }
 
-    res[0 .. aa.keysz] = pkey[0 .. aa.keysz]; // copy key
-    import core.stdc.string : memset; // explicitly use memset (Issue 14458)
-    memset(res + akeysz, 0, aa.valsz); // zero value
+    fastshortcopy(res, pkey, aa.keysz); // copy key
+    fastshortclear(res + akeysz, aa.valsz); // zero value
 
     return res;
 }
--- a/src/rt/aaA.d
+++ b/src/rt/aaA.d
@@ -103,7 +103,7 @@ private:
     {
         for (size_t i = hash & mask, j = 1;; ++j)
         {
-            if (buckets[i].empty || buckets[i].deleted)
+            if (!buckets[i].filled)
                 return &buckets[i];
             i = (i + j) & mask;
         }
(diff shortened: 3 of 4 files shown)


MartinNowak:
Thanks, that's a lot of useful stuff. It's a pity that you redid the effort to split the arrays, but that's a nogo given the cost of a cache miss.

MartinNowak:
Here is a pretty simple benchmark that becomes 30% slower with this patch, my CPU has 6MB L3 cache.

```d
void main(string[] args)
{
    int[int] aa;
    foreach (_; 0 .. 10)
        foreach (i; 0 .. 1_000_000) // 16MB for bucket array
            ++aa[i];
}
```

rainers:
> Here is a pretty simple benchmark that is 30% slower with this patch, my CPU has 6MB of L3 cache.

Thanks. My CPU is similar (i7-2670QM), but shows only a 10% slowdown for Win64. On Win32 I had to double the size to get simliar results. Should we add this test to the benchmark suite?

The more common use case is to have smaller arrays, though. I concede that in a program that does different stuff in between AA operations, a double cache miss is more likely to happen.

MartinNowak:
> Should we add this test to the benchmark suite?

Tests that run a few seconds are somewhat unproductive. Actually the stomper test was supposed to show this, but it's magnitude is much smaller as >50% of the time is spend polluting the cache.

> The more common use case is to have smaller arrays, though. I concede that in a program that does different stuff in between AA operations, a double cache miss is more likely to happen.

It's more common, but both are common enough to matter. Can we split that off into a second pull request?

MartinNowak:
We could ditch the 32-bit hash value and only save 3 extra bits (above the mask) in the lower pointer bits.
That'll work with the 16-byte alignment of the GC and should still detect most collisions before comparing the key, but it means recomputing the hash on growth.

rainers:
> Thanks, that's a lot of useful stuff. It's a pity that you redid the effort to split the arrays, but that's a nogo given the cost of a cache miss.

I've rebased without the split arrays. The improvements are pretty small without it, though.

rainers:
> We could ditch the 32-bit hash value and only save 3 extra bits (above the mask) in the lower pointer bits.
> That'll work with the 16-byte alignment of the GC and should still detect most collisions before comparing the key, but it means recomputing the hash on growth.

Interesting idea (there are 4 unused lower bits in the entry pointer, and 16 upper bits on 64-bit).

rainers:
> Tests that run a few seconds are somewhat unproductive.

If I reduce the iteration count to 4, but for 2_000_000 entries, it runs for 1.5s on my (slower) system. That's less than other benchmarks.

Apropos large number of entries: the new AA limits the number of entries to 2^^32 even on 64-bit. As the Impl struct is already slightly larger than 32 bytes, I think there is no real gain from that limitation and we should remove it.

MartinNowak:
> Apropos large number of entries: the new AA limits the number of entries to 2^^32 even on 64-bit. As the Impl struct is already slightly larger than 32 bytes, I think there is no real gain from that limitation and we should remove it.

Makes sense, as long as the hot stuff fits into a single cacheline. I packed everything when there was still an inline init array.

MartinNowak:
I ran dfmt on the change and replaced CRLF with LF in rt.util.mem.

MartinNowak:
Yay, another bad decision sponsored by dmd's outdated backend.

MartinNowak:
> The memcpy/… changes are probably a slight regression

The length is a runtime value, so that isn't entirely true.

ibuclaw:
> The length is a runtime value, so that isn't entirely true.

I concur with David.  This will result in a slight regression for GDC (no cross-module inlining just yet).

ibuclaw:
Also, memset is recognised as a builtin in GDC, and so already would inline as outlined in this hack at no additional cost!

@MartinNowak - Revert?

rainers:
> Also, mempy/memset is recognised as a builtin in GDC, and so already would inline as outlined in this hack at no additional cost!

I tried both LDC and GDC for Win64: I don't get any inlining for memset/memcpy. What's the supposed code? If it is `rep stosb` or similar it has horrible setup time for small memory blocks. See also https://issues.dlang.org/show_bug.cgi?id=14458

Maybe we should version the code for DMD and other compilers?

ibuclaw:
If you are concerned, maybe you could inline it?  newitemT used to do this:  https://github.com/D-Programming-Language/druntime/commit/79bc91b41334c1805f557ef2f1606de31c6764d1#diff-50840eb79e3292d148af41006e3af980L1115

I see no problem with libc as it usually provides SSE2/3/4 optimized functions, so should have the same overhead as the D arrayops library.

rainers:
> I see no problem with libc as it usually provides SSE2/3/4 optimized functions, so should have the same overhead as the D arrayops library.

Most of the optimizations in memset/memcpy are geared towards large memory chunks, while keys are likely to be just one or two words (maybe also values). It's too bad GDC and LDC don't do cross module inlining, so they don't gain little by the "inlinable" copy and set operations. Maybe we should make them private helpers inside aaa.d, and fall back to memset/memcpy for sizes > 16.

rainers:
> I ran dfmt on the change and replaced CRLF with LF in rt.util.mem.

Sorry for that. I'll try to integrate checkwhitespace into my workflow... 

Is dfmt the defacto standard style now? I don't see many modules using a space after `cast()`.

yebblies:
> Sorry for that. I'll try to integrate checkwhitespace into my workflow...

It's now being run automatically on the autotester.

> Is dfmt the defacto standard style now?

No.

rainers:
> It's now being run automatically on the autotester.

I don't see any target in the druntime/phobos makefiles. I think it is only checked for dmd so far, isn't it?

yebblies:
Yeah, you're right.  I thought this was a dmd pull request.

MartinNowak:
> Is dfmt the defacto standard style now?

No, but it saves me time b/c I no longer fiddle with formatting.

PR #1272, opened 2015-05-14
fix Issue 14573 - classes without destructor leak monitor/mutex

MartinNowak:
- set finalize bit when constructing the monitor
- fix race condition (incorrect double checked locking)
- optimistically init mutex before locking global mutex
  to reduce contention
- remove pointless code duplication of Windows/Posix implementation
- move implementation details into private D functions
- use shared(Monitor)* whenever dealing with an unsynchronized instance
  and Monitor* when it's already externally synchronized
- modernize code for better readability
- drop using STI/STD style C constructors (called by rt_init/rt_term)

[Issue 14573 – [REG2.067] Extreme memory usage when `synchronized( object )` is used](https://issues.dlang.org/show_bug.cgi?id=14573)

### DIFF
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -37,8 +37,8 @@ version (FreeBSD)
     import core.stdc.fenv;
 }
 
-extern (C) void _STI_monitor_staticctor();
-extern (C) void _STD_monitor_staticdtor();
+extern (C) void _d_monitor_staticctor();
+extern (C) void _d_monitor_staticdtor();
 extern (C) void _STI_critical_init();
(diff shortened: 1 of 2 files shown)


schveiguy:
Seems like a good idea, I like it. LGTM, though I'd like to avoid the double-set of the flag if the monitor happens to be set when calling the function.

schveiguy:
Are those the only lines with CRLF? seems weird...

MartinNowak:
Sorry for the huge rewrite, but there was too much wrong with this module.

MartinNowak:
Let's just merge both in #1274.

PR #1273, opened 2015-05-14
fix unittest failures in debug build

MartinNowak:


MartinNowak:
trivial

schveiguy:
I'm not really understanding why the unit test was added, looked in history, and you added it when the new AA implementation was done. I also don't understand why this passes in release mode and fails in debug mode.

So I'm inclined to just accept this PR on that basis :)

Can you give some explanation behind the adding of the test?

I fully agree with the `numToString` change, in fact, I think the original is 2 elements too short, should have been a +1 instead of a -1.

rainers:
I agree with the numToString change, though that function is still wrong for values < -1: it will print duplicate '-'. Plus, it doesn't do proper rounding, though that's probably not so important if it is just used in core.time's unittest.

I'm not so sure about removing the AA unittest: it fails with a range violation (that is not checked in release), because the allocated entry has zero length, so the respective allocation returns null. This looks like a number of operations fail (also `E.init in aa`), though in fact, they did not. Mapping a type with just a single state and zero storage to another similar type doesn't sound very useful, but might happen in some generic code. I think we should at least allocate a single byte for the (key,value) entry to achieve consistent semantics for this edge case, too.

schveiguy:
> I think we should at least allocate a single byte for the (key,value) entry to achieve consistent semantics for this edge case, too

I agree with this.

schveiguy:
Also, why are we not testing in debug mode on druntime, I thought that was happening? It at least happens on phobos.

MartinNowak:
The problem was that zero size keys wouldn't compare equal.

rainers:
I stepped through the code in the debugger and aa[E.init] returned from _aaGetY at https://github.com/MartinNowak/druntime/blob/3af384a89d250e4f04e36068ff65aa4d7efcf9f6/src/rt/aaA.d#L362 with p.entry being null. TypeInfo_StaticArray.equals always returns true for me with 0 elements.

schveiguy:
@rainers I think Martin was talking about why he added the test in the first place.

In any case, I think an AA of this nature is quite useless. I can't see a reason to have an AA with only one possible node, there are better ways to do something like that.

schveiguy:
I'm OK to merge this if you are, @rainers

MartinNowak:
Wait, I'll debug this and see where the range error came from. I added this test originally to maximize coverage.

MartinNowak:
updated

rainers:
LGTM

rainers:
Auto-merge toggled on

PR #1274, opened 2015-05-14
cleanup and rewrite rt.critical_

MartinNowak:
- reuse mutex declarations from rt.monitor_
- fix race condition (incorrect double checked locking)
- modernize code for readability
- use shared(D_CRITICAL_SECTION)* when dealing with
  unsynchronized data
- replace STI/STD style C constructors
- fix Issue 6607
- hopefully fix Issue 14584 - spurious autotester deadlocks

[Issue 14584 – spurious autotester deadlocks](https://issues.dlang.org/show_bug.cgi?id=14584)
DEPENDS ON: #1272

### DIFF
--- a/src/rt/monitor_.d
+++ b/src/rt/monitor_.d
@@ -328,6 +328,10 @@ version( Windows )
         LeaveCriticalSection(&_monitor_critsec);
         if (cs)
             free(cs);
+        // Set the finalize bit so that the monitor gets collected (Bugzilla 14573)
+        import core.memory : GC;
+        if (!(typeid(h).m_flags & TypeInfo_Class.ClassFlags.hasDtor))
+            GC.setAttr(cast(void*)h, GC.BlkAttr.FINALIZE);
         debug(PRINTF) printf("-_d_monitor_create(%p)\n", h);
     }
 
@@ -412,6 +416,10 @@ version( USE_PTHREADS )
         pthread_mutex_unlock(&_monitor_critsec);
         if (cs)
             free(cs);
+        // Set the finalize bit so that the monitor gets collected (Bugzilla 14573)
+        import core.memory : GC;
+        if (!(typeid(h).m_flags & TypeInfo_Class.ClassFlags.hasDtor))
+            GC.setAttr(cast(void*)h, GC.BlkAttr.FINALIZE);
         debug(PRINTF) printf("-_d_monitor_create(%p)\n", h);
     }
 
@@ -440,4 +448,16 @@ version( USE_PTHREADS )
         pthread_mutex_unlock(&getMonitor(h).mon);
         debug(PRINTF) printf("-_d_monitor_release(%p)\n", h);
     }
+}
+
+// Bugzilla 14573
+unittest
+{
+    import core.memory : GC;
+
+    auto obj = new Object;
+    assert(!(GC.getAttr(cast(void*)obj) & GC.BlkAttr.FINALIZE));
+    _d_monitor_create(obj);
+    assert(getMonitor(obj) !is null);
+    assert(GC.getAttr(cast(void*)obj) & GC.BlkAttr.FINALIZE);
 }
--- a/src/rt/monitor_.d
+++ b/src/rt/monitor_.d
@@ -1,23 +1,13 @@
 /**
  * Contains the implementation for object monitors.
  *
- * Copyright: Copyright Digital Mars 2000 - 2011.
+ * Copyright: Copyright Digital Mars 2000 - 2015.
  * License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
- * Authors:   Walter Bright, Sean Kelly
- */
-
-/*          Copyright Digital Mars 2000 - 2011.
- * Distributed under the Boost Software License, Version 1.0.
- *    (See accompanying file LICENSE or copy at
- *          http://www.boost.org/LICENSE_1_0.txt)
+ * Authors:   Walter Bright, Sean Kelly, Martin Nowak
  */
 module rt.monitor_;
 
-//debug=PRINTF;
-
-///////////////////////////////////////////////////////////////////////////////
-// Monitor
-///////////////////////////////////////////////////////////////////////////////
+import core.atomic, core.stdc.stdlib, core.stdc.string;
 
 // NOTE: The dtor callback feature is only supported for monitors that are not
 //       supplied by the user.  The assumption is that any object with a user-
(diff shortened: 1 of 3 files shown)


Safety0ff:
> fix race condition (incorrect double checked locking)

[bug 6607](https://issues.dlang.org/show_bug.cgi?id=6607)?

MartinNowak:
Oh, there is a bug report for that.

Safety0ff:
Is PR #1272 superceded by this one? I see #1272 deadlocks and this one duplicates commits.

MartinNowak:
It's based on this pull request, because it reuses the mutex stuff.

MartinNowak:
Can we please merge this (and it's dependency #1272) ASAP, it fixes the autotester deadlock caused by a newly added synchronized statement in rt.trace, which stalls posix testing and every other PR.

src/rt/monitor_.d
```diff
@@ -1,444 +1,289 @@
...
-nothrow:
+// NOTE: The dtor callback feature is only supported for monitors that are not
+//       supplied by the user.  The assumption is that any object with a user-
+//       supplied monitor may have special storage or lifetime requirements and
+//       that as a result, storing references to local objects within Monitor
+//       may not be safe or desirable.  Thus, devt is only valid if impl is
+//       null.
 
-private
+extern (C) void _d_setSameMutex(shared Object ownee, shared Object owner) nothrow in
```
rainers: I suspect the formatting of the `in` contract is a bug in dfmt?

src/rt/monitor_.d
```diff
@@ -1,444 +1,289 @@
...
     {
-        version = USE_PTHREADS;
+        // let the GC collect the monitor
+        setMonitor(h, null);
     }
-    else version( CRuntime_Bionic )
+    else if (!atomicOp!("-=")(m.refs, cast(size_t) 1))
     {
-        version = USE_PTHREADS;
+        // refcount == 0 means unshared => no synchronization required
```
rainers: This kind of reference counting is not thread safe (consider another thread executing _d_setSameMutex while the current thread is intercepted right here). This problem is not new to this PR, though.

src/rt/monitor_.d
```diff
@@ -1,444 +1,289 @@
...
+    alias destroyMutex = DeleteCriticalSection;
+    alias lockMutex = EnterCriticalSection;
+    alias unlockMutex = LeaveCriticalSection;
+}
+else version (Posix)
+{
+    import core.sys.posix.pthread;
 
-    extern (C) void _d_monitor_lock(Object h)
+    struct Monitor
```
rainers: With the help of alias Mutex, this struct could be moved out of the conditional.

src/rt/monitor_.d
```diff
@@ -1,444 +1,289 @@
...
-        debug(PRINTF) printf("+_d_monitor_release(%p)\n", h);
-        assert(h && h.__monitor && !getMonitor(h).impl);
-        LeaveCriticalSection(&getMonitor(h).mon);
-        debug(PRINTF) printf("-_d_monitor_release(%p)\n", h);
+        pthread_mutex_init(mtx, &gattr);
     }
+
+    alias destroyMutex = pthread_mutex_destroy;
+    alias lockMutex = pthread_mutex_lock;
+    alias unlockMutex = pthread_mutex_unlock;
```
rainers: These posix functions have non-void return values which are ignored (as in the old version). Is this actually ok?

rainers:
Apart from those nitpicks LGTM

src/rt/monitor_.d
MartinNowak: It's safe because you can't call setSameMutex on an object whose refcount just dropped to 0, or otherwise it wouldn't be collected, meaning you no longer own it.

src/rt/monitor_.d
MartinNowak: Apparently already fixed in the latest version :).

src/rt/monitor_.d
MartinNowak: The premise here was, that those function never fail, but let me add an `assert(0)`.

MartinNowak:
done

src/rt/monitor_.d
rainers: Ah, ok. _d_monitordelete is only called when the destructor is called, too. So any dangling reference would be a bug elsewhere.

rainers:
> DEPENDS ON: #1272

I'm slightly confused. Doesn't this PR contain all the changes in #1272?

src/rt/monitor_.d
```diff
@@ -1,444 +1,293 @@
...
+    version (CRuntime_DigitalMars)
     {
-        debug(PRINTF) printf("+_STI_monitor_staticctor()\n");
-        if (!inited)
-        {
-            InitializeCriticalSection(&_monitor_critsec);
-            inited = 1;
-        }
-        debug(PRINTF) printf("-_STI_monitor_staticctor()\n");
+        pragma(lib, "snn.lib");
```
rainers: I have always wondered why these `pragma(lib)` statements are in this module. Didn't you remove them in the previous version?

src/rt/monitor_.d
```diff
@@ -1,444 +1,293 @@
...
+else
+{
+    static assert(0, "Unsupported platform");
+}
+
+struct Monitor
+{
+    IMonitor impl; // for user-level monitors
+    DEvent[] devt; // for internal monitors
+    size_t refs; // reference count
```
rainers: I hate code formatters for breaking white space alignment of consecutive lines.

src/rt/monitor_.d
MartinNowak: Exactly, it's like it says above `0 means unshared`, or more precisely `1 before dropping` means the current object is the last owner.

src/rt/monitor_.d
MartinNowak: And I hate aligned code, pointless exercise, but I also use a proportional font, so I wouldn't even remotely get the idea to align code.

src/rt/monitor_.d
MartinNowak: No I didn't, no idea why they are needed.

MartinNowak:
> I'm slightly confused. Doesn't this PR contain all the changes in #1272?

Yes I initially split them, but then made rt.critical depend on rt.monitor. Since you already reviewed both parts, I'm closing #1272.

src/rt/monitor_.d
MartinNowak: Let me try to remove them, might be related to the STD/STI style initializers.

MartinNowak: As it says here, 759dc8d02fea6ef312a5cd24fd754300fb9e3b90, the C library isn't dragged in otherwise.

MartinNowak:
Seems to work w/o the pragmas, anything left?

rainers:
Sorry for the delay and the bad advice. 

I tried it now: unfortunately, removing those pragmas causes link errors when building DLLs or when using WinMain instead of main. This is caused by snn.lib only being added to an object file for main in https://github.com/D-Programming-Language/dmd/blob/master/src/glue.c#L943. I'm not sure why this has been spread over the source like this, I'd prefer it just being part of the link command line.

MartinNowak:
> I'm not sure why this has been spread over the source like this, I'd prefer it just being part of the link command line.

Thanks, I'll remove that commit. This change doesn't belong here.

src/rt/monitor_.d
ibuclaw: It would be better if we understood the why rather than the what.

ibuclaw:
All looks good to me.  Now I just need to rewrite it for GDC. ;-)

MartinNowak:
> All looks good to me. Now I just need to rewrite it for GDC. ;-)

What's to be done there? I still have a very bad understanding of DMD vs. GDC/LDC differences, so cannot make decisions that help you. Guess I could easily port that change to GDC.

MartinNowak:
Auto-merge toggled on

MartinNowak:
> I'm not sure why this has been spread over the source like this, I'd prefer it just being part of the link command line.

I guess because this is among the first code that gets run for any D code, so it might have seemed sensible to move the pragmas here. They are better located somewhere around rt_init. I'd also prefer a linker switch but only if that doesn't break existing build setups.

ibuclaw:
> What's to be done there? 

It's nothing much.  I just abstract away the thread type to libgcc.gthreads.  Just to support different kinds of configuration (--thread-model=posix, --thread-model=single, etc...)

PR #1276, opened 2015-05-15
reduce excessive unittests and template bloat

MartinNowak:
- move enum name and index function out of them template
- check const conversions once, instead of repeating
  this for every operation
- using a const value guarantees that the method can be
  called for mutable and immutable values as well
- 15% compile time and 10% binary size of unittest library

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -2249,6 +2249,16 @@ struct MonoTimeImpl(ClockType clockType)
     else
         static assert(0, "Unsupported platform");
 
+    // POD value, test mutable/const/immutable conversion
+    unittest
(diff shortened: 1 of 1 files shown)


MartinNowak:
I stumbled over this while trying to bisect dmd compile time increases.

schveiguy:
See #1278, these will probably conflict. I'd recommend pulling that one first.

DmitryOlshansky:
#1278 is in - time to rebase?

MartinNowak:
updated

9rnsr:
Auto-merge toggled on

PR #1277, opened 2015-05-15
Add ARM linux error codes and remove some non-standard constants

joakim-noah:
Some missing pieces for the Android pulls that were recently merged, the second one asked for by Martin.


src/core/sys/posix/time.d
```diff
@@ -126,8 +126,6 @@ else version (Solaris)
 else version (CRuntime_Bionic)
 {
     enum CLOCK_MONOTONIC        = 1;
-    enum CLOCK_MONOTONIC_RAW    = 4;
-    enum CLOCK_MONOTONIC_COARSE = 6;
```
MartinNowak: Better keep them, core.time use those, but nonstandard stuff should live here instead https://github.com/D-Programming-Language/druntime/blob/8d3f96a0cb5d08de7d352d0fdaa8221e46bc8aad/src/core/sys/linux/time.d.

MartinNowak: In fact those constants are a kernel property and should merged under `version (linux)`.
https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/clock_getres.c;hb=b2fb25240813266e4f62f2dd039573f1a042ea2f#l39

joakim-noah: OK, I didn't notice the recent changes to `core.time` to use those.  I had these constants inside `version(CRuntime*)` checks, as opposed to `version(linux)`, because they are also defined by glibc and bionic, and I give precedence to the C runtime in those cases.

But I don't really care, easier to collapse them.  I'll do that.

joakim-noah:
Updated this pull to merge the `CLOCK_MONOTONIC*` sections and fix one recent glibc change from `core.sys.posix.stdio`.

src/core/sys/posix/stdio.d
```diff
@@ -200,7 +200,7 @@ FILE*  popen(in char*, in char*);
 // memstream functions are conforming to POSIX.1-2008.  These functions are
 // not specified in POSIX.1-2001 and are not widely available on other
 // systems.
-version( linux )                             // as of glibc 1.0x
+version( CRuntimeGlibc )                     // as of glibc 1.0x
```
redstar: Typo: `CRuntime_Glibc`

joakim-noah: Ah, thanks, couldn't see that, thought it was a spurious test run.

PR #1278, opened 2015-05-15
Some updates to MonoTime to better initialize the ticks per second

schveiguy:
Also:
- Factor out translation of ClockType enums to POSIX clock types as a parameter to clock_gettime, etc.
- Remove reimplementation of getting MonoTime.ticks from GC code, since the initializer is now run before the gc initializer.
- Factor out translation of clock type to clock type enum index.
- Factor out traits lookup of enum name.

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -312,6 +312,51 @@ else
     static assert(0, "What are the clock types supported by this system?");
 }
 
+// private, used to translate clock type to proper argument to clock_xxx
+// functions on posix systems
+version(Posix)
+{
+    private auto _posixClock(ClockType clockType)
+    {
+        version(linux)
+        {
+            import core.sys.linux.time;
+            with(ClockType) final switch(clockType)
+            {
+            case bootTime: return CLOCK_BOOTTIME;
+            case coarse: return CLOCK_MONOTONIC_COARSE;
+            case normal: return CLOCK_MONOTONIC;
+            case precise: return CLOCK_MONOTONIC;
+            case processCPUTime: return CLOCK_PROCESS_CPUTIME_ID;
+            case raw: return CLOCK_MONOTONIC_RAW;
+            case threadCPUTime: return CLOCK_THREAD_CPUTIME_ID;
+            case second: assert(0);
+            }
+        }
+        else version(FreeBSD)
+        {
+            import core.sys.freebsd.time;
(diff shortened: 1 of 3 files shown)


src/core/time.d
```diff
@@ -312,6 +312,50 @@ else
     static assert(0, "What are the clock types supported by this system?");
 }
 
+// private, used to translate clock type to proper argument to clock_xxx
+// functions on posix systems
+version(Posix)
+{
+    private auto _posixClock(ClockType clockType)
```
MartinNowak: That's a good name, couldn't come up with any.

src/core/time.d
```diff
@@ -312,6 +312,50 @@ else
...
+            case precise: return CLOCK_MONOTONIC;
+            case processCPUTime: return CLOCK_PROCESS_CPUTIME_ID;
+            case raw: return CLOCK_MONOTONIC_RAW;
+            case threadCPUTime: return CLOCK_THREAD_CPUTIME_ID;
+            case second: assert(0);
+            }
+        }
+        else version(FreeBSD)
+        {
+            with(ClockType) final switch(clockType)
```
MartinNowak: Need to import core.sys.freebsd.time here.

MartinNowak:
LGTM

src/core/time.d
schveiguy: Oops. Fixed.

src/core/time.d
```diff
@@ -2595,7 +2618,7 @@ assert(before + timeElapsed == after).
             return "MonoTime(" ~ numToString(_ticks) ~ " ticks, " ~ numToString(ticksPerSecond) ~ " ticks per second)";
         else
         {
-            enum name = __traits(allMembers, ClockType)[_clockIdx];
+            enum name = _clockName;
             return "MonoTimeImpl!(ClockType." ~ name ~ ")(" ~ numToString(_ticks) ~ " ticks, " ~
                    numToString(ticksPerSecond) ~ " ticks per second)";
```
schveiguy: Hm... this is a wasted line...

schveiguy:
I want to do a couple of fixups, and squash. Don't push this yet.

schveiguy:
All set.

schveiguy:
@jmdavis want to take a look at this?

MartinNowak:
Auto-merge toggled on

PR #1279, opened 2015-05-16
replace UnicodeException with UnicodeError

MartinNowak:
- first step to make invalid UTF encodings in strings
  a programming error (ensuring valid encoding has to be
  done when converting raw input data to string)
- attributes for rt.util.utf
- added staticError helper to throw @nogc Errors

### DIFF
--- a/src/core/exception.d
+++ b/src/core/exception.d
@@ -315,14 +315,17 @@ unittest
 }
 
 
+deprecated("UnicodeException is now an Error.")
+alias UnicodeException = UnicodeError;
+
 /**
  * Thrown on a unicode conversion error.
  */
-class UnicodeException : Exception
+class UnicodeError : Error
 {
     size_t idx;
 
-    this( string msg, size_t idx, string file = __FILE__, size_t line = __LINE__, Throwable next = null ) @safe pure nothrow
+    this( string msg, size_t idx, string file = __FILE__, size_t line = __LINE__, Throwable next = null ) @safe pure nothrow @nogc
     {
         super( msg, file, line, next );
         this.idx = idx;
@@ -332,7 +335,7 @@ class UnicodeException : Exception
 unittest
 {
     {
-        auto ue = new UnicodeException("msg", 2);
+        auto ue = new UnicodeError("msg", 2);
         assert(ue.file == __FILE__);
         assert(ue.line == __LINE__ - 2);
         assert(ue.next is null);
@@ -341,7 +344,7 @@ unittest
     }
 
     {
-        auto ue = new UnicodeException("msg", 2, "hello", 42, new Exception("It's an Exception!"));
+        auto ue = new UnicodeError("msg", 2, "hello", 42, new Exception("It's an Exception!"));
         assert(ue.file == "hello");
         assert(ue.line == 42);
         assert(ue.next !is null);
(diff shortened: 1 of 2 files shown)


MartinNowak:
The outcome of [Issue 14519 – [Enh] foreach on strings should return replacementDchar rather than throwing](https://issues.dlang.org/show_bug.cgi?id=14519) is that the programmer has to ensure that strings contains valid unicode data (e.g. by validating raw data as done by [readText](http://dlang.org/phobos/std_file.html#.readText)) and any algorithm working on strings should assert that but not repeat the validation.

src/rt/util/utf.d
```diff
@@ -566,7 +565,7 @@ Checks to see if string is well formed or not. $(D S) can be an array
  of $(D char), $(D wchar), or $(D dchar). Throws a $(D UtfException)
  if it is not. Use to check all untrusted input for correctness.
  */
-void validate(S)(in S s)
+void validate(S)(in S s) pure nothrow @safe @nogc
```
CyberShadow: `readText` calls `validate`, which used to throw `UnicodeException`, so won't this break validating text in `readText`?

MartinNowak: You're right, it's a different validate, but this one shouldn't be nothrow and we still need UTFExcwption for validation.
BTW the signature of validate should become `inout(char)[] validate(input(ubyte)[] str);`.

CyberShadow: I'm still unsure about the whole `ubyte`-for-ASCII idea. I think it's an interesting direction iff we improve Phobos and the language to better support it as well (e.g. allowing comparison of `ubyte[]` and string literals, making `s.startsWith("foo")` work with `ubyte[] s`, etc.)

src/core/exception.d
```diff
@@ -315,14 +315,17 @@ unittest
 }
 
 
+deprecated("UnicodeException is now an Error.")
+alias UnicodeException = UnicodeError;
```
schuetzm: If we still need `UnicodeException` for validation, then this is wrong, isn't it? It mustn't be an `Error` if it's supposed to be catchable.

src/rt/util/utf.d
schuetzm: What do you mean by "ubyte-for-ASCII"? ASCII is a subset of UTF8, `char` is fine for it.

src/core/exception.d
MartinNowak: Yes, it's wrong.

src/rt/util/utf.d
CyberShadow: Sorry, I meant non-UTF 8-bit encodings.

MartinNowak: Well `s.startsWith("foo".representation)` would work.
You can't generally mix `string` with `ubyte[]` because one needs to use autodecoding, while the other can't.

src/core/exception.d
```diff
@@ -315,14 +315,17 @@ unittest
...
+
 /**
  * Thrown on a unicode conversion error.
  */
-class UnicodeException : Exception
+class UnicodeError : Error
```
DmitryOlshansky: Breaking change. How would users fix their code? In fact it may pass all of tests untill it hits bad encoding in the wild and crashes.

DmitryOlshansky:
> added staticError helper to throw @nogc Errors

Useful in and out of itself. Break it out?

MartinNowak:
> Useful in and out of itself. Break it out?

Done, https://github.com/D-Programming-Language/druntime/pull/1325.

d-random-contributor:
```
void validate(S)(in S s) pure @safe
{
  if (!s.invalidDchars.empty)
    throw new UnicodeException();
}

auto invalidDchars(S)(in S s) pure nothrow @safe @nogc;
```
Do we have nothrow @nogc validation?

DmitryOlshansky:
Please rebase, #1325 was pulled.

DmitryOlshansky:
Any updates on this? @MartinNowak

MartinNowak:
It's really to wrong approach to start this transition by replacing some exceptions in druntime.
I'll try to bring this topic up for discussion again, hopefully we can reach a consensus.

PR #1283, opened 2015-05-17
[minor] core.time: Fix SysTime link

CyberShadow:


MartinNowak:
Auto-merge toggled on

PR #1284, opened 2015-05-22
Handle (undocumented) Nk and Nj mangled symbols

ibuclaw:
https://issues.dlang.org/show_bug.cgi?id=14566
https://issues.dlang.org/show_bug.cgi?id=14567


yebblies:
Auto-merge toggled on

PR #1285, opened 2015-05-24
Fix issue 14617 - PTHREAD_MUTEX_INITIALIZER does not work on OSX

schveiguy:
This adds an initializer to pthread_mutex_t so it is correct by default on OSX. Not sure if this will work, it failed my unittest build on OSX with a link error. I think possibly types.d is not included in the unit test build.

### DIFF
--- a/src/core/sys/posix/pthread.d
+++ b/src/core/sys/posix/pthread.d
@@ -162,8 +162,8 @@ else version( OSX )
         PTHREAD_EXPLICIT_SCHED  = 2
     }
 
-    enum PTHREAD_MUTEX_INITIALIZER  = pthread_mutex_t.init;
-    enum PTHREAD_ONCE_INIT          = pthread_once_t.init;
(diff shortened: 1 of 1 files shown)


MartinNowak:
Don't think you want that initializer for the type itself, but only for PTHREAD_MUTEX_INITIALIZER.

andralex:
@MartinNowak why not? Then PTHREAD_MUTEX_INITIALIZER is simply pthread_mutex_t.init. In fact that's exactly what happens on Linux - zero-initialization just works.

schveiguy:
Updated to only change `pthread.d` definition. I also added PTHREAD_ONCE_INIT since it was right there. I don't know about the others.

MartinNowak:
> @MartinNowak why not?

Because it's a bad idea to deviate from the documented and standardized C behavior in a simple C header.

MartinNowak:
Auto-merge toggled on

schveiguy:
The standard behavior is to fill with random data. We don't do that either ;)

PR #1286, opened 2015-05-27
Remove unused object.MemberInfo

klickverbot:
It was supposed to be used for the still unimplemented
ModuleInfo.xgetMembers(), but the latter only returns
void*.


yebblies:
Auto-merge toggled on

PR #1287, opened 2015-05-27
Use osmodel.mak from dmd repo

lionello:
There were slight difference between the logic in `posix.mak` compared to `osmodel.mak` (DRY)

Also see https://github.com/D-Programming-Language/dmd/pull/4684

### DIFF
--- a/osmodel.mak
+++ b/osmodel.mak
@@ -1,0 +1,52 @@
+# This Makefile snippet detects the OS and the architecture MODEL
+# Keep this file in sync between druntime, phobos, and dmd repositories!
+
+ifeq (,$(OS))
+  uname_S:=$(shell uname -s)
+  ifeq (Darwin,$(uname_S))
+    OS:=osx
+  endif
+  ifeq (Linux,$(uname_S))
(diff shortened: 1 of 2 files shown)


andralex:
Fine. I think a future but more controversial step will be to directly refer to the same file.

andralex:
Auto-merge toggled on

WalterBright:
naturally, osmodel.mak was not added to the zip file for druntime. Nobody ever remembers to do that :-(

WalterBright:
Also, there is mak/COPY, mak\DOCS, mak\IMPORTS, mak\MANIFEST, mak\SRCS, and now an include. Are things getting better, or just more complicated?

lionello:
@WalterBright Sorry, but the only way to prevent this is to make building the zip part of the auto-tester. There's no other way to go about it.
I'm fixing it now.

WalterBright:
Merely building a zip won't detect these problems. I find them when I build a zip in order to transfer the whole package to another system so I can build it there. Then, the build fails because some file is missing.

PR #1290, opened 2015-05-30
Two fixes for Bionic

joakim-noah:
lockf isn't provided in Bionic and the ilogb constants are defined differently.


DmitryOlshansky:
LGTM

schveiguy:
Auto-merge toggled on

MartinNowak:
Please don't put non-Posix functions into core.sys.posix, flock should live under core.sys.linux or core.sys.bionic.

PR #703, opened 2014-01-05
Fix Issue 11865 - Unhandled exception does not show crash dialog

CyberShadow:
https://d.puremagic.com/issues/show_bug.cgi?id=11865

If a console window is not present, display uncaught exceptions in a standard Windows message box.

PR #1293, opened 2015-05-31
Windows:unhandled exception dump should not show dialog if stderr is redirected

rainers:
Followup for #703

### DIFF
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -517,7 +517,10 @@ extern (C) void _d_print_throwable(Throwable t)
     // Show a message box instead.
     version (Windows)
     {
-        if (!GetConsoleWindow())
+        // ensure the exception is shown at the beginning of the line, while also
+        // checking whether stderr is a valid file
+        int written = fprintf(stderr, "\n");
+        if (written <= 0)
         {
             static struct WSink
             {


CyberShadow:
Will it ever be `INVALID_HANDLE_VALUE`? IIRC std.process has to check for both `null` and `INVALID_HANDLE_VALUE`: https://github.com/D-Programming-Language/phobos/pull/3148

CyberShadow:
This is an improvement but I think ideally it should display the message box if it tries and fails to write to stderr (e.g. if it points to a closed file). If stderr is reopened by the application, this might not do what the user expects.

rainers:
> Will it ever be INVALID_HANDLE_VALUE?

I used null because of this note in the MSDN doc for GetStdHandle: "If an application does not have associated standard handles, such as a service running on an interactive desktop, and has not redirected them, the return value is NULL."

Checking for INVALID_HANDLE_VALUE would probbly be ok, too.

rainers:
> ideally it should display the message box if it tries and fails to write to stderr

In that case fprintf should return 0, but it's a bit ugly to check because it is only called in the sink delegate. Can we add printing some header?

CyberShadow:
How about prefixing "Unhandled "? Or would that break editors/IDEs matching the error string?

rainers:
I changed it to try writing a newline, so the exception is also guaranteed to be shown at the beginning of a line.

DmitryOlshansky:
LGTM

DmitryOlshansky:
@CyberShadow all good on this one?

CyberShadow:
Yep. Getting used to a newline before stack traces will be a tiny nuisance, but it's justified.

DmitryOlshansky:
Auto-merge toggled on

PR #1294, opened 2015-05-31
detab, tolf, remove trailing whitespace

rainers:
Trying not to commit bad line endings again, I added checkwhitespace to my build batch. This fails for some existing files, though.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -211,9 +211,9 @@ MAKEFILES = $(filter mak/% %.mak,$(MANIFEST))
 NOT_MAKEFILES = $(filter-out $(MAKEFILES),$(MANIFEST))
 
 checkwhitespace:
-	grep -n -U -e "[ \t]$$$$" -e "\r" $(MAKEFILES) ; test "$$?" -ne 0
-	grep -n -U -e " $$$$" -e "\r|\t" $(NOT_MAKEFILES) ; test "$$?" -ne 0
-	
+	grep -n -P "([ \t]$$|\r)" $(MAKEFILES) ; test "$$?" -ne 0
+	grep -n -P "( $$|\r\t)" $(NOT_MAKEFILES) ; test "$$?" -ne 0
+
 detab:
 	detab $(MANIFEST)
 	tolf $(MANIFEST)
--- a/posix.mak
+++ b/posix.mak
@@ -206,6 +206,14 @@ test/%/.run: test/%/Makefile
 	$(QUIET)$(MAKE) -C test/$* MODEL=$(MODEL) OS=$(OS) DMD=$(abspath $(DMD)) \
 		DRUNTIME=$(abspath $(DRUNTIME)) DRUNTIMESO=$(abspath $(DRUNTIMESO)) QUIET=$(QUIET) LINKDL=$(LINKDL)
 
+#################### test for undesired white spaces ##########################
+MAKEFILES = $(filter mak/% %.mak,$(MANIFEST))
+NOT_MAKEFILES = $(filter-out $(MAKEFILES),$(MANIFEST))
+
+checkwhitespace:
+	grep -n -e "[ \t]$$" -e "\r" $(MAKEFILES) ; test "$$?" -ne 0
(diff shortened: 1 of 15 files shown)


andralex:
This is bound to happen again. Do you think you could investigate adding commit hooks to enforce this stuff? Otherwise we'll be spending time again and again pulling PRs like this.

rainers:
A commit hook needs every contributor to setup his environment which will be different on every platform and tooling. I think the simpler and better approach is to run the check on the auto tester very much like it is done for the dmd sources now.

If we can make the checkwhitespace tool available for druntime/phobos aswell, we could add similar targets to the respective makefiles.

joakim-noah:
I think that's what Andrei meant by a "commit hook," to have the check run automatically on the auto-tester when the PR is submitted.  Either way, I agree with you that that is the best approach, failing or amending the PR quickly if it doesn't meet certain criteria.

DmitryOlshansky:
@braddr @CyberShadow - any thoughts on running detab, tolf etc. to detect bad whitespace in auto-tester?

CyberShadow:
Well, all you need to do to make Brad's tester test something is add it to the makefile. I could also add it to mine, in theory, I suppose.

rainers:
I think it needs checkwhitespace from the dmd folder to be precompiled and accessible somewhere on the build server. Then we could add a check to the auto-tester target in posix.mak for druntime and phobos. For druntime I used this:
```
_TOCHECK = $(MANIFEST)
TOCHECK = $(filter-out mak\COPY mak\DOCS mak\IMPORTS mak\SRCS mak\MANIFEST,$(_TOCHECK))

checkwhitespace:
	../tools/checkwhitespace $(TOCHECK)
```

DmitryOlshansky:
@rainers that or conjure something with shell scripts. `grep -P "^.* +$` would work for tailing white space check, `grep -P "\r\n|\t"` for windows line endings and tabs.

rainers:
I've rebased this and tried a makefile rule to be run by the auto tester, but I could not really test it, because grep does some strange things with CR, even with -U. -P also wasn't available with the version I used.

Using grep is probably better than checkwhitespace from the dmd folder because it will give you an idea where the error is.

DmitryOlshansky:
LGTM

DmitryOlshansky:
Auto-merge toggled on

yebblies:
> Using grep is probably better than checkwhitespace from the dmd folder because it will give you an idea where the error is.

That isn't super important because once you know there's a whitespace problem somewhere in a file you can just search for it in your editor.

rainers:
I've added some improvements here: https://github.com/D-Programming-Language/druntime/pull/1315
https://github.com/D-Programming-Language/phobos/pull/3474 adds a similar check to phobos.

rainers:
> That isn't super important because once you know there's a whitespace problem somewhere in a file you can just search for it in your editor.

True, though it is a small inconvenience. There are some other downsides to the approach in dmd though:
- building checkwhitespace every time adds a noticeable delay during a build
- if an invalid UTF8 character creeps in, checkwhitespace just bails out without indication in what file it is in.
- it cannot be used for the files in the mak folder because the makefile detection does not work there.

The grep version has some disadvantages, too:
- it doesn't detect invalid UTF8 charcters
- it doesn't tell you what the actual problem is on the displayed line: CR, tab or trailing whitespace?

yebblies:
> The grep version has some disadvantages, too:

rainers:
> Also doesn't work on windows.

It's a bit more complicated to setup (it needs a unix-make and a "recent" grep, later than 2005), but you can use druntime\posix.mak for the check on windows, too. See #1315.

yebblies:
The spurious druntime failures refer to grep, is the whitespace check possibly involved in a race condition?  Does it depend on any generated files?

rainers:
> Does it depend on any generated files?

No, it just checks files in the MANIFEST variable.

schveiguy:
Just FYI, the spurious failures should get fixed here: https://github.com/D-Programming-Language/druntime/pull/1390

The grep is failing because of a race condition, and some copy paste issues.

PR #1296, opened 2015-06-02
add osmodel.mak to MANIFEST

WalterBright:
Because nobody ever remembers to update MANIFEST when adding files, leading to files missing from the zip file.

grumble, grumble


lionello:
Looks good.

PR #1297, opened 2015-06-03
Ignore real pad in core.internal.convert unittest

smolt:
Change unittest to ignore any pad bytes in a floating point type because
those bits are undefined. Fixes test part of ldc issue #788.

p.s. that's https://github.com/ldc-developers/ldc/issues/788


MartinNowak:
Auto-merge toggled on

MartinNowak:
thx

PR #1299, opened 2015-06-05
add auto-tester specific targets

braddr:
This gives control to the repository rather than the auto-tester.

The choice of target is based on current behavior.  For auto-tester-build, the tester isn't specifying a target, so the behavior is 'whatever is first in the makefile'.  That's 'target' for all the current makefiles.  For auto-tester-test, that's explicitly 'unittest' in the current tester build scripts.

This will need to be merged to the other branches before the tester can switch to using these new targets.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -229,3 +229,10 @@ clean: $(addsuffix /.clean,$(ADDITIONAL_TESTS))
 
 test/%/.clean: test/%/Makefile
 	$(MAKE) -C test/$* clean
+
+.PHONY : auto-tester-build
+auto-tester-build: target
+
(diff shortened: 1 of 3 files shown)


braddr:
Ping?  The dmd and phobos versions of these changes are merged already.

schveiguy:
I'm not sure what this does exactly, but the fact that others have merged it, and it passes the auto tester is good enough for me.

schveiguy:
Auto-merge toggled on

braddr:
Thanks, but in the future, please don't merge changes you can't vouch for the correctness of.

PR #1300, opened 2015-06-07
fix property enforcement

9rnsr:


jmdavis:
Auto-merge toggled on

PR #1301, opened 2015-06-11
DOCS: fix core.thread.Thread.sleep link

John-Colvin:
don't really know if this is the right way to do this, but it's an opportunity to make use of Vladimir's new docs autotester :)

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -382,7 +382,7 @@ unittest
     (e.g. 22 days or 700 seconds).
 
     It is used when representing a duration of time - such as how long to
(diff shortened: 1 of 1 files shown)


schveiguy:
Nice, LGTM! I love the new doc generation preview! I can even live with the OLD/NEW swap bug :)

schveiguy:
Auto-merge toggled on

schveiguy:
I think this is a DDOC issue, not your issue, but in the standard docs (not the ddox), `core_thread.html#Thread.sleep` does not actually go to the sleep function. But `core_thread.html#.Thread.sleep` does. Odd...

PR #1305, opened 2015-06-20
Modify TypeInfo.init to match the spec

JinShil:
Per http://dlang.org/library/object/type_info.init.html

> Return default initializer. If the type should be initialized to all zeros, an array with a null ptr and a length equal to the type size will be returned.

P.S. The TODO in the comment says it should be made into a property. Is this still the plan, or should that comment be removed?

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -286,7 +286,10 @@ class TypeInfo
     /// Return default initializer.  If the type should be initialized to all zeros,
     /// an array with a null ptr and a length equal to the type size will be returned.
     // TODO: make this a property, but may need to be renamed to diambiguate with T.init...
-    const(void)[] init() nothrow pure const @safe @nogc { return null; }
+    const(void)[] init() nothrow pure const @safe @nogc 
+    { 
+        return (cast(void *)null)[0..tsize];
+    }
 
     /// Get flags for type: 1 means GC should scan for pointers,
     /// 2 means arg of this type is passed in XMM register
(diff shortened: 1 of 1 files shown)


yebblies:
I think it would be better to fix the documentation to match the current behavior than to return an invalid array.  Note that this fails the autotester as it's unsafe to slice a pointer.

JinShil:
Well, this didn't turn out to be as straightforward as I thought it would be.  I will pursue it again at a later time.

Still I'd like to have additional opinions on whether to do a documentation fix or a code fix.

JinShil:
I'm not sure, but I think if we do a documentation fix, then [this code in DMD](https://github.com/D-Programming-Language/dmd/blob/master/src/typinf.c#L458) will also need to change, yes?

rainers:
Calling `tsize` adds another virtual function call which should be avoided IMO. Derived classes might even call `init` inside `tsize` causing infinite recursions (TypeInfo_Struct.tsize does it, but implements `init()` aswell).

I think `init` should be properly implemented in the predefined TypeInfo declarations in rt/typeinfo according to the documentation. It's rather useless if some types don't have a proper return value. `TypeInfo.init` might even be made abstract to not create a bad default.

schveiguy:
> Calling tsize adds another virtual function call which should be avoided IMO. 

This is a good point. 

> Derived classes might even call init inside tsize causing infinite recursions (TypeInfo_Struct.tsize does it, but implements init() aswell).

This is not a problem. TypeInfos are not custom user-defined types, we define them. We can simply avoid this problem. Just add some unit tests to the type info modules (if they aren't there already).

> TypeInfo.init might even be made abstract to not create a bad default.

This is probably the best idea. This avoids the issue of having a non-compliant default, and the issue of having a double virtual call.

PR #1306, opened 2015-06-21
Modify TypeInfo.init to match the spec

JinShil:
Per http://dlang.org/library/object/type_info.init.html (or the comment in the source):

> Return default initializer. If the type should be initialized to all zeros, an array with a null ptr and a length equal to the type size will be returned.

Other Notes:
* I also removed the TODO to make it a property, as I don't think that belongs in production code.  If it should be done, then whoever wrote that should have done it.
* Some seem to favor just updating the documentation and leaving the current implementation.  I'm not necessarily against that, but that seems like more of a cop out than due diligence.  Also, I'm not sure, but I think if we do a documentation fix, then [this code in DMD](https://github.com/D-Programming-Language/dmd/blob/master/src/typinf.c#L458) will also need to change.
* It was mentioned that the call to virtual function `tsize` should be avoided because it could negatively affect the derived classes.  I think that that is a problem for the derived classes to solve.
* Also, it was mentioned that this implementation should be deferred to derived classes.  I actually started down that path, but it seems like I was just duplicating the same implementation over and over again, and the code started to smell.

See Also:
* [Bug 2990](https://issues.dlang.org/show_bug.cgi?id=2990)
* https://github.com/D-Programming-Language/druntime/pull/1305

Despite my opinions, I'd be happy to change this pull request to whatever implementation the consensus prefers.  Please advise me so I can get it done right and be done with it.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -285,8 +285,10 @@ class TypeInfo
 
     /// Return default initializer.  If the type should be initialized to all zeros,
     /// an array with a null ptr and a length equal to the type size will be returned.
-    // TODO: make this a property, but may need to be renamed to diambiguate with T.init...
-    const(void)[] init() nothrow pure const @safe @nogc { return null; }
+    const(void)[] init() nothrow pure const @trusted @nogc 
+    {
+        return (cast(void *)null)[0..tsize];
+    }
 
     /// Get flags for type: 1 means GC should scan for pointers,
     /// 2 means arg of this type is passed in XMM register


schveiguy:
> Also, it was mentioned that this implementation should be deferred to derived classes. I actually started down that path, but it seems like I was just duplicating the same implementation over and over again, and the code started to smell.

We can put up with repetitive code smell at this low level. Whatever performs best is the answer, as this is not implementation anyone is going to look at.

I like the solution of having an abstract init stub, and implementing in each of the derived types. One thing it will do immediately is tell you where all the incorrect implementations are :)

JinShil:
> I like the solution of having an abstract init stub, and implementing in each of the derived types.

https://github.com/D-Programming-Language/druntime/pull/1307

schveiguy:
I'll close this, I think #1307 is preferable.

PR #1307, opened 2015-06-23
Modify TypeInfo.init to match the spec - Take 2 - Derived class implementation

JinShil:
This is an alternative implementation for https://github.com/D-Programming-Language/druntime/pull/1306

It makes `TypeInfo.init()` abstract, and defers implementation to the derived classes.

**Motivation** (copied from https://github.com/D-Programming-Language/druntime/pull/1306 for convenience)
Per http://dlang.org/library/object/type_info.init.html (or the comment in the source):

>Return default initializer. If the type should be initialized to all zeros, an array with a null ptr and a length equal to the type size will be returned.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -285,8 +285,7 @@ class TypeInfo
 
     /// Return default initializer.  If the type should be initialized to all zeros,
     /// an array with a null ptr and a length equal to the type size will be returned.
-    // TODO: make this a property, but may need to be renamed to diambiguate with T.init...
-    const(void)[] init() nothrow pure const @safe @nogc { return null; }
+    abstract const(void)[] init() nothrow pure const @safe @nogc;
 
     /// Get flags for type: 1 means GC should scan for pointers,
     /// 2 means arg of this type is passed in XMM register
@@ -338,7 +337,7 @@ class TypeInfo_Typedef : TypeInfo
 
     override @property inout(TypeInfo) next() nothrow pure inout { return base.next; }
     override @property uint flags() nothrow pure const { return base.flags; }
-    override const(void)[] init() nothrow pure const @safe { return m_init.length ? m_init : base.init(); }
+    override const(void)[] init() const { return m_init.length ? m_init : base.init(); }
 
     override @property size_t talign() nothrow pure const { return base.talign; }
 
@@ -464,6 +463,11 @@ class TypeInfo_Array : TypeInfo
     override @property size_t tsize() nothrow pure const
     {
         return (void[]).sizeof;
+    }
+
+    override const(void)[] init() const @trusted
+    {
+        return (cast(void *)null)[0 .. (void[]).sizeof];
     }
 
     override void swap(void* p1, void* p2) const
--- a/src/rt/typeinfo/ti_C.d
+++ b/src/rt/typeinfo/ti_C.d
@@ -63,6 +63,11 @@ class TypeInfo_C : TypeInfo
         return Object.sizeof;
     }
 
+    override const(void)[] init() const @trusted
+    {
+        return (cast(void *)null)[0 .. Object.sizeof];
+    }
+
     override @property uint flags() nothrow pure
     {
         return 1;
--- a/src/rt/typeinfo/ti_void.d
+++ b/src/rt/typeinfo/ti_void.d
@@ -44,6 +44,11 @@ class TypeInfo_v : TypeInfo
         return void.sizeof;
     }
 
+    override const(void)[] init() const @trusted
+    {
+        return (cast(void *)null)[0 .. void.sizeof];
+    }
+
     override void swap(void *p1, void *p2)
     {
         byte t;
--- a/src/rt/typeinfo/ti_byte.d
+++ b/src/rt/typeinfo/ti_byte.d
@@ -44,6 +44,11 @@ class TypeInfo_g : TypeInfo
         return byte.sizeof;
     }
 
+    override const(void)[] init() @trusted
+    {
+        return (cast(void *)null)[0 .. byte.sizeof];
+    }
+
     override void swap(void *p1, void *p2)
     {
         byte t;
--- a/src/rt/typeinfo/ti_cent.d
+++ b/src/rt/typeinfo/ti_cent.d
@@ -52,6 +52,11 @@ class TypeInfo_zi : TypeInfo
         return cent.sizeof;
     }
 
+    override const(void)[] init() const @trusted
+    {
+        return (cast(void *)null)[0 .. cent.sizeof];
+    }
+
     override void swap(void *p1, void *p2)
     {
         cent t;
--- a/src/rt/typeinfo/ti_char.d
+++ b/src/rt/typeinfo/ti_char.d
@@ -53,10 +53,10 @@ class TypeInfo_a : TypeInfo
         *cast(char *)p2 = t;
     }
 
-    override const(void)[] init() nothrow pure
+    override const(void)[] init() const @trusted
     {
         static immutable char c;
 
-        return (cast(char *)&c)[0 .. 1];
+        return (&c)[0 .. 1];
     }
 }
--- a/src/rt/typeinfo/ti_dchar.d
+++ b/src/rt/typeinfo/ti_dchar.d
@@ -53,10 +53,10 @@ class TypeInfo_w : TypeInfo
         *cast(dchar *)p2 = t;
     }
 
-    override const(void)[] init() nothrow pure
+    override const(void)[] init() const @trusted
     {
         static immutable dchar c;
 
-        return (cast(dchar *)&c)[0 .. 1];
+        return (&c)[0 .. 1];
     }
 }
--- a/src/rt/typeinfo/ti_delegate.d
+++ b/src/rt/typeinfo/ti_delegate.d
@@ -50,6 +50,13 @@ class TypeInfo_D : TypeInfo
         *cast(dg *)p2 = t;
     }
 
+    override const(void)[] init() const @trusted
+    {
+        static immutable dg d;
+
+        return (cast(void *)null)[0 .. dg.sizeof];
+    }
+
     override @property uint flags() nothrow pure
     {
         return 1;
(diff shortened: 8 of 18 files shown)


src/object.d
```diff
@@ -338,7 +337,10 @@ class TypeInfo_Typedef : TypeInfo
 
     override @property inout(TypeInfo) next() nothrow pure inout { return base.next; }
     override @property uint flags() nothrow pure const { return base.flags; }
-    override const(void)[] init() nothrow pure const @safe { return m_init.length ? m_init : base.init(); }
+    override const(void)[] init() nothrow pure const @trusted @nogc
+    { 
+        return m_init.length ? m_init : (cast(void *)null)[0..0];
```
schveiguy: You are changing behavior here, I don't think you need to make any modifications.

src/object.d
```diff
@@ -285,8 +285,7 @@ class TypeInfo
 
     /// Return default initializer.  If the type should be initialized to all zeros,
     /// an array with a null ptr and a length equal to the type size will be returned.
-    // TODO: make this a property, but may need to be renamed to diambiguate with T.init...
-    const(void)[] init() nothrow pure const @safe @nogc { return null; }
+    abstract const(void)[] init() nothrow pure const @nogc;
```
schveiguy: Should still be `@safe`.

src/object.d
JinShil: Sorry! I had C# on the brain.  I will revert it.

schveiguy:
Why remove all the `pure` and `@nogc` attributes? Those should all remain, we don't want to make any code that uses `init` currently fail.

JinShil:
> Why remove all the pure and @nogc attributes? Those should all remain, we don't want to make any code that uses init currently fail.

According to http://dlang.org/function.html#FunctionAttributes:
> If a derived class overrides a base class member function with diferrent [FunctionAttributes](http://dlang.org/function.html#FunctionAttributes), the missing attributes will be automatically compensated by the compiler.

src/object.d
JinShil: `@safe` or `@trusted`?  I removed it because I was deferring to the derived class.

schveiguy: I think `@safe`, you can add a `@trusted` section to a safe function by using a function/delegate literal:

```D
override const(void)[] init() /* implied @safe */
{
   return (() @trusted => (cast(void *)null)[0..x])();
}
```

I don't like inferring `@trusted`, since it's very dangerous as an assumption. In fact, I'd say any virtual functions should never be marked `@trusted`.

schveiguy:
Ah, nice. I didn't know that they were inferred, it makes perfect sense.

However, I don't know what happens if you put `@safe` on the base, you may have to put a lot of `@trusted` escapes. Unfortunately, that seems to be what we have to do here.

src/object.d
schveiguy: Just tested it out, it actually does the right thing -- derived functions are inferred `@safe` unless marked with `@trusted`, so go ahead and mark `@safe` and then you can mark the derived functions that need it with `@trusted`

JinShil:
> However, I don't know what happens if you put @safe on the base, you may have to put a lot of @trusted escapes. Unfortunately, that seems to be what we have to do here.

Yeah, due to `TypeInfo_Typedef`, it looks like I'll have to mark it @safe anyway.

Also, should I add unit tests to the derived classes?

For example:
```
unittest
{
    byte b;
    auto init = typeid(b).init();
    assert(init.ptr is null);
    assert(init.length == byte.sizeof);
}
```
It seems a little redundant, but it would help coverage, I guess.

schveiguy:
See my comment inline, I think you can just mark the base `@safe` and then mark the appropriate implementations `@trusted`. Seems to infer `@safe` properly, a pleasant surprise!

schveiguy:
> should I add unit tests to the derived classes?

I don't think so, the implementation is quite simple. the unit tests would be more complex than the code it's testing!

And I don't think there are any unit tests for `init` anywhere else, are there?

JinShil:
> And I don't think there are any unit tests for init anywhere else, are there?

There are a few unit tests for `TypeInfo_Array` classes, but I couldn't find any unit tests for `init()`.

And speaking of `TypeInfo_Array`, I think that implementation is also incorrect.  So, in `TypeInfo_Array` I think I'll change it to...
```
override const(void)[] init() nothrow pure const @trusted @nogc 
{
    return (cast(void *)null)[0 .. 0];
}
```
According to my tests, that will keep the current implementation unchanged.  If you know what it should be, let me know, and I'll do the implementation proper.  Otherwise, I'll leave it until I figure out what it should be.

schveiguy:
> And speaking of TypeInfo_Array, I think that implementation is also incorrect.

I think your original implementation is correct. The one you post in this comment duplicates the issue that we are trying to fix (null array with 0 length instead of null array with length equal to the size of the type).

schveiguy:
To be clear, all `init` functions should return an array that has non-zero length. Ones that have zero initializers should be set to null pointer, ones that have a non-zero initializer should point at the correct initializer.

JinShil:
@schveiguy: Thanks for the review.  I believe I've covered everything.

schveiguy:
Thanks, LGTM.

9rnsr:
To reduce code repetition for `TypeInfo`s, today you can use `rt.util.typeinfo`. For example, you can write a mixin template for the implementation of `init()` member function.

schveiguy:
Hm... the cognitive load of using a mixin generator may be more than it's worth. These are one-liners really. I tend to dislike mixins that create API (although in this case, the API is already defined by the base class) because it's doing something non-obvious to the reader. The boilerplate seems quite minimal and straightforward to understand.

JinShil:
> To reduce code repetition for TypeInfos, today you can use rt.util.typeinfo. For example, you can write a mixin template for the implementation of init() member function.

I see and that's good to know.  I think that same pattern can actually be applied to many of the other functions in these `TypeInfo_` classes (e.g. `toString`, `equals`, `tsize`, etc...).  If I were writing these classes from scratch, I think that would be the best design.  However, I see a lot of disdain in this community for refactoring and disrupting working code, so I'm going to resist the temptation.

mleise:
That would solve https://issues.dlang.org/show_bug.cgi?id=13678 for me.

JinShil:
> That would solve https://issues.dlang.org/show_bug.cgi?id=13678 for me.

Actually, to my great disappointment I don't believe it does.  I tested with the following, and it failed for everything except `StuffDefault`.

```
unittest
{
    struct StuffDefault(T) { T[1024 * 1024] m_data; }
    struct StuffVoid(T) { T[1024 * 1024] m_data = void; }
    struct StuffZero(T) { T[1024 * 1024] m_data = 0; }

    alias Ts = TypeTuple!(
        StuffDefault!void,
        StuffVoid!void,
        StuffDefault!ubyte,
        StuffVoid!ubyte,
        StuffZero!ubyte,
        StuffVoid!char,
        StuffZero!char,
    );

    foreach (T; Ts) 
    {
        assert(typeid(T).init.ptr is null);
    }
}
```
I think some work still needs to be done on the `TypeInfo_Array` class and derived classes to solve that issue.

However, for what this pull request does solve, I believe it can be merged.

src/object.d
```diff
@@ -338,7 +337,7 @@ class TypeInfo_Typedef : TypeInfo
 
     override @property inout(TypeInfo) next() nothrow pure inout { return base.next; }
     override @property uint flags() nothrow pure const { return base.flags; }
-    override const(void)[] init() nothrow pure const @safe { return m_init.length ? m_init : base.init(); }
+    override const(void)[] init()  const { return m_init.length ? m_init : base.init(); }
```
9rnsr: Stylistic nitpick: `init()  const` -> `init() const`

src/rt/typeinfo/ti_C.d
```diff
@@ -62,6 +62,11 @@ class TypeInfo_C : TypeInfo
     {
         return Object.sizeof;
     }
+    
+    override const(void)[] init() const @trusted
+    { 
+        return (cast(void *)null)[0..Object.sizeof];
```
9rnsr: Nitpick: -> `[0 .. Object.sizeof]` for the consistent code style.

src/rt/typeinfo/ti_void.d
```diff
@@ -43,6 +43,11 @@ class TypeInfo_v : TypeInfo
     {
         return void.sizeof;
     }
+    
+    override const(void)[] init() const @trusted
+    { 
+        return (cast(void *)null)[0..void.sizeof];
```
9rnsr: Nitpick: -> `[0 .. void.sizeof]` for the consistent code style.

9rnsr:
OK, the implementation change itself is good.

Now auto-tester is failing because of the trailing spaces. Could you fix that and the nitpicks which I pointed out?

After that' I'll merge this.

9rnsr:
@JinShil Ready to merge?

JinShil:
> @JinShil Ready to merge?

@9rnsr, Yes, ready to merge.  Thank you!

9rnsr:
OK, let's go.

9rnsr:
Auto-merge toggled on

rainers:
I just noticed that this missed to implement init() for TypeInfo_Pointer, TypeInfo_AssociativeArray and a few more in object.d. Is this deliberate?

JinShil:
@rainers  It is not deliberate.  I just did the ones that I understood.  I've been meaning to revisit this and also try to address https://issues.dlang.org/show_bug.cgi?id=13678, but I need to understand a few things a little better.  If you want to provide some advise, it is most welcome.  Then I can do the pull request.

rainers:
I added https://github.com/D-Programming-Language/druntime/pull/1328

@JinShil AFAICT it is currently not specified how void initializers are reflected by the TypeInfo, so changing the init value could break code, e.g. so far you could assume that `init().length` is identical `tsize()` for structs.

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=15185

schveiguy:
I don't think that's a valid regression. We shouldn't have to support user-allocations of TypeInfo. Basically, use at your own risk.

PR #1308, opened 2015-06-29
checkedint functions force inline

burner:
* thanks to dip56 we have forced inlining
* for https://github.com/D-Programming-Language/phobos/pull/3389 dmd,ldc,gdc failed to inline functions of checkedint 

this should fix it

### DIFF
--- a/src/core/checkedint.d
+++ b/src/core/checkedint.d
@@ -46,6 +46,7 @@ pure:
  *      the sum
  */
 
+pragma(inline, true)
 int adds(int x, int y, ref bool overflow)
 {
     long r = cast(long)x + cast(long)y;
@@ -73,6 +74,7 @@ unittest
 }
 
 /// ditto
+pragma(inline, true)
 long adds(long x, long y, ref bool overflow)
 {
     long r = cast(ulong)x + cast(ulong)y;
@@ -114,6 +116,7 @@ unittest
  *      the sum
  */
 
+pragma(inline, true)
 uint addu(uint x, uint y, ref bool overflow)
 {
     uint r = x + y;
@@ -141,6 +144,7 @@ unittest
 }
 
 /// ditto
+pragma(inline, true)
 ulong addu(ulong x, ulong y, ref bool overflow)
 {
     ulong r = x + y;
@@ -181,6 +185,7 @@ unittest
  *      the sum
  */
 
+pragma(inline, true)
 int subs(int x, int y, ref bool overflow)
 {
     long r = cast(long)x - cast(long)y;
@@ -208,6 +213,7 @@ unittest
 }
 
 /// ditto
+pragma(inline, true)
 long subs(long x, long y, ref bool overflow)
 {
     long r = cast(ulong)x - cast(ulong)y;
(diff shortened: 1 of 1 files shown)


ibuclaw:
> this should fix it

If you say so.  Pragma's are made to be allowably ignored though. ;-)

burner:
might < should < will

;-)

JinShil:
LGTM.  LDC and GDC currently implement their own runtime anyway.

DmitryOlshansky:
> Pragma's are made to be allowably ignored though.

In C++ that's true. In D don't think so. Plus this one must fails if it can't or it's just some extra line noise.

JinShil:
>> Pragma's are made to be allowably ignored though.

> In C++ that's true. In D don't think so. 

Please see http://dlang.org/pragma.html#predefined-pragmas

> All implementations must support these, even if by just ignoring them:

DmitryOlshansky:
@JinShil See on the same page:

>Compilers must diagnose an error for unrecognized Pragmas, even if they are vendor specific ones

JinShil:
> Compilers must diagnose an error for unrecognized Pragmas, even if they are vendor specific ones

Touchez!  but...
That is under the Vendor Specific Pragmas heading while `pragma(inline)` is a predefined pragma.  Therefore, I interpret the two statements together to mean that each compiler must know about the predefined pragmas, but are free to do nothing with them (i.e. must not emit an error). However, when encountering a vendor-specific pragma that it does not recognize, the compiler should emit an error.  Hence, vendor-specific pragmas need to be wrapped in `version` to avoid upsetting vendors that the pragma is not intended for.

So, I think this PR is ok, because it only uses a predefined pragma.

DmitryOlshansky:
I'd say PR is ok if we get to see stuff inlined. @burner could you get us disassembly?

ibuclaw:
> Compilers must diagnose an error for unrecognized Pragmas, even if they are vendor specific ones

That is available via a compiler flag to turn that on/off. :)

burner:
I don't have disassemble, but @tsbockman and I both found that add(s|u) is not inlined when imported. If the function is duplicated in the file using it, it is inlined.

ibuclaw:
@burner - See my comments in the other PR regarding use of function pointers.

JinShil:
> but @tsbockman and I both found that add(s|u) is not inlined when imported. If the function is duplicated in the file using it, it is inlined.

I've run into the same problem with GDC.  Apparently [cross-module inlining is not yet supported] (http://forum.dlang.org/post/mzevpbxsyvdrcyjuwhuu@forum.dlang.org), but if you compile both files at the same time (i.e. `gdc file1.d file2.d`) it does inline.

DmitryOlshansky:
> but @tsbockman and I both found that add(s|u) is not inlined when imported. If the function is duplicated in the file using it, it is inlined.

A well-known problem that I find critically important. A workaround is to turn functions into 0-arg templates.

WalterBright:
Auto-merge toggled on

burner:
thanks

PR #1309, opened 2015-06-30
simplify rt.util.string

WalterBright:
Dramatically simplify ToTempString by removing unnecessary templates. Generalized it by making radix a parameter. Added @nogc. Made it @safe instead of @trusted.


WalterBright:
needs https://github.com/D-Programming-Language/dmd/pull/4786

WalterBright:
Moved it to core.util.string so that Phobos can use it.

MartinNowak:
Auto-merge toggled on

PR #1311, opened 2015-06-30
core.cpuid - make ssse3 immutable

WalterBright:
These properties are immutable. They should be made immutable, so that users of them won't need to use static constructors in order to have pure functions, such as in std.digest.sha.

Doing just one to see how it goes.

### DIFF
--- a/src/core/cpuid.d
+++ b/src/core/cpuid.d
@@ -54,11 +54,6 @@
  * Source:    $(DRUNTIMESRC core/_cpuid.d)
  */
 
-/*          Copyright Don Clugston 2007 - 2009.
- * Distributed under the Boost Software License, Version 1.0.
- *    (See accompanying file LICENSE or copy at
- *          http://www.boost.org/LICENSE_1_0.txt)
- */
 module core.cpuid;
 
 @trusted:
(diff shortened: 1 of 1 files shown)


schveiguy:
LGTM

MartinNowak:
> Doing just one to see how it goes.

What do you mean by "how it goes"?
It's unlikely that anyone will follow-up soon and we really don't need more half done stuff in D.

WalterBright:
I mean this is a trivial change, easy to review. If it goes in, then a followup will do the rest of them, which will be a lot of diffs and much harder to review, and not worth the effort if this one is not accepted.

schuetzm:
LGTM

DmitryOlshansky:
@WalterBright it's far simpler then discussing IMHO, just make 'em all immutable? Thanks!

src/core/cpuid.d
```diff
@@ -143,7 +143,7 @@ public:
     /// Is SSE3 supported?
     bool sse3()         {return (miscfeatures&SSE3_BIT)!=0;}
     /// Is SSSE3 supported?
-    bool ssse3()        {return (miscfeatures&SSSE3_BIT)!=0;}
```
DmitryOlshansky: BTW - why not just make miscfeature immutable and leave functions as is? Is it broken cross-module inlining again? If it's the latter just make them templates...

schveiguy: > why not just make miscfeature immutable and leave functions as is

Really good idea. You could make the storage a single immutable struct, then set the value from the shared static ctor based on calling some function.

WalterBright:
> it's far simpler then discussing IMHO, just make 'em all immutable?

Because:

> why not just make miscfeature immutable and leave functions as is?

and:

> You could make the storage a single immutable struct, then set the value from the shared static ctor based on calling some function.

I don't want to do all the work and then have to rework it.

Now, to answer the question about changing the design from immutable variables to functions:

I see zero advantage to using a function that returns an immutable compared with just returning the bool immutable. Yes, the compiler can inline it and it has the same result - but why? This does not need a complex solution to a completely trivial problem.

WalterBright:
> A (small) improvement in binary size?

Would be overshadowed by more code generated.

> There is a lot of code that uses the features as actual functions (i.e. with parens) instead of as properties.

A more persuasive point, though I'm not seeing any such usages in Phobos/Druntime, though I did not look exhaustively.

WalterBright:
> Just make everything set from the static ctor immutable

That's a wholesale rewrite of the module, since immutables can only be set from within a static ctor.

WalterBright:
Making an extra set of variables reduced the amount of rewriting.

DmitryOlshansky:
> Would be overshadowed by more code generated.

As long as it stays inlined probably not. On the contrary - it won't generate code that is not used. Where as initializing every combination of flags in ctor always adds code .
```d
import core.cpuid;

extern(C) void fputc(int c);

void main()
{
  fputc(sse3() ? 0 : 1);
}
```
Here is the current disassembly:
```
__Dmain         proc near               ; DATA XREF: _main+1o
COMDAT01:00000020                 push    eax
COMDAT01:00000021                 mov     eax, dword ptr ds:_D4core5cpuid12miscfeaturesk
COMDAT01:00000026                 and     eax, 1
COMDAT01:0000002B                 xor     eax, 1
COMDAT01:00000030                 push    eax
COMDAT01:00000031                 call    _fputc
COMDAT01:00000036                 add     esp, 4
COMDAT01:00000039                 xor     eax, eax
COMDAT01:0000003B                 pop     ecx
COMDAT01:0000003C                 retn
COMDAT01:0000003C __Dmain         endp
```
I'd say quite good. Getting rid of unused bodies of inlined functions is separate matter with the toolchain that must be sovled anyway.

DmitryOlshansky:
@WalterBright  If you move all functions to be local, inside static ctor would it work?

WalterBright:
> I'd say quite good.

It's more bytes and more code than if it was just testing a bool at the usage site, where if it is not a fast switch between options the user is going to code yet another variable to make it a bool, which is what has been happening.

> If you move all functions to be local, inside static ctor would it work?

No.

DmitryOlshansky:
@WalterBright Then given the situation I agree, making an extra set of immutables is the way.

schveiguy:
> Oh, and to point out the obvious: This change is never going to fly, as it breaks backwards compatibility.

These are all `@property` functions. If they are using them with parentheses, then they aren't using them right :)

schveiguy:
Shouldn't variables like `features` be removed as globals?

WalterBright:
> Shouldn't variables like  features  be removed as globals?

As I mentioned already, a total refactoring of the code in core.cpuid implies breaking it, as there is no test farm with all those CPUs on it. So I am extremely reluctant to mess with it.

schveiguy:
The variables are all private globals. It should be pretty easy to search through the file and make sure it's not used anywhere else.

However, I don't know that it makes a huge deal, we can remove them later.

WalterBright:
Right, they are private globals, they don't leak out anywhere. I think you underestimate the amount of refactoring needed to eliminate them, and there is NO TEST SUITE for this. It's not worth the risk, especially since there is zero effect on user code.

This PR is blocking other improvements. It needs to get pulled.

DmitryOlshansky:
Auto-merge toggled on

WalterBright:
thanks!

MartinNowak:
Why the did we introduce a huge amount of bools for this instead of keeping `_miscfeature & FEAT_FLAG`? Once it was clear that the functions can't be turned into booleans, we should made `_miscfeature` immutable. There is no reason to precompute all the bittests, and now we have twice as much code to maintain and edit.

schveiguy:
@MartinNowak you can fix it if you want, the storage details for the bits are all private.

PR #1310, opened 2015-06-30
[REG2.068a] Issue 14746 - Behavior change with struct destructor and alias this

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=14746

PR #1312, opened 2015-07-01
[REG2.068a] Issue 14746 - Behavior change with struct destructor and alias this

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=14746

On stable branch instead of master (#1310)


MartinNowak:
thx

PR #1313, opened 2015-07-01
fixup Issue 14746

MartinNowak:
- also fix _postblitRecurse
- replace recursive templates with _xdtor and _xpostblit

[Issue 14746 – [REG2.068a] Behavior change with struct destructor and alias this](https://issues.dlang.org/show_bug.cgi?id=14746)

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2228,12 +2228,10 @@ private void _destructRecurse(S)(ref S s)
 {
     import core.internal.traits : hasElaborateDestructor;
 
-    static if (__traits(hasMember, S, "__dtor"))
-    {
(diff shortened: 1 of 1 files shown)


MartinNowak:
As discussed [here](https://github.com/D-Programming-Language/druntime/pull/1181#issuecomment-102780737) the compiler primitives should be used.

9rnsr:
LGTM.

9rnsr:
Auto-merge toggled on

JakobOvrum:
Sorry for being late; thanks for following up with this!

schuetzm:
This change introduced a regression: https://issues.dlang.org/show_bug.cgi?id=15016

PR #1314, opened 2015-07-02
improve diagnostics for circular static construction failure

WalterBright:
This'll say whether it's global or shared constructors, and marks with a '*' which modules in the cycle have a ctor/dtor.


WalterBright:
See also: https://issues.dlang.org/show_bug.cgi?id=14517

schveiguy:
Auto-merge toggled on

PR #1315, opened 2015-07-04
make checkwhitespace runnable on windows

rainers:
- do not abuse MAKEFILES, it is a predefined variable in some versions of make
- make grep overloadable (grep installed with git on windows is too old to support -P)
- add -U to the command line, otherwise \r is not detected when run under windows

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -207,12 +207,13 @@ test/%/.run: test/%/Makefile
 		DRUNTIME=$(abspath $(DRUNTIME)) DRUNTIMESO=$(abspath $(DRUNTIMESO)) QUIET=$(QUIET) LINKDL=$(LINKDL)
 
 #################### test for undesired white spaces ##########################
-MAKEFILES = $(filter mak/% %.mak,$(MANIFEST))
-NOT_MAKEFILES = $(filter-out $(MAKEFILES),$(MANIFEST))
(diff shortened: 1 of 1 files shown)


DmitryOlshansky:
@rainers I look at patterns and the may just work with something other then `-P` - P(erl)-compatible regex. Try `-E` - extended regex, it's more then enough here. Msys git's grep on Windows does seem to work fine with it.

rainers:
It didn't really worked with -E or other options, see https://github.com/D-Programming-Language/phobos/pull/3474#issuecomment-118595475

Already too much time spent with grep, 5 minutes updating checkwhitespace.d yield much better results. Unfortunately, there is no D compiler available that can build it when building druntime by the test servers (HOST_DC does not seem to be set).

DmitryOlshansky:
I suggest to just disable it on Windows for now. It's enough to catch it by some of auto-tester fleet, no need to sweat it.

rainers:
The test isn't running on windows, but for OSs that use posix.mak. OSX and FreeBSD don't seem to have a recent version of grep, though (2.5.4 from 2009 works for me on Windows).

DmitryOlshansky:
Let it be linux-only then, we still catch it either way.

rainers:
Ok, now only grepping on linux.

PR #1317, opened 2015-07-07
fix Issue 11414 - druntime should run debug unittest

MartinNowak:
- change druntime's makefile structure to match phobos
- define ROOT folder as generated/$(OS)/$(BUILD)/$(MODEL)
- unittest rule recursively invokes debug/release unittest
- factor out common.mak from test/*/Makefile and test debug/release

[Issue 11414 – druntime should run debug unittest](https://issues.dlang.org/show_bug.cgi?id=11414)

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -7,6 +7,20 @@ QUIET:=@
 
 include osmodel.mak
 
+# Default to a release built, override with BUILD=debug
+ifeq (,$(BUILD))
+BUILD_WAS_SPECIFIED=0
+BUILD=release
+else
+BUILD_WAS_SPECIFIED=1
+endif
+
+ifneq ($(BUILD),release)
+    ifneq ($(BUILD),debug)
+        $(error Unrecognized BUILD=$(BUILD), must be 'debug' or 'release')
+    endif
+endif
+
 DMD=../dmd/src/dmd
 INSTALL_DIR=../install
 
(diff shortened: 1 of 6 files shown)


MartinNowak:
requires https://github.com/D-Programming-Language/phobos/pull/3478

MartinNowak:
Druntime debug tests will fail until [Issue 14779 – incorrect addressing of arguments in require/in-contract](https://issues.dlang.org/show_bug.cgi?id=14779) gets fixed.

MartinNowak:
This will pass now when the phobos change is merged.
It's a shame that we still can't specify interrepo dependencies of PRs for the auto-tester.

jmdavis:
Unfortunately, it looks like this broke the build on FreeBSD:

https://github.com/D-Programming-Language/druntime/pull/1322

posix.mak
```diff
@@ -23,21 +37,40 @@ else
...
 endif
 
-DFLAGS=$(MODEL_FLAG) -conf= -O -release -dip25 -inline -w -Isrc -Iimport $(PIC)
-UDFLAGS=$(MODEL_FLAG) -conf= -O -release -dip25 -w -Isrc -Iimport $(PIC)
 DDOCFLAGS=-conf= -c -w -o- -Isrc -Iimport -version=CoreDdoc
 
-CFLAGS=$(MODEL_FLAG) -O $(PIC)
+# Set CFLAGS
+CFLAGS=
+ifneq (,$(filter cc% gcc% clang% icc% egcc%, $(CC)))
```
CyberShadow: @MartinNowak Please no whitelists! What is this for? This broke building 32-bit Phobos on a 64-bit machine for me because the model flag is no longer passed to the gcc wrapper (required to work around issue 12572). Otherwise at least please use `basename` to allow setting CC to an absolute path.

CyberShadow: @MartinNowak https://issues.dlang.org/show_bug.cgi?id=15012

MartinNowak: I didn't invent this but took the code from https://github.com/D-Programming-Language/phobos/blob/88fc63d8d3d47ded732e801d4ccacec21eccb8f2/posix.mak#L96? So why did this work with your phobos build? You can still set `CFLAGS` explicitly.

CyberShadow: > So why did this work with your phobos build? You can still set CFLAGS explicitly.

I don't know, maybe the program has to reference the symbols in the .c files to trigger the linker error?

> You can still set CFLAGS explicitly.

Not a very workable solution as 1) I'd have to do it for every account / Digger instance 2) The required value of CFLAGS likely changes across D's history so it'd get in the way of bisecting.

PR #1319, opened 2015-07-07
core.memory add user notes

WalterBright:
For reference: http://blog.experimentalworks.net/2015/07/lessons-learned-writing-a-filesystem-in-d/


andralex:
Thanks!

MartinNowak:
I made a follow-up #1320 to explicitly mention detached threads.

PR #1321, opened 2015-07-09
Make use of invalid MemoryOrder semantics a compile-time error

ibuclaw:
Also removes private ``isHoistOp`` and ``isSinkOp`` templates.

@MartinNowak - we might be able to do away with the ``cmpxchg`` entirely given that load/stores on x86 should always have the desired semantics.

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -105,7 +105,8 @@ version( CoreDdoc )
     /**
      * Loads 'val' from memory and returns it.  The memory barrier specified
      * by 'ms' is applied to the operation, which is fully sequenced by
-     * default.
+     * default.  Valid memory orders are MemoryOrder.raw, MemoryOrder.acq,
+     * and MemoryOrder.seq.
      *
      * Params:
      *  val = The target variable.
@@ -122,6 +123,8 @@ version( CoreDdoc )
     /**
      * Writes 'newval' into 'val'.  The memory barrier specified by 'ms' is
      * applied to the operation, which is fully sequenced by default.
+     * Valid memory orders are MemoryOrder.raw, MemoryOrder.rel, and
+     * MemoryOrder.seq.
      *
      * Params:
      *  val    = The target variable.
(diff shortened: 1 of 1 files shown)


ZombineDev:
A couple of weeks ago I was reading the code and those `isHoistOp` and `isSinkOp`  looked nonsensical to me. So, this change LGTM.

ZombineDev:
BTW, shouldn't all atomic operations have
```D
static assert (__traits(isPOD, T))
//...
```
in their bodies? As far as I can see, this check is only done for `atomicLoad`.

ZombineDev:
Also, can you please update the docs for MemoryOrder, to make them reflect that `atomicLoad` cannot be used with `MemoryOrder.rel` and vice versa?

ibuclaw:
> BTW, shouldn't all atomic operations have static assert (__traits(isPOD, T)) in their bodies?

I guess so, it makes more sense to assert if you attempt to call ``atomicStore`` with a struct that has a cpctor.

(Actually, I will start with just replacing the ``static if``'s with ``static assert``)

ibuclaw:
FYI, this will turn into a second PR, but I'm pretty positive that we can loose the ``lock cmpxchg`` in both ``atomicLoad`` and ``atomicStore`` if we are using a ``MemoryOrder.seq``.

However, an ``mfence`` would still be needed in ``atomicLoad`` to satisfy fully sequenced semantics.

ibuclaw:
Updated comments and replaced static ifs with asserts.

ibuclaw:
Actually, I won't bother touching the assembly, as DMD uses a non-standard calling convention.  But FYI, atomics are potentially heavier than what they need to be on x86.

https://www.cl.cam.ac.uk/~pes20/cpp/cpp0xmappings.html

ibuclaw:
> By the way, how do you treat raw operations in GDC? Like memory_order_relaxed?

Yes.

> Why would a seq load still need a mfence, given that we (should) emit the store as lock xchg?

It's possible to alternately use a ``mov`` and ``mfence`` instead of ``lock xchg``.  This may be subjective to CPU models, but as I understand it, is faster as it does not trigger a redundant memory read like the ``xchg`` which will take time.  It's also as safe as ``lock xchg`` because x86 guarantees strict ordering of writes between threads.

ibuclaw:
> We were talking about loads, not stores? ;)

Oh, I must have confused myself then.  :-)

Don't need it for loads, only need ``lock xchg``/``mfence`` for stores.

PR #1322, opened 2015-07-10
Disable debug build of unit tests.

jmdavis:
The shared library unit tests are currently failing on FreeBSD when
they're built in debug mode, so this disables the recently added debug
unit test build so that the autotester passes again. The debug unit test
build should be re-enabled once the problem with FreeBSD is fixed.

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=90669&dataid=625858

The release unit test build was failing on FreeBSD 9 and 10 already, but the autotester is on 8 still, where the release unit test build seems to pass, but it looks like they all fail the unit tests in the debug build (I've verified that with 10 and I assume that 9 does, since 8 clearly does given the failure on the autotester and the fact that 10 fails). It looks like the failing tests are all in the tests for shared library support, so my guess is that the shared library support for FreeBSD has definite issues on all 3 versions. It's just that the issues don't trigger as many test failures with 8, since the unit test release build works on 8 but not 9 or 10. But without further investigation, we obviously don't know exactly what's wrong. Regardless, we clearly can't leave the debug unit test build enabled right now, since it causes the autotester to fail on FreeBSD.

Pull# 1317 added a debug build for the unit tests, and this does not undo the changes that it had to do to the makefiles to make that work, but it does make it so that the debug unit test build is not built as part of the normal unit test build.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -196,7 +196,9 @@ ifeq (1,$(BUILD_WAS_SPECIFIED))
 unittest : $(UT_MODULES) $(addsuffix /.run,$(ADDITIONAL_TESTS))
 	@echo done
 else
-unittest : unittest-debug unittest-release
+# The unit tests currently fail on FreeBSD in debug mode
+#unittest : unittest-debug unittest-release
+unittest : unittest-release
 unittest-%:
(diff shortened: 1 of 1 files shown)


jmdavis:
Auto-merge toggled on

MartinNowak:
Will check ASAP.

PR #1323, opened 2015-07-10
[reg] fix Issue 14750 - druntime/test/coverage was added to druntime, but n…

WalterBright:
…ot to the MANIFEST - zip file broken again

https://issues.dlang.org/show_bug.cgi?id=14750

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -207,7 +207,7 @@ test/%/.run: test/%/Makefile
 		DRUNTIME=$(abspath $(DRUNTIME)) DRUNTIMESO=$(abspath $(DRUNTIMESO)) QUIET=$(QUIET) LINKDL=$(LINKDL)
 
 #################### test for undesired white spaces ##########################
-CWS_MAKEFILES = $(filter mak/% %.mak,$(MANIFEST))
+CWS_MAKEFILES = $(filter mak/% %.mak %/Makefile,$(MANIFEST))
 NOT_MAKEFILES = $(filter-out $(CWS_MAKEFILES),$(MANIFEST))
 GREP = grep
 
@@ -222,9 +222,13 @@ detab:
 	detab $(MANIFEST)
 	tolf $(MANIFEST)
 
+
+gitzip:
+	git archive --format=zip HEAD > druntime.zip
+
 zip: druntime.zip
 
-druntime.zip: $(MANIFEST) $(IMPORTS)
+druntime.zip: $(MANIFEST)
 	rm -rf $@
 	zip $@ $^
 
--- a/mak/MANIFEST
+++ b/mak/MANIFEST
@@ -291,4 +291,37 @@ MANIFEST=\
 	src\rt\typeinfo\ti_void.d \
 	src\rt\typeinfo\ti_wchar.d \
 	\
-	src\etc\linux\memoryerror.d
+	src\etc\linux\memoryerror.d \
+	\
+	test\coverage\Makefile \
+	test\coverage\src-basic.lst.exp \
+	test\coverage\src-merge.lst.exp \
+	test\coverage\src-merge_true.lst.exp \
+	test\coverage\src\basic.d \
+	test\coverage\src\merge.d \
+	test\coverage\src\merge_true.d \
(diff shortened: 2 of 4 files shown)


DmitryOlshansky:
Anything we can do to prevent this? Add zip packaging tests to auto-tester?

rainers:
We could remove MANIFEST and just use the output of `git ls-tree --name-only -r HEAD` instead.

DmitryOlshansky:
> We could remove MANIFEST and just use the output of git ls-tree --name-only -r HEAD instead.

Yeah, would you go for a pull request or should I?

posix.mak
```diff
@@ -222,9 +222,13 @@ detab:
 	detab $(MANIFEST)
 	tolf $(MANIFEST)
 
+
+gitzip:
```
DmitryOlshansky: Nice

rainers:
> Yeah, would you go for a pull request or should I?

Please do ;-)

I guess @WalterBright is the only one using the zip file. I've never even heard of a version of zip named `zip32`. So it's up to him to decide if it works in his environment.

There are also the detab and checkwhitespace targets using MANIFEST, but fortunately they are only in posix.mak, so we won't have to fight DM make for getting the file list into a make variable.

rainers:
For the test failures, Makefile should be excluded from tab checks by adding `%/Makefile` to the filter in CWS_MAKEFILES.

DmitryOlshansky:
> Please do ;-)

E-hm if only I knew how to make that zip32 accept a file with names...

WalterBright:
The zip file is for systems that do not have git installed on them, so replacing MANIFEST with git commands doesn't work.

WalterBright:
The failure is on Linux:
```
test/shared/Makefile:58:	$(QUIET)$(DMD) -fPIC -shared $(DFLAGS) -of$@ $< $(LINKDL)
test/shared/Makefile:61:	rm -rf $(ROOT)
make: *** [checkwhitespace] Error 1
```
what does that mean?

DmitryOlshansky:
@WalterBright these 2 lines in Makefile contain tabs. Just follow Rainer's suggestion and `%/Makefile` to the filter in posix.mak to ignore makefiles.

WalterBright:
> For the test failures, Makefile should be excluded from tab checks by adding  %/Makefile  to the filter in CWS_MAKEFILES.

Where is CWS_MAKEFILES ? It isn't in druntime.

rainers:
> Where is CWS_MAKEFILES ? It isn't in druntime.

I suspect you are on an older branch here. The check for whitespaces has been merged a couple of days ago.

rainers:
> The zip file is for systems that do not have git installed on them, so replacing MANIFEST with git commands doesn't work.

Moving to a system without git should be no problem. To move a zip file back, you could either "freshen" an existing zip (e.g. -f with info-zip) or just run `make clean` before zipping the whole directory.

WalterBright:
I never trust freshen, as sometimes files get removed from git. `make clean` only removes files it knows about.

WalterBright:
@rainers again, where is CWS_MAKEFILES? I updated my copy of druntime from master, I don't see it.

Also, this sort of error message is embarrassingly bad:
```
test/shared/Makefile:61:	rm -rf $(ROOT)
make: *** [checkwhitespace] Error 1
```

rainers:
> @rainers again, where is CWS_MAKEFILES? I updated my copy of druntime from master, I don't see it.

https://github.com/WalterBright/druntime/blob/d60cf660495a3ef33c3d607314d7aff159acba12/posix.mak#L210

> Also, this sort of error message is embarrassingly bad:

This is the result of running grep. checkwhitespace is not available and cannot be compiled because the build server does not provide a D compiler (with phobos) when building druntime.

rainers:
> I never trust freshen, as sometimes files get removed from git.

You won't notice anyway because you don't have git installed on the system where you use it ;-)

WalterBright:
Oh, I though CWS_MAKEFILES was supposed to be a file.

DmitryOlshansky:
Auto-merge toggled on

PR #1324, opened 2015-07-16
[REG2.066] Issue 14626 - byValue doesn't work with inout AA

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=14626

Until 2.065, compiler had substituted all `inout` qualifiers in the `Key` and `Value` types to `const`, then those had passed to the template struct `AssociativeArray`.

https://github.com/D-Programming-Language/dmd/blob/v2.065.0/src/mtype.c#L4897

This change emulates that.

---

Additionally add an overload of `bykeyValue` function for AA pointer types.

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -53,6 +53,38 @@ template Unqual(T)
     }
 }
 
+// Substitute all `inout` qualifiers that appears in T to `const`
+template substInout(T)
+{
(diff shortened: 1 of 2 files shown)


MartinNowak:
Should be stable branch to go into the release.

9rnsr:
@MartinNowak I opened #1326.

PR #1325, opened 2015-07-17
add staticError helper to throw @nogc Errors

MartinNowak:


DmitryOlshansky:
Hopefully throwing Error while an Error is in flight shouldn't happen.
After all Error is not recoverable so LGTM

src/core/exception.d
```diff
@@ -649,3 +643,21 @@ extern (C)
...
+private void[128] _store;
+
+// only Errors for now as those are rarely chained
+private T staticError(T, Args...)(auto ref Args args)
+    if (is(T : Error))
+{
+    // pure hack, what we actually need is @noreturn and allow to call that in pure functions
+    static T get()
+    {
+        _store[0 .. __traits(classInstanceSize, T)] = typeid(T).init[];
```
schuetzm: Should probably add:
```d
static assert(__traits(classInstanceSize, T) <= _store.length),
              T.stringof ~ " is too large for staticError());
```

DmitryOlshansky:
@MartinNowak - address comment by @schuetzm and it should be good to go

src/core/exception.d
MartinNowak: It would already fail to compile, but with a worse error message.

MartinNowak:
Done

DmitryOlshansky:
Auto-merge toggled on

src/core/exception.d
```diff
@@ -649,3 +643,24 @@ extern (C)
...
+    if (is(T : Error))
+{
+    // pure hack, what we actually need is @noreturn and allow to call that in pure functions
+    static T get()
+    {
+        static assert(__traits(classInstanceSize, T) <= _store.length,
+                      T.stringof ~ " is too large for staticError()");
+
+        _store[0 .. __traits(classInstanceSize, T)] = typeid(T).init[];
+        return cast(T) _store.ptr;
```
jpf91: Is this safe regarding alignment of the `T`/`Error` type?

jpf91: ping @MartinNowak

MartinNowak: You might be right, will check.

MartinNowak: It doesn't seem to be a problem, the array is 16-byte aligned and we don't align classes ([see](https://github.com/D-Programming-Language/druntime/blob/dbd72a2f51d7b70449367947ca556aebdd3ee063/src/rt/lifetime.d#L71)).

CyberShadow:
https://issues.dlang.org/show_bug.cgi?id=14993

src/core/exception.d
klickverbot: Why would 16 byte alignment be guaranteed for the array?

PR #1327, opened 2015-07-18
Issue 2013 - interface to interface dynamic cast is incorrect in some cases

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=2013


9rnsr:
This is a preparation change for the compiler fix https://github.com/D-Programming-Language/dmd/pull/4819.

rainers:
Auto-merge toggled on

PR #1329, opened 2015-07-19
More white space fixes

rainers:
- remove tabs and trailing white space
- add missing files to MANIFEST
- don't rely on MANIFEST for the white space test

### DIFF
--- a/mak/MANIFEST
+++ b/mak/MANIFEST
@@ -112,6 +112,7 @@ MANIFEST=\
 	src\core\sys\osx\mach\thread_act.d \
 	\
 	src\core\sys\osx\sys\cdefs.d \
+	src\core\sys\osx\sys\event.d \
 	src\core\sys\osx\sys\mman.d \
 	\
 	src\core\sys\posix\config.d \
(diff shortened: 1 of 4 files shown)


rainers:
The test failure is caused by a bad unittest in std.random that is bound to fail with some probability.

yebblies:
> The test failure is caused by a bad unittest in std.random that is bound to fail with some probability.

You can trigger a re-run of the failing test by going onto the [run's page](https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=1655387&isPull=true) and clicking the 'deprecate' link.

rainers:
> You can trigger a re-run of the failing test by going onto the run's page and clicking the 'deprecate' link.

Thanks, that helped.

WalterBright:
@rainers All test suite heisenbugs should be submitted to bugzilla so they get fixed.

rainers:
> @rainers All test suite heisenbugs should be submitted to bugzilla so they get fixed.

It is already reported: https://issues.dlang.org/show_bug.cgi?id=13853

WalterBright:
@rainers thank you!

DmitryOlshansky:
LGTM

9rnsr:
Auto-merge toggled on

9rnsr:
Seems no one has complaint with this PR.

PR #1330, opened 2015-07-20
Post-processing for Issue 2013 - Remove temporary branch till the compiler migration

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=2013


src/rt/cast_.d
```diff
@@ -84,37 +84,6 @@ int _d_isbaseof2(ClassInfo oc, ClassInfo c, ref size_t offset)
...
-            if (_d_isbaseof2(iface.classinfo, c, offset))
-            {
-                offset += iface.offset;
-                return true;
-            }
-        }
-      }
-      else
-      {
         // Bugzilla 2013: After the compiler is actually fixed, we can simply use
```
rainers: I'd recommend to change this comment, too. IMO it is no good to refer to bugs that no longer exist.

9rnsr: I updated the comment.
A correct bugzilla number in comment is a good note to know why the code is formed as so, if the code is not yet mature enough.

rainers:
Auto-merge toggled on

PR #1331, opened 2015-07-20
Separate pthreads APIs by OS

joakim-noah:
[According to Walter, all OS APIs should be separated out into version blocks](https://github.com/D-Programming-Language/druntime/pull/1216#issuecomment-121882938), not common blocks like this, that mix potentially different OS APIs together.  I'll submit more PRs for the remaining common blocks in druntime.

### DIFF
--- a/src/core/sys/posix/pthread.d
+++ b/src/core/sys/posix/pthread.d
@@ -440,6 +440,9 @@ else
     static assert(false, "Unsupported platform");
 }
 
+version( CRuntime_Glibc )
+{
+    // These declarations have not been checked.
 int pthread_atfork(void function(), void function(), void function());
 @nogc {
     int pthread_atfork(void function() @nogc, void function() @nogc, void function() @nogc);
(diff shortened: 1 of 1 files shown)


src/core/sys/posix/pthread.d
```diff
@@ -261,14 +261,70 @@ else
...
 
-int pthread_atfork(void function(), void function(), void function());
-int pthread_attr_destroy(pthread_attr_t*);
-int pthread_attr_getdetachstate(in pthread_attr_t*, int*);
-int pthread_attr_getschedparam(in pthread_attr_t*, sched_param*);
-int pthread_attr_init(pthread_attr_t*);
-int pthread_attr_setdetachstate(pthread_attr_t*, int);
-int pthread_attr_setschedparam(in pthread_attr_t*, sched_param*);
-int pthread_cancel(pthread_t);
+version( CRuntime_Glibc )
```
WalterBright: Doesn't FreeBSD use glibc too?

joakim-noah: No, [glibc is the GNU C library](https://en.wikipedia.org/wiki/GNU_C_Library) and is only used on desktop/server linux.  The BSDs have their own libc variants, as do the other UNIX spinoffs.

joakim-noah:
Rebased and updated.

src/core/sys/posix/pthread.d
```diff
@@ -450,6 +528,71 @@ int pthread_atfork(void function(), void function(), void function());
...
+    int pthread_attr_destroy(pthread_attr_t*);
+    int pthread_attr_getdetachstate(in pthread_attr_t*, int*);
+    int pthread_attr_getschedparam(in pthread_attr_t*, sched_param*);
+    int pthread_attr_init(pthread_attr_t*);
+    int pthread_attr_setdetachstate(pthread_attr_t*, int);
+    int pthread_attr_setschedparam(in pthread_attr_t*, sched_param*);
+    int pthread_cancel(pthread_t);
+      }
+}
+else version( CRuntime_Uclibc )
```
rracariu: Actual identifier is `CRuntime_UClibc` with a capital 'C'

src/core/sys/posix/pthread.d
```diff
@@ -440,6 +440,144 @@ else
...
+    int pthread_attr_destroy(pthread_attr_t*);
+    int pthread_attr_getdetachstate(in pthread_attr_t*, int*);
+    int pthread_attr_getschedparam(in pthread_attr_t*, sched_param*);
+    int pthread_attr_init(pthread_attr_t*);
+    int pthread_attr_setdetachstate(pthread_attr_t*, int);
+    int pthread_attr_setschedparam(in pthread_attr_t*, sched_param*);
+    int pthread_cancel(pthread_t);
+      }
+}
+else version( CRuntime_UClibc )
```
ibuclaw: Shouldn't C runtime versions take precedence over OS versions? Then again probably only applies realistically to glibc where there are active ports to FreeBSD and Solaris.

joakim-noah: Yes, you're right, that's why Glibc is always first. However, realistically we don't have to worry about that since linux is the only kernel that uses so many other libc's. If and when we support kFreeBSD or some other new libc/kernel combo, someone will have to do the work of splitting out the FreeBSD libc declarations from the kernel, which are currently bundled together, and then the order won't matter, as kFreeBSD will select for Glibc rather the separated FreeBSD libc.

ibuclaw: Well compilers already support kfreebsd and ksolaris _today_, so thinking about future compatibility over making it someone else's job is preferable 😉

joakim-noah: It's not going to work anyway, as I said, because someone will have to separate out the BSD/Solaris version blocks at that point.  No point in worrying about the order here when it's not going to ever make a difference.

joakim-noah:
Been sitting forever, closing.

PR #1333, opened 2015-07-25
fix FBSD shared library issues

MartinNowak:
- add tests for module collision check
- fix copy relocation check so it can deal with
  __bss_start/_end being defined in a DSO
- workaround FBSD bug ([Issue 14824](https://issues.dlang.org/show_bug.cgi?id=14824))
- reenable druntime-debug unittests

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -196,9 +196,7 @@ ifeq (1,$(BUILD_WAS_SPECIFIED))
 unittest : $(UT_MODULES) $(addsuffix /.run,$(ADDITIONAL_TESTS))
 	@echo done
 else
-# The unit tests currently fail on FreeBSD in debug mode
-#unittest : unittest-debug unittest-release
-unittest : unittest-release
+unittest : unittest-debug unittest-release
 unittest-%:
 	$(MAKE) -f $(MAKEFILE) unittest OS=$(OS) MODEL=$(MODEL) DMD=$(DMD) BUILD=$*
 endif
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -837,6 +837,13 @@ const(void)[] getCopyRelocSection() nothrow
     return bss_start[0 .. bss_size];
 }
 
+/**
+ * Check for module collisions. A module in a shared library collides
+ * with an existing module if it's ModuleInfo is interposed (search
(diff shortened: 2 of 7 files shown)


jmdavis:
Well, I don't know if this was supposed to fix all of the shared library bugs that have been reported for FreeBSD or not, but these changes seem to fix https://issues.dlang.org/show_bug.cgi?id=14776. However, I'm still hitting https://issues.dlang.org/show_bug.cgi?id=14663 on my FreeBSD 10 system, and it looks like the tests are failing on several platforms on the autotester (including on Linux) - around the load_mod_collision test it looks like, which is after the link_linkdep test that is failing on my FreeBSD 10 system.

So, as far as the systems on the auto tester go, you seem to have fixed the issue with FreeBSD and debug builds while creating a new one, and the problem that was failing on FreeBSD 10 in release builds is still failing (and debug builds now hit the same failure instead of the one they were hitting before).

But _thank you_ for looking into this. I was afraid that I was going to have to start looking into the shared library issues, since I've been using FreeBSD of late, but I know nothing about how the shared library support works. And technically, FreeBSD 8.x will be at End-of-Life come August, so without shared library support getting fixed in FreeBSD 9 and/or 10, we soon won't actually be able to successfully run the druntime unit tests on a supported FreeBSD system (which of course doesn't say good things about how well it actually works on those systems). :(

MartinNowak:
> around the load_mod_collision test it looks like,

The test itself works fine, but it seems that the tester doesn't like the pre-pipe command fail with a segfault (`assert(0)`).

MartinNowak:
Found the linux issue. The distribution on the auto-tester uses `--as-needed` by default, and b/c it isn't used by link_mod_collision lib.so was not linked.

MartinNowak:
Please don't let that bitrot, it's kinda difficult to load this topic into my head.

DmitryOlshansky:
Sadly have no idea about this part of druntime. Everything sans the first commit is simple and makes sense. The bulk of changes for copy relocations better be reviewed by somebody competent.

DmitryOlshansky:
Auto-merge toggled on

DmitryOlshansky:
It's been hanging for a while here and on trello so I guess nobody can be bothered to double check it

PR #1334, opened 2015-07-27
Add constants AF_RXRPC and PF_RXRPC on Linux

todayman:
See title.  Pretty simple and self-explanatory.

### DIFF
--- a/src/core/sys/linux/sys/socket.d
+++ b/src/core/sys/linux/sys/socket.d
@@ -6,6 +6,8 @@
  */
 module core.sys.linux.sys.socket;
 
+public import core.sys.posix.sys.socket;
+
 version(linux):
 extern(C):
 @nogc:
--- a/src/core/sys/linux/sys/socket.d
+++ b/src/core/sys/linux/sys/socket.d
@@ -1,0 +1,17 @@
+/**
+ * D header file for GNU/Linux.
+ *
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   Paul O'Neil
+ */
+module core.sys.linux.sys.socket;
+
+version(linux):
+extern(C):
+@nogc:
+nothrow:
+
+enum {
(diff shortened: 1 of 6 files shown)


DmitryOlshansky:
Since these are linux specific no point having them on _posix_ file I guess.
@todayman - could you please follow the convention of putting things into `core.sys.linux.*` if they are linux-specific?

todayman:
There isn't a core.sys.linux.sys.socket.d.  Should I go ahead and create or do you think there's a better place?

DmitryOlshansky:
Time to create ore.sys.linux.sys.socket.d

todayman:
The autotester has been stuck at Pass: 8, Pending: 2 for quite a while now (> 1 week).  What does that mean?  Clicking on the details link shows that there are some failures on Win32 - if those are a real problem I have no way to test / debug those and would appreciate help.

yebblies:
The autotester prioritizes recently-updated pulls, and on some of the slower platforms this means the tester never gets to this pull.  If you rebase on top of master and force-push over this branch it will bump it back up to the top.

todayman:
@DmitryOlshansky core.sys.linux.sys.socket.d is made and the build looks OK.  Is this good to go?

schveiguy:
LGTM

src/core/sys/linux/sys/socket.d
```diff
@@ -0,0 +1,17 @@
...
+ * Authors:   Paul O'Neil
+ */
+module core.sys.linux.sys.socket;
+
+version(linux):
+extern(C):
+@nogc:
+nothrow:
+
+enum {
```
MartinNowak: It should publicly `import core.sys.posix.sys.socket;` so it functions as an overlay.

todayman:
@MartinNowak public import added.

schveiguy:
Still looks good, I assume Martin only had that one issue which is now fixed.

schveiguy:
Auto-merge toggled on

PR #1335, opened 2015-07-30
remove unused _destroy/postblitRecurse for static arrays

MartinNowak:


9rnsr:
Unfortunately we cannot do this, because `_postblitRecurse` is actually used in the `std.conv.emplace` implementation...

MartinNowak:
I didn't knew one could emplace into a static array.

PR #1336, opened 2015-08-03
Fix issue 14863 - CLOCK_BOOTTIME should be optional to support <2.6.39 kernels

schveiguy:
https://issues.dlang.org/show_bug.cgi?id=14863

I realized, MonoTime.currTime is already checking for 0 and asserting (and any use of 0 for ticksPerSecond will have a divide by zero error anyway), we should simply defer to that. I made a note in the code so people don't add the asserts back in later.

This still doesn't eliminate a segfault for anyone who actually *uses* CLOCK_BOOTTIME on an unsupported system, since assert(0) turns into halt statement, but that is really a separate issue.


MartinNowak:
thx

PR #1337, opened 2015-08-04
Fix Issue 14870 Add new "abort" internal module and function

schveiguy:
See discussion here: http://forum.dlang.org/post/mpq93n$mh3$1@digitalmars.com

`abort(msg)` Will ALWAYS print msg, unlike `assert(0, msg)` which does not when compiled in -release mode. `abort` will eventually call `assert(0)`, so the end result is similar, but with a good message for debugging.

The only drawback is that `assert(0, msg)` is pure, but this function cannot be, since it has no mutable references as parameters or return, and the functions it calls to output the message are not pure.

So we can't replace all `assert(0, msg)` calls with `abort(msg)` (I had to leave out one case of `assert(0, msg)` in core.time because it was in a pure function).

In addition, the optimizer may not see that `abort(msg)` does not terminate, I'm not sure.

I have reworked core.time to use this internal module instead. I also had to write (another) integer to string conversion routine (following Walter's lead in his presentation for dconf2015, it doesn't allocate). Is there other needs for that anywhere in druntime? Mabye it should go in core.internal.convert, but that seems to be focused on floating point.

The large changes in core.time were regarding a bizarre "Something threw when nothing should have" try/catch scope, and I found that removing the try/catch did not cause any errors. Probably at one point, the internal/referenced code wasn't properly nothrow-annotated.

If this is approved, I'll work on the other modules in druntime that `assert(0)` with a message, and probably phobos as well.

### DIFF
--- a/mak/MANIFEST
+++ b/mak/MANIFEST
@@ -33,6 +33,7 @@ MANIFEST=\
 	src\core\time.d \
 	src\core\vararg.d \
 	\
+	src\core\internal\abort.d \
 	src\core\internal\convert.d \
 	src\core\internal\hash.d \
 	src\core\internal\string.d \
--- a/src/core/internal/abort.d
+++ b/src/core/internal/abort.d
@@ -1,0 +1,45 @@
+module core.internal.abort;
+
+/*
+ * Use instead of assert(0, msg), since this does not print a message for -release compiled
+ * code, and druntime is -release compiled.
+ */
+void abort(string msg, string filename = __FILE__, size_t line = __LINE__) @nogc nothrow @safe
+{
+    import core.stdc.stdlib: c_abort = abort;
+    // use available OS system calls to print the message to stderr
+    version(Posix)
+    {
+        import core.sys.posix.unistd: write;
+        static void writeStr(const(char)[][] m...) @nogc nothrow @trusted
+        {
+            foreach(s; m)
+                write(2, s.ptr, s.length);
+        }
+    }
+    else version(Windows)
+    {
+        import core.sys.windows.windows: GetStdHandle, STD_ERROR_HANDLE, WriteFile, INVALID_HANDLE_VALUE;
+        auto h = (() @trusted => GetStdHandle(STD_ERROR_HANDLE))();
+        if(h == INVALID_HANDLE_VALUE)
+            // attempt best we can to print the message
+            assert(0, msg);
+        void writeStr(const(char)[][] m...) @nogc nothrow @trusted
+        {
+            foreach(s; m)
+            {
+                assert(s.length <= uint.max);
+                WriteFile(h, s.ptr, cast(uint)s.length, null, null);
+            }
+        }
+    }
+    else
+        static assert(0, "Unsupported OS");
+
+    import core.internal.string;
+    UnsignedStringBuf strbuff;
+
+    // write an appropriate message, then abort the program
+    writeStr("Aborting from ", filename, "(", line.unsignedToTempString(strbuff, 10), ") ", msg);
+    c_abort();
+}
--- a/src/core/internal/string.d
+++ b/src/core/internal/string.d
@@ -27,6 +27,25 @@ char[] unsignedToTempString(ulong value, char[] buf, uint radix) @safe
     return buf[i .. $];
 }
 
+private struct TempStringNoAlloc
+{
+    // need to handle 65 bytes for radix of 2 with negative sign.
+    private char[65] _buf;
+    private ubyte _len;
+    auto get() return
+    {
+        return _buf[$-_len..$];
+    }
+    alias get this;
+}
+
+auto unsignedToTempString(ulong value, uint radix) @safe
+{
+    TempStringNoAlloc result;
+    result._len = unsignedToTempString(value, result._buf, radix).length & 0xff;
+    return result;
+}
+
 unittest
 {
     UnsignedStringBuf buf;
@@ -37,8 +56,81 @@ unittest
     assert(long.sizeof.unsignedToTempString(buf, 10) == "8");
     assert(uint.max.unsignedToTempString(buf, 10) == "4294967295");
     assert(ulong.max.unsignedToTempString(buf, 10) == "18446744073709551615");
+
+    // use stack allocated struct version
+    assert(0.unsignedToTempString(10) == "0");
+    assert(1.unsignedToTempString(10) == "1");
+    assert(12.unsignedToTempString(10) == "12");
+    assert(0x12ABCF .unsignedToTempString(16) == "12abcf");
+    assert(long.sizeof.unsignedToTempString(10) == "8");
+    assert(uint.max.unsignedToTempString(10) == "4294967295");
+    assert(ulong.max.unsignedToTempString(10) == "18446744073709551615");
 }
 
+alias SignedStringBuf = char[20];
+
+auto signedToTempString(long value, char[] buf, uint radix) @safe
+{
+    bool neg = value < 0;
+    if(neg)
+        value = cast(ulong)-value;
+    auto r = unsignedToTempString(value, buf, radix);
+    if(neg)
+    {
+        // about to do a slice without a bounds check
+        assert(r.ptr > buf.ptr);
+        r = (() @trusted => r.ptr[-1..r.length])();
+        r[0] = '-';
+    }
+    return r;
+}
+
+auto signedToTempString(long value, uint radix) @safe
+{
+    bool neg = value < 0;
+    if(neg)
+        value = cast(ulong)-value;
+    auto r = unsignedToTempString(value, radix);
+    if(neg)
+    {
+        r._len++;
+        r.get()[0] = '-';
+    }
+    return r;
+}
+
+unittest
+{
+    SignedStringBuf buf;
+    assert(0.signedToTempString(buf, 10) == "0");
+    assert(1.signedToTempString(buf, 10) == "1");
+    assert((-1).signedToTempString(buf, 10) == "-1");
+    assert(12.signedToTempString(buf, 10) == "12");
+    assert((-12).signedToTempString(buf, 10) == "-12");
+    assert(0x12ABCF .signedToTempString(buf, 16) == "12abcf");
+    assert((-0x12ABCF) .signedToTempString(buf, 16) == "-12abcf");
+    assert(long.sizeof.signedToTempString(buf, 10) == "8");
+    assert(int.max.signedToTempString(buf, 10) == "2147483647");
+    assert(int.min.signedToTempString(buf, 10) == "-2147483648");
+    assert(long.max.signedToTempString(buf, 10) == "9223372036854775807");
+    assert(long.min.signedToTempString(buf, 10) == "-9223372036854775808");
+
+    // use stack allocated struct version
+    assert(0.signedToTempString(10) == "0");
+    assert(1.signedToTempString(10) == "1");
+    assert((-1).signedToTempString(10) == "-1");
+    assert(12.signedToTempString(10) == "12");
+    assert((-12).signedToTempString(10) == "-12");
+    assert(0x12ABCF .signedToTempString(16) == "12abcf");
+    assert((-0x12ABCF) .signedToTempString(16) == "-12abcf");
+    assert(long.sizeof.signedToTempString(10) == "8");
+    assert(int.max.signedToTempString(10) == "2147483647");
+    assert(int.min.signedToTempString(10) == "-2147483648");
+    assert(long.max.signedToTempString(10) == "9223372036854775807");
+    assert(long.min.signedToTempString(10) == "-9223372036854775808");
+    assert(long.max.signedToTempString(2) == "111111111111111111111111111111111111111111111111111111111111111");
+    assert(long.min.signedToTempString(2) == "-1000000000000000000000000000000000000000000000000000000000000000");
+}
 
 int dstrcmp( in char[] s1, in char[] s2 ) @trusted
 {
--- a/src/core/internal/string.d
+++ b/src/core/internal/string.d
@@ -79,7 +79,7 @@ auto signedToTempString(long value, char[] buf, uint radix) @safe
     {
         // about to do a slice without a bounds check
         assert(r.ptr > buf.ptr);
-        r = (() @trusted => r.ptr[-1..r.length])();
+        r = (() @trusted => (r.ptr-1)[0..r.length+1])();
         r[0] = '-';
     }
     return r;
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -82,6 +82,7 @@ import core.exception;
 import core.stdc.time;
 import core.stdc.stdio;
 import core.internal.traits : _Unqual = Unqual;
+import core.internal.string;
 
 version(Windows)
 {
@@ -1861,7 +1862,9 @@ private:
                 unit = "μs";
             else
                 unit = plural ? units : units[0 .. $-1];
-            res ~= numToString(val) ~ " " ~ unit;
+            res ~= signedToTempString(val, 10);
+            res ~= " ";
+            res ~= unit;
         }
 
         if (_hnsecs == 0) return "0 hnsecs";
@@ -1982,6 +1985,14 @@ unittest
         auto t1 = TickDuration.from!U(t1v);
         auto t2 = TickDuration.from!U(t2v);
 
+        auto _str(F)(F val)
+        {
+            static if(is(F == int) || is(F == long))
+                return signedToTempString(val, 10);
+            else
+                return unsignedToTempString(val, 10);
+        }
+
         foreach (F; _TypeTuple!(int,uint,long,ulong,float,double,real))
         {
             F t1f = to!(U,F)(t1);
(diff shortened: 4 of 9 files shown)


DmitryOlshansky:
> I have reworked core.time to use this internal module instead. I also had to write (another) integer to string conversion routine (following Walter's lead in his presentation for dconf2015, it doesn't allocate). Is there other needs for that anywhere in druntime? Mabye it should go in core.internal.convert, but that seems to be focused on floating point.

There is something simillar in core.exception for Exception toString IIRC.

schveiguy:
> There is something simillar in core.exception

Close enough, I found it, actually part of Throwable in object.d. It's in rt/util/string, I'll try and rewrite my code to use it.

schveiguy:
So long story short, I had based my branch on an old version of the master by accident. The rt/util/string.d has been moved, and since I already messed it up, I had to rebuild the branch. In any case, it should be OK now.

jmdavis:
If you're so interested in a message always being printed even in release mode, why not just throw an `Error`?

DmitryOlshansky:
LGTM

schveiguy:
> why not just throw an Error?

I could do that, but this is not dependent on any external code. It can be called from a signal handler, or just about anywhere. And if the system is corrupted, it should still work.

jmdavis:
Well, there are already plenty of places in druntime that throw various Error types, and they're going to have to be caught correctly by _Dmain or whatever to be printed. It seems like `abort` is just specifically for the `assert(0)` case, and I don't see anything special about druntime in that regard. The current language design for it is that it's not going to print anything when built with -release, and no one should be seeing those fail except in highly abnormal circumstances anyway. So, this seems like it's purely designed to work around Walter not being willing to have `assert(0, msg)` print before the HLT instruction in release mode.

In the case of core.time, it probably should be throwing an Error, not asserting 0. It was written with the assumption that things were seriously broken if those calls failed, but since there are legitimate cases where they can fail where it's not a bug in druntime at all, an Error would make more sense, and then you will get a message in that case just like with any of the other Errors that we already throw in druntime.

WalterBright:
This is intended to fix https://issues.dlang.org/show_bug.cgi?id=14870

src/core/internal/abort.d
```diff
@@ -0,0 +1,44 @@
...
+        {
+            foreach(s; m)
+            {
+                assert(s.length <= uint.max);
+                WriteFile(h, s.ptr, cast(uint)s.length, null, null);
+            }
+        }
+    }
+    else
+        static assert(0, "Unsupported output for this OS!");
```
WalterBright: Please, no !. Just `unsupported OS` is plenty good enough.

src/core/internal/abort.d
```diff
@@ -0,0 +1,44 @@
...
+void abort(string msg, string filename = __FILE__, size_t line = __LINE__) @nogc nothrow @safe
+{
+    // use available OS system calls to print the message to stderr
+    version(Posix)
+    {
+        import core.sys.posix.unistd: write;
+        static void writeStr(const(char)[][] m...) @nogc nothrow @trusted
+        {
+            foreach(s; m)
+                cast(void)write(2, s.ptr, s.length);
```
WalterBright: why the cast?

src/core/time.d
```diff
@@ -2645,9 +2650,10 @@ extern(C) void _d_initMonoTime()
         {
             foreach(i, typeStr; __traits(allMembers, ClockType))
             {
+                import core.internal.abort : abort;
                 // ensure we are only writing immutable data once
                 if(tps[i] != 0)
-                    assert(0, "_d_initMonoTime should only be called once!");
+                    abort("_d_initMonoTime should only be called once!");
```
WalterBright: please drop the !, it isn't that exciting :-)

WalterBright: wait a minute, this is not an environmental issue, it's a programming bug. It should remain an assert.

src/core/time.d
```diff
@@ -2662,9 +2668,10 @@ extern(C) void _d_initMonoTime()
             long ticksPerSecond = 1_000_000_000L * info.numer / info.denom;
             foreach(i, typeStr; __traits(allMembers, ClockType))
             {
+                import core.internal.abort : abort;
                 // ensure we are only writing immutable data once
                 if(tps[i] != 0)
-                    assert(0, "_d_initMonoTime should only be called once!");
+                    abort("_d_initMonoTime should only be called once!");
```
WalterBright: leave as assert. No !.

src/core/time.d
```diff
@@ -2681,9 +2688,10 @@ extern(C) void _d_initMonoTime()
                 enum clockArg = _posixClock(__traits(getMember, ClockType, typeStr));
                 if(clock_getres(clockArg, &ts) == 0)
                 {
+                    import core.internal.abort : abort;
                     // ensure we are only writing immutable data once
                     if(tps[i] != 0)
-                        assert(0, "_d_initMonoTime should only be called once!");
+                        abort("_d_initMonoTime should only be called once!");
```
WalterBright: ditto

src/core/time.d
```diff
@@ -4714,34 +4719,29 @@ double _abs(double val) @safe pure nothrow @nogc
   +/
 string numToString(long value) @safe pure nothrow
```
WalterBright: Sigh. Yet another number-to-string conversion function. But fixing this should be a separate PR.

src/core/internal/abort.d
schveiguy: Heh, I original thought to make abort pure (for more utility), but realized later that it cannot be made pure, because it would be considered a strong pure function without any return.

However, while I did have it pure, I had to "fake" write being pure. The compiler complained I wasn't using the output, and so I followed the instructions (put a cast(void) there ;)

It's a relic, will remove.

src/core/internal/abort.d
schveiguy: OK

src/core/time.d
schveiguy: OK.

src/core/time.d
schveiguy: I can fix now. Not too hard.

src/core/internal/abort.d
jmdavis: I didn't even know that you _could_ cast to `void`. What would that even do? Make it so that an expression is a statement but not an expression?

src/core/time.d
jmdavis: If it's really an environmental issue, we should just be throwing Errors IMHO. There might be cases where you wouldn't want to do that, but they'd be pretty atypical, I would think. Really, it looks to me like `abort` is just `assert(0, msg)` that still prints a message when it fails in release mode.

src/core/time.d
WalterBright: Yeah, you're right. Just fix it now :-)

jmdavis:
I think that the cases in core.time aside from those related to `_d_initMonoTime` should just be changed to throw Errors on failure, since they fail when a system call fails, which isn't necessarily a program bug. The program is screwed in that case, but it's probably not its fault.

src/core/time.d
jmdavis: `numToString` does deal with signed integers, so simply using `core.string.unsignedToTempString()` wouldn't cut it, but given that there's a function for handling this in druntime, I agree that we should definitely use that instead. When I wrote `numToString`, I had no clue that the other existed, and I obviously didn't have access to `std.conv.to`. Maybe we should just add `core.string.signedToTempString()` to handle the signed case. In any case, I agree that that should be another PR.

jmdavis:
By the way, for those reviewing this, adding ?w=1 to the end of the URL for the diff definitely helps, because some of the changes altered the indentation, and so a number of the changes are whitespaces changes, and they increase the number of lines in the diff considerably.

src/core/time.d
jmdavis: LOL. Apparently, I didn't reload the page before commenting, and I didn't see either of the two comments that now precede mine. I think that it's better to do the numToString change in another PR, but it's certainly not the end of the world if it goes in this one.

schveiguy:
> number of the changes are whitespaces changes

Yeah, those are the cases where I removed the now-unnecessary try-catch.

src/core/time.d
schveiguy: I am almost done with it :) I've added signed versions, and also versions that don't require a user-supplied buffer (but do not allocate, a la Walter's dconf15 talk).

schveiguy:
OK, updated to take care of comments. Removed `numToString` from standard code, now uses core.internal.string, changed `numToString` for doubles to use core.internal.string (this was a unittest-only function).

One odd note, I have found that:

```D
char[] x = "blah".dup;
auto y = "hi, " ~ x;
assert(is(typeof(y == char[])));
```
This messes up a lot of code, which is why you see idups in places that seemingly shouldn't need them (namely for throwing errors, which only accept string as the message parameter). It would be nice if the compiler simply treated the result of a concatenation as any constancy, since it's clearly a new piece of memory. For cases where the result was used in real code (not unittest), I updated the code to use proper concatenation (it's faster anyway).

src/core/internal/abort.d
schveiguy: What it does is tell the compiler "yes, I know this returns a value and I'm not using it".

For pure functions with no mutable reference parameters, this makes a lot of sense (why call it otherwise?)

schveiguy:
> I think that the cases in core.time aside from those related to _d_initMonoTime should just be changed to throw Errors on failure, since they fail when a system call fails, which isn't necessarily a program bug. The program is screwed in that case, but it's probably not its fault.

It could be a memory corruption bug (if the 0 value came from corruption instead of unsupported clocks). I'm really on the fence for this, we could go either way.

src/core/internal/abort.d
WalterBright: write() can never be pure, and casting it to void doesn't change that. The void cast is an ugly thing from C that I never cared for and don't want to start using in official D code.

ibuclaw:
I'm not really a fan. From the places that are changed in this PR it looks more appropriate to have errors thrown.

ibuclaw:
> It could be a memory corruption bug (if the 0 value came from corruption instead of unsupported clocks).

If it's corruption that could occur through external factors, and not some internal state of core.time, then it should still be an Error to alert of programmer bug.

jmdavis:
> It could be a memory corruption bug (if the 0 value came from corruption instead of unsupported clocks). I'm really on the fence for this, we could go either way.

I don't see much point in worrying about that. If a memory corruption starts screwing with stuff, then all bets are off anyway, and really, it doesn't matter particularly why the system call failed. We can't continue if it does regardless of why it failed. And since it's not a bug in druntime itself at that point, an assertion isn't appropriate. When I wrote that code, I was pretty much assuming that there was no way that the call would fail, and the assertion was just there to kill the program in case things went seriously wrong somehow, and it did fail. But even then, throwing an Error probably would have been better, and we've now seen a case where `clock_gettime` _can_ fail due to a normally valid clock type being used on a system where it's not valid, because that system's kernel is too old to support it. So, regardless of whether an assertion would be appropriate if we could reasonably assume that the call would succeed, we now know that there are cases where the call will legitimately fail, but we can't continue if it does, so I think that falls pretty squarely in Error territory.

assert(0) is really supposed to be for truly unreachable code where it can't be proven that it's unreachable (and thus a serious bug if it is reached), and that's not the case with `clock_gettime` failing.

mak/MANIFEST
```diff
@@ -33,6 +33,7 @@ MANIFEST=\
 	src\core\time.d \
 	src\core\vararg.d \
 	\
+	src\core\internal\abort.d \
```
WalterBright: Yay! Remembered to update MANIFEST!

schveiguy:
> don't see much point in worrying about that. If a memory corruption starts screwing with stuff, then all bets are off anyway, and really, it doesn't matter particularly why the system call failed.

But `abort` can be called even when memory is corrupted. It uses no global state or heap memory, only system calls and new stack space. I think this is the best/safest way to print anything if you have reached a place that should not be reached. Throwing an error requires the GC work properly, and that the stack has not been corrupted.

schveiguy:
> From the places that are changed in this PR it looks more appropriate to have errors thrown.

The usage of abort may or may not be appropriate for those locations in core.time. And I plan to go throughout druntime and phobos and fix all the `assert(0)` that print messages (but don't really because we compile them in release mode). But I think the `abort` function has good use for a "last resort" way to communicate what went wrong to the user. It could be used (for example) in the signal handler for seg fault.

src/core/internal/abort.d
```diff
@@ -0,0 +1,44 @@
...
+    }
+    else
+        static assert(0, "Unsupported OS");
+
+    import core.internal.string;
+    UnsignedStringBuf strbuff;
+
+    // write an appropriate message, then abort the program
+    writeStr("Aborting from ", filename, "(", line.unsignedToTempString(strbuff, 10), ") ", msg);
+    assert(0);
```
MartinNowak: Let's please use abort [¹](http://www.cplusplus.com/reference/cstdlib/abort/),[²](https://msdn.microsoft.com/en-us/library/vstudio/k089yyh0(v=vs.100).aspx) or _Exit [¹](http://www.cplusplus.com/reference/cstdlib/_Exit),[²](https://msdn.microsoft.com/en-us/library/vstudio/6wdz5232(v=vs.100).aspx) instead of the `hlt` instruction.

src/core/internal/string.d
```diff
@@ -27,6 +27,26 @@ char[] unsignedToTempString(ulong value, char[] buf, uint radix) @safe
     return buf[i .. $];
 }
 
+auto unsignedToTempString(ulong value, uint radix) @safe
+{
+    struct Result
+    {
+        // need to handle 65 bytes because radix could be 2, and we
+        // could be signed.
```
MartinNowak: Function takes only unsigned numbers.

src/core/internal/string.d
```diff
@@ -37,8 +57,81 @@ unittest
...
+    bool neg = value < 0;
+    if(neg)
+        value = cast(ulong)-value;
+    auto r = unsignedToTempString(value, radix);
+    if(neg)
+    {
+        r._len++;
+        r.get()[0] = '-';
+    }
+    return r;
```
MartinNowak: Would be much cleaner to declare the Result struct in signedToTempString and use that from unsignedToTempString.

ibuclaw:
> It could be used (for example) in the signal handler for seg fault.

Let us hope that the signal handler isn't recursive. ;-)

src/core/internal/string.d
schveiguy: It's used by `signedToTempString` instead of duplicating the code and creating a new type.

src/core/internal/abort.d
schveiguy: OK, that's probably fine. I'll pick abort.

src/core/internal/string.d
schveiguy: Except `signedToTempString` cannot properly accept `ulong.max` and generate the string for it.

Perhaps the "correct" way to do this is to take the struct into the module space.

schveiguy:
> Let us hope that the signal handler isn't recursive. ;-)

From my understanding, if you trigger a recursive signal, it will simply exit the program.

schveiguy:
Updated to use c `abort` instead of `assert(0)`, moved non-allocating struct for getting temp string outside of the `unsignedToTempString` function because it's used for both signed and unsigned to avoid confusion with comments.

I also marked `core.stdc.stdlib.abort` as `@safe` instead of calling it in a `@trusted` lambda, because the documentation says it basically terminates the program immediately, so I can't see how it could corrupt memory. If this seems too much, I can do it the other way.

DmitryOlshansky:
Fix failures on the auto-tester?

DmitryOlshansky:
> I also marked core.stdc.stdlib.abort as @safe instead of calling it in a @trusted lambda, because the documentation says it basically terminates the program immediately, so I can't see how it could corrupt memory. If this seems too much, I can do it the other way.

Makes sense

schveiguy:
> Fix failures on the auto-tester?

Attempt made, but I also want to fix the issue where the unsigned struct takes into account the negative sign (which reads weird). So hold off for now.

DmitryOlshansky:
Good to go?

schveiguy:
Huh, I guess I already moved the struct. My damned memory :)

So yeah, good to go.

DmitryOlshansky:
Auto-merge toggled on

MartinNowak:
Changelog?

schveiguy:
> Changelog?

Are we using the individual changelogs now, or should I update dlang.org changelog? I see there's a changelog.dd in druntime source...

MartinNowak:
Use the individual ones (now on stable) please.

schveiguy:
Working on this, but I'm wondering, should this even be in the changelog? We don't document core.internal at all.

schveiguy:
ping @MartinNowak

MartinNowak:
I thought it was a public function, no need to mention core.internal and the change to always print errors is quite small too.

DmitryOlshansky:
My bad - I thought abort was a new public primitive.

schveiguy:
It could be made public if it makes sense.

PR #1339, opened 2015-08-07
fix Issue 14882 - [REG] MANIFEST is missing test/common.mak

WalterBright:
https://issues.dlang.org/show_bug.cgi?id=14882


schveiguy:
Going to commit, even with build of docs failing -- doesn't seem like something to do with the MANIFEST, because it seems like a git error that is causing the build failure. @CyberShadow ?

schveiguy:
Auto-merge toggled on

PR #1342, opened 2015-08-11
add put(char) overload

WalterBright:
This trivial change reduces the size of the data section from 3126 to 2726 bytes, and the size of the object file from 26464 to 25041 bytes. There's a lot more that can be done to this module to reduce its footprint.

### DIFF
--- a/src/core/demangle.d
+++ b/src/core/demangle.d
@@ -223,6 +223,12 @@ private struct Demangle
     }
 
 
+    char[] put(char c)
+    {
+        char[1] val = c;
+        return put(val[]);
+    }
+
     char[] put( const(char)[] val )
     {
         if( val.length )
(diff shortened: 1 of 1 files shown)


src/core/demangle.d
```diff
@@ -223,6 +223,12 @@ private struct Demangle
     }
 
 
+    char[] put(char c)
+    {
+        char[1] val = c;
+        return put(val[]);
```
MartinNowak: You could simply use `put((&c)[0 .. 1])`.

WalterBright: true. Will do so next pass.

WalterBright: I decided against that because it makes the code not @safe

yebblies: It's not `@safe` either way, the compiler just lacks the check for slicing the static array.

PR #1343, opened 2015-08-12
core.dmangle: remove another custom int to string converter

WalterBright:
I wonder how many more int=>string converters there are in druntime/phobos. There were two in demangle.

dmangle.d is worth attention because it is linked into every D executable. Also, it is rarely called, so performance is not an issue. But it needs to be as compact as possible.

### DIFF
--- a/src/core/demangle.d
+++ b/src/core/demangle.d
@@ -139,16 +139,6 @@ private struct Demangle
     }
 
 
-    static char val2HexDigit( ubyte val )
-    {
-        if (0x0 <= val && val <= 0x9)
-            return cast(char)(val + '0');
-        if (0xa <= val && val <= 0xf)
-            return cast(ubyte)(val + 'a' - 0xa);
-        assert(0);
-    }
-
-
     //////////////////////////////////////////////////////////////////////////
     // Data Output
     //////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


ibuclaw:
I should probably have a sweep of this code myself to try and fix any of the bugs I raised (that are not present in my version in GCC/GDB)

For instance, does outputting whitespace as-is here considered a problem, or should we attempt to sanitize it?

ibuclaw:
Not in the else branch you changed, but the then branch above it.

WalterBright:
@ibuclaw I don't know what you mean. What branch are you referring to, for example? What whitespace issue? Note that it passes the unit tests.

ibuclaw:
Here: https://github.com/ibuclaw/gdb/blob/30379291886a171e6dc202122bc1c583318c2e17/libiberty/testsuite/d-demangle-expected#L1051

ddemangle outputs:
```
bool std.socket.InternetHost.getHostNoSync!("
	    auto x = htonl(param);
	    auto he = gethostbyaddr(&x, 4, cast(int)AddressFamily.INET);
	", uint).getHostNoSync(uint)
```
instead of:
```
std.socket.InternetHost.getHostNoSync!("\n\t auto x = htonl(param);\n\t auto he = gethostbyaddr(&x, 4, cast(int)AddressFamily.INET);\n\t", uint).getHostNoSync(uint)
```

(By the way, I do encourage using my testsuite in core.demangle if that helps improving it)

WalterBright:
If there are bugs in demangle.d, please submit bug reports for them! Also, please add your test suite to the bug report.

(This PR is only about refactoring, not fixing bugs or changing behavior.)

PR #1344, opened 2015-08-14
Import fallocate in sys/linux/fcntl.d

mrkline:
This adds prototypes for posix_fallocate (which was present but commented out) and the linux-specific fallocate (http://man7.org/linux/man-pages/man2/fallocate.2.html).

### DIFF
--- a/src/core/sys/posix/fcntl.d
+++ b/src/core/sys/posix/fcntl.d
@@ -541,6 +541,9 @@ else
 int fcntl(int, int, ...);
 //int open(in char*, int, ...);
 
+// Generic Posix fallocate
+int posix_fallocate(int, off_t, off_t);
+
 //
 // Advisory Information (ADV)
 //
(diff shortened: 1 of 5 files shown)


src/core/sys/posix/fcntl.d
```diff
@@ -541,6 +541,29 @@ else
 int fcntl(int, int, ...);
 //int open(in char*, int, ...);
 
+// Generic Posix fallocate
+int posix_fallocate(int, off_t, off_t);
+
+// Linux-specific fallocate
+// (http://man7.org/linux/man-pages/man2/fallocate.2.html)
+version( linux ) {
```
schuetzm: Shouldn't this go into core.sys.linux.fcntl instead?

mrkline: I believe you're right.

DmitryOlshansky:
Please squash commits, otherwise LGTM

mrkline:
@DmitryOlshansky Done.

DmitryOlshansky:
Auto-merge toggled on

PR #1345, opened 2015-08-15
replace mangleC with Cprefix

WalterBright:
Removes an unnecessary function and an unnecessary reference to the GC. I've never encountered a C mangling method that did more than simply add a prefix. Having an abstraction for a possibility that will never happen is useless bloat.

### DIFF
--- a/src/core/demangle.d
+++ b/src/core/demangle.d
@@ -1890,36 +1890,15 @@ unittest
     static assert(!__traits(compiles, mangleFunc!(typeof(&fooCPP))("")));
 }
 
-/**
-* Mangles a C function or variable.
-*
-* Params:
-*  sym = The C symbol to mangle.
-*  dst = An optional destination buffer.
-*
-* Returns:
-*  The mangled name for a C function or variable, i.e.
-*  an underscore is prepended or not, depending on the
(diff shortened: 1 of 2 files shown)


ibuclaw:
I am confused by the fact this even exists...

WalterBright:
> I am confused by the fact this even exists...

It's used only by a hack in rt.config to get C name mangling on an identifier declared in a template. The identifier is declared in a template in order for it to be put into a COMDAT section so it can be optionally overridden by a regular declaration. Names in templates are not given C mangling because then they would collide with other names generated by other instantiations of that template.

> Wasn't  mangleC  part of the public API?

Yes. But I suspect it is never used other than for that one hack, as I can think of no other purpose for its existence. It could arguably be marked `deprecated`, but since demangle is inserted into every D executable, I was looking for ways to do a shrink on the executable size, and also ways to reduce the dependence of druntime on the GC.

> In any case,  Cprexfix  should be  cPrefix  according to our naming conventions.

Fixed.

DmitryOlshansky:
Good to merge? @klickverbot @ibuclaw

DmitryOlshansky:
Auto-merge toggled on

PR #1347, opened 2015-08-15
do overflow check correctly

WalterBright:
No need to reinvent this.


9rnsr:
Auto-merge toggled on

PR #1348, opened 2015-08-15
refactor demangle: use range terminology

WalterBright:
Should not invent alternate terminology.


DmitryOlshansky:
LGTM

DmitryOlshansky:
Auto-merge toggled on

PR #1349, opened 2015-08-17
add cmpxchg to core.atomic

WalterBright:
This makes use of the compiler cmpxchg builtin's added in https://github.com/D-Programming-Language/dmd/pull/4900 with a default implementation so it keeps working until dmd is updated.

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -280,78 +280,27 @@ else version( AsmX86_32 )
     {
         static if( T.sizeof == byte.sizeof )
         {
-            //////////////////////////////////////////////////////////////////
             // 1 Byte CAS
-            //////////////////////////////////////////////////////////////////
 
-            asm pure nothrow @nogc
-            {
-                mov DL, writeThis;
-                mov AL, ifThis;
-                mov ECX, here;
-                lock; // lock always needed to make this op atomic
-                cmpxchg [ECX], DL;
-                setz AL;
-            }
+            return cmpxchg(cast(shared(ubyte)*)here, cast(ubyte)ifThis, cast(ubyte)writeThis);
         }
         else static if( T.sizeof == short.sizeof )
         {
-            //////////////////////////////////////////////////////////////////
             // 2 Byte CAS
-            //////////////////////////////////////////////////////////////////
 
-            asm pure nothrow @nogc
-            {
-                mov DX, writeThis;
-                mov AX, ifThis;
-                mov ECX, here;
-                lock; // lock always needed to make this op atomic
-                cmpxchg [ECX], DX;
-                setz AL;
-            }
+            return cmpxchg(cast(shared(ushort)*)here, cast(ushort)ifThis, cast(ushort)writeThis);
         }
         else static if( T.sizeof == int.sizeof )
         {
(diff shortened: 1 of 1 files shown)


src/core/atomic.d
```diff
@@ -1273,6 +1182,146 @@ else version( AsmX86_64 )
     }
 }
 
+/* ***
+ * Functions built in to the compiler, along with a default implementation.
+ */
+
+bool cmpxchg(shared(ubyte)* here, ubyte ifThis, ubyte writeThis) pure nothrow @nogc @trusted
```
MartinNowak: Should be private.

WalterBright: Builtins should be public and available, after all, it's a systems programming language.

klickverbot: They are already available, using `cas`.

WalterBright: `cas` adds another layer. Again, a systems programming language should expose the primitives, not hide them.

ibuclaw: @klickverbot yeah, cas is enough on my side too. I even pass the MemoryOrder value to my backend and it will do the right thing!

Having this as a publicly exposed builtin is as much use as y2lx.

klickverbot: @WalterBright: This is hardly another layer. It's another copy that does exactly the same thing. Please also note that other compilers will implement atomics differently. We'd just have to make cmpxchg forward to `cas`.

MartinNowak: It's a X86 specific whatever that does the same thing as cas. There is no value in exposing that.

MartinNowak: Also we're going to need a lot more intrinsics that that, https://gcc.gnu.org/onlinedocs/gcc-4.9.3/gcc/_005f_005fatomic-Builtins.html. You don't want to add all of those, particularly considering that they are dependent on the compiler.

WalterBright: I'll just reiterate that D being a systems programming language, it should provide a means for the user to get access to low level things like intrinsics.

klickverbot: …which is why we have standalone `cas` and friends in druntime instead of pre-packaged atomic types.

klickverbot: … which is why things like `cas` on arbitrary pointers exist, as opposed to only pre-packaged atomic types and so on.

yebblies: If this doesn't offer anything extra over cas then there's no point in having it public.

MartinNowak: > I'll just reiterate that D being a systems programming language, it should provide a means for the user to get access to low level things like intrinsics.

If that would be your goal then please build a proper API for the intrinsics that is compatible with other compilers, i.e. use https://gcc.gnu.org/onlinedocs/gcc-4.9.3/gcc/_005f_005fatomic-Builtins.html, so users don't have to deal with different intrinsics on every compiler.
Failing that let's please just make core.atomic faster and more useful.

DmitryOlshansky: Nice discussion but the end result is that it drifts dead in the water. Closing?

andralex:
I'll close this for now

PR #1350, opened 2015-08-18
core.cpuid: emit fewer fixups

WalterBright:
The idea is to:

1. consolidate private globals into a struct, reducing the number of pointless symbols in the object file
2. use a pointer to the struct instance to access the globals, which reduces the fixups in the executable image, making for faster loading and a smaller executable file size.

Worthwhile because core.cpuid is compiled into every D executable.

### DIFF
--- a/src/core/cpuid.d
+++ b/src/core/cpuid.d
@@ -311,6 +311,9 @@ public:
     /// The number of cache levels in the CPU.
     @property uint cacheLevels() { return numCacheLevels; }
 private:
+
+struct CpuFeatures
+{
     bool probablyIntel; // true = _probably_ an Intel processor, might be faking
     bool probablyAMD; // true = _probably_ an AMD processor
     string processorName;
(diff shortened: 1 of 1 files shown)


WalterBright:
Needs careful review, because this code is pretty much untestable.

src/core/cpuid.d
```diff
@@ -980,57 +1002,57 @@ shared static this()
...
-    _sse2 =           (features&SSE2_BIT)!=0;
-    _sse3 =           (miscfeatures&SSE3_BIT)!=0;
-    _ssse3 =          (miscfeatures&SSSE3_BIT)!=0;
-    _sse41 =          (miscfeatures&SSE41_BIT)!=0;
-    _sse42 =          (miscfeatures&SSE42_BIT)!=0;
-    _sse4a =          (amdmiscfeatures&SSE4A_BIT)!=0;
-    _aes =            (miscfeatures&AES_BIT)!=0;
-    _hasPclmulqdq =   (miscfeatures&PCLMULQDQ_BIT)!=0;
-    _hasRdrand =      (miscfeatures&RDRAND_BIT)!=0;
+    _vendor =         cast(string)cf.vendorID;
```
DmitryOlshansky: would it make sense to put immutables into a struct too? Then these lines are trivially re-written as `icf = Immortals(cast(string)cf.vendorID, ...); `

WalterBright: Yes, but I wished to limit the scope of this PR to make it easier to review.

DmitryOlshansky: Separate commit will make it just as easy. Well doing the next PR will work too

WalterBright:
rebased

WalterBright:
Any reason why this is not being pulled?

DmitryOlshansky:
Auto-merge toggled on

WalterBright:
thanks! (I don't want to rebase it again!)

DmitryOlshansky:
> thanks! (I don't want to rebase it again!)

No problem, but consider for a moment that a typical pull stays for month or two and survives 3-4 rebases

PR #1351, opened 2015-08-20
Issue 8973 - core.cpuid.coresPerCPU returning incorrect value. 

rsw0x:
maxCores for AMD is currently calculated using the theoretical maximum number of cores, not the actual enabled maximum number of cores.

This causes all bulldozer CPUs(and possibly others) to report the wrong value.

http://support.amd.com/TechDocs/25481.pdf

>The size of this field determines the maximum number of cores (MNC) that the processor could theoretically support, not the actual number of cores that are actually implemented or enabled on the processor, as indicated by CPUID Fn8000_0008_ECX[NC]. 

[issue](https://issues.dlang.org/show_bug.cgi?id=8973)

### DIFF
--- a/src/core/cpuid.d
+++ b/src/core/cpuid.d
@@ -735,15 +735,10 @@ void cpuidX86()
             cpuid;
             mov c, ECX;
         }
-        uint apicsize = (c>>12) & 0xF;
-        if (apicsize == 0) {
-            // use legacy method
-            if (hyperThreadingBit)  maxCores = c & 0xFF;
-            else maxCores = 1;
-        } else {
-            // maxcores = 2^ apicsize
-            maxCores = 1;
-            while (apicsize) { maxCores<<=1; --apicsize; }
+        //http://support.amd.com/TechDocs/25481.pdf pg.36
+        maxCores = 1;
+        if (hyperThreadingBit) {
+            maxCores += c & 0xFF;
         }
     }
 


src/core/cpuid.d
```diff
@@ -735,15 +735,10 @@ void cpuidX86()
...
-            // use legacy method
-            if (hyperThreadingBit)  maxCores = c & 0xFF;
-            else maxCores = 1;
-        } else {
-            // maxcores = 2^ apicsize
-            maxCores = 1;
-            while (apicsize) { maxCores<<=1; --apicsize; }
+        //http://support.amd.com/TechDocs/25481.pdf pg.36
+        maxCores = 1;
+        if (hyperThreadingBit) {
```
DmitryOlshansky: This is reffered to a legacy method. Any reasons the so-called extended method cannot be used?

rsw0x: Using NC is the listed way to do it under the extended method. I believe the 'legacy' is referring to the deprecated LogicalProcessorCount, not NC.

rsw0x: After a reread, I was incorrect. The legacy is referring to MNC only, not NC. MNC is the theoretical core limit, and what is wanted here is the NC(number of cores) which is not legacy. You can see this at the top of page 37 in the reference manual.

DmitryOlshansky: Yes, you are right. All bits were confusingly list in legacy method section, still the only legacy fields are LogicalCpuCount and CmpLegacy

DmitryOlshansky:
Auto-merge toggled on

PR #1352, opened 2015-08-21
output coverage reports to $(ROOT) to avoid test interference

MartinNowak:
- release and debug tests might run in parallel


MartinNowak:
https://issues.dlang.org/show_bug.cgi?id=14946

schveiguy:
OK cool, so there is a way to do this :)

And you opened it 12 hours earlier than me. I didn't even see it :D

schveiguy:
Auto-merge toggled on

PR #1353, opened 2015-08-21
fix issue 14946 - druntime coverage tests fail intermittently

schveiguy:
https://issues.dlang.org/show_bug.cgi?id=14946

The solution is very unix specific, I'm not sure what happens on windows, but it doesn't appear to run those tests.

It's also specific to the way we construct the ROOT directory, but I'm not sure of a good way to do that from ROOT itself. This at least gets the job done.

I'd much rather see a good solution to this involving direction of the runtime on where to write the coverage data.


MartinNowak:
see #1352

PR #1356, opened 2015-08-23
properly fix failing test with missing CLOCK_BOOTTIME support

MartinNowak:
- detect older kernel versions in makefile
- disable ClockType.bootTime in tests

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -27,7 +27,7 @@ INSTALL_DIR=../install
 DOCDIR=doc
 IMPDIR=import
 
-override PIC:=$(if $(PIC),-fPIC,)
+OPTIONAL_PIC:=$(if $(PIC),-fPIC,)
 
 ifeq (osx,$(OS))
 	DOTDLL:=.dylib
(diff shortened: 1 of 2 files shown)


MartinNowak:
Needed to fix Travis-CI tests for ddmd.

schveiguy:
I don't like how the normal (non-test) code is affected by the change. Can you just change it so the unit test skips over that clock?

MartinNowak:
> Might be overkill, though, if this stays the only kink of this sort. I'd imagine the capabilities of other systems to evolve similarly from time to time, though?

I don't, clock APIs are around since ages.

MartinNowak:
> I don't like how the normal (non-test) code is affected by the change. Can you just change it so the unit test skips over that clock?

Done

schveiguy:
LGTM

ibuclaw:
> Needed to fix Travis-CI tests for ddmd.

I wonder why when I just tested gdc-4.9.3 right now that I didn't have this problem...

MartinNowak:
> Possibly because your test ended up running on a box with a newer kernel?

Not sure either, they all use the same kernel.
https://travis-ci.org/D-Programming-Language/dmd/jobs/77026325#L13

ibuclaw:
> Possibly because your test ended up running on a box with a newer kernel?

It's not that.
gdc (pass): https://travis-ci.org/D-Programming-Language/dmd/jobs/77179955#L13
dmd (fail): https://travis-ci.org/D-Programming-Language/dmd/jobs/77179952#L13

ibuclaw:
Hmm... just restarted it and it looks like it will pass now.

MartinNowak:
> Hmm... just restarted it and it looks like it will pass now.

works: https://travis-ci.org/MartinNowak/test_clock_getres/builds/77230921
fails: https://travis-ci.org/MartinNowak/test_clock_getres/builds/77231498

Since the tests are running on the old fully virtualized environment (VirtualBox or so) it might be a lack of support in VirtualBox or the host kernel. Let's merge this and go on.

MartinNowak:
ping

DmitryOlshansky:
Auto-merge toggled on

PR #1357, opened 2015-08-24
Place SEEK_DATA, SEEK_HOLE in core.sys.linux.unistd

mrkline:
Pull #1344 placed them with fallocate. While the three are often used in conjunction, SEEK_DATA and SEEK_HOLE should be placed with SEEK_SET, SEEK_CUR, and SEEK_END.

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -66,6 +66,7 @@ COPY=\
 	$(IMPDIR)\core\sys\linux\termios.d \
 	$(IMPDIR)\core\sys\linux\time.d \
 	$(IMPDIR)\core\sys\linux\tipc.d \
+	$(IMPDIR)\core\sys\linux\unistd.d \
 	\
 	$(IMPDIR)\core\sys\linux\sys\inotify.d \
 	$(IMPDIR)\core\sys\linux\sys\mman.d \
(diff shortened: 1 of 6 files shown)


mrkline:
Does this belong here in core/stdc/stdio.d with the other `SEEK_` definitions?
core/sys/linux/stdio.d since it's Linux-only?
core/sys/posix/unistd.d since it's used with `lseek`?
A new core/sys/linux/unistd.d file since it only works with `lseek` on Linux?

The choices are many. Is there a general guide on where to place functions like this in druntime?

WalterBright:
I don't see what is gained by moving declarations from one file to another.

mrkline:
I put those declarations there last week (so there's this would be breaking in a stable release), and they're in the wrong spot. I put them in the module for `fallocate` when they don't belong there - they're used for `lseek`.

mrkline:
Relevant man page: http://man7.org/linux/man-pages/man2/lseek.2.html

DmitryOlshansky:
> In a new core/sys/linux/unistd.d

Makes more sense then the other two

mrkline:
@DmitryOlshansky Changed and squashed.

DmitryOlshansky:
Auto-merge toggled on

WalterBright:
Causes regression https://issues.dlang.org/show_bug.cgi?id=15223

mrkline:
Sorry about missing the MANIFEST and thanks for patching it up.

PR #1359, opened 2015-08-26
follow up for -profile=gc

MartinNowak:
- add filename functions to core.runtime
- fix null terminated string
- add tests


MartinNowak:
Updated to fix all 32/64 bit issues and with changelog entry. Please merge for 2.068.1 beta.

MartinNowak:
Ping, beta is due for today.

src/core/runtime.d
```diff
@@ -342,9 +346,44 @@ extern (C) void dmd_coverDestPath(string path);
...
+/**
+ * Set the output file name for profile reports (-profile switch).
+ * An empty name will set the output to stdout.
+ *
+ * Params:
+ *  name = file name
+ * Note:
+ *  This is a dmd specific setting.
+ */
+extern (C) void trace_setlogfilename(string name);
```
klickverbot: These should be changed to camel casing before going into a release.

MartinNowak: Let's not repeat the discussion https://github.com/D-Programming-Language/druntime/pull/1225#issuecomment-104062444.

MartinNowak: The functions are already used, so we cannot simply rename them.

klickverbot: I still think that the wrappers should at least use the proper druntime naming convention, but I accept that @WalterBright does not give a damn about having a consistent library experience, so feel free to merge. I strongly believe that inconsistent names are like

MartinNowak:
I'm not too happy with the naming and Walter's disinterest in consistency myself, but those are rarely used functions so you'd likely have to look them up in the docs anyhow.

PR #1361, opened 2015-08-27
Visual Studio 2015: exploit C99-conformant (v)snprintf in core.stdc.stdio

kinke:
And add some comments about changed constants.


rainers:
Thanks. Also fixes a typo initializing stderr.

rainers:
Auto-merge toggled on

PR #1363, opened 2015-08-28
Fix building for COFF32

rainers:
Alternate symbol name declarations need prefix "_" for Win32 builds.


WalterBright:
Auto-merge toggled on

PR #1364, opened 2015-08-28
fix Issue 14926 - Programs compiled using dmd 2.068 are generating dummy profilegc.log files

WalterBright:
https://issues.dlang.org/show_bug.cgi?id=14926

### DIFF
--- a/src/rt/profilegc.d
+++ b/src/rt/profilegc.d
@@ -130,21 +130,24 @@ shared static ~this()
         ++i;
     }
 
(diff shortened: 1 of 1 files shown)


yebblies:
Auto-merge toggled on

MartinNowak:
This should have gone into stable.

MartinNowak:
281bf608c07b0acdfc5ff97e1699057896082155

PR #1366, opened 2015-08-29
Add intrinsic pragmas to builtin functions

yebblies:
Requires: https://github.com/D-Programming-Language/dmd/pull/4984

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -30,6 +30,7 @@ else version (X86)
  *      The bit number of the first bit set.
  *      The return value is undefined if v is zero.
  */
+pragma(DigitalMars_intrinsic, "bsf")
 int bsf(size_t v) pure;
 
 ///
@@ -46,6 +47,7 @@ unittest
  *      The bit number of the first bit set.
  *      The return value is undefined if v is zero.
  */
+pragma(DigitalMars_intrinsic, "bsr")
 int bsr(size_t v) pure;
 
 ///
@@ -84,12 +86,14 @@ int bt(in size_t* p, size_t bitnum) pure @system
 /**
  * Tests and complements the bit.
  */
+pragma(DigitalMars_intrinsic, "btc")
 int btc(size_t* p, size_t bitnum) pure @system;
 
 
 /**
  * Tests and resets (sets to 0) the bit.
  */
+pragma(DigitalMars_intrinsic, "btr")
 int btr(size_t* p, size_t bitnum) pure @system;
 
 
@@ -106,6 +110,7 @@ p[index / (size_t.sizeof*8)] & (1 << (index & ((size_t.sizeof*8) - 1)))
  *      A non-zero value if the bit was set, and a zero
  *      if it was clear.
  */
+pragma(DigitalMars_intrinsic, "bts")
 int bts(size_t* p, size_t bitnum) pure @system;
 
 ///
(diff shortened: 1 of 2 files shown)


src/core/bitop.d
```diff
@@ -276,15 +291,21 @@ version (AnyX86)
  * They may be used to guarantee a write or read cycle occurs at a specified address.
  */
 
-ubyte  volatileLoad(ubyte * ptr);
-ushort volatileLoad(ushort* ptr);  /// ditto
-uint   volatileLoad(uint  * ptr);  /// ditto
-ulong  volatileLoad(ulong * ptr);  /// ditto
+pragma(DigitalMars_intrinsic, "ind")
```
ibuclaw: Need a better name than ``"ind"`` ?

src/core/bitop.d
```diff
@@ -276,15 +291,21 @@ version (AnyX86)
...
+    ushort volatileLoad(ushort* ptr);  /// ditto
+    uint   volatileLoad(uint  * ptr);  /// ditto
+    ulong  volatileLoad(ulong * ptr);  /// ditto
+}
 
-void volatileStore(ubyte * ptr, ubyte  value);   /// ditto
-void volatileStore(ushort* ptr, ushort value);   /// ditto
-void volatileStore(uint  * ptr, uint   value);   /// ditto
-void volatileStore(ulong * ptr, ulong  value);   /// ditto
+pragma(DigitalMars_intrinsic, "eq")
```
ibuclaw: Likewise here with ``"eq"``

src/core/bitop.d
yebblies: These are the dmd backend op names.  Do you have any ideas on a good namespacing scheme for builtins?  Does it make sense to have some common ones?

Something like `digitalmars.*` `gcc.*` `llvm.*` `d.*`?

src/core/bitop.d
```diff
@@ -201,12 +213,15 @@ version (AnyX86)
      * using the X86 SSE4 POPCNT instruction.
      * POPCNT is not available on all X86 CPUs.
      */
+    pragma(DigitalMars_intrinsic, "popcnt")
     ushort _popcnt( ushort x ) pure;
```
ibuclaw: Unrelated to this PR I guess, but why not mark ``popcnt`` as an intrinsic instead?  Why this bizarrely named intrinsic here?

src/core/bitop.d
ibuclaw: Having a namespace for them could be a possibility.  But I would consider ``volatileLoad`` and ``volatileStore`` common between all compilers - unlike ``y2lx`` or ``cmpxchg`` - and so will need a meaningful name.

ibuclaw: Has this been taken to the NG?  I might need a few days to think it over, but maybe we should talk it out there.

yebblies: Hmm, so `dmd.yl2x` and `d.volatileLoad` or something...

With the current PR I have it's possible to do this:
```d
version (GDC)
{
    pragma(intrinsic, "gdc.volatileLoad", ...);
}
else version (DMD)
{
    pragma(intrinsic, "dmd.volatileLoad", ...);
}
```

ibuclaw: Yep.  But we should only deviate when we strictly need to. :-)

ZombineDev: How about `d.*` for intrinsics available on all compilers (e.g. `d.volatileLoad`) and `dmd.*`, `gdc.*`, `ldc.*`, etc. for non cross compiler portable ones?

PR #1368, opened 2015-08-29
Removed Deprecated Methods In core.time

JackStouffer:
These methods were scheduled for removal in June 2015


jmdavis:
Sorry, but no. This isn't how it works. Deprecated symbols don't go away this quickly. I grant you that the documentation says that the symbol will be removed, but it gets removed from the documentation, not the code. It goes from deprecated and documented to deprecated and _un_documented. So, the documentation needs to be removed and replaced with a non-ddoc comment saying that they'll be removed in something like June or August 2016. Then the symbol can be fully removed. So, it won't be documented for the next year, but it will still be around. At that point, hopefully, it will actually be removed, though Walter increasingly doesn't like anything being removed, so I don't know. In either case, we don't want to actually remove the symbol at this point. That's too fast. The way it's been working is that a symbol is deprecated and documented for about a year, and then it's deprecated and undocumented for about a year, and then it's removed. However, I haven't been on top of moving deprecations along as I should (or anything else in D lately, unfortunately).

So, thanks for the PR and the reminder that the date on the deprecation has passed, but this isn't how it's supposed to be handled (and it's not your fault that you don't know that). I'll just go create another PR that updates this and any other deprecations in druntime. I've been meaning to get around to it but haven't. I really need to figure out how to spend more time on D consistently...

jmdavis:
Okay. I created another PR to replace this one where it just removes the docs: https://github.com/D-Programming-Language/druntime/pull/1369

Again, thanks for reminding me about this, and sorry that I didn't get to it earlier.

JackStouffer:
Ok, thanks for removing it :)

PR #1369, opened 2015-08-29
Move deprecations along in core.time.

jmdavis:


src/core/time.d
```diff
@@ -1393,18 +1393,7 @@ public:
...
-              function that gave the desired behavior was $(LREF total). This
-              should make it more explicit and help prevent bugs. This function
-              will be removed in June 2015.)
-
-        Returns the number of the given units in this $(D Duration)
-        (minus the larger units).
-
-        $(D d.get!"minutes"()) is equivalent to $(D d.split().minutes).
-      +/
+    // Explicitly undocumented. It will be removed in August 2016. @@@DEPRECATED_2016-08@@@
```
DmitryOlshansky: Never understood the purpose of not documenting stuff. How would somebody get rid of deprecated thingie if there is no docs?

jmdavis: The deprecation message tells folks what to change their code to use. The documentation is unnecessary for that.

It's undocumented so that no one will see it in the documentation and choose to use it and so that it won't clutter the documentation. In theory, anyone who's actually maintaining their code will have already updated their code to not use this function, but by leaving it around as undocumented at least for now, we give folks more time to fix their code if they haven't yet while still having made it far less likely for new code to mistakenly use it.

In either case, the deprecation message tells folks what to use instead, and they can look at the documentation for that.

DmitryOlshansky: > but by leaving it around as undocumented at least for now, we give folks more time to fix their code if they haven't yet while still having made

And doing so without documentation on what it does in the first place makes porting old stuff that muh harder. Anyway I'm just wasting time on this as our discussion certianly won't change the way things are carried out.

DmitryOlshansky:
Auto-merge toggled on

PR #1370, opened 2015-08-31
fix Issue 14755 - Could -profile=gc also give the number of allocatio…

CyberShadow:
…ns that led to X bytes being allocated?

https://issues.dlang.org/show_bug.cgi?id=14755


DmitryOlshansky:
LGTM

DmitryOlshansky:
Auto-merge toggled on

PR #1371, opened 2015-09-01
fix Issue 14990 No rule to make target src/core/sys/stdio_msvc12.d

WalterBright:
Zip files broken again.

https://issues.dlang.org/show_bug.cgi?id=14990


9rnsr:
Auto-merge toggled on

PR #1372, opened 2015-09-01
mak/SRCS: Add all modules with struct definitions

CyberShadow:
Fixes linker errors when trying to define an array of structs.

Works around [issue 14992](https://issues.dlang.org/show_bug.cgi?id=14992) to fix [issue 14918](https://issues.dlang.org/show_bug.cgi?id=14918). Basically this just adds all modules which define structs to fix this class of errors throughout Druntime, but obviously a compiler fix would be better. Might be worth waiting to see if this can be fixed in the compiler.


CyberShadow:
Superseded by https://github.com/D-Programming-Language/dmd/issues/5014

PR #1373, opened 2015-09-02
Mark functions in core.stdc.math as Pure

JackStouffer:
As far as I know, all of the functions in `core.stdc.math` are all weekly pure, so it is safe to mark them as so. After this is pulled, many of the functions in `std.math` can be marked as pure, which would fix [issue 11320](https://issues.dlang.org/show_bug.cgi?id=11320).

### DIFF
--- a/src/core/stdc/math.d
+++ b/src/core/stdc/math.d
@@ -18,6 +18,7 @@ private import core.stdc.config;
 extern (C):
 @trusted: // All functions here operate on floating point and integer values only.
 nothrow:
+pure:
 @nogc:
 
 ///


yebblies:
`errno`?

ibuclaw:
Yes, errno is the reason why these can never be pure.

ibuclaw:
> After this is pulled, many of the functions in std.math can be marked as pure, which would fix issue 11320.

It wouldn't fix the issue because the code would be unverifiably lying to you.  The best way to make more of std.math pure would be to port or implement the functions yourself so we *know* that there's no side effects.  :-)

I'm going to close this on that note.

PR #1374, opened 2015-09-03
Add core.bitopt.bitswap(ulong)

mathias-lang-sociomantic:
The original (D1) core.bitop module had both the 32 and 64 bits versions, but only the 32 bits version was added to druntime.
This commit contains the original tango code of bitswap(ulong), only slightly modified using feedback from the 32 bits version.

FYI @mihails-strasuns-sociomantic

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -414,3 +414,75 @@ unittest
     foreach(i; 0 .. 32)
         assert(bitswap(1 << i) == 1 << 32 - i - 1);
 }
+
+/**
+ * Reverses the order of bits in a 64-bit integer.
+ */
+ulong bitswap ( ulong x ) pure @trusted
+{
+    version (D_InlineAsm_X86_64)
+    {
+        asm pure nothrow @nogc { naked; }
+
+        version (Win64)
+            asm pure nothrow @nogc { mov RAX, RCX; }
+        else
+            asm pure nothrow @nogc { mov RAX, RDI; }
+
+        asm pure nothrow @nogc
+        {
+            // Author: Tiago Gasiba.
+            mov RDX, RAX;
+            shr RAX, 1;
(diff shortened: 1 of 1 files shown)


yebblies:
> This commit contains the original tango code of bitswap(ulong)

Has this code been re-licensed under boost?

mathias-lang-sociomantic:
> Has this code been re-licensed under boost?

I worked under the assumption it'll be okay given it's the same author for both function, but I guess it depends on the agreement he gave to @WalterBright when the re licensing happened ?

src/core/bitop.d
```diff
@@ -414,3 +414,75 @@ unittest
...
+    version (D_InlineAsm_X86_64)
+    {
+        asm pure nothrow @nogc { naked; }
+
+        version (Win64)
+            asm pure nothrow @nogc { mov RAX, RCX; }
+        else
+            asm pure nothrow @nogc { mov RAX, RDI; }
+
+        asm pure nothrow @nogc
```
DmitryOlshansky: inline assmebly will make this function non-inlinable...

mihails-strasuns-sociomantic: it is same for existing `uint` overload https://github.com/D-Programming-Language/druntime/pull/1374/files#diff-6d81a130202753008e7b435a4ca71ebcL354

mihails-strasuns-sociomantic:
To clarify the context : this function is from same tango module and written by same author ( looks like @complexmath ) as existing http://dlang.org/phobos/core_bitop.html#.bitswap

Only intention of this PR is to minimize patch differences between upstream druntime and our internal distribution.

MartinNowak:
Auto-merge toggled on

MartinNowak:
It's literally a copy of the 32-bit version, so there won't be any problem.

mihails-strasuns-sociomantic:
Thanks!

PR #1377, opened 2015-09-04
Accept identifiers instead of strings in `DECLARE_ALTERNATE_NAME` macro.

denis-sh:
Stringizing operator ('#') adds extra quotes when used with string macro argument.

Causing commit: de0e1e10b4d88cc8e1657f326150f7565f4a9ef3, pull #1363.

### DIFF
--- a/src/rt/stdio_msvc.c
+++ b/src/rt/stdio_msvc.c
@@ -47,9 +47,9 @@ extern const char* _nullfunc = 0;
 #define DECLARE_ALTERNATE_NAME(name, alternate_name)  \
     __pragma(comment(linker, "/alternatename:" C_PREFIX #name "=" C_PREFIX #alternate_name))
 
-DECLARE_ALTERNATE_NAME ("__acrt_iob_func", "_nullfunc");
-DECLARE_ALTERNATE_NAME ("__iob_func", "_nullfunc");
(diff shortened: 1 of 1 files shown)


rainers:
LGTM. You could have also removed the bad `#` inide the macro, but as long as we don't try to map C++ names, this looks cleaner.

rainers:
Auto-merge toggled on

denis-sh:
> as long as we don't try to map C++ names, this looks cleaner.

Preprocessor has no idea about C/C++ and even has its own tokens (like `12abc`) so there is no problem in mapping keywords. Also no macro expansion takes place if there is a stringizing operator `#` (same for `##`) so identifier "conflicts" with defined macros (like `__LINE__`) also doesn't matter.

rainers:
C++ mangled names contain `?` and `@`, I'm not so sure every preprocesor likes these and behaves correctly. We'll see ones we get there...

denis-sh:
> C++ mangled names contain ? and @, I'm not so sure every preprocesor likes these and behaves correctly. We'll see once we get there...

Didn't think about it. Thanks for the explanation.
`?` should work as it is valid preprocessor token and `#` operator (probably) shouldn't add spaces if there is no white-space between tokens. `@` isn't valid *C/C++* token but Microsoft compiler seems to accept it.

PR #1378, opened 2015-09-06
filter on basename of compilers

MartinNowak:
- fixes make CC=/path/to/cc
- fixes Issue 15012

[Issue 15012 – Druntime Makefile whitelists compilers](https://issues.dlang.org/show_bug.cgi?id=15012)

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -40,17 +40,14 @@ endif
 DDOCFLAGS=-conf= -c -w -o- -Isrc -Iimport -version=CoreDdoc
 
 # Set CFLAGS
-CFLAGS=
-ifneq (,$(filter cc% gcc% clang% icc% egcc%, $(CC)))
-	CFLAGS += $(MODEL_FLAG) -fPIC -DHAVE_UNISTD_H
-	ifeq ($(BUILD),debug)
-		CFLAGS += -g
-	else
-		CFLAGS += -O3
-	endif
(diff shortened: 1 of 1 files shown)


MartinNowak:
ping @CyberShadow

CyberShadow:
Thanks Martin. But why is the whitelisting needed in the first place? It doesn't strike me as good practice, anyone can come up with a new compiler or wrapper with gcc-compatible command-line parameters. The "command-line-compatible wrapper" naming convention also seems to have the opposite naming scheme, e.g. `clang-cl` is clang's cl.exe-compatible wrapper.

MartinNowak:
You're right, let's remove this. It's still possible to override CFLAGS if necessary.

ibuclaw:
Pet peeve question, why ``-O3``?  This is an optimisation that has been proven in the past to do aggressive loop unrolling and inlining that make the program no longer fit in the instruction cache, producing worse code than what is done with ``-O2``.

Maybe someone should do some fact-finding with benchmarks... :-)

MartinNowak:
Well there isn't any performance sensitive C code in druntime.

MartinNowak:
Can we also keep the OT discussion out of this PR?

CyberShadow:
Just tested, it fixes everything for me, merging.

CyberShadow:
Oops, forgot to merge via auto-tester, but it was green on both PRs.

PR #1379, opened 2015-09-06
use linker generated section brackets

MartinNowak:


WalterBright:
Auto-merge toggled on

jpf91:
@MartinNowak binutils ld/gold on linux emits these symbols as well: https://sourceware.org/binutils/docs/ld/Orphan-Sections.html Unfortunately binutils does not keep these symbols hidden/local to shared libraries. Just out of curiosity, does Solaris/Android export these as hidden symbols?

MartinNowak:
> @MartinNowak binutils ld/gold on linux emits these symbols as well:

Yes, that's what we're using here.

> Unfortunately binutils does not keep these symbols hidden/local to shared libraries.

The linker uses the visibility of the reference, so they are hidden b/c we reference it like that in dmd.
https://github.com/D-Programming-Language/dmd/blob/6a062c6ca37dce06533167b074277320bc8f9ec2/src/backend/elfobj.c#L3141

PR #1383, opened 2015-09-11
Header file for Linux-specific timerfd API

adamdruppe:
I want to use this in a project... it appeared about the same time as epoll and has a similarly small header file so I just followed that and ported the C headers.

### DIFF
--- a/src/core/sys/linux/timerfd.d
+++ b/src/core/sys/linux/timerfd.d
@@ -1,0 +1,24 @@
+/**
+ * D header file to interface with the Linux timefd API <http://man7.org/linux/man-pages/man2/timerfd_create.2.html>
+ * Available since Linux 2.6
+ *
+ * License : $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ */
+module core.sys.linux.timerfd;
+
+version (linux):
+
+public import core.sys.posix.time;
+
+extern (C):
+@system:
+@nogc:
+nothrow:
+
+int timerfd_create(int clockid, int flags);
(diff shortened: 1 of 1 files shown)


adamdruppe:
Why is the test failing?

```
make[2]: execvp: ./obj/linux/32/both: Permission denied
make[2]: *** [obj/linux/32/both.done] Error 127

diff: ./obj/freebsd/32/bothgc.log: No such file or directory
[snip]
```

I try not to curse anymore, so I think I had best just walk away for tonight.

src/core/sys/linux/timerfd.d
```diff
@@ -0,0 +1,24 @@
...
+ * D header file to interface with the Linux timefd API <http://man7.org/linux/man-pages/man2/timerfd_create.2.html>
+ * Available since Linux 2.6
+ *
+ * License : $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ */
+module core.sys.linux.timerfd;
+
+version (linux):
+
+public import core.sys.posix.time;
```
DmitryOlshansky: public imports are dangerous ;) There is at least one bug with them still open

adamdruppe: On Sat, Sep 12, 2015 at 03:41:57AM -0700, Dmitry Olshansky wrote:
> public imports are dangerous ;) There is at least one bug with them still open

Perhaps we should fix that bug!

I could make it private, but the file is basically useless without that
itimerspec definition, and the file is straight up #included in the C
source so bugs notwithstanding, the public import is the closest match.

DmitryOlshansky:
LGTM

DmitryOlshansky:
Auto-merge toggled on

PR #1384, opened 2015-09-11
Properly free module info when using linker-generated section brackets

joakim-noah:
The recently-merged support for linker-generated section brackets (#1379) causes a segfault on Android, when `rt_term` tries to free module info that is no longer malloced.  I'm guessing it broke for Solaris too, though I haven't tested it.  This PR seems to fix it- all druntime and phobos tests on Android/x86 and linux/x86 pass locally- let me know if any further tweaks are necessary.

### DIFF
--- a/src/core/sys/bionic/fcntl.d
+++ b/src/core/sys/bionic/fcntl.d
@@ -1,0 +1,5 @@
+module core.sys.bionic.fcntl;
+
+version(CRuntime_Bionic) extern(C) nothrow @nogc:
+
+enum LOCK_EX = 2;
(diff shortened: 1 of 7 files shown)


joakim-noah:
Looks like setting `_modules` to `null` in `rt.minfo` was causing a segfault in the debug build on linux, so I got rid of it.  I had only tested locally with a release build, which is why I didn't see the issue, but I can reproduce with a debug build.  Looks like the cause is the invariant in `rt.sections_elf_shared`, which is called after the modules are removed, likely when the destructor is run.  I don't know why it only triggered on one module from the debug build, `std.parallelism`, but getting rid of the null assignment should fix the issue.

joakim-noah:
[As @MartinNowak asked for in a previously-merged PR thread](https://github.com/D-Programming-Language/druntime/pull/1290#issuecomment-140199911), moved some non-Posix bionic declarations into their own directory.

MartinNowak:
Auto-merge toggled on

redstar:
@klickverbot I'll check this on Solaris.

PR #1386, opened 2015-09-16
Bug 15073: statically-initialize SIGRTMIN/MAX

tomerfiliba:
They were aliases to private functions, which meant they never worked. This fixes https://issues.dlang.org/show_bug.cgi?id=15073

### DIFF
--- a/src/core/sys/posix/signal.d
+++ b/src/core/sys/posix/signal.d
@@ -108,10 +108,24 @@ union sigval
 version( Solaris )
 {
     import core.sys.posix.unistd;
-    private int _sigrtmin() { return cast(int) sysconf(_SC_SIGRT_MIN); }
-    private int _sigrtmax() { return cast(int) sysconf(_SC_SIGRT_MAX); }
+
+    @property int SIGRTMIN() nothrow @nogc {
+        static int sig = -1;
+        if (sig == -1) {
+            sig = cast(int)sysconf(_SC_SIGRT_MIN);
+        }
+        return sig;
+    }
+
+    @property int SIGRTMAX() nothrow @nogc {
+        static int sig = -1;
+        if (sig == -1) {
+            sig = cast(int)sysconf(_SC_SIGRT_MAX);
+        }
+        return sig;
+    }
 }
-else version( Posix )
+else version( linux )
 {
     private extern (C) nothrow @nogc
     {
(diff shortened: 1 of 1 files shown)


MartinNowak:
We don't currently compile that module but now it has a module constructor.
You need to add the source to https://github.com/D-Programming-Language/druntime/blob/master/mak/SRCS to make this work.

tomerfiliba:
@MartinNowak - ``core\sys\posix\signal.d`` is already in the ``SRCS`` file (missed that in my second commit). The link errors are on all other modules it imports basically - core.sys.posix.fcntl, core.sys.posix.time, etc. As I'm not knowledgeable enough about the build system, I'll just convert the immutable variables to properties that do it in runtime.

tomerfiliba:
It appears FreeBSD/OSX do not support realtime signals:
https://github.com/PromyLOPh/pianobar/issues/256#issuecomment-5694709
https://books.google.co.il/books?id=K8vUkpOXhN4C&pg=PA1132&lpg=PA1132&dq=osx++sigrtmin&source=bl&ots=OLlhTZSt_w&sig=CfEXCdSglEasRUs3V8863r60elg&hl=en&sa=X&redir_esc=y#v=onepage&q&f=false

MartinNowak:
FreeBSD introduced them with 7.0, but they are constants.
https://github.com/freebsd/freebsd/blob/e79c62ff68fc74d88cb6f479859f6fae9baa5101/sys/sys/signal.h#L117

tomerfiliba:
added those as enums

denis-sh:
`{` should be on its own line. @tomerfiliba, please, open a style fixup pull.

nemanja-boric-sociomantic:
How does this code doesn't introduce race condition? There mere multiple threads assigning value to `sig` as far as I can see.

nemanja-boric-sociomantic:
@MartinNowak could you take a look into this? If more threads require the symbol values at the same time, I suspect this code is prone to race condition. I don't think the previous code was performing bad, and it was correct.

MartinNowak:
Yes, it's a race condition @nemanja-boric-sociomantic, but the value should always be the same and setting an integer (32-bit) is atomic on any relevant architecture (could still use atomicStore!(MemoryOrder.raw)).
I wouldn't want to use `shared static this()` in a C header, and I'm also unsure about the `sysconf` overhead, so lazily caching the result seems sensible.

PR #1389, opened 2015-09-17
fix spurious test failures

MartinNowak:
- use common.mak to test debug/release in different ROOTs


yebblies:
This either doesn't fix them or it's referring to different spurious failures.

schveiguy:
I didn't see this, but I think my fix is more comprehensive: #1390

PR #1392, opened 2015-09-19
Use new version identifier SystemZ instead of S390X.

redstar:
The 64-bit architecture is called System z since 2000.


DmitryOlshansky:
Auto-merge toggled on

PR #1397, opened 2015-09-24
Fix Issue 15104 - Fiber context switch in finally blocks breaks EH

klickverbot:
https://issues.dlang.org/show_bug.cgi?id=15104

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -107,6 +107,17 @@ private
     //
     extern (C) void  rt_moduleTlsCtor();
     extern (C) void  rt_moduleTlsDtor();
+
+    /**
+     * Hook for whatever EH implementation is used to save/restore some data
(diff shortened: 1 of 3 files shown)


ibuclaw:
:+1: - Fix works for gdc too.

MartinNowak:
Can we store it as Fiber field and on the stack instead? I want to get rid of the Context within chain at some point.

MartinNowak:
I want to keep the array of Contexts with stacktop/bottom for the GC, the nested push/pop Context is only needed to find out whether the stack pointer during a GC suspend is in the thread's stack or the current Fiber's stack. This can be done by checking the stack range instead.
But anyhow, keeping this variable on the stack is simple enough.

MartinNowak:
I think you can because the context switching itself does not throw.

PR #1398, opened 2015-09-26
Allow functions that override _d_throwc to call original implementation

Marenz:
Implements what I suggested here: http://forum.dlang.org/thread/uwfacejyzpyjwrzvlspg@forum.dlang.org

Allows users to override `_d_throwc` but still let's them call the original implementation (`onThrow`) in this case. 

I chose `onThrow` because it follows the pattern used for [`_d_assert`](https://github.com/D-Programming-Language/druntime/blob/bce3f9b39396cbdd34569b0ef9e5e0d1f4111041/src/core/exception.d#L605) which calls `OnAssertError`

### DIFF
--- a/src/rt/deh_win32.d
+++ b/src/rt/deh_win32.d
@@ -614,7 +614,7 @@ private void throwImpl(Object h)
                    1, cast(void *)&h);
 }
 
-extern(C) void _d_throwc(Object h)
+extern(C) void onThrow(Object h)
 {
     // set up a stack frame for trace unwinding
     version (AsmX86)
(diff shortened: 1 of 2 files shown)


src/rt/deh_win32.d
```diff
@@ -614,7 +614,7 @@ private void throwImpl(Object h)
                    1, cast(void *)&h);
 }
 
-extern(C) void _d_throwc(Object h)
+extern(C) void onThrow(Object h)
```
klickverbot: Having `onThrow` without any prefix does not seem like a good idea, hygiene-wise.

MartinNowak:
If you want to change the language semantics for throwing errors to abort, then you should lobby for that change. If you don't want to wait please maintain this hack in your own druntime copy.
While I partly agree about this change (and we recently introduced abort), first constructing an Error and then aborting is a weird implementation.

MartinNowak:
See #1337 for the abort function.

PR #1400, opened 2015-10-02
Make the GC proxy usable for proxying.

oridb:
If you want to do tracing of the GC, or anything that actually
proxies the GC, the you need the _gc struct to forward the
calls to. the wrappers by default call through to themselves,
which leads to infinite recursion.

Adding the gc_ getGC() function allows you to acutally call
_gc.foo() from within your proxy.

### DIFF
--- a/src/gc/proxy.d
+++ b/src/gc/proxy.d
@@ -314,6 +314,11 @@ extern (C)
 
     export
     {
+        GC* gc_getGC()
+        {
+            return &_gc;
+        }
+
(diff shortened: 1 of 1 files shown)


DmitryOlshansky:
There were talks of plain removing of GC proxy, no idea if that's good or bad or if it's even still planed .

MartinNowak:
Closed in favor of https://github.com/dlang/druntime/pull/1581.

PR #1401, opened 2015-10-02
fix Issue 15137 - core.time: Support Duration/Duration and Duration%Duration

CyberShadow:
https://issues.dlang.org/show_bug.cgi?id=15137

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -518,13 +518,14 @@ public:
 
 
     /++
-        Adds or subtracts two durations.
+        Adds, subtracts or calculates the modulo of two durations.
 
         The legal types of arithmetic for $(D Duration) using this operator are
 
         $(TABLE
         $(TR $(TD Duration) $(TD +) $(TD Duration) $(TD -->) $(TD Duration))
         $(TR $(TD Duration) $(TD -) $(TD Duration) $(TD -->) $(TD Duration))
+        $(TR $(TD Duration) $(TD %) $(TD Duration) $(TD -->) $(TD Duration))
         $(TR $(TD Duration) $(TD +) $(TD TickDuration) $(TD -->) $(TD Duration))
         $(TR $(TD Duration) $(TD -) $(TD TickDuration) $(TD -->) $(TD Duration))
         )
@@ -533,9 +534,8 @@ public:
             rhs = The duration to add to or subtract from this $(D Duration).
       +/
     Duration opBinary(string op, D)(D rhs) const nothrow @nogc
-        if((op == "+" || op == "-") &&
-           (is(_Unqual!D == Duration) ||
-            is(_Unqual!D == TickDuration)))
+        if(((op == "+" || op == "-" || op == "%") && is(_Unqual!D == Duration)) ||
+           ((op == "+" || op == "-") && is(_Unqual!D == TickDuration)))
     {
         static if(is(_Unqual!D == Duration))
             return Duration(mixin("_hnsecs " ~ op ~ " rhs._hnsecs"));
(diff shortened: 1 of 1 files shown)


MetaLang:
LGTM, but should division by Duration(0) be checked for?

CyberShadow:
Should it?

Personally, I think throwing TimeException when dividing by 0 is a mistake. I really don't think core.time has any business checking for division by 0 when the divisor comes straight from the user. If the user is dividing by zero, let him divide by zero - the language will react quite appropriately to that and throw a "division by zero" error, which is the correct response to such a logic error.

It is too late to change that now, and since I see no merit in adding an additional check in core.time, the only argument in favor that I see is consistency with `opBinary!"/"(long)`.

MetaLang:
I was thinking an assert, but I guess that it will be taken care of down the line anyway, so LGTM.

schveiguy:
Should the result of Duration/Duration or Duration % Duration be another Duration? I would think it should be a straight `long`.

CyberShadow:
Duration/Duration is long, Duration%Duration is Duration. Why would Duration%Duration be long?

schveiguy:
> Why would Duration%Duration be long?

I didn't think about it properly, yeah it should be Duration. I thought it should be the same as the division type (and I mistakenly thought you were saying both should be Duration). But thinking about it, the leftover should be the same scale as the original type (and it is).

LGTM.

src/core/time.d
```diff
@@ -551,23 +553,31 @@ public:
...
 
                 assert((cast(D)Duration(-5)) + (cast(E)Duration(-7)) == Duration(-12));
                 assert((cast(D)Duration(-5)) - (cast(E)Duration(-7)) == Duration(2));
+                assert((cast(D)Duration(-5)) % (cast(E)Duration(7)) == Duration(-5));
                 assert((cast(D)Duration(-7)) + (cast(E)Duration(-5)) == Duration(-12));
                 assert((cast(D)Duration(-7)) - (cast(E)Duration(-5)) == Duration(-2));
+                assert((cast(D)Duration(-7)) % (cast(E)Duration(5)) == Duration(-2));
             }
 
             foreach(T; _TypeTuple!(TickDuration, const TickDuration, immutable TickDuration))
```
schveiguy: Why no tests for % and `TickDuration`?

CyberShadow: TickDuration is going away (replaced by MonoTime), it seems to me that it's pointless writing new code to support it.

src/core/time.d
```diff
@@ -611,7 +622,7 @@ public:
                   subtract this $(D Duration) from.
       +/
     Duration opBinaryRight(string op, D)(D lhs) const nothrow @nogc
-        if((op == "+" || op == "-") &&
+        if((op == "+" || op == "-" || op == "%") &&
             is(_Unqual!D == TickDuration))
     {
         return Duration(mixin("lhs.hnsecs " ~ op ~ " _hnsecs"));
```
schveiguy: Why no tests for this one either?

src/core/time.d
```diff
@@ -698,8 +711,10 @@ public:
         {
             test1!"+="(Duration(5), (cast(E)Duration(7)), Duration(12));
             test1!"-="(Duration(5), (cast(E)Duration(7)), Duration(-2));
+            test1!"%="(Duration(5), (cast(E)Duration(7)), Duration(5));
             test1!"+="(Duration(7), (cast(E)Duration(5)), Duration(12));
             test1!"-="(Duration(7), (cast(E)Duration(5)), Duration(2));
+            test1!"%="(Duration(7), (cast(E)Duration(5)), Duration(2));
 
             test1!"+="(Duration(5), (cast(E)Duration(-7)), Duration(-2));
             test1!"-="(Duration(5), (cast(E)Duration(-7)), Duration(12));
```
schveiguy: here too

src/core/time.d
CyberShadow: As above, plus I think the test coverage in core.time and std.datetime is quite over the top as it is. Adding tests here will not improve test coverage in any way.

src/core/time.d
schveiguy: OK, but it seems rather odd to add code paths that won't be tested.

src/core/time.d
CyberShadow: As above. Operation with signed operands is already tested in above unit tests, additionally as there is no special logic for the operands' sign, we are essentially testing the CPU's signed integer division here.

src/core/time.d
schveiguy: Per your explanation, why even add this update?

CyberShadow: Oops, you're right. I decided it wasn't worth supporting `TickDuration` at the point when I got to writing tests :) Should I take it out?

src/core/time.d
schveiguy: The test is not so much a test that the CPU works, as it is testing the ability to compile properly (IMO). The time tests are somewhat repetitive, I concur on that.

CyberShadow: > as it is testing the ability to compile properly

How so? `opBinary` will receive the same type every time

src/core/time.d
schveiguy: Considering that we just talked about how `typeof(a % b) == typeof(a)`, I'm thinking this support for `TickDuration % Duration -> Duration` is ill advised anyway. I think it will not be a bad thing to see `TickDuration` and this support go.

CyberShadow:
OK, removed all TickDuration support!

src/core/time.d
schveiguy: Moot point now, but as a response, the string parameter isn't the same. Imagine if someone altered the template constraint to somehow reject the call, but we weren't testing it, breaking working code.

schveiguy:
LGTM. I assume @MetaLang is ok with updates.

schveiguy:
Auto-merge toggled on

src/core/time.d
CyberShadow: No, it actually is. Notice that in the prior code, all variation in the lines below is simply changing the sign in the Duration constructor's parameter. The `test1` function is always called as `test1!op(Duration, E, Duration)`. `op` is always `+=`, `-=` or `%=` and we already tested the `%=` case.

schveiguy: This is back when TickDuration was enabled as a code path, but not tested. So the call `test2!op(Duration, TickDuration, Duration, Duration)` wasn't ever tried for op == `%=`. The code has changed, and `Duration %= TickDuration` is no longer enabled.

CyberShadow: Ah OK, I misunderstood you completely. I thought you were talking about the rest of the Duration tests, as I only added two.

schveiguy: Note: the reason I commented up here instead of the context where it really should be is because github doesn't let you comment on sections of code that haven't been modified. So that was probably confusing, sorry.

CyberShadow:
Thanks for the prompt review MetaLang & Steven, hopefully @jmdavis will be OK with this too.

schveiguy:
I can untoggle if you want to wait for him.

CyberShadow:
Yep, probably a good idea. Unless it's an urgent fix, PRs should probably stay open for at least 24 hours so everyone around the globe gets a chance to have a look.

CyberShadow:
Auto-merge toggled off

schveiguy:
> PRs should probably stay open for at least 24 hours so everyone around the globe gets a chance to have a look

right, Sounds good.

jmdavis:
Hmmm. I think that when I originally wrote `Duration`, I didn't see how dividing a `Duration` by a `Duration` made any sense. Thinking it through now, I guess that it does, though I have no clue what you'd be using it for. But if it makes sense from an arithmetic perspective, I see no reason not to add it.

> Personally, I think throwing TimeException when dividing by 0 is a mistake.

It probably was, and honestly, I'd have no problem changing it. The only code that would be negatively affected would be code that does no checks for `0`, expecting to catch the exception in that case, and I seriously doubt that anyone is doing that (and it's arguably bad practice anyway). It's much more likely that code doesn't check for `0` and doesn't try and handle the `TimeException` either. So, there might be some risk in changing it, but I expect that that risk is quite low and worth the improvement to everyone else's code.

CyberShadow:
> though I have no clue what you'd be using it for

FWIW, I provided two examples I ran into in the linked issue.

jmdavis:
Auto-merge toggled on

PR #1402, opened 2015-10-03
Add Windows headers from the win32 bindings project

CyberShadow:
Plan of attack:

- [x] To prevent *any* code breakage, create a comprehensive test of the exact signatures of declarations in core.sys.windows.windows
- [x] Add the win32 headers as-is
- [x] Through a series of commits, incrementally beat them into shape suitable for inclusion in Druntime
- [x] Get all tests to pass (mostly involving conflicts due to duplicate declarations, currently blocked by a DMD regression)
- [x] Add .lib files to the installer to allow 32-bit (OMF) Windows apps to use these bindings (Post-merge)

Hopefully, satisfying the comprehensive test should also imply that Phobos, Druntime and other projects will be able to use the new headers with no changes.

This is similar to https://github.com/D-Programming-Language/druntime/pull/1076 but:

 - uses a more thorough approach to avoid code breakage;
 - all modifications from the win32 bindings are small and self-contained, thus should be reviewable and repeatable.

Open questions:

- [x] Licensing: currently, the headers' license is "public domain". We could change this to Boost.  
Additionally, this excludes all files that were not explicitly listed as public domain (winsock, winsock2 and DirectX headers).
- [x] What changes can be made that must be done now (i.e. which would break code if were to be done later)? E.g. introducing usage D features such as named enums.


CyberShadow:
The autotester failures don't happen on my machine, but I am using a somewhat older compiler... another regression? Reducing now...

CyberShadow:
Found it: https://issues.dlang.org/show_bug.cgi?id=15150

CyberShadow:
Tests are green!

I'd like to do an aggressive check for name collisions due to duplicate declarations before this is merged, though. I think that's the main (only?) possible remaining cause of regressions.

adamdruppe:
This is looking splendid!

DmitryOlshansky:
> I think that's the main (only?) possible remaining cause of regressions.

Well our user may be using a bit of self-defined declarations but even then it should still prefer local decls to imported ones.. I think we are in a good shape here.

ZombineDev:
@CyberShadow Is there anything left to be done, except to set choose license the headers will have, that is blocking this PR?

CyberShadow:
I would like to make a pass through the duplicate symbols. I already wrote the test, just a matter of fixing things. Will do this today/tomorrow-ish.

ZombineDev:
@CyberShadow Nice!

CyberShadow:
OK, I think this is ready.

CyberShadow:
Added .def files and import .lib generation.

Installer pull: https://github.com/D-Programming-Language/installer/pull/157

CyberShadow:
Discussion thread on open questions:
http://forum.dlang.org/post/cakjjasivxcvfdroiadp@forum.dlang.org

AndrejMitrovic:
Will the commits be squashed together or should they all be reviewed individually?

DmitryOlshansky:
I believe it wasn't squashed intentionally to make it easier to review each change

DmitryOlshansky:
@AndrejMitrovic see also:

> Add the win32 headers as-is
> Through a series of commits, incrementally beat them into shape suitable for inclusion in Druntime

CyberShadow:
> I believe it wasn't squashed intentionally to make it easier to review each change

Correct.

CyberShadow:
Following the forum discussion, changed license to Boost. I think I'm done, this is good to merge!

AndrejMitrovic:
I think we should go ahead with this and then submit any fixup PRs when we find issues. I plan on adapting this to the DWinProgramming repo (I'll re-host it soon). 

Otherwise it's way too much code to meticulously review every new line. :)

DmitryOlshansky:
Auto-merge toggled on

DmitryOlshansky:
@AndrejMitrovic agreed. Better yet we have full release cycle to test things out

AndrejMitrovic:
I don't know how many people were involved with this, but thanks goes out to everyone! This took many years, didn't it. :)

yebblies:
https://github.com/D-Programming-Language/dmd/pull/4396 is showing some errors with duplicated declarations from this PR: https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=1792274&isPull=true

WalterBright:
The files were not added to the `mak/MANIFEST`.

Jebbs:
We can close issue 317 with the inclusion of this pr, yes?

CyberShadow:
Ah yes, done.

PR #1404, opened 2015-10-09
forgot to add timerfd to the makefile

adamdruppe:
The druntime build system is error-prone. Having to duplicate the file names over so many places when it really should just do something like *.d means these things are forgotten and almost slip by the beta unnoticed!

Without this, the file is never copied from src/ over to import/ (why does it even use separate folders for that anyway? pointless waste of time), nor is the init blob included in the object file, meaning the structs need to be void initialized to avoid linker errors! (I noticed that with epoll too and shoved its line in here too.)

I recommend these makefiles be rewritten. I knew the Windows one was laughably bad, but I didn't realize the posix one made many of the same mistakes. But until then, this keeps my new addition from being broken on new installs.

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -64,8 +64,10 @@ COPY=\
 	$(IMPDIR)\core\sys\linux\execinfo.d \
 	$(IMPDIR)\core\sys\linux\fcntl.d \
 	$(IMPDIR)\core\sys\linux\link.d \
+	$(IMPDIR)\core\sys\linux\stdio.d \
 	$(IMPDIR)\core\sys\linux\termios.d \
 	$(IMPDIR)\core\sys\linux\time.d \
+	$(IMPDIR)\core\sys\linux\timerfd.d \
 	$(IMPDIR)\core\sys\linux\tipc.d \
 	$(IMPDIR)\core\sys\linux\unistd.d \
 	\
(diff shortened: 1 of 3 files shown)


DmitryOlshansky:
> The druntime build system is error-prone.

Hell yeah.
On the point of the pull itself - LGTM

> I knew the Windows one was laughably bad, but I didn't realize the posix one made many of the same mistakes. But until then, this keeps my new addition from being broken on new installs.

I humbly wait untill reggae or simillar D-based build tool. Even Boost C++ ("just" a library) has its own build tool (Boost.Build) b/c makefiles etc. are major PITA.

DmitryOlshansky:
This likely should target stable branch CC @MartinNowak

MartinNowak:
Picked into stable for 2.070.0 c101e685f41a74daafe909645d96c690523adc79.

tsbockman:
@MartinNowak I think this PR messed up the druntime build on [Win_32](https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=1915232&isPull=true) and [Win_32_64](https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=1915248&isPull=true).

MartinNowak:
Sorry my bad #1466.

Why does the tester spam around lots of green retests instead of waiting for the failing platform to be tested?

braddr:
Because the tester can't know why things are broken.  And even if mainline is broken, chances are that at least one of the open PR's is intended to fix that breakage.

PR #1405, opened 2015-10-10
Treat division by 0 as an error in Duration and consolidate code.

jmdavis:
Duration should never have checked for division by 0 and thrown a
TimeException when it occurred. The OS/hardware already checks for that,
and it should be considered a logic error. Checking for it and possibly
throwing just slows the code down and makes it so that it can't be @nogc.

This is technically a breaking change, but it will only break code which
doesn't bother to avoid dividing by 0 and then catches TimeException
when it occurs, and such code is likely extremely rare, if it exists at
all. Given the fact that the odds of actually breaking code are
extremely low and that this allows us to make Duration fully @nogc except
for toString, I think that it's worth the risk.

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -465,7 +465,6 @@ public:
             return -1;
         if(_hnsecs > rhs._hnsecs)
             return 1;
-
         return 0;
     }
 
@@ -681,7 +680,6 @@ public:
             mixin("_hnsecs " ~ op ~ "= rhs._hnsecs;");
         else if(is(_Unqual!D == TickDuration))
             mixin("_hnsecs " ~ op ~ "= rhs.hnsecs;");
-
         return this;
     }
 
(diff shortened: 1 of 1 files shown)


DmitryOlshansky:
LGTM and nice code consolidation btw

MartinNowak:
Auto-merge toggled on

denis-sh:
IMHO breaking change pull should have a link to a corresponding changelog pull.

schveiguy:
> IMHO breaking change pull should have a link to a corresponding changelog pull.

I agree, but to that end, I believe we now have individual changelog files for each sub-project. We should stop accepting pulls without a changelog entry when one is warranted.

jmdavis:
Enhancement request created and marked as FIXED so that this will show up in the changelog: https://issues.dlang.org/show_bug.cgi?id=15221

PR #1406, opened 2015-10-12
fix Issue 14993 - segfault for InvalidMemoryOperationError

MartinNowak:
- creating a stacktrace currently uses the GC (thus might recurse
  infinetely when the GC throws an exception itself)
- suppress stack traces for OutOfMemoryError,
  InvalidMemoryOperationError, and FinalizeError

[Issue 14993 – Allocating in a destructor segfaults instead of throwing InvalidMemoryOperationError](https://issues.dlang.org/show_bug.cgi?id=14993)

### DIFF
--- a/src/core/exception.d
+++ b/src/core/exception.d
@@ -130,6 +130,7 @@ class FinalizeError : Error
     this( TypeInfo ci, string file = __FILE__, size_t line = __LINE__, Throwable next = null ) @safe pure nothrow @nogc
     {
         super( "Finalization error", file, line, next );
+        super.info = SuppressTraceInfo.instance;
         info = ci;
     }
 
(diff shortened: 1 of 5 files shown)


schuetzm:
Is it possible to suppress the stack traces only if the GC is actually running currently?

MartinNowak:
That's tricky, OutOfMemoryError is the only one used outside of the GC. We could add a bool flag to onOutOfMemory, but a stack trace for when your memory is used up is misleading anyhow b/c the last allocation isn't the cause of the error.

yebblies:
> but a stack trace for when your memory is used up is misleading anyhow b/c the last allocation isn't the cause of the error.

Well, it could be.  Quite often the cause is a much-too-big allocation or much-too-many of the same allocation.  Depending on the application, the stack trace can at least tell you how far it got before running out of memory, which is nice information to have if it's possible.

MartinNowak:
Now only suppresses stack traces for OutOfMemoryError when called by the GC.

MartinNowak:
Going ahead with this so I can build the next beta.
If you have anything else please follow up on this PR.

PR #1412, opened 2015-10-18
issue 14327: move call to init_msvc() into rt_init()

rainers:
https://issues.dlang.org/show_bug.cgi?id=14327

stdin,stdout,stderr handles are only initialized when running d_run_main, but not rt_init.

### DIFF
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -164,6 +164,9 @@ extern (C) int rt_init()
        rt_init. */
     if (atomicOp!"+="(_initCount, 1) > 1) return 1;
 
+    version (CRuntime_Microsoft)
+        init_msvc();
+
     _d_monitor_staticctor();
     _d_critical_init();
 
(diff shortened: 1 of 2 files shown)


rainers:
Not sure if this should be against stable instead. It's not a regression, though we now support a new platform (win32/mscoff) that doesn't work in this use case.

slimshader:
To clarify: I originally reported this issue for x64 app. Now I also tested mscoff32 version to see if things were resolved in 2.069

ZombineDev:
@rainers It's better to target stable, so that it can get more real world testing and what is maybe more important - so we can have better out-of-the-box experience on Windows sooner rather than later, something many have been asking for.

MartinNowak:
I picked it over to stable de2788d911129545531e0e42d1c8630b3f406e7c.

rainers:
> I picked it over to stable de2788d.

Thanks.

> To close the bugzilla ticket you should use fix Issue 14327, then it will also show up in the changelog.

I wasn't sure it completely fixes the issue, because there is also the trouble with linker errors without D main. Rereading the bug report now, that's not exactly in the bug itself, but in a comment.

PR #1413, opened 2015-10-19
add mixin template StandardExceptionConstructor

jamadagni:
Convenience mixin for trivially subclassing exceptions. Documentation and unittest included.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -1760,15 +1760,17 @@ unittest
  *
  * Example:
  * ---
- * class MeaCulpa: Exception {
+ * class MeaCulpa: Exception
+ * {
  *     mixin StandardExceptionConstructor;
  * }
  * assertThrown!MeaCulpa(enforce!MeaCulpa(false));
  *
  * try throw new MeaCulpa;
- * catch (MeaCulpa e) {
+ * catch (MeaCulpa e)
+ * {
  *     assert(e.file == __FILE__);
- *     assert(e.line == __LINE__ - 3);
+ *     assert(e.line == __LINE__ - 4);
  * }
  * ---
  *
@@ -1787,15 +1789,18 @@ unittest
 {
     import std.exception;
 
-    class MeaCulpa: Exception {
+    class MeaCulpa: Exception
+    {
         mixin StandardExceptionConstructor;
     }
     assertThrown!MeaCulpa(enforce!MeaCulpa(false));
 
     try throw new MeaCulpa;
-    catch (MeaCulpa e) {
+    catch (MeaCulpa e)
+    {
         assert(e.file == __FILE__);
-        assert(e.line == __LINE__ - 3);
+        assert(e.line == __LINE__ - 4);
+    }
 }
 
 
--- a/src/object.d
+++ b/src/object.d
@@ -1751,6 +1751,54 @@ unittest
 }
 
 
+/**
+ * Convenience mixin for trivially sub-classing exceptions
+ *
+ * Even trivially sub-classing an exception involves writing boilerplate code
+ * for the constructor to correctly pass in the source file and line number the
+ * exception was thrown from. This mixin makes it easier.
+ *
+ * Example:
+ * ---
+ * class MeaCulpa: Exception {
+ *     mixin StandardExceptionConstructor;
+ * }
+ * assertThrown!MeaCulpa(enforce!MeaCulpa(false));
+ *
+ * try throw new MeaCulpa;
+ * catch (MeaCulpa e) {
+ *     assert(e.file == __FILE__);
+ *     assert(e.line == __LINE__ - 3);
+ * }
+ * ---
+ *
+ * Note that if you want to pass in special arguments to your exception, you
+ * obviously can't use this mixin and have to write the constructor manually.
+ */
+mixin template StandardExceptionConstructor()
+{
+    this(string msg, string file = __FILE__, size_t line = __LINE__)
+    {
+        super(msg, file, line);
+    }
+}
+
+unittest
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -1751,6 +1751,59 @@ unittest
...
+ *
+ * Note that if you want to pass in special arguments to your exception, you
+ * obviously can't use this mixin and have to write the constructor manually.
+ */
+mixin template StandardExceptionConstructor()
+{
+    this(string msg, string file = __FILE__, size_t line = __LINE__)
+    {
+        super(msg, file, line);
+    }
```
jmdavis: I still think that this is a waste of time, because you the constructors that get mixed in can't be documented, but if you're going to do do this, the correct constructors would be

    this(string msg, string file = __FILE__, size_t line = __LINE__, Throwable next = null) @safe pure nothrow
    {
        super(msg, file, line, next);
    }

    this(string msg, Throwable next, string file = __FILE__, size_t line = __LINE__) @safe pure nothrow
    {
        super(msg, file, line, next);
    }

src/object.d
```diff
@@ -1751,6 +1751,59 @@ unittest
...
+}
+
+unittest
```
jmdavis: You can avoid duplicating the example in the documentation by putting a ddoc comment on the unittest block. e.g.

    ///
    unittest
    {
    }

src/object.d
```diff
@@ -1751,6 +1751,59 @@ unittest
...
+{
+    import std.exception;
+
+    class MeaCulpa: Exception
+    {
+        mixin StandardExceptionConstructor;
+    }
+    assertThrown!MeaCulpa(enforce!MeaCulpa(false));
```
jmdavis: Why show the exception being thrown via `enforce`? What's it supposed to convey?

src/object.d
```diff
@@ -1751,6 +1751,59 @@ unittest
...
+
+    try throw new MeaCulpa;
```
jmdavis: `MeaCulpa` has no default constructor, so I this shouldn't compile. Additionally, if you're trying to show that the `file` and `line` number get set correctly with the constructors that are mixed in, you should probably show that the exception's `msg` get set correctly as well, which would mean not only constructing the exception with a message but also checking that the caught exception's `msg` matches what you constructed it with.

src/object.d
jamadagni: Well the next parameter is *not* intended to be used from user code as you can see at: http://dlang.org/phobos/object.html#.Exception: "The `next` parameter is used internally and should always be null when passed by user code", and likewise the second overload is only needed for convenience of providing the `next` right after the message without having to manually include `__FILE__` and `__LINE__`, so there's no point in including that overload either.

As for documentation, I thought using the mixin would actually enable it if you include a doc-comment block before it? Besides, if all you are doing is a trivial constructor for purposes of identification, then there should be no *need* for documentation since it's constructed in no way different than a regular exception.

src/object.d
jamadagni: Don't get you – are you saying that there is no need to have the `MeaCulpa` after the `assertThrown` since all that is needed is to show that *an* exception was thrown, or there is no need for this whole line at all?

src/object.d
jamadagni: Oh! But then why was the std.regex example which I have proposed to update via https://github.com/D-Programming-Language/phobos/pull/3731/files duplicated in both the unittest and comments? Should the example be removed from the comments there too?

src/object.d
jmdavis: Okay. I just tested it, and it looks like if you put a ddoc comment on the constructor, then that ddoc comment will actually show up in the exception's documentation. I did not think that that was the case. So, you're going to need to document the constructor(s) in the mixin.

As for `next`, every exception in druntime declares a constructor which takes it as do most of the Phobos exceptions. It's used for exception chaining. Please declare the constructors like so

    /++
        Params:
            msg  = The message for the exception.
            file = The file where the exception occurred.
            line = The line number where the exception occurred.
            next = The previous exception in the chain of exceptions, if any.
      +/
    this(string msg, string file = __FILE__, size_t line = __LINE__, Throwable next = null) @safe pure nothrow
    {
        super(msg, file, line, next);
    }

    /++
        Params:
            msg  = The message for the exception.
            next = The previous exception in the chain of exceptions.
            file = The file where the exception occurred.
            line = The line number where the exception occurred.
      +/
    this(string msg, Throwable next, string file = __FILE__, size_t line = __LINE__) @safe pure nothrow
    {
        super(msg, file, line, next);
    }

That _is_ the standard way to do it, and the only exceptions that don't do that are ones which have alternate constructors, because they're not simply taking a message.

src/object.d
jmdavis: ddoc-ed `unittest` blocks are put into the preceding ddoc comment as an example so that the example is tested while only having to be typed once. Code in Phobos that explicitly has the example in a function's documentation and then has a non-ddoc-ed `unittest` block to test it pre-dates the ddoc `unittest` block feature.

src/object.d
jmdavis: I don't see any point in the `assertThrown` line at all. You're showing your exception type being thrown immediately after this.

jmdavis:
Also, this code has no business in object.d. It's a helper template, not something that needs to be pulled into every piece of D code ever. I'd have commented on that before, but I wasn't paying enough attention to where you put it. It should go in std.exception.

jmdavis:
So, feel free to create a PR with this template mixin in std.exception, but I'm closing this PR, because it doesn't below here.

PR #1414, opened 2015-10-19
eliminate more custom int to string conversions

WalterBright:


jmdavis:
Auto-merge toggled on

WalterBright:
Once again, unittests (for numDigits) found a bug in my "so obvious I don't need to test it" implementation.

PR #1415, opened 2015-10-19
add missing files to MANIFEST

WalterBright:
fix https://issues.dlang.org/show_bug.cgi?id=15223


9rnsr:
Auto-merge toggled on

yebblies:
Can this be automatically checked?

adamdruppe:
Or better yet, get rid of this awful broken and error-prone build setup entirely?

PR #1417, opened 2015-10-26
Render TypeInfo_Struct.xopCmp ABI-agnostic wrt. parameter order

kinke:


ZombineDev:
What's the status of this @kinke?

kinke:
Hmm it's been a long time - haven't some functions been templatized recently? If these 2 are, the problem should be solved.

ibuclaw:
I'm not aware of these being converted to templates.

You may want to check dmd/clone.d also.

ZombineDev:
(Sorry, I just found out I have accidentally closed this PR from my phone)

@kinke The only two template-ized operations are array comparison (https://github.com/dlang/dmd/pull/6597/files) and array ops (https://github.com/dlang/dmd/pull/7032) so it looks like this PR still may be needed. I don't know how soon we'll be able to get rid of of the use of TypeInfo for this kind of stuff, but probably we should continue to template-ize druntime so it can work in a more pay as you go manner, solving the issue @kinke is trying to address here as nice side effect.

ibuclaw:
Tested against gdc, and confirmed that this works for us too.

ibuclaw:
Rebased this in #2025

PR #1419, opened 2015-10-27
core.thread: Only use pthread to store Thread.sm_this when building shared library

klickverbot:
This avoids calling into pthread on switching fiber contexts
with a static runtime, which can be (comparatively) very
expensive (taking a kernel spinlock).

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -41,6 +41,12 @@ version( Solaris )
 // this should be true for most architectures
 version = StackGrowsDown;
 
+// On Posix (excluding OSX), pthread_key_t is explicitly used to
+// store and access thread reference. This is needed
+// to avoid TLS access in signal handlers (malloc deadlock)
+// when using shared libraries, see issue 11981.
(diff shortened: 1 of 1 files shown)


MartinNowak:
It would be better to address the root cause, we shouldn't signal threads that are just about to start.
I'll see if it can be done easily w/o running into issues with the messy thread startup on Windows.

MartinNowak:
Of course, all I'm saying is it looks fairly simple to do better, so I'll spend the hour trying to fix it properly.
In particularly b/c core.thread is in a horrible state w/ lots of ugly documented hacks.
For example I instantly found a bug just by thinking about this problem [Issue 15268](https://issues.dlang.org/show_bug.cgi?id=15268).

MartinNowak:
See #1421 and #1422, 50% reduction of messiness in core.thread ;).
I'm actually surprised it worked out so well (b/c I tried before). But once I decided to touch everything it was a pretty straightforward change.

PR #1420, opened 2015-10-28
Fix imports

9rnsr:
`core.sys.windows.winnt` privately imports `core.sys.windows.basetyps`, so `GUID` is not selective though `basetyps` module.

Required by: https://github.com/D-Programming-Language/dmd/pull/3416


9rnsr:
Thanks!

PR #1422, opened 2015-10-31
fix Issue 15270 - use TLS to store Thread.getThis

MartinNowak:
- as signals are no longer send during thread startup we can use
  a TLS variable to store Thread.getThis()
- make sure to set the variable (and probably trigger lazy TLS
  allocation) before adding the Thread to the global list

[Issue 15270 – use TLS to store Thread.getThis (pthread_getspecific causes heavy lock contention)](https://issues.dlang.org/show_bug.cgi?id=15270)

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -267,8 +267,8 @@ else version( Posix )
             obj.m_tlsgcdata = rt_tlsgc_init();
 
             atomicStore!(MemoryOrder.raw)(obj.m_isRunning, true);
-            Thread.setThis(obj);
-            Thread.add(obj);
+            Thread.setThis(obj); // allocates lazy TLS (see Issue 11981)
+            Thread.add(obj);     // can only receive signals from here on
             scope (exit)
             {
                 Thread.remove(obj);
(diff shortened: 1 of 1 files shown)


MartinNowak:
Depends on https://github.com/D-Programming-Language/druntime/pull/1421, please merge that beforehand.

MartinNowak:
This is a much nicer solution than what we tried before, meaning we can now retackle parallel suspend much easier.
https://github.com/D-Programming-Language/druntime/pull/1116
https://github.com/D-Programming-Language/druntime/pull/1164
https://trello.com/c/lddgsl6f/14-parallel-suspend

MartinNowak:
Yes, I have to debug why the removal from the aboutToStart array fails.

MartinNowak:
> Let me know if you need help with the OS X part (although I would imagine you have access to a Mac anyway).

Yes, VirtualBox.

Rats, looks like a race condition.
https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=1813353&isPull=true
https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=1813843&isPull=true

MartinNowak:
Found it, only took a few hours :(. The stupid pthread_create implementation on OSX doesn't assign `m_addr` before launching the thread, which tripped a rather superfluous assertion.

PR #1428, opened 2015-11-03
core.sys.windows.windef: Remove min/max functions

CyberShadow:
These are not part of the Windows API and will just conflict with the std.algorithm ones.


DmitryOlshansky:
LGTM

PR #1430, opened 2015-11-13
OSX/Darwin - Fix ticksPerSecond

smolt:
denom and numer were backwards in this case.  It works the wrong way on Intel MacOS only because both denom and numer are 1, but not so on other Darwin family OS like iOS, and tvOS.

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -2636,7 +2636,7 @@ extern(C) void _d_initMonoTime()
         mach_timebase_info_data_t info;
         if(mach_timebase_info(&info) == 0)
         {
-            long ticksPerSecond = 1_000_000_000L * info.numer / info.denom;
+            long ticksPerSecond = (1_000_000_000L * info.denom) / info.numer;
             foreach(i, typeStr; __traits(allMembers, ClockType))
             {
                 // ensure we are only writing immutable data once


schveiguy:
Yep, the formula to get the number of nanoseconds for a specific tick value is:

```
x (ns) = ticks * (numer / denom)
```
So, if we know we want `x` to equal 1 second to find `ticks`:
```
1_000_000_000 = ticks * (numer / denom)
```
To solve for `ticks`, we multiply by `denom` and divide by `numer`:
```
1_000_000_000  * (denom / numer) = ticks
```
So this is a valid fix.

schveiguy:
Auto-merge toggled on

schveiguy:
Auto-merge toggled off

schveiguy:
Can we put an `abort` (https://github.com/D-Programming-Language/druntime/blob/master/src/core/internal/abort.d) in here if the ticks per second ends up being non-integral? It should be always valid I would think (several online posts say iPhone is 125/3, which ends up being an integer ticks/second), but there are definitely possible values for numer and denom that would result in non-integer ticks per second. An `abort` check would likely never trigger, but would prevent something from calculating slightly inaccurate times (this would be a really crappy subtle problem).

schveiguy:
To expand, basically I'd like to see this:

```D
long ticksPerSecond = 1_000_000_000L * info.denom;
if(ticksPerSecond % info.numer != 0)
    abort("Non integral ticksPerSecond from mach_timebase_info");
ticksPerSecond /= info.numer
...
```

schveiguy:
Also, can you create an issue for this in the bug tracker? http://issues.dlang.org

smolt:
I like the check.  You are right about iOS have ratio of 125/3, which is how I ever stumbled into this, but what might the future hold?  Other solutions I've seen punt and use floating point.

There was a PR earlier this year because the same calc is used elsewhere: bce9c563b56f5db1b7cc39740ae30cbdd4062603

Probably want one func to calc ticksPerSec that both can use.

I'll do the bug issue when off work.

schveiguy:
> There was a PR earlier this year because the same calc is used elsewhere: bce9c56

Huh, that is the same place, but MonoTime has been rewritten to allow multiple clocks. It's likely that Jonathan's pull to change that predated your fix, and so the fix was lost. It was not very obvious that the numer/denom order was changed.

In fact...

https://github.com/D-Programming-Language/druntime/pull/990/files#diff-b1d9cd2f54f89877fa85992be121dd6cL2304

Yep, that's exactly what happened. Sorry that this got screwed up again! Please make sure to flag the bug as a regression, and I hate to do this, but can you reopen this pull request against the stable branch?

Instructions on how to do so are here: http://wiki.dlang.org/Pull_Requests#Stable_Branch

smolt:
https://issues.dlang.org/show_bug.cgi?id=15334

PR #1432, opened 2015-11-14
Fix issue 15334 - OS X core.time ticksPerSecond calculation is incorrect

smolt:
Fix for https://issues.dlang.org/show_bug.cgi?id=15334

The `_d_initMonoTime()` `version(OSX)` calculation for ticksPerSecond is using a flipped mach_timebase_info ratio numer/denom when it should be denom/numer.

This is a redo of #1430 on stable per @schveiguy with some of his suggestions.

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -2682,18 +2682,14 @@ extern(C) void _d_initMonoTime()
     }
     else version(OSX)
     {
-        mach_timebase_info_data_t info;
-        if(mach_timebase_info(&info) == 0)
+        immutable long ticksPerSecond = machTicksPerSecond();
+        foreach(i, typeStr; __traits(allMembers, ClockType))
         {
-            long ticksPerSecond = 1_000_000_000L * info.numer / info.denom;
-            foreach(i, typeStr; __traits(allMembers, ClockType))
-            {
-                // ensure we are only writing immutable data once
-                if(tps[i] != 0)
-                    // should only be called once
-                    assert(0);
-                tps[i] = ticksPerSecond;
-            }
+            // ensure we are only writing immutable data once
+            if(tps[i] != 0)
+                // should only be called once
+                assert(0);
+            tps[i] = ticksPerSecond;
         }
     }
     else version(Posix)
@@ -2993,17 +2989,7 @@ struct TickDuration
         }
         else version(OSX)
         {
-            static if(is(typeof(mach_absolute_time)))
-            {
-                mach_timebase_info_data_t info;
-
-                if(mach_timebase_info(&info))
-                    ticksPerSec = 0;
-                else
-                    ticksPerSec = (1_000_000_000L * info.denom) / info.numer;
-            }
-            else
-                ticksPerSec = 1_000_000;
+            ticksPerSec = machTicksPerSecond();
         }
         else version(Posix)
         {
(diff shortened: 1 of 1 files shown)


src/core/time.d
```diff
@@ -4732,6 +4718,21 @@ unittest
     static assert(!__traits(compiles, nextLargerTimeUnits!"years"));
 }
 
+version(OSX)
+long machTicksPerSecond()
+{
+    // Be optimistic that ticsPerSecond (1e9*denom/numer) is integral.  So far
```
jmdavis: Not terribly important, but you misspelled `ticksPerSecond` in the comment.

jmdavis:
I swear that this code becomes weirder and uglier ever time I look at it. The number of hoops that we've had to jump through to support multiple clock types is ludicrous. :|

In any case, I don't see anything obviously wrong with this change. I wish that we could have a unit test to verify this bug, but I can't think of one at the moment.

schveiguy:
@jmdavis if you read the previous PR that @smolt made, when we merged the big change to generic `MonoTime`, we accidentally undid a previous fix of his for this that occurred between when you opened the PR and when it was merged. We just didn't notice the ordering of the numer/denom change. See this:

https://github.com/D-Programming-Language/druntime/pull/990/files#diff-b1d9cd2f54f89877fa85992be121dd6cL2304

> I wish that we could have a unit test to verify this bug

The bug is straightforward to see, I don't know if there is a way to test it without mocking `mach_timebase_info`. Putting all the `ticksPerSecond` calculations in one place is probably best (although `TickDuration` is going away eventually).

One thing we could do is put a unit test that sleeps for 1 second, then verifies the ticks change roughly `ticksPerSecond`, or at least that the mach time corresponds to `gettimeofday`, but even with this, you would only see the bug on a system where the ratio is not 1:1 (e.g. iOS).

Dan, can you fix the comment typo and I will merge.

jmdavis:
Oh, the bug is easy enough to see when you look carefully, but it always sucks when a unit test doesn't catch a bug, so ideally, we'd add a test that would. But in this case, I don't think we have a reasonable test that we can run. I certainly don't want to be adding sleeps to tests if we can avoid it.

src/core/time.d
smolt: Got it.  I'll fix and redo the commit.

smolt:
I fix spelling and added a period to end of new abort message to match other aborts in this module.

schveiguy:
> I don't think we have a reasonable test that we can run.

Thinking outside the box, what about this:

```D
version(Linux) version(unittest) version = hasMachTimeStuff;
version(OSX) version = hasMachTimeStuff;
...

version(hasMachTimeStuff)
{
   version(Linux) version(unittest)
   {
       struct mach_timebase_info_data_t {...}
       bool mach_timebase_info(mach_timebase_info_data_t *info)
       {
              /* return 125 / 3 */
              ...
       }
       unittest { assert(machTicksPerSecond() == 24_000_000);}
   }

   long machTicksPerSecond() { ... }
}
```

Ugly hack, yes :)

schveiguy:
Auto-merge toggled on

smolt:
Do I need a separate PR to get this into master or does that happen without me doing anything?

schveiguy:
Martin will merge the branch into master every once in a while.

PR #1434, opened 2015-11-15
Mark GC.addRoot and GC.removeRoot as @nogc

Philpax:


MartinNowak:
Please update [gcstub](https://github.com/D-Programming-Language/druntime/blob/master/src/gcstub/gc.d) as well.

Philpax:
I would have, but the commit that added `@nogc` to `addRange` didn't update those: https://github.com/D-Programming-Language/druntime/commit/db7fe0eb9d3af3154d1873380b00853c7b977c81

Should I update gcstub/proxy for both addRoot and addRange?

DmitryOlshansky:
Seems to fallen through cracks. I guess this needs to update gcstub and proxy for both add/remove ranges

MartinNowak:
Going to merge this anyhow, whoever uses gcstub next can update it.

MartinNowak:
Auto-merge toggled on

leandro-lucarella-sociomantic:
Stupid question, shouldn't it be all the code in the GC @nogc? :)

It happened to me in the past that I used `assert()` in the GC, which triggered an allocation, which called GC code and you can imagine what happened next...

schveiguy:
@leandro-lucarella-sociomantic I think we have to mark public API as not `@nogc`. Otherwise direct calls to GC.malloc would be `@nogc`!

PR #1435, opened 2015-11-19
OS X - Use 64-bit inode stat and dirent functions

smolt:
This updates the stat and dirent family of types and functions (dirent.h and sys/stat.h) to use the newer 64-bit inode versions.  These functions are the default for C compiled code on OS X > 10.5 and provide extra fields and longer filenames.

This also removes the 32-bit inode versions as I believe they are intended for compatibility with binaries build pre 64-bit inode OS X.  I would think all new code should use 64-bit inode versions.

But just in case that isn't preferred, I do have two other versions of this PR that allows:
- selection of 64-bit or 32-bit, configurable at compile time, or
- both 64-bit and 32-bit coexist, where 32-bit is selected by `import core.sys.osx.sys.stat_ino32`

### DIFF
--- a/src/core/sys/posix/dirent.d
+++ b/src/core/sys/posix/dirent.d
@@ -96,14 +96,18 @@ else version( OSX )
         DT_WHT      = 14
     }
 
-    align(4)
+    // _DARWIN_FEATURE_64_BIT_INODE dirent is default for Mac OSX >10.5 and is
+    // only meaningful type for other OS X/Darwin variants (e.g. iOS).
+    // man dir(5) has some info, man stat(2) gives details.
     struct dirent
     {
         ino_t       d_ino;
+        alias       d_fileno = d_ino;
+        ulong       d_seekoff;
         ushort      d_reclen;
+        ushort      d_namlen;
         ubyte       d_type;
-        ubyte       d_namlen;
-        char[256]   d_name;
+        char[1024]  d_name;
     }
 
     struct DIR
@@ -111,7 +115,9 @@ else version( OSX )
         // Managed by OS
     }
 
-    dirent* readdir(DIR*);
+    // OS X maintains backwards compatibility with older binaries using 32-bit
+    // inode functions by appending $INODE64 to newer 64-bit inode functions.
+    pragma(mangle, "readdir$INODE64") dirent* readdir(DIR*);
 }
 else version( FreeBSD )
 {
@@ -205,10 +211,30 @@ else
     static assert(false, "Unsupported platform");
 }
 
-int     closedir(DIR*);
-DIR*    opendir(in char*);
-//dirent* readdir(DIR*);
-void    rewinddir(DIR*);
+version( OSX )
+{
+    version( D_LP64 )
+    {
+        int closedir(DIR*);
+        pragma(mangle, "opendir$INODE64")   DIR* opendir(in char*);
+        pragma(mangle, "rewinddir$INODE64") void rewinddir(DIR*);
+    }
+    else
+    {
+        // 32-bit mangles __DARWIN_UNIX03 specific functions with $UNIX2003 to
+        // maintain backward compatibility with binaries build pre 10.5
+        pragma(mangle, "closedir$UNIX2003")          int closedir(DIR*);
+        pragma(mangle, "opendir$INODE64$UNIX2003")   DIR* opendir(in char*);
+        pragma(mangle, "rewinddir$INODE64$UNIX2003") void rewinddir(DIR*);
+    }
+}
+else
+{
+    int     closedir(DIR*);
+    DIR*    opendir(in char*);
+    //dirent* readdir(DIR*);
+    void    rewinddir(DIR*);
+}
 
 //
 // Thread-Safe Functions (TSF)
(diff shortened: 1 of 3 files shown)


smolt:
Ah, didn't think OS X 32-bit would have a problem because iOS 32-bit uses 64-bit inodes.

smolt:
32-bit functions need addition mangling to use modern OS X __DARWIN_UNIX03 functions.  Should pass 32-bit testers now (fingers crossed).

src/core/sys/posix/dirent.d
```diff
@@ -276,8 +302,20 @@ else version( FreeBSD )
...
+    version ( D_LP64 )
+    {
+        pragma(mangle, "seekdir$INODE64") void seekdir(DIR*, c_long);
+        pragma(mangle, "telldir$INODE64") c_long telldir(DIR*);
+    }
+    else
+    {
+        // 32-bit mangles __DARWIN_UNIX03 specific functions with $UNIX2003 to
+        // maintain backward compatibility with binaries build pre 10.5
+        pragma(mangle, "seekdir$INODE64$UNIX2003") void seekdir(DIR*, c_long);
```
MartinNowak: Why would you use the 32-bit functions just b/c you're executable is 32-bit?

MartinNowak:
> But just in case that isn't preferred, I do have two other versions of this PR that allows:
> 
> selection of 64-bit or 32-bit, configurable at compile time, or

Just select the 64-bit versions as we do on every other platform.

> Ah, didn't think OS X 32-bit would have a problem because iOS 32-bit uses 64-bit inodes.

It shouldn't, please try to fix this. I'd guess that you need to replace c_longs with longs.

src/core/sys/posix/dirent.d
smolt: yes, 32-bit executable using 64-bit inode structs.

smolt:
@MartinNowak - the current commit is correct for both `-m64` and `-m32` and matches output of clang.  The original 32-bit issue was that I missed the `__DARWIN_ALIAS` macro on some of the prototypes in the C header.  For `-m32` code, this appends "$UNIX2003" to function symbols.

BTW I notice while looking through the MacOSX SDK header files that many function prototypes have `__DARWIN_ALIAS` macro. Has anybody thought to review D OS X prototypes so that the modern API is used?  As-is, could lead to subtle difference between D and C code.  @jacob-carlborg are you aware of this?

jacob-carlborg:
@smolt no. But I looked up the `__DARWIN_ALIAS` macro to see what it actually does and found this in the documentation [1].

"Because the \__LP64__ environment only supports UNIX03 semantics it causes __DARWIN_UNIX03 to be defined, but causes __DARWIN_ALIAS to do no symbol mangling"

The test below shows as well that the mangling only occurs when building for 32bit.

```
$ cat main.c
#include <time.h>

int main (int c, char** v)
{
    long a = timezone;
    return 0;
}
$ clang main.c -o main && nm main
0000000100000000 T __mh_execute_header
0000000100000f70 T _main
                 U _timezone
                 U dyld_stub_binder
$ clang main.c -o main -m32 && nm main
00001000 T __mh_execute_header
00001f80 T _main
         U _timezone$UNIX2003
         U dyld_stub_binder
```

[1] http://www.opensource.apple.com/source/xnu/xnu-1456.1.26/bsd/sys/cdefs.h

smolt:
Yup, `man 5 compat` has good info too.

There is a lot of stuff going on in cdefs.h.  It appears to me 64-bit functions are mostly without special mangling, although select and pselect also get a "$1050" suffix due to `__DARWIN_1050` macro.  Here though we might want to make the unlimited select version with "$DARWIN_EXTSN" suffix the default.

```
$ man 2 select
...
COMPATIBILITY
     select() now returns with errno set to EINVAL when nfds is greater
     than FD_SETSIZE.  Use a smaller value for nfds or compile with
     -D_DARWIN_UNLIMITED_SELECT.
```

Seems like a review of OS X declarations would be good.  I'll catch what I can as I poke around.  I just noticed last night that core.stdc.time `CLOCKS_PER_SEC` should be 1_000_000 instead of 100 as of OS X 10.5.

src/core/sys/posix/dirent.d
MartinNowak: But the parameter is using `c_long` (which is 32-bit with -m32).

smolt: Hi Martin!  Yes, that is the way it is prototyped in direct.h, and the return for `telldir` is also a C long, so just a 32-bit directory stream location when compiled with -m32.   Prototype in dirent.h is:

```C
void seekdir(DIR *, long) __DARWIN_ALIAS_I(seekdir);
```

where `__DARWIN_ALIAS_I` from cdef.h mangles specifically for 32-bit arch using 64-bit inodes (`seekdir$INODE64$UNIX2003`)

It can be confirmed without have macosx SDK installed by viewing https://opensource.apple.com/source/Libc/Libc-1044.1.2/include/dirent.h.

smolt:
ping

MartinNowak:
Auto-merge toggled on

PR #1436, opened 2015-11-20
Issue 15367 - array of delegates comparison fails

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=15367


yebblies:
Auto-merge toggled on

PR #1438, opened 2015-11-21
create output dir before building implibs

MartinNowak:


MartinNowak:
Auto-merge toggled on

MartinNowak:
Merging now as it breaks the build script and nightlies.

PR #1440, opened 2015-11-24
update links to spec pages

aG0aep6G:
This is a companion PR to <https://github.com/D-Programming-Language/dlang.org/pull/896>. Do not pull without that one.


CyberShadow:
Auto-merge toggled on

PR #1441, opened 2015-11-26
use DDSUBLINK instead of LINK2

aG0aep6G:
DDSUBLINK uses ROOT_DIR which is more robust than a relative path.

Together with https://github.com/D-Programming-Language/dlang.org/pull/1159 this fixes the link in the online DDox docs.


CyberShadow:
Auto-merge toggled on

PR #1442, opened 2015-11-28
purge already released changelog entries

MartinNowak:


9rnsr:
Auto-merge toggled on

schveiguy:
Shouldn't the master be 2.070 and the stable be 2.069.2?

PR #1443, opened 2015-11-29
Use __c_long_double for CRuntime_Microsoft in general, not just DMD

kinke:
We need this in LDC too; I guess GDC and SDC too (if they support `version(CRuntime_Microsoft)` at all).

### DIFF
--- a/src/core/stdc/config.d
+++ b/src/core/stdc/config.d
@@ -86,27 +86,27 @@ else version( Posix )
   }
 }
 
-version( DigitalMars )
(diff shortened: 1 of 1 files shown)


ibuclaw:
GDC's ``real`` always maps to C's ``long double``, or whatever the target requests it should be.

In any case, this is harmless as I only ever define the gnu runtime version condition.

rainers:
Ok, then let's get this in.

rainers:
Auto-merge toggled on

PR #1446, opened 2015-11-29
MSVC: Provide alternate implementations for 32-bit math functions

kinke:
The x86 runtime lacks most *f versions (asinf(), acosf() etc.), i.e., they are only inline-declared in the C headers (and use the 64-bit double version).
The available symbols apparently depend upon whether one's linking against the static or dynamic MS runtime (libcmt.lib vs. msvcrt.lib), and possibly upon the Visual Studio version too.

So provide alternate implementations here in druntime to prevent linker errors. I came up with these 19 missing functions about a year ago by inspecting the VS 2013 headers (these should be all inline-declared ones, iirc).

Pinging @rainers.

### DIFF
--- a/src/rt/msvc.c
+++ b/src/rt/msvc.c
@@ -6,7 +6,7 @@
 * License: Distributed under the
 *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
 *    (See accompanying file LICENSE)
-* Source:    $(DRUNTIMESRC rt/_stdio_msvc.d)
+* Source:    $(DRUNTIMESRC rt/_msvc.c)
 * Authors:   Rainer Schuetze
 */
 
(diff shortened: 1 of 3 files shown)


src/rt/msvc.c
```diff
@@ -6,10 +6,25 @@
 * License: Distributed under the
 *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
 *    (See accompanying file LICENSE)
-* Source:    $(DRUNTIMESRC rt/_stdio_msvc.d)
+* Source:    $(DRUNTIMESRC rt/_msvc.c)
```
kinke: Not sure if this leading underscore is required.

rainers: No documentation is generated from the rt package, so it's unlikely to make a difference. It's used everywhere so better keep it.

src/rt/msvc.c
```diff
@@ -156,3 +160,55 @@ int  _msvc_fileno(FILE* stream)
...
+
+
+
+/*** math ***/
+
+// The x86 runtime lacks most 32-bit float *f math functions.
+// Provide alternate implementations using the 64-bit double version.
+#if defined _M_IX86
+
+#define IMPL(baseName) \
```
rainers: I'd recommend to use more descriptive names for these macros, e.g. `IMPL_ALT_MATHFNF`.

rainers:
I tried this change with VS2013, and the size of "hello.exe" using printf increased from 239kB to 313kB. 
Maybe we can put the alternate names into msvc.c, but the actual implementations into another module. Or just ensure these are generated into COMDATs?

kinke:
> increased from 239kB to 313kB

Incredible. I completely forgot how primitive linkers still work in 2015 ;) - but `/OPT:REF` should be able to trim it down a bit. Anyway, the math stuff is now in its own module.

rainers:
> Incredible. I completely forgot how primitive linkers still work in 2015 ;) - but /OPT:REF should be able to trim it down a bit. 

My results were with the default, which seems /OPT:REF as long as /DEBUG is not passed, too. with /OPT:NOREF the change is from 346kB to 423kB.

> Anyway, the math implementations are now in their own module.

Better, but still not optimal, as you get the full additional code when referring to one symbol.

I tried compiling msvc_math.c with -Gy to put each function in its own COMDAT, but that did not change much: only the used `_msvc_...` function is found in the map file, but all math functions referred from the module, too. This does not happen if I refer to the double versions.

rainers:
BTW: if the single precision functions are now available for all VS versions, we should add them to core.stdc.math, too.

src/rt/msvc_math.c
```diff
@@ -0,0 +1,52 @@
+/**
+* This module provides alternate implementations of single-precision math
+* functions missing in at least some 32-bit x86 MS VC runtimes
+*
+* Copyright: Copyright Digital Mars 2015.
+* License: Distributed under the
+*      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+*    (See accompanying file LICENSE)
+* Source:    $(DRUNTIMESRC rt/_msvc_math.c)
```
rainers: I think the module names would be more consistent with `sections_*.d` and `deh_*.d` if this module is called `math_msvc.d`.

kinke:
> BTW: if the single precision functions are now available for all VS versions, we should add them to core.stdc.math, too.

They already are: https://github.com/D-Programming-Language/druntime/blob/master/src/core/stdc/math.d#L987

> Better, but still not optimal, as you get the full additional code when referring to one symbol.

Sufficient in my eyes. ;)

> This does not happen if I refer to the double versions.

I guess the double versions are each in their own module.

src/rt/msvc_math.c
kinke: But there are no other math modules, contrary to sections and deh. To me, this extra module is part of `msvc.c` and only a separate file to help the linker somewhat.

rainers:
> They already are: https://github.com/D-Programming-Language/druntime/blob/master/src/core/stdc/math.d#L987

Ah, I forgot about that.

>>    Better, but still not optimal, as you get the full additional code when referring to one symbol.
> 
> Sufficient in my eyes. ;)

Yeah, not so bad in this case. I wonder why putting functions into COMDATs doesn't work. It's the approach used by dmd aswell for all functions.

> 
>     This does not happen if I refer to the double versions.
> 
> I guess the double versions are each in their own module.

Unfortunately there is no source code for these, but the library dump suggests that the guess is correct. For example `ceil` is implemented in an asm file with nothing else in there.

rainers:
Auto-merge toggled on

PR #1153, opened 2015-02-03
spinlock for GC

MartinNowak:
- less overhead than pthread_mutex
- uses test and test-and-set algorithm with configurable backoff

PR #1447, opened 2015-12-01
spinlock for GC

MartinNowak:
- less overhead than pthread_mutex
- uses test and test-and-set algorithm with configurable backoff

Had to reopen [spinlock for GC by MartinNowak · Pull Request #1153](https://github.com/D-Programming-Language/druntime/pull/1153) after force pushing, please look the existing review before repeating any questions.

chart

### DIFF
--- a/src/core/internal/spinlock.d
+++ b/src/core/internal/spinlock.d
@@ -1,0 +1,103 @@
+/**
+ * SpinLock for runtime internal usage.
+ *
+ * Copyright: Copyright Digital Mars 2015 -.
+ * License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Authors:   Martin Nowak
+ * Source: $(DRUNTIMESRC core/internal/_spinlock.d)
+ */
+module core.internal.spinlock;
+
+import core.atomic, core.thread;
+
+shared struct SpinLock
+{
+    /// for how long is the lock usually contended
+    enum Contention : ubyte
+    {
+        brief,
+        medium,
+        lengthy,
+    }
+
+@trusted @nogc nothrow:
+    this(Contention contention)
+    {
+        this.contention = contention;
+    }
+
+    void lock()
+    {
+        if (cas(&val, size_t(0), size_t(1)))
+            return;
+        // Try to reduce the chance of another cas failure
+        // TTAS lock (https://en.wikipedia.org/wiki/Test_and_test-and-set)
+        immutable step = 1 << contention;
+        while (true)
+        {
+            for (size_t n; atomicLoad!(MemoryOrder.raw)(val); n += step)
+                yield(n);
+            if (cas(&val, size_t(0), size_t(1)))
+                return;
+        }
+    }
+
+    void unlock()
+    {
+        atomicStore!(MemoryOrder.rel)(val, size_t(0));
+    }
+
+    /// yield with backoff
+    void yield(size_t k)
+    {
+        if (k < pauseThresh)
+            return pause();
+        else if (k < 32)
+            return Thread.yield();
+        Thread.sleep(1.msecs);
+    }
+
+private:
+    version (D_InlineAsm_X86)
+        enum X86 = true;
+    else version (D_InlineAsm_X86_64)
+        enum X86 = true;
+    else
+        enum X86 = false;
+
+    static if (X86)
+    {
+        enum pauseThresh = 16;
+        void pause()
+        {
+            asm @trusted @nogc nothrow
+            {
+                // pause instruction
+                rep;
+                nop;
+            }
+        }
+    }
+    else
+    {
+        enum pauseThresh = 4;
+        void pause()
+        {
+        }
+    }
+
+    size_t val;
+    Contention contention;
+}
+
+// aligned to cacheline to avoid false sharing
+shared align(64) struct AlignedSpinLock
+{
+    this(SpinLock.Contention contention)
+    {
+        impl = shared(SpinLock)(contention);
+    }
+
+    SpinLock impl;
+    alias impl this;
+}
--- a/src/core/runtime.d
+++ b/src/core/runtime.d
@@ -494,6 +494,11 @@ Throwable.TraceInfo defaultTraceHandler( void* ptr = null )
     else version( Solaris )
         import core.sys.solaris.execinfo;
 
+    // avoid recursive GC calls in finalizer, trace handlers should be made @nogc instead
+    import core.memory : gc_inFinalizer;
+    if (gc_inFinalizer)
+        return null;
+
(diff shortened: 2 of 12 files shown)


MartinNowak:
Closing again until fixed, seems to deadlock a dmd test.

MartinNowak:
Ready now, after fixing a dead-lock caused by onOutOfMemoryError.
This still provides a substantial speed-up, in particular for concurrent code.

MartinNowak:
updated to resolve merge conflict

src/core/internal/spinlock.d
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    void lock()
+    {
+        // TTAS lock (https://en.wikipedia.org/wiki/Test_and_test-and-set)
+        while (true)
+        {
+            for (size_t n; atomicLoad!(MemoryOrder.raw)(val); ++n)
+                yield(n);
+            if (cas(&val, false, true))
```
andralex: Don't you want to lead with optimistically start with the cas?

src/core/internal/spinlock.d
```diff
@@ -0,0 +1,99 @@
...
+    }
+    else
+    {
+        enum pauseThresh = 4;
+        void pause()
+        {
+        }
+    }
+
+    bool val;
```
andralex: Have you tried other data sizes? It's possible that 32-bit may be nicer for the CPU.

src/core/internal/spinlock.d
```diff
@@ -0,0 +1,99 @@
...
+    Contention cont;
```
andralex: Same here, you may want to consider a different data size.

src/core/runtime.d
```diff
@@ -494,6 +494,11 @@ Throwable.TraceInfo defaultTraceHandler( void* ptr = null )
     else version( Solaris )
         import core.sys.solaris.execinfo;
 
+    // HACK: avoid recursive GC calls in finalizer, trace handlers should be made @nogc instead
```
andralex: I think that's fine, i.e. doesn't qualify as a hack.

src/gc/gc.d
```diff
@@ -494,12 +483,12 @@ struct GC
         assert(gcx);
         //debug(PRINTF) printf("gcx.self = %x, pthread_self() = %x\n", gcx.self, pthread_self());
 
-        if (gcx.running)
-            onInvalidMemoryOperationError();
-
         auto p = gcx.alloc(size + SENTINEL_EXTRA, alloc_size, bits);
         if (!p)
+        {
+            gcLock.unlock();
```
andralex: heh, looks like a bugfix

src/gc/gc.d
```diff
@@ -1543,6 +1519,13 @@ struct Gcx
      */
     void runFinalizers(in void[] segment) nothrow
     {
+        GC.inFinalizer = true;
+        scope (failure) // unlock on finalize errors
+        {
+            GC.inFinalizer = false;
+            GC.gcLock.unlock();
```
andralex: This is confusing - it doesn't pair with a visible lock call in the same scope. Couldn't the same entity that locked also take care of the unlock?

src/gc/gc.d
```diff
@@ -1969,7 +1954,11 @@ struct Gcx
             enum initSize = 64 * 1024; // Windows VirtualAlloc granularity
             immutable ncap = _cap ? 2 * _cap : initSize / Range.sizeof;
             auto p = cast(Range*)os_mem_map(ncap * Range.sizeof);
-            if (p is null) onOutOfMemoryErrorNoGC();
+            if (p is null)
+            {
+                GC.gcLock.unlock();
```
andralex: same here - this style of pairing lock/unlock across different scopes is dangerous

src/gc/gc.d
```diff
@@ -2412,7 +2397,17 @@ struct Gcx
             start = stop;
         }
 
-        immutable freedLargePages = sweep();
+        GC.inFinalizer = true;
+        size_t freedLargePages=void;
+        {
+            scope (failure) // unlock on finalize errors
```
andralex: comment says nothing - it repeats what the code does without explaining why

src/gc/proxy.d
```diff
@@ -307,6 +311,13 @@ extern (C)
         return proxy.gc_runFinalizers( segment );
     }
 
+    bool gc_inFinalizer() nothrow
+    {
+        if( proxy is null )
```
andralex: All code you added uses `if (proxy is null)` (i.e. different spacing), please fix this one

src/gcstub/gc.d
```diff
@@ -349,6 +353,13 @@ extern (C) void gc_runFinalizers( in void[] segment )
         proxy.gc_runFinalizers( segment );
 }
 
+extern (C) bool gc_inFinalizer()
+{
+    if( proxy !is null )
```
andralex: spacing

andralex:
I was to pull this, but raises enough flags to warrant another pass.

MartinNowak:
Update to address the review points.
spinlock
Also rebenchmarked, as before this is mostly an improvement for multi-threaded programms (see conmsg).

src/gc/proxy.d
MartinNowak: Same as in the rest of the file.

src/core/internal/spinlock.d
MartinNowak: Done, hardly measurable.

src/core/internal/spinlock.d
MartinNowak: I used size_t and read the contention value only once now, hardly measurable.

src/gc/gc.d
MartinNowak: I changed everything to use `runLocked` and added a `scope (failure)` there.
This will cost a few cycles, b/c dmd can't currently inline a function using exception handling.

src/core/internal/spinlock.d
```diff
@@ -0,0 +1,103 @@
...
+
+@trusted @nogc nothrow:
+    this(Contention contention)
+    {
+        this.contention = contention;
+    }
+
+    void lock()
+    {
+        if (cas(&val, size_t(0), size_t(1)))
```
andralex: I think restructuring this function as a do/while so there's only one cas() might generate tighter code. But not urgent.

src/gc/proxy.d
andralex: Not like in the code you just added e.g. in line 499. But I'll leave this for a later pass.

andralex:
Auto-merge toggled on

src/core/internal/spinlock.d
MartinNowak: I wanted to avoid accessing the `contention` field.

PR #1448, opened 2015-12-06
Support iOS, tvOS, and watchOS x86 in druntime

smolt:
[edited to describe latest commit]
This sets up a version structure for Darwin variants iOS, tvOS, and watchOS.  This is intended to play with D-Programming-Language/dmd#5231.

The approach uses version Darwin in most places for the common code which is set for either OSX, iOS, TVOS, or WatchOS versions.  There are a couple spots where the OSX version is using older functions, thus requiring an alternate version from the new variants.  At some point the OSX versions could be brought up to date.  Also, creates `core.sys.darwin` modules as a copy of `core.sys.osx` and the osx modules have been changed to public import the darwin modules.

There are no iOS, tvOS, or watchOS x86 devices but the iOS Simulator is x86 based, so with these changes, druntime can be compiled and used in the simulator.  Later commits can add the ARM and AArch64 stuff if that is ok in this repo.


src/core/stdc/config.d
```diff
@@ -120,6 +120,8 @@ else version( DigitalMars )
             alias real c_long_double;
         else version( OSX )
             alias real c_long_double;
+        else version( iOS )
```
jacob-carlborg: Why not combining with `OSX`?

smolt: No real reason besides some of the one liners I got lazy on.  I suppose with watchOS and tvOS getting support later, probably should follow the same pattern of defining version Darwin and using that.

joakim-noah:
I'd use an 'else' between the two versions, to signify it has to be one or the other and possibly make compiling quicker.  I think you should merge the two core.sys.* packages into core.sys.darwin- no reason to delay- and fix some of those TODOs, like checking if sem_timedwait exists.  Other than that, LGTM.

Looks like you need to fix the Windows makefiles not to try compiling core.sys.ios, nobody will review a PR that's failing the auto-tester.

As for adding ARM/64, this main druntime repo already has versioned architecture blocks for everything from MIPS to SystemZ, no reason that won't go in here.

smolt:
Thanks for the review

> I'd use an 'else' between the two versions, to signify it has to be one or the other and possibly make compiling quicker.

Sure.

> I think you should merge the two core.sys.* packages into core.sys.darwin- no reason to delay- and fix some of those TODOs, like checking if sem_timedwait exists. Other than that, LGTM.

`sem_timedwait`doesn't show up in the MacOSX 10.11 SDK, and is unresolved when I declare and link, but I thought it might be a historic thing for older OS versions.  Perhaps OSX cleanup might be better on its own PR so it is not buried along with adding iOS?
 
> As for adding ARM/64, this main druntime repo already has versioned architecture blocks for everything from MIPS to SystemZ, no reason that won't go in here.

Is the druntime plan to have all arch stuff here except for inline assembly ala GDC or LDC?  I know that a lot of the LDC ARM stuff is missing here.

joakim-noah:
Yes, OS X cleanup can be separate and that's the plan for druntime.  All ldc's druntime ARM work should go in here, provided it doesn't conflict with gdc's support.

ibuclaw:
The only assembly you should need to do can be external (threadasm.S)

smolt:
Anybody against combining `core.sys.osx` and `core.sys.ios` into `core.sys.darwin`?

jpf91:
> The only assembly you should need to do can be external (threadasm.S)

There's also the ASM in std.math to read/set the FPU flags and IIRC core.cpuid uses ASM as well.

ibuclaw:
> core.cpuid uses ASM as well.

Ah, core.cpuid, that is one module that needs to be re-written in extended assembler.

jacob-carlborg:
> Anybody against combining core.sys.osx and core.sys.ios into core.sys.darwin?

I would be in favor of doing that.

joakim-noah:
I'm pretty sure Walter isn't against it and was talking about something else.

Of course, I'm saying Dan should check all those APIs and make sure it's all Darwin, but keep stuff that's OS-specific in core.sys.osx, not simply rename core.sys.osx to core.sys.darwin without checking.

smolt:
For grins, made that core.sys.darwin package to see how it looks.

smolt:
It might be worth adding the `WatchOS` and `TVOS` versions now.  The Darwin checks at the top of modules would become something like this:
```D
version (OSX)
    version = Darwin;
else version (iOS)
    version = Darwin;
else version (TvOS)
    version = Darwin;
else version (WatchOS)
    version = Darwin;
```

joakim-noah:
Have you actually tested on TvOS and WatchOS?  No point in adding them till they've been tested.

Also, please squash at least last 3 commits in this branch once it's decided to use core.sys.darwin, no need for the extra commits.

smolt:
> Have you actually tested on TvOS and WatchOS? No point in adding them till they've been tested.

Just testing on the TV and Watch x86 simulators which works fine.  I can wait until I've tested on H/W though.

I will eventually have to change `thread_suspend` and `thread_get_state` for these two because these mach calls used by druntime are considered "prohibited" for TVOS and WatchOS.  The mach  calls exist and work but have an attribute attached to the include declaration that makes calling them an error in C.

> Also, please squash at least the last 3 commits in this branch once it's decided to use core.sys.darwin, no need for the extra commits.

Yeah, was planning to.  I still have to incorporate your other suggestions too.  Just waiting until there is some sort of consensus on having a single darwin package.

P.S. Thanx to all who have looked at this.  It's going to be fun having these devices supported by D.

joakim-noah:
Simulators are good enough for me, cool!  Looking forward to the first person to report running D on their watch. :)

smolt:
> Anybody against combining core.sys.osx and core.sys.ios into core.sys.darwin?

I discovered that `core.sys.osx` does have to remain for some period of time because dmd, being D, refers to it.  The osx modules could just public import the darwin ones.

Looking at all of this again, it is kind of messy.  Everything is much simpler when `version (OSX)` is defined by compiler when targeting any of Darwin family (iOS, tvOS, and watchOS) and `core.sys.osx` is also for entire Darwin family.  Very few druntime changes then.

jacob-carlborg:
>  Everything is much simpler when version (OSX) is defined by compiler when targeting any of Darwin family (iOS, tvOS, and watchOS)

How would you identify OS X (that is not iOS, tvOS, and watchOS) then?

joakim-noah:
>I discovered that core.sys.osx does have to remain for some period of time because dmd, being D, refers to it. The osx modules could just public import the darwin ones.

You need to keep `core.sys.osx` around because it's a public API, so it's possible that outside code might be importing it, other than just dmd.  But you can simply have it publicly import the darwin modules and mark it as deprecated.  As for dmd, you could modify your existing dmd PR to have the one module that imports it changed to darwin, and have that PR and this one merged simultaneously to avoid tripping the auto-tester.  @yebblies has done that for me in the past.

>Looking at all of this again, it is kind of messy. Everything is much simpler when version (OSX) is defined by compiler when targeting any of Darwin family (iOS, tvOS, and watchOS) and core.sys.osx is also for entire Darwin family. Very few druntime changes then.

I don't think minimizing druntime changes is what's important, it's minimizing developer confusion by using the appropriate OS names.  If you've been waiting for these PRs before moving ahead with further iOS work, I suggest you don't.  D PRs are notorious for taking a long time to merge; the wait on these PRs is no indication that they're not wanted.

smolt:
> How would you identify OS X (that is not iOS, tvOS, and watchOS) then?

Well, it was simple in my fork when only iOS was there because just had to check iOS first,  but yes, that becomes cumbersome with tvOS and watchOS.

> As for dmd, you could modify your existing dmd PR to have the one module that imports it changed to darwin, and have that PR and this one merged simultaneously to avoid tripping the auto-tester. 

I think as long as an "older" dmd is used to compile dmd src, then core.sys.osx should be module path imported by dmd.  Otherwise older dmds can't be used.

> I don't think minimizing druntime changes is what's important, it's minimizing developer confusion by using the appropriate OS names. If you've been waiting for these PRs before moving ahead with further iOS work, I suggest you don't. D PRs are notorious for taking a long time to merge; the wait on these PRs is no indication that they're not wanted.

I am holding off adding anything iOS to LDC master until this gets in.  I don't want to make a merging upstream too difficult for LDC.

joakim-noah:
>I think as long as an "older" dmd is used to compile dmd src, then core.sys.osx should be module path imported by dmd. Otherwise older dmds can't be used.

Ah, yes, missed that aspect, so you can put off that dmd tweak till the next dmd/druntime with this PR is released, and not mark that single module from the core.sys.osx package as deprecated till then either.

>I am holding off adding anything iOS to LDC master until this gets in. I don't want to make a merging upstream too difficult for LDC.

That's up to you, I was talking about not waiting for this to work on and push stuff to your own github forks.

ZombineDev:
BTW, after this is merged, you may want to update [std.system](http://dlang.org/phobos/std_system.html).

smolt:
> BTW, after this is merged, you may want to update std.system.

Yes, thanks.  I have that waiting in the wings with a few other changes for phobos once the version names, etc are settled on.  On this this topic, does anybody know if the ordinals matter?  I did this (below), but wonder if new versions should always be added at the bottom.
```D
    enum OS
    {
        win32 = 1, /// Microsoft 32 bit Windows systems
        win64,     /// Microsoft 64 bit Windows systems
        linux,     /// All Linux Systems
        osx,       /// Mac OS X
        iOS,       /// iPhone OS (iPad, iPod)
        tvOS,      /// Apple TV OS
        watchOS,   /// Apple Watch OS
        freeBSD,   /// FreeBSD
        solaris,   /// Solaris
        android,   /// Android
        otherPosix /// Other Posix Systems
    }
```

smolt:
@joakim-noah - how does the module deprecation process work?

smolt:
> That's up to you; I was talking about not waiting for this PR to work on and push stuff to your own github forks.

Yeah, I won't wait.  I have some days off soon and was going to spend some of it integrating with watch and Apple TV hardware.  The goal is D for watchOS and tvOS by end of the year!

joakim-noah:
No, ordinals don't matter, I simply stuck android at the end because it seemed to be in chronological order, but win64 broke that.  Btw, I see that linux comment is wrong, it should say "All linux systems except Android."  As for module deprecation, take a look at [my PR deprecating std.c, including the discussion](https://github.com/D-Programming-Language/phobos/pull/2337) and subsequent PR from Vladimir, that started it off with a doc comment.  Which reminds me, that should be turned into the original deprecation warning I had now, will submit another PR for that.  Good to hear you have hardware, send us a photo of the D test output on-screen, alongside a dime for scale. ;)

joakim-noah:
Maybe add a doc comment to the core.sys.osx modules saying they will be deprecated soon.  Or should we just keep them around for legacy reasons, what do others think?

Go ahead and merge that block from core.stdc.math: only separate what's different.  If you're worried about breaking something, don't, the reviewers are here to catch that.  I looked over the PR again, other than those two issues, LGTM.

smolt:
@joakim-noah - I don't have hardware yet, I am waiting until next week.

> Maybe add a doc comment to the core.sys.osx modules saying they will be deprecated soon. Or should we just keep them around for legacy reasons, what do others think?

Ya, I'd like to see what others think.
 
> Go ahead and merge that block from core.stdc.math: only separate what's different. If you're worried about breaking something, don't, the reviewers are here to catch that.

Ok

joakim-noah:
Auto-tester fails now: need to remove `mak/MANIFEST` change and merge `core.stdc.math` blocks and we can try to get this in.

smolt:
It has decayed hasn't it.  I kind of forgot about this and moved on to other stuff.  Ok, I'll bring it up-to-date.

joakim-noah:
Not really, it's just one file that has since been removed and one module whose version blocks you could have edited a month ago.  I don't know why this PR's been sitting for two months, but druntime hasn't changed much in the meantime.  It just moves slower than phobos or dmd.

smolt:
Why would FreeBSD 32-bit fail in std.datetime and then pass on the same commit?

yebblies:
Could be ghosts?

smolt:
:ghost:  in the machine or the code?

schveiguy:
Right in the failure message:
```
--- std.socket(482) test fails depending on environment ---
 (core.exception.AssertError@std/socket.d(484): Assertion failure
```

smolt:
It was real failure in std.datetime.  But I pushed branch twice in short period, maybe it messed up earlier FreeBSD test.

src/core/stdc/math.d
```diff
@@ -609,8 +618,6 @@ else version( OSX )
         FP_NORMAL      = 4,
         ///
         FP_SUBNORMAL   = 5,
-        ///
-        FP_SUPERNORMAL = 6,
```
smolt: This was only used by PPC

smolt:
This is ready for review.

For the curious that have the iPhoneOS SDK laying around, you can actually build and run druntime unit tests against it because the iOS Simulator is x86-based.

Hack in iOS version in mars.d:
```D
VersionCondition.addPredefinedGlobalIdent("iOS");
//VersionCondition.addPredefinedGlobalIdent("OSX");
```

Then run druntime unittests against the iPhoneOS SDK:
```
$ IPHONEOS_DEPLOYMENT_TARGET=9.0 SDKROOT=`xcrun --sdk iphonesimulator -show-sdk-path` make -f posix.mak MODEL=32 clean unittest
```

joakim-noah:
Some minor tweaks listed below, other than that LGTM.

- `core.sys.darwin.sys.event` and its OS X predecessor were left out of `mak/COPY` and the windows makefiles, probably should fix that for Darwin now.
- 'nothrow' is missing for several of the `core.sys.darwin` imports, should add `@nogc` to all of them also.

smolt:
> Some minor tweaks

Updated.

smolt:
I have phobos updates for iOS, TVOS, and WatchOS but want this to go first to make sure the philosophy is correct.

mak/COPY
```diff
@@ -93,6 +108,7 @@ COPY=\
 	$(IMPDIR)\core\sys\osx\mach\thread_act.d \
 	\
 	$(IMPDIR)\core\sys\osx\sys\cdefs.d \
+	$(IMPDIR)\core\sys\osx\sys\event.d \
```
joakim-noah: Since this was never added and it's only a shell now, maybe leave it out?

src/core/sys/darwin/sys/cdefs.d
```diff
@@ -0,0 +1,25 @@
...
+    version = Darwin;
+
+version (Darwin):
+
+public import core.sys.posix.config;
+
+// http://www.opensource.apple.com/source/xnu/xnu-2422.115.4/bsd/sys/cdefs.h
+enum _DARWIN_C_SOURCE = true;
+
+enum __DARWIN_C_FULL = 900000L;
```
joakim-noah: Underscore would be nice here.

mak/COPY
smolt: Would that be confusing though?  To me it seems better to consistently have imports directory for core.sys.osx match core.sys.darwin.  I think it was a mistake before that added event.d without copying it to import dir.

joakim-noah: Confusing to whom?  Because it was never added to `mak/COPY`, it never made it into the `import/` directory in an official release.  The only way someone would use it is if they added `druntime/src` to their path for some reason.

I'm leaning towards just deprecating the `core.sys.osx` modules over time, starting now by sticking doc comments in there indicating they will be deprecated and the `core.sys.darwin` equivalent should be used instead.

smolt: Wait, did you mean just remove the copy commands for core/sys/osx/sys/event.d from make files or also remove the file core/sys/osx/sys/event.d?

joakim-noah: I meant just remove it from `mak/COPY`, as it never shipped with an official dmd release and now there's the darwin equivalent.  As for removing the file, I just checked and the header _was_ shipped with ldc, so probably best to treat it like every other `core.sys.osx` header you moved over to `core.sys.darwin`, ie add a doc comment saying it will be deprecated or whatever we decide to do with those.

smolt:
@joakim-noah I suppose I need to poke at this every so often and see what happens.  I made your suggested revisions and added deprecation comments for `core.sys.osx` modules.  I just picked a date, can someone tell me what it should be?

smolt:
yeah! this moves ldc-developers/ldc#1081 along!

PR #1449, opened 2015-12-08
add core.sys.unwind

WalterBright:
Interface to libunwind. Needed for dwarf exception handling.

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -43,6 +43,8 @@ COPY=\
 	$(IMPDIR)\core\stdc\wchar_.d \
 	$(IMPDIR)\core\stdc\wctype.d \
 	\
+	$(IMPDIR)\core\sys\unwind.d \
+	\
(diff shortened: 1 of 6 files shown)


yebblies:
Auto-merge toggled on

WalterBright:
I was unsure where to put it, as technically it is compiler dependent thing, not an operating system thing. But gcc and clang essentially define the same header. It is only used by one file (the thrower and unwinder). My goal is to get dwarf eh working for Linux 64, and then see where we go from there. It could reasonably be moved then, but for now it just has to be somewhere.

yebblies:
Does it have to be public? Or can it be hidden in `rt.*` like the rest of the runtime calls?

ibuclaw:
Yes, this should have gone in ``rt.*``

MartinNowak:
Could someone please move the file.

PR #1450, opened 2015-12-11
Fixes #15434: object.d must not import from rt

yazd:
https://issues.dlang.org/show_bug.cgi?id=15434

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -1548,9 +1548,11 @@ const:
 
     static int opApply(scope int delegate(ModuleInfo*) dg)
     {
-        import rt.minfo;
+        import core.internal.traits : externDFunc;
+        alias moduleinfos_apply = externDFunc!("rt.minfo.moduleinfos_apply",
+                                              int function(scope int delegate(immutable(ModuleInfo*))));
         // Bugzilla 13084 - enforcing immutable ModuleInfo would break client code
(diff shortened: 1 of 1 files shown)


yazd:
The auto-tester fails for some reason.. uhmm. I cannot reproduce on either a 32-bit or a 64-bit linux.

schveiguy:
Seems it was a spurious failure, passed now.

schveiguy:
I've not seen externDFunc before, interesting! Is that really the only way to declare an external d function? Almost all other runtime extern functions are hooked via C linkage. I'm not sure which way is better.

yazd:
It is used several times in this file to link to `rt` functions.

schveiguy:
OK, then I guess I just haven't read through it in a while :)
LGTM

ZombineDev:
@yazd thanks for fixing this!

Btw I wonder how it worked in debug (no inline) mode.

@schveiguy this is a regression from DMD 2.067 (when the code used to work), so I hope this PR will be merged in time for the next release.

yazd:
@ZombineDev :+1: 
I think it works with no inline because the compiler only uses the declaration and links directly to the compiled druntime (it doesn't recompile druntime). I might be wrong.

Is there anything else to address to get this pulled?
Thanks.

PR #1451, opened 2015-12-26
fix nightly builds

MartinNowak:
- build minit.obj explicitly
- fixes sporadic build failures caused by different timestamps after
  checkout
- requires to explicitly rebuild minit.obj, but we haven't touched the
  asm code since ages

### DIFF
--- a/win32.mak
+++ b/win32.mak
@@ -1175,7 +1175,8 @@ lib\win32\winspool.lib : def\winspool.def
 errno_c_$(MODEL).obj : src\core\stdc\errno.c
 	$(CC) -c -o$@ $(CFLAGS) src\core\stdc\errno.c
 
-src\rt\minit.obj : src\rt\minit.asm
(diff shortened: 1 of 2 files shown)


rainers:
Without dependency, minit.obj will be rebuilt every time.

yebblies:
Can we just make `minit.obj` unbuildable?  Anyone who actually wants to build it can just look up the line in the makefile.

rainers:
Yeah, the target could also be something different than the object file name, e.g. build_minit_obj, to be invoked explicitely.
BTW: win64 and win32mscoff builds don't use minit.obj, so it can be removed from win64.mak.

MartinNowak:
> Without dependency, minit.obj will be rebuilt every time.

Of course Walter's make has to do something unexpected.
I changed it to a dedicated minit target instead.

yebblies:
Auto-merge toggled on

PR #1452, opened 2015-12-28
Add bsf(ulong), bsr(ulong), and popcnt(ulong).

tsbockman:
This PR adds `ulong` versions of `bsf()`, `bsr()`, and `popcnt()`. (It also fixes a small error in the `_popcnt()` docs.)

I use these `core.bitop` functions fairly often, and am always annoyed that the maximum bit-ness of the input depends on what architecture I'm compiling for; it makes using these ops from generic code painful.

There is no good reason for this, as even on 32-bit systems the 64-bit op can be trivially implemented in terms of a couple of calls to the 32-bit op.

**EDIT:** I have created a [companion PR for Phobos](https://github.com/D-Programming-Language/phobos/pull/3888). While working on it, I discovered two different private implementations of `bsr()` for `ulong`, one of which is bugged.

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -60,10 +60,9 @@ int bsf(size_t v) pure;
 /// ditto
 int bsf(ulong v) pure
 {
-    static if(size_t.sizeof == ulong.sizeof)
-        return bsf(cast(size_t)v);
-    else
-    static if(size_t.sizeof == uint.sizeof)
+    static if (size_t.sizeof == ulong.sizeof)
+        return bsf(cast(size_t) v);
+    else static if (size_t.sizeof == uint.sizeof)
     {
         const sv = Split64(v);
         return (sv.lo == 0)?
@@ -95,10 +94,9 @@ int bsr(size_t v) pure;
 /// ditto
 int bsr(ulong v) pure
 {
-    static if(size_t.sizeof == ulong.sizeof)
-        return bsr(cast(size_t)v);
-    else
-    static if(size_t.sizeof == uint.sizeof)
+    static if (size_t.sizeof == ulong.sizeof)
+        return bsr(cast(size_t) v);
+    else static if (size_t.sizeof == uint.sizeof)
     {
         const sv = Split64(v);
         return (sv.hi == 0)?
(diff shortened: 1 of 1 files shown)


tsbockman:
I was able to massively speed up `popcnt()` in most circumstances by:

1. Generalizing the old software algorithm to directly cover 64-bit as well. (Roughly x+3 ops for the 64-bit version; much better than the naive (2*x)+1 ops.)
2. Only using 64-bit ops if `size_t.sizeof == ulong.sizeof`.
3. On DMD, use `_popcnt()` if it is available. There is a small performance penalty associated with checking `core.cpuid.hasPopcnt`, but a **massive** speed-up if it returns `true`, allowing the use of the SSE4 `POPCNT` instruction.

ibuclaw:
Considering this comment: https://github.com/D-Programming-Language/druntime/blob/4ed3b721cc9c6b99f9400817853e430487e3d45b/src/core/checkedint.d#L18-L20

I don't know why @WalterBright introduced ``_popcnt``.  Would have been better off recognising ``popcnt`` as an intrinsic, and ignoring the body.

tsbockman:
> I don't know why @WalterBright introduced `_popcnt`.

I suspect it's related to the fact that you can't count on every CPU having SSE4 support. The other `core.bitop` intrinsics should work on any x86 CPU supported by D, right?

tsbockman:
I have tested the new `popcnt()` algorithm in commit `b460981b06b51d6756e742d405bafa139a5d6890` exhaustively against the old one with the following:
```D
void verify32(string funStr)() {
    mixin("alias fun = " ~ funStr ~ ";");
    uint badCnt = 0;

    uint i = 0;
    while(true) {
        if((i & 0xFFF_FFFF) == 0) {
            write(i >>> 28);
            write(';');
            stdout.flush();
        }

        if(popcnt(i) != fun(i))
            ++badCnt;
        if(i == uint.max)
            break;
        ++i;
    }

    writefln("\r\n" ~ funStr ~ " fails: %s", badCnt);
}
void verify64(string funStr)() {
    mixin("alias fun = " ~ funStr ~ ";");
    ulong badCnt = 0uL;

    for(ulong i = 0; i <= uint.max; ++i)
    {
        if((i & 0xFFF_FFFF) == 0) {
            write(i >>> 28);
            write(';');
            stdout.flush();
        }

        const pi = popcnt(cast(uint)i);
        if(pi != fun(i))
            ++badCnt;

        const ib = i << 12;
        if(pi != fun(ib))
            ++badCnt;

        const ic = ~(i << 17);
        if((64 - pi) != fun(ic))
            ++badCnt;

        const id = Split64(ib ^ ic);
        if((popcnt(id.lo) + popcnt(id.hi)) != fun(id.u64))
            ++badCnt;
    }

    writefln("\r\n" ~ funStr ~ " fails: %s", badCnt);
}
```
Passed so far (all on an Intel Haswell running linux): DMD (64- and 32-bit)

tsbockman:
Micro-benchmark results:
```
// DMD 64-bit:
SSE4 `_popcnt(uint)` :   82 ms
Old   `popcnt(uint)` : 5902 ms
New   `popcnt(uint)` :  205 ms

SSE4 `_popcnt(ulong)`:  147 ms
New   `popcnt(ulong)`:  179 ms

// DMD 32-bit:
SSE4 `_popcnt(uint)` :   83 ms
Old   `popcnt(uint)` : 7677 ms
New   `popcnt(uint)` :  126 ms

New   `popcnt(ulong)`: 2403 ms

// GDC 64-bit:
Old   `popcnt(uint)` : 2369 ms
New   `popcnt(uint)` : 1711 ms
New   `popcnt(ulong)`: 2282 ms

// GDC 32-bit:
Old   `popcnt(uint)` : 4874 ms
New   `popcnt(uint)` : 1632 ms
New   `popcnt(ulong)`: 5056 ms

// LDC 64-bit:
Old   `popcnt(uint)` : 1844 ms
New   `popcnt(uint)` : 1904 ms
New   `popcnt(ulong)`: 2404 ms
```
The error bars on these numbers are huge for various reasons, but hopefully this is sufficient to demonstrate that the new algorithm is an improvement.

ibuclaw:
> // GDC 64-bit:
> Old   `popcnt(uint)` : 2369 ms
> New   `popcnt(uint)` : 1711 ms
> New   `popcnt(ulong)`: 2282 ms
>
> // GDC 32-bit:
> Old   `popcnt(uint)` : 4874 ms
> New   `popcnt(uint)` : 1632 ms
> New   `popcnt(ulong)`: 5056 ms

Challenge accepted. :-)

src/core/bitop.d
```diff
@@ -194,10 +256,149 @@ version (DigitalMars) version (AnyX86) @system // not pure
...
+
+    x += (x>>16);
+    enum mask16 = cast(N)0x0000_FFFF_0000_FFFFL;
+    x &= mask16;
+
+    static if(is(N == ulong)) {
+        x += (x>>32);
+        enum mask32 = cast(N)0x0000_0000_FFFF_FFFFL;
+        x &= mask32;
+    }
```
ibuclaw: Why not use multiply method?
```
x = (x + (x >> 4)) & cast(N)0x0F0F_0F0F_0F0F_0F0FL;

static if (is(N == uint))
    enum shiftbits = 24;
else
    enum shiftbits = 56;

x = (x * cast(N)0x0101_0101_0101_0101L) >> shiftbits;
```

src/core/bitop.d
```diff
@@ -194,10 +256,149 @@ version (DigitalMars) version (AnyX86) @system // not pure
...
+private int soft_popcnt(N)(N x) pure
+    if(is(N == uint) || is(N == ulong))
+{
+    // Avoid branches, and the potential for cache misses which
+    // could be incurred with a table lookup.
+
+    // We need to mask alternate bits to prevent the
+    // sum from overflowing.
+    // add neighbouring bits. Each bit is 0 or 1.
+    enum mask1 = cast(N)0x5555_5555_5555_5555L;
```
ibuclaw: Do we really need a enum name for each step in the process?

tsbockman: > Do we really need a enum name for each step in the process?

No. I just thought it was a little easier to read this way. I can change it if you want.

tsbockman:
> Why not use multiply method?

I benchmarked your method. On my machine (Intel Haswell), it's precisely identical in speed for `ulong`, and 1.3-4.9x faster for `uint`. (I have also verified correctness using the same test as before.)

So, this PR is now strictly faster on my machine than the old implementation, even without using the SSE4 `POPCNT`; the small overhead of checking `core.cpuid.hasPopcnt` is overcome by the improved `soft_popcnt!uint()`.

tsbockman:
The auto-tester failure is unrelated to my PR:
```
Exception - file 'lexer.h': lexer.h: No such file or directory
Exception - file 'parse.h': parse.h: No such file or directory
Error - file 'doc.d' contains trailing whitespace at line 2526
Exception - file 'gluestub.d': gluestub.d: No such file or directory
```

tsbockman:
@ibuclaw Does this need more work, or would you say it's ready if I can get a second reviewer to approve it?

andralex:
This adds yet another punctuation style to the several ones we have... please stick with space after `if`, follow the existing style for `else static if` etc.

tsbockman:
> This adds yet another punctuation style to the several ones we have... please stick with space after `if`, follow the existing style for else `static if` etc.

@andralex If you want to enforce arbitrary requirements for semantically irrelevant stuff like whitespace usage, please add them to the [official style guide](http://dlang.org/dstyle.html) or the [Phobos contribution guidelines](http://wiki.dlang.org/Phobos_Quality).

I have updated the `if` and `static if` stuff like you asked, but I'm not even going to try to guess what all "**etc.**" might encompass since, as you yourself admitted, there is no consistent "existing style" to follow.

JakobOvrum:
@tsbockman, filed style update: D-Programming-Language/dlang.org/pull/1191

tsbockman:
> filed style update: [D-Programming-Language/dlang.org#1191](https://github.com/D-Programming-Language/dlang.org/pull/1191)

My changes should meet the whitespace requirements now. (Unlike the rest of the file...)

andralex:
OK. Shall we return `uint` when there's no possibility of a negative?

tsbockman:
> Shall we return `uint` when there's no possibility of a negative?

That would be a breaking change:
* Someone might be relying on the current return type by capturing it with `auto`.
* Stuff like [Issue 259](https://issues.dlang.org/show_bug.cgi?id=259) makes changing API types between signed and unsigned *very* dangerous.

(I've heard `null` called "*the billion dollar mistake*", but is it really worse than this?
```D
int  a = -1;
uint b =  1;
assert(a > b);
```
At least trying to dereference `null` usually generates a segfault or something. And unlike `null`, the fundamental fix is so simple, regardless of whether you implement it in hardware or software...)

tsbockman:
More generally, because of the way that integer type promotion works currently, *outputs* should always be signed, if their range is guaranteed to fit in an appropriate signed type. Mixed signed/unsigned operations are always safe - well, as safe as same-signedness operations, anyway - provided the signed value is provably positive.

Unfortunately, I don't think there is any safe default choice for *inputs* with the current fail-silently-as-fast-as-possible rules.

andralex:
Auto-merge toggled on

tsbockman:
@andralex Since this is merged now, please consider taking a look at the companion PR for Phobos: https://github.com/D-Programming-Language/phobos/pull/3888

tsbockman:
I broke `popcnt()` in CTFE with this PR. Fix: https://github.com/D-Programming-Language/druntime/pull/1475

denis-sh:
@tsbockman, `size_t` isn't some "magic" type in D, it's just an `alias` for a corresponding integer type so `int bsf(size_t v)` is actually `int bsf(ulong v)` if pointers are 64 bits. As a result `bsf(cast(size_t) v)` is equal to `bsf(v)` if `size_t.sizeof == ulong.sizeof` so your function implementation recursively calls itself.
Opened fixup pull #1486.

##### P.S.
Looks like we don't have enough human resources to carefully review pull requests.

tsbockman:
> your function implementation recursively calls itself.

This claim confused me greatly, until I saw [the explanation you posted elsewhere](https://github.com/D-Programming-Language/druntime/pull/1486#issue-129376824):
> It didn't cause tester failure because dmd uses intrinsic even if there is a function implementation.

PR #1453, opened 2015-12-28
add src/rt/dwarfeh.d

WalterBright:


src/rt/dwarfeh.d
```diff
@@ -0,0 +1,752 @@
...
+     */
+    static void free(ExceptionHeader* eh)
+    {
+        //printf("free(%p)\n", eh);
+        /* Smite contents even if subsequently free'd,
+         * to ward off dangling pointer bugs.
+         */
+        *eh = ExceptionHeader.init;
+        if (eh != &ehstorage)
+            core.stdc.stdlib.free(eh);
```
klickverbot: Won't this break when moving stacks with an in-flight exception to a different thread (where it would then try to `free` the static buffer from a different thread)? (the alpha for the current LDC release has a similar problem too, although it's very unlikely to occur)

MartinNowak:
Thanks, moved the card to done.
https://trello.com/c/ndzxBBqP/141-move-core-sys-unwind-to-rt-unwind

MartinNowak:
> @MartinNowak: It shouldn't have been. The implementation still uses the core.sys.unwind "header".

Wait, this doesn't delete the public interface :o, why was it merged then?

MartinNowak:
I made a checklist, just in case.
https://trello.com/c/ndzxBBqP/141-move-core-sys-unwind-to-rt-unwind

PR #1454, opened 2015-12-30
Dwarf EH: add _d_eh_swapContextDwarf

WalterBright:
Added function needed to swap contexts for fibers. For transitional porpoises, support both old and new eh functionality at the same time, detect which at runtime.

This is blocking https://github.com/D-Programming-Language/dmd/pull/5324

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -118,6 +118,50 @@ private
      *         is switched in for the first time.
      */
     extern(C) void* _d_eh_swapContext(void* newContext) nothrow;
+
+    version (DigitalMars)
+    {
+        version (Windows)
+            alias _d_eh_swapContext swapContext;
+        else
+        {
+            extern(C) void* _d_eh_swapContextDwarf(void* newContext) nothrow;
+
+            void* swapContext(void* newContext) nothrow
+            {
+                /* Detect at runtime which scheme is being used.
+                 * Eventually, determine it statically.
+                 */
+                static int which = 0;
+                final switch (which)
+                {
+                    case 0:
+                    {
+                        assert(newContext == null);
+                        auto p = _d_eh_swapContext(newContext);
+                        auto pdwarf = _d_eh_swapContextDwarf(newContext);
+                        if (p)
+                        {
+                            which = 1;
+                            return p;
+                        }
+                        else if (pdwarf)
+                        {
+                            which = 2;
+                            return pdwarf;
+                        }
+                        return null;
+                    }
+                    case 1:
+                        return _d_eh_swapContext(newContext);
+                    case 2:
+                        return _d_eh_swapContextDwarf(newContext);
+                }
+            }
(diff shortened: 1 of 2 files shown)


WalterBright:
The problem is, it is damn difficult to debug EH when the libraries cannot be built or are non-functional, which is the case when EH does not work. So I made it so that they both exist, and switchable at runtime. This enables the libraries to work with the old EH and the test suite to be run with the new. This is a purely temporary arrangement until Dwarf EH is working on OSX, FreeBSD, and Linux 32.

You can make it work with other compiler runtimes by adding:
```
void* _d_eh_swapContextDwarf(void*) { return null; }
```
All this will be removed once Dwarf EH works.

WalterBright:
> are non-functional
For example, in druntime the demangler relies (unfortunately, but that's the way it was coded) on exceptions working correctly. Unit tests fail to work, assert bombs, etc.

WalterBright:
I know doing Dwarf EH is disruptive. I chose this approach because it minimizes the disruption until it gets reliable and everywhere on Posix.

yebblies:
@WalterBright You can use branches to debug something that isn't in master!

WalterBright:
The `version(DigitalMars)` sounds reasonable. Done.

src/core/thread.d
```diff
@@ -118,6 +118,50 @@ private
...
+                    case 1:
+                        return _d_eh_swapContext(newContext);
+                    case 2:
+                        return _d_eh_swapContextDwarf(newContext);
+                }
+            }
+        }
+    }
+    else
+        alias _d_eh_swapContext swapContext;
```
MartinNowak: This is a public alias on module level, please make it private.
I know you don't like this but our module system leaks private identifiers (and swapContext is a very common name), so please prefix it `_swapContext` or come up with a very specialized name.

MartinNowak:
Rest looks OK, clever state machine.

src/core/thread.d
WalterBright: It's already in a `private` block starting on line 21.

WalterBright:
Pls pull, then, as this is blocking further testing of the Dwarf EH code.

MartinNowak:
> Pls pull, then, as this is blocking further testing of the Dwarf EH code.

Sure, once the namespace pollution is addressed.
https://github.com/D-Programming-Language/druntime/pull/1454#discussion_r48627544

WalterBright:
> Sure, once the namespace pollution is addressed.

Already did:

"It's already in a  `private`  block starting on line 21."

andralex:
noice

MartinNowak:
> "It's already in a private block starting on line 21."

Ah didn't got that comment in time (github's often caching too aggressively).
Let's please still prefix the introduced name https://github.com/D-Programming-Language/druntime/pull/1456.

PR #1455, opened 2015-12-30
Example code to start thread with an anonymous function

tcak:
Adds an example code into unittest to show how to create a new thread and start it with an anonymous function.

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -1808,6 +1808,8 @@ unittest
     // create and start instances of each type
     auto derived = new DerivedThread().start();
     auto composed = new Thread(&threadFunc).start();
+
+    new Thread( (){ writeln("Hello world!"); } ).start();
 }
 
 unittest
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -1808,8 +1808,7 @@ unittest
     // create and start instances of each type
     auto derived = new DerivedThread().start();
     auto composed = new Thread(&threadFunc).start();
-
-    new Thread( (){ writeln("Hello world!"); } ).start();
+    new Thread({ writeln("Hello world!"); }).start();
 }
 
 unittest
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -1808,7 +1808,9 @@ unittest
     // create and start instances of each type
     auto derived = new DerivedThread().start();
     auto composed = new Thread(&threadFunc).start();
-    new Thread({ writeln("Hello world!"); }).start();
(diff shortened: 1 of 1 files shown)


src/core/thread.d
```diff
@@ -1808,6 +1808,8 @@ unittest
     // create and start instances of each type
     auto derived = new DerivedThread().start();
     auto composed = new Thread(&threadFunc).start();
+
+    new Thread( (){ writeln("Hello world!"); } ).start();
```
MartinNowak: You can leave away the `()` parens, and please avoid the extra spaces.
```d
new Thread({ writeln("Hello world!"); }).start();
```

Geod24:
Please squash the commits into 1.

tcak:
I am sorry but I am very new to GitHub, and I neither know what squashing commits to 1 mean, nor how to do it. Do you mean to make the last commit in 1 line only maybe?

etcimon:
The squash command is done by using `git rebase -i HEAD~3` where 3 is the number of commits you made that you wish to merge into a single commit for readability. In the vim window that opens, you replace `pick` with `squash` for the last 2 commits in the list.

Here's a more complete explanation:
http://makandracards.com/makandra/527-squash-several-git-commits-into-a-single-commit

Geod24:
> I am sorry but I am very new to GitHub, and I neither know what squashing commits to 1 mean, nor how to do it. Do you mean to make the last commit in 1 line only maybe?

Currently your patches appear in 3 'commits', since you did it incrementally (first version, then fixes). However, such incremental work should makes the history hard to read, since you see 3 sets of changes (you can see them separately in the 'Commits' tab) for a single change.
Luckily git provides an easy way to turn 3 commits into one easily ('squashing' them).
You can type `git rebase -i HEAD~3` (make sure you are on your branch), and then mark the last 2 lines with `f`

You'll end up with something like:
```
pickup c4f6acf Example code to start thread with an anonymous function
f 0fd2b6 Parentheses and redundant spaces are removed.
f 9868cd writeln is removed
```

For this case, one simpler way would be to just use `git reset --soft HEAD~3` and then `git commit`

If you want to learn more about git and how to use it, I definitely recommend spending some time reading https://git-scm.com/book/en/v2 .

MartinNowak:
Don't sweat the small stuff, it'll be a single merge commit. No need to advice noobs to interactively rebase their first PR.

MartinNowak:
Auto-merge toggled on

PR #1456, opened 2015-12-30
prefix swapContext to avoid namespace pollution

MartinNowak:
- fixup for #1454
- workaround missing DIP22

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -122,12 +122,12 @@ private
     version (DigitalMars)
     {
         version (Windows)
-            alias _d_eh_swapContext swapContext;
+            alias _d_eh_swapContext _swapContext;
         else
         {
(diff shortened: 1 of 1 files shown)


src/core/thread.d
```diff
@@ -122,12 +122,12 @@ private
     version (DigitalMars)
     {
         version (Windows)
-            alias _d_eh_swapContext swapContext;
+            alias _d_eh_swapContext _swapContext;
```
andralex: switch to the new syntax since you touched the line anyway?

andralex:
Not sure I get the motivation - the alternate name is just as polluting. If you want to insulate yourself from any clash with user-defined names, use a double underscore e.g. `__swapContext`.

MartinNowak:
Let's not establish double underscore prefixes. I hope we're getting to [DIP22](http://wiki.dlang.org/DIP22) at some point.

PR #1457, opened 2015-12-31
Dwarf EH: terminate instead of assert

WalterBright:
The trouble with using assert(0) in dwarfeh.d is that it throws an exception inside an exception, which then goes into an infinite loop. This makes things hard to debug. So have it produce a seg fault instead, which is easy to find out whodunit.

This is needed to help me figure out what is going wrong with https://github.com/D-Programming-Language/dmd/pull/5324 as I cannot reproduce it on my machine.

### DIFF
--- a/src/rt/dwarfeh.d
+++ b/src/rt/dwarfeh.d
@@ -71,7 +71,7 @@ struct ExceptionHeader
         {
             eh = cast(ExceptionHeader*)core.stdc.stdlib.calloc(ExceptionHeader.sizeof, 1);
             if (!eh)
-                assert(0);              // out of memory while throwing - not much else can be done
+                terminate(__LINE__);              // out of memory while throwing - not much else can be done
         }
         eh.object = o;
         eh.exception_object.exception_class = dmdExceptionClass;
@@ -144,7 +144,7 @@ extern(C) Throwable __dmd_begin_catch(_Unwind_Exception* exceptionObject)
 
     // Pop off of chain
     if (eh != ExceptionHeader.pop())
-        assert(0);                      // eh should have been at top of stack
+        terminate(__LINE__);                      // eh should have been at top of stack
 
     _Unwind_DeleteException(&eh.exception_object);      // done with eh
     return o;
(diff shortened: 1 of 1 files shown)


tsbockman:
Why is it, in some places, this:
```D
terminate(__LINE__);
assert(0);
```
And others just this:
```D
terminate(__LINE__);
```
?

src/rt/dwarfeh.d
```diff
@@ -192,7 +192,8 @@ extern(C) void _d_throwdwarf(Throwable o)
             case _URC_FATAL_PHASE1_ERROR:       // unknown error code
             case _URC_FATAL_PHASE2_ERROR:       // probably corruption
             default:                            // uh-oh
-                assert(0);                      // C++ calls terminate() instead
+                terminate(__LINE__);            // C++ calls terminate() instead
```
tsbockman: C++ calls `terminate()` instead of `terminate()`?

WalterBright:
Because the compiler knows that assert(0) does not return, while it thinks that terminate() does return.

src/rt/dwarfeh.d
WalterBright: C++ has a core library function called terminate().

src/rt/dwarfeh.d
```diff
@@ -765,3 +774,10 @@ enum LsdaResult
     cleanup,    // cleanup found (i.e. finally or destructor)
     handler,    // handler found (i.e. a catch)
 }
+
+void terminate(uint line) @nogc
+{
+    static char* p;
+    printf("dwarfeh(%u) fatal error\n", line);
+    *p = 0;     // generate seg fault
```
MartinNowak: It might be better to use `_Exit` instead, like [here](https://github.com/D-Programming-Language/druntime/blob/59ff0a44c98ffe90b94aee36df9485032e56a915/src/rt/sections_elf_shared.d#L882), instead of spamming some user's system with core dumps.

WalterBright: Actually, the seg fault doesn't even work here, because etc.linux.memoryerror installs a signal handler for seg faults which, you guessed it, converts it into a thrown exception and then we get the recursive crash again. So I'm switching it to `abort()`.

MartinNowak: OK, abort is even better than _Exit here b/c it still flushes files.

PR #1458, opened 2015-12-31
add indication of which test is being started

WalterBright:
Because when a test crashes, I have no idea which one it was, because the logging only logs upon completion.

### DIFF
--- a/src/test_runner.d
+++ b/src/test_runner.d
@@ -29,6 +29,8 @@ bool tester()
         try
         {
             immutable t0 = MonoTime.currTime;
+            printf("STARTING %.*s\n",
+                cast(uint)name.length, name.ptr);
             fp();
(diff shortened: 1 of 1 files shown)


src/test_runner.d
```diff
@@ -29,6 +29,8 @@ bool tester()
         try
         {
             immutable t0 = MonoTime.currTime;
+            printf("STARTING %.*s\n",
+                cast(uint)name.length, name.ptr);
```
andralex: y u two lines? one is enough...

andralex:
Auto-merge toggled on

src/test_runner.d
WalterBright: did it to match the formatting of the following lines.

WalterBright:
thanks, this should help

PR #999, opened 2014-10-24
Add ability to clear all elements from AA

schveiguy:
I couldn't find a bug report on this, but it has been a thorn in many people's sides for a long time. And it's really easy to fix. Given that I just looked at the AA code for something else, I figured I'd just do it.

Most recent discussion: http://forum.dlang.org/thread/xovoxwvzdnuvcjljwhaq@forum.dlang.org

Note, I would like to have called this `clear`, but we currently have `clear` as an alias to `destroy` scheduled for deprecation this month. We can probably change this to `clear` in the future, but we need to make `clear` a compiler error for a while after deprecation to avoid silently changing code that uses `clear` to mean `destroy`.

PR #1459, opened 2016-01-03
Add ability to clear AA of all elements

schveiguy:
This is a reworking of #999 using new AA implementation.

Note, clear has been gone as a deprecated alias since [March 2015](https://github.com/D-Programming-Language/druntime/commit/fa624c472f2357e375d3b52eaa473dad709c725a).

I also changed firstUsed cache code to be more accurate.

@MartinNowak, I'm curious why you keep a running total of deleted elements (and flag them differently)? What purpose does this serve? It seems to make the code more complex than it needs to be.

### DIFF
--- a/changelog.dd
+++ b/changelog.dd
@@ -7,7 +7,8 @@ $(BUGSTITLE Library Changes,
 
 $(LI $(RELATIVE_LINK2 TypeInfo.initializer, `TypeInfo.init` has been renamed to
     `TypeInfo.initializer`.))
-)
+$(LI $(RELATIVE_LINK2 AA.clear, Associative array `clear` method to remove all
+    elements has been added.))
 )
 
 $(BUGSTITLE Library Changes,
@@ -26,6 +27,15 @@ $(LI $(LNAME2 TypeInfo.initializer, `TypeInfo.init` has been renamed to
         2.072.0 release, and is going to be `@disable`d with the 2.073.0
         release. Finally, the special casing is going to be removed with the
         2.074.0 release, so that the type property `init` takes over.
+    )
+)
+$(LI $(LNAME2 AA.clear, Associative array `clear` method to remove all
+    elements has been added.)
+
+    $(P One can now use `aa.clear()` to remove all elements from an
+        associative array. This allows one to remove all elements from all
+        references to the same array (setting to `null` just cleared the
+        reference, not removed the elements).
     )
 )
 )
--- a/changelog.dd
+++ b/changelog.dd
@@ -7,8 +7,8 @@ $(BUGSTITLE Library Changes,
 
 $(LI $(RELATIVE_LINK2 TypeInfo.initializer, `TypeInfo.init` has been renamed to
     `TypeInfo.initializer`.))
-$(LI $(RELATIVE_LINK2 AA.clear, Associative array `clear` method to remove all
-    elements has been added.))
+$(LI $(RELATIVE_LINK2 aa-clear, A `clear` method has been added to associative
+    arrays to remove all elements.))
 )
 
 $(BUGSTITLE Library Changes,
@@ -29,14 +29,21 @@ $(LI $(LNAME2 TypeInfo.initializer, `TypeInfo.init` has been renamed to
         2.074.0 release, so that the type property `init` takes over.
     )
 )
-$(LI $(LNAME2 AA.clear, Associative array `clear` method to remove all
-    elements has been added.)
+$(LI $(LNAME2 aa-clear, A `clear` method has been added to associative
+    arrays to remove all elements.)
 
     $(P One can now use `aa.clear()` to remove all elements from an
-        associative array. This allows one to remove all elements from all
-        references to the same array (setting to `null` just cleared the
-        reference, not removed the elements).
+        associative array. This allows removing all elements from all
+        references to the same array (setting to `null` just reset the
+        reference, but did not removed the elements from other references).
     )
+
+-------
+auto aa = ["first" : "entry", "second": "entry"];
+auto aa2 = aa1; // reference to the same AA
+aa.clear(); // still retains its current capacity for faster imports
+assert(aa2.length == 0); // other references affected
+-------
 )
 )
 
--- a/src/object.d
+++ b/src/object.d
@@ -1857,6 +1857,7 @@ extern (C)
     inout(void)[] _aaValues(inout void* p, in size_t keysize, in size_t valuesize, const TypeInfo tiValArray) pure nothrow;
     inout(void)[] _aaKeys(inout void* p, in size_t keysize, const TypeInfo tiKeyArray) pure nothrow;
     void* _aaRehash(void** pp, in TypeInfo keyti) pure nothrow;
+    void _aaClear(void* p) pure nothrow;
 
     // alias _dg_t = extern(D) int delegate(void*);
     // int _aaApply(void* aa, size_t keysize, _dg_t dg);
@@ -1889,6 +1890,16 @@ void* aaLiteral(Key, Value)(Key[] keys, Value[] values) @trusted pure
 }
 
 alias AssociativeArray(Key, Value) = Value[Key];
+
+void clear(T : Value[Key], Value, Key)(T aa)
+{
+    _aaClear(*cast(void **)&aa);
+}
+
+void clear(T : Value[Key], Value, Key)(T* aa)
+{
+    _aaClear(*cast(void **)aa);
+}
 
 T rehash(T : Value[Key], Value, Key)(T aa)
 {
--- a/src/object.d
+++ b/src/object.d
@@ -1893,12 +1893,12 @@ alias AssociativeArray(Key, Value) = Value[Key];
 
 void clear(T : Value[Key], Value, Key)(T aa)
 {
-    _aaClear(*cast(void **)&aa);
+    _aaClear(*cast(void **) &aa);
 }
 
 void clear(T : Value[Key], Value, Key)(T* aa)
 {
-    _aaClear(*cast(void **)aa);
+    _aaClear(*cast(void **) aa);
 }
 
 T rehash(T : Value[Key], Value, Key)(T aa)
--- a/src/rt/aaA.d
+++ b/src/rt/aaA.d
@@ -142,17 +142,12 @@ private:
     {
         auto obuckets = buckets;
         buckets = allocBuckets(ndim);
-        auto newFirstUsed = cast(uint) ndim;
 
         foreach (ref b; obuckets[firstUsed .. $])
             if (b.filled)
-            {
-                auto x = findSlotInsert(b.hash);
-                newFirstUsed = min(newFirstUsed, cast(uint) (x - buckets.ptr));
-                *x = b;
-            }
+                *findSlotInsert(b.hash) = b;
 
-        firstUsed = newFirstUsed;
+        firstUsed = 0;
         used -= deleted;
         deleted = 0;
         GC.free(obuckets.ptr); // safe to free b/c impossible to reference
--- a/src/rt/aaA.d
+++ b/src/rt/aaA.d
@@ -142,13 +142,13 @@ private:
     {
         auto obuckets = buckets;
         buckets = allocBuckets(ndim);
-        uint newFirstUsed = cast(uint)ndim;
+        auto newFirstUsed = cast(uint) ndim;
 
-        foreach (ref b; obuckets[firstUsed..$])
+        foreach (ref b; obuckets[firstUsed .. $])
             if (b.filled)
             {
                 auto x = findSlotInsert(b.hash);
-                newFirstUsed = min(newFirstUsed, cast(uint)(x - buckets.ptr));
+                newFirstUsed = min(newFirstUsed, cast(uint) (x - buckets.ptr));
                 *x = b;
             }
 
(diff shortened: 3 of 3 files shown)


schveiguy:
Note, I will update docs as a separate pull request since those are done in dlang.org for AA.

schveiguy:
What? I tried to rebase because there was a merge conflict, and it destroyed all my changes.

Luckily I have time machine backup. Opening a new PR. This is very weird.

schveiguy:
Somehow able to resurrect this PR. And the merge conflicts are gone. Very strange.

schuetzm:
In case you didn't know it yet, git has a kind of built-in backup that can be used when rebasing goes wrong:
http://gitready.com/intermediate/2009/02/09/reflog-your-safety-net.html

schveiguy:
I read about that after I'd already used time machine. But I'm thinking if I didn't do it my way, the merge conflict would still be there :) Not sure why it had one.

src/rt/aaA.d
```diff
@@ -142,16 +142,36 @@ private:
...
-        firstUsed = 0;
+        firstUsed = newFirstUsed;
         used -= deleted;
         deleted = 0;
         GC.free(obuckets.ptr); // safe to free b/c impossible to reference
     }
+
+    void clear() pure nothrow
+    {
+        foreach(ref b; buckets[firstUsed..$])
```
andralex: space after `foreach`, `if` etc :)

src/object.d
```diff
@@ -1890,6 +1891,16 @@ void* aaLiteral(Key, Value)(Key[] keys, Value[] values) @trusted pure
 
 alias AssociativeArray(Key, Value) = Value[Key];
 
+void clear(T : Value[Key], Value, Key)(T aa)
+{
+    _aaClear(*cast(void **)&aa);
+}
+
+void clear(T : Value[Key], Value, Key)(T* aa)
```
andralex: cant we do without this?

src/rt/aaA.d
andralex: spaces around `..`

src/rt/aaA.d
```diff
@@ -142,16 +142,36 @@ private:
...
+        {
+            if(b.filled)
+            {
+                b.hash = HASH_DELETED;
+                b.entry = null;
+            }
+        }
+        deleted = used;
+        firstUsed = cast(uint)dim;
```
andralex: space after `)`

andralex:
Thx! Just a few nits and I'm approving this.

src/rt/aaA.d
yebblies: Since when?  The phobos codebase is roughtly 50/50 on this, and the dmd codebase consistently uses no space after cast.

andralex: In druntime: 

```bash
$ grep 'cast *([^)]*)' src/**/*.d | wc -l
  3371
$ grep 'cast *([^)]*) ' src/**/*.d | wc -l
  1233
```

Guess you're right.

yebblies: Can we make it official policy? Dfmt can be configured for either.

src/object.d
schveiguy: I don't know, I was following precedent, the other methods (e.g. rehash, dup) also do this.

schveiguy:
updated formatting. On the question of adding trailing space to cast(x), I did so, though I think it looks odd.

src/object.d
MartinNowak: The other methods are imitating a builtin property and UFCS doesn't dereference pointers. I'd say keep this for clear as well.

src/rt/aaA.d
MartinNowak: Let's just put the dfmt config into .editorconfig.

src/rt/aaA.d
```diff
@@ -142,16 +142,36 @@ private:
...
         buckets = allocBuckets(ndim);
+        auto newFirstUsed = cast(uint) ndim;
 
-        foreach (ref b; obuckets)
+        foreach (ref b; obuckets[firstUsed .. $])
             if (b.filled)
-                *findSlotInsert(b.hash) = b;
+            {
+                auto x = findSlotInsert(b.hash);
+                newFirstUsed = min(newFirstUsed, cast(uint) (x - buckets.ptr));
```
MartinNowak: This change needs to pass our AA benchmark suite, AA resizing is very performance critical and you're adding a bunch of instructions in the hot loop.
My gut feeling would be that it's cheaper to compute the firstUsed cache after repopulating the new table, and even cheaper to do it lazily next time it's used.

src/rt/aaA.d
yebblies: That works too.

src/rt/aaA.d
schveiguy: I'll put it back to setting it to 0, because that change is orthogonal to this PR. However, no reason to ignore the cache when referencing the original array. Will update soon.

schveiguy:
I removed the update of cache.

I'm very suspect that the cache does anything at all. In the original AA code, it was updated whenever possible. In this code, it seems that it will never get set properly (once you expand, it's set to 0, and never gets updated). Even code that may benefit from it (e.g. apply) will not update the cache for the next time through. I suggest that this is a regression, but the original complaint was performance based, so the comparison is difficult.

I'm also am dubious that when you are copying elements from one bucket array to the other, it will be a performance problem to update a stack variable with a minimize call (it could be written as `a < b ? a : b` if the `min` is the problem). Potentially, the pointer arithmetic can be saved for the end.

schveiguy:
> I'm curious why you keep a running total of deleted elements (and flag them differently)? What purpose does this serve? It seems to make the code more complex than it needs to be.

@MartinNowak any answer to above question?

MartinNowak:
> I'm also am dubious that when you are copying elements from one bucket array to the other, it will be a performance problem to update a stack variable with a minimize call (it could be written as a < b ? a : b if the min is the problem). Potentially, the pointer arithmetic can be saved for the end.

Look at [findSlotInsert](https://github.com/D-Programming-Language/druntime/pull/1459/files#diff-fdc0da51523ff831dd6cbe33a5bb8b4cR102). It's very efficient but requires quite a few registers.
Now if you steal a few of those registers (from dmd's not too good optimizer) to execute a hard to predict conditional it's going to have an effect.

> @MartinNowak, I'm curious why you keep a running total of deleted elements (and flag them differently)? What purpose does this serve? It seems to make the code more complex than it needs to be.

Those are tombstones, they are skipped during lookup but refilled during insertion.

> In this code, it seems that it will never get set properly (once you expand, it's set to 0, and never gets updated). Even code that may benefit from it (e.g. apply) will not update the cache for the next time through.

Seems you're right, but the range and apply functions are const, so they can't modify the cache (remember we want to emit AAs to the code segment at some point).
The value is set for literals but also precomputing it after every resize cache would penalize everyone just so that some people might probably run O(N^2) `while (aa.length) aa.remove(aa.byKey.front)`. Ultimately a hashtable isn't an array and cannot be used like one.
Also this AA does shrink, so you'd recompute it more often.

schveiguy:
> Those are tombstones, they are skipped during lookup but refilled during insertion.

why not just set them to unfilled? Then you could avoid keeping track of the number of deletions altogether.

src/rt/aaA.d
```diff
@@ -152,6 +152,21 @@ private:
...
         GC.free(obuckets.ptr); // safe to free b/c impossible to reference
     }
+
+    void clear() pure nothrow
+    {
+        foreach (ref b; buckets[firstUsed .. $])
+        {
+            if (b.filled)
+            {
+                b.hash = HASH_DELETED;
```
MartinNowak: If you're deleting every entry you should use HASH_EMPTY here. Or even better just memset(0) the whole bucket array.

schveiguy:
> requires quite a few registers.

In fact, it uses virtual call `keyti.equals` which I would expect to clean out all the registers and be pretty hard to predict, no? Maybe I'm not understanding how it works.

> Also this AA does shrink, so you'd recompute it more often.

The old one did too.

src/rt/aaA.d
```diff
@@ -152,6 +152,21 @@ private:
...
+                b.entry = null;
+            }
+        }
+        deleted = used;
+        firstUsed = cast(uint) dim;
+        resize(INIT_NUM_BUCKETS);
```
MartinNowak: Shrinking the AA to it's initial size defeats one of the main use cases for clear, it should work like appender.clear and keep the buffer so it can be fully refilled w/o resizing.
The AA will adapt it's size to the workload (when there are more deletions) and there is still aa.rehash to shrink w/o deleting.

src/rt/aaA.d
schveiguy: > just memset(0) the whole bucket array

OK, I wasn't sure if I was missing something on the HASH_DELETED record keeping, so I played it safe. I can use memset instead.

src/rt/aaA.d
schveiguy: Well, that's definitely not a good reason to use clear. How would you even do that? You can't reserve any bucket data, and it would be silly to add a whole bunch of elements and then clear so you could reserve the buckets. A whim of the implementation might just wreck that "reserve" that you did later on.

I'm not opposed to leaving the array intact, as it doesn't necessarily hurt to keep the memory. But not for your reasoning.

Also keep in mind that if one had a very very large AA, and you called clear on it, you may expect it to relinquish as much memory as possible.

schveiguy: Note, it would be nice to be able to specify such behaviors for the implementation.

MartinNowak: > Well, that's definitely not a good reason to use clear. How would you even do that? You can't reserve any bucket data, and it would be silly to add a whole bunch of elements and then clear so you could reserve the buckets.

The use-case goes like this, parse a document and insert 1M entries, process the entries, clear the AA, continue w/ the next 1M inserts. Not having to resize the AA (and there are many resizes from 8 to 1M) would be a huge plus here.
Maybe you're right and it should be part of a separate reserve and capacity API, but killing the capacity on clear when we have a dedicated rehash for that would be remain weird.

schveiguy:
@MartinNowak updated per your comments

MartinNowak:
> > Also this AA does shrink, so you'd recompute it more often.

> The old one did too.

No, it did not shrink on deletion.
https://github.com/schveiguy/druntime/blob/7dd76e02a44b0892daedece3c4e2057e07cf9667/src/rt/aaA.d#L397

MartinNowak:
> > requires quite a few registers.

> In fact, it uses virtual call keyti.equals which I would expect to clean out all the registers and be pretty hard to predict, no? Maybe I'm not understanding how it works.

Yes virtual function calls are a huge problem for the AA implementation, but please look at the code before making up arguments. Resizing the AA (findSlotInsert in particular) does not perform any key comparison or rehashing, it just used a different modulo/mask on the precomputed hash.

changelog.dd
```diff
@@ -7,7 +7,8 @@ $(BUGSTITLE Library Changes,
 
 $(LI $(RELATIVE_LINK2 TypeInfo.initializer, `TypeInfo.init` has been renamed to
     `TypeInfo.initializer`.))
-)
+$(LI $(RELATIVE_LINK2 AA.clear, Associative array `clear` method to remove all
+    elements has been added.))
 )
```
MartinNowak: Please use `RELATIVE_LINK aa-clear`, anchors shouldn't be camel cased.
How about "A `clear` method to remove all elements from an associative array has been added"?

changelog.dd
```diff
@@ -28,6 +29,15 @@ $(LI $(LNAME2 TypeInfo.initializer, `TypeInfo.init` has been renamed to
...
     )
 )
+$(LI $(LNAME2 AA.clear, Associative array `clear` method to remove all
+    elements has been added.)
+
+    $(P One can now use `aa.clear()` to remove all elements from an
+        associative array. This allows one to remove all elements from all
+        references to the same array (setting to `null` just cleared the
+        reference, not removed the elements).
+    )
```
MartinNowak: I think the second part is a bit hard to understand and would be better illustrates by a concise code example.
```d
auto aa = ["first": "entry", " second": "entry"];
aa.clear(); // still retains it's current capacity for faster inserts
assert(aa.length == 0);
aa.rehash(); // explicitly reset the capacity
```
```d
auto aa1 = ["first": "entry", " second": "entry"];
auto aa2 = aa1; // reference to the same AA
aa1.clear(); // clear the referenced AA
assert(aa2.length == 0);
```

MartinNowak:
Code looks good now, thanks.

schveiguy:
> please look at the code before making up arguments.

I did. Your link went to "findSlotLookup", or at least that's what github showed me. I assumed that's what you meant. Sorry for the confusion.

I had to click on the "open more code" link to see "findSlotInsert".

schveiguy:
> No, it did not shrink on deletion.

I would have sworn it did. I looked back to 2010 and it still did not. Sorry for the misinformation, I apparently did not understand the code as well as I thought.

changelog.dd
schveiguy: Gah, I hate ddoc. I can never understand what all the macros mean.

schveiguy: Wait, did you intend to remove the "2" from RELATIVE_LINK2 ? I compiled it with just RELATIVE_LINK and ddoc doesn't seem to know what that means.

Of course, it's impossible to actually build these freaking docs without lots of manual manipulation, so I have no idea if that's some externally defined macro...

schveiguy:
Fixed nits for docs.

MartinNowak:
Auto-merge toggled on

MartinNowak:
Thanks @schveiguy, this is a very useful addition.

jmdavis:
> Thanks @schveiguy, this is a very useful addition.

Indeed. And it seems to have taken a lot more work than I would have thought, so it deserves that much more of a thank you.

schveiguy:
Thanks, I will work on the doc update.

PR #1460, opened 2016-01-10
Dwarf EH: add class Object.__cpp_type_info

WalterBright:
See discussion https://github.com/D-Programming-Language/dmd/pull/5342 and http://www.digitalmars.com/d/archives/digitalmars/D/Catching_C_Exceptions_in_D_277799.html#N277849

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -1846,6 +1846,13 @@ unittest
     }
 }
 
+/* Used in Exception Handling LSDA tables to 'wrap' C++ type info
+ * so it can be distinguished from D TypeInfo
+ */
+class __cpp_type_info_ptr
+{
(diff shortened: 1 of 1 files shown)


ibuclaw:
I think my personal preference would be to have someone get this working on GDC or LDC too, then we can compare notes on how we did it / the best approach.

andralex:
Auto-merge toggled on

andralex:
@ibuclaw guess this PR shouldn't be too contentious.

ibuclaw:
> @ibuclaw guess this PR shouldn't be too contentious 

That depends on whether it's strictly necessary.  I still think that I can make it work without

PR #1461, opened 2016-01-10
Add support for native TLS on OS X

jacob-carlborg:
Runtime part of https://github.com/D-Programming-Language/dmd/pull/5346.

### DIFF
--- a/src/rt/osx_tls.c
+++ b/src/rt/osx_tls.c
@@ -1,0 +1,51 @@
+/**
+ * Helpers for determining TLS memory ranges on OS X.
+ *
+ * This unfortunately cannot be entirely done in D, as the OS X API uses
+ * the Apple-specific blocks C extension.
+ *
+ * Copyright: David Nadlinger, 2012.
+ * License:   <a href="http://www.boost.org/LICENSE_1_0.txt">Boost License 1.0</a>.
+ * Authors:   David Nadlinger
+ */
+
+#ifndef __APPLE__
+#ifndef __BLOCKS__
+ #error "Need a C compiler with Apple Blocks support – not building on OS X with Clang?"
+#endif
+#endif
+
+#include <assert.h>
+#include <stddef.h>
+#include <stdio.h>
+
+/*
+ * Declarations from dyld_priv.h, available on 10.7+.
+ */
+enum dyld_tlv_states {
+    dyld_tlv_state_allocated = 10,
+    dyld_tlv_state_deallocated = 20
+};
+typedef struct {
+    size_t info_size;
+    void * tlv_addr;
+    size_t tlv_size;
+} dyld_tlv_info;
+typedef void (^dyld_tlv_state_change_handler)(enum dyld_tlv_states state, const dyld_tlv_info *info);
+extern void dyld_register_tlv_state_change_handler(enum dyld_tlv_states state, dyld_tlv_state_change_handler handler);
+extern void dyld_enumerate_tlv_storage(dyld_tlv_state_change_handler handler);
+
+void _d_dyld_getTLSRange(void* arbitraryTLSSymbol, void** start, size_t* size) {
+    dyld_enumerate_tlv_storage(
+        ^(enum dyld_tlv_states state, const dyld_tlv_info *info) {
+            assert(state == dyld_tlv_state_allocated);
+            if (info->tlv_addr <= arbitraryTLSSymbol &&
+                arbitraryTLSSymbol < (info->tlv_addr + info->tlv_size)
+            ) {
+                // Found the range.
+                *start = info->tlv_addr;
+                *size = info->tlv_size;
+            }
+        }
+    );
+}
--- a/src/rt/sections_osx_x86_64.d
+++ b/src/rt/sections_osx_x86_64.d
@@ -1,18 +1,18 @@
 /**
  * Written in the D programming language.
- * This module provides OSX-specific support for sections.
+ * This module provides OS X x86-64 specific support for sections.
  *
- * Copyright: Copyright Digital Mars 2008 - 2012.
+ * Copyright: Copyright Digital Mars 2016.
  * License: Distributed under the
  *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
  *    (See accompanying file LICENSE)
- * Authors:   Walter Bright, Sean Kelly, Martin Nowak
- * Source: $(DRUNTIMESRC src/rt/_sections_osx.d)
+ * Authors: Jacob Carlborg
+ * Source: $(DRUNTIMESRC src/rt/_sections_osx_x86_64.d)
  */
-
-module rt.sections_osx;
+module rt.sections_osx_x86_64;
 
 version(OSX):
+version(X86_64):
 
 // debug = PRINTF;
 import core.stdc.stdio;
@@ -72,7 +72,6 @@ __gshared bool _isRuntimeInitialized;
  */
 void initSections()
 {
-    pthread_key_create(&_tlsKey, null);
     _dyld_register_func_for_add_image(&sections_osx_onAddImage);
     _isRuntimeInitialized = true;
 }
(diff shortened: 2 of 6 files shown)


jacob-carlborg:
Added back emulate TLS for 32bit.

src/rt/sections_osx.d
```diff
@@ -23,6 +23,13 @@ import core.sys.osx.mach.getsect;
 import rt.deh, rt.minfo;
 import rt.util.container.array;
 
+version (D_LP64)
+    import rt.sections_osx64;
+else
+    import rt.sections_osx32;
+
+mixin SectionsOSX;
```
bitwise-github: sorry if this is a nitpick..but couldn't you just use public import instead of a mixin?

jacob-carlborg: The reason I did it like this is because the code in `sections_osxXX` depends on code in `sections_osx`. I tried to avoid a circular dependency without changing too much of the code. Not sure if this is better though.

andralex:
So, what's the deal with https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=1918184&isPull=true

src/rt/sections_osx.d
bitwise-github: I probably would have opted for creating a third/common file. The mixin approach has two problems: The mixin could change meaning depending on where it's used, and while editing the contents of the mixin, intellisense systems will not know where to look for the symbols. Newcomers looking at sections_osxXX may also be confused as well. I'm not sure how prevalent this type of thing is in the druntime code, but if it can be avoided, I think it should.

jacob-carlborg:
@andralex this PR for druntime is required https://github.com/D-Programming-Language/druntime/pull/1461.

src/rt/sections_osx.d
jacob-carlborg: So you mean a fourth file? There's already `section_osx.d`, `section_osx32.d` and `section_osx64.d`.

I'll wait for some additional input before making any changes.

bitwise-github: Actually, taking a fresh look at things this morning, these three files are so short that I don't think it really makes sense to split them up. Can we just keep it all in one file?

jacob-carlborg: I just tried to avoid `version (D_LP64) {} else {}` which Walter is not so found of, IIRC.

bitwise-github: What if we keep the three files, and import sections_osx.d into sections_osxXX so we can do this is sections.d?

    else version(OSX)
        version(D_LP64)
            public import rt.sections_osx64
        else
            public import rt.sections_osx32

jacob-carlborg: I guess that could work. I would still like some input form someone else as well.

bitwise-github: Ok, but I strongly believe that any of the solutions that I've suggested will have us in a better spot than a mixin.

bitwise-github: Actually, Looking at my code for shared libraries, I would have to change the SectionGroup struct and the onAddImage function as well. And per Walter's response in the forums, shared libraries should be made available for 64bit only as well. So this eliminates/differentiates most of what's still being shared between the two files sections_osx32.d and sections_osx64.d.

So could we go with my last suggestion, and have one file for each platform, and replace the imports in sections.d with this?

    else version(OSX)
    {
        version(D_LP64)
            public import rt.sections_osx64
        else
            public import rt.sections_osx32
    }

jacob-carlborg: @bitwise-github I've updated the code now, structured according to your suggestion.

bitwise-github: Ok, thanks.

src/rt/osx_tls.c
```diff
@@ -0,0 +1,49 @@
...
+    size_t info_size;
+    void * tlv_addr;
+    size_t tlv_size;
+} dyld_tlv_info;
+typedef void (^dyld_tlv_state_change_handler)(enum dyld_tlv_states state, const dyld_tlv_info *info);
+extern void dyld_register_tlv_state_change_handler(enum dyld_tlv_states state, dyld_tlv_state_change_handler handler);
+extern void dyld_enumerate_tlv_storage(dyld_tlv_state_change_handler handler);
+
+void _d_dyld_getTLSRange(void* arbitraryTLSSymbol, void** start, size_t* size) {
+    dyld_enumerate_tlv_storage(
```
MartinNowak: Please don't use weird vendor extensions.

src/rt/osx_tls.c
```diff
@@ -0,0 +1,49 @@
...
+        ^(enum dyld_tlv_states state, const dyld_tlv_info *info) {
```
MartinNowak: Please don't use weird vendor extensions, for sure you can do it with a plain callback or a struct.

src/rt/osx_tls.c
```diff
@@ -0,0 +1,49 @@
...
+ * Helpers for determining TLS memory ranges on OS X.
+ *
+ * This unfortunately cannot be entirely done in D, as the OS X API uses
+ * the Apple-specific blocks C extension.
+ *
+ * Copyright: David Nadlinger, 2012.
+ * License:   <a href="http://www.boost.org/LICENSE_1_0.txt">Boost License 1.0</a>.
+ * Authors:   David Nadlinger
+ */
+
```
MartinNowak: Check for OSX, not for lambda support.

src/rt/osx_tls.c
```diff
@@ -0,0 +1,49 @@
...
+#ifndef __BLOCKS__
```
MartinNowak: Check for OSX, not for lambda support.

src/rt/sections_osx_x86_64.d
```diff
@@ -0,0 +1,193 @@
...
+    if (ehsect != null)
+    {
+        debug(PRINTF) printf("  deh_eh\n");
+        auto p = cast(immutable(FuncTable)*)ehsect.ptr;
+        immutable len = ehsect.length / (*p).sizeof;
+
+        _sections._ehTables = p[0 .. len];
+    }
+
+    auto tlssect = getSection(h, slide, "__DATA", "__tls_data");
```
MartinNowak: You should connect this info to the actual allocate TLS range above.

src/rt/osx_tls.c
```diff
@@ -0,0 +1,49 @@
...
+};
+typedef struct {
+    size_t info_size;
+    void * tlv_addr;
+    size_t tlv_size;
+} dyld_tlv_info;
+typedef void (^dyld_tlv_state_change_handler)(enum dyld_tlv_states state, const dyld_tlv_info *info);
+extern void dyld_register_tlv_state_change_handler(enum dyld_tlv_states state, dyld_tlv_state_change_handler handler);
+extern void dyld_enumerate_tlv_storage(dyld_tlv_state_change_handler handler);
+
```
MartinNowak: A dummy symbol will only get you a single section, try to connect this to __tls_data below.

src/rt/sections_osx_x86_64.d
```diff
@@ -0,0 +1,193 @@
...
+ * Boolean flag set to true while the runtime is initialized.
+ */
+__gshared bool _isRuntimeInitialized;
+
+/****
+ * Gets called on program startup just before GC is initialized.
+ */
+void initSections()
+{
+    _dyld_register_func_for_add_image(&sections_osx_onAddImage);
```
MartinNowak: Well unless something changed this function is still a dead-end because you cannot deregister the callback, meaning that this change makes D shared libraries (a VST plugin) ununloadable.

bitwise-github: I have a fix on the way to add linux ctors/dtors to all modules and replace this callback usage.

src/rt/osx_tls.c
klickverbot: The API is only offered using a blocks-based callback. The need to do that is in fact the only reason why the code exists in an extra `.c` file. I briefly considered reverse-engineering the Blocks ABI to use it from D, but just using it the way it is supposed to seems like the much easier and more robust solution.

src/rt/osx_tls.c
jacob-carlborg: There's a theoretical possibility that someone will try to compile this with GCC. So perhaps check for both OS X and blocks? @klickverbot will it be possible for LDC to use this functionality? In that case what, preprocessor constant can I use to cover all Apple platforms?

src/rt/sections_osx_x86_64.d
jacob-carlborg: > Well unless something changed this function is still a dead-end because you cannot deregister the callback

@MartinNowak this PR won't fix dynamic libraries. It implements native TLS, nothing else. But native TLS is a prerequisite to fix dynamic libraries.

> meaning that this change makes D shared libraries (a VST plugin) ununloadable.

I don't see how this PR would affect the way dynamic libraries work. The state of dynamic libraries will not be any worse than they are now.

src/rt/sections_osx_x86_64.d
jacob-carlborg: `__tls_data` is not used anymore with the native TLS on x86-64. Previously this code was shared, I forgot to remove it when I made it x86-64 specific.

src/rt/osx_tls.c
jacob-carlborg: I'm not sure I understand this comment. Is it still relevant when `__tls_data` is not used/removed?

klickverbot: There should be only a single section given that shared libraries are not supported on OS X yet.

jacob-carlborg: How does this work for LDC? It does support dynamic libraries?

src/rt/osx_tls.c
klickverbot: > @klickverbot will it be possible for LDC to use this functionality?

Huh? The file is from LDC's druntime, so of course we use it already. If your question is about using Blocks from D, then I don't think anybody is planning to add this to the ObjC interop implementation just yet.

In either case, the file is supposed to be excluded from the build on all other platforms anyway, so strictly speaking, the `#ifndef` is completely unnecessary. However, since the error messages produced by tools that don't support Blocks can be quite puzzling (especially if you are not aware that Blocks even exist), I added the extra check to get a clean preprocessor error. Feel free to change it so that it refers to another version, or remove it entirely; I don't think a strong argument is to be made either way. You'd need to look into what other preprocessor constant to use yourself, though.

src/rt/osx_tls.c
klickverbot: We don't really do yet.

src/rt/sections_osx_x86_64.d
jacob-carlborg: @MartinNowak removed `__tls_data` on 64bit.

src/rt/osx_tls.c
jacob-carlborg: @MartinNowak checking for both OS X and blocks now.

src/rt/osx_tls.c
MartinNowak: You should've already figured out how generic native tls accesses work (for the dmd part), so why introducing sth. that's incompatible with shared libraries?

src/rt/sections_osx_x86_64.d
MartinNowak: You can use D shared libraries as plugins for C hosts (statically linking against phobos), this change breaks the use-case. So please finish your work on ctors/dtors first, or come up w/ a different solution.

jacob-carlborg: Master is using `_dyld_register_func_for_add_image`, this change is still using `_dyld_register_func_for_add_image `. Please explain how this change breaks the shared libraries?

jacob-carlborg: And if you want to support shared libraries, theres should be a test for that.

MartinNowak: Sorry, didn't remember that we've always used those callbacks, would still be great to replace them soon.

jacob-carlborg: > Sorry, didn't remember that we've always used those callbacks

No worries :smiley:.

> would still be great to replace them soon.

Absolutely. But let's take this one step at the time.

bitwise-github: > Absolutely. But let's take this one step at the time.

Indeed, I would like to go at the ctor problem knowing we've got a clean slate(all tests passing)

jacob-carlborg:
New PR is: https://github.com/D-Programming-Language/druntime/pull/1523

PR #1469, opened 2016-01-16
Revert "Merge pull request #1445 from WalterBright/message"

MartinNowak:
This reverts commit 74e5d87148028134e6927c31e99c293e0e2f91bd, reversing
changes made to c36cc5180a3ad59b099dec91bf0055d2e6508753.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -1689,7 +1689,6 @@ class Throwable : Object
         sink("@"); sink(file);
         sink("("); sink(sizeToTempString(line, tmpBuff, 10)); sink(")");
 
-        auto msg = message();
         if (msg.length)
         {
             sink(": "); sink(msg);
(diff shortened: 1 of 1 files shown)


MartinNowak:
As discussed delaying the addition to after 2.070.

src/object.d
```diff
@@ -1689,7 +1689,6 @@ class Throwable : Object
         sink("@"); sink(file);
         sink("("); sink(sizeToTempString(line, tmpBuff, 10)); sink(")");
 
-        auto msg = message();
```
MartinNowak: BTW, this wouldn't work if `message()` would allocate. We use the sink based toString to print exception messages when the GC cannot be used.

andralex:
This has been vetted by Sociomantic. We'll have an e-meeting with Sociomantic on Monday to make plans for a permanent solution.

PR #1470, opened 2016-01-17
Add catching C++ exception support to dwarfeh.d

WalterBright:
Requires:
https://github.com/D-Programming-Language/dmd/pull/5342

### DIFF
--- a/src/rt/dwarfeh.d
+++ b/src/rt/dwarfeh.d
@@ -1,7 +1,7 @@
-/*
+/**
  * Exception handling support for Dwarf-style portable exceptions.
  *
- * Copyright: Copyright 2015 D Language Foundation
+ * Copyright: Copyright (c) 2015-2016 by D Language Foundation
  * License: Distributed under the
  *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
  *    (See accompanying file LICENSE)
@@ -275,6 +275,7 @@ extern (C) _Unwind_Reason_Code __dmd_personality_v0(int ver, _Unwind_Action acti
                _Unwind_Context* context)
 {
     //printf("__dmd_personality_v0(actions = x%x, eo = %p, context = %p)\n", cast(int)actions, exceptionObject, context);
+    //printf("exceptionClass = x%08lx\n", exceptionClass);
     if (ver != 1)
       return _URC_FATAL_PHASE1_ERROR;
     assert(context);
@@ -283,7 +284,6 @@ extern (C) _Unwind_Reason_Code __dmd_personality_v0(int ver, _Unwind_Action acti
     int handler;
     _Unwind_Ptr landing_pad;
 
-    ExceptionHeader *eh = ExceptionHeader.toExceptionHeader(exceptionObject);
     //for (auto ehx = eh; ehx; ehx = ehx.next)
         //printf(" eh: %p next=%014p lsda=%p '%.*s'\n", ehx, ehx.next, ehx.languageSpecificData, ehx.object.msg.length, ehx.object.msg.ptr);
 
@@ -304,7 +304,7 @@ extern (C) _Unwind_Reason_Code __dmd_personality_v0(int ver, _Unwind_Action acti
     auto result = scanLSDA(language_specific_data, ip - Start, exceptionClass,
         (actions & _UA_FORCE_UNWIND) != 0,          // don't catch when forced unwinding
         (actions & _UA_SEARCH_PHASE) != 0,          // search phase is looking for handlers
-        getClassInfo(eh),
+        exceptionObject,
         landing_pad,
         handler);
     landing_pad += Start;
@@ -343,11 +343,13 @@ extern (C) _Unwind_Reason_Code __dmd_personality_v0(int ver, _Unwind_Action acti
             assert(!(actions & _UA_FORCE_UNWIND));
             if (actions & _UA_SEARCH_PHASE)
             {
-                assert(exceptionClass == dmdExceptionClass);   // Can't do handlers for anything but D
-                eh.handler = handler;
-                eh.languageSpecificData = language_specific_data;
-                eh.landingPad = landing_pad;
-
+                if (exceptionClass == dmdExceptionClass)
+                {
+                    auto eh = ExceptionHeader.toExceptionHeader(exceptionObject);
+                    eh.handler = handler;
+                    eh.languageSpecificData = language_specific_data;
+                    eh.landingPad = landing_pad;
+                }
                 return _URC_HANDLER_FOUND;
             }
             break;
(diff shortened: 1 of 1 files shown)


src/rt/dwarfeh.d
```diff
@@ -780,3 +807,99 @@ void terminate(uint line) @nogc
...
+    else
+        p = cast(void*)(exceptionObject + 1);           // thrown object is immediately after it
+
+    const tt = (cast(CppExceptionHeader*)p - 1).typeinfo;
+
+    if (tt.__is_pointer_p())
+        p = *cast(void**)p;
+
+    // Pointer adjustment may be necessary due to multiple inheritance
+    return (sti.__do_catch(tt, &p, 1)) ? p : null;
```
ibuclaw: Maybe add a pointer comparison before this to shortcut calling do_catch?  If ``sti == tt`` then it's a direct match.

IgorStepanov: It's done inside `__do_catch`:
```
bool __class_type_info::
__do_catch (const type_info *thr_type,
            void **thr_obj,
            unsigned outer) const
{
  if (*this == *thr_type)
    return true;
  if (outer >= 4)
    // Neither `A' nor `A *'.
    return false;
  return thr_type->__do_upcast (this, thr_obj);
}
```

src/rt/dwarfeh.d
```diff
@@ -780,3 +807,99 @@ void terminate(uint line) @nogc
...
+    void* p;    // pointer to thrown object
+    if (exceptionObject.exception_class & 1)
+        p = CppExceptionHeader.toExceptionHeader(exceptionObject).ptr;
+    else
+        p = cast(void*)(exceptionObject + 1);           // thrown object is immediately after it
+
+    const tt = (cast(CppExceptionHeader*)p - 1).typeinfo;
+
+    if (tt.__is_pointer_p())
+        p = *cast(void**)p;
```
IgorStepanov: Why do you do it here?
If C++ code throws pointer to a class, you should do upcast before (via `__do_catch`).

```

try
{
    throwCppException(); //throw new std::bad_exception();
}
catch(std.exception e) // we need to cast std::bad_exception** to std::exception** before the exception pointer dereferencing.
{
}
```

src/rt/dwarfeh.d
ibuclaw: That is not actually true.  ``(*this == *thr_type)`` calls ``operator==``.

src/rt/dwarfeh.d
ibuclaw: If you really think this is problematic, raise a bug against g++.  They do the same thing. :-)

src/rt/dwarfeh.d
IgorStepanov: `operator==` in my libstdc++ looks like
```
bool std::type_info::
operator== (const std::type_info& arg) const
{
  return (&arg == this) || (__builtin_strcmp (name (), arg.name ()) == 0);
}
```
However in other versions it may differ.

src/rt/dwarfeh.d
IgorStepanov: You are right, `get_adjusted_ptr` has the similar implementation. It seems I did not understand the purpose of this code.

src/rt/dwarfeh.d
ibuclaw: You may find that it's been that way for a very long time.  C++ is not something that simply changes.

WalterBright: > Maybe add a pointer comparison before this to shortcut calling do_catch? If  sti == tt  then it's a direct match.

Done, but given all the other code executed for an exception, I doubt the difference would be measurable.

IgorStepanov:
BTW, will be this exception handling work with some another c++ compilers? clang++? intel c++ compiler?

WalterBright:
> will be this exception handling work with some another c++ compilers? clang++? intel c++ compiler?

My understanding is that they are binary compatible with g++, so they should work.

IgorStepanov:
>My understanding is that they are binary compatible with g++, so they should work.

Are they use the same libstdc++ implementation? Otherwise there can be another virtual method order in std::type_info for example.

WalterBright:
You can be binary compatible and yet have a different implementation. Again, my understanding is they are binary compatible.

IgorStepanov:
>You can be binary compatible and yet have a different implementation. Again, my understanding is they are binary compatible.

Sounds good. It remains to deal with dmc and Visual Studio :)

src/rt/dwarfeh.d
```diff
@@ -343,11 +343,13 @@ extern (C) _Unwind_Reason_Code __dmd_personality_v0(int ver, _Unwind_Action acti
...
             if (actions & _UA_SEARCH_PHASE)
             {
-                assert(exceptionClass == dmdExceptionClass);   // Can't do handlers for anything but D
-                eh.handler = handler;
-                eh.languageSpecificData = language_specific_data;
-                eh.landingPad = landing_pad;
-
+                if (exceptionClass == dmdExceptionClass)
+                {
+                    ExceptionHeader *eh = ExceptionHeader.toExceptionHeader(exceptionObject);
```
andralex: this line in desperate need of `auto`

src/rt/dwarfeh.d
```diff
@@ -357,50 +359,54 @@ extern (C) _Unwind_Reason_Code __dmd_personality_v0(int ver, _Unwind_Action acti
...
 
     // Figure out what to do when there are multiple exceptions in flight
-    while (eh.next)
+    if (exceptionClass == dmdExceptionClass)
     {
-        ExceptionHeader* ehn = eh.next;
-
-        // Don't combine when the exceptions are from different functions
-        if (language_specific_data != ehn.languageSpecificData)
+        ExceptionHeader *eh = ExceptionHeader.toExceptionHeader(exceptionObject);
```
andralex: this too

src/rt/dwarfeh.d
```diff
@@ -357,50 +359,54 @@ extern (C) _Unwind_Reason_Code __dmd_personality_v0(int ver, _Unwind_Action acti
...
-                handler = ehn.handler;
+                //printf("break: %p %p\n", language_specific_data, ehn.languageSpecificData);
+                break;
+            }
 
-                eh.handler = handler;
-                eh.languageSpecificData = language_specific_data;
-                eh.landingPad = landing_pad;
+            Error e = cast(Error)eh.object;
+            if (e !is null && (cast(Error)ehn.object) is null)
```
andralex: is that pair of parens needed?

andralex: I'd write it simply as `if (e && cast(Error)ehn.object)`

src/rt/dwarfeh.d
```diff
@@ -752,8 +765,22 @@ int actionTableLookup(ClassInfo catchType, uint actionRecordPtr, const(ubyte)* p
...
-        if (_d_isbaseof(catchType, ci))
+        if (ci.classinfo is __cpp_type_info_ptr.classinfo)
+        {
+            if (exceptionClass == cppExceptionClass || exceptionClass == cppExceptionClass1)
+            {
+                // sti is catch clause type_info
+                auto sti = cast(CppTypeInfo)((cast(__cpp_type_info_ptr)cast(void*)ci).ptr);
+                auto p = getCppPtrToThrownObject(exceptionObject, sti);
+                if (p)
+                {   // found it
```
andralex: brace on its own line

src/rt/dwarfeh.d
```diff
@@ -752,8 +765,22 @@ int actionTableLookup(ClassInfo catchType, uint actionRecordPtr, const(ubyte)* p
...
+                    auto eh = CppExceptionHeader.toExceptionHeader(exceptionObject);
```
andralex: zat's da spirit

andralex:
thx!

PR #1473, opened 2016-01-19
add core.stdcpp.exception and core.stdcpp.typeinfo

WalterBright:
gotta start somewhere

### DIFF
--- a/src/core/stdcpp/exception.d
+++ b/src/core/stdcpp/exception.d
@@ -1,0 +1,106 @@
+// Written in the D programming language.
+
+/**
+ * Interface to C++ <exception>
+ *
+ * Copyright: Copyright (c) 2016 D Language Foundation
+ * License:   $(WEB boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Authors:   $(WEB digitalmars.com, Walter Bright)
+ * Source:    $(DRUNTIMESRC core/stdcpp/_exception.d)
+ */
+
+module core.stdcpp.exception;
+
+version (CRuntime_DigitalMars)
+{
+    import core.stdcpp.typeinfo;
+
+    extern (C++, std)
+    {
+        alias void function() unexpected_handler;
+        unexpected_handler set_unexpected(unexpected_handler f) nothrow;
+        void unexpected();
+
+        alias void function() terminate_handler;
+        terminate_handler set_terminate(terminate_handler f) nothrow;
+        void terminate();
+
+        bool uncaught_exception();
+
+        class exception
+        {
+            this() nothrow { }
+            this(const exception) nothrow { }
+            //exception operator=(const exception) nothrow { return this; }
+            //virtual ~this() nothrow;
+            void dtor() { }
+            const(char)* what() const nothrow;
+        }
+
+        class bad_exception : exception
+        {
+            this() nothrow { }
+            this(const bad_exception) nothrow { }
+            //bad_exception operator=(const bad_exception) nothrow { return this; }
+            //virtual ~this() nothrow;
+            override const(char)* what() const nothrow;
+        }
+    }
+}
+else version (CRuntime_Glibc)
+{
+    extern (C++, std)
+    {
+        alias void function() unexpected_handler;
+        unexpected_handler set_unexpected(unexpected_handler f) nothrow;
+        void unexpected();
+
+        alias void function() terminate_handler;
+        terminate_handler set_terminate(terminate_handler f) nothrow;
+        void terminate();
+
+        void unexpected();
+        pure bool uncaught_exception();
+
+        class exception
+        {
+            this();
+            //virtual ~this();
+            void dtor1();
+            void dtor2();
+            const(char)* what() const;
+        }
+
+        class bad_exception : exception
+        {
+            this();
+            //virtual ~this();
+            const(char)* what() const;
+        }
+    }
+}
+else version (CRuntime_Microsoft)
+{
+    extern (C++, std)
+    {
+        class exception
+        {
+            this();
+            this(const exception);
+            //exception operator=(const exception) { return this; }
+            //virtual ~this();
+            void dtor() { }
+            const(char)* what() const;
+
+          private:
+            const(char)* mywhat;
+            bool dofree;
+        }
+
+        class bad_exception : exception
+        {
+            this(const(char)* msg = "bad exception");
+            //virtual ~this();
+        }
+    }
+}
--- a/src/core/stdcpp/typeinfo.d
+++ b/src/core/stdcpp/typeinfo.d
@@ -1,0 +1,146 @@
+// Written in the D programming language.
+
+/**
+ * Interface to C++ <typeinfo>
+ *
+ * Copyright: Copyright (c) 2016 D Language Foundation
+ * License:   $(WEB boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Authors:   $(WEB digitalmars.com, Walter Bright)
+ * Source:    $(DRUNTIMESRC core/stdcpp/_typeinfo.d)
+ */
+
+module core.stdcpp.typeinfo;
+
+version (CRuntime_DigitalMars)
+{
+    import core.stdcpp.exception;
+
+    extern (C++, std)
+    {
+        class type_info
+        {
+            void* pdata;
+
+          public:
+            //virtual ~this();
+            void dtor() { }     // reserve slot in vtbl[]
+
+            //bool operator==(const type_info rhs) const;
+            //bool operator!=(const type_info rhs) const;
+            final bool before(const type_info rhs) const;
+            final const(char)* name() const;
+          protected:
+            //type_info();
+          private:
+            //this(const type_info rhs);
+            //type_info operator=(const type_info rhs);
+        }
+
+        class bad_cast : core.stdcpp.exception.std.exception
+        {
+            this() nothrow { }
+            this(const bad_cast) nothrow { }
+            //bad_cast operator=(const bad_cast) nothrow { return this; }
+            //virtual ~this() nothrow;
+            override const(char)* what() const nothrow;
+        }
+
+        class bad_typeid : core.stdcpp.exception.std.exception
+        {
+            this() nothrow { }
+            this(const bad_typeid) nothrow { }
+            //bad_typeid operator=(const bad_typeid) nothrow { return this; }
+            //virtual ~this() nothrow;
+            override const (char)* what() const nothrow;
+        }
+    }
+}
+else version (CRuntime_Microsoft)
+{
+    import core.stdcpp.exception;
+
+    struct __type_info_node
+    {
+        void* _MemPtr;
+        __type_info_node* _Next;
+    }
+
+    extern __gshared __type_info_node __type_info_root_node;
+
+    extern (C++, std)
+    {
+        class type_info
+        {
+
+          public:
+            //virtual ~this();
+            void dtor() { }     // reserve slot in vtbl[]
+            //bool operator==(const type_info rhs) const;
+            //bool operator!=(const type_info rhs) const;
+            final bool before(const type_info rhs) const;
+            final const(char)* name(__type_info_node* p = &__type_info_root_node) const;
+
+          private:
+            void* pdata;
+            char _name[1];
+            //this(const type_info rhs);
+            //type_info operator=(const type_info rhs);
+        }
+
+        class bad_cast : core.stdcpp.exception.std.exception
+        {
+            this(const(char)* msg = "bad cast") { }
+            this(const bad_cast) { }
+            //virtual ~this();
+        }
+
+        class bad_typeid : core.stdcpp.exception.std.exception
+        {
+            this(const(char)* msg = "bad typeid") { }
+            this(const bad_typeid) { }
+            //virtual ~this();
+        }
+    }
+}
+else version (CRuntime_Glibc)
+{
+    import core.stdcpp.exception;
+
+    extern (C++, __cxxabiv1)
+    {
+        class __class_type_info;
+    }
+
+    extern (C++, std)
+    {
+        class type_info
+        {
+            void dtor1();                           // consume destructor slot in vtbl[]
+            void dtor2();                           // consume destructor slot in vtbl[]
+            final const(char)* name();
+            final bool before(const type_info) const;
+            //bool operator==(const type_info) const;
+            bool __is_pointer_p() const;
+            bool __is_function_p() const;
+            bool __do_catch(const type_info, void**, uint) const;
+            bool __do_upcast(const __cxxabiv1.__class_type_info, void**) const;
+
+            const(char)* _name;
+            this(const(char)*);
+        }
+
+        class bad_cast : core.stdcpp.exception.std.exception
+        {
+            this();
+            //~this();
+            const(char)* what() const;
+        }
+
+        class bad_typeid : core.stdcpp.exception.std.exception
+        {
+            this();
+            //~this();
+            const(char)* what() const;
+        }
+    }
+}
--- a/mak/COPY
+++ b/mak/COPY
@@ -43,6 +43,9 @@ COPY=\
 	$(IMPDIR)\core\stdc\wchar_.d \
 	$(IMPDIR)\core\stdc\wctype.d \
 	\
+	$(IMPDIR)\core\stdcpp\typeinfo.d \
+	$(IMPDIR)\core\stdcpp\exception.d \
+	\
 	$(IMPDIR)\core\sys\unwind.d \
 	\
 	$(IMPDIR)\core\sys\freebsd\dlfcn.d \
--- a/win32.mak
+++ b/win32.mak
@@ -199,6 +199,7 @@ $(IMPDIR)\core\sync\semaphore.di : src\core\sync\semaphore.d
 
 copydir: $(IMPDIR)
 	mkdir $(IMPDIR)\core\stdc
+	mkdir $(IMPDIR)\core\stdcpp
 	mkdir $(IMPDIR)\core\internal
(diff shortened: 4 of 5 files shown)


src/core/stdcpp/exception.d
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * Copyright: Copyright (c) 2016 D Language Foundation
+ * License:   $(WEB boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Authors:   $(WEB digitalmars.com, Walter Bright)
+ * Source:    $(DRUNTIMESRC core/stdcpp/_exception.d)
+ */
+
+module core.stdcpp.exception;
+
+version (DigitalMars) version (_Win32)
```
rainers: Debug code? CRuntime_DigitalMars is predefined, and _Win32 is not.

rainers:
No update to MANIFEST? I guess we can nuke it then ;-)

WalterBright:
> No update to MANIFEST? I guess we can nuke it then ;-)

Yeah, I gave up on it. If you want to do a PR to nuke it, I'll pull it.

src/core/stdcpp/typeinfo.d
```diff
@@ -0,0 +1,146 @@
...
+        class __class_type_info;
+    }
+
+    extern (C++, std)
+    {
+        class type_info
+        {
+            void dtor1();                           // consume destructor slot in vtbl[]
+            void dtor2();                           // consume destructor slot in vtbl[]
+            final const(char)* name();
```
IgorStepanov: If you declares final method, you add dependence from libstdc++, because this external symbol should be resolved by linker. Is it ok?

src/core/stdcpp/typeinfo.d
```diff
@@ -0,0 +1,146 @@
...
+            final bool before(const type_info) const;
+            //bool operator==(const type_info) const;
+            bool __is_pointer_p() const;
+            bool __is_function_p() const;
+            bool __do_catch(const CppTypeInfo, void**, uint) const;
```
IgorStepanov: You want to say `std.type_info`?

src/core/stdcpp/typeinfo.d
```diff
@@ -0,0 +1,146 @@
...
+            bool __do_upcast(const __cxxabiv1.__class_type_info, void**) const;
+
+            const(char)* _name;
+            this(const(char)*);
```
IgorStepanov: Is external `C++` constructors can be correctly used from D? Do we really need to declare class internals (instead of using of `extern(C++) interface`? Anyway we will not move all C++ type_info hierarchy into druntime, as far as I understand.

src/core/stdcpp/exception.d
```diff
@@ -0,0 +1,106 @@
...
+
+        void unexpected();
+        pure bool uncaught_exception();
+
+        class exception
+        {
+            this();
+            //virtual ~this();
+            void dtor();
+            void dtor();
```
IgorStepanov: dtor2 ?

src/core/stdcpp/typeinfo.d
```diff
@@ -0,0 +1,146 @@
...
+        {
+            this() nothrow { }
+            this(const bad_typeid) nothrow { }
+            //bad_typeid operator=(const bad_typeid) nothrow { return this; }
+            //virtual ~this() nothrow;
+            override const (char)* what() const nothrow;
+        }
+    }
+}
+else version (CRuntime_Microsoft)
```
IgorStepanov: BTW, do you plan to implement exception catching for VC++? I have some ideas, how we may implement exception catching and dynamic casting (which needed for polymorphic types catching), but I totally haven't free time until February.

src/core/stdcpp/typeinfo.d
```diff
@@ -0,0 +1,146 @@
...
+            bool __is_pointer_p() const;
+            bool __is_function_p() const;
+            bool __do_catch(const CppTypeInfo, void**, uint) const;
+            bool __do_upcast(const __cxxabiv1.__class_type_info, void**) const;
+
+            const(char)* _name;
+            this(const(char)*);
+        }
+
+        class bad_cast : core.stdcpp.exception.std.exception
```
jpf91: Can we use `core.stdcpp.exception` or at least `std.exception` or does that require the alias workarounds discussed in @TurkeyMan newsgroup thread? (I didn't read the complete discussion)

src/core/stdcpp/exception.d
```diff
@@ -0,0 +1,106 @@
...
+        class exception
+        {
+            this();
+            this(const exception);
+            //exception operator=(const exception) { return this; }
+            //virtual ~this();
+            void dtor() { }
+            const(char)* what() const;
+
+          private:
```
IgorStepanov: Strange intendation, 10 spaces. Is it ok?

src/core/stdcpp/typeinfo.d
```diff
@@ -0,0 +1,146 @@
...
+          private:
+            void* pdata;
+            char _name[1];
+            //this(const type_info rhs);
+            //type_info operator=(const type_info rhs);
+        }
+
+        class bad_cast : core.stdcpp.exception.std.exception
+        {
+            this(const(char)* msg = "bad cast") { }
```
IgorStepanov: This constructor do nothing. When you declared ` core.stdcpp.exception.std.bad_exception` you declared constructor as external: `this(const(char)* msg = "bad exception");`.
What way is right? How external C++ constructors works in D code? Who should allocate memory, С++ ctor, or D code?

src/core/stdcpp/typeinfo.d
```diff
@@ -0,0 +1,146 @@
...
+            this(const bad_cast) { }
+            //virtual ~this();
+        }
+
+        class bad_typeid : core.stdcpp.exception.std.exception
+        {
+            this(const(char)* msg = "bad typeid") { }
```
IgorStepanov: Same here.

src/core/stdcpp/typeinfo.d
IgorStepanov: I think we should summon @yebblies .

src/core/stdcpp/typeinfo.d
WalterBright: That's why it is currently not compiled into the library. In the future, it should be compiled into the .a, but not the .so.

src/core/stdcpp/typeinfo.d
yebblies: If you declare a constructor in the D code and call it from the D code, it will be allocated on the D side (ie with the GC).  It will not currently link to the C++ ctor, and if it did they wouldn't agree about ABI details (eg D assumes init[] has been blitted over)

IgorStepanov: I asked about [external](https://github.com/D-Programming-Language/druntime/pull/1473/files#diff-df057fd1f678baa077a5e4097364ccdaR102) constructor. Is it an error?

yebblies: What is an external constructor?  If you `new` a C++ class from D it will be allocated using the GC.

IgorStepanov: >What is an external constructor?

Look by the reference.
Walter have written the follofing code:
```
class bad_exception : exception
{
    this(const(char)* msg = "bad exception");
    //virtual ~this();
 }
```
This ctor haven't body and at the first look it should be linked with original C++ ctor and `new bad_exception()` should call C++ ctor. Does this magic works? I think, it may be implemented, because C++ has a placement constructor and there should be possibility to allocate memory in D and pass it to C++ ctor. Moreover, D compiler may generate D dtor, which may call C++ dtor ("complete object destructor" which doesn't do memory deallocation) and call it at garbarge collection.
However, what happens in this case now?

yebblies: As I said, D does not match C++ constructor mangling or ABI, so it shouldn't link, and if it did then it would corrupt memory and/or crash.  It could be made to work.

IgorStepanov: >if it did then it would corrupt memory and/or crash.

Looks like the "complete object constructor/destructor" (C1/D1 mangling) has the similar signature with D dtor/ctor: `abi::__cxa_cdtor_return_type (*)(void*)` where __cxa_cdtor_return_type is `void` or `void*`. Thus we should may possibility to call C++ ctor as D ctor:
```
void* __d_create_cpp_object(size_t size, void function(void*) ctor)
{
    auto obj = GC.calloc(size);
    //blitting init[] is possible but isn't needed.
    ctor(obj);
    return obj;
}
```

yebblies: Sure it's possible, but not implemented or tested.  And calling constructors written in D from C++ would require changes to ctor codegen IIRC.

IgorStepanov: What another limitation of using extern(C++) classes, which defined in D? As I understand, `dynamic_cast` can't be applied to D-born classes, because D can't generate correct pointer to C++ RTTI. Moreover, С++ code shouldn't throw those classes. Is there other limitations?

src/core/stdcpp/typeinfo.d
WalterBright: > Is external  C++  constructors can be correctly used from D?

Not at the moment, but perhaps we can improve on that.

> Do we really need to declare class internals (instead of using of  extern(C++) interface ?

Yes, if you're going to make a class derived from it.

src/core/stdcpp/typeinfo.d
WalterBright: > do you plan to implement exception catching for VC++?

Yes, but it's not high on the agenda. Anyone who wants to champion this in the meantime is welcome to have a go.

src/core/stdcpp/typeinfo.d
WalterBright: Let's leave that debate in the n.g. and out of this thread, please.

src/core/stdcpp/exception.d
WalterBright: I've always used a half-indent for public/private.

src/core/stdcpp/typeinfo.d
WalterBright: Constructors are something we can improve on, but later. I put it there mostly as a note for the future.

IgorStepanov: I suggest to declare them with empty bodies (`{ }`) to avoid of undefined reference error. When C++ ctors/dtors will work correctly we may rewrite druntime code.

src/core/stdcpp/typeinfo.d
IgorStepanov: >Yes, if you're going to make a class derived from it.

Idea of inheritance with extern(C++) classes scared me.
We can't generate correct C++ RTTI thus we can't use C++ features based on RTTI (for example dynamic_cast). As result, we shouldnt pass those classes to C++ code.
When we define all hierarchy of those classes in D there are no problem with casting. If part of hierarchy defined in C++ and declared in D as `extern(C++) interface`, we can't make dynamic casting in D, because we lost information about base class fields (and, I think, we don't want to move multiple inheritance ABI to D compiler). If we will use native C++ dynamic_cast in D (that would like to make), we should have C++ RTTI.
How we can define safe rules of using extern(C++) classes in D.
Sorry for semi-offtopic.

yebblies: We could implement C++ RTTI, but using C++ without RTTI is not that strange.

IgorStepanov: If C++ code uses RTTI, and we pass our D-constructed object, it may cause strange unexpected error:
```
//C++
void thridLibraryFunction(std::exception *ex)
{
    if (std::bad_typeid *ebti = dynamic_cast<std::bad_typeid>(ex))
    {
        //do something
    }
    else
    {
         //do something else
    }
}


//D code:

extern(C++) void thridLibraryFunction(std.exception ex);

extern(C++) class d_exception : std.exception
{
     override const (char)* what() const nothrow
     {
          return "From D with love";
     }
}

void main()
{
    thridLibraryFunction(new d_exception()); //unexpected segfault on dynamic_cast
}
```

>We could implement C++ RTTI

It may be difficult for VC++, because VC++ rtti is undocumented.
clang++ probably resolves this trouble and I have reverse-enginered scheme of VC++ RTTI, but do we want to take on this headache?

yebblies: It's an expected segfault if you know that RTTI is not supported.  IIRC most C++ compilers have an option to disable RTTI, so it's not that strange.

> It may be difficult for VC++, because VC++ rtti is undocumented.
clang++ probably resolves this trouble and I have reverse-enginered scheme of VC++ RTTI, but do we want to take on this headache?

I doubt it's as difficult to implement as EH.  And as I've said I think we can probably live without C++ RTTI support.

src/core/stdcpp/typeinfo.d
jpf91: Sounds complicated. Then you can use `core.stdcpp` with static linking, but not with dynamic linking?
But as druntime won't use `core.stdcpp` internally, how about putting all `core.stdcpp` modules into a extra library, `stdcpp`. We can still ship this library with dmd. Then users using `core.stdcpp` simply add `-lstdcpp` to their linker flags and users not needing `core.stdcpp` never get a dependency on `libstdc++`.

src/core/stdcpp/typeinfo.d
jpf91: I don't want to reopen that debate. I'd simply suggest to use `std.exception` or maybe `core.stdcpp.exception.exception` instead if possible. `core.stdcpp.exception.std.exception` is a quite ugly / long name. According to some newsgroups posts it should be possible to simply use the C++ namespace form and that would look much nicer.

src/core/stdcpp/exception.d
klickverbot: It seems like this style isn't used anywhere else in druntime, though.

src/core/stdcpp/exception.d
```diff
@@ -0,0 +1,106 @@
...
+        alias void function() terminate_handler;
+        terminate_handler set_terminate(terminate_handler f) nothrow;
+        void terminate();
+
+        void unexpected();
+        pure bool uncaught_exception();
+
+        class exception
+        {
+            this();
```
IgorStepanov: I suggest to add empty bodies or bodies with `{assert(0);}` to all costructors declarations.
Also we may use `@disable` attribute for all unavailable methods. It will be better then strange linkage error on constructor using.

src/core/stdcpp/exception.d
```diff
@@ -0,0 +1,106 @@
...
+        {
+            this() nothrow { }
+            this(const bad_exception) nothrow { }
+            //bad_exception operator=(const bad_exception) nothrow { return this; }
+            //virtual ~this() nothrow;
+            override const(char)* what() const nothrow;
+        }
+    }
+}
+else version (CRuntime_Glibc)
```
IgorStepanov: Small nitpick: You wrote version `(CRuntime_Glibc)` before version `(CRuntime_Microsoft)` here, but in the next file you wrote version `(CRuntime_Microsoft)` earler. Maybe the similar order will be better? 
Perfectionism post :)

src/core/stdcpp/typeinfo.d
```diff
@@ -0,0 +1,146 @@
...
+    {
+        class __class_type_info;
+    }
+
+    extern (C++, std)
+    {
+        class type_info
+        {
+            void dtor1();                           // consume destructor slot in vtbl[]
+            void dtor2();                           // consume destructor slot in vtbl[]
```
IgorStepanov: This dtors also may be annotated with `@disable` to avoid thoughtless calling.

src/core/stdcpp/typeinfo.d
IgorStepanov: >I doubt it's as difficult to implement as EH.

Your optimism makes me happy. Ok.

>And as I've said I think we can probably live without C++ RTTI support.

Anyway, we need to implement dynamic casting over C++ classes for D. For example, when we have a code:
```
extern(C++){
interface A
{
    void foo();
}

interface B
{
    void boo();
}

interface C : A, B
{
    void bar();
}

C getC();

}

void main()
{
    auto c = getC();
    B b = c; //fail
}
```
we can't implement casting by using the information that D compiler has. We need to access to C++ RTTI (to `__do_upcast` method e. g.).
What should be do, if we declare class D, inherited from C?
```
extern(C++) class D : C 
{
    int a;
    int b;
}
```
What the layout of this class, when we don't know sizeof(C)?

And the last point: most С++ projects use RTTI. We will not provide attractive mechanism of moving C++ projects to D until we don't think of a way to get along with С++ RTTI.

rainers:
>> No update to MANIFEST? I guess we can nuke it then ;-)
> 
> Yeah, I gave up on it. If you want to do a PR to nuke it, I'll pull it.

Great! Here it is: https://github.com/D-Programming-Language/druntime/pull/1476

src/core/stdcpp/typeinfo.d
JakobOvrum: @IgorStepanov, it's better that it issues a linker error than silently failing. Calling an empty D constructor doesn't do the right thing.

JakobOvrum:
I agree that `@disable` is marginally better than the current linker errors for unsupported operators like constructors.

src/core/stdcpp/exception.d
andralex: urgh

src/core/stdcpp/typeinfo.d
```diff
@@ -0,0 +1,146 @@
...
+{
+    import core.stdcpp.exception;
+
+    extern (C++, std)
+    {
+        class type_info
+        {
+            void* pdata;
+
+          public:
```
andralex: yeah half indent has to go

andralex:
Auto-merge toggled on

PR #1474, opened 2016-01-20
dwarfeh: EDX and RDX are different register numbers for dwarf

WalterBright:
because having Dwarf register numbers line up with the CPU register numbers wouldn't make any sense (!)


MartinNowak:
This should have gone to stable of it's necessary for the 2.070.0 release.

ibuclaw:
Or just use the compiler intrinsics directly. :)

PR #1475, opened 2016-01-21
Fix CTFE popcnt() regression

tsbockman:
This fixes CTFE `popcnt()` which I accidentally broke in [#1452](https://github.com/D-Programming-Language/druntime/pull/1452).

That PR added optimized `popcnt()` by having it check `core.cpuid.hasPopcnt` to see if it can use the `_popcnt()` intrinsic. However, `hasPopcnt` is not accessible at compile time.

This PR simply skips that step if `__ctfe` is set, causing it to fall back to `soft_popcnt()` which works fine in CTFE.

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -271,9 +271,12 @@ int popcnt(uint x) pure
         {
             static if (is(typeof(_popcnt(uint.max))))
             {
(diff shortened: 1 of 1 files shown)


MartinNowak:
Auto-merge toggled on

PR #1476, opened 2016-01-21
get rid of mak/MANIFEST

rainers:
remove mak/MANIFEST, use "git ls-tree" instead

approved by @WalterBright in a recent comment: https://github.com/D-Programming-Language/druntime/pull/1473#issuecomment-172796230

I could not test the win32/win64 zip targets, because I have no clue where to get `zip32`, so I just expect it accepts the response file syntax. Passing all the files on the command line leads to a "command line too long" error anyway.


WalterBright:
Auto-merge toggled on

andralex:
Nice. Love the red. The command is run every time `MANIFEST` is used, is that often? If more than a couple times, use `MANIFEST:=...` instead.

rainers:
I wanted the shell command not to be excuted eagerly, so the makefile is usable without having git installed. It's now only called if you make the respective targets, i.e. twice for checking whitespaces (no changes from previous version), twice for detab and once for the zip file. Maybe I failed for the last one, as it might not by lazily evaluated when used as a dependency?

MartinNowak:
> I wanted the shell command not to be excuted eagerly, so the makefile is usable without having git installed.

Avoiding git should not be a concern on a non-Windows machine, and even on Windows it's almost a standard tool.

PR #1477, opened 2016-01-23
Issue 15595 - Interface cast on collection may fail

etcimon:
This solves an issue where the GC collection cycle would take out a vtable pointer before another destructor tried to access it when calling destroy() on the object. This covered 99.9% of the remaining causes of segmentation faults in my server and was unpredictable.

### DIFF
--- a/src/rt/cast_.d
+++ b/src/rt/cast_.d
@@ -53,7 +53,11 @@ void* _d_interface_cast(void* p, ClassInfo c)
     if (!p)
         return null;
 
-    Interface* pi = **cast(Interface***) p;
+    Interface* pi;
+    if (auto pi_ = *cast(Interface***)p)
+        pi = *pi_;
+    if (!pi)
+        return null;
 
     debug(cast_) printf("\tpi.offset = %d\n", pi.offset);
     return _d_dynamic_cast(cast(Object)(p - pi.offset), c);


src/rt/cast_.d
```diff
@@ -53,7 +53,12 @@ void* _d_interface_cast(void* p, ClassInfo c)
     if (!p)
         return null;
 
-    Interface* pi = **cast(Interface***) p;
+    Interface* pi;
+    if (auto pi_ = *cast(Interface***)p)
+    {
+        pi = *pi_;
+    }
+    if (!pi) return null;
```
tsbockman: Mixing `if` styles like that is awkward. How about:
```D
Interface* pi;
if (auto pi_ = *cast(Interface***)p)
    pi = *pi_;
if (!pi)
    return null;
```
**?**

etcimon: Sorry that was a copy paste. I use a mix of styles depending on the length of the line. Reasoning is: if it reads well in one line, why not? If not, might as well use the `{ }` as well to make it obvious it's not a one liner.

MartinNowak:
Please add a test case.

MartinNowak:
So you want to make casting of an interface that belongs to an already destroyed object more forgiving?
Not sure about that, you might not generally access other objects in finalizers b/c finalization is deliberately unordered.

etcimon:
My issue was when the `TLSClient` object is destroyed by the GC in certain circumstances. It has say, a `Unique!Hash` member object which may be an implementation of `SHA512` and allocated on the GC. So, once the GC collects `TLSClient`, the destructor of `Unique!Hash` will call `destroy` which fails in this cast. I still need to find a simple way to reproduce this..

MartinNowak:
You're not using that crappy Unique, are you?
In this case Unique is not the real owner of the object (b/c it's also allocated on the GC) and must not destroy it in a finalizer.
According to the docs [Unique.release](https://dlang.org/library/std/typecons/unique.release.html) is your friend.

etcimon:
Yes, a customization of unique in memutils. I use it on manually managed allocations, but high-level objects are allowed to be managed in the interface. For obvious reasons, I need determinate lifetimes for these high-level objects when working with TLS.

MartinNowak:
Have you been able to resolve your problem?

etcimon:
It would require a rewrite that moves everything out of the GC. I think it would be simpler if the cast forgave null pointers

MartinNowak:
It would silence a bug in your code and hide a race condition once we parallelize finalization.

etcimon:
In that case, the .destroy() call would need to be thread-safe to be compatible with GC parallelized collection shouldn't it?

etcimon:
An alternative fix for me is to check in `Unique.~this` whether `bool rt_isInGC()` returns `true` before calling destroy.

Here is the implementation in lifetime.d :

```D
bool g_in_gc;

extern (C) bool rt_isInGC() {
	return g_in_gc;
}
extern (C) void rt_finalizeFromGC(void* p, size_t size, uint attr)
{
	g_in_gc = true;
	scope(exit) g_in_gc = false;
```

etcimon:
This will be solved by using the gc_inFinalizers call from the GC SpinLock pull request

PR #1478, opened 2016-01-23
Optimize array cast for equivalence

etcimon:
In web servers and crypto, there is frequent back-and-forth casting between ubyte[] and string. This created a bottleneck where perf gave this function a 7% sampling prevalence. After this optimization, it went down <1%.

### DIFF
--- a/src/rt/arraycast.d
+++ b/src/rt/arraycast.d
@@ -25,6 +25,7 @@ extern (C)
 @trusted nothrow
 void[] _d_arraycast(size_t tsize, size_t fsize, void[] a)
 {
+    if (tsize == fsize) return a;
     auto length = a.length;
 
     auto nbytes = length * fsize;


yebblies:
Seems like we could do this on the compiler side.

MartinNowak:
Auto-merge toggled on

MartinNowak:
Yes, please make a ticket to do this on the call site.

MartinNowak:
Wait, in fact it's already done https://github.com/D-Programming-Language/dmd/blob/c6ac10aed0f7dc76112e576204e3a1e8ea307dbd/src/e2ir.c#L3962.
Can you please provide a test case for this @etcimon?

yebblies:
So it is!  This change should be 100% dead code.

MartinNowak:
Maybe it's not yet done for gdc or ldc, waiting for more input from @etcimon before reverting this.

etcimon:
I assumed this was the line that dropped the load on the server, but the commit actually removed a modulo comparison operation as well:
https://github.com/etcimon/druntime/commit/f938683cf83473f20f7fcd9e7a948e2f612b3b46#diff-7165ac077e578d70f8d6f9153b344bdaR32

I'll isolate this on the server today

etcimon:
It looks like the modulo might be responsible for this to be 35% slower, but it only slows down from 1.0% to 1.35% in the perf diagnostics because I delegated the monty reduce operations to openssl, so my bigint operations don't use the array cast. I could do an optimization in that library instead once I get around to optimizing it correctly, it was 10x slower than openssl in monty because of some inline issues I suppose. So, this can be reverted, sorry about that

etcimon:
bottlenecks-with-modulo

MartinNowak:
The modulo could be replaced with `length * tbytes != nbytes`, leaving only a single division.
And we could optimize the IR gen for common scale factors to use a cheap shift.

PR #1480, opened 2016-01-25
More core.bitop improvements

tsbockman:
* Fixed `bsf(ulong)` and `bsr(ulong)` for 32-bit CTFE - union repainting doesn't work in CTFE.
* Added `bswap(ulong)` - this should be made an intrinsic on 64-bit, but that's obviously outside the scope of this PR.
* Fixed CTFE `bitswap()` - `asm` doesn't work in CTFE.
* Consolidated both non-`asm` implementations of `bitswap()` into a single `private` `softBitswap()` template.
* `softBitswap()` now uses `bswap` for the final step, like the optimized assembly versions do.

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -3,7 +3,7 @@
  *
  * Copyright: Copyright Don Clugston 2005 - 2013.
  * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
- * Authors:   Don Clugston, Sean Kelly, Walter Bright, Alex Rønne Petersen
+ * Authors:   Don Clugston, Sean Kelly, Walter Bright, Alex Rønne Petersen, Thomas Stuart Bockman
  * Source:    $(DRUNTIMESRC core/_bitop.d)
  */
 
@@ -40,28 +40,26 @@ private union Split64
             uint lo;
         }
     }
-}
 
-pragma(inline, true)
-private auto split64(ulong u64) pure
-{
-    if(__ctfe)
+    pragma(inline, true)
+    this(ulong u64) @safe pure nothrow @nogc
     {
-        // union repainting doesn't work in CTFE
-        Split64 ret;
-        ret.lo = cast(uint)u64;
-        ret.hi = cast(uint)(u64 >>> 32);
-        return ret;
-    } else
-        return Split64(u64);
+        if (__ctfe)
+        {
+            lo = cast(uint) u64;
+            hi = cast(uint) (u64 >>> 32);
+        }
+        else
+            this.u64 = u64;
+    }
 }
 
 unittest
 {
-    const rt = split64(1);
+    const rt = Split64(1);
     assert((rt.lo == 1) && (rt.hi == 0));
 
-    enum ct = split64(1);
+    enum ct = Split64(1);
     assert((ct.lo == rt.lo) && (ct.hi == rt.hi));
 }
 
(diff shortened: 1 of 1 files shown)


tsbockman:
Minor update to better conform to the official style guide.

WalterBright:
Remember that the bitop functions need to be recognized by the compiler so that the custom bit instructions can be emitted. "Fixing" them may break that - you'll need to check that this is not the case.

tsbockman:
@WalterBright How should I do that? Was there any specific change in this PR that concerned you?

src/core/bitop.d
```diff
@@ -548,37 +667,11 @@ void volatileStore(ulong * ptr, ulong  value);   /// ditto
...
-        x = ((x >> 2) & 0x3333_3333) | ((x & 0x3333_3333) << 2);
-        // swap nibbles
-        x = ((x >> 4) & 0x0F0F_0F0F) | ((x & 0x0F0F_0F0F) << 4);
-        // swap bytes
-        x = ((x >> 8) & 0x00FF_00FF) | ((x & 0x00FF_00FF) << 8);
-        // swap 2-byte long pairs
-        x = ( x >> 16              ) | ( x               << 16);
-        return x;
-
-    }
```
MartinNowak: Did anyone ever run a microbenchmark for the asm code?
Current compilers seem to generate slightly better code https://goo.gl/VcBUWg.

src/core/bitop.d
```diff
@@ -216,6 +246,24 @@ int bts(size_t* p, size_t bitnum) pure @system;
...
+
+    const temp = sv.lo;
+    sv.lo = bswap(sv.hi);
+    sv.hi = bswap(temp);
+
+    if (__ctfe)
+        return (cast(ulong) sv.hi << 32) | sv.lo;
+    else
+        return sv.u64;
+}
```
MartinNowak: Just always use the CTFE version, going through a store and different load is difficult for compilers to figure out and might result in stack usage.

src/core/bitop.d
tsbockman: @MartinNowak I'm not an assembly expert, so someone else should file a separate pull request if the `asmBitswap` functions need an optimization pass.

I'm mostly just trying to standardize the `core.bitop` API so that all functions are available, regardless of the execution environment.

MartinNowak:
Auto-merge toggled on

tsbockman:
@MartinNowak 
> Auto-merge toggled on

Earlier @WalterBright [expressed concern](https://github.com/D-Programming-Language/druntime/pull/1480#issuecomment-174761367) that I might break the intrinsic detection with my changes. I [asked how to verify that it's still working](https://github.com/D-Programming-Language/druntime/pull/1480#issuecomment-174762148), and never got an answer.

Did you take care of this for me? Or did you just decide to ignore the issue and merge anyway?

MartinNowak:
> Earlier @WalterBright expressed concern that I might break the intrinsic detection with my changes. I asked how to verify that it's still working, and never got an answer.
> 
> Did you take care of this for me? Or did you just decide to ignore the issue and merge anyway?

Walter was either referring to [this](https://github.com/D-Programming-Language/dmd/blob/ce5374df58516af4d72f0c3e9fb2530e022c2e03/src/toir.c#L363) or [this](https://github.com/D-Programming-Language/druntime/blob/bc8b4cdf403b10c70364420b8150a7e858a0cbb4/src/core/bitop.d#L150). You're changes shouldn't have broken any of those detections.

PR #1481, opened 2016-01-26
There's no reason to hide what the makefile is doing

WalterBright:
When the makefile fails, like it does here:

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=1929728&isPull=true

it's really hard to figure out what was being executed when it went wrong when in quiet mode. I have no idea where this desire to hide what commands are being executed comes from.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -3,7 +3,7 @@
 #    pkg_add -r gmake
 # and then run as gmake rather than make.
 
-QUIET:=@
+QUIET:=
 
 include osmodel.mak
 


schveiguy:
I don't see any harm in this. I mostly ignore/can't read everything the build says anyway. LGTM.

MartinNowak:
Ideally it would be hidden until an error happens.
The nightly build logs are completely unreadable for example.
http://nightlies.dlang.org/dmd-nightly/build.html

Disabling QUIET for druntime doesn't seem to add much output, so it should be fine w/ @braddr who has to store all the log data.

MartinNowak:
Auto-merge toggled on

PR #1482, opened 2016-01-26
essential code should not be in assert's

WalterBright:
The call to `unloadLibrary()` should not be in an assert, because when compiled with `-release` it gets elided.

This appears to be blocking https://github.com/D-Programming-Language/dmd/pull/5359 so please pull.

### DIFF
--- a/test/shared/src/finalize.d
+++ b/test/shared/src/finalize.d
@@ -28,6 +28,7 @@ extern (C) alias SetFinalizeCounter = void function(shared(size_t*));
 
 void main(string[] args)
 {
+    printf("STARTING finalize\n");
     auto name = args[0];
     assert(name[$-9 .. $] == "/finalize");
     name = name[0 .. $-8] ~ "lib.so";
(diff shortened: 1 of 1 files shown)


JakobOvrum:
Why the move to `assert(0)`?

MartinNowak:
> The call to unloadLibrary() should not be in an assert, because when compiled with -release it gets elided.

It's a unittest, so the asserts should always be compiled, right?

WalterBright:
It gets compiled with -release, meaning the library does not get unloaded.

MartinNowak:
Auto-merge toggled on

MartinNowak:
Someone just recently told me asserts were run as soon as `-unittest` is passed. But I vaguely remember we have two assert variants for inside and outside of unittest blocks (and apparently only the former is run in release tests).

schveiguy:
It sure seems like we should be running all this code in a unittest instead of in a main function.

I agree with moving the unload library out of an assert. Having any action inside an assert is a no-no. But the others are just checks of values.

test/shared/src/finalize.d
```diff
@@ -47,8 +48,14 @@ void main(string[] args)
...
+    if (!r)
+        assert(0);
+    if (finalizeCounter != 4)
+        assert(0);
+    if (nf1._finalizeCounter)
+        assert(0);
+    if (nf2._finalizeCounter)
+        assert(0);
+    printf("PASS finalize\n");
 }
```
andralex: Nicer idiom: 
```d
Runtime.unloadLibrary(h) || assert(0);
finalizeCounter == 4 || assert(0);
```
etc.

rainers:
> Someone just recently told me asserts were run as soon as -unittest is passed. But I vaguely remember we have two assert variants for inside and outside of unittest blocks (and apparently only the former is run in release tests).

After being disabled with -release all asserts are reenabled with the -unittest switch: https://github.com/D-Programming-Language/dmd/blob/master/src/mars.d#L1087.
What's different in unittest blocks is that other functions are called instead of the usual assert: https://github.com/D-Programming-Language/dmd/blob/master/src/e2ir.c#L1904. Not sure what's the reasoning, though.

PR #1484, opened 2016-01-27
remove QUIET from Makefile

WalterBright:
The logs are so useless with QUIET on, might as well not even bother having logs.


MartinNowak:
There are 4 subfolders in test each having a similar makefile.

quickfur:
LGTM

MartinNowak:
Let's keep it, QUIET is configurable, defaults to non-quiet, noisy log output is a pain if you want to scroll back in your shell or have to store the files. What's important is to get good diagnostic output on failure.
Also your PR is incomplete.

PR #1485, opened 2016-01-27
More extensive ioctl support for OSX

sanjayss:
I followed the .h structure on OSX and split the ioctl functionality into ioccom.d (for some basic macros), ttycom.d (that separates out terminal functionality) and filio.d (which has some file related ioctl commands).

### DIFF
--- a/src/core/sys/posix/sys/ioctl.d
+++ b/src/core/sys/posix/sys/ioctl.d
@@ -323,13 +323,7 @@ else version (OSX)
     import core.sys.posix.termios; // termios
     import core.sys.posix.sys.time; // timeval
 
-    struct winsize
-    {
-        ushort ws_row;
-        ushort ws_col;
-        ushort ws_xpixel;
-        ushort ws_ypixel;
-    }
+    import core.sys.posix.sys.ttycom; // Terminal related ioctls
 
     struct ttysize
     {
@@ -338,6 +332,11 @@ else version (OSX)
         ushort ts_xxx;
         ushort ts_yyy;
     }
+
+    enum uint TIOCGSIZE = TIOCGWINSZ;
+    enum uint TIOCSSIZE = TIOCSWINSZ;
+
+    import core.sys.posix.sys.filio; // File related ioctls
 
     int ioctl(int fildes, c_ulong request, ...);
 }
--- a/src/core/sys/posix/sys/ioctl.d
+++ b/src/core/sys/posix/sys/ioctl.d
@@ -323,7 +323,7 @@ else version (OSX)
     import core.sys.posix.termios; // termios
     import core.sys.posix.sys.time; // timeval
 
-    import core.sys.posix.sys.ttycom; // Terminal related ioctls
+    public import core.sys.posix.sys.ttycom; // Terminal related ioctls
 
     struct ttysize
     {
@@ -336,7 +336,7 @@ else version (OSX)
     enum uint TIOCGSIZE = TIOCGWINSZ;
     enum uint TIOCSSIZE = TIOCSWINSZ;
 
-    import core.sys.posix.sys.filio; // File related ioctls
+    public import core.sys.posix.sys.filio; // File related ioctls
 
     int ioctl(int fildes, c_ulong request, ...);
 }
--- a/mak/COPY
+++ b/mak/COPY
@@ -122,6 +122,8 @@ COPY=\
 	$(IMPDIR)\core\sys\posix\netinet\in_.d \
 	$(IMPDIR)\core\sys\posix\netinet\tcp.d \
 	\
+	$(IMPDIR)\core\sys\posix\sys\filio.d \
+	$(IMPDIR)\core\sys\posix\sys\ioccom.d \
 	$(IMPDIR)\core\sys\posix\sys\ioctl.d \
 	$(IMPDIR)\core\sys\posix\sys\ipc.d \
 	$(IMPDIR)\core\sys\posix\sys\mman.d \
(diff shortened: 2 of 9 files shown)


src/core/sys/posix/sys/ioctl.d
```diff
@@ -323,13 +323,7 @@ else version (OSX)
     import core.sys.posix.termios; // termios
     import core.sys.posix.sys.time; // timeval
 
-    struct winsize
```
adamdruppe: Where is winsize defined now for non-OSX posixes?

src/core/sys/posix/sys/ioctl.d
```diff
@@ -323,13 +323,7 @@ else version (OSX)
...
-    {
-        ushort ws_row;
-        ushort ws_col;
-        ushort ws_xpixel;
-        ushort ws_ypixel;
-    }
+    import core.sys.posix.sys.ttycom; // Terminal related ioctls
```
adamdruppe: This might want to be a public import. In C, would you just write `#include<sys/ioctl.h>` or would you explicitly do `#include<sys/ttycom.h>`? If it is the former, you'd need to public import in D so the file is exposed here too.

src/core/sys/posix/sys/ioctl.d
```diff
@@ -339,6 +333,11 @@ else version (OSX)
         ushort ts_yyy;
     }
 
+    enum uint TIOCGSIZE = TIOCGWINSZ;
+    enum uint TIOCSSIZE = TIOCSWINSZ;
+
+    import core.sys.posix.sys.filio; // File related ioctls
```
adamdruppe: same on the import vs public impor as above.

src/core/sys/posix/sys/ioctl.d
sanjayss: Not sure what you mean -- I only moved the 'struct winsize' from the version (OSX) block into ttycom.d. This is to reflect how it is placed in the C header files on OSX so someone moving to D from C on OSX knows where to look for this based on C documentation.

adamdruppe: oh, I see what you mean, I didn't realize the context of the file in the diff. cool.

src/core/sys/posix/sys/ioctl.d
sanjayss: On OSX, you would include both ttycom.h and ioctl.h -- I think the separation originated from the fact that for kernel development you would use only ttycom.h, but for userland development you need the ioctl() prototype from ioctl.h.

Like I mentioned in the other comment, I kept the structuring of this to be the same as the .h file structuring so someone reading C documentation on OSX (it's different on Linux though!) knows where to find them.

src/core/sys/posix/sys/ioctl.d
sanjayss: Same reason as above (following OSX .h structure). People include "filio.h" if they need the file specific IOCTL defines. So in D, the would need to include the filio.d module and the ioctl.d module.

I am new to D, so maybe I am missing something related to modules that you are pointing out -- would it be problematic for people to include both, say, filio.d and ioctl.d if they want access to the file related ioctls?

adamdruppe: On Thu, Jan 28, 2016 at 08:44:12PM -0800, sanjayss wrote:
> I am new to D, so maybe I am missing something related to modules that you are pointing out -- would it be problematic for people to include both, say, filio.d and ioctl.d if they want access to the file related ioctls?

No, it is ok to import both, just in C any #includes inside an included file are available without needing to be included again, whereas on D, you only get that behavior on public imports... so just want to make sure you're aware of that and you are doing what you want it to do.

sanjayss: Yes, that is what I intended -- that anyone wanting say the 'struct winsize' to do an ioctl would include both ttycom.d and ioctl.d mimicing what they would do in C on OSX (and think all the BSDs).

src/core/sys/posix/sys/ioctl.d
sanjayss: I will move winsize back into ioctl.d. Even though I decided to follow the OSX header file model (and I will for the remaining things if there are no objections), I realized there may be existing D programs that include only ioctl.d and would have expected winsize to be present in ioctl.d itself (i.e for backwards compatibility).

sanjayss: Instead of getting winsize back into ioctl.d, I reverted to our original suggestion and made the ttycom and filio imports public. This way, now people can simply include ioctl module and don't need to include the additional ttycom and filio modules. This avoids having to include ioctl in ttycom module, but also keeps the separation of definitions like in the C headers.

adamdruppe: On Sun, Jan 31, 2016 at 07:39:35PM -0800, sanjayss wrote:
> Instead of getting winsize back into ioctl.d, I reverted to our original suggestion and made the ttycom and filio imports public. This way, now people can simply include ioctl module and don't need to include the additional ttycom and filio modules. This avoids having to include ioctl in ttycom module, but also keeps the separation of definitions like in the C headers.

Cool, I think that will please everyone.

DmitryOlshansky:
What's the state of this? @adamdruppe

sanjayss:
If the question was directed to me -- there is nothing more to add or
change and the pull request is in it's final state unless someone points
out any issues.

DmitryOlshansky:
Auto-merge toggled on

PR #1486, opened 2016-01-28
Do not recursively call itself from `bsf`/`bsr` bodies if pointers are 64 bits.

denis-sh:
It didn't cause tester failure because dmd uses intrinsic even if there is a function implementation.

This is a fixup for pull #1452.

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -72,20 +72,18 @@ unittest
  */
 int bsf(size_t v) pure;
 
+version(CoreDdoc)
+{
 /// ditto
+int bsf(ulong v) pure;
+}
+else version(D_LP64) { } else
 int bsf(ulong v) pure
 {
-    static if (size_t.sizeof == ulong.sizeof)
-        return bsf(cast(size_t) v);
-    else static if (size_t.sizeof == uint.sizeof)
-    {
-        const sv = Split64(v);
-        return (sv.lo == 0)?
-            bsf(sv.hi) + 32 :
-            bsf(sv.lo);
-    }
-    else
-        static assert(false);
+    const sv = Split64(v);
+    return (sv.lo == 0)?
+        bsf(sv.hi) + 32 :
+        bsf(sv.lo);
 }
 
 ///
@@ -112,20 +110,18 @@ unittest
  */
 int bsr(size_t v) pure;
 
+version(CoreDdoc)
+{
 /// ditto
+int bsr(ulong v) pure;
+}
+else version(D_LP64) { } else
 int bsr(ulong v) pure
 {
-    static if (size_t.sizeof == ulong.sizeof)
-        return bsr(cast(size_t) v);
-    else static if (size_t.sizeof == uint.sizeof)
-    {
-        const sv = Split64(v);
-        return (sv.hi == 0)?
-            bsr(sv.lo) :
-            bsr(sv.hi) + 32;
-    }
-    else
-        static assert(false);
+    const sv = Split64(v);
+    return (sv.hi == 0)?
+        bsr(sv.lo) :
+        bsr(sv.hi) + 32;
 }
 
 ///
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -75,7 +75,8 @@ int bsf(size_t v) pure;
 version(CoreDdoc)
 {
 /// ditto
-int bsf(ulong v) pure;
+int bsf(ulong v) pure
+{ assert(0); } // Workaround dmd @@@BUG15818@@@
 }
 else version(D_LP64) { } else
 int bsf(ulong v) pure
@@ -113,7 +114,8 @@ int bsr(size_t v) pure;
 version(CoreDdoc)
 {
 /// ditto
-int bsr(ulong v) pure;
+int bsr(ulong v) pure
+{ assert(0); } // Workaround dmd @@@BUG15818@@@
 }
 else version(D_LP64) { } else
 int bsr(ulong v) pure


src/core/bitop.d
```diff
@@ -72,20 +72,17 @@ unittest
  */
 int bsf(size_t v) pure;
 
+version(CoreDdoc)
 /// ditto
+int bsf(ulong v) pure;
+
+version(D_LP64) { } else
```
tsbockman: `static if (size_t.sizeof != ulong.sizeof)` seems more appropriate to me, since the point is, literally, to detect when the above `int bsf(size_t v)` is not equivalent to `int bsf(ulong v)`.

Also, `static if` is used for this purpose everywhere else in this file; this is a gratuitous inconsistency.

denis-sh: `size_t.sizeof` is just a consequence of `D_LP64` version. For me it's a bad style to use such huge expressions like `static if (size_t.sizeof != ulong.sizeof)` and also create a third `else static assert(0);` case. I just don't want to refactor all the module.

tsbockman: Omit the `static assert (0);` if you want.

(Although, it's not just fluff - it is documenting the fact that this code depends upon the precise bit-ness of the types involved, and will probably break if the assumption that `size_t` is either 32- or 64-bit is ever broken on some future architecture.)

But, trying to change the style of a module by updating some random small piece of it just increases the chaos. I was already [asked not to do that](https://github.com/D-Programming-Language/druntime/pull/1452#issuecomment-170805228) on the very PR you're fixing. (Admittedly, the context was somewhat different.)

denis-sh: > it is documenting the fact that this code depends upon the precise bit-ness of the types involved

And this is what `D_LP64` for. If someone don't know about `D_LP64` it's not our problem. It's silly to not use language constructs just because someone can be uninformed about it.

> But, trying to change the style of a module by updating some random small piece of it just increases the chaos.

Yes, but I will not use the style I'm against of.

tsbockman: > It's silly to not use language constructs just because someone can be uninformed about it.

I already explained why I used `static if`. Also, some of the similar uses of `static if` are from **Walter Bright** himself.

Just because *you* wouldn't do it the way I did, doesn't mean I'm "uninformed".

> Yes, but I will not use the style I'm against of.

Then fix the whole file, if you think it's that important.

denis-sh: > Then fix the whole file, if you think it's that important.

I'll prefer to wait for druntime devs to chain in. It's also possible they request to change my style to `static if` as they like it too just like you and I'm the minority.

tsbockman: > And this is what `D_LP64` for. If someone don't know about `D_LP64` it's not our problem.

Looking at this again, I see that you actually missed my point about the "precise bit-ness of the types".

Leaving aside the issue of `static if` verses `version`, the `else static assert(0);` is there because the code which you have marked as "anything but 64-bit" is really intended *specifically* for 32-bit.

So, instead of `version (D_LP64) { } else`, it should really just be `version (D_LP32)` (or the equivalent `static if`).

MartinNowak:
Both would work. I'd prefer version on a module level b/c it doesn't require semantic analysis or interpretation (which isn't a problem for the size_t check but can cause forward reference issues for more complex stuff).

src/core/bitop.d
```diff
@@ -72,20 +72,17 @@ unittest
  */
 int bsf(size_t v) pure;
 
+version(CoreDdoc)
 /// ditto
+int bsf(ulong v) pure;
```
MartinNowak: Look at the doc build failure, you need this.
```d
version (CoreDDoc)
   decl;
else version (D_LP64) {}
else
   impl;

**Comment by denis-sh** [comment]
Fixed DDoc failure.

File: src/core/bitop.d [hunk 3]
```diff
@@ -72,20 +72,18 @@ unittest
  */
 int bsf(size_t v) pure;
 
+version(CoreDdoc)
+{
 /// ditto
+int bsf(ulong v) pure;
+}
+else version(D_LP64) { } else
```
tsbockman: Please just make this `else version(D_LP32)`. My implementation is not good for anything other than 32-bit.

denis-sh: There is no `D_LP32` standard version. `size_t` is 32 bit unless `D_LP64` is set.

tsbockman: Well, I still think this should be `static if(size_t.sizeof == 4)` then, because your `version` constraint actually has a different meaning than the one it replaces.

But I will not argue the point further if you insist. (That'll teach me not to make mistakes in future PRs though, I guess...)

denis-sh: > There is no `D_LP32` standard version. `size_t` is 32 bit unless `D_LP64` is set.

So `static if(size_t.sizeof == 4)` is equal to `version(D_LP64) { } else`. This is because `size_t` can be either 32 bit or 64 bit and there is no plans to support anything else (and if there were such plans something like `D_LP32` version would be added at first).

tsbockman: > This is because size_t can be either 32 bit or 64 bit and there is no plans to support anything else

Plans change.

> (and if there were such plans something like D_LP32 version would be added at first).

To the standard `version` specifiers, but not to this code. Everyone will have forgotten about this conversation by then.

tsbockman:
Inspired by our discussion: [DMD PR #5377](https://github.com/D-Programming-Language/dmd/pull/5377) and [dlang.org PR #1220](https://github.com/D-Programming-Language/dlang.org/pull/1220).

tsbockman:
Another question that occurred to me while looking into the `version` issue - what is `size_t` on `version(D_X32)`?

I expect that `size_t.sizeof` == `uint.sizeof`, but that would mean that all the `core.bitop` intrinsics that take `size_t` need to be changed to take `ulong` on `version(D_X32)`, since registers are still 64-bit.

ibuclaw:
All compilers that support X32 (one) overrides uses of all core.bitop functions and 'just does the right thing depending on input size'

ibuclaw:
I can't see why there aren't two distinct overloads. One for uint and the other for ulong.   If the compiler can treat either as a builtin then great!  Otherwise have soft implementations for both.

ibuclaw:
Then again, I don't treat any of these functions as functions for assembler opcodes, so my opinion is biased.

MartinNowak:
Ping, the docs are still broken.

denis-sh:
> Ping, the docs are still broken.

Added dmd [bug 15818](https://issues.dlang.org/show_bug.cgi?id=15818) workaround.

MartinNowak:
Isn't the problem overloading `size_t` against `ulong` on 64-bit?

denis-sh:
> Isn't the problem overloading size_t against ulong on 64-bit?

The problem is documenting both `size_t` and `ulong` functions. As `size_t` is documented as `size_t`, not as `ulong`, `ulong` declaration should also be documented. Also there is no overloading on 64-bit, just two equal declarations.

tsbockman:
> The problem is documenting both `size_t` and `ulong` functions.

Will the compiler still recognize the intrinsics, if you just drop the use of `size_t` in favor of separate `uint` and `ulong` overloads, like ibuclaw suggested earlier?

(The body of the `uint` version would just forward to the `ulong` one on 64-bit.)

MartinNowak:
Auto-merge toggled on

MartinNowak:
> Will the compiler still recognize the intrinsics, if you just drop the use of size_t in favor of separate uint and ulong overloads, like ibuclaw suggested earlier?

Yes it will

tsbockman:
@MartinNowak 
> Yes it will.

That sounds like a much cleaner solution to me, then.

(Although I'm still OK with merging this as is, if everyone would rather just move on.)

MartinNowak:
> Will the compiler still recognize the intrinsics, if you just drop the use of size_t in favor of separate uint and ulong overloads, like ibuclaw suggested earlier?

Yes it will, that would be the cleaner solution.

tsbockman:
So basically, this:
```D
/**
 * Scans the bits in v starting with bit 0, looking
 * for the first set bit.
 * Returns:
 *      The bit number of the first bit set.
 *      The return value is undefined if v is zero.
 */
int bsf(uint v) pure
{
    version(D_LP64)
        bsf(cast(ulong)v);
    else
    {
		for(int b = 0; b < 32; ++b)
		{
			if(v & (1u << b))
				return b;
		}
		
		return -1;
    }
}
/// ditto
int bsf(ulong v) pure
{
    const sv = Split64(v);
    return (sv.lo == 0)?
        bsf(sv.hi) + 32 :
        bsf(sv.lo);
}
```

```D
/**
 * Scans the bits in v from the most significant bit
 * to the least significant bit, looking
 * for the first set bit.
 * Returns:
 *      The bit number of the first bit set.
 *      The return value is undefined if v is zero.
 */
int bsr(uint v) pure
{
	version(D_LP64)
		bsr(cast(ulong)v);
	else
	{
		for(int b = 31; b >= 0; --b)
		{
			if(v & (1u << b))
				return b;
		}
		
		return -1;
	}
}
/// ditto
int bsr(ulong v) pure
{
    const sv = Split64(v);
    return (sv.hi == 0)?
        bsr(sv.lo) :
        bsr(sv.hi) + 32;
}
```
But with a more efficient fall-back implementation for `uint` than I wrote just now?

Should I just submit a new pull request, or does @denis-sh want to update this one?

denis-sh:
> Yes it will, that would be the cleaner solution.

IMO, no.

> Should I just submit a new pull request, or does @denis-sh want to update this one?

Submit a new pull request, I'm against this solution as for me it is clearly useless. There is no reason for `uint` overload in 64 bit mode as it gives us nothing and may waste performance on cropping its argument.

tsbockman:
> There is no reason for `uint` overload in 64 bit mode as it gives us nothing and may waste performance on cropping its argument.

Why would it have to "crop its argument", given that the `uint` overload would only be selected when the input is already 32-bit?

denis-sh:
> Why would it have to "crop its argument", given that the `uint` overload would only be selected when the input is already 32-bit?

I was thinking about some imaginary `uint` operation sequence which can be optimized as `ulong` operation sequence on 64 bit machines. I'm not sure it's really possible, but also I can't proof it is impossible.

tsbockman:
That seems like a weak reason to reject the more straightforward solution that the rest of us favour. I really doubt that would be a problem, as long as the `uint` overload inlines properly - which I can verify with `pragma(inline, true)`.

tsbockman:
I have submitted an [alternative pull request](https://github.com/D-Programming-Language/druntime/pull/1517).

When I thought about it more, I realized that it makes no sense to fix the recursive call issue - which can cause practical problems only if the intrinsic is not available - without also providing a working non-intrinsic code path to follow. So, I added a proper software fallback implementation of the bit scan algorithm, too.

denis-sh:
> That seems like a weak reason to reject the more straightforward solution that the rest of us favour.

Yes, lets add needless `uint` overload which has no reason to exist. I probably need to visit some psychiatrist as I definitely loose connection with reality as I still sure there must be a reason for function to exist in a library.

PR #1489, opened 2016-01-30
Fix issue 15628: exception handling in fibers with /SAFESEH

rainers:
/SAFESEH is not supported by DMD, but with LDC.

https://issues.dlang.org/show_bug.cgi?id=15628


rainers:
This might also fix https://issues.dlang.org/show_bug.cgi?id=13821

PR #1490, opened 2016-01-31
Android update

joakim-noah:
Pull in some necessary declarations from bionic, exclude some `CRuntime_Glibc` definitions, add ARM support extracted from my work with ldc, and replace all uses of `Android` with `CRuntime_Bionic` or `linux`.

### DIFF
--- a/src/core/stdc/math.d
+++ b/src/core/stdc/math.d
@@ -1967,7 +1967,8 @@ else version(CRuntime_Bionic)
     double  ldexp(double x, int exp);
     ///
     float   ldexpf(float x, int exp);
-    // alias for double: real    ldexpl(real x, int exp);
+    ///
+    real    ldexpl(real x, int exp);
 
     ///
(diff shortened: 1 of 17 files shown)


joakim-noah:
Ping, no problems with the auto-tester for three weeks.  I think this and #1384, which has been sitting for more than five months, can be merged.

joakim-noah:
I'd like to get [this and a couple other PRs merged into the ldc branches](https://github.com/ldc-developers/ldc/issues/1264#issuecomment-183027405).  Should I do that or would you prefer to do it?

joakim-noah:
Regarding function implementations, I guess you're talking about `__tls_get_addr`?  The two function prototypes are slightly different, because `X86` uses an extra underscore and I think the dmd patch dereferences the passed-in pointer, so only the last three lines are the same.  Rather than extract those out, I just cut-n-pasted it.

PR #1491, opened 2016-02-01
replace doLock/doUnlock with runLocked

MartinNowak:
- pass function via alias parameter to allow inlining

### DIFF
--- a/src/gc/gc.d
+++ b/src/gc/gc.d
@@ -345,40 +345,29 @@ struct GC
         gcx.disabled++;
     }
 
-
-    mixin template doLock()
+    auto runLocked(alias func, alias time, alias count, Args...)(auto ref Args args)
     {
-        debug(PROFILE_API)
-        {
-            long tm = (GC.config.profile > 1 ? currTime.ticks : 0);
-        }
+        debug(PROFILE_API) {} else pragma(inline, true);
+        debug(PROFILE_API) immutable tm = (GC.config.profile > 1 ? currTime.ticks : 0);
+        gcLock.lock();
+        debug(PROFILE_API) immutable tm2 = (GC.config.profile > 1 ? currTime.ticks : 0);
 
-        bool locked = (gcLock.lock(), true);
-
-        debug(PROFILE_API)
-        {
-            long tm2 = (GC.config.profile > 1 ? currTime.ticks : 0);
-        }
-    }
-
-    mixin template doUnlock(alias time, alias count)
-    {
-        debug(PROFILE_API)
-        {
-            bool unlocked = () {
-                if (GC.config.profile > 1)
-                {
-                    count++;
-                    long now = currTicks;
-                    lockTime += tm2 - tm;
-                    time += now - tm2;
-                }
-                gcLock.unlock();
-                return true;
-            }();
-        }
+        static if (is(typeof(func(args)) == void))
+            func(args);
         else
-            bool unlocked = (gcLock.unlock(), true);
+            auto res = func(args);
+
+        debug(PROFILE_API) if (GC.config.profile > 1)
+        {
+            count++;
+            immutable now = currTime.ticks;
+            lockTime += tm2 - tm;
+            time += now - tm2;
+        }
+        gcLock.unlock();
+
+        static if (!is(typeof(func(args)) == void))
+            return res;
     }
 
     /**
@@ -391,7 +380,7 @@ struct GC
             return 0;
         }
 
-        uint go() nothrow
+        static uint go(Gcx* gcx, void* p) nothrow
         {
             Pool* pool = gcx.findPool(p);
             uint  oldb = 0;
@@ -406,10 +395,7 @@ struct GC
             return oldb;
         }
 
-        mixin doLock!();
-        uint rc = go();
-        mixin doUnlock!(otherTime, numOthers);
-        return rc;
+        return runLocked!(go, otherTime, numOthers)(gcx, p);
     }
 
 
@@ -423,7 +409,7 @@ struct GC
             return 0;
         }
 
-        uint go() nothrow
+        static uint go(Gcx* gcx, void* p, uint mask) nothrow
         {
             Pool* pool = gcx.findPool(p);
             uint  oldb = 0;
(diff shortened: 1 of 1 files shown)


rainers:
I tried the benchmarks for win32 and didn't notice any performance regressions or improvements, so I'm fine with the refactoring.
Unfortunately compiling with -debug=PROFILE_API results in
```
..\src\gc\gc.d(349): Error: function gc.gc.GC.runLocked!(mallocNoSync, mallocTime, numMallocs, uint, uint, uint, const(TypeInfo)).runLocked cannot inline function
..\src\gc\gc.d(349): Error: function gc.gc.GC.runLocked!(sizeOfNoSync, otherTime, numOthers, void*).runLocked cannot inline function
..\src\gc\gc.d(349): Error: function gc.gc.GC.runLocked!(queryNoSync, otherTime, numOthers, void*).runLocked cannot inline function
..\src\gc\gc.d(349): Error: function gc.gc.GC.runLocked!(go, otherTime, numOthers, Gcx*, void*).runLocked cannot inline function
..\src\gc\gc.d(349): Error: function gc.gc.GC.runLocked!(go, otherTime, numOthers, Gcx*, void*).runLocked cannot inline function
..\src\gc\gc.d(349): Error: function gc.gc.GC.runLocked!(go, otherTime, numOthers, Gcx*, void*, uint, const(TypeInfo)).runLocked cannot inline function
..\src\gc\gc.d(349): Error: function gc.gc.GC.runLocked!(go, otherTime, numOthers, Gcx*, void*).runLocked cannot inline function
```

BTW: With the current version I could not measure any difference anymore when compiling with PROFILE_API (but not using it). Maybe we can always enable it into the build.

MartinNowak:
Stupid inliner, will try to resolve this.

We can reduce it to a single `GC.config.profile > 1` branch, so to use only a single branch predictor slot, then it should be fine to always enable the profile API.

ibuclaw:
People should stop marking everything with pragma inline.

MartinNowak:
> People should stop marking everything with pragma inline.

Not at all, we have measurable problem (a few %) if this isn't inlined.

ibuclaw:
> Not at all, we have measurable problem (a few %) if this isn't inlined

As I understood it, marking things as inline changes nothing.  Your only turning non-inlineable code into an error.

MartinNowak:
> As I understood it, marking things as inline changes nothing.

At some point it's supposed to become an inline hint.

> Your only turning non-inlineable code into an error.

I explicitly added it b/c it is a performance regression to change the function in a way that it becomes non-inlineable.

MartinNowak:
Fixed for `debug(PROFILE_API)`, dmd isn't capable of inlining the code, so I separated the function.
Until the inlining issues isn't resolved we can't enable PROFILE_API by default.

yebblies:
Why don't you just disable the `pragma(inline)` when PROFILE_API is on?

rainers:
> Why don't you just disable the pragma(inline) when PROFILE_API is on?

`pragma(inline)` only accepts tokens `true` and `false` as argument, not a compile time expression, making it very clumsy to use conditionally. But IIRC it has the rather unexpected version where you can put the pragma into the function body.

yebblies:
Yes, using it as a statement should work.

MartinNowak:
updated

rainers:
Thanks.

rainers:
Auto-merge toggled on

9rnsr:
This is now blocking my inlining improvement https://github.com/D-Programming-Language/dmd/pull/5292. With dmd git master, the instantiation `runLocked!(reallocNoSync, ...)` at src/gc/gc.d L552 is not inlined because `reallocNoSync` returns void and is called directly on `ExpStatement` at L356. By the dmd PR 5292, the call will also be tried to inline forcibly, but `reallocNoSync` is too big to expand its body, then it fails.

In my humble opinion: druntime is built with `-inline` by default. So the new `runLocked` API is *needlessly* limited by the compiler's inliner capability due to the use of `pragma(inline, true)`. Removing it would have no problem.

MartinNowak:
I don't care too much about reallocNoSync, but runLocked itself.

rainers:
> With dmd git master, the instantiation runLocked!(reallocNoSync, ...) at src/gc/gc.d L552 is not inlined because reallocNoSync returns void and is called directly on ExpStatement at L356. By the dmd PR 5292, the call will also be tried to inline forcibly, but reallocNoSync is too big to expand its body, then it fails.

Why would it forcibly inline reallocNoSync? This is not the sematics of the pragma(inline) inside the function body (albeit a bad aberration from other attributes).

> In my humble opinion: druntime is built with -inline by default. So the new runLocked API is needlessly limited by the compiler's inliner capability due to the use of pragma(inline, true). Removing it would have no problem.

If runLocked cannot be inlined we have seen a performance regressions in the past, with the druntime benchmarks running a few percents slower. The most problematic case is the `malloc`, though, so we wouldn't care too much for all the other calls.

9rnsr:
> Why would it forcibly inline reallocNoSync? This is not the sematics of the pragma(inline) inside the function body (albeit a bad aberration from other attributes).

In short, it's current behavior. (Note that, today `pragma(inline, true) auto func() { ... }` and `auto func() { pragma(inline, true); ... }` are exactly same behavior.)

> If runLocked cannot be inlined we have seen a performance regressions in the past, 

The code performance with `-inline` is only limited by the compiler's inlining capability. `pragma(inline, true)` would introduce no additinal performance. See also: https://github.com/D-Programming-Language/druntime/pull/1496#issuecomment-181650286

MartinNowak:
But it fails to compile when a function (runLocked) cannot be inlined, which is what we want here to prevent someone from introducing a performance regression.

MartinNowak:
Maybe we should leave that few percent to more serious backends?

9rnsr:
> But it fails to compile when a function (runLocked) cannot be inlined, which is what we want here to prevent someone from introducing a performance regression.

But current dmd inliner has known problems. By using pragma(inline, true), druntime is now relying on its bug-prone behavior. And sticking on current performance would prevent further inliner improvements. Actually my https://github.com/D-Programming-Language/dmd/pull/5292 is now dying.

PR #1495, opened 2016-02-06
core.bitops: Add bitwise rotate functions

CounterPillow:
Adds functions for a bitwise rotate left and rotate right operation. This operation is common enough that some architectures such as x86 have it as its own instruction; while many compilers will detect the used pattern and optimise it down to one instruction, some compilers (e.g. DMD) won't do so, in which case having this as a part of druntime allows for one hand-written assembly implementation later on.

The choice of `uint` as the type for the number of shifts is a bit arbitrary; any unsigned integer type would do, as the number of places to shift is usually fairly small (and wraps at `T.sizeof` anyway).

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -708,36 +708,53 @@ version (D_InlineAsm_X86_64)
 }
 
 /**
- *  Bitwise rotates unsigned arithmetic types
+ *  Bitwise rotates unsigned integer types
  */
-pure T rotl(T)(in T u, in uint s)
-    if (__traits(isArithmetic, T) && __traits(isUnsigned, T))
+pure T rol(T)(in T value, in uint count)
+    if (__traits(isIntegral, T) && __traits(isUnsigned, T))
 {
-    return cast(T) ((u << s) | (u >> (-s & (T.sizeof * 8 - 1))));
+    assert(count < 8 * T.sizeof);
+    return cast(T) ((value << count) | (value >> (-count & (T.sizeof * 8 - 1))));
+}
+/// ditto
+pure T ror(T)(in T value, in uint count)
+    if (__traits(isIntegral, T) && __traits(isUnsigned, T))
+{
+    assert(count < 8 * T.sizeof);
+    return cast(T) ((value >> count) | (value << (-count & (T.sizeof * 8 - 1))));
+}
+/// ditto
+pure T rol(uint count, T)(in T value)
+    if (__traits(isIntegral, T) && __traits(isUnsigned, T))
+{
+    static assert(count < 8 * T.sizeof);
+    return cast(T) ((value << count) | (value >> (-count & (T.sizeof * 8 - 1))));
+}
+/// ditto
+pure T ror(uint count, T)(in T value)
+    if (__traits(isIntegral, T) && __traits(isUnsigned, T))
+{
+    static assert(count < 8 * T.sizeof);
+    return cast(T) ((value >> count) | (value << (-count & (T.sizeof * 8 - 1))));
 }
 
-pure T rotr(T)(in T u, in uint s)
-    if (__traits(isArithmetic, T) && __traits(isUnsigned, T))
-{
-    return cast(T) ((u >> s) | (u << (-s & (T.sizeof * 8 - 1))));
-}
-
+///
 unittest
 {
     ubyte a = 0b10101010U;
     ulong b = ulong.max;
 
-    assert(rotl(a, 1) == 0b01010101);
-    assert(rotr(a, 1) == 0b01010101);
-    assert(rotl(a, 3) == 0b01010101);
-    assert(rotr(a, 3) == 0b01010101);
+    assert(rol(a, 1) == 0b01010101);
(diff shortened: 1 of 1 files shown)


ibuclaw:
Yep, I can train gdc to recognise and rewrite this as an lrotate or rrotate expression. :+1:

CounterPillow:
Any opinions on `isArithmetic` being used instead of `isIntegral`? I don't know whether there are valid use cases for bit rotating a floating point value (I'm not too familiar with bit twiddling hacks). Is there any downside to allowing floating point values to be bit rotated?

Also, can anybody think of more edge cases to add to the unit tests? Especially the `cast(T)` is making me nervous (and it was required to get types shorter than `int` working).

src/core/bitop.d
```diff
@@ -706,3 +706,38 @@ version (D_InlineAsm_X86_64)
...
+/**
+ *  Bitwise rotates unsigned arithmetic types
+ */
+pure T rotl(T)(in T u, in uint s)
+    if (__traits(isArithmetic, T) && __traits(isUnsigned, T))
+{
+    return cast(T) ((u << s) | (u >> (-s & (T.sizeof * 8 - 1))));
+}
+
+pure T rotr(T)(in T u, in uint s)
```
MartinNowak: Use `/// ditto` if both functions should share the same documentation.

src/core/bitop.d
```diff
@@ -706,3 +706,38 @@ version (D_InlineAsm_X86_64)
...
+    if (__traits(isArithmetic, T) && __traits(isUnsigned, T))
+{
+    return cast(T) ((u >> s) | (u << (-s & (T.sizeof * 8 - 1))));
+}
+
+unittest
```
MartinNowak: Make this a [documented unittest](http://dlang.org/spec/unittest.html#documented-unittests).

MartinNowak:
The instructions are called `ror` and `rol` by Intel, and `ror` by ARM, let's stick with that.

src/core/bitop.d
```diff
@@ -706,3 +706,38 @@ version (D_InlineAsm_X86_64)
...
     }
 }
+
+/**
+ *  Bitwise rotates unsigned arithmetic types
+ */
+pure T rotl(T)(in T u, in uint s)
+    if (__traits(isArithmetic, T) && __traits(isUnsigned, T))
+{
+    return cast(T) ((u << s) | (u >> (-s & (T.sizeof * 8 - 1))));
```
MartinNowak: What happens when `s` is bigger than `8 * T.sizeof`? If that behavior is inconsistent across architectures, an `assert(s < 8 * T.sizeof)` would be adequate.

src/core/bitop.d
```diff
@@ -706,3 +706,38 @@ version (D_InlineAsm_X86_64)
         }
     }
 }
+
+/**
+ *  Bitwise rotates unsigned arithmetic types
+ */
+pure T rotl(T)(in T u, in uint s)
```
MartinNowak: I'd prefer `n` or `count` instead of `s`, and `val` or `value` instead of `u`.
Any particular reason for naming them `u` and `s`?

MartinNowak:
There should also be an overload that takes the shift number as template parameter, so more efficient instructions w/ immediate argument can be used.

src/core/bitop.d
```diff
@@ -706,3 +706,38 @@ version (D_InlineAsm_X86_64)
...
+    if (__traits(isArithmetic, T) && __traits(isUnsigned, T))
```
MartinNowak: Bit operations on floats are only possible after reinterpreting the representation to an integer, so please `isIntegral` here.

MartinNowak: There are no unsigned floats btw.

ibuclaw: Wouldn't rotate only work for integers?

CounterPillow:
Pushed a new commit to address the raised concerns. I'll squash them when it's ready to merge. Time for round 2 of suggestions.

Does anybody happen to know whether it's allowed to use std.exception in druntime unittests to get `assertThrown`? That way I could include a couple more unittests.

MartinNowak:
> Does anybody happen to know whether it's allowed to use std.exception in druntime unittests to get assertThrown? That way I could include a couple more unittests.

No, it's not possible to use phobos in druntime.

MartinNowak:
I've slightly tweaked the documentation, see af2fa26b1434ae82c617ea554bcc7bb5c49a32bb.

src/core/bitop.d
```diff
@@ -706,3 +706,55 @@ version (D_InlineAsm_X86_64)
         }
     }
 }
+
+/**
+ *  Bitwise rotates unsigned integer types
```
MartinNowak: It's better to use `rotate` here, b/c the function doesn't do anything on its own account.

CounterPillow:
I see, you merged it right away without squashing the commits, though I guess the commit history isn't too bad with just two commits.

Also, I forgot to add my name to the authors list on the top, though I'm not sure if there needs to be some minimum required amount of contribution to the module before that's put there.

PR #1496, opened 2016-02-08
Remove pragma(inline, true) in runLocked function template.

9rnsr:
druntime is compiled with -inline by default, so it's needlessly limited by the compiler's inlining capability. Removing it would have no problem.

From: https://github.com/D-Programming-Language/druntime/pull/1491#issuecomment-181357376

### DIFF
--- a/src/gc/gc.d
+++ b/src/gc/gc.d
@@ -347,7 +347,6 @@ struct GC
 
     auto runLocked(alias func, alias time, alias count, Args...)(auto ref Args args)
     {
-        debug(PROFILE_API) {} else pragma(inline, true);
         debug(PROFILE_API) immutable tm = (GC.config.profile > 1 ? currTime.ticks : 0);
         gcLock.lock();
         debug(PROFILE_API) immutable tm2 = (GC.config.profile > 1 ? currTime.ticks : 0);


MartinNowak:
The point of this is to ensure that runLocked itself is inlined, as not doing so costs a few percent.

ibuclaw:
Benchmarks?

rainers:
> Benchmarks?

Here are the last 6 benchmarks for Win32 on my i7-2670QM:
```
                noinline,  inline
slist            3.340 s,  3.266 s
testgc3          2.197 s,  2.122 s
tree1            0.902 s,  0.900 s
tree2            1.828 s,  1.749 s
vdparser         3.569 s,  3.460 s
words            1.741 s,  1.757 s
```
Test results are reproducible within 1-2%, most other tests don't show a larger difference.

ibuclaw:
Seems pretty negligible to me.

9rnsr:
@MartinNowak `pragma(inline, true)` affects not only for the direct body of runLocked itself, and also for the *expanded any code* in it. So, if the called `func` condtantly  has big body, the runLocked inlining would fail always.

@klickverbot Yes, if `-inline` is not specified, currently `pragma(inline, true)` has no-op.

9rnsr:
@ibuclaw @rainers It's not surprising to me that there's no performance difference. Say again, `pragma(inline, true)` is enabled only when `-inline` is specified. And, by default `-inline` tries to expand all compiled functions bodies. Therefore, if a function has been succeeded to compile with `-inline`, adding `pragma(inline, true)` to it has no effect - the function has been already inlined.

In short, `pragma(inline, true)` is *not* a feature to increase code performance by using inlining. It's not designed for that purpose. From the view, @MartinNowak seems to misunderstand about its use case. The only one use case I can see is, to observe compiler's inlining behavior - stop compilation when function body expansion is *unintentionally* failed.

rainers:
> Seems pretty negligible to me.

> It's not surprising to me that there's no performance difference

There is a difference in some tests of 3-4%. We've been adding up these tiny performance steps in the past to get the GC quite a bit faster.

> The only one use case I can see is, to observe compiler's inlining behavior - stop compilation when function body expansion is unintentionally failed.

That's exactly the use case here. What's pretty unexpected is that it is influenced by a function called by runLocked. Take for example a simple wrapper:
```
void condFunc(bool b)
{
   if(b)
      largeFunc();
}
```
I would very much want this function to be inlined no matter what largeFunc does.

Maybe you can just annotate reallocNoSync with `pragma(inline,false)`.

rainers:
> In short, pragma(inline, true) is not a feature to increase code performance by using inlining. It's not designed for that purpose.

According to the docs, it is designed for that purpose, but not ready yet. I would have hoped that it would at least increase the cost threshold for inlining already.

9rnsr:
> There is a difference in some tests of 3-4%. We've been adding up these tiny performance steps in the past to get the GC quite a bit faster.

How the benchmark is measured? `-inline` switch difference is out of the discussion table.

> I would very much want this function to be inlined no matter what largeFunc does.

Your expect does not fit to current compiler's behavior. Walter's first implementation sometimes is not well implemented, compared from other people's expectation.

> According to the docs, it is designed for that purpose,

You meant that `pragma(inline, true)` is designed to add more performance? I cannot get read it from the current doc: http://dlang.org/spec/pragma.html#inline

Anyway, if an addition of `pragma(inline, true)` increases performance, it would be a result of dmd inliner's whim (or bug). From my knowledge about dmd, `pragma(inline, true)` would not work such.

schveiguy:
@9rnsr are you saying that pragma(inline, true) *requires* the given function and all functions it calls to be inlined? That's ridiculous. It should only affect the function tagged. @rainers is absolutely right, if it doesn't do that, I consider that a bug.

9rnsr:
@schveiguy Yes, currently it works as so. If you say it's a bug, please file it in bugzilla.

rainers:
> How the benchmark is measured? -inline switch difference is out of the discussion table.

I used `pragma(inline,false)` to actually measure the effect of inlining.

If it works as it does right now, `pragma(inline)` is too unreliable, so I'm ok with removing it. @MartinNowak, ok to merge?

rainers:
> You meant that pragma(inline, true) is designed to add more performance? I cannot get read it from the current doc: http://dlang.org/spec/pragma.html#inline

It says "This is expected to be improved in the future to causing functions to always be inlined regardless of compiler switch settings" which I interpret as "force inline".

schveiguy:
> Yes, currently it works as so.

I don't think it works this way. I created a very long function, then called the function from another function.

```D
shared int x;
void longFunc()
{
    version(bad)
        pragma(inline, true);
    import std.random;
    x = uniform!int();
    x = uniform!int();
    x = uniform!int();
    x = uniform!int();
    x = uniform!int();
    x = uniform!int();
    x = uniform!int();
} 

void foo(alias func)()
{
    pragma(inline, true);
    func();
}

void main()
{
    foo!longFunc();
}
```

So this compiles with `-inline`, not with `-inline -version=bad`

What gets inlined here, what does not? I'm completely unsure how this feature works.

ibuclaw:
@schveiguy - the same functions that are inlineable continue to be inlined with or without the pragma.  What changes is the compiler complaining when it can not inline a marked function.

ibuclaw:
I'm not sure about the 'flatten' effect reported here, maybe it's an artifact of how dmd's inliner works.

schveiguy:
@ibuclaw No, I get that.

What I'm trying to figure out is, why is this PR necessary? If inlining works the way I suspected it did (and it sure appears to based on my test), the function that was passed as an alias to `runLocked` shouldn't matter. It's OK if func isn't inlined, but the call to `runLocked` must be.

Perhaps there is an interaction artifact that I'm not understanding. Like for instance if the compiler chooses to inline `func` into `runLocked`, and then can't inline `runLocked`. That's a poor decision on the compiler, since it could meet the constraints given if it chose to inline `runLocked` and not `func`.

schveiguy:
> Like for instance if the compiler chooses to inline `func` into `runLocked`, and then can't inline `runLocked`.

OK, I modified my example, and that indeed is the problem.

```D
import std.random;
shared int x;
void longFunc()
{
    x = uniform!int();
    x = uniform!int();
    x = uniform!int();
    x = uniform!int();
    x = uniform!int();
    x = uniform!int();
} 

void foo(alias func)()
{
    pragma(inline, true);
    func();
    x = uniform!int();
}

void main()
{
    foo!longFunc();
}
```
Fails to compile.

DMD sees that `longFunc` can be inlined inside the `foo` function. But then when it tries to inline `foo` in function `main`, `foo` has become too long (note that I figured out how many calls to `uniform` would prevent inlining, so I just moved one call from `longFunc` into `foo`).

So the compiler isn't smart enough to pick which functions to inline based on the pragmas. If it chose to NOT inline `longFunc` inside `foo`, then `foo` would always be inlineable.

I'll file an enhancement request. I agree this should be merged.

schveiguy:
Auto-merge toggled on

ibuclaw:
Probably inlining starts at the lowest denominator and works it's way up the call stack.

9rnsr:
Thanks! I'll also see the enhancement for `pragma(inline)`.

MartinNowak:
Ah, just wanted to merge this. I think we fundamentally need to redo inlining in dmd, and move it from frontend to IR or even the backend. Given all the deficiencies dmd's optimizer has it might not even be worth the trouble.

PR #1498, opened 2016-02-15
Issue 15680 - TypeInfo broken for typeof(null)

9rnsr:
Add `TypeInfo_n` for `typeid(typeof(null))`.

Needs to be merged before dmd codegen fix #5459.

### DIFF
--- a/src/rt/typeinfo/ti_n.d
+++ b/src/rt/typeinfo/ti_n.d
@@ -1,0 +1,58 @@
+/**
+ * TypeInfo support code.
+ *
+ * Copyright: Copyright Digital Mars 2016.
+ * License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Authors:   Kenji Hara
+ */
+
+/*          Copyright Digital Mars 2016.
+ * Distributed under the Boost Software License, Version 1.0.
+ *    (See accompanying file LICENSE or copy at
+ *          http://www.boost.org/LICENSE_1_0.txt)
+ */
+module rt.typeinfo.ti_n;
+
+// typeof(null)
+
+class TypeInfo_n : TypeInfo
+{
+    override string toString() const @safe { return "typeof(null)"; }
+
+    override size_t getHash(in void* p) const
+    {
+        return 0;
+    }
+
+    override bool equals(in void* p1, in void* p2) const @trusted
+    {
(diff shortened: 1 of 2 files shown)


src/rt/typeinfo/ti_n.d
```diff
@@ -0,0 +1,58 @@
...
+
+    override size_t getHash(in void* p) const
+    {
+        return 0;
+    }
+
+    override bool equals(in void* p1, in void* p2) const @trusted
+    {
+        //return *cast(typeof(null)*)p1 is *cast(typeof(null)*)p2;
+        return true;
```
ibuclaw: Is this right?

9rnsr: From the logical consequence, yes. `typeof(null)` can have only one value `null`, so two typeof(null)` values are always equals, comparison returns 0, and swap has no effect.

ibuclaw: Would you mind documenting in all places why you return a literal then?  Rather than leaving the original code stubbed out.

9rnsr: What I'm concern is, how these TypeInfo member functions should work. As I already said, returning constant is logically valid. But if they should compare the actual runtime values, they are not good. For example:

```d
int x, y;
auto n1 = cast(typeof(null))&x;
auto n2 = cast(typeof(null))&y;
assert(typeid(typeof(null)).equals(&n1, &n2) == ?);
```

WalterBright: Casting a non-null value to null is undefined behavior as far as I know. So we don't care what it equals, only `null` can be of type null and so they should all compare equal.

src/rt/typeinfo/ti_n.d
```diff
@@ -0,0 +1,58 @@
...
+/*          Copyright Digital Mars 2016.
+ * Distributed under the Boost Software License, Version 1.0.
+ *    (See accompanying file LICENSE or copy at
+ *          http://www.boost.org/LICENSE_1_0.txt)
+ */
+module rt.typeinfo.ti_n;
+
+// typeof(null)
+
+class TypeInfo_n : TypeInfo
```
WalterBright: should be `TypeInfo_null`

WalterBright:
Auto-merge toggled on

PR #1499, opened 2016-02-27
fixes for reading EH info

WalterBright:
Required for https://github.com/D-Programming-Language/dmd/pull/5469


MartinNowak:
Some reference would be nice (also in the commit message), it reduces maintenance overhead not having to search the whole internet to figure out why those constants where used.

MartinNowak:
Auto-merge toggled on

WalterBright:
Those constants are used in the EH table format documentation, such as it is. Interestingly, googling on the constant name will take you pretty much right to it.

MartinNowak:
Must be your own search bubble, but on page 4 I found this nice overview article http://www.airs.com/blog/archives/460.

PR #1500, opened 2016-03-02
odbc32.def Unicode functions

john-sevsk:
Unicode versions of functions that accept strings arguments
+ add some missing functions
+ delete wrongly defined duplicates for SQLExecDirect and SQLGetData


MartinNowak:
Auto-merge toggled on

PR #1501, opened 2016-03-04
Linux/PPC: Add errno constants.

redstar:


MartinNowak:
Auto-merge toggled on

PR #1502, opened 2016-03-06
Fix epoll_event for non-X86 architectures

jpf91:
X86 is the only architecture where epoll_event is actually packed. All other architectures define epoll_event as a normal struct:
http://code.metager.de/source/search?q=__EPOLL_PACKED&path=%2Fgnu%2Fglibc%2F&project=gnu
See also: http://bugzilla.gdcproject.org/show_bug.cgi?id=214

### DIFF
--- a/src/core/sys/linux/epoll.d
+++ b/src/core/sys/linux/epoll.d
@@ -45,11 +45,35 @@ enum
     EPOLL_CTL_MOD = 3, // Change file descriptor epoll_event structure.
 }
 
-align(1) struct epoll_event
+version (X86)
 {
-align(1):
-    uint events;
-    epoll_data_t data;
+    align(1) struct epoll_event
+    {
+    align(1):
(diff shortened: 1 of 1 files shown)


yebblies:
We should really consider adding automatic struct layout tests at some point...

yebblies:
Auto-merge toggled on

ibuclaw:
We should also consider having a list of primary targets to support in druntime, rather than breaking builds for eg: PPC after someone has gone through the effort of porting all other parts.

jpf91:
> We should also consider having a list of primary targets to support in druntime, rather than breaking builds for eg: PPC after someone has gone through the effort of porting all other parts.

That's true, we should have a list somewhere. @redstar is maintaining the PPC port IIRC? also ping @klickverbot @kinke

redstar:
See #1504

PR #1503, opened 2016-03-06
use timelimit to kill deadlocked tests

MartinNowak:
- restrict any test to 10s
- don't trigger the 1h timeout for the auto-tester
- get a make error message w/ a specific test on timeout failure
- the timelimit program is used b/c the auto-tester already relies on it

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -97,6 +97,9 @@ LINKDL=$(if $(findstring $(OS),linux),-L-ldl,)
 
 MAKEFILE = $(firstword $(MAKEFILE_LIST))
 
+# use timelimit to avoid deadlocks if available
+TIMELIMIT:=$(if $(shell which timelimit 2>/dev/null || true),timelimit -t 10 ,)
+
 ######################## All of'em ##############################
 
 ifneq (,$(SHARED))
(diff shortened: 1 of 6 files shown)


MartinNowak:
Because the STARTING message (589e5c298623033e2a291f9214252e50845e6908) didn't actually solve the problem that it's difficult to figure out which test deadlocked.

braddr:
Great in concept, but the 'command' command doesn't exist on the linux platforms:
  make: command: Command not found

The 'which' command ought to do the trick though.  The '-s' option will suppress output and just set the return value.

MartinNowak:
> Great in concept, but the 'command' command doesn't exist on the linux platforms:
> make: command: Command not found

How can that not exist if it's a bash builtin?
http://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html#index-command
Updated to use which instead.

> The '-s' option will suppress output and just set the return value.

My which doesn't have that and `$(if` only looks for empty/non-empty string anyhow.

braddr:
Auto-merge toggled on

braddr:
valid question, don't have a good answer off the top of my head.

MartinNowak:
Now that this is done we can remove the pointless STARTING output https://github.com/D-Programming-Language/druntime/pull/1505.

PR #1505, opened 2016-03-09
remove the pointless STARTING output during testing

MartinNowak:
- we now have make non-quiet by default and apply timelimit
  to each test so it's easy to figure out deadlocked tests
- with parallel make it wasn't easily possible to see which
  STARTING/PASS lines belong together anyhow


DmitryOlshansky:
Auto-merge toggled on

PR #1507, opened 2016-03-10
Fix15513 (revised)

puneet:
Fix for https://issues.dlang.org/show_bug.cgi?id=15513

As pointed out by @schuetzm, I did something wrong while merging a rebase from the master branch. So I closed PR 1506 and am creating this new one.

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -305,6 +305,10 @@ else version( Posix )
             }
             assert( obj );
 
+            // loadedLibraries need to be inherited from parent thread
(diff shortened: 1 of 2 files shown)


MartinNowak:
Thanks, there are already plenty of tests that should've catched this, e.g. [this](https://github.com/D-Programming-Language/druntime/blob/bbd797aac223cf734cfd822f64db7450999ec66e/test/shared/src/plugin.d#L24), but they weren't working, see #1510.

PR #1508, opened 2016-03-11
fix Issue 15353 - ignore freeing during finalization

MartinNowak:


changelog.dd
```diff
@@ -42,6 +43,12 @@ Macros:
     COREMODREF = <a href="phobos/core_$1.html">$2</a>
     XREF = <a href="phobos/std_$1.html#$2">$2</a>
     CXREF = <a href="phobos/core_$1.html#$2">$2</a>
+    OXREF = <a href="phobos/object.html#$2">$2</a>
+    NXREF = <a href="phobos/std_$1.html#.$2">$2</a>
```
andralex: Are these other added macros used?

andralex:
Nice, thanks. I'll pull this although there's a question about the extra macros.

andralex:
Auto-merge toggled on

changelog.dd
MartinNowak: I just keep them in sync w/ what's available in dlang.org, so people don't fall back to plain html.

src/gc/gc.d
```diff
@@ -3250,6 +3250,23 @@ unittest // bugzilla 14467
...
+    static struct Foo
+    {
+        ~this()
+        {
+            GC.free(buf); // ignored in finalizer
+        }
+
+        void* buf;
+    }
+    new Foo(GC.malloc(10));
```
CyberShadow: This should probably be in a loop, to make sure there is no left-over pointer in the registers or stack.

PR #1509, opened 2016-03-12
remove more pointless start/pass output

MartinNowak:
- the makefile already prints a Testing finalize


DmitryOlshansky:
Auto-merge toggled on

PR #1510, opened 2016-03-12
fix regression tests for fixed Issue 15513

MartinNowak:
- the existing tls GC test weren't working b/c the scanned stack region
  still contained references to the allocated values


MartinNowak:
Auto-merge toggled on

PR #1511, opened 2016-03-12
core.math: Add unit test for 128bit reals.

redstar:
Required for AArch64 port.


tsbockman:
Looks good to me.

tsbockman:
@klickverbot 
> the exponent range is not bigger than 80 bit reals for this one?

[Correct.](https://en.wikipedia.org/wiki/Floating_point#Internal_representation)

PR #1514, opened 2016-03-19
core.stdc.stdarg should be nothrow

WalterBright:


rainers:
LGTM. I recently needed this, too.

rainers:
Auto-merge toggled on

PR #1516, opened 2016-03-21
local module name string hides object.string

MartinNowak:


rainers:
Auto-merge toggled on

PR #1517, opened 2016-03-22
Implement a software fallback algorithm for core.bitop.bsf() and bsr()

tsbockman:
This pull request is intended to supersede [PR #1486](https://github.com/D-Programming-Language/druntime/pull/1486#issuecomment-199689712).

This fixes an issue identified by @denis-sh with the `ulong` versions that [I previously added](https://github.com/D-Programming-Language/druntime/pull/1452) to `core.bitop` for `bsf()` and `bsr()`: the `ulong` overloads were defined to forward to the `size_t` versions on 64-bit, but then `size_t` is just an alias for `ulong`, which would make the function infinitely recursive.

This was not a problem in testing, because on 64-bit the `ulong` function body is replaced by a compiler intrinsic on all current DMD platforms.

Still, this pull request solves this potential issue by adding a software fallback algorithm for bit scans - `softScan()` - just in case someone wants to use `bsf()` or `bsr()` on a platform where they are not intrinsics.

There is no longer an explicit `size_t` overload of `bsf()` or `bsr()`; instead there are separate `uint` and `ulong` overloads, each of which forwards to the overload which is expected to be an intrinsic on the current platform. The expected intrinsic, in turn, falls back to `softScan()` if needed.

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -70,14 +70,21 @@ unittest
  *      The bit number of the first bit set.
  *      The return value is undefined if v is zero.
  */
-int bsf(size_t v) pure;
+int bsf(uint v) pure
+{
+    static if (size_t.sizeof == ulong.sizeof)
+    {
+        pragma(inline, true);
+        return bsf(cast(ulong) v);
+    }
+    else
+        return softBsf!uint(v);
+}
 
 /// ditto
 int bsf(ulong v) pure
 {
-    static if (size_t.sizeof == ulong.sizeof)
-        return bsf(cast(size_t) v);
-    else static if (size_t.sizeof == uint.sizeof)
+    static if (size_t.sizeof == uint.sizeof)
     {
         const sv = Split64(v);
         return (sv.lo == 0)?
@@ -85,7 +92,7 @@ int bsf(ulong v) pure
             bsf(sv.lo);
     }
     else
-        static assert(false);
+        return softBsf!ulong(v);
 }
 
 ///
@@ -110,14 +117,21 @@ unittest
  *      The bit number of the first bit set.
  *      The return value is undefined if v is zero.
  */
-int bsr(size_t v) pure;
+int bsr(uint v) pure
+{
+    static if (size_t.sizeof == ulong.sizeof)
+    {
+        pragma(inline, true);
+        return bsr(cast(ulong) v);
+    }
+    else
+        return softBsr!uint(v);
+}
 
 /// ditto
 int bsr(ulong v) pure
 {
-    static if (size_t.sizeof == ulong.sizeof)
-        return bsr(cast(size_t) v);
-    else static if (size_t.sizeof == uint.sizeof)
+    static if (size_t.sizeof == uint.sizeof)
     {
         const sv = Split64(v);
         return (sv.hi == 0)?
@@ -125,7 +139,7 @@ int bsr(ulong v) pure
             bsr(sv.hi) + 32;
     }
     else
-        static assert(false);
+        return softBsr!ulong(v);
 }
 
 ///
@@ -140,6 +154,110 @@ unittest
     // Make sure bsr() is available at CTFE
     enum test_ctfe = bsr(ulong.max);
     assert(test_ctfe == 63);
+}
+
+private alias softBsf(N) = softScan!(N, true);
+private alias softBsr(N) = softScan!(N, false);
+
+/* Shared software fallback implementation for bit scan foward and reverse.
+
+If forward is true, bsf is computed (the index of the first set bit).
+If forward is false, bsr is computed (the index of the last set bit).
+
+-1 is returned if no bits are set (v == 0).
+*/
+private int softScan(N, bool forward)(N v) pure
+    if(is(N == uint) || is(N == ulong))
+{
+    // bsf() and bsr() are officially undefined for v == 0.
+    if (!v)
+        return -1;
+
+    // This is essentially an unrolled binary search:
+    enum mask(ulong lo) = forward ? cast(N) lo : cast(N)~lo;
+    enum inc(int up) = forward ? up : -up;
+
+    N x;
+    int ret;
+    static if (is(N == ulong))
+    {
+        x = v & mask!0x0000_0000_FFFF_FFFFL;
+        if (x)
+        {
+            v = x;
+            ret = forward ? 0 : 63;
+        }
+        else
+            ret = forward ? 32 : 31;
+
+        x = v & mask!0x0000_FFFF_0000_FFFFL;
+        if (x)
+            v = x;
+        else
+            ret += inc!16;
+    }
+    else static if (is(N == uint))
+    {
+        x = v & mask!0x0000_FFFF;
+        if (x)
+        {
+            v = x;
+            ret = forward ? 0 : 31;
+        }
+        else
+            ret = forward ? 16 : 15;
+    }
+    else
+        static assert(false);
+
+    x = v & mask!0x00FF_00FF_00FF_00FFL;
+    if (x)
+        v = x;
+    else
+        ret += inc!8;
+
+    x = v & mask!0x0F0F_0F0F_0F0F_0F0FL;
+    if (x)
+        v = x;
+    else
+        ret += inc!4;
+
+    x = v & mask!0x3333_3333_3333_3333L;
+    if (x)
+        v = x;
+    else
+        ret += inc!2;
+
+    x = v & mask!0x5555_5555_5555_5555L;
+    if (!x)
+        ret += inc!1;
+
+    return ret;
+}
+
+unittest
+{
+    assert(softBsf!uint(0u) == -1);
+    assert(softBsr!uint(0u) == -1);
+    assert(softBsf!ulong(0uL) == -1);
+    assert(softBsr!ulong(0uL) == -1);
+
+    assert(softBsf!uint(0x0031_A000) == 13);
+    assert(softBsr!uint(0x0031_A000) == 21);
+    assert(softBsf!ulong(0x0000_0001_8000_0000L) == 31);
+    assert(softBsr!ulong(0x0000_0001_8000_0000L) == 32);
+
+    foreach (b; 0 .. 64)
+    {
+        if(b < 32)
+        {
+            assert(softBsf!uint(1u << b) == b);
+            assert(softBsr!uint(1u << b) == b);
+        }
+
+        assert(softBsf!ulong(1uL << b) == b);
+        assert(softBsr!ulong(1uL << b) == b);
+    }
 }
 
 /**


denis-sh:
Yes, let's add more useless functions and increase code size with `softScan`. The more the better, isn't it?

tsbockman:
> Yes, let's add more useless functions and increase code size with `softScan`.

Why did you even complain and make your pull request in the first place, if you don't care about making `bsf()`/`bsr()` work without the intrinsics? Just to be pedantic?

Your PR doesn't actually *do* anything as far as I can tell.

denis-sh:
> Why did you even complain and make your pull request in the first place, if you don't care about making bsf()/bsr() work without the intrinsics? Just to be pedantic?

Two reasons:

1. There should be no useless code in the library.
2. There should be no wrong code in a library even if it's unused now (for unused code see 1).

denis-sh:
> Your PR doesn't actually do anything as far as I can tell.

My PR removes 64-bit `bsf`/`bsr` implementation in 64 bit mode, i.e. removes unused code.

tsbockman:
OK. I'll let the reviewers decide whether they want my fallback implementation or not.

CyberShadow:
> Yes, let's add more useless functions and increase code size with softScan. The more the better, isn't it?

Please be nice.

src/core/bitop.d
```diff
@@ -142,6 +156,74 @@ unittest
     assert(test_ctfe == 63);
 }
 
+private alias softBsf(N) = softScan!(N, true);
```
WalterBright: Function is not documented.

src/core/bitop.d
```diff
@@ -142,6 +156,74 @@ unittest
...
+        else
+            ret = forward? 32 : 31;
+
+        x = v & mask!0x0000_FFFF_0000_FFFFL;
+        if (x)
+            v = x;
+        else
+            ret += inc!16;
+    }
+    else
```
WalterBright: else what values of N are acceptable? Clearly `ubyte` won't work, but there needs to be a test for what the code is designed to work with.

src/core/bitop.d
```diff
@@ -142,6 +156,74 @@ unittest
...
+private int softScan(N, bool forward)(N v) pure
+    if(is(N == uint) || is(N == ulong))
+{
+    enum mask(ulong lo) = forward? cast(N) lo : cast(N)~lo;
+    enum inc(int up) = forward? up : -up;
+
+    if (!v)
+        return -1;
+
+    N x = void;
```
WalterBright: `= void` is not necessary, as standard dead assignment analysis will remove it.

src/core/bitop.d
```diff
@@ -142,6 +156,74 @@ unittest
...
+    int ret = void;
```
WalterBright: ditto

src/core/bitop.d
WalterBright: Ah, I missed it must be `uint`. Perhaps add `else static if (is(N == uint))` and then an `else static assert(0);`

src/core/bitop.d
```diff
@@ -142,6 +156,74 @@ unittest
...
+    else
+        ret += inc!2;
+
+    x = v & mask!0x5555_5555_5555_5555L;
+    if (!x)
+        ret += inc!1;
+
+    return ret;
+}
+
```
WalterBright: Need unit tests. No reason not to get 100% line coverage.

tsbockman:
I added documentation and a fairly thorough `unittest` for `softScan`, as well as made the requested style tweaks.

@WalterBright, I believe I have addressed all of your comments, unless the `softBsf` and `softBsr` aliases really need to be documented separately from `softScan`? I can't think why they would though, since these are all private and there's nothing to say about the aliases that isn't already right there in the code.

src/core/bitop.d
```diff
@@ -142,6 +156,110 @@ unittest
...
+private int softScan(N, bool forward)(N v) pure
+    if(is(N == uint) || is(N == ulong))
+{
+    // bsf() and bsr() are officially undefined for v == 0.
+    if (!v)
+        return -1;
+
+    // This is essentially an unrolled binary search:
+    enum mask(ulong lo) = forward? cast(N) lo : cast(N)~lo;
+    enum inc(int up) = forward? up : -up;
```
ibuclaw: What is this, ruby? :-)

Though I'm not immediately sure of any coding convention imposed on condition expressions in druntime/phobos.

tsbockman: > What is this, ruby? :-)

For what it's worth, I've never written so much as one line of ruby in my life. I do like the ternary operator, though.

schuetzm: He means that you should insert a space before the `?`. (In Ruby, method names can end in `?`.)

tsbockman:
@ibuclaw D-rubyfied.

DmitryOlshansky:
LGTM Anything else @WalterBright?

DmitryOlshansky:
Auto-merge toggled on

tsbockman:
@DmitryOlshansky Since you merged this, #1486 should be closed now?

9il:
This should be reverted. Please see #1694

tsbockman:
> This should be reverted.

How about we *fix* it instead?

9il:
I think it should be reverted first

tsbockman:
Have you made any effort to understand why the intrinsics aren't working? Where's your test code? What compiler(s) are you having problems with?

9il:
I have - http://forum.dlang.org/post/ajifmjpyrijddhiblaky@forum.dlang.org. Why do you think they work (i mean BSR/BSF instructions for x86)? Where is your test code?

tsbockman:
> I think it should be reverted first

That would be a breaking change. The practical motivation for this PR was to get `bsf()` and `bsr()` working at CTFE.

> Why do you think they work (i mean BSR/BSF instructions for x86)?

Because I tested them when I submitted this PR. I probably only tested the intrinsic aspect with DMD though. (LDC had it's own druntime implementation at the time, with many of the differences relating to intrinsic handling. I figured they would figure out how to integrate it, as they had obviously done for past intrinsic work.)

> Where is your test code?

Gone. I didn't keep it once this got pulled.

> I have - http://forum.dlang.org/post/ajifmjpyrijddhiblaky@forum.dlang.org.

Thanks. I'll take a look.

tsbockman:
> I have - http://forum.dlang.org/post/ajifmjpyrijddhiblaky@forum.dlang.org.

That's not a valid test. Intrinsics are matched to fully qualified names; once you copy-paste the code to another module it stops matching.

If you actually `import core.bitop;` as intended, it works fine on DMD and GDC. LDC additionally requires `-enable-cross-module-inlining`, as noted by [Johan Engelen](http://forum.dlang.org/post/gbqfcshdayxgcxcmrgkx@forum.dlang.org).

9il:
No, LDC and GDC have not recent Druntime. Please provide poof for recent DMD

tsbockman:
> Please provide poof for recent DMD

It's currently working correctly for `bsr(ulong)` on 64-bit:
```
0000000000427cc8 <_Dmain>:
_Dmain():
/home/tsbockman/Code/d-scratch/source/app.d:5
module app;

import core.bitop, core.time;

int main() {
  427cc8:	55                   	push   rbp
  427cc9:	48 8b ec             	mov    rbp,rsp
  427ccc:	48 83 ec 10          	sub    rsp,0x10
/home/tsbockman/Code/d-scratch/source/app.d:6
    const rtval = MonoTime.currTime.ticks();
  427cd0:	e8 87 05 00 00       	call   42825c <_D4core4time42__T12MonoTimeImplVE4core4time9ClockTypei0Z12MonoTimeImpl8currTimeFNbNdNiNeZS4core4time42__T12MonoTimeImplVE4core4time9ClockTypei0Z12MonoTimeImpl>
  427cd5:	48 89 45 f8          	mov    QWORD PTR [rbp-0x8],rax
  427cd9:	48 8d 7d f8          	lea    rdi,[rbp-0x8]
  427cdd:	e8 ca 06 00 00       	call   4283ac <_D4core4time42__T12MonoTimeImplVE4core4time9ClockTypei0Z12MonoTimeImpl5ticksMxFNaNbNdNiNfZl>
  427ce2:	48 89 45 f0          	mov    QWORD PTR [rbp-0x10],rax
/home/tsbockman/Code/d-scratch/source/app.d:7
    return bsr(rtval);
  427ce6:	48 0f bd 45 f0       	bsr    rax,QWORD PTR [rbp-0x10]
/home/tsbockman/Code/d-scratch/source/app.d:8
}
  427ceb:	c9                   	leave  
  427cec:	c3                   	ret    
  427ced:	0f 1f 00             	nop    DWORD PTR [rax]
```
But **NOT** for `bsr(uint)` on 64-bit for some reason:
```
0000000000427d08 <_Dmain>:
_Dmain():
/home/tsbockman/Code/d-scratch/source/app.d:5
module app;

import core.bitop, core.time;

int main() {
  427d08:	55                   	push   rbp
  427d09:	48 8b ec             	mov    rbp,rsp
  427d0c:	48 83 ec 10          	sub    rsp,0x10
/home/tsbockman/Code/d-scratch/source/app.d:6
    const rtval = cast(uint) MonoTime.currTime.ticks();
  427d10:	e8 53 06 00 00       	call   428368 <_D4core4time42__T12MonoTimeImplVE4core4time9ClockTypei0Z12MonoTimeImpl8currTimeFNbNdNiNeZS4core4time42__T12MonoTimeImplVE4core4time9ClockTypei0Z12MonoTimeImpl>
  427d15:	48 89 45 f8          	mov    QWORD PTR [rbp-0x8],rax
  427d19:	48 8d 7d f8          	lea    rdi,[rbp-0x8]
  427d1d:	e8 96 07 00 00       	call   4284b8 <_D4core4time42__T12MonoTimeImplVE4core4time9ClockTypei0Z12MonoTimeImpl5ticksMxFNaNbNdNiNfZl>
  427d22:	89 45 f0             	mov    DWORD PTR [rbp-0x10],eax
/home/tsbockman/Code/d-scratch/source/app.d:7
    return bsr(rtval);
  427d25:	8b 7d f0             	mov    edi,DWORD PTR [rbp-0x10]
  427d28:	e8 53 01 00 00       	call   427e80 <_D4core5bitop3bsrFNaNbNiNfkZi>
/home/tsbockman/Code/d-scratch/source/app.d:8
}
  427d2d:	c9                   	leave  
  427d2e:	c3                   	ret    
  427d2f:	90                   	nop
```
```
0000000000427e80 <_D4core5bitop3bsrFNaNbNiNfkZi>:
_D4core5bitop3bsrFNaNbNiNfkZi():
  427e80:	55                   	push   rbp
  427e81:	48 8b ec             	mov    rbp,rsp
  427e84:	48 83 ec 10          	sub    rsp,0x10
  427e88:	89 7d f8             	mov    DWORD PTR [rbp-0x8],edi
  427e8b:	8b 7d f8             	mov    edi,DWORD PTR [rbp-0x8]
  427e8e:	e8 05 00 00 00       	call   427e98 <_D4core5bitop19__T8softScanTmVbi0Z8softScanFNaNbNiNfmZi>
  427e93:	48 8b e5             	mov    rsp,rbp
  427e96:	5d                   	pop    rbp
  427e97:	c3                   	ret    
```
I will have to investigate further and see if I can figure out why it's taking the wrong `static if` branch for `uint`.

tsbockman:
It seems that DMD's intrinsic detection fails randomly if there is a function body available: https://issues.dlang.org/show_bug.cgi?id=16743

It looks like a compiler issue, not a druntime one; GDC and LDC are unaffected.

tsbockman:
It turns out that intrinsics and inlining don't mix. A fix is in the works:

[DMD PR #6278](https://github.com/dlang/dmd/pull/6278)
[druntime PR #1695](https://github.com/dlang/druntime/pull/1695)

PR #1518, opened 2016-03-22
core.sys.windows.winuser: Provide valid GetMonitorInfo overloads

kinke:
LDC isn't fooled by using the same mangled name for multiple C function declarations with differing signatures (see [DMD bugzilla 15217](https://issues.dlang.org/show_bug.cgi?id=15217)).

In the MS header, the ...EX structs derive from the non-EX ones, so a pointer to an EX struct is implicitly converted to a pointer to the base struct.

Provide `extern(D)` wrappers performing the explicitly required cast in D.


rainers:
For C++ the MS headers declare MONITORINFOEX as a derived class of MONITORINFO, which allows to call GetMonitorInfo with both structs without casting. I guess this has been tried to emulate. Should we try to do this too, maybe with "alias this"?

There are other places in druntime where function signatures for extern(C) functions don't match. Try to build with -singleobj! I'm not sure if LDC is too picky about this because it might be used to circumvent some compiler checks, e.g. to pretent a function is (strongly) pure while the implementation actually does not compile as pure, but is carefully verified to be.

kinke:
> I guess this has been tried to emulate.

Yep, that's what I though too. But there should be WAY more WinAPI functions taking a pointer to a normal or EXtended struct; for those there's no extra overload.

>  I'm not sure if LDC is too picky about this because it might be used to circumvent some compiler checks, e.g. to pretent a function is (strongly) pure while the implementation actually does not compile as pure, but is carefully verified to be.

In this case, LDC only complained about non-matching function signatures for the same mangled name. I'm pretty sure D purity isn't reflected in the IR for `extern(C)` functions, so that shouldn't be a problem.

But dmd accepting 2 different signatures for an `extern(Windows)` function is definitely not okay. I hope this hack is not considered a feature!

kinke:
I've re-added the convenience overloads.

JohanEngelen:
It is a DMD bug: https://issues.dlang.org/show_bug.cgi?id=15217

JohanEngelen:
@kinke I applied a simpler fix in LDC merge-2.070 before I saw your PR here. Ping me once this is merged so I can update LDC's druntime.

kinke:
No need to ping, I'll just cherry-pick it myself once and if this gets merged.
Please don't keep on doing so much directly on the branches. Open PRs instead, as I wouldn't have let your 'fix' slip through; this needs to be fixed here upstream.

rainers:
I wonder if the original committers @CyberShadow and @adamdruppe dislike adding extern(D) wrappers for this overload. The C headers have a special case, too, i.e. they use inheritence for the implicit conversion. If there are no objections I'm fine with merging (with the Ex removed from the overload).

kinke:
Thx Rainer, updated.

CyberShadow:
LGTM.

> I wonder if the original committers @CyberShadow and @adamdruppe dislike adding extern(D) wrappers for this overload.

It's really no different than how a macro would be translated.

PR #1520, opened 2016-03-23
GC: remove condition that is never false

rainers:
The condition should have been `if (npools > 1)` but I removed it because we should not optimize for the unlikely state that the application is using less than 1MB of memory. Instead this improves the situation when more moemory is used and scan speed matters more.

### DIFF
--- a/src/gc/gc.d
+++ b/src/gc/gc.d
@@ -2020,29 +2020,21 @@ struct Gcx
                     continue;
 
                 Pool* pool = void;
-                if (npools > 0)
+                size_t low = 0;
+                size_t high = highpool;
+                while (true)
                 {
-                    size_t low = 0;
-                    size_t high = highpool;
-                    while (true)
-                    {
-                        size_t mid = (low + high) >> 1;
-                        pool = pools[mid];
-                        if (p < pool.baseAddr)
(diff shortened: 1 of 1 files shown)


schveiguy:
Technically, you could pull that if check out of the loop altogether (the number of pools doesn't change inside the scan, right?)

What I think might be better is to have a threshold where linear search is used instead of binary with small enough number of pools (maybe 8?) which is checked outside the scan loop. In essence, have 2 different scan mechanisms, optimized to the situation at the outset.

schveiguy:
In any case, trivially, this LGTM, it doesn't change anything, but removes an always-true if check.

DmitryOlshansky:
Auto-merge toggled on

MartinNowak:
> What I think might be better is to have a threshold where linear search is used instead of binary with small enough number of pools (maybe 8?) which is checked outside the scan loop. In essence, have 2 different scan mechanisms, optimized to the situation at the outset.

Not sure that'll be better, the binary search has a really low overhead, but it's easy enough to try.
In the longer run we want to replace the pools with a page allocator, combined w/ page to metadata hashtable.
https://trello.com/c/0WxZP9KL/11-freelist-for-large-objects-page-heap

PR #1525, opened 2016-03-27
Fix issue 15838 	Many Win32 API callback functions miss extern(Windows)

qchikara:
Add extern (Windows) to aliases and struct fields.


src/core/sys/windows/commctrl.d
```diff
@@ -3869,7 +3869,7 @@ static if (_WIN32_WINNT >= 0x501) {
     }
     alias LVINSERTGROUPSORTED* PLVINSERTGROUPSORTED;
 
-    alias int function(INT, INT, VOID*) PFNLVGROUPCOMPARE;
+    extern (Windows) alias int function(INT, INT, VOID*) PFNLVGROUPCOMPARE;
```
DmitryOlshansky: Shouldn't it be `alias PFNLVGROUPCOMPARE = extern(Windows) int function(INT, INT, VOID*);` ?

qchikara: This is OK,
and extern(Windows){  aliases ...  }, too.

qchikara: Ah, old style aliases should be fixed ...?

DmitryOlshansky: Yes, that'd be great as a separate commit.

qchikara: I think the change of styles should be made at all.
So, these core.sys.windows.* modules have thousands of aliases, mostly in old styles.
I misses such a skill to do efficiently :P

DmitryOlshansky: Okay let's ignore them for now.

DmitryOlshansky:
LGTM

DmitryOlshansky:
Auto-merge toggled on

PR #1526, opened 2016-03-28
TraceInfo: add a frames property

tomerfiliba:
http://forum.dlang.org/post/rqwxzjcyrccgzkbaqgsc@forum.dlang.org

### DIFF
--- a/src/core/exception.d
+++ b/src/core/exception.d
@@ -701,6 +701,7 @@ package class SuppressTraceInfo : Throwable.TraceInfo
     override int opApply(scope int delegate(ref const(char[]))) const { return 0; }
     override int opApply(scope int delegate(ref size_t, ref const(char[]))) const { return 0; }
     override string toString() const { return null; }
+    override @property const(void*)[] frames() const {return null;};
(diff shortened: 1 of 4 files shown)


Geod24:
An overload of `opApply` that expose a `Frame` would be better suited / encapsulated IMO.

Also, please :
- provide a clear explanation on why this is needed as part of the commit message: when reading the history, it might not obvious why this is needed, and one would have to jump through various hoops to understand this change (commit -> P.R. -> forum post). Some of those steps might also not be available (because the user is offline / the reference is gone).
- Add documentation for the new function(s) you add (at least for the base class, we should have a way to inherit documentation).
- Add a `static assert` after `else version (Windows)` so that folks working with more platform will get a clear and loud "Fix this" message.

Finally, it could also be worth adding a test, at least to make sure the behaviour is the same on all `Posix` platform (Which platform did you test this on?).

MartinNowak:
> An overload of opApply that expose a Frame would be better suited / encapsulated IMO.

Yes, we need to make stacktraces `@nogc` at some point and returning a slice would no longer be safe.

PR #1527, opened 2016-03-29
Remove invalid extern(Windows) overloads.

JohanEngelen:
DMD accepts this invalid code, see: https://issues.dlang.org/show_bug.cgi?id=15217
but LDC errors on it.

In this particular case, calls to the overloaded functions will function correctly because a pointer is passed and the structs themselves contain their own size such that Windows can distinguish the two calls.
After this PR, the user can call GetMonitorInfoA with a MONITORINFOEXA using a `cast(LPMONITORINFO)`.


JohanEngelen:
Duplicate https://github.com/D-Programming-Language/druntime/pull/1518

PR #1528, opened 2016-03-30
Fix Issue 1180 - the GC failes to handle large allocation requests propperly 

rainers:
report out-of-memory for allocation a bit below size_t.max

even though reported as a regression, this doesn't seem like anything new, so it might not be necessary to merge into stable.


DmitryOlshansky:
LGTM

DmitryOlshansky:
Auto-merge toggled on

schveiguy:
Auto-merge toggled off

schveiguy:
Re-enable auto merge after fix for win64 build

schveiguy:
Auto-merge toggled on

PR #1529, opened 2016-03-31
benchmark for array ops

MartinNowak:
- including a plotting script

### DIFF
--- a/benchmark/arrayops/arrayops.d
+++ b/benchmark/arrayops/arrayops.d
@@ -94,8 +94,34 @@ void runOp(string op)()
             getLatencies!(T, op), getThroughput!(T, op));
 }
 
+version (X86)
+    version = SSE;
+else version (X86_64)
+    version = SSE;
+else
+    static assert(0, "unimplemented");
+
+void ignoreDenormals()
+{
+    version (SSE)
+    {
+        uint mxcsr = void;
+        asm
+        {
+            stmxcsr mxcsr;
+        }
+        mxcsr |= 0x8000 | 0x0040; // FTZ and DAZ
+        asm
+        {
+            ldmxcsr mxcsr;
+        }
+    }
+}
+
 void main()
 {
+    ignoreDenormals();
+
     writefln("type, op, %(latency%s, %), %-(throughput%s, %)", iota(6)
         .map!(i => 1 << i), ["8KB", "32KB", "512KB", "32MB"]);
     foreach (op; mixin("AliasSeq!(%(%s, %))".format(genOps)))
--- a/benchmark/arrayops/arrayops.d
+++ b/benchmark/arrayops/arrayops.d
@@ -1,0 +1,103 @@
+/**
+ * Benchmark for array ops.
+ *
+ * Copyright: Copyright Martin Nowak 2016 -.
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:    Martin Nowak
+ */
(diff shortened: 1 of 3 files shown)


MartinNowak:
Example plots.

array_ops_latency
array_ops_throughput

rainers:
Could you add a comment to plot.R how to run it?

On win32, it crashes after `ubyte, a += b + c` with "out of memory".

On win64, it takes considerable time, so it should be excluded from the runbench tests. We could add some disable mechanism to runbench, or the benchmark should be moved into another directory.

benchmark/arrayops/arrayops.d
```diff
@@ -0,0 +1,103 @@
...
+    foreach (i, ref latency; latencies)
+    {
+        auto len = 1 << i;
+        foreach (_; 1 .. 32)
+        {
+            auto sw = StopWatch(AutoStart.yes);
+            foreach (off; 0 .. 1_024)
+            {
+                off &= 127;
+                enum op = op.replace("const", "1").replace("a",
```
rainers: Maybe '1' isn't the best generic constant, as there might be special optimizations for `mul` or `div`. I'm not sure how to avoid overflow and underflow with other constants for all types, though.

MartinNowak: Good point, will use some primes instead.

MartinNowak:
All done

rainers:
I suspect that you get a lot of zero's, denormals and overflow with `*=` and `/=` operators now. Reinitializing the arrays before each benchmark might avoid some of these, but not all.

MartinNowak:
I enabled flush-to-zero which fixed an issue w/ the `a /= b / c` and `a /= b / const` latency benchmarks (plots above are updated).
Also uses malloc/free now.

rainers:
Thanks. I stepped through some versions and the throughput tests seem ok, max values are 1e+18. /= operations on integers result in the buffer becoming zero after 1 or 2 iterations. Reinitializing the buffer before each iteration could help.

The latency test for "a *= b" on floats overflows to "inf" in the first iteration, though. I expect similar issues with most *= and /= operations.

benchmark/arrayops/arrayops.d
```diff
@@ -0,0 +1,147 @@
+/**
+ * Benchmark for array ops.
+ *
+ * Copyright: Copyright Martin Nowak 2016 -.
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:    Martin Nowak
+ */
```
WalterBright: Add `Source:` tag.

benchmark/arrayops/arrayops.d
```diff
@@ -0,0 +1,147 @@
+/**
+ * Benchmark for array ops.
+ *
+ * Copyright: Copyright Martin Nowak 2016 -.
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:    Martin Nowak
```
WalterBright: Line up columns.

benchmark/arrayops/arrayops.d
MartinNowak: No need for those in benchmarks.

benchmark/arrayops/arrayops.d
MartinNowak: It's just copied from the other benchmarks, and I won't do this differently only in a single file.
Also an almost OT review comment while we're still discussing the actual implementation.

MartinNowak:
Finally updated to avoid FPU exceptions.

PR #1530, opened 2016-04-02
Fix issue 15863: add length to AA.byKey/Value/KeyValue

yshui:
Make it easier to pass those ranges to functions that requires
hasLength.

* * *

This is my first D pull request, so if I got anything wrong please help me fix it.

Thanks.

issue https://issues.dlang.org/show_bug.cgi?id=15863

### DIFF
--- a/src/rt/aaA.d
+++ b/src/rt/aaA.d
@@ -643,6 +643,7 @@ struct Range
 {
     Impl* impl;
     size_t idx;
+    size_t length;
     alias impl this;
 }
 
@@ -656,14 +657,14 @@ extern (C) pure nothrow @nogc
         foreach (i; aa.firstUsed .. aa.dim)
         {
             if (aa.buckets[i].filled)
-                return Range(aa.impl, i);
+                return Range(aa.impl, i, aa.length);
         }
-        return Range(aa, aa.dim);
+        return Range(aa, aa.dim, aa.length);
     }
 
     bool _aaRangeEmpty(Range r)
     {
-        return r.impl is null || r.idx == r.dim;
+        return r.length == 0;
     }
 
     void* _aaRangeFrontKey(Range r)
@@ -678,6 +679,7 @@ extern (C) pure nothrow @nogc
 
     void _aaRangePopFront(ref Range r)
     {
+        r.length--;
         for (++r.idx; r.idx < r.dim; ++r.idx)
         {
             if (r.buckets[r.idx].filled)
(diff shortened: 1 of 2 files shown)


ZombineDev:
Can you add a changelog entry [here](https://github.com/D-Programming-Language/druntime/blob/master/changelog.dd)? You would need remove the current entries and bump the version to 2.072.0.

Otherwise, LGTM.

schveiguy:
> Can you add a changelog entry

No need. you just need to identify the bug in the commit message, and it will close the issue and auto generate a changelog entry.

src/rt/aaA.d
```diff
@@ -684,6 +684,10 @@ extern (C) pure nothrow @nogc
                 break;
         }
     }
+    size_t _aaRangeLen(Range r)
+    {
+        return r.impl is null ? 0 : r.impl.length;
```
schveiguy: This is wrong. The length should reflect how many items are left in the range, not the total elements in the AA.

MartinNowak:
Don't we have anything in phobos to combine a forward range w/ a know length?
Some people might not need length and would have to pay for the extra `size_t` and `--length` of a correct implementation.

src/rt/aaA.d
yshui: @schveiguy Hmm, but Impl.length returns 'used - deleted', so I thought that's the one?

schveiguy: That's length of aa, not range. This should work:

```D
auto r = aa.byKey;
auto len = r.length;
r.popFront;
assert(r.length == len - 1);
```

yshui: @schveiguy So used - deleted - idx?

schveiguy: No. You need a new field. Idx is just the bucket and doesn't have any relation to the number of elements in the range

yshui:
I can't think of a clean way of keeping track of the length of the range without adding a new member, so that's what I did.

MartinNowak:
I'd really like to see this as more general construct in std.range (augmenting a forward range with a length similar to take(n)) rather than slowing down byKey for everyone.

yshui:
@MartinNowak If that's what you want I think take(n) already works?

aa.byKey.take(aa.length)

MartinNowak:
> @MartinNowak If that's what you want I think take(n) already works?
> aa.byKey.takeExactly(aa.length)

Great, we usually favor choice and composability over already making those choices or providing different variations.
Might be worth to add the example to the docs.

schveiguy:
@MartinNowak agreed. @yshui here is the page to update if you want to add the example to the docs: https://github.com/D-Programming-Language/dlang.org/edit/master/spec/hash-map.dd

PR #1532, opened 2016-04-05
Add new module for Linux header file sys/auxv.

redstar:
This header file provides access to the auxiliary vector. It was added
to glibc version 2.16. See https://lwn.net/Articles/519085/ for some
background information.

Also adds the missing AT_HWCAP2 index to the elf module.

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -92,6 +92,7 @@ COPY=\
 	$(IMPDIR)\core\sys\linux\tipc.d \
 	$(IMPDIR)\core\sys\linux\unistd.d \
 	\
+	$(IMPDIR)\core\sys\linux\sys\auxv.d \
 	$(IMPDIR)\core\sys\linux\sys\inotify.d \
 	$(IMPDIR)\core\sys\linux\sys\mman.d \
 	$(IMPDIR)\core\sys\linux\sys\signalfd.d \
--- a/src/core/sys/linux/sys/auxv.d
+++ b/src/core/sys/linux/sys/auxv.d
@@ -1,0 +1,163 @@
+/**
+ * D header file for GNU/Linux.
+ *
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   Kai Nacke
+ */
+module core.sys.linux.sys.auxv;
+
+import core.stdc.config;
+
+version (linux):
+extern (C):
+
+c_ulong getauxval(c_ulong type) nothrow pure @nogc @system;
+
+version(PPC)
+{
+  // See https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/powerpc/bits/hwcap.h
+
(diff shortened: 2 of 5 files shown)


DmitryOlshansky:
LGTM

redstar:
Anything preventing a merge? @klickverbot @ibuclaw @MartinNowak

ibuclaw:
Looks fine, do the makefiles need to be updated?

redstar:
Good hint. I missed to add the file to `mak/COPY`. There should be no need to update other make related files.

ibuclaw:
Are you sure? I thought there was a MANIFESTS file.

mak/COPY
```diff
@@ -75,6 +75,7 @@ COPY=\
 	$(IMPDIR)\core\sys\freebsd\sys\mman.d \
 	$(IMPDIR)\core\sys\freebsd\time.d \
 	\
+	$(IMPDIR)\core\sys\linux\auxv.d \
```
ibuclaw: Ahh, I see, you missed the second `\sys` in `sys\linux\sys\aux.d`

redstar:
Added the missing `sys`.

ibuclaw:
@redstar apparently you need an explicit copy rule in the win32 and win64 make files.

redstar:
Rebased on master. Includes all fixes.

src/core/sys/linux/sys/auxv.d
```diff
@@ -0,0 +1,165 @@
...
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   Kai Nacke
+ */
+module core.sys.linux.sys.auxv;
+
+import core.sys.posix.sys.types;
+
+version (linux):
+extern (C):
+@system:
```
WalterBright: Since it accepts and returns a ulong, perhaps it should be @safe ?

src/core/sys/linux/sys/auxv.d
```diff
@@ -0,0 +1,165 @@
...
+nothrow:
+
```
WalterBright: add `@nogc`

src/core/sys/linux/sys/auxv.d
```diff
@@ -0,0 +1,165 @@
...
+c_ulong getauxval(c_ulong type);
```
WalterBright: Since this is the only function declared in the module, perhaps it's best to not use the : syntax for the attributes, but apply them directly to the function.

WalterBright: I suspect that `getauxval()` can be `pure`:
http://man7.org/linux/man-pages/man3/getauxval.3.html

src/core/sys/linux/sys/auxv.d
redstar: Internally the function uses a lot of pointer manipulation. The can't be safe.

WalterBright:
Auto-merge toggled on

PR #1533, opened 2016-04-06
fix Issue 15886 - Add an uncaught exception handler

CyberShadow:
https://issues.dlang.org/show_bug.cgi?id=15886

I'm not sure how to do the win32 part. @llucax?

### DIFF
--- a/src/rt/deh.d
+++ b/src/rt/deh.d
@@ -24,6 +24,25 @@ extern (C)
             t.info = _d_traceContext(context);
         }
     }
+
+    /// Uncaught exception handler.
+    /// Should only ever be executed if rt_trapExceptions was 0 when
+    /// the runtime was initialized.
+    /// Not supported on Win32 as stack unwinding and uncaught
+    /// exceptions are handled by the OS there.
+    /// If this function returns, a platform-specific action occurs
+    /// (e.g. abort() is called, a HLT instruction is executed, or the
+    /// exception is passed to the operating system).
+    __gshared void function(Object o) _d_uncaughtException =
+        &_d_uncaughtExceptionDefaultHandler;
+
+    /// Default uncaught exception handler.
+    /// It should be possible to breakpoint this function to stop the
+    /// debugger on uncaught exceptions before the stack has been unwound
+    /// (except on Win32 - see above).
+    void _d_uncaughtExceptionDefaultHandler(Object o)
+    {
+    }
 }
 
 version (Win32)
--- a/src/rt/deh_win64_posix.d
+++ b/src/rt/deh_win64_posix.d
@@ -36,6 +36,7 @@ extern (C)
     Throwable.TraceInfo _d_traceContext(void* ptr = null);
     int _d_isbaseof(ClassInfo oc, ClassInfo c);
     void _d_createTrace(Object o, void* context);
+    extern __gshared void function(Object o) _d_uncaughtException;
 }
 
 alias int function() fp_t;   // function pointer in ambient memory model
@@ -491,6 +492,7 @@ extern (C) void _d_throwc(Object h)
             }
         }
     }
+    _d_uncaughtException(h);
     terminate();
 }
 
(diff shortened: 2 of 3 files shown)


CyberShadow:
For some reason the exception passed to the handler doesn't have a stack trace (when printed with `_d_print_throwable`), even though `_d_createTrace` is called on the object. Not sure why. @Walte

llucax:
> I'm not sure how to do the win32 part. @llucax?

If you are asking me about win32, you are barking at the wrong tree :)

llucax:
So in Linux, if no special handler is used, `abort()` is called but **after** the stack was unwound?

CyberShadow:
> If you are asking me about win32, you are barking at the wrong tree :)

Ah, git-blame pointed me to https://github.com/D-Programming-Language/druntime/commit/04e29d17959a8e14a8b2f5c06dfdf72f87f9206a, sorry. @donc? :)

> So in Linux, if no special handler is used, abort() is called but after the stack was unwound?

Well, if there are no exception frames, there is nowhere to unwind to. Normally there is a top-level frame in `_d_run_main`, which can be disabled by setting `rt_trapExceptions` to 0.

Marenz:
> So in Linux, if no special handler is used, abort() is called but after the stack was unwound?

As far as I can tell that is not true, output of a small test:

```
(gdb) break _d_run_main
(gdb) set rt_trapExceptions = 0
(gdb) cont
Continuing.

uncaught exception
dwarfeh(224) fatal error

0x00007ffff740fcc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56	../nptl/sysdeps/unix/sysv/linux/raise.c: Datei oder Verzeichnis nicht gefunden.
(gdb) bt
#0  0x00007ffff740fcc9 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffff74130d8 in __GI_abort () at abort.c:89
#2  0x0000000000423a56 in _d_throwdwarf ()
#3  0x0000000000421bb8 in _d_assert ()
#4  0x0000000000421918 in main.__assert() ()
#5  0x00000000004218ae in D main () at main.d:15
#6  0x00000000004238af in rt.dmain2._d_run_main() ()
#7  0x0000000000423827 in rt.dmain2._d_run_main() ()
#8  0x000000000042386b in rt.dmain2._d_run_main() ()
#9  0x0000000000423827 in rt.dmain2._d_run_main() ()
#10 0x000000000042374a in _d_run_main ()
#11 0x0000000000421968 in main ()
```

Not sure why I see `dwarfeh(224) fatal error` though

CyberShadow:
> Not sure why I see dwarfeh(224) fatal error though

https://github.com/D-Programming-Language/druntime/blob/master/src/rt/dwarfeh.d#L223-L224

which calls

https://github.com/D-Programming-Language/druntime/blob/master/src/rt/dwarfeh.d#L805-L809

Marenz:
Ah, makes sense now.

Marenz:

rainers:
> I'm not sure how to do the win32 part.

SetUnhandledExceptionFilter will probably help, see https://msdn.microsoft.com/en-us/library/ms680634.aspx

CyberShadow:
> SetUnhandledExceptionFilter will probably help, see https://msdn.microsoft.com/en-us/library/ms680634.aspx

Thanks. That's probably going too far for this pull request, so it's probably better to declare that this feature is unsupported on Win32.

CyberShadow:
A test would be much easier to write once https://github.com/dlang/druntime/pull/1538 is in.

Burgos:
See how I got the test for the same functionality, albeit only for dwarfeh: https://github.com/dlang/druntime/pull/1673/files#diff-eb6feae96258d70e7e610bffc21427bc

I was not aware of this PR when writing my own, I guess we shouldn't leave them too long to rot :(.

CyberShadow:
This would be nice to have for certain situations but does need some more thinking (especially how to approach the platform disparity).

PR #1534, opened 2016-04-06
If malloc fails, prevent getting Segmentation Fault

tcak:
When the memory is limited for the process, malloc fails, and data is null. Hence, it creates Segmentation Fault by trying to write a null pointer.

### DIFF
--- a/src/rt/tlsgc.d
+++ b/src/rt/tlsgc.d
@@ -32,6 +32,7 @@ struct Data
 void* init()
 {
     auto data = cast(Data*).malloc(Data.sizeof);
+    if( data is null ) return null;
     *data = Data.init;
 
     // do module specific initialization
--- a/src/rt/tlsgc.d
+++ b/src/rt/tlsgc.d
@@ -32,7 +32,8 @@ struct Data
 void* init()
 {
     auto data = cast(Data*).malloc(Data.sizeof);
-    if( data is null ) return null;
+    import core.exception;
+    if( data is null ) core.exception.onOutOfMemoryError();
     *data = Data.init;
 
     // do module specific initialization


tcak:
I would be calling core.exception.onOutOfMemoryError function instead of returning null, but then thought, returning null might be a better idea. Actually, calling that function would enable to catch memory problems much easily.

tcak:
If throwing won't be any problem, please note it here, and cancel this patch. So, I commit a new patch with exception.

rainers:
Returning null will result in missing roots during a GC scan, so I think throwing out-of-memory would be more appropriate. The source of error is a lot more obvious then.
You don't need to open another PR, just amend the change to this commit.

schuetzm:
Please squash, the LGTM

tcak:
This squash thing is going to kill me. I have spent about 1 hour, and couldn't managed to do it. I did the changes on code directly on GitHub. Cloned on my local computer, ran rebase -i HEAD~x things etc, but no luck.

schuetzm:
A squash button on Github would be really useful. What are the problems you were having?

tcak:
As I said, I made the changes directly on the GitHub's editor. Then it has put it into my fork as far as I understand. I clone my own fork, and check it with rebase. No matter how big number I type for HEAD~, it only shows my last commit, the previous is not that. So I cannot squash. On the other hand, I try same thing by cloning the D runtime itself, I even cannot find my commit in rebase list. So, I basically gave up.

yebblies:
Assuming you have `upstream` set up pointing to `D-Programming-Language/druntime`, you should be able to rebase it using:
```
git checkout <branch>
git fetch upstream master
git rebase -i upstream/master
```

rainers:
LGTM. There is no notification if the branch changes, so please ping on update.

rainers:
Auto-merge toggled on

PR #1536, opened 2016-04-07
fix issue 15111 - hashOf fails for structs that have an alias this to a dynamic array

BBasile:


DmitryOlshansky:
LGTM

BBasile:
This is a bit _of topic_ but will it be a nightmare if _multiple alias this_ is finally added to D ? Already with one alias something can be an array and struct. 

_std.traits_ will loose its mind with them.

Maybe this fix is a bit too pragmatic.

DmitryOlshansky:
Auto-merge toggled on

PR #1537, opened 2016-04-07
Make rt_trapExceptions apply to fibers as well

Marenz:


Marenz:
Ping.

Marenz:
Ping

Marenz:
It would be nice to get some feedback or at least _any_ reaction to this.

Marenz:

adamdruppe:
So, I'm not as confident here because I don't use fibers but I think it is a good idea too.

mathias-baumann-sociomantic:
@MartinNowak ping ping ping

mathias-baumann-sociomantic:
Only documentation I know of for this is

https://github.com/dlang/druntime/blob/15a227477a344583c4748d95492703901417f4f8/src/rt/dmain2.d#L142-L144

I can change the docu for `Fiber.call` to mention the behavior when rt_trapExceptions  is set

PR #1538, opened 2016-04-07
	Add runtime config option to change exception trapping behavior

Marenz:
This way you can do

```
extern extern (C) __gshared bool rt_trapExceptions;
static this ( )
{
        rt_trapExceptions = false;
}
```

to change the runtime behavior before main() is called.

### DIFF
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -139,8 +139,7 @@ extern (C) string[] rt_args()
 // make arguments passed to main available for being filtered by runtime initializers
 extern(C) __gshared char[][] _d_main_args = null;
 
-// This variable is only ever set by a debugger on initialization so it should
-// be fine to leave it as __gshared.
+// This variable can be set by a debugger or using the rt config option noTrapExceptions
 extern (C) __gshared bool rt_trapExceptions = true;
 
 alias void delegate(Throwable) ExceptionHandler;
@@ -430,17 +429,15 @@ extern (C) int _d_run_main(int argc, char **argv, MainFunc mainFunc)
         args = argsCopy[0..j];
     }
 
-    bool trapExceptions = rt_trapExceptions;
-
     version (Windows)
     {
         if (IsDebuggerPresent())
-            trapExceptions = false;
+            rt_trapExceptions = false;
     }
 
     void tryExec(scope void delegate() dg)
     {
-        if (trapExceptions)
+        if (rt_trapExceptions)
         {
             try
             {
@@ -475,6 +472,13 @@ extern (C) int _d_run_main(int argc, char **argv, MainFunc mainFunc)
 
         if (!rt_term())
             result = (result == EXIT_SUCCESS) ? EXIT_FAILURE : result;
+    }
+
+    import rt.config : rt_configOption;
+
+    if (rt_configOption("noTrapExceptions") !is null)
+    {
+        rt_trapExceptions = false;
     }
 
     tryExec(&runAll);
--- a/test/exceptions/line_trace.exp
+++ b/test/exceptions/line_trace.exp
@@ -2,8 +2,8 @@ object.Exception@src/line_trace.d(17): exception
 ----------------
 src/line_trace.d:17 void line_trace.f1() [ADDR]
 src/line_trace.d:5 _Dmain [ADDR]
-src/rt/dmain2.d:472 _D2rt6dmain211_d_run_mainUiPPaPUAAaZiZ6runAllMFZ9__lambda1MFZv [ADDR]
-src/rt/dmain2.d:447 void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [ADDR]
-src/rt/dmain2.d:472 void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll() [ADDR]
-src/rt/dmain2.d:447 void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [ADDR]
-src/rt/dmain2.d:480 _d_run_main [ADDR]
+src/rt/dmain2.d:469 _D2rt6dmain211_d_run_mainUiPPaPUAAaZiZ6runAllMFZ9__lambda1MFZv [ADDR]
+src/rt/dmain2.d:444 void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [ADDR]
+src/rt/dmain2.d:469 void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll() [ADDR]
+src/rt/dmain2.d:444 void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [ADDR]
+src/rt/dmain2.d:484 _d_run_main [ADDR]


CyberShadow:
Interesting, I thought `rt_trapExceptions` applies to unit tests / constructors as well. I think it should.

Perhaps update its documentation to reflect that it is now also settable from a static constructor?

Marenz:
Updated documentation for the variable

Marenz:
Updated unit test to account for changed line numbers

Marenz:
Ping.

Marenz:
Ping

Marenz:
It would be nice to get some feedback or at least _any_ reaction to this.

rainers:
I have not tried, but I don't think it works to set the flag in a ctor: there is an outer `tryExec(&runAll)` that will still catch excetions and errors (including those from static ctors).
I guess it might make sense together with applying the same flag to fibers, though.

CyberShadow:
> I have not tried, but I don't think it works to set the flag in a ctor: there is an outer tryExec(&runAll) that will still catch excetions and errors (including those from static ctors).

That's what I thought too, but it appears to be (no longer?) so.

Marenz:
You were right, it didn't work. I did test it and something did work before but I guess not this.
I updated the PR so that it is a runtime config option now. I tested it with 

```
void main()
{
    assert(1==2);
}
```
```
dub build
Performing "debug" build using dmd for x86.
test ~master: building configuration "application"...
Linking...
```

```
(gdb) r
Starting program: /home/marenz/projekte/test/test 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/i386-linux-gnu/i686/cmov/libthread_db.so.1".
core.exception.AssertError@source/app.d(3): Assertion failure
----------------
??:? _d_assert [0x806654e]
??:? void app.__assert(int) [0x8066494]
source/app.d:3 _Dmain [0x806644c]
??:? _D2rt6dmain211_d_run_mainUiPPaPUAAaZiZ6runAllMFZ9__lambda1MFZv [0x80668aa]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x80667ff]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll() [0x8066866]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x80667ff]
??:? _d_run_main [0x8066792]
??:? main [0x80664d3]
??:? __libc_start_main [0xb7d9ea62]
[Inferior 1 (process 28082) exited with code 01]
(gdb) bt
No stack.
```
```
(gdb) r --DRT-noTrapExceptions=1
Starting program: /home/marenz/projekte/test/test --DRT-noTrapExceptions=1
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/i386-linux-gnu/i686/cmov/libthread_db.so.1".
uncaught exception
dwarfeh(224) fatal error

0xb7fdcd40 in __kernel_vsyscall ()
(gdb) bt
#0  0xb7fdcd40 in __kernel_vsyscall ()
#1  0xb7db3367 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#2  0xb7db4a23 in __GI_abort () at abort.c:89
#3  0x08066a78 in _d_throwdwarf ()
#4  0x0806654f in _d_assert ()
#5  0x08066495 in app.__assert() ()
#6  0x0806644d in D main () at source/app.d:3
#7  0x080668ab in rt.dmain2._d_run_main() ()
#8  0x08066838 in rt.dmain2._d_run_main() ()
#9  0x08066867 in rt.dmain2._d_run_main() ()
#10 0x08066838 in rt.dmain2._d_run_main() ()
#11 0x08066793 in _d_run_main ()
#12 0x080664d4 in main ()
```

Or outside GDB:

```
➜  test  ./test --DRT-noTrapExceptions=1
uncaught exception
dwarfeh(224) fatal error
[1]    28317 abort (core dumped)  ./test --DRT-noTrapExceptions=1
➜  test  ./test                         
core.exception.AssertError@source/app.d(3): Assertion failure
----------------
??:? _d_assert [0x806654e]
??:? void app.__assert(int) [0x8066494]
source/app.d:3 _Dmain [0x806644c]
??:? _D2rt6dmain211_d_run_mainUiPPaPUAAaZiZ6runAllMFZ9__lambda1MFZv [0x80668aa]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x80667ff]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll() [0x8066866]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x80667ff]
??:? _d_run_main [0x8066792]
??:? main [0x80664d3]
??:? __libc_start_main [0xb74a2a62]

```

Marenz:

adamdruppe:
Yes, yes, yes, merge this!

BTW if it isn't merged, since it is an extern(C) variable, you can also redefine it yourself and instruct the linker to use your version instead of the druntime one. hacky sure but hey.

`extern(C) __gshared bool rt_trapExceptions = false;`

compile:
`dmd test.d -L-zmuldefs -g`


but this is a trivial change that has clear value, looks very good to me.

Marenz:
@klickverbot I wrote it above already: you pass ` --DRT-noTrapExceptions=1` to your application (one of the ways) and any assert failure will cause a coredump now instead of just outputting the stack trace.

Marenz:
Do you want to test the switch itself or just whether the runtime behaves correctly when the option is enabled?

Marenz:
I am asking because the parsing etc of that switch is all functionality that is already present.

Marenz:
Will that test need to be cross-platform?

src/rt/dmain2.d
```diff
@@ -477,6 +474,13 @@ extern (C) int _d_run_main(int argc, char **argv, MainFunc mainFunc)
             result = (result == EXIT_SUCCESS) ? EXIT_FAILURE : result;
     }
 
+    import rt.config : rt_configOption;
+
+    if (rt_configOption("noTrapExceptions") !is null)
```
leandro-lucarella-sociomantic: I suggest using options without negative: `trapExceptions`

mathias-baumann-sociomantic: Gladly, but that means the default will be changed, right? Is that acceptable @MartinNowak

leandro-lucarella-sociomantic: No, that's not necessary (nor desirable, I think). Just make it default to 1, not 0.

CyberShadow: Certainly not desirable :) It would mean that programs will segfault instead of showing an exception stack trace on uncaught exceptions by default.

mathias-baumann-sociomantic: We have different definitions of "desirable", but sure, I see if I can change it

adamdruppe:
why is this still in limbo?

CyberShadow:
Looks good to me. @MartinNowak ?

src/rt/dmain2.d
CyberShadow: It looks like you're using the result of `rt_configOption` like a bool here... which means that `--DRT-noTrapExceptions=1` and `--DRT-noTrapExceptions=0` will both do the same thing. This is not ideal.

I looked to see if we have a precedent for this, and here is how #1798 added `callStructDtorsDuringGC`, also a boolean option: https://github.com/dlang/druntime/blob/639878dc9c02dfff9603f024988ccd4d5ecee2cb/src/rt/lifetime.d#L49-L58

I suggest to parse this option in a similar way, which would also allow e.g. overriding the default behaviour of disabling `rt_trapExceptions` when running under a debugger on Windows. That also solves the "double negation" issue. If you're feeling up for it, refactoring the bool checks into a new function in the `rt.config` module, which also rejects invalid values, would be a bonus.

wilzbach:
Superseded by https://github.com/dlang/druntime/pull/2035

wilzbach:
AFAICT this stagnated/orphaned PR was about allowing `--DRT-noTrapExceptions=1` (or environment config) and #2035 is exactly about this too (just that a different naming was used).
`__gshared bool rt_trapExceptions` already exists in DRuntime.

PR #1539, opened 2016-04-11
Mark @nogc attribute for OSX semaphore and pthread functions

p0nce:


schuetzm:
LGTM

ZombineDev:
LGTM^2

schveiguy:
I think this looks good, but should we do non-osx specific ones too? e.g. https://github.com/p0nce/druntime/blob/master/src/core/sys/posix/pthread.d

schveiguy:
Wow, that seems to have broken things, Looks like there may be some function type issues?

joakim-noah:
OS X changes not needed because of recently merged #1448.

p0nce:
I guess this can be closed then?

joakim-noah:
Depends on whether you want to try and make `core.sys.posix.pthread` `@nogc` too by looking into why it's failing the auto-tester.  It appears that it's breaking some template instantiation that shouldn't have worked in the first place?

p0nce:
Well no I didn't want to do that.

PR #1541, opened 2016-04-16
GDC support for core.cpuid

mleise:
- Supports GDC based on front-end release 2.066.1
- Supports both AT&T and Intel syntax assemblers
- Supports x86, amd64 and x32
- Moves up the `stepping`, `model` and `family` calculation, as the `uint a` (holding the bits) got overwritten in the original code before being evaluated (See line 725 in the original code)
- Wrapper functions do **NOT** inline, that would require accessing the cached constants directly as GDC does not support cross module inlining

### DIFF
--- a/src/core/cpuid.d
+++ b/src/core/cpuid.d
@@ -81,14 +81,11 @@ nothrow:
 // AMD K10    --   + isX86_64()
 // Cyrix 6x86 -- preferPentium1()
 //    6x86MX  --   + mmx()
-version(D_InlineAsm_X86)
-{
-    version = InlineAsm_X86_Any;
-}
-else version(D_InlineAsm_X86_64)
-{
-    version = InlineAsm_X86_Any;
-}
+
+// GDC support uses extended inline assembly:
+//   https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html        (general information and hints)
+//   https://gcc.gnu.org/onlinedocs/gcc/Simple-Constraints.html  (binding variables to registers)
+//   https://gcc.gnu.org/onlinedocs/gcc/Machine-Constraints.html (x86 specific register short names)
 
 public:
 
@@ -429,7 +426,22 @@ CpuFeatures* getCpuFeatures() @nogc nothrow
     }
 
 
-version(InlineAsm_X86_Any) {
+version(GNU) {
+    version(X86)
+        enum supportedX86 = true;
+    else version(X86_64)
+        enum supportedX86 = true;
+    else
+        enum supportedX86 = false;
+} else version(D_InlineAsm_X86) {
+    enum supportedX86 = true;
+} else version(D_InlineAsm_X86_64) {
+    enum supportedX86 = true;
+} else {
+    enum supportedX86 = false;
+}
+
+static if (supportedX86) {
 // Note that this code will also work for Itanium in x86 mode.
 
 __gshared uint max_cpuid, max_extended_cpuid;
@@ -497,7 +509,9 @@ void getcacheinfoCPUID2()
     // for old single-core CPUs.
     uint numinfos = 1;
     do {
-        asm pure nothrow @nogc {
+        version(GNU) asm pure nothrow @nogc {
+            "cpuid" : "=a" a[0], "=b" a[1], "=c" a[2], "=d" a[3] : "a" 2;
+        } else asm pure nothrow @nogc {
             mov EAX, 2;
             cpuid;
             mov a, EAX;
@@ -539,7 +553,9 @@ void getcacheinfoCPUID4()
     int cachenum = 0;
     for(;;) {
         uint a, b, number_of_sets;
-        asm pure nothrow @nogc {
+        version(GNU) asm pure nothrow @nogc {
+            "cpuid" : "=a" a, "=b" b, "=c" number_of_sets : "a" 4, "c" cachenum : "edx";
+        } else asm pure nothrow @nogc {
             mov EAX, 4;
             mov ECX, cachenum;
             cpuid;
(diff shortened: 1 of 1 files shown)


mleise:
@WalterBright Although, you said you are against removing the wrappers and that they should not be called in hot code, I believe people would still do it. Compare for example https://github.com/dlang/phobos/blob/master/std/digest/sha.d#L225 where you call `ssse3` once for every 64 bytes of data.

@ibuclaw If you have the time, could you look over the asm blocks please? Until you posted it on the NG I didn't know you could write code compatible with AT&T and Intel at the same time and there is hardly anyone else around here that familiar with GCC asm.

src/core/cpuid.d
```diff
@@ -344,7 +341,7 @@ CpuFeatures cpuFeatures;
  */
 CpuFeatures* getCpuFeatures() @nogc nothrow
 {
-    pragma(inline, false);
+    version (GNU) {} else pragma(inline, false);
```
ibuclaw: Why this?

src/core/cpuid.d
```diff
@@ -497,7 +509,9 @@ void getcacheinfoCPUID2()
     // for old single-core CPUs.
     uint numinfos = 1;
     do {
-        asm pure nothrow @nogc {
+        version(GNU) asm {
```
ibuclaw: `pure nothrow @nogc`

src/core/cpuid.d
mleise: The version of GDC mentioned above does not support pragma(inline, ...) unless you disable unknown pragmas. Does the comment above the function apply to GCC in the same way? I could use you noinline attribute from gcc.attributes.

src/core/cpuid.d
mleise: For whatever reason I believed this syntax doesn't work. I'll add them everywhere.

src/core/cpuid.d
ibuclaw: I'd say assume you're supporting the current language version of dmd, not gdc.  Just because it doesn't support pragma inline now, doesn't mean that we should honour compatibility with 2.067 in upstream.

src/core/cpuid.d
ibuclaw: Supported since 2.067

mleise:
I will squash any extra commits that accumulate before the merge.

src/core/cpuid.d
mleise: It actually works with the 2.066.1 release, too.

ibuclaw:
Looks sound from a technical aspect.  I'll leave it up to you to decide whether one-liner statements are better than multi-line.

i.e:
```
asm pure nothrow @nogc {
    "cpuid"
    : "=a" dummy, "=c" numcores  /* outputs  */
    : "a" 0x8000_0008            /* inputs   */
    : "ebx", "edx";              /* clobbers */
}
```

src/core/cpuid.d
```diff
@@ -651,130 +676,157 @@ void getCpuInfo0B()
...
         }
     }
 
     if (max_extended_cpuid >= 0x8000_0004) {
-        char *procptr = cf.processorNameBuffer.ptr;
-        version(D_InlineAsm_X86)
+        uint* pnb = cast(uint*)cf.processorNameBuffer.ptr;
+        version(GNU)
+        {
+            asm pure nothrow @nogc { "cpuid" : "=a" pnb[0], "=b" pnb[1], "=c" pnb[ 2], "=d" pnb[ 3] : "a" 0x8000_0002; }
```
mleise: I find this style useful here, where output and input are obvious and the pattern is easy on the eyes. What do others think ?

src/core/cpuid.d
```diff
@@ -651,130 +676,157 @@ void getCpuInfo0B()
...
+            asm pure nothrow @nogc { "cpuid" : "=a" pnb[4], "=b" pnb[5], "=c" pnb[ 6], "=d" pnb[ 7] : "a" 0x8000_0003; }
+            asm pure nothrow @nogc { "cpuid" : "=a" pnb[8], "=b" pnb[9], "=c" pnb[10], "=d" pnb[11] : "a" 0x8000_0004; }
+        }
```
mleise: I find this style useful here, where output and input are obvious and the pattern is easy on the eyes. What do others think ?

DmitryOlshansky: LGTM reads rather fluently

ibuclaw:
:+1:

DmitryOlshansky:
Apparently yes, per Walter's suggestion

mleise:
Let's not wake sleeping dogs.

mleise:
I know that and the historical arguments haven't disappeared, they are just buried in memory today. Sometimes when you want something fixed, you need to trick people into thinking it was their own idea and trigger strong emotions that work towards your goal. If you look at the thread my first statement was vague: "What we have now doesn't work." Walter usually points at something in druntime/Phobos that is supposed to get the job done and any deficiencies should be reported as bugs/enhancements. So he mentioned core.cpuid, and I explained why it cannot work for GDC. I talked about how this wont be working for the "years to come" and how this should be fixed, trying to use bold claims and complaints without taking action to trigger the emotions I needed to cloud the rational thought for long enough to get past the realization that the proposed fix would run against his past stances on foreign compiler code in druntime. This could have worked out if you didn't remind him of why he refused LDC's patches in the past. So I'll quickly squash the commits and hope someone merges this! (GDC's cross-module-inlining can hopefully be fixed on their end some day.)

mleise:
P.S.: The last comment wasn't entirely serious. It could have been that way, but frankly I just wanted to go sure the work has a chance of getting merged and it does seem to be controversial after all.

@redstar I just heard you talk at DConf. Does your getauxval() work completely supersede the current core.cpuid? There is quite some code in the module for different generations of x86 and especially determining the cache levels, that's not covered by auxv. Do you plan on splitting CPU feature and cache detection or have the x86 specific code override the auxv information ?

DmitryOlshansky:
Auto-merge toggled on

ibuclaw:
@mleise, pr for gdc? ;-)

mleise:
@ibuclaw Ok, against which branch and which version of the file? It would be the easiest for me to create a PR with the druntime master version, but you probably prefer a clean patch against the version in the gdc repo, am I right?

ibuclaw:
Everything goes into master first.

PR #1542, opened 2016-04-19
fix Issue 15911 - undefined __Unwind_GetIPInfo for x86_64

MartinNowak:
- only use __Unwind_GetIPInfo w/ glibc


WalterBright:
Auto-merge toggled on

andralex:
good catch!

PR #1545, opened 2016-04-19
Deprecate TypeInfo.init

aG0aep6G:
As planned.

### DIFF
--- a/changelog.dd
+++ b/changelog.dd
@@ -5,9 +5,22 @@ $(COMMENT Pending changelog for 2.072.0.
 
 $(BUGSTITLE Library Changes,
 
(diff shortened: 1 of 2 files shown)


schveiguy:
LGTM.

ZombineDev:
LGTM too, but can you add a deprecation message that tells the users to use `initializer`?

schveiguy:
A deprecation message would be useful, and costs basically nothing. I think it's a good idea.

aG0aep6G:
> can you add a deprecation message that tells the users to use initializer?

done

schveiguy:
Auto-merge toggled on

PR #1547, opened 2016-04-22
Fix deadlock on runtime errors

WalterWaldron:
I found this issue investigating other deadlocks using an older D release.
I'll open an issue in the tracker (I haven't yet.)

The primary problem is that ```assert``` could trigger a GC allocation while the GC lock is held by another thread which is waiting for threads to suspend.

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -431,7 +431,8 @@ else version( Posix )
         extern (C) void thread_suspendHandler( int sig ) nothrow
         in
         {
-            assert( sig == suspendSignalNumber );
+            if( sig != suspendSignalNumber )
+                onThreadError( "Unexpected suspend signal" );
         }
         body
         {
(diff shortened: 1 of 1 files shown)


schveiguy:
LGTM. Only possibility is that you missed some asserts.

It may be the case that we should simply always call onThreadError for private functions inside core.thread.

WalterWaldron:
> LGTM. Only possibility is that you missed some asserts.

I did not remove other potentially problematic ```asserts``` not involved in the failure case I was working on to keep the commit atomic and avoid excess churn in the PR.
> It may be the case that we should simply always call ```onThreadError``` for private functions inside core.thread.

This would be a good policy to prevent ```assert``` from creeping into sensitive code paths.

schveiguy:
Why was this closed?

WalterWaldron:
> Why was this closed?

I want to re-review the code and consolidate some changes.
My attention/time has temporarily shifted and I'll get back to this.

schveiguy:
OK, thanks. I just wanted to make sure there wasn't something wrong. Carry on!

PR #1549, opened 2016-04-23
Add Virtual memory manager module

DmitryOlshansky:
A reboot of previous attempt.

The idea in a nut shell is to provide some portable recipes for working with virtual memory subsystem, it's not intended to be exhaustive but cases like:
- allocating executable memory
- commiting/decommit ranges of pages
- locking memory

can and should be portably supported.

### DIFF
--- a/src/core/vmm.d
+++ b/src/core/vmm.d
@@ -55,19 +55,22 @@ enum MemoryOptions : uint
 /**
     System's virtual memory manager capabilities flags,
     indicate whether an operation is supported natively.
-    If some operation is not supported native, the "next best thing" is done
+    If some operation is not supported natively, the "next best thing" is done where applicable
     such as e.g. madvising memory as "won't use" instead of forcibly decommiting pages on FreeBSD.
 */
 enum MemoryCapability
 {
     commit = 0x1,
     decommit = 0x2,
-    reset = 0x4
+    reset = 0x4,
+    hugePages = 0x8,
+    sharedMapping = 0x10
 };
 
 version(CoreDdoc)
     enum MemoryCapability memoryCapability =
-        MemoryCapability.commit | MemoryCapability.decommit | MemoryCapability.reset;
+        MemoryCapability.commit | MemoryCapability.decommit | MemoryCapability.reset
+        | MemoryCapability.hugePages | MemoryCapability.sharedMapping;
 else version(Windows)
     enum MemoryCapability memoryCapability =
         MemoryCapability.commit | MemoryCapability.decommit | MemoryCapability.reset;
@@ -269,8 +272,8 @@ unittest
     assert(commit(slice[$/2..$]));
     slice[$-1] = 0xEA;
     assert(slice[$-1] == 0xEA);
-//    assert(decommit(slice[$/2..$]));
-//    assert(commit(slice[$/2..$]));
+    decommit(slice[$/2..$]);
+    commit(slice[$/2..$]);
     //decommit - commit leaves memory zeroed-out on Linux/Windows
     version(Windows)
         assert(slice[$-1] == 0);
--- a/src/core/vmm.d
+++ b/src/core/vmm.d
@@ -1,0 +1,449 @@
+module core.vmm;
+
+/**
+    $(P The module aims to provide the most commonly useful features of
+    target OS virtual memory manager subsystem. )
+
+    $(P The primary goal is not to provide full potential of every OS as
+        that is next impossible due to inherent differences.
+        Instead the goal is to cover $(STRONG the most common) scenarios
+        in an $(STRONG optimal) way. In particular issuing the absolute minimum
+        of system calls for each scenario.
+    )
+
+    Major use cases covered:
+    $(LI Getting platform's page size as cheap as reading a global. See $(LREF pageSize).
+    It takes a single system call at application start up.)
+    $(LI Allocating a memory block that has executable permission. A must have for JITs.
+    See $(LREF allocate) to allocate a block and set permissions in one call.)
+    $(LI Reserving a large virtual memory range and gradually committing memory from it.
+    Important for data-heavy applications. The permissions are set to read/write.
+    See $$(LREF reserve) and $(LREF commit).)
+    $(LI Locking a range of memory pages to persist in physical memory
+        (avoid being paging out by OS). See $(LREF lock) and $(LREF unlock).)
+
+    $(P API provides a way to perform less common operations but may have some overhead
+    compared to writing platform-specific code.)
+
+*/
+
+import core.thread;
+
+/// Flags that specify desired kind of access to the memory.
+enum MemoryAccess : uint
+{
+    none =      0b0000_0000,
+    read =      0b0000_0001,
+    write =     0b0000_0010,
+    execute =   0b0000_0100,
+    readWrite = read | write,
+    readWriteExecute = read | write | execute,
+};
+
+/**
+    Flags that specify additional memory options.
+    BUGS: None of these are implemented yet.
+*/
+enum MemoryOptions : uint
+{
+    none =      0b0000_0000,
+    hugePages = 0b0001_0000
+};
+
+/// The page size as reported by the OS at the application start.
+public alias pageSize = PAGESIZE;
+
+//TODO: provide the same for huge page size
+
+/**
+    Allocate a block of $(D size) bytes of virtual memory rounded up to
+    nearest multiple of page size. The memory is allocated
+    with specified permissions and options.
+
+    $(D base) specifies the desired base address of the block, $(D null)
+    means any base address.
+
+    To release a block of memory allocated by this function use $(LREF free).
+
+    Note: This operation commits virtual memory.
+    To have fine grained control over reserve and commit phases
+    use $(LREF reserve) and $(LREF commit).
+*/
+void[] allocate(void* base, size_t size,
+    MemoryAccess access = MemoryAccess.readWrite,
+    MemoryOptions options = MemoryOptions.none)
+{
+    size = roundUpToPage(size);
+    void* ptr = osMap(base, size, access, options);
+    return ptr ? ptr[0..size] : null;
+}
+
+/**
+    Reserve a block of $(D size) bytes of virtual memory rounded up
+    to nearest multiple of page size. The memory is not commited
+    and/or OS is not explicitly asked to assign memory pages to it.
+
+    $(D base) specifies the desired base address of the block, $(D null)
+    means any base address.
+
+    To release a virtual memory range reserved by this function use $(LREF free).
+
+    Note: This function and $(LREF commit)
+    set read/write permissions but not execute.
+    The reason is that some OS would set permissions on reserve
+    while others on commit. Also Posix makes the commit stage almost transparent.
+    Use $(LREF protect) to override permissions as required.
+*/
+void[] reserve(void* base, size_t size)
+{
+    size = roundUpToPage(size);
+    void* ptr = osReserve(base, size);
+    return ptr ? ptr[0..size] : null;
+}
+
+/**
+    Commit a set of pages that form the specified $(D range) of virtual memory.
+    Only after applying this call is any memory obtained via $(LREF reserve)
+    guaranteed to be accessible.
+
+    Note: This function and $(LREF reserve)
+    set read/write permissions but not execute.
+    The reason is that some OS' would set permissions on reserve
+    while others on commit. Also Posix makes the commit stage almost transparent.
+    Use $(LREF protect) to override permissions as required.
+*/
+bool commit(void[] range)
+{
+    return osCommit(range.ptr, range.length);
+}
+
+/**
+    Resets a set of pages that form the specified $(D range) of virtual memory.
+    This instructs an OS that data in the pages is no loger of interest to the application,
+    so it's free to reuse any of page involved while keeping mapping intact.
+    Subsequent access will result in either reading a prior data or a fresh page.
+
+*/
+bool reset(void[] range)
+{
+    return osReset(range.ptr, range.length);
+}
+
+
+/**
+    Decommit a set of pages that form the specified $(D range) of virtual memory.
+    While system retains the address space range, accessing memory inside of it
+    is illegal. The exact consequences of such illegal access are platform
+    dependent.
+
+*/
+bool decommit(void[] range)
+{
+    return osDecommit(range.ptr, range.length);
+}
+
+/**
+    Decommit and releases pages that were previously allocated and/or reserved.
+    $(D range) must contain exactly the same address range as the one
+    returned by $(LREF allocate) or $(LREF reserve).
+
+    It doesn't matter what the state of each page was (reserved vs commited),
+    the OS decommits pages as needed.
+*/
+bool free(void[] range)
+{
+    return osUnmap(range.ptr, range.length);
+}
+
+/**
+    Set permissions on the set of pages that form $(D range).
+*/
+bool protect(void[] range, MemoryAccess access)
+{
+    return osProtect(range.ptr, range.length, access);
+}
+
+/**
+    Lock pages that form $(D range) to prevent their paging out.
+*/
+bool lock(void[] range)
+{
+    return osLockPages(range.ptr, range.length);
+}
+
+/**
+    Undo the effect of locking (with $(LREF lock)) pages that form $(D range).
+*/
+bool unlock(void[] range)
+{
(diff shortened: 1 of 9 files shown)


DmitryOlshansky:
https://github.com/dlang/druntime/pull/653

ZombineDev:
Since the interface is meant to be consumed by D code, why not use `void[]`, instead of `void* base, size_t size`?

ZombineDev:
Since every OS provides slightly different set of functionality, I think it would be useful to have an enum that is the union of all the possible capabilities and then you could query at CT what the target OS can offer. For example I think that in both Linux and Windows you can create mmaps that can be shared between processes.

DmitryOlshansky:
> I think it would be useful to have an enum that is the union of all the possible capabilities and then you could query at CT what the target OS can offer

That's a great idea! Meanwhile I added `reset` primitive which is kind of nice and supported everywhere but Linux.

src/core/vmm.d
```diff
@@ -0,0 +1,485 @@
...
+    BUGS: None of these are implemented yet.
+*/
+enum MemoryOptions : uint
+{
+    none =      0b0000_0000,
+    hugePages = 0b0001_0000
+};
+
+/// The page size as reported by the OS at the application start.
+static immutable size_t pageSize;
```
MartinNowak: We already have a pageSize somewhere in core.thread.

src/core/vmm.d
```diff
@@ -0,0 +1,485 @@
...
+    private shared int fzero;
+
+    shared static this()
+    {
+        pageSize = sysconf(_SC_PAGESIZE);
+        version(linux) //temporary
+        {
+            import core.sys.posix.fcntl;
+            fzero = open("/dev/zero", O_RDWR);
+            assert(fzero != -1);
```
MartinNowak: ???

MartinNowak: Doesn't mmap come with a private, not file backed option?
How about MAP_ANONYMOUS or fd == -1?

src/core/vmm.d
```diff
@@ -0,0 +1,485 @@
...
+    $(D base) specifies the desired base address of the block, $(D null)
+    means any base address.
+
+    To release a block of memory allocated by this function use $(LREF free).
+
+    Note: This operation commits virtual memory.
+    To have fine grained control over reserve and commit phases
+    use $(LREF reserve) and $(LREF commit).
+*/
+static void[] allocate(void* base, size_t size,
```
MartinNowak: static free functions?

MartinNowak:
Looks good, also like the enum support idea.
I'd go with `ubyte[]` instead of `void[]` b/c it's not GC tracked memory. If you want to store pointers in raw memory, you're entering unsafe territory and need casts anyhow.

ZombineDev:
@MartinNowak I don't follow what has `ubyte[]` vs `void[]` has to do with GC tracking. If you want the memory range to be scanned you need to register it with `GC.addRange`, otherwise it won't be, regardless if it's `ubyte[]` or `void[]`.
I prefer `void[]` and this is what `std.experimental.allocator` uses too. Since you're acquiring raw memory you need casts anyway and I see no benefit in typing elements as `ubyte`, as it will be the wrong type anyway.

Geod24:
> @MartinNowak I don't follow what has ubyte[] vs void[] has to do with GC tracking. If you want the memory range to be scanned you need to register it with GC.addRange, otherwise it won't be, regardless if it's ubyte[] or void[].

By default `void[]` is scanned for pointers, `ubyte[]` isn't. See [Void Arrays](https://dlang.org/spec/arrays.html).

ZombineDev:
@Geod24 this is only true if the memory has been allocated from the GC. Since the memory comes straight from the OS, it won't be scanned unless it is explicitly registered. The same is true for: `Mallocator.instance.allocate` which also returns `void[]`.

ZombineDev:
I think a common use case would be:
```
pure nothrow @nogc
unittest
{
    import std.experimental.allocator : make, dispose;

    int[] array = VMMAllocator.instance.makeArray!int(4, 2);
    scope(exit) VMMAllocator.instance.dispose(arr):

    // use array...
    arr[] *= 3;
    assert (arr == [6, 6, 6, 6]);
}

struct VMMAllocator
{
    // not shared, because it interacts poorly with pure code
    static VMMAllocator instance;

    void[] allocate(size_t size) pure nothrow @nogc
    {
        import core.vmm : allocate;
        return allocate(null, size);
    }

    // See https://github.com/dlang/druntime/pull/198
    // for a discussion on pure memory management.
    bool deallocate(void[] memory) pure nothrow @nogc
    {
        import core.vmm : free;
        return free(memory);
    }

    //... other allocator methods
}
```

So maybe it is a good idea to provide a mid-level `std.experimental.allocator` compatible interface and also provide low-level access to `os*` functions. That way the functionality could be plugged directly under one of the allocators, but the other useful functionality is still accessible.

MartinNowak:
If `std.experimental.allocator` uses it, then stick with it.

DmitryOlshansky:
> Looks good, also like the enum support idea.

All comments addressed, still sticking with `void[]`.

DmitryOlshansky:
> I think a common use case would be

It should be possible to create an allocator with this, but it's not the only use case nor I plan to add it in this pull.

DmitryOlshansky:
Dang FreeBSD fails to lock memory cause ZFS locks up to the limits. I just dealt with it on my VM, should I just disable the test on FreeBSD or do I get to configure auto-tester machines?

ZombineDev:
> nor do I plan to add it in this pull

No problem and sorry if was off topic.

But can you add some attributes, so those functions could be usable from an allocator? At least `nothrow` and `@nogc`.
`round*ToPage`, `protectionFlags`, `memoryOptions`, `memoryFlags` should also be `pure` and `@safe`.
Maybe `allocate` and `deallocate` could be also `pure @system`, so they could be used to implement a `@trusted @nogc nothrow pure RefCounted!Allocator` (RefCounted won't use those functions directly, but the `ParentAllocator` of `Allocator` might), but if this is too controversial, it would be better to leave the discussion for a another pull request.

DmitryOlshansky:
> But can you add some attributes, so those functions could be usable from an allocator? At least nothrow and @nogc.

Done

DmitryOlshansky:
> Looks good, also like the enum support idea.

Did it, by default all functions fallback to the most sensible "next best thing" but at least now it's possible to check if they do at compile time.

src/core/vmm.d
```diff
@@ -0,0 +1,486 @@
...
+    System's virtual memory manager capabilities flags,
+    indicate whether an operation is supported natively.
+    If some operation is not supported native, the "next best thing" is done
+    such as e.g. madvising memory as "won't use" instead of forcibly decommiting pages on FreeBSD.
+*/
+enum MemoryCapability
+{
+    commit = 0x1,
+    decommit = 0x2,
+    reset = 0x4
```
ZombineDev: I think adding `hugePages` to list (and not changing anything else) is a nice way to indicate that are not (yet) supported.

src/core/vmm.d
```diff
@@ -0,0 +1,486 @@
...
+};
+
+version(CoreDdoc)
+    enum MemoryCapability memoryCapability =
```
ZombineDev: A comment along these lines would be nice:
```
/** Using this enum you can query at compile-time whether the target platform
 * supports a certain feature.
 */
```

src/core/vmm.d
ZombineDev: Also I wonder if `sharedMemory` should be included in this list, even if not supported at the moment.

ZombineDev:
Thanks, @DmitryOlshansky. Overall LGTM, I have just some minor comments about the `memoryCapabilities`.

DmitryOlshansky:
Ping @ZombineDev @MartinNowak Anything else? Do we do a voting for this kind of additions?

schveiguy:
> Do we do a voting for this kind of additions?

I would err on the side of needing a vote. At the very least, because we are adding another module and feature to druntime.

If this were Phobos, I'd say it has to go into std.experimental. Should we make a core.experimental?

UplinkCoder:
I am in favor of adding this.

src/core/vmm.d
```diff
@@ -0,0 +1,489 @@
...
+    }
+
+    bool osCommit(void* ptr, size_t size)
+    {
+        ptr = roundDownToPage(ptr);
+        size = roundUpToPage(size);
+        version(linux)
+            return madvise(ptr, size, MADV_WILLNEED) == 0;
+        else
+            return posix_madvise(ptr, size, POSIX_MADV_WILLNEED) == 0;
```
Burgos: Why not just call `posix_madvise` on Linux as well?

src/core/vmm.d
```diff
@@ -0,0 +1,489 @@
...
+    }
+
+    bool osDecommit(void* ptr, size_t size)
+    {
+        ptr = roundDownToPage(ptr);
+        size = roundUpToPage(size);
+        version(linux)
+            return madvise(ptr, size, MADV_DONTNEED) == 0;
+        else
+            return posix_madvise(ptr, size, POSIX_MADV_DONTNEED) == 0;
```
Burgos: On error, `posix_madvise` _returns_ an error (as opposed to setting `errno`), which is lost by bool conversion.

src/core/vmm.d
Burgos: Yeah, I see the point in case of `POSIX_MADV_DONTNEED`. Fair enough.

src/core/vmm.d
```diff
@@ -0,0 +1,489 @@
...
+
+/// Flags that specify desired kind of access to the memory.
+enum MemoryAccess : uint
+{
+    none =      0b0000_0000,
+    read =      0b0000_0001,
+    write =     0b0000_0010,
+    execute =   0b0000_0100,
+    readWrite = read | write,
+    readWriteExecute = read | write | execute,
```
CyberShadow: Worth noting that some operating systems (OpenBSD) or OS configurations (Linux+grsec) forbid this combination for security reasons, and perhaps we should discourage it as well by not providing a predefined constant for it. How about `readExecute` instead?

src/core/vmm.d
```diff
@@ -0,0 +1,489 @@
...
+    such as e.g. madvising memory as "won't use" instead of forcibly decommiting pages on FreeBSD.
+*/
+enum MemoryCapability
+{
+    commit = 0x1,
+    decommit = 0x2,
+    reset = 0x4,
+    hugePages = 0x8,
+    sharedMapping = 0x10
+};
```
CyberShadow: Perhaps there should be some basic descriptions for the enum members. What's `reset`?

src/core/vmm.d
```diff
@@ -0,0 +1,489 @@
...
+    Only after applying this call is any memory obtained via $(LREF reserve)
+    guaranteed to be accessible.
+
+    Note: This function and $(LREF reserve)
+    set read/write permissions but not execute.
+    The reason is that some OS' would set permissions on reserve
+    while others on commit. Also Posix makes the commit stage almost transparent.
+    Use $(LREF protect) to override permissions as required.
+*/
+bool commit(void[] range)
```
CyberShadow: I'm slightly concerned that the `void[]` parameter (which will match any mutable array) and the nonspecific names may cause name clashes with other modules. Perhaps the function names could be more explicit, e.g. `commitMemory`? It might also make the user code more readable.

src/core/vmm.d
```diff
@@ -0,0 +1,489 @@
...
+    to the nearest multiple of platform's page size.
+*/
+T roundDownToPage(T)(T arg)
+    if (is(T : void*) || is(T : size_t))
+{
+    return cast(T)(cast(size_t)arg & ~(pageSize-1));
+}
+
+//this is not even nearly covering all cases
+//some things are hard to check (how to check that segmentation fault is generated?)
```
CyberShadow: FWIW, it's possible to test whether a page is readable or writable without causing a segmentation fault on Windows using `IsBadRead/WritePtr` and on UNIX [using a `read`/`write` syscall with a pipe](http://stackoverflow.com/questions/7134590/how-to-test-if-an-address-is-readable-in-linux-userspace-app). Perhaps these can be worked into primitives for this module, as well.

CyberShadow:
If the huge pages will not be implemented in the first version of the module, I think they should be removed from the interface. They can be added back when the implementation is ready, right?

src/core/vmm.d
DmitryOlshansky: I don't expect lots of users to import this module to begin with. Still maybe we should go ahead and append Memory to all of the names.

src/core/vmm.d
DmitryOlshansky: reset is a soft version of uncommit except that pages are not returned to the OS immediately , in fact if you commit them again they may happen to be the same pages with same content.

CyberShadow: Thanks for the explanation, I saw that the `reset` function is documented below. Anyway, that remark was from the point of view of someone reading the documentation. Perhaps instead of repeating the descriptions for the operations here, there could simply be a comment mentioning that the operation descriptions are available in the documentation of the functions themselves.

src/core/vmm.d
```diff
@@ -0,0 +1,489 @@
...
+    $(D base) specifies the desired base address of the block, $(D null)
+    means any base address.
+
+    To release a block of memory allocated by this function use $(LREF free).
+
+    Note: This operation commits virtual memory.
+    To have fine grained control over reserve and commit phases
+    use $(LREF reserve) and $(LREF commit).
+*/
+void[] allocate(void* base, size_t size,
```
WalterBright: Meta comment: no `Returns:` documentation blocks anywhere. No indication of what happens when the OS function call fails. No ability to determine why it failed (when it does fail).

WalterBright: No `Params:` blocks, either.

src/core/vmm.d
```diff
@@ -0,0 +1,489 @@
...
+        return mlock(ptr, size) == 0;
+    }
+
+    bool osUnlockPages(void* ptr, size_t size)
+    {
+        return munlock(ptr, size) == 0;
+    }
+}
+else
+    static assert(false);
```
WalterBright: good!

ZombineDev:
@DmitryOlshansky  with the recent interest in improving and/or rewriting the GC, I think now is a great time to get this merged. @MartinNowak does this look good to you? AFAIR, there were only documentation nits to address. In the interest of time, I can open a PR to address them after the PR is merged (if there's anything left to do).

DmitryOlshansky:
> @DmitryOlshansky with the recent interest in improving and/or rewriting the GC, I think now is a great time to get this merged. @MartinNowak does this look good to you? AFAIR, there were only documentation nits to address. In the interest of time, I can open a PR to address them after the PR is merged (if there's anything left to do).

I suggest we don't rush this in just yet. In fact I'd first evaluate the needs for VMM in my upcoming GC to test this proposal.

src/core/vmm.d
nemanja-boric-sociomantic: Ping on this one - `posix_memadvise` returns error code, which you can't get if `osDecommit` fails (you get just the information that it failed, `errno` is not set - perhaps you could set it here?

DmitryOlshansky: Well, I think it's clear by now that the module should return some kind of ErrorCode instead of bool on most operations. The tricky one is allocate though that returns `void[]`.

ZombineDev:
Alright, sounds good.

src/core/vmm.d
nemanja-boric-sociomantic: Yeah, sorry, I got under impression that my review comment (under different alias) was lost in the noise :-) Yeah, I'm afraid I have no good solution for the `allocate` - you either pass slice as `ref` or the error code as `out`, but that strikes me as highly verbose. I will be interested to see with what you did come out in the end.

DmitryOlshansky:
> I suggest we don't rush this in just yet. In fact I'd first evaluate the needs for VMM in my upcoming GC to test this proposal.

Speaking of which I will defientely need portable shared memory mappings as well.

UplinkCoder:
> I suggest we don't rush this in just yet. In fact I'd first evaluate the needs for VMM in my upcoming GC to test this proposal.

@DmitryOlshansky I need vmm in druntime for the `newCTFE`-JIT.

Therefore I'd be quite happy to see this merged as soon it's ready (not before of course).

DmitryOlshansky:
This needs more work and will likely be revived once I catch up with my GC plans.

PR #1550, opened 2016-04-26
Fix Issue 15958 - Missing extern(Windows) of core.sys.windows functions

qchikara:
added some extern (Windows).


CyberShadow:
LGTM

DmitryOlshansky:
LGTM

DmitryOlshansky:
Auto-merge toggled on

denis-sh:
What is the reason to not add `extern(Windows):` at top of each module and instead search all the modules for function declarations?

CyberShadow:
I think one reason is that C macros translated as D functions shouldn't be `extern(Windows)`.

denis-sh:
> I think one reason is that C macros translated as D functions shouldn't be extern(Windows).

It's a minor case and `extern(D)` will help here. Also I don't see a strong reason even for `extern(D)` as there is nothing wrong in being `extern(Windows)` for *D* functions.

So looks like it's just an incorrect design.

CyberShadow:
> Also I don't see a strong reason even for extern(D) as there is nothing wrong in being extern(Windows) for D functions.

It will pollute the global symbol namespace, perhaps even cause collisions. The calling convention is also less efficient. I'm not sure how it will affect inlining.

> So looks like it's just an incorrect design.

You're welcome to have a go at it and see what breaks as a result.

PR #1552, opened 2016-04-29
Fix Issue 15959 - core.sys.windows modules should be modified for x64

qchikara:
Fix Win32-x64 DataTypes

Changed such as DWORD into DWORD_PTR, and so on.

These Data Types are defined separately between x86/x64,
so existing Win32-x86 programs get _no_ change.


rainers:
Thanks, looks great. Did you verify these systematically or just corrected functions that failed for your programs?

rainers:
Auto-merge toggled on

qchikara:
Sorry that NOT verfied systematically.
But some of them have been woking well in my application.
The others are small fixes and listed in MSDN. (as linked from Issue 15959)
So I trust.

PR #1554, opened 2016-04-30
core.stdc.fenv: Add platform-specific FE_XXX flags

ibuclaw:
It has been a pet peeve of mine for a while that these are defined as a different name in `std.math`, but not reflected here.

### DIFF
--- a/src/core/stdc/fenv.d
+++ b/src/core/stdc/fenv.d
@@ -321,20 +321,253 @@ version( CRuntime_Microsoft )
 }
 else
 {
-    enum
+    version (X86)
     {
-        FE_INVALID      = 1, ///
-        FE_DENORMAL     = 2, /// non-standard
-        FE_DIVBYZERO    = 4, ///
-        FE_OVERFLOW     = 8, ///
-        FE_UNDERFLOW    = 0x10, ///
-        FE_INEXACT      = 0x20, ///
-        FE_ALL_EXCEPT   = 0x3F, ///
-        FE_TONEAREST    = 0, ///
-        FE_UPWARD       = 0x800, ///
-        FE_DOWNWARD     = 0x400, ///
-        FE_TOWARDZERO   = 0xC00, ///
+        // Define bits representing the exception.
+        enum
+        {
+            FE_INVALID      = 0x01, ///
+            FE_DENORMAL     = 0x02, /// non-standard
+            FE_DIVBYZERO    = 0x04, ///
+            FE_OVERFLOW     = 0x08, ///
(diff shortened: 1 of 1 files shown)


src/core/stdc/fenv.d
```diff
@@ -265,20 +265,122 @@ version( CRuntime_Microsoft )
...
+            FE_DENORMAL     = 0x02, /// non-standard
+            FE_DIVBYZERO    = 0x04, ///
+            FE_OVERFLOW     = 0x08, ///
+            FE_UNDERFLOW    = 0x10, ///
+            FE_INEXACT      = 0x20, ///
+            FE_ALL_EXCEPT   = 0x3F, ///
+            FE_TONEAREST    = 0, ///
+            FE_DOWNWARD     = 0x400, ///
+            FE_UPWARD       = 0x800, ///
+            FE_TOWARDZERO   = 0xC00, ///
```
wilzbach: Just a dumb question - why do we duplicate this? Can't we combine these two?

src/core/stdc/fenv.d
```diff
@@ -265,20 +265,122 @@ version( CRuntime_Microsoft )
...
+        enum
+        {
+            FE_INVALID      = 0x01, ///
+            FE_DENORMAL     = 0x02, /// non-standard
+            FE_DIVBYZERO    = 0x04, ///
+            FE_OVERFLOW     = 0x08, ///
+            FE_UNDERFLOW    = 0x10, ///
+            FE_INEXACT      = 0x20, ///
+            FE_ALL_EXCEPT   = 0x3F, ///
+            FE_TONEAREST    = 0, ///
```
wilzbach: Nitpick: not aligned with the other statements

src/core/stdc/fenv.d
ibuclaw: The use of `X86_Any` or `PPC_Any` is discouraged.

ibuclaw:
Ping? Anything else apart from nitpicking that needs addressing?

ibuclaw:
Rebased, added more targets.

ZombineDev:
> Error: `dub search dub` failed

@ibuclaw The dub test failed on jenkins because of network issues, which usually go away if you restart the test (I've just done so). If that happens in the future you login on Jenkins with your GitHub account and restart it yourself.

ZombineDev:
(I missed the fact that Jenkins is not required when I added the `auto-merge` label, but given result of the previous Jenkins test, this change is extremely unlikely to cause any problems.)

PR #1555, opened 2016-05-01
fix issue 15975: misaligned TLS on linux main thread

rainers:
might fix spurious crashes on objects collected by the GC, though they are still referenced.

https://issues.dlang.org/show_bug.cgi?id=15975

### DIFF
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -723,6 +723,11 @@ void scanSegments(in ref dl_phdr_info info, DSO* pdso)
             assert(!pdso._tlsSize); // is unique per DSO
             pdso._tlsMod = info.dlpi_tls_modid;
             pdso._tlsSize = phdr.p_memsz;
+
+            // align to multiple of size_t to avoid misaligned scanning
+            // (size is subtracted from TCB address to get base of TLS)
+            immutable mask = size_t.sizeof - 1;
+            pdso._tlsSize = (pdso._tlsSize + mask) & ~mask;
             break;
 
         default:


rainers:
The problem is that the last element is smaller than size_t.sizeof and causes the TLS section to have an odd size (0x329 in my test program according to readelf). The code in getTLSRange does just subtract this size from the base of the TCB, but that does not seem to be what glibc does to initialize it, i.e. aligning. 

Adding a preinitialized value and printing it looks ok after this change, too.

rainers:
> If the TLS storage does not respect the content alignment (as in the Bugzilla code snippet), how does modifying the scanning range help at all?

Ouch.Trying to avoid having to go back to the broken version, I faked the result of the printf from memory and srewed it completely. I actually printed the scan range and it was badly aligned, not the variable itself. Sorry. I'll update the bug report.

src/rt/sections_elf_shared.d
```diff
@@ -723,6 +723,11 @@ void scanSegments(in ref dl_phdr_info info, DSO* pdso)
             assert(!pdso._tlsSize); // is unique per DSO
             pdso._tlsMod = info.dlpi_tls_modid;
             pdso._tlsSize = phdr.p_memsz;
+
+            // align to multiple of size_t to avoid misaligned scanning
+            // (size is subtracted from TCB address to get base of TLS)
+            immutable mask = size_t.sizeof - 1;
+            pdso._tlsSize = (pdso._tlsSize + mask) & ~mask;
```
MartinNowak: But that's just the size, not the beginning of the array. See https://github.com/rainers/druntime/blob/a000c499f477439e1cacbd12b30d7eadd1fea19d/src/rt/sections_elf_shared.d#L969
Any idea how this happens? First few bytes are bools?

rainers: Ah sorry, I wasn't aware that the patch is only necessary for the LDC branch due to this optimization that is not in upstream: https://github.com/ldc-developers/druntime/blob/ldc/src/rt/sections_elf_shared.d#L996
I'll close this and resubmit against LDC's fork.

PR #1556, opened 2016-05-01
fix issue 15976: swap out explicit TLS values during impersonating thread

rainers:
This used to cause trouble with LDC built upon older druntime versions, but can also happen for user code.

### DIFF
--- a/src/core/sys/windows/threadaux.d
+++ b/src/core/sys/windows/threadaux.d
@@ -285,8 +285,39 @@ struct thread_aux
             return;
 
         curteb[11] = tlsarray;
+
+        // swap out the TLS slots aswell
+        version(Win64)
+        {
+            enum TEB_offset_TlsSlots = 0x1480;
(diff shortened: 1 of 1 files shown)


src/core/sys/windows/threadaux.d
```diff
@@ -285,8 +285,39 @@ struct thread_aux
...
         dg();
+
         curteb[11] = curtlsarray;
+
+        // copy the TLS slots back in case they have been changed in dg
+        memcpy(tlsSlotsAdr(teb), tlsSlotsAdr(curteb), slots.sizeof);
+        tlsExpansionSlots(teb) = tlsExpansionSlots(curteb);
+
+        memcpy(tlsSlotsAdr(curteb), slots.ptr, slots.sizeof);
+        tlsExpansionSlots(curteb) = extraSlots;
```
MartinNowak: The delegate should be made nothrow, or we need a `scope(exit)`.

MartinNowak:
Auto-merge toggled on

PR #1557, opened 2016-05-02
Fix minor memory leak shown in valgrind.

GilesBathgate:
When debugging a trivial program using the memory leak profiler valgrind the d runtime leaks 88bytes of memory. While this is a small amount that the os will cleanup anyway on program shutdown, it is a bug nonetheless and is misleading when looking for real memory leaks in larger programs. The cause of the issue is that the main thread is not garbage collected so special care has to be taken to ensure its resources are freed.

The trivial program is as follows

``` d
void main() { }
```

This produces the following output when run with `valgrind --leak-check=full ./memtest`

```
==11350== Memcheck, a memory error detector
==11350== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==11350== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==11350== Command: ./memtest
==11350== 
==11350== 
==11350== HEAP SUMMARY:
==11350==     in use at exit: 88 bytes in 2 blocks
==11350==   total heap usage: 27 allocs, 25 frees, 45,320 bytes allocated
==11350== 
==11350== 16 bytes in 1 blocks are definitely lost in loss record 1 of 2
==11350==    at 0x4C2BBCF: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==11350==    by 0x41326A: _D2rt5tlsgc4initFZPv (in /home/giles/dlanguage/memtest)
==11350==    by 0x4032C7: thread_attachThis (in /home/giles/dlanguage/memtest)
==11350==    by 0x4031C6: thread_init (in /home/giles/dlanguage/memtest)
==11350==    by 0x40C72F: gc_init (in /home/giles/dlanguage/memtest)
==11350==    by 0x401F80: rt_init (in /home/giles/dlanguage/memtest)
==11350==    by 0x402355: _d_run_main (in /home/giles/dlanguage/memtest)
==11350==    by 0x401F37: main (in /home/giles/dlanguage/memtest)
==11350== 
==11350== 72 bytes in 1 blocks are definitely lost in loss record 2 of 2
==11350==    at 0x4C2DB95: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==11350==    by 0x40CF53: _D2rt8monitor_13ensureMonitorFNbC6ObjectZPOS2rt8monitor_7Monitor (in /home/giles/dlanguage/memtest)
==11350==    by 0x40D095: _d_monitorenter (in /home/giles/dlanguage/memtest)
==11350==    by 0x4034F7: thread_joinAll (in /home/giles/dlanguage/memtest)
==11350==    by 0x4020FF: rt_term (in /home/giles/dlanguage/memtest)
==11350==    by 0x40238F: _d_run_main (in /home/giles/dlanguage/memtest)
==11350==    by 0x401F37: main (in /home/giles/dlanguage/memtest)
==11350== 
==11350== LEAK SUMMARY:
==11350==    definitely lost: 88 bytes in 2 blocks
==11350==    indirectly lost: 0 bytes in 0 blocks
==11350==      possibly lost: 0 bytes in 0 blocks
==11350==    still reachable: 0 bytes in 0 blocks
==11350==         suppressed: 0 bytes in 0 blocks
==11350== 
==11350== For counts of detected and suppressed errors, rerun with: -v
==11350== ERROR SUMMARY: 2 errors from 2 contexts (suppressed: 0 from 0)
```

The 16byte block is lost because on `thread_init` the bytes are assigned by `rt_tlsgc_init`, but since the main thread is created with the GC disabled, the destructor of the main Thread is never called, and so `rt_tlsgc_destroy` doesn't `free()`. A solution is to call free directly in `thread_term`.

The 72byte block is lost because in `thread_joinAll` a call is made to see if `t.isDaemon`, this property wraps `m_isDaemon` in a synchronize block. However the synchronize block requires the GC to ensure that the Mutex primitives are cleaned up via `_d_monitordelete`, again because the thread was created with the GC disabled this never happens. A solution is to read `t.m_isDaemon` directly. The rational is that a) all writes to m_isDaemon were via the member and not the synchronized property anyway, and b) There appears to be a global slock in thread_joinAll anyway.

These are my proposed fixes, I have little idea what the impact will be due to inexperience with the d runtime.

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -2032,6 +2032,7 @@ extern (C) void thread_term()
         free(Thread.pAboutToStart);
         Thread.pAboutToStart = null;
     }
+    free(Thread.sm_main.m_tlsgcdata);
     Thread.termLocks();
 }
 
@@ -2293,7 +2294,7 @@ extern (C) void thread_joinAll()
             Thread.remove(t);
             t = tn;
         }
-        else if (t.isDaemon)
+        else if (t.m_isDaemon)
         {
             t = t.next;
         }


GilesBathgate:
I believe the underlying issue was partly the cause of this SO question: https://stackoverflow.com/questions/14802212/memory-leak-in-minimal-d-program

GilesBathgate:
I have also noticed people use valgrind suppression to work around this issue: https://github.com/PhilippeSigaud/Pegged/blob/master/pegged.valgrind

GilesBathgate:
Fixes the bugzilla issue 6830: https://issues.dlang.org/show_bug.cgi?id=6830

src/core/thread.d
```diff
@@ -2032,6 +2032,7 @@ extern (C) void thread_term()
         free(Thread.pAboutToStart);
         Thread.pAboutToStart = null;
     }
+    free(Thread.sm_main.m_tlsgcdata);
```
MartinNowak: Please use `rt_tls_destroy` instead, there is more to free.

src/core/thread.d
```diff
@@ -2293,7 +2294,7 @@ extern (C) void thread_joinAll()
             Thread.remove(t);
             t = tn;
         }
-        else if (t.isDaemon)
```
MartinNowak: Now the check whether sth. is a daemon thread is no longer synchronized w/ setting this flag. But in fact the old synchronization wasn't correct either. It's Thread.join that must synchronize w/ daemonizing a thread, in order to eliminate any chance for join to wait infinitely on a daemon thread.

MartinNowak: I think daemon should become an argument to thread start and should not be changed afterwards. But that would be quite some change.

src/core/thread.d
GilesBathgate: Unfortunately:

```bash
giles@tikka:~/ldc/runtime/druntime$ git diff
diff --git a/src/core/thread.d b/src/core/thread.d
index 4fe05e2..2bebc4b 100644
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -2018,6 +2018,7 @@ extern (C) void thread_term()
         free(Thread.pAboutToStart);
         Thread.pAboutToStart = null;
     }
+    rt_tlsgc_destroy(Thread.sm_main.m_tlsgcdata);
     Thread.termLocks();
 }
giles@tikka:~/ldc/runtime/druntime$ cd ~/dlanguage/
giles@tikka:~/dlanguage$ cat memtest.d
void main() { }
giles@tikka:~/dlanguage$ ldc2 memtest.d
giles@tikka:~/dlanguage$ ./memtest 
Segmentation fault (core dumped)
giles@tikka:~/dlanguage$
```

src/core/thread.d
GilesBathgate: Perhaps the synchronization can be done with something else that doesn't rely on the GC to clean up the Mutex objects.

UplinkCoder:
what is the status on this ?

GilesBathgate:
@UplinkCoder The patch works to the extent that it fixes the issue in my local tests. I was trying to justify my changes to @MartinNowak, and explain that what might be considered more clean solutions (in terms of code symmetry) do not appear to work.

BBasile:
ping, even it's only a constant leak .Can't you fix this at last ?

GilesBathgate:
@BBasile I don't know what is expected of me here. The patch works.

I was told to make various amendments by @MartinNowak but they don't work.

JackStouffer:
@GilesBathgate Try sending him an email reminding him

BBasile:
If the fix is okay then LGTM.

BBasile:
DUnrtime members, you can close now i think. See #1857.

PR #1558, opened 2016-05-04
Fix Issue 15987 - core.sys.windows.msacm remains pseudo definitions

qchikara:
modified constants for the buffer sizes used from some structs,
whiche have been pseudo value(256).


MartinNowak:
Where did you got the real values from? Any online reference?

qchikara:
I got them from msacm.h in Windows SDK included with Visual Studio 2015.
I am verifying the sizes of structures in windows headers.

MartinNowak:
Auto-merge toggled on

PR #1564, opened 2016-05-07
remove old tango imports

wilzbach:
I realized that there were still two imports to Tango in our codebase. I assume they ended up by mistake?


DmitryOlshansky:
Auto-merge toggled on

PR #1567, opened 2016-05-08
Solaris: fix misprint in ld_start64.

nykytenko:


MartinNowak:
Auto-merge toggled on

PR #1568, opened 2016-05-09
Add documentation to GC struct

Jebbs:
Documentation was pulled from the core.memory module since there is a direct relationship between that module and this struct.

### DIFF
--- a/src/gc/gc.d
+++ b/src/gc/gc.d
@@ -269,7 +269,12 @@ struct GC
     static gcLock = shared(AlignedSpinLock)(SpinLock.Contention.lengthy);
     static bool inFinalizer;
 
-    // lock GC, throw InvalidMemoryOperationError on recursive locking during finalization
+    /*
+     * Lock the GC.
+     *
+     * Throws:
+     *  InvalidMemoryOperationError on recursive locking during finalization.
+     */
     static void lockNR() @nogc nothrow
     {
         if (inFinalizer)
@@ -279,6 +284,13 @@ struct GC
 
     __gshared Config config;
 
+
+    /*
+     * Initialize the GC based on command line configuration.
+     *
+     * Throws:
+     *  OutOfMemoryError if failed to initialize GC due to not enough memory.
+     */
     void initialize()
     {
         config.initialize();
@@ -297,12 +309,6 @@ struct GC
 
     void Dtor()
     {
-        version (linux)
-        {
-            //debug(PRINTF) printf("Thread %x ", pthread_self());
-            //debug(PRINTF) printf("GC.Dtor()\n");
-        }
-
         if (gcx)
         {
             gcx.Dtor();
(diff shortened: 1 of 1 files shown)


Jebbs:
I figured there was no harm in this since I was going through and trying to understand the code better anyway. I'll be adding more documentation to this file soon.

src/gc/gc.d
```diff
@@ -525,7 +608,25 @@ struct GC
 
 
     /**
+     * Requests an aligned block of managed memory from the garbage collector,
+     * which is initialized with all bits set to zero.  This memory may be
+     * deleted at will with a call to free, or it may be discarded and cleaned
+     * up automatically during a collection run.  If allocation fails, this
+     * function will call onOutOfMemory which is expected to throw an
+     * OutOfMemoryError.
```
MartinNowak: This should reference malloc instead of repeating the same text.

MartinNowak:
I'm not entirely convinced this is a good idea. This is user facing documentation, where the GC needs implementation specific docs.

src/gc/gc.d
```diff
@@ -432,6 +484,8 @@ struct GC
                 p = sentinel_sub(p);
                 auto biti = cast(size_t)(p - pool.baseAddr) >> pool.shiftBy;
 
+                //Based on the documetnation, shouldn't we set the
+                //bits first and then get the new value of the bit field?
```
leandro-lucarella-sociomantic: Is it a good idea to include questions in the code? At least I find it super strange when I find questions like this in the code.

src/gc/gc.d
```diff
@@ -1476,7 +1719,7 @@ struct Gcx
     /**
      *
      */
-    void removeRoot(void *p) nothrow @nogc
+    void removeRoot(void *p) nothrow
```
leandro-lucarella-sociomantic: Why are `@nogc` being removed? I think those were added recently...

Jebbs:
@MartinNowak 
>I'm not entirely convinced this is a good idea. This is user facing documentation, where the GC needs implementation specific docs, e.g. there is no point in telling someone that works w/ the GC about the differences between malloc/calloc/realloc, but an extend function is uncommon.

That's fair. I originally was writing documentation from scratch for this since it felt like a bad idea to not have any, but later noticed that the functions found in this struct are called by their counterparts in core.memory. IMO it should have the same documentation as its core.memory counterparts while Gcx gets the implementation specific documentation. 

I can go back to the documentation I was originally writing though. It was based off of the documentation in core.memory, but was a little bit less formal I guess you could say. It might be more appropriate since it would be more for internal use than for users.

src/gc/gc.d
Jebbs: I probably should have mentioned this when I opened the request, but my intent here was to bring awareness to these various bits of code for the reasons stated in the questions.

I was planning on removing these before anything was actually committed, but now that I think about it it should probably have been brought up in a different PR or something.

src/gc/gc.d
Jebbs: It must have gotten removed on accident, which is strange because I saw the PR that added them in and I was trying to avoid them getting taken out again by mine.

MartinNowak:
Let's rebase and get this in.

MartinNowak:
Gah, substituted by #1573. Please close PRs if they're no longer relevant.

PR #1569, opened 2016-05-10
Add core.thread.fork

CyberShadow:
For https://github.com/dlang/phobos/pull/4294

Needs ~~https://github.com/dlang/phobos/pull/4302~~ https://github.com/dlang/phobos/pull/4305

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -5369,3 +5369,29 @@ version (Windows)
 else
 version (Posix)
     alias ThreadID = pthread_t;
+
+version (Posix)
+{
+    private extern (C) void gc_lock() nothrow @nogc;
+    private extern (C) void gc_unlock() nothrow @nogc;
+
+    int fork() //nothrow @nogc
+    {
+        // Acquire the GC lock for ourselves to ensure it is not held by another thread
+        gc_lock();
+        auto id = core.sys.posix.unistd.fork();
+        gc_unlock();
+
+        if (id < 0)
+            return id;
+        if (id == 0)
+        {
+            // Unregister all threads other than self
+            auto thisThread = Thread.getThis();
+            foreach (Thread t; Thread)
+                if (thisThread && t !is thisThread)
+                    thread_detachInstance(t);
+        }
+        return id;
+    }
+}
--- a/src/gc/proxy.d
+++ b/src/gc/proxy.d
@@ -62,6 +62,9 @@ private
             void function(in void[]) gc_runFinalizers;
 
             bool function() gc_inFinalizer;
+
+            void function() nothrow @nogc gc_lock;
+            void function() nothrow @nogc gc_unlock;
         }
     }
 
@@ -100,6 +103,9 @@ private
         pthis.gc_runFinalizers = &gc_runFinalizers;
 
         pthis.gc_inFinalizer = &gc_inFinalizer;
+
+        pthis.gc_lock = &gc_lock;
+        pthis.gc_unlock = &gc_unlock;
     }
 }
 
@@ -316,6 +322,20 @@ extern (C)
         if( proxy is null )
             return _gc.inFinalizer;
         return proxy.gc_inFinalizer();
+    }
+
+    void gc_lock() nothrow @nogc
+    {
+        if( proxy is null )
+            return _gc.lock();
+        return proxy.gc_lock();
+    }
+
+    void gc_unlock() nothrow @nogc
+    {
+        if( proxy is null )
+            return _gc.unlock();
+        return proxy.gc_unlock();
     }
 
     Proxy* gc_getProxy() nothrow
--- a/src/gc/gc.d
+++ b/src/gc/gc.d
@@ -338,6 +338,16 @@ struct GC
         runLocked!(go, otherTime, numOthers)(gcx);
     }
 
+    void lock() nothrow @nogc
+    {
+        gcLock.lock();
+    }
+
+    void unlock() nothrow @nogc
+    {
+        gcLock.unlock();
+    }
+
     auto runLocked(alias func, Args...)(auto ref Args args)
     {
         debug(PROFILE_API) immutable tm = (GC.config.profile > 1 ? currTime.ticks : 0);


rainers:
I'm not a fan of exposing the lock. This could make optimizations like lock-free allocation a lot more complicated or even impossible.

Not sure whether this has been discussed elsewhere, but why do you care for the GC lock, but not other locks e.g. inside malloc?

leandro-lucarella-sociomantic:
Mmm, this is leaking way too much implementation detail. I think it would be better to include a `gc_fork()` or `gc_pre_fork()`/`gc_post_fork()`. I wonder if even [pthread_atfork](http://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_atfork.html) could be used.

CyberShadow:
> Not sure whether this has been discussed elsewhere, but why do you care for the GC lock, but not other locks e.g. inside malloc?

Actually I think glibc's `fork` will do the same thing for malloc's lock as we do with the GC lock. But also I documented this limitation in the Phobos PR.

> I think it would be better to include a gc_fork() or gc_pre_fork()/gc_post_fork().

OK, I could change it that way. It could also be changed to use undocumented C functions.

leandro-lucarella-sociomantic:
> I wonder if even pthread_atfork could be used.

I actually wonder if pthread_atfork **should** be used, as the locks in the child process should have to be re-created to be posix compliant (IIRC).

schveiguy:
What about exposing GC.runLocked instead? https://github.com/dlang/druntime/blob/master/src/gc/gc.d#L341

CyberShadow:
Using scopes and delegates makes me a bit uncomfortable, it is too easy to trigger unwanted behavior such as memory allocation or exception handling when dealing with such a low-level facility. This function also goes via an `extern(C)` interface, it seems unnatural to pass a D delegate through it.

CyberShadow:
OK, I removed the public lock/unlock GC interface and instead moved most of the `fork` implementation from Phobos to Druntime. This allows keeping the Druntime-specific `fork` code a Druntime implementation detail, and avoid exposing GC internals.

CyberShadow:
Merge https://github.com/dlang/phobos/pull/4302 first to fix autotester failures for this one.

src/core/thread.d
```diff
@@ -5369,3 +5369,29 @@ version (Windows)
 else
 version (Posix)
     alias ThreadID = pthread_t;
+
+version (Posix)
+{
+    extern (C) void gc_lock() nothrow @nogc;
+    extern (C) void gc_unlock() nothrow @nogc;
```
schveiguy: These should be private to avoid public access, no?

src/gc/proxy.d
```diff
@@ -318,6 +321,20 @@ extern (C)
...
+            return _gc.lock();
+        return proxy.gc_lock();
+    }
+
+    void gc_unlock() nothrow @nogc
+    {
+        if( proxy is null )
+            return _gc.unlock();
+        return proxy.gc_unlock();
+    }
```
schveiguy: Existing proxies that don't have functions for these calls will cause a crash, right? Or is it required to implement these for compilation?

src/core/thread.d
CyberShadow: Yep, thanks

src/gc/proxy.d
CyberShadow: Hmm...

I thought this was used exclusively for connecting multiple Druntime instances (from multiple shared objects). Is this also used for third-party GC implementations? ( @rainers , @leandro-lucarella-sociomantic )?

If it is only used for the former, then it should be a non-issue, because AFAIK we don't consider the Druntime ABI binary-stable, and mixing different Druntime versions in one application is unsupported.

Still, if I understand the intended use of this proxy thing correctly, this code does need to be here. When we are in a shared object and forking the program, we do need to acquire the exclusive lock of the main program's GC instance.

schveiguy: Heh, I guess this is fixed if you update initProxy :)

CyberShadow: Thanks :)

CyberShadow:
D'oh, I missed one: https://github.com/dlang/phobos/pull/4305

src/gc/proxy.d
leandro-lucarella-sociomantic: I never understood the proxy thing of the GC, I completely got rid of it in CDGC :P

leandro-lucarella-sociomantic:
What about https://github.com/dlang/druntime/pull/1569#issuecomment-218131386?

leandro-lucarella-sociomantic:
I think [pthread_atfork rationale](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_atfork.html#tag_16_402_08) is describing exactly this situation...

CyberShadow:
Yeah, that does sound like the way to go. Though, I am a bit concerned that this would be a global change, and might interfere with existing applications using `fork` in some way or another. I am getting out of my depth here.

Using `pthread_atfork` does have the advantage of making the Druntime and Phobos changes orthogonal - Druntime would protect itself against all usages of the C `fork` call, and Phobos would just provide a `Pid`-returning wrapper.

src/gc/proxy.d
rainers: > I thought this was used exclusively for connecting multiple Druntime instances (from multiple shared objects). Is this also used for third-party GC implementations?

IMO, the proxy never was good enough for its proposed use, i.e. dynamic linking against another runtie (because of other objects not being shared). 
It could be used to swap out the GC implementation if we wouldn't compile it together with the rest of druntime. Right now, inlining binds it to the current implementation.

rainers: If you add functions to gc/proxy.d, you should also do this in gcstub/gc.d.

rainers:
`pthread_atfork` looks preferrable to me, too. I'm not a unix user, though, so I don't have any experience with it.

CyberShadow:
From IRC:

```
[2016-05-12 01:18:52] [w_w] CyberShadow: I think there are other druntime locks you need to acquire in fork() (e.g. slock, maybe criticalRegionLock but I'm not familiar with that code)
[2016-05-12 01:19:09] [w_w] 100% sure no threads have druntime locks
[2016-05-12 01:19:13] [w_w] *to be
```

leandro-lucarella-sociomantic:
> From IRC:

This is why I think every lock should protect itself against forking(). Forking with threading is always a mess. :-/

CyberShadow:
The way forward seems to be:

- Add fork to Phobos, which simply calls the C `fork` function.
- Add fork handling to Druntime, which hooks into `atfork`.

These two are mutually independent.
I no longer have a use case, so I'm going to close this.

PR #1570, opened 2016-05-10
fix Issue 16007 - Some Win32 API structs has wrong definitions

qchikara:
This is a set of bug fixes in core.sys.windows modules.


qchikara:
I know the autotester failed, and why?
I tried to remove space chars from that empty line.

src/core/sys/windows/winnt.d
```diff
@@ -3316,13 +3325,28 @@ struct IMAGE_LOAD_CONFIG_DIRECTORY64 {
...
 
-struct IMAGE_RUNTIME_FUNCTION_ENTRY {
+version (Win64) {
+    alias IMAGE_LOAD_CONFIG_DIRECTORY64 IMAGE_LOAD_CONFIG_DIRECTORY;
+} else {
+    alias IMAGE_LOAD_CONFIG_DIRECTORY32 IMAGE_LOAD_CONFIG_DIRECTORY;
+}
+alias IMAGE_LOAD_CONFIG_DIRECTORY* PIMAGE_LOAD_CONFIG_DIRECTORY;
+
+struct IMAGE_ALPHA_RUNTIME_FUNCTION_ENTRY { // for Alpha...need?
```
rainers: I think we can just replace it with the correct definition. Having an Alpha declaration, but no Alpha64 or ARM declaration seems pointless.

rainers:
Otherwise LGTM

qchikara:
Fixed the IMAGE_RUNTIME_FUNCTION_ENTRY as @rainers mentiond,
I rejected useless Alpha version and added a comment.

d-random-contributor:
Doesn't Windows run on ARM?

qchikara:
Probably it would require more and more to do. It is not my business.
My that comment meant that I've rejected accidentaly merged one.

qchikara:
Rebased.

ZombineDev:
@d-random-contributor I think developing for Windows on ARM is officially only possible by tageting Windows Store / WinRT which is a very different API. Access to the "old" Win32 API is not allowed, but also not [impossible](https://surfsec.wordpress.com/2013/01/06/circumventing-windows-rts-code-integrity-mechanism/).
Anyway, I think it's better to focus WinRT, but that's a topic for another discussion.

rainers:
Auto-merge toggled on

PR #1573, opened 2016-05-12
Add documentation to GC struct (Alt)

Jebbs:
This is an alternative PR to #1568.

This pull request doesn't always use the user facing documentation found in core.memory, but does where applicable(the documentation is small). Otherwise I wrote a small tidbit for the function along with Params, Returns, and Throws sections for consistency. I tried to not go into too much detail as more of the actual implementation is in Gcx and should be documented there. 

I also deleted a bunch of commented out debug statements that don't add anything to the code and should be removed.

### DIFF
--- a/src/core/internal/gc/impl/conservative/gc.d
+++ b/src/core/internal/gc/impl/conservative/gc.d
@@ -450,7 +450,6 @@ class ConservativeGC : GC
      * Params:
      *  size = The desired allocation size in bytes.
      *  bits = A bitmask of the attributes to set on this block.
-     *  alloc_size = The actuall size allocated in bytes.
      *  ti = TypeInfo to describe the memory.
      *
      * Returns:
@@ -459,7 +458,7 @@ class ConservativeGC : GC
      * Throws:
      *  OutOfMemoryError on allocation failure
      */
-    void *malloc(size_t size, uint bits = 0, size_t *alloc_size = null, const TypeInfo ti = null) nothrow
+    void *malloc(size_t size, uint bits = 0, const TypeInfo ti = null) nothrow
     {
         if (!size)
         {
@@ -538,7 +537,6 @@ class ConservativeGC : GC
      * Params:
      *  size = The desired allocation size in bytes.
      *  bits = A bitmask of the attributes to set on this block.
-     *  alloc_size = The actuall size allocated in bytes.
      *  ti = TypeInfo to describe the memory.
      *
      * Returns:
@@ -547,7 +545,7 @@ class ConservativeGC : GC
      * Throws:
      *  OutOfMemoryError on allocation failure.
      */
-    void *calloc(size_t size, uint bits = 0, size_t *alloc_size = null, const TypeInfo ti = null) nothrow
+    void *calloc(size_t size, uint bits = 0, const TypeInfo ti = null) nothrow
     {
         if (!size)
         {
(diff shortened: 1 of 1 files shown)


src/gc/gc.d
```diff
@@ -384,7 +407,15 @@ struct GC
     }
 
     /**
+     * Returns a bit field representing all block attributes set for the memory
+     * referenced by p.
      *
+     * Params:
+     *  p = A pointer to the root of a valid memory block or to null.
```
rainers: Please use `base` instead of `root` here and in the following comments. It might get confusing when talking about root pointers to life objects, e.g. in `addRoot`.

src/gc/gc.d
```diff
@@ -551,7 +626,25 @@ struct GC
...
+     *  size = The desired allocation size in bytes.
+     *  bits = A bitmask of the attributes to set on this block.
+     *  alloc_size = The actuall size allocated in bytes.
+     *  ti = TypeInfo to describe the memory.
+     *
+     * Returns:
+     *  A reference to the allocated memory on success or null if size is
+     *  zero.  On failure, the original value of p is returned.
```
rainers: >  On failure, the original value of p is returned.

No. It just crashes if you pass in a non-GC pointer and a size, and throws OutOfMemoryError if allocation fails.

src/gc/gc.d
```diff
@@ -1090,7 +1228,7 @@ struct GC
 
 
     /**
-     *
+     * Returns an interator allowing roots to be traversed via a foreach loop.
```
rainers: `interator` -> `iterator`

src/gc/gc.d
```diff
@@ -498,15 +563,12 @@ struct GC
...
 
     //
-    //
+    // Implementation for malloc and calloc.
     //
     private void *mallocNoSync(size_t size, uint bits, ref size_t alloc_size, const TypeInfo ti = null) nothrow
     {
         assert(size != 0);
-
-        //debug(PRINTF) printf("GC::malloc(size = %d, gcx = %p)\n", size, gcx);
```
rainers: I'd recommend keeping the commented logging of modifying API calls, they can be helpful when trying to find bugs. Some need to be corrected, though.

rainers:
Apart from nitpicks, looks good. I like this version better than #1568.

src/gc/gc.d
Jebbs: >If p references memory not originally allocated by this garbage collector, or if it points to the interior of a memory block, no action will be taken.

This is from the user facing documentation. Does that mean that should be changed as well?

src/gc/gc.d
Jebbs: Whoops!

src/gc/gc.d
rainers: > This is from the user facing documentation. Does that mean that should be changed as well?

Not sure. Implementation and documenttion don't match, but I don't think realloc is well defined to begin with.

> On failure, the original value of p is returned.

If "failure" means an "invalid" pointer is passed, this is not a good return value, as it is the same if the memory has been successfully resized in place. "invalid" pointers are not detected at the moment.

> Throws: OutOfMemoryError on allocation failure.

This is the only current failure.

In addition I don't like that reallocating with size 0 frees the memory immediately, but leaves it for garbage collection for other sizes. This will be surprising in generic code.

Jebbs:
Not sure if anything else should be done for the realloc documentation, but everything else was fixed.

MartinNowak:
Rebase and go.

RazvanN7 (approved):

PR #1574, opened 2016-05-13
Fix issue 15959 additional fixes

qchikara:
https://issues.dlang.org/show_bug.cgi?id=15959

A number of simple replaces,
which are required by x64 Win32 programs.

Mostly in this form:
DWORD/int/uint/ -> DWORD_PTR/ULONG_PTR/INT_PTR

This should make no change in x86 programs.


qchikara:
I've got this by boring repetition of greps. (sigh)

qchikara:
Oops.
> which requires for x64 Win32 programs.
===> which are required by x64 Win32 programs.

qchikara:
rebased

rainers:
I haven't verified every change, but I trust you on these. Thanks for doing this.

rainers:
Auto-merge toggled on

PR #1575, opened 2016-05-18
More fix of issue 15959 (size_t)

qchikara:
These are simple replaces,
which are required by x64 Win32 programs.

Mostly in this form:
DWORD/int/uint/ULONG -> SIZE_T

This should make no change in x86 programs.

### DIFF
--- a/src/core/sys/windows/dll.d
+++ b/src/core/sys/windows/dll.d
@@ -241,6 +241,7 @@ struct dll_aux
     }
 
     // the following structures can be found here: http://undocumented.ntinternals.net/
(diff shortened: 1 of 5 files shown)


qchikara:
Added mswsock.d, which was fixed in my local branch days ago :P.

schveiguy:
This certainly seems reasonable. But I'm not familiar enough with how the internals of windows work to know if this is correct. I tried searching for these definitions on my Windows installation, can't find them. MSDN matches your updates, so I'd say LGTM based on that. I guess we can always undo if this breaks things :)

DmitryOlshansky:
Auto-merge toggled on

PR #1576, opened 2016-05-20
Fix issue 16049 - core.sys.windows structs have wrong sizes and aligns

qchikara:
https://issues.dlang.org/show_bug.cgi?id=16049

A set of small fixes of Win32 structs.

- Add/Remove align specifiers
- Paddings under some conditions
- Helper macro 'AlignedStr' to switch alignment between x64/x86 environment

I detect and verify these correct sizes by  comparing with ones of Windows SDK in MSVC.

### DIFF
--- a/src/core/sys/windows/basetsd.d
+++ b/src/core/sys/windows/basetsd.d
@@ -49,6 +49,20 @@ package template DECLARE_HANDLE(string name, base = HANDLE) {
 }
 alias HANDLE* PHANDLE, LPHANDLE;
 
+// helper for aligned structs
+// alignVal 0 means the default align.
+// _alignSpec as parameter does not pollute namespace.
+package mixin template AlignedStr(int alignVal, string name, string memberlist,
+                                    string _alignSpec = !alignVal ? "align" : "align("~alignVal.stringof~")" )
+{
+    mixin( _alignSpec ~ " struct " ~ name ~" { " ~ _alignSpec ~":"~ memberlist~" }" );
+}
+
+version (unittest) {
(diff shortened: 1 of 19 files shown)


qchikara:
Rebased and added a module mmsystem.

CyberShadow:
That's quite a large patch. 

Are all of these changes necessary?

How have you verified that the changes are correct, and how have you verified which changes are necessary?

The way I would solve this is write a D program which iterates over all public struct declarations (`__traits(allMembers, ...)`), and emits a C program that verifies that the D and C struct sizes and field offsets match - and fix things on the D side when there's a mismatch.

qchikara:
> Are all of these changes necessary?

Yes. This is a set of bug fixes.

> How have you verified that the changes are correct, and how have you verified which changes are necessary?

I collected all structs defined in core.sys.windows with my private tool.
Then wrote dump list of sizes of structs[1] in VS2015, and made a checker[2] from it.
And then, [3]updated old struct holding extra members, which can be added by some newer platform.
( you know, first member is typically 'cbSize'=size of struct )
FInally, we can find mismatches in definitions and fix them.
[1][2] attached file in https://issues.dlang.org/show_bug.cgi?id=16049
[3] my custom repo https://github.com/qchikara/druntime/tree/fetch1574

qchikara:
Note that I did not check about offset of each field,
and that do not modify the updated structs -- [3] in above -- for now.

But..., yeah, I try to verify all offsets too.

CyberShadow:
OK, I see you were sufficiently meticulous about it :)

CyberShadow:
Auto-merge toggled on

CyberShadow:
Thanks!

PR #1577, opened 2016-05-23
Set deployment target to the minimum supported version of OS X

jacob-carlborg:
Since the update to use native TLS on OS X the minimum deployment
target is now OS X Lion (10.7).


MartinNowak:
Stable or not?

MartinNowak:
Auto-merge toggled on

jacob-carlborg:
Not sure what you mean with "stable", the "stable" branch? It should be merged into whatever branch contains the changes to use native TLS.

jacob-carlborg:
There's a PR for Phobos as well. https://github.com/dlang/phobos/pull/4347

PR #1578, opened 2016-05-26
remove the deprecated wiki macros

wilzbach:
WIKI is no longer used

> These tags are back from the days when the official D wiki was http://www.prowiki.org/wiki4d/wiki.cgi.

See also

https://github.com/dlang/phobos/pull/4369
https://github.com/dlang/dlang.org/pull/1310

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2,8 +2,6 @@
  * Forms the symbols available to all D programs. Includes Object, which is
  * the root of the class object hierarchy.  This module is implicitly
  * imported.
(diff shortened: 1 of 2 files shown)


aG0aep6G:
Looks good to me.

wilzbach:
@klickverbot I honestly don't know (and can't dig deepter atm), but I filled a bug, s.t. it doesn't get lost

https://issues.dlang.org/show_bug.cgi?id=16101

PR #1579, opened 2016-05-28
core.sys.linux.sys.inotify: Fix typo in identifier

CyberShadow:
This typo seems to be D specific, coming all the way from Tango:
https://www.google.com/search?q=%22IN_UMOUNT%22

### DIFF
--- a/src/core/sys/linux/sys/inotify.d
+++ b/src/core/sys/linux/sys/inotify.d
@@ -34,7 +34,7 @@ enum: uint
     IN_DELETE        = 0x00000200,
     IN_DELETE_SELF   = 0x00000400,
     IN_MOVE_SELF     = 0x00000800,
-    IN_UMOUNT        = 0x00002000,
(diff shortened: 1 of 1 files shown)


src/core/sys/linux/sys/inotify.d
```diff
@@ -48,6 +48,9 @@ enum: uint
     IN_ALL_EVENTS    = 0x80000FFF,
 }
 
+// Old typo, preserved for compatibility
+enum IN_UMOUNT = IN_UNMOUNT;
```
schuetzm: How about deprecating it?

CyberShadow: Maybe later. We need at least one release where you can use either an old feature or its replacement without any warnings.

DmitryOlshansky:
LGTM

DmitryOlshansky:
Auto-merge toggled on

PR #1581, opened 2016-06-01
Allow GC implementations to be swappable.

Jebbs:
This wraps up GC implementations in a proxy so that they can easily be picked and initialized at runtime before the user program starts according to the config options passed or embedded.

This also adds a secondary "GC implementation" to swap to based on gc_stub. This implementation allows memory to be managed in a manual style using 'new' and 'delete' instead of relying on the GC.


Jebbs:
@MartinNowak 
@LightBender

src/gc/proxy.d
```diff
@@ -13,13 +13,16 @@
...
+import malloc = gc.impl.malloc.gc;
+import gc.config;
 import gc.stats;
 import core.stdc.stdlib;
 
 private
 {
-    __gshared GC _gc;
+    __gshared conservative.GC _conservativeGC;
+    __gshared malloc.GC _mallocGC;
```
MartinNowak: Those instances should be in the correspond gc implementation modules.

src/gc/proxy.d
```diff
@@ -62,57 +68,158 @@ private
...
+
+        if (config.malloc)
+        {
+            initMallocGCProxy();
+        }
+        else
+        {
+            _conservativeGC.initialize();
+            initConservativeGCProxy();
+        }
```
MartinNowak: Always call the init functions of all GCs and let them do the right thing based on the config.
```
config.initialize();
conservative.instance.initialize(); // instance being the singleton
malloc.instance.initialize();
```

src/gc/proxy.d
```diff
@@ -13,13 +13,16 @@
  */
 module gc.proxy;
 
-import gc.gc;
+import conservative = gc.impl.conservative.gc;
+import malloc = gc.impl.malloc.gc;
```
MartinNowak: We shouldn't use that name in the real of memory management, rather call it manual or so.

src/gc/proxy.d
```diff
@@ -62,57 +68,158 @@ private
...
+
         }
     }
 
     __gshared Proxy  pthis;
     __gshared Proxy* proxy;
 
-    void initProxy()
+
+    void initConservativeGCProxy()
```
MartinNowak: Then within `conservative.GC.initialize()` check whether `config.gc == "conservative"`.

src/gc/config.d
```diff
@@ -63,6 +66,7 @@ struct Config
     profile:0|1|2  - enable profiling with summary when terminating program (%d)
     precise:0|1    - enable precise scanning (not implemented yet)
     concurrent:0|1 - enable concurrent collection (not implemented yet)
+    malloc:0|1     - enable a manual management implementation
```
MartinNowak: This idea doesn't scale to link new unknown GCs, let's use a string option `gc:conservative|malloc|concurrent` instead.

MartinNowak: We need a `parse(T:const(char)[])` specialization for that. Just munch an indentifier `[A-Za-z][A-Za-z_0-9]` therin.

src/gc/proxy.d
Jebbs: Proxy struct is currently private. Should it be made public then?

src/gc/config.d
Jebbs: Do you want to only list the implementations available right now (conservative|manual) or include the future ones as well?

LightBender: My opinion would be to only show the currently available options.

rainers:
The proxy struct is emulating a virtual function table. Has it been discussed to make it an interface that each GC variant implements? This would remove all the boilerplate with function literals.

That's actually what I considered the way to go when Martin converted the GC class to a struct.

leandro-lucarella-sociomantic:
Even when I think being able to specify completely different implementations, for the case of manual vs. conservative, I wonder if it doesn't make more sense to just make it a feature, not a full implementation. Is there any practical difference between conservative and manual other than trying a collection before getting more memory when the heap is exhausted?

Jebbs:
@leandro-lucarella-sociomantic 
Well, two big differences would be that no memory is reserved and freed memory is returned to the host instead of sticking around. There is also no collection that happens when the program exits, so this is a completely manual implementation.

leandro-lucarella-sociomantic:
Do you really expect to someone for use that implementation or is it there just an example of a minimal GC implementation?

Jebbs:
A bit of both, really. It probably won't be common, but it is fully usable and there are cases when using a gc would be more than necessary.

Do you have anything against it being accessible?

Jebbs:
Did some updates based on comments. The function I added to config.d to parse string arguments will probably need to be changed though. It works, but I don't know what is expected for this situation.

MartinNowak:
> Do you really expect to someone for use that implementation or is it there just an example of a minimal GC implementation?

That a replacement for gc_stub and one of it's usecases is to run valgrind (if you have enough memory).

src/gc/config.d
```diff
@@ -186,6 +186,47 @@ body
...
+{
+
+    int start;
+    int nscanned;
+
+    for(;start<str.length; start++)
+    {
+        if(!isspace(str[start]))
+            break;
+    }
```
MartinNowak: The options are already space separated, so this will never do anything useful. There were also 2 helpers `skip!isspace` and `find!isspace` to avoid the loop.

MartinNowak: Let's just use the whole str as-is.

src/gc/config.d
```diff
@@ -186,6 +186,47 @@ body
...
+
+    for(nscanned=start; nscanned<str.length; nscanned++)
+    {
+        if(isspace(str[nscanned]))
+            break;
+    }
+
+    switch(str[start .. nscanned])
+    {
+        case "manual":
```
MartinNowak: Let's rather use an if-else-ladder that switch/case, b/c the latter requires quite some runtime machinery, and the runtime isn't yet initialized.

src/gc/impl/manual/gc.d
```diff
@@ -0,0 +1,216 @@
...
+            {
+                ranges[i] = ranges[--nranges];
+                return;
+            }
+        }
+        assert( false );
+    }
+}
+
+
```
MartinNowak: For new code, please use [dfmt](https://github.com/Hackerpilot/dfmt) `dfmt -i file.d` and make sure to read the [The D Style Guide](https://dlang.org/dstyle.html).

src/gc/proxy.d
```diff
@@ -13,106 +13,79 @@
...
-            void    function(void*) gc_free;
-
-            void*   function(void*) gc_addrOf;
-            size_t  function(void*) gc_sizeOf;
-
-            BlkInfo function(void*) gc_query;
-
-            void function(void*) gc_addRoot;
-            void function(void*, size_t, const TypeInfo ti) gc_addRange;
+    __gshared Proxy*  pthis;
```
MartinNowak: What is `pthis` for? You should only need a single GC ever.

src/gc/impl/conservative/gc.d
```diff
@@ -292,6 +297,70 @@ struct GC
...
+        pthis.gc_addRoot = function void(void* p) { gcInstance.addRoot(p);};
+        pthis.gc_addRange = function void(void* p, size_t sz, const TypeInfo ti) { gcInstance.addRange( p, sz, ti );};
+
+        pthis.gc_removeRoot = function void(void* p) { gcInstance.removeRoot(p);};
+        pthis.gc_removeRange = function void(void*p) { gcInstance.removeRange(p);};
+        pthis.gc_runFinalizers = function void(in void[] segment) { gcInstance.runFinalizers(segment);};
+
+        pthis.gc_inFinalizer = function bool() { return gcInstance.inFinalizer;};
+
+        pthis.gc_setProxy = function void(Proxy* p) {
```
MartinNowak: Let's rather add the `rootIter` and `rangeIter` interfaces than this confusing `gc_setProxy` stuff.

MartinNowak:
Mostly looks good now.

> Has it been discussed to make it an interface that each GC variant implements?

I agree, that's a better idea. Let's do that as next step.

src/gc/impl/manual/gc.d
```diff
@@ -0,0 +1,216 @@
...
+    __gshared void** roots  = null;
+    __gshared size_t nroots = 0;
+
+    __gshared Range* ranges  = null;
+    __gshared size_t nranges = 0;
+
+    void initialize()
+    {
+        if(config.gc != "manual")
```
rainers: I don't agree with Martin here (but not very strongly). I think it's better to make the decision which GC implementation to initialize in gc_init, because it's more obvious that only one of them is being selected.

src/gc/proxy.d
Jebbs: This was due to the set_proxy and clr_proxy functions. Should we just remove those since there should only be one GC at a time?

rainers: Maybe just pass the proxy pointer to the initialization function? gc_setProxy has a slightly different purpose (moving roots and ranges to another GC).

Jebbs:
>> Has it been discussed to make it an interface that each GC variant implements?

> I agree, that's a better idea. Let's do that as next step.

What is the correct way to allocate a class in druntime without a gc?

src/gc/proxy.d
Jebbs: What I meant was, if you set a new proxy and then call gc_clrProxy, pthis is what is fallen back on after the current proxy is cleared.

rainers:
> What is the correct way to allocate a class in druntime without a gc?

Allocate memory and blit the initializer over it:
``` 
void*p = malloc(__traits(classInstanceSize,GC));
GC gc = cast(GC)memcpy(p, GC.init.ptr, GC.init.length);
```

src/gc/impl/conservative/gc.d
Jebbs: The manual GC doesn't have iterators though. Should I add them?

andralex:
Nice. Is my understanding correct that this replaces the old GC hot swappability introduced by @complexmath ? He already had an interface based on pointers to functions.

andralex:
And yes it would be nice if we took advantage of vtables instead of manual tables of pointers to functions.

src/gc/impl/manual/gc.d
MartinNowak: Nope! One goal of the implementation is that it's possible to link in new GCs, previously unknown to the runtime. Therefor it __must__ be done like that.
We'll later switch to C constructors or run `gc.` module constructors earlier to make this work.

src/gc/impl/conservative/gc.d
MartinNowak: Yes

MartinNowak:
Classes and an Interface would be already much cleaner, you want to do it as part of this PR @Jebbs?

src/gc/impl/manual/gc.d
rainers: Hmm, not sure how this will work. Right now, gc_init needs to know and even import each implementation.

If you register implementations from a C constructor, you can just pass the name. That would be helpful to list available GCs, too.

Jebbs:
Yeah, I like that a lot.

Jebbs:
Might need some more formatting, but this is ready for more reviewing. 

Some key things:

1. I had some issue trying to think of a good way to free the GC instance. This is currently being done in the Dtor function. Any advice here is greatly appreciated. 

2. The Range structure is bigger than it was previously so that the manual and conservative implementations use the same structure. I'm not sure if this should be changed or if it slows down the current GC, I haven't benchmarked yet.


Other than that, it is more or less at the same spot the last commit was except now it uses an interface instead of the Proxy structure.

rainers:
This looks much cleaner.
Unfortunately, there is a disconnect for renamed files which makes it hard to review changes in gc.d. I guess these have just been removed and readded while they should have been git-moved.

src/gc/config.d
```diff
@@ -20,12 +20,13 @@ extern extern(C) __gshared bool rt_envvars_enabled;
 extern extern(C) __gshared bool rt_cmdline_enabled;
 extern extern(C) __gshared string[] rt_options;
 
+__gshared Config config;
+
 struct Config
-{
+{   bool initialized;
```
rainers: is `initialized` used anywhere? BTW: bad old dmd-style formatting.

rainers:
> I had some issue trying to think of a good way to free the GC instance. This is currently being done in the Dtor function. Any advice here is greatly appreciated.

That might indeed cause trouble when compiling with invariants, as they are also run when exiting `Dtor`. Maybe just make `Dtor` static and call it on every implementation. This would be symmetric to `initialize`.

rainers:
> The Range structure is bigger than it was previously so that the manual and conservative implementations use the same structure. I'm not sure if this should be changed or if it slows down the current GC, I haven't benchmarked yet.

Each GC can opt out of using the default range struct internally. The range iterator would then just fill a local range object with the necessary information (and leave the type information blank, for example) and call the delegate on this. So I think this is ok.

src/gc/config.d
Jebbs: Whoops! Nope, this isn't used anywhere. I had intended to test something with this, but forgot to double check the diff before I pushed and it wasn't deleted.

Jebbs:
>Unfortunately, there is a disconnect for renamed files which makes it hard to review changes in gc.d. I guess these have just been removed and readded while they should have been git-moved.

Yeah, I had no idea that I was going to be readding the a gc.d file. This shouldn't be a problem if everything is squashed at the end though, right?

Jebbs:
>>The Range structure is bigger than it was previously so that the manual and conservative implementations use the same structure. I'm not sure if this should be changed or if it slows down the current GC, I haven't benchmarked yet.

>Each GC can opt out of using the default range struct internally. The range iterator would then just fill a local range object with the necessary information (and leave the type information blank, for example) and call the delegate on this. So I think this is ok.

That's what I was thinking I could do. It just seems strange that the current GC doesn't care about the type information when the manual one does, so I thought I would mention this. Why does manual implementation(the gc stub) need to track the type info for ranges anyway?

Jebbs:
>>I had some issue trying to think of a good way to free the GC instance. This is currently being done in the Dtor function. Any advice here is greatly appreciated.

>That might indeed cause trouble when compiling with invariants, as they are also run when exiting `Dtor`. Maybe just make `Dtor` static and call it on every implementation. This would be symmetric to `initialize`.

Yeah, I had some trouble with invariants with a few other things I tried and this was one of the only things that didn't explode on me. I left it this way instead of doing a static method like you suggested because I was thinking of the possibility that there could be a GC implementation someone is using that isn't one of the built in ones added via `setGC` and that would need to get cleaned up as well. Perhaps I'm just severely over complicating it though.

I also considered using the [interface offset](https://dlang.org/phobos/object.html#.Interface.offset) in order to get the correct address of the memory. It worked as far as I could tell, but I couldn't do it in a way that I found clean looking enough.

rainers:
> Yeah, I had no idea that I was going to be readding the a gc.d file. This shouldn't be a problem if everything is squashed at the end though, right?

Might be, but reviewing is almost impossible on these files. Replacing a file with something completely different brakes almost any git-compare tool without renaming. Please try to reestablish that connection. Otherwise git-history is also lost.

rainers:
> I left it this way instead of doing a static method like you suggested because I was thinking of the possibility that there could be a GC implementation someone is using that isn't one of the built in ones added via setGC and that would need to get cleaned up as well.

Might be a sensible use case. To avoid the invariant being called, the GC instance could return a delegate/function that is called to do the destruction. Seems a bit overcomplicated, too ;-)

> I also considered using the interface offset in order to get the correct address of the memory. It worked as far as I could tell, but I couldn't do it in a way that I found clean looking enough.

Using an abstract base class instead of an interface might be a tiny bit more efficient as it avoids having to deal with these interface offset to begin with.

rainers:
> Why does manual implementation(the gc stub) need to track the type info for ranges anyway?

The original purpose of the stub was to collect some initialization information, just to be swapped out by another GC instance shortly after, so it should provide as much information as possible to the new GC.

WalterBright:
> This wraps up GC implementations in a proxy so that they can easily be picked and initialized at runtime before the user program starts according to the config options passed or embedded.

Why do this at runtime? I never understood this. The GC should be a library thing - just link with a different GC implementation to have a different GC. Having it be a runtime thing means all interactions will go through a pointer, which is a big speed bump.

Jebbs:
>Why do this at runtime? I never understood this. The GC should be a library thing - just link with a different GC implementation to have a different GC. Having it be a runtime thing means all interactions will go through a pointer, which is a big speed bump.

As far as I understand it, this is done at runtime so that no matter the GC implementation all D programs can share a common druntime library.

Jebbs:
@rainers, did a rebase. I think that should fix things as far as the history goes. Git mentioned a rename for gc.d -> impl/conservative/gc.d.

rainers:
> Any idea how to fix that?

What I would try:
- make a backup copy of the affected file
- reset the branch to the state of master
- git-mv the file gc.d to it's new location
- restore the backup files
- commit and check the history
- double check again before pushing

I use an interactive tool for this (TortoiseGit) though, so I'm not too familiar with the actual command line calls.

Jebbs:
Thanks for the info. I'll play around with it and see what I can do.

rainers:
> Having it be a runtime thing means all interactions will go through a pointer, which is a big speed bump.

GC has been a class until a couple of releases ago (with an extra check for an installed proxy), and the change to a struct made just a tiny difference in the benchmarks. Removing the proxy check probably has a similar affect (I'm not a fan of the proxy stuff, but if we keep it, it should just be an implementation of the GC interface). The benchmarks should be rerun on any change to the implementation, though.

> The GC should be a library thing - just link with a different GC implementation to have a different GC.

Having the GC functions in gc.proxy for being replaced at link time means they must not be inlined. Instead each call needs to execute an extern(C) function wrapper.

Jebbs:
So after a lot of time searching and failed attempts at getting this to work, it looks like moving a file won't keep the history. Even though it detects a rename, it considers it a new file. 

I did find a way to "preserve" the history, but from what I've read it is frowned upon. We would have to essentially rewrite the entire git history for this file making it seem that the file `gc/gc.d` had always been `gc/impl/conservative/gc.d`. While it does keeps the history of the file, it changes 4393 previous commits to make this work. You can see that it does keep the history if you [check out](https://github.com/Jebbs/druntime/blob/gc_swap_test2/src/gc/impl/conservative/gc.d) my test. (I had to do a commit myself in order to get the history to show up correctly on github)

So we can either bite the bullet and change all of these commits in order to preserve the history of this file (possibly breaking things for people trying to search through the history), or allow the history to be lost for this file. Unless anyone has any other suggestions?

LightBender:
I've done it before. It was a small team with relatively few commits. It was pure, unrestricted pain. Rewriting the history is a path of Woe and Sorrow. Do do it, unless under extreme need. Which I don't think this situation rises to.

Jebbs:
I agree, I think it is a terrible idea. I would much rather lose the history on a single file than mess with the repo like this.

Jebbs:
>Using an abstract base class instead of an interface might be a tiny bit more efficient as it avoids having to deal with these interface offset to begin with.

I'm ok with doing this. I played around a little bit and using an abstract class I was able to change things to free the memory in `gc_term` without any hassles that using an interface caused. Is there any downside to using an abstract class as opposed to an interface that I'm not aware of? Other than putting `abstract` and `override` everywhere?

rainers:
Thanks, David, for explanation.

Adding a commit with the renames only seems to work (I tried here: https://github.com/rainers/druntime/commits/gc_swa). It breaks the log, but it is at least visible in the history what happened. git-blame is also able to track history correctly.

While we are at it: 
- I'm not sure it's necessary to rename gc.proxy.
- do we need the extra "impl" sub package? will there be other directories besides "impl"?

Jebbs:
That's what I'll do then. 

>While we are at it:
>+ I'm not sure it's necessary to rename gc.proxy.

I'm fine with this. The interface (or abstract class) serves the same function as the proxy, and we can at least keep the history this way.

>+ do we need the extra "impl" sub package? will there be other directories besides "impl"?

I did this in order to be explicit in the directory layout where the implementations were kept. Since it is only used in a couple of places in druntime, I don't see any reason to change it. I also don't know what would be a better layout.

MartinNowak:
> Why do this at runtime? The GC choice should be a library thing - just link with a different GC implementation to have a different GC. Having it be a runtime thing means all interactions will go through a pointer, which is a big speed bump.

Because we want the GC implementation to be selectable at runtime and link-time. Also a vtbl jump costs almost nothing in comparison to an allocation.

MartinNowak:
As Rainer said, having the rename in an extra commit should fix the rename detection, b/c it's done on a per commit base.

Jebbs:
Ok, let's try this again.

Minor changes, but they include:
- GC interface is now located in gc/proxy.d and named changed to Proxy because of the filename. Things related to this name have been changed back (gc_setProxy/gc_clearProxy/gc_getProxy)
- The way the GC is freed was changed to be a static method of each GC class, and all are called in gc_term.
- Copyright updates

Looks like I'm still getting some failures in the auto-tester, mostly 32 bit FreeBSD machines (though some pass?). Any idea how I might tackle those? Should I set up a FreeBSD virtual computer?

rainers:
> Ok, let's try this again.

Thanks, it's digestible now.

> GC interface is now located in gc/proxy.d and named changed to Proxy because of the filename. Things related to this name have been changed back (gc_setProxy/gc_clearProxy/gc_getProxy)

I don't like `Proxy` as the name of the GC interface. I'd recommend calling it `GCInterface` and putting it into a separate module. The proxy functionality is something different.

src/gc/impl/conservative/gc.d
```diff
@@ -267,21 +273,53 @@ struct GC
...
+        auto p = cstdlib.malloc(__traits(classInstanceSize,ConservativeGC));
+
+        if(!p)
+            onOutOfMemoryErrorNoGC();
+
+        instance = cast(ConservativeGC)memcpy(p, typeid(ConservativeGC).initializer.ptr, typeid(ConservativeGC).initializer.length);
+
+        instance.__ctor();
+
+        gc_setProxy(instance);
```
rainers: This call is causing the test failure on the windows build server: it causes the exported function to be included in the binary, which makes the linker produce an import library. I'd rather return the instance and assign it to the global. (I still don't like checking `config.gc` in each GC, but that could be a matter of taste).

src/gc/impl/conservative/gc.d
```diff
@@ -383,9 +396,49 @@ struct GC
             return res;
     }
 
-    /**
-     *
-     */
+
+    void enable()
```
rainers: It would be easier to review if you could avoid moving functions. If you think it's necessary it would be nicer in a separate commit.

src/gc/impl/conservative/gc.d
```diff
@@ -1003,68 +1029,64 @@ struct GC
...
-    /**
-     * add p to list of roots
-     */
-    void addRoot(void *p) nothrow @nogc
+        stats.poolsize = psize;
+        stats.usedsize = bsize - flsize;
+        stats.freelistsize = flsize;
+    }
+
+    void addRoot(void* p) nothrow
```
rainers: You have stripped @nogc here from the function signture. This should not be possible for an overriding function, must be a dmd bug. Same for the following functions.

rainers: BTW: I don't think it was a good idea to add @nogc to these functions, because it doesn't allow using garbage collected memory to store the roots, i.e. you cannot easily implement this (mostly) lock free (due to the complications freeing memory).

Jebbs:
>I don't like Proxy as the name of the GC interface. I'd recommend calling it GCInterface and putting it into a separate module. The proxy functionality is something different.

I don't like the name much either, but since they served similar functions I thought it wasn't too bad an idea. I also wasn't sure what else to call the file other than gc.d, so I thought this was fine for now. No problems changing it, but I want to make sure it has an agreeable name.

src/gc/impl/conservative/gc.d
Jebbs: That's fair. The only reason I did it was for consistency of order with the interface and the other implementation. I can change it back for reviewing.

src/gc/proxy.d
```diff
@@ -126,229 +257,180 @@ extern (C)
...
-        thread_term();
 
-        _gc.Dtor();
+        instance.collectNoStack(); // not really a 'collect all' -- still scans
+                                   // static data area, roots, and ranges.
+
+        ManualGC.finalize();
+        ConservativeGC.finalize();
+
+        thread_term();
```
rainers: `thread_term` should be called before finalizing the GCs.

src/gc/proxy.d
```diff
@@ -126,229 +257,180 @@ extern (C)
...
+                instance = ithis = inst;
+                return;
             }
-            proxy = p;
-            foreach (r; _gc.rootIter)
-                proxy.gc_addRoot( r );
 
-            foreach (r; _gc.rangeIter)
-                proxy.gc_addRange( r.pbot, r.ptop - r.pbot, null );
+            foreach(root; inst.rootIter)
```
rainers: This moves roots from `inst` to `inst`. One of these should be `instance`.

src/gc/impl/conservative/gc.d
Jebbs: Oh, interesting! Thanks for the help here.

rainers:
> I also wasn't sure what else to call the file other than gc.d, so I thought this was fine for now. No problems changing it, but I want to make sure it has an agreeable name.

`gc.proxy` is ok for the proxy functionality of the extern(C) interface and the get/setProxy functions, but the interface definition should be in a separate file, e.g. `gc.gcinterface` or `gc._interface`.

Jebbs:
gc.gcinterface isn't bad. I'll probably go with that.

Jebbs:
Worked on stuff @rainers commented on, should be mostly good now.

Now let's see what I can do about those tests...

rainers:
Thanks, looks much cleaner now. I'd recommend making `RangeIterator` and `RootIterator` public as they are part of the interface. That avoids repeating them in each GC implementation.

src/gc/gcinterface.d
```diff
@@ -0,0 +1,196 @@
...
+ * Distributed under the Boost Software License, Version 1.0.
+ *    (See accompanying file LICENSE or copy at
+ *          http://www.boost.org/LICENSE_1_0.txt)
+ */
+module gc.gcinterface;
+import gc.stats;
+
+private
+{
+    static import core.memory;
```
ZombineDev: Can you do this the other way around - define `BlkInfo` here and import it from `core.memory`? Otherwise there may be cyclic dependency - `core.memory` imports the module and vice versa. Also, is `BlkInfo` common for all possible GCs, or is it specific for the current conservative one?

rainers: I agree, if `BlkInfo` is part of the interface (`qalloc` uses it), it should be defined here as well. But I'd avoid broadening the impact of this PR, let's defer this to another PR.

rainers:
> Now let's see what I can do about those tests...

They magically just pass now ;-) Instead you might want to run the druntime benchmarks to verify there's no performance penalty.

src/gc/impl/manual/gc.d
```diff
@@ -0,0 +1,305 @@
...
+}
+
+extern (C) void onOutOfMemoryError(void* pretend_sideffect = null) @trusted pure nothrow @nogc; /* dmd @@@BUG11461@@@ */
+
+__gshared ManualGC instance;
+
+class ManualGC : GC
+{
+    __gshared Root* roots = null;
+    __gshared size_t nroots = 0;
```
ZombineDev: pointer + size = slice
Why not use `Root[]` instead? AFAICT, it's dynamic array, not a linked-list or something like that.

src/gc/impl/manual/gc.d
```diff
@@ -0,0 +1,305 @@
...
+
+    __gshared Range* ranges = null;
+    __gshared size_t nranges = 0;
```
ZombineDev: Same as above - use `Range[]` instead.

src/gc/impl/manual/gc.d
Jebbs: Because this can't use the GC.

src/gc/impl/manual/gc.d
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    GCStats stats() nothrow
+    {
+        return GCStats.init;
+    }
+
+    void addRoot(void* p) nothrow @nogc
+    {
+        Root* r = cast(Root*) cstdlib.realloc(roots, (nroots + 1) * roots[0].sizeof);
```
ZombineDev: Growing on every insertion is slow. It's better to grow by e.g. factor of 2, so that reallocations are amortized.
BTW, I noticed that there is a [`rt.util.containrr.array`](https://github.com/dlang/druntime/blob/master/src/rt/util/container/array.d) module. Maybe you can use it, instead of managing this yourself? Though, I'm not sure if it's reallocations are optimal either.

Jebbs:
>They magically just pass now

Mostly. It looks like there's still some red mixed in there, still mostly 32bit FreeBSD machines. Shouldn't I make sure every test passes?

src/gc/impl/manual/gc.d
LightBender: @ZombineDev The purpose of this GSoC project is not to improve the existing GC but to improve the infrastructure such that we can pull in Rainer's Precise GC. As such I would like to keep the student focused on that goal. You are more than welcome to make the changes you are recommending and put them up in a separate PR.

LightBender:
Those red lines are from prior runs of the AutoTester. The latest one shows all Greens.

src/gc/impl/manual/gc.d
```diff
@@ -0,0 +1,305 @@
...
+    void removeRoot(void* p) nothrow @nogc
+    {
+        for (size_t i = 0; i < nroots; ++i)
+        {
+            if (roots[i] is p)
+            {
+                roots[i] = roots[--nroots];
+                return;
+            }
+        }
```
ZombineDev: Is this correct? Don't you need to move all of the elements after and including the removed one to the left? I think the correct way should be something along the lines:

```
size_t idx;
for (; roots[idx] !is p && idx < nroots; idx++) { }
assert (roots[idx] is p);
roots[idx .. nroots - 1] = roots[idx + 1 .. nroots]; // move elemnts to the left
```

src/gc/impl/manual/gc.d
```diff
@@ -0,0 +1,305 @@
...
+            if (result)
+                break;
+        }
+
+        return result;
+    }
+
+    void addRange(void* p, size_t sz, const TypeInfo ti = null) nothrow @nogc
+    {
+        Range* r = cast(Range*) cstdlib.realloc(ranges, (nranges + 1) * ranges[0].sizeof);
```
ZombineDev: Same suggestion as for `addRoot` - don't manage this manually, DRY.

src/gc/impl/manual/gc.d
```diff
@@ -0,0 +1,305 @@
...
+    void removeRange(void* p) nothrow @nogc
+    {
+        for (size_t i = 0; i < nranges; ++i)
+        {
+            if (ranges[i].pbot is p)
+            {
+                ranges[i] = ranges[--nranges];
+                return;
+            }
+        }
```
ZombineDev: Same suggestion as for `removeRoot` - verify correctness and preferably use a container.

src/gc/impl/manual/gc.d
ZombineDev: Slices are not related to the GC. They are the same as pointer+size. That's why they are usable in `@nogc` code. The only thing requires the GC is concatenation and setting `slice.length`.

src/gc/impl/manual/gc.d
ZombineDev: I know, I was just pointing that it would be better to reuse an existing container implementation, because it would be easier, should minimize the probability of bugs and it maybe faster.

src/gc/gcinterface.d
ZombineDev: Alright, leaving this for a different PR, works for me.

src/gc/impl/manual/gc.d
ZombineDev: E.g. the idiom when doing manual memory management is:
```
T[] slice = (cast(T*)malloc(T.sizeof * length))[0 .. length];
scope (exit) free(slice.ptr);
```

ZombineDev:
@Jebbs nice work! Overall I like the changes in this PR (especially the GC interface and the ManualGC). I made a couple of minor suggestions, I hope you don't mind. I'm ok with leaving other improvements for future PRs, the goal is to get this one merged ASAP.

The only bigger question for this PR that I have is:
Why does `gc.proxy` need to know about all the GC implementations and why does it need to initialize them?

I think it makes more sense if the interface was something along the lines of this:
```
// user code somewhere close to rt_init, extern(C) main

import gc.impl.manual : ManualGC;
auto gc = ManualGC.createInstance();

imort gc.proxy : set_proxy;
set_proxy(gc);
```

ZombineDev:
BTW, @Jebbs I think it would be better to leave non-critical changes (like most of the formatting / documentation changes in src/gc/impl/conservative/gc.d) in order to speed of the reviewing process. Walter usually recommends placing formatting, refactoring and bug fixes all in separate PRs.

complexmath:
```
Nice. Is my understanding correct that this replaces the old GC hot swappability introduced by @complexmath ? He already had an interface based on pointers to functions.

And yes it would be nice if we took advantage of vtables instead of manual tables of pointers to functions.
```
Just a bit of background. The function pointer stuff was imported from some work Walter did to support DLLs. For backwards compatibility reasons, it wasn't possible to use something like interfaces. Presumably this is no longer a requirement, but someone should double check.

And yes, Druntime has always supported link-time GC swapping. In fact, this was a design goal since before it was called Druntime, and it originally shipped with compiler runtime (rt.*) GC (gc.*) and library core (core.*) as separate entities. The intent was to recognize that compiler writers, standard library writers, and GC designers are all typically distinct teams and that managing a single large repo with everything wasn't scalable in the long term. However, no one liked the idea of having three separate but necessary runtime libraries on their link line, and since DMD was basically the only game in town back in the day, people felt that this was needless complexity, and so the three pieces were eventually merged into one per user request. Perhaps (hopefully) D is approaching the point where this may be a valid library design.

And on Druntime in general, the rule is that everything is allowed to import core.*, but the gc.* and rt.* namespaces are private and should never be externally referenced. This recognizes that there is only one standard library, while the compiler and GC support code should be interchangeable.

Jebbs:
@ZombineDev, Thanks for taking the time to review my work. Looks like you mostly gave feedback on the manual GC implementation, but aside from porting it ( it was originally [this](https://github.com/dlang/druntime/blob/master/src/gcstub/gc.d)) I didn't really touch it. I figured that since it was already in druntime I didn't need to worry about it.

>Why does gc.proxy need to know about all the GC implementations and why does it need to initialize them?

@rainers suggested that the extern(C) interface stay in gc.proxy.d, and I went with it. Because gc_init is one of those functions, it currently needs to know about them. I think we're going to be changing that in the future, but it should be fine for this PR. 

>BTW, @Jebbs I think it would be better to leave non-critical changes (like most of the formatting / documentation changes in src/gc/impl/conservative/gc.d) in order to speed up the reviewing process.

Yeah, that came up when @rainers was doing some reviewing... I'll definitely strive for that next time.

src/gc/impl/manual/gc.d
Jebbs: AFAICT, the order doesn't matter, so I think it should be ok.

src/gc/gcinterface.d
Jebbs: >Can you do this the other way around - define BlkInfo here and import it from core.memory?

Is this a good idea? I don't think anything outside of the gc packager is supposed to import anything in it. 

>Otherwise there may be cyclic dependency - core.memory imports the module and vice versa. 

core.memory doesn't import anything from the package. It just calls the extern(c) functions defined in gc.proxy.

Jebbs:
>Those red lines are from prior runs of the AutoTester. The latest one shows all Greens.

Wow, I did not even notice that.

Jebbs:
>Instead you might want to run the druntime benchmarks to verify there's no performance penalty.

I ran benchmarks for this a couple of days ago. As far as I could see there wasn't a discernible difference. The mean, median, and mode for a set of 50 runs were almost the same. Should I post my results before this is merged?

rainers:
> The mean, median, and mode for a set of 50 runs were almost the same

`druntime/benchmark/runbench` takes the minimum of 10 runs, which is considered a good measure because noise usually adds to the actual load.

> Should I post my results before this is merged?

Yes, please do.

Jebbs:
Not sure what format to put them in, so here's a table.

| | Current | PR |
| --- | --- | ---|
| resize | 0.262 | 0.263 |
| string | 0.063 | 0.063 |
| bulk | 0.242 | 0.242 |
| int | 0.123 | 0.121 |
| l3cache | 0.589 | 0.587 |
| ptr | 0.084 | 0.076 |
| bigval | 0.439 | 0.44 |
| stomper | 0.518 | 0.518 |
| obj | 0.126 | 0.124 |
| tree2 | 1.067 | 1.047 |
| slist | 1.774 | 1.76 |
| conalloc | 0.069 | 0.068 |
| huge_single | 0.004 | 0.004 |
| conappend | 0.014 | 0.014 |
| vdparser | 1.922 | 1.921 |
| rand_small | 0.645 | 0.643 |
| concpu | 0.072 | 0.072 |
| conmsg | 1.241 | 1.238 |
| words | 1.041 | 1.035 |
| rand_large | 0.276 | 0.277 |
| dlist | 1.872 | 1.858 |
| tree1 | 0.631 | 0.641 |
| testgc3 | 1.912 | 1.91 |

Jebbs:
Anything other thoughts or feedback on this?

LightBender:
One minor nit. Should the authors section include the prior authors since this was a re-org and not a rewrite?

Other than that, looks good. 
LGTM!

rainers:
> Not sure what format to put them in, so here's a table.

Thanks, looks good. It even shows slightly better performance (though probably within accuracy of measurements).

rainers:
> Just a bit of background.

@complexmath Thanks, Sean. This PR doesn't change anything with respect to linkage (yet). The extern(C) gc_* functions are still there.

src/gc/proxy.d
```diff
@@ -1,118 +1,53 @@
...
-
-        pthis.gc_addrOf = &gc_addrOf;
-        pthis.gc_sizeOf = &gc_sizeOf;
-
-        pthis.gc_query = &gc_query;
-
-        pthis.gc_addRoot = &gc_addRoot;
-        pthis.gc_addRange = &gc_addRange;
+    __gshared GC instance;
+    __gshared GC ithis;
```
rainers: One more nit: could you please add a comment about the difference of `instance` and `ithis`? Maybe `ithis` isn't a good name to begin with...

Jebbs: `ithis` is what it was called before, so I didn't think to change it.

Jebbs:
@LightBender , do you mean in gcinterface.d?

Jebbs:
@rainers, it might even be faster after I figure out how to get the range opApply figured out for the conservative implementation. It is still using the larger structure.

Jebbs:
Original authors' names were added to gcinterface.d and `ithis` was changed and comments were added. Any other nits or suggestions?

rainers:
Thanks. I'm ready to merge, I was hoping for @MartinNowak to also show his content.

ZombineDev:
One last design question:
What do you all think about placing the GC interface in `core.memory` (and replacing the GC struct with static  methods) and leaving only `extern(C)` wrappers in `gc.proxy`? I think this would be the cleanest design, considering what @complexmath said.

ZombineDev:
We can even get away without a deprecation cycle, if add:
```
@property IGC GC() { return get_current_gc(); }
```

MartinNowak:
> Just a bit of background. The function pointer stuff was imported from some work Walter did to support DLLs. For backwards compatibility reasons, it wasn't possible to use something like interfaces. Presumably this is no longer a requirement, but someone should double check.

Yes, vtbls and interfaces should just work when loaded dynamically, no problem here.

> What do you all think about placing the GC interface in core.memory (and replacing the GC struct with static methods) and leaving only extern(C) wrappers in gc.proxy?

Let's keep gc stuff in gc.* until we know more. We still want to support runtime GC selection which requires to preinitialize all linked GCs, using the module constructors of the `gc.*` package would be one option for that.

src/gc/gcinterface.d
```diff
@@ -0,0 +1,194 @@
...
+
+struct Range
+{
+    void* pbot;
+    void* ptop;
+    TypeInfo ti; // should be tail const, but doesn't exist for references
+    alias pbot this; // only consider pbot for relative ordering (opCmp)
+}
+
+const uint GCVERSION = 1; // increment every time we change interface
```
MartinNowak: No need for that, you can use `__VERSION__` to get the dmd frontend version, and we're releasing druntime w/ that as well.
Also it's our duty to keep this interface somewhat backwards compatible.

MartinNowak: Seems like that already existed before, but isn't used anywhere.

src/gc/impl/manual/gc.d
MartinNowak: Use rt.util.Array!Root instead.

MartinNowak:
Except from https://github.com/dlang/druntime/pull/1581#discussion_r68642017, LGTM.

src/gc/impl/manual/gc.d
Jebbs: For GC.removeRoot and GC.removeRange, do you want me to use Array.remove? Adjusting the array is going to slow things down as opposed to how it is currently done.

src/gc/proxy.d
```diff
@@ -1,118 +1,53 @@
...
-
-        pthis.gc_addrOf = &gc_addrOf;
-        pthis.gc_sizeOf = &gc_sizeOf;
-
-        pthis.gc_query = &gc_query;
-
-        pthis.gc_addRoot = &gc_addRoot;
-        pthis.gc_addRange = &gc_addRange;
+    __gshared GC currentGC;  //used for making the GC calls
+    __gshared GC initialGC; //used to reset currentGC if gc_clrProxy was called
```
MartinNowak: We shouldn't need initialGC. You can only ever use a single GC during a program lifetime.

Jebbs: What about in the case of gc_clrProxy?

Jebbs:
I updated those bits to use `rt.util.container.array.Array`, but I kept the `removeRoot` and `removeRange` functions mostly unchanged in how they function.

MartinNowak:
I made a PR for your branch with a few things I found.
https://github.com/Jebbs/druntime/pull/2

src/gc/proxy.d
MartinNowak: You're right it's needed for the whacky Windows DLL support, I renamed and added comments.
https://github.com/Jebbs/druntime/pull/2/commits/8f99b220b4b22c1041df2d6107f450d8a0a4a68d

LightBender:
LGTM

MartinNowak:
Auto-merge toggled on

MartinNowak:
I'll take care of the support for unknown GCs that get linked into a program. It requires quite a bit of low-level linker understanding.

MartinNowak:
This is still lacking a changelog entry and updated [documentation](https://dlang.org/spec/garbage.html) on dlang.org @Jebbs, could you please add those.

Jebbs:
Yep, I'll take care of that tomorrow.

andralex:
@Jebbs thx!

Jebbs:
Since it wasn't in a change log before this, should I add it to changelog.dd for pending 2.072 changes?

MartinNowak:
> Since it wasn't in a change log before this, should I add it to changelog.dd for pending 2.072 changes?

Yes @Jebbs, it hasn't yet been released and will be part of 2.072.

PR #1582, opened 2016-06-01
Port osx_tls.c to D

jacob-carlborg:


Geod24:
Since it's OSX-specific, should it be in stdc ?

src/rt/sections_osx_x86_64.d
```diff
@@ -114,11 +113,62 @@ void scanTLSRanges(void[] rng, scope void delegate(void* pbeg, void* pend) nothr
...
+        assert(state == dyld_tlv_states.allocated);
+        if (info.tlv_addr <= tlsSymbol && tlsSymbol <
+            (info.tlv_addr + info.tlv_size)
+        )
+        {
+            range = TLSRange(info.tlv_addr, info.tlv_size);
+        }
+    });
+
+    dyld_enumerate_tlv_storage(&handler);
```
Geod24: This handler doesn't escape, does it ?
In which case the delegate literal at line 156 should be made a `scope` delegate to avoid needless memory allocation.
What is blocks semantic in that regard ? Is it mostly / only used for callback or can it be returned as a true delegate as well ?

src/core/stdc/clang_block.d
```diff
@@ -0,0 +1,198 @@
...
+    // Imported variables go here
+    R delegate(Params) dg;
+}
+
+Block!(R, Params) block(R, Params...)(R delegate(Params) dg)
+{
+    static if (Params.length == 0)
+        enum flags = 1342177280;
+    else
+        enum flags = 1073741824;
```
Geod24: Could you document the source ?

src/core/stdc/clang_block.d
```diff
@@ -0,0 +1,198 @@
...
+    int flags;
+    int reserved;
+    extern(C) R function(Block*, Params) invoke;
+    Descriptor* descriptor;
+
+    // Imported variables go here
+    R delegate(Params) dg;
+}
+
+Block!(R, Params) block(R, Params...)(R delegate(Params) dg)
```
Geod24: Should `function` be supported as well ?

Geod24:
Out of curiosity, what's the motivation behind the porting ?

jacob-carlborg:
> Out of curiosity, what's the motivation behind the porting ?

* Dogfooding
* I got some [complains](https://github.com/dlang/druntime/pull/1461#discussion_r56579759) first time this was added and was not written in D
* Why is the rest of druntime, DMD (half of it), or anything wirtten in D? 😃 
* Because I can 😃

jacob-carlborg:
> Since it's OSX-specific, should it be in stdc ?

I don't know, suggestions are welcome. I had to put it somewhere. I quick search on google suggest that it works on Linux as well with the correct runtime libraries installed. I can do a test and confirm.

src/rt/sections_osx_x86_64.d
jacob-carlborg: > This handler doesn't escape, does it ?

I really hope it doesn't because `block` returns by value and `handler` is on the stack, which will be gone after `getTLSRange` completes. I can look at the source code to confirm, but I would be very surprised.

> In which case the delegate literal at line 156 should be made a scope delegate to avoid needless memory allocation.

Is that possible to do on the call site?

> What is blocks semantic in that regard ? Is it mostly / only used for callback or can it be returned as a true delegate as well ?

What do you mean with "true delegate"? A closure that captures local variables which can be used after a function returns? Yes.

src/core/stdc/clang_block.d
jacob-carlborg: I got these numbers from an example in C where I pass in a Block to a function, where the block parameter was typed as the struct `Block` above, and then reading the `flags` field. I tried to use the flags which are documented [here](http://clang.llvm.org/docs/Block-ABI-Apple.html), but it never matched the values that were actually received.

src/core/stdc/clang_block.d
jacob-carlborg: I don't know. I tried to keep it simple. What's required for supporting both, a union? Or should I just force a function pointer into the delegate, i.e. `dg.funcptr = ...`?

jacob-carlborg:
I just did a test in a Debian container, according to these instructions [1], and it works perfectly fine.

[1] http://stackoverflow.com/a/11097610

jacob-carlborg:
I'll add some code to make it possible to call the block from the D side as well.

jacob-carlborg:
Suggestions for the naming convention would be appreciated as well. I think the official name is "Blocks" but I called it "Clang Blocks" for now. To avoid confusion with a regular block. Perhaps "Block" with a capital letter is enough to avoid any confusion.

jacob-carlborg:
@MartinNowak osx_tls.c is now ported to D.

src/core/stdc/clang_block.d
rainers: I have no clue what the flags mean, but using the hex values 0x50000000 and 0x40000000 will probably look less magic.

jacob-carlborg: Fair enough.

jacob-carlborg:
> I'll add some code to make it possible to call blocks from the D side as well.

I'll skip this because the `Block` struct need to look different depending on if any local variables are captured on the C side. This seems a bit too complicated.

jacob-carlborg:
Ok, all tests are passing.

src/rt/osx_tls.c
```diff
@@ -1,51 +0,0 @@
...
- * Copyright: David Nadlinger, 2012.
- * License:   <a href="http://www.boost.org/LICENSE_1_0.txt">Boost License 1.0</a>.
- * Authors:   David Nadlinger
- */
-
-#ifndef __APPLE__
-#ifndef __BLOCKS__
- #error "Need a C compiler with Apple Blocks support – not building on OS X with Clang?"
-#endif
-#endif
```
schveiguy: Is there an equivalent test needed in the D version? That is, what happens if you are building on something other than OSX, or linking with non-clang libc?

I totally admit, I'm not sure what should be done, I'm just pointing out a difference between the C code and the D code.

jacob-carlborg: The content of this file is moved into `sections_osx_x86_64.d`, which is already properly guarded for OS X 64bit.

The code which makes calling Blocks possible, `clang_block.d`, is cross-platform. Using that code only make sense if you have a C function taking Block you want to call. That C code needs to be compiled with Clang, and linking that on non-Apple platforms requires the Blocks runtime, which is noted in the documentation. We can't do anything, as far as I know, on the D side to enforce this.

If you try to compile the C code with something else than Clang you'll get a syntax error. Linking without the Blocks runtime will cause undefined symbol errors.

jacob-carlborg: So far the tests for `clang_block.d` is only enabled on OS X because it has everything required out of the box. If we install Clang and the Blocks runtime on the other machines used by the autotester I can enable it for those other platforms as well.

src/rt/sections_osx_x86_64.d
Geod24: > Is that possible to do on the call site?

Yes. Either you can use:
```D
scope dg = (dyled_tlv_states state, dyld_tlv_info* info) { /*...  */ }
auto handler = block(dg);
```

You could change the definition of `block` to take a parameter of type `scope R delegate(Params) dg` which will mean delegate literals passed directly will never allocate. But see the next point.

> What do you mean with "true delegate"? A closure that captures local variables which can be used after a function returns? Yes.

Quite a poor phrasing on my side, the question was actually "can the block escape the lexical scope in which it was declared", to which the answers seems to be yes (and so you can't use `scope delegate` for `block`'s prototype).

src/core/stdc/clang_block.d
Geod24: Both would work, though the `union` will complicate things. It's up to you if you want to support it, just though I would raise the point.

Geod24:
> I don't know, suggestions are welcome. I had to put it somewhere. I quick search on google suggest that it works on Linux as well with the correct runtime libraries installed. I can do a test and confirm.

I would definitely not put it in `stdc`, as its for C standard library.

Either we can have a `core.clang` package, but that seems like a dangerous precedent, or as part of `core.sys.osx[.clang]`. Last but not least, it could be in an internal module, like `rt`.
I am leaning towards the later option.

src/core/stdc/clang_block.d
```diff
@@ -0,0 +1,206 @@
...
+ * $(CCODE
+ *      $ clang foo.c -o foo.o
+ *      $ dmd main.d foo.o -L-lBlocksRuntime
+ * )
+ *
+ * Copyright: Copyright Jacob Carlborg 2016.
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   Jacob Carlborg
+ * Source:    $(DRUNTIMESRC core/_clang_block.d)
+ */
```
Geod24: Kudos for documentation !

src/rt/sections_osx_x86_64.d
jacob-carlborg: > `scope dg = ...`

Aha, I've never though of doing that for anything like but classes. I'll update that.

src/core/stdc/clang_block.d
jacob-carlborg: @rainers changed.

jacob-carlborg:
> Either we can have a core.clang package, but that seems like a dangerous precedent, or as part of core.sys.osx[.clang]. Last but not least, it could be in an internal module, like rt.
I am leaning towards the later option.

It works on other platforms than OS X (that's why I put it in core.stdc) but I guessing it's not going to be used on any non-Apple platform. I did not want to put it in the `rt` package because it's useful outside of the runtime.

Geod24:
> It works on other platforms than OS X (that's why I put it in core.stdc) but I guessing it's not going to be used on any non-Apple platform. I did not want to put it in the rt package because it's useful outside of the runtime.

I think we should separate what is used for implementing the runtime and what druntime expose.
There is a great deal of overlap thanks to libc, but I don't think we should expose this one.

jacob-carlborg:
> I think we should separate what is used for implementing the runtime and what druntime expose.
There is a great deal of overlap thanks to libc

There's a specific reason why I chose to expose the Blocks related code. That's because it's very useful to be able to interface with functions taking Blocks on OS X.

> but I don't think we should expose this one.

I don't see a reason why not.

src/rt/sections_osx_x86_64.d
jacob-carlborg: @Geod24 changed delegate to be `scope`.

MartinNowak:
Blocks aren't standard C, so they don't belong into core.stdc, use core.sys.osx._something_.
If the ported D version doesn't rely on hacks (looks like http://clang.llvm.org/docs/Block-ABI-Apple.html), then replacing the C code is fine with me.

jacob-carlborg:
> If the ported D version doesn't rely on hacks (looks like http://clang.llvm.org/docs/Block-ABI-Apple.html), then replacing the C code is fine with me.

Not sure what you mean with "hacks". But yes, it uses/follows http://clang.llvm.org/docs/Block-ABI-Apple.html

jacob-carlborg:
> sys.osx isn't really correct either, though, since they are not OS X-specific (even though obviously developed by Apple). It's a Clang thing.

Exactly. If we install Clang and the Blocks Runtime on the autotester machines I could enable the tests on more platforms.

PR #1583, opened 2016-06-02
Add @nogc to pthread functions.

tmccombs:
Fixes Issue 14601

### DIFF
--- a/src/core/sys/posix/pthread.d
+++ b/src/core/sys/posix/pthread.d
@@ -33,7 +33,6 @@ else version (WatchOS)
 version (Posix):
 extern (C)
 nothrow:
-@nogc:
 
 //
 // Required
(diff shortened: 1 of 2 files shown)


MartinNowak:
You need to move the function pointer types above the `@nogc:` declaration.
The way we did it in other modules was to create overloads for both `@nogc` and normal function pointer, then deprecated the normal ones.

tmccombs:
Could you give me an example of a module where this was done?

MartinNowak:
> Could you give me an example of a module where this was done?

It took me a while to remember which it was, core.sys.posix.signal.
Here is the [declaration for the signal handler  functions](https://github.com/dlang/druntime/blob/cf7a816dccd84e9a1353748c70357b03ea09b3cf/src/core/sys/posix/signal.d#L94-L101) and an [example for an overload](https://github.com/dlang/druntime/blob/cf7a816dccd84e9a1353748c70357b03ea09b3cf/src/core/sys/posix/signal.d#L1418-L1424).
If the type is part of the public API,

tmccombs:
Thank you

MartinNowak:
Auto-merge toggled on

PR #1584, opened 2016-06-05
XREF purge: CXREF, custom SXREF, changelog.dd

aG0aep6G:
This removes the last traces of XREF from druntime.

Be aware that removing the macros from changelog.dd may be controversial. I suggest to discuss this in a central place over at phobos: https://github.com/dlang/phobos/pull/4404


CyberShadow:
Auto-merge toggled on

PR #1588, opened 2016-06-07
fix Issue 6333 - The 'capacity' function is not pure/nothrow/@safe

WalterBright:
Can't think of a reason why `capacity()` shouldn't be trusted.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2943,12 +2943,12 @@ private
  *
  * Note: The capacity of a slice may be impacted by operations on other slices.
  */
-@property size_t capacity(T)(T[] arr) pure nothrow
+@property size_t capacity(T)(T[] arr) pure nothrow @trusted
 {
     return _d_arraysetcapacity(typeid(T[]), 0, cast(void *)&arr);
 }
 ///
-unittest
+@safe unittest
 {
     //Static array slice: no capacity
     int[4] sarray = [1, 2, 3, 4];


schuetzm:
IMO, `_d_arraysetcapacity` should be the function that is marked as `@trusted`. However, it is far from obvious that it is actually safe...

schveiguy:
> IMO, `_d_arraysetcapacity` should be the function that is marked as `@trusted`

No, this can't be `@trusted` because you can violate memory expectations. As far as the runtime is concerned, when you `assumeSafeAppend`, the memory you release at the end of the array is now free data. If you still have references to it, then bad things can happen. It may be arguable that those bad things still won't violate memory issues, but, for instance, a smarter GC may stop scanning the unallocated data, thereby collecting things that you incorrectly retained access to.

schveiguy:
Auto-merge toggled on

schveiguy:
Auto-merge toggled off

schveiguy:
This needs a unit test, then it's good to go.

schuetzm:
My point was that if `_d_arraysetcapacity` can't be `@trusted`, then `capacity` (which calls it) is even less trustworthy.

schveiguy:
Oops, I'm incorrect!

`_d_arraysetcapacity` should be `@trusted`, I'm confusing it with `_d_arrayshrinkfit`.

Yes, let's fix this, and then both `capacity` and `reserve` should be inferred `@safe` (we can remove `@trusted` from `reserve`). I'll work on this.

JackStouffer:
Why did you close this? Why not let @WalterBright work on this in this PR?

schveiguy:
> Why not let @WalterBright work on this in this PR?

Because I am more familiar with the lifetime.d code.

I was working on making `_d_arraysetcapacity` `@trusted`, but I ran into a snag: the general case for `_d_arraysetcapacity` can call postblit. So we would have to solve that first, and even in that case, I wouldn't mark the low-level function `@trusted`. But the specific call used by capacity should be safe.

@WalterBright, please add unit test, then we can merge.

schveiguy:
> the general case for _d_arraysetcapacity can call postblit.

Note, this is the case for reserve. It really should not be `@trusted` unless the postblit is `@safe`.

WalterBright:
Added unit test.

schveiguy:
Auto-merge toggled on

schveiguy:
@WalterBright thanks!

PR #1589, opened 2016-06-09
fix Issue 14117 - core.atomic should be @safe

WalterBright:
Added @trusted and @nogc where necessary, too.

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -2,17 +2,12 @@
  * The atomic module provides basic support for lock-free
  * concurrent programming.
  *
- * Copyright: Copyright Sean Kelly 2005 - 2010.
+ * Copyright: Copyright Sean Kelly 2005 - 2016.
  * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
  * Authors:   Sean Kelly, Alex Rønne Petersen
  * Source:    $(DRUNTIMESRC core/_atomic.d)
  */
 
-/*          Copyright Sean Kelly 2005 - 2010.
- * Distributed under the Boost Software License, Version 1.0.
- *    (See accompanying file LICENSE or copy at
- *          http://www.boost.org/LICENSE_1_0.txt)
- */
 module core.atomic;
 
 version( D_InlineAsm_X86 )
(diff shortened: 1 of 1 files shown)


src/core/atomic.d
```diff
@@ -1341,16 +1336,15 @@ version( unittest )
...
         testCAS!(T)( val );
         testLoadStore!(MemoryOrder.seq, T)( val );
         testLoadStore!(MemoryOrder.raw, T)( val );
     }
 
-
-    //@@@BUG@@@ http://d.puremagic.com/issues/show_bug.cgi?id=8081
-    /+pure nothrow+/ unittest
+@safe
+    pure nothrow unittest
```
ZombineDev: Nitpick: strange formatting

ZombineDev:
LGTM

ZombineDev:
BTW I was thinking about doing the same for core.sync.* but @nogc is hard because all the synchronization primitives are implemented as classes. Furthermore missing support for something like head_const / head_immutable prevents making
ReadWriteMutex.reader and ReadWriteMutex.writter @safe shared (otherwise you would need to use atomicLoad, which is clearly unnecessary).

src/core/atomic.d
schveiguy: The added `@safe` here is causing problems in some builds. But I can't figure out why. The error is on line 1386, can't take the address of a local in a `@safe` function.

src/core/atomic.d
```diff
@@ -161,34 +156,34 @@ version( CoreDdoc )
...
 
-    private HeadUnshared!(T) atomicFetchSub(T, V1)( ref shared T val, V1 mod ) pure nothrow @nogc
+    private HeadUnshared!(T) atomicFetchSub(T, V1)( ref shared T val, V1 mod ) pure nothrow @nogc @safe
         if( T.sizeof <= 4 && V1.sizeof <= 4)
     {
         return atomicFetchAdd(val, -mod);
     }
 
-    HeadUnshared!(T) atomicOp(string op, T, V1)( ref shared T val, V1 mod ) pure nothrow @nogc
+    HeadUnshared!(T) atomicOp(string op, T, V1)( ref shared T val, V1 mod ) pure nothrow @nogc @trusted
```
klickverbot: What about `T` being a struct with overloaded operators?

src/core/atomic.d
WalterBright: fixed

src/core/atomic.d
WalterBright: As far as I can tell, the only thing done with `val` is take its address. No T operators.

There's a mistake in the design of `cas` in that it takes a pointer rather than a reference, this messes with safety checks. But fixing that is beyond the scope of this PR for now.

klickverbot: Huh? This comment is on `atomicOp`, which might e.g. execute `return atomicLoad(…) + mod;`.

klickverbot: (i.e. I don't see why the `op` parts of the function can't end up forwarding to custom operators on `T`/`V1`, which is what you need to be able to rule out to make `@trusted` correct)

WalterBright:
Switching argument passing from `&` to `ref` resolved a lot of safety violations.

rainers:
> Switching argument passing from & to ref resolved a lot of safety violations.

Why is that? Pointers are just as good and allowed in safe code as long as they are not modified. Do you mean it's simpler to manually verify?

Otherwise LGTM.

WalterBright:
Taking the address of a local is not allowed in @safe code, but taking a ref is. That's because of the escaping pointer problem.

rainers:
> Taking the address of a local is not allowed in @safe code, but taking a ref is.

Thanks for explanation. That's sometimes hard to verify, though (e.g. `return localArray[].ptr;`).

rainers:
Auto-merge toggled on

PR #1591, opened 2016-06-13
Exposes minimal GC stats via gc_usage

Dicebot:
https://issues.dlang.org/show_bug.cgi?id=16019

This is one very minor utility that we have in Sociomantic GC API that makes it different from current upstream one - and which is used quite extensively. One of most important use cases of having these stats is to be able write a test that memory consumption is O(1) for a function that can't be `@nogc`.

Ping @MartinNowak

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -130,6 +130,8 @@ private
     extern (C) void*   gc_addrOf( void* p ) pure nothrow;
     extern (C) size_t  gc_sizeOf( void* p ) pure nothrow;
 
+    extern (C) void gc_usage( out size_t used, out size_t free) nothrow;
+
     struct BlkInfo_
     {
         void*  base;
@@ -787,5 +789,21 @@ struct GC
     static void runFinalizers( in void[] segment )
     {
         gc_runFinalizers( segment );
+    }
+
+    /**
+        Basic stats about current GC state
+
+        Exact meaning of reported stats may very depending on underlying
+        GC implementation but must be consistent for the same implementation.
+
+        Params:
+            used = will contain amount of bytes marked in GC as allocated
+            free = will contain amount of bytes allocated from OS but marked
+                in GC as free
+     */
+    static void usage( out size_t used, out size_t free) nothrow
+    {
+        gc_usage(used, free);
     }
 }
--- a/src/gc/proxy.d
+++ b/src/gc/proxy.d
@@ -53,6 +53,7 @@ private
             size_t  function(void*) gc_sizeOf;
 
             BlkInfo function(void*) gc_query;
+            void function(out size_t, out size_t) gc_usage;
 
             void function(void*) gc_addRoot;
             void function(void*, size_t, const TypeInfo ti) gc_addRange;
@@ -91,6 +92,7 @@ private
         pthis.gc_sizeOf = &gc_sizeOf;
 
         pthis.gc_query = &gc_query;
+        pthis.gc_usage = &gc_usage;
 
         pthis.gc_addRoot = &gc_addRoot;
         pthis.gc_addRange = &gc_addRange;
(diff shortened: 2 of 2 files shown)


schveiguy:
Whitespace issue.

schveiguy:
1. Why does the more general `gc_stats` function not suffice?
2. I'm not sure `freeblocks` is of the same unit as `usedsize`. Your function gives the impression that `free` is in bytes, but `GCStats` says `freeblocks` is in "blocks", whatever that means.

schveiguy:
> It is currently not exposed being current GC implementation detail.

What I think may be a better option than creating a redundant, less powerful, but more general and stable function, is to document which of the `GCStats` members are finalized. So for example, free bytes and used bytes is likely a stat we require ALL GC implementations to provide, so they must fill in those pieces. If we want to change the other stats, or add more (or make them customizable, etc.), then that can happen without affecting code that is written to use `GCStats.freebytes` and `GCStats.usedbytes`. You just need an alias from `freelistbytes` to more general `freebytes`.

The gc_stats function itself can remain undocumented for now (and sociomantic just knows that it's there). This can't be any worse than your current situation.

src/core/memory.d
```diff
@@ -130,6 +130,8 @@ private
     extern (C) void*   gc_addrOf( void* p ) pure nothrow;
     extern (C) size_t  gc_sizeOf( void* p ) pure nothrow;
 
+    extern (C) void gc_usage( out size_t used, out size_t free) nothrow;
```
MartinNowak: Let's return a struct instead of using out parameters.

MartinNowak: In fact it's already there, so please just forward to gc_stats here.

MartinNowak:
I'd opt for
```d
struct GCStats
{
    size_t heapsize;        /// total size of GC heap
    size_t freesize;        /// free bytes on the GC heap (might only get updated after a collection)
}
```
atm.

src/core/memory.d
Dicebot: > In fact it's already there

? There is no `GCStats` in `core.memory`, it is only defined in gc stats module. This is why I am not sure about how to proceed here.

How about introducing new `core.gcapi` module that can be imported by both GC implementations and `core.memory`? This whole business of manually maintaining `extern(C)` declarations in sync feels like legacy.

Dicebot: Ping @MartinNowak

ZombineDev:
@Dicebot you may want to have a look at PR https://github.com/dlang/druntime/pull/1581 as there will be a merge conflict in of the two PRs, depending on which PR is merged first.

rainers:
> So what API would you envision for actual public function? 

A common technique is to have a field with flags for valid info fields in the info struct, for example https://technet.microsoft.com/en-us/library/bb146197(v=vs.110).aspx
The call could also just request the fields it is interested in, so you don't have to pay for extensive heap analysis if you just want to get the overall used memory.


> One additional problem with exposing GCStats is that right now relation between core.memory

Importing `core` from the GC should be no problem, it already does that for other modules.

src/core/memory.d
rainers: > How about introducing new core.gcapi module that can be imported by both GC implementations and core.memory? 

That's more or less what gc.gcinterface does in #1581.

> This whole business of manually maintaining extern(C) declarations in sync feels like legacy.

I agree. If the extern(C) interface is to stay (it is barely useful to begin with), it should be opt-in provided by a GC proxy. Having the calls go through an extern(C) interface also inhibits some inlining for the common case.

ZombineDev:
> It looks semantically wrong though

I thought so too, but it turns out it is by design. See:
https://github.com/dlang/druntime/pull/1581#issuecomment-227846234

MartinNowak:
>> A common technique is to have a field with flags for valid info fields ...

> Sounds like a good approach indeed.

Let's not overcomplicate things, we can go with a minimal API first and extend it later, e.g. by adding a name argument and a separate float value for that field.

> so you don't have to pay for extensive heap analysis if you just want to get the overall used memory

We won't add any expensive stat functions. The values might be outdated and/or estimates.

The path forward here is the following.
- change the fields of GCStats (this has never been public API)
- move gc.stats.GCStats to core.memory
- add core.memory.GC.stats to forward to gc.proxy.gc_stats
- reopen against master (new features don't go into stable)

PR #1593, opened 2016-06-15
[critical] change hashOf() to safe interface

WalterBright:
When this gets pulled, all users of hashOf() need to be changed, and the old way deprecated.

### DIFF
--- a/src/rt/util/hash.d
+++ b/src/rt/util/hash.d
@@ -1,15 +1,10 @@
 /**
- * This module contains the default hash implementation.
+ * The default hash implementation.
  *
- * Copyright: Copyright Sean Kelly 2009 - 2009.
+ * Copyright: Copyright Sean Kelly 2009 - 2016.
  * License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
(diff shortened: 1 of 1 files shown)


src/rt/util/hash.d
```diff
@@ -1,15 +1,10 @@
 /**
- * This module contains the default hash implementation.
+ * The default hash implementation.
  *
- * Copyright: Copyright Sean Kelly 2009 - 2009.
+ * Copyright: Copyright Sean Kelly 2009 - 2016.
```
wilzbach: How about `$(YEAR)`?

WalterBright: That would be inaccurate, as the year should be the last year the code was worked on rather than the year the documentation was generated.

WalterBright:
This turns out to be a critical bug fix, as many uses of hashOf() are incorrect in that they are passing '0' as the length of the hash!! Just goes to show the evils of default arguments coupled with multiple incompatible definitions of hashOf().

This doesn't actually fix the bugs I found, but is a precursor to getting them fixed.

Please don't delay pulling this.

schveiguy:
Auto-merge toggled on

PR #1595, opened 2016-06-17
use array interface to hashOf()

WalterBright:
Because the array interface is safer.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -214,11 +214,11 @@ class TypeInfo
     {
         import core.internal.traits : externDFunc;
         alias hashOf = externDFunc!("rt.util.hash.hashOf",
-                                    size_t function(const(void)*, size_t, size_t) @trusted pure nothrow);
+                                    size_t function(const(void)[], size_t) @trusted pure nothrow @nogc);
         try
         {
             auto data = this.toString();
-            return hashOf(data.ptr, data.length, 0);
+            return hashOf(data, 0);
         }
         catch (Throwable)
         {
@@ -255,7 +255,15 @@ class TypeInfo
         return ti && this.toString() == ti.toString();
     }
 
-    /// Returns a hash of the instance of a type.
+    /**
+     * Computes a hash of the instance of a type.
+     * Params:
+     *    p = pointer to start of instance of the type
+     * Returns:
+     *    the hash
+     * Bugs:
+     *    fix https://issues.dlang.org/show_bug.cgi?id=12516 e.g. by changing this to a truly safe interface.
+     */
     size_t getHash(in void* p) @trusted nothrow const { return cast(size_t)p; }
 
     /// Compares two instances for equality.
@@ -1101,7 +1109,7 @@ class TypeInfo_Struct : TypeInfo
                     this.initializer().length == s.initializer().length;
     }
 
-    override size_t getHash(in void* p) @safe pure nothrow const
+    override size_t getHash(in void* p) @trusted pure nothrow const
     {
         assert(p);
         if (xtoHash)
@@ -1112,8 +1120,8 @@ class TypeInfo_Struct : TypeInfo
         {
             import core.internal.traits : externDFunc;
             alias hashOf = externDFunc!("rt.util.hash.hashOf",
-                                        size_t function(const(void)*, size_t, size_t) @trusted pure nothrow);
-            return hashOf(p, initializer().length, 0);
+                                        size_t function(const(void)[], size_t) @trusted pure nothrow @nogc);
+            return hashOf(p[0 .. initializer().length], 0);
         }
     }
 
@@ -3167,7 +3175,7 @@ struct Test
 size_t hashOf(T)(auto ref T arg, size_t seed = 0)
 {
     import core.internal.hash;
-    return core.internal.hash.hashOf(arg, seed);
+    return core.internal.hash.hashOf((cast(void*)&arg)[0 .. T.sizeof], seed);
 }
 
 bool _xopEquals(in void*, in void*)
@@ -3238,9 +3246,9 @@ private size_t getArrayHash(in TypeInfo element, in void* ptr, in size_t count) 
 
     import core.internal.traits : externDFunc;
     alias hashOf = externDFunc!("rt.util.hash.hashOf",
-                                size_t function(const(void)*, size_t, size_t) @trusted pure nothrow);
+                                size_t function(const(void)[], size_t) @trusted pure nothrow @nogc);
     if(!hasCustomToHash(element))
-        return hashOf(ptr, elementSize * count, 0);
+        return hashOf(ptr[0 .. elementSize * count], 0);
 
     size_t hash = 0;
     foreach(size_t i; 0 .. count)
--- a/src/rt/typeinfo/ti_Ag.d
+++ b/src/rt/typeinfo/ti_Ag.d
@@ -27,8 +27,8 @@ class TypeInfo_Ag : TypeInfo_Array
 
     override size_t getHash(in void* p) @trusted const
     {
-        byte[] s = *cast(byte[]*)p;
-        return rt.util.hash.hashOf(s.ptr, s.length * byte.sizeof);
+        const s = *cast(const void[]*)p;
+        return rt.util.hash.hashOf(s, 0);
     }
 
     override bool equals(in void* p1, in void* p2) const
--- a/src/rt/typeinfo/ti_Aint.d
+++ b/src/rt/typeinfo/ti_Aint.d
@@ -28,8 +28,8 @@ class TypeInfo_Ai : TypeInfo_Array
 
     override size_t getHash(in void* p) @trusted const
     {
-        int[] s = *cast(int[]*)p;
-        return rt.util.hash.hashOf(s.ptr, s.length * int.sizeof);
+        const s = *cast(const int[]*)p;
+        return rt.util.hash.hashOf(s, 0);
     }
 
     override bool equals(in void* p1, in void* p2) const
--- a/src/rt/typeinfo/ti_Along.d
+++ b/src/rt/typeinfo/ti_Along.d
@@ -26,8 +26,8 @@ class TypeInfo_Al : TypeInfo_Array
 
     override size_t getHash(in void* p) @trusted const
     {
-        long[] s = *cast(long[]*)p;
-        return rt.util.hash.hashOf(s.ptr, s.length * long.sizeof);
+        const s = *cast(const long[]*)p;
+        return rt.util.hash.hashOf(s, 0);
     }
 
     override bool equals(in void* p1, in void* p2) const
--- a/src/rt/typeinfo/ti_Ashort.d
+++ b/src/rt/typeinfo/ti_Ashort.d
@@ -26,8 +26,8 @@ class TypeInfo_As : TypeInfo_Array
 
     override size_t getHash(in void* p) @trusted const
     {
-        short[] s = *cast(short[]*)p;
-        return rt.util.hash.hashOf(s.ptr, s.length * short.sizeof);
+        const s = *cast(const short[]*)p;
+        return rt.util.hash.hashOf(s, 0);
     }
 
     override bool equals(in void* p1, in void* p2) const
--- a/src/rt/util/hash.d
+++ b/src/rt/util/hash.d
@@ -17,14 +17,14 @@ version( AnyX86 )
     version = HasUnalignedOps;
 
 
-@trusted pure nothrow
+@trusted pure nothrow @nogc
 size_t hashOf( const(void)[] buf, size_t seed = 0 )
 {
     return hashOf(buf.ptr, buf.length, seed);
 }
 
-@trusted pure nothrow
-size_t hashOf( const(void)* buf, size_t len, size_t seed = 0 )
+@system pure nothrow @nogc
+size_t hashOf( const(void)* buf, size_t len, size_t seed )
 {
     /*
      * This is Paul Hsieh's SuperFastHash algorithm, described here:
@@ -32,7 +32,7 @@ size_t hashOf( const(void)* buf, size_t len, size_t seed = 0 )
      * It is protected by the following open source license:
      *   http://www.azillionmonkeys.com/qed/weblicense.html
      */
-    static uint get16bits( const (ubyte)* x ) pure nothrow
+    static uint get16bits( const (ubyte)* x ) pure nothrow @nogc
     {
         // CTFE doesn't support casting ubyte* -> ushort*, so revert to
         // per-byte access when in CTFE.
@@ -96,14 +96,14 @@ size_t hashOf( const(void)* buf, size_t len, size_t seed = 0 )
 }
 
 // Check that hashOf works with CTFE
-unittest
+@nogc nothrow pure unittest
 {
     size_t ctfeHash(string x)
     {
-        return hashOf(x.ptr, x.length);
+        return hashOf(x.ptr, x.length, 0);
     }
 
     enum test_str = "Sample string";
     enum size_t hashVal = ctfeHash(test_str);
-    assert(hashVal == hashOf(test_str.ptr, test_str.length));
+    assert(hashVal == hashOf(test_str.ptr, test_str.length, 0));
 }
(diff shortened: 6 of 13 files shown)


src/object.d
```diff
@@ -1101,7 +1101,7 @@ class TypeInfo_Struct : TypeInfo
                     this.initializer().length == s.initializer().length;
     }
 
-    override size_t getHash(in void* p) @safe pure nothrow const
+    override size_t getHash(in void* p) @trusted pure nothrow const
```
WalterBright: Because the slicing is an unsafe operation. The older code was unsafe as well.

src/object.d
```diff
@@ -3167,7 +3167,7 @@ struct Test
 size_t hashOf(T)(auto ref T arg, size_t seed = 0)
 {
     import core.internal.hash;
-    return core.internal.hash.hashOf(arg, seed);
```
WalterBright: Note that this code has a bug in it - the `seed` is interpreted as the length of `arg` !!

src/object.d
klickverbot: How can this interface (dereferencing a `void*`) even be `@safe` in the first place?

ZombineDev:
The changes LGTM. Interfaces with `void*` don't look `@safe` to me either, but I suggest we leave this topic for discussion in other PR, as this one is a clear improvement.

src/rt/util/hash.d
```diff
@@ -24,7 +24,7 @@ size_t hashOf( const(void)[] buf, size_t seed = 0 )
 }
 
 @trusted pure nothrow
-size_t hashOf( const(void)* buf, size_t len, size_t seed = 0 )
```
WalterBright: get rid of the default argument, it just causes bugs

src/object.d
WalterBright: `getHash` is supposed to take a pointer to the object matching the `TypeInfo`.

klickverbot: "is supposed" is something entirely different than "is guaranteed to". You could of course argue that the interface was broken before, but it still remains a safety violation that should at least be recorded in Bugzilla (and subsequently be deprecated/made un-@safe).

klickverbot: > Please do add a bugzilla issue for it. Such is beyond the scope of this PR, however, and should not impede it.

It's your responsibility. The PR adds a `@trusted` annotation that is clearly wrong.

WalterBright: The @safe was equally wrong. This PR does **not** address (or change) the interface to getHash(), the change was made to reflect the unsafe operation inside the function. The issue you brought up is with the interface to getHash(), not the interface to hashOf().

> your responsibility

It is the responsibility of people who find bugs to post them to bugzilla. The bug you found has nothing to do with this PR, which is to fix the interface to `hashOf()`. You should take the credit for finding the issue by posting it to bugzilla.

Meanwhile, this PR is **only** concerned with the interface to `hashOf`.

klickverbot: > The @safe was equally wrong.

No; `@safe` can never be wrong as in breaking the type system, it can just cause code not to compile. You are adding a new, incorrect instance of `@trusted`, which *does* add an extra type system violation.

As I'm sure you realise, this is a huge deal because instances where things are mistakenly marked as `@safe` can be detected automatically, which is impossible for mistaken uses of `@trusted`. The only way to find those is to painstakingly go through all occurrences of `@trusted` and manually audit each one of them.

Just trust me on that one. I've found a large share of the Phobos safety bugs you went through recently, this PR would add another such bug, I point it out, you fix/document it *before* it goes in. That's how pull request review works. You can't justify knowingly introducing an issue without any documentation, etc. by pointing out that the code base contained bugs before.

WalterBright:
It is not incorrect. The revised code will not compile with @safe, even though the original, despite being marked @safe, was not detected as being unsafe. The unsafety was in the interface to hashOf(). By fixing the interface to hashOf(), the unsafety of the getHash() interface is exposed.

This is similar to a discussion I had with @yebblies some time back about adding const correctness incrementally. Adding const tends to be viral, i.e. an all-or-nothing proposition. But trying to const-correct the entire program in on PR is thoroughly impractical, so I was doing it incrementally by adding unsafe casts here and there to satisfy the type system.

As const spread through the program in successive PRs, the casts were removed.

yebblies:
> As const spread through the program in successive PRs, the casts were removed.

Are they all gone now?

Of course, you could just do what every other contributor does and FINISH the changes in a branch before pushing little bits to master.

WalterBright:
> I just found out that a report of this has existed since more than two years ago.

I just came here to report the same thing! So what's stopping you from pulling this?

schveiguy:
The correct marking of `getHash` is `@system`, period. No public API which takes a `void *` and assumes it points to valid memory of a given length should ever be marked `@trusted`. Whatever breaks should be then marked `@trusted` as follows:
1. If you are 100% certain that the `void *` you are passing into `getHash` is the correct type, then mark the function `@trusted`.
2. Otherwise, mark it `@system` and find the next level of breakage. Repeat algorithm.

Note that item 1 doesn't include any of the calls being modified here.

Please, let's not obfuscate where the problem lies. If this breaks code, it's code worth breaking. Anyone who cares about `@safe`-ty doesn't want the compiler to just gloss over such details.

WalterBright:
Redoing the interface is `getHash` and its consequences are way beyond the scope of this PR. This PR is solely about fixing the interface to `hashOf, which already affects 13 files.

WalterBright:
This PR does not change the interface to getHash. Trying to lump this PR in with another problem is a mistake and completely off-topic. If you want to raise the visibility of https://issues.dlang.org/show_bug.cgi?id=12516 that's fine, but it has nothing to do with this PR and should not block it.

WalterBright:
> Yes, the ticket does exist, but you neither mentioned it in code,

That's not what we do. Bug reports go in Bugzilla.

> nor in the commit message (where I would really want to see it at the very least).

It has nothing to do with this PR. It is an entirely separate problem.

schveiguy:
> You are adding an incorrect `@trusted` attribute where there was none before.

IMO, a `@trusted` function that can be abused is no worse or better than a `@safe` function that can be abused. Whether `getHash` commits the egregious behavior or `hashOf` does, that is not the issue. This PR does nothing to further safety. Except for the actual bug fix, this should be closed.

WalterBright:
>  why the broken code was introduced in the first place.

Broken code was not introduced by this PR. If you believe it does, please re-examine it.

Furthermore, this is a misunderstanding of @trusted. @trusted is a statement about the contents of the function, not the interface. The interface did not change, and the function was never safe, and it was never safe because of the unsafe interface to hashOf. Fixing hashOf, the lowest level code, is the first step to improving the safety of the call chain.

> a  @trusted  function that can be abused is no worse or better than a  @safe  function that can be abused.

That is correct.

> This PR does nothing to further safety.

This is incorrect. It fixes the unsafe interface of `hashOf`. One thing at a time per PR.

> this should be closed.

An all-or-nothing approach to fixing safety problems means that zero progress can be made.

schveiguy:
> It fixes the unsafe interface of `hashOf`

I think you mean something different, like that it avoids using the unsafe interface?

But in any case, this doesn't push the problem to the right level. Passing a memory-invalid array to a `@safe` function isn't any safer. The right level is where you know the `void *` actually points at an `int` (or whatever). Intermediate fixes are rubble bouncing.

> An all-or-nothing approach to fixing safety problems means that zero progress can be made.

There is no reason to avoid calling the unsafe interface when the operation itself is actually unsafe. IMO, this PR is already achieving zero progress.

It doesn't have to be all or nothing. Fix calls to `getHash` to be properly marked `@trusted` at the right level, and then you can properly mark `getHash` as `@system`, because it is.

WalterBright:
> But in any case, this doesn't push the problem to the right level.

Of course it does. There's no reason for hashOf to be trusted instead of safe, and this fixes the problem with hashOf.

> I think you'll find it hard to argue that this is not incorrect code.

The problem was always there, it was just hidden in the call to hashOf due to the unsafe interface to hashOf. This PR does not introduce a problem.

> In any case, would you mind separating out the actual one-line bug fix (zero length vs. seed) into a second PR so I can merge it right away?

No, because it's time to settle this point, not defer it. The hashOf interface needs to be fixed, and this PR fixes it and does **NOT** degrade anything else.

Adding a requirement for PRs to only be "full stack" fixes is unworkable in general. What needs to happen is progress, piece by piece. Otherwise druntime/phobos will simply **never** get fixed to be safe.

It's exactly the same issue one runs into when trying to retrofit const-correctness into an existing code base. Doing it incrementally, bottom up, is the only practical way, and requires the introduction of temporary `cast(char*)` bits of code. If a principled stand is taken against that, the reality is the code will never become const-correct.

andralex:
I have two questions about this:

* Are there improvements in the compiler that detect `@safe` is misapplied to `hashOf` and render it incompilable?
* What is the plan after this step?

src/rt/util/hash.d
```diff
@@ -17,22 +17,22 @@ version( AnyX86 )
...
-@trusted pure nothrow
+@trusted pure nothrow @nogc
 size_t hashOf( const(void)[] buf, size_t seed = 0 )
 {
     return hashOf(buf.ptr, buf.length, seed);
 }
 
-@trusted pure nothrow
-size_t hashOf( const(void)* buf, size_t len, size_t seed = 0 )
+@trusted pure nothrow @nogc
```
klickverbot: (Somewhat tangential to the main discussion, this should be changed to `@system` right away. `rt.util` is a druntime-internal module, and there is nothing to be gained from keeping the broken attribute around.)

klickverbot: (And unless you make this `@system`, you can hardly claim that it makes any code more safe – more `@safe`, anyway; there is an argument to be made that removing the default value already makes client code more resilient against bugs.)

WalterBright:
> your change spreads out the mistakenly @trusted bits further, 

No, it does not. It simply identifies code that should have been marked @trusted to begin with, because it was not @safe despite being marked @safe.

> It was only a suggestion to resolve this impasse

The consequences will be to avoid making a decision and never fix hashOf.

> I'm not going to comment on the (non-)applicability of your comparison to const-correctness

It's an apt comparison. Insisting on a "full stack" approach to fixing viral issues means nothing will get done.

> your apparent unwillingness to even minimally document your work.

I see no purpose to documenting things that are listed as bugs in bugzilla. You're putting forth a new policy from existing practice. You might as well require that all phobos bugs in bugzilla also get PRs to document them in the source code. Remember, this PR is not about fixing bugs in the interface to getHash, it is about making the interface to hashOf safe. Conflating this with other problems is not how PRs are supposed to work. Grepping a codebase for FIXME is the old way of doing things. We use bugzilla now.

And besides, simply marking it as @trusted is supposed to be a flag for reviewers. It is not hidden.

WalterBright:
> Are there improvements in the compiler that detect @safe is misapplied to hashOf and render it incompilable?

The @safe was misapplied to getHash, not hashOf. hashOf was marked @trusted although it did not provide a trusted interface to what it was doing. It's like marking `free(void*)` as trusted. Switching to an array interface makes it possible to be trusted.

> What is the plan after this step?

My plan was grepping the druntime sources for `.ptr` and reviewing/improving them one by one. I'd be surprised if there wasn't more crap in there like `hashOf`.

src/rt/util/hash.d
WalterBright: I left the old function in there because removing it may break existing code, adding system to it may break existing code, and breaking existing code should be done separately if possible. Not lumped in with other things. Even if `hashOf` is an internal function, there's nothing stopping anyone from using it. There being another `hashOf` in object.d is unfortunate and further obfuscates things, but again, that is not the topic of this PR.

WalterBright:
Keep in mind that this PR is very simple and straightforward:

It adds an overload to `hashOf` that provides a safe array interface rather than an unsafe ptr/length interface. All callers to it in druntime were fixed to use the new array interface.

It's a self-contained, incremental step towards improving the overall safety of druntime/Phobos.

src/rt/util/hash.d
klickverbot: > Even if hashOf is an internal function, there's nothing stopping anyone from using it.

`rt.*` is not distributed to users.

schveiguy:
> It adds an overload to hashOf that provides a safe array interface rather than an unsafe ptr/length interface. All callers to it in druntime were fixed to use the new array interface.

This PR does not do that. I already pulled #1593

src/rt/typeinfo/ti_Ag.d
```diff
@@ -27,8 +27,8 @@ class TypeInfo_Ag : TypeInfo_Array
 
     override size_t getHash(in void* p) @trusted const
     {
-        byte[] s = *cast(byte[]*)p;
-        return rt.util.hash.hashOf(s.ptr, s.length * byte.sizeof);
+        const s = *cast(void[]*)p;
```
andralex: please cast to `const void[]` instead

src/rt/typeinfo/ti_Aint.d
```diff
@@ -28,8 +28,8 @@ class TypeInfo_Ai : TypeInfo_Array
 
     override size_t getHash(in void* p) @trusted const
     {
-        int[] s = *cast(int[]*)p;
-        return rt.util.hash.hashOf(s.ptr, s.length * int.sizeof);
+        const s = *cast(int[]*)p;
```
andralex: please cast to `const int[]` instead

src/rt/typeinfo/ti_Along.d
```diff
@@ -26,8 +26,8 @@ class TypeInfo_Al : TypeInfo_Array
 
     override size_t getHash(in void* p) @trusted const
     {
-        long[] s = *cast(long[]*)p;
-        return rt.util.hash.hashOf(s.ptr, s.length * long.sizeof);
+        const s = *cast(long[]*)p;
```
andralex: please cast to `const long[]` instead

src/rt/typeinfo/ti_Ashort.d
```diff
@@ -26,8 +26,8 @@ class TypeInfo_As : TypeInfo_Array
 
     override size_t getHash(in void* p) @trusted const
     {
-        short[] s = *cast(short[]*)p;
-        return rt.util.hash.hashOf(s.ptr, s.length * short.sizeof);
+        const s = *cast(short[]*)p;
```
andralex: please cast to `const short[]` instead

src/rt/util/hash.d
andralex: @WalterBright response?

WalterBright:
> This PR does not do that. I already pulled #1593

You're right. This one fixes the calls to it.

> it can only be mistakenly allowed due to compiler bugs. Neither of it is the case here.

This is incorrect. The problem was that hashOf was marked @trusted, even though hashOf did not have a safe interface. Trusted can be used to hide unsafety from the compiler, and was in this case, and there is no compiler technology that can detect that in the general case.

> the code shouldn't have been marked @trusted to begin with, but @system. I suspect (and dearly hope) that this is obvious to you

It is obvious to me. But adding @system to getHash is an API-breaking change, and needs to be in a separate PR and be evaluated for its breaking effects before doing so. Such is beyond the scope of this PR, which is solely about hashOf.

andralex:
This is a viable PR that marks progress on an issue. The intensity of the discussion is somewhat proportional; clearly something hanky-panky was going on in the virtual `getHash` method. Currently we can't fix that without breaking compatibility, but it's self-evidently good to mark that as `@trusted` instead of `@safe` (e.g. when searching for fishy code you grep for the former, not the latter).

@WalterBright please add this just before the `getHash` declaration:

```D
// TODO: fix https://issues.dlang.org/show_bug.cgi?id=12516 e.g. by changing this to a truly safe interface.
```

I'll pull after @WalterBright addresses the nits.

WalterBright:
@andralex done

src/rt/util/hash.d
WalterBright: Anyone can import it - it's in `import/core/internal/hash.d`

Changing that is properly a separate PR.

klickverbot: > Ah, you're right, I had it confused.

No worries – all the functions having (almost) identical names make it easy to mix them up.

klickverbot: > Ah, you're right, I had it confused.

No worries – all the functions having (almost) identical names make it easy to mix them up.

andralex:
alrighty then

WalterBright:
thanks!

PR #1596, opened 2016-06-17
remove long dead code from arraycast.d

WalterBright:
So old it is from D1.


leandro-lucarella-sociomantic:
Who said D1 is old? :trollface:

WalterBright:
> Who said D1 is old?

The dead code was so old it was even `version (none)` in D1!

PR #1597, opened 2016-06-18
Walter Andrei Action List

burner:
http://forum.dlang.org/post/nk37j7$2h6q$1@digitalmars.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,3 +57,8 @@ Building
 --------
 
 See the [wiki page](http://wiki.dlang.org/Building_DMD) for build instructions.
+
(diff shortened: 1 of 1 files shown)


WalterBright:
I'm not too enamored with the wording on that. It's a bit hostile. It isn't a "Walter Andrei" list, either, it is a "Druntime" action list.

burner:
https://github.com/dlang/phobos/pull/4435 same complaint there. I asked @wilzbach to contribute a "nicer" version. But I have to say, it is still your action list not Druntime's.

PR #1598, opened 2016-06-26
update comments in object.d

WalterBright:
1. remove redundant header
2. replace open BUG comments with a more useful link to the bug report
3. remove fixed BUG comments

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -8,11 +8,6 @@
  * Authors:   Walter Bright, Sean Kelly
  */
 
-/*          Copyright Digital Mars 2000 - 2011.
- * Distributed under the Boost Software License, Version 1.0.
- *    (See accompanying file LICENSE or copy at
- *          http://www.boost.org/LICENSE_1_0.txt)
- */
 module object;
 
 private
@@ -2125,7 +2120,7 @@ pure nothrow unittest
     }
 }
 
-pure /*nothrow @@@BUG5555@@@*/ unittest
+pure /*nothrow */ unittest
 {
     auto a = [ 1:"one", 2:"two", 3:"three" ];
     auto b = a.dup;
@@ -2477,7 +2472,7 @@ void _postblitRecurse(E, size_t n)(ref E[n] arr)
 
     struct InnerMiddle {}
 
-    version(none) // @@@BUG@@@ 14242
+    version(none) // https://issues.dlang.org/show_bug.cgi?id=14242
     struct InnerElement
     {
         static char counter = '1';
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -3257,8 +3252,6 @@ private size_t getArrayHash(in TypeInfo element, in void* ptr, in size_t count)
 }
 
 
-// @@@BUG5835@@@ tests:
-
```
schveiguy: Why eliminate this ref?

WalterBright: I didn't see a point to it.

https://issues.dlang.org/show_bug.cgi?id=5835

After fixes have been pulled, it takes a strong reason to keep references to the bugzilla entry in the source code. (Otherwise things would fill up with clutter.)

schveiguy: Usually when I add a unit test to verify a fix, I tag the test with the issue number. I thought the tag should last forever as a reminder of why the test is there. This is different than the comments identifying why some workaround is employed. In any case this is inconsistent with your leaving the bug ref for 10574 above

WalterBright: https://issues.dlang.org/show_bug.cgi?id=10574 is still open, https://issues.dlang.org/show_bug.cgi?id=5835 is closed.

> the tag should last forever as a reminder of why the test is there.

Only if there's something special about it.

schveiguy: It's not that there's something special, it's a link back to the impetus to create it. For example, if I looked at the unit tests for 10574, I would have no idea why someone added those tests. The comment allows me to avoid searching for it.

A comment is there to help solidify the reason for having it. Indeed, this unit test is confusing without the context of the comment.

BTW, why are the asserts marked `//fails` ? Clearly they don't fail, since they are passing unit tests!

WalterBright: > it's a link back to the impetus to create it

This is not normal practice - if it were, the source code would be packed with them. There needs to be a special reason for such a link, and you agreed there isn't anything special here. It's the same as removing dead code - the history is what github is for.

>  this unit test is confusing without the context of the comment.

The bugzilla entry is a meandering thread. If you want to suggest a proper comment, I'd be happy to put it in. The bugzilla comment is not helpful.

> why are the asserts marked  //fails  ?

I don't know. It's not in the scope of this PR.

schveiguy: > If you want to suggest a proper comment, I'd be happy to put it in.

"ensure `TypeInfo_Array.getHash` uses element hash functions instead of hashing array data"



> I don't know. It's not in the scope of this PR.

Can you remove them while you are adding the explanation suggested above? If there is a definition of cruft, this is it -- a comment that has no explanation, and is obviously untrue.

aG0aep6G: > This is not normal practice

But it is. We put bugzilla issue numbers next to corresponding unittests all the time.

WalterBright: Rarely compared with the number of bugzilla issues fixed & closed.

schveiguy:
Auto-merge toggled on

PR #1599, opened 2016-06-26
remove old hashOf overload

WalterBright:
1. remove the old unsafe interface
2. remove default argument so it won't cause any more bugs
3. add decent unittest

### DIFF
--- a/src/rt/util/hash.d
+++ b/src/rt/util/hash.d
@@ -18,13 +18,7 @@ version( AnyX86 )
 
 
 @trusted pure nothrow @nogc
-size_t hashOf( const(void)[] buf, size_t seed = 0 )
-{
-    return hashOf(buf.ptr, buf.length, seed);
-}
-
-@system pure nothrow @nogc
-size_t hashOf( const(void)* buf, size_t len, size_t seed )
+size_t hashOf( const(void)[] buf, size_t seed )
 {
     /*
      * This is Paul Hsieh's SuperFastHash algorithm, described here:
@@ -47,7 +41,8 @@ size_t hashOf( const(void)* buf, size_t len, size_t seed )
 
     // NOTE: SuperFastHash normally starts with a zero hash value.  The seed
     //       value was incorporated to allow chaining.
-    auto data = cast(const (ubyte)*) buf;
+    auto data = cast(const(ubyte)*) buf.ptr;
+    auto len = buf.length;
     auto hash = seed;
 
     if( len == 0 || data is null )
@@ -95,15 +90,18 @@ size_t hashOf( const(void)* buf, size_t len, size_t seed )
     return hash;
 }
 
-// Check that hashOf works with CTFE
-@nogc nothrow pure unittest
+unittest
 {
-    size_t ctfeHash(string x)
-    {
-        return hashOf(x.ptr, x.length, 0);
-    }
-
     enum test_str = "Sample string";
-    enum size_t hashVal = ctfeHash(test_str);
-    assert(hashVal == hashOf(test_str.ptr, test_str.length, 0));
+    size_t hashval = hashOf(test_str, 5);
+
+    //import core.stdc.stdio;
+    //printf("hashval = %lld\n", cast(long)hashval);
+
+    if (hashval.sizeof == 4)
+        assert(hashval == 528740845);
+    else if (hashval.sizeof == 8)
+        assert(hashval == 8106800467257150594L);
+    else
+        assert(0);
 }
--- a/src/rt/typeinfo/ti_ulong.d
+++ b/src/rt/typeinfo/ti_ulong.d
@@ -28,7 +28,7 @@ class TypeInfo_m : TypeInfo
 
     override size_t getHash(in void* p)
     {
-        return rt.util.hash.hashOf(p[0 .. ulong.sizeof]);
+        return rt.util.hash.hashOf(p[0 .. ulong.sizeof], 0);
     }
 
     override bool equals(in void* p1, in void* p2)


src/rt/util/hash.d
```diff
@@ -95,15 +90,18 @@ size_t hashOf( const(void)* buf, size_t len, size_t seed )
...
 {
-    size_t ctfeHash(string x)
-    {
-        return hashOf(x.ptr, x.length, 0);
-    }
-
     enum test_str = "Sample string";
-    enum size_t hashVal = ctfeHash(test_str);
-    assert(hashVal == hashOf(test_str.ptr, test_str.length, 0));
+    size_t hashval = hashOf(test_str, 5);
```
ZombineDev: Can you leave this as `enum`? So the test would verify that the computation happens at CT.

src/rt/util/hash.d
```diff
@@ -95,15 +90,18 @@ size_t hashOf( const(void)* buf, size_t len, size_t seed )
...
+
+    //import core.stdc.stdio;
+    //printf("hashval = %lld\n", cast(long)hashval);
+
+    if (hashval.sizeof == 4)
```
ZombineDev: Can you change this and the code below to `static if` and `static assert`?

src/rt/util/hash.d
```diff
@@ -95,15 +90,18 @@ size_t hashOf( const(void)* buf, size_t len, size_t seed )
     return hash;
 }
 
-// Check that hashOf works with CTFE
-@nogc nothrow pure unittest
```
ZombineDev: Why are you removing those attributes?

src/rt/util/hash.d
WalterBright: No, because it doesn't work at CT due to a bug in the interpreter. There are no CT uses of it, either.

src/rt/util/hash.d
WalterBright: In this case, it makes no difference.

src/rt/util/hash.d
WalterBright: Because of the printf. hashOf has those attributes anyway, so there's nothing lost.

src/rt/util/hash.d
ZombineDev: Yeah the only difference is no unnecessary code included in the final executable, though probably that's not important for unittest code.

src/rt/util/hash.d
ZombineDev: Oh, that's too bad. Can we postpone this PR, until the bug is fixed, or that may take a long time to figure out?

src/rt/util/hash.d
ZombineDev: OK, but I don't see why the printf is necessary (and it can be put under `debug` so nothing is lost). Anyway can you just leave a link to bugzilla issue for the CTFE problem?

WalterBright: The printf is used to see what the hash value is so it can be put in the assert.
Attributes for unittests make sense when they are verifying correct attribute inference for templates. It is pointless here.

src/rt/util/hash.d
WalterBright: > Can we postpone this PR

We'll never make any progress doing things like that. CTFE is irrelevant to hashOf, there are no uses of it. One thing at a time.

src/rt/util/hash.d
WalterBright: The compiler removes the unnecessary code regardless.

src/rt/util/hash.d
ZombineDev: > One thing at a time

Alright, LGTM, then.

src/rt/util/hash.d
klickverbot: Let's keep it simple for unit test code.

src/rt/util/hash.d
klickverbot: Note that this is in contradiction to [your Phobos `@safe`-ty call to action](http://forum.dlang.org/post/nkn25t$btk$1@digitalmars.com).

src/rt/util/hash.d
WalterBright: Don't know what you mean by keep it simple.

src/rt/util/hash.d
WalterBright: Yes, I was uncomfortably aware of that when I removed the attributes. When I wrote the n.g. post, I had forgotten that the point of attributes on unittests is to test the attribute inference of the functions/templates it calls. In this case, the attributes of hashOf are fully specified, so adding them to the unittest does nothing, and is inconvenient because of the printf.

`debug` offers an escape from purity not @safe. Perhaps it would be a reasonable enhancement to add that.

And I'm gratified that somebody reads my broadsides :-)

WalterBright:
The CTFE issue: https://issues.dlang.org/show_bug.cgi?id=16207

src/rt/util/hash.d
klickverbot: @WalterBright: I.e. I'm agreeing with you.

WalterBright: > @WalterBright: I.e. I'm agreeing with you. ;)

I'm noting this on my calendar of momentous events!

src/rt/util/hash.d
ZombineDev: Having attributed unittests still has value (even when they don't verify attribute inference), because it makes it easier to completely replace one implementation with another and ensure that interface and behavior is unchanged, i.e. prevents breaking changes.
`debug` would help with `pure` and `printf` is already `nothrow` and `@nogc`, so only `@safe is problematic`, but could be easily worked around with a trusted lambda. But TBH I don't think `printf` adds any value (even for future maintenance) because its trivial.

ZombineDev:
> The CTFE issue: https://issues.dlang.org/show_bug.cgi?id=16207

Thanks!

MartinNowak:
Auto-merge toggled on

PR #1600, opened 2016-06-27
rt.util.utf: add attributes

WalterBright:
Trivial. Just added annotations that the compiler would pass.


andralex:
Auto-merge toggled on

schveiguy:
Note that we have to be careful about adding attributes "that pass", especially where `@safe` is involved. In this case, `toUtf16z` is returning a pointer to a null-terminated string. While the "trick" passed in [this other PR](https://github.com/dlang/druntime/pull/1616) works, it's disingenuous. A `@safe` function that returns a pointer is only supposed to be returning a pointer to the single element, not an array as `toUTF16z` is doing.

see this recent discussion: #2053 

I feel like `toUTF16z` was improperly marked `@safe`, but not sure what we can do about it now.

PR #1601, opened 2016-06-28
rt.util.random: add attributes

WalterBright:
trivial


andralex:
Auto-merge toggled on

PR #1602, opened 2016-06-28
Fix issue 16211 - Cyclic dependencies broken again

schveiguy:
~~Note: you may want to look just at the first commit. I left the old algorithm there to prevent the diff from comparing two unrelated functions. The second commit removes the old algorithm from the source.~~ No longer the case, there are a lot of commits, and the code has changed significantly since the first commit.

The new algorithm finds cycles that have an intermediate module in the cycle twice. For example:
`a -> b -> c -> b -> a`, where only a and c have module ctors, but b does not. the old algorithm would short circuit the evaluation at b the second time, thinking it was already visited. This unit test has been added.

Essentially, the real graph of modules is between only the modules with static ctors (relevant modules), and the "edges" are really which other relevant modules can be reached. The fact that you have to go through non-relevant modules isn't important, and the non-relevant modules can be used more than once. For this reason, the algorithm needs to be very different than a straight cycle detection algorithm.

The new algorithm is a copy of the original algorithm I wrote in 2010, but enhanced to deal with the current module system.

I also implemented a crude dijkstra search to try and print the shortest cycle possible given the two offending modules. Works rather well, actually.

Unfortunately, there is still an ugly wart of a linear search whenever a module index has to be looked up, because the module info is now all immutable (previous version stored the index inside the module info). Also note that a module can import another module multiple times (see https://issues.dlang.org/show_bug.cgi?id=16208), so I try to minimize the looking up, but I can't avoid it there without allocating an edge map.

Finally, I added a bit better mechanism for figuring out which unit test fails when doing module sorting, because it's all done within a wrapper (and the stack trace doesn't always show line numbers).

pinging @MartinNowak since you are most familiar with this code (and wrote the old version).

This requires dlang/phobos#4493

Note: even though this is a regression fix, this will likely cause code breakage outside phobos, which is why I did not target stable.


JackStouffer:
Why was this closed? The cycle detection is still broken.

schveiguy:
@JackStouffer correct. This should now pass

schveiguy:
Yay a windows specific cycle! I'll have to fix that first

src/rt/minfo.d
```diff
@@ -60,6 +60,339 @@ struct ModuleGroup
...
+        // use this to hold the error message.
+        string errmsg = null;
+
+        void print(string[] msgs...)
+        {
+            foreach (m; msgs)
+            {
+                // save to the error message. Note that if we are throwing an
+                // exception, we don't care about being careful with using
+                // stack memory. Just use the GC/runtime.
```
klickverbot: Can you be sure that everything is already brought up at this point when using druntime as a shared library itself?

schveiguy: I expect that @MartinNowak already solved that issue when he changed the code to using module groups. I think cycle detection is run on each shared objects' group, which can't have upstream dependencies.

schveiguy:
> but I'm curious how many more projects will discover issues with cycles now.

I expect there will be some. But it may be fixing bugs in their code! However, based on how blunt a tool cycle detection is, it's likely not helping at all (most cycle detection problems are frivolous). This is going to be controversial, but it's still a bug fix. Don't know how to do it any better, open to ideas.

schveiguy:
> GitHub auto-closed this

I suspected so, because the comment was weird. I wonder how that happened?

src/rt/minfo.d
schveiguy: Wait, I get what you are saying, the GC may not have been set up yet? In the old version, when it was decided to throw an Error with the cycle, the code used GC concatenation, so I assumed this was OK to use. Isn't the GC inside druntime, so I would think it is available by now?

schveiguy:
I'm concerned here -- this cycle doesn't seem to be windows specific. Need to investigate the reason why it doesn't fail on OSX or others.

schveiguy:
There appears to be a bug in dmd or druntime. On OSX, `std.internal.math.biguintcore` only lists `core.cpuid` as a depencency, but it clearly imports many other modules.

schveiguy:
Needs dlang/phobos#4571

schveiguy:
Nuts. New shiny object to chase. I'm going to get a proper Windows test build going before going any further

schveiguy:
OK, I think dlang/phobos#4580 will make this pass.

MartinNowak:
I'm a bit concerned about this going in the wrong direction. The cycle detection is already overly conservative and annoys people, b/c it forces them through weird hoops, even though their code just works fine. Just look through bugzilla how many issues we have related to this topic.

MartinNowak:
There were two reasons to replace the old cycle detector. The code was very hard to maintain, and the algorithm had a fairly bad runtime (O(N^2) IIRC). We shouldn't go back to that if possible.
Shouldn't it just suffice to tweak the conditions here https://github.com/dlang/druntime/blob/90bd014d38ec0bbc693660747bc8b5ede304896e/src/rt/minfo.d#L180-L195, e.g. marking a module w/o ctor but with imports as ctorstart?

schveiguy:
I had a realization last night -- imports in unit tests shouldn't play any role in cycles, because they cannot be called from static ctors. If we remove unit test imports from the cycle detection, then we can work around most of the problems I have fixed in Phobos. We still need to handle imports inside functions, because static ctors can call them.

In terms of how complex the algorithm has to be, it just has to be. Otherwise the cycle detection is annoying *and* incorrect.

The O(n^2) requirement is expected, as we need to construct a reachability graph between all nodes. My algorithm avoids the `O(n^2)` space requirements as much as possible (it still takes `O(n^2)` space, but only for reachability between relevant nodes).

schveiguy:
> marking a module w/o ctor but with imports as ctorstart?

No, this is how it was before I fixed it in 2010. It doesn't find cycles that involve a non-ctor-containing node twice. Such as a -> b -> c -> b -> a, where a and c have ctors. You can't count the irrelevant nodes during cycle detection, they are just a way to get to the next relevant node.

schveiguy:
> I think we need more support from the compiler to detect actual cycles, i.e. figure out symbols used in a ctor/dtor and just get their module dependencies.

this would be awesome.

schveiguy:
> imports in unit tests shouldn't play any role in cycles

@MartinNowak Just filed an enhancement on this, I figure might be an easy improvement instead of doing full-blown dependency detection: https://issues.dlang.org/show_bug.cgi?id=16265

MartinNowak:
> imports in unit tests shouldn't play any role in cycles
> @MartinNowak Just filed an enhancement on this, I figure might be an easy improvement instead of doing full-blown dependency detection: https://issues.dlang.org/show_bug.cgi?id=16265

We should definitely do this and maybe a few more things, before tightening the cyclic dependency detector. I'm currently working on fixing 16211 based on the existing code. It just requires a few adjustments to allow cycles between modules w/o ctors, but it's also a good opportunity to clean up the code a bit.

schveiguy:
When revamping the cyclic dependency detector (with full analysis), let's collaborate on a design. There are still some holes in cycle detection, especially when it comes to templates. I should file some more bugs :)

MartinNowak:
> When revamping the cyclic dependency detector (with full analysis), let's collaborate on a design. There are still some holes in cycle detection, especially when it comes to templates. I should file some more bugs :)

What design? This just looks like an algorithmic problem to me. We get a graph (now with a restricted kind of cycles) and need to sort it in a way that modules with ctor/dtor can't be used before/after construction/destruction.

schveiguy:
What I mean by design is how you are going to create/store/read the graph. We can go down to the level of marking which functions each static ctor depends on, and then tracing the functions to see whether they depend on any variables that are initialized by static ctors. But I think that means our graph will be orders of magnitude bigger, require too many resources at startup. So there has to be some compromise somewhere. I think we will probably need language change for this.

All I really care about is that you ping me when you make your new patch (both for the runtime fix you are working on and the design of how the new dependency graph works), so I don't miss it.

BTW, this is now passing on Windows (auto tester finally got around to it), so it can be merged. If you can't figure out a better way, please consider merging.

src/rt/minfo.d
MartinNowak: Yes this will work. Modules are only initialized/sorted after rt_init, i.e. shared libraries aren't constructed in dso_registry until the runtime is initialized.

src/rt/minfo.d
```diff
@@ -60,6 +60,339 @@ struct ModuleGroup
...
+        // dependencies that also have static ctor/dtors.
+        void _checkModCtors2(int curidx)
+        {
+            assert(curidx != -1);
+            immutable ModuleInfo* current = _modules[curidx];
+
+            // we only get called if current has a dtor or a ctor, so no need to
+            // check that.  First, determine what non-trivial elements are
+            // reachable.
+            reachable[] = 0;
```
MartinNowak: Can use a bit vector for reachability.

src/rt/minfo.d
```diff
@@ -60,6 +60,339 @@ struct ModuleGroup
...
+            // allocate the dependencies on the stack
+            auto p = cast(int *)alloca(nmodules * int.sizeof);
+            auto dependencies = p[0 .. nmodules];
+            uint depidx = 0;
+            // fill in the dependencies
+            foreach (int i, r; reachable)
+            {
+                if(r && i != curidx)
+                {
+                    if((flags[i] & (MIctor | MIdtor)) && !(flags[i] & MIstandalone))
```
MartinNowak: This could already be filtered out during reachability computation, no need for a second array.
We could just use `auto reachable = findDependencies(current)` and let it return a bit vector (Array!ubyte or so).

src/rt/minfo.d
```diff
@@ -60,6 +60,339 @@ struct ModuleGroup
...
+            reachable[idx] = 1;
+            if(!orig && (flags[idx] & (MIctor | MIdtor)) && !(flags[idx] & MIstandalone))
+                // non-trivial, stop here
+                return result + 1;
+            foreach (m; current.importedModules)
+            {
+                auto midx = findModule(m);
+                if(midx != -1)
+                    // not part of this DSO, don't consider it.
+                    result += _findDependencies(midx, false);
```
MartinNowak: Should check reachable before recursing.

MartinNowak: I'd recommend `if (!bts(reachable, mids)) recurse`.

src/rt/minfo.d
```diff
@@ -60,6 +60,339 @@ struct ModuleGroup
...
+            // add this module to the construction order list
+            ctors[ctoridx++] = current;
+        }
+
+        void _checkModCtors3()
+        {
+            foreach (int idx, m; _modules)
+            {
+                // TODO: Should null ModuleInfo be allowed?
+                if (m is null) continue;
```
MartinNowak: We no longer have nulls in the module lists.

src/rt/minfo.d
MartinNowak: Or just during iteration.

src/rt/minfo.d
MartinNowak: Might make sense to replace the recursion with a loop and an allocated stack to save on stack usage.

src/rt/minfo.d
klickverbot: LDC still does, IIRC.

src/rt/minfo.d
schveiguy: > Should check reachable before recursing.

yes, I can do this. And you are right, I can switch from flags array to bit arrays like you had before.

> Might make sense to replace the recursion with a loop and an allocated stack to save on stack usage.

I can do this. It will save runtime to avoid calls as well. Largest stack we need in this case is the size of the module array.

src/rt/minfo.d
schveiguy: OK

src/rt/minfo.d
schveiguy: The reason I did it this way is to allocate the array of dependencies on the stack using `alloca`. To do that, I needed to count all the dependencies all at once. However, I can cache this test before even running anything and store the result in a bit array.

One other thing I can do is cut short the search for reachable nodes if I detect a cycle during that time (i.e. ctorstart is set).

MartinNowak:
After all the nitpicks, there is one bigger optimization the old algorithm misses. When all modules w/ ctor reachable from a module w/o ctor are constructed (marked ctordone), we can also mark the module w/o ctor as ctordone and no longer need to look at it's imports. This will help to avoid the O(N^2) behavior.

schveiguy:
Ok. I think I know how to do this. Nice idea

MartinNowak:
> Ok. I think I know how to do this. Nice idea

That's the main reason why I rewrote the implementation back then, but of course I overlooked the case leading to Issue 16211.
It might be possible to squeeze this into my version of the algorithm, but it gets a bit messy. If you can find a good way to do it with your version, go on. I'm a bit challenged to invest much time into this atm.

schveiguy:
@MartinNowak I fixed pretty much everything discussed, please take a look (starting with commit called "More improvements"). Also rebased.

How I handle pruning non-relevant nodes is to mark them as started when I first encounter them, then mark them as done after that module's search is finished. From that point on, all relevant nodes reached via that node have been checked, so we can mark it done (and not look at it again).

I think this is the only possible way to do that. Unfortunately, this means that they aren't pruned if they are detected very early.

WalterBright:
> For example:  a -> b -> c -> b -> a, where only a and c have module ctors, but b does not. the old algorithm would short circuit the evaluation at b the second time, thinking it was already visited.

Hmm, I don't think the original did that, as `b` would look at both `c` and `a` the first time `b` was encountered.

schveiguy:
> I don't think the original did that

The original, the one I fixed, did not detect the cycle. Then in 2.058, the bad behavior returned. Try compiling the cycle-containing code in the bug report (the old one: https://issues.dlang.org/show_bug.cgi?id=4384) with any compiler from 2.058 to 2.071, you will find it does not detect the cycle. Depending on the order of compilation, you get different results.

schveiguy:
Hm... I could handle non-relevant nodes at the deepest level instead of the shallowest. This should make the behavior much more performant. It means allocating dependencies for them at each level of recursion. I wonder if it's better to just duplicate the reachable array instead of allocating an array of indexes. That wouldn't affect the complexity, but uses much less space.

I'll change this, give me a few minutes.

schveiguy:
OK, now I think it's pretty well optimized.

MartinNowak:
Will review in the next few days. We'll prolly have to release this with just a warning in the beginning, and/or should also implement a `--DRT-ignoreCycles` switch (in another PR though).

MartinNowak:
Just a tip, if you add `Issue 16211` to the Trello cards, dlang-bot will take care of cross-linking Trello with GH and Bugzilla. It will also move the card once a PR gets opened/merged.

schveiguy:
ping @MartinNowak any time to look at this? I'm worried that I will have to relearn this again if we don't fix it soon, it's such an obscure algorithm 😄

MartinNowak:
Yes, I'll look at it tomorrow, in particularly testing several libraries and projects.

MartinNowak:
Looks technically correct and is fairly easy to grasp now.
Only implementation review point, let's please use malloc/free for the recursive allocation of the reachable bitvector, at least if it's a bigger allocation. Since we have shared libraries it's possible that this code runs on a Fiber stack. Same for the "stack" which is 2KB with 512 modules.
Might want to time, e.g. sorting of all modules in the shared Phobos library, but I doubt that malloc is a lot slower than alloca for this use case.

Style review:
- Please use dfmt or at least at the missing space between if/for/foreach/while and the paren
- Can we move the path finding out of this function? It's a lot to read over to get to the gist.

src/rt/minfo.d
```diff
@@ -62,174 +62,382 @@ struct ModuleGroup
...
+            return ctors[0 .. ctoridx];
         }
 
-        /* Do two passes: ctor/dtor, tlsctor/tlsdtor
-         */
-        sort(_ctors, MIctor | MIdtor);
-        sort(_tlsctors, MItlsctor | MItlsdtor);
+        // finally, do the sorting for both shared and tls ctors.
+        _ctors = doSort((immutable ModuleInfo * m) => (m.flags & MIstandalone) | ((m.dtor || m.ctor)? MIctor : 0));
+        _tlsctors = doSort((immutable ModuleInfo * m) => (m.flags & MIstandalone) | ((m.tlsdtor || m.tlsctor)? MIctor : 0));
```
MartinNowak: Just passing the mask as done on the left (side-by-side diff) is simpler than this callback.

src/rt/minfo.d
```diff
@@ -62,174 +62,382 @@ struct ModuleGroup
...
+            clearFlags(ctorstart);
+            clearFlags(ctordone);
+
+            // pre-allocate enough space to hold all modules.
+            ctors = (cast(immutable(ModuleInfo)**).malloc(len * (void*).sizeof))[0 .. len];
+            ctoridx = 0;
+            foreach (int idx, m; _modules)
+            {
+                // TODO: Should null ModuleInfo be allowed?
+                if (m is null) continue;
```
MartinNowak: We no longer have null modules, please remove.

src/rt/minfo.d
```diff
@@ -62,174 +62,382 @@ struct ModuleGroup
...
+            if(m > maxMod)
+                maxMod = m;
+        }
 
         int findModule(in ModuleInfo* mi)
         {
+            // short circuit linear search if possible.
+            if(mi < minMod || mi > maxMod)
+                return -1;
+            // binary search would be nice...
```
MartinNowak: Could use a hash table if you wanted to, there is one in rt.util.containers.

src/rt/minfo.d
```diff
@@ -62,174 +62,382 @@ struct ModuleGroup
...
+                    processMod(i);
+                }
+            }
+
+            // now mark this node, and all nodes reachable from this module as done.
+            bts(ctordone, curidx);
+            btr(ctorstart, curidx);
+            foreach (int i; 0 .. len)
+            {
+                if(bt(reachable, i))
```
MartinNowak: We have a similar problem w/ the current precise GC scan loop, iterating and testing every bit should be improvable using sth. like bsf/bsr.

src/rt/minfo.d
klickverbot: LDC IIRC still has them.

src/rt/minfo.d
klickverbot: Is it really worth it, though? Seems like we'd rather fix the algorithm complexity first.

MartinNowak: There isn't much to fix, b/c we have a graph with cycles between modules w/o ctor, a cheap O(N) sorting for DAGs doesn't work. We do a lot of caching so the normal case should be fairly linear, but the worst case has O(N) recursions, each of which reaching all the remaining modules through importing the next, so it's O(N^2) for the reachability computation.
Sorting Phobos took some 10s of microseconds last time I measured.

src/rt/minfo.d
schveiguy: I was working based on my original code. I didn't realize there were flags that were now set by the compiler.

src/rt/minfo.d
schveiguy: In essence, what we want is a DAG between modules that matter. The "edges", however, can involve any number of non-relevant modules, and contain cycles of their own. So you need one step to establish the edges (which is O(N^2)), and one step that sorts the DAG in O(N) (or finds a cycle). The trick is to avoid creating a large O(N^2) data structure to store the edges, at least avoid creating it unless necessary. Applying heuristics at this point is basically all we have left.

src/rt/minfo.d
schveiguy: Well, looking back at the current released code, null modules weren't checked for, and would definitely have crashed the runtime. Are you sure?

klickverbot: Ah, yes, the `null` module "padding" we unfortunately end up with is now skipped in `sections_elf_shared` before constructing the `ModuleGroup`. Sorry for the noise.

schveiguy:
@MartinNowak Addressed all your points. I removed all `alloca` in favor of `malloc`/`free`. I added `HashTab` to search for module info indexes. Even though I think it reads poorer, I removed the cycle generation code and shortest path into another private method.

Note: I added a basic `BitRange` type to `core.bitop`, to address iterating through all bits to see which are set. Maybe you can use this in the GC. I don't love that I have to keep track of so much in the range (I'm storing 4 words), improvement suggestions welcome.

schveiguy:
Ugh... segfaults. And on a test I'm not normally running. I'll see if I can figure it out.

schveiguy:
Hm... I think it's because of the realloc. On Linux, there are times where the module ctors/dtors length is 0 (Probably because of shared libraries). I'll push a fix momentarily. Just trying to verify on Linux that this works.

src/rt/minfo.d
schveiguy: @klickverbot thanks, removed the check based on that.

schveiguy:
Passing now.

schveiguy:
Note: I'll squash once this is approved.

src/rt/minfo.d
```diff
@@ -393,9 +393,6 @@ struct ModuleGroup
             ctoridx = 0;
             foreach (int idx, m; _modules)
             {
-                // TODO: Should null ModuleInfo be allowed?
-                if (m is null)
-                    continue;
```
MartinNowak: Wait, as @klickverbot said, it might still happen with LDC. Maybe just change to comment to mention LDC.

src/core/bitop.d
```diff
@@ -356,6 +356,143 @@ int bts(size_t* p, size_t bitnum) pure @system;
 }
 
 /**
+ * Range over bit set. Each element is the bit number that is set.
+ *
+ * This is more efficient than testing each bit in a sparsely populated bit
+ * set. Note that the first bit in the bit set would be bit 0.
+ */
+struct BitRange
```
MartinNowak: Nice

src/rt/minfo.d
schveiguy: I'm pretty sure if ldc passed null modules here, programs would deterministically crash with the current implementation. I don't think we need to worry about it

klickverbot: I looked it up a couple of days ago, and your sections_elf_shared code now actually trims null modules at the beginning and end of the list. In other words, we shouldn't get any null modules here for LDC either.

schveiguy:
Ugh. Another CI tester? How do I get it to pass?

wilzbach:
> Ugh. Another CI tester?

Yes we transition(ed) from Travis to CircleCi for all the non-critical stuff (style checks, code coverage, ...) as dmd and its PRs use all the five concurrent builds ;-)

> How do I get it to pass?

Just rebase to master

schveiguy:
Rebased

src/rt/minfo.d
```diff
@@ -53,6 +53,128 @@ struct ModuleGroup
...
+                if (impidx != -1 && impidx != i)
+                    if (!bts(reachable, impidx))
+                        edges[i] ~= impidx;
+            }
+        }
+
+        // determine the shortest path between two modules. Uses dijkstra
+        // without a priority queue. (we can be a bit slow here, in order to
+        // get a better printout).
+        void shortest(int start, int target)
```
ibuclaw: Could this be a private function rather than a nested closure function?

schveiguy: Yes. But much easier to share all the state this way

schveiguy: Btw not a closure.

ibuclaw: I didn't check hard enough to see if there are nested references. If there are none probably won't make a dent in the cost either way.

schveiguy: All the edges, the cycle array that records the path, and the distance array are referenced from the enclosing function. But the distance array is only used locally, so it could potentially be allocated for each run.

In any case, I'm not sure it's too important. Nothing else will use this function, as it's very specific to the cycle detection.

MartinNowak:
OK, didn't found any new cycles in a bunch of projects I tested.

MartinNowak:
Auto-merge toggled on

schveiguy:
Awesome, thanks!

schveiguy:
> Only takes 1ms to sort phobos, even twice as fast as the current implementation.

that's nice to have tested.

schveiguy:
> Could we easily identify the cases that are handled differently w/ this fix, instead of just making all cycles non-fatal?

The cases where this behaves differently because of this fix are not the proper distinction here. In other words, the fact that the cycle is harmless has nothing to do with the fact that the cycle is now detected where it wasn't before.

In fact, I'm surprised the above cycle wasn't found by older algorithm -- there's no repeated nodes in the cycle!

IMO, the original cycle detection was hopelessly broken, and compiling things in different orders, or importing sub-modules instead of packages, changed whether the cycle is truly detected.

I'm fine for now with just printing that there is a cycle and not causing a fatal crash. Having a messy printout is an annoying enough thing that people will still try and fix the problem, but it doesn't prevent bad issues from happening. I know from experience that working around cycle issues can be very tricky.

One thing I see happening is when ordering *is* important for initialization, but the linker order happens to make it correct. Then, at some later time, the ordering is jumbled and all of a sudden things aren't initialized in the right order. There is a fundamental reason why the cycle detection must happen.

We should figure out if there is a better way to do this, including allowing tagging of static ctors to override order detection.

schveiguy:
Also, while you are paying attention to this, can you take a peek at dlang/dmd#6076

schveiguy:
If we just print the cycle, how do we decide the order of the static ctors? In other words, what static ctor gets run first in the cycle above, `ddmd.traits` or `ddmd.cond`?

MartinNowak:
> I'm fine for now with just printing that there is a cycle and not causing a fatal crash.

I think we need to figure this out anyhow, an option to disable this being fatal was requested a couple of times (and can be supported via the -DRT- options, e.g. `-DRT-nocyclecheck`).
Adding such an option might be enough for the transition phase (not sure).

> If we just print the cycle, how do we decide the order of the static ctors? In other words, what static ctor gets run first in the cycle above, ddmd.traits or ddmd.cond?

That's a tricky question and I haven't yet thought long enough about this to give an answer. My intuition says we should run ddmd.traits first b/c it's at the bottom of the depedency tree (before the cycle ascends back up). Though a much simpler resolution would be to mark the end of the cycle `ddmd.cond` as resolved and continue where we left of.
OTOH we can't guarantee any working order and know nothing about the importance, so any order is a good as any other, and using the simplest implementation would be a reasonable solution, i.e. print the cycle, ignore the module at the end (`ddmd.cond`) and go on (replace throw w/ printf).

schveiguy:
> My intuition says we should run ddmd.traits first b/c it's at the bottom of the depedency tree

That's only because the module order happened to put ddmd.cond first in the list of modules :)

If it was the other way around, the cycle would be traits -> cond -> traits. I wonder if it would be worth sorting the module names first to create a deterministic ordering, even in the event of cycles. We are already putting the names into a hash table, instead if we just put it in a sorted list, and then use binary search to take the place of the hashing, we could get similar performance and more determinism.

The fundamental problem here is that the cycle detection means there is a real determinism problem, where there is no right answer. All we can do is run the constructors in whichever order we pick (I think from an  algorithm standpoint, it makes sense to print the cycle and prune that branch as normal, so in the cycle above ddmd.traits would run first), and hope that it will not create issues. We for sure need a switch to enable this behavior.

MartinNowak:
> I wonder if it would be worth sorting the module names first to create a deterministic ordering, even in the event of cycles. 

Not to keen on presorting the modules, its add even more complexity, startup time, and code into every binary and we already print the cycle, so people know that sth. doesn't work as expected.
Can we recommend a simple workaround for people w/ harmless cycles?

MartinNowak:
Don't want to push, but we should include such a switch in 2.072.0.

schveiguy:
> Don't want to push, but we should include such a switch in 2.072.0.

I will try to get to it by tomorrow.

schveiguy:
> I will try to get to it by tomorrow.

Obviously didn't get there, I've been crazy busy. I am traveling today, and will not get to this until at least Saturday, and if not, it will be next week. If you start releasing betas, I'll put it into stable branch.

MartinNowak:
Let's use `-DRT-cyclecheck=printonly`, `-DRT-cyclecheck=no`, and `-DRT-cyclecheck=yes` (default, but also explicit to override other cyclecheck settings).

schveiguy:
I had settled on this scheme:
```
-DRT-onCycle=abort -> print cycle and abort (current behavior)
-DRT-onCycle=print -> print cycle but make best effort to sort ctors
-DRT-onCycle=ignore -> ignore cycles, don't print anything
```
I can switch easily to your scheme. Almost done with the PR.

schveiguy:
@MartinNowak having issues building stable branch of dmd:

```
CC=clang++ dmd -ofdmd -m64 -vtls -J. -L-lstdc++ -version=MARS -wi access.d aggregate.d aliasthis.d apply.d argtypes.d arrayop.d arraytypes.d attrib.d builtin.d canthrow.d clone.d complex.d cond.d constfold.d cppmangle.d ctfeexpr.d dcast.d dclass.d declaration.d delegatize.d denum.d dimport.d dinifile.d dinterpret.d dmacro.d dmangle.d dmodule.d doc.d dscope.d dstruct.d dsymbol.d dtemplate.d dversion.d entity.d errors.d escape.d expression.d func.d globals.d hdrgen.d id.d identifier.d impcnvtab.d imphint.d init.d inline.d intrange.d json.d lexer.d lib.d link.d mars.d mtype.d nogc.d nspace.d opover.d optimize.d parse.d sapply.d sideeffect.d statement.d staticassert.d target.d tokens.d traits.d utf.d visitor.d typinf.d utils.d statementsem.d safe.d objc.d libmach.d scanmach.d irstate.d toelfdebug.d toctype.d glue.d gluelayer.d todt.d tocsym.d toir.d dmsc.d backend/bcomplex.d backend/cc.d backend/cdef.d backend/cgcv.d backend/code.d backend/dt.d backend/el.d backend/global.d backend/obj.d backend/oper.d backend/outbuf.d backend/rtlsym.d backend/ty.d backend/type.d tk/dlist.d root/aav.d root/array.d root/ctfloat.d root/file.d root/filename.d root/man.d root/outbuffer.d root/port.d root/response.d root/rmem.d root/rootobject.d root/speller.d root/stringtable.d newdelete.o glue.a backend.a
Undefined symbols for architecture x86_64:
  "symboldata(unsigned long long, unsigned int)", referenced from:
      el_ptr(Symbol*) in backend.a(el.o)
      el_convstring(elem*) in backend.a(el.o)
      out_readonly_sym(unsigned int, void*, int) in backend.a(out.o)
      Obj::sym_cdata(unsigned int, char*, int) in backend.a(machobj.o)
  "_align(unsigned long long, unsigned long long)", referenced from:
      codgen() in backend.a(cgcod.o)
      stackoffsets(int) in backend.a(cgcod.o)
      outjmptab(block*) in backend.a(cod3.o)
      outswitab(block*) in backend.a(cod3.o)
      type_paramsize(TYPE*) in backend.a(type.o)
      alignOffset(int, unsigned long long) in backend.a(out.o)
      cdfunc(elem*, unsigned int*) in backend.a(cod1.o)
      ...
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
--- errorlevel 1
make[1]: *** [dmd] Error 1
make: *** [all] Error 2
```
Any ideas? I've tried several versions of dmd, including 2.069 and 2.071

MartinNowak:
> I can switch easily to your scheme. Almost done with the PR.

Looks great, let's not start with camel cases for command lines and stick to `-DRT-oncycle=`.

schveiguy:
@MartinNowak I created the PR: https://github.com/dlang/druntime/pull/1668. I didn't see your message about downloading dmd stable. I'll try that.

MartinNowak:
Unfortunately this change did broke more projects than initially tested, at least Higgs, vibe.d/libasync, and an intermediate state of ddmd. I'd expect that it hit a lot more projects not tested by us.
We had the idea of temporarily adding back the replaced implementation as a fallback, for when the new code reports an error, so we can properly deprecate this cycle detection change.
Flags were a nice supplemental addition (already requested a couple of times), but don't really replace a soft-fault deprecate-then-error transition.

schveiguy:
I may be able to "run" the old algorithm, or at least adapt the new algorithm to run like the old one did, if errors are seen. Alternatively, we can resurrect the entire original code, and just call it if the corrected algorithm fails.

It's an unfortunate perception that this is "breaking" projects -- those projects are already broken in terms of having a cycle. Whether the cycle is harmless or not, the truth is that the compiler doesn't give us enough information to determine whether it's truly broken or not.

CyberShadow:
> It's an unfortunate perception that this is "breaking" projects -- those projects are already broken in terms of having a cycle.

No, that's a fallacy. You could say "broken" about a lot of things that are written in ways that under some circumstances can end up being buggy, but that's just not a useful definition. Relying on that makes for an unusable language because each version makes most existing code unusable as-is. If a piece of software has been tested, known to work correctly for all foreseeable inputs and circumstances, and has been successfully put in production, it is not broken, it's working as expected. A compiler update that causes correct software to stop working *is* broken.

I don't want to start a debate, but this is an argument that has been brought up a few times. D needs to be *useful* foremost, facilitating creation of correct programs is just one side of that.

schveiguy:
@CyberShadow There is no need to debate, we definitely can go through deprecation state.

D claims to handle the proper initialization of globals via dependency sorting at runtime. The sorting was broken. The answer to running the program *should* have been (and is now), "I don't know how to properly intialize things". Essentially, we can embrace undefined behavior or we can reject it. Which is more useful to a user? In my experience a broken compiler that spits out UB is far worse than one that doesn't compile (or in this case run) my broken code.

The *most* unfortunate thing in this whole mess is the fact that cycles are left for runtime with minimal information (the import dependencies) to go on. It's really a problem that should be solved by the compiler/linker.

CyberShadow:
> Which is more useful to a user? In my experience a broken compiler that spits out UB is far worse than one that doesn't compile (or in this case run) my broken code.

Yes - but only for **new programs**. Making working software stop working is the problem. This is why we can't reinvent things every year and are likely stuck with the http://wiki.dlang.org/Language_issues list forever.

schveiguy:
We can agree to disagree. I personally don't want my programs blowing up at some future date due to a forseeable initialization ambiguity. Others might be OK with that I guess.

This is not a reinvention. I fixed this problem in 2010, and it resurfaced within a year. I should have put a test in for it at the time, we are much better about that now.

In any case, we can stop the pointless philosophical discussion and just put in the deprecation mechanism.

CyberShadow:
> We can agree to disagree.

Just to clarify, this is not just about a debate of personal preferences. It became clear a long time ago that the point where we can break working code willy-nilly in D is long past. As such, this is not negotiable: we all must take code breakage seriously.

A method of notifying the user without actually breaking their code is the correct solution here. For this we usually have deprecation paths, or in this particular case, a message printed on startup is good too.

schveiguy:
Again, not willy-nilly. The compiler was generating code that creates undefined behavior. I think we can all agree that any fix to any broken code of this nature has the chance to "break" code that depended on the brokenness that's being fixed. It's a question of where we draw the line. This is why it's not a winnable argument either way. Either you have code that crashes the mars lander, or you have fixes that set the project back months. Which is worse depends on what your job is.

The issues surrounding this fix are very woolly, and we have a path forward, let's just implement it.

ibuclaw:
> The most unfortunate thing in this whole mess is the fact that cycles are left for runtime with minimal information (the import dependencies) to go on. It's really a problem that should be solved by the compiler/linker.

@schveiguy - what information would be useful to you?

schveiguy:
Well, an actual dependency map for the ctors. Right now you have things that are listed as dependencies simply because they are imported somewhere in the module. But the compiler knows which variables depend on which other modules because it can see what functions are called in the ctor.

schveiguy:
With that, the cycle detection would be closer to reality. Most cycles are harmless.

ibuclaw:
> Right now you have things that are listed as dependencies simply because they are imported somewhere in the module. But the compiler knows which variables depend on which other modules because it can see what functions are called in the ctor.

Well, it is just an array of imported modules, which may not be the best indicator of dependencies between them.  I can't think of any technical blockers that would prevent adding new information about which imported modules are used in the construction/destruction phases, it could even be an int that just masks which indexes in the importes modules array are used by the ctor.

MartinNowak:
All the ideas and thoughts on how to improve this are fairly spread over multiple Bugzilla tickets, forum entries, and github comments. Can we please aggregate all of this in a single Bugzilla ticket (or didn't we already have one).
Even the Trello card isn't a good starting place (https://trello.com/c/V8WAxAlD/61-redesign-of-cycle-detection-issue-16211).

One related idea was that ctors only using private variables can be marked as `MIstandalone`, that would work for just initializing `immutable` variables.

MartinNowak:
See #1717

PR #1605, opened 2016-07-03
fix issue 16230 - core.atomic.atomicLoad removes shared from aggregat…

aG0aep6G:
…e types too eagerly

https://issues.dlang.org/show_bug.cgi?id=16230

This breaks two places in phobos. Pull request follows.


aG0aep6G:
> This breaks two places in phobos. Pull request follows.

https://github.com/dlang/phobos/pull/4551

The phobos changes need to be pulled first. But before that, the changes here should be reviewed.

ZombineDev:
LGTM

MartinNowak:
Looks goods, could we do it w/o the breakage, e.g. by instantiating a deprecated HeadUnsharedDeprecated for the old conversion?
IIRC there was a bug that kept dmd from emitting deprecation warnings for template instantiations.

src/core/atomic.d
```diff
@@ -35,9 +35,13 @@ private
...
         static if( is( T U : shared(U*) ) )
-            alias shared(U)* HeadUnshared;
+            alias HeadUnshared = shared(U)*;
+        else static if( is( T U : shared(U[]) ) )
+            alias HeadUnshared = shared(U)[];
+        else static if (is (shared T : T))
+            alias HeadUnshared = T;
         else
-            alias T HeadUnshared;
+            alias HeadUnshared = shared T;
```
MartinNowak: This didn't remove shared before your PR, so the bug description is confusing.

aG0aep6G: > HeadUnshared didn't remove shared from structs before your PR, so the bug description is confusing.

The problem is that HeadUnshared is never instantiated with fully shared types, only ever with tail-shared ones or completely unshared ones. That happens because atomicLoad itself strips shared already, and only then does it apply HeadUnshared. That means, the instantiations of HeadUnshared have been nops.

Now, shared needs to be re-added at some point to fix the issue with structs. I hadn't really thought this all through before, so my solution was a bit hacky.

I've now renamed HeadUnshared to TailShared and made it so that it constructs a type with a shared tail, ignoring T's given shared-ness. If T has no tail (no indirections), an unshared T is returned. If T's tail cannot be typed independently (aggregates with indirections), a shared T is returned.

Hope it makes more sense now. Tell me if this version is better. I'd squash the commits then.

aG0aep6G:
> Looks goods, could we do it w/o the breakage, e.g. by instantiating a deprecated HeadUnsharedDeprecated for the old conversion?

I don't know how a deprecated atomicLoad could be distinguished from the fixed one. The parameters aren't changing, only the return type does.

MartinNowak:
> I don't know how a deprecated atomicLoad could be distinguished from the fixed one. The parameters aren't changing, only the return type does.

Well, I though we can attach a deprecation message to template instantiations of TailShared that will change their behavior.
Your change has a simple consequence, it returns shared class references instead of plain class references. So you can even catch this with a deprecated atomicLoad specialization.
In any case we can't merge this with the breakage.

src/core/atomic.d
```diff
@@ -76,10 +126,10 @@ version( CoreDdoc )
      * Returns:
      *  The result of the operation.
      */
-    HeadUnshared!(T) atomicOp(string op, T, V1)( ref shared T val, V1 mod ) pure nothrow @nogc @safe
+    TailShared!T atomicOp(string op, T, V1)( ref shared T val, V1 mod ) pure nothrow @nogc @safe
```
MartinNowak: Can't do ops on classes, so this is more permissive than before.
Can you recheck that `TailShared!T` is always implicitly convertible to `shared T`?

src/core/atomic.d
```diff
@@ -119,9 +169,9 @@ version( CoreDdoc )
      * Returns:
      *  The value of 'val'.
      */
-    HeadUnshared!(T) atomicLoad(MemoryOrder ms = MemoryOrder.seq,T)( ref const shared T val ) pure nothrow @nogc @safe
+    TailShared!T atomicLoad(MemoryOrder ms = MemoryOrder.seq,T)( ref const shared T val ) pure nothrow @nogc @safe
```
MartinNowak: This should get a deprecated overload with `if (is(T == class))`.

src/core/atomic.d
aG0aep6G: > Can't do ops on classes, so this is more permissive than before.

Sorry, I don't follow. What am I allowing that wasn't allowed before?

> Can you recheck that TailShared!T is always implicitly convertible to shared T?

I can put `static assert(is(TailShared : shared T));` at the end of TailShared. Is that what you have in mind?

By the way, are the implicit conversion rules on shared sound? D lets me copy large types (that can't be loaded/stored atomically) to and from shared. Shouldn't that be disallowed?

src/core/atomic.d
aG0aep6G: > This should get a deprecated overload with if (is(T == class)).

But calling atomicLoad on a class isn't being deprecated.

aG0aep6G:
> Well, I though we can attach a deprecation message to template instantiations of TailShared that will change their behavior.

A deprecation means that the thing should not be used, and that it's going to get removed. But the changed parts should be good to use, and their removal is not planned.

If a deprecation stage is a must, then I think we have to add a new function that replaces atomicLoad (and probably the same for atomicOp). Then atomicLoad and atomicOp can be meaningfully deprecated.

src/core/atomic.d
MartinNowak: > > Can't do ops on classes, so this is more permissive than before.

> Sorry, I don't follow. What am I allowing that wasn't allowed before?

This function `atomicOp` has nothing to do w/ classes, `TailShared!T` is less restrictive than `HeadUnshared!T` (IIRC), so that function doesn't need any deprecation.

src/core/atomic.d
MartinNowak: Because `TailShared!T` is more restrictive than `HeadUnshared!T` for classes, we should keep an overload with `HeadUnshared!T` for classes and deprecate it.
Not yet sure how to implement the deprecation though.

MartinNowak:
> A deprecation means that the thing should not be used, and that it's going to get removed. But the changed parts should be good to use, and their removal is not planned.

Yes, but it also means that we're going the break code, so `deprecated("Cast away shared if necessary")` HeadUnshared!T atomicLoad() if (is(T == class))` does make sense.

src/core/atomic.d
aG0aep6G: > This function atomicOp can't be used w/ classes, TailShared!T is less restrictive than HeadUnshared!T (IIRC), so that function doesn't need any deprecation.

Ok, so atomicOp cannot possibly work correctly with classes, right?

The current code, using HeadUnshared, fails on the cas call. Looks like a lucky coincidence that the atomicOp call is rejected.

With TailShared it goes through if and only if the class has both a shared and an unshared opOpCall, because the template constraint checks the unshared case, and the actual operation is done on the result of TailShared, which is shared for classes.

The same thing happens with structs that have both variants of opOpAssign.

So, it's not just classes. atomicOp can't work with types that have an inseparable head and tail. Can't call an unshared opOpAssign, because it would assume that the referenced data is unshared. Shouldn't call a shared opOpAssign, because atomicOp would just add overhead then, and the shared opOpAssign may be implemented with locks instead of atomic operations.

So, add a constraint `!is(TailShared!T == shared)`? Does that make sense?

aG0aep6G:
> Yes, but it also means that we're going the break code, so deprecated("Cast away shared if necessary") HeadUnshared!T atomicLoad() if (is(T == class))` does make sense.

But then I have no way to call the fixed version. So I can't fix my code properly until the old version is actually removed and the new one takes over. And until then I have to live with the deprecation message.

To me that would be more frustrating than just getting an error.

aG0aep6G:
@MartinNowak (or anyone else), it's been about a year. Can we find a way forward for this? This is a safety issue that should be resolved.

I'd love to mark the old behavior as `deprecated`, but I don't see how.

The only way I see is to add the new behavior under a new name, and deprecate the old name. Eventually, when the old stuff has been removed, the new stuff can get the old name. If I should go for this, please say so.

Otherwise, I think I need it spelled out in detail how you'd do this.

aG0aep6G:
> @MartinNowak (or anyone else), it's been about a year.

Another week later. Anyone?

ZombineDev:
@aG0aep6G I'll take a look shortly.

src/core/atomic.d
```diff
@@ -32,12 +32,62 @@ else
...
+    /* Construct a type with a shared tail, and if possible with an unshared
+    head. */
+    template TailShared(T)
     {
-        static if( is( T U : shared(U*) ) )
-            alias shared(U)* HeadUnshared;
+        alias S = shared T;
+
+        static if (is(S U == shared U)) {}
+        else static assert(false);
```
MetaLang: Needs an error message

MetaLang:
@ZombineDev not sure if you've seen it but this has a corresponding Phobos pull that should be reviewed and pulled as well this one: https://github.com/dlang/phobos/pull/4551

src/core/atomic.d
aG0aep6G: > Needs an error message

Added a message: "S should be shared." But I'm not sure if it's any good. The assert shouldn't ever be hit, no matter what `T` is. So I'm not sure what expectation to express in the message.

MetaLang: I would specify that it should never be hit in the error message and the reason why (the latter of it looks like you've already done).

aG0aep6G: > I would specify that it should never be hit in the error message and the reason why

I've added some more text. I may have gone overboard. If it's too verbose now, I'd appreciate a concrete suggestion.

ZombineDev:
@aG0aep6G Sorry for taking so long to get back to you, I had limited time.

This PR looks good to me, essentially making this module work as I imagined it would before looking at the code. I have only one request - can you please add a changelog entry detailing the difference, as this is an important change IMO.

aG0aep6G:
> can you please add a changelog entry

done

aG0aep6G:
@CyberShadow, I don't understand the DAutoTest failure. Looks like it tries building phobos 2.075 with druntime master?

Regarding the Jenkins failure, I can't see what's wrong at all.

CyberShadow:
> Looks like it tries building phobos 2.075 with druntime master?

Yes, that seems exactly right.

Which is of course completely wrong.

https://github.com/dlang/dlang.org/pull/1852

aG0aep6G:
The Win64 failure seems to be caused by [issue 17665](https://issues.dlang.org/show_bug.cgi?id=17665) which is getting fixed by #1884. So this is blocked for the moment.

ZombineDev:
@andralex does this look good to you too?

aG0aep6G:
> @andralex does this look good to you too?

Ping. If this needs approval by @andralex, maybe assign him as a reviewer, or how do we get his attention?

src/core/atomic.d
```diff
@@ -32,12 +32,65 @@ else
...
-            alias shared(U)* HeadUnshared;
+        alias S = shared T;
+
+        static if (is(S U == shared U)) {}
+        else static assert(false, "Should never be triggered. The `static " ~
+            "if` declares `U` as the unshared version of the shared type " ~
+            "`S`. `S` is explicitly declared as shared, so getting `U` " ~
+            "should always work.");
+
+        static if (is(S : U))
```
andralex: Under what circumstances is U different from T?

changelog/atomicLoad-returns-shared.dd
```diff
@@ -0,0 +1,34 @@
...
+
+shared int tail = 1;
+auto shs = shared S(2, &tail);
+
+void main()
+{
+    import core.atomic : atomicLoad;
+    shared S s = atomicLoad(shs);
+
+    ++(cast(S) s).head; // Ok, reading/writing thread-local head.
```
aG0aep6G: > Did you test it?

Tested it now. Works for me: https://run.dlang.io/is/HXG743

`& cast(S) s` doesn't work, because "cast(S)s is not an lvalue". But the compiler accepts `&(cast(S) s).head` no problem, and it's the same pointer as `&s.head`. No idea, if that makes sense.

src/core/atomic.d
```diff
@@ -32,12 +32,65 @@ else
...
     {
-        static if( is( T U : shared(U*) ) )
-            alias shared(U)* HeadUnshared;
+        alias S = shared T;
+
+        static if (is(S U == shared U)) {}
+        else static assert(false, "Should never be triggered. The `static " ~
+            "if` declares `U` as the unshared version of the shared type " ~
+            "`S`. `S` is explicitly declared as shared, so getting `U` " ~
+            "should always work.");
```
aG0aep6G: > S is defined as shared T so this static if essentially extracts the T back under the name U. What is the purpose?

U is different from T when T is shared. T may be shared or not. S is always shared. U is always unshared.

src/core/atomic.d
aG0aep6G: Answered above.

src/core/atomic.d
andralex: I see, so it T is e.g. shared(int), then S is also shared(int) and U is int. This is worth placing in an explanatory comment.

changelog/atomicLoad-returns-shared.dd
andralex: This may as well be a bug in the compiler. The cast should create an rvalue in all circumstances. I'd be in your debt if you could submit that as a bug report. For this code to be robust, use: `++((cast(S*) &s).head)`

aG0aep6G: > For this code to be robust, use: ++((cast(S*) &s).head)

done

src/core/atomic.d
aG0aep6G: > This is worth placing in an explanatory comment.

Added a comment.

changelog/atomicLoad-returns-shared.dd
aG0aep6G: > The cast should create an rvalue in all circumstances. I'd be in your debt if you could submit that as a bug report.

https://issues.dlang.org/show_bug.cgi?id=17729

aG0aep6G:
> Cool. @MartinNowak ?

ping

ZombineDev:
It seems Martin is too busy at the moment. Jonathan, could you take a look?

MartinNowak:
Wrapping a struct is somewhat tricky for private fields.

ZombineDev:
> First of all, this still breaks a prominent use-case of core.atomic. [...]

I don't think we need to discuss why this code is broken. `shared` is not mature language feature, so I don't think we can guarantee it's stability in the short term (1-2 years), although I strongly agree that we shouldn't go cowboy style and trigger-happy follow the "break my code" advice from users. But in this case, I would rather help users see their potential mistakes rather than people seeing D as a language with no focus on safety and with ugly legacy baggage (functions that don't quite work correct but weren't fixed).  Today most of druntime/phobos doesn't work correctly with `shared`. If this PR is merged, we will have one less wrong thing to worry about, and one that is quite important.

> Second, I'm somewhat warily of the recent trend to enforce more usage of shared. While it's good to be correct, shared still isn't a full mature language feature.

I see this the other way around. For someone coming off from TDPL, `shared` promises a bright future, but in reality it under-delivers. People are not trying to use `shared` "just because". Instead we are trying to make `shared` more mature and useful. We can't present `shared` to people unless at least we have it working correctly with the standard library. `core.atomic` plays a fundamental role in building the library stack.

> Now this PR changes the basic primitive to interact with shared data from one brokenness to a different one, i.e. from overly loose (but possibly correct depending on the programmer) to overly restrictive (but requiring the programmer to add correct casts).

I see your point, but I don't agree with you. It's like saying:
"This proposal to the language C - disallowing implicit casts from `void*` to `T*` changes the basic primitive to interact with pointers from one brokenness to a different one, i.e. from overly loose (but possibly correct depending on the programmer) to overly restrictive (but requiring the programmer to add correct casts).

> Ultimately I'm worried a bit this will drive users to a cast() attitude, making shared basically useless.

Looking at the project tester it looks like this is already the case :( . People either don't use `shared` correctly, put casts everywhere, or ignore it completely by using `__gshared`. The only way forward is to make using `shared` practical (by making adding needed functionality to the library) and useful (by catching potential mistakes).

---

> How about we try a different correct approach, implement a HeadUnshared that actually wraps the return type (similar to Rebindable), and has @trusted @property ref getters, that returns the appropriate field types. Mostly structs returned from atomicLoad are just plain data fields.
Implicit conversion via alias this would still return a shared struct/class, to avoid any holes by calling unshared methods.

I also thought about writing a proper typecons for shared (in my case it was `Synchronized(T, Mutex = core.sync.mutex.Mutex)` meant to emulate the peeling of `shared` that `synchronized` blocks are meant to do), and in general I like the idea, but I don't know if it's doable in practice for `atomicLoad`.
* May break generic code that introspects the return type in unforeseeable ways
* Aggregate scoped imports can't be introspected (AFAIK), so we can't emulate them
* Not sure if it could be made to work with classes.
* What would these properties return? In the example `struct S { int head; int* tailPointer; }`, what would the `ref tailPointer() @property` return if you call `atomicLoad!(S*)` - `shared(int*)` or `shared(int)*`? If it would be  `shared(int)*`, then this property would need to implicitly call `atomicLoad`, but then how would you atomically set it? If the answer is `shared(int*)`, then there would be no point in having a wrapper. And returning `int*` is just wrong.

I'm not saying that I'm in principle against this idea. I just think that this need to be very carefully designed, it is out of scope for this PR and could easily be a breaking change to atomicLoad's return type. Probably sounds more like a job for something called `ImplicitlyAtomic(T)`.

MartinNowak:
> What would these properties return? In the example struct S { int head; int* tailPointer; }, what would the ref tailPointer() @property return if you call atomicLoad!(S*) - shared(int*) or shared(int)*? If it would be shared(int)*, then this property would need to implicitly call atomicLoad, but then how would you atomically set it? If the answer is shared(int*), then there would be no point in having a wrapper. And returning int* is just wrong.

```d
struct S { int head; int* tailPointer; }
atomicLoad!(shared(S*)) -> shared(S)*
atomicLoad!(shared(S)) -> HeadUnshared!S

struct HeadUnshared!S
{
    @property ref int head() { return impl.head; }
    @property ref shared(int*) tailPointer() { return impl.tailPointer; } // just use atomic* on the second layer yourself
private:
    shared S __impl;
}
```

Sketch
```d
struct HeadUnshared(T)
{
    static foreach (mem; __traits(getMembers, T))
    {
        alias MemT = typeof(__traits(getMember, T.init, mem));
        static if (hasIndirections!MemT)
            mixin("@property ref shared(MemT) " ~ mem ~ "() @safe { return __impl. " ~ mem ~ "; }");
        else
            mixin("@property ref MemT " ~ mem ~ "() @trusted { return *cast(MemT*)&__impl. " ~ mem ~ "; }");
    }
private:
    shared T __impl;
}
```

> I'm not saying that I'm in principle against this idea. I just think that this need to be very carefully designed, it is out of scope for this PR and could easily be a breaking change to atomicLoad's return type. Probably sounds more like a job for something called ImplicitlyAtomic(T).

Sure, but *this* PR is already a breaking change, so it's a good time to look at where we want to go.

aG0aep6G:
As far as I see, the wrapper idea can only work as intended with structs. With a class, atomicLoad only loads the class reference. All fields have to be treated as shared. Doesn't matter if they have indirections or not. I don't see `Wrapper!C` being an improvement over `shared(C)` then. All it achieves is removing the `shared` qualifier from the top level, at the cost of being a completely different type.

For structs, it has more appeal. Instead of generating properties, we could also generate a struct type that has the same fields, but applies `TailShared` to their types:

```d
template TailShared2000(T)
{
    static if (is(T == struct) && is(TailShared!T == shared))
    {
        struct TailShared2000
        {
            private import std.traits : hasIndirections, isAggregateType;
            
            static foreach (i; 0 .. T.tupleof.length)
            {
                mixin("TailShared!(typeof(T.tupleof[i])) " ~
                    __traits(identifier, T.tupleof[i]) ~ ";");
                //TODO: alignment
            }
            
            @property shared(T) _fullyShared()
            {
                return * cast(shared(T)*) &this;
            }
            alias _fullyShared this;
        }
        static assert(TailShared2000.sizeof == T.sizeof);
    }
    else alias TailShared2000 = TailShared!T;
}
unittest
{
    struct S { int head; int* tailPointer; }
    TailShared2000!S s;
    static assert(!is(typeof(s) == shared));
    static assert(is(typeof(s.head) == int));
    static assert(is(typeof(s.tailPointer) == shared(int)*));
    static assert(!is(typeof(s) : S));
    static assert(is(typeof(s) : shared S));
}
```

ZombineDev:
Keep in mind that atomic loading a whole struct is rather rare, since at most you can load 2 * size_t.sizeof, one most architectures other than x86, you can load only a single size_t.

MartinNowak:
> As far as I see, the wrapper idea can only work as intended with structs. With a class, atomicLoad only loads the class reference.

Sure, for classes it doesn't provide anything, the example above wasn't too useful.

> Keep in mind that atomic loading a whole struct is rather rare, since at most you can load 2 * size_t.sizeof.
> On most architectures other than x86, you can load only a single size_t.
> (for easy ref: https://github.com/tc39/ecmascript_sharedmem/blob/master/MACHINES.md)

Yes, the use-cases are mostly lock-free data structures and such. They are limited to rather simple data types, hence the wrapping approach might work.

> For structs, it has more appeal. Instead of generating properties, we could also generate a struct type that has the same fields, but applies TailShared to their types:

If that still supports implicit conversion to the shared original type (e.g. for method calls), that would be fine as well. Preserving the exact struct layout (alignment) is critical though.

aG0aep6G:
> Preserving the exact struct layout (alignment) is critical for this.

Yeah, and it's a good bit more complicated than generating getters. So I'm backing away from that idea.

@ZombineDev, @andralex: Are you on board with returning a wrapper for structs when we can't just strip `shared`?

This is what I have at the moment:

```d
template TailShared2001(T)
{
    static if (is(T == struct) && is(TailShared!T == shared))
    {
        struct TailShared2001
        {
            static foreach (i, alias field; T.tupleof)
            {
                mixin("
                    @property ref " ~ __traits(identifier, field) ~ "()
                    {
                        alias R = TailShared!(typeof(field));
                        return * cast(R*) &_impl.tupleof[i];
                    }
                ");
            }
            shared(T) _impl;
            alias _impl this;
        }
    }
    else alias TailShared2001 = TailShared!T;
}
unittest
{
    struct S { int head; int* tailPointer; }
    TailShared2001!S s;
    static assert(!is(typeof(s) == shared));
    static assert(is(typeof(s.head) == int));
    static assert(is(typeof(s.tailPointer) == shared(int)*));
    static assert(!is(typeof(s) : S));
    static assert(is(typeof(s) : shared S));
}
```

MartinNowak:
> `@property ref " ~ __traits(identifier, field) ~ "()`

Should be `@trusted`, otherwise looks good.

aG0aep6G:
I've pushed the wrapper.

As far as i can tell, private fields are not an issue, because `tupleof` ignores that.

One limitation of the wrapper is with code like this: `&atomicLoad(s).field`. Since `field` is a method, you now get a delegate instead of a pointer to the returned field. A struct with the same the same layout but `TailShared` fields wouldn't have that problem. Maybe I'll get back to that idea.

Unfortunately, this still works with the wrapper: `int* p = atomicLoad(s).tailPointer;`. But that's a compiler bug: [issue 17769](https://issues.dlang.org/show_bug.cgi?id=17769).

MartinNowak:
> Since field is a method, you now get a delegate instead of a pointer to the returned field.

That's what `@property` is for ;), see https://run.dlang.io/is/cNC6UA.

src/core/atomic.d
```diff
@@ -32,12 +32,117 @@ else
...
+                }
+                static bool canFind(string[] haystack, string needle)
+                {
+                    foreach (candidate; haystack)
+                    {
+                        if (candidate == needle) return true;
+                    }
+                    return false;
+                }
+                while (canFind(fieldNames, name)) name ~= "_";
```
MartinNowak: Just use `__` here, they are reserved for internal (compiler&std lib) usage. To use a suffix that won't clash w/ phobos, how about `__tailSharedImpl`.

changelog/atomicLoad-return-types.dd
```diff
@@ -0,0 +1,45 @@
...
+bug ($(BUGZILLA 16230)). `atomicLoad` only loads the head. The tail remains in
+shared memory, and must keep the `shared` qualifier.
+
+When loading a struct that contains indirections, `atomicLoad` now returns a
+wrapper that provides getters which return properly typed values.
+
+When loading a class reference, `atomicLoad` now leaves the `shared` qualifier
+on.
+
+---
```
CyberShadow: Spurious code block delimiter

andralex: Great catch! Do we have a bugzilla for the missing line number in the error message "Error: unmatched --- in DDoc comment"?

aG0aep6G:
> > Since field is a method, you now get a delegate instead of a pointer to the returned field.
> 
> That's what @property is for ;), see https://run.dlang.io/is/cNC6UA.

That prints:

```
int delegate()
int delegate() @property
```

I.e., it shows that `@property` doesn't help.

PR #1606, opened 2016-07-05
OpenBSD: Add OpenBSD definitions to core.stdc

redstar:
Add OpenBSD definitions to core.stdc.

This is the first part of my changes for OpenBSD.


WalterBright:
Check the DAutoTest failure?

redstar:
I don't understand the failure. Why is `} else version ( OpenBSD )` an unexpected declaration?

redstar:
With a closer look I found the problem.

JackStouffer:
Why not use the new alias syntax since the old one is (supposedly) scheduled for deprecation?

PR #1607, opened 2016-07-06
rt.util.utf: remove 'return'

WalterBright:
Slicing a ref static array is equivalent to taking the address of a ref, which is not allowed in safe code.

### DIFF
--- a/src/rt/util/utf.d
+++ b/src/rt/util/utf.d
@@ -583,7 +583,7 @@ void validate(S)(in S s)
 /* =================== Conversion to UTF8 ======================= */
 
 @safe pure nothrow @nogc
-char[] toUTF8(return out char[4] buf, dchar c)
+char[] toUTF8(char[] buf, dchar c)
     in
     {
         assert(isValidDchar(c));
@@ -692,7 +692,7 @@ string toUTF8(in dchar[] s)
 /* =================== Conversion to UTF16 ======================= */
 
 @safe pure nothrow @nogc
-wchar[] toUTF16(return out wchar[2] buf, dchar c)
+wchar[] toUTF16(wchar[] buf, dchar c)
     in
     {
         assert(isValidDchar(c));


WalterBright:
This is blocking https://github.com/dlang/dmd/pull/5909

aG0aep6G:
This opens the door for bugs where a `buf` is passed that's too small. Maybe just make the functions `@system` instead?

WalterBright:
That's what array bounds checking is for.

aG0aep6G:
> That's what array bounds checking is for.

Bounds checking prevents catastrophic fallout, but it doesn't prevent one from writing buggy code.

[std.utf.encode](https://dlang.org/phobos/std_utf.html#.encode) takes a static array but returns a length instead of a slice. Makes it a bit more cumbersome to use, but it's `@safe` and you can't pass too small a buffer. Could be an option here.

Anyway, taking dynamic arrays isn't too bad, I suppose. At least put asserts about the lengths of the buffers, so that the `toUTF8`/`toUTF16` fail early. Otherwise we're going to get code that works for ASCII and maybe European Unicode characters but fails with stuff that needs more space.

WalterBright:
Returning a length means more executable code by putting the buffer overflow check in every of the 22 callers rather than once in the callee.

> Otherwise we're going to get code that works for ASCII and maybe European Unicode characters but fails with stuff that needs more space.

The unittests are supposed to cover that. Adding asserts to overlap with array bounds checking is overkill.

aG0aep6G:
> The unittests are supposed to cover that.

Wishful thinking.

> Adding asserts to overlap with array bounds checking is overkill.

Relying on tests here is sloppy and is going to lead to more bugs.

andralex:
Yeah, the old API wasn't good at all - the moral equivalent of taking a struct by reference and returning a pointer to a member. The change improves things in the sense that it leaves it to the caller to deal with stack-allocated arrays. @WalterBright what is the plan for annotating these functions to clarify they will return a subslice of the array?

WalterBright:
> what is the plan

Support converting a `ref` into a `scope`, but that's towards the end of getting this all to work, as both the `ref` and `scope` stuff has to be fully working first.

andralex:
Auto-merge toggled on

WalterBright:
thanks!

PR #1608, opened 2016-07-07
another incorrect use of return ref

WalterBright:
Blocking https://github.com/dlang/dmd/pull/5909


aG0aep6G:
What's incorrect here? https://github.com/dlang/dmd/pull/5909 says: "'return' without 'ref' implies 'scope'". But here it's `return` with `ref`, so `scope` isn't implied. And there's no `@safe` attribute that's being broken, either.

WalterBright:
I planned to implement that functionality, but this step broke too much code. So I withdraw the patch.

PR #1609, opened 2016-07-09
fix --DRT-gcopt string argument parsing

MartinNowak:
- would match more than just an identifier, e.g.
  'gc:conservative profile:1' was parsed as gc='conservative profile:1'
- was restricted to known GC names, instead we now check
  that one GC was initialized
- parse const(char)[] arguments as simple identifier [_a-zA-Z][_a-zA-Z0-9]*

### DIFF
--- a/src/gc/config.d
+++ b/src/gc/config.d
@@ -26,7 +26,7 @@ struct Config
 {
     bool disable;            // start disabled
     ubyte profile;           // enable profiling with summary when terminating program
-    const(char)[] gc = "conservative"; // select gc implementation conservative|manual
+    string gc = "conservative"; // select gc implementation conservative|manual
 
     size_t initReserve;      // initial reserve (MB)
     size_t minPoolSize = 1;  // initial and minimum pool size (MB)
@@ -74,7 +74,7 @@ struct Config
                cast(long)maxPoolSize, cast(long)incPoolSize, heapSizeFactor);
     }
 
-    bool parseOptions(const(char)[] opt)
+    bool parseOptions(string opt)
     {
         opt = skip!isspace(opt);
         while (opt.length)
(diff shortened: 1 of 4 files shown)


src/gc/config.d
```diff
@@ -26,7 +26,7 @@ struct Config
 {
     bool disable;            // start disabled
     ubyte profile;           // enable profiling with summary when terminating program
-    string gc = "conservative"; // select gc implementation conservative|manual
+    const(char)[] gc = "conservative"; // select gc implementation conservative|manual
```
WalterBright: why this change?

MartinNowak: If we don't know all GCs that can be selected (b/c one w/ a new name can be linked in), then we can't have all names as string symbols. I'm now just taking a slice of the DRT-gcopt argument, which can come from the command line, a string constant in the binary, or if enabled from the environment (see [gc_config](https://dlang.org/spec/garbage.html#gc_config)).
We could malloc and copy the string to make it immutable, but at the time gc_init runs, no usercode has run, so it's hardly possible to change cmdline or env variables. And if someone does it, they simply have to deal w/ this.

MartinNowak:
Ping, also @Jebbs who wrote the former code.

src/gc/config.d
```diff
@@ -190,27 +190,10 @@ bool parse(T:const(char)[])(const(char)[] optname, ref const(char)[] str, ref T
...
-    }
-    else
-    {
-        return parseError("conservative or manual", optname, str);
-    }
-
+    // [_a-zA-Z][_a-zA-Z0-9]*]
+    if (str[0] != '_' && !isalpha(str[0]))
+        return parseError("an identifier", optname, str);
+    res = str[0 .. $ - str[1 .. $].find!(c => c != '_' && !isalnum(c)).length];
```
rainers: We are pretty permissive about characters in option names, we should not be more picky with the name of the GC. I think everything but the same terminators should be accepted (i.e. ' ', ':' and '='). This avoids having to camel-case names composed of multiple words, e.g. `mark-and-sweep`.

src/gc/config.d
rainers: `parse` in `gc.config.initialize` receives a string, doesn't that allow to keep the options as strings aswell? It would be surprising if the referred string would be changed after initialization...

MartinNowak:
Both done

rainers:
Auto-merge toggled on

PR #1610, opened 2016-07-09
Expose gc_stats as core.memory.GC.stats

Dicebot:
Replaces https://github.com/dlang/druntime/pull/1591 with
changes as suggested by @MartinNowak


WalterBright:
By removing several fields, it's become rather useless. I put those stats in there because they were useful for debugging and tuning the implementation.

MartinNowak:
> By removing several fields, it's become rather useless. I put those stats in there because they were useful for debugging and tuning the implementation.

Those are not portable to other GCs, not even to the already planned GC changes. We want to get rid of pools and move the freelists to thread local caches.
This coarse API is intended for programs to monitor GC usage, not for GC implementors to debug their code.
If you want implementation specific stats from the GC, you can already use `--DRT-gcopt=profile:2`.

src/core/memory.d
```diff
@@ -159,6 +159,17 @@ struct GC
...
 
     /**
+     * Aggregation of GC stats to be exposed via public API
+     */
+    static struct Stats
+    {
+        /// total size of GC heap
+        size_t heapsize;
+        /// free bytes on the GC heap (might only get updated after a collection)
+        size_t freesize;
```
leandro-lucarella-sociomantic: Any reason for the `alllowercase`? Since we are adding something new, we might follow the current style guide, right?

src/gc/impl/conservative/gc.d
```diff
@@ -1214,7 +1214,7 @@ class ConservativeGC : GC
         size_t bsize = 0;
 
         //debug(PRINTF) printf("getStats()\n");
-        memset(&stats, 0, GCStats.sizeof);
+        memset(&stats, 0, core.memory.GC.Stats.sizeof);
```
ZombineDev: Nitpick: remove the `memset` because it is redundant for `out` parameters. (See https://dlang.org/spec/function.html#parameters)

Dicebot: It is a bad idea to put unrelated changes in pull requests. Please open a new one.

src/gc/impl/conservative/gc.d
```diff
@@ -1244,9 +1240,8 @@ class ConservativeGC : GC
 
         usize = bsize - flsize;
 
-        stats.poolsize = psize;
-        stats.usedsize = bsize - flsize;
-        stats.freelistsize = flsize;
+        stats.heapsize = bsize - flsize;
```
MartinNowak: That's misleading, you're computing used size here.
Either rename the stats parameter or change the computation.

MartinNowak:
Some improvements in https://github.com/Dicebot/druntime/pull/1.

ZombineDev:
Does it make sense to add a test for this in https://github.com/dlang/druntime/tree/master/test? Perhaps in a separate PR?

MartinNowak:
> Does it make sense to add a test for this in https://github.com/dlang/druntime/tree/master/test? Perhaps in a separate PR?

Writing a test makes sense, but it shouldn't be too sensitive to different values getting reported.

MartinNowak:
Auto-merge toggled on

mihails-strasuns-sociomantic:
Yeah I am not sure what exactly to test as the stats are supposed to show GC internal memory data, not match exactly allocation volume from the user code. Checking that used stats increase somewhat if GC allocation happens is pretty much only reliable thing.

schveiguy:
> where should I submit release note update PR

If you close the bugzilla issue, it will automatically get included in the changelog. If you want to make a more detailed entry in the changelog, update the `changelog.dd` file in this repo.

PR #1612, opened 2016-07-11
Check for overflow before incrementing length in insertBack

ibuclaw:
`insertBack` doesn't have overflow protection, and so it can cause undefined behaviour or a crash to occur.  e.g:
```
auto ary = Array!size_t(null, -1);
ary.insertBack(0);  // SEGV.
```

This change actually helps GDC produce better code, see small test below.

Before: http://goo.gl/zp3Dto
After: http://goo.gl/LSvUgP

### DIFF
--- a/src/rt/util/container/array.d
+++ b/src/rt/util/container/array.d
@@ -33,12 +33,22 @@ nothrow:
 
     @property void length(size_t nlength)
     {
-        if (nlength < length)
-            foreach (ref val; _ptr[nlength .. length]) common.destroy(val);
-        _ptr = cast(T*)common.xrealloc(_ptr, nlength * T.sizeof);
-        if (nlength > length)
-            foreach (ref val; _ptr[length .. nlength]) common.initialize(val);
-        _length = nlength;
+        import core.checkedint : mulu;
+
+        bool overflow = false;
+        size_t reqsize = mulu(T.sizeof, nlength, overflow);
+        if (!overflow)
+        {
+            if (nlength < length)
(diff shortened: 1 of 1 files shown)


schveiguy:
Wait, how does this type even work? There's no public members or ctor.

At the very least, this needs a new unit test to verify the behavior

schveiguy:
This seems rather coner-case-ish. What about:

```D
auto ary = Array!size_t(null, -2);
ary.insertBack(0);
```

ibuclaw:
> auto ary = Array!size_t(null, -2);

Well, that doesn't trigger an overflow for a start, but `realloc` would never allocate that much memory, meaning the second loop will SEGV too.

ibuclaw:
Ah, even better. I can use `onOutOfMemoryError` for the test.

There would be two cases to check for.
- Overflow length
- Failed realloc.

MartinNowak:
That's rather pointless. This data structure is for internal druntime usage, and if we had arrays spanning the whole address space we had a fairly different.
Also insertAppend is not performance sensitive, either you preallocate whatever space you need, or you don't care about performance (e.g. in tests) and live with reallocations anyhow.
Rather than adding quite so e ugly code, a new dependency, and another problematic Error catching test, let's rather close this.

ibuclaw:
This is a blocker for building unittest with `-Werror` on my side: http://goo.gl/QHBVI0

WalterBright:
Checking for overflow is a reality we all need to get used to if we want to convince people that D is good for safe code.

WalterBright:
Auto-merge toggled on

PR #1613, opened 2016-07-17
test_runner can run all tests as well as individual modules

atilaneves:


WalterBright:
Auto-merge toggled on

MartinNowak:
What's the purpose of this?

atilaneves:
Purpose: making the test runner better - I for one don't think we should have as much logic as we do in the makefiles

joakim-noah:
Does running all tests actually work?  I had to [disable two test blocks in rt.lifetime when running all the tests in a single process](https://gist.github.com/joakim-noah/069036835dd92cf648a8219bfe4d68c3#file-druntime_1-0-0_ldc_arm-L32), which this PR enables but doesn't exercise in the auto-tester.

atilaneves:
No, running all tests in the same process doesn't work. I get a failure in std.stdio, but without this patch we wouldn't even know that.

MartinNowak:
> No, running all tests in the same process doesn't work. I get a failure in std.stdio, but without this patch we wouldn't even know that.

But it's more likely to be a bad test than a real problem ;).

PR #1616, opened 2016-07-19
utf.d: use safe construct for .ptr instead

WalterBright:
Blocking https://github.com/dlang/dmd/pull/5860

### DIFF
--- a/src/rt/util/utf.d
+++ b/src/rt/util/utf.d
@@ -767,7 +767,7 @@ wptr toUTF16z(in char[] s)
         }
     }
     r ~= '\000';
-    return r.ptr;
+    return &r[0];
 }
 
 /** ditto */


andralex:
noice

andralex:
Auto-merge toggled on

schveiguy:
Hm... since we *know* that r has a 0 termination, we know that r[0] is valid.

Seems like a wasteful bounds check. But of course, this can be a further enhancement, needs a `@trusted` escape.

ugh, and then there's the `r.length = s.length + 1` followed by `r.length = 0`. Which does nothing except throw away an allocation :)

WalterBright:
> ugh, and then there's the r.length = s.length + 1 followed by r.length = 0. Which does nothing except throw away an allocation :)

It's meant to reserve the space, not throw away the allocation.

aG0aep6G:
> It's meant to reserve the space, not throw away the allocation.

Would need assumeSafeAppend to make that work.

PR #1617, opened 2016-07-26
core.time.d: fix unsafe use of .ptr

WalterBright:
blocks https://github.com/dlang/dmd/pull/5860


jmdavis:
Auto-merge toggled on

PR #1618, opened 2016-07-30
Enable uint128_t/int128_t if ucent/cent type is implemented.

redstar:
This is a better approach than just commenting out the lines.


redstar:
My question is if I should also add the other types/values, e.g. `INT128_MAX`.

WalterBright:
Auto-merge toggled on

PR #1619, opened 2016-07-30
Centify core.checkedint

redstar:
Adds support for cent/ucent types. This is required for an D frontend
which supports cent/ucent.


JackStouffer:
Looks like a lot of repeated code. Why not use templates?

PR #1620, opened 2016-07-31
Enable coverage statistics

wilzbach:
I copied the `.travis.yml` from Phobos over and kept the code style bits. For the simple style checks there exist regexes that can be applied in future work, see e.g. https://github.com/dlang/phobos/pull/4385 and enabling Dscanner is also an open task (volunteers wanted!).

### Other open tasks for future work:

- coverage of the separate `test` files isn't included at the moment. While it is possible to set the coverage directory with `dmd_coverDestPath`, a global way to set the coverage destination dir is missing. Maybe we can add a CLI flag to DMD and modify the `common.mak` or has anyone a better idea?
- coverage for 32-bit - can we simply run with `-m32` on Travis?
- coverage for more platforms ([codecov-bash](https://github.com/codecov/codecov-bash) should be run on the AutoTester too)

edit: a preview of the coverage view can be found [here](https://codecov.io/gh/wilzbach/druntime/tree/afeeedf3ec2d9303b31a988b17c115d85c23f855)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,37 @@
+sudo: false
+os: linux
+language: d
+d: dmd-2.071.0
+git:
+  depth: 1
+install:
+ - mkdir druntime
+ - ls -1 | grep -v ^druntime | xargs -I{} mv {} druntime
+ - git clone --depth=1 https://github.com/dlang/dmd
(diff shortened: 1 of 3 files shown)


WalterBright:
Auto-merge toggled on

PR #1621, opened 2016-08-04
add toFloatPrec/toDoublePrec/toRealPrec

WalterBright:
This is necessary but not entirely sufficient. The compiler needs to recognize them as intrinsics in order for CTFE to also respect the precision.

Also, being intrinsics will avoid the function call overhead.

### DIFF
--- a/src/core/math.d
+++ b/src/core/math.d
@@ -164,3 +164,101 @@ unittest
     }
 }
 
+/*************************************
+ * Round argument to float precision.
+ *
+ * D language types specify a minimum precision, not
+ * a maximum. toFloatPrec() forces float precision.
+ * Params:
+ *      f = value to convert
+ * Returns:
+ *      f in float precision
+ */
+
+@safe pure nothrow
+float toFloatPrec(float f)  { pragma(inline, false); return f; }
+
+/// ditto
+@safe pure nothrow
+float toFloatPrec(double f) { pragma(inline, false); return cast(float)f; }
+
+/// ditto
+@safe pure nothrow
+float toFloatPrec(real f)   { pragma(inline, false); return cast(float)f; }
+
+@safe unittest
+{
+    static float f = 1.1f;
+    static double d = 1.1;
+    static real r = 1.1L;
+    f = toFloatPrec(f + f);
+    f = toFloatPrec(d + d);
+    f = toFloatPrec(r + r);
+}
+
+
+/*************************************
+ * Round argument to double precision.
+ *
+ * D language types specify a minimum precision, not
+ * a maximum. toDoublePrec() forces double precision.
+ * Params:
+ *      f = value to convert
+ * Returns:
+ *      f in double precision
+ */
+
+@safe pure nothrow
+double toDoublePrec(float f)  { pragma(inline, false); return f; }
+
+/// ditto
+@safe pure nothrow
+double toDoublePrec(double f) { pragma(inline, false); return f; }
+
+/// ditto
+@safe pure nothrow
+double toDoublePrec(real f)   { pragma(inline, false); return cast(double)f; }
+
+@safe unittest
+{
+    static float f = 1.1f;
+    static double d = 1.1;
+    static real r = 1.1L;
+    d = toDoublePrec(f + f);
+    d = toDoublePrec(d + d);
+    d = toDoublePrec(r + r);
+}
+
+
+/*************************************
+ * Round argument to real precision.
+ *
+ * D language types specify a minimum precision, not
+ * a maximum. toRealPrec() forces real precision.
+ * Params:
+ *      f = value to convert
+ * Returns:
+ *      f in real precision
+ */
+
+@safe pure nothrow
+real toRealPrec(float f)  { pragma(inline, false); return f; }
+
+/// ditto
+@safe pure nothrow
+real toRealPrec(double f) { pragma(inline, false); return f; }
+
+/// ditto
+@safe pure nothrow
+real toRealPrec(real f)   { pragma(inline, false); return f; }
+
+@safe unittest
+{
+    static float f = 1.1f;
+    static double d = 1.1;
+    static real r = 1.1L;
+    r = toRealPrec(f + f);
+    r = toRealPrec(d + d);
+    r = toRealPrec(r + r);
+}
+
--- a/changelog/toFloat.dd
+++ b/changelog/toFloat.dd
@@ -1,0 +1,11 @@
+add toFloatPrec(), toDoublePrec(), and toRealPrec()
+
+Some floating point algorithms, such as
+Kahan-Babuska-Neumaier Summation, require rounding
+to specific precisions. Rounding to precision after
+every operation, however, loses overall precision in
+the general case and is a runtime performance problem.
+
+Adding these functions guarantee the rounding at required
+points in the code, and document where in the algorithm
+the requirement exists.


schveiguy:
Why is `float(f)` not sufficient?

WalterBright:
> Why is float(f) not sufficient?

Because the type and the precision are not the same thing in D semantics.

src/core/math.d
```diff
@@ -164,3 +164,101 @@ unittest
...
+ *
+ * D language types specify a minimum precision, not
+ * a maximum. toFloat() forces float precision.
+ * Params:
+ *      f = value to convert
+ * Returns:
+ *      f in float precision
+ */
+
+@safe pure nothrow
```
schveiguy: `@nogc`

src/core/math.d
```diff
@@ -164,3 +164,101 @@ unittest
...
+
+@safe unittest
+{
+    static float f = 1.1f;
+    static double d = 1.1;
+    static real r = 1.1L;
+    f = toFloat(f + f);
+    f = toFloat(d + d);
+    f = toFloat(r + r);
+}
```
schveiguy: I'm not sure I understand the point of these unit tests... Shouldn't you test something that verifies the precision?

src/core/math.d
WalterBright: @nogc is there at the top of the file

WalterBright:
I renamed them to `toFloatPrec`, etc., to clarify what it does.

src/core/math.d
WalterBright: I initially tried that, but got different results on different platforms :-(. The compiler needs to recognize them as intrinsics to get it consistent.

src/core/math.d
```diff
@@ -164,3 +164,101 @@ unittest
...
+float toFloatPrec(real f)   { pragma(inline, false); return cast(float)f; }
+
+@safe unittest
+{
+    static float f = 1.1f;
+    static double d = 1.1;
+    static real r = 1.1L;
+    f = toFloatPrec(f + f);
+    f = toFloatPrec(d + d);
+    f = toFloatPrec(r + r);
```
qznc: Shouldn't we `assert(f+f == toFloatPrec(d + d));` or something like that?

src/core/math.d
```diff
@@ -164,3 +164,101 @@ unittest
...
+
+/// ditto
+@safe pure nothrow
+float toFloatPrec(real f)   { pragma(inline, false); return cast(float)f; }
+
+@safe unittest
+{
+    static float f = 1.1f;
+    static double d = 1.1;
+    static real r = 1.1L;
```
Temtaime: Static for which purpose ?

src/core/math.d
```diff
@@ -164,3 +164,101 @@ unittest
...
+float toFloatPrec(float f)  { pragma(inline, false); return f; }
+
+/// ditto
+@safe pure nothrow
+float toFloatPrec(double f) { pragma(inline, false); return cast(float)f; }
+
+/// ditto
+@safe pure nothrow
+float toFloatPrec(real f)   { pragma(inline, false); return cast(float)f; }
+
```
nomad-software: Why use overloading when these could be one templated function?

src/core/math.d
WalterBright: I did that initially, but got different results on some platforms.

src/core/math.d
WalterBright: So it isn't constant folded.

src/core/math.d
WalterBright: Because template functions are a pain to turn into compiler recognized intrinsics.

MartinNowak:
> Why is float(f) not sufficient?

Just wondered about that myself for a second, b/c the temporary precision even for a float can be real or double (depending on which FPU is used). This function should really force the rounding.

MartinNowak:
For reference [Issue 9937 – CTFE floats don't overflow correctly](https://issues.dlang.org/show_bug.cgi?id=9937), [Issue 13474 – Discard excess precision for float and double (x87)](https://issues.dlang.org/show_bug.cgi?id=13474).

src/core/math.d
jmdavis: You might want to add a comment to that effect so that no enterprising individual changes it later - especially if there's a decent chance that the tests will pass either way.

schveiguy: > there's a decent chance that the tests will pass either way.

Nothing is being tested :) So of course it will pass either way!

MartinNowak:
Changelog entry missing. Talk to our users!

> This is necessary but not entirely sufficient. The compiler needs to recognize them as intrinsics in order for CTFE to also respect the precision.

Will we get the CTFE part in the next 2 weeks? Please prepare the dmd PR already.
I don't want to release only half of the feature.

ibuclaw:
> Because the type and the precision are not the same thing in D semantics.

Next question, shouldn't an inlined cast already be doing this?

MartinNowak:
> Next question, shouldn't an inlined cast already be doing this?

DMD "optimizes" some floating point casts, even though they might have side effects (like setting precision exception flags).
Both gdc and ldc already support the obvious C semantic and are not optimizing away float casts.
```d
real test(real a, real b)
{
    return a + cast(float)b;
}
```
Actually ended up here b/c the current gdc test failure is actually caused by dmd ignoring those casts.
https://github.com/dlang/dmd/blob/386ef6d8c8e09efbc9a5bd755859e078e6a0dfd0/src/todt.d#L306

I'm inclined to say that adding an intrinsic is a bad workaround and dmd should just make the explicit casts work.

ibuclaw:
> Actually ended up here b/c the current gdc test failure is actually caused by dmd ignoring those casts.

Really? Wow.

MartinNowak:
Have you added the necessary intrinsics to the compiler yet?

MartinNowak:
@WalterBright

src/core/math.d
quickfur: I think he meant to document the reason `static` is there, so that somebody doesn't come along later, decide the `static` is redundant and delete it without knowing why it was done in the first place.

WalterBright:
I don't see much point to adding the intrinsics to the compiler until after this is pulled, because they won't work without this PR.

wilzbach:
How about setting all the intrinsics to private and just exposing a `toPrec!double` template to the user? This would
- be similar to `std.conv.to`
- probably be necessary in generic code anyways
- hide the ugly/redundant intrinsics API documentation from the user

WalterBright:
A template wrapper over this adds little value, and subtracts value due to the user wondering if the template is different from the intrinsic.

It would be nice if we could move forward and pull this. It's an often asked for capability, that has now sat around for over a year.

andralex:
Hm, I could edit other PRs but not this one. Github gurus, why?

jmdavis:
> Hm, I could edit other PRs but not this one. Github gurus, why?

When you own a PR, there's a checkbox on the right side labeled "Allow edits from maintainers." It's checked by default, but if the PR owner unchecks it, then folks with commit access to the repo can't mess with your PR like they can when it's checked. That may or may not be what's happening here, but it's one possible explanation.

schveiguy:
It's also a relatively new feature. Perhaps when this PR was created the checkbox didn't exist.

MartinNowak:
> Hm, I could edit other PRs but not this one. Github gurus, why?

Cause it's old and thus the option wasn't set, only the PR opener can change that flag.

andralex:
@WalterBright you may want to try checking "Allow edits from maintainers", I have a few minor ones. Thx!

WalterBright:
> @andralex Allow edits from maintainers

Now allowed! (I was unaware of that switch, thanks for pointing it out.)

WalterBright:
@MartinNowak I prefer to keep intrinsics as simple as possible, because they are warty, brittle things in the compiler. Much as I tried to make templates simple, they aren't.

andralex:
Three traditional intrinsic functions fronted by a template seems like a good way to go. @WalterBright ?

MartinNowak:
Seems like my point wasn't really understood, it's as trivial as

https://github.com/MartinNowak/druntime/blob/4db59f1949ff7ab32a7f7653cf3742f688eff428/src/core/math.d#L180-L205

and

https://github.com/MartinNowak/dmd/blob/b53ecdd3bb4a34819914e4f10dd7080e282f95cf/src/ddmd/builtin.d#L360-L377

WalterBright:
Because of the way the x87 works, this has to be a compiler intrinsic. The intrinsic itself should be a simple function, to make understanding it easier, and the compiler implementation as straightforward as possible.

If you want to make a template wrapper for it, that is something that can be layered on by phobos, analogously to the checkedint stuff.

WalterBright:
To clarify a bit, the idea of druntime is to provide the part of the library that needs to be adjusted when porting D to different platforms, it supplies the "core", so to speak, and the under-the-hood parts. They aren't meant to be user facing stuff.

Compiler intrinsics are really about extending the instruction set of the language

The user facing stuff goes in Phobos, and the checkedint stuff is a particularly nice example of this.

This dichotomy isn't always followed, of course, but we should at least try to maintain it when adding new stuff.

ibuclaw:
My thoughts are that any new intrinsic should have consensus with other vendors. This means finding a middle ground where all can implement, which typically does end up being easier to use than the original.

There's nothing particularly wrong about the template solution IMO.

MartinNowak:
Unbelievable, is anyone actually reading what I write?
There is nothing simpler about a function than about a template function, both are simply recognized as intrinsics by matching their mangling, see my examples above(https://github.com/dlang/druntime/pull/1621#issuecomment-330524874). In fact this is all there is to the whole implementation.
BTW, [`std.math.isFinite`](https://dlang.org/library/std/math/is_finite.html) and [`std.math.isNaN`](https://dlang.org/library/std/math/is_nan.html) are precedences of templated intrinsics, I don't think anyone has a problem understanding those.
https://github.com/dlang/dmd/blob/241236394b303a1ea396fe8e20a0c25b5c09a90e/src/ddmd/builtin.d#L310

MartinNowak:
See #1941

PR #1624, opened 2016-08-07
add @nogc to bytesHash

aG0aep6G:
See <http://forum.dlang.org/post/gmeowrdivjpjhezbdror@forum.dlang.org>.

By the way, `bytesHash` should not be `@trusted`. It takes pointer and length separately.

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -432,10 +432,10 @@ version(AnyX86)
     }
 }
 
-@trusted pure nothrow
+@trusted pure nothrow @nogc
 size_t bytesHash(const(void)* buf, size_t len, size_t seed = 0)
 {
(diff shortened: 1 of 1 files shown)


src/core/internal/hash.d
```diff
@@ -432,18 +432,18 @@ version(AnyX86)
     }
 }
 
-@trusted pure nothrow
+@trusted pure nothrow @nogc
 size_t bytesHash(const(void)* buf, size_t len, size_t seed = 0)
 {
-    static uint rotl32(uint n)(in uint x) pure nothrow @safe
+    static uint rotl32(uint n)(in uint x) pure nothrow @safe @nogc
```
WalterBright: this function already exists in std.core.bitop

PR #1625, opened 2016-08-07
Make Coverage checking more stable

wilzbach:
After @MartinNowak's nice work at https://github.com/dlang/dmd/pull/6022, here's a similar setup for Druntime.

Short summary
- Travis clogs our PR queue -> use CircleCi for coverage testing
- The coverage testing script now allows to run for 64-bit & 32-bit (a run takes only 3 minutes, so I think we can enable here two containers too)
- moved style check to the `posix.mak` as `style` target


wilzbach:
> moved style check to the posix.mak as style target

Oh sorry ... `checkwhitespace` was already a defined target in the Makefile,  so I changed to `style` require it. `style` will be used for a similar PR at Phobos, s.t. we have clear naming and the style target at druntime could/should be expanded anyhow ;-)

MartinNowak:
Auto-merge toggled on

MartinNowak:
Can we also please enable CircleCI for druntime and Phobos @WalterBright @andralex.
https://circleci.com/add-projects

andralex:
@MartinNowak are we good now?

wilzbach:
> @MartinNowak are we good now?

Yep we are all set & all projects are enabled, only the configuration for Phobos in pending
See it in "live" e.g here:

https://github.com/dlang/druntime/pull/1602

PR #1626, opened 2016-08-10
hash.d: make .ptr @safe

WalterBright:
Blocking https://github.com/dlang/dmd/pull/5860


WalterBright:
@klickverbot You're right, I'll look into a revision. But in the meantime, this is no longer a blocker.

WalterBright:
https://github.com/dlang/druntime/pull/1627

PR #1627, opened 2016-08-10
@trusted function is really @system

WalterBright:
The default argument is also a bad idea.


andralex:
Auto-merge toggled on

PR #1628, opened 2016-08-11
add codecov config for dlang-bot

wilzbach:
As we transition from Travis to CircleCi the script was never run there, so I just open this PR to verify everything is ok. Sorry for the noise.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,4 +1,7 @@
 # https://github.com/codecov/support/blob/master/codecov.yml
+codecov:
+  bot: dlang-bot
+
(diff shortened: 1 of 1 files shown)


wilzbach:
> As we transition from Travis to CircleCi the script was never run there, so I just open this PR to verify everything is ok. Sorry for the noise.

Everything worked fine, so I just rebased this PR to add `dlang-bot` to the CodeCov file, s.t. it can post updates to the CI status board.

CC @MartinNowak 

See also: https://github.com/dlang/dmd/pull/6046

PR #1630, opened 2016-08-12
Fix FP_ILOGB constants for GlibC and non-X86

jpf91:
These constants are actually architecture dependent:
http://osxr.org:8080/glibc/ident?v=glibc-2.19&_i=FP_ILOGBNAN


ibuclaw:
Someone should make Codecov a soft fail.

ibuclaw:
Auto-merge toggled on

PR #1631, opened 2016-08-13
Add badges to README

wilzbach:
Similar as https://github.com/dlang/phobos/pull/4728 and https://github.com/dlang/dmd/pull/6057, this adds nice badges to the README. It's a small thing, but it matters quite a bit for "the first five minutes" ;-)


andralex:
Auto-merge toggled on

PR #1632, opened 2016-08-14
CircleCi: fix for non-PR builds & add git merge with upstream

wilzbach:
Follow-up to https://github.com/dlang/dmd/pull/6066

I also took the opportunity
- to provide a default base branch (in case of rate limiting) and
- merge the latest upstream branch into the current branch (if this fails, all other CI & Github will show an error too)

CC @klickverbot

### DIFF
--- a/circleci.sh
+++ b/circleci.sh
@@ -51,11 +51,17 @@ clone() {
 }
 
 coverage() {
-    if [ -n ${CIRCLE_PR_NUMBER:-} ]; then
-        local base_branch=$(curl -fsSL https://api.github.com/repos/dlang/dmd/pulls/$CIRCLE_PR_NUMBER | jq -r '.base.ref')
+    base_branch="master"
+    if [ -n "${CIRCLE_PR_NUMBER:-}" ]; then
+        base_branch=$(curl -fsSL https://api.github.com/repos/dlang/dmd/pulls/$CIRCLE_PR_NUMBER | jq -r '.base.ref')
     else
-        local base_branch=$CIRCLE_BRANCH
+        base_branch=$CIRCLE_BRANCH
     fi
+
+    # merge with latest upstream branch
+    git remote add upstream git@github.com:dlang/druntime.git
+    git fetch upstream
+    git merge upstream/$base_branch
 
     clone https://github.com/dlang/dmd.git ../dmd $base_branch
 


wilzbach:
> merge the latest upstream branch into the current branch (if this fails, all other CI & Github will show an error too)
...
> 73.91% (+0.05%) compared to 0745846

Hmm turns out that CodeCov maps the PR to latest ref to the origin branch it is pointing to and not the latest commit on master/stable

CC @MartinNowak

wilzbach:
> Is this something we can fix/change in the CodeCov settings?

I doubt that's possible, but pinging @stevepeak for help :)

stevepeak:
@wilzbach @klickverbot we are working on a new merge commit process: https://github.com/codecov/support/wiki/Merge-Commits

circleci.sh
```diff
@@ -51,12 +51,18 @@ clone() {
 }
 
 coverage() {
-    if [ -n ${CIRCLE_PR_NUMBER:-} ]; then
-        local base_branch=$(curl -fsSL https://api.github.com/repos/dlang/dmd/pulls/$CIRCLE_PR_NUMBER | jq -r '.base.ref')
+    base_branch="master"
```
MartinNowak: You can keep the local, don't want that variable outside of the function.

circleci.sh
```diff
@@ -51,12 +51,18 @@ clone() {
...
+    if [ -n "${CIRCLE_PR_NUMBER:-}" ]; then
+        base_branch=$(curl -fsSL https://api.github.com/repos/dlang/dmd/pulls/$CIRCLE_PR_NUMBER | jq -r '.base.ref')
```
MartinNowak: If curl fails, this will be set to an empty base branch.
Better check for empty string below `if [ -z $base_branch ]; then base_branch=master; fi`.

circleci.sh
```diff
@@ -51,12 +51,18 @@ clone() {
...
     else
-        local base_branch=$CIRCLE_BRANCH
+        base_branch=$CIRCLE_BRANCH
     fi
 
+    # merge with latest upstream branch
+    git remote add upstream git@github.com:dlang/druntime.git
+    git fetch upstream
+    git merge upstream/$base_branch
```
MartinNowak: Instead of merging here, we could just checkout github's merge reference, e.g. https://gist.github.com/piscisaureus/3342247#gistcomment-608797.

andralex:
@wilzbach ?

MartinNowak:
Let's wait for codecov's merge commit support first.
The non-PR was just fixed in #1658.

PR #1634, opened 2016-08-16
Fix issue 16394 -- document the intializer function behavior correctly.

schveiguy:
The Static array typeinfo only returns an initializer for a single element. This documents that behavior.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -286,9 +286,12 @@ class TypeInfo
     null if none. */
     @property inout(TypeInfo) next() nothrow pure inout @nogc { return null; }
 
-    /** Return default initializer.  If the type should be initialized to all
-    zeros, an array with a null ptr and a length equal to the type size will be
-    returned. */
+    /**
+     * Return default initializer.  If the type should be initialized to all
+     * zeros, an array with a null ptr and a length equal to the type size will
+     * be returned. For static arrays, this returns the default initializer for
+     * a single element of the array, use `tsize` to get the correct size.
+     */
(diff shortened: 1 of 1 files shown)


schveiguy:
See also dlang/phobos#4736

aG0aep6G:
> @aG0aep6G: The PR was automatically assigned to you, please reassign it if you were identified mistakenly.

For the record, I have not actually been assigned, and I can't reassign anything, because I don't have merge rights. [Issue is known.](https://github.com/dlang-bots/mention-bot/issues/7)

On topic:

I still think that a better fix would be to make `initializer` return the correct initializer for the array. But everyone seems to agree that a new method should be added that doesn't need weird exceptions like this. So if that's going to happen, I don't care too much about what `initializer` does, and documenting the weirdness is of course better than not doing so.

src/object.d
```diff
@@ -286,9 +286,12 @@ class TypeInfo
     null if none. */
     @property inout(TypeInfo) next() nothrow pure inout @nogc { return null; }
 
-    /** Return default initializer.  If the type should be initialized to all
-    zeros, an array with a null ptr and a length equal to the type size will be
-    returned. */
+    /**
+     * Return default initializer.  If the type should be initialized to all
+     * zeros, an array with a null ptr and a length equal to the type size will
+     * be returned. For static arrays, this represents a single element of the
```
aG0aep6G: Very minor remark: Is "represents" clear enough? How about "For static arrays, this returns the default initializer for a single element ..."?

schveiguy: I can do that.

PR #1637, opened 2016-08-29
add 'scope' to object.dup()

WalterBright:


ZombineDev:
LGTM

MetaLang:
Just out of curiosity, can `scope` not be inferred here?

WalterBright:
Inferring scope on a parameter is problematic due to its chicken-and-egg nature internal to the function.

andralex:
nice nice

denis-sh:
Sorry if I'm wrong, but currently `scope` is [documented](https://dlang.org/spec/function.html#parameters) as

> references in the parameter cannot be escaped

and `dup` is called for unrestricted set of types so type's postblit can assign references from the parameter to global variable thus violating `scope` storage class assumptions as there is no such restrictions for postblits.

Filed [issue 16504](https://issues.dlang.org/show_bug.cgi?id=16504).

MartinNowak:
What about the criticism? What was the purpose of this change, neither commit message nor PR description list any intentions?
What does "chicken-and-egg nature internal to the function" refer to? Dup is supposed to derive it's attributes from the postblit. Not sure how scope inference works, but dup doesn't escape anything other than calling the postblit.
I guess it's still not possible to mark the `this` pointer in member methods as scope (`this(this) scope {}`), is it?

@denis-sh your argument would be more convincing if it came with a realistic example of ref-escaping postblits.

MartinNowak:
Ping @WalterBright, this needs a clarification. It seems that `scope` can't be applied to all `.dup` arguments.

PR #1638, opened 2016-08-29
[TRIVIAL IMPORTANT] first fixes for static __ctfe

UplinkCoder:
Hi this is a small fix deblocking  https://github.com/dlang/dmd/pull/6098

### DIFF
--- a/src/rt/util/hash.d
+++ b/src/rt/util/hash.d
@@ -30,13 +30,20 @@ size_t hashOf( const(void)[] buf, size_t seed )
     {
         // CTFE doesn't support casting ubyte* -> ushort*, so revert to
         // per-byte access when in CTFE.
-        version( HasUnalignedOps )
-        {
-            if (!__ctfe)
+        if (!__ctfe) {
+            version( HasUnalignedOps )
+            {
                 return *cast(ushort*) x;
+            }
+            else
+            {
+                return ((cast(uint) x[1]) << 8) + (cast(uint) x[0]);
+            }
         }
-
-        return ((cast(uint) x[1]) << 8) + (cast(uint) x[0]);
+        else
+        {
+            return ((cast(uint) x[1]) << 8) + (cast(uint) x[0]);
+        }
     }
 
     // NOTE: SuperFastHash normally starts with a zero hash value.  The seed
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -794,13 +794,18 @@ void volatileStore(ulong * ptr, ulong  value);   /// ditto
 pragma(inline, true)
 uint bitswap( uint x ) pure
 {
-    if (!__ctfe)
+    static if (is(typeof(asmBitswap32(x))))
     {
-        static if (is(typeof(asmBitswap32(x))))
-            return asmBitswap32(x);
-    }
 
-    return softBitswap!uint(x);
+        if (!__ctfe)
+            return asmBitswap32(x);
+        else
+            return softBitswap!uint(x);
+    }
+    else
+    {
+        return softBitswap!uint(x);
+    }
 }
 
 unittest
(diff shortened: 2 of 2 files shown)


src/rt/util/hash.d
```diff
@@ -33,10 +33,14 @@ size_t hashOf( const(void)[] buf, size_t seed )
         version( HasUnalignedOps )
         {
             if (!__ctfe)
+            {
                 return *cast(ushort*) x;
+            }
```
schveiguy: In the case of `version(HasUnalignedOps)`, and `__ctfe == true`, this function has no body.

schveiguy:
I'm a bit puzzled. Why is this needed? Is it because of constant folding?

UplinkCoder:
@schveiguy yes. 
Otherwise the code will be flagged as unreachable.

schveiguy:
Hm... the non-DRY-ness of this looks awful. I suppose this is necessary to work with both compile time and runtime `__ctfe` definitions.

UplinkCoder:
@schveiguy one the static __ctfe goes in.
It can be pretty!

UplinkCoder:
@WalterBright or @andralex please merge.

schveiguy:
I think this may be a big problem. All current code uses `if(__ctfe)`, and if this is not just a drop-in replacement, we are talking about breaking a lot of code for zero benefit. If this PR is any indication, the "fix" is not trivial.

Regardless of the obvious benefits, I think your dmd patch needs to continue to treat `__ctfe` as a magic runtime variable when used at runtime, and a compile-time constant when used at compile time.

UplinkCoder:
The user would get a depreciation warning about unreachable code.
And that warning is totally legitimate.

UplinkCoder:
The only reason this kludge is needed is because the d-auto-tester compiles with -w.
And the dmd pr is not yet in.
once the pr is in the fix it to change all if(__ctfe) to static if (__ctfe) done!

schveiguy:
> The only reason this kludge is needed is because the d-auto-tester compiles with -w.

And so do many many projects. -w finds bugs.

> And that warning is totally legitimate.

That's like saying a function like this:

```D
char foo(bool flag)
{
    if(flag)
        return 'y';
    return 'n';
}
```

when called via `foo(true)` and -inline should flag a "legitimate" unreachable code error. It's useless noise.

UplinkCoder:
Yes you are right. 
However that is a different problem. 
And one which has been brought up before.

schveiguy:
It is not a different problem. The code is reachable, just not with that specific call. I happen to think actually that we are too strict on that requirement. For example, a template with a boolean should not trigger unreachable code -- sometimes the boolean is true, sometimes false. It shouldn't say the code is unreachable, just because it's unreachable in a certain instantiation.

But in this case, you are going to flag lots of existing code that works perfectly fine and is *understood* to be unreachable at compile time or at runtime.

I'm wondering if `__ctfeStatic` or something would be better to introduce, so as not to break existing code.

andralex:
I agree this could be problematic. I routinely mark code that uses the redundant `else` in code reviews.

UplinkCoder:
With dmd's current control-flow-analysis. The else is not redundant. 
I thought about `__ctfeStatic` and discarded it. 
We need to clean the unreachable situation up. 
Not introduce hacks.

jmdavis:
> With dmd's current control-flow-analysis. The else is not redundant. 

As long as you're dealing with runtime checks, if the if returns, the else is redundant, and it's useless noise to have it. It's only with static if that it makes sense, because it actually affects which code is compiled in and not just what is run.  `__ctfe` is a weird beast, because it's affecting what gets run at compile time, but it's still affecting what gets run and not what gets compiled in.

> The only reason this kludge is needed is because the d-auto-tester compiles with -w.

How about making `__ctfe` work in both static if statements and if statements where it has the current behavior in if statements and the new behavior in static if statements and then we deprecate the use of `__ctfe` as a runtime variable. Then you don't run afoul of -w, since deprecation is separate from warnings. Anyone using `__ctfe` will be bugged to update their code, but their code won't break. And once the deprecation is done with, you'll just have `__ctfe`, so it will be clean like you're looking for, but it won't break code in the interim.

UplinkCoder:
@klickverbot making __ctfe a CT-Entity, is fairly useful.
please elaborate on why it is a mistake,

andralex:
@UplinkCoder to make sure I understand, consider a library with the code:

```D
string fun(int x)
{
    if (x < 0) return "negative";
    ... a bunch of code ...
    return "positive";
}
```

Now, say in a particular program `fun` is evaluated during compilation and only with negative values. Does your code introduce a warning about unreachable code?

UplinkCoder:
No not at all.
All my code does is to change `__ctfe` from a variable into a literal. 
However that does not work with the way function are currently handled.

PR #1640, opened 2016-08-30
refactor: remove unnecessary object._rawDup() template

WalterBright:
Reduce the thicket of templates in object.d

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3323,8 +3323,8 @@ unittest
 /// ditto
 @property T[] dup(T:void)(scope const(T)[] a) @trusted
 {
-    if (__ctfe) assert(0, "Cannot dup a void[] array at compile time.");
-    return cast(T[])_rawDup(a);
+    import core.internal.traits : Unconst;
+    return _dup!(const(T), Unconst!T)(a);
 }
 
 /// Provide the .idup array property.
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -3355,28 +3355,30 @@ private U[] _dup(T, U)(scope T[] a) // pure nothrow depends on postblit
...
     {
-        U[] res;
-        foreach (ref e; a)
-            res ~= e;
-        return res;
+        static if (is(T : void))
+            assert(0, "Cannot dup a void[] array at compile time.");
+        else
+        {
+            U[] res;
```
andralex: This code is uncovered.

src/object.d
```diff
@@ -3355,28 +3355,30 @@ private U[] _dup(T, U)(scope T[] a) // pure nothrow depends on postblit
...
+            foreach (ref e; a)
+                res ~= e;
```
andralex: aw, this is quite inefficient, we should look at this soonish

andralex:
OK but you should know that the net number of line actually increases...

src/object.d
WalterBright: Yes, but it was before, as this is a refactoring.

src/object.d
WalterBright: That branch is only executed when doing ctfe.

PR #1642, opened 2016-09-02
remove redundant object.dup(T:void)

WalterBright:


andralex:
red is good

src/object.d
```diff
@@ -3323,12 +3323,6 @@ unittest
         return _dup!(const(T), T)(a);
 }
 
-/// ditto
-@property T[] dup(T:void)(scope const(T)[] a) @trusted
-{
-    import core.internal.traits : Unconst;
-    return _dup!(const(T), Unconst!T)(a);
```
schveiguy: This is more than just removing extra overloads. This fixes a safety issue.

I could compile the following code on the current compiler:

```D
void main() @safe
{
    immutable(int *)[] arr;
    arr ~= new int(4);
    const(void[]) b = arr;
    auto c = b.dup;
    pragma(msg, typeof(c)); // void[]
    int*[] arr2;
    arr2 ~= null;
    void[] x = arr2;
    x[] = c[];
    *arr2[0] = 5;
    assert(*arr[0] == 5);
}
```

andralex: @schveiguy interesting. Could you please add a "negative" unittest with the code above? (e.g. `static assert(!is(typeof(@safe { ... big lambda ... })));`

WalterBright: The safety improves I've implemented but are blocked yields:
```
test.d(11): Error: cannot copy void[] to void[] in @safe code
```

schveiguy: Something doesn't sit right with me, when you can `dup` a `const(void)[]` or an `immutable(void)[]` into a `void[]` without any complaint. Even in non-`@safe` code.

Yes, we have sort of fixed the `void[]` copy because it's dangerous in `@safe` code, even without violating immutability. But that's because the types could be anything, and you may be copying non-pointer data into pointer data. However, in this case, the basic types match, so that isn't the issue. That is, you won't get a dangling pointer. So it's more of a const-safety issue than memory safety.

I don't know if a unittest is appropriate here. I'm not sure what the intent of the `dup` function is at this point.

WalterBright: This PR does not attempt to fix dup(), it just simplifies the template bloat. I recommend making a bugzilla issue out of the @safe issue.

schveiguy:
This PR does change what can be called from `@safe` code. In particular, before this change, `void[]` can be dup'd. After, it cannot.

So this is not a trivial template bloat change, and I think maybe this increased limitation is a good thing, but obviously may break code (that should be broken). I'll file an issue about the larger problem.

PR #1643, opened 2016-09-03
[blocks __ctfeWriteln]change __ctfeWrite/ln to a signature which is easier to support

UplinkCoder:
This change simplifies the function signature of __ctfeWrite and __ctfeWriteln.
This is beneficial for the compiler implementation.
For context see:  https://github.com/dlang/dmd/pull/6101

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3183,9 +3183,12 @@ bool _xopCmp(in void*, in void*)
 {
     throw new Error("TypeInfo.compare is not implemented");
 }
-
-void __ctfeWrite(T...)(auto ref T) {}
-void __ctfeWriteln(T...)(auto ref T values) { __ctfeWrite(values, "\n"); }
+void __ctfeWrite(const string s) {}
+void __ctfeWriteln(const string s)
+{
+  __ctfeWrite(s);
+  __ctfeWrite("\n");
+}
 
 /******************************************
  * Create RTInfo for type T


src/object.d
```diff
@@ -3184,8 +3184,8 @@ bool _xopCmp(in void*, in void*)
     throw new Error("TypeInfo.compare is not implemented");
 }
 
-void __ctfeWrite(T...)(auto ref T) {}
-void __ctfeWriteln(T...)(auto ref T values) { __ctfeWrite(values, "\n"); }
+void __ctfeWrite(const string s) {}
+void __ctfeWriteln(const string s) {}
```
ZombineDev: `const(char)[]` (or `in char[]`) would be more flexible for the user. A common pattern with string mixins is to append to a `char[]` in a loop. Requiring immutability is unnecessary.

UplinkCoder:
It is required because of the implementation on the compiler-side.

src/object.d
ZombineDev: Also, why not `const(char)[]...`, as @kliverbot suggested in the DMD PR?

UplinkCoder:
Because the the compiler treats string and const (char[]) as very different expressions.

UplinkCoder:
@klickverbot anything blocking a merge ?

src/object.d
```diff
@@ -3183,9 +3183,12 @@ bool _xopCmp(in void*, in void*)
 {
     throw new Error("TypeInfo.compare is not implemented");
 }
-
-void __ctfeWrite(T...)(auto ref T) {}
-void __ctfeWriteln(T...)(auto ref T values) { __ctfeWrite(values, "\n"); }
+void __ctfeWrite(const string s) {}
+void __ctfeWriteln(const string s)
```
ZombineDev: Should these functions have `pure`, `@safe`, `nothrow` and `@nogc` attributes?

ZombineDev:
@klickverbot as I mentioned in an earlier comment, both function should take `const(char)[]` arguments. "const" in `const string` is meaningless and "string" is needlessly restrictive. One more point against "string" - since Phobos imports are frowned upon in the dmd testsuite, the only way to conver an integer to string (apart from DIY) is to use sprintf + an unsafe cast from char[] to string. One may argue that the current policy for the dmd testsuite is too restrictive, but I think that is indicative of some of hoops one has to go when `@nogc` is a hard requirement and/or when targeting more constrained platforms where using Phobos would bring too much weight.

If making `__ctfeWrite` work with variable number of arguments (like `pragma (msg, ...)`), or at least accepting a single arugment of type `T` is hard for the time being, can we at least lift the requirement for immutable strings? It makes the user experience unnecessary bad in some situations.

UplinkCoder:
@ZombineDev making it work with const(char)[] makes the code more complex.

ZombineDev:
It makes the compiler's code more complex, not the user's and that's what's more important when evaluating language features, IMO (at least as long as the feature is not un-implementable).

Can some of the code that handles `pragma (msg,   ...)` be reused for `__ctfeWrite[ln]`?

PR #1644, opened 2016-09-04
fix Issue 14439 - aa's keys, values, byKey, byValue not usable in @sa…

WalterBright:
…fe context

The mechanics of the aa's are trustable. This leaves the effect of whether postblit is safe or not intact.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -1923,12 +1923,13 @@ class __cpp_type_info_ptr
 extern (C)
 {
     // from druntime/src/rt/aaA.d
+    private:
 
     // size_t _aaLen(in void* p) pure nothrow @nogc;
-    private void* _aaGetY(void** paa, const TypeInfo_AssociativeArray ti, in size_t valuesize, in void* pkey) pure nothrow;
+    void* _aaGetY(void** paa, const TypeInfo_AssociativeArray ti, in size_t valuesize, in void* pkey) pure nothrow;
     // inout(void)* _aaGetRvalueX(inout void* p, in TypeInfo keyti, in size_t valuesize, in void* pkey);
-    inout(void)[] _aaValues(inout void* p, in size_t keysize, in size_t valuesize, const TypeInfo tiValArray) pure nothrow;
-    inout(void)[] _aaKeys(inout void* p, in size_t keysize, const TypeInfo tiKeyArray) pure nothrow;
+    inout(void)[] _aaValues(inout void* p, in size_t keysize, in size_t valuesize, const TypeInfo tiValArray) pure nothrow @trusted;
+    inout(void)[] _aaKeys(inout void* p, in size_t keysize, const TypeInfo tiKeyArray) pure nothrow @trusted;
     void* _aaRehash(void** pp, in TypeInfo keyti) pure nothrow;
     void _aaClear(void* p) pure nothrow;
 
@@ -1938,10 +1939,10 @@ extern (C)
     // alias _dg2_t = extern(D) int delegate(void*, void*);
     // int _aaApply2(void* aa, size_t keysize, _dg2_t dg);
 
-    private struct AARange
+    struct AARange
     {
     private:
-        this(void* impl, size_t idx) { this.impl = impl; this.idx = idx; }
+        this(void *impl) { this.impl = impl; }
         void* impl;
         size_t idx;
     }
@@ -2045,7 +2046,7 @@ V[K] dup(T : V[K], K, V)(T* aa)
     return (*aa).dup;
 }
 
-auto byKey(K, V)(const(V[K]) aa) pure nothrow @nogc
+auto byKey(T : V[K], K, V)(T aa) pure nothrow @nogc @trusted
 {
     import core.internal.traits : substInout;
 
@@ -2055,20 +2056,21 @@ auto byKey(K, V)(const(V[K]) aa) pure nothrow @nogc
 
     pure nothrow @nogc @trusted:
         @property bool empty() { return _aaRangeEmpty(r); }
-        @property ref front() { return *cast(substInout!K*)_aaRangeFrontKey(r); }
-        void popFront() { _aaRangePopFront(r); }
+        @property ref front() { assert(!empty); return *cast(substInout!K*)_aaRangeFrontKey(r); }
+        void popFront() { assert(!empty); _aaRangePopFront(r); }
         @property Result save() { return this; }
     }
 
-    return Result(((typeof(aa) aa) @trusted => _aaRange(cast(void*)aa))(aa));
+    static assert(!__traits(compiles, __traits(getAliasThis, T)), "cannot have alias this of associative array");
+    return Result(_aaRange(cast(void*)aa));
 }
 
-auto byKey(T : V[K], K, V)(T* aa)
+auto byKey(T : V[K], K, V)(T* aa) pure nothrow @nogc
 {
     return (*aa).byKey();
 }
 
-auto byValue(K, V)(const(V[K]) aa) pure nothrow @nogc
+auto byValue(T : V[K], K, V)(T aa) pure nothrow @nogc @trusted
 {
     import core.internal.traits : substInout;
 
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -1964,15 +1964,15 @@ V[K] dup(T : V[K], K, V)(T* aa)
...
+auto byKey(T : V[K], K, V)(T aa) pure nothrow @nogc @trusted
 {
     import core.internal.traits : substInout;
 
     static struct Result
     {
         AARange r;
 
-    pure nothrow @nogc:
+    pure nothrow @nogc @trusted:
```
schuetzm: Seeing `@trusted` followed by `:` makes me nervous... IMO it's better to add it to each function signature individually.

JackStouffer: I agree; besides the safety issue, I have run into over applied attribute issues in other PRs.

WalterBright: Given that the functions it applies to are all right there and one liners, this isn't an issue, and implementing it would add clutter.

src/object.d
```diff
@@ -1852,8 +1852,8 @@ extern (C)
     // size_t _aaLen(in void* p) pure nothrow @nogc;
     private void* _aaGetY(void** paa, const TypeInfo_AssociativeArray ti, in size_t valuesize, in void* pkey) pure nothrow;
     // inout(void)* _aaGetRvalueX(inout void* p, in TypeInfo keyti, in size_t valuesize, in void* pkey);
-    inout(void)[] _aaValues(inout void* p, in size_t keysize, in size_t valuesize, const TypeInfo tiValArray) pure nothrow;
-    inout(void)[] _aaKeys(inout void* p, in size_t keysize, const TypeInfo tiKeyArray) pure nothrow;
+    inout(void)[] _aaValues(inout void* p, in size_t keysize, in size_t valuesize, const TypeInfo tiValArray) pure nothrow @trusted;
```
klickverbot: This is obviously wrong, the signature (dereferencing a void*) cannot possibly be safe.

Note that this actually introduces a safety hole into user code since the declaration is not even private. Just make the calls `@trusted` with a lambda.

src/object.d
```diff
@@ -1987,15 +1987,15 @@ auto byKey(T : V[K], K, V)(T* aa) pure nothrow @nogc
...
+auto byValue(T : V[K], K, V)(T aa) pure nothrow @nogc @trusted
 {
     import core.internal.traits : substInout;
 
     static struct Result
     {
         AARange r;
 
-    pure nothrow @nogc:
+    pure nothrow @nogc @trusted:
```
klickverbot:
```d
int main() @safe {
  auto a = ["foo": 1];
  auto r = a.byValue;
  r.popFront;
  return r.front; // oops
}
```

JackStouffer: To be fair, this is also applies to `std.range.front` for normal arrays when compiling in release mode without bounds checking, and that's marked as safe.

src/object.d
aG0aep6G: > To be fair, this is also applies to std.range.front for normal arrays when compiling in release mode without bounds checking, and that's marked as safe.

But that's `-boundscheck=off`'s fault. It flat-out breaks `@safe`. An `@safe`/`@trusted` function may be unsafe with the switch, but it must be safe without it.

src/object.d
WalterBright: fixed

src/object.d
WalterBright: Made it private, which should do it.

andralex:
@klickverbot is that code sample submitted as a bug? Seems to me `opCast` is not checked, and it should.

WalterBright:
> I've certainly been able to find safety issues in dlang PRs fairly consistently – including several of yours – by just mechanically applying them and having a closer look at code that violates them.

That's great. Keep 'em coming!

WalterBright:
> disallowing AliasThis breaks perfectly valid code

I understand, but the trouble is, alias this can be used to override various behaviors (not just opCast) that the implementation relies on - i.e. the cast to void* that is a type known to the implementation could get corrupted. And even if the user supplied opCast was safe, that doesn't mean it left the AA in a state recognizable by the implementation.

The correct fix to this is to create a fully templated implementation of the AAs. We've aspired to do this, but have repeatedly failed. Anyhow, doing that is way beyond the scope of this PR. As long as the implementation has a non-templated void* interface, the only pragmatic way forward for now is to disallow alias this usage.

> make the signatures look more like this

Doesn't work because then the values will all be const.

> you can also assign aa to a V[K] type inside the implementation

That was my first try. Unfortunately, that does not work (again, const issues, the test suite failed miserably).

> we now still need to audit all of object.d for incorrect uses of the functions.

I'm not terribly concerned about that. There are just a couple uses, and there aren't unsafe template arguments mucking it up. I don't even like it being in object.d at all, it should be encapsulated elsewhere.

> If you just used @trusted the way it is supposed to be – as a local escape for non-checkable code –, we wouldn't have this issue in the first place, at no additional cost.

It wouldn't have solved the alias this problem, and it would have looked awful because it is pretty much every line of code. The buildin AAs do not follow the usual typing rules in D, this is a known issue, and can only be solved by redoing the whole thing with templates. Until then I think we're stuck with this admittedly hackish approach - but at least it solves the particular bug report problem.

WalterBright:
> How about this as a starting point?

Won't work. The trouble is that once data is cast to void*, ALL type information is lost, including any alias this information. Making it a trusted cast won't help. The purpose of alias this is to override certain behaviors, and even the layout, of the underlying type. This will never work safely with an implementation that assumes exact layout and behavior coming from a void*.

WalterBright:
> still has a safety issue:

Interesting. Will work on it.

WalterBright:
Hole fixed. Next? :-)

andralex:
@klickverbot will leave approval of this to you following consensus with @WalterBright. Let me add we're very appreciative of your work!

WalterBright:
> Note that in my example the parameter is typed as an AA, so what you are pointing out does not apply. The AliasThis is evaluated at the function call site, which also side-steps @safety issues.

It does apply, otherwise the cast override in your first example would not have happened. Secondly, even if did not apply, it still is a hole. Let me reiterate that the purpose of wrapping is to override behaviors. Overriding behaviors means the `void*` interface can no longer rely on having had complete control over the data structure. For example, a field could be added before the aa, and then boom, the supposedly trustworthy `void*` cast points to that field rather than the aa.

> -release is not supposed to disable bounds-checking in @safe code, so you might want to use version (D_NoBoundsChecks) instead of asserts. It's probably a bit of a murky area, though.

The bounds checking is for arrays, not associative arrays. Even so, to add this requirement implies reconsidering the pervasive use of asserts in ranges, starting with:

https://github.com/dlang/phobos/blob/master/std/range/primitives.d#L2267

which is a whole 'nother discussion and way outside the scope of this PR.

WalterBright:
At this point, it would seem more efficient if you wrote and submitted another PR rather than trying to fold in all those diffs, as you've essentially rewritten it completely anyway.

WalterBright:
If you don't want to make a PR with your file, how about emailing it to me? That would be a lot easier than trying to go line by line with the above diff, and far less error prone.

MartinNowak:
Thanks for the review @klickverbot.
The PR smells a bit, but it would also be nice if we can make those methods safe before the long library AA transition.

> The correct fix to this is to create a fully templated implementation of the AAs. We've aspired to do this, but have repeatedly failed.

Just to correct your perception, we __had__ failed twice, a promising third attempt was mostly stuck by lack of feedback, see #1282. Though you didn't seem to fully understand the topic, I'm taking your affirmation https://github.com/dlang/druntime/pull/1282#issuecomment-234758959 as a reason to put some more resources into this.
Hope I can get more useful information out of this thread http://forum.dlang.org/post/anhtpsftcedurgskfsgx@forum.dlang.org, otherwise would have to go through a DIP.

WalterBright:
@klickverbot I have posted your version, adding in the test cases from the bug report.

WalterBright:
Your version fails the test I added in:
---
unittest
{
    static struct Foo {
        int[int] aa;
        auto opCast() pure nothrow @nogc { *cast(uint*)0xdeadbeef = 0xcafebabe; return null; }
        alias aa this;
    }

    static void test() @safe {
        Foo f;
        static assert(!__traits(compiles, { int i = !f.byKey.empty; }));
    }
}
---
which is your test case :-( posted above.

WalterBright:
Fails with:
```
runnable/interpret.d(1975): Error: template object.keys cannot deduce function from argument types !()(immutable(bool[int])), candidates are:
../../druntime/import/object.d(2058):        object.keys(Value, Key)(Value[Key] aa)
../../druntime/import/object.d(2068):        object.keys(T : Value[Key], Value, Key)(T* aa)
runnable/interpret.d(1979):        called from here: foo101()
```

WalterBright:
I changed it back to the version I originally wrote because it passes the test suite. It does have the issue that it cannot be wrapped, but I don't see a way around it at the moment. @klickverbot 's version could be wrapped, but didn't pass the tests, and the reason for that is the same reason the templates are written a little oddly in the original.

somzzz:
@WalterBright @andralex is this PR still required?
```byKey```, ```byValue``` and ```byKeyValue``` were made @safe in https://github.com/dlang/druntime/pull/1944

RazvanN7:
byKey and byValue are @safe now and keys/values are being made @safe in this PR: https://github.com/dlang/druntime/pull/3528 . So I'm closing this, but feel free to give feedback to https://github.com/dlang/druntime/pull/3528

PR #1646, opened 2016-09-06
Move deprecations along.

jmdavis:


jmdavis:
> Trusting the deprecation czar that he got the timing right. :)

LOL. Well, the deleted code _does_ specifically say August 2016, so it's not hard to see what the correct timing was, and if anything, it's slightly late (though since we haven't had a full release in that timeframe, it's not like doing this PR a month earlier would have changed much), but it's not like being a bit late hurts much. It's being early that's a problem.

In any case, thanks for pulling it - though it looks like an unrelated error on OS X is blocking it for the moment.

PR #1647, opened 2016-09-09
change AA to avoid a calling the equals function

UplinkCoder:
... since it causes major data-cache pressure.
And op-equals is often a pretty costly function!


UplinkCoder:
This change will effect template instanciation positively.
And it does make AA's a viable alternative to StringTable.d

UplinkCoder:
Aww Shoot. We cannot avoid calling equals ....
Because the hash could collide by accident.

PR #1652, opened 2016-09-11
rt.util.container: Use outOfMemoryErrorNoGC

klickverbot:
Array and HashTab are used before the GC is even initialised.


ZombineDev:
LGTM. The CircleCI failure is unrelated; looks like it's due to bad configuration.

PR #1654, opened 2016-09-11
ctfe-code is often annotated with pure @safe we want to print anyway

UplinkCoder:
As per title.
This signature change is needed to make the test work.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3186,12 +3186,8 @@ bool _xopCmp(in void*, in void*)
 {
     throw new Error("TypeInfo.compare is not implemented");
 }
+
 void __ctfeWrite(const string s) @nogc @safe pure nothrow {}
-void __ctfeWriteln(const string s) @nogc @safe pure nothrow
-{
-  __ctfeWrite(s);
-  __ctfeWrite("\n");
-}
 
 /******************************************
  * Create RTInfo for type T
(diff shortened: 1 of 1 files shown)


UplinkCoder:
@klickverbot this is needed to get the test to run.
I forgot to add the attributes in the first PR.

UplinkCoder:
It is a special debug tool. 
We will always special case.

UplinkCoder:
I disagree with @safe.

ZombineDev:
I agree with @klickverbot about `@safe`. Those functions don't do anything unsafe that needs to be trusted.

While we're at that, can we change the signature to `const(string)[]`? C.f. my rant about immutable in the [original PR](https://github.com/dlang/druntime/pull/1643#issuecomment-246271082).

UplinkCoder:
@klickverbot it's @safe now.

WalterBright:
Do we really need the "ln" variant?

UplinkCoder:
@WalterBright not necessarily however since it only takes one string argument the ln variant takes quite some typing.

UplinkCoder:
Furthermore the ln Variant was there before.
I don't see why we should remove it.

UplinkCoder:
@klickverbot @WalterBright merge or close ?

MartinNowak:
The signature can also be used w/ formattedWrite, formattedWrite((in char[] c) => __ctfeWrite(c.idup), "");.

MartinNowak:
Auto-merge toggled on

PR #1655, opened 2016-09-11
sections_elf_shared: Do not access TLS of dead thread in finiTLSRanges()

klickverbot:
finiTLSRanges() is called from the destructor of core.Thread. At this point,
the OS thread has already ceased to exist, so what was formerly a pointer
to _loadedDSOs is no longer valid.

In other words, `tdsos.reset()` was a use-after-free bug. It is unclear why
the issue didn't surface on Linux/FreeBSD yet; for example, glibc might not
actually re-use the TLS address range after a thread exits. On OS X, however,
this did quite frequently trigger a crash when running the Phobos unit tests,
since `tdsos` would have already been overwritten with unrelated contents.

@MartinNowak

### DIFF
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -149,7 +149,10 @@ version (Shared)
 
     void finiTLSRanges(Array!(ThreadDSO)* tdsos)
     {
-        tdsos.reset();
+        // Nothing to do here. tdsos used to point to the _loadedDSOs instance
+        // in the dying thread's TLS segment and as such is not valid anymore.
+        // The memory for the array contents was already reclaimed in
+        // cleanupLoadedLibraries().
     }
 
     void scanTLSRanges(Array!(ThreadDSO)* tdsos, scope ScanDG dg) nothrow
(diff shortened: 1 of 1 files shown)


WalterBright:
Fixing random crashes is always good!

WalterBright:
Auto-merge toggled on

MartinNowak:
Your analysis is correct, finiTLSRanges gets called from the Thread finalizer (`~this()`) through rt.tlsgc.destroy, i.e. after the thread might have exited.
The fix is a bit wonky though. The underlying problem is that we use a TLS array and store a reference to that outside of the Thread (so that it can get GCed). What we should do instead is storing an array outside of the Thread and a reference to that in TLS, then the design in rt.tlsgc should work.

pjotrp:
Turns out finiTLSRanges() does crash randomly in sambamba on Linux. Both in the original version and the later fixed version of druntime as is discussed in https://github.com/lomereiter/sambamba/issues/219. 

When I removed the line in the static version of the library with this patch https://github.com/genenetwork/guix-bioinformatics/blob/master/ldc-druntime-finiTLSRanges.patch the segfault no longer happens. We are still running tests, but so far it looks good.

PR #1656, opened 2016-09-21
Add example for Runtime.moduleUnitTester

Abscissa:


Abscissa:
The example (at least the body of customModuleUnitTester) was ripped straight from the default implemention in Druntime itself. Not sure exactly what needs changed.

jmdavis:
I assuming that he's talking about the spacing on the loop, if statements, and catch. e.g. the foreach should be

    foreach (m; ModuleInfo)

The style guide now requires that the keyword in those cases be followed by a space, and while the guide does not actually indicate the spacing inside the parens, almost none of Phobos has spaces around what's inside the parens.

But there's definitely code floating around still that doesn't quite follow the current style guide, which is presumably why the code that you pulled doesn't match the current style guide. I think that someone went through most of Phobos and updated after the bit about the keywords in if statements and loops and whatnot needing a space and made that change using dfix, but I'm not sure that anyone did that with druntime.

MartinNowak:
Yes, the main point here is to be consistent in user documentation.

MartinNowak:
Auto-merge toggled on

PR #1658, opened 2016-09-23
fix building non-PR builds

MartinNowak:
- i.e. fix checking for empty/unbound CIRCLECI_PR_NUMBER

### DIFF
--- a/circleci.sh
+++ b/circleci.sh
@@ -51,7 +51,7 @@ clone() {
 }
 
 coverage() {
-    if [ -n ${CIRCLE_PR_NUMBER:-} ]; then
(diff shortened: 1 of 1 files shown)


MartinNowak:
Auto-merge toggled on

MartinNowak:
> I could have sworn I submitted a similar patch some while ago. Probably it was on another repo, and the three different copies got out of sync.

You mean https://github.com/dlang/dmd/pull/6066, it's a pity that we still don't have a common place for sharing makefiles/scripts between our repos.
I'm still a proponent of a top level dlang repo referencing the others as submodules, maybe there is a way we can move @CyberShadow's automated [bitbucket repo](https://bitbucket.org/cybershadow/d/src) to githbub w/o the commit referencing issue?

PR #1660, opened 2016-09-26
[TRIVIAL] Remove use of implicit string concatenation

mathias-lang-sociomantic:
This feature will hopefully be deprecated soon.


schveiguy:
Auto-merge toggled on

PR #1663, opened 2016-09-27
remove EH table support - no longer used

WalterBright:
Not needed since druntime switched to Dwarf exception handling tables.


WalterBright:
Will do. I was just waiting to see if anything would fail, first :-)

WalterBright:
After this is pulled, I'll remove it from the compiler as well.

WalterBright:
Ready to pull.

WalterBright:
Companion dmd pull: https://github.com/dlang/dmd/pull/6157

PR #1664, opened 2016-09-29
Fix range violation in multithreaded application compiled with -profile=gc option

japplegame:
Possible related issue https://issues.dlang.org/show_bug.cgi?id=14511

### DIFF
--- a/src/rt/profilegc.d
+++ b/src/rt/profilegc.d
@@ -94,8 +94,13 @@ static ~this()
                 // Merge
                 foreach (name, entry; newCounts)
                 {
-                    globalNewCounts[name].count += entry.count;
-                    globalNewCounts[name].size += entry.size;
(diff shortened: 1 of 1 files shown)


JackStouffer:
Please follow the [D style guide](https://dlang.org/dstyle.html)

UplinkCoder:
This will not fix anything in the long run.
The range violation happens because of race-conditions.
You would have to use a Mutex.

japplegame:
No. The range violation happens because of wrong merging several TLS arrays into one. There is no race conditions and no need of additional mutexes, because the code is already synchronized, see [line 90](https://github.com/japplegame/druntime/blob/a6f139313b612223c2b035bd22abd9a7cd4503d1/src/rt/profilegc.d#L90)

japplegame:
Ping

PR #1665, opened 2016-10-04
fix Issue 16594 - static dtors called twice w/ exception

MartinNowak:
- must call finiSections when an exception is thrown in a destructor
- marks runtime as shut down, so final unloading
  of executable doesn't try to run module dtors again

### DIFF
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -207,7 +207,6 @@ extern (C) int rt_term()
         thread_joinAll();
         rt_moduleDtor();
         gc_term();
-        finiSections();
         return 1;
     }
     catch (Throwable t)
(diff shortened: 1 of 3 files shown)


MartinNowak:
Coverage info is written with the module dtor, hence we don't see any lines covered after running that.

schveiguy:
LGTM. What about gc_term and others?

MartinNowak:
> LGTM. What about gc_term and others?

You can't cleanly shutdown after an Error was thrown, running finalizers could easily deadlock, and freeing the GC memory is not important. The finiSections just does a few lowlevel C-ish things and sets the _runtimeIsInitialized flag to false.

schveiguy:
Alrighty then

schveiguy:
Auto-merge toggled on

PR #1666, opened 2016-10-05
Added some missing UUID's

TurkeyMan:
I couldn't link because these were missing.
There are lots more missing, but it'll be a HUGE task rounding them all up.


MartinNowak:
Why don't we have those in dinput, but in uuid?
https://sourceforge.net/p/mingw-w64/mingw-w64/ci/38d327fd43032c672f08591bb944ce3b60dde45d/tree/mingw-w64-headers/direct-x/include/dinput.h#l53

PR #1668, opened 2016-10-05
Allow configuration of cycle detection via command line parameters.

schveiguy:
Changes in this:
1. Allow ignoring/just printing of cycles
2. Remove all GC usage except when throwing an exception.
3. Because edges are needed every time we print a cycle, I'm mallocing them at the beginning, and only use the hashtable there (findModule goes away).

~~NOTE: I have NOT tested this properly, because I cannot build dmd on MacOS, see https://issues.dlang.org/show_bug.cgi?id=16536. I'd like to add tests for the switches, but I'm not sure how to do this, especially since I cannot build. I'm throwing this in the mix to at least test the default functionality.~~

I will add a changelog once this is more solid.

Update: Was able to download binary of MacOS dmd and run unit tests.

### DIFF
--- a/src/rt/minfo.d
+++ b/src/rt/minfo.d
@@ -360,7 +360,7 @@ struct ModuleGroup
                                     // print the message
                                     buildCycleMessage(idx, midx, (string x) {
                                                       import core.stdc.stdio : fprintf, stderr;
-                                                      fprintf(stderr, "%.*s", x.length, x.ptr);
+                                                      fprintf(stderr, "%.*s", cast(int) x.length, x.ptr);
                                                       });
                                     // continue on as if this is correct.
                                     break;
--- a/src/rt/minfo.d
+++ b/src/rt/minfo.d
@@ -250,13 +250,14 @@ struct ModuleGroup
                 // preallocate enough space to store all the indexes
                 int *edge = cast(int*)malloc(int.sizeof * _modules.length);
                 size_t nEdges = 0;
-                foreach (e; m.importedModules)
+                foreach (imp; m.importedModules)
                 {
-                    if(auto impidx = e in modIndexes)
+                    if (imp is m) // self-import
+                        continue;
+                    if (auto impidx = imp in modIndexes)
                     {
-                        if (*impidx != i)
-                            if (!bts(reachable, *impidx))
-                                edge[nEdges++] = *impidx;
+                        if (!bts(reachable, *impidx))
+                            edge[nEdges++] = *impidx;
                     }
                 }
                 // trim space to what is needed.
--- a/src/rt/minfo.d
+++ b/src/rt/minfo.d
@@ -57,42 +57,24 @@ struct ModuleGroup
     // cycle path, and then creates it. It is a precondition that src and
     // target modules are involved in a cycle.
     //
+    // The return value is malloc'd using C, so it must be freed after use.
+    //
     // The delegate is a helper to map module info pointers to index into the modules array
-    private int[] genCyclePath(int srcidx, int targetidx,
-            scope int delegate(immutable(ModuleInfo)*) findModule)
+    private size_t[] genCyclePath(size_t srcidx, size_t targetidx, int[][] edges)
     {
         import core.bitop : bt, btc, bts;
 
-        // set up all the arrays. Use the GC, we are going to exit anyway.
-        int[] cyclePath;
-        int[] distance;
-        int[][] edges;
-        distance.length = _modules.length;
-        edges.length = _modules.length;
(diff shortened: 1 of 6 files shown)


schveiguy:
Added tests for new switches. Working on changelog...

schveiguy:
Hm... changelog is already cleared. @MartinNowak, can you set up the changelog properly for this? I'll work on updating the docs.

schveiguy:
OK, I can set up changelog when I add docs. So I can do all this. This PR is probably OK to merge, I just need to do docs/changelog.

schveiguy:
Changelog and docs done.

src/rt/minfo.d
```diff
@@ -283,7 +349,23 @@ struct ModuleGroup
...
+                                    string errmsg = "";
+                                    buildCycleMessage(idx, midx, (string x) {errmsg ~= x;});
+                                    throw new Error(errmsg, __FILE__, __LINE__);
+                                case ignore:
+                                    break;
+                                case print:
+                                    // print the message
+                                    buildCycleMessage(idx, midx, (string x) {
+                                                      import core.stdc.stdio : fprintf, stderr;
+                                                      fprintf(stderr, "%.*s", x.length, x.ptr);
```
MartinNowak: `cast(int)x.length`

src/rt/minfo.d
```diff
@@ -235,12 +231,82 @@ struct ModuleGroup
             memset(flags, 0, flagbytes);
         }
 
+
+        // build the edges between each module. We may need this for printing,
+        // and also allows avoiding keeping a hash around for module lookups.
+        int[][] edges = (cast(int[]*)malloc((int[]).sizeof * _modules.length))[0 .. _modules.length];
```
MartinNowak: Is that really such a good idea?
Space for all edges could be quite big, right?
And we almost never need to print cycles.

MartinNowak: Will test against phobos and a vibe-d project.

src/rt/minfo.d
```diff
@@ -235,12 +231,82 @@ struct ModuleGroup
...
+
+            foreach (i, m; _modules)
+            {
+                // use bit array to prevent duplicates
+                // https://issues.dlang.org/show_bug.cgi?id=16208
+                clearFlags(reachable);
+                // preallocate enough space to store all the indexes
+                int *edge = cast(int*)malloc(int.sizeof * _modules.length);
+                size_t nEdges = 0;
+                foreach (e; m.importedModules)
```
MartinNowak: I think `imp` would be easier to understand.

src/rt/minfo.d
```diff
@@ -235,12 +231,82 @@ struct ModuleGroup
...
+                {
+                    if(auto impidx = e in modIndexes)
+                    {
+                        if (*impidx != i)
```
MartinNowak: You could continue before the hash lookup.
`if (e is m) continue; // self-import`

test/cycles/Makefile
```diff
@@ -0,0 +1,26 @@
...
+
+.PHONY: all clean
+all: $(addprefix $(ROOT)/,$(addsuffix .done,$(TESTS)))
+
+$(ROOT)/cycle_ignore.done: RETCODE=0
+$(ROOT)/cycle_ignore.done: LINES=0
+$(ROOT)/cycle_abort.done: RETCODE=1
+$(ROOT)/cycle_abort.done: LINES=5
+$(ROOT)/cycle_print.done: RETCODE=0
+$(ROOT)/cycle_print.done: LINES=8
```
MartinNowak: Nice

MartinNowak:
> Hm... changelog is already cleared. @MartinNowak, can you set up the changelog properly for this? I'll work on updating the docs.

Yes, shouldn't have done that. Need to improve the changed tool to automatically gather the project changelog files.

src/rt/minfo.d
MartinNowak: 11KB phobos, 7KB dub-registry (uses pretty much all of vibe.d).
Time for sorting phobos went from 1000µs -> 500µs.

MartinNowak:
Auto-merge toggled on

src/rt/minfo.d
```diff
@@ -360,7 +360,7 @@ struct ModuleGroup
                                     // print the message
                                     buildCycleMessage(idx, midx, (string x) {
                                                       import core.stdc.stdio : fprintf, stderr;
-                                                      fprintf(stderr, "%.*s", x.length, x.ptr);
+                                                      fprintf(stderr, "%.*s", cast(int) x.length, x.ptr);
```
schveiguy: thanks!

PR #1669, opened 2016-10-09
add core.simd.prefetch declaration

WalterBright:
Necessary to support making it a compiler intrinsic.


WalterBright:
In support of https://issues.dlang.org/show_bug.cgi?id=16560

MartinNowak:
Exactly matches [GCC's __builtin_prefetch](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-g_t_005f_005fbuiltin_005fprefetch-4287) :+1:.

MartinNowak:
Auto-merge toggled on

9il:
Wow, that was fast!

Nitpicks:

1. LLVM <cache type> parameter is missing. This is not used now, but likely be used in the future. For example intel CPU has non shared L1 / L2 caches. If GDC BE has not this feature, it can just ignore it.
2. Only `__prefetch` should be presented, and it would be nice to have it as a part of the language instead of DRuntime hack. Making it a function breaks final fastmath optimization for LDC. So, only real intrinsics should be provided for user without any function shells. Underscored `__names` looks good to me. Also this will help to BetterC mode.
3. A clean API without encodings is required for `__prefetch` in context of (2).

ibuclaw:
> Exactly matches GCC's __builtin_prefetch :+1:.

Shame it is hidden behind `version(D_SIMD)` then. ;-)

> LLVM parameter is missing. This is not used now, but likely be used in the future. For example intel CPU has non shared L1 / L2 caches. If GDC BE has not this feature, it can just ignore it.

Looks like It's exposed in gcc.builtins as:
```
extern pure nothrow @nogc @safe void __builtin_prefetch(in void*, ...);
```
So you could have ten parameters for all I care. There are zero changes needed to be done on my side to support this.

WalterBright:
> Shame it is hidden behind version(D_SIMD) then. ;-)

Maybe that should be fixed.

WalterBright:
> LLVM (EDIT markdown) cache type parameter is missing. This is not used now, but likely be used in the future. For example intel CPU has non shared L1 / L2 caches. If GDC BE has not this feature, it can just ignore it.

I left it off because LDC did not have it in the documentation I found. Perhaps I found old documentation? I also could find nothing on what might trigger the generation of the `PREFETCHWT1` opcode.

> Only __prefetch should be presented, and it would be nice to have it as a part of the language instead of DRuntime hack.

Using a template enables forcing the parameters to be constants and getting easy checking of the values. My testing of it shows it does compile directly to the instruction - no function call - when inlining is turned on.

> Making it a function breaks final fastmath optimization for LDC. So, only real intrinsics should be provided for user without any function shells.

I don't understand - doesn't LDC inline it?

> Underscored __names looks good to me. Also this will help to BetterC mode.

It shouldn't impair betterC at all, since the template is instantiated in the user code, not druntime.

> A clean API without encodings is required for __prefetch in context of (2).

I don't understand this comment.

9il:
> LLVM (EDIT markdown) cache type parameter is missing. This is not used now, but likely be used in the future. For example intel CPU has non shared L1 / L2 caches. If GDC BE has not this feature, it can just ignore it.
> I left it off because LDC did not have it in the documentation I found. Perhaps I found old documentation? I also could find nothing on what might trigger the generation of the PREFETCHWT1 opcode.

From the same link you have posted on the Issue Tracking System:
```
‘llvm.prefetch‘ Intrinsic

Syntax:

declare void @llvm.prefetch(
  i8* <address>, 
  i32 <rw>,
  i32 <locality>,
  i32 <cache type>)
```
The last one parameter is what i mean.
```
Arguments:

... The cache type specifies whether the prefetch
is performed on the data (1) or instruction (0) cache. ...
```
> 
> Only __prefetch should be presented, and it would be nice to have it as a part of the language instead of DRuntime hack.
> Using a template enables forcing the parameters to be constants and getting easy checking of the values. My testing of it shows it does compile directly to the instruction - no function call - when inlining is turned on.
> 
> Making it a function breaks final fastmath optimization for LDC. So, only real intrinsics should be provided for user without any function shells.
> I don't understand - doesn't LDC inline it?
> 
> Underscored __names looks good to me. Also this will help to BetterC mode.
> It shouldn't impair betterC at all, since the template is instantiated in the user code, not druntime.
> 
> A clean API without encodings is required for __prefetch in context of (2).
> I don't understand this comment.

9il:
> Only __prefetch should be presented, and it would be nice to have it as a part of the language instead of DRuntime hack.
> Using a template enables forcing the parameters to be constants and getting easy checking of the values. My testing of it shows it does compile directly to the instruction - no function call - when inlining is turned on.
> 
> Making it a function breaks final fastmath optimization for LDC. So, only real intrinsics should be provided for user without any function shells.
> I don't understand - doesn't LDC inline it?

Yes, but code may be less optimised comparing with code which uses exact intrinsics call. Example with `fastmath` can be found [here](http://forum.dlang.org/post/ywoukacyupehzcmjnhld@forum.dlang.org). The last paragraph describes fastmath problem.

9il:
> Underscored __names looks good to me. Also this will help to BetterC mode.
> It shouldn't impair betterC at all, since the template is instantiated in the user code, not druntime.
> 
> A clean API without encodings is required for __prefetch in context of (2).
> I don't understand this comment.

The better C is not only about linking. It is also about transparency for users. If this be a part of the language, D capabilities will be more clear for users.

9il:
> A clean API without encodings is required for __prefetch in context of (2).
> I don't understand this comment.

I want to use `__prefetch`, please make its API simple. For example it may looks like `prefetch` or like LLVM variant. I mean without this encoding:
```d
        static if (writeFetch)
            __prefetch(address, 4);
        else static if (locality < 4)
            __prefetch(address, 3 - locality);
```

I am not a specialist in prefetch, but current code constraints looks less portable comparing with LLVM

src/core/simd.d
```diff
@@ -447,4 +447,36 @@ version ( D_SIMD )
...
+   *    locality = 0..3 (0 meaning least local, 3 meaning most local)
+   * Note:
+   *    The Intel mappings are:
+   *    $(TABLE
+   *    $(THEAD writeFetch, locality, Instruction)
+   *    $(TROW false, 0, prefetchnta)
+   *    $(TROW false, 1, prefetch2)
+   *    $(TROW false, 2, prefetch1)
+   *    $(TROW false, 3, prefetch0)
+   *    $(TROW false, 0, prefetchw)
```
klickverbot: Typo (`false`).

9il:
> This doesn't apply here.

This does

ibuclaw:
> Yes, but code may be less optimised comparing with code which uses exact intrinsics call.

I'm not sure I follow your reasoning here.  The compiler has full control when it comes to intrinsics.

> I am not a specialist in prefetch, but current code constraints looks less portable comparing with LLVM

I've only just noticed that `__prefetch` accepted an encoded parameter.  While I'm not surprised Walter has decided to do something a little different, I'd just ignore what the body is doing - that's DMD-specific.

From the implementer side, it's also possible to override the call to the template, so whatever you say about inlining is a non-issue.

9il:
> This doesn't apply here.
> This does

I mean if ldc will use a function shell on top of intrinsics this may break fastmath.

9il:
> From the implementer side, it's also possible to override the call to the template, so whatever you say about inlining is a non-issue.

How this can be implemented?

ibuclaw:
> How this can be implemented?

GDC has template builtins for `va_start` and `va_arg`, for instance.  How you detect these would be compiler dependent.

9il:
OK then

ibuclaw:
> How you detect these would be compiler dependent.

From a high level overview, you can check if a function is an instantiated decl.  Then do a match on it's template instance.  In this way, you mark all instantiated decls as you generate them.

PR #1671, opened 2016-10-13
add getdelim & getline

9il:
http://pubs.opengroup.org/onlinepubs/9699919799/functions/getdelim.html


schveiguy:
These are not available on Windows AFAIK.

9il:
Add appropriate version, rebased

jmdavis:
If it's not part of standard C, it doesn't belong in `core.stdc.stdio`. The man page clearly indicates that these are POSIX functions, so they belong somewhere in `core.sys.posix` - probably `core.sys.posix.stdio` given that the include file for them is `stdio.h`.

PR #1672, opened 2016-10-13
Line 593 comparing with a constant

varunranga:
Changing 'buffer.sizeof' to 16 because the result of 'buffer.sizeof' will always result to 16 bytes. Comparison with constant will be faster.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -590,7 +590,7 @@ class TypeInfo_StaticArray : TypeInfo
         ubyte[16] buffer;
         void* pbuffer;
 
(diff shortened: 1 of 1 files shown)


mathias-lang-sociomantic:
> Comparison with constant will be faster.

Do you have anything to back that claim ?

varunranga:
Won't specifying the size (16) directly be faster (not by much, but still) than accessing buffer object and finding its length?

varunranga:
Alright thanks :)

PR #1673, opened 2016-10-18
Print exception and abort if no handler is found

Burgos:
Previously, dwarf exception handler was relying on a fact that running
user code is always enclosed in the catch-all exception handler.  That
might not be the case if the user (with debugger, for example) change
value of `rt_trapExceptions` to false. To cover this scenario, instead
of issuing fatal error, we wil just print exception info and abort the
program.

### DIFF
--- a/changelog/unhandled.dd
+++ b/changelog/unhandled.dd
@@ -1,0 +1,8 @@
+Print unhandled exception's trace and abort if rt_trapExceptions is set
+
+D exception handling was assuming the top level exception handler,
+so all exceptions would be handled somewhere. In case rt_trapExceptions
+is cleared, this handler would not be enabled, and this assertion
+would fail, aborting the program, but without any information
+about the exception. This is now changed, so the exception information
+will be printed to the stderr, followed by abort.
--- a/src/rt/dwarfeh.d
+++ b/src/rt/dwarfeh.d
@@ -13,6 +13,7 @@ module rt.dwarfeh;
 
 version (Posix):
 
+import rt.dmain2: _d_print_throwable;
 import rt.unwind;
 import core.stdc.stdio;
 import core.stdc.stdlib;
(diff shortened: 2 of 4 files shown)


Burgos:
Example with `rt_trapException` cleared:

```
burgos@ubuntu-vbox:~$ dmd-dev -run test2.d
uncaught exception
dwarfeh(224) fatal error
Error: program killed by signal 6
burgos@ubuntu-vbox:~$  # with this patch
burgos@ubuntu-vbox:~$ dmd-dev -run test2.d
uncaught exception
object.Exception@test2.d(3): test exception
----------------
Error: program killed by signal 6
```

Burgos:
I'm not sure why coverage is complaining. I've added tests for this line to be hit. Could it be that coverage is ignoring this test?

wilzbach:
>I'm not sure why coverage is complaining. I've added tests for this line to be hit. Could it be that coverage is ignoring this test? 

The following command is run to check for coverage:

```
 TEST_COVERAGE="1" make -j$N -C . -f posix.mak MODEL=$MODEL unittest-debug
```
https://github.com/dlang/druntime/blob/master/circleci.sh#L81

adamdruppe:
i like it

nemanja-boric-sociomantic:
I'll do rebase (to fix Jenkins issue) and coverage enhancement later today.

Burgos:
Rebased and updated.

Only thing confuses me now is the coverage report:

https://codecov.io/gh/dlang/druntime/compare/f4095df4c07b0fefba76f5bd0b72972d901b9f86...8eefed3a9877d7029480cc82e4b5d2cf460d8b20/src/src/rt/dwarfeh.d#L225...227

You can see it reports these lines as not covered. But, they indeed are covered (see the test in the commit). Why this shows as not covered is because program writes out the coverage report at the end of the program. But in case of this test, program must abort, so it doesn't write these lines. I'm not sure if I can do anything about it?

Burgos:
Thanks @Calrama !

ZombineDev:
@schveiguy @jmdavis @MartinNowak can you please take a look? This PR should be straightforward to review.

ZombineDev:
@Burgos CodeCov says that it has not coverage data on `test/exceptions/src/rt_trap_exceptions.d` which probably explain why the lines you changed in `src/rt/dwarfeh.d` are not covered. Your changes are fine, the problem is most probably with the CodeCov setup.

nemanja-boric-sociomantic:
Thanks!

PR #1674, opened 2016-10-19
Allow rt_trapExceptions to be set from command line

Burgos:
This allows rt_trapExceptions to be set from the command line
using --DRT-trap-exceptions={false,true}.

In case this is set to false, D runtime will not install top
level handler for the exceptions, but it will allow the exception
to propagate through entire call stack and to abort the program
(which may be convinient for tracking them with the debugger).

See https://issues.dlang.org/show_bug.cgi?id=15894

### DIFF
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -13,6 +13,7 @@ module rt.dmain2;
 
 private
 {
+    import rt.config: rt_configOption;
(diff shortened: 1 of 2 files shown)


Burgos:
Closing this, as there are already similar PRs (#1538, #1811)

PR #1675, opened 2016-10-19
Add some more overflow checks inside array allocation

Burgos:
https://issues.dlang.org/show_bug.cgi?id=16470

### DIFF
--- a/src/rt/lifetime.d
+++ b/src/rt/lifetime.d
@@ -274,9 +274,17 @@ bool __setArrayAllocLength(ref BlkInfo info, size_t newlength, bool isshared, co
 
     if(info.size <= 256)
     {
-        if(newlength + SMALLPAD + typeInfoSize > info.size)
+        import core.checkedint;
+
+        bool overflow;
+        auto newlength_padded = addu(newlength,
+                                     addu(SMALLPAD, typeInfoSize, overflow),
+                                     overflow);
+
+        if(newlength_padded > info.size || overflow)
             // new size does not fit inside block
             return false;
+
         auto length = cast(ubyte *)(info.base + info.size - typeInfoSize - SMALLPAD);
         if(oldlength != ~0)
         {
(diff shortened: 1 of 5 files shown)


src/rt/lifetime.d
```diff
@@ -274,7 +274,8 @@ bool __setArrayAllocLength(ref BlkInfo info, size_t newlength, bool isshared, co
 
     if(info.size <= 256)
     {
-        if(newlength + SMALLPAD + typeInfoSize > info.size)
+        auto newlength_padded = newlength + SMALLPAD + typeInfoSize;
+        if(newlength + SMALLPAD + typeInfoSize > info.size || newlength_padded < newlength)
```
ntrel: DRY: `if (newlength_padded > info.size ||`

test/allocations/src/overflow_from_zero.d
```diff
@@ -0,0 +1,6 @@
+void main()
+{
+    void[] buffer;
+    buffer.length = 1;
+    buffer.length = size_t.max;
```
dgileadi: This file seems to have exactly the same content as `overflow_from_existing.d`. Is that on purpose?

WalterBright:
Yes, please use `core.checkedint`. This kind of thing is exactly what it is for. It's been documented by Regher that getting overflow checking right can be surprisingly tricky, and most programmers get it wrong. Hence, the motivation for `core.checkedint` where it only has to be gotten right once, instead of every time. It'll also save on the grief of anyone reviewing the ad-hoc implementations to make sure they are correct which, again, is very tricky.

Burgos:
Well, this is all nice to see on a PR :-) 

I'll check and and use the `checkedint`, if not today, then over the weekend.

test/allocations/src/overflow_from_zero.d
Burgos: Good catch, I completely missed it!

Burgos:
Updated to use `core.checkedint`, and test-case fix.

Burgos:
And commit title to trigger bugzilla integration.

Burgos:
Friendly ping :-)

WalterBright:
Auto-merge toggled on

PR #1676, opened 2016-10-19
Add more checks for monitor objects

Burgos:
https://issues.dlang.org/show_bug.cgi?id=11594


Burgos:
I might have made a mistake in the first commit's terminology (the name of the object passed to `synchronized`)

src/core/sync/mutex.d
```diff
@@ -92,7 +92,8 @@ class Mutex :
     this( Object o ) nothrow @trusted
     in
     {
-        assert( o.__monitor is null );
+        assert ( o, "Valid object instance muss be passed." );
```
ZombineDev: muss -> must

Or maybe, "Error: the argument passed to Mutex.this(Object o) is null!"

src/rt/monitor_.d
```diff
@@ -57,6 +57,11 @@ extern (C) void _d_monitordelete(Object h, bool det)
 }
 
 extern (C) void _d_monitorenter(Object h)
+in
+{
+    assert(h !is null, "Synchronized lock must not be null.");
```
ZombineDev: lock -> object

The argument of `_d_monitorenter` can be any class object. Synchronization primitives (like Mutex) are just a special case of objects, whose `__monitor` field indirectly points to themselves.

Burgos: Thanks!

src/core/sync/mutex.d
Burgos: Pardon my German :-)

Burgos:
Updated to address @ZombineDev comments.

Burgos:
Rebased to get the tests running. I've also removed one of the commits, as it was fixed in meantime by some other PR (commit: https://github.com/dlang/druntime/commit/0291197f119f2df8a896fe88352d31138c5de7fc)

src/core/sync/mutex.d
```diff
@@ -122,6 +122,8 @@ class Mutex :
         if (is(Q == Mutex) || is(Q == shared Mutex))
     in
     {
+        assert(obj !is null,
+            "Error: the argument passed to Mutex.this(Object obj) is null.");
```
Calrama: Failed asserts are always an `Error`, so I would drop the "Error: " prefix and in keeping with the preexisting error message for `obj.__monitor` go with simply "The provided object must not be null.".

nemanja-boric-sociomantic: Thanks, will update!

PR #1677, opened 2016-10-20
Add linux-specific `core.sys.linux.file` module

Burgos:
This adds core.sys.linux.file module which contains
flock function definition and values of the parameters
to pass.

### DIFF
--- a/src/core/sys/linux/sys/file.d
+++ b/src/core/sys/linux/sys/file.d
@@ -1,0 +1,34 @@
+/**
+ * D header file for Linux file ops.
+ *
+ * Copyright: Copyright Nemanja Boric 2016.
+ * License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Authors:   Nemanja Boric
+ */
+module core.sys.linux.sys.file;
+
+version (linux):
+extern (C):
+nothrow:
+@nogc:
+
+/* Operations for the `flock' call. */
+/// Shared lock
+enum LOCK_SH              = 0x01;
+/// Exclusive lock
+enum LOCK_EX              = 0x02;
+/// Unlock
+enum LOCK_UN              = 0x08;
+
+/// Don't block when locking.
+/// Can be OR'd into one of the above.
+enum LOCK_NB              = 0x04;
+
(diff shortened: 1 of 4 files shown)


CyberShadow:
I believe this should be `core.sys.linux.sys.file` to follow the naming scheme for `sys/file.h`.

CyberShadow:
BTW, unless you specifically need non-advisory locks, we already have definitions for fcntl locks, also exposed via `std.stdio.File.lock()` etc.

CyberShadow:
Makefiles and file lists need to be modified when adding new modules. See all occurrences of an existing module name for an example, e.g. `git grep timerfd | grep -v ^src/`.

Burgos:
> I believe this should be core.sys.linux.sys.file to follow the naming scheme for sys/file.h.

Thanks, I got fooled by the `sys` prefix.

Updated with moved location and updated `make/COPY`, `win32.mak` and `win64.mak` files.

src/core/sys/linux/sys/file.d
```diff
@@ -0,0 +1,35 @@
...
+
+/* Operations for the `flock' call. */
+/// Shared lock
+enum LOCK_SH              = 0x01;
+/// Exclusive lock
+enum LOCK_EX              = 0x02;
+/// Unlock
+enum LOCK_UN              = 0x08;
+/// Atomic update
+enum __LOCK_ATOMIC        = 0x16;
```
CyberShadow: This one isn't in my `file.h`. Is it Debian-specific?

CyberShadow: I think this one is only implemented in the GNU Hurd kernel.

Burgos: Hmm, I got it from mine (Ubuntu's) file.h, but I don't see it on glibc's upstream file.h, so I'll remove it.

Thanks and good catch!

Burgos:
Updated with `__LOCK_ATOMIC` removed.

DmitryOlshansky:
Auto-merge toggled on

DmitryOlshansky:
Looks good apart from horrible windows makefiles... but that's separate problem.

andralex:
thx!

PR #1678, opened 2016-10-20
Ensure that the writable data segment is aligned.

ibuclaw:
~~**NOTE**: I'll close this PR once tests have ran, or if some fail, I'll add some extra debugging for comparison.~~

~~I am curious to know if DMD generates code that could suffer from a heisenberg I'm seeing with GC scanning in dynamic executables.~~

Sometimes the p_vaddr for a data segment may not start at an address that is aligned for the target (in my case, x86_64, and the start address is aligned to 4 bytes).

```
#0  0x00002accccd3ff2d in gc.gc.Gcx.mark() (this=..., pbot=0x600ce4, ptop=0x601000) at /home/runner/GDC/libphobos/libdruntime/gc/gc.d:2060
(gdb) p pbot
$1 = (void *) 0x600ce4
(gdb) p *(void**) 0x600ce4
$2 = (void *) 0x4008a000000000
(gdb) p (void**)(((size_t)pbot + 7) & ~7)
$3 = (void **) 0x600ce8
(gdb) p *(void**) 0x600ce8
$4 = (void *) 0x4008a0 <frame_dummy>
```

The difference between the start and the first aligned piece of data is just zeroes for padding.

```
(gdb) x/8b pbot
0x600ce4:       0       0       0       0       -96     8       64      0
```

The end of the data segment is aligned to 8 bytes.  But because the for loop in `Gcx.mark()` iterates at each 8 byte interval (32bit would be 4 byte). The last iteration of the loop scans the last four bytes of the segment plus some excess program data.

```
(gdb) x/8b p1
0x600ffc:       0       0       0       0       0       0       0       0
```

Sometimes, this excess is not accessible, which results in a segfault.

```
(gdb) x/8b p1
0x600ffc:       0       0       0       0       Cannot access memory at address 0x601000
(gdb) p *p1
Cannot access memory at address 0x601000
```

I can only seem to reproduce this in a very controlled test environment, any intervention and the problem goes away.  It is unreproducible outside the test environment.  Otherwise it is very consistent in happening 100% of the time.

This small test is really just a sanity check for DMD.  If unaligned, I believe it will mean that the GC will very likely find nothing in the data segment, where references to GC infact do exist.

### DIFF
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -676,7 +676,7 @@ version (Shared)
         {
             if (phdr.p_type == PT_DYNAMIC)
             {
-                auto p = cast(ElfW!"Dyn"*)(info.dlpi_addr + phdr.p_vaddr);
+                auto p = cast(ElfW!"Dyn"*)(info.dlpi_addr + (phdr.p_vaddr & ~(size_t.sizeof - 1)));
                 dyns = p[0 .. phdr.p_memsz / ElfW!"Dyn".sizeof];
                 break;
             }
@@ -745,12 +745,12 @@ void scanSegments(in ref dl_phdr_info info, DSO* pdso) nothrow @nogc
         case PT_LOAD:
             if (phdr.p_flags & PF_W) // writeable data segment
             {
-                auto beg = cast(void*)(info.dlpi_addr + phdr.p_vaddr);
+                auto beg = cast(void*)(info.dlpi_addr + (phdr.p_vaddr & ~(size_t.sizeof - 1)));
                 pdso._gcRanges.insertBack(beg[0 .. phdr.p_memsz]);
             }
             version (Shared) if (phdr.p_flags & PF_X) // code segment
             {
-                auto beg = cast(void*)(info.dlpi_addr + phdr.p_vaddr);
+                auto beg = cast(void*)(info.dlpi_addr + (phdr.p_vaddr & ~(size_t.sizeof - 1)));
                 pdso._codeSegments.insertBack(beg[0 .. phdr.p_memsz]);
             }
             break;


ibuclaw:
FYI @MartinNowak - maybe you have some insight into this. :-)

ibuclaw:
Great, looks like no failures then.

ibuclaw:
https://github.com/D-Programming-GDC/GDC/pull/250

Yep, that works.  But should it?

@MartinNowak, I'll get hold of you tonight. :-)

rainers:
Not sure if it is related, but this reminded me of alignment problems with TLS in LDC: https://github.com/ldc-developers/druntime/commit/740c1bb86b36725cf2d671a7ac8f77de692899be

ibuclaw:
Very similar for sure. I'm just not sure whether it's the compiler or the linker that makes the start address unaligned.

rainers:
The problem was that the runtime allocated the TLS segment backwards from `FS:[0]` (top of the stack IIRC), so if the size of TLS is not a multiple of the size of a pointer, it caused bad alignment: https://github.com/ldc-developers/druntime/blob/740c1bb86b36725cf2d671a7ac8f77de692899be/src/rt/sections_elf_shared.d#L982

ibuclaw:
For the data section, it appears that it need not be aligned (link provided by @jpf91)

https://src.chromium.org/viewvc/native_client?revision=8781&view=revision

ibuclaw:
Updated commit to align all accesses of `phdr.p_vaddr`.

Ping @MartinNowak

ibuclaw:
Ping?

ibuclaw:
Actually wait, we found a problem when updating druntime to a snapshot prior to 2.069 release.

src/rt/sections_elf_shared.d
```diff
@@ -711,12 +711,12 @@ void scanSegments(in ref dl_phdr_info info, DSO* pdso)
...
             {
-                auto beg = cast(void*)(info.dlpi_addr + phdr.p_vaddr);
+                auto beg = cast(void*)(info.dlpi_addr + (phdr.p_vaddr & ~(size_t.sizeof - 1)));
                 pdso._gcRanges.insertBack(beg[0 .. phdr.p_memsz]);
             }
             version (Shared) if (phdr.p_flags & PF_X) // code segment
             {
-                auto beg = cast(void*)(info.dlpi_addr + phdr.p_vaddr);
+                auto beg = cast(void*)(info.dlpi_addr + (phdr.p_vaddr & ~(size_t.sizeof - 1)));
                 pdso._codeSegments.insertBack(beg[0 .. phdr.p_memsz]);
```
jpf91: I'm not sure if this is always 100% correct: `phdr.p_vaddr & ~(size_t.sizeof - 1)` rounds down the address to the next smaller aligned address. If this happens `beg[0 .. phdr.p_memsz]` will in theory not include the complete section and miss up to `size_t.sizeof - 1` bytes at the end. Being pedantic this needs to be `beg[0 .. phdr.p_memsz + (phdr.p_vaddr & (size_t.sizeof - 1))]`.

Example: p_vaddr = 6, p_memsz = 4, size_t.sizeof = 4 correct range: [6..10] after rounding: [4..8], correct and aligned range : [4..10]

ibuclaw: I don't think that's quite correct.
- Only the address space between `[dlpi_addr .. p_vaddr+p_memsz]` is accessible.  Attempting to read outside this results in a segfault (at least, in our hardened CI environment)
- The end address should be on a pointer aligned boundary, and if it isn't, we don't want to read it - else we'll end up segfaulting in the GC.

In your example, your "correct" case would do the following:
1. Set `codeSegments` length to 6
2. GC will scan the first pointer `[0 .. 4]`, move to the next pointer position.
3. GC will scan the second pointer `[4 .. 8]` - oops, out of bounds.

jpf91: > Only the address space between [dlpi_addr .. p_vaddr+p_memsz] is accessible. Attempting to read outside this results in a segfault (at least, in our hardened CI environment)

Agreed. However, the fix to align the `_codeSegments[0]` address rounds down. So theoretically this could already cause a out of bounds read. In reality the real start of the section / end of a protected memory segment is probably aligned, so rounding down should be safe. This increases the size of the memory segment though, and increasing the length by the amount of these extra bytes will only cause the  `_codeSegments[$-1]` address to match the real end of the section. So this shouldn't be out of bounds.

> The end address should be on a pointer aligned boundary, and if it isn't, we don't want to read it - else we'll end up segfaulting in the GC.

So I guess the GC simply casts the `_codeSegments` to `size_t[]` as indicated by your example? If this is the case we should indeed not include the trailing bytes. And as we always miss less than `size_t.sizeof` it can't contain a pointer either. So your code should be fine indeed.

ibuclaw: > Agreed. However, the fix to align the _codeSegments[0] address rounds down. So theoretically this could already cause a out of bounds read.

Shouldn't be, as it's still within range of `[dlpi_addr .. p_vaddr]` unless `dlpi_addr` is unaligned, and `p_vaddr` is 0.  Which I don't think it can ever be if `(p_vaddr & p_align) == (p_offset & p_align)` is to be satisfied.

> So I guess the GC simply casts the _codeSegments to size_t[] as indicated by your example?

This is simple pointer arithmetic, you increment a pointer, its address with be incremented by `size_t.sizeof`.
```
void *p = 0x80;
p++; // p == 0x84 or 0x88
p++; // p == 0x88 or 0x90
```
And as all pointers in the data section should be guaranteed aligned to `size_t.sizeof`, you should make sure that the start address is aligned on the same boundary, otherwise the GC will be scanning for pointers at an unaligned offset.  My assumption is that in this case, runtime will find no pointers to GC memory.

jpf91: OK, makes sense now. Thanks for explaining 👍

ibuclaw:
Pong.

src/rt/sections_elf_shared.d
Calrama: > Which I don't think it can ever be if (p_vaddr & p_align) == (p_offset & p_align) is to be satisfied.

@ibuclaw Could you add an explanation for why you can draw that conclusion as a code comment?

ibuclaw: Why rounding down won't cause a segfault?

Calrama: Not all of it. Just why `(p_vaddr & p_align) == (p_offset & p_align)` implies `!(dlpi_addr is unaligned && p_vaddr == 0)`.

ibuclaw: > Not all of it. Just why (p_vaddr & p_align) == (p_offset & p_align) implies !(dlpi_addr is unaligned && p_vaddr == 0).

I've just noticed that should be modulo (`%`), not bitwise and (`&`).  This is probably a sign I wrote that snippet as late at night as I'm reading right now.  Probably a sign to stop.

In any case, `p_align` is the alignment of `dlpi_addr`, and `p_align` should always be a power of 2.

ibuclaw:
PING

n8sh:
Does this change also need to be made on line 835 in `findSegmentForAddr`?

ibuclaw:
> Does this change also need to be made on line 835 in findSegmentForAddr?

It looks ok to me as the `beg` var is not dereferenced - the pointer comparison is valid even if the pointer is unaligned.

All others in this patch are potentially problematic as if the result of `info.dlpi_addr + phdr.p_vaddr` is an unaligned address, if you don't get a segfault on the GC scan, you could instead have the GC collecting memory that still has live references, or not calling finalizers.

ibuclaw:
Maybe...  This has been the usual way that I've been aligning data for years now (https://github.com/dlang/druntime/blob/master/src/core/stdc/stdarg.d#L68).

ibuclaw:
If it wasn't clear from the original post - I was _consistently_ and _reproducibly_ getting segfaults in druntime until this patch was applied.  So there's no question about how it works when you do have an unaligned data segment.

All data between `dlpi_addr[0 .. p_vaddr + p_memsz]` would be readable by the process, so aligning down is harmless.

PR #1679, opened 2016-10-21
Add tests for GC.removeRoot

Burgos:
Calling `removeRoot` on something that was not added
by addRoot should be possible.

https://issues.dlang.org/show_bug.cgi?id=9275

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -3368,3 +3368,12 @@ unittest
     assert(thrown);
 */
 }
+
+unittest
+{
+    import core.memory;
+
+    // https://issues.dlang.org/show_bug.cgi?id=9275
+    GC.removeRoot(null);
+    GC.removeRoot(cast(void*)13);
+}


Burgos:
Fails on CI because runner can't download DMD:

```
Failed to download 'http://nightlies.dlang.org/dmd-2.068.2/LATEST'
+ source ''
./circleci.sh: line 31: : No such file or directory
```

WalterWaldron:
Nice, you've added a place for GC unit tests (they used to get put at the bottom of src/rt/lifetime.d.)

wilzbach:
> Fails on CI because runner can't download DMD:

Wow good catch!
This was introduced by https://github.com/dlang/installer/pull/200 (support for building feature branches for dmd). Seems like we should have a bit more eyes on our installation scripts.
I submitted a quick fix: https://github.com/dlang/installer/pull/201

Burgos:
Rebased and now the tests are passing. Except the code coverage test, though. However, this just adds tests, and I don't know how it could this affect the rwmonitor.d and thread.d coverage, given that it should at least increase the coverage (not that it's touching any of that).

Burgos:
Oh, that actually looks like it's showing the lines that are not covered, not delta. I guess it just has more lines with these tests, so the ratio is smaller.

rainers:
Not a bad idea to start a separate test suite for the GC. There's a problem with the test folder, though: it's never build on Windows.

Similar unittests are currently in src/gc/impl/conservative/gc.d. If you actually care about this test only, it might be better to just add it there.

Burgos:
Ah, wasn't aware of that. I'll fix it tonight, thanks!

Burgos:
@rainers I've updated the code, so that I've moved the tests just into the unittest in the suggested file. Thanks!

PR #1680, opened 2016-10-22
Make FD_* functions for select pure.

jmdavis:
This fixes it so that the FD_* functions (which are macros in C/C++) are
pure on POSIX systems. The Windows versions (in
core.sys.windows.winsock2) are already pure.

### DIFF
--- a/src/core/sys/posix/sys/select.d
+++ b/src/core/sys/posix/sys/select.d
@@ -1,16 +1,10 @@
 /**
  * D header file for POSIX.
  *
- * Copyright: Copyright Sean Kelly 2005 - 2009.
+ * Copyright: Copyright Sean Kelly 2005 - 2016.
  * License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
  * Authors:   Sean Kelly, Alex Rønne Petersen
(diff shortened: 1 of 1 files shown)


jmdavis:
> Maybe take this chance to mark them nothrow/nogc/safe as applicable as well?

They're already marked `nothrow` and `@nogc` by a label at the top, but I couldn't do the same with `pure`, because the `select` and `pselect` functions shouldn't be marked as `pure`, and we have no way to mark a function as being impure.

As for `@trusted`/`@safe`, the FD_* functions are doing pointer manipulation, using the `fd` to generate an index, and their @safety really depends on stuff like the caller passing an `fd` that would result in the pointer arithmetic giving a correct result. A bad `fd` would mean that the FD_* functions were doing something unsafe even if the pointer were fine. And I don't know what our policy is with marking functions like that as `@trusted`. Given what the functions do though, it should be clear to the caller when they can mark what they're doing as `@trusted`.

jmdavis:
> Ah, sorry, apparently I didn't expand enough context to see the attributes at the top.

That's part of why I'm not a big fan of mass applying attributes - it's too easy to not realize that they're there. It does seem like it's a popular thing to do in druntime though.

PR #1682, opened 2016-10-23
[trivial] core.time: Make private helper private

CyberShadow:
Fix conflict with other `copy` functions when user programs are built with `-unittest`.

CC @jmdavis


schveiguy:
Auto-merge toggled on

jmdavis:
Weird. Normally a function like that would have been put at the bottom of the file among the private symbols. I wonder what I was thinking at the time.

PR #1684, opened 2016-10-24
Revert "Merge pull request #1637 from WalterBright/scope-dup"

MartinNowak:
This reverts commit 35cd63567095bf865e0c7b4bb8bb92945b6244a5, reversing
changes made to aee373fb541dafce2769792b4430be82ef25bd2b.

- fixes Issue 16504 - dup can't use storage class scope for its parameter in general

Original PR #1637

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3547,3 +3547,19 @@ unittest
     S[] arr;
     auto a = arr.dup;
 }
+
+unittest
+{
+    // Bugzilla 16504
+    static struct S
+    {
+        __gshared int* gp;
+        int* p;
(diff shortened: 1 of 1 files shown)


WalterBright:
I think this should come with a test case, if only to provide documentation for why.

MartinNowak:
Done

MartinNowak:
BTW, once escape analysis is implemented, it should be possible to add an overload that checks the postblit in a template constraint.

MartinNowak:
Going to merge this for 2.072 as the required change was done.

PR #1685, opened 2016-10-27
Modify unittest handling

schveiguy:
Currently, if unit tests run and all pass, the program is started. Typically users expect the main function to exit immediately, so they instrument their main program with

``` D
version(unittest) { void main() {} }
else int main(string[] args)
{
   ...
}
```

But this sucks. D shouldn't be running the main program after unit testing (by default). This can be an opt-in for the user, but should default to exiting.

This modification changes the unit test handler to return `int` instead of `bool` to signify different handling of the unit test result. If the handler function returns 0, then the main function is run. If the handler returns nonzero, the main function is not run. If the result is `int.min`, then the program exits with a failure exit code. If the result is negative, the program prints the negated value as the number of modules that failed, and exits with a failure exit code. If the result is positive, that number is printed as the number of modules that were tested, and exits with a success exit code.

the API for setting the handler routine should be backwards compatible, as I allow setting of the handler routine as a bool return or an int return. The one place where it may be incorrect is if you fetch the handler routine and you set a bool version. This code casts it to an int-returning routine. The effect may be very negligible, as I believe both would return using a register. I did the cast mainly to allow code that checks to see which exact routine is installed will work. I'm expecting there's almost no usage of the getter property there anyway.

Possible improvements -- we may want to only enable this behavior with a `-DRT` switch. Or we may want to allow the previous behavior with such a switch. We may want to separate the bool/int handlers into a different API so the getter makes more sense if complaints arise.

I'm adding this for testing and also to allow for discussion.

### DIFF
--- a/src/core/runtime.d
+++ b/src/core/runtime.d
@@ -367,11 +367,11 @@ struct Runtime
      *
      * Example:
      * ---------
-     * version (unittest) shared static this()
+     * shared static this()
      * {
      *     import core.runtime;
      *
-     *     Runtime.extModuleUnitTester = &customModuleUnitTester;
+     *     Runtime.extendedModuleUnitTester = &customModuleUnitTester;
      * }
      *
      * UnitTestResult customModuleUnitTester()
@@ -403,15 +403,21 @@ struct Runtime
      *             }
      *         }
      *     }
-     *     if(result.executed != result.passed)
+     *     if (result.executed != result.passed)
+     *     {
+     *         result.runMain = false;  // don't run main
      *         result.summarize = true; // print failure
+     *     }
      *     else
-     *         result.runMain = true; // all UT passed
+     *     {
+     *         result.runMain = true;    // all UT passed
+     *         result.summarize = false; // be quiet about it.
+     *     }
      *     return result;
      * }
      * ---------
      */
-    static @property void extModuleUnitTester( ExtendedModuleUnitTester h )
+    static @property void extendedModuleUnitTester( ExtendedModuleUnitTester h )
     {
         sm_extModuleUnitTester = h;
     }
@@ -434,7 +440,6 @@ struct Runtime
      *  The current legacy module unit tester handler or null if none has been
      *  set.
      */
-
     static @property ModuleUnitTester moduleUnitTester()
     {
         return sm_moduleUnitTester;
@@ -450,8 +455,7 @@ struct Runtime
      *  The current  module unit tester handler or null if none has been
      *  set.
      */
-
-    static @property ExtendedModuleUnitTester extModuleUnitTester()
+    static @property ExtendedModuleUnitTester extendedModuleUnitTester()
     {
         return sm_extModuleUnitTester;
     }
--- a/src/core/runtime.d
+++ b/src/core/runtime.d
@@ -37,9 +37,58 @@ extern(C) int rt_init();
 /// C interface for Runtime.terminate, returns 1/0 instead of bool
 extern(C) int rt_term();
 
+/**
+ * This type is returned by the module unit test handler to indicate testing
+ * results.
+ */
+struct UnitTestResult
+{
+    /**
+     * Number of modules which were tested
+     */
+    size_t executed;
+
+    /**
+     * Number of modules passed the unittests
+     */
+    size_t passed;
+
+    /**
+     * Should the main function be run or not? This is ignored if any tests
+     * failed.
+     */
+    bool runMain;
+
+    /**
+     * Should we print a summary of the results?
+     */
+    bool summarize;
+
+    /**
+     * Simple check for whether execution should continue after unit tests
+     * have been run. Works with legacy code that expected a bool return.
+     *
+     * Returns:
+     *    true if execution should continue after testing is complete, false if
+     *    not.
+     */
+    bool opCast(T : bool)() const
+    {
+        return runMain && (executed == passed);
+    }
+
+    /// Simple return code that says unit tests pass, and main should be run
+    enum UnitTestResult pass = UnitTestResult(0, 0, true, false);
+    /// Simple return code that says unit tests failed.
+    enum UnitTestResult fail = UnitTestResult(1, 0, false, false);
+}
+
+/// Legacy module unit test handler
+alias bool function() ModuleUnitTester;
+/// Module unit test handler
+alias UnitTestResult function() ExtendedModuleUnitTester;
 private
 {
-    alias bool function() ModuleUnitTester;
     alias bool function(Object) CollectHandler;
     alias Throwable.TraceInfo function( void* ptr ) TraceHandler;
 
@@ -300,8 +349,21 @@ struct Runtime
      * value of this routine indicates to the runtime whether the tests ran
      * without error.
      *
+     * There are two options for handlers. The `bool` version is deprecated but
+     * will be kept for legacy support. Returning `true` from the handler is
+     * equivalent to returning `UnitTestResult.pass` from the extended version.
+     * Returning `false` from the handler is equivalent to returning
+     * `UnitTestResult.fail` from the extended version.
+     *
+     * See the documentation for `UnitTestResult` to see how you should set up
+     * the return structure.
+     *
+     * See the documentation for `runModuleUnitTests` for how the default
+     * algorithm works, or read the example below.
+     *
      * Params:
-     *  h = The new unit tester.  Set to null to use the default unit tester.
+     *  h = The new unit tester.  Set both to null to use the default unit
+     *  tester.
      *
      * Example:
      * ---------
@@ -309,17 +371,17 @@ struct Runtime
      * {
      *     import core.runtime;
      *
-     *     Runtime.moduleUnitTester = &customModuleUnitTester;
+     *     Runtime.extModuleUnitTester = &customModuleUnitTester;
      * }
      *
-     * bool customModuleUnitTester()
+     * UnitTestResult customModuleUnitTester()
      * {
      *     import std.stdio;
      *
      *     writeln("Using customModuleUnitTester");
      *
      *     // Do the same thing as the default moduleUnitTester:
-     *     size_t failed = 0;
+     *     UnitTestResult result;
      *     foreach (m; ModuleInfo)
      *     {
      *         if (m)
@@ -328,45 +390,78 @@ struct Runtime
      *
      *             if (fp)
      *             {
+     *                 ++result.executed;
      *                 try
      *                 {
      *                     fp();
+     *                     ++result.passed;
      *                 }
      *                 catch (Throwable e)
      *                 {
      *                     writeln(e);
-     *                     failed++;
      *                 }
      *             }
      *         }
      *     }
-     *     return failed == 0;
+     *     if(result.executed != result.passed)
+     *         result.summarize = true; // print failure
+     *     else
+     *         result.runMain = true; // all UT passed
+     *     return result;
      * }
      * ---------
      */
+    static @property void extModuleUnitTester( ExtendedModuleUnitTester h )
+    {
+        sm_extModuleUnitTester = h;
+    }
+
+    /// Ditto
     static @property void moduleUnitTester( ModuleUnitTester h )
     {
         sm_moduleUnitTester = h;
     }
 
-
     /**
-     * Gets the current module unit tester.
+     * Gets the current legacy module unit tester.
+     *
+     * This property should not be used, but is supported for legacy purposes.
+     *
+     * Note that if the extended unit test handler is set, this handler will
+     * be ignored.
      *
      * Returns:
-     *  The current module unit tester handler or null if none has been set.
+     *  The current legacy module unit tester handler or null if none has been
+     *  set.
      */
+
     static @property ModuleUnitTester moduleUnitTester()
     {
         return sm_moduleUnitTester;
     }
 
+    /**
+     * Gets the current module unit tester.
+     *
+     * This handler overrides any legacy module unit tester set by the
+     * moduleUnitTester property.
+     *
+     * Returns:
+     *  The current  module unit tester handler or null if none has been
+     *  set.
+     */
+
+    static @property ExtendedModuleUnitTester extModuleUnitTester()
+    {
+        return sm_extModuleUnitTester;
+    }
 
 private:
 
     // NOTE: This field will only ever be set in a static ctor and should
     //       never occur within any but the main thread, so it is safe to
     //       make it __gshared.
+    __gshared ExtendedModuleUnitTester sm_extModuleUnitTester = null;
     __gshared ModuleUnitTester sm_moduleUnitTester = null;
 }
 
@@ -440,14 +535,45 @@ extern (C) void profilegc_setlogfilename(string name);
 
 /**
  * This routine is called by the runtime to run module unit tests on startup.
- * The user-supplied unit tester will be called if one has been supplied,
+ * The user-supplied unit tester will be called if one has been set,
  * otherwise all unit tests will be run in sequence.
  *
+ * If the extended unittest handler is registered, this function returns the
+ * result from that handler directly.
+ *
+ * If a legacy boolean returning custom handler is used, `false` maps to
+ * `UnitTestResult.fail`, and `true` maps to `UnitTestResult.pass`. This was
+ * the original behavior of the unit testing system.
+ *
+ * If no unittest custom handlers are registered, the following algorithm is
+ * executed (the behavior can be affected by the `--DRT-testmode` switch
+ * below):
+ * 1. Run all unit tests, tracking tests executed and passes. For each that
+ *    fails, print the stack trace, and continue.
+ * 2. If there are no failures, set the summarize flag to false, and the
+ *    runMain flag to true.
+ * 3. If there are failures, set the summarize flag to true, and the runMain
+ *    flag to false.
+ *
+ * See the documentation for `UnitTestResult` for details on how the runtime
+ * treats the return value from this function.
+ *
+ * If the switch `--DRT-testmode` is passed to the executable, it can have
+ * one of 3 values:
+ * 1. "run-main": even if unit tests are run (and all pass), main is still run.
+ *    This is currently the default.
+ * 2. "test-or-main": any unit tests present will cause the program to
+ *    summarize the results and exit regardless of the result. This will be the
+ *    default in 2.080.
+ * 3. "test-only", the runtime will always summarize and never run main, even
+ *    if no tests are present.
+ *
+ * This command-line parameter does not affect custom unit test handlers.
+ *
  * Returns:
- *  true if execution should continue after testing is complete and false if
- *  not.  Default behavior is to return true.
+ *   A `UnitTestResult` struct indicating the result of running unit tests.
  */
-extern (C) bool runModuleUnitTests()
+extern (C) UnitTestResult runModuleUnitTests()
 {
     // backtrace
     version( CRuntime_Glibc )
@@ -494,32 +620,60 @@ extern (C) bool runModuleUnitTests()
         }
     }
 
-    if( Runtime.sm_moduleUnitTester is null )
+    if (Runtime.sm_extModuleUnitTester !is null)
+        return Runtime.sm_extModuleUnitTester();
+    else if (Runtime.sm_moduleUnitTester !is null)
+        return Runtime.sm_moduleUnitTester() ? UnitTestResult.pass : UnitTestResult.fail;
+    UnitTestResult results;
+    foreach( m; ModuleInfo )
     {
-        size_t failed = 0;
-        foreach( m; ModuleInfo )
+        if( m )
         {
-            if( m )
-            {
-                auto fp = m.unitTest;
+            auto fp = m.unitTest;
 
-                if( fp )
+            if( fp )
+            {
+                ++results.executed;
+                try
                 {
-                    try
-                    {
-                        fp();
-                    }
-                    catch( Throwable e )
-                    {
-                        _d_print_throwable(e);
-                        failed++;
-                    }
+                    fp();
+                    ++results.passed;
+                }
+                catch( Throwable e )
+                {
+                    _d_print_throwable(e);
                 }
             }
         }
-        return failed == 0;
     }
-    return Runtime.sm_moduleUnitTester();
+    import rt.config : rt_configOption;
+    if (results.passed != results.executed)
+    {
+        // by default, we always print a summary if there are failures.
+        results.summarize = true;
+    }
+    else switch (rt_configOption("testmode"))
+    {
+    case "":
+        // By default, run main. Switch to only doing unit tests in 2.080
+    case "run-main":
+        results.runMain = true;
+        break;
+    case "test-only":
+        // Never run main, always summarize
+        results.summarize = true;
+        break;
+    case "test-or-main":
+        // only run main if there were no tests. Only summarize if we are not
+        // running main.
+        results.runMain = (results.executed == 0);
+        results.summarize = !results.runMain;
+        break;
+    default:
+        throw new Error("Unknown --DRT-testmode option: " ~ rt_configOption("testmode"));
+    }
+
+    return results;
 }
 
 
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -40,6 +40,16 @@ version (NetBSD)
     import core.stdc.fenv;
 }
 
+// not sure why we can't define this in one place, but this is to keep this
+// module from importing core.runtime.
+struct UnitTestResult
+{
+    size_t executed;
+    size_t passed;
+    bool runMain;
+    bool summarize;
+}
+
 extern (C) void _d_monitor_staticctor();
 extern (C) void _d_monitor_staticdtor();
 extern (C) void _d_critical_init();
@@ -52,7 +62,7 @@ extern (C) void rt_moduleTlsCtor();
 extern (C) void rt_moduleDtor();
 extern (C) void rt_moduleTlsDtor();
 extern (C) void thread_joinAll();
-extern (C) bool runModuleUnitTests();
+extern (C) UnitTestResult runModuleUnitTests();
 extern (C) void _d_initMonoTime();
 
 version (OSX)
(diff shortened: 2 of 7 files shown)


schveiguy:
ping @wilzbach circleci seems very sensitive to line number changes. Can we fix this? I don't like the idea of having the build break because you added some lines of code!

jacob-carlborg:
Not sure I fully follow when the different int values will be used, but can't you use an enum?

schveiguy:
> Not sure I fully follow when the different int values will be used

So let's say 5 modules are tested. In the default routine, if all 5 unit test routines pass, then the function returns 5. Then the runtime sees that some unit tests were run, prints "5 Module unit tests passed", and exits.

If 2 of the module unit tests fail, stack traces are printed, then the routine returns -2. The runtime then prints "2 Module unit tests failed", and exits.

I could change to enum, but then I'll just print "All unit tests passed". The extra info of how many modules passed may be trivia, but it may also be useful. I'm open to being convinced 😉

schveiguy:
Fixed issue with trace expectation. This will probably fix circleci tests too.

jacob-carlborg:
I see. Another thing, how does this work with custom unit test runners? For example, a unit test runner that knows how many actual unit tests were run instead of the number of tested modules. That unit test runner would most likely not want print the number of modules tested.

schveiguy:
The customer runner can return 0 or int.min, and nothing will be printed. However, you may want the runtime not to run main, but still return 0 without printing anything. That case isn't handled. Perhaps int.max should do that.

schveiguy:
> Perhaps int.max should do that

Now does that. Code looks more even also.

schveiguy:
BTW, part of the WIP, I need to add testing for these cases.

jacob-carlborg:
The `int.max` addition should be documented as well.

schveiguy:
@jacob-carlborg updated docs.

src/core/runtime.d
```diff
@@ -300,6 +301,30 @@ struct Runtime
      * value of this routine indicates to the runtime whether the tests ran
      * without error.
      *
+     * There are two options for handlers. The `bool` version is deprecated but
```
jacob-carlborg: Shouldn't this say "schedule for deprecation" unless `deprecated` is added to the deprecated overload?

schveiguy: This is going to be permanently left in place, in my opinion. It doesn't cause much grief, the translation to the new version is backwards compatible, and so we don't have to do anything more.

We may want to simply undocument it at some point.

jacob-carlborg: Ok, fair enough.

deadalnix:
I'm not convinced this is the right approach. The thing will still fail to link when no main function is provided.

IMO, it is better to do this as proposed by basil. I plan to do a DMD PR.

schveiguy:
> The thing will still fail to link when no main function is provided.

dmd -main should work just fine.

schveiguy:
@wilzbach the issue actually wasn't circleci, it was the build itself (see my later [comment](https://github.com/dlang/druntime/pull/1685#issuecomment-256743553)). I didn't realize this right away, so false alarm!

src/core/runtime.d
```diff
@@ -300,6 +301,30 @@ struct Runtime
...
+     * print a message giving the value as the number of modules that passed
+     * unit tests. `main` is not executed, and success is returned.
+     *
+     * If a negative number is returned, that number is printed as the number
+     * of modules that failed unit tests. `main` is not executed, and failure
+     * is returned.
+     *
+     * The default unit tester does not return `int.min` or `int.max`, but
+     * rather the number of tests passed or failed (or 0 if none are run).
+     *
```
andralex: OK, so for the new handler:

* 0 means all good run main
* int.min failure with no extra information
* int.max all good don't run main
* >0 these many unittests passed, none failed, don't run main
* <0 the negation of these many unittests failed, print and exit(1)

That's subtle but nothing we can't handle. How about an enum as was suggested already:

How about this:

```d
enum UnittestResult { failStop = int.min, successStop, successContinue = 0 }
```

Then everything outside these values would obey the <0 / >0 rules. For nice symmetry `successStop` could be `int.max`.

src/core/runtime.d
```diff
@@ -328,46 +354,58 @@ struct Runtime
...
+     * module unit tester is set, it will be returned, even though the type is
+     * not exactly the same.
      *
      * Returns:
      *  The current module unit tester handler or null if none has been set.
      */
     static @property ModuleUnitTester moduleUnitTester()
     {
+        if(sm_moduleUnitTester == null)
+            return cast(ModuleUnitTester)sm_boolModuleUnitTester;
```
andralex: Instead of the cast use a small lambda?

src/core/runtime.d
```diff
@@ -492,32 +538,34 @@ extern (C) bool runModuleUnitTests()
...
-                    {
-                        _d_print_throwable(e);
-                        failed++;
-                    }
+                    fp();
+                }
+                catch( Throwable e )
+                {
+                    _d_print_throwable(e);
+                    --failed;
```
andralex: cool :) prolly clearer if the name were `negativeFailed`

src/rt/dmain2.d
```diff
@@ -468,8 +468,26 @@ extern (C) int _d_run_main(int argc, char **argv, MainFunc mainFunc)
...
+        {
+            int utResult = runModuleUnitTests();
+            if (utResult == 0)
+            {
+                tryExec({ result = mainFunc(args); });
+            }
+            else if (utResult > 0)
+            {
+                if(utResult != int.max)
+                    .fprintf(.stderr, "%d Module unit tests passed\n", utResult);
```
andralex: use the actual "unittest" keyword:

```d
.fprintf(.stderr, "%d unittests passed\n", utResult);
```

src/rt/dmain2.d
```diff
@@ -468,8 +468,26 @@ extern (C) int _d_run_main(int argc, char **argv, MainFunc mainFunc)
...
+                result = EXIT_SUCCESS;
+            }
+            else
+            {
+                if (utResult != int.min)
+                    .fprintf(.stderr, "%d Module Unit tests failed\n", -utResult);
```
andralex: similar here, maybe FAILED in uppercase

schveiguy:
I will finish this up with some testing soon!

andralex:
So this would be complemented by a separate flag (e.g. '-unittest-only') in the compiler.

CyberShadow:
> So this would be complemented by a separate flag (e.g. '-unittest-only') in the compiler.

Would it be possible to make it so that DMD's `-main` and rdmd's `--main`, when combined with `-unittest`, both continue to do the right thing after the change?

andralex:
@CyberShadow I was thinking one uses either `-unittest` or -unittest-only` but not both.

src/core/runtime.d
schveiguy: I will still make the return be int, but also define an enum that can be returned for convenience.

src/core/runtime.d
schveiguy: So I'm a bit concerned about doing it that way. I don't know what code exists out there, but there is the possibility that code is checking to see if their particular unit tester is set. If I wrap in a lambda, that will always be false.

Now, code that checks against the function address of wrong type is going to fail to compile. But there will be no way to rectify the situation. At least with the way I have implemented, you can cast to the correct type to figure out if you've set the function.

Alternatively, I can just expose the "new" style unit tester via a new property (and make them mutually exclusive).

schveiguy: Also note, that there isn't any danger of improperly calling the function, since it takes no parameters. Just the return type is different, and won't crash the code that uses the wrong function type.

src/core/runtime.d
schveiguy: I will fix the code to increment instead (and return `-failed` at the end).

src/rt/dmain2.d
schveiguy: ok

src/rt/dmain2.d
schveiguy: ok

schveiguy:
Made suggested edits.
Added test harness for custom unittest handler
Added support for `--DRT-unittestmode` option:
* `runmain`: run main even on passing unittests (currently default, change in 2 versions).
* `unittestonly`: if any unittests present, do not run main. (will be default on 2.074)

schveiguy:
Note, automated coverage is not going to work here, because of the nature of how we have to test these things. The added tests should cover all the new code however.

src/core/runtime.d
```diff
@@ -492,32 +562,48 @@ extern (C) bool runModuleUnitTests()
...
     }
-    return Runtime.sm_moduleUnitTester();
+    import rt.config : rt_configOption;
+    if (failed != 0)
+        return -failed;
+    else switch (rt_configOption("unittestmode"))
+    {
+    case "":
+        // By default, run main. Switch to only doing unit tests in 2.074
+    case "runmain":
```
jacob-carlborg: Could we please call this `run-main`.

src/core/runtime.d
```diff
@@ -492,32 +562,48 @@ extern (C) bool runModuleUnitTests()
...
                 }
             }
         }
-        return failed == 0;
     }
-    return Runtime.sm_moduleUnitTester();
+    import rt.config : rt_configOption;
+    if (failed != 0)
+        return -failed;
+    else switch (rt_configOption("unittestmode"))
```
jacob-carlborg: Could we please call this `unit-test-mode` or `unittest-mode`.

src/rt/dmain2.d
```diff
@@ -468,8 +468,26 @@ extern (C) int _d_run_main(int argc, char **argv, MainFunc mainFunc)
...
+        if (rt_init())
+        {
+            int utResult = runModuleUnitTests();
+            if (utResult == 0)
+            {
+                tryExec({ result = mainFunc(args); });
+            }
+            else if (utResult > 0)
+            {
+                if(utResult != int.max)
```
jacob-carlborg: * Space after `if`
* How about using the enum instead of `int.max/min`

src/core/runtime.d
schveiguy: Assuming you are backing out of that request?

src/core/runtime.d
schveiguy: I'm game for anything in terms of naming, I agree it's long but I couldn't think of something better. My original does follow precedent of all lowercase no-space or hyphen.

CC @MartinNowak thoughts?

src/core/runtime.d
jacob-carlborg: Yes. I accidentally commented on the wrong line. I left the comment because I hate when I get a notification and then I get find it on GitHub because it was removed.

src/rt/dmain2.d
schveiguy: > Space after if

ok



> How about using the enum

The enum is in the public interface, not imported here. I could re-create the enum, but I don't think it's worth doing that. I will put a comment to clarify that it is attached to the enum.

schveiguy:
Fixed formatting nit, and added comments to clarify enum usage. Also squashed into 2 commits.

src/core/runtime.d
jacob-carlborg: I don't think the name is too long, I just like to have a hyphen (or something else) to separate the words. I see three existing usages of `rt_configOption`, two with lowercase no-space or hyphen `gcopt` and `oncycle`  and one with camel case no-space or hyphen `callStructDtorsDuringGC`.

src/rt/dmain2.d
jacob-carlborg: > The enum is in the public interface, not imported here. I could re-create the enum, but I don't think it's worth doing that. I will put a comment to clarify that it is attached to the enum.

Works for me. I didn't noticed they were in separate modules.

src/core/runtime.d
schveiguy: Hm... if there is precedent, I'd prefer the camel case. I was going off [this comment](https://github.com/dlang/druntime/pull/1602#issuecomment-251509511) when I was doing the cycle detection parameters.

jacob-carlborg: I cannot recall ever seeing a program using camel case for the command line flags, most use a hyphen. I don't feel strongly about it, I just thought we should do what most other programs are doing.

test/unittest/Makefile
```diff
@@ -0,0 +1,44 @@
...
+$(ROOT)/na.done: TESTTEXT=
+
+$(ROOT)/%.done: customhandler.d
+	@echo Testing $*
+	$(QUIET)$(DMD) $(DFLAGS) -of$(ROOT)/tester_$(patsubst %.done,%, $(notdir $@)) customhandler.d -version=$(VER)
+	$(QUIET)$(TIMELIMIT)$(ROOT)/tester_$(patsubst %.done,%, $(notdir $@)) > $@ 2>&1; test $$? -eq $(RETCODE)
+	$(QUIET)test $(HASMAIN) -eq 0 || grep -q main $@
+	$(QUIET)test $(HASMAIN) -eq 1 || ! grep -q main $@
+	$(QUIET)test -z "$(TESTTEXT)" || grep -q "unittests $(TESTTEXT)" $@
+	$(QUIET)test -n "$(TESTTEXT)" || ! grep -q "unittests" $@
```
schveiguy: BTW, if some Make guru out there knows of a better way to do these grep acrobatics, I'm so very willing to fix it :)

andralex: seems par for the course :)

schveiguy:
Ping

andralex:
The return codification seems still arcane and insufficient. E.g. you can figure out how many failed but not the total run so you can't print "5/160 unittests failed". How about using some good old `ref uint` parameters that are filled with the successful and the failed unit tests instead?

cc @MartinNowak for pulling this

schveiguy:
> The return codification seems still arcane and insufficient.

I'm not trying to rule the unittest world here. The purpose of the return value change is to allow 3 different states: "pass and exit", "fail and exit", or "pass and continue". This is vs. the existing `bool` return which just allows "pass and continue" or "fail and exit". The addition of the number of tests passed or failed was just a "bonus" to give a bit of extra info since we have to return something more anyway (I also never liked the fact that when unit tests passed, no indication of passing was printed).

Full unittest suites are going to have their own mechanism of printing the results vs. the crude and limited built-in features of druntime.

andralex:
Understood. All I'm saying is let's use two integrals, one for passed and one for failed. It's reasonable and reasonably easy to implement. Thanks.

schveiguy:
OK, I will try it out and see what happens.

src/core/runtime.d
wilzbach: I agree with @jacob-carlborg here. Command line flags usually use a hyphen.

wilzbach:
> OK, I will try it out and see what happens.

Btw as no one mentioned it yet, why don't we return a simple `UnittestResult` `struct`?

schveiguy:
> Btw as no one mentioned it yet, why don't we return a simple UnittestResult struct?

May as well if we are breaking the API anyway. @andralex any objection?

andralex:
@schveiguy I approve

MartinNowak:
> wilzbach added this to the 2.074.0 milestone 9 days ago

Please don't add milestones to PRs unless it __has__ to be part of the release.

schveiguy:
I have updated the code (boy that took a while for me to get to), and now the API is as follows:

1. Instead of overloading the moduleUnitTester property, I just added an `extModuleUnitTester` property, to avoid issues with casting function pointers.
2. If the new property is set the old is ignored.
3. The new property now returns a struct, which has lots of cool goodies in it, making the behavior configurable quite easily.
4. I added an opCast!bool, so any code that currently checks `if(runModuleUnitTests)` or the opposite, then it should work.
5. I *hope* all the documentation is properly updated. I'll take a look once the doc generator runs.

schveiguy:
ping @andralex @jacob-carlborg @CyberShadow

src/core/runtime.d
```diff
@@ -37,9 +37,58 @@ extern(C) int rt_init();
...
+    /**
+     * Should the main function be run or not. Should be ignored if any tests
+     * failed.
+     */
+    bool runMain;
+
+    /**
+     * Should we print a summary of the results. Ignored if 0 tests executed.
+     */
+    bool summarize;
```
jacob-carlborg: I'm not sure what the summary actually contains but many frameworks print the number of tests run even if 0 tests were run.

src/core/runtime.d
```diff
@@ -440,14 +535,40 @@ extern (C) void profilegc_setlogfilename(string name);
...
  *
+ * If the extended unittest handler is registered, this function returns the
+ * result from that handler directly.
+ *
+ * If a legacy boolean returning custom handler is used, `false` maps to
+ * `UnitTestResult.fail`, and `true` maps to `UnitTestResult.pass`. This was
+ * the original behavior of the unit testing system.
+ *
+ * If no unittest custom handlers are registered, the following algorithm is
+ * executed (the behavior can be affected by the `--DRT-unittestmode` switch
```
jacob-carlborg: I'm not sure what the naming conventions are when it comes to `--DRT` flags but I think it should be called `--DRT-unit-test-mode` since that's what most Posix commands use.

src/core/runtime.d
```diff
@@ -440,14 +535,40 @@ extern (C) void profilegc_setlogfilename(string name);
...
+ *    runMain flag to true.
+ * 3. If there are failures, set the summarize flag to true, and the runMain
+ *    flag to false.
+ *
+ * See the documentation for `UnitTestResult` for details on how the runtime
+ * treats the return value from this function.
+ *
+ * If the --DRT-unittestmode is set to "runmain", then even if unit tests are
+ * run (and all pass), main is still run. This is currently the default. If
+ * --DRT-unittestmode is set to "unittestonly", then any unit tests present
```
jacob-carlborg: Same here, I think `unittestonly` should be `unit-test-only`.

andralex: well unittest is already considered a special word so prolly unittest-only etc

src/core/runtime.d
schveiguy: OK, so D's determination if we are in "unittest mode" is whether there exists any unit tests at all. So if there are no unit tests run, the runtime thinks that you are not testing.

But thinking about this some more, we can easily add an option for "don't ever run main, even without unit tests" as well, as we are already processing a runtime switch. I'll add it, and let you know.

BTW, the summary is done in the `dmain2` file, I added it in this PR, look for the `printf` statements

src/core/runtime.d
schveiguy: OK, will do.

schveiguy:
I added a third option for the DRT parameter. I changed the parameter to just say "test" and not "unittest", since it was very verbose otherwise. Let me know what you think. I still need to make a changelog entry.

src/core/runtime.d
```diff
@@ -328,45 +390,78 @@ struct Runtime
...
+     * Gets the current module unit tester.
+     *
+     * This handler overrides any legacy module unit tester set by the
+     * moduleUnitTester property.
+     *
+     * Returns:
+     *  The current  module unit tester handler or null if none has been
+     *  set.
+     */
+
```
jacob-carlborg: Please remove empty newline between Ddoc comment and method declaration. It's not consistent with the already existing declarations.

src/core/runtime.d
```diff
@@ -328,45 +390,78 @@ struct Runtime
...
-     *     return failed == 0;
+     *     if(result.executed != result.passed)
+     *         result.summarize = true; // print failure
+     *     else
+     *         result.runMain = true; // all UT passed
+     *     return result;
      * }
      * ---------
      */
+    static @property void extModuleUnitTester( ExtendedModuleUnitTester h )
```
jacob-carlborg: I would prefer that the name, `ext`, was not shortened. Most likely you only going to call this method once, there's no reason to make it less clear to save a few letters.

src/core/runtime.d
```diff
@@ -440,14 +535,45 @@ extern (C) void profilegc_setlogfilename(string name);
...
  *
+ * If the extended unittest handler is registered, this function returns the
+ * result from that handler directly.
+ *
+ * If a legacy boolean returning custom handler is used, `false` maps to
+ * `UnitTestResult.fail`, and `true` maps to `UnitTestResult.pass`. This was
+ * the original behavior of the unit testing system.
+ *
+ * If no unittest custom handlers are registered, the following algorithm is
+ * executed (the behavior can be affected by the `--DRT-testmode` switch
```
jacob-carlborg: I would have hoped for `--DRT-testmode` to be renamed to ``--DRT-test-mode` as well.

src/core/runtime.d
```diff
@@ -494,32 +620,60 @@ extern (C) bool runModuleUnitTests()
...
-    return Runtime.sm_moduleUnitTester();
+    import rt.config : rt_configOption;
+    if (results.passed != results.executed)
+    {
+        // by default, we always print a summary if there are failures.
+        results.summarize = true;
+    }
+    else switch (rt_configOption("testmode"))
+    {
+    case "":
```
jacob-carlborg: Shouldn't everything in this block be indented one level?

src/rt/dmain2.d
```diff
@@ -472,8 +482,27 @@ extern (C) int _d_run_main(int argc, char **argv, MainFunc mainFunc)
     //       thrown during cleanup, however, will abort the cleanup process.
     void runAll()
     {
-        if (rt_init() && runModuleUnitTests())
-            tryExec({ result = mainFunc(args); });
+        if (rt_init())
+        {
+            auto utResult = runModuleUnitTests();
+            assert(utResult.passed <= utResult.executed);
+            if(utResult.passed == utResult.executed)
```
jacob-carlborg: Space after `if` before the opening parenthesis .

src/rt/dmain2.d
```diff
@@ -472,8 +482,27 @@ extern (C) int _d_run_main(int argc, char **argv, MainFunc mainFunc)
...
+                    .fprintf(.stderr, "%d unittests passed\n", cast(int)utResult.passed);
+                if(utResult.runMain)
+                    tryExec({ result = mainFunc(args); });
+                else
+                    result = EXIT_SUCCESS;
+            }
+            else
+            {
+                if (utResult.summarize)
+                    .fprintf(.stderr, "%d/%d unittests FAILED\n", cast(int)(utResult.executed - utResult.passed), cast(int)utResult.executed);
```
jacob-carlborg: Most test frameworks will print the same summary regardless if something failed or not. That is, they usually print the total number of tests, the number of passed and the number of failed. I'm not sure how we would like it to behave.

jacob-carlborg:
> I added a third option for the DRT parameter.

👍 

> I changed the parameter to just say "test" and not "unittest", since it was very verbose otherwise. 

I like that, then we don't have to argue if it should be `unit-test` or `unittest` 😃.

jacob-carlborg:
Please squash all commits when done.

src/rt/dmain2.d
schveiguy: If nothing failed, then the number of tests run and passed is printed (a few lines above this).

This behavior is totally up for grabs if someone wants to make the effort to repaint the bikeshed. My goal is simply to stop the runtime from always executing main after running unit tests.

src/core/runtime.d
schveiguy: Hm... I generally don't indent switch case labels. I'm not sure what the style guide says. I see switch statements elsewhere that don't have case labels indented: https://github.com/dlang/druntime/blob/master/src/core/time.d#L352

src/core/runtime.d
schveiguy: So far, the DRT options are gcopt, oncycle, and it looks like some other newer camelCase ones: callStructDtorsDuringGC, scanDataSeg

So I don't know which precedent to follow. ping @MartinNowak @andralex

src/core/runtime.d
schveiguy: I wished to call it `moduleUnitTester`, but since there is an accessor, I'm reluctant to do that (the previous version of this PR did that, but the function pointer was returning a value compatible with bool).

I can change this.

src/core/runtime.d
schveiguy: I'll note none of the precedents have dashes.

src/core/runtime.d
jacob-carlborg: Yeah, I was surprised now to see how many place in Phobos that don't indent cases in switch statements. I always indent the content of a block.

src/core/runtime.d
jacob-carlborg: I try to follow what I see in most command line applications. In these case I like to look at how Git is doing things. Git contains quite a lot of subcommands and flags. There's also this, which I think is from the Posix standard.

"The arguments that consist of <hyphen-minus> characters and single letters or digits" [1].

[1] http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html

jacob-carlborg: I can't recall that I've event seen command line flags that use camel casing.

src/core/runtime.d
```diff
@@ -37,9 +37,58 @@ extern(C) int rt_init();
...
+
+    /**
+     * Simple check for whether execution should continue after unit tests
+     * have been run. Works with legacy code that expected a bool return.
+     *
+     * Returns:
+     *    true if execution should continue after testing is complete, false if
+     *    not.
+     */
+    bool opCast(T : bool)() const
```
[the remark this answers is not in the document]
schveiguy: It's not to be cute :)

It's to not break code. Existing code that may have their own runtime startup could easily be written like:

```D
if(runModuleUnitTests())
{
   dmain();
}
```

src/core/runtime.d
```diff
@@ -328,45 +390,78 @@ struct Runtime
...
-     *                     failed++;
      *                 }
      *             }
      *         }
      *     }
-     *     return failed == 0;
+     *     if(result.executed != result.passed)
+     *         result.summarize = true; // print failure
+     *     else
+     *         result.runMain = true; // all UT passed
```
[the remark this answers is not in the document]
schveiguy: Hm... I didn't set `summarize` or `runMain` if they were to be `false` (I did originally, similarly to what you had, but realized it defaults to `false` anyway). I can be more explicit.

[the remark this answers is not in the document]
schveiguy: Oops, didn't realize this was the example. The real code is more complex due to the `DRT` options. I've updated it anyway.

src/core/runtime.d
```diff
@@ -328,45 +390,78 @@ struct Runtime
...
      *                 catch (Throwable e)
      *                 {
      *                     writeln(e);
-     *                     failed++;
      *                 }
      *             }
      *         }
      *     }
-     *     return failed == 0;
+     *     if(result.executed != result.passed)
```
andralex: space after if, mmmmkay? :)

schveiguy: You could use torture on me, and I still won't ever remember to do this 😆

schveiguy:
OK, I think this is good. Please let me know if I missed any spacing issues!
I added a changelog entry, please review also.

PR #1686, opened 2016-10-30
Added private qualifier to one-off unittest helper isnan in object.d

pineapplemachine:
I noticed this when I attempted to run unittests for a module in which I defined a method called `isnan` and received a conflicts-with error. This method should definitely not be public.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2927,7 +2927,7 @@ version(unittest) unittest
 
 version (unittest)
 {
-    bool isnan(float x)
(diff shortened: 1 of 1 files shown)


yebblies:
@MartinNowak Do we need to bump the time limit, or is the freebsd32 timeout caused by something else?  That box is really slow.

yebblies:
This function could just be inlined at the one call site, but this fix works.

yebblies:
Auto-merge toggled on

PR #1687, opened 2016-11-01
Convert exception message to Windows console codepage

tramker:
This converts an exception message to the active codepage when writing to the Windows console.

- used for writes to the console window only. Output to pipe is not affected.
- works around [bug 1448](https://issues.dlang.org/show_bug.cgi?id=1448) for exception messages
- generally recommended practice of just using codepage 65001 is not enough, as it depends on a user setting outside the scope of program and there are some other bugs with UTF8 output.

I'd like to eventually propose something similar for stdio.

A question to debate is whether to convert an output to pipe too. It would make the following work correctly in any codepage:
`prog 2>&1 | find "error"`

### DIFF
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -513,39 +513,52 @@ private void formatThrowable(Throwable t, void delegate(in char[] s) nothrow sin
 extern (C) void _d_print_throwable(Throwable t)
 {
     // On Windows, a console may not be present to print the output to.
-    // Show a message box instead.
+    // Show a message box instead. If the console is present, convert to
+    // the correct encoding.
     version (Windows)
     {
+        static struct WSink
+        {
+            wchar_t* ptr; size_t len;
+
+            void sink(in char[] s) nothrow
+            {
+                if (!s.length) return;
+                int swlen = MultiByteToWideChar(
+                        CP_UTF8, 0, s.ptr, cast(int)s.length, null, 0);
+                if (!swlen) return;
+
+                auto newPtr = cast(wchar_t*)realloc(ptr,
+                        (this.len + swlen + 1) * wchar_t.sizeof);
+                if (!newPtr) return;
+                ptr = newPtr;
+                auto written = MultiByteToWideChar(
+                        CP_UTF8, 0, s.ptr, cast(int)s.length, ptr+len, swlen);
+                len += written;
+            }
+
+            wchar_t* get() { if (ptr) ptr[len] = 0; return ptr; }
+
+            void free() { .free(ptr); }
+        }
+
+        HANDLE windowsHandle(int fd)
+        {
+            version (CRuntime_Microsoft)
+                return cast(HANDLE)_get_osfhandle(fd);
+            else
+                return _fdToHandle(fd);
+        }
+
+        auto hStdErr = windowsHandle(fileno(stderr));
+        CONSOLE_SCREEN_BUFFER_INFO sbi;
+        bool isConsole = GetConsoleScreenBufferInfo(hStdErr, &sbi) != 0;
+
         // ensure the exception is shown at the beginning of the line, while also
         // checking whether stderr is a valid file
         int written = fprintf(stderr, "\n");
         if (written <= 0)
         {
-            static struct WSink
-            {
-                wchar_t* ptr; size_t len;
-
-                void sink(in char[] s) nothrow
-                {
-                    if (!s.length) return;
-                    int swlen = MultiByteToWideChar(
-                        CP_UTF8, 0, s.ptr, cast(int)s.length, null, 0);
-                    if (!swlen) return;
-
-                    auto newPtr = cast(wchar_t*)realloc(ptr,
-                            (this.len + swlen + 1) * wchar_t.sizeof);
-                    if (!newPtr) return;
-                    ptr = newPtr;
-                    auto written = MultiByteToWideChar(
-                            CP_UTF8, 0, s.ptr, cast(int)s.length, ptr+len, swlen);
-                    len += written;
-                }
-
-                wchar_t* get() { if (ptr) ptr[len] = 0; return ptr; }
-
-                void free() { .free(ptr); }
-            }
-
             WSink buf;
             formatThrowable(t, &buf.sink);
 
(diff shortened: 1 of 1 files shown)


tramker:
There is one downside of using WriteFile function, stderr can be reassigned. This will just write to STD_ERROR_HANDLE. I'll think about it.

rainers:
Is there a bug report for this?

I agree any output to the screen should use the proper encoding. The definition of CPSink looks like a lot of code duplication with WSink, though. Can't you use WriteConsoleW with the output of WSink instead?

tramker:
If the (possible) pipe conversion is not desired, than WriteConsoleW is probably better option. There isn't a bug report specific to druntime exception handler, should I create one ?

tramker:
@rainers so what do you think about the current version ?

rainers:
> @rainers so what do you think about the current version ?

Sorry, I hadn't noticed the update. Looks much better now. 

I've run it under Windows 10 in both console windows (I switched back to legacy mode to simulate earlier OS versions). With my default codepage (850) everything works ok. When switching to codepage 65001 (UTF8)
- the new console window prints bad characters. I cannot even enter UTF8 characters on the command line (e.g. öäü). It's likely a problem of the console window.
- the legacy mode console window does not print the exception at all. Umlauts are not displayed on the command line, too.

It should work with the recommended practice, but it seems switching to code page 65001 is pretty broken in Windows 10. Do you run an older version to verify?

tramker:
The problem you describe happens with raster fonts only. TT fonts should work correctly in all cases.
Raster fonts have OEM codepage characters only. So when you select an unsupported codepage, they will display incorrect or broken characters. The program output should be the same like what you type in the console window.

The only additional problem I just noticed is, that that WriteConsole version will not show anything in CP65001 + raster font combo. WriteFile version will display incorrect characters.

WriteFile version: https://github.com/tramker/druntime/commit/4fd35c7df315c5c8ecb405ab3bf2cfaca91111bf
WriteConsole version: https://github.com/tramker/druntime/commit/0aa02b91984067798230e72aa4fddbfd1944a4f3

I will try both on Windows XP, but I believe it should work as described.

tramker:
Tested on Windows XP and Windows 7 (Czech)
- both versions work correctly with TrueType font in any codepage including 65001.
- cp 65001 with raster font is positively broken. The console window closes just after typing "dir" on Win7 and Win8. Interestingly, both versions print the exception in WinXP and Win7 (with broken or blank characters).

Conclusion: UTF8 with raster font is undefined behaviour and likely broken even without this patch. Keeping this in mind, WriteConsole version is slightly better, as it prints the correct characters with TT font independently of codepage setting.

rainers:
> The problem you describe happens with raster fonts only

Yeah,I forgot about this.

> Tested on Windows XP and Windows 7 (Czech)

Thanks.

> cp 65001 with raster font is positively broken.

In the Win10 legacy console window, cp65001 seems to be broken with TT fonts, too: the exception works fine with WriteConsole, but printing `Hällö` with writeln yields an exception in stdio (when building against the MSVC libraries with -m64 or -m32mscoff).

My remaining reservation is that we might impair a common case: if the user selected CP65001, but still uses raster fonts, no exception might be displayed where he just saw a few bad characters before. Can we detect that situation and print a warning (or disable conversion)? Could even make sense to print a warning at program start.

The functionality implemented here could also be helpful in general whenever you might output a string to the console. We should consider moving it elsewhere in druntime, but let's delay it into another PR.

tramker:
> the exception works fine with WriteConsole, but printing Hällö with writeln yields an exception in stdio (when building against the MSVC libraries with -m64 or -m32mscoff).

I guess this is [dmd bug 15761](https://issues.dlang.org/show_bug.cgi?id=15761). That's why stdio needs some improvements too.

> we might impair a common case: if the user selected CP65001, but still uses raster fonts, no exception might be displayed where he just saw a few bad characters before.

I repeat: this is broken. On my Win7 and Win8, _just typing_ `dir` closes (crashes) the console window. No program needs to be run. Unfortunately, I don't think it possible to detect the font. I was experimenting with GetCurrentConsoleFont, but it seems failing.

rainers:
> I guess this is bug 15761. That's why stdio needs some improvements too. And that's why I don't recommend using CP65001 any more.

Yes. fwrite doesn't return the expected number of characters written.

> I was experimenting with GetCurrentConsoleFontEx, but it seems failing.

I thought the same, but it worked for me after setting the size of the struct before calling GetCurrentConsoleFontEx. It isn't available on XP though.

> If we want to be conservative, let's go with the WriteFile version.

According to the reactos sources, WriteFile calls WriteConsoleA for console handles anyway. It seems WriteConsoleW is the broken version.

BTW: do you have experiencce with _setMode(_O_U8TEXT)? Can this help with a general solution?

tramker:
So should I change this PR back to WriteFile ?

rainers:
> So should I change this PR back to WriteFile ?

Yeah, it's probably the safer choice. What I didn't like about the WriteFile version is the verboseness of having both  WSink and CPSink. Can't you combine these to a template with argument char/wchar?

Sorry for the indecisiveness, it's not so simple to get the best possible result when connecting several broken components ;-)

tramker:
Updated. I don't know how to combine WSink and CPSink. The API call is different. It would look messy. This may be verbose, but it's clear what is going on.

rainers:
Yeah, combining was not the best idea. CPSink is overcomplicating things, though. I'd reuse WSink and convert the result:
```
        else if (isConsole)
        {
            WSink buf;
            formatThrowable(t, &buf.sink);

            if (buf.ptr)
            {
                uint codepage = GetConsoleOutputCP();
                int slen = WideCharToMultiByte(codepage, 0,
                    buf.ptr, cast(int)buf.len, null, 0, null, null);

                auto ptr = cast(char*)malloc(slen * char.sizeof);
                if (ptr)
                {
                    WideCharToMultiByte(codepage, 0,
                        buf.ptr, cast(int)buf.len, ptr, slen, null, null);
                    WriteFile(hStdErr, ptr, slen, null, null);
                    free(ptr);
                }
                buf.free();
            }
            return;
        }
```

tramker:
OK, done.

rainers:
Auto-merge toggled on

rainers:
> OK, done.

Thanks.

PR #1688, opened 2016-11-02
free the member Object when rwmutex destroy

dushibaiyu:
ifix if use the  manual GC, the rwmutex will memory leak

### DIFF
--- a/src/core/sync/rwmutex.d
+++ b/src/core/sync/rwmutex.d
@@ -104,6 +104,19 @@ class ReadWriteMutex
         m_reader = new Reader;
         m_writer = new Writer;
     }
+    
+    ~this()
+    {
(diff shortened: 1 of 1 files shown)


WalterWaldron:
No, the GC should collect and finalize these objects automatically.

dushibaiyu:
if use the manual GC. It will memory leak.
In 2.072, the manual GC not collect automatically

schveiguy:
Not sure what you are trying to fix here, but destructors (especially of classes) cannot access GC-allocated members, let alone free them.

It's possible there is an issue, but this needs to be clearly identified first. Some example code that demonstrates the issue is needed.

In any case, manually freeing GC members in a destructor is not valid in D.

PR #1691, opened 2016-11-06
Issue 12125: do not enforce specific MS C runtime

rainers:
druntime.lib is agnostic to the used MS C runtime, but preselects one. Creating the automatic reference when building main/WinMain/DllMain is good enough.

### DIFF
--- a/src/rt/monitor_.d
+++ b/src/rt/monitor_.d
@@ -155,11 +155,6 @@ version (Windows)
     {
         pragma(lib, "snn.lib");
     }
-    else version (CRuntime_Microsoft)
-    {
-        pragma(lib, "libcmt.lib");
-        pragma(lib, "oldnames.lib");
-    }
     import core.sys.windows.windows;
 
     alias Mutex = CRITICAL_SECTION;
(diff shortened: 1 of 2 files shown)


rainers:
I guess the win64 build fails because the auto tester still tries to patch win64.mak, but fails because the same line is modified in this PR. @braddr Any plan to remove the patching and use command line options instead?

TurkeyMan:
Looks amazing to me!

ZombineDev:
Ping @MartinNowak this seems to be ready to to.

rainers:
Now that the other PRs are included in Beta 2.073.0-b1, this change should also be merged to complete https://issues.dlang.org/show_bug.cgi?id=11961 and https://issues.dlang.org/show_bug.cgi?id=12125

Does this have to be retargetted to stable now?

rainers:
Rebased to stable. The new option -mscrt (http://dlang.org/dmd-windows.html#switch-mscrt) is dysfunctional without this.

MartinNowak:
> Rebased to stable. The new option -mscrt (http://dlang.org/dmd-windows.html#switch-mscrt) is dysfunctional without this.

The new options isn't in the changelog! Please add it (dmd/stable), see https://github.com/dlang/dmd/tree/stable/changelog for the new format.

MartinNowak:
Auto-merge toggled on

rainers:
Thanks, changelog entry https://github.com/dlang/dmd/pull/6467

> make -C ../dlang.org -f posix.mak pending_changelog

This target just prints "This command will be available soon." both on stable and master, so I can't preview it locally, though.

PR #1692, opened 2016-11-13
fix issue 1761 - TypeInfo.toString for function types always indicates no-parameter function

somzzz:
https://issues.dlang.org/show_bug.cgi?id=1761

Please review, thanks!

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -784,11 +784,11 @@ class TypeInfo_Function : TypeInfo
 
 unittest
 {
-    class C
+    abstract class C
     {
-       void func() {}
-       void func(int a) {}
-       int func(int a, int b) { return 0; }
+       void func();
+       void func(int a);
+       int func(int a, int b);
     }
 
     alias functionTypes = typeof(__traits(getVirtualFunctions, C, "func"));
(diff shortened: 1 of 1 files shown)


andralex:
OK, good going. Now you may want to add a unittest covering the code you just added. That will make the codecov checks pass.

andralex:
BTW, the codecov extension for Chrome will show the uncovered lines in dark red in the diff view.

src/object.d
```diff
@@ -782,6 +782,21 @@ class TypeInfo_Function : TypeInfo
...
 }
 
+unittest
+{
+    class C
+    {
+       void func() {}
+       void func(int a) {}
+       int func(int a, int b) { return 0; }
+    }
```
andralex: I think you don't need to define these functions, just declare them - so your coverage won't be affected. Try it out.

```d
class C
{
        void func();
        void func(int a);
        int func(int a, int b);
}
```

somzzz: I made the class `C ` abstract and only declared the functions as you suggested. Is this better?

andralex:
Great, thanks. The auto-tester has failed because of a timeout, I'll merge this.

PR #1693, opened 2016-11-16
fix Issue 16558 - [Mir] Generic unaligned load/store like (like LDC l…

WalterBright:
…oadUnaligned and storeUnaligned)

### DIFF
--- a/src/core/simd.d
+++ b/src/core/simd.d
@@ -479,4 +479,63 @@ version ( D_SIMD )
   }
 
   private void __prefetch(const(void*) address, ubyte encoding);
+
+  /*************************************
+   * Load unaligned vector from address.
+   * This is a compiler intrinsic.
+   * Params:
+   *    p = pointer to vector
+   * Returns:
+   *    vector
+   */
+
+  V loadUnaligned(V)(const V* p)
+        if (is(V == void16) ||
+            is(V == byte16) ||
+            is(V == ubyte16) ||
+            is(V == short8) ||
+            is(V == ushort8) ||
+            is(V == int4) ||
+            is(V == uint4) ||
+            is(V == long2) ||
+            is(V == ulong2))
+  {
+        pragma(inline, true);
+        static if (is(V == double2))
+            return cast(V)__simd(XMM.LODUPD, *cast(const void16*)p);
+        else static if (is(V == float4))
+            return cast(V)__simd(XMM.LODUPS, *cast(const void16*)p);
+        else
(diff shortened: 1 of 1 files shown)


src/core/simd.d
```diff
@@ -479,4 +479,65 @@ version ( D_SIMD )
...
+   * Store vector to unaligned address.
+   * This is a compiler intrinsic.
+   * Params:
+   *    p = pointer to vector
+   *    value = value to store
+   * Returns:
+   *    value
+   */
+
+  V storeUnaligned(V)(const V* p, V value)
```
9il: `(const V* p, V value, const V* p)` order is more convenient, IMHO

src/core/simd.d
```diff
@@ -479,4 +479,65 @@ version ( D_SIMD )
...
+                        is(V == ubyte16) ||
+                        is(V == short8) ||
+                        is(V == ushort8) ||
+                        is(V == int4) ||
+                        is(V == uint4) ||
+                        is(V == long2) ||
+                        is(V == ulong2))
+            return cast(V)__simd(XMM.LODDQU, *cast(const void16*)p);
+        else
+            static assert(0);
```
andralex: The conditions should be moved to the template constraint. Basically `static assert(0)` in a stdlib function is an antipattern.

src/core/simd.d
andralex: How do you mean that? (p apppears twice).

WalterBright:
I don't understand. I see `p` only once.

src/core/simd.d
WalterBright: ok

src/core/simd.d
andralex: @WalterBright was asking @9il

andralex: I'm confused as to why the target is `const` although it will be obviously overwritten. I assume there must be a reason, so it would be great to state that in the documentation or at least in a comment.

WalterBright:
1. The const was a mistake. Fixed.

2. The order is there to mimic the order operands appear for assignment,
```
a = b;
storeUnaligned(&a, b);
```
It's also the same order as core.bitop.outp(), and in general it works like setters for structs,
```
s.setValue(v);
```

ZombineDev:
I also prefer the `storeUnaligned(V)(V* p, V value)` signature. It also allows the following syntax:
```D
(&a).storeUnaligned = b;
```
:smile:

9il:
Ok

PR #1694, opened 2016-11-22
Revert "Implement a software fallback algorithm for core.bitop.bsf() and bsr()"

9il:
Reverts dlang/druntime#1517

This was intrinsics because BSF/BSR assembler instruction. Another one reason to use only LDC

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -70,21 +70,14 @@ unittest
  *      The bit number of the first bit set.
  *      The return value is undefined if v is zero.
  */
-int bsf(uint v) pure
-{
-    static if (size_t.sizeof == ulong.sizeof)
-    {
-        pragma(inline, true);
-        return bsf(cast(ulong) v);
-    }
-    else
-        return softBsf!uint(v);
-}
+int bsf(size_t v) pure;
 
 /// ditto
 int bsf(ulong v) pure
 {
-    static if (size_t.sizeof == uint.sizeof)
+    static if (size_t.sizeof == ulong.sizeof)
+        return bsf(cast(size_t) v);
+    else static if (size_t.sizeof == uint.sizeof)
     {
         const sv = Split64(v);
         return (sv.lo == 0)?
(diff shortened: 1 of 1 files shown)


UplinkCoder:
Rational for reverting ?

9il:
> Rational for reverting ?

Original bsr/bsf uses hardware realisation which is faster.

9il:
Because this PR I can not use bsr/bsf from core for https://github.com/libmir/mir-random . mir-random should work without linking with DRuntime

UplinkCoder:
If it's only a fallback you should be fine with just importing the intrinsics.

9il:
> If it's only a fallback you should be fine with just importing the intrinsics.

It not intrinsics anymore.

tsbockman:
Things that are broken should be *fixed*, not just deleted. The fall-back algorithms are used by CTFE, and this PR would be a breaking change.

It's also a completely unnecessary change, as the presence of software fall-backs does not prevent functions from acting as intrinsics when hardware acceleration is available. The intrinsics were tested and verified to still work on DMD at the time of my original PR; it would not have been accepted otherwise.

If they're not working on LDC, that's because LDC handles intrinsics differently. `core.bitop` should be patched to include the necessary LDC support code, instead of just setting it all on fire.

I imagine all that is necessary is to give `bsf()` and `bsr()` the same treatment as `popcnt()`:
```D
 /**
  *  Calculates the number of set bits in an integer.
  */
 int popcnt(uint x) pure
 {
     // Select the fastest method depending on the compiler and CPU architecture
     version(LDC)
     {
         return _popcnt(x);
     }
     else
     {
         // ...
     }
}
```

9il:
All work correctly. But it mast be BSR/BSF for x86 :P

WalterBright:
Relevant thread:

http://www.digitalmars.com/d/archives/digitalmars/D/Please_add_intrinsics_module_for_DMD_DRuntime_295063.html

Post showing that dmd does correctly treat `bsf()` as an intrinsic:

http://www.digitalmars.com/d/archives/digitalmars/D/Please_add_intrinsics_module_for_DMD_DRuntime_295063.html#N295094

9il:
Thank you all and sorry for the noise !

PR #1695, opened 2016-11-24
fix Issue 16743 - Intrinsic recognition sometimes fails if a software…

WalterBright:
… implementation is available

This is a partial fix. dmd fix coming shortly.

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -72,27 +72,28 @@ unittest
  */
 int bsf(uint v) pure
 {
-    static if (size_t.sizeof == ulong.sizeof)
-    {
-        pragma(inline, true);
-        return bsf(cast(ulong) v);
-    }
-    else
-        return softBsf!uint(v);
+    pragma(inline, false);  // so intrinsic detection will work
+    return softBsf!uint(v);
 }
 
 /// ditto
 int bsf(ulong v) pure
 {
-    static if (size_t.sizeof == uint.sizeof)
+    static if (size_t.sizeof == ulong.sizeof)  // 64 bit code gen
     {
+        pragma(inline, false);   // so intrinsic detection will work
+        return softBsf!ulong(v);
+    }
+    else
+    {
+        /* intrinsic not available for 32 bit code,
+         * make do with 32 bit bsf
+         */
         const sv = Split64(v);
         return (sv.lo == 0)?
             bsf(sv.hi) + 32 :
             bsf(sv.lo);
     }
-    else
-        return softBsf!ulong(v);
 }
 
 ///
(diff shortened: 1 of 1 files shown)


src/core/bitop.d
```diff
@@ -72,27 +72,35 @@ unittest
  */
 int bsf(uint v) pure
 {
-    static if (size_t.sizeof == ulong.sizeof)
+    pragma(inline, false);  // so intrinsic detection will work
+    static if (size_t.sizeof == ulong.sizeof)  // 64 bit code gen
```
tsbockman: I originally used a `static if` here because `bsf(uint)` was not an intrinsic on 64-bit, and the **intrinsic** `bsf(ulong)` is faster than `softBsf!uint()`.

However, the software fallback `softBsf!ulong()` is slower, so you should ditch the `static if` and just always call `softBsf!uint()` for the fallback.

src/core/bitop.d
```diff
@@ -119,27 +127,35 @@ unittest
  */
 int bsr(uint v) pure
 {
-    static if (size_t.sizeof == ulong.sizeof)
+    pragma(inline, false);  // so intrinsic detection will work
+    static if (size_t.sizeof == ulong.sizeof)  // 64 bit code gen
```
tsbockman: As with `bsf(uint)`, above, you should ditch the `static if` and just always take the `false` branch, which is faster than your revised `true` branch.

tsbockman:
We should avoid creating a double indirection for the software fallback:

Originally, I set things up so that `bsf()` and `bsr()` would inline, while the handling of `softScan()` would be left to the compiler's discretion.

Since you are now forbidding `bsf()` and `bsr()` to inline, and (once my other two comments, above, are addressed) each instantiation of `softScan()` can only ever be called from exactly one place, `softScan()` should be marked `pragma(inline, true);` to avoid unnecessary function call overhead.

WalterBright:
The trouble with the original design is that:

1. inlining would prevent intrinsic recognition.
2. the `static if` ordering was switched between the two bsf's, making it very confusing to follow.
3. the forwarding of one bsf to another was confusing and unnecessary

I don't understand your comments. The functions are short enough - why not post what you think they should be?

tsbockman:
> why not post what you think they should be?

Like this:
```D
int bsf(uint v) pure
{
    pragma(inline, false);
    return softBsf!uint(v);
}
```
It's faster, simpler, and does the same thing.

WalterBright:
Ok, that makes sense. Done.

tsbockman:
`softScan()` should have `pragma(inline, true);` (on line 174) if the compiler will allow it: none of its instantiations are ever called from more than one place so there's no point to paying another layer of function call overhead to reach its body.

WalterBright:
> softScan() should have

1. This is kinda out of the scope of this PR, which is to fix the inlining problem

2. The reason the compiler has a `-inline` switch is it is worthwhile to be able to turn inlining on and off for various reasons. I'm reluctant to override these decisions without really strong reasons. Should we in general start going through druntime and inserting `pragma(inline, true)` here and there? Perhaps, but that would be an entirely separate issue from the purpose of this PR and deserves its own discussion.

andralex:
Auto-merge toggled on

andralex:
Hm, we have an issue with codecov - the soft implementation is not covered but I suspect it's exactly because it doesn't get used?

WalterBright:
Exactly!

PR #1696, opened 2016-11-24
fix issue 5995 - string append negative integer causes segfault

somzzz:
Appending an invalid UTF8 character now throws a UnicodeException:
core.exception.UnicodeException@src/rt/lifetime.d(2031): Invalid UTF-8 sequence

https://issues.dlang.org/show_bug.cgi?id=5995

### DIFF
--- a/src/rt/lifetime.d
+++ b/src/rt/lifetime.d
@@ -2043,16 +2043,24 @@ unittest
 {
     import core.exception : UnicodeException;
 
-    try
+    static void assertThrown(T : Throwable = Exception, E)(lazy E expr, string msg)
+    {
+        try
+            expr;
+        catch (T e) {
+            assert(e.msg == msg);
+            return;
+        }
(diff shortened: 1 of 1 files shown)


src/rt/lifetime.d
```diff
@@ -2036,6 +2039,30 @@ extern (C) void[] _d_arrayappendcd(ref byte[] x, dchar c)
...
+
+    static void assertThrown(T : Throwable = Exception, E)(lazy E expr, string msg)
+    {
+        try
+            expr;
+        catch (T e) {
+            assert(e.msg == msg);
+            return;
+        }
+    }
```
andralex: OK, please add a comment summarizing why it was necessary to add this template instead of just using the inline code.

somzzz: Done, also created & referenced an issue for it.

PR #1697, opened 2016-11-24
Add bindings for POSIX iconv library

nemanja-boric-sociomantic:
This adds bindings for `iconv_open`, `iconv` and `iconv_close`.

### DIFF
--- a/src/core/sys/posix/iconv.d
+++ b/src/core/sys/posix/iconv.d
@@ -1,0 +1,44 @@
+/*******************************************************************************
+
+    D binding for the POSIX iconv library.
+
+    Defines external functions required to use iconv codeset conversion
+    function.
+
+    iconv_open(3)   Allocates the descriptor for code conversion
+    iconv(3)        Performs the conversion
+    iconv_close(3)  Deallocates allocated resources
+
+    Copyright:  Copyright (c) 2016 Sociomantic Labs. All rights reserved.
+    License:    $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+    Authors:    Nemanja Boric
+    Standards:  POSIX.1-2001, POSIX.1-2008
+    See_Also:
+       http://pubs.opengroup.org/onlinepubs/009695399/functions/iconv_open.html
+
+*******************************************************************************/
+
+module core.sys.posix.iconv;
+
+version (Posix):
+extern (C):
+nothrow:
+@nogc:
+
+
+alias void* iconv_t;
+
+/// Allocate descriptor for code conversion from codeset FROMCODE to
+/// codeset TOCODE.
+iconv_t iconv_open (in char* tocode, in char* fromcode);
+
+/// Convert at most *INBYTESLEFT bytes from *INBUF according to the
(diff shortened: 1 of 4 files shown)


nemanja-boric-sociomantic:
dautotest failure:

```
git clone -b v1.078.0 --depth=1 https://github.com/dlang/druntime ../druntime-1.078.0
Cloning into '../druntime-1.078.0'...
warning: Could not find remote branch v1.078.0 to clone.
fatal: Remote branch v1.078.0 not found in upstream origin
posix.mak:344: recipe for target '../druntime-1.078.0' failed
make: *** [../druntime-1.078.0] Error 128
Build failed: object.Exception@ae/sys/d/manager.d(673): Command ["make", "-f", "posix.mak", "all", "kindle", "pdf", "verbatim", "NODATETIME=nodatetime.ddoc", "LATEST=1.078.0"] failed with status 2
```

src/core/sys/posix/iconv.d
```diff
@@ -0,0 +1,37 @@
...
+
+    See_Also:
+       http://pubs.opengroup.org/onlinepubs/009695399/functions/iconv_open.html
+
+*******************************************************************************/
+
+module core.sys.posix.iconv;
+
+version (Posix):
+extern (C):
```
ZombineDev: BTW,

> This function is a possible cancellation point and therefore not marked with __THROW.  

no need to worry about that. Other functions that are cancellation points are still marked with `nothrow` in druntime. There would be no point in not marking them with `nothrow` because `abi::__forced_unwind` exceptions are not meant to be caught anyway (I'm not even sure if it is possible to catch them in D).

nemanja-boric-sociomantic: About `pure`, I don't _think_ this qualifies, as it is changing the global state via  `errno`, which is not passed via parameters. Is my reasoning correct, or am I missing something?

nemanja-boric-sociomantic:
Updated.

src/core/sys/posix/iconv.d
ZombineDev: Yep, your right everything that touches mutable global state (such as `errno`) is definitely not pure.

src/core/sys/posix/iconv.d
```diff
@@ -0,0 +1,33 @@
+/*******************************************************************************
+
+    D binding for the POSIX iconv library.
+
+    See_Also:
+       http://pubs.opengroup.org/onlinepubs/009695399/functions/iconv_open.html
+
```
ZombineDev: I'm not sure, but I think every new file should have copyright and license, similar to e.g. https://github.com/dlang/druntime/blob/master/src/core/sys/posix/mqueue.d

@WalterBright is probably the right person to ask.

nemanja-boric-sociomantic: It looks like it indeed should have a header, thanks! I've updated this PR with the hearer like in `mqueue.d` module.

nemanja-boric-sociomantic:
I'll add for this and all others.

nemanja-boric-sociomantic:
Updated with changelog entry (not sure if it's all right, though).

PR #1698, opened 2016-11-24
Add Fiber's guard page in Posix as well

nemanja-boric-sociomantic:
Windows Fiber implementation already uses the fiber stack guard page.
This brings the similar protection for mmaped fiber stacks, using
mprotect.

### DIFF
--- a/changelog/fiber.dd
+++ b/changelog/fiber.dd
@@ -1,0 +1,9 @@
+Add Fiber's stack-protection page for Posix.
+
+The feature already existing for Windows' fiber implementation is now added to
+the systems using mmap to allocate fibers' stacks: After (or before) the last
+page used for the Fiber's stack, the page is allocate which is protected for
+any kind of access. This will cause system to trap immediately on the fiber's
+stack overflow. If in debugger session, one can inspect contents of the memory
+before or after stack pointer and it can be seen if it contains END OF FIBER
+string pattern.
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -4429,6 +4429,9 @@ private:
 
             static if( __traits( compiles, mmap ) )
             {
+                // Allocate PAGESIZE more for the memory guard
+                sz += PAGESIZE;
+
                 m_pmem = mmap( null,
                                sz,
                                PROT_READ | PROT_WRITE,
@@ -4458,13 +4461,31 @@ private:
             {
                 m_ctxt.bstack = m_pmem + sz;
                 m_ctxt.tstack = m_pmem + sz;
+                void* guard = m_pmem;
             }
             else
             {
                 m_ctxt.bstack = m_pmem;
                 m_ctxt.tstack = m_pmem;
+                void* guard = m_pmem + sz - PAGESIZE;
             }
             m_size = sz;
+
+            static if( __traits( compiles, mmap ) )
+            {
+                // Mark end of stack
+                for ( ubyte* g = cast(ubyte*)guard; g < guard + PAGESIZE; g+= 32)
+                    g[0 .. 32] = cast(ubyte[]) "END OF FIBER -- END OF FIBER -- ";
+
+                // protect end of stack
+                if ( mprotect(guard, PAGESIZE, PROT_NONE) == -1 )
+                    abort();
+            }
+            else
+            {
+                // Supported only for mmap allocated memory - results are
+                // undefined if applied to memory not obtained by mmap
+            }
         }
 
         Thread.add( m_ctxt );
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -4484,10 +4484,6 @@ private:
             {
                 if (guardPageSize)
                 {
-                    // Mark end of stack
-                    for ( ubyte* g = cast(ubyte*)guard; g < guard + guardPageSize; g+= 32)
-                        g[0 .. 32] = cast(ubyte[]) "END OF FIBER -- END OF FIBER -- ";
-
                     // protect end of stack
                     if ( mprotect(guard, guardPageSize, PROT_NONE) == -1 )
                         abort();
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -3976,21 +3976,21 @@ class Fiber
      * Params:
      *  fn = The fiber function.
      *  sz = The stack size for this fiber.
-     *  guard_page_size = size of the guard page to trap fiber's stack
+     *  guardPageSize = size of the guard page to trap fiber's stack
      *                    overflows
      *
      * In:
      *  fn must not be null.
      */
     this( void function() fn, size_t sz = PAGESIZE*4,
-          size_t guard_page_size = PAGESIZE ) nothrow
+          size_t guardPageSize = PAGESIZE ) nothrow
     in
     {
         assert( fn );
     }
     body
     {
-        allocStack( sz, guard_page_size );
+        allocStack( sz, guardPageSize );
         reset( fn );
     }
 
@@ -4002,21 +4002,21 @@ class Fiber
      * Params:
      *  dg = The fiber function.
      *  sz = The stack size for this fiber.
-     *  guard_page_size = size of the guard page to trap fiber's stack
+     *  guardPageSize = size of the guard page to trap fiber's stack
      *                    overflows
      *
      * In:
      *  dg must not be null.
      */
     this( void delegate() dg, size_t sz = PAGESIZE*4,
-          size_t guard_page_size = PAGESIZE ) nothrow
+          size_t guardPageSize = PAGESIZE ) nothrow
     in
     {
         assert( dg );
     }
     body
     {
-        allocStack( sz, guard_page_size);
+        allocStack( sz, guardPageSize);
         reset( dg );
     }
 
@@ -4361,7 +4361,7 @@ private:
     //
     // Allocate a new stack for this fiber.
     //
-    final void allocStack( size_t sz, size_t guard_page_size ) nothrow
+    final void allocStack( size_t sz, size_t guardPageSize ) nothrow
     in
     {
         assert( !m_pmem && !m_ctxt );
@@ -4386,7 +4386,7 @@ private:
         {
             // reserve memory for stack
             m_pmem = VirtualAlloc( null,
-                                   sz + guard_page_size,
+                                   sz + guardPageSize,
                                    MEM_RESERVE,
                                    PAGE_NOACCESS );
             if( !m_pmem )
@@ -4394,7 +4394,7 @@ private:
 
             version( StackGrowsDown )
             {
-                void* stack = m_pmem + guard_page_size;
+                void* stack = m_pmem + guardPageSize;
                 void* guard = m_pmem;
                 void* pbase = stack + sz;
             }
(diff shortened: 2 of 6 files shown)


src/core/thread.d
```diff
@@ -4427,13 +4430,26 @@ private:
...
+            static if( __traits( compiles, mmap ) )
+            {
+                // Mark end of stack
+                for ( ubyte* g = cast(ubyte*)guard; g < guard + PAGESIZE; g+= 32)
+                    g[0 .. 32] = cast(ubyte[]) "END OF FIBER -- END OF FIBER -- ";
+
+                // protect end of stack
+                if ( mprotect(guard, PAGESIZE, PROT_NONE) == -1 )
+                    onOutOfMemoryError();
+            }
```
Dicebot: If `valloc`/`malloc` cases are not supported, I'd suggest putting explicit empty `else` clause with a comment about it (to

nemanja-boric-sociomantic: `(to` ... ? :-)

Dicebot: .. to better match previous `static if` block)

nemanja-boric-sociomantic:
Updated with the `else` block and the comment saying that results are defined only for memory obtained by `mmap`

nemanja-boric-sociomantic:
Not sure why `DAutoTest` failed:

```
make[1]: Leaving directory '/home/dtest/DAutoTest/work/repo/phobos'
for f in `find "/home/dtest/DAutoTest/work/repo/dlang.org/web/phobos-prerelease-verbatim" -iname '*.html'`; do mv $f `dirname $f`/`basename $f .html`.verbatim; done
mv /home/dtest/DAutoTest/work/repo/dlang.org/web/phobos-prerelease-verbatim/* /home/dtest/DAutoTest/work/repo/dlang.org/web/phobos-prerelease/
rm -r /home/dtest/DAutoTest/work/repo/dlang.org/web/phobos-prerelease-verbatim
fatal: 'refs/ae-sys-d-cache/website-2fa232357638ad439f82243dfc7f9d1c2f237c46-f798e1d288b93f2f15229c70de8009f8' - not a valid ref
```

ZombineDev:
The DAutoTest is about testing the documentation build, and so its failure is unrelated to your pull request.

src/core/thread.d
```diff
@@ -4427,13 +4430,26 @@ private:
...
+
+            static if( __traits( compiles, mmap ) )
+            {
+                // Mark end of stack
+                for ( ubyte* g = cast(ubyte*)guard; g < guard + PAGESIZE; g+= 32)
+                    g[0 .. 32] = cast(ubyte[]) "END OF FIBER -- END OF FIBER -- ";
+
+                // protect end of stack
+                if ( mprotect(guard, PAGESIZE, PROT_NONE) == -1 )
+                    onOutOfMemoryError();
```
nemanja-boric-sociomantic: Yeah, but I was just following the module's style, which uses this for anything that can fail (see the same thing in Windows: https://github.com/nemanja-boric-sociomantic/druntime/blob/68bf497033575bdcb41a9392e72bdcf794c66016/src/core/thread.d#L4385-L4386)

MartinNowak:
Maybe we should make the guard size configurable, including a size of 0, much as the stack size is configurable.
Not everyone might want to spent his address space and the (tiny) page mapping overhead, and a single page might not be enough for all memory corruption concerns.

src/core/thread.d
```diff
@@ -4427,13 +4430,31 @@ private:
...
                 m_ctxt.tstack = m_pmem;
+                void* guard = m_pmem + sz - PAGESIZE;
             }
             m_size = sz;
+
+            static if( __traits( compiles, mmap ) )
+            {
+                // Mark end of stack
+                for ( ubyte* g = cast(ubyte*)guard; g < guard + PAGESIZE; g+= 32)
+                    g[0 .. 32] = cast(ubyte[]) "END OF FIBER -- END OF FIBER -- ";
```
MartinNowak: Not acceptable to write to that page.
The point of a guard page is that it's not accessible, not backed by real memory, and causes a (page) fault when being accessed.

src/core/thread.d
MartinNowak: Well, the other ones look like they allocate sth., b/c VirtualAlloc is used to change flags, and should be changed to fail instead of using onOutOfMemoryError.

src/core/thread.d
Dicebot: It is not a guarded at this point and writing recognizable string is the main purpose if this PR. It makes possible to easily debug stack overflow in a fiber by printing few bytes around rsp in gdb and checking if it resembles a guard string. Similar code has existed in original tango runtime and I have no idea how one is expected to work with heavy usage of fibers without it - pretty much everyone relies on it in daily workflow at Sociomantic :(

src/core/thread.d
nemanja-boric-sociomantic: OK, fair enough, will change to abort.

nemanja-boric-sociomantic:
> Maybe we should make the guard size configurable, including a size of 0, much as the stack size is configurable.
Not everyone might want to spent his address space and the (tiny) page mapping overhead, and a single page might not be enough for all memory corruption concerns.

Ok, sure. Do you prefer to do this via druntime config argument?

nemanja-boric-sociomantic:
Updated with the:

- Change `onOutOfMemoryError` to `abort` in  case of `mprotect` failure
- Allow configuring the size of the stack guard (on both Windows and Posix systems) with `--DRT-stack-guard-size` runtime option.

nemanja-boric-sociomantic:
Ah, FreeBSD failed because by the time it got to testing, changelog.dd started containing merge confict.

nemanja-boric-sociomantic:
Updated. @wilzbach could you advise if I got the new changelog format right?

wilzbach:
> Ah, FreeBSD failed because by the time it got to testing, changelog.dd started containing merge confict.

The idea is at that you will never have to rebase again due to changelog conflicts ;-)

https://github.com/dlang/phobos/pull/4228

> Updated. @wilzbach could you advise if I got the new changelog format right?

Yup - the idea was also that it's less complicated ;-)

nemanja-boric-sociomantic:
Thanks!

MartinNowak:
Sorry haven't looked at this for a while.
- Why can't the guard page be write protected? That would be a lot better. In fact actually using a page is a huge overhead, not to mention the cost of writing to main memory.
- Druntime config switches are fairly ugly and only intended for runtime "boot" configuration, doesn't seem necessary here, simply passing the guard size to the constructor of Fiber (like the stack size) should work, but might require to save the size somewhere.

src/core/thread.d
MartinNowak: Not going to rediscuss why Exceptions for async failures are wrong, unwinding is hardly possible for those, and having the compiler assume that each instruction could throw requires complex unwind information or conservative (slow) code.
In any case, neither an exception nor a segfault for a stack overflow tells you much else that that the stack overflowed.

nemanja-boric-sociomantic:
> Why can't the guard page be write protected? That would be a lot better. In fact actually using a page is a huge overhead, not to mention the cost of writing to main memory.

I'm not sure I follow this - the guard page is both write and read protected (see here: https://github.com/dlang/druntime/pull/1698/files#diff-8bb12ed976acf0a5132e877ec5a01ea8R4488)

I'll take a look later at other issues mentioned. Thanks!

src/core/thread.d
nemanja-boric-sociomantic: Sorry, I wasn't referring about the D exceptions, but the operating system's exception. For example, Windows builds will, this same runtime will, raise `STATUS_GUARD_PAGE` exception - which is very useful - it prevents you from silently corrupting your memory _and_ it will tell you that your fiber's stack overflowed (if you're familiar with the runtime, though).

On POSIX, I'm not sure what else I can do here. The most reliable way to report the cause of the segmentation fault here is to install `SIGSEGV` handler, and to compare the `siginfo_t.si_addr` to see if this is done by fiber stack's overflow, eventually print info to stderr and die. 

It is true that segmentation fault doesn't tell you why your program failed, but it least prevents you from silently corrupting other pages of the memory, making your program fail in completely unrelated code, just because some other fiber corrupted the memory. Sadly, this is not theoretical concern (and the code is already available on Windows), but this is something that we're seeing in practice every now and then.

However, even with the segmentation fault, if the core file is generated, you can load it in debugger and simply do `x/s $sp-64` and see if the contents of the memory is just `END OF FIBER`. This is, btw, the reason why we're filling this page with data.

nemanja-boric-sociomantic: (I forgot to mention @MartinNowak to pass email filters :-) )

nemanja-boric-sociomantic:
> Druntime config switches are fairly ugly and only intended for runtime "boot" configuration, doesn't seem necessary here, simply passing the guard size to the constructor of Fiber (like the stack size) should work, but might require to save the size somewhere.

I've updated the code so that this is configurable using Fiber's contructor's argument (with an ability to turn it off by setting `0`). There wasn't need to save the size anywhere, since it's just counted in the stack's size (in POSIX's case), or it was deallocated automatically (like before in Window's case).

src/core/thread.d
```diff
@@ -4458,13 +4470,34 @@ private:
...
             }
             m_size = sz;
+
+            static if( __traits( compiles, mmap ) )
+            {
+                if (guard_page_size)
+                {
+                    // Mark end of stack
+                    for ( ubyte* g = cast(ubyte*)guard; g < guard + guard_page_size; g+= 32)
+                        g[0 .. 32] = cast(ubyte[]) "END OF FIBER -- END OF FIBER -- ";
```
MartinNowak: Here you're writing to the guard page which will cause it to actually get wired. Seems unnecessary, when the page is write protected below.

src/core/thread.d
```diff
@@ -3976,18 +3976,21 @@ class Fiber
      * Params:
      *  fn = The fiber function.
      *  sz = The stack size for this fiber.
+     *  guard_page_size = size of the guard page to trap fiber's stack
```
MartinNowak: Change to `guardPageSize` to match our naming.

src/core/thread.d
Burgos: Yeah, the intention is to easily check if the stackoverflow has happened, simply by inspecting the memory pointed by sp in the debugger. I guess one could also check the mappings from /proc to confirm that the page is actually protected. If this writing is a deal breaker I would remove it.

src/core/thread.d
Burgos: Will do. Thanks for the review!

MartinNowak: I sent you an PR @Burgos.

Burgos:
@MartinNowak I've pushed a small commit with a trivial tidying up the test case, if you don't mind.

Burgos:
Green!

changelog/fiber.dd
```diff
@@ -0,0 +1,9 @@
+Add Fiber's stack-protection page for Posix.
+
+The feature already existing for Windows' fiber implementation is now added to
+the systems using mmap to allocate fibers' stacks: After (or before) the last
+page used for the Fiber's stack, the page is allocate which is protected for
+any kind of access. This will cause system to trap immediately on the fiber's
+stack overflow. If in debugger session, one can inspect contents of the memory
+before or after stack pointer and it can be seen if it contains END OF FIBER
```
klickverbot: This isn't true anymore, see #1821 for a fix.

PR #1699, opened 2016-11-24
Add missing POSIX <libgen.h> header

leandro-lucarella-sociomantic:


src/core/sys/posix/libgen.d
```diff
@@ -0,0 +1,20 @@
...
+ * Authors:   Leandro Lucarella
+ * Standards: The Open Group Base Specifications Issue 6, IEEE Std 1003.1, 2004 Edition
+ */
+
+/*          Copyright Sociomantic Labs GmbH 2016.
+ * Distributed under the Boost Software License, Version 1.0.
+ *    (See accompanying file LICENSE or copy at
+ *          http://www.boost.org/LICENSE_1_0.txt)
+ */
+module ocean.stdc.posix.libgen;
```
ZombineDev: This should be `core.sys.posix.libgen`, right?

src/core/sys/posix/libgen.d
```diff
@@ -0,0 +1,20 @@
...
+
+extern (C):
```
ZombineDev: Please add `@nogc nothrow pure:`

src/core/sys/posix/libgen.d
```diff
@@ -0,0 +1,20 @@
...
+
+char* basename(char*);
+char* dirname(char*);
```
ZombineDev: Just a personal preference:
```D
char* basename(char* path);
char* dirname(char* path);
```

src/core/sys/posix/libgen.d
ZombineDev: I'm not 100% sure about `pure`, but according to http://pubs.opengroup.org/onlinepubs/7908799/xsh/libgen.h.html those functions do not touch `errno` or other mutable global state, at least AFAICS.

src/core/sys/posix/libgen.d
leandro-lucarella-sociomantic: Oh, crap. Yes, of course :)

src/core/sys/posix/libgen.d
leandro-lucarella-sociomantic: The linux man page says:

>       Both dirname() and basename() may modify the contents of path, so it may be desirable to pass  a  copy when calling one of these functions.

Is this really `pure`? I think adding `pure` might be stretching this too much as different implementations could use static data for example.

Also: http://pubs.opengroup.org/onlinepubs/009695399/functions/basename.html

> The basename() function need not be reentrant. A function that is not required to be reentrant is not required to be thread-safe.

Definitely can't guarantee that all implementations out there are pure.

src/core/sys/posix/libgen.d
leandro-lucarella-sociomantic: Noted, but that's not the style of most C headers in druntime, so I prefer to stick with the common practices in the project :)

leandro-lucarella-sociomantic:
Updated with fixed `module` name and added `@nogc nothrow`.

leandro-lucarella-sociomantic:
Also added missing bits of makefiles.

leandro-lucarella-sociomantic:
BTW, those makefiles needs some work, lots of repeated stuff :unamused:

src/core/sys/posix/libgen.d
```diff
@@ -0,0 +1,21 @@
...
+
+/*          Copyright Sociomantic Labs GmbH 2016.
+ * Distributed under the Boost Software License, Version 1.0.
+ *    (See accompanying file LICENSE or copy at
+ *          http://www.boost.org/LICENSE_1_0.txt)
+ */
+module core.sys.posix.libgen;
+
+@nogc nothrow:
+extern (C):
```
leandro-lucarella-sociomantic: Updated!

ZombineDev:
Ping @mihails-strasuns-sociomantic @Dicebot this looks ready to go.

PR #1701, opened 2016-11-25
posix/time: Add timer manipulation macros

leandro-lucarella-sociomantic:


leandro-lucarella-sociomantic:
Updated to fix the build issues.

leandro-lucarella-sociomantic:
Green :)

@MartinNowak maybe have a quick look? Is a small one :)

src/core/sys/posix/sys/time.d
```diff
@@ -28,7 +28,55 @@ else version (WatchOS)
     version = Darwin;
 
 version (Posix):
-extern (C) nothrow @nogc:
+nothrow @nogc:
+
+
+/* macros provided to operate on timeval structures
+ * they are extern(D) because they are not really C symbols, just macros
+ */
```
schveiguy: This is nit picking, but why not just `extern(D)` these functions?

src/core/sys/posix/sys/time.d
```diff
@@ -28,7 +28,55 @@ else version (WatchOS)
...
+
+pure @safe void timeradd(const timeval* a, const timeval* b,
+        timeval* result)
+{
+    result.tv_sec = a.tv_sec + b.tv_sec;
+    result.tv_usec = a.tv_usec + b.tv_usec;
```
schveiguy: `assert(a.tv_usec < 1_000_000 && b.tv_usec < 1_000_000);` perhaps? otherwise, the result doesn't make sense.

src/core/sys/posix/sys/time.d
```diff
@@ -28,7 +28,55 @@ else version (WatchOS)
...
+        ++result.tv_sec;
+        result.tv_usec -= 1000000;
+    }
+}
+
+pure @safe void timersub(const timeval* a, const timeval* b,
+        timeval *result)
+{
+    result.tv_sec = a.tv_sec - b.tv_sec;
+    result.tv_usec = a.tv_usec - b.tv_usec;
```
schveiguy: Same assert here

src/core/sys/posix/sys/time.d
```diff
@@ -28,7 +28,55 @@ else version (WatchOS)
...
+{
+    (tvp.tv_sec = tvp.tv_usec = 0);
+}
+
+pure @safe int timerisset(timeval* tvp)
+{
+    return cast(int) (tvp.tv_sec || tvp.tv_usec);
+}
+
+pure @safe int timercmp(char[] CMP)(const timeval* a, const timeval* b)
```
schveiguy: should accept a string instead of char[]

src/core/sys/posix/sys/time.d
schveiguy: In fact, you could do:

```D
extern(D) pure @safe {
...
}
```

src/core/sys/posix/sys/time.d
```diff
@@ -28,7 +28,55 @@ else version (WatchOS)
...
+{
+    return cast(int)
+           mixin("((a.tv_sec == b.tv_sec) ?" ~
+                 "(a.tv_usec" ~ CMP ~ "b.tv_usec) :" ~
+                 "(a.tv_sec"  ~ CMP ~ "b.tv_sec))");
+}
```
schveiguy: nice!

src/core/sys/posix/sys/time.d
schveiguy: Also I'd like to see the underscores for all the large numeric constants.

src/core/sys/posix/sys/time.d
leandro-lucarella-sociomantic: I could do that if you find it more readable.

src/core/sys/posix/sys/time.d
leandro-lucarella-sociomantic: > otherwise, the result doesn't make sense.

Well, it makes sense, just a weird sense, just the same as having usecs > 1_000_000. This is based on a C macro, I know is not as purely correct as one might think, but we are wrapping C here, so I'm not sure if it's a good idea to depart from it. If we want better timer manipulation functions, they could be written in D, but this is just an interface to C functions.

But looking into this, I just realized this is not really POSIX:
```
CONFORMING TO

       Not in POSIX.1.  Present on most BSD derivatives.
```

http://man7.org/linux/man-pages/man3/timeradd.3.html

OK, I have to rethink this, and probably move it to the linux/bsd specific bindings...

src/core/sys/posix/sys/time.d
leandro-lucarella-sociomantic: Ha! I was in C mode :)

src/core/sys/posix/sys/time.d
leandro-lucarella-sociomantic: Again, this is taken almost literally from C macros. Is the closest D translation I could think of, I'm not particularly proud of it :P

leandro-lucarella-sociomantic:
Thanks for the review BTW.

src/core/sys/posix/sys/time.d
schveiguy: I like it though :)

src/core/sys/posix/sys/time.d
schveiguy: It's a nit, and not critical. But it does allow you to avoid attribute verbosity.

src/core/sys/posix/sys/time.d
schveiguy: > Well, it makes sense, just a weird sense, just the same as having usecs > 1_000_000

Right, not how one might expect! For instance, if you had:

```D
auto ts1 = timeval(1, 5_000_000);
auto ts2 = timeval(5, 1_000_000);
timeval ts3;
timer_add(&ts1, &ts2, &ts3);
// a reasonable expectation:
assert(ts3.tv_sec == 12 && ts3.tv_usec == 0);
```
Thinking about it a bit, it's also possible that the current code works on timevals where the individual usecs might be above 1 million, but added together are still under 2 million, which would result in a valid timeval (ironically!).

> OK, I have to rethink this, and probably move it to the linux/bsd specific bindings...

I'm assuming this is an aid to porting C code. But how bad would it be to change the lines that use this function to instead call a more d-ish function?

In any case, even if the timer math functions are not posix, timeval is, and this function would be valid on all platforms that have the type.

leandro-lucarella-sociomantic: I think this, being an interface to C code, just have to do the exact same as C does. Sometimes is not just a matter of porting C code that then will be just "pure D", sometimes, at least at Sociomantic, we just prefer to use C functions because they are proven and we know what to expect. OK, this is a grey zone because the actual code will be generated by the D compiler, not linked, but having the exact same implementation as in C has some value IMHO.

Remember these are just plain C bindings, not phobos, things here just have to be as close to C as possible, not be D idioms to achieve the same you could achieve using the C counterparts but more conveniently and safely.

leandro-lucarella-sociomantic:
I see there is a `CRuntime_Glibc` version identifier, since this is coming from glibc, I guess this can be reused in any platform that comes with glibc, but I don't know how are you handling modules in this case. I was going to put it in `core.sys.linux.sys.time`, but it might apply to other glibc platforms? But there is no `core.sys.glibc` for example, how do you deal with that? (there are many stdc extensions in glibc).

For now I will include it in linux and make sure glibc runtime is used to make sure we are mimicking the actually used C runtime.

leandro-lucarella-sociomantic:
Updated

src/core/sys/linux/sys/time.d
```diff
@@ -0,0 +1,72 @@
...
+module core.sys.linux.sys.time;
+
+private import core.sys.linux.config;
+public import core.sys.posix.sys.types;  // timeval
+
+version (linux)
+else static assert(false, "Unsupported platform");
+
+version( CRuntime_Glibc )
+else static assert(false, "Unsupported platform");
```
schveiguy: There's a trivial error here, you need empty braces for the `version` statement to apply to something.

But I'm wondering if we need all this anyway. Looking at other examples in the sys/linux directory, I think you can just put:
```D
version (linux): version (CRuntime_Glibc):
```
Having imports that explode a compilation means it just pushes the version statements outside the lib and into user code.

schveiguy: In fact, that reveals an error in your code, the import of this file from `core.sys.posix.sys.time` is versioned on `linux` but not `CRuntime_Glibc`. On platforms that define only the former, druntime would not compile.

leandro-lucarella-sociomantic: Good catch. Is tricky with this `CRuntime_Glibc`.

leandro-lucarella-sociomantic: Mmm, it's very weird to have:
```d
version (linux) version (CRuntime_Glibc)
    public import core.sys.linux.sys.time;
```
At that file we don't really know about restrictions on that particular module and glibc runtime.

I see 3 possible solutions:
1. Ignore the `CRuntime_Glibc` completely and just hope nobody uses linux with a different (and incompatible) runtime.
2. Create a `core.sys.glibc.sys.time` and make it only versioned with `CRuntime_Glibc` (not mention linux in there and hope all glibc implementations in different platforms are the same).
3. Move the code to `core.sys.posix.sys.time` and put it versioned with both `CRuntime_Glibc` and `linux`, but then that's not really posix (which makes me think that, even when in C one would do a regular `#include <sys/time.h>`, maybe the automatic public import in `core.sys.posix.sys.time` shouldn't be done if there is some guarantee that if you stay in `core.sys.posix` your code won't call non-standard functions.

And I don't like any of them 100% :sweat: :gun:

leandro-lucarella-sociomantic: Well, about 3, I guess if the idea of `core.sys` is to mimic C, then I guess the implicit import / moving the code is fine. In C you don't know what extensions you might be using if you don't read the man page.

schveiguy: Well, you could actually just version the whole core.sys.linux.time file like I said (with the two versions required), and then if someone imports the files and tries to use the functions they won't be defined. Look for instance at: https://github.com/dlang/druntime/blob/master/src/core/sys/linux/sys/auxv.d at the top. The precedent is pretty clear.

I do agree that people aren't going to have thought of the difference between glibc and non glibc. Most people just say version(linux) and call it a day.

However, in this case, you are doing macros. Glibc isn't really required. I'm not sure we need it.

leandro-lucarella-sociomantic: OK, I will keep it simple then.

leandro-lucarella-sociomantic: And yeah, if `CRuntime_Glibc` is not defined by the compiler, it doesn't make any sense to use that version statement, the user shouldn't be require to set that version manually.

leandro-lucarella-sociomantic:
Updated.

leandro-lucarella-sociomantic:
Updated with fixes to windows makefiles. I wonder why files to be copies are repeated **3 times** in different makefiles :unamused:

schveiguy:
Yep, looks good to me!

> Updated with fixes to windows makefiles. I wonder why files to be copies are repeated 3 times in different makefiles

Grumble...

schveiguy:
Hm... Just realized, you have the functions still in the posix file!

leandro-lucarella-sociomantic:
> Hm... Just realized, you have the functions still in the posix file!

I think I messed it up with the commits. I had 2 commits for some reason, I just assumed I wanted to squash them so I did without checking. I hope next time I finally get it right! :S

src/core/sys/linux/sys/time.d
```diff
@@ -0,0 +1,68 @@
...
+
+private import core.sys.linux.config;
+public import core.sys.posix.sys.time;  // timeval
+
+version (linux):
+
+/* macros provided to operate on timeval structures
+ * they are extern(D) because they are not really C symbols, just macros
+ */
+extern (D) pure @safe @nogc {
```
ZombineDev: `nothrow` is missing.

leandro-lucarella-sociomantic: Fixed.

PR #1702, opened 2016-11-26
fix issue 16797 - Zero clock resolution lead to division by zero

joakim-brannstrom:
I haven't tried it in production yet so unable to confirm if this is the only fix that is needed.
It do contains an additional check for negative time which I've also observed.

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -2443,12 +2443,14 @@ extern(C) void _d_initMonoTime()
                         assert(0);
 
                     // For some reason, on some systems, clock_getres returns
-                    // a resolution which is clearly wrong (it's a millisecond
-                    // or worse, but the time is updated much more frequently
-                    // than that). In such cases, we'll just use nanosecond
-                    // resolution.
-                    tps[i] = ts.tv_nsec >= 1000 ? 1_000_000_000L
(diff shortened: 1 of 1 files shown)


jmdavis:
That's an awfully verbose fix. It's a lot of if statements that do the same thing. An equivalent fix would be to do

    tps[i] = tv.tv_sec != 0 || ts.tv_nsec <= 0 || ts.tv_nsec >= 1000
           ? 1_000_000_000L : 1_000_000_000L / ts.tv_nsec;

Please change it to that. And wow do some folks have some messed up computers. It just seems crazy that something like `clock_getres` would be reporting such clearly wrong values. And it seems pretty absurd that we have to put in all of these extra checks because of misbehaving systems, but I don't know what else to do if folks are really seeing this sort of thing in the wild. However, I have to think that at some point, if a system is going to misbehave, we should just let it misbehave rather than work around it and affect everyone else. Division by zero is obviously really bad though.

joakim-brannstrom:
Done :)

jmdavis:
Auto-merge toggled on

denis-sh:
@jmdavis, I'd say this bug shows why one shouldn't merge his own code to a repository (this is what you just did as @joakim-brannstrom just copied code from your comment above).

PR #1703, opened 2016-11-27
Fix bug caused by pull #1702.

denis-sh:
`_ticksPerSecond` was always set to `1_000_000_000L` in `version(Posix)` branch because of 09d570f47cd1c727298e4b0fe1ad668f1dd28e4e.

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -2449,7 +2449,7 @@ extern(C) void _d_initMonoTime()
                     //  - it's negative
                     //  - it's zero
                     // In such cases, we'll just use nanosecond resolution.
-                    tps[i] = ts.tv_sec != 0 || ts.tv_nsec <= 0 || ts.tv_nsec >= 1000
+                    tps[i] = ts.tv_nsec <= 0 || ts.tv_nsec >= 1000
                         ? 1_000_000_000L : 1_000_000_000L / ts.tv_nsec;
                 }
             }


joakim-brannstrom:
Interesting. There are systems that report a clock resolution of 1s or more?

schveiguy:
How does this fix anything? A tick that lasts more than a second is going to result in wrong ticks per second anyway. Please outline the bug this is fixing.

jmdavis:
Yeah, a system that's reporting a `tv_sec` value of more than `0` is sounds seriously borked. And it's not like the changes prior to pull #1702 would have handled it correctly anyway. `MonoTime` doesn't use `tv_sec` to calculate the clock resolution (because it makes no sense for a clock to reporting a resolution that low), and so previously, it just ignored `tv_sec` entirely, meaning that this change doesn't do anything to improve the correctness of a system that's reporting a non-zero value for `tv_sec`. It's wrong either way.

`MonoTime` is checking `tv_sec` now on the theory that if `tv_sec` is non-zero, then `clock_res` is too borked to be trusted. `1_000_000_000` is then the best value that we have, because the tick resolution doesn't even really matter for Linux and FreeBSD anyway (aside from trying to be informative to the programmer), because `clock_gettime` reports everything in nanoseconds regardless of the actual clock resolution, which is why it was previously made to just use `1_000_000_000` when `clock_res` reported that the clock had millisecond resolution or worse. Claiming nanosecond resolution wasn't correct, but it was more correct than actually using the value that `clock_getres` reported, since the clock clearly had a higher resolution than one millisecond.

If there's a system where `clock_getres` is setting `tv_sec` to anything other than zero, I don't see how we could trust what it's reporting, and the only way that using what it reports in `tv_nsec` would be correct is if it set that correctly while setting `tv_sec` incorrectly, and we have no way of knowing that it's actually doing that. Or is have you actually found a system that has a clock resolution of worse than 1 second? If so, this is not the correct fix. We'd need to adjust how `ticksPerSecond` worked so that it would be possible for it to deal with fractional ticks per second, which would definitely not be good for performance, since it would mean extra calculations that were pointless on almost all machines.

And as far as `MonoTime` functioning correctly goes, we could totally get away with just always setting the ticks per second to `1_000_000_000` on Linux and FreeBSD, and everything would be fine. It's just that if the programmer looked at `ticksPerSecond`, they wouldn't be getting the actual ticks per second of the system clock. As it stands, we at least sometimes get the correct system clock resolution, but clearly, on some systems, what `clock_getres` reports is just plain borked. And in that case, setting it to `1_000_000_000` works, even if it's not ideal.

denis-sh:
> How does this fix anything? A tick that lasts more than a second is going to result in wrong ticks per second anyway. Please outline the bug this is fixing. i.e. Which clock on which system comes out wrong, and what are the timespec fields set to?

Seriously? I can only repeat pull description here:

> `_ticksPerSecond` was always set to `1_000_000_000L` in `version(Posix)` branch because of 09d570f47cd1c727298e4b0fe1ad668f1dd28e4e.

If for some reason it's not enough, spoiler: I just fixed `a != 0 || a <= 0` code (which is (the second spoiler!) always `true`).

joakim-brannstrom:
Not the same.
```d
ts.tv_sec != 0 || ts.tv_nsec <= 0
```

denis-sh:
> Not the same.

This means I should carefully read the code. Stupid me, sorry for the noise.

joakim-brannstrom:
You did good. Got everyone thinking twice :)
I was unsure of the fix because of the lack of testing to rely upon :/

denis-sh:
Yes, no test-cases situation is really a bad one.
What was really good is I have shown (at least to myself) that when you stop contributing *dlang* for a considerable time you start doing code reading mistakes like this one.

jmdavis:
> Yes, no test-cases situation is really a bad one.

Yeah, unfortunately, in this case, there really isn't any way to test it aside from someone who previously had problems running it on their box, since testing it requires running it on a box that somehow manages to give bad results with `clock_getres`.

> What was really good is I have shown (at least to myself) that when you stop contributing dlang for a considerable time you start doing code reading mistakes like this one.

Well, in this case, if the problem was thinking that `tv_sec` was `tv_nsec`, that's easy to do if you don't read the code carefully, no matter how often you're programming or how often you're programming in D.

d-random-contributor:
As for a test:
```d
version(unittest) alias clock_getres=test_clock_getres;
else import core.posix.time:clock_getres;
```
Or something like that.

schveiguy:
> As for a test:

No, we need the clock resolutions to work during unit testing of druntime!

There are probably ways to get this working only for one unit test (perhaps by passing clock_getres as an alias to the actual function).

d-random-contributor:
```d
test_clock_getres
{
  if(work)core.posix.time.clock_getres(clock,time);
  else (test code);
}
```

PR #1705, opened 2016-11-29
Add Linux-specific sched_setaffinity and sched_getaffinity

nemanja-boric-sociomantic:


nemanja-boric-sociomantic:
Updated with changelog entry (not sure if it's all right, though).

PR #1706, opened 2016-12-01
Fixup PR #1687 - wrong buffer length used

tramker:
This corrects a wrong length used in #1687. Sorry for the mess-up.
Ping @rainers.


rainers:
Oops, should have noticed in the last review...

rainers:
Auto-merge toggled on

PR #1710, opened 2016-12-12
Make AssertError a singleton

andralex:
This makes AssertError a singleton, which allows people to use `assert` in higher-level applications without needing to link in the garbage collector.

This could break code that calls `onAssertError` repeatedly and assumes it returns separate objects. That's quite an unusual way to go about things, and it seems we need not guarantee it.

If this approach works we can (a) publish `singleton` so as to make it available to clients, and (b) convert `RangeError` to a singleton as well so we can throw range errors without worrying about memory allocations.

### DIFF
--- a/src/core/exception.d
+++ b/src/core/exception.d
@@ -68,11 +68,11 @@ class AssertError : Error
     }
 }
 
-unittest
+@safe unittest
 {
     {
         auto ae = singleton!AssertError;
-        assert(typeid(ae) == typeid(AssertError));
+        assert(typeid(ae) is typeid(AssertError));
         assert(ae.file is null);
         assert(ae.line == 0);
     }
@@ -436,16 +436,19 @@ Different threads own different instances of `C`.
 
 The destructor of the singleton object is called during application's exit.
 */
-private @trusted C singleton(C)() if (is(C == class) && is(typeof(new C)))
+private C singleton(C)() if (is(C == class) && is(typeof(new C)))
 {
     static size_t[1 + (__traits(classInstanceSize, C) - 1) / size_t.sizeof] b;
     static bool initialized = false;
-    auto result = cast(C) b.ptr;
+    auto result = () @trusted { return cast(C) b.ptr; }();
     if (!initialized)
     {
         assert(typeid(C).initializer.length == b.length * size_t.sizeof);
         import core.stdc.string;
-        memcpy(b.ptr, typeid(C).initializer.ptr, b.length);
+        () @trusted
+        {
+            memcpy(b.ptr, typeid(C).initializer.ptr, b.length);
+        }();
         result.__ctor;
         import core.internal.traits;
         static if (hasElaborateDestructor!C)
--- a/src/core/exception.d
+++ b/src/core/exception.d
@@ -47,6 +47,11 @@ unittest
  */
 class AssertError : Error
 {
+    @safe pure nothrow this()
+    {
+        super(null);
+    }
+
     @safe pure nothrow this( string file, size_t line )
     {
         this(cast(Throwable)null, file, line);
@@ -65,6 +70,12 @@ class AssertError : Error
 
 unittest
 {
+    {
+        auto ae = singleton!AssertError;
+        assert(typeid(ae) == typeid(AssertError));
+        assert(ae.file is null);
+        assert(ae.line == 0);
+    }
     {
         auto ae = new AssertError("hello", 42);
         assert(ae.file == "hello");
@@ -421,6 +432,22 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
 // Overridable Callbacks
 ///////////////////////////////////////////////////////////////////////////////
 
+private C singleton(C)() if (is(C == class) && is(typeof(new C)))
+{
+    static size_t[1 + (__traits(classInstanceSize, C) - 1) / size_t.sizeof] b;
+    static bool initialized = false;
+    auto result = cast(C) b.ptr;
+    if (!initialized)
+    {
+        assert(typeid(C).init.length == b.length * size_t.sizeof);
+        import core.stdc.string;
+        memcpy(b.ptr, typeid(C).init.ptr, b.length);
+        static if (is(typeof(result.__ctor())))
+            result.__ctor;
+        initialized = true;
+    }
+    return result;
+}
 
 /**
  * A callback for assert errors in D.  The user-supplied assert handler will
@@ -433,9 +460,7 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
  */
 extern (C) void onAssertError( string file = __FILE__, size_t line = __LINE__ ) nothrow
 {
-    if( _assertHandler is null )
-        throw new AssertError( file, line );
-    _assertHandler( file, line, null);
+    onAssertErrorMsg(file, line, null);
 }
 
 
@@ -452,7 +477,13 @@ extern (C) void onAssertError( string file = __FILE__, size_t line = __LINE__ ) 
 extern (C) void onAssertErrorMsg( string file, size_t line, string msg ) nothrow
 {
     if( _assertHandler is null )
-        throw new AssertError( msg, file, line );
+    {
+        auto e = singleton!AssertError;
+        e.msg = msg;
+        e.file = file;
+        e.line = line;
+        throw e;
+    }
     _assertHandler( file, line, msg );
 }
 
--- a/src/core/exception.d
+++ b/src/core/exception.d
@@ -427,27 +427,44 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
     assertHandler = h;
 }
 
+/*
+Returns a singleton object of class type `C`, i.e. repeated calls to
+`singleton!C` within a given thread return the same instance. For that reason,
+no constructor parameters are accepted; `C` must allow default construction.
 
-///////////////////////////////////////////////////////////////////////////////
-// Overridable Callbacks
-///////////////////////////////////////////////////////////////////////////////
+Different threads own different instances of `C`.
 
-private C singleton(C)() if (is(C == class) && is(typeof(new C)))
+The destructor of the singleton object is called during application's exit.
+*/
+private @trusted C singleton(C)() if (is(C == class) && is(typeof(new C)))
 {
     static size_t[1 + (__traits(classInstanceSize, C) - 1) / size_t.sizeof] b;
     static bool initialized = false;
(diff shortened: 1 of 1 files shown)


src/core/exception.d
```diff
@@ -433,9 +460,7 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
  */
 extern (C) void onAssertError( string file = __FILE__, size_t line = __LINE__ ) nothrow
 {
-    if( _assertHandler is null )
-        throw new AssertError( file, line );
-    _assertHandler( file, line, null);
+    onAssertErrorMsg(file, line, null);
```
mathias-lang-sociomantic: That should really be done on the compiler level one day

src/core/exception.d
```diff
@@ -421,6 +432,22 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
 // Overridable Callbacks
 ///////////////////////////////////////////////////////////////////////////////
 
+private C singleton(C)() if (is(C == class) && is(typeof(new C)))
```
mathias-lang-sociomantic: Documentation and unittests please

src/core/exception.d
andralex: @mathias-lang-sociomantic yes, and it will

src/core/exception.d
andralex: unittests are there... documentation not needed for private functions :)

andralex:
Added support for destruction via `atexit`.

andralex:
,,, @mathias-lang-sociomantic and dox :)

tsbockman:
> This makes AssertError a singleton

When I read that, I immediately wondered about the potential race condition if multiple threads were to `assert` simultaneously. Then I saw:

> repeated calls to `singleton!C` **within a given thread** return the same instance

Is it correct to refer to an object as a "*singleton*" when there's actually one per thread? It's very confusing, etymologically speaking.

andralex:
@tsbockman good point thx, changed

src/core/exception.d
WalterBright: why?

src/core/exception.d
```diff
@@ -416,12 +427,48 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
     assertHandler = h;
 }
 
+/*
+Returns a perThreadInstance object of class type `C`, i.e. repeated calls to
```
WalterBright: This is not the standard Ddoc function documentation style.

src/core/exception.d
mathias-lang-sociomantic: @WalterBright : Having two function seems unnecessary, as the case of `assert(cond)` could just be treated as `assert(cond, null)`. That's one less function between DMD and druntime, so less complication.
Also it's a common source of frustration that when debugging we need to break both on `_d_assert` and `_d_assert_msg`.

src/core/exception.d
```diff
@@ -416,12 +427,48 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
...
+    {
+        assert(typeid(C).initializer.length == b.length * size_t.sizeof);
+        import core.stdc.string;
+        () @trusted
+        {
+            memcpy(b.ptr, typeid(C).initializer.ptr, b.length);
+        }();
+        result.__ctor;
+        import core.internal.traits;
+        static if (hasElaborateDestructor!C)
```
mathias-lang-sociomantic: According to a quick test:
```D
import std.traits, core.exception;
static assert(hasElaborateDestructor!AssertError);

// Error: static assert  (hasElaborateDestructor!(AssertError)) is false
```

So it would be nice to have this line covered. Otherwise, LGTM.

andralex: Thanks. I think we can defer coverage of that to the time we get to use this private function elsewhere.

src/core/exception.d
andralex: yah, that's what was happening anyway (with duplicated code)

andralex:
@mathias-lang-sociomantic added coverage

src/core/exception.d
andralex: added nicer dox

dushibaiyu:
I think other Error should be singleton too.

src/core/exception.d
```diff
@@ -416,12 +427,65 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
...
+        }();
+        static if (is(typeof(result.__ctor())))
+            result.__ctor;
+        import core.internal.traits;
+        static if (hasElaborateDestructor!C)
+        {
+            import core.stdc.stdlib;
+            static void destroy()
+            {
+                (cast(C) b.ptr).__dtor;
```
ZombineDev: If you don't want to leverage `object.destroy`, you should at least use `rt_finalize2`, because that's what `object.destroy` [does for classes](https://github.com/dlang/druntime/blob/v2.072.1/src/object.d#L2718).
The problem with calling `obj.__dtor` directly is that:
+ It does a non-virtual call (see [1](https://godbolt.org/g/y2zSnr) and [2](https://godbolt.org/g/e7Am3H)) to `C.~this()`, without calling any of `C`'s base classes' dtors.
+ It does not destroy `C`'s `__monitor` field.
+ It does not clear the memory, causing the GC to attempt destroy (this time "properly") the object again - see [this](http://rextester.com/ROLT52146) (reproduces perfectly on newer compilers too).

src/core/exception.d
```diff
@@ -416,12 +427,65 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
...
+The destructor of the perThreadInstance object is called during application's
+exit.
+
+Returns: A thread-local instance of type `C`.
+
+Throws: If the default constructor of `C` throws.
+*/
+private C perThreadInstance(C)() if (is(C == class) && is(typeof(new C)))
+{
+    static size_t[1 + (__traits(classInstanceSize, C) - 1) / size_t.sizeof] b;
```
ZombineDev: It would have been simpler to do `ubyte[__traits(classInstanceSize, C] b`. Why do you want to round up to a multiple of `size_t`

src/core/exception.d
```diff
@@ -416,12 +427,65 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
...
+    static bool initialized = false;
+    auto result = () @trusted { return cast(C) b.ptr; }();
+    if (!initialized)
+    {
+        assert(typeid(C).initializer.length == b.length * size_t.sizeof);
+        import core.stdc.string;
+        () @trusted
+        {
+            memcpy(b.ptr, typeid(C).initializer.ptr, b.length);
```
ZombineDev: In theory the following could be more efficient `auto ci = typeid(C); b[0 .. ci.initializer.length] = ci.initializer[]` (assuming `b` is of type `ubyte[]`), because the compiler knows at compile time `typeid(C).initializer.length` and can make a better decision about the codegen (e.g. potentially inline the memcpy for smaller types).

src/core/exception.d
```diff
@@ -416,12 +427,65 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
...
+        }();
+        static if (is(typeof(result.__ctor())))
+            result.__ctor;
```
andralex: Hmmm... this is not GC-allocated memory so it does not have block attributes. No?

src/core/exception.d
```diff
@@ -416,12 +427,48 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
...
+Different threads own different instances of `C`.
+
+The destructor of the singleton object is called during application's exit.
+*/
+private C singleton(C)() if (is(C == class) && is(typeof(new C)))
+{
+    static size_t[1 + (__traits(classInstanceSize, C) - 1) / size_t.sizeof] b;
+    static bool initialized = false;
+    auto result = () @trusted { return cast(C) b.ptr; }();
+    if (!initialized)
```
andralex: I'm glad you're giving the same advice I often do :). I'd say when the if is short things are fine either way.

src/core/exception.d
andralex: this code will be executed once per run really

ZombineDev:
Ping @andralex. I need `perThreadInstance` to continue my quest on making more of druntime `@nogc` and `nothrow` (see https://github.com/dlang/druntime/pull/1726 and https://github.com/dlang/druntime/pull/1728).

src/core/exception.d
schveiguy: @andralex correct.

src/core/exception.d
schveiguy: Agree, this was actually a problem with destroy before we fixed it.

src/core/exception.d
```diff
@@ -416,12 +427,65 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
...
+        assert(typeid(C).initializer.length == b.length * size_t.sizeof);
+        import core.stdc.string;
+        () @trusted
+        {
+            memcpy(b.ptr, typeid(C).initializer.ptr, b.length);
+        }();
+        static if (is(typeof(result.__ctor())))
+            result.__ctor;
+        import core.internal.traits;
+        static if (hasElaborateDestructor!C)
```
schveiguy: Hm... I think just remove this static if. There is always a destructor for a class (i.e. rt_finalize2 will always call destructors, even for Object).

schveiguy:
Note, this will close https://issues.dlang.org/show_bug.cgi?id=4587

schveiguy:
> Please use/adopt staticError, it was build for that purpose.

Hm... I think actually `staticError` may be better off using `perThreadInstance`. It solves some problems (i.e. the note about chaining), and uses exact memory size vs. guessing for all possible Error types.

@andralex, looking at how `staticError` works, I think using a `void` static array may be cleaner than `size_t` array. two reasons: 1) no math needed for size, so it looks cleaner. 2) precise scanning may care one day.

src/core/exception.d
klickverbot: To be sufficiently paranoid, better add a check for the instance alignment not being larger than size_t then (SIMD vectors, …).

rainers:
> precise scanning may care one day.

It already does, see https://github.com/dlang/druntime/pull/1710. For the buffer to still be scanned with proper alignment, it should be an array of void*.

rainers:
I think that there should also be a test showing what happens if an assert triggers during unwinding an AssertError. It doesn't have to support perfect chaining, but it should somehow pass the initial error along.

andralex:
superseded

PR #1711, opened 2016-12-12
On range error

andralex:
This also makes RangeError a singleton. We have an additional problem here - this makes onRangeError impure. Hence https://github.com/dlang/phobos/pull/4949.


andralex:
superseded

PR #1712, opened 2016-12-13
Fix missing bypassed chain in Error

yshui:
Issue: https://issues.dlang.org/show_bug.cgi?id=16968
Forum discussion: https://forum.dlang.org/post/o2n7f8$2p1t$1@digitalmars.com

### DIFF
--- a/src/rt/dwarfeh.d
+++ b/src/rt/dwarfeh.d
@@ -373,16 +373,11 @@ extern (C) _Unwind_Reason_Code __dmd_personality_v0(int ver, _Unwind_Action acti
     if (exceptionClass == dmdExceptionClass)
     {
         auto eh = ExceptionHeader.toExceptionHeader(exceptionObject);
+        auto currentLsd = language_specific_data;
+        bool bypassed = false;
         while (eh.next)
         {
             ExceptionHeader* ehn = eh.next;
-
-            // Don't combine when the exceptions are from different functions
-            if (language_specific_data != ehn.languageSpecificData)
-            {
-                //printf("break: %p %p\n", language_specific_data, ehn.languageSpecificData);
-                break;
-            }
 
             Error e = cast(Error)eh.object;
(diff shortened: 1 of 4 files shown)


src/rt/dwarfeh.d
```diff
@@ -373,25 +373,34 @@ extern (C) _Unwind_Reason_Code __dmd_personality_v0(int ver, _Unwind_Action acti
     if (exceptionClass == dmdExceptionClass)
     {
         auto eh = ExceptionHeader.toExceptionHeader(exceptionObject);
+        auto current_lsd = language_specific_data;
```
andralex: use `currentLsd` please

src/rt/dwarfeh.d
```diff
@@ -418,6 +427,14 @@ extern (C) _Unwind_Reason_Code __dmd_personality_v0(int ver, _Unwind_Action acti
             //printf("delete %p\n", ehn);
             _Unwind_DeleteException(&ehn.exception_object); // discard ehn
         }
+        if (bypassed)
+        {
+            eh = ExceptionHeader.toExceptionHeader(exceptionObject);
+
```
andralex: why the whitespace?

test/exceptions/src/chain.d
```diff
@@ -0,0 +1,64 @@
+// Author: Ali Çehreli
+// See https://forum.dlang.org/post/o2n7f8$2p1t$1@digitalmars.com
+
+import core.stdc.stdio;
+
+class TestException : Exception {
```
andralex: follow brace-on-its-line convention please

yshui:
Hmm... I think I still don't understand why the check ```language_specific_data == ehn.languageSpecificData``` is needed

andralex:
I was worried about the naming convention but then I noticed a different convention is used here.

andralex:
Auto-merge toggled on

PR #1713, opened 2016-12-13
simplify _ArrayEq to make it work with the new ctfe engine

UplinkCoder:
The 2 argument foreach is not yet well supported.
In the new ctfe engine.

The following change removes it form the code and makes the codegen more straightforward.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3139,6 +3139,8 @@ bool _ArrayEq(T1, T2)(T1[] a1, T2[] a2)
     if (a1.length != a2.length)
         return false;
 
+    // This is function is used as a compiler intrinsic and explicitly written
+    // in a lowered flavor to use as few CTFE instructions as possible.
     auto idx = 0;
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -3138,9 +3138,12 @@ bool _ArrayEq(T1, T2)(T1[] a1, T2[] a2)
...
     if (a1.length != a2.length)
         return false;
-    foreach(i, a; a1)
+
+    auto length = a1.length;
+
+    while(length--)
     {
-        if (a != a2[i])
+        if (a1[length] != a2[length])
```
MartinNowak: Name it `idx` or `i` please, `[length]` used to mean `[$]`.

MartinNowak:
A straightforward for loop might fulfill the same purpose without reverse iteration, which might not be as fast on all architectures.

UplinkCoder:
@MartinNowak I addressed your comments.

PR #1714, opened 2016-12-14
Fix Issue 4587 - Assert exception should not allocate

Darredevil:
Work in Progress 
https://issues.dlang.org/show_bug.cgi?id=4587

### DIFF
--- a/src/core/exception.d
+++ b/src/core/exception.d
@@ -433,8 +433,15 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
  */
 extern (C) void onAssertError( string file = __FILE__, size_t line = __LINE__ ) nothrow
 {
-    if( _assertHandler is null )
-        throw new AssertError( file, line );
+    import core.stdc.string : memcpy;
+    char[1024] buffer;
+    if( _assertHandler is null ) {
+        enum size = __traits(classInstanceSize, AssertError);
+        memcpy(buffer.ptr, typeid(AssertError).initializer().ptr, size);
+        auto obj = cast(AssertError) buffer.ptr;
+        obj.__ctor( file, line );
+        throw obj;
+    }
     _assertHandler( file, line, null);
 }
 


Darredevil:
PS. I noticed there is another PR opened on this topic (https://github.com/dlang/druntime/pull/1710).
Will discuss this with @andralex before moving forward.

src/core/exception.d
```diff
@@ -433,8 +433,15 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
  */
 extern (C) void onAssertError( string file = __FILE__, size_t line = __LINE__ ) nothrow
 {
-    if( _assertHandler is null )
-        throw new AssertError( file, line );
+    import core.stdc.string : memcpy;
+    char[1024] buffer;
```
lindt: declare this buffer just within the if

1024 is random and a magic number, you could be concrete with char[__traits(classInstanceSize, AssertError)] buffer;

src/core/exception.d
```diff
@@ -433,8 +433,15 @@ deprecated void setAssertHandler( AssertHandler h ) @trusted nothrow @nogc
...
+    if( _assertHandler is null ) {
+        enum size = __traits(classInstanceSize, AssertError);
+        memcpy(buffer.ptr, typeid(AssertError).initializer().ptr, size);
+        auto obj = cast(AssertError) buffer.ptr;
```
lindt: cast is ugly, could emplace be used?

src/core/exception.d
DmitryOlshansky: Not to mention the alignment issue - I don't think there is guarantee that char[x] array is properly aligned.

schveiguy: Not to also mention -- this is a local buffer which you are throwing outside the function. This results in corruption. Make it a static buffer.

src/core/exception.d
schveiguy: emplace is from Phobos, cannot be used.

schveiguy:
I think we should close this in favor of #1710

andralex:
Yah, let's focus on #1710 and friends.

PR #1715, opened 2016-12-15
[REG2.068] Fix Issue 16974 - Equal associative arrays with associative array keys are considered unequal

denis-sh:
Use correct overload of `hashOf` in `_aaGetHash`.

This incorrect usage is possible because of [Issue 16973 - `hashOf` has error-prone signature as `(T, seed)` may be confused with `(ptr, length)`](https://issues.dlang.org/show_bug.cgi?id=16973).


WalterWaldron:
Since this is a regression could you change the branch to ```dlang:stable``` and modify the title to include ```[REG2.068]``` so that it gets proper treatment.

WalterWaldron:
Unrelated failure in phobos, [fixed in ```master```](https://github.com/dlang/phobos/pull/4838)

PR #1717, opened 2016-12-18
fixup for #1602 - deprecate fixed cycle check

MartinNowak:
- only print a deprecation warning for the fixed cycle detection
  cases (cycles over modules without ctor/dtor)

### DIFF
--- a/src/rt/minfo.d
+++ b/src/rt/minfo.d
@@ -16,6 +16,12 @@ import core.stdc.stdlib;  // alloca
 import core.stdc.string;  // memcpy
 import rt.sections;
 
+// Fixing the cycle detection would break some programs relying on that
+// bug. Only print such cycles with a deprecation warning for now.
+// https://issues.dlang.org/show_bug.cgi?id=16211
+// https://github.com/dlang/druntime/pull/1602
+enum deprecationForBugzilla16211 = true;
+
 enum
 {
     MIctorstart  = 0x1,   // we've started constructing it
@@ -168,6 +174,7 @@ struct ModuleGroup
         enum OnCycle
         {
             abort,
+            deprecate,
             print,
             ignore
         }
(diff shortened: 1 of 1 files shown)


schveiguy:
Looks close, but the issue is not that cycles with a non-ctor are a problem, the issue is with a cycle that goes through the same non-ctor module *twice*. The previous algorithm skipped that path.

So for instance with module a and c having ctors, and module b not, the following cycle was previously correctly detected:

a -> b -> c -> a

But this was not:

a -> b -> c -> b -> a.

I think that's the only thing to look for, a module cycle with 2 of the same non-ctor module in it. There could be more cases, but I think it's close enough to be acceptable.

If you want to add this extra detection, I will merge it.

MartinNowak:
That doesn't fix Higgs, I still get another fatal cycle after restricting the deprecation to the first one.

```sh
Deprecated: Cyclic dependency between module ir.inlining and ir.iir
ir.inlining* ->
ir.ast ->
ir.iir* ->
ir.ops ->
jit.ops* ->
ir.ast ->
ir.inlining*
object.Error@src/rt/minfo.d(346): Cyclic dependency between module jit.ops and ir.iir
jit.ops* ->
ir.ast ->
ir.iir* ->
ir.ops ->
jit.ops*
```

The problem was more generally that modules w/o ctors were immediately sorted in before all their deps were checked, so the module could already occur outside of the cycle.
`a -> b* -> a -> c* -> d -> b*`
Doesn't seem like we can detect this from the current implementation. Also running somewhat out of time on this.

MartinNowak:
An additional bitvector deprecatedCtorDone for non-relevant modules would work, but not that elegant.

schveiguy:
There also is the issue of the ordering of the modules. I believe if you reordered the modules, in some cases a cycle was then detected.

I think it's probably inelegant, but we may want to just run the old algorithm if we see any cycles, and act accordingly.

MartinNowak:
> I think it's probably inelegant, but we may want to just run the old algorithm if we see any cycles, and act accordingly.

Yes, that was what I wanted to do, but then had _this_ idea. Would you have a bit time for that?

schveiguy:
I will look into doing this later tonight if I can.

schveiguy:
OK, I think this should do it: #1720

PR #1720, opened 2016-12-22
fixup for cycle detection

schveiguy:
Adds new "deprecate" mode for cycle handling: if cycle is detected, runs original flawed algorithm. If that algorithm finds no cycle, continue execution of the program with old sorted order, but print a deprecation warning to stderr.

Still needs doc update. Ping @MartinNowak


schveiguy:
Ping @Dicebot @CyberShadow

src/rt/minfo.d
```diff
@@ -339,8 +353,7 @@ struct ModuleGroup
                 else
                 {
                     auto midx = edges[sp.curMod][sp.curDep];
-                    // if midx is -1, then this isn't part of this DSO.
-                    if (midx != -1 && !bts(reachable, midx))
+                    if (!bts(reachable, midx))
```
MartinNowak: Seems correct but slightly unrelated (though you prolly double checked for the old implementation).
That means we could change the type of edges to `uint`, right?

MartinNowak:
Thanks, fixes Higgs and vibe-d/libasync.

src/rt/minfo.d
schveiguy: Yes, I realized this was a remnant from the implementation that still used the `findModule` function. I don't put cross-module dependencies as -1 in the `edges` array, so no point in checking for -1 here.

schveiguy:
> fixes Higgs and vibe-d/libasync

curious, I'm assuming you get the deprecation printout?

src/rt/minfo.d
schveiguy: > That means we could change the type of edges to uint, right?

Hm... probably could. I don't think it's too important though, no chance of having over 2 billion modules to sort. I'm leery of unsigned math, too.

PR #1721, opened 2016-12-23
Partial fix for running druntime unittest with PIC=1.

jmdavis:
This fixes it so that the builds in the test folder are built with
-fPIC when the unittest build is run with PIC=1. Unfortunately, the
line_trace test in test/exceptions is then failing (at least on my
system), because it doesn't actually generate a stacktrace. So, this
doesn't seem to completely fix it so that the PIC=1 unittest build
works, but it's close.

### DIFF
--- a/test/common.mak
+++ b/test/common.mak
@@ -14,11 +14,13 @@ SRC:=src
 GENERATED:=./generated
 ROOT:=$(GENERATED)/$(OS)/$(BUILD)/$(MODEL)
 
+OPTIONAL_PIC:=$(if $(PIC),-fPIC,)
+
 ifneq (default,$(MODEL))
 	MODEL_FLAG:=-m$(MODEL)
 endif
 CFLAGS:=$(MODEL_FLAG) -Wall
-DFLAGS:=$(MODEL_FLAG) -w -I../../src -I../../import -I$(SRC) -defaultlib= -debuglib= -dip1000
+DFLAGS:=$(MODEL_FLAG) -w -I../../src -I../../import -I$(SRC) -defaultlib= -debuglib= -dip1000 $(OPTIONAL_PIC)
 # LINK_SHARED may be set by importing makefile
 DFLAGS+=$(if $(LINK_SHARED),-L$(DRUNTIMESO),-L$(DRUNTIME))
 ifeq ($(BUILD),debug)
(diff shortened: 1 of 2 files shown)


test/common.mak
```diff
@@ -14,11 +14,13 @@ SRC:=src
 GENERATED:=./generated
 ROOT:=$(GENERATED)/$(OS)/$(BUILD)/$(MODEL)
 
+OPTIONAL_PIC:=$(if $(PIC),-fPIC,)
```
MartinNowak: I'd prefer to explicitly pass `OPTIONAL_PIC=$(OPTIONAL_PIC)` from the top makefile as is done for the other variables above.
See https://github.com/dlang/druntime/blob/4af3423c2898a1f61014dfbb03e8d8d6f5712cf2/posix.mak#L264-L266.
The fact that explict CLI arguments (like `PIC=1`) are passed to proper submake invocations is fairly subtle to rely on.

andralex: ditto

jmdavis: Okay. I just added it as an argument.

jmdavis:
Rebased.

test/common.mak
Burgos: @jmdavis why not pass `OPTIONAL_PIC` (as @MartinNowak suggested), instead of `PIC`, so the `OPTIONAL_PIC` has to be calculated again?

Burgos:
With my comment applied, or as is, I think we should merge this. I've installed Arch today and spent an hour trying to figure out why druntime unittests are failing - it lead to #1863, but I'll close it now as a duplicate of this. cc @MartinNowak @andralex

test/common.mak
jmdavis: I thought that I had done what was requested, but I could have misunderstood, and my makefile-foo is quite weak.

nemanja-boric-sociomantic:
Also needs rebase now :(

wilzbach:
> Also needs rebase now :(

Rebased.

wilzbach:
Btw as `-fPIC` is the default for dmd on Posix since 2.072, why don't we enable `-fPIC` for druntime as default as well? More an more OSes are moving towards this hardening anyways.
FYI: I also proposed to do this for Phobos: https://github.com/dlang/phobos/pull/5586

PR #1724, opened 2016-12-24
Don't run main after unittests

andralex:
This change in behavior may disrupt some folks but we need to push things forward. It makes no sense to run the program after unittests.

### DIFF
--- a/src/core/runtime.d
+++ b/src/core/runtime.d
@@ -494,7 +494,7 @@ extern (C) bool runModuleUnitTests()
 
     if( Runtime.sm_moduleUnitTester is null )
     {
-        size_t failed = 0;
+        size_t total = 0;
         foreach( m; ModuleInfo )
         {
             if( m )
@@ -503,6 +503,7 @@ extern (C) bool runModuleUnitTests()
 
                 if( fp )
                 {
+                    ++total;
                     try
                     {
                         fp();
@@ -510,12 +511,13 @@ extern (C) bool runModuleUnitTests()
                     catch( Throwable e )
                     {
                         _d_print_throwable(e);
-                        failed++;
                     }
                 }
             }
         }
-        return failed == 0;
+        // Run main if and only if there was no unittest run. This is new
+        // behavior; previously, main would run if all unittests passed.
+        return total == 0;
     }
     return Runtime.sm_moduleUnitTester();
 }


andralex:
I suspect now everybody will love the current behavior :)

andralex:
cc @MartinNowak @CyberShadow @WalterBright

wilzbach:
There is also prior art: https://github.com/dlang/druntime/pull/1685

MartinNowak:
Please figure out a way that does __not__ break every D project out there!

One example, add a runtime switch `--DRT-stop_after_tests=`, and ask people to explicitly make a choice before switching the default. Maybe you could also figure out how to allow a choice for [Runtime.moduleUnitTester - multiple declarations](http://dlang.org/library/core/runtime/runtime.module_unit_tester.html).
Another way might be to not insist on a main function (might be tricky).

Also this needs to be coordinated with what dub and `dub test` do.

WalterBright:
Yeah, it's a good change, but I worry about breaking code and if this is really worth it.

schveiguy:
#1685 does a similar thing in a backwards compatible way.

adamdruppe:
Does anybody seriously care about this kind of "breakage"? Virtually everyone complains about and has numerous hacks to bypass the existing behavior toward what this does, and it is trivial to handle.

adamdruppe:
> This is bullshit

Amen. The other PR (or one of the ones before it, this isn't new) should have been merged without argument a long time ago. We have a chance to make that right: just merge it and close this, or merge this and close it.

Someone just make an executive decision. This isn't worth arguing over and it is so frustrating to see this pattern over and over again, someone does the work, it gets ignored, someone else does the work again, it gets debated pointlessly, forgotten again... then later on someone is guaranteed to say "hey contributors, want change? start doing great work". Now *that* is spit in the face.

Just pick a PR, I don't care which, and merge it. Close the other one immediately. Then you can write your --DRT switch patch on top of it.

andralex:
All: not to worry, this was just to start at the upper bound of changing everything silently and see how much is liable to break. 

@wilzbach @schveiguy cool! didn't know about that work, it's pretty much how I intended to continue this. For backwards compatibility, I'm thinking we could just add a flag `-unittest-only` that when used instead of `-unittest` does not run main. Then we can phase off `-unittest` if necessary.

@adamdruppe the executive decision is we should move forward with this, at best with @schveiguy's work.

@Dicebot could we please tone down the language? Thanks.

Merry Christmas!

andralex:
Closing this to shift focus on https://github.com/dlang/druntime/pull/1685

adamdruppe:
@andralex splendid! Let's see more decisive action taking concrete steps in 2017, I have high hopes for that year.

PR #1725, opened 2016-12-25
CircleCi: use master as fallback for DMD

wilzbach:
Similar to https://github.com/dlang/phobos/pull/4996 - this should make CircleCi more resilient when used with branches that don't exist at DMD / Druntime, so this shouldn't be the usual case.


wilzbach:
I added more fallbacks to this PR as currently done at Phobos: https://github.com/dlang/phobos/pull/5012

PR #1729, opened 2016-12-30
linux/prctl: Added support for prctl.

carun:
On Linux it is possible to perform various operations like setting and getting a process name via `prctl`.

This patch creates a convenience wrapper over GNU/Linux's `prctl` implementation, inspired from `inotify.d`.

Please let me know if we need any improvements on this.

### DIFF
--- a/src/core/sys/linux/sys/prctl.d
+++ b/src/core/sys/linux/sys/prctl.d
@@ -9,6 +9,7 @@ module core.sys.linux.sys.prctl;
 version (linux):
 extern (C):
 @system:
+@nogc:
 nothrow:
 
 enum: uint
@@ -142,4 +143,20 @@ struct prctl_mm_map
     uint     exe_fd;
 };
 
-int prctl(int option, ...);
+int prctl(int option, size_t arg2, size_t arg3, size_t arg4, size_t arg5);
+
+//
+// Example usage to set and get the task name.
+//
+// byte[16] oldname = cast(byte[]) "1234567890123456";
+// oldname[oldname.length-1] = 0;
+// prctl(PR_SET_NAME, cast(size_t) oldname.ptr, cast(size_t) null, cast(size_t) null, cast(size_t) null);
+// byte[16] newname;
+// prctl(PR_GET_NAME, cast(size_t) newname.ptr, cast(size_t) null, cast(size_t) null, cast(size_t) null);
+// int i;
+// foreach (b; newname)
+// {
+//         assert(b == oldname[i]);
+//             i++;
+// }
+// writeln(cast(string) newname);
--- a/src/core/sys/linux/sys/prctl.d
+++ b/src/core/sys/linux/sys/prctl.d
@@ -1,0 +1,145 @@
+/**
+ * D header file for GNU/Linux.
+ *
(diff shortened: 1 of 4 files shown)


src/core/sys/linux/sys/prctl.d
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   Arun Chandrasekaran <aruncxy@gmail.com>
+ */
+module core.sys.linux.sys.prctl;
+
+version (linux):
+extern (C):
+@system:
+nothrow:
```
schveiguy: `@nogc`

src/core/sys/linux/sys/prctl.d
```diff
@@ -0,0 +1,145 @@
...
+    ulong    arg_start;
+    ulong    arg_end;
+    ulong    env_start;
+    ulong    env_end;
+    ulong*   auxv;
+    uint     auxv_size;
+    uint     exe_fd;
+};
+
+int prctl(int option, ...);
```
schveiguy: According to http://man7.org/linux/man-pages/man2/prctl.2.html the prototype should be:
```C
int prctl(int option, unsigned long arg2, unsigned long arg3,
                 unsigned long arg4, unsigned long arg5);
```

Now, I think generally unsigned long turns into uint?

I think the vararg style does not have the same ABI as just regular parameters. You can wrap the C binding elsewhere, but the C binding should reflect what's actually in the library.

carun: > Now, I think generally unsigned long turns into uint?

I think it has to be `ulong` as pointers can be passed as argument. So I am not quite sure about this. Please suggest. Typical, `prctl` usage is as follows:

```d
string name = "12345678901234567890";
prctl(PR_SET_NAME, name.ptr, null, null, null);

byte[16] newname;
prctl(PR_GET_NAME, newname.ptr, null, null, null);
writeln(cast(string) newname);
```
It would print `012345678901234`

carun: Removing the varargs results in below. Is this acceptable? If yes, I can edit the changes. IMHO, this code hurts the eyes with `cast`s.

```d

byte[16] name = cast(byte[]) "1234567890123456";
name[name.length-1] = 0;
prctl(PR_SET_NAME, cast(uint) name.ptr, cast(uint) null, cast(uint) null, cast(uint) null);

byte[16] newname;
prctl(PR_GET_NAME, cast(uint) newname.ptr, cast(uint) null, cast(uint) null, cast(uint) null);

int i;
foreach (b; newname)
{
	assert(b == name[i]);
	i++;
}
writeln(cast(string) newname);

```

schveiguy: If pointers are meant to be passed, then it should be `size_t`, as that is the platform pointer sized word.

Yes, you need to wrap it that way. But you can always write a nice D interface around that!

carun:
Implemented review comments. Please review.

schveiguy:
Thanks LGTM.

schveiguy:
Auto-merge toggled on

PR #1732, opened 2017-01-06
Turn the SECURITY_ enum into a uint.

LemonBoy:
Instead of turning the initializers into `true`/`false` I've just changed its type to `uint`, the values are usually used in bitwise operations and it makes not much sense to keep those as `bool`. I think.

### DIFF
--- a/src/core/sys/windows/winerror.d
+++ b/src/core/sys/windows/winerror.d
@@ -2226,7 +2226,7 @@ enum : HRESULT {
 }
 
 
-enum : bool {
+enum : uint {
     SEVERITY_SUCCESS = 0,
     SEVERITY_ERROR   = 1
 }


ZombineDev:
For reference: https://en.m.wikipedia.org/wiki/HRESULT

jmdavis:
It looks like the continuous integration build is failing for some reason.

schveiguy:
@LemonBoy you need to rebase this, then the jenkins test should pass. I had the same issue.

I think this will work, but not 100% sure, I always have to look it up:
```
git fetch upstream
git rebase upstream/master
```

schveiguy:
Wow, this is such a simple change. May be better to just make a new PR...

wilzbach:
> Wow, this is such a simple change. May be better to just make a new PR...

@schveiguy as Jenkins was disabled for DMD and Phobos, I just disabled it here as well. Ideally now closing & reopening the PR should work.

braddr:
Manually merging due to having long ago exceeded the max number of status updates (1000) a single context can provide.  A direct side effect of a prioritized build that isn't being merged.

braddr:
Manually merging due to exceeding 1000 status updates due to being prioritized for a month without being merged.

PR #1733, opened 2017-01-09
compile with -dip1000

WalterBright:


WalterBright:
Blocked by https://github.com/dlang/dmd/pull/6420

andralex:
Auto-merge toggled on

WalterBright:
Hmm, looks like I have a little more work to do here.

src/rt/config.d
```diff
@@ -58,7 +58,7 @@ import core.stdc.string : strlen;
 
 extern extern(C) string[] rt_args() @nogc nothrow;
 
-alias rt_configCallBack = string delegate(string) @nogc nothrow;
+alias rt_configCallBack = string delegate(return scope string) @nogc nothrow;
```
ZombineDev: You probably also need to update https://github.com/dlang/druntime/blob/v2.073.0-b1/src/gc/config.d#L43

WalterBright:
Now blocked by https://github.com/dlang/dmd/pull/6429

PR #1736, opened 2017-01-13
Workaround Issue 17081 - Bodies in extern cpp functions in D files are not linked

Darredevil:
https://issues.dlang.org/show_bug.cgi?id=17081

### DIFF
--- a/src/core/stdcpp/typeinfo.d
+++ b/src/core/stdcpp/typeinfo.d
@@ -117,8 +117,15 @@ else version (CRuntime_Glibc)
         {
             void dtor1();                           // consume destructor slot in vtbl[]
             void dtor2();                           // consume destructor slot in vtbl[]
-            final const(char)* name();
-            final bool before(const type_info) const;
+            final const(char)* name()() const nothrow {
+                return _name[0] == '*' ? _name + 1 : _name;
+            }
+            final bool before()(const type_info _arg) const {
+                import core.stdc.string : strcmp;
+                return (_name[0] == '*' && _arg._name[0] == '*')
+                    ? _name < _arg._name
+                    : strcmp(_name, _arg._name) < 0;
+            }
             //bool operator==(const type_info) const;
             bool __is_pointer_p() const;
             bool __is_function_p() const;


kinke:
Why do you implement these functions? Isn't adding the `const` qualifier to `name()` enough to fix the linking error? Additionally, you don't check whether `_arg` for `before()` is null, but claim that the function is `nothrow`...

Darredevil:
@kinke adding `const` was not enough to fix the problem, see the bug tracker link for example. Implementing them was the only way to fix the problem. 
As for not checking for null, it's because I translated the exact source code from the GNU stdcpp, however you are correct, the `nothrow` will be removed.

src/core/stdcpp/typeinfo.d
```diff
@@ -117,8 +117,15 @@ else version (CRuntime_Glibc)
         {
             void dtor1();                           // consume destructor slot in vtbl[]
             void dtor2();                           // consume destructor slot in vtbl[]
-            final const(char)* name();
-            final bool before(const type_info) const;
+            final const(char)* name()() const nothrow {
+                return _name[0] == '*' ? _name + 1 : _name;
+            }
+            final bool before()(const type_info _arg) const {
+                import core.stdc.string:strcmp;
```
andralex: spaces around `:`, cc @wilzbach

src/core/stdcpp/typeinfo.d
```diff
@@ -117,8 +117,15 @@ else version (CRuntime_Glibc)
...
+                return (_name[0] == '*' && _arg._name[0] == '*')
+                    ? _name < _arg._name
+                    : strcmp (_name, _arg._name) < 0;
```
andralex: no space before `(` in function call (i.e. use math-like notation), cc @wilzbach

andralex:
BTW this shall not be considered a fix, so I'll change the title.

Darredevil:
@andralex Fixed spacing, waiting final tests for merge

andralex:
Auto-merge toggled on

andralex:
@Darredevil great, thx. Next step would be probably to actually see the problem with https://issues.dlang.org/show_bug.cgi?id=17081. 

Also please push this again, looks like the autotester failed for no cause. cc @MartinNowak @braddr

kinke:
@andralex: Related & trivial, but sitting there for more than a month already and mentioning more C++ interop issues: #1709

PR #1737, opened 2017-01-13
Always use same-type pointers when subtracting them

LemonBoy:
> If both operands are pointers, and the operator is -, the pointers are subtracted and the result is divided by the size of the type pointed to by the operands. **It is an error if the pointers point to different types**. The type of the result is ptrdiff_t.

See DMD's PR [here](https://github.com/dlang/dmd/pull/6440)


DmitryOlshansky:
LGTM

DmitryOlshansky:
Auto-merge toggled on

PR #1740, opened 2017-01-16
string.d: add return and scope

WalterBright:
Taking advantage of `-dip1000`

### DIFF
--- a/src/core/stdc/string.d
+++ b/src/core/stdc/string.d
@@ -29,80 +29,80 @@ nothrow:
 @nogc:
 
 ///
-pure void* memchr(in void* s, int c, size_t n);
+pure void* memchr(return const void* s, int c, size_t n);
 ///
-pure int   memcmp(in void* s1, in void* s2, size_t n);
+pure int   memcmp(scope const void* s1, scope const void* s2, size_t n);
 ///
-pure void* memcpy(void* s1, in void* s2, size_t n);
+pure void* memcpy(return void* s1, scope const void* s2, size_t n);
 version (Windows)
 {
     ///
-    int memicmp(in char* s1, in char* s2, size_t n);
+    int memicmp(scope const char* s1, scope const char* s2, size_t n);
 }
 ///
-pure void* memmove(void* s1, in void* s2, size_t n);
+pure void* memmove(return void* s1, scope const void* s2, size_t n);
 ///
-pure void* memset(void* s, int c, size_t n);
+pure void* memset(return void* s, int c, size_t n);
 
 ///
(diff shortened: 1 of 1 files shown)


WalterBright:
Looks like a job for `inout`.

WalterBright:
@MartinNowak changes made!

src/core/stdc/string.d
```diff
@@ -29,80 +29,80 @@ nothrow:
 @nogc:
 
 ///
-pure void* memchr(in void* s, int c, size_t n);
+pure void* memchr(return const void* s, int c, size_t n);
 ///
-pure int   memcmp(in void* s1, in void* s2, size_t n);
+pure int   memcmp(scope const void* s1, scope const void* s2, size_t n);
```
ZombineDev: So `in` != `scope const`. If I remember correctly, you said that you did that because it would break too much code.
Do you expect that to change? I ask because when I write code I need to know the meaning of `in` and so that my code would be compatible with DIP1000.

src/core/stdc/string.d
```diff
@@ -29,80 +29,80 @@ nothrow:
...
+int    strcoll(scope const char* s1, scope const char* s2);
 ///
-pure int    strncmp(in char* s1, in char* s2, size_t n);
+pure int    strncmp(scope const char* s1, scope const char* s2, size_t n);
 ///
-size_t strxfrm(char* s1, in char* s2, size_t n);
+size_t strxfrm(scope char* s1, scope const char* s2, size_t n);
 ///
-pure char*  strchr(in char* s, int c);
+pure char*  strchr(return char* s, int c);
```
ZombineDev: This should also use `inout`, because the mutability of the return type should depend on the mutability if `s`. It's nice that a problem that C struggled for so much time (for [reference](https://www.lysator.liu.se/c/dmr-on-noalias.html)) with can be solved in D so elegantly.

ZombineDev: BTW, I'm OK with leaving this for another time.

src/core/stdc/string.d
WalterBright: `in` pretty much just means `const`.

src/core/stdc/string.d
WalterBright: I added some `inout`s

src/core/stdc/string.d
ZombineDev: Then https://dlang.org/spec/function.html#parameters needs to be updated.

nordlow: Specifically the `return` storage class is not mentioned.

PR #1741, opened 2017-01-16
Change aliases to new syntax in several files

JackStouffer:
Modernization and clarity.

### DIFF
--- a/src/core/demangle.d
+++ b/src/core/demangle.d
@@ -1980,12 +1980,12 @@ version(unittest)
 
     template staticIota(int x)
     {
-        template Seq(T...){ alias T Seq; }
+        template Seq(T...){ alias Seq = T; }
 
         static if (x == 0)
-            alias Seq!() staticIota;
+            alias staticIota = Seq!();
         else
-            alias Seq!(staticIota!(x - 1), x - 1) staticIota;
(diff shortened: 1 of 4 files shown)


joakim-noah:
FYI, #1002.

JackStouffer:
Martin seems to be in favor of the new syntax.

The area I disagree with him on is the idea that people will just incrementally change the code to the new syntax. Sorry to be a cynic but if it hasn't happened after two years of pull requests it isn't going to happen.

ZombineDev:
> Sorry to be a cynic but if it hasn't happened after two years of pull requests it isn't going to happen.

https://github.com/dlang/druntime/commit/d7581e817b37ccc0f4a502704ddcdb375a3b095a#diff-955ac66e3ac3a2edbe727ac0cf9fba4dR72 :P

Though I agree that it's better to do it once and be done with it.

JackStouffer:
Ping @wilzbach. Do you think you could merge this?

CyberShadow:
What's happening with this useless coverage thing, argh

wilzbach:
> Ping @wilzbach. Do you think you could merge this?

Sorry - totally slipped through my radar. Rebased & auto-merging now, finally :)

PR #1743, opened 2017-01-16
Remove a catch of Throwable in object.d

JackStouffer:
Catching `Throwable` is a bad idea most of the time.

* In the first case, it was a relic that can be safely removed
* ~~In the second case, there's no obvious reason why we would want to catch `Error`s from user code generated by the `TraceInfo.opApply`~~

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -207,23 +207,12 @@ class TypeInfo
         return typeid(this).name;
     }
 
-    override size_t toHash() @trusted const
+    override size_t toHash() @trusted const nothrow
     {
         import core.internal.traits : externDFunc;
         alias hashOf = externDFunc!("rt.util.hash.hashOf",
                                     size_t function(const(void)[], size_t) @trusted pure nothrow @nogc);
-        try
-        {
-            auto data = this.toString();
-            return hashOf(data, 0);
-        }
-        catch (Throwable)
-        {
-            // This should never happen; remove when toString() is made nothrow
-
-            // BUG: this prevents a compacting GC from working, needs to be fixed
-            return cast(size_t)cast(void*)this;
-        }
+        return hashOf(this.toString(), 0);
     }
 
     override int opCmp(Object o)


src/object.d
```diff
@@ -1737,7 +1728,7 @@ class Throwable : Object
                     sink("\n"); sink(t);
                 }
             }
-            catch (Throwable)
+            catch (Exception)
             {
```
wilzbach: Is it even at any time a good idea to silently catch Exceptions?  It seems to me a bit suspicious that `toString` doesn't expose the `nothrow` guarantee publicly, but in the other hand does this silent catching..

JackStouffer: `Exception` implies that the program is still in a recoverable state. `Error`s imply some unrecoverable state.

wilzbach: Yeah I know, but I was trying to say that silently swallowing them as it is done in the existing code isn't really a good practice.

JackStouffer: Ah, ok :)

JackStouffer:
Had to remove the second change as it's error-ing and I don't have time to debug right now.

JackStouffer:
Pinging people with merge rights

@schveiguy @jmdavis @DmitryOlshansky

src/object.d
```diff
@@ -207,23 +207,14 @@ class TypeInfo
...
-        }
-        catch (Throwable)
-        {
-            // This should never happen; remove when toString() is made nothrow
 
-            // BUG: this prevents a compacting GC from working, needs to be fixed
-            return cast(size_t)cast(void*)this;
-        }
+        auto data = this.toString();
+        return hashOf(data, 0);
```
jmdavis: I know that it was like this before, but is there any reason to create a variable here? It seems like a pointless extra line.

JackStouffer:
@jmdavis Fixed

schveiguy:
Just wasted time looking at this, because github doesn't send me notifications of labels being added. Grrr...

wilzbach:
> Just wasted time looking at this, because github doesn't send me notifications of labels being added. Grrr...

@schveiguy: https://github.com/dlang-bots/dlang-bot/issues/32

PR #1745, opened 2017-01-16
style fix - remove wrong usages of the final attribute

BBasile:
Using a [new dscanner checker](https://github.com/Hackerpilot/Dscanner/pull/395), this PR removes several wrong usage of the `final` attribute:
- final method while the current visibility is private.
- final method while the method is actually templatized.

### DIFF
--- a/src/core/stdcpp/typeinfo.d
+++ b/src/core/stdcpp/typeinfo.d
@@ -112,10 +112,12 @@ else version (CRuntime_Glibc)
         {
             void dtor1();                           // consume destructor slot in vtbl[]
             void dtor2();                           // consume destructor slot in vtbl[]
-            final const(char)* name()() const nothrow {
+            final const(char)* name() const nothrow @nogc
+            {
                 return _name[0] == '*' ? _name + 1 : _name;
             }
-            final bool before()(const type_info _arg) const {
+            final bool before(const type_info _arg) const nothrow @nogc
+            {
                 import core.stdc.string : strcmp;
                 return (_name[0] == '*' && _arg._name[0] == '*')
                     ? _name < _arg._name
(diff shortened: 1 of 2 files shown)


src/core/stdcpp/typeinfo.d
```diff
@@ -112,10 +112,10 @@ else version (CRuntime_Glibc)
         {
             void dtor1();                           // consume destructor slot in vtbl[]
             void dtor2();                           // consume destructor slot in vtbl[]
-            final const(char)* name()() const nothrow {
+            final const(char)* name() const nothrow {
```
JackStouffer: `@nogc`

src/core/stdcpp/typeinfo.d
```diff
@@ -112,10 +112,10 @@ else version (CRuntime_Glibc)
...
                 return _name[0] == '*' ? _name + 1 : _name;
             }
-            final bool before()(const type_info _arg) const {
+            final bool before(const type_info _arg) const {
```
JackStouffer: `nothrow @nogc`

src/core/stdcpp/typeinfo.d
BBasile: I don't know, this is not the aim of the PR.

wilzbach:
@BBasile what's the reason for closing this? While PR tend to stall here, usually at some point they manage to find their way into upstream.

src/core/stdcpp/typeinfo.d
```diff
@@ -112,10 +112,12 @@ else version (CRuntime_Glibc)
         {
             void dtor1();                           // consume destructor slot in vtbl[]
             void dtor2();                           // consume destructor slot in vtbl[]
-            final const(char)* name()() const nothrow {
+            final const(char)* name() const nothrow @nogc
```
klickverbot: Weren't these templates on purpose?

PR #1746, opened 2017-01-17
Add pure wrappers for C style allocation

nordlow:
In response to https://github.com/dlang/phobos/pull/4832#issuecomment-272195784

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -809,3 +809,103 @@ struct GC
         gc_runFinalizers( segment );
     }
 }
+
+/**
+ * Pure variants of C's memory allocation functions `malloc`, `calloc` and
+ * `realloc`. Purity is achieved via resetting the `errno` to it's value prior
+ * to being called, removing the function's global state mutation.
+ *
+ * See_Also:
+ *     $(LINK2 https://dlang.org/spec/function.html#pure-functions, D's rules for purity),
+ *     which allow for memory allocation under specific circumstances.
+ */
+void* pureMalloc(size_t size) @trusted pure @nogc nothrow
+{
+    const errno = fakePureGetErrno();
+    void* ret = fakePureMalloc(size);
+    if (!ret || errno != 0)
+    {
+        cast(void)fakePureSetErrno(errno);
+    }
+    return ret;
+}
+/// ditto
+void* pureCalloc(size_t nmemb, size_t size) @trusted pure @nogc nothrow
+{
+    const errno = fakePureGetErrno();
+    void* ret = fakePureCalloc(nmemb, size);
+    if (!ret || errno != 0)
+    {
+        cast(void)fakePureSetErrno(errno);
+    }
+    return ret;
+}
+/// ditto
+void* pureRealloc(void* ptr, size_t size) pure @nogc nothrow
+{
+    const errno = fakePureGetErrno();
+    void* ret = fakePureRealloc(ptr, size);
+    if (!ret || errno != 0)
+    {
+        cast(void)fakePureSetErrno(errno);
+    }
+    return ret;
+}
+
+///
+nothrow @nogc unittest
+{
+    import core.stdc.stdlib : free;
+
+    ubyte[] fun(size_t n) pure
+    {
+        void* p = pureMalloc(n);
+        p !is null || n == 0 || assert(0);
+        scope(failure) p = pureRealloc(p, 0);
+        p = pureRealloc(p, n *= 2);
+        p !is null || n == 0 || assert(0);
+        return cast(ubyte[]) p[0 .. n];
+    }
+
+    auto buf = fun(100);
+    assert(buf.length == 200);
+    free(buf.ptr);
+}
+
+pure @nogc nothrow unittest
+{
+    const int errno = fakePureGetErrno();
+
+    void* x = pureMalloc(10);            // normal allocation
+    assert(errno == fakePureGetErrno()); // errno shouldn't change
+    assert(x !is null);                   // allocation should succeed
+
+    x = pureRealloc(x, 10);              // normal reallocation
+    assert(errno == fakePureGetErrno()); // errno shouldn't change
+    assert(x !is null);                   // allocation should succeed
+
+    fakePureFree(x);
+
+    void* y = pureCalloc(10, 1);         // normal zeroed allocation
+    assert(errno == fakePureGetErrno()); // errno shouldn't change
+    assert(y !is null);                   // allocation should succeed
+
+    fakePureFree(y);
+
+    void* z = pureMalloc(size_t.max); // won't affect `errno`
+    assert(errno == fakePureGetErrno()); // errno shouldn't change
+    assert(z is null);
+}
+
+// locally purified for internal use here only
+extern (C) private pure @system @nogc nothrow
+{
+    pragma(mangle, "getErrno") int fakePureGetErrno();
+    pragma(mangle, "setErrno") int fakePureSetErrno(int);
+
+    pragma(mangle, "malloc") void* fakePureMalloc(size_t);
+    pragma(mangle, "calloc") void* fakePureCalloc(size_t nmemb, size_t size);
+    pragma(mangle, "realloc") void* fakePureRealloc(void* ptr, size_t size);
+
+    pragma(mangle, "free") void fakePureFree(void* ptr); // needed by unittests
+}


src/core/stdc/stdlib.d
```diff
@@ -159,6 +159,17 @@ else
 ///
 void*   malloc(size_t size) pure;
 ///
+void*   pureMalloc(size_t size) pure
+{
+    const errno = pureGetErrno();
+    typeof(return) ret = malloc(size);
```
ZombineDev: Why not use `void*` or `auto` here? `malloc`'s return type isn't going to change anytime soon .

src/core/stdc/stdlib.d
```diff
@@ -159,6 +159,17 @@ else
...
+    cast(void)pureSetErrno(errno);
+    return ret;
+}
+extern (C) pure pragma(mangle, "getErrno") int pureGetErrno(); // for internal use
+extern (C) pure pragma(mangle, "setErrno") int pureSetErrno(int); // for internal use
```
ZombineDev: Checking in with LDC developers @kinke @JohanEngelen, does this work on LDC?

BTW, I bumped into a related `extern (C)` issue on LDC - https://github.com/dlang/druntime/pull/1728#discussion_r96115755

ZombineDev: Also, since those functions are for internal use, why not make them `private` or `package`, or declare them inside `pureMalloc`? They should be used only in a controlled manner, and reviewed case-by-case.

JackStouffer:
Andrei stated this should go in core/memory.d

src/core/stdc/stdlib.d
JackStouffer: > Also, since those functions are for internal use,

I think the plan is to make this public

kinke: > does this work on LDC?

Multiple `extern (C)` forward-declarations with the same mangled name aren't a problem. [This](http://ldc.acomirei.ru/#compilers:!((compiler:ldc,options:'',source:'extern+(C)+int+foo()%3B%0A%0Apragma+(mangle,+%22foo%22)%0Aextern+(C)+int+bar()+pure%3B%0A%0Avoid+main()%0A%7B%0A++foo()%3B%0A++bar()%3B%0A%7D%0A')),filterAsm:(commentOnly:!t,directives:!t,labels:!t),version:3) works.

nordlow: So are you saying I should rename these declarations to `getErrno()` and `setErrno()`?

src/core/stdc/stdlib.d
nordlow: Change return type to `void*`.

src/core/stdc/stdlib.d
kinke: Well just saying that https://github.com/ldc-developers/ldc/issues/1020 doesn't apply to these `extern (C)` fwd declarations. While declaring them with their real name should work too, I think the explicit `pure` in the function name makes a lot of sense in this case.

nordlow: Ok. I keep it as that for now.

nordlow:
So, should I move it to `core/memory.d`, @andralex?

src/core/stdc/stdlib.d
nordlow: This currently fails to compile as 

```
stdlib.d(166,23): Error: pure function 'core.stdc.stdlib.pureMalloc' cannot call impure function 'core.stdc.stdlib.malloc'
```

How do I fix that?

nordlow:
Should I add a similar wrapper for `calloc` and `realloc`?

JackStouffer:
> Should I add similar wrappers for calloc and realloc?

Yes

> So, should I move it to core/memory.d

https://github.com/dlang/druntime/pull/1735#issuecomment-272656349

nordlow:
Done.

kinke:
What about potential race conditions when restoring `errno`? The [docs here](http://www.cplusplus.com/reference/cstdlib/malloc/) don't mention `errno` at all, instead:

> Data races: Only the storage referenced by the returned pointer is modified. No other storage locations are accessed by the call.

nordlow:
Yeah, I thought of that too :) @andralex ?

JackStouffer:
Pretty sure errno is thread local.

nordlow:
http://stackoverflow.com/questions/1694164/is-errno-thread-safe#1694170

src/core/memory.d
```diff
@@ -809,3 +809,41 @@ struct GC
         gc_runFinalizers( segment );
     }
 }
+
+/// Pure variant of `malloc` that doesn't affect global variable `errno`.
+void* pureMalloc(size_t size) pure
+{
+    const errno = _pureGetErrno();
+    void* ret = _pureMalloc(size);
+    cast(void)_pureSetErrno(errno);
```
andralex: space after closing paren in cast - cc @wilzbach

src/core/memory.d
```diff
@@ -809,3 +809,41 @@ struct GC
         gc_runFinalizers( segment );
     }
 }
+
+/// Pure variant of `malloc` that doesn't affect global variable `errno`.
+void* pureMalloc(size_t size) pure
+{
+    const errno = _pureGetErrno();
```
andralex: Does druntime use this convention of prefixing private symbols with one underscore?

src/core/memory.d
```diff
@@ -809,3 +809,41 @@ struct GC
...
+// locally purified for internal use here only
+extern (C) private pure
+{
+    pragma(mangle, "getErrno") int _pureGetErrno();
+    pragma(mangle, "setErrno") int _pureSetErrno(int);
+
+    pragma(mangle, "malloc") void* _pureMalloc(size_t);
+    pragma(mangle, "calloc") void* _pureCalloc(size_t nmemb, size_t size);
+    pragma(mangle, "realloc") void* _pureRealloc(void* ptr, size_t size);
+}
```
andralex: please add one unittest to cover all, or at best short ddoc unittests showing the use of each inside a pure function

andralex:
OK, this is good progress. Thanks!

andralex:
One idea for documentation: simply group all three together with documentation such as "Pure wrappers for the C stdlib memory allocation primitives `malloc`, `calloc`, and `realloc`." Then you use one ddoc unittest featuring all. That way you get nice compact documentation and one unifying example.

andralex:
I suspect `pureCalloc` is safe, too. Awesomeness.

src/core/memory.d
```diff
@@ -809,3 +809,41 @@ struct GC
         gc_runFinalizers( segment );
     }
 }
+
+/// Pure variant of `malloc` that doesn't affect global variable `errno`.
```
JackStouffer: Please replace this with the following:

```
/**
 * Pure variants of C's memory allocation functions. Purity is achieved via
 * resetting the `errno` to it's value prior to being called, removing the
 * function's global state mutation.
 *
 * See_Also:
 *     $(LINK2 https://dlang.org/spec/function.html#pure-functions, D's rules for purity),
 *     which allow for memory allocation under specific circumstances.
 */
```

IMO it addresses the intuitiveness of a pure memory allocation function for the reader.

andralex: The teleological description is better than the description in terms of implementation details. I.e. if a platform offered access to an errno-less malloc (as e.g. would be easy to hack with jemalloc), pureMalloc would use that.

src/core/memory.d
nordlow: Renamed it.

src/core/memory.d
nordlow: Done.

src/core/memory.d
```diff
@@ -810,25 +810,31 @@ struct GC
...
+ *
+ * See_Also:
+ *     $(LINK2 https://dlang.org/spec/function.html#pure-functions, D's rules for purity),
+ *     which allow for memory allocation under specific circumstances.
+ */
 void* pureMalloc(size_t size) @trusted pure
 {
     const errno = fakePureGetErrno();
     void* ret = fakePureMalloc(size);
     cast(void)fakePureSetErrno(errno);
```
andralex: I'd say it's more efficient to guard the call like this:

```d
if (!ret || errno != ERR_OK) cast(void)fakePureSetErrno(errno);
```

The common case is malloc succeeds and the previous state was ERR_OK so no need to reset errno.

nordlow:
Ready for squashing?

andralex:
lgtm

nordlow:
Squashed.

nordlow:
BTW: Is manual squashing still needed by PR authors?

andralex:
>BTW: Is manual squashing still needed by PR authors?

I forgot. cc @MartinNowak @CyberShadow

andralex:
One more thing - please add a doc unittest featuring the use of all three. Thanks!

wilzbach:
>> BTW: Is manual squashing still needed by PR authors?
> I forgot. cc @MartinNowak @CyberShadow

GitHub squashing is already enabled and the dlang bot supports squashed merges. @CyberShadow had some concerns related to digger and PR number inference, but IIRC they have been resolved as the PR number is part of the commit message?

nordlow:
How should the three functions be tested?

CyberShadow:
> I heard somewhere that Github had added an automatic squash option upon merge.

It does something a little different (specifically, doesn't create a merge commit).

> @CyberShadow had some concerns related to digger and PR number inference, but IIRC they have been resolved as the PR number is part of the commit message?

I *think* it should be OK now, but I'm not 100% sure. In any case, please never use "Rebase and merge". We should disable that for all repositories (and consider enabling "Squash and merge" instead).

CyberShadow:
> GitHub squashing is already enabled and the dlang bot supports squashed merges.

Is this something other than "Squash and merge"? Because "Squash and merge" is clearly disabled, at least for druntime/phobos.

The problem with "Squash and merge" is that the operation is destructive. This is fine when the change is small and all the commits after the first one are fixups. So, when commits are not squashed when they should have been, we get some untidy history in the repo. However, when the commits are squashed when they shouldn't have been, information about the commits is lost, which makes blaming, bisecting, and generally examining history more difficult. So, I think that enabling "Squash and merge" would be fine if we knew that it would always be used properly.

andralex:
@nordlow you're overthinking it :). The unittest as written is not suitable for documentation. User code cannot call the private undocumented function `fakeGetErrno`. Also doc unittests don't show their attribute so the example isn't clear.

Just add a simple example of a pure function using these primitives and you're good. E.g.:

```d
///
unittest
{
    pure ubyte[] fun(size_t n)
    {
        void* p = pureMalloc(n);
        p !is null || n == 0 || assert(0);
        scope(failure) pureFree(p);
        // ...
        pureRealloc(p, n *= 2);
        p !is null || n == 0 || assert(0);
        // ...
        return cast(ubyte[]) p[0 .. n];
    }
    auto buf = fun(100);
    assert(buf.length == 200);
    pureFree(buf.ptr);
}

**Comment by nordlow** [comment]
Should I still keep my unittest as undocumented along with yours?

File: src/core/memory.d [hunk 8]
```diff
@@ -809,3 +809,113 @@ struct GC
...
+    const errno = fakePureGetErrno();
+    void* ret = fakePureRealloc(ptr, size);
+    if (!ret || errno != 0)
+    {
+        cast(void)fakePureSetErrno(errno);
+    }
+    return ret;
+}
+
+/**
```
andralex: Please group documentation with all others.

src/core/memory.d
```diff
@@ -809,3 +809,113 @@ struct GC
...
+ * Pure variant of C's memory deallocation function `free`.
+ */
+void pureFree(void* ptr) pure @nogc nothrow
+{
+    fakePureFree(ptr);
```
andralex: Now I presume there's no need for the additional wrapper? Just rename fakePureFree to pureFree and move it here.

andralex:
I really want to merge this and have us all move on with our respective lives, but now we run afoul of @schveiguy's dire predictions on how bad `pureFree` is. I was wrong to use it in the example, let's just not use it and get this over with. Thanks.

nordlow:
Ok, one more round, @and

src/core/memory.d
```diff
@@ -809,3 +809,106 @@ struct GC
...
+    }
+    return ret;
+}
+
+///
+nothrow @nogc unittest
+{
+    import core.stdc.stdlib : free;
+
+    ubyte[] fun(size_t n)
```
andralex: Add `pure` here, otherwise the example is confusing as heck. Check http://dtest.thecybershadow.net/artifact/website-82f057bfd26992f61d739be5df65183831e83aa2-fe5ae0cbeba94c4622d946034746e8f7/web/library-prerelease/core/memory/pure_malloc.html to see what it looks like right now.

src/core/memory.d
```diff
@@ -809,3 +809,106 @@ struct GC
...
+    {
+        void* p = pureMalloc(n);
+        assert(n == 0 || p !is null);
```
andralex: The way I'd written it, the `assert` will halt the program. This version has undefined behavior in release mode if `malloc` returns null.

src/core/memory.d
```diff
@@ -809,3 +809,106 @@ struct GC
...
+
+        scope(failure) free(p);
+
+        p = pureRealloc(p, n *= 2);
+        assert(n == 0 || p !is null);
```
andralex: ditto

src/core/memory.d
```diff
@@ -809,3 +809,106 @@ struct GC
...
+///
+nothrow @nogc unittest
+{
+    import core.stdc.stdlib : free;
+
+    ubyte[] fun(size_t n)
+    {
+        void* p = pureMalloc(n);
+        assert(n == 0 || p !is null);
+
```
andralex: I'd inserted `// ...` here to suggest some operations may fail. So please either insert those or (probably better) use `enforce(n == 0 || p !is null)` instead of any assert. That way we throw a legit exception if allocation fails, and the scope statement has a good purpose.

src/core/memory.d
nordlow: Cool, how did you find that link?

nordlow: We cannot declare `fun pure` because `stdlib.free` is not pure...

src/core/memory.d
nordlow: Done.

src/core/memory.d
nordlow: Done.

src/core/memory.d
```diff
@@ -809,3 +809,107 @@ struct GC
...
+unittest
+{
+    import std.exception : enforce;
+    import core.stdc.stdlib : free;
+
+    ubyte[] fun(size_t n) // TODO pure?
+    {
+        void* p = pureMalloc(n);
+        enforce(n == 0 || p !is null);
+
```
andralex: Vertical space in a five-liner is a net negative (ironically your implementations themselves don't have useless vertical space, and indeed they shouldn't). There is no value added by it.

src/core/memory.d
```diff
@@ -809,3 +809,107 @@ struct GC
...
+        scope(failure) free(p);
```
andralex: Use `p = pureRealloc(p, 0);` to free memory in a pure manner.

src/core/memory.d
```diff
@@ -809,3 +809,107 @@ struct GC
...
+    return ret;
+}
+
+///
+unittest
+{
+    import std.exception : enforce;
+    import core.stdc.stdlib : free;
+
+    ubyte[] fun(size_t n) // TODO pure?
```
andralex: one way or another the example must be pure!

src/core/memory.d
```diff
@@ -809,3 +809,107 @@ struct GC
...
+    {
+        void* p = pureMalloc(n);
+        enforce(n == 0 || p !is null);
+
+        scope(failure) free(p);
+
+        p = pureRealloc(p, n *= 2);
+        enforce(n == 0 || p !is null);
+
```
andralex: space, the final frontier - let's not

nordlow: Done.

src/core/memory.d
nordlow: Done.

src/core/memory.d
nordlow: Done.

src/core/memory.d
nordlow: Done.

src/core/memory.d
```diff
@@ -809,3 +809,103 @@ struct GC
...
+    }
+    return ret;
+}
+
+///
+nothrow @nogc unittest
+{
+    import core.stdc.stdlib : free;
+
+    ubyte[] fun(size_t n) // TODO pure?
```
andralex: make pure

nordlow: Done.

nordlow:
Needed to use assert instead. enforce is in Phobos :)

src/core/memory.d
```diff
@@ -809,3 +809,103 @@ struct GC
...
+    {
+        void* p = pureMalloc(n);
+        p !is null || n == 0 || assert(0);
+        scope(failure) { p = pureRealloc(p, 0); }
```
andralex: braces are redundant - think of an inlined `if`, would one put braces?

nordlow: Done.

nordlow:
Ok no, @andralex ?

andralex:
thx!

nordlow:
Great!

nordlow:
Why is pureFree disallowed when pureRealloc can be used to free aswell?

andralex:
Ask @schveiguy :)

nordlow:
@schveiguy ?

schveiguy:
It's a good question. I would question it the other way, why is `pureRealloc` allowed when `pureFree` is not.

Technically, `realloc(ptr, 0)` is essentially the same as `free(ptr)`. However, one is not likely to `realloc` without wanting to re-assign to the original pointer. That is, if one has a function as has been discussed like:

```D
void foo(immutable int *ptr) pure { ... }
```

Using `realloc` in such a function doesn't make a whole lot of sense, since you potentially destroy the original pointer, and have no way to inform the caller of the new location.

So while technically `realloc` can be used as a crude instrument to achieve pure `free`, the likelihood that someone does this is quite low. I would say you are misusing `realloc` if you are doing it that way.

IMO, `realloc` is ok to mark `pure`.

rainers:
There are no auto-tester builds for druntime anymore?
This PR broke master because the Digital Mars malloc is able to allocate size_t.max bytes! (In fact it just adds 2 to the size before calling the Windows API functions.)

CyberShadow:
Wow, the autotester did not report any status for the last commit.

I'm going to enable the autotester being green as a requirement for merging Druntime PRs.

nordlow:
Oops, sorry about that. Should we just remove this test case, @andralex?

CyberShadow:
Done. @rainers Can you code up a fix or should we revert?

nordlow:
I can remove the part where `size_t.max` is allocated.

rainers:
Subtracting 2 works, too: https://github.com/dlang/druntime/pull/1754

nordlow:
Thanks.

schveiguy:
> Wow, the autotester did not report any status for the last commit.

The merge was done without auto-tester requirement (i.e. merge button was pushed, not auto-merge).

> I'm going to enable the autotester being green as a requirement for merging Druntime PRs.

This is probably a good requirement, but I'd much rather the auto-merge feature be used, since there is usually a lag between some other merge happening and github deciding that it's no longer green.

CyberShadow:
I see... so the problem is that @andralex did not use auto-merge.

> This is probably a good requirement, but I'd much rather the auto-merge feature be used, since there is usually a lag between some other merge happening and github deciding that it's no longer green.

We might be able to fix this by enabling the GitHub-side requirement that PR branches can only be merged if they are rebased off the latest commit of the target branch, and have the autotester do that rebase on top of the target branch commit it tested with before attempting to merge the PR. Not sure it would be worth the effort, though - have we ever had breakages due to this?

andralex:
I saw "All tests passed" so I merged directly. It looks like a race condition.

schveiguy:
> It looks like a race condition.

Yes, this is the lag I spoke of. github can take some time to sync with the status of the auto tester, or the auto tester could possibly not report the "invalid" status quickly enough. In any case, using the auto-tester auto-merge feature is the safest path -- if all tests have passed, it will merge immediately, so it's no different.

PR #1749, opened 2017-01-22
core.stdc.stdlib: annotate with 'scope'

WalterBright:
For -dip1000. Also corrects problem where `strtold` can be used to convert pointer to immutable to pointer to mutable.

### DIFF
--- a/src/core/stdc/stdlib.d
+++ b/src/core/stdc/stdlib.d
@@ -32,11 +32,11 @@ extern (C):
 /* Placed outside @nogc in order to not constrain what the callback does.
  */
 ///
-alias int function(in void*, in void*) _compare_fp_t;
+alias int function(scope const void*, scope const void*) _compare_fp_t;
 ///
-void*   bsearch(in void* key, in void* base, size_t nmemb, size_t size, _compare_fp_t compar);
+inout(void)* bsearch(scope const void* key, scope inout(void)* base, size_t nmemb, size_t size, _compare_fp_t compar);
 ///
-void    qsort(void* base, size_t nmemb, size_t size, _compare_fp_t compar);
+void    qsort(scope void* base, size_t nmemb, size_t size, _compare_fp_t compar);
 
 
 nothrow:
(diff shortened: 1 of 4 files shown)


andralex:
Auto-merge toggled on

src/core/stdc/stdlib.d
```diff
@@ -81,55 +81,55 @@ else version(CRuntime_Bionic) enum RAND_MAX = 0x7fffffff;
...
 ///
-ulong   strtoull(in char* nptr, char** endptr, int base);
+ulong   strtoull(scope inout(char)* nptr, scope inout(char)** endptr, int base);
 
 version (CRuntime_Microsoft)
 {
     // strtold exists starting from VS2013, so we make this a template to avoid link errors
     ///
-    real strtold()(in char* nptr, char** endptr)
+    real strtold()(scope inout(char)* nptr, inout(char)** endptr)
```
MartinNowak: Fails on Win64 w/
```
std\conv.d(3163): Error: template core.stdc.stdlib.strtold cannot deduce function from argument types !()(string, typeof(null)), candidates are:
..\druntime\import\core\stdc\stdlib.d(109):        core.stdc.stdlib.strtold()(scope inout(char)* nptr, inout(char)** endptr)
```

src/core/stdc/stdlib.d
```diff
@@ -81,55 +81,55 @@ else version(CRuntime_Bionic) enum RAND_MAX = 0x7fffffff;
...
+double  atof(scope const char* nptr);
 ///
-int     atoi(in char* nptr);
+int     atoi(scope const char* nptr);
 ///
-c_long  atol(in char* nptr);
+c_long  atol(scope const char* nptr);
 ///
-long    atoll(in char* nptr);
+long    atoll(scope const char* nptr);
```
MartinNowak: It was circulated for quite a while that `in` is supposed to mean `scope const`, in the compiler it was always just `const` but lots of people have used it as `scope const`. Have you thought about actually changing the semantic of `in`?

jmdavis: http://forum.dlang.org/post/o6m17i$1jh7$1@digitalmars.com

src/core/stdc/stdlib.d
```diff
@@ -81,55 +81,55 @@ else version(CRuntime_Bionic) enum RAND_MAX = 0x7fffffff;
...
 
 ///
-double  strtod(in char* nptr, char** endptr);
+double  strtod(scope inout(char)* nptr, scope inout(char)** endptr);
 ///
-float   strtof(in char* nptr, char** endptr);
+float   strtof(scope inout(char)* nptr, scope inout(char)** endptr);
 ///
-c_long  strtol(in char* nptr, char** endptr, int base);
+c_long  strtol(scope inout(char)* nptr, scope inout(char)** endptr, int base);
```
rainers: This seems to have broken the travis builds for dmd, e.g. https://travis-ci.org/dlang/dmd/jobs/201062823#L5271

The change seems ok, but it breaks code that tried to work around the wrong annotation. The problem with fixing dmd is that it is built with both dmd 2.073 and git-master...

WalterBright: The error message:
```
mars.d(420): Error: function core.stdc.stdlib.strtol (scope inout(char)* nptr, scope inout(char)** endptr, int base) is not callable using argument types (const(char)*, char**, int)
```
This is the problem I was trying to correct. The old declaration allowed a const pointer to be converted to a mutable one. Any code that relies on such needs to be fixed.

rainers: Sure. dmd doesn't rely on the old signature, it had to workaround it with a cast. See https://github.com/dlang/dmd/pull/6539

rainers: I suspect we need to readd the old signatures as deprecated, otherwise every other project will fail to compile.

MartinNowak:
It says `@system` at the top of the module.
https://github.com/WalterBright/druntime/blob/fe160cb22d035f21f5bb39ec86553728bec45f91/src/core/stdc/stdlib.d#L30
So all of those APIs are marked as unsafe and the compiler won't check scope for them anyhow :o.

PR #1750, opened 2017-01-26
Add bindings for POSIX getdelim and getline

nemanja-boric-sociomantic:


src/core/sys/posix/stdio.d
```diff
@@ -344,3 +344,7 @@ unittest
     assert(memcmp(ptr, testdata.ptr, testdata.length*wchar_t.sizeof) == 0);
     assert(fclose(f) == 0);
 }
+
+
+ssize_t getdelim (char** lineptr, size_t* n, int delimiter, FILE* stream);
+ssize_t getline (char** lineptr, size_t* n, FILE* stream);
```
ZombineDev:
```
ssize_t getdelim(scope char** lineptr, scope size_t* n, int delimiter, FILE* stream);
ssize_t getline(scope char** lineptr, scope size_t* n, FILE* stream);
```

Adding `scope` will allow passing pointers to local variables in `@safe` code. See also #1740, #1748 and #1749.

nemanja-boric-sociomantic: Ah, that's a new one ;-). I'll look into it and add it, thanks!

nemanja-boric-sociomantic: I obviously need to do a reading on DIP 1000. I've updated the PR, but this question is making me wonder:

```
void bar(scope char** c, scope int* n) @trusted
{

}

void foo() @safe
{
    char[100] buf;
    char* c = &buf[0];
    int n;

    bar(&c, &n);
}

void main()
{
    foo();
}
```

```
test.d(12): Error: cannot take address of scope local c in @safe function foo
```

Moving `char[100] buf` ouside of a scope works.

ZombineDev: Using DMD 2.073.0 and compiling with `-dip1000` I got the following results:
```
void main() @safe
{
    char[100] buf;
    char* c;
    int n;

    bar(&c, &n); // Error
}

void bar(char** c, int* n) @safe
{
}
```

> dip1000.d(7): Error: reference to local variable c assigned to non-scope parameter c calling dip1000.bar
dip1000.d(7): Error: reference to local variable n assigned to non-scope parameter n calling dip1000.bar

```
void main() @safe
{
    char[100] buf;
    char* c;
    int n;

    bar(&c, &n); // OK
}

void bar(scope char** c, scope int* n) @safe
{
}
```

> [No errors]

```
void main() @safe
{
    char[100] buf;
    char* c = &buf[0];
    int n;

    bar(&c, &n);
}

void bar(scope char** c, scope int* n) @safe
{
}
```

> dip1000.d(7): Error: cannot take address of scope local c in @safe function main

(same result as yours)

So while DIP1000 is not a 100% solution (yet), using `scope` is still an improvement for `@safe` code, because it allows to pass pointers to *some* local variables.

ZombineDev: I just saw the [forum post](http://forum.dlang.org/thread/iioxwplvezvfsrjfdtwh@forum.dlang.org) (thanks for bringing this to Walter's attention) and I was wondering if we should change the declarations to:

```D
ssize_t getdelim(ref scope char* lineptr, scope size_t* n, int delimiter, FILE* stream);
ssize_t getline(ref scope char* lineptr, scope size_t* n, FILE* stream);
```

relying on the fact that the calling convention would remain the same. @WalterBright  do you think this is a good idea?

nemanja-boric-sociomantic:
Ping :-)

src/core/sys/posix/stdio.d
nemanja-boric-sociomantic: Ping @WalterBright

MartinNowak: No, `scope char** lineptr` is correct here, the function doesn't escape the value of the `char**` pointer passed in. In any case it hardly seems to matter b/c those functions are inherently unsafe, after the call `*lineptr` will point to some internal buffer of the stream.

nemanja-boric-sociomantic:
Thanks a lot!

PR #1752, opened 2017-01-29
[scope] add 'scope' to delegates

WalterBright:
This is to resolve the following error messages:
```
src/rt/dmain2.d(507): Error: none of the overloads of 'toString' are callable using argument types (void delegate(const(char[]) s) nothrow), candidates are:
src/object.d(1701):        object.Throwable.toString()
src/object.d(1714):        object.Throwable.toString(scope void delegate(const(char[])) scope sink)
src/rt/dmain2.d(499): Error: none of the overloads of 'toString' are callable using argument types (void delegate(const(char[]) s) nothrow), candidates are:
src/object.d(1701):        object.Throwable.toString()
src/object.d(1714):        object.Throwable.toString(scope void delegate(const(char[])) scope sink)
```
from compiling with -dip1000.

### DIFF
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -492,7 +492,7 @@ extern (C) int _d_run_main(int argc, char **argv, MainFunc mainFunc)
     return result;
 }
 
-private void formatThrowable(Throwable t, void delegate(in char[] s) nothrow sink)
+private void formatThrowable(Throwable t, scope void delegate(in char[] s) nothrow sink)
 {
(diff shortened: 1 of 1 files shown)


WalterBright:
@MartinNowak ping

JackStouffer:
@WalterBright ZombieDev, while his/her reviews are very appreciated, is not a DMD member. Please don't merge unless the review check mark is green. It's grey when the user is not a member.

PR #1753, opened 2017-01-29
[scope] add 'scope' to delegates

WalterBright:
This is to resolve the following -dip1000 generated errors:
```
src/gc/impl/conservative/gc.d(2150): Error: function core.thread.thread_scanAll (scope void delegate(void*, void*) nothrow scope scan) is not callable using argument types (void delegate(void* pbot, void* ptop) nothrow)
src/gc/impl/conservative/gc.d(2397): Error: function core.thread.thread_processGCMarks (scope int delegate(void* addr) nothrow scope isMarked) is not callable using argument types (int delegate(void* addr) nothrow)
```
I.e. a non-scope delegate cannot be used where a scope delegate is expected.

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -1952,7 +1952,7 @@ struct Gcx
     /**
      * Search a range of memory values and mark any pointers into the GC pool.
      */
-    void mark(void *pbot, void *ptop) nothrow
+    void mark(void *pbot, void *ptop) scope nothrow
     {
         void **p1 = cast(void **)pbot;
         void **p2 = cast(void **)ptop;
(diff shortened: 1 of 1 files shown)


WalterBright:
Auto-merge toggled on

WalterBright:
Auto-merge toggled off

WalterBright:
@martinnowak ping

MartinNowak:
Auto-merge toggled on

MartinNowak:
Also see https://github.com/dlang/dmd/pull/6495#issuecomment-277932866

WalterBright:
CyberShadow/DAutoTest is failing with the same error on every PR of mine. It's clearly something broken that has been merged in somewhere, and has nothing to do with this PR.

PR #1754, opened 2017-01-30
fix pureMalloc test for Win32

rainers:


CyberShadow:
Auto-merge toggled on

PR #1755, opened 2017-02-04
[trivial] more consistent aApply

UplinkCoder:
no need for a return we can just break since the return at the end of the function will be hit


yebblies:
a) Some of the breaks are inside two nested loops, and break will only break out of the inner.
b) The return makes it clear that in that case the function returns immediately, a break doesn't.

UplinkCoder:
@yebblies Thanks I did overlook the nesting level ...
closing for now

PR #1757, opened 2017-02-07
make toString() @safe in core.sys.windows.stacktrace

WalterBright:
Need to move towards making Object.toString() `@safe`.


andralex:
I'm going to go on a limb here (and elsewhere) and ignore DAutoTest...

andralex:
Auto-merge toggled on

PR #1758, opened 2017-02-07
make formatThrowable() @safe

WalterBright:
Moving towards making Object.toString() `@safe`

### DIFF
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -496,11 +496,12 @@ extern (C) int _d_run_main(int argc, char **argv, MainFunc mainFunc)
     return result;
 }
 
-private void formatThrowable(Throwable t, scope void delegate(in char[] s) nothrow sink)
+@safe private void formatThrowable(Throwable t, scope void delegate(in char[] s) @safe nothrow sink)
 {
     for (; t; t = t.next)
     {
-        t.toString(sink); sink("\n");
+        () @trusted { t.toString(sink); } ();
+        sink("\n");
 
         auto e = cast(Error)t;
         if (e is null || e.bypassedException is null) continue;
@@ -508,7 +509,8 @@ private void formatThrowable(Throwable t, scope void delegate(in char[] s) nothr
         sink("=== Bypassed ===\n");
         for (auto t2 = e.bypassedException; t2; t2 = t2.next)
         {
-            t2.toString(sink); sink("\n");
+            () @trusted { t2.toString(sink); } ();
+            sink("\n");
         }
         sink("=== ~Bypassed ===\n");
     }
@@ -523,9 +525,11 @@ extern (C) void _d_print_throwable(Throwable t)
     {
         static struct WSink
         {
+          private:
             wchar_t* ptr; size_t len;
 
-            void sink(in char[] s) scope nothrow
+          public:
+            @trusted void sink(scope const char[] s) scope nothrow
             {
                 if (!s.length) return;
                 int swlen = MultiByteToWideChar(
@@ -613,7 +617,7 @@ extern (C) void _d_print_throwable(Throwable t)
         }
     }
 
-    void sink(in char[] buf) scope nothrow
+    @trusted void sink(scope const char[] buf) scope nothrow
     {
         fprintf(stderr, "%.*s", cast(int)buf.length, buf.ptr);
     }


JackStouffer:
Why are you assuming `toString` is going to be safe?

Won't this change allow this to be merged?

```d
class ThrowMe : Throwable
{
    char* seq;
    ubyte len;

    override string toString() @system
    {
        string ret;
        int i;
        auto start = seq;

        for (i = 0; i <= len; i++) // oops, used <=, buffer overflow
        {
            ret ~= start;
            ++start;
        }
    }
}

main() @safe
{
    auto c = new ThrowMe();
    formatThrowable(c, someSink);
}
```

WalterBright:
> Why are you assuming toString is going to be safe?

Assuming, no, driving it that direction, yes. All of Object's methods need to be safe, otherwise `@safe` programming is a bit of a joke.

WalterBright:
> But they currently aren't

They aren't marked @safe, correct. But I haven't found one yet that is not actually @safe or could be easily made so.

> can you provide another justification for your assumption of Throwable.toString being safe to call? If not, the joke's on you.

It needs to be safe to call - if you agree that there's any point to `@safe` at all.

JackStouffer:
Why not make this a template that can infer `@safe`-ty

jmdavis:
Well, the issues with forcing particular attributes like `@safe` or `const` are why it was decided some time ago that we should remove functions like `toString` from `Object`: https://issues.dlang.org/show_bug.cgi?id=9772

However, there hasn't been much done towards that. The only work that I'm aware of in that direction is the work that Martin has done towards replacing the AA implementation (without which, we can't do it), and the attempts to templatize `opEquals` that kept getting blocked by compiler bugs, and the current PR for it has just lingered: https://github.com/dlang/druntime/pull/1439. So, while a decision was made, it hasn't had much practical effect.

And unfortunately, even if we solve that problem for `Object`, we're _still_ stuck with it with `Throwable`, since the nature of `Throwable` is such that we need `toString` (or something like it) on `Throwable` and thus have to then either force it to be `@safe`/`@trusted` or allow it to be `@system` and then have anything which prints exception would have to be `@trusted`.

Now, fortunately, I think that forcing `@safe` in this situation is probably far more tenable that forcing `const` tends to be. It at least _seems_ like it would be unreasonable to have a `Throwable` with an `@system` `toString` that can't be `@trusted`, but it's also not all that uncommon that someone comes up with a valid use case that seems to defy what seems obvious.

Regardless, while I would expect that moving towards `@safe` is the right way to go, I'm not sure how we do it without `@trust`ing code that is potentially very much not `@safe` - which is what this is doing. And as David points out, this is just internal stuff that we can likely afford to let be `@system`, even if that's not ideal.

There was talk at one point though of replacing `toString` to allow exceptions more freedom in how they generate the text to be printed (e.g. not having to allocate a `string` to print a message), which would allow us the opportunity to change how `@safe` is involved with `Throwable` without breaking code or `@trust`ing `@system` code which potentially already exists - though depending on how that new mechanism worked, it might not be guaranteed to be `@safe`, since it would certainly allow for more freedom in how it was implemented, since that would be the whole point. But it does seem pretty bad if we can't have a way to guarantee that printing an exception's message is `@safe`.

andralex:
Looks like a catch-22, and this doesn't seem like the best first step.

MartinNowak:
Yes, and we'll keep getting stuck on those. There is a bigger general problem with attribute transitions, in particular on interfaces and non-final classes.
Think we need to come up with some deprecation mechanism for those.
Maybe the `@future` proposal (https://github.com/dlang/DIPs/pull/55) might be a fit for that.

WalterBright:
We'll get nowhere if we continue like this. One has to start making progress somewhere.

WalterBright:
It's progress because nobody that calls formatThrowable directly or via other functions can be made @safe. In trying to make Phobos @safe, I often run into A->B->C->D->A again. At some point some part in this circle has to be made @safe, because trying to make the entire circle @safe in one PR is impractical.

As I said before, it's just like the viral nature of making things const-correct.

If you have a better way, please let me know.

yebblies:
@WalterBright I think the ability to override `Throwable.toString` as `@safe` will make this redundant, right?  Either that or making `Throwable` derive from `SafeObject`.

RazvanN7:
This PR doesn't seem to be going anywhere, so I'm going to close. @WalterBright please reopen if you still think this should be pursued.

PR #1760, opened 2017-02-10
core.exception: add _d_assertxx() function

WalterBright:
The idea is to fix the bloat of these 3 functions being generated for every module:
```
_D5bug177__arrayZ:
                push    EBX
                push    EAX
                mov     ECX,offset FLAT:_D5bug1712__ModuleInfoZ[0Eh]
                mov     EBX,7
                push    ECX
                push    EBX
                call    near ptr __d_arraybounds

_D5bug178__assertFiZv:
                push    EBX
                push    EAX
                mov     ECX,offset FLAT:_D5bug1712__ModuleInfoZ[0Eh]
                mov     EBX,7
                push    ECX
                push    EBX
                call    near ptr __d_assert

_D5bug1715__unittest_failFiZv:
                push    EBX
                push    EAX
                mov     ECX,offset FLAT:_D5bug1712__ModuleInfoZ[0Eh]
                mov     EBX,7
                push    ECX
                push    EBX
                call    near ptr __d_unittest
                add     ESP,0Ch
                pop     EBX
                ret
```
by only generating one function that calls `_d_assertxx()`. The string parameter is replaced by a `ModuleInfo` pointer which saves an additional 2 instructions.

Compiler changes are necessary to complete this.

### DIFF
--- a/src/core/exception.d
+++ b/src/core/exception.d
@@ -374,9 +374,9 @@ unittest
 }
 
 
-///////////////////////////////////////////////////////////////////////////////
+// /////////////////////////////////////////////////////////////////////////////
 // Overrides
-///////////////////////////////////////////////////////////////////////////////
+// /////////////////////////////////////////////////////////////////////////////
 
 
 // NOTE: One assert handler is used for all threads.  Thread-local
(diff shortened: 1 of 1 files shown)


WalterBright:
Don't pull this, I'm not sure I want to go this way yet.

PR #1761, opened 2017-02-10
add _d_assertp, _d_unittestp, _d_arrayboundsp

WalterBright:
The "m" functions are no longer used. The "p" versions use fewer bytes of code to call. The efficiency loss with the strlen is irrelevant.

### DIFF
--- a/src/core/exception.d
+++ b/src/core/exception.d
@@ -621,9 +621,10 @@ extern (C)
 
     /* One of these three is called upon an assert() fail.
      */
-    void _d_assertm(immutable(ModuleInfo)* m, uint line)
+    void _d_assertp(immutable(char)* file, uint line)
     {
-        onAssertError(m.name, line);
+        import core.stdc.string : strlen;
+        onAssertError(file[0 .. strlen(file)], line);
     }
 
     void _d_assert_msg(string msg, string file, uint line)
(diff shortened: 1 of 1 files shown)


src/core/exception.d
```diff
@@ -621,9 +621,10 @@ extern (C)
 
     /* One of these three is called upon an assert() fail.
      */
-    void _d_assertm(immutable(ModuleInfo)* m, uint line)
+    void _d_assertp(immutable(char)* file, uint line)
```
jacob-carlborg: Can `file` be passed as a D string directly?

klickverbot: I believe the idea is to reduce the amount of machine code needed at the call site.

jacob-carlborg: And a D string would require this? I mean, I have no idea about these things.

klickverbot: There would be an extra register push (on x86_64, or depending on whatever other target ABI a stack store) to set up the length argument.

jacob-carlborg: I see, thanks.

WalterBright: That's right.

WalterBright:
Blocking https://github.com/dlang/dmd/pull/6530

PR #1762, opened 2017-02-12
line_trace.exp: make more resistant to 'scope' changes

WalterBright:
This makes it work with and without `scope`.

Blocks https://github.com/dlang/dmd/pull/6495


andralex:
Auto-merge toggled on

PR #1763, opened 2017-02-12
scan DATA and TLS according to compiler generated pointer infos

rainers:
Implement precise scanning of the DATA and TLS segment for Win32, based on pointer information written by https://github.com/dlang/dmd/pull/6534


MartinNowak:
I guess we might regret it once looking at DLLs again, but let's try it as experiment.
Can you (or someone else) gather non-trivial benchmarks?

MartinNowak:
Auto-merge toggled on

rainers:
Thanks. I use this in the Visual D build (a DLL, but not sharing phobos), the map file shows that the number of pointers is similar to phobos unittest (357 data, 246 tls), but in about 800 kB.

MartinNowak:
> Thanks. I use this in the Visual D build (a DLL, but not sharing phobos), the map file shows that the number of pointers is similar to phobos unittest (357 data, 246 tls), but in about 800 kB.

Does it measurably speed up scanning?

MartinNowak:
Still seems to fail.

rainers:
> Does it measurably speed up scanning?

I don't have any benchmarks on that. My concern in Visual D is editing files like std.datetime eating memory like crazy.

rainers:
> Still seems to fail.

Seems to crash for Win64. That works for me locally.

rainers:
I'm going to disable the precise scanning for now. Should we make enabling precise DATA/TLS scanning a DRT-option?

PR #1764, opened 2017-02-16
Push back deprecation of bad cycle detection one more major release

schveiguy:
We missed the window to switch to abort in 2.073, and actually in hindsight, the master branch should have been updated early on in the cycle to give people time to deal with it.

This just pushes it to 2.074 (and updates the message appropriately). I will add a PR to switch to abort in the master branch once this is pulled and merged into master.

### DIFF
--- a/src/rt/minfo.d
+++ b/src/rt/minfo.d
@@ -182,7 +182,7 @@ struct ModuleGroup
             ignore
         }
 
-        // Change default to .abort in 2.073
+        // Change default to .abort in 2.074
         auto onCycle = OnCycle.deprecate;
 
         switch(cycleHandling) with(OnCycle)
@@ -517,7 +517,7 @@ struct ModuleGroup
             import core.stdc.stdio : fprintf, stderr;
             fprintf(stderr, "Deprecation 16211 warning:\n"
                 ~ "A cycle has been detected in your program that was undetected prior to DMD\n"
-                ~ "2.072. This program will continue, but will not operate when using DMD 2.073\n"
+                ~ "2.072. This program will continue, but will not operate when using DMD 2.074\n"
                 ~ "to compile. Use runtime option --DRT-oncycle=print to see the cycle details.\n");
 
         }


schveiguy:
The circleci test failure makes no sense. Coverage should not have changed at all.

schveiguy:
Ping @MartinNowak

schveiguy:
Also not sure of the reason for the doc build failure.

wilzbach:
> The circleci test failure makes no sense. Coverage should not have changed at all.

As you have used a branch directly at the `dlang` repo, the script tries to clone `dmd` with the same branch. This is for example used for the `stable` branch.
However, in theory it should fallback to the `master` branch at DMD.

> We missed the window to switch to abort in 2.073, and actually in hindsight, the master branch should have been updated early on in the cycle to give people time to deal with it.
> This just pushes it to 2.074

Since the last release @MartinNowak has enabled GitHub milestones for every release with which you can tag a PR. The idea behind this is exactly to prevent such  oversights.

wilzbach:
> However, in theory it should fallback to the master branch at DMD.

https://github.com/dlang/druntime/pull/1725 (it hangs in the queue)

schveiguy:
> Since the last release @MartinNowak has enabled GitHub milestones for every release with which you can tag a PR.

That doesn't help as the PR to update to the `.abort` default didn't exist, there was nothing to tag!

schveiguy:
> As you have used a branch directly at the dlang repo

Just curious if there is a way to prevent this? I have no options when making a "quickie" PR by editing a file in the github web UI to create a branch for my user, it defaults to creating it on the dlang repo.

wilzbach:
> That doesn't help as the PR to update to the .abort default didn't exist, there was nothing to tag!

Oh sorry. Fair enough ;-)

> Just curious if there is a way to prevent this? I have no options when making a "quickie" PR by editing a file in the github web UI to create a branch for my user, it defaults to creating it on the dlang repo.

The only way I know is to manually change the URL. So if I click on "edit" e.g. at https://github.com/dlang/phobos/edit/master/.gitignore, I can replace `dlang` with `schveiguy` in the URL bar and thus get `https://github.com/dlang/phobos/edit/master/.gitignore`. This will create a branch from your fork then, so it requires the fork to be moderately up-to-date for this to work.
In any case the problem with CircleCi was fixed long ago at Phobos, it's just stuck in the queue here.

schveiguy:
Hm... maybe editing the stable branch on my account I can then retarget the PR to the upstream branch? I'll try that next time, thanks.

schveiguy:
@MartinNowak just released 2.073.1, so not sure if this is viable any more.

wilzbach:
> @MartinNowak just released 2.073.1, so not sure if this is viable any more.

There is still 2.073.2

schveiguy:
Thanks, @MartinNowak. Here is the promised change to master: #1765

PR #1765, opened 2017-02-17
Update handling of previously undetected cycles to abort execution.

schveiguy:
This completes the deprecation. In a future version, we may remove the "deprecate" option.

Note, I didn't change the message/version number to 2.074, because that was done in #1764, and should get into master soon.


schveiguy:
associated doc change: dlang/dlang.org#1574

PR #1766, opened 2017-02-17
@disable TypeInfo.init

aG0aep6G:
As planned (one release behind schedule).

### DIFF
--- a/changelog/disable-TypeInfo.init.dd
+++ b/changelog/disable-TypeInfo.init.dd
@@ -1,0 +1,6 @@
+TypeInfo.init has been `@disable`d.
+
+This is a step towards removal of `TypeInfo.init`, which is necessary to resolve
+a name clash with the type property $(GLINK2 property, init).
+
+Use $(REF_OBJECT_SHORT TypeInfo.initializer) instead.
--- a/changelog/disable-TypeInfo.init.dd
+++ b/changelog/disable-TypeInfo.init.dd
@@ -1,6 +1,12 @@
-TypeInfo.init has been `@disable`d.
+`TypeInfo.init` has been `@disable`d.
 
-This is a step towards removal of `TypeInfo.init`, which is necessary to resolve
-a name clash with the type property $(GLINK2 property, init).
+$(REF_OBJECT_SHORT TypeInfo.init) has been `@disable`d. Use
+$(REF_OBJECT_SHORT TypeInfo.initializer) instead.
 
-Use $(REF_OBJECT_SHORT TypeInfo.initializer) instead.
+`TypeInfo.init` is a legacy alias of the method that is now called
+$(REF_OBJECT_SHORT TypeInfo.initializer). The name change is necessary because
+the name "init" clashes with the type property of the same name
+($(GLINK2 property, init)).
+
+The legacy alias has been deprecated since 2.072.0. It's going to be removed in
+2.075.0.
--- a/src/object.d
+++ b/src/object.d
@@ -296,11 +296,9 @@ class TypeInfo
      */
     abstract const(void)[] initializer() nothrow pure const @safe @nogc;
 
-    /// $(RED Scheduled for deprecation.) Please use `initializer` instead.
-    deprecated("Please use initializer instead.") alias init = initializer;
-        // since 2.072
-    version(none) @disable static const(void)[] init(); // planned for 2.073
-    /* Planned for 2.074: Remove init, making way for the init type property,
+    /// $(RED Removed.) Please use `initializer` instead.
+    @disable static const(void)[] init(); // since 2.074
+    /* Planned for 2.075: Remove init, making way for the init type property,
     fixing issue 12233. */
 
     /** Get flags for type: 1 means GC should scan for pointers,


aG0aep6G:
I'm not sure if I got the changelog entry right. The README.md didn't actually tell me to create a file, but that's the plan, right? When I build ../dlang.org/web/changelog/pending.html as instructed, my entry doesn't show up. The doc tester doesn't show it either. Maybe the tooling isn't there yet?

wilzbach:
> I'm not sure if I got the changelog entry right. The README.md didn't actually tell me to create a file, but that's the plan, right?

Yep - see Phobos for a couple of examples: https://github.com/dlang/phobos/tree/master/changelog

> When I build ../dlang.org/web/changelog/pending.html as instructed, my entry doesn't show up.

Your problem is that the changelog entry doesn't end with `.dd`. It seems that the `druntime` changelog `README` is a bit outdated in comparison to Phobos: https://github.com/dlang/phobos/tree/master/changelog

For reference, that's how it looks once the file has the proper extension:

image

> The doc tester doesn't show it either. Maybe the tooling isn't there yet?

Yep: https://github.com/dlang/dlang.org/pull/1549

aG0aep6G:
Blocked by dmd tests using TypeInfo.init - <https://github.com/dlang/dmd/pull/6547>.

> Your problem is that the changelog entry doesn't end with .dd.

Fixed, thanks.

aG0aep6G:
> Blocked by dmd tests using TypeInfo.init - dlang/dmd#6547.

That PR has been merged. auto-tester is turning green. So this should be good to go. I've squashed the fixup commit.

rainers:
You should also update dmd not to use TypeInfo.init, as the travis build for dmd rebuilds itself, see for example https://travis-ci.org/dlang/dmd/jobs/202809672#L5264.

aG0aep6G:
> You should also update dmd not to use TypeInfo.init, as the travis build for dmd rebuilds itself, see for example https://travis-ci.org/dlang/dmd/jobs/202809672#L5264.

https://github.com/dlang/dmd/pull/6550

aG0aep6G:
Alright, we have a problem here. dlang/dmd#6550 fails because the test bots use relatively old releases to build dmd. Those don't have TypeInfo.initializer, so they fail.

I'm not sure, is there an older release that's deemed stable, which is supposed to be able to build all newer releases? If so, can we bump that to a newer one or make a point release for it? If not, can we just make the bots use newer versions?

wilzbach:
> I'm not sure, is there an older release that's deemed stable, which is supposed to be able to build all newer releases? If so, can we bump that to a newer one or make a point release for it? If not, can we just 
make the bots use newer versions?

That would be 2.072.2. For Travis and CircleCi this is already used at the `dmd` repo (though not for druntime and phobos, but the bump is in the queue: https://github.com/dlang/phobos/pull/5140, https://github.com/dlang/druntime/pull/1725)

However as the auto-tester (and GDC) still use an old frontend, you should definitely see this discussion http://forum.dlang.org/thread/FA142E6B-DD11-43F6-8EF9-EEC72AA1A7F2@klickverbot.at

aG0aep6G:
> That would be 2.072.2.

2.072 is not a long-term support version, is it? Anyway, 2.072 is fine, as TypeInfo.initializer has been [added in 2.070.0](https://github.com/dlang/druntime/commit/04700217971964fdc8edb1798eaa669ec37fa7e9). Trouble is with anything older than that.

> However as the auto-tester (and GDC) still use an old frontend

The auto-tester and the doc tester are the real problem. If gdc is supposed to be able to build dmd, then it needs updating too. Question is how do I get TypeInfo.initializer into the testers (and maybe gdc)?

ZombineDev:
@ibuclaw @jpf91 we would like to move forward with removing `TypeInfo.init`. Any chance of cherrypicking the druntime change from #1403 into GDC, so we can bump the minimum compiler version that is used to build DMD?

CC @MartinNowak

ibuclaw:
GDC is not affected by this as far as I can see, the dmd builds are still passing on travisci.

aG0aep6G:
dlang/dmd#6550 has been merged. Testers are green. Regarding dmd, this should be good to go.

gdc still uses the C frontend, so it should not be affected.

ldc will have to incorporate dlang/dmd#6550 to be be buildable by dmd 2.074. I've tried to cherry pick it, but the path to rmem.d is different in ldc and that makes things weird. @klickverbot: Please advise or take over.

changelog/disable-TypeInfo.init.dd
```diff
@@ -0,0 +1,6 @@
+TypeInfo.init has been `@disable`d.
```
klickverbot: Perhaps write this out as `init()` to make clear that this is referring to the property function?

changelog/disable-TypeInfo.init.dd
```diff
@@ -0,0 +1,6 @@
...
+
+This is a step towards removal of `TypeInfo.init`, which is necessary to resolve
```
klickverbot: Maybe phrase this in a more direct and comprehensive way to give people who have not been following the process enough to make sense of the change? Something like:

`TypeInfo.init()` (deprecated since 2.0XX) collides with the built-in type property of the same name and will be removed in XXX. To get the default initializer of the described type, use TypeInfo.initializer instead.

aG0aep6G:
I've tweaked the changelog entry a bit, reacting to @klickverbot's comments, hopefully making things clearer. I don't really care about the exact words. Please feel free to just edit it, if you think it can be improved.

----

> This is fine from an LDC perspective.

Alright. Looks like we're all set then.

MartinNowak:
> If we wanted to be cautious about reducing the number of future bootstrapping steps, we would make sure that jump straight past this change from the last C++ compiler

Bootstrapping will really only remain feasible via cross-compilation. It's just unrealistic to assume that we can rebuild every intermediate compiler in 10 years (not to mention the crazy amount of steps).

PR #1768, opened 2017-02-19
workaround sscanf bug in old snn.lib

WalterBright:
Explanation in comments.

### DIFF
--- a/src/gc/config.d
+++ b/src/gc/config.d
@@ -180,8 +180,32 @@ body
     assert(n > 4 && n < fmt.length);
 
     int nscanned;
-    if (sscanf(str.ptr, fmt.ptr, &res, &nscanned) < 1)
-        return parseError("a float", optname, str);
+    version (Win32)
+    {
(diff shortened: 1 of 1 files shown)


src/gc/config.d
```diff
@@ -180,8 +180,32 @@ body
     assert(n > 4 && n < fmt.length);
 
     int nscanned;
-    if (sscanf(str.ptr, fmt.ptr, &res, &nscanned) < 1)
-        return parseError("a float", optname, str);
+    version (Win32)
```
sprinkle131313: If the issue is with DMC, shouldn't this be version(DMC) then?

klickverbot: Yep, it should be.

MartinNowak:
Mmh, this is just a workaround for testing right? Because the next release containing that workaround will already come with an updated snn.lib.
And the auto-tester were updated already as well, so this is intended for the other testers?

MartinNowak:
Auto-merge toggled on

PR #1769, opened 2017-02-21
Fix wrong version name in core.sys.linux.ifaddrs

tomerfiliba:


nemanja-boric-sociomantic:
Whoops, sorry and thanks!

PR #1770, opened 2017-02-21
Remove support for deprecated built-in array.reverse property.

quickfur:
Depends on: https://github.com/dlang/dmd/pull/6554


quickfur:
Note: I couldn't remove `_adSort` et call because apparently sorting is required by the treap implementation, and I couldn't see any easy replacement without needing to import Phobos, which is a no-no.

jmdavis:
The dmd tests are failing on the autotester, though glancing at the results, it's not clear to me why they're failing.

aG0aep6G:
> The dmd tests are failing on the autotester, though glancing at the results, it's not clear to me why they're failing.

dmd/test/runnable/test20.d uses `.reverse` ([see lines 734 through 759](https://github.com/dlang/dmd/blob/5cfc8d982df28930a49a6ec0bd73061ec25bad30/test/runnable/test20.d#L734-L759)).

PR #1772, opened 2017-02-24
Fix issue 17136 - dictionary get(value, defaultValue) should be nothrow

somzzz:
The error message indicated ```lazy``` parameter ```defaultValue``` in the ```get``` function could ```throw```. This solution returns ```null``` if ```defaultValue``` throws (but this shouldn't happen).

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2129,10 +2129,16 @@ unittest
     assert(T.count == 2);
 }
 
-inout(V) get(K, V)(inout(V[K]) aa, K key, lazy inout(V) defaultValue)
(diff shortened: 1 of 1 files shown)


wilzbach:
Can't we fix the underlying problem:

https://issues.dlang.org/show_bug.cgi?id=12647

There's even a PR for it:

https://github.com/dlang/dmd/pull/6348

somzzz:
I wasn't aware of that PR, thanks for pointing it out!

andralex:
@somzzz you may want to take over dlang/dmd#6348, we haven't heard from @LemonBoy in a while. Thanks!

somzzz:
I will close this since it is a duplicate of  https://github.com/dlang/dmd/pull/6348 .

PR #1774, opened 2017-02-25
core.sys.windows: Allow use on non-Windows platforms

CyberShadow:
The Windows API binding headers contain declarations which are
occasionally useful even on non-Windows platforms, such as:

- Windows / Portable Executable file format structures, which are used
  for Windows executables, but also portable Common Language
  Infrastructure executables as used by .NET and Mono.

- Bitmap file format structures, used in .bmp image files which can be
  used on any platform.

Limiting the contents of the Windows headers to be available only when
compiling for the Windows platform is generally an artificial
limitation, since one can conceivably think of uses for most
declarations. Implementations of the Win32 API exist on other
platforms and can be used alongside that platform's native APIs (see
e.g. Winelib).

As such, this change removes the `version(Windows):` from the top of
all modules, and instead only limits the `pragma(lib)` linker
directives and Windows-specific helper/wrapper code to the Windows
target platform.

### DIFF
--- a/src/core/sys/windows/accctrl.d
+++ b/src/core/sys/windows/accctrl.d
@@ -8,7 +8,6 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_accctrl.d)
  */
 module core.sys.windows.accctrl;
-version (Windows):
 
 version (ANSI) {} else version = Unicode;
 
--- a/src/core/sys/windows/aclapi.d
+++ b/src/core/sys/windows/aclapi.d
@@ -8,10 +8,9 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_aclapi.d)
  */
 module core.sys.windows.aclapi;
-version (Windows):
 
 version (ANSI) {} else version = Unicode;
-pragma(lib, "advapi32");
+version (Windows) pragma(lib, "advapi32");
 
 import core.sys.windows.windows, core.sys.windows.accctrl;
 
--- a/src/core/sys/windows/aclui.d
+++ b/src/core/sys/windows/aclui.d
@@ -8,8 +8,7 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_aclui.d)
  */
 module core.sys.windows.aclui;
-version (Windows):
-pragma(lib, "aclui");
+version (Windows) pragma(lib, "aclui");
 
 private import core.sys.windows.w32api;
 /*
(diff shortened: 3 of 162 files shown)


CyberShadow:
CC @smjgordon @adamdruppe @qchikara

adamdruppe:
Do you know what the effect is of the various struct.inits or other static data in this? Before-and-after size of druntime lib file good enough to check. I would guess it is negligible but would like to be sure. I presume you already took care of versioning anything that could cause other linker errors.

Otherwise though, I think this is pretty harmless and indeed potentially useful.

CyberShadow:
> Before-and-after size of druntime lib file good enough to check.

Before:
```
-rw-r--r-- 1 vladimir vladimir 10160152 Feb 26 03:23 libdruntime.a
-rw-r--r-- 1 vladimir vladimir  2920784 Feb 26 03:23 libdruntime.so.a
```
After:
```
-rw-r--r-- 1 vladimir vladimir 32850064 Feb 26 03:25 libdruntime.a
-rw-r--r-- 1 vladimir vladimir  6009080 Feb 26 03:25 libdruntime.so.a
```

Actually holy crap :)

adamdruppe:
Yeah, yikes, that's a lot more than I thought. It isn't necessarily a dealbreaker, but that is a fairly significant difference for something that we can live without... prolly all ModuleInfo and TypeInfo we could avoid generating if dmd could skip those :S

CyberShadow:
Here's the treemap: http://thecybershadow.net/d/mapview/view.php?id=58b2577a2a099

The total size doesn't match the .a file, but the proportions seem right. I guess the less is the symbol listing, debug information, and padding.

I was hoping there would be a few big ones we could take down, but that doesn't seem to be the case - it's spread all over.

Anyway, I thought DMD didn't generate `.init` if it was all zeroes?

Maybe as a middle ground we could just do the ones that are immediately obviously useful... such as `winnt` and `wingdi`, and their dependencies.

CyberShadow:
> Anyway, I thought DMD didn't generate .init if it was all zeroes?

Possibly related bug report: https://issues.dlang.org/show_bug.cgi?id=15130

RazvanN7:
@CyberShadow  I wonder if the size issue has been fixed in the meanwhile, however, it's gonna be pretty boring to rebase this. Do you still think this addition is worthwhile? I can then rebase it and see what the differences are.

CyberShadow:
No, I maintain the win32 Dub package for this purpose:

https://code.dlang.org/packages/win32
https://github.com/CyberShadow/win32

Also, this PR is a dupe of #2615

RazvanN7:
Thanks!

PR #1776, opened 2017-02-26
adding Jenkinsfile to hook with new ci2.dawg.eu instance

MartinNowak:
- should transfer the ownership of the ci repo to dlang/ci

Example: https://ci2.dawg.eu/blue/organizations/jenkins/dlang/detail/dlang/150/pipeline

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,11 @@
+#!/bin/env groovy
+
+node {
+    stage ("Load CI Scripts") {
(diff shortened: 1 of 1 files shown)


MartinNowak:
Ah, didn't work b/c the Jenkinsfile is only loaded from trusted repos (i.e. not from forks), so it was using master here for testing which doesn't have a Jenkinsfile yet.

PR #1779, opened 2017-03-01
Updated makefile to use generated/os/release/model/dmd

RazvanN7:


wilzbach:
> Does this handle multilib builds, i.e. the building 32 bit libraries on a 64 bit system part?

Yes, `$(MODEL)` is set based on the system whereas `$(MODEL_FLAG)` (by default set to `$(MODEL)`) is used for building the libraries.

This is part of the [`osmodel.mak`](https://github.com/dlang/druntime/blob/master/osmodel.mak#L35)

andralex:
Unclear on what the autotester issue is, but it looks like a one-off.

andralex:
Auto-merge toggled on

andralex:
@RazvanN7 please push again your repo so it kicks off the autotester

andralex:
I'll take a risk to actually merge this

PR #1780, opened 2017-03-02
osmodel.mak: Mention tools + dlang.org repos in the update reminder

wilzbach:
- The `osmodel.mak` files weren't in sync
- Since a couple of weeks the `tools` and `dlang.org` repo use the same file

Btw this manual sync seems very error-prone and labour-intensive.
Why can't we do an `include $(DMD_DIR)/src/osmodel.mak`?


wilzbach:
> Why can't we do an include $(DMD_DIR)/src/osmodel.mak?

So I change the PR to do this as we require the `dmd` repo to build anyways and it's hard to get an outdated file, if there's only one ;-)

Btw additionally we could even let the Makefile clone `dmd` automatically if it doesn't exists, but that's not directly relevant for this PR.

MartinNowak:
> Btw additionally we could even let the Makefile clone dmd automatically if it doesn't exists, but that's not directly relevant for this PR.

Please no, it's already a nightmare to work with the dlang.org Makefile in builds scripts. Let's not turn more Makefiles into their own build scripts.

MartinNowak:
We have been over this, see https://github.com/dlang/druntime/pull/698 and https://github.com/dlang/phobos/pull/1806.
Having separate repos for all components turned be a pain in many places, but a single (or meta) repo would have it's own downsides (though likely less).
If we can figure out a good way to store common makefiles and script components somewhat, that'd sound reasonable. We also still want to move the dlangci repo to dlang/ci.

Maybe we'd rather add a `dlang/build` or `dlang/common` repo that all components can rely on.

wilzbach:
> Maybe we'd rather add a dlang/build or dlang/common repo that all components can rely on.

Well, the idea was to use `tools` as common repo, because it avoids to create yet another repo & tools is already required at Phobos for some targets (e.g. `style`).

Anyway this takes much more effort than initially thought, so I am closing this.
If anyone wants to pick this up & push for a less redundant build system, please do so!

PR #1781, opened 2017-03-02
Array Compare Template Function

somzzz:
This is an initial implementation of the template function to replace all overloads of ```TypeInfo.compare```.

The purpose is to have the compiler generate calls to this template, rather than a function (adCmp2 in this case https://github.com/dlang/druntime/blob/master/src/rt/adi.d#L445)
Array comparisons will be lowered to this template after a ```dmd``` patch. An example:
```
bool fun(int[] a, int[] b) {
        return b[] < a[]; // generates call to the template
}

void main()
{
    import std.stdio;
    writeln(fun([1,2,3], [4, 1, 5]));
    writeln(fun([4,6,5], [1, 2, 3]));
}
```

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3241,6 +3241,9 @@ template RTInfo(T)
     enum RTInfo = null;
 }
 
+// This function is called by the compiler when dealing with array
+// comparisons in the semantic analysis phase of CmpExp. The ordering
+// comparison is lowered to a call to this template.
 int __cmp(T1, T2)(T1[] s1, T2[] s2)
 {
     import core.internal.traits : Unqual;
@@ -3364,7 +3367,8 @@ int __cmp(T1, T2)(T1[] s1, T2[] s2)
 
                 if (e1 < e2)
                     return -1;
-                else if (e1 > e2)
+
+                if (e1 > e2)
                     return 1;
             }
 
@@ -3391,9 +3395,9 @@ int __cmp(T1, T2)(T1[] s1, T2[] s2)
                 if (c != 0)
                     return c < 0 ? -1 : 1;
             }
-            else static if (__traits(compiles, _adCmp(e1, e2)))
+            else static if (__traits(compiles, __cmp(e1, e2)))
             {
-                auto c = _adCmp(e1, e2);
+                auto c = __cmp(e1, e2);
                 if (c != 0)
                     return c < 0 ? -1 : 1;
             }
@@ -3401,7 +3405,8 @@ int __cmp(T1, T2)(T1[] s1, T2[] s2)
             {
                 if (e1 < e2)
                     return -1;
-                else if (e1 > e2)
+
+                if (e1 > e2)
                     return 1;
             }
         }
--- a/src/object.d
+++ b/src/object.d
@@ -3241,71 +3241,13 @@ template RTInfo(T)
     enum RTInfo = null;
 }
 
-private template Floating1(T)
-if (is(T == cfloat) || is(T == cdouble) || is(T == creal))
-{
-    // Use rt.cmath2._Ccmp instead ?
-    int compare(T f1, T f2)
-    {
-        int result;
-
-        if (f1.re < f2.re)
-            result = -1;
-        else if (f1.re > f2.re)
-            result = 1;
-        else if (f1.im < f2.im)
-            result = -1;
-        else if (f1.im > f2.im)
-            result = 1;
-        else
-            result = 0;
-        return result;
-    }
-}
-
-private template Floating1(T)
-if (is(T == float) || is(T == double) || is(T == real))
-{
-    int compare(T d1, T d2)
-    {
-        if (d1 != d1 || d2 != d2) // if either is NaN
-        {
-            if (d1 != d1)
-            {
-                if (d2 != d2)
-                    return 0;
-                return -1;
-            }
-            return 1;
-        }
-        return d1 < d2 ? -1 : (d1 > d2);
-    }
-}
-
-private template Array1(T)
-if (is(T ==  float) || is(T ==  double) || is(T ==  real) ||
-    is(T == cfloat) || is(T == cdouble) || is(T == creal))
-{
-    int compare(T[] s1, T[] s2)
-    {
-        auto len = s1.length;
-        if (s2.length < len)
-            len = s2.length;
-
-        foreach (const u; 0 .. len)
-        {
-            if (int c = Floating1!T.compare(s1[u], s2[u]))
-                return c;
-        }
-        return s1.length < s2.length ? -1 : (s1.length > s2.length);
-    }
-}
-
-int _adCmpT(T1, T2)(T1[] s1, T2[] s2)
+int __cmp(T1, T2)(T1[] s1, T2[] s2)
 {
     import core.internal.traits : Unqual;
     alias U1 = Unqual!T1;
     alias U2 = Unqual!T2;
+
+    static @trusted ref R at(R)(R[] r, size_t i) { return r.ptr[i]; }
 
     // objects
     // Old code went to object.d/TypeInfo_array.compare and then to
@@ -3313,35 +3255,35 @@ int _adCmpT(T1, T2)(T1[] s1, T2[] s2)
     // which combined result in the same code as below.
     static if (is(U1 : Object) && is(U2 : Object))
     {
-        auto c  = cast(sizediff_t)(s1.length - s2.length);
-        if (c == 0)
+        auto len = s1.length;
+
+        if (s2.length < len)
+            len = s2.length;
+
+        foreach (const u; 0 .. len)
         {
-            for (size_t u = 0; u < s1.length; u++)
+            auto o1 = at(s1, u);
+            auto o2 = at(s2, u);
+
+            if (o1 is o2)
+                continue;
+
+            // Regard null references as always being "less than"
+            if (o1)
             {
-                auto o1 = () @trusted { return s1.ptr[u]; }();
-                auto o2 = () @trusted { return s2.ptr[u]; }();
-
-                if (o1 is o2)
-                    continue;
-
-                // Regard null references as always being "less than"
-                if (o1)
-                {
-                    if (!o2)
-                        return 1;
-                    c = o1.opCmp(o2);
-                    if (c != 0)
-                        break;
-                }
-                else
-                {
-                    return -1;
-                }
+                if (!o2)
+                    return 1;
+                auto c = o1.opCmp(o2);
+                if (c != 0)
+                    return c < 0 ? -1 : 1;
+            }
+            else
+            {
+                return -1;
             }
         }
-        return c < 0 ? -1 : (c > 0);
+        return s1.length < s2.length ? -1 : (s1.length > s2.length);
     }
-
     // floating point types
     else static if (__traits(isFloating, U1))
     {
@@ -3350,9 +3292,52 @@ int _adCmpT(T1, T2)(T1[] s1, T2[] s2)
         else static if (is(U1 == ireal)) alias F = real;
         else alias F = U1;
 
-        return () @trusted { return Array1!F.compare(cast(F[])s1, cast(F[])s2); }();
-    }
+        static int compare(F f1, F f2)
+        {
+            static if (is(F == cfloat) || is(F == cdouble) || is(F == creal))
+            {
+                // Use rt.cmath2._Ccmp instead ?
+                int result;
 
+                if (f1.re < f2.re)
+                    result = -1;
+                else if (f1.re > f2.re)
+                    result = 1;
+                else if (f1.im < f2.im)
+                    result = -1;
+                else if (f1.im > f2.im)
+                    result = 1;
+                else
+                    result = 0;
+                return result;
+            }
+            else static if (is(F == float) || is(F == double) || is(F == real))
+            {
+                // d1 is NaN 2^0 + d2 is NaN 2^1
+                auto NaNs = (f1 != f1) | ((f2 != f2) << 1);
+
+                return (NaNs == 3) ? 0 :
+                    (NaNs == 2) ? 1 :
+                    (NaNs == 1) ? -1 :
+                    (f1 < f2) ? -1 : (f1 > f2);
+            }
+            else static assert(false, "Internal error");
+        }
+
+        auto len = s1.length;
+        if (s2.length < len)
+            len = s2.length;
+
+        auto fpArray1 = () { return cast(F[])s1; }();
+        auto fpArray2 = () { return cast(F[])s2; }();
+
+        foreach (const u; 0 .. len)
+        {
+            if (int c = compare(fpArray1[u], fpArray2[u]))
+                return c;
+        }
+        return s1.length < s2.length ? -1 : (s1.length > s2.length);
+    }
     // char types = > dstrcmp
     else static if ((is(U1 == ubyte) && is(U2 == ubyte))
         || (is(U1 == void) && is(U2 == void))
@@ -3374,8 +3359,8 @@ int _adCmpT(T1, T2)(T1[] s1, T2[] s2)
 
             foreach (const u; 0 .. len)
             {
-                auto e1 = () @trusted { return s1.ptr[u]; }();
-                auto e2 = () @trusted { return s2.ptr[u]; }();
+                auto e1 = at(s1, u);
+                auto e2 = at(s2, u);
 
                 if (e1 < e2)
                     return -1;
@@ -3386,8 +3371,7 @@ int _adCmpT(T1, T2)(T1[] s1, T2[] s2)
             return s1.length < s2.length ? -1 : (s1.length > s2.length);
         }
     }
-
-    // integral
+    // integral, struct, nested arrays
     else
     {
         auto len = s1.length;
@@ -3397,15 +3381,30 @@ int _adCmpT(T1, T2)(T1[] s1, T2[] s2)
 
         foreach (const u; 0 .. len)
         {
-            auto e1 = () @trusted { return s1.ptr[u]; }();
-            auto e2 = () @trusted { return s2.ptr[u]; }();
+            auto e1 = at(s1, u);
+            auto e2 = at(s2, u);
 
-            if (e1 < e2)
-                return -1;
-            else if (e1 > e2)
-                return 1;
+            // structs
+            static if (__traits(compiles, e1.opCmp(e2)))
+            {
+                auto c = e1.opCmp(e2);
+                if (c != 0)
+                    return c < 0 ? -1 : 1;
+            }
+            else static if (__traits(compiles, _adCmp(e1, e2)))
+            {
+                auto c = _adCmp(e1, e2);
+                if (c != 0)
+                    return c < 0 ? -1 : 1;
+            }
+            else
+            {
+                if (e1 < e2)
+                    return -1;
+                else if (e1 > e2)
+                    return 1;
+            }
         }
-
         return s1.length < s2.length ? -1 : (s1.length > s2.length);
     }
 }
@@ -3418,8 +3417,8 @@ unittest
         T[2] a = [T.max, T.max];
         T[2] b = [T.min, T.min];
 
-        assert(_adCmpT(a, b) > 0);
-        assert(_adCmpT(b, a) < 0);
+        assert(__cmp(a, b) > 0);
+        assert(__cmp(b, a) < 0);
     }
 
     compareMinMax!int;
@@ -3441,8 +3440,8 @@ unittest
         T[2] a = [T.max, T.max];
         T[2] b = [T.min, T.min];
 
-        assert(_adCmpT(a, b) > 0);
-        assert(_adCmpT(b, a) < 0);
+        assert(__cmp(a, b) > 0);
+        assert(__cmp(b, a) < 0);
     }
 
     compareMinMax!ubyte;
@@ -3452,8 +3451,8 @@ unittest
 
     string s1 = "aaaa";
     string s2 = "bbbb";
-    assert(_adCmpT(s2, s1) > 0);
-    assert(_adCmpT(s1, s2) < 0);
+    assert(__cmp(s2, s1) > 0);
+    assert(__cmp(s1, s2) < 0);
 }
 
 // fp types
@@ -3464,8 +3463,8 @@ unittest
         T[2] a = [T.max, T.max];
         T[2] b = [T.min_normal, T.min_normal];
 
-        assert(_adCmpT(a, b) > 0);
-        assert(_adCmpT(b, a) < 0);
+        assert(__cmp(a, b) > 0);
+        assert(__cmp(b, a) < 0);
     }
 
     compareMinMax!real;
@@ -3500,8 +3499,8 @@ unittest
     auto c1 = new C(1);
     auto c2 = new C(2);
 
-    assert(_adCmpT([c1, c1], [c2, c2]) < 0);
-    assert(_adCmpT([c2, c2], [c1, c1]) > 0);
+    assert(__cmp([c1, c1], [c2, c2]) < 0);
+    assert(__cmp([c2, c2], [c1, c1]) > 0);
 }
 
 
--- a/src/object.d
+++ b/src/object.d
@@ -3241,6 +3241,226 @@ template RTInfo(T)
     enum RTInfo = null;
 }
 
+int _adCmpT(T1, T2)(T1[] s1, T2[] s2)
+{
+    // U1 gets the unqualified version of T1
+    static if (is(T1 W == immutable W)) alias U1 = W;
+    else static if (is(T1 W == const W)) alias U1 = W;
+    else alias U1 = T1;
+
+    // U2 gets the unqualified version of T2
+    static if (is(T2 U == immutable U)) alias U2 = U;
+    else static if (is(T2 U == const U)) alias U2 = U;
+    else alias U2 = T2;
+
+    // objects
+    static if (is(U1 : Object) && is(U2 : Object))
+    {
+        auto c  = cast(sizediff_t)(s1.length - s2.length);
+        if (c == 0)
+        {
+            for (size_t u = 0; u < s1.length; u++)
+            {
+                Object o1 = s1[u];
+                Object o2 = s2[u];
+
+                if (o1 is o2)
+                    continue;
+
+                // Regard null references as always being "less than"
+                if (o1)
+                {
+                    if (!o2)
+                        return 1;
+                    c = o1.opCmp(o2);
+                    if (c == 0)
+                        continue;
+                    break;
+                }
+                else
+                {
+                    return -1;
+                }
+            }
+        }
+        return c < 0 ? -1 : c > 0 ? 1 : 0;
+    }
+
+    // floating point types
+    static if ((is(U1 == float) && is(U2 == float))
+        || (is(U1 == double)   && is(U2 == double))
+        || (is(U1 == real)     && is(U2 == real))
+        || (is(U1 == ifloat)   && is(U2 == ifloat))
+        || (is(U1 == idouble)  && is(U2 == idouble))
+        || (is(U1 == ireal)    && is(U2 == ireal))
+        || (is(U1 == cfloat)   && is(U2 == cfloat))
+        || (is(U1 == cdouble)  && is(U2 == cdouble))
+        || (is(U1 == creal)    && is(U2 == creal)))
+    {
+        template Floating(T)
+        if (is(T == cfloat) || is(T == cdouble) || is(T == creal))
+        {
+          pure nothrow @safe:
+            int compare(T f1, T f2)
+            {
+                int result;
+
+                if (f1.re < f2.re)
+                    result = -1;
+                else if (f1.re > f2.re)
+                    result = 1;
+                else if (f1.im < f2.im)
+                    result = -1;
+                else if (f1.im > f2.im)
+                    result = 1;
+                else
+                    result = 0;
+                return result;
+            }
+        }
+
+        template Floating(T)
+        if (is(T == float) || is(T == double) || is(T == real))
+        {
+          pure nothrow @safe:
+            int compare(T d1, T d2)
+            {
+                if (d1 != d1 || d2 != d2) // if either are NaN
+                {
+                    if (d1 != d1)
+                    {
+                        if (d2 != d2)
+                            return 0;
+                        return -1;
+                    }
+                    return 1;
+                }
+                return (d1 == d2) ? 0 : ((d1 < d2) ? -1 : 1);
+            }
+        }
+
+        template Array(T)
+        if (is(T ==  float) || is(T ==  double) || is(T ==  real) ||
+            is(T == cfloat) || is(T == cdouble) || is(T == creal))
+        {
+          pure nothrow @safe:
+            int compare(T[] s1, T[] s2)
+            {
+                size_t len = s1.length;
+                if (s2.length < len)
+                    len = s2.length;
+                for (size_t u = 0; u < len; u++)
+                {
+                    if (int c = Floating!T.compare(s1[u], s2[u]))
+                        return c;
+                }
+                if (s1.length < s2.length)
+                    return -1;
+                else if (s1.length > s2.length)
+                    return 1;
+                return 0;
+            }
+        }
+
+        static if (is(U1 == ifloat)) alias F = float;
+        else static if (is(U1 == idouble)) alias F = double;
+        else static if (is(U1 == ireal)) alias F = real;
+        else alias F = U1;
+
+        return () @trusted { return Array!F.compare(cast(F[])s1, cast(F[])s2); }();
+    }
+
+    // integral, char types
+    else
+    {
+        size_t len = s1.length;
+
+        if (s2.length < len)
+            len = s2.length;
+
+        for (size_t u = 0; u < len; u++)
+        {
+            if (s1[u] < s2[u])
+                return -1;
+            else if (s1[u] > s2[u])
+                return 1;
+        }
+
+        if (s1.length < s2.length)
+            return -1;
+        else if (s1.length > s2.length)
+            return 1;
+
+        return 0;
+    }
+}
+
+// integral types
+unittest
+{
+    void compareMinMax(T)()
+    {
+        T[] a = [T.max, T.max];
+        T[] b = [T.min, T.min];
+
+        assert(a > b);
+        assert(b < a);
+    }
+
+    compareMinMax!int;
+    compareMinMax!uint;
+    compareMinMax!long;
+    compareMinMax!ulong;
+    compareMinMax!short;
+    compareMinMax!ushort;
+    compareMinMax!byte;
+    compareMinMax!ubyte;
+    compareMinMax!bool;
+}
+
+// char types
+unittest
+{
+    void compareMinMax(T)()
+    {
+        T[] a = [T.max, T.max];
+        T[] b = [T.min, T.min];
+
+        assert(a > b);
+        assert(b < a);
+    }
+
+    compareMinMax!char;
+    compareMinMax!dchar;
+    compareMinMax!wchar;
+}
+
+// fp types
+unittest
+{
+    void compareMinMax(T)()
+    {
+        T[] a = [T.max, T.max];
+        T[] b = [T.min_normal, T.min_normal];
+
+        assert(a > b);
+        assert(b < a);
+    }
+
+    compareMinMax!real;
+    compareMinMax!float;
+    compareMinMax!double;
+    compareMinMax!ireal;
+    compareMinMax!ifloat;
+    compareMinMax!idouble;
+    compareMinMax!creal;
+    //compareMinMax!cfloat;
+    compareMinMax!cdouble;
+
+    // qualifiers
+    compareMinMax!(const real);
+    compareMinMax!(immutable real);
+}
 
 // Helper functions
 
--- a/src/object.d
+++ b/src/object.d
@@ -3241,19 +3241,76 @@ template RTInfo(T)
     enum RTInfo = null;
 }
 
+private template Floating1(T)
+if (is(T == cfloat) || is(T == cdouble) || is(T == creal))
+{
+    // Use rt.cmath2._Ccmp instead ?
+    int compare(T f1, T f2)
+    {
+        int result;
+
+        if (f1.re < f2.re)
+            result = -1;
+        else if (f1.re > f2.re)
+            result = 1;
+        else if (f1.im < f2.im)
+            result = -1;
+        else if (f1.im > f2.im)
+            result = 1;
+        else
+            result = 0;
+        return result;
+    }
+}
+
+private template Floating1(T)
+if (is(T == float) || is(T == double) || is(T == real))
+{
+    int compare(T d1, T d2)
+    {
+        if (d1 != d1 || d2 != d2) // if either is NaN
+        {
+            if (d1 != d1)
+            {
+                if (d2 != d2)
+                    return 0;
+                return -1;
+            }
+            return 1;
+        }
+        return d1 < d2 ? -1 : (d1 > d2);
+    }
+}
+
+private template Array1(T)
+if (is(T ==  float) || is(T ==  double) || is(T ==  real) ||
+    is(T == cfloat) || is(T == cdouble) || is(T == creal))
+{
+    int compare(T[] s1, T[] s2)
+    {
+        auto len = s1.length;
+        if (s2.length < len)
+            len = s2.length;
+
+        foreach (const u; 0 .. len)
+        {
+            if (int c = Floating1!T.compare(s1[u], s2[u]))
+                return c;
+        }
+        return s1.length < s2.length ? -1 : (s1.length > s2.length);
+    }
+}
+
 int _adCmpT(T1, T2)(T1[] s1, T2[] s2)
 {
-    // U1 gets the unqualified version of T1
-    static if (is(T1 W == immutable W)) alias U1 = W;
-    else static if (is(T1 W == const W)) alias U1 = W;
-    else alias U1 = T1;
-
-    // U2 gets the unqualified version of T2
-    static if (is(T2 U == immutable U)) alias U2 = U;
-    else static if (is(T2 U == const U)) alias U2 = U;
-    else alias U2 = T2;
+    import core.internal.traits : Unqual;
+    alias U1 = Unqual!T1;
+    alias U2 = Unqual!T2;
 
     // objects
+    // Old code went to object.d/TypeInfo_array.compare and then to
+    // object.d/TypeInfo_Class.compare
+    // which combined result in the same code as below.
     static if (is(U1 : Object) && is(U2 : Object))
     {
         auto c  = cast(sizediff_t)(s1.length - s2.length);
@@ -3261,8 +3318,8 @@ int _adCmpT(T1, T2)(T1[] s1, T2[] s2)
         {
             for (size_t u = 0; u < s1.length; u++)
             {
-                Object o1 = s1[u];
-                Object o2 = s2[u];
+                auto o1 = () @trusted { return s1.ptr[u]; }();
+                auto o2 = () @trusted { return s2.ptr[u]; }();
 
                 if (o1 is o2)
                     continue;
@@ -3273,9 +3330,8 @@ int _adCmpT(T1, T2)(T1[] s1, T2[] s2)
                     if (!o2)
                         return 1;
                     c = o1.opCmp(o2);
-                    if (c == 0)
-                        continue;
-                    break;
+                    if (c != 0)
+                        break;
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


andralex:
@9il you may be interested - this lifts array comparisons from druntime into templateland. Also cc @WalterBright @MartinNowak @CyberShadow.

andralex:
To clarify, the plan is: 

* merge this template, which for the time being is not used by generated code
* merge the dmd frontend change, which will insert calls to this template

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
     enum RTInfo = null;
 }
 
+int _adCmpT(T1, T2)(T1[] s1, T2[] s2)
+{
+    // U1 gets the unqualified version of T1
+    static if (is(T1 W == immutable W)) alias U1 = W;
+    else static if (is(T1 W == const W)) alias U1 = W;
+    else alias U1 = T1;
```
CyberShadow: `core.internal.traits.Unqual` ?

andralex: @CyberShadow thx I missed that, thought we only have `Unqual` in Phobos. @somzzz is there a problem if you add a local import here for `core.internal.traits`?

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+    // objects
+    static if (is(U1 : Object) && is(U2 : Object))
+    {
+        auto c  = cast(sizediff_t)(s1.length - s2.length);
+        if (c == 0)
+        {
+            for (size_t u = 0; u < s1.length; u++)
+            {
+                Object o1 = s1[u];
+                Object o2 = s2[u];
```
andralex: Use `auto` for `o1` and `o2` here so you avoid qualifier trouble.

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+
+                // Regard null references as always being "less than"
+                if (o1)
+                {
+                    if (!o2)
+                        return 1;
+                    c = o1.opCmp(o2);
+                    if (c == 0)
+                        continue;
+                    break;
```
andralex: These last three lines are simply:

```d
if (c != 0)
    break;
```

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+                }
+                else
+                {
+                    return -1;
+                }
+            }
+        }
+        return c < 0 ? -1 : c > 0 ? 1 : 0;
```
CyberShadow: A `return` here causes "Statement is not reachable" warnings lower

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+    // U2 gets the unqualified version of T2
+    static if (is(T2 U == immutable U)) alias U2 = U;
+    else static if (is(T2 U == const U)) alias U2 = U;
+    else alias U2 = T2;
+
+    // objects
+    static if (is(U1 : Object) && is(U2 : Object))
+    {
+        auto c  = cast(sizediff_t)(s1.length - s2.length);
+        if (c == 0)
```
andralex: I'm a bit confused here (I know it's the existing implementation, but just for my edification). This is not lexicographical compare because if an array is shorter than the other it always counts as smaller. For example, if we have an array `[hugeWidget]`, it always counts as smaller than `[littleWidget, littleWidget]`. Why is this the case?

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+            }
+        }
+
+        template Floating(T)
+        if (is(T == float) || is(T == double) || is(T == real))
+        {
+          pure nothrow @safe:
+            int compare(T d1, T d2)
+            {
+                if (d1 != d1 || d2 != d2) // if either are NaN
```
andralex: s/are/is/

src/object.d
somzzz: It should be fine, I missed it too. Will make the change, thanks!

src/object.d
CyberShadow: This is a breaking change. Previous behavior was to compare lexicographically (up to the first difference or end of the shortest array).

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+                {
+                    if (d1 != d1)
+                    {
+                        if (d2 != d2)
+                            return 0;
+                        return -1;
+                    }
+                    return 1;
+                }
+                return (d1 == d2) ? 0 : ((d1 < d2) ? -1 : 1);
```
andralex: I think this may be a tad more efficient: `return d1 < d2 ? -1 : (d1 > d2);`

It actually seems to be: https://godbolt.org/g/XhUSPm, https://godbolt.org/g/T1qwSr

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+        || (is(U1 == idouble)  && is(U2 == idouble))
+        || (is(U1 == ireal)    && is(U2 == ireal))
+        || (is(U1 == cfloat)   && is(U2 == cfloat))
+        || (is(U1 == cdouble)  && is(U2 == cdouble))
+        || (is(U1 == creal)    && is(U2 == creal)))
+    {
+        template Floating(T)
+        if (is(T == cfloat) || is(T == cdouble) || is(T == creal))
+        {
+          pure nothrow @safe:
```
CyberShadow: (Nit) unusual indentation style?

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+            int compare(T f1, T f2)
```
CyberShadow: `rt.cmath2._Ccmp`?

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+                for (size_t u = 0; u < len; u++)
+                {
+                    if (int c = Floating!T.compare(s1[u], s2[u]))
+                        return c;
+                }
+                if (s1.length < s2.length)
+                    return -1;
+                else if (s1.length > s2.length)
+                    return 1;
+                return 0;
```
andralex: same here, `return s1.length < s2.length ? -1 : (s1.length > s2.length);`

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+                    result = -1;
+                else if (f1.re > f2.re)
+                    result = 1;
+                else if (f1.im < f2.im)
+                    result = -1;
+                else if (f1.im > f2.im)
+                    result = 1;
+                else
+                    result = 0;
+                return result;
```
andralex: see comments below for making this faster

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+            else if (s1[u] > s2[u])
+                return 1;
+        }
+
+        if (s1.length < s2.length)
+            return -1;
+        else if (s1.length > s2.length)
+            return 1;
+        
+        return 0;
```
andralex: `return s1.length < s2.length ? -1 : (s1.length > s2.length);`

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+// integral types
+unittest
+{
+    void compareMinMax(T)()
+    {
+        T[] a = [T.max, T.max];
+        T[] b = [T.min, T.min];
+        
+        assert(a > b);
+        assert(b < a);
```
andralex: For this PR, please use calls to _adCmp directly. Then we know the function being tested is actually used. Then, after merging, we go back to using operators. I know, tedious!

```d
assert(_opCmpT(a, b) > 0);
assert(_opCmpT(b, a) < 0);
```

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+// char types
+unittest
+{
+    void compareMinMax(T)()
+    {
+        T[] a = [T.max, T.max];
+        T[] b = [T.min, T.min];
+
+        assert(a > b);
+        assert(b < a);
```
andralex: same here

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+// fp types
+unittest
+{
+    void compareMinMax(T)()
+    {
+        T[] a = [T.max, T.max];
+        T[] b = [T.min_normal, T.min_normal];
+        
+        assert(a > b);
+        assert(b < a);
```
andralex: same here

src/object.d
wilzbach: OT: Btw why don't we move `std.meta` and `std.traits` entirely into druntime? This copy/pasting is a bit weird :/

src/object.d
somzzz: This is the same code as before for  ```Object[]``` type.
https://github.com/dlang/druntime/blob/master/src/rt/typeinfo/ti_AC.d#L66

CyberShadow:
I'm not up to speed with current Druntime code but it bothers me somewhat that this implements comparison for not just arrays, but also a bunch of primitive types, as inline functions. I think ideally we would have one single function or overload set `int cmp(A, B)(A a, B b)` which would then have appropriate implementations for various types (thus the inline functions would be moved up as individual overloads of the above). That would also be really useful for all user `opCmp` implementations. Same with equality and hash calculations.

andralex:
@CyberShadow yah, this would be a good step in that direction.

src/object.d
andralex: @somzzz yah all indents should be multiples of 4. As an aside: I don't think we need any of those annotations, they're all deduced. Maybe for clarity.

src/object.d
CyberShadow: Not sure what's going on here but a trivial test indicates that this actually does seem to be a breaking change. https://dpaste.dzfl.pl/21cf68839945 Did I make a mistake?

src/object.d
andralex: I spoke to @somzzz and advised that we should take a larger step once the bit of duplication becomes painful. Since her work will intensify the use of templates in object.d, that's likely to happen.

src/object.d
CyberShadow: Yep, see https://github.com/dlang/druntime/pull/1781#discussion_r104008245

andralex: @WalterBright please illuminate us

quickfur:
Yeah, in order to implement array comparisons we need comparisons for array elements, i.e., basically *all* types.  Why not just unify the logic into a single template that specializes for each concrete type?

src/object.d
andralex: By the comments at https://github.com/dlang/druntime/blob/master/src/rt/typeinfo/ti_AC.d#L19:

```
NOTE: It is sometimes used for arrays of
classes or (incorrectly) interfaces.
But naked `TypeInfo_Array` is mostly used.
See @@@BUG12303@@@.
```

So there's also https://issues.dlang.org/show_bug.cgi?id=12303 to look at.

This seems to suggest that the "wrong" function and `TypeInfo_AC ` as a whole are not used for normal comparison, but perhaps for hash tables or not at all.

andralex:
@quickfur getting there! For now we are carefully replacing one lowering with another.

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+    compareMinMax!wchar;
+}
+
+// fp types
+unittest
+{
+    void compareMinMax(T)()
+    {
+        T[] a = [T.max, T.max];
+        T[] b = [T.min_normal, T.min_normal];
```
UplinkCoder: If you define em like this they will be on the heap.
T[2] please.

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+        return 0;
+    }
+}
+
+// integral types
+unittest
+{
+    void compareMinMax(T)()
+    {
+        T[] a = [T.max, T.max];
```
UplinkCoder: T[2]

src/object.d
WalterBright: This is done to make the comparison faster. No need to load the string into the memory cache if the lengths can be compared.

andralex: @WalterBright I'm not sure you're right. Per my comment above there may be a shorter array that is greater than a longer array. Please recheck.

CyberShadow: That makes sense for equality but comparison? And why is it a breaking change?

UplinkCoder:
@WalterBright short-circuiting on the length only works for opEquals. not for opCmp.

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
     enum RTInfo = null;
 }
 
+int _adCmpT(T1, T2)(T1[] s1, T2[] s2)
```
CyberShadow: It might make sense to turn this into a stub that then calls the implementation with `U1` and `U2` to lower the number of template instantiations, if possible.

src/object.d
WalterBright: Length checks should be first, before the loop, not last.

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+        }
+
+        template Array(T)
+        if (is(T ==  float) || is(T ==  double) || is(T ==  real) ||
+            is(T == cfloat) || is(T == cdouble) || is(T == creal))
+        {
+          pure nothrow @safe:
+            int compare(T[] s1, T[] s2)
+            {
+                size_t len = s1.length;
```
WalterBright: `const` instead of `size_t`

src/object.d
schveiguy: I agree with @CyberShadow, "abc" < "z", you can't short circuit the length.

> By the comments at ...

Note that when I click on that link, the coverage reported by github for the whole file is 0%. I don't think that function is ever called.

schveiguy: BTW, I applaud this PR. And that note is *exactly* why we should be doing this. Nobody knows what is called when.

WalterBright: The existing code for comparing arrays of Objects is:
```
    override int compare(in void* p1, in void* p2) const
    {
        Object[] s1 = *cast(Object[]*)p1;
        Object[] s2 = *cast(Object[]*)p2;
        auto     c  = cast(sizediff_t)(s1.length - s2.length);
        if (c == 0)
        {
            for (size_t u = 0; u < s1.length; u++)
            {
                Object o1 = s1[u];
                Object o2 = s2[u];

                if (o1 is o2)
                    continue;

                // Regard null references as always being "less than"
                if (o1)
                {
                    if (!o2)
                        return 1;
                    c = o1.opCmp(o2);
                    if (c == 0)
                        continue;
                    break;
                }
                else
                {
                    return -1;
                }
            }
        }
        return c < 0 ? -1 : c > 0 ? 1 : 0;
    }
```
So I don't see a breaking change.

WalterBright: See druntime/src/rt/typeinfo/ti_AC.d

CyberShadow: Please see the linked comment. I provided a test program.

WalterBright: For non-Objects, the length is checked after the contents (for the current implementation). For Objects, the length is checked first.

CyberShadow: Then `is(U1 : Object) && is(U2 : Object)` need to be changed to `is(U1 == Object) && is(U2 == Object)`?

CyberShadow: @WalterBright By `Object` do you mean the variable type or the instance type? It's still a change in behavior when the variable is `Object` but the instance isn't, so never mind my comment above, it's not the correct fix.

jpf91:
> merge the dmd frontend change, which will insert calls to this template

Just a small request for this and similar changes: Make DMD fail with some nice error message when the template can't be found iff it actually needs to be used by some array comparison. This will please the `betterC` crowd as they'll get a nice error when a custom runtime doesn't implement some function instead of the linker error we get now. Additionally as long as they don't use array comparisons they won't have to implement this template.

src/object.d
schveiguy: Note, if I grep for `TypeInfo_Array` in dmd source, I get many hits. If I grep for `TypeInfo_AC`, I get nothing. When is this TypeInfo used?

schveiguy: > When is this TypeInfo used?

Answer: never #1782

src/object.d
schveiguy: In fact, you need non-qualified objects. `Object.opCmp` requires it.

andralex:
Let's also see how https://github.com/dlang/druntime/pull/1783 fares.

src/object.d
WalterBright: It looks like I was wrong. The following code:
```
int cmp(Object[] a, Object[] b)
{
    return a < b;
}
```
generates:
```
_D18TypeInfo_AC6Object6__initZ  comdat
        dd      offset FLAT:_D14TypeInfo_Array6__vtblZ
        db      000h,000h,000h,000h     ;....
        dd      offset FLAT:_D6Object7__ClassZ

_D3bug3cmpFAC6ObjectAC6ObjectZi comdat
        assume  CS:_D3bug3cmpFAC6ObjectAC6ObjectZi
L0:             mov     EAX,offset FLAT:_D18TypeInfo_AC6Object6__initZ
                push    EAX
                push    dword ptr 0Ch[ESP]
                push    dword ptr 0Ch[ESP]
                push    dword ptr 01Ch[ESP]
                push    dword ptr 01Ch[ESP]
                call    near ptr __adCmp2
                shr     EAX,01Fh
                add     ESP,014h
                ret     010h
```
meaning it is calling `TypeInfo_Array` rather than `TypeInfo_AC`. The latter is never called any more that I can see. I don't know when this changed, it's been a long time since I've looked at it. This dead code should be removed from druntime. The same goes for `TypeInfo_Ag`.

The code to generate a typeinfo looks like this, in typinf.d:
```
extern (C++) TypeInfoDeclaration getTypeInfoDeclaration(Type t)
{
    //printf("Type::getTypeInfoDeclaration() %s\n", t.toChars());
    switch (t.ty)
    {
    case Tpointer:
        return TypeInfoPointerDeclaration.create(t);
    case Tarray:
        return TypeInfoArrayDeclaration.create(t);
    case Tsarray:
        return TypeInfoStaticArrayDeclaration.create(t);
    case Taarray:
        return TypeInfoAssociativeArrayDeclaration.create(t);
    case Tstruct:
        return TypeInfoStructDeclaration.create(t);
    case Tvector:
        return TypeInfoVectorDeclaration.create(t);
    case Tenum:
        return TypeInfoEnumDeclaration.create(t);
    case Tfunction:
        return TypeInfoFunctionDeclaration.create(t);
    case Tdelegate:
        return TypeInfoDelegateDeclaration.create(t);
    case Ttuple:
        return TypeInfoTupleDeclaration.create(t);
    case Tclass:
        if ((cast(TypeClass)t).sym.isInterfaceDeclaration())
            return TypeInfoInterfaceDeclaration.create(t);
        else
            return TypeInfoClassDeclaration.create(t);

    default:
        return TypeInfoDeclaration.create(t);
    }
}
```
and from todt.d:
```
    override void visit(TypeInfoArrayDeclaration d)
    {
        //printf("TypeInfoArrayDeclaration.toDt()\n");
        verifyStructSize(Type.typeinfoarray, 3 * Target.ptrsize);

        dtb.xoff(toVtblSymbol(Type.typeinfoarray), 0);    // vtbl for TypeInfo_Array
        dtb.size(0);                                     // monitor

        assert(d.tinfo.ty == Tarray);

        TypeDArray tc = cast(TypeDArray)d.tinfo;

        genTypeInfo(tc.next, null);
        dtb.xoff(toSymbol(tc.next.vtinfo), 0); // TypeInfo for array of type
    }
```
where there are no special cases for certain array types.

andralex: Ha ha ha, great minds think alike! Just sent out https://github.com/dlang/druntime/pull/1783 before seeing this!

schveiguy:
> Let's also see how #1783 fares.

#1782 (GMTA) shows that it passes all tests (at least on two platforms before something was merged). I think `TypeInfo_AC` is not used anywhere.

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+
+        static if (is(U1 == ifloat)) alias F = float;
+        else static if (is(U1 == idouble)) alias F = double;
+        else static if (is(U1 == ireal)) alias F = real;
+        else alias F = U1;
+
+        return () @trusted { return Array!F.compare(cast(F[])s1, cast(F[])s2); }();
+    }
+
+    // integral, char types
```
WalterBright: The troubles with this generic code:

1. it is slow, especially for strings. Contrast this with `dstrcmp()` which is currently used for strings which uses `memcmp` which is heavily optimized

2. it creates separate implementations for sbyte and char and immutable(char) and const(char), although those implementations will be identical. It also creates these implementations over and over and over, because everyone uses strings. This is not a runtime problem because the duplicate string implementations are merged, but it is a problem for compilation speed.

3. The use of s1[u] means an array bounds check is (uselessly) generated, further slowing things down.

I suggest that many cases can be combined, and then forwarded to a heavily optimized non-template function that implements it, such as `dstrcmp()`.

WalterBright: For example, float, ifloat, const(float), immutable(float), const(ifloat), immutable(ifloat) should all be forwarded to one non-template function. Ditto for void, byte, char, and the qualified variants of them. Etc.

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+    else
+    {
+        size_t len = s1.length;
+
+        if (s2.length < len)
+            len = s2.length;
+
+        for (size_t u = 0; u < len; u++)
```
WalterBright: `foreach (const u; 0 .. len)

src/object.d
andralex: @WalterBright yah that's the next good step. Until now @somzzz has worked toward getting together something that works amid all subtleties discussed here and elsewhere.

src/object.d
andralex: Point stands - `auto` ensures unique point of maintenance.

andralex:
BTW I figured a completely jump-free 3-way compare: `(a > b) - (a < b)`. This generates no jumps with gcc: https://godbolt.org/g/RVJIqw.

src/object.d
WalterBright: Next step, sure, but I am concerned that the next step doesn't get taken soon enough and we release a version that has substantially degraded performance. There's no way the manual loop will come close to `memcmp()`.

WalterBright:
I prefer to not use urls that may not last for a while. The relevant bits of https://godbolt.org/g/RVJIqw are:
```
int test3(double a, double b) {
  return (a > b) - (a < b);
}
```
and:
```
        xor     eax, eax
        ucomisd xmm0, xmm1
        seta    al
        xor     edx, edx
        ucomisd xmm1, xmm0
        seta    dl
        sub     eax, edx
        ret
```

UplinkCoder:
@andralex check the performance. seta is very expensive on some circumstances.

CyberShadow:
> `(a > b) - (a < b)`

This is fine only for primitive types. For non-primitive types, this causes two calls to `opCmp` or the Druntime comparison operators. If the types themselves then contain other non-primitive types, this will in turn cause a total of four calls (per instance) to them, thus it becomes 2^^(nesting depth of the data structure) comparisons.

I'm not sure how the old code worked, but it's possible that to avoid performance regressions, we **must** call `opCmp` or the old Druntime comparison functions directly instead of doing two comparisons.

src/object.d
jacob-carlborg: `immutable`?

src/object.d
```diff
@@ -3241,6 +3241,226 @@ template RTInfo(T)
...
+    // floating point types
+    static if ((is(U1 == float) && is(U2 == float))
+        || (is(U1 == double)   && is(U2 == double))
+        || (is(U1 == real)     && is(U2 == real))
+        || (is(U1 == ifloat)   && is(U2 == ifloat))
+        || (is(U1 == idouble)  && is(U2 == idouble))
+        || (is(U1 == ireal)    && is(U2 == ireal))
+        || (is(U1 == cfloat)   && is(U2 == cfloat))
+        || (is(U1 == cdouble)  && is(U2 == cdouble))
+        || (is(U1 == creal)    && is(U2 == creal)))
```
jacob-carlborg: Perhaps extract constraints to a separate template?

jacob-carlborg:
* Would it be useful as a transition, if possible, to forward the existing implementation to this function without modifying the compiler?

* A generic note, I think that the `object` module is getting very crowed. I'm leaning towards that we should completely avoid declarations directly in the `object` module. Instead defining the declarations in other modules and forward them to `object` using `public import` and similar

9il:
> @9il you may be interested - this lifts array comparisons from druntime into templateland. Also cc @WalterBright @MartinNowak @CyberShadow.

Thanks for the ping!

+1 to @WalterBright  performance comments.

andralex:
@CyberShadow yes that form only applies to built-in types.

CyberShadow:
> @CyberShadow yes that form only applies to built-in types.

Well, actually that comment is misleading. It also applies to arrays and structs in the current implementation.

src/object.d
andralex: @WalterBright it's changed later.

src/object.d
andralex: @WalterBright thx!

1. Calling `dstrcmp` didn't work in this approach because it was invoked during compilation. So we wanted to have something simple that passes muster. At this point @somzzz will look into special casing `dstrcmp` with `if (__ctfe)`.
2. Fixing (1) will also fix (2) - the template becomes an inlined one-liner.
3. using `s1.ptr[u]` has a different issue - it's not deducible as @safe. I've explained @somzzz to use the idiom for generic trusted code.

>For example, float, ifloat, const(float), immutable(float), const(ifloat), immutable(ifloat) should all be forwarded to one non-template function.

That is the case already.

>Next step, sure, but I am concerned that the next step doesn't get taken soon enough and we release a version that has substantially degraded performance. There's no way the manual loop will come close to memcmp().

There actually is if the size is short. On a larger note, this is a long-haul project not a drive-by refactoring so no need to worry.

src/object.d
andralex: Some of the code is taken from other parts of the system. I'm all for refactoring code during the movaroo, but this is against your (and my) philosophy that moving is one step and refactoring is another.

I'm fine with this particular nit. but I am getting nervous that people see in this PR the ultimate opportunity to improve everything starting with simple refactoring all the way up to a complete reorganization of `object.d`. Thanks!

src/object.d
andralex: things are fine here

src/object.d
CyberShadow: Could you link to the previous location of this code then for review please?

andralex: @somzzz is out for a couple hours but the code should be easy to find - too many instances actually!

```
$ git grep 'for (size_t u = 0; u < len; u++)'
src/object.d:        for (size_t u = 0; u < len; u++)
src/object.d:        for (size_t u = 0; u < len; u++)
src/object.d:        for (size_t u = 0; u < len; u++)
src/rt/typeinfo/ti_Ag.d:        for (size_t u = 0; u < len; u++)
src/rt/typeinfo/ti_Aint.d:        for (size_t u = 0; u < len; u++)
src/rt/typeinfo/ti_Aint.d:        for (size_t u = 0; u < len; u++)
src/rt/typeinfo/ti_Along.d:        for (size_t u = 0; u < len; u++)
src/rt/typeinfo/ti_Along.d:        for (size_t u = 0; u < len; u++)
src/rt/typeinfo/ti_Ashort.d:        for (size_t u = 0; u < len; u++)
src/rt/typeinfo/ti_Ashort.d:        for (size_t u = 0; u < len; u++)
src/rt/util/typeinfo.d:        for (size_t u = 0; u < len; u++)
src/rt/util/typeinfo.d:        for (size_t u = 0; u < len; u++)
```

andralex:
>@andralex check the performance. seta is very expensive on some circumstances.

@UplinkCoder what would be a few relevant links discussing that? Thx!

src/object.d
CyberShadow: Thanks. Depending on how careful we want to be, it might be prudent to list exactly what code this replaces, so it's possible to check that there are no behavior or performance regressions.

jacob-carlborg: > all the way up to a complete reorganization of `object.d`

If you refer to my comment, I don't think we should refactor `object.d` in this PR, just that the function could have been added to a different module. It's a function that should be called by the compiler, why would it go in `object.d`?

src/object.d
jacob-carlborg: Or even better `__traits(isFloating)`.

src/object.d
CyberShadow: > It's a function that should be called by the compiler, why would it go in object.d?

Non-template functions can be "called" directly because the compiler can insert call to the function's mangled name. However, this doesn't work for templates. I understand that the relevant compiler change will simply cause lower comparisons to an instantiation + invocation of the templated function introduced here, which is why it must be available to all D code that has to make such comparisons.

jacob-carlborg: At least it should be possible to put it in a separate module and import it in `object.d`?

src/object.d
andralex: @jacob-carlborg very nice! cc @somzzz

src/object.d
andralex: @jacob-carlborg moving the function elsewhere creates a visibility definition of its own. As it is right now:

* compiler lowers high-level call to a low-level template
* template is defined in object.d and found by the language rules
* template takes care of the rest

I don't want to divert this into changing the language to define special visibility rules.

andralex: @jacob-carlborg we're good here. What we can do (and are already doing) is import things inside the template if necessary.

jacob-carlborg: @andralex I don't see why visibility rules would need to change. As long as the compiler has seen a template it can instantiate and call it? Worst case scenario we add a public import in `object.d`.

andralex: @jacob-carlborg Where has the compiler seen said template? On a larger topic, what are you trying to accomplish? What's wrong with object.d? You do realize that importing another file for a few hundred lines of code is a large performance penalty in compilation speed?

jacob-carlborg: > Where has the compiler seen said template?

I guess an import is needed.

> On a larger topic, what are you trying to accomplish? What's wrong with object.d?

It's messy, too much code in one place.

> You do realize that importing another file for a few hundred lines of code is a large performance penalty in compilation speed?

Have you benchmarked this specific case?

andralex: >> Where has the compiler seen said template?

>I guess an import is needed.

Where would the import take place? You do understand how lowering works, right?

* Compiler lowers high-level code to lower-level code within the same language
* Whatever the lower code is, it obeys the usual language rules

So if the compiler generates a call to `_adCmpT` then it must be found somewhere? Well, if only there was a module that is automatically imported so fundamental lowered artifacts would be found in it!

>> On a larger topic, what are you trying to accomplish? What's wrong with object.d?

>It's messy, too much code in one place.

We were good and as of this PR continue being so. As more stuff gets moved to templates by @somzzz 's work, yes, we'll take the usual software engineering measures.

>> You do realize that importing another file for a few hundred lines of code is a large performance penalty in compilation speed?

>Have you benchmarked this specific case?

We are wasting time here.

jpf91: I agree with jacob. We should not add more and more stuff to `object.d` and pollute the global namespace if we can easily avoid it. Lowering does not require the symbol to be visible to user written D code. We already do something similar in GDC to load the shared library types symbol from `rt.sections`, no need to place it into `object.d`:
```c++
  // Try to find the required types and functions in druntime
  Dsymbol *mref = NULL, *compiler_dso_type = NULL, *dso_registry_func = NULL;
  // Ignore error if the file can not be found and simply don't emit Moduleinfo
  unsigned errors = global.startGagging();
  Module *m = Module::load(Loc(), NULL, Id::sectionsModule); //Just load the module in the compiler frontend
  global.endGagging(errors);//

  if (m)
    {
      m->importedFrom = m;
      m->importAll (NULL);
      m->semantic(NULL);
      m->semantic2(NULL);
      m->semantic3(NULL);
      // These symbols are normally private to the module they're declared in,
      // but for internal compiler lookups, their visibility is discarded.
      int sflags = IgnoreErrors | IgnoreSymbolVisibility;
      mref = m->search(Loc(), Id::Dmodule_ref, sflags);
      compiler_dso_type = m->search(Loc(), Id::compiler_dso_type, sflags);
      dso_registry_func = m->search(Loc(), Id::dso_registry_func, sflags);
    }
```

This can certainly work for templates to in some case. From a high level POV this just means looking up these special symbols does not require an explicit import.

CyberShadow: @jpf91 @jacob-carlborg There is no problem moving it away from object.d after this PR is merged, right? If so, let's take things one step at a time.

andralex: I don't see a matter of creeping pollution, especially given that access to the low-level three-way comparison may be useful to client code as well.

As a matter of principle I am opposed to the compiler having special prerogatives (i.e. lookup here). It is fundamental to lowering that the lowered code is _within_ the same language as the higher-level code and therefore can be reasoned about without knowing about special treatment. So yes, using runtime support under the hood should be fine for certain cases but we will use lowering as the preferred course of action.

jacob-carlborg: > There is no problem moving it away from object.d after this PR is merged, right?

@CyberShadow  No. But in my experience that rarely happens if you don't do it immediately. It took `std.algorithm` several years and at least on language change. It still haven't happened to `std.datetime`.

jacob-carlborg: > So if the compiler generates a call to _adCmpT then it must be found somewhere? Well, if only there was a module that is automatically imported so fundamental lowered artifacts would be found in it!

The import would be placed in `object.d`.

jacob-carlborg: > especially given that access to the low-level three-way comparison may be useful to client code as well

How would that not be accessible if it was located in a different module? Just import it like ant other module.

> As a matter of principle I am opposed to the compiler having special prerogatives.

@andralex I don't think the compiler should have special prerogatives either. Please explain why that is necessary. As I've said above, the import would be placed in `object.d`.

somzzz:
Thank you everybody for your comments and feedback!

I tried to address most nits and the optimizations suggested. For the time being I did not move the code from ```object.d```.

Please let me know what you think.

src/object.d
somzzz: I forgot an ```else```. Fixed.

src/object.d
somzzz: Fixed.

src/object.d
somzzz: Done!

src/object.d
```diff
@@ -3241,6 +3241,269 @@ template RTInfo(T)
...
+    // which combined result in the same code as below.
+    static if (is(U1 : Object) && is(U2 : Object))
+    {
+        auto c  = cast(sizediff_t)(s1.length - s2.length);
+        if (c == 0)
+        {
+            for (size_t u = 0; u < s1.length; u++)
+            {
+                auto o1 = () @trusted { return s1.ptr[u]; }();
+                auto o2 = () @trusted { return s2.ptr[u]; }();
```
JackStouffer: I've noticed in many cases that DMD still fails to inline trusted anonymous functions. Please make sure that this is inlined.

src/object.d
```diff
@@ -3241,6 +3241,269 @@ template RTInfo(T)
     enum RTInfo = null;
 }
 
+private template Floating1(T)
```
andralex: Regarding the fact that this is at top level (ideally it should be nested within the only place where it's necessary): https://issues.dlang.org/show_bug.cgi?id=17242

andralex: I see @WalterBright replied to that providing a rationale. At any rate, something we can do right now:

```d
    // inside _adCmpT
    static int compare(X f1, X f2)
    {
        static if (is(T == cfloat) || is(T == cdouble) || is(T == creal))
        {
            ...
        }
        else static if (is(T == float) || is(T == double) || is(T == real))
        {
            ...
        }
        else static assert(false, "Internal error");
    }
```

Then use compare.

src/object.d
```diff
@@ -3241,6 +3241,269 @@ template RTInfo(T)
...
+                    return 0;
+                return -1;
+            }
+            return 1;
+        }
+        return d1 < d2 ? -1 : (d1 > d2);
+    }
+}
+
+private template Array1(T)
```
andralex: This, too, should be nestable after the above. There's no need for a template, just write a static function inside _adCmpT.

src/object.d
```diff
@@ -3241,6 +3241,269 @@ template RTInfo(T)
...
+                }
+                else
+                {
+                    return -1;
+                }
+            }
+        }
+        return c < 0 ? -1 : (c > 0);
+    }
+
```
andralex: this empty line between if and else is unusual

src/object.d
```diff
@@ -3241,6 +3241,269 @@ template RTInfo(T)
...
+        else alias F = U1;
+
+        return () @trusted { return Array1!F.compare(cast(F[])s1, cast(F[])s2); }();
+    }
+
+    // char types = > dstrcmp
+    else static if ((is(U1 == ubyte) && is(U2 == ubyte))
+        || (is(U1 == void) && is(U2 == void))
+        || (is(U1 == bool) && is(U2 == bool))
+        || (is(U1 == char) && is(U2 == char)))
```
andralex: Add `byte` as well

src/object.d
```diff
@@ -3241,6 +3241,269 @@ template RTInfo(T)
...
+    {
+        if (!__ctfe)
+        {
+            import core.internal.string : dstrcmp;
+            return () @trusted { return dstrcmp(cast(char[])s1, cast(char[])s2); }();
+        }
+        else
+        {
+            // pretty ugly...
```
andralex: Yah, it's difficult to get rid of this duplication. One possibility:

```
... from line 3357 above ...
else
{
    static int lexicographicalCompare(U1 a, U2 b) { ... }
    static if ((is(U1 == ubyte) && is(U2 == ubyte))
         || (is(U1 == byte) && is(U2 == byte))
         || (is(U1 == void) && is(U2 == void))
         || (is(U1 == bool) && is(U2 == bool))
         || (is(U1 == char) && is(U2 == char)))
            if (!__ctfe) return () @trusted { return dstrcmp(cast(char[])s1, cast(char[])s2); }();
    return lexicographicalCompare(s1, s2);
}
```

src/object.d
```diff
@@ -3241,6 +3241,269 @@ template RTInfo(T)
...
+                if (e1 < e2)
+                    return -1;
+                else if (e1 > e2)
+                    return 1;
+            }
+
+            return s1.length < s2.length ? -1 : (s1.length > s2.length);
+        }
+    }
+
```
andralex: remove

src/object.d
andralex: >The import would be placed in object.d.

Think it over - that wouldn't work (unless you make the import public, which I understand you'd want to avoid).

CyberShadow: > It took std.algorithm several years and at least one language change. 

I don't think that's a fair comparison, because we're not dealing with a public interface here.

> It still haven't happened to std.datetime.

I think that was only because no one was working on it. Jonathan was at some point, but I guess it was too much work to do properly.

src/object.d
andralex: Yah, sadly that's the case. One mildly unpleasant way to get this done is to define a static function inside _adCmp:

```d
static @trusted ref X at(X)(X[] r, size_t i) { return r.ptr[i]; }
```

Then use `s1.at(u)` etc.

src/object.d
andralex: Actually now that I think of it there's no need to define a separate function?

src/object.d
somzzz: Byte falls under ``TypeInfo_Ag``` which uses the default comparison under ```else```. I tried to stick with that behavior for this PR.

https://github.com/dlang/druntime/blob/master/src/rt/typeinfo/ti_Ag.d#L20

src/object.d
somzzz: Without optimization, the compiler still generates a call to a function:
```
int test1(Object[] a, Object[] b) {
    static @trusted ref R at(R)(R[] r, size_t i) { return r.ptr[i]; }
	auto o1 = at(a, 0);
	auto o2 = at(b, 0);
    return 0;
}

int test2(Object[] a, Object[] b) {
	auto e1 = () @trusted { return a.ptr[0]; }();
	auto e2 = () @trusted { return b.ptr[0]; }();
  
  	if (e1 != e2) {}
  	return 0;
}
```

```
---
call    pure nothrow ref @nogc @trusted Object example.test1(Object[], Object[]).at!(Object).at(Object[], ulong)
---
```

From https://godbolt.org/g/IykECY

src/object.d
WalterBright: bytes are signed. ubyte, bool, void, char are all unsigned, and so compare differently

src/object.d
WalterBright: > I've noticed in many cases that DMD still fails to inline trusted anonymous functions.

Please enumerate them and file a bug report.

JackStouffer: @WalterBright 
Issues with nested functions:
https://issues.dlang.org/show_bug.cgi?id=17210 
https://issues.dlang.org/show_bug.cgi?id=15159

Other inlining issues:
https://issues.dlang.org/show_bug.cgi?id=7625
https://issues.dlang.org/show_bug.cgi?id=14943
https://issues.dlang.org/show_bug.cgi?id=16360

WalterBright:
Looks good to me, with @andralex 's changes

src/object.d
andralex: @somzzz for efficiency assessment we're interested only in optimized builds, i.e. `-release -inline -O`. Please advise if `at` helps those cases, thanks!

ibuclaw:
> A generic note, I think that the object module is getting very crowed. I'm leaning towards that we should completely avoid declarations directly in the object module.

I don't think that there would be anything stopping object.d being turned into a package.

src/object.d
jacob-carlborg: > I don't think that's a fair comparison, because we're not dealing with a public interface here.

@CyberShadow  What is indicating this is not a public interface? It's public (in the default imported module), not comments indicating otherwise (not comments at all!!) and it's not even using a double underscore prefix which usually indicates internal symbols.

> I think that was only because no one was working on it. Jonathan was at some point, but I guess it was too much work to do properly.

Even more reason to start before the module gets even bigger.

CyberShadow: > What is indicating this is not a public interface?

Common sense. 

We are changing the implementation of the comparison operator, which is invoked by compiler-generated code directly.

All of the Druntime underscore symbols are internal, implementation-specific and not part of the language.

Hypothetically, would it make you happy if we'd pretend another underscore to the template name?

> Even more reason to start before the module gets even bigger.

It is also important to separate concerns. Instead of deciding on merging one change now we'll be deciding on two changes, the second one being a bikeshed issue. Please open a PR after this is merged or let's take the discussion elsewhere.

jacob-carlborg: > Think it over - that wouldn't work (unless you make the import public, which I understand you'd want to avoid).

@andralex I've already suggested `public import` twice.

What's even more concerning than the code side, is that it adds additional public symbols to the module that is by default imported. That can cause conflicts with existing symbols. What if I already have a function with the same name and a matching signature, will the compiler use that one instead? Or will it use a fully qualified name when invoking it?

What would be even better, and solve both of these problems, is to put the internal symbols in a separate module. When the compiler needs to access an internal symbol it would add a import in the lowering process for that module. It would use a renamed import to a compiler generated symbol. Example:

```d
import __someCompilerGeneratedSymbol123 = rt.internal;
__someCompilerGeneratedSymbol123._adCmpT(a, b);
```

It could even, if necessary, combine that with a selective import to only import what's exactly needed.

andralex: >All of the Druntime underscore symbols are internal, implementation-specific and not part of the language.

Actually `_adCmpT` is a legit symbol. I was thinking we need to replace it with a double-underscore symbol. @somzzz could you please replace it with `__cmp` here and in the upcoming compiler patch? Thanks!

andralex: >What would be even better, and solve both of these problems, is to put the internal symbols in a separate module.

Sorry, no.

src/object.d
CyberShadow: Why the name? Never would guess that something called `Floating1` deals with comparison.

src/object.d
```diff
@@ -3241,6 +3241,269 @@ template RTInfo(T)
...
+private template Floating1(T)
+if (is(T == float) || is(T == double) || is(T == real))
+{
+    int compare(T d1, T d2)
+    {
+        if (d1 != d1 || d2 != d2) // if either is NaN
+        {
+            if (d1 != d1)
+            {
+                if (d2 != d2)
```
CyberShadow: Is there really no better way to do this than the worst case of 4 comparisons?

src/object.d
andralex: It was carried from somewhere else. Anyhow, I think all that code can be massaged with relative ease inside `_adCmpT`, pardon, `__cmp`.

src/object.d
```diff
@@ -3241,6 +3241,269 @@ template RTInfo(T)
...
+    alias U2 = Unqual!T2;
+
+    // objects
+    // Old code went to object.d/TypeInfo_array.compare and then to
+    // object.d/TypeInfo_Class.compare
+    // which combined result in the same code as below.
+    static if (is(U1 : Object) && is(U2 : Object))
+    {
+        auto c  = cast(sizediff_t)(s1.length - s2.length);
+        if (c == 0)
```
CyberShadow: This breaking change hasn't been addressed yet.

src/object.d
```diff
@@ -3241,6 +3241,269 @@ template RTInfo(T)
...
+                    return -1;
+                else if (e1 > e2)
+                    return 1;
+            }
+
+            return s1.length < s2.length ? -1 : (s1.length > s2.length);
+        }
+    }
+
+    // integral
```
CyberShadow: As before, this isn't just integral, is it? Also structs and (nested) arrays.

src/object.d
```diff
@@ -3241,6 +3241,269 @@ template RTInfo(T)
...
+
+        foreach (const u; 0 .. len)
+        {
+            auto e1 = () @trusted { return s1.ptr[u]; }();
+            auto e2 = () @trusted { return s2.ptr[u]; }();
+
+            if (e1 < e2)
+                return -1;
+            else if (e1 > e2)
+                return 1;
```
CyberShadow: Again, two comparisons leading to exponential complexity for non-basic types.

src/object.d
andralex: Struck my eye too. This code, too, was carried from legacy parts of the system.

src/object.d
CyberShadow: If we are going to turn more things into template instantiations, we probably can't keep putting them in `object.d` forever, so we will probably need to move things out at some point ... just not now.

andralex: @CyberShadow yah, the ticket is local `import`s that are loaded on demand if user code triggers this or that functionality. Pay as you go at its best.

src/object.d
andralex: @somzzz when you fix this please add unittests that wouldn't work with the current code and will pass with the fixed code. Thanks!

andralex:
>I don't think that there would be anything stopping object.d being turned into a package.

Compilation speed is a major concern.

src/object.d
CyberShadow: Here's some tests I wrote:

https://dump.thecybershadow.net/2801c21c14f64b3cc72bd5bbb7622888/test.d

Perhaps these should be added to the DMD test suite.

jacob-carlborg:
> Compilation speed is a major concern.

Turning all type info code into templates is more likely to decrease the compilation speed than a separate module.

CyberShadow:
That's a fallacy... if an overall desired event has an undesired outcome, that doesn't mean we should stop caring about other instances of that outcome.

src/object.d
andralex: @CyberShadow cool, thanks! Since druntime actually implements the logic, the tests belong with it. Can you please submit them as a PR? Thanks!

CyberShadow: OK (didn't know Druntime has a non-unittest testsuite): https://github.com/dlang/druntime/pull/1784

src/object.d
andralex: @CyberShadow what would be the simplest way to address this?

CyberShadow: Well, do what the old code did. Which I would guess would be: essentially for all types for which we can get an integer comparison result, we should do that - call `opCmp` for structs/classes and / `__cmp` (i.e. the template introduced here) for arrays.

ibuclaw:
> Compilation speed is a major concern.

My only response to that would be std.datetime.

src/object.d
somzzz: I've just checked: using ```-release -inline -O``` doesn't generate calls to ```at```.
Interesting enough, there still is an unused function stub for ```at``` in the executable. It is never called/jumped at as far as I can see.

andralex: @somzzz thanks! then code using `at` should be used throughout.

src/object.d
somzzz: It is strange that these tests passed with that faulty implementation as well. It should be fixed now. Thanks!

src/object.d
```diff
@@ -3241,6 +3241,268 @@ template RTInfo(T)
     enum RTInfo = null;
 }
```
andralex: Please add a regular (not documentation) comment stating that this function is called by the lowering of ordering comparisons.

src/object.d
```diff
@@ -3241,6 +3241,268 @@ template RTInfo(T)
...
+            {
+                auto c = _adCmp(e1, e2);
+                if (c != 0)
+                    return c < 0 ? -1 : 1;
+            }
+            else
+            {
+                if (e1 < e2)
+                    return -1;
+                else if (e1 > e2)
```
andralex: remove the else

src/object.d
```diff
@@ -3241,6 +3241,268 @@ template RTInfo(T)
...
+    }
+    // floating point types
+    else static if (__traits(isFloating, U1))
+    {
+        static if (is(U1 == ifloat)) alias F = float;
+        else static if (is(U1 == idouble)) alias F = double;
+        else static if (is(U1 == ireal)) alias F = real;
+        else alias F = U1;
+
+        static int compare(F f1, F f2)
```
andralex: Cool! This is getting ready, but I now realized I gave you the wrong advice when I told you to move these functions inside `__cmp`. This has to do with a subtlety in the way nested static functions are generated: even though you constrained `compare` properly to only accept certain types, the compiler will still generate one instance for each distinct outer template.

I'll approve this (if there are no other issues) and will follow with a fixup.

src/object.d
```diff
@@ -3241,6 +3241,268 @@ template RTInfo(T)
...
+            auto e2 = at(s2, u);
+
+            // structs
+            static if (__traits(compiles, e1.opCmp(e2)))
+            {
+                auto c = e1.opCmp(e2);
+                if (c != 0)
+                    return c < 0 ? -1 : 1;
+            }
+            else static if (__traits(compiles, _adCmp(e1, e2)))
```
andralex: Wouldn't this be `__cmp`?

andralex:
@CyberShadow I plan to follow up with a bit of factorization, any lingering issues for this PR?

src/object.d
somzzz: I used replace in my editor, it failed.  Fixing.

src/object.d
somzzz: Is this code better?
```
 if (e1 == e2)
   continue;
 return e1 - e2;
```

src/object.d
```diff
@@ -3364,7 +3367,8 @@ int __cmp(T1, T2)(T1[] s1, T2[] s2)
 
                 if (e1 < e2)
                     return -1;
-                else if (e1 > e2)
+
+                if (e1 > e2)
```
andralex: noice

src/object.d
```diff
@@ -3241,6 +3241,273 @@ template RTInfo(T)
...
+        {
+            auto e1 = at(s1, u);
+            auto e2 = at(s2, u);
+
+            // structs
+            static if (__traits(compiles, e1.opCmp(e2)))
+            {
+                auto c = e1.opCmp(e2);
+                if (c != 0)
+                    return c < 0 ? -1 : 1;
```
CyberShadow: Any reason not to do `return c`?

CyberShadow:
> any lingering issues for this PR?

Looks like my outstanding complaints are resolved, I don't see anything else so far. Would be good to see the results from Martin's project tester for the PR that enables using this code.

CyberShadow:
> Is this code better?
> ```
>  if (e1 == e2)
>    continue;
>  return e1 - e2;
> ```

I can't see where that comment was made any more, but if that's an option, what about `auto c = e1 - e2; if (c) return c;`

somzzz:
@CyberShadow I missunderstood something @andralex said when I wrote that comment. It's irrelevant.

jacob-carlborg:
So here's the first question about these new functions [1]. This what happens when internal functionality is not properly hidden. Unfortunately the functions that were added later did not have the comment warning about that the function is only used by the compiler. I would not be surprised if someone starts to call them manually soon.

[1] http://forum.dlang.org/post/oddrluarorcmvepgbpcs@forum.dlang.org

andralex:
@jacob-carlborg there's no need for changes at user level. How do you mean functionality is not hidden? Symbols starting with a double underscore are reserved, and the compiler lowers comparisons automatically.

jacob-carlborg:
> there's no need for changes at user level

Exactly. But it's nothing that indicates that except the double underscore, hence the question was asked in the newsgroup. There's not even a comment indicating it shouldn't be called directly by user code.

> How do you mean functionality is not hidden.

The compiler doesn't stop you from accessing the symbol.

andralex:
@jacob-carlborg I see, thanks. I think the double underscore is perfect - it's warning against unwitting use, yet it still makes the symbol accessible for people who know what they're doing and are willing to forgo portability.

PR #1784, opened 2017-03-04
Add tests for class/struct/array comparisons

CyberShadow:
PR #1781 is moving array comparisons to a new template; the initial implementation included changes which may have introduced performance and behavior regressions.

Add explicit tests for these cases to detect such regressions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@ druntime.json
 .DS_Store
 trace.def
 trace.log
-Makefile
+/Makefile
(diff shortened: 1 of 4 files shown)


MartinNowak:
Auto-merge toggled on

andralex:
Thanks!

CyberShadow:
This would be one case where squashing is inappropriate :) Martin already enabled the old auto-merge.

PR #1786, opened 2017-03-04
Remove dmd.conf target from CircleCi

wilzbach:
After the ddmd -> dmd transition the `dmd.conf` target won't be needed as well - in fact will probably error.

See also:
https://github.com/dlang/dmd/pull/6586#discussion_r104297885

CC @RazvanN7 @MartinNowak


wilzbach:
@MartinNowak merging as Dlang-Bot (without admin rights) is only possible once a PR has at least one approval.

PR #1787, opened 2017-03-06
One more pass through __cmp

andralex:
A few more improvements to `__cmp` to reduce number of templates generated and tighten the code.

### DIFF
--- a/src/core/internal/string.d
+++ b/src/core/internal/string.d
@@ -192,16 +192,22 @@ unittest
 
 int dstrcmp( scope const char[] s1, scope const char[] s2 ) @trusted
 {
-    import core.stdc.string : memcmp;
+    immutable len = s1.length <= s2.length ? s1.length : s2.length;
+    if (__ctfe)
+    {
+        foreach (const u; 0 .. len)
+        {
+            if (s1[u] != s2[u])
+                return s1[u] > s2[u] ? 1 : -1;
+        }
+    }
+    else
+    {
+        import core.stdc.string : memcmp;
 
-    int  ret = 0;
-    auto len = s1.length;
-    if( s2.length < len )
-        len = s2.length;
-    if( 0 != (ret = memcmp( s1.ptr, s2.ptr, len )) )
-        return ret;
-    return s1.length >  s2.length ? 1 :
-           s1.length == s2.length ? 0 : -1;
+        const ret = memcmp( s1.ptr, s2.ptr, len );
+        if( ret )
+            return ret;
+    }
+    return s1.length < s2.length ? -1 : (s1.length > s2.length);
 }
-
-
--- a/src/object.d
+++ b/src/object.d
@@ -3248,13 +3248,13 @@ if (__traits(isFloating, F))
     static if (is(F == cfloat) || is(F == cdouble) || is(F == creal))
     {
         // Use rt.cmath2._Ccmp instead ?
-        auto r = __cmp(lhs.re, lhs.re);
-        if (!r) r = __cmp(lhs.im, lhs.im);
+        auto r = __cmp(lhs.re, rhs.re);
+        if (!r) r = __cmp(lhs.im, rhs.im);
         return r;
     }
     else
     {
-        return (lhs > rhs) - (lhs > rhs);
+        return (lhs > rhs) - (lhs < rhs);
     }
 }
 
--- a/src/object.d
+++ b/src/object.d
@@ -3258,11 +3258,31 @@ if (is(Obj : Object))
 int __cmp(T)(const T[] lhs, const T[] rhs) @trusted
 if (__traits(isScalar, T))
 {
-    static if (is(T == ubyte) || is(T == void) || is(T == bool)
-        || is(T == char))
+    // Compute U as the implementation type for T
+    static if (is(T == ubyte) || is(T == void) || is(T == bool))
+        alias U = char;
+    else static if (is(T == wchar))
+        alias U = ushort;
+    else static if (is(T == dchar))
+        alias U = uint;
+    else static if (is(T == ifloat))
+        alias U = float;
+    else static if (is(T == idouble))
+        alias U = double;
+    else static if (is(T == ireal))
+        alias U = real;
+    else
+        alias U = T;
+
+    static if (is(U == char))
     {
         import core.internal.string : dstrcmp;
         return dstrcmp(cast(char[]) lhs, cast(char[]) rhs);
+    }
+    else static if (!is(U == T))
+    {
+        // Reuse another implementation
+        return __cmp(cast(U[]) lhs, cast(U[]) rhs);
     }
     else
     {
--- a/src/object.d
+++ b/src/object.d
@@ -3297,7 +3297,7 @@ if (__traits(isScalar, T))
                 {
                     // Use rt.cmath2._Ccmp instead ?
                     auto r = (a.re > b.re) - (a.re < b.re);
-                    if (!r) r = (a.im > b.im) - (a.im < a.im);
+                    if (!r) r = (a.im > b.im) - (a.im < b.im);
                 }
                 else
                 {
--- a/src/object.d
+++ b/src/object.d
@@ -3242,19 +3242,19 @@ template RTInfo(T)
 }
 
 // Compare floating point numbers for ordering (one instantiation per width).
-private int __cmp(F)(const F f1, const F f2)
+private int __cmp(F)(const F lhs, const F rhs)
 if (__traits(isFloating, F))
 {
     static if (is(F == cfloat) || is(F == cdouble) || is(F == creal))
     {
         // Use rt.cmath2._Ccmp instead ?
-        auto r = __cmp(f1.re, f2.re);
-        if (!r) r = __cmp(f1.im, f2.im);
+        auto r = __cmp(lhs.re, lhs.re);
+        if (!r) r = __cmp(lhs.im, lhs.im);
         return r;
     }
     else
     {
-        return (f1 > f2) - (f2 > f1);
+        return (lhs > rhs) - (lhs > rhs);
     }
 }
 
@@ -3262,21 +3262,14 @@ if (__traits(isFloating, F))
 private int __cmp(Obj)(Obj lhs, Obj rhs)
 if (is(Obj : Object))
 {
-    if (o1 is o2)
+    if (lhs is rhs)
         return 0;
     // Regard null references as always being "less than"
-    if (o1)
-    {
-        if (!o2)
-            return 1;
-        auto c = o1.opCmp(o2);
-        if (c != 0)
-            return c;
-    }
-    else
-    {
+    if (!lhs)
         return -1;
-    }
+    if (!rhs)
+        return 1;
+    return lhs.opCmp(rhs);
 }
 
 // This function is called by the compiler when dealing with array
@@ -3413,14 +3406,14 @@ int __cmp(T1, T2)(T1[] s1, T2[] s2)
 }
 
 //objects
-unittest
+@safe unittest
 {
     class C
     {
         int i;
         this(int i) { this.i = i; }
 
-        override int opCmp(Object c) const
+        override int opCmp(Object c) const @safe
         {
             return i - (cast(C)c).i;
         }
@@ -3428,11 +3421,15 @@ unittest
 
     auto c1 = new C(1);
     auto c2 = new C(2);
+    assert(__cmp(c1, null) > 0);
+    assert(__cmp(null, c1) < 0);
+    assert(__cmp(c1, c1) == 0);
+    assert(__cmp(c1, c2) < 0);
+    assert(__cmp(c2, c1) > 0);
 
-    assert(__cmp([c1, c1], [c2, c2]) < 0);
+    assert(__cmp([c1, c1][], [c2, c2][]) < 0);
     assert(__cmp([c2, c2], [c1, c1]) > 0);
 }
-
 
 // Helper functions
 
--- a/src/object.d
+++ b/src/object.d
@@ -3272,10 +3272,34 @@ if (is(Obj : Object))
     return lhs.opCmp(rhs);
 }
 
+int __cmp(T)(const T[] lhs, const T[] rhs) @trusted
+if (__traits(isScalar, T))
+{
+    immutable len = lhs.length <= rhs.length ? lhs.length : rhs.length;
+    foreach (const u; 0 .. len)
+    {
+        static if (__traits(isFloating, T))
+        {
+            auto r = __cmp(lhs.ptr[u], rhs.ptr[u]);
+            if (r) return r;
+        }
+        else static if (is(T == ubyte) || is(T == void) || is(T == bool)
+            || is(T == char))
+        {
+            import core.internal.string : dstrcmp;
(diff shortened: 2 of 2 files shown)


andralex:
cc @WalterBright @somzzz @CyberShadow @MartinNowak

src/core/internal/string.d
```diff
@@ -192,16 +192,22 @@ unittest
 
 int dstrcmp( scope const char[] s1, scope const char[] s2 ) @trusted
 {
-    import core.stdc.string : memcmp;
```
CyberShadow: This (dstrcmp change) really ought to be in its own commit... currently there is no rationale. I see it's to add CTFE support?

src/object.d
```diff
@@ -3287,135 +3302,57 @@ int __cmp(T1, T2)(T1[] s1, T2[] s2)
...
     }
-    // floating point types
-    else static if (__traits(isFloating, U1))
+    // Imaginary types
+    else static if (is(U1 == ifloat) || is(U1 == idouble) || is(U1 == ireal))
     {
+        // Special-case imaginary types to use comparison for "concrete" types
         static if (is(U1 == ifloat)) alias F = float;
         else static if (is(U1 == idouble)) alias F = double;
         else static if (is(U1 == ireal)) alias F = real;
```
CyberShadow: BTW, should `__cmp` perhaps have a constraint that restricts the two arguments to have the same unqualified type? The compiler seems to enforce that when you do the comparison, however if we do somehow end up with an instantiation with e.g. `float[]` and `double[]`, such as an explicit user invocation, then slightly wrong things will happen (such as casting the `double`s to `float`s).

CyberShadow:
Hmm, diff isn't very useful, a bit difficult to review. In such cases splitting changes across multiple commits is helpful.

src/object.d
CyberShadow: Actually casting entire arrays is more than slightly wrong :)

src/object.d
```diff
@@ -3287,135 +3302,57 @@ int __cmp(T1, T2)(T1[] s1, T2[] s2)
...
-                    return 1;
-            }
-
-            return s1.length < s2.length ? -1 : (s1.length > s2.length);
-        }
+        import core.internal.string : dstrcmp;
+        return (() @trusted => dstrcmp(cast(char[])s1, cast(char[])s2))();
     }
-    // integral, struct, nested arrays
+    // Everything else
```
CyberShadow: The previous comment was more descriptive.

src/object.d
```diff
@@ -3287,135 +3302,57 @@ int __cmp(T1, T2)(T1[] s1, T2[] s2)
...
-        auto len = s1.length;
-
-        if (s2.length < len)
-            len = s2.length;
+        immutable len = s1.length <= s2.length ? s1.length : s2.length;
 
         foreach (const u; 0 .. len)
         {
-            auto e1 = at(s1, u);
-            auto e2 = at(s2, u);
```
CyberShadow: Good catch, this avoids a copy (and in fact wouldn't work with `@disable this(this)` structs, would it?)

src/object.d
```diff
@@ -3287,135 +3302,57 @@ int __cmp(T1, T2)(T1[] s1, T2[] s2)
...
-            else static if (__traits(compiles, __cmp(e1, e2)))
+            else static if (__traits(compiles, __cmp(at(s1, u), at(s2, u))))
             {
-                auto c = __cmp(e1, e2);
+                auto c = __cmp(at(s1, u), at(s2, u));
                 if (c != 0)
-                    return c < 0 ? -1 : 1;
+                    return c;
             }
             else
```
CyberShadow: Do you think it would make sense to add another `static if` with `if (__traits(compiles, { int i = at(s1, u) - at(s2, u); })`?

By the way, one thing we need to keep in mind is efficient comparison of structs that don't have an `opCmp`, and those that have an auto-generated one (such as when one of the fields has one). That's a thing, right?

CyberShadow:
> A few more improvements to `__cmp` to reduce number of templates generated and tighten the code.

From that description I thought this would involve making `__cmp` a stub which normalizes constness and then calls an implementation (or itself). That would make sense, right? So as to avoid multiple instantiations for `ubyte[]` / `ubyte[]`, `ubyte[]` / `const(ubyte)[]`, `const(ubyte)[]` / `immutable(ubyte)[]`...

src/core/internal/string.d
andralex: Affirmative.

src/object.d
```diff
@@ -3241,6 +3241,23 @@ template RTInfo(T)
...
+    static if (is(F == cfloat) || is(F == cdouble) || is(F == creal))
+    {
+        // Use rt.cmath2._Ccmp instead ?
+        auto r = __cmp(f1.re, f2.re);
+        if (!r) r = __cmp(f1.im, f2.im);
+        return r;
+    }
+    else static if (is(F == float) || is(F == double) || is(F == real))
+    {
+        return (f1 > f2) - (f2 > f1);
```
andralex: This generates nice code with dmd and gdc, and also much faster than the previous version (and also a bit faster than the obvious alternatives).

src/object.d
andralex: The relationship between `<` and `-` does not apply to all types.

So does the compiler generate `opCmp` for structs always/when necessary?

src/object.d
andralex: Currently I planted a static assert at the beginning of `__cmp`. From the looks of it the qualifiers are passed by the compiler.

src/object.d
CyberShadow: Hmm, doesn't look like it. Curiously, for a struct without `opCmp`, `S(1) < S(2)` doesn't compile, but `[S(1)] < [S(2)]` does. If S has a field with an `opCmp`, it doesn't get called. So I guess that's not something we need to worry about right now.

Though, there's still the question of structs without `opCmp`. I think the compiler just does a `memcmp` right now, but someone would need to check. Ideally we'd want to avoid doing that `memcmp` twice.

src/object.d
CyberShadow: > From the looks of it the qualifiers are passed by the compiler.

I think that should either be enforced or accounted for, especially if we want to expose this function to users as well (and we do want that, as it would be useful for defining `opCmp` in user types).

andralex: Not sure what the action item is.

src/object.d
andralex: It had become untrue due to the changes in this diff. There are only a few leading special cases for objects and imaginary types, so... "Everything that is not a class/interface or imaginary type..."

src/object.d
andralex: @CyberShadow looks like worth raising a bug report over!

src/object.d
CyberShadow: Oops, that was a reply to a different comment (normalizing constness). In which case the action item is either `static assert(is(T1 == const(U1)) && is(T2 == const(U2)))` or `static if (!is(T1 == const(U1)) || !is(T2 == const(U2)) { return __cmp(cast(const)s1, cast(const)s2)); } else { ...  }` OSLT.

src/object.d
CyberShadow: I see, but it would still be prudent to be explicit, which documents that we've tried to ensure that the code below should work for the given kinds of types. At this point it is a nitpick though.

src/object.d
CyberShadow: Sorry, what's the bug? Is it for the feature request to auto-generate `opCmp`?

src/object.d
andralex: @CyberShadow there'd be immutable as well. Doesn't the assert that U1 must be the same as U2 cover it?

src/object.d
andralex: I'm thinking the fact that `S(1) < S(2)` does not compile yet `[S(1) < S(2)]` compiles is definitely a browraiser.

src/object.d
CyberShadow: Sorry, I don't understand how. `Unqual` strips immutable so we're still looking at unnecessary template instantiations for all the various constnesses.

src/object.d
CyberShadow: Not `[S(1) < S(2)]` but `[S(1)] < [S(2)]`. Is that what you meant?

andralex:
Had a bit more of a good time taking `__cmp` out for class/interface objects. Then the generic loop implementation works just fine, no more special casing needed.

src/object.d
```diff
@@ -3251,11 +3252,31 @@ if (__traits(isFloating, F))
         if (!r) r = __cmp(f1.im, f2.im);
         return r;
     }
-    else static if (is(F == float) || is(F == double) || is(F == real))
+    else
     {
         return (f1 > f2) - (f2 > f1);
     }
-    else static assert(false, "Internal error");
```
CyberShadow: Why this change? Being explicit is generally better than hoping that the catch-all (else/default) does the right thing.

src/object.d
```diff
@@ -3241,6 +3241,7 @@ template RTInfo(T)
     enum RTInfo = null;
 }
 
+// Compare floating point numbers for ordering (one instantiation per width).
```
CyberShadow: I'd suggest enforcing "one instantiation per width" using e.g. `static assert(is(Unqual!F == F));`

src/object.d
andralex: Oh. Yah, that's problematic too - you can't order lexicographically unless elements are ordered.

src/object.d
andralex: I'll let folks call `__cmp` with idouble if they have the need. Just made sure it's not called from the array compare.

src/object.d
andralex: The parameters already have `const`.

src/object.d
```diff
@@ -3321,23 +3310,22 @@ int __cmp(T1, T2)(T1[] s1, T2[] s2)
...
+    // Everything else: class, interface, struct, other built-ins
     else
     {
         immutable len = s1.length <= s2.length ? s1.length : s2.length;
 
         foreach (const u; 0 .. len)
         {
-            // structs
-            static if (__traits(compiles, at(s1, u).opCmp(at(s2, u))))
+            static if (__traits(compiles, __cmp(at(s1, u), at(s2, u))))
```
CyberShadow: Why the swap? Just curious.

src/object.d
andralex: @CyberShadow I guess we're talking past each other. What do you want this PR to accomplish?

src/object.d
andralex: I wanted to make sure `__cmp` gets first crack so we don't call `a.opCmp(b)` where a is a class type and its value is null.

src/object.d
CyberShadow: This is a tricky one: the `const` in the parameters actually does not change `F` - only the type of `f1` / `f2`.

src/object.d
CyberShadow: Let's say we have:

```d
int[] a;
const(int)[] b;
immutable(int)[] c;

writeln(a < b);
writeln(b < c);
writeln(a < c);
```

You said that "the qualifiers are passed by the compiler". I understood that the compiler strips qualifiers before instantiating `__cmp`, so there will be only one `__cmp` instantiation? If so, this should be enforced on the `__cmp` side.

If, however, there are three template instantiations, two of those should short-circuit to calling the third to avoid redundant code gen. I also mentioned this in Lucia's PR and this is what I thought this PR did from your initial description :)

src/object.d
andralex: @CyberShadow got a repro? Just tried this:

```d
void fun(F)(const F x)
{
	pragma(msg, F.stringof);
}

void main()
{
	float a;
	fun(a);
	const float b = 1;
	fun(b);
}
```

Only one copy is generated.

src/object.d
CyberShadow: https://issues.dlang.org/show_bug.cgi?id=17244

src/object.d
CyberShadow: Ah, OK - it doesn't apply to floats because the compiler strips head const on call. Only types with indirections.

CyberShadow: I.e. `F` will still be non-const but that's not a problem because it will always be non-const :)

src/object.d
andralex: Yah, that's my understanding as well - three copies will be generated. I can work on folding those together in this PR. Stay tuned.

somzzz:
I opened the PR for the ```dmd``` lowering to ```__cmp```.
https://github.com/dlang/dmd/pull/6597

There are tests failing in DMD because of comparisons of  arrays of ```struct``` types without ```opCmp``` defined. I talked with @andralex and it seems that this is because of an old issue with using ```memcmp``` for comparing arrays of structs, which the current implementation does't preserve.

A simple example/test:
```
struct VariantN(size_t maxDataSize)
{
    ~this() {}
}
alias Variant4 = VariantN!(4);
void test()
{
    Variant4[] a4;
    assert(a4 < a4);
}
void main() {}
```

This could either be fixed here or I could open a new PR after this is merged. What do you think?

JackStouffer:
@somzzz One thing at a time.

andralex:
@JackStouffer the latest commit should take care of things.

src/object.d
```diff
@@ -3241,181 +3241,105 @@ template RTInfo(T)
...
+        return 0;
+    // Regard null references as always being "less than"
+    if (!lhs)
+        return -1;
+    if (!rhs)
+        return 1;
+    return lhs.opCmp(rhs);
+}
+
+int __cmp(T)(const T[] lhs, const T[] rhs) @trusted
```
andralex: @CyberShadow ok, so this overload should generate only one copy per scalar type. There are still a couple of opportunities:

* merge wchar with short implementation
* merge dchar with uint implementation
* merge ifloat with float implementation
* merge idouble with real implementation
* merge ireal with real implementation

I've let these be because they're relatively rare.

CyberShadow: > ok, so this overload should generate only one copy per scalar type

Are you sure? This doesn't even compile:

```d
int __cmp(T)(const T[] lhs, const T[] rhs) @trusted {}

int[] a;
immutable(int)[] b;
__cmp(a, b);
```

andralex: @CyberShadow take that back, I decided to implement the whole shebang anyway, it doesn't take much code.

andralex: worksforme:

```d
void main()
{
	int __cmp(T)(const T[] lhs, const T[] rhs) @trusted { return 0; }

	int[] a;
	immutable(int)[] b;
	__cmp(a, b);
}
```

src/object.d
```diff
@@ -3258,12 +3258,32 @@ if (is(Obj : Object))
...
+    static if (is(U == char))
     {
         import core.internal.string : dstrcmp;
         return dstrcmp(cast(char[]) lhs, cast(char[]) rhs);
     }
+    else static if (!is(U == T))
+    {
+        // Reuse another implementation
+        return __cmp(cast(U[]) lhs, cast(U[]) rhs);
+    }
```
CyberShadow: :+1:

src/object.d
CyberShadow: Nice but constness issue remains, I think that would be good as a test BTW.

andralex:
Since this is approved I'll merge.

PR #1792, opened 2017-03-15
_equals druntime template for structs

somzzz:
This is similar to what happened to ```adCmp```, or ```__cmp```, only that it's a first step, rather than the whole implementation. This is the template function for testing equality on arrays of structs (basetype is struct).

I also worked on a ```dmd``` patch lowering ```EqualExp``` to this code for arrays of structs. Locally, ```Druntime``` and ```Phobos``` tests are passing with the new implementation.

```__equals``` is a bit more interesting than ```__cmp``` because for arrays of basic types the compiler makes an optimization and generates a call too "memcmp" directly rather than call a ```druntime``` function. See here: https://github.com/dlang/dmd/blob/master/src/ddmd/e2ir.d#L2282

I avoided changing that bit for the time being until we can agree on what the best approach is. This is one of the reasons this PR only does struct array comparisons. The other would be that a smaller PR is easier to review.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3241,6 +3241,71 @@ template RTInfo(T)
     enum RTInfo = null;
 }
 
+// lhs == rhs lowers to __equals(lhs, rhs) for arrays of all struct types.
+// old path: Typeinfo_array => TypeInfo_struct
+bool __equals(T1, T2)(T1[] lhs, T2[] rhs)
+{
+    if (lhs.length != rhs.length)
+        return false;
+
+    import core.internal.traits : Unqual;
+    static assert(is(Unqual!T1 == Unqual!T2), "Internal error.");
+
+    static @trusted ref R at(R)(R[] r, size_t i) { return r.ptr[i]; }
+
+    foreach (const u; 0 .. lhs.length)
+    {
+        static if (__traits(compiles, at(lhs, u).opEquals(at(rhs, u))))
+        {
+            if (!at(lhs, u).opEquals(at(rhs, u)))
+                return false;
+        }
+        else
+        {
+            if (at(lhs, u).tupleof != at(rhs, u).tupleof)
+                return false;
+        }
+    }
+
+    return true;
+}
+
+unittest
+{
+    struct A
+    {
+        int a;
+    }
+
+    auto arr1 = [A(0), A(2)];
+    auto arr2 = [A(0), A(1)];
+    auto arr3 = [A(0), A(1)];
+
+    assert(arr1 != arr2);
+    assert(arr2 == arr3);
+}
+
+unittest
+{
+    struct A
+    {
+        int a;
+        int b;
+
+        bool opEquals(const A other)
+        {
+            return this.a == other.b && this.b == other.a;
+        }
+    }
+
+    auto arr1 = [A(1, 0), A(0, 1)];
+    auto arr2 = [A(1, 0), A(0, 1)];
+    auto arr3 = [A(0, 1), A(1, 0)];
+
+    assert(arr1 != arr2);
+    assert(arr2 == arr3);
+}
+
 // Compare class and interface objects for ordering.
 private int __cmp(Obj)(Obj lhs, Obj rhs)
 if (is(Obj : Object))

### CODE REVIEW

File: src/object.d [hunk 1]
```diff
@@ -3241,6 +3241,40 @@ template RTInfo(T)
     enum RTInfo = null;
 }
 
+bool __equals(T1, T2)(T1[] lhs, T2[] rhs)
+{
+    if (lhs.length != rhs.length)
+        return 0;
```
JackStouffer: stylistic nitpick: if you're going to use `return true;` at the end, then make this `return false;`

src/object.d
```diff
@@ -3241,6 +3241,40 @@ template RTInfo(T)
     enum RTInfo = null;
 }
 
+bool __equals(T1, T2)(T1[] lhs, T2[] rhs)
```
andralex: Please add a comment before this function a la `// lhs == rhs lowers to __equals(lhs, rhs) for arrays of all struct types.`

src/object.d
```diff
@@ -3241,6 +3241,40 @@ template RTInfo(T)
...
+{
+    if (lhs.length != rhs.length)
+        return 0;
+
+    import core.internal.traits : Unqual;
+    alias U1 = Unqual!T1;
+    alias U2 = Unqual!T2;
+    static assert(is(U1 == U2), "Internal error.");
```
andralex: No need to define names because we're never using them, `static assert(is(Unqual!T1 == Unqual!T2), "Internal error.");` should suffice

src/object.d
```diff
@@ -3241,6 +3241,40 @@ template RTInfo(T)
...
+
+    static @trusted ref R at(R)(R[] r, size_t i) { return r.ptr[i]; }
+
+    //old path: Typeinfo_array => TypeInfo_struct
+    //static if (is(U1 == struct) || is(U1 == enum))
+    foreach (const u; 0 .. lhs.length)
+    {
+        auto s1 = at(lhs, u);
+        auto s2 = at(rhs, u);
```
andralex: Don't create temporaries here, just invoke `at` inline; the structs may not be copyable

src/object.d
```diff
@@ -3241,6 +3241,40 @@ template RTInfo(T)
...
+
+        static if (__traits(compiles, s1.opEquals(s2)))
+        {
+            auto result = (s1).opEquals(s2);
```
andralex: No need for parens around s1

src/object.d
```diff
@@ -3241,6 +3241,40 @@ template RTInfo(T)
...
+            if (!result)
+                return result;
+        }
+        else
+        {
+            return s1.tupleof == s2.tupleof;
```
andralex:
```d
if (s1.tupleof != s2.tupleof)
    return false;
```

andralex: You should write a unittest that fails with the current code and passes with the correct one.

src/object.d
```diff
@@ -3241,6 +3241,40 @@ template RTInfo(T)
...
+    foreach (const u; 0 .. lhs.length)
+    {
+        auto s1 = at(lhs, u);
+        auto s2 = at(rhs, u);
+
+        static if (__traits(compiles, s1.opEquals(s2)))
+        {
+            auto result = (s1).opEquals(s2);
+            if (!result)
+                return result;
```
andralex: A bit much pomp and circumstance. This should be enough:

```d
if (!s1.opEquals(s2))
    return false;
```

src/object.d
somzzz: I'm not sure I understand what you mean here.

somzzz: Nvm, got it.

somzzz: Nice. I had a case where both ``` assert(a != b);``` and ```assert(a==b);``` passed with that code because I forgot to do this in the compiler:
```
                if (op == TOKnotequal)
                {
                    __equals = new NotExp(loc, __equals);
                }
```

src/object.d
```diff
@@ -3241,6 +3241,50 @@ template RTInfo(T)
...
+    foreach (const u; 0 .. lhs.length)
+    {
+        static if (__traits(compiles, at(lhs, u).opEquals(at(rhs, u))))
+        {
+            if (!at(lhs, u).opEquals(at(rhs, u)))
+                return false;
+        }
+        else
+        {
+            if(!(at(lhs, u).tupleof == at(rhs, u).tupleof))
```
andralex: Interesting. Did `!=` not work?

src/object.d
```diff
@@ -3241,6 +3241,50 @@ template RTInfo(T)
...
+        int a;
+    }
+
+    auto arr1 = [A(0), A(2)];
+    auto arr2 = [A(0), A(1)];
+    auto arr3 = [A(0), A(1)];
+
+    assert(arr1 != arr2);
+    assert(arr2 == arr3);
+}
```
andralex: Need to also add a unittest for a struct that defines opEquals for coverage.

PR #1793, opened 2017-03-16
Fix issue 8411 - add opCast!bool support for Duration.

schveiguy:
Checking whether a duration is zero or not is a handy mechanism. Furthermore, it allows one to check if a duration is zero, and allows declaring a new duration variable at the same time:

```D
if(d = calculateDuration())
{
   // do something with d
}
```

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -1104,6 +1104,21 @@ public:
         }
     }
 
+    /++
+        Returns `true` if this duration is non-zero.
+      +/
+    bool opCast(T : bool)() const nothrow @nogc
+    {
+        return _hnsecs != 0;
+    }
+
+    unittest
+    {
+        auto d = 10.minutes;
+        assert(d);
+        assert(!(d - d));
+        assert(d + d);
+    }
 
     //Temporary hack until bug http://d.puremagic.com/issues/show_bug.cgi?id=5747 is fixed.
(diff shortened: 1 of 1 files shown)


src/core/time.d
```diff
@@ -1104,6 +1104,21 @@ public:
         }
     }
 
+    /++
+        Returns `true` if this duration is non-zero.
```
JackStouffer: `Returns: `

schveiguy: Fixed

schveiguy:
Not sure what the test error is, looks like a fluke segfault in unrelated std.parallelism, so I deprecated it. In any case, thank you all for the approvals. Can I get an automerge from one of you? 😉

JackStouffer:
The green check mark vs the grey one shows who has merge permissions.

schveiguy:
> The green check mark vs the grey one shows who has merge permissions.

Ah, I forget you all don't have druntime cred. I will just push the button, don't think this is controversial.

schveiguy:
Auto-merge toggled on

WalterWaldron:
> looks like a fluke segfault in unrelated std.parallelism,

My fix for that is https://github.com/dlang/phobos/pull/5004 and it's been gathering dust for ~4 months.

PR #1798, opened 2017-04-02
windows: make precise scanning of DATA and TLS optional

rainers:
This makes use of the pointer reference information written by the compiler. If it passes I think it would make sense to add it to the release. It should be opt-in, though, so I'd change the default before merging.


rainers:
Changed back to master and added changelog.

rainers:
Finally figured one/the cause for the crashes: https://github.com/dlang/phobos/pull/5357

rainers:
Now that this has passed the auto tester for a couple of days the spurious failures are likely gone after the fix to phobos. I changed the default to conservative scanning, so this can be considered for merging.

rainers:
Thanks. Added a documentation PR here: https://github.com/dlang/dlang.org/pull/1724

PR #1799, opened 2017-04-04
Fix cannot throw scoped exception

WalterBright:
because that would leak it and cause corruption.

### DIFF
--- a/src/rt/minfo.d
+++ b/src/rt/minfo.d
@@ -167,13 +167,6 @@ struct ModuleGroup
         import core.bitop : bts, btr, bt, BitRange;
         import rt.util.container.hashtab;
 
-        // used to unwind stack for printing deprecation message.
-        static class DeprecatedCycleException : Exception
-        {
-            this() { super(""); }
-        }
-        scope deprecation = new DeprecatedCycleException();
-
         enum OnCycle
         {
             deprecate,
@@ -318,7 +311,8 @@ struct ModuleGroup
         // trivial modules to get at the non-trivial ones.
         //
         // If a cycle is detected, returns the index of the module that completes the cycle.
-        void findDeps(size_t idx, size_t* reachable)
+        // Returns: true for success, false for a deprecated cycle error
+        bool findDeps(size_t idx, size_t* reachable)
         {
             static struct stackFrame
             {
@@ -367,7 +361,7 @@ struct ModuleGroup
                                     if(sortCtorsOld(edges))
                                     {
                                         // unwind to print deprecation message.
-                                        throw deprecation;
+                                        return false;   // deprecated cycle error
                                     }
                                     goto case abort; // fall through
                                 case abort:
@@ -408,6 +402,7 @@ struct ModuleGroup
                 // next dependency
                 ++sp.curDep;
             }
+            return true; // success
         }
 
         // The list of constructors that will be returned by the sorting.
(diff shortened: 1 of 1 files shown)


WalterBright:
Blocking https://github.com/dlang/dmd/pull/6675 which checks for this kind of bug.

schveiguy:
This is not a bug. The exception is caught inside the scope always. The idea is to avoid any gc activity. Is there a better way to do this?

WalterBright:
> The exception is caught inside the scope always.

Yes, you are right. I missed that.

> The idea is to avoid any gc activity. Is there a better way to do this?

Yes. Return an error code (by return value or an `out` parameter). All `DeprecatedCycleException` does is flag an error, it doesn't even have a message.

WalterBright:
Fixed it to use error code for cycles instead of exception.

WalterBright:
@schveiguy you're the right person to review this

schveiguy:
I'll take a look. The convenience of the stack unwinding is what I liked about that mechanism.

src/rt/minfo.d
```diff
@@ -505,12 +513,11 @@ struct ModuleGroup
...
 
         // finally, do the sorting for both shared and tls ctors.
-        try
-        {
-            _ctors = doSort(MIctor | MIdtor);
-            _tlsctors = doSort(MItlsctor | MItlsdtor);
-        }
-        catch(DeprecatedCycleException)
+        bool cycle = false;
+        _ctors = doSort(MIctor | MIdtor, cycle);
```
schveiguy: You need to do this a weird way. The old algorithm sets `_ctors` and `_tlsctors` internally, so the assignment cannot happen. The exception prevented the assignment. You need to do something like this:

```D
auto ctorTemp = doSort(MIctor | MIdtor, cycle);
if(!cycle)
{
    _ctors = ctorTemp;
    ctorTemp = doSort(MItlsctor | MItlsdtor, cycle);
    if(!cycle)
        _tlsctors = ctorTemp;
}
```

WalterBright: I implemented the suggested behavior, but changed the return to a ref parameter which simplifies the logic.

In general, the GC should not be used in druntime if it can be avoided.

Exceptions are expensive (both to throw and in code size) and should be avoided where reasonable.

src/rt/minfo.d
```diff
@@ -457,10 +457,15 @@ struct ModuleGroup
 
             // add this module to the construction order list
             ctors[ctoridx++] = current;
+            return true;
         }
 
-        immutable(ModuleInfo)*[] doSort(size_t relevantFlags)
+        // `cycle` is set to `true` if deprecated cycle error otherwise set `result`.
+        void doSort(size_t relevantFlags, ref bool cycle, ref immutable(ModuleInfo)*[] result)
```
schveiguy: Hm... what about returning `bool`, and then if statement below reads:

```D
if (doSort(...) || doSort(...))
```

In any case, the new code is correct at least.

WalterBright: Returning both a bool and a ref with the same value strikes me as odd and a bit eyebrow raising. I'd need a better reason to use it.

WalterBright: Let's just get this merged and move on.

schveiguy: Not sure what you mean, you just would return a bool. But not important right now.

schveiguy:
Auto-merge toggled on

src/rt/minfo.d
ibuclaw: Indeed, the ref parameter is no longer required if doSort just returns a bool.

PR #1800, opened 2017-04-04
Add eventfd bindings

nemanja-boric-sociomantic:
This adds bindings for linux-specific eventfd_* calls.

### DIFF
--- a/src/core/sys/linux/sys/eventfd.d
+++ b/src/core/sys/linux/sys/eventfd.d
@@ -1,0 +1,85 @@
+/**
+ * D header file for GNU/Linux.
+ *
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   Nemanja Boric
+ */
+module core.sys.linux.sys.eventfd;
+
+version (linux):
+extern (C):
+@nogc:
+@system:
+nothrow:
+
+import core.stdc.stdint: uint64_t;
+
+/// Type for the event counter
+alias uint64_t eventfd_t;
+
+/* Return file descriptor for generic event channel.  Set initial
+   value to count.  */
+int eventfd (uint count, int flags);
+
+/* Read event counter and possibly wait for events.  */
+int eventfd_read (int fd, eventfd_t* value);
+
+/* Increment event counter.  */
+int eventfd_write (int fd, eventfd_t value);
+
+version (X86)
+{
+    enum EFD_SEMAPHORE = 1;
+    enum EFD_CLOEXEC = 0x80000; // octal!2000000
+    enum EFD_NONBLOCK = 0x800; // octal!4000
+}
+else version (X86_64)
+{
+    enum EFD_SEMAPHORE = 1;
+    enum EFD_CLOEXEC = 0x80000; // octal!2000000
+    enum EFD_NONBLOCK = 0x800; // octal!4000
+}
+else version (MIPS32)
+{
+    enum EFD_SEMAPHORE = 1;
+    enum EFD_CLOEXEC = 0x80000; // octal!2000000
+    enum EFD_NONBLOCK = 0x80; // octal!200
+}
+else version (MIPS64)
+{
+    enum EFD_SEMAPHORE = 1;
+    enum EFD_CLOEXEC = 0x80000; // octal!2000000
+    enum EFD_NONBLOCK = 0x80; // octal!200
+}
+else version (PPC)
+{
+    enum EFD_SEMAPHORE = 1;
+    enum EFD_CLOEXEC = 0x80000; // octal!2000000
+    enum EFD_NONBLOCK = 0x800; // octal!4000
+}
+else version (PPC64)
+{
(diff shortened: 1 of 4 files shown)


src/core/sys/linux/sys/eventfd.d
```diff
@@ -0,0 +1,79 @@
...
+extern (C):
+@system:
+nothrow:
+
+/* Return file descriptor for generic event channel.  Set initial
+   value to count.  */
+int eventfd (uint count, int flags);
+
+/* Read event counter and possibly wait for events.  */
+int eventfd_read (int fd, uint* value);
```
ZombineDev: Where did you get those signatures from?
According to https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/sys/eventfd.h;hb=HEAD they should be:
```C
/* Type for event counter.  */
typedef uint64_t eventfd_t;

/* Return file descriptor for generic event channel.  Set initial
   value to COUNT.  */
extern int eventfd (unsigned int __count, int __flags) __THROW;

/* Read event counter and possibly wait for events.  */
extern int eventfd_read (int __fd, eventfd_t *__value);

/* Increment event counter.  */
extern int eventfd_write (int __fd, eventfd_t __value);
```
I suggest you add and use the `eventfd_t` typedef accordingly.

nemanja-boric-sociomantic: This is what I was wondering - should I add this typedef. I'll update.

nemanja-boric-sociomantic:
Updated with `nogc` and `eventfd_t` being alias to `uint64_t`. @ZombineDev is alias enough, or should I go with a full-blown typedef?

ZombineDev:
@nemanja-boric-sociomantic `alias` is more than enough - D2 doesn't have `typedef` as a language feature and `Typedef(T)` is part of Phobos, so it can't be used in druntime.

nemanja-boric-sociomantic:
Yeah, that's what I thought, but simple grep found some `Typedef`s in some runtime modules:

```
nemanjaboric@labs-129:/home/nemanjaboric/work/druntime  git:(eventfd*) $ ag -l TypeDef .               
src/core/sys/windows/ole.d
src/core/sys/windows/mmsystem.d
src/core/sys/windows/rpcdcep.d
src/core/sys/windows/basetsd.d
src/core/sys/windows/vfw.d
src/core/sys/windows/sqltypes.d
src/core/sys/windows/ras.d
src/core/sys/windows/rpcndr.d
src/core/sys/windows/winuser.d
src/core/sys/windows/wingdi.d
```

ZombineDev:
AFAICS, this was something that @CyberShadow [did](https://github.com/dlang/druntime/blob/v2.074.0-b2/src/core/sys/windows/basetsd.d#L26) and is used only for the Win32 bindings. My understanding of this is that the Win32 bindings were originally written in D1 and were later ported to D2 while keeping D1 backwards compatibility and now this is no longer needed. Also none of the other system bindings (stdc/, sys/) use this technique. @CyberShadow can you shed some light on this?

CyberShadow:
That TypeDef definition was from back when the win32 package was still only available separately, on DSource. At some point `typedef` was a proper language feature in D2, however it became deprecated (in favor of the Phobos definition) and later removed, thus that change.

Although strong typing for the various integral (or `void*`-aliased) Windows types would be nice, it is far too late to change it now, and would make porting code from C/C++ more onerous. So, the use and definition of `TypeDef` in the `core.sys.windows` package can now be removed.

ZombineDev:
Thanks for the explanation @CyberShadow. Are you OK with merging this PR?

CyberShadow:
What's the reason for the large version switch? Looking at the Linux source code, these don't seem to be architecture-dependent constants.

nemanja-boric-sociomantic:
They are architecture-dependent constants, see https://github.molgen.mpg.de/git-mirror/glibc/blob/master/sysdeps/unix/sysv/linux/sparc/bits/eventfd.h https://github.molgen.mpg.de/git-mirror/glibc/blob/master/sysdeps/unix/sysv/linux/alpha/bits/eventfd.h https://github.molgen.mpg.de/git-mirror/glibc/blob/master/sysdeps/unix/sysv/linux/hppa/bits/eventfd.h https://github.molgen.mpg.de/git-mirror/glibc/blob/master/sysdeps/unix/sysv/linux/mips/bits/eventfd.h

and the common one https://github.molgen.mpg.de/git-mirror/glibc/blob/master/sysdeps/unix/sysv/linux/bits/eventfd.h

I chose the same way `inotify` bindings are done, where common one is still enumerated by architecture .

CyberShadow:
Huh, so they are in glibc. Not in the Linux kernel though, as far as I could tell.

nemanja-boric-sociomantic:
Yeah, I'll give it another look to be 100% sure.

CyberShadow:
Well there's no harm in doing it the glibc way, so LGTM.

nemanja-boric-sociomantic:
Thanks!

PR #1801, opened 2017-04-05
Clean up awkward inner function for cycle sorting.

schveiguy:
Followup to #1799 

ping @WalterBright this is what I meant.

### DIFF
--- a/src/rt/minfo.d
+++ b/src/rt/minfo.d
@@ -460,12 +460,9 @@ struct ModuleGroup
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


src/rt/minfo.d
```diff
@@ -495,33 +492,29 @@ struct ModuleGroup
...
-                    }
+                        return false;
                 }
             }
 
             if (ctoridx == 0)
             {
                 // no ctors in the list.
                 .free(ctors);
-                result = null;
```
schveiguy: Note, this wasn't needed, as the ctor list is `null` by default.

PR #1803, opened 2017-04-06
[REG] Simple multithreaded program + "-profile=gc" = crash

mihails-strasuns-sociomantic:
Fixes issue 15947 (https://issues.dlang.org/show_bug.cgi?id=15947)
Fixes issue 14511 (https://issues.dlang.org/show_bug.cgi?id=14511)

Root problem was trying to increment non-existent AA field upon merging profile
stats. The code was also simplified to remove unnecessary condition.

### DIFF
--- a/src/rt/profilegc.d
+++ b/src/rt/profilegc.d
@@ -89,18 +89,14 @@ static ~this()
     {
         synchronized
         {
-            if (globalNewCounts.length)
+            foreach (name, entry; newCounts)
             {
-                // Merge
-                foreach (name, entry; newCounts)
-                {
-                    globalNewCounts[name].count += entry.count;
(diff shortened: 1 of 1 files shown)


mihails-strasuns-sociomantic:
FYI @WalterBright @MartinNowak

mihails-strasuns-sociomantic:
Thanks!

@MartinNowak is it possible to do merge of stable into master without waiting for next point release? I am working on `-profile=gc` enhancements that require this fix.

japplegame:
I made a similar PR many months ago - https://github.com/dlang/druntime/pull/1664
But no one bothered to accept or decline it.
This makes me angry.

WalterBright:
You're right and you're justified in being annoyed. I apologize.

PR #1806, opened 2017-04-10
New implementation of -profile=gc

mihails-strasuns-sociomantic:
Requires merge of `stable` into `master` to remove  a975720 commit.

This is an attempt to fix https://issues.dlang.org/show_bug.cgi?id=17294 - it is mostly finished but there are few improvements to make which I am not 100% sure about:

- to be 100% reliable, trace module needs to reuse existing GC lock instead of introducing own one, because direct calls to `GC.malloc` can still interfere with stats
- it would be really good to trace those `malloc` calls in log too, but that requires some enhancements on compiler side too, to let runtime know that `-profile=gc` is enabled. I will probably delay that to separate PR.

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -187,3 +187,4 @@ template hasElaborateCopyConstructor(T...)
     else
         enum bool hasElaborateCopyConstructor = false;
 }
+
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -165,9 +165,12 @@ struct GC
     static struct Stats
     {
         /// number of used bytes on the GC heap (might only get updated after a collection)
-        size_t usedSize;
+        ulong usedSize;
         /// number of free bytes on the GC heap (might only get updated after a collection)
-        size_t freeSize;
+        ulong freeSize;
+        /// number of bytes freed during collections through program lifetime so
+        /// far (will count same memory multiple times if re-used)
+        ulong totalCollected;
     }
 
     /**
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -49,6 +49,7 @@ import cstdlib = core.stdc.stdlib : calloc, free, malloc, realloc;
 import core.stdc.string : memcpy, memset, memmove;
 import core.bitop;
 import core.thread;
+import core.sync.mutex;
 static import core.memory;
 
 version (GNU) import gcc.builtins;
@@ -252,6 +253,10 @@ debug (LOGGING)
 
 class ConservativeGC : GC
 {
+    bool profiling_enabled;
+    shared Mutex profiler_lock;
+    shared void[__traits(classInstanceSize, Mutex)] profiler_lock_bytes;
+
     // For passing to debug code (not thread safe)
     __gshared size_t line;
     __gshared char*  file;
@@ -316,6 +321,10 @@ class ConservativeGC : GC
             gcx.reserve(config.initReserve << 20);
         if (config.disable)
             gcx.disabled++;
+
+        this.profiler_lock_bytes[0..$] = typeid(Mutex).initializer[];
+        this.profiler_lock = cast(shared(Mutex)) this.profiler_lock_bytes.ptr;
+        this.profiler_lock.__ctor();
     }
 
 
@@ -359,6 +368,13 @@ class ConservativeGC : GC
 
     auto runLocked(alias func, Args...)(auto ref Args args)
     {
+        if (this.profiling_enabled)
+        {
+            this.profiler_lock.lock_nothrow();
+            scope(exit)
+                this.profiler_lock.unlock_nothrow();
+        }
+
         debug(PROFILE_API) immutable tm = (config.profile > 1 ? currTime.ticks : 0);
         lockNR();
         scope (failure) gcLock.unlock();
@@ -1229,6 +1245,16 @@ class ConservativeGC : GC
         stats.usedSize -= freeListSize;
         stats.freeSize += freeListSize;
         stats.totalCollected = .totalCollectedPages * PAGESIZE;
+    }
+
+    void enableProfiling() nothrow @nogc
+    {
+        this.profiling_enabled = true;
+    }
+
+    shared(Mutex) profilerLock() nothrow @nogc
+    {
+        return this.profiler_lock;
     }
 }
 
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -115,6 +115,7 @@ __gshared long numFrees;
 __gshared long numReallocs;
 __gshared long numExtends;
 __gshared long numOthers;
+__gshared long totalCollectedPages;
 __gshared long mallocTime; // using ticks instead of MonoTime for better performance
 __gshared long freeTime;
 __gshared long reallocTime;
@@ -1227,6 +1228,7 @@ class ConservativeGC : GC
 
         stats.usedSize -= freeListSize;
         stats.freeSize += freeListSize;
+        stats.totalCollected = .totalCollectedPages * PAGESIZE;
     }
 }
 
@@ -2434,7 +2436,9 @@ struct Gcx
 
         updateCollectThresholds();
 
-        return freedLargePages + freedSmallPages;
+        auto total = freedLargePages + freedSmallPages;
+        totalCollectedPages += total;
+        return total;
     }
 
     /**
--- a/src/rt/profilegc.d
+++ b/src/rt/profilegc.d
@@ -89,18 +89,14 @@ static ~this()
     {
         synchronized
         {
-            if (globalNewCounts.length)
+            foreach (name, entry; newCounts)
             {
-                // Merge
-                foreach (name, entry; newCounts)
-                {
-                    globalNewCounts[name].count += entry.count;
-                    globalNewCounts[name].size += entry.size;
-                }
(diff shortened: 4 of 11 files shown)


mihails-strasuns-sociomantic:
FYI @leandro-lucarella-sociomantic @mathias-lang-sociomantic

src/rt/tracegc.d
```diff
@@ -91,65 +91,47 @@ extern (C) void[] _d_newarraymiTX(const TypeInfo ti, size_t[] dims);
 extern (C) void* _d_newitemT(in TypeInfo _ti);
 extern (C) void* _d_newitemiT(in TypeInfo _ti);
 
-extern (C) Object _d_newclassTrace(string file, int line, string funcname, const ClassInfo ci)
+private string generatePrintf ( )
```
nemanja-boric-sociomantic: nitpick: suggest `generatePrintfTrace`

src/core/internal/traits.d
```diff
@@ -187,3 +187,4 @@ template hasElaborateCopyConstructor(T...)
     else
         enum bool hasElaborateCopyConstructor = false;
 }
+
```
nemanja-boric-sociomantic: extra newline

src/rt/tracegc.d
nemanja-boric-sociomantic: Btw, it seems that this change is completely removed in one of the next commits (it's using old style of tracing printf).

src/rt/tracegc.d
```diff
@@ -88,243 +84,156 @@ extern (C) void[] _d_newarrayT(const TypeInfo ti, size_t length);
...
+// will call wrapped function with all required arguments.
+//
+// Placed here as a separate string constant to simplify maintenance as it is
+// much more likely to be modified than rest of generation code.
+enum accumulator = q{
+    import rt.profilegc : accumulate;
+    import core.memory : GC;
+
+    static if (is(typeof(ci)))
+        string name = ci.name;
```
nemanja-boric-sociomantic: suggest: `allocation_type` or similar

src/rt/tracegc.d
```diff
@@ -88,243 +84,156 @@ extern (C) void[] _d_newarrayT(const TypeInfo ti, size_t length);
...
 }
 
-extern (C) byte[] _d_arrayappendcTXTrace(string file, int line, string funcname, const TypeInfo ti, ref byte[] px, size_t n)
+private string generateWrapper(alias Declaration)()
 {
-    mixin(generatePrintf());
-    accumulate(file, line, funcname, ti.toString(), ti.next.tsize * n);
-    return _d_arrayappendcTX(ti, px, n);
-}
+    static size_t findParamIndex(string s)
```
nemanja-boric-sociomantic: Would be very nice if these helper methods would have ddoc, or at least `generateWrapper`.

mihails-strasuns-sociomantic:
Note: failures are most likely related to the mentioned issue with lack of single lock with GC implementation - there seems to be a data race on multi-threaded allocation test case.

src/core/memory.d
```diff
@@ -165,9 +165,12 @@ struct GC
     static struct Stats
     {
         /// number of used bytes on the GC heap (might only get updated after a collection)
-        size_t usedSize;
+        ulong usedSize;
         /// number of free bytes on the GC heap (might only get updated after a collection)
-        size_t freeSize;
+        ulong freeSize;
+        /// number of bytes freed during collections through program lifetime so
+        /// far (will count same memory multiple times if re-used)
```
leandro-lucarella-sociomantic: What's the point of keeping track of this? It should be mentioned at least in the commit message, but probably also in the changelog and the code docs.

src/rt/tracegc.d
```diff
@@ -91,65 +91,47 @@ extern (C) void[] _d_newarraymiTX(const TypeInfo ti, size_t[] dims);
...
-            funcname.length, funcname.ptr
-            );
+        return q{
+            printf("%s file = '%.*s' line = %d function = '%.*s'\n",
+                __FUNCTION__.ptr,
+                file.length, file.ptr,
+                line,
+                funcname.length, funcname.ptr
+                );
+        };
```
leandro-lucarella-sociomantic: Why a `mixin` and not just a function with a `version` statement inside? The optimizer should remove the empty function call I guess...

src/rt/tracegc.d
```diff
@@ -88,243 +84,156 @@ extern (C) void[] _d_newarrayT(const TypeInfo ti, size_t length);
...
-                __FUNCTION__.ptr,
-                file.length, file.ptr,
-                line,
-                funcname.length, funcname.ptr
-                );
-        };
-    }
-    else
-        return "";
-}
```
leandro-lucarella-sociomantic: I agree with @nemanja-boric-sociomantic, is weird that you almost remove everything you did in the previous commit here, but also for me is very distracting that you are reworking all the tracing again mixed with what the commit is supposed to do "Rework how -profile=gc output is calculated". Is hard to review those changes like this.

test/profile/myprofilegc.log.exp
```diff
@@ -1,8 +1,8 @@
 bytes allocated, allocations, type, function, file:line
            2560	             10	core.thread.Thread D main src/profilegc.d:77
             640	             10	int[] profilegc.main.bar src/profilegc.d:73
-            288	              1	immutable(char)[][int] D main src/profilegc.d:34
             240	              4	core.thread.Thread[] D main src/profilegc.d:77
+            288	              1	immutable(char)[][int] D main src/profilegc.d:34
```
leandro-lucarella-sociomantic: I don't understand, so, bytes allocated is how much is supposed to be allocated each time? How is something like `auto x = new int[n];` handled, since the bytes allocated on each call will vary depending on `n`?

src/rt/profilegc.d
```diff
@@ -21,7 +21,7 @@ import core.stdc.string;
 
 import core.exception : onOutOfMemoryError;
 
-struct Entry { size_t count, size; }
+struct Entry { ulong count, size; }
```
leandro-lucarella-sociomantic: Why is changing `size_t` to `ulong` is necessary or better? Needs clarification in the commit message at least, but maybe even here too, so nobody is tempted to change it back.

src/gc/impl/conservative/gc.d
```diff
@@ -252,6 +253,10 @@ debug (LOGGING)
 
 class ConservativeGC : GC
 {
+    bool profiling_enabled;
+    shared Mutex profiler_lock;
```
leandro-lucarella-sociomantic: Why creating this new mutex instead of using the pre-existing GC global lock for everything?

mihails-strasuns-sociomantic: Currenty main GC implementation uses non re-entrant spinlock. Anyway, all lock/sync hacking is more of proof of concept to see if I can make it work that way (have failed so far).

src/rt/tracegc.d
mihails-strasuns-sociomantic: `__FUNCTION__`

src/rt/tracegc.d
mihails-strasuns-sociomantic: Funnily, I did exactly to simplify reviewing, minimizing amount of lines removed in the commit that does interesting stuff (as tracing copy-paste bits have been removed in previous commit) :)

test/profile/myprofilegc.log.exp
mihails-strasuns-sociomantic: No, `bytes allocated` should be total amount throught the life of the program.

src/rt/profilegc.d
mihails-strasuns-sociomantic: Was trying to fix differences in stats between 32-bit and 64-bit builds, stupid idea that will need to be reverted (I have later switched to using two different log files to compare against in test runner).

mihails-strasuns-sociomantic:
> On a more general topic, I don't understand why is all the calculation done in the runtime functions in the first place instead of directly in the GC

It is simply impossible with current `-profile=gc` design, as it requires either changing signatures of involved `extern(C)` functions to also return gc stats diff or somehow passing file/line/function info to those GC functions (== also changing signatures). Both are nasty breaking changes while approach in this PR is a simple enhancement.

Main problem with existing implementation is that I couldn't manage to get determenistic stats as soon as threads are involved, even with dump global lock for every operation (it seems like amount of memory allocated by `new Thread` itself may vary). It is not important for our Sociomantic use case so I haven't invested too much into trying to nail the issue but will need to be addressed if this approach to be considered for upstream.

test/profile/myprofilegc.log.exp
leandro-lucarella-sociomantic: I don't understand, the commit says sort by total size first, but this is putting first 240 and 280 afterwards...

leandro-lucarella-sociomantic:
Back to my question about why not doing all this tracking inside the GC instead of the allocating functions in the runtime... Is this because it will be a complete rewrite of this feature, or is there any other fundamental problem with that approach that I'm not seeing?

mihails-strasuns-sociomantic:
> Is this because it will be a complete rewrite of this feature, or is there any other fundamental problem with that approach that I'm not seeing?

Only former - I actually do agree that would be fundamentally more correct approach. But is much more changes, including plenty of breaking changes, thus I was reluctant to pursue that path :(

mihails-strasuns-sociomantic:
Closing until we come up with solution that can be used upstream

PR #1809, opened 2017-04-13
Update core.sys.freebsd.sys.event

Calrama:


src/core/sys/freebsd/sys/event.d
```diff
@@ -28,10 +30,11 @@ enum
     EVFILT_PROC     =  -5, /* attached to struct proc */
     EVFILT_SIGNAL   =  -6, /* attached to struct proc */
     EVFILT_TIMER    =  -7, /* timers */
-    // EVFILT_NETDEV   =  -8, /* no longer supported */
+    EVFILT_PROCDESC =  -8, /* attached to process descriptors */
```
nemanja-boric-sociomantic: I would keep both and add deprecation note to this one (can one deprecate enum field?) so it get's removed in the next release and not break code immediately.

Calrama: If I changed existing `EVFILT_NETDEV` to `EVFILT_PROVDESC` I would understand that, but `EVFILT_NETDEV` was already commented out and not usable by any code?

nemanja-boric-sociomantic: Ah, I'm sorry, I've missed that it was already commented. Thank you for pointing that out!

src/core/sys/freebsd/sys/event.d
```diff
@@ -15,6 +15,8 @@ module core.sys.freebsd.sys.event;
 
 version (FreeBSD):
 extern (C):
+nothrow:
+@nogc:
```
MartinNowak: Be careful with those in C headers, sometime callback types are declared which would become more restrictive by that and a deprecated overload taking the old callbacks would be in order.

Calrama: With regards to `@nogc` I see your point, but `nothrow`? Is throwing from a D function called by C even defined behaviour?

PR #1810, opened 2017-04-13
Update core.sys.darwin.sys.event

Calrama:


src/core/sys/darwin/sys/event.d
```diff
@@ -86,29 +122,70 @@ enum
...
     NOTE_FFNOP      = 0x00000000, /* ignore input fflags */
-    NOTE_FFAND      = 0x40000000, /* AND fflags */
-    NOTE_FFOR       = 0x80000000, /* OR fflags */
+    NOTE_FFAND      = 0x40000000, /* and fflags */
+    NOTE_FFOR       = 0x80000000, /* or fflags */
     NOTE_FFCOPY     = 0xc0000000, /* copy fflags */
     NOTE_FFCTRLMASK = 0xc0000000, /* masks for operations */
     NOTE_FFLAGSMASK = 0x00ffffff,
 
-    NOTE_TRIGGER    = 0x01000000, /* Cause the event to be
```
DmitryOlshansky: Should probably be deprecated at first?

Burgos: It's just moved up, to line 162, if I can see correctly at this time of a day.

DmitryOlshansky: Oh, indeed.

Calrama: @DmitryOlshansky I moved that up with separate documentation to more closely match the original C headers present on macOS Sierra.

Calrama: @DmitryOlshansky Do you want me to move it back down (since your review is marked as `changes requested`)?

Burgos: Ping @DmitryOlshansky

src/core/sys/darwin/sys/event.d
```diff
@@ -43,7 +43,6 @@ enum : short
     EVFILT_FS       =  -9, /* filesystem events */
     EVFILT_USER     = -10, /* User events */
     EVFILT_VM       = -12, /* virtual memory events */
-    EVFILT_SYSCOUNT =  11
```
Burgos: Perhaps we should deprecate this constant first and remove in the next release cycle?

src/core/sys/darwin/sys/event.d
DmitryOlshansky: No, it's unchanged let's just go ahead.

src/core/sys/darwin/sys/event.d
Calrama: The reasons why I didn't are:
- It's essentially a [version number](https://github.com/opensource-apple/xnu/blob/8cf668b09f7c419fadf6081d1f1bdd6c5033e708/bsd/kern/kern_event.c#L346):
  Any value we define is wrong on some (supported) system, so unless we grab the correct value from the OS at compile time (which would require the D frontend to be able to parse C header files), we should simply not expose it.  It also follows that any D code depending on `EVFILT_SYSCOUNT`'s value being correct for the current OS is broken and should not compile.
- It's used internally only (e.g. [here](https://github.com/opensource-apple/xnu/blob/8cf668b09f7c419fadf6081d1f1bdd6c5033e708/bsd/kern/kern_event.c#L2491)), AFAICT you can't use it in API calls.

So essentially, if you were using `EVFILT_SYSCOUNT` from D, your code was already broken.

nemanja-boric-sociomantic: OK, if you can't use it in API calls, I think this is fine to remove without deprecation cycle.

ZombineDev:
@Calrama please rebase.

Calrama:
@ZombineDev Done

ZombineDev:
Thanks, the CIs are now looking positive. One thing that I'm not certain about is about the documentation added in this PR. The problem is that while interfaces are copyright free, comments / documentation is protected by copyright and the license to use it may now be compatible with our Boost license. AFAIK, this the reason most of the header files are devoid of any documentation. Probably @WalterBright can clarify. Apart from that everything looks good to me.

Calrama:
> One thing that I'm not certain about is about the documentation added in this PR. The problem is that while interfaces are copyright free, comments / documentation is protected by copyright and the license to use it may now be compatible with our Boost license.

Hm, I took the comments originally from the macOS Sierra headers, which are under [Apple Public Source License](https://opensource.apple.com/apsl). Regardless, the easiest way is indeed to just not include them, so I've amended the commit.

ZombineDev:
Let's see if can resurrect DAutoTest

jmdavis:
Aside from the potential copyright issues, I think that in general, we don't take the comments from the C headers, because we want folks looking at the OS documentation for what this stuff means, not us. We're not in the business of maintaining OS APIs, and we're already at a pretty high risk of having maintenance issues with the bindings themselves precisely because they simply link to someone else's code that we don't maintain. Fortunately, OS APIs don't change much, so we're generally okay,  but I think that it's just safer to not include any comments from the header files, regardless of the license. Anyone who wants to know what these mean really should be looking at the documentation for the actual API and not our bindings. So, I think that the comments should be removed.

Calrama:
> So, I think that the comments should be removed.

The amended https://github.com/dlang/druntime/pull/1810/commits/696d83f74b64d954f4aba4de341d0768d6c7e7c1 from earlier also removes any such previously existing commentary from `core.sys.darwin.sys.event`.

The next question should be, if another dedicated PR is due in order to remove all such C inherited comments from druntime in other modules.

PR #1814, opened 2017-04-18
__equals for floats & objects & integral types

somzzz:
Full implementation for __equals.

Handles integral types, objects, fp types, void arrays, arrays of arrays.


UplinkCoder:
@somzzz I like it.
You do avoid unnecessary template-instanciations.
That is good.

PR #1817, opened 2017-04-25
[RISC-V] Add some definitions necessary for LDC to compile programs to RISC-V targets.

e-y-e:
This PR simply eliminates some errors due to undefined constants and types that occur when using LDC to compile programs using RISC-V as a target.


schveiguy:
Why does this version not include the `c_long_double` definitions that were in the LDC PR?

e-y-e:
I was hesitant to do so as none of the previous architectures specific to LDC did that in upstream. But if you feel it is necessary I am not at a computer so if you could modify the PR that would be great.

schveiguy:
It doesn't hurt AFAIK, but it would actually make sense just to copy all the definitions in a separate PR. It doesn't make sense to just have RISC-V version and not all the others. In general, the closer and more consistent all the compilers' druntimes are, the better.

So long story short, we can ignore that change, and do that in a separate PR.

e-y-e:
Ok, that sounds like a logical approach.

e-y-e:
I assume the auto-tester is failing on master, since this code should affect nothing. If so, could this be pulled so that the matching PR in LDC druntime can progress?

schveiguy:
The failure on the auto tester is a sporadic one (looks like a make subprocess was killed for some reason). I marked it to rerun.

e-y-e:
Many thanks.

PR #1818, opened 2017-04-27
[STABLE] Add missing binding in SRCS

Geod24:
Some files were not part of `mak/SRCS`, and thus were not compiled it.

Since they are C bindings, it was unnoticed for a long time,
as most operations won't require the rare symbols this module defined.
However for things such as static array declarations,
the `__init` of the struct might be required, which triggers a linker error.

One simple way to trigger the issue was to use the function `timersub` defined in
`core.sys.linux.sys.time`:

```
src/ocean/core/UnitTestRunner.d:671: undefined reference to 'core.sys.linux.sys.time.timersub(const(core.sys.posix.sys.time.timeval*), const(core.sys.posix.sys.time.timeval*), core.sys.posix.sys.time.timeval*)'
```

The patch is a superset of: https://github.com/sociomantic-tsunami/ocean/blob/fef725ca9643af4ffcccb44bcc8a2ff7598c4e24/docker/dmd-transitional/patches/druntime/0010-Add-missing-binding-in-SRCS.patch

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -1,5 +1,6 @@
 COPY=\
 	$(IMPDIR)\object.d \
+	\
 	$(IMPDIR)\core\atomic.d \
 	$(IMPDIR)\core\attribute.d \
 	$(IMPDIR)\core\bitop.d \
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -23,10 +23,15 @@ SRCS=\
 	src\core\internal\string.d \
 	src\core\internal\traits.d \
 	\
+	src\core\stdc\complex.d \
 	src\core\stdc\config.d \
 	src\core\stdc\ctype.d \
 	src\core\stdc\errno.d \
 	src\core\stdc\fenv.d \
+	src\core\stdc\float_.d \
+	src\core\stdc\inttypes.d \
+	src\core\stdc\limits.d \
+	src\core\stdc\locale.d \
 	src\core\stdc\math.d \
 	src\core\stdc\signal.d \
 	src\core\stdc\stdarg.d \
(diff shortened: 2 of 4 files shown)


schveiguy:
LGTM

mathias-lang-sociomantic:
Win32 error: `Error: command line too long`

Great. I guess it's time to use this little DMD feature.

Geod24:
Didn't seem to help. @braddr , am I correct that Win_32_64 is using DigitalMars' Make ? Could you offer some advice on this issue (Command too long) ?

rainers:
That's likely a limitation of the Digital Mars make: the win64 build command line is slightly above 10000 characters, while the win32 is just a couple of characters below. It used to be 8192 in the command line interpreter, but it's 32768 for the CreateProcess function (https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425(v=vs.85).aspx).

The usual workaround is to place the command line arguments into a temporary file and use `@tmpfile` on the command line.

Geod24:
> The usual workaround is to place the command line arguments into a temporary file and use @tmpfile on the command line.

According to [the manual](http://www.digitalmars.com/ctg/make.html) that's what using `*` and `~` should do.

rainers:
> According to the manual that's what using * and ~ should do.

Ah, I wasn't aware of this functionlity. It actually seems to work here until a higher length (probably 32768 as the limit of environment strings).

mathias-lang-sociomantic:
CC @WalterBright : I tried using `*` and `~` inside the Win32 Makefile so DigitialMars' MAKE would pass the command line args via `@_CMDLINE`, but it doesn't seem to have any effect.

schveiguy:
If the issue here is windows make, can't we do something different for it? Clearly windows doesn't need linux headers.

CyberShadow:
> According to [the manual](http://www.digitalmars.com/ctg/make.html) that's what using `*` and `~` should do.

No, that's something else. Those cause DigitalMars Make to pass the command line in an *environment variable*, not a file. This is functionality specific to the DigitalMars toolchain, and although it is apparently supported by OPTLINK, it is not supported by dmd itself.

I suggest trying manually write the list to a temporary file in the Makefile target rule.

rainers:
> it is not supported by dmd itself.

Unfortunately, it does support it: https://github.com/dlang/dmd/blob/master/src/ddmd/root/response.d#L65

Nevertheless, I'd also suggest to use an explicit temporary file, too.

mathias-lang-sociomantic:
We had a discussion with @WalterBright at DConf about it. He offered to give me access to DM Make to track the problem down. I'd rather fix the issue than add another workaround.

rainers:
The trouble is just that you have applied the "~" fix to win32.mak, but not to win64.mak.

Geod24:
> The trouble is just that you have applied the "~" fix to win32.mak, but not to win64.mak.

Oh damn, it was that simple, thanks !
Fixed, but Jenkins is failing with a seemingly unrelated error:
```
[0;33m[INFO] Running /var/lib/jenkins/dlang_projects@3/dlang/dub/test/interactive-remove.sh...[0m
Fetching dub 0.9.20...
Fetching dub 0.9.21...
Cannot remove package 'dub', there are multiple possibilities at location
'user'.
Available versions:
  0.9.20
  ~master
  1.1.0
Please specify a individual version using --version=... or use the wildcard --version=* to remove all versions.
[0;31mScript failure.[0m
readlink -f ${BASH_SOURCE[0]}
```

wilzbach:
@Geod24: Jenkins is Linux only (ATM) and unfortunately broken since a couple of days. It's not related to this PR.

Geod24:
So it won't prevent a merge, despite being `Required` ?

wilzbach:
It does, but @CyberShadow can disable it

MartinNowak:
Was a temporary failure with the interactive dub remove test, not exactly sure why it failed. I just reran it, need to change Jenkins auth to allow more people to rerun tests (but am still traveling atm.).

MartinNowak:
Was a temporary failure with the interactive dub remove test, not exactly sure why it failed. I just reran it, need to change Jenkins auth to allow more people to rerun tests (but am still traveling atm.).

PR #1819, opened 2017-05-03
Add some Bionic declarations needed in latest phobos, plus two fixes

joakim-noah:
All prior Android declarations were taken from Android API 9, which is [soon to be deprecated](https://github.com/android-ndk/ndk/wiki/Changelog-r15-beta1), so I note the few that were needed by Phobos and taken from a newer API with a comment.

### DIFF
--- a/src/core/sys/posix/signal.d
+++ b/src/core/sys/posix/signal.d
@@ -159,15 +159,27 @@ else version( CRuntime_Glibc )
     }
 }
 else version (FreeBSD) {
+    // Note: it appears that FreeBSD (prior to 7) and OSX do not support realtime signals
     // https://github.com/freebsd/freebsd/blob/e79c62ff68fc74d88cb6f479859f6fae9baa5101/sys/sys/signal.h#L117
     enum SIGRTMIN = 65;
     enum SIGRTMAX = 126;
 }
-// Note: it appears that FreeBSD (prior to 7) and OSX do not support realtime signals
 else version(NetBSD)
 {
     enum SIGRTMIN = 33;
     enum SIGRTMAX = 63;
+}
+else version (CRuntime_Bionic)
+{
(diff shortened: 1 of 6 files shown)


src/core/sys/posix/sys/stat.d
```diff
@@ -1110,6 +1110,10 @@ else version( CRuntime_Bionic )
     extern (D) bool S_ISREG( uint mode )  { return S_ISTYPE( mode, S_IFREG );  }
     extern (D) bool S_ISLNK( uint mode )  { return S_ISTYPE( mode, S_IFLNK );  }
     extern (D) bool S_ISSOCK( uint mode ) { return S_ISTYPE( mode, S_IFSOCK ); }
+
+    // Added since Lollipop
+    int utimensat(int dirfd, const char *pathname,
+        ref const(timespec)[2] times, int flags);
```
joakim-noah: Needed by [this code path](https://github.com/dlang/phobos/blob/6fa5e0e969bd6a4a885932e9e13dbbeda1990c1b/std/file.d#L1260), which gets [this test of higher precision on linux](https://github.com/dlang/phobos/blob/6fa5e0e969bd6a4a885932e9e13dbbeda1990c1b/std/file.d#L1339) to pass on Android.

src/core/sys/posix/signal.d
```diff
@@ -159,16 +159,28 @@ else version( CRuntime_Glibc )
...
+    enum SIGRTMIN = 32;
+    version(ARM)
+        enum SIGRTMAX = 64;
+    else version(X86)
+        enum SIGRTMAX = 64;
+    else version(MIPS32)
+        enum SIGRTMAX = 128;
+    else
+        static assert(false, "Architecture not supported.");
+}
```
joakim-noah: Adding in preparation for #1565, whenever that's merged.

PR #1822, opened 2017-05-07
[DConf] GC TLS cache

DmitryOlshansky:
A first shot at resolving lock contention in the GC.

And fixing bugs in concurrent GC benchmarks.


DmitryOlshansky:
Closing in favor of more in depth approaches.

PR #1824, opened 2017-05-07
Array equals overload for dynamic arrays only

somzzz:
Fixes some bugs from https://github.com/dlang/druntime/pull/1814

Requires https://github.com/dlang/dmd/pull/6697 to be reverted first. Due to CTFE issues (https://issues.dlang.org/show_bug.cgi?id=17351) with static arrays passed as ref, we cannot support static array __equals yet. Remains to be discussed.

Dmd patch to enable this behaviour to be posted soon.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3229,41 +3229,21 @@ template RTInfo(T)
     enum RTInfo = null;
 }
 
-bool __equals(T1, T2, size_t n1)(auto ref T1[n1] lhs, T2[] rhs)
-{
-    T1[] lhs1 = lhs;
-    return __equals(lhs1, rhs);
-}
-
-bool __equals(T1, T2, size_t n2)(T1[] lhs, auto ref T2[n2] rhs)
-{
-    T2[] rhs1 = rhs;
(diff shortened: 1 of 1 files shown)


UplinkCoder:
@somzzz squash the whitespace fix please ;)

wilzbach:
@uplinkcoder there is no need to do this as long as the PR hasn't been approved and even then can the squashing easily been done by maintainers or even the DLang-Bot

src/object.d
```diff
@@ -3277,12 +3257,27 @@ bool __equals(T1, T2)(T1[] lhs, T2[] rhs)
...
+        foreach (const u; 0 .. lhs.length)
+        {
+            if (at(lhs, u) != at(rhs, u))
+                return false;
+        }
+        return true;
+    }
     else static if (__traits(isIntegral, U1))
     {
+        //pragma(msg, "__equals integral");
```
klickverbot: Please also remove these debug statements (they aren't good documentation, since the condition one line above them says exactly the same thing).

PR #1826, opened 2017-05-11
NetBSD: port struct statvfs from netbsd source

nrTQgc:


src/core/sys/posix/sys/statvfs.d
```diff
@@ -86,21 +86,61 @@ version(CRuntime_Glibc) {
 }
 else
 {
-    struct statvfs_t
+    version(NetBSD)
```
joakim-noah: Better to just create your own whole new `NetBSD` block, rather than stick this in here, ie replace the `else` just above with `else version(NetBSD)`, fill it with everything you need, and leave this last `else` block alone.

nrTQgc: fixed

src/core/sys/posix/sys/statvfs.d
```diff
@@ -84,6 +84,53 @@ version(CRuntime_Glibc) {
...
+        int f_owner;
+        int[4] f_spare;
+        char[_VFS_NAMELEN] f_fstypename;
+        char[_VFS_MNAMELEN] f_mntonname;
+        char[_VFS_MNAMELEN] f_mntfromname;
+    }
+
+    enum FFlag
+    {
+        ST_RDONLY = 1,        /* Mount read-only.  */
```
joakim-noah: Oh, is this comment copied from the header?  If so, remove it.  We don't allow copied comments for potential copyright reasons.

nrTQgc: I copied it from "else {}" block. It was here before me.

joakim-noah: OK, probably fine then.

joakim-noah:
ping, @klickverbot, small change.

joakim-noah:
Needs to be marked as reviewed by someone with commit access, my review doesn't cut it because I'm not  member.

PR #1828, opened 2017-05-22
correct mistake which caused idx to be inferrred as int

UplinkCoder:
as per title.
I found this mistake using `-vcg-ast`


UplinkCoder:
@andralex could you auto-merge this ?

andralex:
thx!

wilzbach:
@UplinkCoder shouldn't this (as a bug fix) be targeted at `stable`?

PR #1830, opened 2017-05-27
core.demangle: add support for type, identifer and symbol back references

rainers:
support for https://github.com/dlang/dmd/pull/5855 and a couple of fixes.

### DIFF
--- a/src/core/demangle.d
+++ b/src/core/demangle.d
@@ -100,6 +100,8 @@ pure @safe:
 
     static void error( string msg = "Invalid symbol" ) @trusted /* exception only used in module */
     {
+        pragma(inline, false); // tame dmd inliner
+
         //throw new ParseException( msg );
         debug(info) printf( "error: %.*s\n", cast(int) msg.length, msg.ptr );
         throw __ctfe ? new ParseException(msg)
@@ -110,6 +112,8 @@ pure @safe:
 
     static void overflow( string msg = "Buffer overflow" ) @trusted /* exception only used in module */
     {
+        pragma(inline, false); // tame dmd inliner
+
         //throw new OverflowException( msg );
         debug(info) printf( "overflow: %.*s\n", cast(int) msg.length, msg.ptr );
         throw cast(OverflowException) cast(void*) typeid(OverflowException).initializer;
@@ -176,6 +180,8 @@ pure @safe:
     // move val to the end of the dst buffer
     char[] shift( const(char)[] val )
     {
+        pragma(inline, false); // tame dmd inliner
+
         if( val.length && !mute )
         {
             assert( contains( dst[0 .. len], val ) );
@@ -196,6 +202,8 @@ pure @safe:
     // remove val from dst buffer
     void remove( const(char)[] val )
     {
+        pragma(inline, false); // tame dmd inliner
+
         if( val.length )
         {
             assert( contains( dst[0 .. len], val ) );
@@ -210,6 +218,8 @@ pure @safe:
 
     char[] append( const(char)[] val )
     {
+        pragma(inline, false); // tame dmd inliner
+
         if( val.length && !mute )
         {
             if( !dst.length )
@@ -252,6 +262,8 @@ pure @safe:
 
     char[] put( const(char)[] val )
     {
+        pragma(inline, false); // tame dmd inliner
+
         if( val.length )
         {
             if( !contains( dst[0 .. len], val ) )
--- a/src/core/demangle.d
+++ b/src/core/demangle.d
@@ -585,7 +585,12 @@ pure:
             return;
         }
         auto n = decodeNumber();
-        if( !n || n > buf.length || n > buf.length - pos )
+        if( n == 0 )
+        {
+            put( "__anonymous" );
+            return;
+        }
+        if( n > buf.length || n > buf.length - pos )
             error( "LName must be at least 1 character" );
         if( '_' != front && !isAlpha( front ) )
             error( "Invalid character in LName" );
@@ -2332,7 +2337,8 @@ unittest
     {
         auto mngl = mangleFunc!(int function(Object))("object.Object.opEquals");
         assert(mngl == "_D6object6Object8opEqualsFC6ObjectZi");
-        assert(reencodeMangled(mngl) == "_D6object6Object8opEqualsFCQtZi");
+        auto remngl = reencodeMangled(mngl);
+        assert(remngl == "_D6object6Object8opEqualsFCQsZi");
     }
     // trigger back tracking with ambiguity on '__T', template or identifier
     assert(reencodeMangled("_D3std4conv4conv7__T3std4convi") == "_D3std4convQf7__T3stdQpi");
@@ -2468,6 +2474,10 @@ version(unittest)
          "pure @safe int std.format.getNth!(\"integer width\", std.traits.isIntegral, int, uint, uint).getNth(uint, uint, uint)"],
         ["_D3std11parallelism42__T16RoundRobinBufferTDFKAaZvTDxFNaNdNeZbZ16RoundRobinBuffer5primeMFZv",
          "void std.parallelism.RoundRobinBuffer!(void delegate(ref char[]), bool delegate() pure @property @trusted const).RoundRobinBuffer.prime()"],
+        // Lname '0'
+        ["_D3std9algorithm9iteration__T9MapResultSQBmQBlQBe005stripTAAyaZQBi7opSliceMFNaNbNiNfmmZSQDiQDhQDa__TQCtSQDyQDxQDq00QCmTQCjZQDq",
+         "pure nothrow @nogc @safe std.algorithm.iteration.MapResult!(std.algorithm.iteration.__anonymous.strip, "
+        ~"immutable(char)[][]).MapResult std.algorithm.iteration.MapResult!(std.algorithm.iteration.strip, immutable(char)[][]).MapResult.opSlice(ulong, ulong)"],
 
         // back references
         ["_D4core4stdc5errnoQgFZi", "int core.stdc.errno.errno()"], // identifier back reference
--- a/src/core/demangle.d
+++ b/src/core/demangle.d
@@ -1483,21 +1483,21 @@ private struct Demangle
     TemplateInstanceName:
         Number __T LName TemplateArgs Z
     */
-    void parseTemplateInstanceName()
+    void parseTemplateInstanceName(bool hasNumber)
     {
         debug(trace) printf( "parseTemplateInstanceName+\n" );
         debug(trace) scope(success) printf( "parseTemplateInstanceName-\n" );
 
         auto sav = pos;
         scope(failure) pos = sav;
-        auto n = decodeNumber();
+        auto n = hasNumber ? decodeNumber() : 0;
         auto beg = pos;
         match( "__T" );
         parseLName();
         put( "!(" );
         parseTemplateArgs();
         match( 'Z' );
-        if( pos - beg != n )
+        if( hasNumber && pos - beg != n )
             error( "Template name length mismatch" );
         put( ')' );
     }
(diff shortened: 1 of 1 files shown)


ibuclaw:
Notably missing unittests, but I guess these will be added after compiler support is in.

rainers:
> Notably missing unittests, but I guess these will be added after compiler support is in.

The demangling is tested with the compiler test suite as I changed some of the mangling tests to just compare the result of demangling.

I guess I should some tests here, though, so we'll get some coverage locally before merging.

rainers:
Added some test cases.

Trying to demangle all symbols from phobos unittest, I'm hitting the amgibuity on Pascal function types, see https://github.com/dlang/dmd/pull/6702. The current implementation runs into a stack overflow. I should add some sanity checks anyway...

rainers:
As a test field to work on I have extracted all D symbols from the map file of the phobos unittest. For a build with master, there are **127172** symbols with a maximum length of **416133** characters and an average length of **369** characters.

With back references the same symbols have a maximum length of **1114** characters with an average of **117**.

I updated demangling to properly support back references, but a number of failures for demangling symbols with back references also affected symbols as emitted by master, so I fixed these, too.

Results:

branch   | failed | incomplete | max time   | average time
---------|--------|------------|------------|-------------
master   |   4192 |      19042 | 19 min 32 s| 140 ms
shiftopt |   4192 |      19042 | 19 ms      | 33 us
this_pr  |      0 |         25 | 36 ms      | 21 us

- shiftopt: master with https://github.com/dlang/druntime/pull/1830/commits/99ae2c56a1a9c7ffeb3494aa4dd8c1f7e5e60d96
- failed: core.demangle reports error (does not include bad results)
- incomplete: `__T` still found in demangled name (indicating a template is left mangled)
- max time: maximum time to decode one symbol

I cancelled the master test after about 3 hours when it was about half done (timings taken from that run). 
The optimization for branch shiftopt didn't change results of demangling, so the number of failed/incomplete demanglings is copied to the master results.

For the back refernce symbols, results only change a little:

branch   | failed | incomplete | max time   | average time
---------|--------|------------|------------|-------------
this_pr  |     36 |          0 | 43 ms      | 23 us

src/core/internal/traits.d
```diff
@@ -88,7 +88,12 @@ private template substInoutForm(T)
 /// used to declare an extern(D) function that is defined in a different module
 template externDFunc(string fqn, T:FT*, FT) if(is(FT == function))
 {
-    static if (is(FT RT == return) && is(FT Args == function))
+    static if (__traits(compiles, __traits(externDmangle, "foo", void function())))
+    {
+        alias _wrap(alias T) = T; // fool the parser
+        alias externDFunc = _wrap!(__traits(externDmangle, fqn, FT));
```
MartinNowak: We should be fixing `mangleFunc!T(fqn)` instead.

rainers: Unfortunately the qualified name in mangleFunc is a runtime parameter. No luck using a trait there...

rainers:
> Unfortunately the qualified name in mangleFunc is a runtime parameter.

As mangleFunc is a public documented function this quite a bummer. If we want to keep the runtime argument as is I don't see an alternative to actually implementing the mangler in druntime, too.

I'm not sure if this is possible with the recently added traits, but would be rather pretty pessimistic with respect to template arguments. As the qualified name didn't support these so far, it might not be a large breaking change to restrict the function type to non-template types, too.

wilzbach:
> Thanks for your pull request, @rainers! We are looking forward to reviewing it, and you should be hearing from a maintainer soon.

FYI: this is a newly deployed feature on the dlang-bot and it will only post _once_ for each PR and then update its comment with (hopefully) helpful messages, see e.g.
- https://github.com/dlang-bots/dlang-bot/pull/44
- https://github.com/dlang-bots/dlang-bot/pull/57

All bugs or feature requests can be reported [dlang-bot's repo](https://github.com/dlang-bots/dlang-bot).

rainers:
Planning to throw out symbol back references and just go with type and symbol back references, I've now implemented mangleFunc with the help of the demangler and a little bit of DbI.

I had to add some `pure` and `@safe` hacks, though. The latter can be narrowed to a number of places if required.

rainers:
Rebases with symbol back references removed.

rainers:
I've restricted the trusted code to a few small functions, so the demangler can be annotated safe. Apart from adding support for back references this PR now fixes some other issues, too:

- https://issues.dlang.org/show_bug.cgi?id=14319 core.demangle does not support member function attributes 
- https://issues.dlang.org/show_bug.cgi?id=14563 core.demangle: Does not demangle type modifers
- https://issues.dlang.org/show_bug.cgi?id=14576 core.demangle unable to handle ambiguity in symbols
- https://issues.dlang.org/show_bug.cgi?id=16664 core.demangle functions are not callable from @safe or pure code

wilzbach:
> Apart from adding support for back references this PR now fixes some other issues, too:

FYI: you can tag as many issues as you want to the dlang-bot / changelog in one commit.
See also: [the regex used by Bugzilla and the bot](https://github.com/dlang-bots/dlang-bot#automated-references).

rainers:
> FYI: you can tag as many issues as you want to the dlang-bot / changelog in one commit.

Ok, rebased with a couple of issues added to the commit messages.

rainers:
@wilzbach @CyberShadow I think its pretty demotivating seeing almost every PR in the overview marked with a red cross, even though these are caused by spurious or unrelated failures of coverage and jenkins builds. Maybe the resulting build state should be evaluated from the required tests only. Is this possible?

CyberShadow:
@rainers I agree completely, and I've been complaining about this since forever to anyone who can listen, but I'm not sure what I can do at this point. If you have time to spare, tracking down and fixing the causes of the random failures or coverage fluctuations would be much appreciated.

@wilzbach What is the significance of "74.92% of diff hit (target 75.38%)"? Is it to say that of the lines added in this PR, only 74% are covered? If so, why the oddly specific 75.38% threshold?

CyberShadow:
> Maybe the resulting build state should be evaluated from the required tests only. Is this possible?

Unfortunately I don't think it's possible. It would be nice if we could get rid of the codecov statuses and only use their bot's comments or integrate them with dlang-bot's comments, though.

wilzbach:
> I think its pretty demotivating seeing almost every PR in the overview marked with a red cross, even though these are caused by spurious or unrelated failures of coverage and jenkins builds.

I fully agree.

> Maybe the resulting build state should be evaluated from the required tests only. Is this possible?

Well actually Jenkins is supposed to be required, but we had quite some issues with it over the last few weeks.
And as @Cybershadow mentioned it's not possible to exclude status checks on the GitHub summary. Sadly if one check is red, the PR is marked as red.
At least for Jenkins I have good news as Martin and Sönke are tracking down the issues and have already fixed two.

> but I'm not sure what I can do at this point. If you have time to spare, tracking down and fixing the causes of the random failures or coverage fluctuations would be much appreciated.

Yeah, I took us quite a long time to fix all spots at Phobos.

> @wilzbach What is the significance of "74.92% of diff hit (target 75.38%)"? Is it to say that of the lines added in this PR, only 74% are covered? 

Yes.

> If so, why the oddly specific 75.38% threshold?

Target is the overall coverage that the project had before. I think the motivation is that you want to avoid decreasing the overall coverage with every PR.

> Unfortunately I don't think it's possible. It would be nice if we could get rid of the codecov statuses and only use their bot's comments or integrate them with dlang-bot's comments, though.

This shouldn't be too difficult, but I am not sure whether people would really care about the bot's comment. CodeCov's solution is to send a new comment for every new coverage event which is quite noisy. Unfortunately the GH status API really sucks and we can't automatically receive the status hook and remove it, but would have to crawl CodeCov for updates.

CyberShadow:
> This shouldn't be too difficult, but I am not sure whether people would really care about the bot's comment.

It seems that it boils down to one of the following:
- Red X means "sometimes it's OK to merge" instead of "DO NOT MERGE";
- We adjust CodeCov settings so that no reasonable PR would ever have a "failed" status.

IMO the second one has more value, as lack of trust in CI is not a problem we want to have, and seems more important than the artificial coverage metric.

I understand that even if we were to set the target to a very low value or even 0, codecov will still continue creating the status updates, and thus it should still be possible to see the difference in coverage at a glance, right? If so, then to me that seems like the preferable option.

wilzbach:
> We adjust CodeCov settings so that no reasonable PR would ever have a "failed" status.

I don't know how that would be possible, the only thing that comes to my mind that we can do for now is https://github.com/dlang/druntime/pull/1874 (disable the status notifications).

> I understand that even if we were to set the target to a very low value or even 0, codecov will still continue creating the status updates, and thus it should still be possible to see the difference in coverage at a glance, right? If so, then to me that seems like the preferable option.

AFAICT is the target the current project coverage. We could try to manually set the target coverage to `0` (see e.g. https://github.com/codecov/support/wiki/Codecov-Yaml), but at least for the next two or three weeks I don't have the time to play around with this. However, I just saw that CodeCov allows to specify an arbitrary webhook URL, so I can use the opportunity to plugin the bot, s.t. it can repost things (with a green status) in the future.

rainers:
> I don't know how that would be possible, the only thing that comes to my mind that we can do for now is #1874 (disable the status notifications).

Thanks for working on this. Is that supposed to make a difference for new builds already?

I have tried to avoid some indeterministic results here: https://github.com/dlang/druntime/pull/1875

src/core/demangle.d
```diff
@@ -164,43 +177,57 @@ private struct Demangle
...
-            for( size_t n = 0; n < val.length; n++ )
-            {
-                for( size_t v = val.ptr - dst.ptr; v + 1 < len; v++ )
-                {
-                    exch( v, v + 1 );
-                }
-            }
+            if (len + val.length > dst.length)
+                overflow();
+            size_t v = &val[0] - &dst[0];
```
andralex: nice

src/core/demangle.d
```diff
@@ -296,6 +325,13 @@ private struct Demangle
         return char.init;
     }
 
+    char peek( size_t n )
+    {
+        if( pos + n < buf.length )
+            return buf[pos + n];
```
andralex: I speculate speed is important here - perhaps you can make this trusted and skip bounds checking.

andralex:
I'll boldly go where only @rainers has gone and merge this.

rainers:
@andralex Thanks for merging
@wilzbach This is now merged to branch mangle, not master. Is that as expected?

wilzbach:
> Thanks for working on this. Is that supposed to make a difference for new builds already?

Yes, but only if they are based on a master commit that contains these changes (unfortunately `codecov.yml` is rather stupid).
For example:

image

https://github.com/dlang/druntime/pull/1877

It doesn't show up anymore, but it's built and analyzed:

https://codecov.io/gh/dlang/druntime/commit/d4015f4d403bd22ecee8b81bb61378d33e7df7df

> I have tried to avoid some indeterministic results here: #1875

Thanks a lot.

> @wilzbach This is now merged to branch mangle, not master. Is that as expected?

Yes, I didn't anticipate that this would be merged so quickly (it could have been rechanged to `master`), but luckily @MartinNowak helped out with https://github.com/dlang/druntime/pull/1877

rainers:
> CodeCov doesn't show up anymore, but it's built and analyzed:

Thanks, looks good.
 
> https://codecov.io/gh/dlang/druntime/commit/d4015f4d403bd22ecee8b81bb61378d33e7df7df

Can the link be displayed somewhere? Or will it be necessary to build it manually or copy it from circleCI's output?


> Yes, I didn't anticipate that this would be merged so quickly

Nor did I ;-)

wilzbach:
> Can the link be displayed somewhere? Or will it be necessary to build it manually or copy it from circleCI's output?

https://github.com/dlang-bots/dlang-bot/pull/142 (for now we will just hard-code it to success and display it in the status bar). In the future, we might also show the message in the comment, but that's slightly more complicated as we would need to parse the existing comment when try to modify it)

PR #1833, opened 2017-06-05
Make mulu faster when not defined as an intrinsic

andralex:
Define overloads taking mixed widths, and guard division with the (frequent) case when both operands fit within one word.

### DIFF
--- a/src/core/checkedint.d
+++ b/src/core/checkedint.d
@@ -655,9 +655,9 @@ pragma(inline, true)
 uint mulu(uint x, uint y, ref bool overflow)
 {
     ulong r = ulong(x) * ulong(y);
-    if (r > uint.max)
+    if (r >> 32)
         overflow = true;
-    return cast(uint)r;
+    return cast(uint) r;
 }
 
 unittest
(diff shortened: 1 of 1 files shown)


src/core/checkedint.d
```diff
@@ -682,10 +682,27 @@ unittest
 
 /// ditto
 pragma(inline, true)
+ulong mulu(uint x, ulong y, ref bool overflow)
+{
+    ulong r = x * y;
+    if (y >> 32 && x && (r / x) != y)
+        overflow = true;
```
WalterBright: needs a test case!

src/core/checkedint.d
```diff
@@ -682,10 +682,27 @@ unittest
 
 /// ditto
 pragma(inline, true)
+ulong mulu(uint x, ulong y, ref bool overflow)
+{
+    ulong r = x * y;
+    if (y >> 32 && x && (r / x) != y)
```
WalterBright: separating this into 3 lines:
```
if (y >> 32 &&
x &&
(r/x)!=y)
```
will enable code coverage testing to check each line. I think thoroughly testing these case is particularly important for this function.

andralex: Ugh. I'd so much rather improve the codecov support...

wilzbach: @andralex this is D's coverage output which will insert `000001` as soon as this line is hit (no condition coverage).

Other languages do support this, e.g.
https://github.com/codecov/support/issues/227

andralex: @wilzbach I know... we should get better at this either by count or by supporting columns.

andralex:
Updated to use a simpler test (same speed) and added tests.

src/core/checkedint.d
WalterBright: If a condition has 3 clauses, `A && B && C`, then if all 3 are hit then the line count will be `3`. So if the line above has 1 hit, and the conditional has 3, you know all 3 are hit. If the line count is `1`, then you know only the first conditional was hit.
Not the greatest, but still more information than just "was any clause hit".

andralex: @WalterBright simpler would be better

WalterBright: I agree. But redoing the coverage analyzer is a longer term goal. In the meantime, let's just get those 3 clauses covered. It's important because overflow detection is an often misunderstood and misprogrammed issue. Having good coverage will inure us from someone "fixing" it in the future.

src/core/checkedint.d
```diff
@@ -741,4 +763,4 @@ unittest
     assert(mulu(0UL, 0UL, overflow) == 0);
     assert(overflow);                   // sticky
 }
-}
```
WalterBright: What's up with this?

andralex: That was the last line of the file. My editor inserted a CR after it.

andralex:
Ran some measurements, and with dmd and gdc we're looking at dramatic improvements of `mulu(ulong, ulong, ref bool)` on random `uint`s - one order of magnitude. For ldc somehow the relative improvement is much smaller, but it doesn't matter because ldc implements mulu as an intrinsic.

src/core/checkedint.d
```diff
@@ -682,29 +682,75 @@ unittest
...
+        overflow = true;
+    return r;
+}
+
+/// ditto
+pragma(inline, true)
+ulong mulu(ulong x, uint y, ref bool overflow)
+{
+    return mulu(y, x, overflow);
+}
```
ibuclaw: I don't think `(ulong, uint, ref bool)` override is really needed, we don't do this for others.

andralex: ok, it's just for convenience but druntime is not about convenience

src/core/checkedint.d
```diff
@@ -655,9 +655,9 @@ pragma(inline, true)
 uint mulu(uint x, uint y, ref bool overflow)
 {
     ulong r = ulong(x) * ulong(y);
-    if (r > uint.max)
+    if (r >> 32)
```
andralex: This generates smaller code.

WalterBright: Not no more: https://github.com/dlang/dmd/pull/6865

klickverbot: Fun fact: LLVM actually canonicalises the `if (shift)` into an unsigned comparison first, and then goes on to emit the `shr` in the backend.

andralex:
Well... somebody please approve something!

ibuclaw:
You still have the override for `(ulong, uint, ref bool)` - is that really needed for the tests to pass?

andralex:
>You still have the override for (ulong, uint, ref bool) - is that really needed for the tests to pass?

That is useful because it takes advantage of static knowledge - does one OR and one comparison less.

ibuclaw:
Fine, I just need to remember to add the signature to the compiler to recognize it as an intrinsic.

andralex:
thx!

src/core/checkedint.d
WalterBright: Line ending problems is why I run my submissions through a line ending fixer (tolf in the D tools).

WalterBright: tolf is also used in druntime's posix.mak to canonicalize the line endings.
Other than that, I approve this PR.

PR #1834, opened 2017-06-09
Unify Makefile variables for dlang.org: Rename DOCDIR -> DOC_OUTPUT_DIR, DOCFMT -> STDDOC

wilzbach:
In [an attempt](https://github.com/dlang/dlang.org/pull/1687) to reduce the mess in the dlang.org Makefile, it makes sense and helps to use the same variable names for the same things.
`DOC_OUTPUT_DIR` and `STDDOC` are already used at Phobos + DMD.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -28,7 +28,8 @@ endif
 DMD=$(DMD_DIR)/generated/$(OS)/$(BUILD)/$(MODEL)/dmd
 INSTALL_DIR=../install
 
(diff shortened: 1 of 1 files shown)


WalterBright:
Some documentation in the makefile itself is necessary for these additions. Document what the variable is for and what its expected values are and who should be setting/overriding it.

RazvanN7 (approved):

PR #1836, opened 2017-06-11
Add pureFree to core.memory

nordlow:
To be used in contexts such as `RefCounted`-destructor instead of having to define local overloads such as here https://github.com/dlang/phobos/pull/4832/files#diff-4e008aedb3026d4a84f58323e53bf017R5087.

AFAICT, `free` doesn't set `errno`.

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -811,9 +811,11 @@ struct GC
 }
 
 /**
- * Pure variants of C's memory allocation functions `malloc`, `calloc` and
- * `realloc`. Purity is achieved via resetting the `errno` to it's value prior
- * to being called, removing the function's global state mutation.
+ * Pure variants of C's memory allocation functions `malloc`, `calloc`, and
+ * `realloc` and deallocation function `free`.
+ *
+ * Purity is achieved by saving and restoring the value of `errno`, thus
+ * behaving as if it were never changed.
  *
  * See_Also:
  *     $(LINK2 https://dlang.org/spec/function.html#pure-functions, D's rules for purity),
@@ -841,7 +843,7 @@ void* pureCalloc(size_t nmemb, size_t size) @trusted pure @nogc nothrow
     return ret;
 }
 /// ditto
-void* pureRealloc(void* ptr, size_t size) pure @nogc nothrow
+void* pureRealloc(void* ptr, size_t size) @system pure @nogc nothrow
 {
     const errno = fakePureGetErrno();
     void* ret = fakePureRealloc(ptr, size);
@@ -851,12 +853,17 @@ void* pureRealloc(void* ptr, size_t size) pure @nogc nothrow
     }
     return ret;
 }
+/// ditto
+void pureFree(void* ptr) @system pure @nogc nothrow
+{
+    const errno = fakePureGetErrno();
+    fakePureFree(ptr);
+    cast(void)fakePureSetErrno(errno);
+}
 
 ///
-nothrow @nogc unittest
+@system pure nothrow @nogc unittest
 {
-    import core.stdc.stdlib : free;
-
     ubyte[] fun(size_t n) pure
     {
         void* p = pureMalloc(n);
@@ -869,10 +876,10 @@ nothrow @nogc unittest
 
     auto buf = fun(100);
     assert(buf.length == 200);
-    free(buf.ptr);
+    pureFree(buf.ptr);
 }
 
-pure @nogc nothrow unittest
+@system pure nothrow @nogc unittest
 {
     const int errno = fakePureGetErrno();
 
@@ -909,5 +916,5 @@ extern (C) private pure @system @nogc nothrow
     pragma(mangle, "calloc") void* fakePureCalloc(size_t nmemb, size_t size);
     pragma(mangle, "realloc") void* fakePureRealloc(void* ptr, size_t size);
 
-    pragma(mangle, "free") void fakePureFree(void* ptr); // needed by unittests
+    pragma(mangle, "free") void fakePureFree(void* ptr);
 }


nordlow:
What's wrong with the **continuous-integration/jenkins/pr-merge**?

wilzbach:
>  What's wrong with the continuous-integration/jenkins/pr-merge?

It's unrelated - it's a regression introduced in https://github.com/dlang/phobos/pull/5427

src/core/memory.d
```diff
@@ -851,12 +853,15 @@ void* pureRealloc(void* ptr, size_t size) pure @nogc nothrow
     }
     return ret;
 }
+/// ditto
+void pureFree(void* ptr) @system pure @nogc nothrow
```
Calrama: `free` is [not guaranteed not to set `errno` by current standards](https://sourceware.org/ml/libc-alpha/2015-02/msg00067.html); libc implementations may choose to set errno for both valid and invalid pointers.
POSIX.1 Issue 8 (still WIP), [is going to forbid  `free` to set `errno` for valid pointers](http://austingroupbugs.net/view.php?id=385), but explicitly allows setting it for an invalid pointers.
AFAIK there's no way for us to know whether the used libc will be linked to statically or dynamically and in the latter case there's no way AFAIK to know if the libc does or does not set `errno`.
It follows that `pureFree` must save&restore the errno value the same way that the allocation functions do:
```d
const errno = fakePureGetErrno();
fakePureFree(ptr);
if (errno != 0) fakePureSetErrno(errno);
```

src/core/memory.d
```diff
@@ -811,9 +811,11 @@ struct GC
 }
 
 /**
- * Pure variants of C's memory allocation functions `malloc`, `calloc` and
- * `realloc`. Purity is achieved via resetting the `errno` to it's value prior
- * to being called, removing the function's global state mutation.
+ * Pure variants of C's memory allocation functions `malloc`, `calloc`, and
+ * `realloc` and deallocation function `free`.
+ *
+ * Purity for allocation is achieved via resetting the `errno` to it's value
```
Calrama: Since deallocation must also save&restore errno (see comment at `pureFree`):
Purity for allocation ->  Purity for allocation and deallocation

JackStouffer:
I thought we all agreed that `free` can't be `pure`?

Calrama:
> AFAICT, free doesn't set errno

This may be true for your libc, but it's not guaranteed by ISO C or POSIX. The WIP POSIX.1 Issue 8 even explicitly mentions that implementations are free to set `errno` if they receive an invalid pointer as input.

wilzbach:
> AFAICT, `free` doesn't set `errno`.
> free is not guaranteed not to set errno by current standards; 

FWIW it doesn't matter for the `pure`ness considerations. For example many functions in [std.math](https://dlang.org/phobos/std_math.html) can set `errno` as well, but they are considered as `pure`.

wilzbach:
> AFAICT, `free` doesn't set `errno`.

Linking to the previous threaD: https://github.com/dlang/druntime/pull/1746

JackStouffer:
>FWIW it doesn't matter for the pureness considerations. For example many functions in std.math can set errno as well, but they are considered as pure.

That's a bug then. Can you please make a bugzilla issue?

Calrama:
> FWIW it doesn't matter for the pureness considerations. For example many functions in std.math can set errno as well, but they are considered as pure.

`errno` is global state. Changing is is impure, so those are bugs.

Calrama:
> I thought we all agreed that free can't be pure?

I have read [this discussion](http://forum.dlang.org/thread/32f556d7-fd1f-048a-ebf0-22c98ff894e3@puremagic.com) and the thread @wilzbach linked to, but I saw (so far) no convincing argument as to why `free` should not be pure (in D's definition of pure, which is weakly pure w.r.t. functional PL's pure)

JackStouffer:
@Calrama Check out this discussion https://github.com/dlang/druntime/pull/1718

Ping @schveiguy

Calrama:
@JackStouffer Thank you. AFAICT the described issue lies with people casting away `immutable` inside a `pure` function and then passing that to `free`, because the compiler is free to remove the call to `free` as it has no return value (and thus no effect by the rules of `pure`)?
What I don't understand there is this: `pure` functions are *required* to *only* take `immutable` arguments, so shouldn't casting away `immutable` from an argument inside `pure` be illegal, i.e. produce a compiler error?

jmdavis:
> What I don't understand there is this: pure functions are required to only take immutable arguments, so shouldn't casting away immutable from an argument inside pure be illegal, i.e. produce a compiler error?

`pure` functions have not been required to take `immutable` arguments for years now. I would suggest that you read this: http://klickverbot.at/blog/2012/05/purity-in-d/

Calrama:
> pure functions have not been required to take immutable arguments for years now.

Sorry, momentary confusion caused by looking at the [FAQ, which still states that](https://dlang.org/const-faq.html#invariant).

andralex:
@Calrama could you please post an issue against https://dlang.org/const-faq.html#invariant? Better yet, a PR fixing it. Thanks!

src/core/memory.d
andralex: Yah, let's set and restore it for now regardless. We may change that later as an optimization. Thanks!

src/core/memory.d
andralex: s/Purity for allocation is achieved via resetting the `errno` to it's value/Purity for allocation is achieved by saving and restoring the value of `errno`, thus behaving as if it were never changed/

andralex:
cc @nordlow

Calrama:
@andralex Sure, [PR is open](https://github.com/dlang/dlang.org/pull/1847)

schveiguy:
The issue with `pure free` is that it could be elided by the compiler. `pure malloc` doesn't have the same problem because of the return type.

What is going to happen is that someone will need to free immutable or const data. They will cast away the immutable and call `pureFree`. Then that function will be elided - and memory will leak.

The only grace of this is that we are not actually marking the real `free` function, so when we have to rescind this in the future due to subtle nasty memory leaks, not much will be affected, or at least code that is affected will have opted in to the idea that `free` can be pure.

I'm not going to continually have this argument (please see the previous PR), it seems enough people want to try this doomed experiment, so go ahead.

nordlow:
Why should the compiler elide successive calls to a function taking mutable references?

andralex:
>I'm not going to continually have this argument (please see the previous PR), it seems enough people want to try this doomed experiment, so go ahead.

Steve, I'm hoping for a different tack on this. Yes, it is clear there is a problem. The problem stands in the way of people doing things that are useful. So we need to find a solution to the problem. Stating that the problem is not solvable and "can't do that" has low informational entropy - a bunch of folks will confirm they can't solve it, starting with the mailman ringing the door right now. What is interesting and good and what we should aim for is the high information response - a creative solution that solves the problem. What are your thoughts on that?

schveiguy:
I didn't say the problem wasn't solvable, I said that solving it this way will likely lead to subtle hard-to-detect leaks.

My recommendation (as before) is to let people tackle this using private declarations of `pureFree` in their own applications. For instance using it in the `RefCounted` struct. I think with that level of control, you can make sure it's not elided.

Note that this all really depends on the type, not the call. If the type is not fully immutable/const, then the compiler cannot elide `pure` calls which use it. In the case of `RefCounted`, the reference count is always mutable, so I'm assuming that the compiler sees this and won't allow elision.

We can't make such guarantees if we simply provide an easy way to do it. So we let the user make the guarantees on their own.

It's like lock-free programming. There's lots of wrong ways to do it, and in most cases the wrong ways still work most of the time. It's those times it doesn't work where you are in big trouble.

schveiguy:
> Why should the compiler elide successive calls to a function taking mutable references?

It's not the call to `pureFree` that would be elided, but the call to a function that casts away immutable and calls `pureFree`.

And it's not successive calls either, it would be all the calls. A pure function that takes no mutable references and returns nothing can simply be elided by the compiler 100% of the time.

Calrama:
> It's not the call to pureFree that would be elided, but the call to a function that casts away immutable and calls pureFree.

I see the problem, but there is one question I have: The [spec states](https://dlang.org/spec/const3.html#const_and_immutable) that
>> Immutable data values, once constructed, remain the same for the duration of the program's execution.

From that I would assume that casting away `immutable` in order to free an object *is illegal* in D (and as such UB), i.e. an immutable object must remain alive for the entirety of the program's remaining execution time.

schveiguy:
> From that I would assume that casting away immutable in order to free an object is illegal in D

If there is no other reference, and you are about to forget the last reference (i.e. you are freeing it), then how is it illegal? For all intents and purposes, that piece of memory that you no longer have access to may very well still have the value assigned to it. In fact, free may not even change the value, but just store it away for safe keeping!

We are already accepting that malloc/free isn't really modifying global data, even though it is. We have to have some level of practicality that you can free immutable data as long as you no longer reference it.

Note that freeing immutable objects already cast away immutable (the destructor is not marked immutable).

It would be a good idea to update the spec for this.

Calrama:
> If there is no other reference, and you are about to forget the last reference (i.e. you are freeing it), then how is it illegal?

Because the spec - as it is currently worded - explicitly forbids the freeing of immutable objects. Once constructed they must remain alive for the remainder of the program's execution, irrespective of whether you are still using them.

schveiguy:
> Because the spec - as it is currently worded - explicitly forbids the freeing of immutable objects.

Then the GC is in violation of the spec. Again, we have to put aside these kinds of requirements when talking about allocators. This is like the old question, if a tree falls in a forest but nobody is around, does it make a sound? Similarly, if nothing can access a piece of immutable memory, is it illegal to modify it?

schveiguy:
> In my opinion yes, because immutable data need not even be stored in writable memory

Sorry I thought in this context we were only talking about allocated RAM, not ROM.

Calrama:
> Sorry I thought in this context we were only talking about allocated RAM, not ROM.

For one thing, you can have write protected RAM, using e.g. [`mprotect`](https://linux.die.net/man/2/mprotect) on Linux; it would be perfectly valid to store immutable objects in memory pages that are write protected and are only temporarily writeable for construction of new immutable objects.
For another, when looking at an immutable object's address in a process' virtual address space, how would you be able to differentiate between whether that address is mapped into ROM or RAM?

src/core/memory.d
```diff
@@ -851,12 +853,15 @@ void* pureRealloc(void* ptr, size_t size) pure @nogc nothrow
...
 }
+/// ditto
+void pureFree(void* ptr) @system pure @nogc nothrow
+{
+    fakePureFree(ptr);
+}
 
 ///
-nothrow @nogc unittest
+pure nothrow @nogc unittest
```
nordlow: Why should it be marked as `@system`?

src/core/memory.d
```diff
@@ -869,10 +874,10 @@ nothrow @nogc unittest
 
     auto buf = fun(100);
     assert(buf.length == 200);
-    free(buf.ptr);
+    pureFree(buf.ptr);
 }
 
-pure @nogc nothrow unittest
+pure nothrow @nogc unittest
```
nordlow: Done.

src/core/memory.d
nordlow: Changed to `its`.

nordlow:
Updates made according to comments.

src/core/memory.d
nordlow: Done.

jmdavis:
It seems to me that given the issues with call elision that Steven brought up, `pureFree` needs a warning on it explaining the issue, because it's exactly the sort of thing that most folks won't have a clue about. They'll just blindly use it when they want to use `free` in `pure` code without understanding when it's safe to use it or not, and they'll end up leaking memory. The fact that Steven keeps having to explain it highlights the fact that it's not obvious.

andralex:
@jmdavis a documentation PR would be very welcome - thanks!

FeepingCreature:
General point: pure functions *cannot* be elided. *Total* functions can be elided; pure functions are not necessarily total, ie. they do not necessarily return. For instance, this is a pure function:

    pure void foo() {
      while (true) {
      }
    }

schveiguy:
It all depends on how D defines it. A pure function that never returns isn't an interesting or desirable behavior. We could simply say that pure functions that don't ever return are undefined behavior, and let the compiler assume it must return at some point, allowing the obvious optimization. We could also dictate that it must be called the first time, but this means we will waste cycles calling functions that easily could be elided. Either way, it's not really a *fundamental* property of purity in general that the first call must be done.

After all, pure functions model mathematics, and I don't know if there's a math book that describes functions that never return ;)

aG0aep6G:
> We could simply say that pure functions that don't ever return are undefined behavior, and let the compiler assume it must return at some point, allowing the obvious optimization.

What about inferred `pure`?

```d
void foo()() /* a template now */
{
  while (true) {}
}
```

The compiler will infer `pure`. Surely, we don't want `foo()` to have undefined behavior.

schveiguy:
> Surely, we don't want `foo()` to have undefined behavior.

I'd be fine with the compiler assuming that pure functions always return. I'm not convinced that it's detrimental for a function that never returns to be elided if it does nothing observable inside.

I'm trying to see the benefit of having an opaque function like `void pure foo();` not be elided. "because it might have an infinite loop" doesn't seem worth giving up on that optimization benefit.

FeepingCreature:
I want to emphasize that Phobos has had actual production bugs due to confusing pure and total functions.

schveiguy:
Great, it should be easy to show then! Can you point me at the bug? The bug report I saw you posted on the forums doesn't have anything to do with pure functions.

schveiguy:
OK, so if I can be a bit pedantic, what I think you are saying is that something like `true.repeat.find(false)` would somehow be optimized into returning something instead of going into the infinite loop it should have? What would it optimize into, just an empty range?

I'm not sure if I got it right, but I'm still not sure where a strong-pure function makes sense to run just because it might contain an infinite loop (which IMO is clearly a programming error).

FeepingCreature:
That's exactly what happens, since `find` returns the remaining range, which is statically known (enum) to be infinite.

schveiguy:
OK, I see what you are saying (I think):

`true.repeat` is comprised solely of enum and no data. So if you have a function that returns that, the return value basically cannot vary based on the inputs.

Since all the inputs and outputs are immutable or enum, this is tagged as a strong pure function. The logic in the compiler could deduce that it must ALWAYS return that range, which indicates the value (`false`) was found. But the real answer is that it cannot find that item, and it should never actually return.

I see the flaw assuming that there is no infinite loop in any strong-pure function, and definitely this cannot be allowed to happen. Thanks for giving the clear example.

I still am concerned with applying the same restriction to void-returning functions. In this case, the control flow of the program (assuming it returns) is not affected based on anything the function does. One cannot assume incorrectly anything based on the return value. It truly is a "do nothing" call. It's a different thing to assume the return value based on the type than it is to assume it returns at all. In one case, the two are linked, due to the immutable property of the type, but in the other case, the return value doesn't exist, so it's not relevant. Literally the ONLY thing a strong-pure void-returning function can do is either use a finite amount of time to execute, or never return.

I'd prefer simply never calling a void strong-pure function to calling it only the first time. This allows generic programming that doesn't need to be hand-optimized based on some admittedly strange criteria.

FeepingCreature:
It does seem like a void-returning strong pure function should not fulfill any purpose. This does feel like the setup for a really subtle bug somewhere though.

PR #1837, opened 2017-06-12
Eliminate one shared static this from thread.d

andralex:
The Windows implementation does not need any initialization.

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -909,39 +909,172 @@ class Thread
     // Thread Priority Actions
     ///////////////////////////////////////////////////////////////////////////
 
+    version( Windows )
+    {
+        @property static int PRIORITY_MIN() @nogc nothrow pure @safe
+        {
+            return THREAD_PRIORITY_IDLE;
+        }
 
-    /**
-     * The minimum scheduling priority that may be set for a thread.  On
-     * systems where multiple scheduling policies are defined, this value
-     * represents the minimum valid priority for the scheduling policy of
-     * the process.
-     */
-    __gshared const int PRIORITY_MIN;
+        @property static const(int) PRIORITY_MAX() @nogc nothrow pure @safe
+        {
+            return THREAD_PRIORITY_TIME_CRITICAL;
+        }
 
+        @property static int PRIORITY_DEFAULT() @nogc nothrow pure @safe
+        {
+            return THREAD_PRIORITY_NORMAL;
+        }
+    }
+    else
+    {
+        private struct Priority
+        {
+            int PRIORITY_MIN = int.min;
+            int PRIORITY_DEFAULT = int.min;
+            int PRIORITY_MAX = int.min;
+        }
 
-    /**
-     * The maximum scheduling priority that may be set for a thread.  On
-     * systems where multiple scheduling policies are defined, this value
-     * represents the maximum valid priority for the scheduling policy of
-     * the process.
-     */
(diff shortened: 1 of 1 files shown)


andralex:
Ah, spoke too fast. NetBSD is Posix so it uses the same implementation.

andralex:
ping

andralex:
ping

CyberShadow:
CC @nrTQgc @nemanja-boric-sociomantic 

@andralex I think this PR would be easier to review if you described the changes in more detail (ideally in the commit message), as the diff is hard to make sense of even with ?w=1.

src/core/thread.d
```diff
@@ -909,39 +909,159 @@ class Thread
...
+        {
+            Priority result;
+            version( Solaris )
+            {
+                pcparms_t pcParms;
+                pcinfo_t pcInfo;
 
+                pcParms.pc_cid = PC_CLNULL;
+                if (priocntl(idtype_t.P_PID, P_MYID, PC_GETPARMS, &pcParms) == -1)
+                    assert( 0, "Unable to get scheduling class" );
```
nemanja-boric-sociomantic: Nit pick: these asserts could have the unified syntax with `version(Posix)` - no `if`s

src/core/thread.d
```diff
@@ -909,39 +909,159 @@ class Thread
...
+        private struct Priority
+        {
+            int PRIORITY_MIN = int.min;
+            int PRIORITY_DEFAULT = int.min;
+            int PRIORITY_MAX = int.min;
+        }
 
+        static int loadGlobal(string which)() //@nogc nothrow pure @safe
+        {
+            shared Priority cache;
```
nemanja-boric-sociomantic: Can this be `static shared Priority cache`?

nemanja-boric-sociomantic:
Other than that LGTM.

nemanja-boric-sociomantic:
Just to check, this comment:

```
The NetBSD version was elaborately just returning zero it seems, so I simplified that
```

is not valid anymore, right? (I can't see any changes done for `NetBSD`, which your second comment seems to confirm).

Burgos:
Ping :-)

src/core/thread.d
andralex: @nemanja-boric-sociomantic ah, that was the whole idea :). Thanks!

src/core/thread.d
andralex: ok

andralex:
@nemanja-boric-sociomantic yah, that comment about NetBSD is not valid, sorry for the confusion.

src/core/thread.d
andralex: Looking over code, I don't get the idea for refactoring.

andralex:
@nemanja-boric-sociomantic done

andralex:
@CyberShadow added comments

src/core/thread.d
Burgos: Yeah, me neither anymore. All good, sorry.

Burgos: Ah, I see it now: https://github.com/dlang/druntime/pull/1837/files#diff-8bb12ed976acf0a5132e877ec5a01ea8R1021

`priocntl(...) != -1 || assert (0, "Unable to get scheduling class.");`

I don't think it's important, though.

Burgos:
Looks good, thanks!

nemanja-boric-sociomantic:
Just needs squash @andralex

PR #1838, opened 2017-06-13
GetMarshalSizeMax takes a DWORD as second last arg

WebFreak001:
see https://msdn.microsoft.com/en-us/library/windows/desktop/ms680062(v=vs.85).aspx, it took a `DWORD*` before


rainers:
Thanks. It's also wrong in the MinGW headers that were the basis for these files.

PR #1839, opened 2017-06-15
add core.stdc.assert_

WalterBright:
For those who would want to call the C assert function rather than the druntime one.

### DIFF
--- a/src/core/stdc/assert_.d
+++ b/src/core/stdc/assert_.d
@@ -1,0 +1,70 @@
+/**
+ * D header file for C99.
+ *
+ * $(C_HEADER_DESCRIPTION pubs.opengroup.org/onlinepubs/009695399/basedefs/_assert.h.html, _assert.h)
+ *
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Source:    $(DRUNTIMESRC core/stdc/_assert_.d)
+ * Standards: ISO/IEC 9899:1999 (E)
+ */
+
+/****************************
+ * These are the various functions called by the assert() macro.
+ * They are all noreturn functions, although D doesn't have a specific attribute for that.
+ */
+
+module core.stdc.assert_;
+
+extern (C):
+@trusted:
+nothrow:
+@nogc:
+
+version (CRuntime_DigitalMars)
+{
+    /***
+     * Assert failure function in the Digital Mars C library.
+     */
+    void _assert(const(void)* exp, const(void)* file, uint line);
+}
+else version (CRuntime_Microsoft)
+{
+    /***
+     * Assert failure function in the Microsoft C library.
+     * `_assert` is not in assert.h, but it is in the library.
+     */
+    void _wassert(const(wchar)* exp, const(wchar)* file, uint line);
+    ///
+    void _assert(const(char)* exp, const(char)* file, uint line);
+}
+else version (OSX)
+{
+    /***
+     * Assert failure function in the OSX C library.
+     */
+    void __assert_rtn(const(char)* func, const(char)* file, uint line, const(char)* exp);
+}
+else version (FreeBSD)
+{
+    /***
+     * Assert failure function in the FreeBSD C library.
+     */
+    void __assert(const(char)* exp, const(char)* file, uint line);
+}
+else version (CRuntime_Glibc)
+{
+    /***
+     * Assert failure functions in the GLIBC library.
+     */
+    void __assert(const(char)* exp, const(char)* file, uint line);
+    ///
+    void __assert_fail(const(char)* exp, const(char)* file, uint line, const(char)* func);
+    ///
+    void __assert_perror_fail(int errnum, const(char)* file, uint line, const(char)* func);
+}
+else
+{
+    static assert(0);
+}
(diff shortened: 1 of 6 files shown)


win64.mak
```diff
@@ -17,8 +17,8 @@ IMPDIR=import
 
 MAKE=make
 
-DFLAGS=-m$(MODEL) -conf= -O -release -dip1000 -inline -w -Isrc -Iimport
-UDFLAGS=-m$(MODEL) -conf= -O -release -dip1000 -w -Isrc -Iimport
+DFLAGS=-m$(MODEL) -conf= -O -release -dip1000 -dip1008 -inline -w -Isrc -Iimport
+UDFLAGS=-m$(MODEL) -conf= -O -release -dip1000 -dip1008 -w -Isrc -Iimport
```
rainers: Adding -dip1008 seems unrelated. There are also a couple more unrelated changes to win64.mak that cause build failures.

src/core/stdc/assert_.d
```diff
@@ -0,0 +1,52 @@
...
+ * They are all noreturn functions, although D doesn't have a specific attribute for that.
+ */
+
+version (CRuntime_DigitalMars)
+{
+    void _assert(const(void)* exp, const(void)* file, uint line);
+}
+else version (CRuntime_Microsoft)
+{
+    void _assert(const(char)* exp, const(char)* file, uint line);
```
rainers: There is no _assert in the MS runtime, they use _wassert with wide strings only:
```
void __cdecl _wassert(const wchar_t * _Message, const wchar_t *_File, unsigned _Line);
```

WalterBright: Actually, there is. It just isn't in the .h file.

rainers: I see. A comment would be nice for the next time I stumble over this.

Should we add the _wassert version, too, similar to the various declaration on the other platforms?

rainers:
Not a blocker, but the documentation only lists the glibc version declarations. This can be pretty confusing for other platforms. The ddox documentation doesn't list any function at all.

src/core/stdc/assert_.d
```diff
@@ -0,0 +1,52 @@
...
+ * Source:    $(DRUNTIMESRC core/stdc/_assert_.d)
+ * Standards: ISO/IEC 9899:1999 (E)
+ */
+
+module core.stdc.assert_;
+
+extern (C):
+@trusted:
+nothrow:
+@nogc:
```
ZombineDev: `pure`?

WalterBright: I'm not sure that a function never returning means it is `pure`.

src/core/stdc/assert_.d
WalterBright: yes

src/core/stdc/assert_.d
ZombineDev: As far as I understand, this provides an alternative to built-in `assert`. Since the built-in `assert` is usable from `pure` functions, I think this should also be.

> I'm not sure that a function never returning means it is pure.

That should never happen in correct programs. But in general `pure` functions are allowed to terminate the program, [according to the language spec](http://dlang.org/spec/function.html#pure-functions). So I think the C `assert` functions fit the bill.

rainers: I agree with @ZombineDev, should be `pure`. The betterC purists will come back at you anyway without it.

rainers:
I think this would be more useful with a template `c_assert` or similar that translates the C macro `assert` for the different platforms.

WalterBright:
> I think this would be more useful with a template c_assert or similar that translates the C macro assert for the different platforms.

The purpose of these imports is not to fix the interface, but to simply declare it.

src/core/stdc/assert_.d
WalterBright: Pure functions that return void can be elided by the compiler.

rainers:
> Pure functions that return void can be elided by the compiler.

If these functions are actually elided, then we'll have to go without `pure` for now. I guess we need a solution for this, though. There are already unintuitive hacks in druntime to work around this, e.g. `onOutOfMemoryError`.

rainers:
> The purpose of these imports is not to fix the interface, but to simply declare it.

The C interface is `assert`, though, not `__assert_rtn`. Other core.stdc files translate the macros, too. See https://github.com/dlang/druntime/blob/master/src/core/stdc/stdio.d#L1068 for an example.

WalterBright:
> I guess we need a solution for this, though.

The solution is a `noreturn` function attribute of some sort.

WalterBright:
> The C interface is assert, though, not __assert_rtn.

We also declare the internal functions in the `core.stdc.*` imports.

rainers:
> We also declare the internal functions in the core.stdc.* imports.

Sure, I never said these should be removed. But without an implementation of the C assert macro, everyone will have to implement their own version forwarding to the platform specific functions.

rainers:
> The solution is a noreturn function attribute of some sort.

This will also need a `mightnotreturn` attribute if the termination is conditional.

WalterBright:
> This will also need a mightnotreturn attribute if the termination is conditional.

Every function is mightnotreturn. It's OT anyway.

rainers:
> Every function is mightnotreturn. It's OT anyway.

If that's true you must not elide any pure function if you want to keep the side effect of `noreturn` pure calls.

PR #1842, opened 2017-06-18
fix deprecation messages when compiling typeinfo.d

WalterBright:
```
Deprecation: identity comparison of static arrays implicitly coerces them to slices, which are compared by reference
```


wilzbach:
Seems like the testsuite in Druntime is highly non-deterministic. As mentioned at DMD, there are basically two ways to "fix" the coverage report (i.e. so that it doesn't show unrelated changes).

A) add tests to always cover these lines
B) postprocess the coverage​ output before it's sent to CodeCov and always remove coverage in these lines (we do this for std.concurrency at Phobos)

PR #1843, opened 2017-06-18
osmodel.mak: add comments

WalterBright:
Makefiles need comments, too.

### DIFF
--- a/osmodel.mak
+++ b/osmodel.mak
@@ -1,5 +1,15 @@
-# This Makefile snippet detects the OS and the architecture MODEL
-# Keep this file in sync between druntime, phobos, and dmd repositories!
+#   osmodel.mak
+#
+# Detects and sets the macros:
+#
+#   OS         = one of {osx,linux,freebsd,openbsd,netbsd,solaris}
+#   MODEL      = one of { 32, 64 }
+#   MODEL_FLAG = one of { -m32, -m64 }
+#
+# Note:
+#   Keep this file in sync between druntime, phobos, and dmd repositories!
+# Source: https://github.com/dlang/druntime/blob/master/osmodel.mak
+
 
 ifeq (,$(OS))
   uname_S:=$(shell uname -s)


osmodel.mak
```diff
@@ -1,5 +1,15 @@
...
+#
+# Detects and sets the macros:
+#
+#   OS         = one of {osx,linux,freebsd,openbsd,netbsd,solaris}
+#   MODEL      = one of { 32, 64 }
+#   MODEL_FLAG = one of { -m32, -m64 }
+#
+# Note:
+#   Keep this file in sync between druntime, phobos, and dmd repositories!
+# Source: https://github.com/dlang/druntime/blob/master/osmodel.mak
```
CyberShadow: I don't understand the purpose of this source link.

osmodel.mak
```diff
@@ -1,5 +1,15 @@
...
+#   osmodel.mak
+#
+# Detects and sets the macros:
+#
+#   OS         = one of {osx,linux,freebsd,openbsd,netbsd,solaris}
+#   MODEL      = one of { 32, 64 }
+#   MODEL_FLAG = one of { -m32, -m64 }
+#
+# Note:
+#   Keep this file in sync between druntime, phobos, and dmd repositories!
```
CyberShadow: I guess this needs accompanying PRs to dmd and phobos as per the comment.

osmodel.mak
WalterBright: We have them in all the dmd source code files - it's for the same reason - an easy way to find out things like the history of the file.

osmodel.mak
WalterBright: Yes, but after this one.

andralex:
codecov should not consider this for coverage. Why is jenkins failing? cc @MartinNowak

ZombineDev:
https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdruntime/detail/PR-1843/1/pipeline#step-480-log-330 - looks like configuration / environment issue. We can try restarting the Jenkins by closing & reopening the PR.

andralex:
let's see

osmodel.mak
wilzbach: > I guess this needs accompanying PRs to dmd and phobos as per the comment.

FWIW Phobos and Druntime already depend on DMD and I see no logical reason to have redundant copies of this files...
I actually tried to push for this once:

https://github.com/dlang/phobos/pull/5228

But was stopped by people not wanting to have the DMD Makefile auto-checking out the tools repo.
However, we could go with the small scale solution and at least get rid of the osmodel files at Phobos, Druntime ...
Dlang.org was controversial as well due to automatic git clone

andralex: I wonder if there's a way to fix this from the github side - is it possible to link a file across projects?

CyberShadow: > I wonder if there's a way to fix this from the github side - is it possible to link a file across projects?

Yes, it's called submodules, and they come with their share of problems. (You need to remember to do `git submodule update` any time you move HEAD.)

osmodel.mak
wilzbach: `git log osmodel.mak`, no?

PR #1845, opened 2017-06-19
Tweak object.d docs

ntrel:
* Show `destroy(T)` and `destroy(ref T)` instead of only showing `destroy` with class constraint.
* Minor doc tweaks.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2730,18 +2730,12 @@ unittest
     does is done and so that it no longer references any other objects. It does
     $(I not) initiate a GC cycle or free any GC memory.
   +/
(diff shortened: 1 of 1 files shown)


CyberShadow:
> Show destroy(T) and destroy(ref T) instead of only showing destroy with class constraint.

Can `ditto` not be used there?

CyberShadow:
> Should I expose all 5 overloads with ditto instead?

Not sure; paging @andralex

andralex:
Yes please, all overloads under the same doc would be great. Thanks!

ntrel:
Done.

PR #1848, opened 2017-06-20
Add sd_notify binding to core/sys/linux/systemd/daemon.d

nemanja-boric-sociomantic:
This adds family of functions used from a service to notify systemd
of a startup completion and other service status changes.


nemanja-boric-sociomantic:
Closing since these bindings are not of the kernel interface (not yet ;-) ), so we believe they don't belong here.

PR #1821, opened 2017-05-06
Fixup fiber guard page changelog

klickverbot:
During the review of GitHub #1698, the "END OF FIBRE" marker data
was removed to avoid dirtying the guard page. Hence, the corresponding
comment in the release notes doesn't apply anymore.

PR #1849, opened 2017-06-22
Reset pending changelog after 2.075.0

wilzbach:
We should probably include this into the release pipeline ...
CC @MartinNowak


nemanja-boric-sociomantic:
This needs to be pulled in: https://github.com/dlang/druntime/pull/1821

MartinNowak:
Right now I've done it on stable after a release, then purged master when merging back.
https://gist.github.com/MartinNowak/a471fe7ddbfeef205cdf04c93a94c6d0#file-release-sh-L153
That way we only purge things after they've actually been released. Should be fine to have those changelog files still in master for a week to two, no?

MartinNowak:
Let's stick with that scheme for the moment. Purging changelogs after sth. has been released seems more sound and might turn out necessary if we start to backport fixes at some point.

CyberShadow:
@MartinNowak The problem with the current approach is that once you create the _pre.dd changelog file on dlang.org master, it becomes impossible to preview any new changelog entries until the pending version is released. Would it be possible to adjust the release workflow to accommodate this?

PR #1851, opened 2017-06-24
Last tweaks for Android

joakim-noah:
Tested with @kinke's WIP ldc branch merge-2.074, these are the last remaining changes for a fully working druntime on Android.

### DIFF
--- a/src/core/sync/mutex.d
+++ b/src/core/sync/mutex.d
@@ -372,6 +372,9 @@ unittest
     // by checking that locking is not possible. This assumes
     // that the underlying implementation is well behaved
     // and makes the object non-lockable upon destruction.
+    // For example, Bionic doesn't appear to do so, so this test is
+    // not run on Android.
+    version (CRuntime_Bionic) {} else
     assert(!mtx.tryLock_nothrow());
 
     free(cast(void*) mtx);
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -1972,6 +1972,10 @@ extern (C) void thread_init()
     //       functions to detect the condition and return immediately.
 
     Thread.initLocks();
+    // The Android VM runtime intercepts SIGUSR1 and apparently doesn't allow
+    // its signal handler to run, so swap the two signals on Android, since
+    // thread_resumeHandler does nothing.
+    version( Android ) thread_setGCSignals(SIGUSR2, SIGUSR1);
 
     version( Darwin )
     {
--- a/src/rt/sections_android.d
+++ b/src/rt/sections_android.d
@@ -67,7 +67,7 @@ void initSections() nothrow @nogc
     auto mend = cast(immutable ModuleInfo**)&__stop_minfo;
     _sections.moduleGroup = ModuleGroup(mbeg[0 .. mend - mbeg]);
 
-    auto pbeg = cast(void*)&_tls_end;
+    auto pbeg = cast(void*)&_tlsend;
     auto pend = cast(void*)&__bss_end__;
     _sections._gcRanges[0] = pbeg[0 .. pend - pbeg];
 }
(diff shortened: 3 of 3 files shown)


src/core/sync/mutex.d
```diff
@@ -372,6 +372,9 @@ unittest
     // by checking that locking is not possible. This assumes
     // that the underlying implementation is well behaved
     // and makes the object non-lockable upon destruction.
+    // For example, Bionic doesn't appear to do so, so this test is
+    // not run on Android.
+    version (CRuntime_Bionic) {} else
```
joakim-noah: This test asserts on Android 6.0 Marshmallow, as [it doesn't appear that `pthread_mutex_destroy` really destroys the mutex there](https://android.googlesource.com/platform/bionic/+/marshmallow-release/libc/bionic/pthread_mutex.cpp#639).  I notice that if I comment out the line that sets `PTHREAD_MUTEX_RECURSIVE` in this druntime module, the test passes again, but I haven't dug into why it only doesn't work for recursive mutexes.  It doesn't, so this test can't be run as is.

src/core/thread.d
```diff
@@ -1972,6 +1972,10 @@ extern (C) void thread_init()
     //       functions to detect the condition and return immediately.
 
     Thread.initLocks();
+    // The Android VM runtime intercepts SIGUSR1 and apparently doesn't allow
+    // its signal handler to run, so swap the two signals on Android, since
+    // thread_resumeHandler does nothing.
+    version( Android ) thread_setGCSignals(SIGUSR2, SIGUSR1);
```
joakim-noah: The Android VM runtime [intercepts `SIGUSR1` in its signal-catcher thread](https://android.googlesource.com/platform/art/+/master/runtime/signal_catcher.cc#247) and [apparently doesn't run its signal-handler](https://android.googlesource.com/platform/art/+/master/runtime/signal_catcher.cc#214).  This doesn't matter when D executables are run as standalone processes on the Android command-line, but it causes problems when threads are suspended in an Android app that runs D code in a shared library called from the VM runtime, as that's how native code normally has to be run on Android.

When the unit tests for `core.thread` are run in an Android app without this change, they hang at [`pthread_kill` when it tries to suspend a thread](https://github.com/dlang/druntime/blob/master/src/core/thread.d#L2600).  My guess is this is because the VM runtime intercepts `SIGUSR1` and [the thread_suspendHandler](https://github.com/dlang/druntime/blob/master/src/core/thread.d#L431) is never run.  If I swap the two signals, it still resumes fine, presumably because [the thread_resumeHandler does nothing](https://github.com/dlang/druntime/blob/master/src/core/thread.d#L488), so it doesn't matter if it doesn't run.

I've been using Ilya's #1565 also, which works fine.  However, that pull has been stuck in limbo for more than a year, so I'd like to get this in in the meantime.  Whenever that pull is approved, this can be removed.

src/rt/sections_android.d
```diff
@@ -63,11 +63,11 @@ void initSections() nothrow @nogc
 {
     pthread_key_create(&_tlsKey, null);
 
-    auto mbeg = cast(immutable ModuleInfo**)&__start_minfo;
-    auto mend = cast(immutable ModuleInfo**)&__stop_minfo;
+    auto mbeg = cast(immutable ModuleInfo**)&__start___minfo;
+    auto mend = cast(immutable ModuleInfo**)&__stop___minfo;
```
joakim-noah: I don't know where these magic linker symbols come from, a `grep` of the Android NDK for them turns up nothing.  But these are the symbols ldc uses and the extra underscores are needed for this to actually work with the NDK's linker.

JohanEngelen: This may help: https://www.airs.com/blog/archives/56
We set the section name to `"__minfo"`, so that's where the extra underscores come from: https://github.com/ldc-developers/ldc/blob/0b8d09d517ce2155496bf9e8f39ae57eff0b6ad0/gen/modules.cpp#L375

joakim-noah: Thanks for the link.  I thought a `grep` of the linker scripts would turn them up, but that's obviously not going to work for a section name we make up.

joakim-noah:
Since Johan pointed out that the different linker symbols are because ldc names its moduleinfo section differently, took out that change and will just submit it separately to ldc's druntime.

src/core/sync/mutex.d
ZombineDev: As the author of this test (and the test below) I'm now wondering if it might have been a mistake to rely on implementation defined behavior here (actually undefined behavior according to the [Posix standard](http://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_mutex_destroy.html)). I assumed that [`pthread_mutex_trylock`](http://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_mutex_trylock.html) would return `EINVAL` on all platforms if called with a previously destroyed mutex (otherwise what would be the point of specifying such return value in the standard?).

I would love to find a better way to do this, but I can't think of any other alternatives at the moment.

PR #1144, opened 2015-02-01
[WIP] Optimize GC bit handling for small object pools

rainers:
This continues work of https://github.com/D-Programming-Language/druntime/pull/1127

The idea is to extract the actual bit handling into the different type of pools, so the small object pool can have a very different approach than the large object pool.

The actual change to the bit layout is in commit https://github.com/rainers/druntime/commit/42c133134974f2edd9a2e6ec4604624e6366f44a where the bits are stored in the same page as the objects they describe. For simplicity they are stored as one byte per object, saving space for objects larger than 16 bytes due to the removal of bit strides.

Unfortunately, there is no noticable performance improvement from this change yet. Intermediate version had a small loss caused by maintaining freebits instead of rebuilding them before a collection.

PR #1854, opened 2017-06-30
GC hashed pooltable

DmitryOlshansky:
This turned out not to be the low hanging fruit I was looking for but I believe a smallish improvement of around 10% in pool lookup itself. To that end I added a benchmark that does 25M GC.query to validate that. In dedicated pool lookup benchmark I get measly 4% less total time spent.

Full data is here:
https://1drv.ms/x/s!Ag1cr_x9Z667grMBZG3MnwCRLLYOkg

For my own sanity I did 2 runs with Old GC which highlights that words and conappend are unreliable. words spents 4-5ms in GC anyway, what kind of benchmark is that?

### DIFF
--- a/src/rt/util/container/hashtab.d
+++ b/src/rt/util/container/hashtab.d
@@ -328,3 +328,183 @@ unittest
     assert(thrown);
     assert(tab[3] == 3);
 }
+
+/// An open-addressing hash table with quadratic probing
+/// roughly follows the description of Google's dense_hash_map.
+/// WARNING: supports only POD types and no removal by key for now.
+struct FlatHashTab(K, V, alias nullValue, alias hashFunc=hashOf)
+{
+nothrow:
+    this(size_t capacity)
+    {
+        import core.bitop : bsr;
+        if(!capacity)
+            capacity = 32;
+        else
+            capacity = 1<<bsr(capacity);
+        table.length = capacity;
+    }
+
+    @disable this(this);
+
+    ~this()
+    {
+        if(table.length) common.free(table.ptr);
+    }
+
+    void reset()
+    {
+        table[] = Entry.init;
+        items = 0;
+    }
+
+    @property size_t length() const
+    {
+        return items;
+    }
+
+    @property bool empty() const
+    {
+        return items == 0;
+    }
+
+    void opIndexAssign(V value, K key)
+    in
+    {
+        assert(value != nullValue);
+    }
+    body
+    {
+        if(items * 2 == table.length) rehash();
+        size_t h = hashFunc(key);
+        size_t mask = (table.length-1);
+        size_t i = h & mask;
+        if (table[i].value == nullValue)
+        {
+            table[i] = Entry(key, value);
+            items++;
+            return;
+        }
+        if (table[i].key == key)
+        {
+            table[i].value = value;
+            return;
+        }
+        for (size_t n = 1; ; n++)
+        {
+            size_t j = (i + triangular(n)) & mask;
+            if (table[j].value == nullValue)
+            {
+                table[j] = Entry(key, value);
+                items++;
+                return;
+            }
+            if (table[j].key == key)
+            {
+                table[j].value = value;
+                return;
+            }
+        }
+    }
+
+    inout(V) opIndex(K key) inout
+    in
+    {
+        assert((table.length & (table.length-1)) == 0);
+    }
+    body
+    {
+        size_t h = hashFunc(key);
+        size_t mask = table.length - 1;
+        size_t i = h & mask;
+        return table[i].key == key ? table[i].value : 
+            (table[i].value == nullValue ? nullValue : slowLookup(key, i));
+        
+    }
+
+    inout(V) slowLookup(K key, size_t i) inout
+    {
+        size_t mask = table.length - 1;
+        for (size_t n = 1; ; n++)
+        {
+            size_t j = (i + triangular(n)) & mask;
+            if (table[j].value == nullValue) return nullValue;
+            if (table[j].key == key) return table[j].value;
+        }
+    }
+
+    int opApply(scope int delegate(ref K, ref V) nothrow dg)
+    {
+        foreach (e; table)
+        {
+            if (e.value != nullValue)
+            {
+                if(auto res = dg(e.key, e.value))
+                    return res;
+            }
+        }
+        return 0;
+    }
+
+private:
+    struct Entry
+    {
+        K key;
+        V value = nullValue;
+    }
+
+    Entry[] table;
+    size_t items;
+
+    static size_t triangular(size_t n){ return n * (n + 1) / 2; }
+
+    void rehash()
+    {
+        import core.stdc.stdio;
+        Entry[] old = table;
+        size_t newSize = items == 0 ? 32 : table.length * 2;
+        table = (cast(Entry*)common.xmalloc(Entry.sizeof * newSize))[0..newSize];
+        table[] = Entry.init;
+        items = 0;
+        foreach(e; old)
+        {
+            if(e.value != nullValue) this[e.key] = e.value;
+        }
+        if(old.length) common.free(old.ptr);
+    }
+}
+
+unittest
+{
+    import core.stdc.stdio;
+    FlatHashTab!(ushort, ushort, ushort.max, x => x) htab;
+    // Make collisions
+    htab[1] = 1;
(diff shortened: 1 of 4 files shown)


rainers:
I've run the benchmarks on my desktop system (Win32 and Win64, i7 4790K) but they didn't show a notable difference above the error delta (+/- 1-2% for most benchmarks). I profiled the rand_small benchmark which showed that the mark phase takes just about 6% of the overall time.
With so little pools to search, the method of lookup doesn't really seem to matter. Even linear search was hardly measurable in the benchmarks.

Please note that the results vary quite a bit with the platform, so please specify your system coordinates. Win64 and Win32 produce very different results for some tests, for example. Optimzations on AMD processors might produce not so great results on Intel (see e.g. https://github.com/dlang/druntime/pull/1603#issuecomment-241331882). 

The dmd optimizer can lead to a lot of frustrations, so LDC should also be used for benchmarks.

> words spents 4-5ms in GC anyway, what kind of benchmark is that?

GC takes about 15% of the overall time for me, not very different from other tests.

BTW: during profiling, I noticed that the GC lock seems very expensive on Win64 (took about 30% of overall time) and a lot of the remaining time was spent with bit fiddling. Some time ago I tried to reorganize that, but without much effect on performance: https://github.com/dlang/druntime/pull/1144

DmitryOlshansky:
> With so little pools to search, the method of lookup doesn't really seem to matter. Even linear search was hardly measurable in the benchmarks.

I guess this is it. Well for my new GC the number of pools will be way bigger as I plan to segregate by size class. Closing this.

MartinNowak:
I'd recommend to stay away from the current GC. We spent endless hours on micro-optimizations and there isn't much to find except for maybe performance bugs on some systems.

> I noticed that the GC lock seems very expensive on Win64 (took about 30% of overall time)

How comes, we did replace it with a spinlock, should be platform agnostic. How did you measure? Should take quite a lot of contention to run into a `Thread.yield` call https://github.com/dlang/druntime/blob/68cce532a2ce7954528f801eebd23cc51b410efd/src/core/internal/spinlock.d#L56.

src/rt/util/container/hashtab.d
```diff
@@ -328,3 +328,183 @@ unittest
...
+            table[i] = Entry(key, value);
+            items++;
+            return;
+        }
+        if (table[i].key == key)
+        {
+            table[i].value = value;
+            return;
+        }
+        for (size_t n = 1; ; n++)
```
MartinNowak: [Here](https://github.com/dlang/druntime/blob/68cce532a2ce7954528f801eebd23cc51b410efd/src/rt/aaA.d#L104) is how you loop over triangular numbers.
```d
for (size_t i = h & mask, step = 1;; i = (i + step++) & mask)
{
}
```
No stop condition needed as it will visit every slot.

MartinNowak:
@DmitryOlshansky

I still think that allocating smaller pools (or pages associated to spans for metadata) instead of say 64KB pools will be the better approach. You'd have to lookup metadata with a radix tree to make this feasible.
In order to directly use a hashtable that remains cashable, pool granularity needs to be rather big, e.g. 64KB which can lead to lots of internal fragmentation.

Quick back of the envelope calculation.
hashmap
1GB heap / 64KB = 16384 entries = 16384 * (4 bytes hash + 4 bytes pool index) = 512KB L3 cache :/
radix tree with 6 bits per level (assuming upper bits are equal, i.e. contiguous 1GB):
1GB heap / 4KB = 16384 entries = (up to 64 - 12) bits / 6 bits / level * 512 byte / level + (4 parent entries * 64 leave nodes) * 512 byte / node = 132.5KB L3 cache

BTW, a page is just a pool with 4KB size ;).

The main reason why you don't want big granularity of pools, in particular when using them strictly for one size class, is that just a single alive object can keep the whole pool alive. Without compaction a 32 byte allocation can consume 64KB. In particular when the memory usage of a program varies a lot this might lead to inflated mem usage.

With smaller granularity pools (e.g. 4KB pages) you can just return them to the page/pool allocator, which can sort them in and merge adjoined pages/pools. Those can be returned to the OS or be used to serve further allocation requests.

DmitryOlshansky:
> The main reason why you don't want big granularity of pools, in particular when using them strictly for one size class, is that just a single alive object can keep the whole pool alive. Without compaction a 32 byte allocation can consume 64KB. In particular when the memory usage of a program varies a lot this might lead to inflated mem usage.

I belive the correct way is to keep mapping around but madvise the physical pages away. This way you don't need to worry about many pool being underutilized.

> Quick back of the envelope calculation.
> hashmap
> 1GB heap / 64KB = 16384 entries = 16384 * (4 bytes hash + 4 bytes pool index) = 512KB L3 cache :/
> radix tree with 6 bits per level (assuming upper bits are equal, i.e. contiguous 1GB):
> 1GB heap / 4KB = 16384 entries = (up to 64 - 12) bits / 6 bits / level * 512 byte / level + (4 parent entries * 64 leave nodes) * 512 byte / node = 132.5KB L3 cache

Yeah, but radix tree is dog slow.

src/rt/util/container/hashtab.d
DmitryOlshansky: Bookmarked

WalterBright:
@DmitryOlshansky keep in mind that as data grows into the multiple gigabytes, an optimization that has little effect on a < 1G data set may have a large effect for > 1G.

(The GC's algorithms are little changed from when it was designed, which was when memory was much, much smaller.)

MartinNowak:
> @DmitryOlshansky keep in mind that as data grows into the multiple gigabytes, an optimization that has little effect on a < 1G data set may have a large effect for > 1G.

Don't think it's possible to scale a conservative (or precise) GC to multi GB heaps. Squeezing all of that through limited memory bandwidth already incurs too big latencies (30ms / GB at todays 34.1GB / s memory bandwidth). To support multi-GB a form of partial collection is needed (generational or [CBGC](https://www.cs.purdue.edu/homes/hosking/690M/cbgc.pdf)).
A useful GC for D would be a good sidekick to complement manual memory management for more demanding applications and bigger heap requirements.


> I belive the correct way is to keep mapping around but madvise the physical pages away.

Believe and correct is a problematic mixture. Make sure to run some tests to figure out whether the OS can deal well with a swiss cheese mappings, in the end they got to do the bookkeeping for you and also just use data structures. Does madvise reduce the shown RSS?
Sounds like it could cause some problems with huge pages, which are a useful optimization to many programs (to reduce TLB misses) and likely keep becoming more prominent.

MartinNowak:
Just tested, madvise works, also nicely for **transparent** huge pages (kernel 4.11.8-200.fc25.x86_64).
For explicitly mapped `MAP_HUGETLB` madvise `MADV_DONTNEED` doesn't seem to work at all though.

```d
import core.thread, core.stdc.stdlib, core.sys.linux.sys.mman;

enum PAGE_SIZE = 4096;
enum HUGE_PAGE_SIZE = 2048 * 1024;
enum MADV_HUGEPAGE = 14;

void main()
{
    enum size = 2UL ^^ 32;
    enum stride = PAGE_SIZE;
    auto mem = (cast(ubyte*) .mmap(null, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANON, -1, 0))[0 .. size];
    scope (exit) .munmap(mem.ptr, mem.length);
    // use transparent huge pages to reduce TLB trashing
    .madvise(mem.ptr, mem.length, MADV_HUGEPAGE);
    foreach (u; 0 .. size / stride)
    {
        mem[u * stride] = 0;
        if (u % (size / stride / 2048) == 0)
            Thread.sleep(5.seconds / 2048);
    }
    foreach (u; 0 .. size / stride)
    {
        if (u % 2 == 0) // works great, kernel converts huge pages to normal pages if we peek holes into them
            .madvise(&mem[u * stride], stride, MADV_DONTNEED);
        if (u % (size / stride / 2048) == 0)
            Thread.sleep(5.seconds / 2048);
    }
}
```
Use `watch -d 0.5 grep -Fi hugepages /proc/meminfo` to monitor usage of AnonHugePages.

I found [Transparent Huge Pages and Alternative Memory Allocators: A Cautionary Tale](https://blog.digitalocean.com/transparent-huge-pages-and-alternative-memory-allocators/) which could apply to an older kernel. Definitely lots of dbs still recommend to disable transparent huge pages.
We could at least make it a GC option, whether it defaults to true or not is a different question.

PR #1855, opened 2017-07-03
Remove osmodel.mak in favor of ../dmd/osmodel.mak

wilzbach:
A small step at cleaning up the Makefiles.

Similarly to https://github.com/dlang/tools/pull/227


circle.yml
```diff
@@ -6,8 +6,8 @@ dependencies:
 
 test:
   override:
-    - make -f posix.mak style
     - ./circleci.sh setup-repos
+    - make -f posix.mak style
```
andralex: Is this related?

wilzbach: Yes, DMD is now needed before other Makefile tasks can be run.

PR #1857, opened 2017-07-06
Fix leaking main thread held memory

Calrama:
Solves the problem described in https://github.com/dlang/druntime/pull/1557 (which got stalled)

- Release such memory explicitly by `destroy`ing
  the main thread on threading module termination
- Prior to this PR `rt.sections_elf_shared.finiTLSRanges` already contains the necessary logic to `reset` `_tlsRanges` to zero length in the not `Shared` version, it just isn't used; instead the `_tlsRanges` (which only have one element) are cleared by calling `_tlsRanges.popBack` in `_d_dso_registry`.
  This PR removes that `popBack` call and instead uses the preexisting logic in `finiTLSRanges` via
  core.thread.thread_term -> ~Thread on the main thread -> rt.tlsgc.destroy -> rt.sections.finiTLSRanges

### DIFF
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -461,8 +461,6 @@ extern(C) void _d_dso_registry(CompilerDSOData* data)
         else
         {
             // static DSOs are unloaded in reverse order
-            assert(pdso._tlsSize == _tlsRanges.back.length);
-            _tlsRanges.popBack();
             assert(pdso == _loadedDSOs.back);
             _loadedDSOs.popBack();
         }
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -2043,7 +2043,7 @@ extern (C) void thread_init()
  */
 extern (C) void thread_term()
 {
-    try destroy(Thread.sm_main); catch (Exception) {}
+    destroy(Thread.sm_main);
     Thread.sm_main = null;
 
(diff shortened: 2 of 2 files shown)


src/rt/sections_elf_shared.d
```diff
@@ -461,8 +461,6 @@ extern(C) void _d_dso_registry(CompilerDSOData* data)
         else
         {
             // static DSOs are unloaded in reverse order
-            assert(pdso._tlsSize == _tlsRanges.back.length);
-            _tlsRanges.popBack();
```
nemanja-boric-sociomantic: Just to check - this is called once per the executable, not once per thread, right? Meaning, `_tlsRanges` will have just one element.

Calrama: Before this PR, yes. With this PR it will have no elements at that time (and the `.popBack` will thus result in an `AssertError`, which is why I removed it), because it was already `.clear`ed via `thread_term` (as per PR description).

nemanja-boric-sociomantic: Yes, just wanted to confirm that!

Burgos:
Just FYI, since I saw this mentioned on the newsgroup :). I'm back from a holiday next week, and will give this a more detailed look then, if nobody does by then. It looked fine to me last time I checked.

ZombineDev:
@Burgos, do you have time to give this PR a more detailed look?

PR #1858, opened 2017-07-06
core.atomic: avoid comparing nan bit patterns

WalterBright:
Different NAN patterns can be generated by various code paths, and comparing the bit patterns is a mistake. In this case, a signalling nan is compared against a quiet nan.

Use `0` instead.

This is blocking https://github.com/dlang/dmd/pull/6974

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -1304,8 +1304,8 @@ version( unittest )
     }
     body
     {
-        T         base;
-        shared(T) atom;
+        T         base = cast(T)null;
+        shared(T) atom = cast(shared(T))null;
 
         assert( base !is val, T.stringof );
         assert( atom is base, T.stringof );


ibuclaw:
Looks fine, but I'll just double check something first. It's not immediately obvious why the signalling bit would differ.

WalterBright:
Every time the signalling bit goes through the x87, it gets turned off. In one path in the code, the initialization is done through the x87. In the other path, it is bit compared with the initializer. Hence the difference. (You can see this if you disassemble the code.)

ibuclaw:
Yeah, I was wondering why this wasn't triggered in gdc.  In any case, there's no problem with this on its own.  However it seems like you've introduced a regression in the compiler though if this triggers.

ibuclaw:
Unfortunately, this caused a regression in GDC.

WalterBright:
I'm not terribly surprised, having signalling nan support in D seemed like a good idea, but it just doesn't work, and weird inconsistent bugs will result. It's best to purge GDC of such dependencies, too.

ibuclaw:
OK.

I also found another unhandled case.  `auto f = cast(float)null` was being treated as a cast from `void*` to `float`.  I guess where `typeof(null)` is the from type, the right thing to do is generate whatever represents zero of the destination type?

WalterBright:
I'd treat that as a float with all bits 0. Though I think casting a null to float would require Lucy to do some 'splainin'.

ibuclaw:
Yes, Lucy, you do have some 'splainin' to do.  I guess the following would have the same desired effect without the null to float cast.

```
T         base = cast(T) 0;
shared(T) atom = cast(shared(T)) 0;
```

PR #1859, opened 2017-07-06
core.memory: Fix typo (useable -> usable)

CyberShadow:


CyberShadow:
Err, maybe not a typo. https://english.stackexchange.com/questions/69709/is-useable-preferred-in-certain-regions-or-just-an-alternate-spelling-of-usa

WalterBright:
"usable" seems more common, so let's do it.

PR #1860, opened 2017-07-07
atomicOp!(double) unittests depend on has64BitCAS

ibuclaw:
Should be self-explanatory.


WalterBright:
@ibuclaw since you're up to speed on this section of code, how about reviewing https://github.com/dlang/druntime/pull/1858 ? It's blocking my other work, and it's a trivial fix.

ibuclaw:
Two independent approvals, so I'm adding auto-merge.

PR #1862, opened 2017-07-08
Fis issue 16856: Don't use dlopen from the fini sections

Burgos:
In case finalizers are called from the runtime linker
we shouldn't try to get handle to and reference the
dying shared object. This was used just for the asserting,
so this simply removes the assert making it working on the
platforms where this is strictly forbiden (e.g. FreeBSD 12).

### DIFF
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -454,7 +454,6 @@ extern(C) void _d_dso_registry(CompilerDSOData* data)
                 }
             }
 
-            assert(pdso._handle == handleForAddr(data._slot));
             unsetDSOForHandle(pdso, pdso._handle);
             pdso._handle = null;
         }


Burgos:
CC @MartinNowak 

@jmdavis with this I get druntime passing the tests, and phobos' tests as well (well, not all, they are still running, but it looks promising :-) )

MartinNowak:
They normally explicity unlock/relock around calling init/fini, cause user code can load/unload other libraries.
https://github.com/freebsd/freebsd/blob/683404221f539272641cfea4d2c20f4c039b7915/libexec/rtld-elf/rtld.c#L2544
https://github.com/freebsd/freebsd/blob/683404221f539272641cfea4d2c20f4c039b7915/libexec/rtld-elf/rtld.c#L2574-L2578
But indeed looks like it's forbidden https://github.com/freebsd/freebsd/commit/3ff2e66ecba2094f5c1c1efe7f2d009649527195 and kind of make sense, you don't want half a finalized library to linger around. Thankfully that only a problem within the fini calls.

aG0aep6G:
You've got a typo in the commit message: "Fi**s** issue 16856". Should be "Fi**x**", of course.

Burgos:
Thanks @aG0aep6G ! Updated the commit message (which sadly removed `auto-merge` flag).

wilzbach:
> Updated the commit message (which sadly removed auto-merge flag).

Yeah, that's a feature - not a bug. Happy to reapply though ;-)

nemanja-boric-sociomantic:
Haha, yes :-)

Thanks, @wilzbach !

PR #1863, opened 2017-07-09
Fix issue 17375: Use OPTIONAL_PIC in the test suite

Burgos:
Test suite didn't compile the executable with fPIC or with -cov no
matter if that was set in the top makefile (since it is a) overriding
DFLAGS, and b) because OPTIONAL_PIC was not exported from the main make
file to sub make files).

This makes test suite compile executables with the fPIC if requested
with PIC. Incidentally, this fixes a problem compiling shared test suite
on arch linux with binutils greater than v2.28, since the executable
itself will be compiled with -fPIC, not just shared library.

N.B.: the -cov flag is present, but not exported and used, because
some shared library tests are segfaulting when compiled with -cov.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -32,6 +32,8 @@ IMPDIR=import
 OPTIONAL_PIC:=$(if $(PIC),-fPIC,)
 OPTIONAL_COVERAGE:=$(if $(TEST_COVERAGE),-cov,)
 
+export OPTIONAL_PIC
+
 ifeq (osx,$(OS))
 	DOTDLL:=.dylib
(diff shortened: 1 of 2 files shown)


Burgos:
Edit: removed "Fix issue" annotation, since it just worksaround it so that testsuite may work.

Burgos:
Added `-dip1000` flag to test-suite while I was there as well.

Burgos:
Bah, now I found out #1721 which is staled from December. :grumpy_cat:. Closing as duplicate, will reopen a new one with the `-dip1000` now.

PR #1864, opened 2017-07-09
Add missing errno codes in existing linux platforms.

ibuclaw:
Just so happened to notice that EDEADLOCK was wrong on PPC, so went through every CPU header that GDC builds for and spot checked each one just incase, and found a few more (`asm/errno.h` and overrides in `bits/errno.h`).

Added/Modified:
- x86: `ERFKILL(132)`, `EWPOISON(133)`.
- x86_64: `ERFKILL(132)`, `EWPOISON(133)`.
- ARM: `ERFKILL(132)`, `EWPOISON(133)`.
- AArch64: `ENOTSUP(EOPNOTSUPP)`, `ENOTRECOVERABLE(131)`.
- MIPS64: `ENOTSUP(EOPNOTSUPP)`.
- PPC: `EDEADLOCK(58)`, `ENOTSUP(EOPNOTSUPP)`, `ENOTRECOVERABLE(131)`.
- PPC64: `EDEADLOCK(58)`, `ENOTSUP(EOPNOTSUPP)`, `ENOTRECOVERABLE(131)`.
- SystemZ: `ENOTSUP(EOPNOTSUPP)`, `ENOTRECOVERABLE(131)`.

Don't have RISCV headers to hand, so did not verify them.

### DIFF
--- a/src/core/stdc/errno.d
+++ b/src/core/stdc/errno.d
@@ -216,6 +216,8 @@ else version( linux )
         enum EKEYREJECTED       = 129;        ///
         enum EOWNERDEAD         = 130;        ///
         enum ENOTRECOVERABLE    = 131;        ///
+        enum ERFKILL            = 132;        ///
+        enum EHWPOISON          = 133;        ///
     }
     else version(X86_64)
     {
@@ -317,6 +319,8 @@ else version( linux )
         enum EKEYREJECTED       = 129;        ///
         enum EOWNERDEAD         = 130;        ///
         enum ENOTRECOVERABLE    = 131;        ///
+        enum ERFKILL            = 132;        ///
+        enum EHWPOISON          = 133;        ///
     }
     else version(ARM)
     {
@@ -418,6 +422,8 @@ else version( linux )
         enum EKEYREJECTED       = 129;        ///
         enum EOWNERDEAD         = 130;        ///
         enum ENOTRECOVERABLE    = 131;        ///
+        enum ERFKILL            = 132;        ///
+        enum EHWPOISON          = 133;        ///
     }
     else version(AArch64)
     {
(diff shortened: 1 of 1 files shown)


MartinNowak:
Looks like they use the [generic errno.h](https://github.com/riscv/riscv-glibc/blob/ce8a6c47923b461f09167bc312ca8cbecada083f/sysdeps/unix/sysv/linux/bits/errno.h). I'd be in favor of restructuring these headers to match glibc's generic linux parts as it's a lot more maintainable, see [mman.d](https://github.com/dlang/druntime/blob/3e10db56668477c6d746390686c48be8d7131a70/src/core/sys/linux/sys/mman.d) for how this would look like.

ibuclaw:
> Looks like they use the generic errno.h. I'd be in favor of restructuring these headers to match glibc's generic linux parts as it's a lot more maintainable, see mman.d for how this would look like.

You probably want to remove a layer or two however, this seems to be the spaghetti layout of things in Linux.
```
errno.h
-> #include bits/errno.h
   Define codes that older linux's don't have.
 -> #include linux/errno.h
  -> #include asm/errno.h
   ...either...
      Define arch-specific codes.
   -> #include asm-generic/errno-base.h
       Define posix codes common to everyone.
   ...or...
   -> #include asm-generic/errno.h
       Define codes common to all architectures that aren't backwards compatible with some legacy OS.
    -> #include asm-generic/errno-bash.h
        Define posix codes common to everyone.
```

ibuclaw:
> Looks like they use the generic errno.h. 

The right place to look would be [here](https://github.com/riscv/riscv-linux/tree/riscv-next/arch/riscv/include/uapi/asm). I see no errno.h, so yes you are right.  However I suspect that the only archs that deviate are the ones older than linux itself. ;-)

PR #1865, opened 2017-07-09
Add support for MIPS32 and SPARC64 architectures in stdc/errno.d

ibuclaw:
MIPS32 is identical to MIPS64.

SPARC64 claims to be mostly identical with SunOS in the lower digits, but interestingly shares nothing in common with Solaris.

This is part of minimal work to get the D2 testsuite passing on all platforms in our CI.  More complete druntime support will be later.


src/core/stdc/errno.d
```diff
@@ -1030,6 +1135,111 @@ else version( linux )
         enum ERFKILL            = 132;        ///
         enum EHWPOISON          = 133;        ///
     }
+    else version(SPARC64)
+    {
+        enum EWOULDBLOCK        = EAGAIN,     ///
```
ibuclaw: Whoops, added commas instead of semicolons.

PR #1868, opened 2017-07-09
Add missing -dip1000 flag to test-suite DFLAGs

Burgos:


Burgos:
Part of #1863 which complements #1721 .

PR #1869, opened 2017-07-10
Add support for SPARC64 in sys/posix

ibuclaw:
This should get all compilable tests in the testsuite to pass gdc's CI.

### DIFF
--- a/src/core/sys/posix/sys/stat.d
+++ b/src/core/sys/posix/sys/stat.d
@@ -534,6 +534,83 @@ version( CRuntime_Glibc )
         else
             static assert(stat_t.sizeof == 128);
     }
+    else version (SPARC64)
+    {
+        private
+        {
+            alias __dev_t = ulong;
+            alias __ino_t = c_ulong;
+            alias __ino64_t = ulong;
+            alias __mode_t = uint;
+            alias __nlink_t = uint;
+            alias __uid_t = uint;
+            alias __gid_t = uint;
+            alias __off_t = c_long;
+            alias __off64_t = long;
+            alias __blksize_t = c_long;
+            alias __blkcnt_t = c_long;
+            alias __blkcnt64_t = long;
+            alias __timespec = timespec;
+            alias __time_t = time_t;
+        }
+        struct stat_t
+        {
+            __dev_t st_dev;
+            ushort __pad1;
+            __ino_t st_ino;
+            __mode_t st_mode;
+            __nlink_t st_nlink;
+            __uid_t st_uid;
+            __gid_t st_gid;
+            __dev_t st_rdev;
+            ushort __pad2;
+
+            static if(!__USE_FILE_OFFSET64)
+            {
+                __off_t st_size;
+            }
+            else
+            {
+                __off64_t st_size;
+            }
+            __blksize_t st_blksize;
+
+            static if(!__USE_FILE_OFFSET64)
+            {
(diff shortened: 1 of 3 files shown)


src/core/sys/posix/sys/stat.d
```diff
@@ -534,6 +534,83 @@ version( CRuntime_Glibc )
...
+            alias __blkcnt_t = c_long;
+            alias __blkcnt64_t = long;
+            alias __timespec = timespec;
+            alias __time_t = time_t;
+        }
+        struct stat_t
+        {
+            __dev_t st_dev;
+            ushort __pad1;
+            __ino_t st_ino;
```
Calrama: Looking at glibc 2.23 I see
```c
#if __WORDSIZE == 64 || !defined __USE_FILE_OFFSET64
    unsigned short int __pad1;
    __ino_t st_ino;			/* File serial number.	*/
#else
    __ino64_t st_ino;			/* File serial number.	*/
#endif
```
in `sysdeps/unix/sysv/linux/sparc/bits/stat.h`, so the above should be


```d
static if(!__USE_FILE_OFFSET64)
{
    ushort __pad1;
    __ino_t st_ino;
}
else
{
    __ino64_t st_ino;
}
```

src/core/sys/posix/sys/stat.d
```diff
@@ -534,6 +534,83 @@ version( CRuntime_Glibc )
...
+            static if(!__USE_FILE_OFFSET64)
+            {
+                __blkcnt_t st_blocks;
+            }
+            else
+            {
+                __blkcnt64_t st_blocks;
+            }
+
+            static if( __USE_MISC || __USE_XOPEN2K8)
```
Calrama: I don't see the `__USE_MISC` in glibc 2.23 here, only `__USE_XOPEN2K8`. Which version did you use?

src/core/sys/posix/sys/stat.d
ibuclaw: This is `version SPARC64`. Your suggestions are only relevant for 32-bit SPARC.

ibuclaw: Hint, wordsize can only be 64, and so the first branch is always taken.

Calrama: > Hint, wordsize can only be 64, and so the first branch is always taken.

Point taken. I'd still like to know where the `__USE_MISC`(below)  comes from, since I can't see that in the source.

ibuclaw: This entire block is a copy pasto from `version (AArch64)`, I just diff'd the C headers and changed what was different.  Can remove, it is always false anyway.

Something should probably be done about those always false identifiers in sys/posix/config. But not now.

Calrama: > This entire block is a copy pasto from version (AArch64), I just diff'd the C headers and changed what was different. Can remove, it is always false anyway.

In glibc 2.25 the only `struct stat` user of `__USE_MISC` I can still see is `microblaze`, so what we have for AArch64 (and possibly others) will (eventually) need adjusting, anyway, but that's not your issue.

ZombineDev:
@ibuclaw please rebase.

ibuclaw:
> Nevermind, I rebased your PR myself.

OK, you'd have to wait until next week anyway if waiting for me.

PR #1872, opened 2017-07-13
Fix issue 17413: Prevent deadlock in the GC init 

Burgos:
If the program during initialization in the pressing memory
environment died with the Error thrown by the GC, GC would
not be usable anymore. However, dso registry unregistration
will try to removeRanges. This would cause a deadlock in the GC,
preventing the exit of the program. This commit prevents GC entering
the recursive lock from the chain GC.fullcollect() -> Error -> dso_registry
-> GC.removeRange. Another deadlock would be thread_attachThis -> fullcollect ->
assert -> new AssertError -> scope(exit) GC.enable which is solved by the first
two commits.


Burgos:
I'm not sure how flaky the test case would be, since it relies on the fact that we don't have enough address space for the runtime to be initialised, but enough for shared libraries to be loaded.

Burgos:
Looks like it could be that vibe.d hanging could be not related this this PR:

https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdruntime/detail/PR-1721/17/pipeline/
https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdruntime/detail/PR-1870/1/pipeline/
https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdruntime/detail/PR-1869/3/pipeline/

(last five druntime PRs and bunch phobos ones are hanging)

ZombineDev:
@Burgos the vibe.d failures are not related to your PR. A fix for that will hopefully be deployed soon - see: 
* https://github.com/rejectedsoftware/vibe.d/pull/1837
* https://github.com/rejectedsoftware/vibe.d/pull/1838

ZombineDev:
At least the first part of the call chain (GC.fullcollect() -> Error) could be tested pretty reliably by attempting to allocate memory from a class dtor executed during collection, though I'm not sure if this will hit the case you describe.

About the other call chain (thread_attachThis -> fullcollect ->
assert -> new AssertError -> scope(exit) GC.enable) - how easy it is to reproduce?

src/gc/impl/conservative/gc.d
```diff
@@ -259,12 +259,12 @@ class ConservativeGC : GC
 
     import core.internal.spinlock;
     static gcLock = shared(AlignedSpinLock)(SpinLock.Contention.lengthy);
-    static bool _inFinalizer;
+    static bool _isLocked;
```
MartinNowak: That's easily confusable with the global GC lock? Also it's purpose is not locking but detecting reentrancy from finalizers. I think the name is rather appropriate.

src/gc/impl/conservative/gc.d
```diff
@@ -2408,12 +2408,12 @@ struct Gcx
             start = stop;
         }
 
-        ConservativeGC._inFinalizer = true;
+        ConservativeGC._isLocked = true;
         size_t freedLargePages=void;
         {
-            scope (failure) ConservativeGC._inFinalizer = false;
+            scope (failure) ConservativeGC._isLocked = false;
             freedLargePages = sweep();
```
MartinNowak: `inFinalizer` is set around sweep b/c that's where we call object finalizers.

src/gc/impl/conservative/gc.d
```diff
@@ -2374,12 +2374,14 @@ struct Gcx
...
         {
             // lock roots and ranges around suspending threads b/c they're not reentrant safe
+            ConservativeGC._isLocked = true;
             rangesLock.lock();
             rootsLock.lock();
             scope (exit)
             {
                 rangesLock.unlock();
                 rootsLock.unlock();
+                ConservativeGC._isLocked = false;
```
MartinNowak: From your description the failure seems to be that we're not unlocking those ranges on failure, so it's just broken cleanup on Error, but setting the reentrant flag is quite an ugly hackaround.
An explicit `try / catch (Error) / cleanup / rethrow` works.

nemanja-boric-sociomantic: You're right, of course. I'll update.

MartinNowak:
Ping @Burgos

Burgos:
@MartinNowak sorry, I went to vacation, then had quite some catch-up to do. Will address this tomorrow, thanks for ping!

rainers:
Ping @Burgos, please update as suggested by Martin.

leandro-lucarella-sociomantic:
@nemanja-boric-sociomantic ? :)

PR #1873, opened 2017-07-14
Add constants for F_OFT_* locks

nemanja-boric-sociomantic:
This adds constants for Linux's F_OFD_GETLK, F_OFD_SETLK and
F_OFD_SETLKW.

### DIFF
--- a/src/core/sys/linux/fcntl.d
+++ b/src/core/sys/linux/fcntl.d
@@ -29,6 +29,31 @@ enum {
     FALLOC_FL_UNSHARE_RANGE = 0x40
 }
 
+// From asm-generic/fcntl.h
+/**
+
+Open File Description locks
+
+Usually record locks held by a process are released on *any* close and are
+not inherited across a fork().
+
+These cmd values will set locks that conflict with process-associated
+record  locks, but are "owned" by the open file description, not the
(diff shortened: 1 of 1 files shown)


src/core/sys/linux/fcntl.d
```diff
@@ -17,6 +17,24 @@ enum {
     FALLOC_FL_UNSHARE_RANGE = 0x40
 }
 
+// From asm-generic/fcntl.h
+/*
```
wilzbach: Any reason why you didn't copy the title as well: "Open File Description Locks"?

wilzbach: I see that the other definition here isn't marked with a public Ddoc comment, but I don't see any reason why it shouldn't be exposed in the docs?

Burgos: Not really, I'll update tomorrow with the ddoc (will fix entire file) and add the title. Thanks!

nemanja-boric-sociomantic:
Updated with the ddoc done on the file.

wilzbach:
> Updated with the ddoc done on the file.

Thanks a lot. Still LGTM - anyone objecting? Feel free to merge this in two days with your @Burgos account if no further words are heard ;-)

PR #1876, opened 2017-07-16
Merge mangle into master

andralex:


rainers:
This looks like the wrong branch and direction (the GC coverage change is not required for the mangling). Martin has opened another PR: https://github.com/dlang/druntime/pull/1877

PR #1878, opened 2017-07-16
profile test: allow both the old and the new mangling

rainers:
needed for https://github.com/dlang/dmd/pull/6998

### DIFF
--- a/test/profile/Makefile
+++ b/test/profile/Makefile
@@ -33,7 +33,8 @@ $(ROOT)/both.done: $(ROOT)/%.done: $(ROOT)/%
 	$(QUIET)$(TIMELIMIT)$(ROOT)/$* $(ROOT)/both.log $(ROOT)/both.def $(ROOT)/bothgc.log
 	$(QUIET)$(GREP) -q '1 .*_Dmain' $(ROOT)/both.log
 	$(QUIET)$(GREP) -q '1000 .*both.Num\* both.foo(uint)' $(ROOT)/both.log
-	$(QUIET)$(DIFF) both.def.exp $(ROOT)/both.def
+	# (temporarily) allow both the new and the old mangling in both.def
+	$(QUIET)$(DIFF) both.def.exp $(ROOT)/both.def || $(DIFF) bothnew.def.exp $(ROOT)/both.def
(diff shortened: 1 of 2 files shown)


test/profile/Makefile
```diff
@@ -33,7 +33,7 @@ $(ROOT)/both.done: $(ROOT)/%.done: $(ROOT)/%
 	$(QUIET)$(TIMELIMIT)$(ROOT)/$* $(ROOT)/both.log $(ROOT)/both.def $(ROOT)/bothgc.log
 	$(QUIET)$(GREP) -q '1 .*_Dmain' $(ROOT)/both.log
 	$(QUIET)$(GREP) -q '1000 .*both.Num\* both.foo(uint)' $(ROOT)/both.log
-	$(QUIET)$(DIFF) both.def.exp $(ROOT)/both.def
+	$(QUIET)$(DIFF) both.def.exp $(ROOT)/both.def || $(DIFF) both.def.exp2 $(ROOT)/both.def
```
wilzbach: Maybe adding a new comment that this allows the old & new mangling and can be removed in the future wouldn't hurt?

test/profile/both.def.exp2
```diff
@@ -0,0 +1,5 @@
+
+FUNCTIONS
+	_Dmain
```
wilzbach: You have tabs here ... (you need to either replace them or make the check [more specific](https://github.com/dlang/druntime/blob/master/posix.mak#L279))..

rainers: Those tabs are generated as part of the profile output and are also in both.exp.

test/profile/Makefile
rainers: Amended a comment

test/profile/both.def.exp2
wilzbach: I know - I was just FYIing the CI error.
I didn't add the whitespace check here, but it seems that [`.exp` files are explicitly filtered out](https://github.com/dlang/druntime/blob/master/posix.mak#L274). So the simple fix is to rename the file ...

rainers: Sorry, misunderstood at first. Renamed now.

wilzbach: Thanks!

PR #1880, opened 2017-07-18
Enable -fPIC by default

wilzbach:
Regarding https://github.com/dlang/druntime/pull/1721

> Unfortunately, the
line_trace test in test/exceptions is then failing (at least on my
system), because it doesn't actually generate a stacktrace. So, this
doesn't seem to completely fix it so that the PIC=1 unittest build
works, but it's close.

The test works on my system.

See also:
- [ ] https://github.com/dlang/phobos/pull/5586
- [ ] https://github.com/dlang/dmd/pull/7002

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -29,7 +29,12 @@ INSTALL_DIR=../install
 DOCDIR=doc
 IMPDIR=import
 
-OPTIONAL_PIC:=$(if $(PIC),-fPIC,)
+# -fPIC is enabled by default and can be disabled with DISABLE_PIC=1
+ifeq (,$(DISABLE_PIC))
+    PIC_FLAG:=-fPIC
+else
+    PIC_FLAG:=
+endif
 OPTIONAL_COVERAGE:=$(if $(TEST_COVERAGE),-cov,)
 
 ifeq (osx,$(OS))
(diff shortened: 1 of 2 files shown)


wilzbach:
:tada: - everything is passing :)

posix.mak
```diff
@@ -29,7 +29,12 @@ INSTALL_DIR=../install
 DOCDIR=doc
 IMPDIR=import
 
-OPTIONAL_PIC:=$(if $(PIC),-fPIC,)
+# -fPIC is enabled by default and can be disabled with PIC=0
+ifneq ($(PIC),0)
+    DEFAULT_PIC:=-fPIC
```
atilaneves: `DEFAULT_PIC` seems misleading to me, how about `PIC_FLAG`?

wilzbach: That's what @MartinNowak suggested here: https://github.com/dlang/phobos/pull/5586#discussion_r127688645

I really don't care about the name of this parameter ;-)
If necessary we can always change the name, but I highly doubt wants to deactivate `PIC` (e.g. on Phobos PIE is hard-coded by default for the C files)

atilaneves: I don't care _that_ much either ;)
But someone else reading this in the future might get confused, because it's not really a default if it can be changed on the command line.

wilzbach:
> DEFAULT_PIC seems misleading to me, how about PIC_FLAG?

Okay changed it to `PIC_FLAG`

atilaneves:
LGTM

PR #1885, opened 2017-07-20
Revert "Add SO_REUSEPORT constant"

wilzbach:
Temporarily reverts dlang/druntime#1879

Allows the Project Tester (and other projects that test `dmd-nightly`) to be green. We can revert the revert and keep it in the queue until this has been fixed at the projects. As a general policy Jenkins (the project tester) should always be green.

CC @CyberShadow

PR #1886, opened 2017-07-20
Revert "Add SO_REUSEPORT constant"

Burgos:
Remade #1885  to make Jenkins test it with dmd master.

> Temporarily reverts dlang/druntime#1879

> Allows the Project Tester (and other projects that test dmd-nightly) to be green. We can revert the revert and keep it in the queue until this has been fixed at the projects. As a general policy Jenkins (the project tester) should always be green.


Burgos:
Nevermind, I didn't see `jenkins/pr-merge` test running for some reason.

PR #1888, opened 2017-07-23
[Trivial] core.thread: Fix missing alignment of global

kinke:


kinke:
Are pulls against `stable` discouraged these days? CI keeps having problems with it.

wilzbach:
> Are pulls against stable discouraged these days? CI keeps having problems with it.

I fixed this yesterday: https://github.com/dlang/dlang.org/pull/1839

Sadly you need to rebase the PR, so that DAutoTest reruns. The trick is to do sth like this:

```
git commit --amend--no-edit
```

To get a new sha hash.

kinke:
Thanks, done. Requires a new auto-merge OK though.

wilzbach:
> Requires a new auto-merge OK though.

Done  :)

wilzbach:
DAutoTest has been fixed in https://github.com/dlang/dlang.org/issues/1842
As this is a trivial change, it passes on all CIs and for some reason I can't force push here, I am manually overwriting this to avoid yet another ping-pong loop (this one-line PR has had much more activity than needed already).

PR #1889, opened 2017-07-26
core.mutex: @nogc lock/unlock

wilzbach:
I couldn't see any reason why they shouldn't be `@nogc` - their implementations are `@nogc`.


ZombineDev:
See https://github.com/dlang/druntime/pull/1728#discussion_r96087687 for why you can't add attributes to those functions. Also the vibe.d failure is quite indicative.

New code should use the `core.sync.mutex.Mutex.*_nothrow()` functions.

PR #1891, opened 2017-07-26
templated array ops

MartinNowak:
- runtime implementation of templated array operations w/ vectorization
- expresion tree is passed in RPN to encode operator precendence
- converts the expression to a simple loop
- for dmd the loop is vectorized
- gdc/ldc rely on auto-vectorization
- huge performance and latency improvements for some operations
- small throughput decreases due to loop overhead with dmd's optimizer
  and the `floatArray[] / scalar` change mentioned in the changelog
- supports vectorization of more complex operations

## latency
array_ops_latency
## throughput
array_ops_throughput

### DIFF
--- a/src/core/internal/arrayop.d
+++ b/src/core/internal/arrayop.d
@@ -155,15 +155,18 @@ __vector(T[N]) unaop(string op, T, size_t N)(in __vector(T[N]) a) if (op[0] == '
 // mixin gen
 
 // filter out ops without matching SSE/SIMD instructions (could be composed of several instructions though)
-bool vectorizeableOps(E)(string[] ops)
+template vectorizeableOps(E, ops...) if (ops.length > 1)
 {
-    // dfmt off
-    return !(
-        ops.contains("/", "/=") && __traits(isIntegral, E) ||
-        ops.contains("*", "*=") && __traits(isIntegral, E) && E.sizeof != 2 ||
-        ops.contains("%", "%=")
-    );
-    // dfmt on
+    enum vectorizeableOps = vectorizeableOps!(E, ops[0 .. $ / 2])
+            && vectorizeableOps!(E, ops[$ / 2 .. $]);
+}
+
+template vectorizeableOps(E, string op)
+{
+    static if (isUnaryOp(op))
+        enum vectorizeableOps = is(typeof((vec!E a) => mixin(op[1 .. $] ~ " a")));
+    else
+        enum vectorizeableOps = is(typeof((vec!E a, vec!E b) => mixin("a " ~ op ~ " b")));
 }
 
 // filter out things like float[] = float[] / size_t[]
--- a/src/core/internal/arrayop.d
+++ b/src/core/internal/arrayop.d
@@ -53,45 +53,22 @@ private:
 
 // SIMD helpers
 
-version (GNU)
-    import gcc.builtins;
-else version (LDC)
+version (DigitalMars)
 {
-    import ldc.simd;
-    import ldc.gccbuiltins_x86;
-}
-else version (DigitalMars)
     import core.simd;
-else
-    static assert(0, "unimplemented");
 
-template vec(T)
-{
-    enum regsz = 16; // SSE2
-    enum N = regsz / T.sizeof;
-    alias vec = __vector(T[N]);
-}
-
-void store(T, size_t N)(T* p, in __vector(T[N]) val)
-{
-    pragma(inline, true);
-    alias vec = __vector(T[N]);
-
-    version (LDC)
+    template vec(T)
     {
-        storeUnaligned!vec(val, p);
+        enum regsz = 16; // SSE2
+        enum N = regsz / T.sizeof;
+        alias vec = __vector(T[N]);
     }
-    else version (GNU)
+
+    void store(T, size_t N)(T* p, in __vector(T[N]) val)
     {
-        static if (is(T == float))
-            __builtin_ia32_storeups(p, val);
-        else static if (is(T == double))
-            __builtin_ia32_storeupd(p, val);
-        else
-            __builtin_ia32_storedqu(cast(char*) p, val);
-    }
-    else version (DigitalMars)
-    {
+        pragma(inline, true);
+        alias vec = __vector(T[N]);
+
         static if (is(T == float))
             cast(void) __simd_sto(XMM.STOUPS, *cast(vec*) p, val);
         else static if (is(T == double))
@@ -99,28 +76,14 @@ void store(T, size_t N)(T* p, in __vector(T[N]) val)
         else
             cast(void) __simd_sto(XMM.STODQU, *cast(vec*) p, val);
     }
-}
 
-const(__vector(T[N])) load(T, size_t N)(in T* p)
-{
-    pragma(inline, true);
-    alias vec = __vector(T[N]);
+    const(__vector(T[N])) load(T, size_t N)(in T* p)
+    {
+        import core.simd;
 
-    version (LDC)
-    {
-        return loadUnaligned!vec(cast(T*) p);
-    }
-    else version (GNU)
-    {
-        static if (is(T == float))
-            return __builtin_ia32_loadups(p);
-        else static if (is(T == double))
-            return __builtin_ia32_loadupd(p);
-        else
-            return __builtin_ia32_loaddqu(cast(const char*) p);
-    }
-    else version (DigitalMars)
-    {
+        pragma(inline, true);
+        alias vec = __vector(T[N]);
+
         static if (is(T == float))
             return __simd(XMM.LODUPS, *cast(const vec*) p);
         else static if (is(T == double))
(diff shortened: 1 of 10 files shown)


MartinNowak:
Jeez, got broken by a7867ca72bf51a6db9e3ed44428e142a21d3de75.

src/core/internal/arrayop.d
```diff
@@ -0,0 +1,453 @@
...
+version (X86)
+{
+    version (DigitalMars)
+        version = NoSIMD; // DMD-x86 does not support SIMD
+    else version = X86_OR_X64;
+}
+else version (X86_64)
+    version = X86_OR_X64;
+else
+    static assert(0, "unimplemented");
```
ibuclaw: Don't break ARM, please.

src/core/internal/arrayop.d
```diff
@@ -0,0 +1,453 @@
...
+    void store(T, size_t N)(T* p, in __vector(T[N]) val)
+    {
+        pragma(inline, true);
+        alias vec = __vector(T[N]);
+
+        version (LDC)
+        {
+            storeUnaligned!vec(val, p);
+        }
+        else version (GNU)
```
ibuclaw: You have `version (GNU)` *inside* `DigitalMars`

src/core/internal/arrayop.d
```diff
@@ -0,0 +1,453 @@
...
+            return loadUnaligned!vec(cast(T*) p);
+        }
+        else version (GNU)
+        {
+            static if (is(T == float))
+                return __builtin_ia32_loadups(p);
+            else static if (is(T == double))
+                return __builtin_ia32_loadupd(p);
+            else
+                return __builtin_ia32_loaddqu(cast(const char*) p);
```
ibuclaw: Why?

src/core/internal/arrayop.d
```diff
@@ -0,0 +1,453 @@
...
+            do
+            {
+                mixin(vectorExp!Args ~ ";");
+                pos += vec.length;
+            }
+            while (--n);
+        }
+    }
+    for (; pos < res.length; ++pos)
+        mixin(scalarExp!Args ~ ";");
```
ibuclaw: Does this generate something comparable to what the compiler currently generates for `fd->isArrayOp` functions?

ibuclaw: I checked, and the loop is identical.  https://explore.dgnu.org/g/sNV7Bo  :+1:

src/core/internal/arrayop.d
MartinNowak: Oh, just a leftover, I removed the GNU/LDC stuff later on, when I first tested this a few month ago auto-vectorization didn't work too well.

src/core/internal/arrayop.d
MartinNowak: Yes, we should figure out which template instances are needed at runtime, affects all compilers and could be done in the frontend.

MartinNowak:
A bit frustrating how much effort was only spent because of dmd's backend.

src/core/internal/arrayop.d
ibuclaw: Yeah, there's a very *very* limited set of conditions that allow it to generate small and optimal code.  In the best case for parameters, you might at least have a branch generated that is ran if all conditions are met.

Otherwise I guess we could generate a generic simd load using:
```
auto v1 = *cast(float4*) a1.ptr;
```
Loop as necessary, then do single element operations after that.

klickverbot: `*cast(float4*) a1.ptr`

That would assume alignment to `float4.alignof`.

ibuclaw: > That would assume alignment to float4.alignof.

Right.  I should have remembered the segfault in dmd I recently encountered.  Because that is precisely what would happen. ;-)

src/core/internal/traits.d
```diff
@@ -187,3 +187,25 @@ template hasElaborateCopyConstructor(T...)
     else
         enum bool hasElaborateCopyConstructor = false;
 }
+
+template Filter(alias pred, TList...)
```
WalterBright: Please add comment saying what `Filter` does.

src/object.d
```diff
@@ -3634,6 +3634,12 @@ if (!__traits(isScalar, T1))
     assert(__cmp([c2, c2], [c1, c1]) > 0);
 }
 
+template _arrayOp(Args...)
```
WalterBright: The purpose of this is mysterious. Is it to just forward to `core.internal.arrayop.arrayOp` ? If so, why not just use alias? Please add documentation comment.

WalterBright:
@ibuclaw is your review satisfied?

WalterBright:
> A bit frustrating how much effort was only spent because of dmd's backend.

Yeah, but your improvements made it worth the effort!

src/core/internal/arrayop.d
ibuclaw: Two really important pieces of information that could be most beneficial are:
1. Was this instantiated in only by CTFE? We never need these.
2. Was this instantiated inside a function or module/class scope?  With instantiations from a function it should be fine to discard inlined and unreferenced functions in this current compilation.

ibuclaw:
> @ibuclaw is your review satisfied?

Yes, I tested a copy of and it works on all gdc targets. Just the comment on the extra template bloat of entirely unreferenced functions. But that is a compiler concern and not a blocker for this.

src/object.d
MartinNowak: It's a template so we don't unnecessarily import internal modules.

MartinNowak:
Done @WalterBright 

> Yeah, but your improvements made it worth the effort!

Hardly, neither GDC nor LDC need any of this. The vector code in *this* PR (which required the various dmd backend fixes for SIMD) basically just does the same as auto-vectorization in GDC/LDC.

At least we can stop maintaining a huge amout of hand-written assembly code.

MartinNowak:
Anything left @WalterBright, @ibuclaw?

ibuclaw:
I have no problems with this.

ibuclaw:
Do we have an open issue about template instantiation in general?

MartinNowak:
> Do we have an open issue about template instantiation in general?

It's not even clear whether it's an actual problem (and how big it is).
[Issue 17719 – compiler generates code for CTFE-only templates](https://issues.dlang.org/show_bug.cgi?id=17719)
Intuitively I'd say the codegen for CTFE instantiations is less of a problem compared to the semantic speed of template instantiations in general. Lots of instantiations will be used at runtime, and distinguishing which are not used at runtime is not exactly trivial.

ibuclaw:
> Lots of instantiations will be used at runtime, and distinguishing which are not used at runtime is not exactly trivial.

indeed.  But I think we (possibly meaning I and @klickverbot, not dmd) could probably get away with just adding one new field that represents the least restrictive scope that a template was instantiated in.  If an instantiation is only ever used inside a function, then I could allow my backend to discard inlined and/or unreferenced templates.  However if it were instantiated inside a top-level type or module scope, then everything will need to be emitted.

Maybe I'm only thinking of contrived / simple examples though.

ibuclaw:
@WalterBright - please review. :-)

src/core/internal/arrayop.d
```diff
@@ -0,0 +1,428 @@
...
+enum compatibleVecTypes(E, Types...) = compatibleVecTypes!(E, Types[0 .. $ / 2])
+        && compatibleVecTypes!(E, Types[$ / 2 .. $]);
+
+template vectorizeable(E : E[], Args...)
+{
+    static if (is(vec!E))
+        enum vectorizeable = vectorizeableOps!E([Filter!(not!isType, Args)])
+                && compatibleVecTypes!(E, Filter!(isType, Args));
+    else
+                    enum vectorizeable = false;
```
MartinNowak: It's a bug in https://github.com/dlang-community/dfmt/issues/286, fixed manually for now.

src/core/internal/arrayop.d
```diff
@@ -0,0 +1,422 @@
...
+        return false;
+    }
+}
+
+bool isBinaryAssignOp(string op)
+{
+    return op.length == 2 && op[1] == '=' && isBinaryOp(op[0 .. 1]);
+}
+
+string scalarExp(Args...)()
```
WalterBright: Desperately needs documentation - for example, what is the format of the RPN string? What are the Args? Where does the RPN string come from?

src/core/internal/arrayop.d
```diff
@@ -183,6 +194,9 @@ bool isBinaryAssignOp(string op)
     return op.length == 2 && op[1] == '=' && isBinaryOp(op[0 .. 1]);
 }
 
+// Generate mixin expression to perform scalar arrayOp loop expression, assumes
+// `pos` to be the current slice index, `args` to contain operand values, and
+// `res` the target slice.
```
WalterBright: I don't see pos, args, or res in the parameter list or even in the function body. Also, when documenting parameters, please use Ddoc conventions, i.e. a `Params:` block.

MartinNowak: It's just a small helper function that generates a mixin string for the only public method in this module `_arrayOps`.

PR #1898, opened 2017-08-10
Added system-specific bindings for netinet/in.h

jmdavis:
Both FreeBSD and Linux add quite a few items to netinet/in.h which are not part of POSIX (some being part of other standards, some not). I'd be shocked if the other *nix systems out there didn't do the same, but I don't have any such systems set up, so I haven't dealt with them. This adds core.sys.{freebsd,linux}.netinet.in_ which include the non-POSIX declarations and publicly import core.sys.posix.netinet.in_. It includes quite a few of the constants that are in std.c.linux.socket (which is deprecated and overdue for removal), including ones that vibe.d uses and has never bothered to redeclare itself or add to druntime in order to not use the deprecated symbols. With this, hopefully, it can stop using the deprecated symbols.

Not all of the symbols from those header files have been included (some were marked as deprecated or obsolete and some have been included as commented out, because they required symbols that we don't have yet or were macros which were too much of a pain to figure out how to port at the moment), but most of the symbols have been included.

Also, I added a number of files to mak/SRCS that were missing for some reason.

### DIFF
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -52,8 +52,19 @@ SRCS=\
 	src\core\sync\rwmutex.d \
 	src\core\sync\semaphore.d \
 	\
+	src\core\sys\freebsd\dlfcn.d \
 	src\core\sys\freebsd\execinfo.d \
+	src\core\sys\freebsd\sys\_bitset.d \
+	src\core\sys\freebsd\sys\_cpuset.d \
+	src\core\sys\freebsd\sys\cdefs.d \
+	src\core\sys\freebsd\sys\elf_common.d \
+	src\core\sys\freebsd\sys\elf.d \
+	src\core\sys\freebsd\sys\elf32.d \
+	src\core\sys\freebsd\sys\elf64.d \
 	src\core\sys\freebsd\sys\event.d \
+	src\core\sys\freebsd\sys\link_elf.d \
+	src\core\sys\freebsd\sys\mman.d \
+	src\core\sys\freebsd\time.d \
 	\
 	src\core\sys\linux\stdio.d \
 	src\core\sys\linux\tipc.d \
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -54,6 +54,7 @@ SRCS=\
 	\
 	src\core\sys\freebsd\dlfcn.d \
 	src\core\sys\freebsd\execinfo.d \
+	src\core\sys\freebsd\netinet\in_.d \
 	src\core\sys\freebsd\sys\_bitset.d \
 	src\core\sys\freebsd\sys\_cpuset.d \
 	src\core\sys\freebsd\sys\cdefs.d \
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -52,6 +52,8 @@ SRCS=\
 	src\core\sync\rwmutex.d \
 	src\core\sync\semaphore.d \
 	\
+	src\core\sys\darwin\netinet\in_.d \
+	\
 	src\core\sys\freebsd\dlfcn.d \
 	src\core\sys\freebsd\execinfo.d \
 	src\core\sys\freebsd\netinet\in_.d \
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -67,6 +67,7 @@ SRCS=\
 	src\core\sys\freebsd\sys\mman.d \
 	src\core\sys\freebsd\time.d \
 	\
+	src\core\sys\linux\netinet\in_.d \
 	src\core\sys\linux\stdio.d \
 	src\core\sys\linux\tipc.d \
 	src\core\sys\linux\sys\inotify.d \
--- a/src/core/sys/darwin/netinet/in_.d
+++ b/src/core/sys/darwin/netinet/in_.d
@@ -1,0 +1,611 @@
+//Written in the D programming language
+
+/++
+    D header file for Darwin's extensions to POSIX's netinet/in.h.
+
+    Copyright: Copyright 2017 -
+    License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+    Authors:   Jonathan M Davis
+ +/
+module core.sys.darwin.netinet.in_;
+
+import core.sys.darwin.sys.cdefs;
+
+public import core.sys.posix.netinet.in_;
+
+version(OSX)
+    version = Darwin;
+else version(iOS)
+    version = Darwin;
+else version(TVOS)
+    version = Darwin;
+else version(WatchOS)
+    version = Darwin;
+
+version(Darwin):
+extern(C) nothrow @nogc:
+
+enum IPPROTO_IP = 0;
+static if (_DARWIN_C_SOURCE)
+    enum IPPROTO_HOPOPTS = 0;
+enum IPPROTO_ICMP = 1;
+static if (_DARWIN_C_SOURCE)
+{
+    enum IPPROTO_IGMP = 2;
+    enum IPPROTO_GGP  = 3;
+    enum IPPROTO_IPV4 = 4;
+    enum IPPROTO_IPIP = IPPROTO_IPV4;
+}
+enum IPPROTO_TCP = 6;
+static if (_DARWIN_C_SOURCE)
+{
+    enum IPPROTO_ST     = 7;
+    enum IPPROTO_EGP    = 8;
+    enum IPPROTO_PIGP   = 9;
+    enum IPPROTO_RCCMON = 10;
+    enum IPPROTO_NVPII  = 11;
+    enum IPPROTO_PUP    = 12;
+    enum IPPROTO_ARGUS  = 13;
+    enum IPPROTO_EMCON  = 14;
+    enum IPPROTO_XNET   = 15;
+    enum IPPROTO_CHAOS  = 16;
+}
+enum IPPROTO_UDP = 17;
+static if (_DARWIN_C_SOURCE)
+{
+    enum IPPROTO_MUX    = 18;
+    enum IPPROTO_MEAS   = 19;
+    enum IPPROTO_HMP    = 20;
+    enum IPPROTO_PRM    = 21;
+    enum IPPROTO_IDP    = 22;
+    enum IPPROTO_TRUNK1 = 23;
+    enum IPPROTO_TRUNK2 = 24;
+    enum IPPROTO_LEAF1  = 25;
+    enum IPPROTO_LEAF2  = 26;
+    enum IPPROTO_RDP    = 27;
+    enum IPPROTO_IRTP   = 28;
+    enum IPPROTO_TP     = 29;
+    enum IPPROTO_BLT    = 30;
+    enum IPPROTO_NSP    = 31;
+    enum IPPROTO_INP    = 32;
+    enum IPPROTO_SEP    = 33;
+    enum IPPROTO_3PC    = 34;
+    enum IPPROTO_IDPR   = 35;
+    enum IPPROTO_XTP    = 36;
+    enum IPPROTO_DDP    = 37;
+    enum IPPROTO_CMTP   = 38;
+    enum IPPROTO_TPXX   = 39;
+    enum IPPROTO_IL     = 40;
+}
+enum IPPROTO_IPV6 = 41;
+static if (_DARWIN_C_SOURCE)
+{
+    enum IPPROTO_SDRP      = 42;
+    enum IPPROTO_ROUTING   = 43;
+    enum IPPROTO_FRAGMENT  = 44;
+    enum IPPROTO_IDRP      = 45;
+    enum IPPROTO_RSVP      = 46;
+    enum IPPROTO_GRE       = 47;
+    enum IPPROTO_MHRP      = 48;
+    enum IPPROTO_BHA       = 49;
+    enum IPPROTO_ESP       = 50;
+    enum IPPROTO_AH        = 51;
+    enum IPPROTO_INLSP     = 52;
+    enum IPPROTO_SWIPE     = 53;
+    enum IPPROTO_NHRP      = 54;
+    enum IPPROTO_ICMPV6    = 58;
+    enum IPPROTO_NONE      = 59;
+    enum IPPROTO_DSTOPTS   = 60;
+    enum IPPROTO_AHIP      = 61;
+    enum IPPROTO_CFTP      = 62;
+    enum IPPROTO_HELLO     = 63;
+    enum IPPROTO_SATEXPAK  = 64;
+    enum IPPROTO_KRYPTOLAN = 65;
+    enum IPPROTO_RVD       = 66;
+    enum IPPROTO_IPPC      = 67;
+    enum IPPROTO_ADFS      = 68;
+    enum IPPROTO_SATMON    = 69;
+    enum IPPROTO_VISA      = 70;
+    enum IPPROTO_IPCV      = 71;
+    enum IPPROTO_CPNX      = 72;
+    enum IPPROTO_CPHB      = 73;
+    enum IPPROTO_WSN       = 74;
+    enum IPPROTO_PVP       = 75;
+    enum IPPROTO_BRSATMON  = 76;
+    enum IPPROTO_ND        = 77;
+    enum IPPROTO_WBMON     = 78;
+    enum IPPROTO_WBEXPAK   = 79;
+    enum IPPROTO_EON       = 80;
+    enum IPPROTO_VMTP      = 81;
+    enum IPPROTO_SVMTP     = 82;
+    enum IPPROTO_VINES     = 83;
+    enum IPPROTO_TTP       = 84;
+    enum IPPROTO_IGP       = 85;
+    enum IPPROTO_DGP       = 86;
+    enum IPPROTO_TCF       = 87;
+    enum IPPROTO_IGRP      = 88;
+    enum IPPROTO_OSPFIGP   = 89;
+    enum IPPROTO_SRPC      = 90;
+    enum IPPROTO_LARP      = 91;
+    enum IPPROTO_MTP       = 92;
+    enum IPPROTO_AX25      = 93;
+    enum IPPROTO_IPEIP     = 94;
+    enum IPPROTO_MICP      = 95;
+    enum IPPROTO_SCCSP     = 96;
+    enum IPPROTO_ETHERIP   = 97;
+    enum IPPROTO_ENCAP     = 98;
+    enum IPPROTO_APES      = 99;
+    enum IPPROTO_GMTP      = 100;
+    enum IPPROTO_PIM       = 103;
+    enum IPPROTO_IPCOMP    = 108;
+    enum IPPROTO_PGM       = 113;
+    enum IPPROTO_SCTP      = 132;
+    enum IPPROTO_DIVERT    = 254;
+}
+enum IPPROTO_RAW = 255;
+
+static if (_DARWIN_C_SOURCE)
+{
+    enum IPPROTO_MAX  = 256;
+    enum IPPROTO_DONE = 257;
+}
+
+static if (_DARWIN_C_SOURCE)
+{
+    enum IPPORT_RESERVED = 1024;
+    enum IPPORT_USERRESERVED = 5000;
+
+    enum IPPORT_HIFIRSTAUTO = 49152;
+    enum IPPORT_HILASTAUTO  = 65535;
+
+    enum IPPORT_RESERVEDSTART = 600;
+}
+
+static if (_DARWIN_C_SOURCE)
+{
+    extern(D) bool IN_CLASSA(in_addr_t i) pure @safe { return (i & 0x80000000) == 0; }
+    enum IN_CLASSA_NET    = 0xff000000;
+    enum IN_CLASSA_NSHIFT = 24;
+    enum IN_CLASSA_HOST   = 0x00ffffff;
+    enum IN_CLASSA_MAX    = 128;
+
+    extern(D) bool IN_CLASSB(in_addr_t i) pure @safe { return (i & 0xc0000000) == 0x80000000; }
+    enum IN_CLASSB_NET    = 0xffff0000;
+    enum IN_CLASSB_NSHIFT = 16;
+    enum IN_CLASSB_HOST   = 0x0000ffff;
+    enum IN_CLASSB_MAX    = 65536;
+
+    extern(D) bool IN_CLASSC(in_addr_t i) pure @safe { return (i & 0xe0000000) == 0xc0000000; }
+    enum IN_CLASSC_NET    = 0xffffff00;
+    enum IN_CLASSC_NSHIFT = 8;
+    enum IN_CLASSC_HOST   = 0x000000ff;
+
+    extern(D) bool IN_CLASSD(in_addr_t i) pure @safe { return (i & 0xf0000000) == 0xe0000000; }
+    enum IN_CLASSD_NET    = 0xf0000000;
+    enum IN_CLASSD_NSHIFT = 28;
+    enum IN_CLASSD_HOST   = 0x0fffffff;
+    extern(D) bool IN_MULTICAST(in_addr_t i) pure @safe { return IN_CLASSD(i); }
+
+    // The fact that these are identical looks suspicious (they're not quite
+    // identical on Linux). However, this _is_ how they're defined in FreeBSD
+    // and on OS X. So, while it _might_ be a bug, if it is, it's an upstream
+    // one, and we're compatible with it.
+    extern(D) bool IN_EXPERIMENTAL(in_addr_t i) pure @safe { return (i & 0xf0000000) == 0xf0000000; }
+    extern(D) bool IN_BADCLASS(in_addr_t i) pure @safe { return (i & 0xf0000000) == 0xf0000000; }
+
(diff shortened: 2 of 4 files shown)


mak/SRCS
```diff
@@ -52,8 +52,19 @@ SRCS=\
 	src\core\sync\rwmutex.d \
```
nemanja-boric-sociomantic: Re: commit message: Did you forget to include change to `pthread_np` into the patch?

nemanja-boric-sociomantic: (and here)

src/core/sys/freebsd/netinet/in_.d
```diff
@@ -0,0 +1,581 @@
+//Written in the D programming language
```
nemanja-boric-sociomantic: Is this line really necessary, feels like noise :-) ?

src/core/sys/freebsd/netinet/in_.d
```diff
@@ -0,0 +1,581 @@
...
+
+    static if (__BSD_VISIBLE)
+    {
+        enum IPV6PORT_RESERVED    = 1024;
+        enum IPV6PORT_ANONMIN     = 49152;
+        enum IPV6PORT_ANONMAX     = 65535;
+        enum IPV6PORT_RESERVEDMIN = 600;
+        enum IPV6PORT_RESERVEDMAX = IPV6PORT_RESERVED - 1;
+
+// TODO
```
nemanja-boric-sociomantic: TODO on purpose?

src/core/sys/linux/netinet/in_.d
```diff
@@ -0,0 +1,387 @@
...
+extern(D) bool IN_BADCLASS(in_addr_t i) pure @safe { return (i & 0xf0000000) == 0xf0000000; }
+
+enum IN_LOOPBACKNET = 127;
+
+enum INADDR_UNSPEC_GROUP    = 0xe0000000;
+enum INADDR_ALLHOSTS_GROUP  = 0xe0000001;
+enum INADDR_ALLRTRS_GROUP   = 0xe0000002;
+enum INADDR_MAX_LOCAL_GROUP = 0xe00000ff;
+
+// TODO
```
nemanja-boric-sociomantic: ditto todo

mak/SRCS
jmdavis: No. I removed it purposefully, because I was afraid that it would break something to just change the module name even though it was wrong. I just forgot to update the commit message. So, I'll fix that shortly. I'll figure out what to do about pthread_np in a different PR.

src/core/sys/freebsd/netinet/in_.d
jmdavis: We normally put that at the top of all of the D files. AFAIK, all of the .d files in druntime and Phobos have it.

nemanja-boric-sociomantic: Hmm, looks like it's half/half for phobos, and 23 vs 450 for druntime (that's why it looked strange to me), but fair enough :+1:

src/core/sys/freebsd/netinet/in_.d
jmdavis: I guess that I can add a note about it if that helps, but yes, on purpose. I didn't want to take the time to figure out how to convert it, though thinking about it now, I'm pretty sure that I know how. So, I'll see if I'm right and update it if so and comment it for later if not. The one on the Linux side is for the same reason

nemanja-boric-sociomantic: I was just curious if you left this for latter and then forgot (which happens to me on a regular basis). No, I don't think there's need to add this note - after the review is finished, I don't think the note would carry any information.

jmdavis: I just added the implementation. It was pretty straightforward. For some reason, I had been thinking that something more involved was going to be required.

src/core/sys/freebsd/netinet/in_.d
jmdavis: I think that Walter likes people to put in on their files for searchability when google comes trauling along...

jmdavis:
Okay. I managed to get the OS X headers from someone so that I was able to add them into the mix, though I obviously couldn't run the OS X build locally, so I may have to deal with some autotester failures before I get it compiling on OS X. I didn't realize that the OS X stuff had been moved to core.sys.darwin, though the dmd PR where that was discussed makes all of that sound like a mess. :|  In any case, I just made them work for `version(OSX)` for now, since while I expect that they're valid for the other "darwin" OSes, I don't know for sure. Someone who actually deals with those OSes will have to take care of that.

I also changed it so that linux uses `version(CRuntime_Glibc)` now, since it seems that that's what we used for core.sys.posix.netinet.in_, though based on what I understand of the reasoning for using `linux` verses `CRuntime_Glibc`, it seems like it almost makes `linux` useless, because applications almost always deal with libc and not the kernel. :|

src/core/sys/freebsd/netinet/in_.d
wilzbach: Yes, Walter insists on this whenever he happens to catch a PR with a new module addition.
I don't think it makes any difference in Google optimization nowadays, but it definitely doesn't hurt.

jacob-carlborg:
@jmdavis Do you know which version of the SDK the macOS headers are from?

jacob-carlborg:
I diffed the macOS headers and compared to all the other Apple platforms in the Xcode 9 SDK (still in beta) . There are no differences.

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+    enum IN_CLASSD_NSHIFT = 28;
+    enum IN_CLASSD_HOST   = 0x0fffffff;
+    extern(D) bool IN_MULTICAST(in_addr_t i) pure @safe { return IN_CLASSD(i); }
+
+    // The fact that these are identical looks suspicious (they're not quite
+    // identical on Linux). However, this _is_ how they're defined in FreeBSD
+    // and on OS X. So, while it _might_ be a bug, if it is, it's an upstream
+    // one, and we're compatible with it.
+    extern(D) bool IN_EXPERIMENTAL(in_addr_t i) pure @safe { return (i & 0xf0000000) == 0xf0000000; }
+    extern(D) bool IN_BADCLASS(in_addr_t i) pure @safe { return (i & 0xf0000000) == 0xf0000000; }
```
jacob-carlborg: These are equal in the header file as well.

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+
```
jacob-carlborg: The following to constants are missing:

```c
#define	INADDR_LOOPBACK		(u_int32_t)0x7f000001

#define	INADDR_NONE		0xffffffff		/* -1 return */
```

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+               (i & 0xffff0000) == 0xc0a80000;
+    }
+
+    extern(D) bool IN_LOCAL_GROUP(in_addr_t i) pure @safe { return (i & 0xffffff00) == 0xe0000000; }
+
+    extern(D) bool IN_ANY_LOCAL(in_addr_t i) pure @safe { return IN_LINKLOCAL(i) || IN_LOCAL_GROUP(i); }
+
+    enum IN_LOOPBACKNET = 127;
+
+
```
jacob-carlborg: This is missing:

```c
#define INET_ADDRSTRLEN                 16
```

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+    enum IP_DROP_MEMBERSHIP = 13;
+    enum IP_MULTICAST_VIF   = 14;
+    enum IP_RSVP_ON         = 15;
+    enum IP_RSVP_OFF        = 16;
+    enum IP_RSVP_VIF_ON     = 17;
+    enum IP_RSVP_VIF_OFF    = 18;
+    enum IP_PORTRANGE       = 19;
+    enum IP_RECVIF          = 20;
+
+    enum IP_IPSEC_POLICY = 21;
```
jacob-carlborg: Missing: `#define	IP_FAITH		22   /* deprecated */`.

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+    };
+
+    struct group_source_req
+    {
+        align(4):
+        uint             gsr_interface;
+        sockaddr_storage gsr_group;
+        sockaddr_storage gsr_source;
+    };
+
```
jacob-carlborg: This is missing, but the header says it's private:

```c
/*
 * The following structure is private; do not use it from user applications.
 * It is used to communicate IP_MSFILTER/IPV6_MSFILTER information between
 * the RFC 3678 libc functions and the kernel.
 */
struct __msfilterreq {
	uint32_t		 msfr_ifindex;	/* interface index */
	uint32_t		 msfr_fmode;	/* filter mode for group */
	uint32_t		 msfr_nsrcs;	/* # of sources in msfr_srcs */
	uint32_t		__msfr_align;	
	struct sockaddr_storage	 msfr_group;	/* group address */
	struct sockaddr_storage	*msfr_srcs;
};
```

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+    int bindresvport_sa(int, sockaddr*);
+}
+
+// =============================================================================
+// What follows is from netinet6/in6.h, but since netinet6/in6.h specifically
+// says that it should only be #included by #including netinet/in.h, it makes
+// more sense to put them in here so that the way they're imported corresponds
+// with the correct way of #including them in C/C++.
+// =============================================================================
+
```
jacob-carlborg: Missing, not sure if it's needed:

```c
/*
 * Identification of the network protocol stack
 * for *BSD-current/release: http://www.kame.net/dev/cvsweb.cgi/kame/COVERAGE
 * has the table of implementation/integration differences.
 */
#define	__KAME__
#define	__KAME_VERSION		"2009/apple-darwin"
```

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+    enum IPV6PORT_RESERVED    = 1024;
+    enum IPV6PORT_ANONMIN     = 49152;
+    enum IPV6PORT_ANONMAX     = 65535;
+    enum IPV6PORT_RESERVEDMIN = 600;
+    enum IPV6PORT_RESERVEDMAX = IPV6PORT_RESERVED - 1;
+}
+
+
+    enum IN6ADDR_ANY_INIT = in6_addr.init;
+    enum IN6ADDR_LOOPBACK_INIT = in6_addr([0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
```
jacob-carlborg: Indentation is not correct.

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+static if (_DARWIN_C_SOURCE)
+{
+    enum IPV6PORT_RESERVED    = 1024;
+    enum IPV6PORT_ANONMIN     = 49152;
+    enum IPV6PORT_ANONMAX     = 65535;
+    enum IPV6PORT_RESERVEDMIN = 600;
+    enum IPV6PORT_RESERVEDMAX = IPV6PORT_RESERVED - 1;
+}
+
+
```
jacob-carlborg: Missing:

```c
/*
 * Socket address for IPv6
 */
#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
#define	SIN6_LEN
#endif /* (_POSIX_C_SOURCE && !_DARWIN_C_SOURCE) */
```

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+enum IPV6_MULTICAST_HOPS = 10;
+enum IPV6_MULTICAST_LOOP = 11;
+enum IPV6_JOIN_GROUP     = 12;
+enum IPV6_LEAVE_GROUP    = 13;
+
+static if (_DARWIN_C_SOURCE)
+{
+    enum IPV6_PORTRANGE    = 14;
+    enum ICMP6_FILTER      = 18;
+
```
jacob-carlborg: Missing:

```c
#define	IPV6_2292PKTINFO	19 /* bool; send/recv if, src/dst addr */
#define	IPV6_2292HOPLIMIT	20 /* bool; hop limit */
#define	IPV6_2292NEXTHOP	21 /* bool; next hop addr */
#define	IPV6_2292HOPOPTS	22 /* bool; hop-by-hop option */
#define	IPV6_2292DSTOPTS	23 /* bool; destinaion option */
#define	IPV6_2292RTHDR		24 /* ip6_rthdr: routing header */

/* buf/cmsghdr; set/get IPv6 options [obsoleted by RFC3542] */
#define	IPV6_2292PKTOPTIONS	25

#ifdef __APPLE_USE_RFC_2292
#define	IPV6_PKTINFO	IPV6_2292PKTINFO
#define	IPV6_HOPLIMIT	IPV6_2292HOPLIMIT
#define	IPV6_NEXTHOP	IPV6_2292NEXTHOP
#define	IPV6_HOPOPTS	IPV6_2292HOPOPTS
#define	IPV6_DSTOPTS	IPV6_2292DSTOPTS
#define	IPV6_RTHDR	IPV6_2292RTHDR
#define	IPV6_PKTOPTIONS	IPV6_2292PKTOPTIONS
#endif /* __APPLE_USE_RFC_2292 */
```

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+    enum IPV6_CHECKSUM = 26;
+}
+enum IPV6_V6ONLY = 27;
+static if (_DARWIN_C_SOURCE)
+{
+    enum IPV6_BINDV6ONLY = IPV6_V6ONLY;
+
+    enum IPV6_IPSEC_POLICY = 28;
+
```
jacob-carlborg: Missing: `#define	IPV6_FAITH		29 /* deprecated */`.

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+    enum IPV6_PATHMTU = 44;
+
+    enum IPV6_PKTINFO  = 46;
+    enum IPV6_HOPLIMIT = 47;
+    enum IPV6_NEXTHOP  = 48;
+    enum IPV6_HOPOPTS  = 49;
+    enum IPV6_DSTOPTS  = 50;
+    enum IPV6_RTHDR    = 51;
+
```
jacob-carlborg: Missing:

```c
#define	IPV6_PKTINFO	IPV6_3542PKTINFO
#define	IPV6_HOPLIMIT	IPV6_3542HOPLIMIT
#define	IPV6_NEXTHOP	IPV6_3542NEXTHOP
#define	IPV6_HOPOPTS	IPV6_3542HOPOPTS
#define	IPV6_DSTOPTS	IPV6_3542DSTOPTS
#define	IPV6_RTHDR	IPV6_3542RTHDR
```

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+    enum IPV6_AUTOFLOWLABEL = 59;
+
+    enum IPV6_DONTFRAG = 62;
+
+    enum IPV6_PREFER_TEMPADDR = 63;
+
```
jacob-carlborg: Missing:

```c
/*
 * The following option is private; do not use it from user applications.
 * It is deliberately defined to the same value as IP_MSFILTER.
 */
#define	IPV6_MSFILTER		74 /* struct __msfilterreq; */
#endif /* __APPLE_USE_RFC_3542 */
```

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+    enum IPV6CTL_DEFMTU             = 4;
+    enum IPV6CTL_FORWSRCRT          = 5;
+    enum IPV6CTL_STATS              = 6;
+    enum IPV6CTL_MRTSTATS           = 7;
+    enum IPV6CTL_MRTPROTO           = 8;
+    enum IPV6CTL_MAXFRAGPACKETS     = 9;
+    enum IPV6CTL_SOURCECHECK        = 10;
+    enum IPV6CTL_SOURCECHECK_LOGINT = 11;
+    enum IPV6CTL_ACCEPT_RTADV       = 12;
+
```
jacob-carlborg: Missing: `#define	IPV6CTL_KEEPFAITH	13	/* deprecated */`.

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+    enum IPV6CTL_MCAST_PMTU = 44;
+
+    enum IPV6CTL_NEIGHBORGCTHRESH      = 46;
+    enum IPV6CTL_MAXIFPREFIXES         = 47;
+    enum IPV6CTL_MAXIFDEFROUTERS       = 48;
+    enum IPV6CTL_MAXDYNROUTES          = 49;
+    enum ICMPV6CTL_ND6_ONLINKNSRFC4861 = 50;
+
+    enum IPV6CTL_MAXID = 51;
+
```
jacob-carlborg: Missing:

```c
extern int inet6_option_space(int);
extern int inet6_option_init(void *, struct cmsghdr **, int);
extern int inet6_option_append(struct cmsghdr *, const __uint8_t *, int, int);
extern __uint8_t *inet6_option_alloc(struct cmsghdr *, int, int, int);
extern int inet6_option_next(const struct cmsghdr *, __uint8_t **);
extern int inet6_option_find(const struct cmsghdr *, __uint8_t **, int);
```

jacob-carlborg:
I found it a bit difficult to review since some symbols that are defined in `in.h` is missing here but instead are defined in the Posix `in_.d` module.

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+ +/
+module core.sys.darwin.netinet.in_;
+
+import core.sys.darwin.sys.cdefs;
+
+public import core.sys.posix.netinet.in_;
+
+// These are probably the same on the other "Darwin" OSes, but someone will
+// need to verify that.
+version(OSX):
```
jacob-carlborg: Verified, please change to `Darwin`.

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
...
+
+static if (_DARWIN_C_SOURCE)
+{
+    extern(D) bool IN_CLASSA(in_addr_t i) pure @safe { return (i & 0x80000000) == 0; }
+    enum IN_CLASSA_NET    = 0xff000000;
+    enum IN_CLASSA_NSHIFT = 24;
+    enum IN_CLASSA_HOST   = 0x00ffffff;
+    enum IN_CLASSA_MAX    = 128;
+
+    extern(D) bool IN_CLASSB(in_addr_t i) pure @safe { return (i & 0xc0000000) == 0x80000000; }
```
jacob-carlborg: All of these macros use `u_int32_t` instead of `in_addr_t `. They all correspond to the same type though.

jacob-carlborg: Wouldn't a closer translation of these macros be?

```d
bool IN_CLASSB(T)(T i) { return (cast(uint) i & 0xc0000000) == 0x80000000; }
```

src/core/sys/darwin/netinet/in_.d
jmdavis: That's what the comment says.

src/core/sys/darwin/netinet/in_.d
```diff
@@ -0,0 +1,605 @@
+//Written in the D programming language
+
+/++
+    D header file for FreeBSD's extensions to POSIX's netinet/in.h.
```
jmdavis: Yay, for copy-paste errors... :|

src/core/sys/darwin/netinet/in_.d
jmdavis: Because they're in `core.sys.posix.netinet._in`, because they're POSIX, unlike the stuff in this PR.

src/core/sys/darwin/netinet/in_.d
jmdavis: Yes, because it says it's deprecated. I saw no point in including declarations where the header file said that they were deprecated or obsolete. A number of the declarations in in.h were marked that way, and some were marked that way on one system, so I left them out on all of the systems. I don't think that anything that's missing is not missing on purpose, and everything that you've mentioned as missing in a comment is purposefully missing. More declarations can be added later if we need to, whereas removing them later is more of a pain. I don't expect that we'll need the ones that I left out.

src/core/sys/darwin/netinet/in_.d
jmdavis: Per the linux version of in.h, they come from an obsolete RFC, so I didn't put them in either the FreeBSD or OS X bindings.

src/core/sys/darwin/netinet/in_.d
jmdavis: I translated them the way that similar macros were translated in `core.sys.posix.netinet.in_`. Also, this way, they won't accidentally accept an `in6_addr`.

jmdavis:
Okay. I fixed the typos in the OS X stuff, and switched it to use `version(Darwin)` as requested. As mentioned response to some of the comments, the missing declarations are missing on purpose - mostly because the declarations were marked as deprecated or obsolete, so in theory, no one should be using them, and it's easier to add more later than to remove them later.

jacob-carlborg:
In general I don't think it's our job to decide which symbols should or should not be included in bindings. We should just expose all the symbols. It's fine to remove internal or private symbols, but I do think we should include symbols for obsolete protocols, if they are present in the header file by default.

jacob-carlborg:
Wouldn't it be better to move all the symbols from the Posix file to the platform specific ones now when they exist? The Posix file is already versioned by platform and it's more complicated to maintain when some of the symbols are in one file and some in another file.

This would go in a separate pull request.

jacob-carlborg:
Since I requested changes, can I approve them now somewhere? Or can't I do that because I don't have write access?

nemanja-boric-sociomantic:
@jacob-carlborg you can approve them, go to "Files changed" tab and then "Review" in the top right corner.

jmdavis:
> In general I don't think it's our job to decide which symbols should or should not be included in bindings. We should just expose all the symbols. It's fine to remove internal or private symbols, but I do think we should include symbols for obsolete protocols, if they are present in the header file by default.

I don't agree, but you're free to create a PR to add the other symbols.

> Wouldn't it be better to move all the symbols from the Posix file to the platform specific ones now when they exist? The Posix file is already versioned by platform and it's more complicated to maintain when some of the symbols are in one file and some in another file.

And then how would someone import just the POSIX declarations? If code imports just the POSIX declarations and just uses those, it should be platform-independent between POSIX systems, whereas if it imports everything from a platform, then it risks accidentally using something that's not actually POSIX. Those who want or need to import the system-specific files obviously can, and this PR's whole purpose is to make that possible, but by having the POSIX stuff be separate, it makes it easier to avoid creating platform-dependent code by accident.

I wouldn't expect a lot of maintenance to go into these files anyway. Most of this stuff doesn't change, and it's likely that something will be added to these files only if a declaration is added to the C headers and someone wants to use them from D and finds that they aren't here. I very much doubt that anyone is going to be going to the effort of periodically trying to go over every declaration to make sure that the C and D declarations are in sync. And even if someone does want to do that, it's easy enough to grep for the symbol in question or just search `core.sys.posix.netinet.in_`. And the POSIX stuff is even less likely to change than the rest.

jacob-carlborg:
> And then how would someone import just the POSIX declarations?

That's a pretty good point. I didn't think of that.

> I wouldn't expect a lot of maintenance to go into these files anyway.

That's probably true. It just made my life a bit more difficult when I reviewed this 😃.

MartinNowak:
The added file is missing in mak/COPY.

> Also, I added a number of files to mak/SRCS that were missing for some reason.

Those haven't been there b/c they are header only files and don't need to be compiled :/.

> In general I don't think it's our job to decide which symbols should or should not be included in bindings. We should just expose all the symbols. It's fine to remove internal or private symbols, but I do think we should include symbols for obsolete protocols, if they are present in the header file by default.

Yes, headers should be one-to-one translations. In fact they should be diff-able against the C headers. You might want to reread the Deimos guidelines (https://jkm.github.io/phobos/deimos.html).

For OS-specific headers I've tried to establish to translate the full header, but comment out stuff that's Posix.
https://github.com/dlang/druntime/blob/77b70a76f474422d4375a6432944e51daca95edf/src/core/sys/linux/sys/mman.d
https://github.com/dlang/druntime/blob/77b70a76f474422d4375a6432944e51daca95edf/src/core/sys/linux/dlfcn.d

You might also have overlooked that the bits/ folder in glibc is for platform/asm specific parts, i.e. needs to be versioned for the translated platform/architecture. Never translate the headers from your host OS, instead use glibc's source and please also add links to their glibc.git.

PR #1899, opened 2017-08-10
add missing ^^ and ^^= arrayop support

MartinNowak:
fixup for #1891, required by dlang/dmd#7032.
Unfortunately not really possible to test those with druntime's unittests as PowExp depends on phobos' std.math.pow.


ibuclaw:
It may not be a bad idea to move it to druntime.  As phobos should be optional.

MartinNowak:
> It may not be a bad idea to move it to druntime. As phobos should be optional.

Yes indeed.

PR #1900, opened 2017-08-13
Add overrides for __simd intrinsics

ibuclaw:
This is blocking https://github.com/dlang/dmd/pull/7065

### DIFF
--- a/src/core/simd.d
+++ b/src/core/simd.d
@@ -384,14 +384,58 @@ version ( D_SIMD )
    * Returns:
    *      result of opcode
    */
-  pure @safe void16 __simd(XMM opcode, void16 op1, void16 op2);
+  pure @safe V1 simd(XMM opcode, V1, V2)(V1 op1, V2 op2)
+      if (is(V1 == __vector) && is(V2 == __vector))
+  {
+      pragma(inline, true);
+      return cast(V1)__simd(opcode, op1, op2);
+  }
+
+  pure @safe void16 __simd(XMM opcode, void16 op1, void16 op2); // intrinsic
+
+  ///
+  unittest
+  {
+      float4 a;
+      a = simd!(XMM.PXOR)(a, a);
+  }
 
   /**
    * Unary SIMD instructions.
    */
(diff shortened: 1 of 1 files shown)


ibuclaw:
Replaced with templates.

Can't add unittests because of https://issues.dlang.org/show_bug.cgi?id=17751

ibuclaw:
Marking as blocked. I'll rebase and add unittests so the compiler ICE gets triggered (so that we'll know when dmd is fixed).

@MartinNowak is this fine from a high level design?

@WalterBright don't suppose you can have a look at the backens related ICE?

WalterBright:
I don't know what the rationale of this is. There's no bugzilla issue referenced, no explanation here, no comments in the code?

ibuclaw:
@WalterBright see dmd pr and linked bug there.  I want to make it illegal to implicitly cast to void vector, but these intrinsics block the patch.

JinShil:
> Can't add unittests because of https://issues.dlang.org/show_bug.cgi?id=17751

Bug 17751 was fixed with https://github.com/dlang/dmd/pull/7216.  Is this still "Blocked"?

ibuclaw:
Fix turned my proposal into an error.  If only dmd did constant propagation it would not be an error though...

The imm8 template should instead have the following signature.
```
pure @safe V1 simd(XMM opcode, ubyte imm8, V1, V2)(V1 op1, V2 op2)
    if (is(V1 == __vector) && is(V2 == __vector))
```

ZombineDev:
I've restarted CircleCI, hopefully the failure was intermittent.

wilzbach:
> I've restarted CircleCI, hopefully the failure was intermittent.

Probably caused by a very old config file. I rebased to `upstream/master`. This typically helps for old PRs.

ibuclaw:
_Raises Eyebrows_.

I think I will have to go through some of the tests though to make sure they are OK.  I don't recall whether coverage is complete here, I can see a couple commented out tests.

I guess I can switch over the XMM test file in dmd to use these templates and see what blows up. :-)

PR #1903, opened 2017-08-17
two fixups for the arrayop implementation

MartinNowak:
- add support for mixed vector types
- fix argument index in initScalarVecs

### DIFF
--- a/src/core/internal/arrayop.d
+++ b/src/core/internal/arrayop.d
@@ -297,16 +297,17 @@ string scalarExp(Args...)()
 // `args` to contain operand values.
 string initScalarVecs(Args...)()
 {
-    size_t scalarsIdx;
+    size_t scalarsIdx, argsIdx;
     string res;
-    foreach (aidx, arg; Args)
+    foreach (arg; Args)
     {
         static if (is(arg == T[], T))
         {
+            ++argsIdx;
         }
         else static if (is(arg))
             res ~= "immutable vec scalar" ~ scalarsIdx++.toString ~ " = args["
-                ~ aidx.toString ~ "];\n";
+                ~ argsIdx++.toString ~ "];\n";
     }
     return res;
 }
@@ -318,7 +319,7 @@ string vectorExp(Args...)()
 {
     size_t scalarsIdx, argsIdx;
     string[] stack;
-    foreach (i, arg; Args)
+    foreach (arg; Args)
     {
         static if (is(arg == T[], T))
             stack ~= "load(&args[" ~ argsIdx++.toString ~ "][pos])";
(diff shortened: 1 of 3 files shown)


src/core/internal/arrayop.d
```diff
@@ -224,7 +283,7 @@ string scalarExp(Args...)()
         }
         else static if (isBinaryOp(arg))
         {
-            stack[$ - 2] = "(cast(T)(" ~ stack[$ - 2] ~ " " ~ arg ~ " " ~ stack[$ - 1] ~ "))";
```
MartinNowak: The existing arrayop behavior only casts the whole RHS before assignment, but not intermediate expressions. This is fixed here.

src/core/internal/arrayop.d
```diff
@@ -302,6 +362,25 @@ template not(alias tmlp)
 {
     enum not(Args...) = !tmlp!Args;
 }
+/// find index for which pred!haystack[index] is true or return -1
```
WalterBright: Please use Ddoc style comments, with `Params:` and `Returns:` sections. It's not just for readability, it's so tools can be used on the documentation.

src/core/internal/traits.d
```diff
@@ -210,3 +210,23 @@ template Filter(alias pred, TList...)
                 Filter!(pred, TList[$/2 ..  $ ]));
     }
 }
+
+// std.meta.staticMap
```
WalterBright: Looking at the documentation for std.meta.staticMap, all it says is:
```
Evaluates to $(D AliasSeq!(F!(T[0]), F!(T[1]), ..., F!(T[$ - 1]))).
```
but that is not what this template body does.

src/core/internal/arrayop.d
```diff
@@ -224,7 +283,7 @@ string scalarExp(Args...)()
         }
         else static if (isBinaryOp(arg))
         {
-            stack[$ - 2] = "(cast(T)(" ~ stack[$ - 2] ~ " " ~ arg ~ " " ~ stack[$ - 1] ~ "))";
+            stack[$ - 2] = "(" ~ stack[$ - 2] ~ " " ~ arg ~ " " ~ stack[$ - 1] ~ ")";
```
WalterBright: Any way this and the following red lines can be turned to green?

MartinNowak: No, it's CTFE only functions.

src/core/internal/traits.d
MartinNowak: It's exactly what this template does, in fact I've copied the implementation.
We still need to rename TypeTuple to AliasSeq in this module though.

src/core/internal/arrayop.d
MartinNowak: Done.
Note that this already was a ddoc comment and also it's just a private internal helper template with only 5 lines of implementation.

MartinNowak:
Done, thanks for the review @WalterBright.

PR #1905, opened 2017-08-19
remove emulated TLS for OSX

MartinNowak:
- native TLS supported since 2.072.0


ZombineDev:
CyberShadow/DAutoTest fails due to https://github.com/dlang/dlang.org/pull/1796.

jacob-carlborg:
@MartinNowak 32bit is still using emulated TLS (unless someone recently changed that).

MartinNowak:
Yes, smelled a bit as if 32-bit was still relying on this. Any reference to the OSX TLS implementation? Can't find any in https://github.com/dlang/dmd/pull/5346, but I guess you didn't reverse engineer all of it @jacob-carlborg.
I found https://stackoverflow.com/a/29929949/2371032 which mentions TLS support in XCode 8.

MartinNowak:
> Decide what to do about OSX 32bit.

Keep as long as it's not much to keep.

jacob-carlborg:
> Any reference to the OSX TLS implementation? 

There's no official reference, as far as I know. What I found out is documented here [1][2].

[1] https://github.com/dlang/dmd/blob/master/src/ddmd/toobj.d#L1212
[2] https://github.com/dlang/dmd/blob/master/src/ddmd/backend/el.c#L1285

jacob-carlborg:
TLS was supported before Xcode 8. `__thread` was supported but not `thread_local`, IIRC.

MartinNowak:
> There's no official reference, as far as I know. What I found out is documented here [1][2].

Unfortunately doesn't document from what you derived this. Guess from their runtime linker and llvm/clang, right?

jacob-carlborg:
> Unfortunately doesn't document from what you derived this. Guess from their runtime linker and llvm/clang, right?

From dyld and looking at the assembly and object dumps from Clang.

ZombineDev:
Darwin_64_32 fails consistently on autotester, so it looks we can't move forward with this PR. Please reopen, if I'm misunderstanding.

PR #1906, opened 2017-08-21
provide missing Ddoc for object.ModuleInfo

WalterBright:
It's notably missing from http://dlang.org/phobos/object.html as the existing Ddoc comment was malformed.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -1415,12 +1415,7 @@ class TypeInfo_Inout : TypeInfo_Const
     }
 }
 
-
-///////////////////////////////////////////////////////////////////////////////
-// ModuleInfo
-///////////////////////////////////////////////////////////////////////////////
-
-
+// Contents of Moduleinfo._flags
 enum
 {
     MIctorstart  = 0x1,   // we've started constructing it
@@ -1439,10 +1434,15 @@ enum
     MIname       = 0x1000,
 }
 
-
+/*****************************************
+ * An instance of ModuleInfo is generated into the object file for each compiled module.
+ *
+ * It provides access to various aspects of the module.
+ * It is not generated for betterC.
+ */
 struct ModuleInfo
 {
-    uint _flags;
+    uint _flags; // MIxxxx
     uint _index; // index into _moduleinfo_array[]
 
     version (all)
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -1526,41 +1526,73 @@ const:
 
     @property uint flags() nothrow pure @nogc { return _flags; }
 
+    /************************
+     * Returns:
+     *  module constructor for thread locals, `null` if there isn't one
```
wilzbach: Isn't the convention to start with uppercase here?

src/object.d
```diff
@@ -1571,6 +1603,10 @@ const:
         return null;
     }
 
+    /****************
+     * Returns:
+     *  array of TypeInfo_Class references for classes defined in this module
```
JinShil: "TypeInfo_Class" should probably be a link.  For example `$(DDLINK	phobos/object, TypeInfo_Class,
$(D object.TypeInfo_Class))`

src/object.d
```diff
@@ -1526,41 +1526,73 @@ const:
...
+     *  module destructor for thread locals, `null` if there isn't one
+     */
     @property void function() tlsdtor() nothrow pure @nogc
     {
         return flags & MItlsdtor ? *cast(typeof(return)*)addrOf(MItlsdtor) : null;
     }
 
+    /*****************************
+     * Returns:
+     *  address of a module's `const(MemberInfo)[] getMembers(string)` function, `null` if there isn't one
```
JinShil: I'd love to be able to read more about the `MemberInfo` type, but I can't find any information about anywhere D's documentation.

src/object.d
WalterBright: There is no convention. In any case, non-sentences shouldn't be capitalized nor end in periods.

src/object.d
WalterBright: Generally don't do that for symbols in the same file.

src/object.d
```diff
@@ -1526,41 +1526,73 @@ const:
...
+     *  module order independent constructor, `null` if there isn't one
+     */
     @property void function() ictor() nothrow pure @nogc
     {
         return flags & MIictor ? *cast(typeof(return)*)addrOf(MIictor) : null;
     }
 
+    /*************
+     * Returns:
+     *  address of function that runs the module's unittests, `null` if there isn't one
```
MetaLang: Ditto.

Also this is inconsistent, as you spelled out the function signature of the address of the function returned by `xgetMembers`, but you omitted it in this case. Please either add the same here or remove it from the documentation of `xgetMembers`.

src/object.d
```diff
@@ -1526,41 +1526,73 @@ const:
...
+     */
     @property void function() unitTest() nothrow pure @nogc
     {
         return flags & MIunitTest ? *cast(typeof(return)*)addrOf(MIunitTest) : null;
     }
 
+    /****************
+     * Returns:
+     *  array of pointers to the ModuleInfo's of modules imported by this one
```
MetaLang: You didn't mention that this function can return null.

Again, brevity hurts clarity in this case. You should probably start the sentence with "_an_ array of pointers".

Please add a particle to specify which "modules" are being referred to ("_the_ modules", "_all_ modules", "_any_ module", etc.).

Also recommended: "...modules imported by this one" -> "...modules imported by this module". Clarity, explicitness and uniformity are more valuable than brevity when it comes to documentation.

Finally, "ModuleInfo's" is bad style and should be written as ModuleInfos in the plural case (you can write it as `ModuleInfo`s to make it clear that the 's' is only for marking plurality and not part of the symbol name).

src/object.d
MetaLang: "_an_ array"
"_the_ classes"

src/object.d
```diff
@@ -1581,6 +1617,10 @@ const:
         return null;
     }
 
+    /********************
+     * Returns:
+     *  name of module, `null` if no name
```
MetaLang: Add an 'or', 'otherwise', etc. after the comma.

MetaLang: Also, "_the_ name of _the_ module"

src/object.d
kuettler: I feel many of these requests do not improve the quality. This might be a case that warrants a quick merge. I am all for it.

src/object.d
WalterBright: What I wrote is normal prose for such comments. Making such changes does not improve it.

src/object.d
WalterBright: These are sentence fragments, not complete sentences. They are clear and appropriate for this use as is.

src/object.d
WalterBright: It is not inconsistent. The returned function signature here is spelled out by the signature of the function. Repeating it in the Ddoc comment is redundant. The `xgetMembers` return pointer is specified as `void*` so pointing out what it actually is is appropriate in the Ddoc comment for it.

src/object.d
WalterBright: > You didn't mention that this function can return null.

Didn't need to, as an array of 0 length is null.

`modules imported by this one` seems clear enough without further qualification

I use the ' to distinguish the identifier from the plural, which is common practice for code. Highlighting it instead looks awkward to me.

PR #1909, opened 2017-08-25
fix reference escape

WalterBright:
Passing static array `buf` to `new` causes the message:
```
src/gc/impl/conservative/gc.d(3252): Error: returning `&buf` escapes a reference to local variable `buf`
```
with PR https://github.com/dlang/dmd/pull/7101 which it blocks.

While the allocated memory is not actually escaped, the destructor has access to the expired stack frame for `buf`, and so needs to be corrected.

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -3236,7 +3236,7 @@ unittest // bugzilla 15822
 {
     import core.memory : GC;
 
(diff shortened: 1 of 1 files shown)


ZombineDev:
> Error: returning `&buf` escapes a reference to local variable `buf`

We definitely need to improve the error message, as it is quite confusing, since there's not a single `return` statement in the whole unit test.

PR #1910, opened 2017-08-27
fix issue 17788: MSCOFF: TLS broken with VS2017 15.3.1

rainers:
with VS2017 15.3.1 segment .tls is no longer written to a separate image section, so _tls_start and _tls_end no longer happen to work as both __gshared and TLS section offsets.

### DIFF
--- a/src/rt/sections_win64.d
+++ b/src/rt/sections_win64.d
@@ -112,7 +112,12 @@ void[] initTLSRanges() nothrow @nogc
 {
     void* pbeg;
     void* pend;
-    version(Win32)
+    // with VS2017 15.3.1, the linker no longer puts TLS segments into a
+    //  separate image section. That way _tls_start and _tls_end no
+    //  longer generate offsets into .tls, but DATA.
+    // Use the TEB entry to find the start of TLS instead and read the
+    //  length from the TLS directory
+    version(D_InlineAsm_X86)
     {
         asm @nogc nothrow
         {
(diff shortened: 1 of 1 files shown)


ZombineDev:
@rainers how confident are you with this solution, should we try to include it the next beta of 2.076.0?

src/rt/sections_win64.d
```diff
@@ -108,12 +108,40 @@ void finiSections() nothrow @nogc
...
-    auto pend = cast(void*)&_tls_end;
-    return pbeg[0 .. pend - pbeg];
-}
+void[] initTLSRanges() nothrow @nogc
+{
+    void* pbeg;
+    void* pend;
+    version(Win32)
+    {
+        asm @nogc nothrow
```
ibuclaw: Shouldn't this be guarded by D_InlineAsm too?

rainers:
> @rainers: Does this affect LDC as well? It seems like we are using the same code.

Yes, see https://github.com/ldc-developers/ldc/pull/2279#issuecomment-323512428

> @rainers how confident are you with this solution, should we try to include it the next beta of 2.076.0?

Should work for all VC at least back to VS2005, they all use _tls_used for the TLS directory. I'll retarget stable if it passes. There are also other issues with this VS update that break Visual D, though.

> Shouldn't this be guarded by D_InlineAsm too?

Yeah, probably better. I replaced Win32 with D_InlineAsm_X86, etc.

ZombineDev:
> Should work for all VC at least back to VS2005, they all use _tls_used for the TLS directory. I'll retarget stable if it passes. There are also other issues with this VS update that break Visual D, though.

Great to hear. I've added this to the 2.076.0 milestone. CC @MartinNowak

rainers:
rebased to stable.

kinke:
We'll 'need' this for LDC 1.4 final.

PR #1911, opened 2017-09-01
Fix a static if (__USE_MISC) section in core.sys.linux.netinet.in_.

jmdavis:
Apparently, I made it cover more than it was supposed to. I think that
the fact that the #endif ending that section in the header file was
right after a #define threw me off, and I incorrectly thought that
the #endif went with the #define even though that doesn't make sense. In
either case, this should fix my mistake.

And man does the way that github show this diff look weird. All it does is add a brace in order to end a static if earlier, remove the brace that used to end the static if, and then fix the indentation. :|

### DIFF
--- a/src/core/sys/linux/netinet/in_.d
+++ b/src/core/sys/linux/netinet/in_.d
@@ -162,62 +162,62 @@ version(CRuntime_Glibc)
         {
             return group_filter.sizeof - sockaddr_storage.sizeof + numsrc * sockaddr_storage.sizeof;
         }
+    }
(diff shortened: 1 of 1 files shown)


ZombineDev:
> And man does the way that github show this diff look weird. All it does is add a brace in order to end a static if earlier, remove the brace that used to end the static if, and then fix the indentation. :|

Adding `?w=1` to the end of the URL is your friend - https://github.com/dlang/druntime/pull/1911/files?w=1

PR #1914, opened 2017-09-02
core.stdc.stdio - change stdout, etc., to enum for -betterC

WalterBright:
Having `stdout`, etc., be a variable means druntime needs to be linked in, meaning they do not work for `-betterC`. Switching to an enum makes it work.

### DIFF
--- a/src/core/stdc/stdio.d
+++ b/src/core/stdc/stdio.d
@@ -705,15 +705,15 @@ version( CRuntime_DigitalMars )
     private extern shared FILE[_NFILE] _iob;
 
     ///
-    shared stdin  = &_iob[0];
+    enum stdin  = &_iob[0];
     ///
-    shared stdout = &_iob[1];
+    enum stdout = &_iob[1];
     ///
-    shared stderr = &_iob[2];
+    enum stderr = &_iob[2];
     ///
-    shared stdaux = &_iob[3];
+    enum stdaux = &_iob[3];
     ///
-    shared stdprn = &_iob[4];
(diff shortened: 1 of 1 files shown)


WalterBright:
> Doesn't this break &stdin

Given the vagaries of how `stdin` is implemented in various C implementations (none of them are the same), and code relying on C `stdin` being an lvalue is not going to be portable anyway. Sometimes the implementations are a function, not a variable, where taking the address won't give what you presumably want.

> (and compatibility with other runtimes)?

No, this one is specific to the Digital Mars C runtime.

Note that this change is necessary to make betterC code able to access `stdin` at all. In general, `core.stdc` implementations should try to find ways that do not require linking in druntime.

schveiguy:
ping @CyberShadow any ideas on the doc build failure? They look internal to the build system itself, but I'm not 100% sure.

wilzbach:
> They look internal to the build system itself, but I'm not 100% sure.

I think this has been fixed one or two days ago. Unfortunately the only way to rekick DAutoTest is to rebase (unfortunately I can't help here as I'm without a computer atm).

CyberShadow:
@schveiguy It's green now so I'm not sure what the error was, but judging from other build failures in the past few days, it was almost certainly due to code.dlang.org being down.

schveiguy:
> I'm not sure what the error was

This was the failed build: http://dtest.dlang.io/results/209be21a8f62782bb5b8f66edc7f9f8d62e1c439/2b5450cb012263aa23aad9c24f9528adefbba887/

PR #1915, opened 2017-09-08
Fix issue 17952 - destroy() and @nogc

Biotronic:
The current version of `destroy()` is not marked `@nogc` even when the class to be destroyed has a destructor with that attribute. This code checks if the type passed has a `@nogc` destructor, and does a cast to get the correct `@nogc` attribute on `destroy`.

This requires that the destructor be explicitly marked `@nogc`, and thus that classes with empty destructors explicitly implement one.

While it seems logical to also make `destroy()` `@nogc` when no destructor is defined, this fails when a subclass implements a non-`@nogc` destructor.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2790,7 +2790,54 @@ unittest
   +/
 void destroy(T)(T obj) if (is(T == class))
 {
-    rt_finalize(cast(void*)obj);
+    static if (__traits(hasMember, obj, "__dtor"))
+    {
+        static if (contains("@nogc", [__traits(getFunctionAttributes, typeof(__traits(getMember, obj, "__dtor")))]))
+        {
+            auto cleanup = cast(void function(void *, bool) @nogc @system)&rt_finalize;
+            cleanup(cast(void*)obj, true);
+        }
+        else
+        {
+            rt_finalize(cast(void*)obj);
+        }
+    }
+    else
+    {
+            rt_finalize(cast(void*)obj);
+    }
+}
+
+version(unittest) unittest
+{
+    class C {
+        @nogc
+        ~this()
+        {
+        }
+    }
+    class D {
+    }
+
+    @nogc
+    void cleanup(T)(T t) {
+        destroy(t);
+    }
+
+    assert(__traits(compiles, cleanup(new C())));
+    assert(!__traits(compiles, cleanup(new D())));
+}
+
+bool contains(T)(T needle, T[] haystack)
+{
+    foreach (e; haystack)
+    {
+        if (e == needle)
+        {
+            return true;
+        }
+    }
+    return false;
 }
 
 /// ditto


src/object.d
```diff
@@ -2790,7 +2790,54 @@ unittest
   +/
 void destroy(T)(T obj) if (is(T == class))
 {
-    rt_finalize(cast(void*)obj);
+    static if (__traits(hasMember, obj, "__dtor"))
```
BBasile: That cant work properly. destroy can be called for a class that has a nogc dtor but after downcasting.

Biotronic: I thought at first you meant that `destroy()` might not be marked `@nogc` when a subclass is `@nogc`. This is by design, and the only possible way to do it. I then realized you might mean this:
```
class C {
    @nogc ~this() {}
    @nogc void foo() {}
}

class D : C {
    ~this() {}
    override void foo() {}
}

// tuple("@nogc", "@system")
pragma(msg, __traits(getFunctionAttributes, __traits(getMember, C, "__dtor")));
// tuple("@system")
pragma(msg, __traits(getFunctionAttributes, __traits(getMember, D, "__dtor")));
// tuple("@nogc", "@system")
pragma(msg, __traits(getFunctionAttributes, C.foo));
// tuple("@nogc", "@system")
pragma(msg, __traits(getFunctionAttributes, D.foo));
```
Clearly this is a hole in the type system - if a base function is marked `@nogc`, derived functions must be as well (and are, in the case of regular virtual functions).

Does this invalidate the PR? I disagree.

BBasile: class destructors are not virtual. That's a part of the problem. destructors are found and called using the TypeInfo_Class, not by inheritance, so

> derived functions must be as well

is unfortunately wrong. A real fix would involve huge changes. Try to work more on the problem and you'll see ;)

Biotronic: I love being wrong - it means I get to learn something! :D

Looking at `rt_finalize2`, it seems the only issue is destructors - monitor deletion doesn't touch the GC. In other words, if `@nogc` on destructors were inherited, we could do the test I do in this PR, and it would give correct results (if a `@nogc` destructor is defined, it's safe to assume that all derived destructors are `@nogc`, in all other cases that's an unsafe assumption**). Of course, since it's not, it won't.

There may very well be something I've missed - are there any discussions on this topic that you can recommend? It seems weird to me that destructor attributes aren't considered part of the public interface of a class, but I can accept there being reasons for it - I just don't know what they are.

** Actually, for final classes with no destructor, it would also be safe.

BBasile: >  are there any discussions on this topic that you can recommend?

No.  i see this problem as a dead end. In first place constructor and destructor not being virtual is a fundamental design error imo. likely unfixable.

Biotronic: There is actually a subset of classes for which destroy can safely be `@nogc` today: final classes where all fields and base classes define `@nogc` destructors or no destructors at all.

PR #1916, opened 2017-09-14
Add --DRT-covopt: merge, srcpath and dstpath options

leandro-lucarella-sociomantic:
```
    Make coverage reports configurable via --DRT-covopt
    
    The coverage module has a couple of options that can be set through some
    weirdly looking dmd_coverSetMerge(), dmd_coverSourcePath() and
    dmd_coverDestPath() functions (I guess the dmd prefix comes from
    pre-druntime times).
    
    These options don't only use non standard names, they also don't use the
    standard --DRT-xxx / DRT_XXX runtime options passing mechanism, and they
    can only be overridden in code.
    
    This commit plugs the runtime options mechanism and adds support to set
    these options via --DRT-covopt (and matching environment variables),
    providing the sub-options merge, dstpath and srcpath.
    
    The old functions are kept for backwards compatibility (they are used by
    DMD itself for example), but should probably be removed in some distant
    future.
```

### DIFF
--- a/changelog/covopts.dd
+++ b/changelog/covopts.dd
@@ -1,0 +1,13 @@
+The runtime learned a few new options to customize how coverage reports are created.
+
+The format is the same as with the GC but using `covopt` as suffix, for example
+you can pass options separating them with spaces like `--DRT-covopt "merge:1
+dstpath:/tmp"`.
+
+These are the currently accepted options:
+
+$(DL
+  $(DT merge) $(DD Merge the current run with existing reports if 1, or overwrite the existing reports if 0.)
+  $(DT srcpath) $(DD Set path to where source files are located.)
+  $(DT dstpath) $(DD Set path to where listing files are to be written.)
+)
--- a/src/rt/cover.d
+++ b/src/rt/cover.d
@@ -24,7 +24,6 @@ private
     import core.stdc.stdio;
     import core.stdc.stdlib;
     import rt.util.utf;
-    import core.internal.parseoptions;
 
     struct BitArray
     {
@@ -71,17 +70,18 @@ private
 
         bool initialize()
         {
+            import core.internal.parseoptions : initConfigOptions;
             return initConfigOptions(this, this.errorName);
         }
 
         void help()
         {
-            string s = "Code coverage options are specified as white space separated assignments:
+            string s = "Code coverage options are specified as whitespace separated assignments:
     merge:0|1      - 0 overwrites existing reports, 1 merges current run with existing coverage reports (default: %d)
-    dstpath:<PATH> - write code coverage reports to <PATH> (default: current
+    dstpath:<PATH> - writes code coverage reports to <PATH> (default: current
             working directory)
-    srcpath:<PATH> - set the path where the source files are located to <PATH>
-    (defaul: current working directory)
+    srcpath:<PATH> - sets the path where the source files are located to <PATH>
+    (default: current working directory)
 ";
             printf(s.ptr, merge);
         }
--- a/src/rt/cover.d
+++ b/src/rt/cover.d
@@ -24,6 +24,7 @@ private
     import core.stdc.stdio;
     import core.stdc.stdlib;
     import rt.util.utf;
+    import core.internal.parseoptions;
 
     struct BitArray
     {
@@ -57,9 +58,35 @@ private
     __gshared
     {
         Cover[] gdata;
+        Config config;
+    }
+
+    struct Config
+    {
         string  srcpath;
         string  dstpath;
         bool    merge;
+
+    @nogc nothrow:
+
+        bool initialize()
+        {
+            return initConfigOptions(this, this.errorName);
+        }
+
+        void help()
+        {
+            string s = "Code coverage options are specified as white space separated assignments:
+    merge:0|1      - 0 overwrites existing reports, 1 merges current run with existing coverage reports (default: %d)
+    dstpath:<PATH> - write code coverage reports to <PATH> (default: current
+            working directory)
+    srcpath:<PATH> - set the path where the source files are located to <PATH>
+    (defaul: current working directory)
+";
+            printf(s.ptr, merge);
+        }
+
+        string errorName() { return "covopt"; }
     }
 }
 
@@ -72,7 +99,7 @@ private
  */
 extern (C) void dmd_coverSourcePath( string pathname )
 {
-    srcpath = pathname;
+    config.srcpath = pathname;
 }
 
 
@@ -84,7 +111,7 @@ extern (C) void dmd_coverSourcePath( string pathname )
  */
 extern (C) void dmd_coverDestPath( string pathname )
 {
-    dstpath = pathname;
+    config.dstpath = pathname;
 }
 
 
@@ -97,7 +124,7 @@ extern (C) void dmd_coverDestPath( string pathname )
  */
 extern (C) void dmd_coverSetMerge( bool flag )
 {
-    merge = flag;
+    config.merge = flag;
 }
 
 
@@ -149,6 +176,11 @@ uint parseNum(const(char)[] s)
 T min(T)(T a, T b) { return a < b ? a : b; }
 T max(T)(T a, T b) { return b < a ? a : b; }
 
+shared static this()
+{
+    config.initialize();
+}
+
 shared static ~this()
 {
     if (!gdata.length) return;
@@ -161,14 +193,14 @@ shared static ~this()
 
     foreach (c; gdata)
     {
-        auto fname = appendFN(dstpath, addExt(baseName(c.filename), "lst"));
+        auto fname = appendFN(config.dstpath, addExt(baseName(c.filename), "lst"));
         auto flst = openOrCreateFile(fname);
         if (flst is null)
             continue;
         lockFile(fileno(flst)); // gets unlocked by fclose
         scope(exit) fclose(flst);
 
-        if (merge && readFile(flst, buf))
+        if (config.merge && readFile(flst, buf))
         {
             splitLines(buf, lines);
 
@@ -176,7 +208,7 @@ shared static ~this()
                 c.data[i] += parseNum(line);
         }
 
-        if (!readFile(appendFN(srcpath, c.filename), buf))
+        if (!readFile(appendFN(config.srcpath, c.filename), buf))
             continue;
         splitLines(buf, lines);
 
--- a/src/gc/config.d
+++ b/src/gc/config.d
@@ -33,7 +33,7 @@ struct Config
 
     void help() @nogc nothrow
     {
-        string s = "GC options are specified as white space separated assignments:
+        string s = "GC options are specified as whitespace separated assignments:
     disable:0|1    - start disabled (%d)
     profile:0|1|2  - enable profiling with summary when terminating program (%d)
     gc:conservative|manual - select gc implementation (default = conservative)


src/rt/cover.d
```diff
@@ -149,10 +149,80 @@ uint parseNum(const(char)[] s)
 T min(T)(T a, T b) { return a < b ? a : b; }
 T max(T)(T a, T b) { return b < a ? a : b; }
 
+version (unittest) __gshared bool inUnittest;
+
+string parseMergeOpt(string opt) @nogc nothrow
```
nemanja-boric-sociomantic: Care to add some documentation? :-)

leandro-lucarella-sociomantic:
Added changelog entry.

src/rt/cover.d
leandro-lucarella-sociomantic: Anywhere besides the changelog where this can be documented?

src/rt/cover.d
```diff
@@ -149,10 +149,80 @@ uint parseNum(const(char)[] s)
...
+    switch (opt)
+    {
+        case "no", "No", "NO", "n", "N", "0":
+            merge = false;
+            return "0";
+        case "", "yes", "Yes", "YES", "y", "Y", "1":
+            merge = true;
+            return "1";
+        default:
+            if (!inUnittest)
```
nemanja-boric-sociomantic: Given this here, this would not compile outside unittests, right? Why not `version (unittest)` on this method?

src/rt/cover.d
```diff
@@ -149,10 +149,80 @@ uint parseNum(const(char)[] s)
...
+    rt_configOption("covmerge", &parseMergeOpt);
+    rt_configOption("covsrcpath", (string s) { srcpath = s; });
+    rt_configOption("covdstpath", (string s) { dstpath = s; });
+}
+
 shared static ~this()
 {
     if (!gdata.length) return;
 
+    updateDRTOpts();
```
nemanja-boric-sociomantic: Weird that `updateDRTOpts` is called in the desctructor, but I guess you don't need it before.

changelog/covopts.dd
```diff
@@ -0,0 +1,7 @@
+The runtime learned a few new options to customize how coverage reports are created:
+
+$(DL
+  $(DT covmerge) $(DD Set merge mode; either N/n/0 or Y/y/1.)
+  $(DT covsrcpath) $(DD Set path to where source files are located.)
+  $(DT covdstpath) $(DD Set path to where listing files are to be written.)
```
mathias-lang-sociomantic: I would make the usage a bit more explicit. E.g. first line:
>  $(DT --DRT-covmerge=Y) $(DD Set merge mode; Accepts N/n/0 or Y/y/1. Default to Y.)

Also for this specific part (covmerge), it might be more consistent to use {false,0} / {true,1}

src/rt/cover.d
leandro-lucarella-sociomantic: Yeah, all is done in the destructor, and I thought adding a constructor to do the setup when it's never used before the destructor was adding too much complications for nothing (and adding a constructor means we need to care about ordering issues with the previous functions to set the options).

src/rt/cover.d
leandro-lucarella-sociomantic: What do you mean? Why wouldn't it compile outside unittests?

src/rt/cover.d
```diff
@@ -149,10 +149,83 @@ uint parseNum(const(char)[] s)
...
+        case "no", "No", "NO", "n", "N", "0":
+            merge = false;
+            return "0";
+        case "", "yes", "Yes", "YES", "y", "Y", "1":
+            merge = true;
+            return "1";
+        default:
+            if (!inUnittest)
+                fprintf(stderr, "DRT Error: Runtime option 'cov-merge' should "
+                        "be 1/y/yes or 0/n/no, got '%.*s'.\n",
```
mathias-lang-sociomantic: No implicit string concatenation please.

changelog/covopts.dd
leandro-lucarella-sociomantic: I'm copying how `gcopt` parse options (more or less).

src/rt/cover.d
nemanja-boric-sociomantic: Am I not seeing right, or the variable `inUnittest` is only defined if `version (unittest)` (see few lines above)?

src/rt/cover.d
leandro-lucarella-sociomantic: Argh! I saw that and thought I had fixed it already.

src/rt/cover.d
mathias-lang-sociomantic: `private` also

src/rt/cover.d
leandro-lucarella-sociomantic: Updated.

changelog/covopts.dd
leandro-lucarella-sociomantic: I was just about to try to generalize how the `gc` parse the options, but it was a much bigger task and I'm not sure I fully understand the restrictions on druntime to do it properly (for example, I don't know why this `rt_configOption` is not just imported normally instead of using this `externDFunc` thing), and then I didn't know if it will get accepted.

At some point I think it would be nice to have a generalized option parsing mechanism like the one in `src/gc/config.d`.

src/rt/cover.d
nemanja-boric-sociomantic: I'm not the only one, travis is also complaining here :-)

src/rt/cover.d
```diff
@@ -149,10 +149,83 @@ uint parseNum(const(char)[] s)
 T min(T)(T a, T b) { return a < b ? a : b; }
 T max(T)(T a, T b) { return b < a ? a : b; }
 
+version (unittest) __gshared bool inUnittest;
```
mathias-lang-sociomantic: Meh this global is truly ugly. Why not make it a parameter to the function instead ?

src/rt/cover.d
leandro-lucarella-sociomantic: Right, I see it now. Copy&paste error :)

leandro-lucarella-sociomantic: Updated.

src/rt/cover.d
nemanja-boric-sociomantic: Ok, there is camp that says that `private` functions doesn't need a documentation, but I wouldn't mind just a single line comment above this function to says what it is doing.

src/rt/cover.d
leandro-lucarella-sociomantic: Also taken from `src/gc/config.d`. I guess in this case an argument can be passed, although to be honest, if the problem here is ugliness, I don't think passing an extra pointless argument in every call to `parseMergeOpts` will add any beauty.

nemanja-boric-sociomantic: You could avoid it completely and then just in the unittests do this:

```
   assert(parseMergeOpt("yEs")[0.."DRT Error".length] == "DRT Error" && !merge);
```

nemanja-boric-sociomantic: Whoops, that's wrong, since you're printing to `stderr`, not returning this.

mathias-lang-sociomantic: Yep that was my first thought as well. But here you are declaring a public global used as a flag for a single function from a single calling point. How else can I say that a global is one of the worst way to do it, unless you're a C programmer from the 60's ?

src/rt/cover.d
leandro-lucarella-sociomantic: Ah, OK.

src/rt/cover.d
leandro-lucarella-sociomantic: Why is the worst way to do it. I don't see any real technical reason. Just saying something is ugly or old is not good enough for a reason to me. Tell me why is it bad.

leandro-lucarella-sociomantic: To explain a bit more, I know many ways in which globals could be bad. I will tell you why I think a global is good in this case:

* Avoids polluting the function signature with something that's only useful for unit testing.
* It gets completely eliminated when not unit testing, not only the global but all the code that is used for unit testing.
* It is completely encapsulated, is private to the module and used only when unit testing.

nemanja-boric-sociomantic: Would it be better if you marked this global `private`?

nemanja-boric-sociomantic: > It is completely encapsulated, is private to the module and used only when unit testing.
 
I think this is wrong.

leandro-lucarella-sociomantic:
I'm still getting these errors and I don't understand why, my D-foo is not that great these days:
```
src/rt/cover.d(215): Error: function core.internal.traits.externDFunc!("rt.config.rt_configOption", string function(string, scope string delegate(string) nothrow @nogc, bool) nothrow @nogc, nothrow @nogc string(string, scope string delegate(string) nothrow @nogc, bool)).externDFunc (string, scope string delegate(string) nothrow @nogc, bool) is not callable using argument types (string, string function(string opt) nothrow @nogc @system)
```

The only difference I see between what is passed and expected is the `@system`, but I don't see any `@trusted` or `@safe` in the module defining this callback so I can't see why it should take anything other than the default `@system`.

leandro-lucarella-sociomantic:
Argh, the default bool is not passed :roll_eyes:

src/rt/cover.d
mathias-lang-sociomantic: >  I think this is wrong.

Actually there's a `private:` a couple of lignes before so it's okay-ish.

My reasons mainly boils down to SoC. You mention:

> Avoids polluting the function signature with something that's only useful for unit testing.

But this information *actually* belongs to the function signature ! Because it only affects that function, and not the whole module. When I see a global in a module, I expect it to be used in several places, and I truly hope it's read-only (or at least written-once) because that complicates understanding of the code a great deal.

If you really want something clean, make it take a sink which either does nothing or forward to `fprintf`.

> It gets completely eliminated when not unit testing, not only the global but all the code that is used for unit testing.

Good point though, but does it really matter in that case ?

If you feel strongly about it, I'm totally okay to drop the ball, I was just really surprised by this code.

leandro-lucarella-sociomantic: I'm a bit indifferent, I don't care much either way, I'm just lazy and repeating a practice I already found in the code. But I do think is a simple solution, I don't think it complicates reading the code, is super straight-forward to read `inUnittest` and know what is it about and when is true and when it's false. When I see it in the original code I didn't see anything wrong and it was perfectly clear to me.

I'll wait to see what the druntime maintainers think and change it only they prefer it gone.

leandro-lucarella-sociomantic:
Ping @MartinNowak maybe? :innocent:

leandro-lucarella-sociomantic:
Or @ZombineDev / @wilzbach ...

src/rt/cover.d
```diff
@@ -149,10 +149,84 @@ uint parseNum(const(char)[] s)
...
+}
+
+// Parse options taken from the standard runtime options passing via
+// --DRT-xxx / DRT_XXX.
+void updateDRTOpts()
+{
+    import core.internal.traits : externDFunc;
+    alias rt_configCallBack = string delegate(string) @nogc nothrow;
+    alias fn_configOption = string function(string opt, scope rt_configCallBack dg, bool reverse) @nogc nothrow;
+    alias rt_configOption = externDFunc!("rt.config.rt_configOption", fn_configOption);
```
wilzbach: Why not simply `import rt.config : rt_configOption;`

src/rt/cover.d
wilzbach: > I'll wait to see what the druntime maintainers think and change it only they prefer it gone.

 Another solution would be to define a specific version which is only set during druntime's unittest compilation

src/rt/cover.d
leandro-lucarella-sociomantic: I don't know. I just copied what it's done in `gc/config.d` thinking that there was a good reason for it.

src/rt/cover.d
leandro-lucarella-sociomantic: This is not good enough, because then other tests hitting this code will print to stderr. the version should be only set while this particular unittest is running. AFAIK that can't be done with a `version`, right?

leandro-lucarella-sociomantic:
> FYI: the LDC guys proposed to add this as a compiler option (ldc-developers/ldc#1672), but imho this isn't conflicting and their are valid use cases for which a compiler flag would be simpler (e.g. compiling the binary with coverage reporting enabled once and then running a couple of different tests with it - even DMD does this)), but imho this could be added independently in the future.

For me this really looks more like a runtime option than a compiler option. Where reports are written, if they should be merged or not, and where to find the source code is something that should be set when running the program, not when compiling it. As you said you could want to run the program with different flags and store the reports in different locations each time.

rainers:
> I don't know. I just copied what it's done in gc/config.d thinking that there was a good reason for it.

It's done with externDFunc there to avoid an import dependency from core/gc to the rt package. rt is usually not in the generated import folder so not accessible if other modules import a core module.

For a module in the rt package importing rt.config is fine.

leandro-lucarella-sociomantic:
OK, I'll update to make a simple import then.

leandro-lucarella-sociomantic:
Updated.

ibuclaw:
> This options don't only use non standard names, they also don't use the
standard --DRT-xxx / DRT_XXX runtime options passing mechanism, and they
can only be overridden in code.

This entire module is to support a dmd feature anyway.  No point trying to make it more "standard", no one else is using it.

leandro-lucarella-sociomantic:
I would really appreciate if people could either:

* Approve the PR if they don't think there is anything else to improve/change
* Tell what needs to be done for this PR to get accepted

Thanks :)

src/rt/cover.d
wilzbach: Yes, it can't, but is there any other bit of unittest code at Druntime that should `printf` instead of "err"?

src/rt/cover.d
```diff
@@ -149,10 +149,75 @@ uint parseNum(const(char)[] s)
...
 shared static ~this()
 {
     if (!gdata.length) return;
 
+    // Parse options taken from the standard runtime options passing via
+    // --DRT-xxx / DRT_XXX.
+    import rt.config: rt_configOption;
+    rt_configOption("covmerge", (string opt) => parseMergeOpt(opt), false);
+    rt_configOption("covsrcpath", (string opt) => srcpath = opt, false);
+    rt_configOption("covdstpath", (string opt) => dstpath = opt, false);
```
MartinNowak: The format is a lot different from `DRT-gcopt`.
How about something more consistent?
```sh
-DRT-coverage=help

merge:0|1 - Set whether to merge data with listings from previous runs.
srcpath:path - Set path to where source files are located.
dstpath:path - Set path to where listing files are to be written.
```
So for example `-DRT-coverage='merge:1 srcpath:C:\foo\bar dstpath:outdir'` would be used for all configs.
Also an opportunity to the existing parsing code from `gc/config.d` to `core/internal/drt_config.d` or so.

src/rt/cover.d
MartinNowak: It's a common pattern, e.g. see cpuid.

src/rt/cover.d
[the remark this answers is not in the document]
MartinNowak: Adding a `static this()` wouldn't be much of a problem for an internal `rt.` module, but alternatively feel free to make the DRT parsing more modular and bootstrap it from the rt startup code.

leandro-lucarella-sociomantic: I tried to start like that, but then I found it was quite difficult to abstract out the code from `gcopts` because of this *no imports of rt from gc*. Then also, if there is already mechanisms to easily add `--DTR-xxx` options, why bothering complicating stuff by adding another layer of parsing?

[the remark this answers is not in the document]
leandro-lucarella-sociomantic: > You want to parse those options on startup instead of reporting an error after all tests have been run.

OK, fair point. There is still the ordering issue if someone decides to call the `dmd_cover*()` directly in some other module constructor, but it is a pretty obscure case and then if we could remove those functions in the future, I guess is not worth much to care about it.

Updated.

leandro-lucarella-sociomantic: To be more precise, what I mean is the code in the gc is generic to be able to parse a struct, but I don't know if I can import from the GC those templates if they are moved to the runtime (where they should belong if it will be a standard way to add options). If this can be done, I would need some help because I'm not familiar with it.

Why is this restriction about importing `rt.*` modules inside the GC again?

leandro-lucarella-sociomantic: But I still think is silly to have 2 layers of parsing in options, if we go generic I think it might be better to head for something like `--DRT-gc-disable=1 --DRT-cov-merge=0` and maybe add a `--DRT-help` to get all the options.

I don't think this PR is the place to go for that refactoring thought.

leandro-lucarella-sociomantic:
Ping. Please don't let this die :)

rainers:
> I don't think this PR is the place to go for that refactoring thought.

https://github.com/dlang/druntime/pull/1967 might help.

leandro-lucarella-sociomantic:
:heart_eyes:

Thanks for taking on that! I'll gladly rebase and re-do this PR on top of #1967 once it's merged!

leandro-lucarella-sociomantic:
Adapted to use `core.internal.parseoptions` introduced by https://github.com/dlang/druntime/pull/1967.

leandro-lucarella-sociomantic:
Please re-review :innocent:

wilzbach:
> Thanks for taking on that! I'll gladly rebase and re-do this PR on top of #1967 once it's merged!

@leandro-lucarella-sociomantic #1967 has been merged ;-)

src/rt/cover.d
```diff
@@ -24,6 +24,7 @@ private
     import core.stdc.stdio;
     import core.stdc.stdlib;
     import rt.util.utf;
+    import core.internal.parseoptions;
```
wilzbach: Could be a selective import here: `: initConfigOptions`

src/rt/cover.d
```diff
@@ -57,9 +58,35 @@ private
...
+        }
+
+        void help()
+        {
+            string s = "Code coverage options are specified as white space separated assignments:
+    merge:0|1      - 0 overwrites existing reports, 1 merges current run with existing coverage reports (default: %d)
+    dstpath:<PATH> - write code coverage reports to <PATH> (default: current
+            working directory)
+    srcpath:<PATH> - set the path where the source files are located to <PATH>
+    (defaul: current working directory)
```
wilzbach: default

src/rt/cover.d
```diff
@@ -57,9 +58,35 @@ private
...
+    @nogc nothrow:
+
+        bool initialize()
+        {
+            return initConfigOptions(this, this.errorName);
+        }
+
+        void help()
+        {
+            string s = "Code coverage options are specified as white space separated assignments:
```
wilzbach: [Whitespace](https://en.wikipedia.org/wiki/Whitespace_character)

src/rt/cover.d
```diff
@@ -57,9 +58,35 @@ private
...
+    merge:0|1      - 0 overwrites existing reports, 1 merges current run with existing coverage reports (default: %d)
+    dstpath:<PATH> - write code coverage reports to <PATH> (default: current
```
wilzbach: You use the third person for the explanation of the merge option - consistency would be great.

src/rt/cover.d
```diff
@@ -57,9 +58,35 @@ private
...
+            working directory)
+    srcpath:<PATH> - set the path where the source files are located to <PATH>
```
wilzbach: ditto about consistency

src/rt/cover.d
ZombineDev: If we go for a selective import, we should go all the way and make it local.
Though I'm fine with keeping as it is, for consistency.

leandro-lucarella-sociomantic: That was exactly my take, the current style is an old style, I wasn't sure if having a mixed style is better. If it's a requirement for anyone, I can change it.

src/rt/cover.d
leandro-lucarella-sociomantic: Do you mean I should write "whitespace" instead of "white space"? This was a copy&paste from the gc code, so if that's so that should be fixed too :)

leandro-lucarella-sociomantic:
Updated.

@MartinNowak ?

leandro-lucarella-sociomantic:
> Looks great! A couple of nits and I think we should document this on dlang.org too. For the GC, there's https://dlang.org/spec/garbage.html#gc_config, but there's an equivalent page for coverage in the D spec (in fact coverage is only mentioned twice). There's an article about Code Coverage on dlang.org, but it's well hidden, so how about creating a new page which is solely about listing all settings of --DRT?

Sorry if I'm being excessively lazy, but can you point me to the source code of the files I should update?

src/rt/cover.d
leandro-lucarella-sociomantic: Updated the GC help too.

leandro-lucarella-sociomantic:
Also, is a bit shameful, but as I'm not using myself D2 yet, this is untested so far, so before merging it would be good if someone can test it. I'll try to do some testing myself, but it will probably better if people that use code coverage daily can test this too.

wilzbach:
> Sorry if I'm being excessively lazy, but can you point me to the source code of the files I should update?

Oh I was more thinking about creating a `spec/druntime` page which will provide such an overview or something similar like this, but I wasn't sure either on the best place for this. Anyhow this shouldn't block this PR, I was just pointing out that there's no overview page, which lists all the options configurable via `--DRT`.

leandro-lucarella-sociomantic:
https://github.com/dlang/dlang.org/pull/1945. I think a new doc about druntime config would be awesome but it is beyond the scope of this PR.

rainers:
I guess @MartinNowak's concerns have also been addressed, so if there are no more objections, we can merge this.
There is also a related bug report (https://issues.dlang.org/show_bug.cgi?id=17300), you might want to add "fix issue 17300" to some commit message.

src/rt/cover.d
wilzbach: Well, we want to gradually update the style and at least all new additions should try to avoid old-school global imports, but I can change this later today myself and then this PR should finally be good to go.

wilzbach:
> Well, we want to gradually update the style and at least all new additions should try to avoid old-school global imports, but I can change this later today myself and then this PR should finally be good to go.

Done.

> There is also a related bug report (https://issues.dlang.org/show_bug.cgi?id=17300), you might want to add "fix issue 17300" to some commit message.

Done.

jondegenhardt:
This is great! I'm looking forward to removing the custom code I had write to aggregate code coverage reports for Travis-CI and CodeCov. Thanks!

PR #1918, opened 2017-09-17
Fix leaking memory in shared druntime

Calrama:
When built as a shared library, druntime allocates memory for storing DSO dependencies in a custom dynamic array via [`_d_dso_registry`](https://github.com/dlang/druntime/blob/95fd6e1e395e6320284a22f5d19fa41de8e1dcbb/src/rt/sections_elf_shared.d#L355) (init path) -> [`getDependencies`](https://github.com/dlang/druntime/blob/95fd6e1e395e6320284a22f5d19fa41de8e1dcbb/src/rt/sections_elf_shared.d#L650) -> [`deps.insertBack`](https://github.com/dlang/druntime/blob/95fd6e1e395e6320284a22f5d19fa41de8e1dcbb/src/rt/sections_elf_shared.d#L694), which never gets freed. This explicitly frees that memory in `_d_dso_registry`'s deinit path by resetting that array.

Side note: This together with the now merged https://github.com/dlang/druntime/pull/1857 means that valgrind no longer shows any memory lost in druntime on Linux 64bit AFAICT. There's still some reachable memory in the shared part on account of the [`_handleToDSO`](https://github.com/dlang/druntime/blob/95fd6e1e395e6320284a22f5d19fa41de8e1dcbb/src/rt/sections_elf_shared.d#L306) `HashTab` not being cleared on runtime shutdown, though.

### DIFF
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -456,6 +456,8 @@ extern(C) void _d_dso_registry(CompilerDSOData* data)
 
             unsetDSOForHandle(pdso, pdso._handle);
             pdso._handle = null;
+
+            pdso._deps.reset();
         }
         else
         {
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -455,9 +455,6 @@ extern(C) void _d_dso_registry(CompilerDSOData* data)
             }
 
             unsetDSOForHandle(pdso, pdso._handle);
-            pdso._handle = null;
-
-            pdso._deps.reset();
         }
         else
         {
@@ -611,7 +608,12 @@ version (Shared) void runFinalizers(DSO* pdso)
 void freeDSO(DSO* pdso) nothrow @nogc
 {
     pdso._gcRanges.reset();
-    version (Shared) pdso._codeSegments.reset();
+    version (Shared)
+    {
+        pdso._codeSegments.reset();
+        pdso._deps.reset();
+        pdso._handle = null;
+    }
     .free(pdso);
 }
 
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -468,7 +468,17 @@ extern(C) void _d_dso_registry(CompilerDSOData* data)
 
         freeDSO(pdso);
 
-        if (_loadedDSOs.empty) finiLocks(); // last DSO
+        if (_loadedDSOs.empty)
+        {
+            // last DSO
+            version (Shared)
+            {
+                !pthread_mutex_lock(&_handleToDSOMutex) || assert(0);
+                _handleToDSO.reset();
+                !pthread_mutex_unlock(&_handleToDSOMutex) || assert(0);
+            }
+            finiLocks();
+        }
     }
 }
 
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -465,14 +465,13 @@ extern(C) void _d_dso_registry(CompilerDSOData* data)
 
         freeDSO(pdso);
 
+        // last DSO being unloaded => shutdown registry
         if (_loadedDSOs.empty)
         {
-            // last DSO
             version (Shared)
             {
-                !pthread_mutex_lock(&_handleToDSOMutex) || assert(0);
+                assert(_handleToDSO.empty);
                 _handleToDSO.reset();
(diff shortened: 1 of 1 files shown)


Calrama:
Second commit now frees the memory held by `_handleToDSO` right before the mutex used to access `_handleToDSO` gets destroyed.

This reduces the reachable memory blocks (which strictly speaking aren't necessarily worrisome, just annoying) to only one (and that one seems to be a [valgrind bug](https://bugs.kde.org/show_bug.cgi?id=358980)):

```
==22648== Memcheck, a memory error detector
==22648== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==22648== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==22648== Command: ./test
==22648== 
==22648== 
==22648== HEAP SUMMARY:
==22648==     in use at exit: 32 bytes in 1 blocks
==22648==   total heap usage: 1,188 allocs, 1,187 frees, 1,286,283 bytes allocated
==22648== 
==22648== 32 bytes in 1 blocks are still reachable in loss record 1 of 1
==22648==    at 0x4C2CF15: calloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
==22648==    by 0x5CD2736: _dlerror_run (dlerror.c:141)
==22648==    by 0x5CD2120: dlopen@@GLIBC_2.2.5 (dlopen.c:87)
==22648==    by 0x5231221: _d_dso_registry (in /home/calrama/Projects/dlang/install/linux/lib64/libphobos2.so.0.76.0)
==22648==    by 0x51DDE1D: ??? (in /home/calrama/Projects/dlang/install/linux/lib64/libphobos2.so.0.76.0)
==22648==    by 0x400F4A9: call_init.part.0 (dl-init.c:72)
==22648==    by 0x400F5BA: call_init (dl-init.c:30)
==22648==    by 0x400F5BA: _dl_init (dl-init.c:120)
==22648==    by 0x4000BA9: ??? (in /lib64/ld-2.23.so)
==22648== 
==22648== LEAK SUMMARY:
==22648==    definitely lost: 0 bytes in 0 blocks
==22648==    indirectly lost: 0 bytes in 0 blocks
==22648==      possibly lost: 0 bytes in 0 blocks
==22648==    still reachable: 32 bytes in 1 blocks
==22648==         suppressed: 0 bytes in 0 blocks
```

Somewhat unrelated note: I'm not sure, but I think [`_loadedDSOs`](https://github.com/dlang/druntime/blob/95fd6e1e395e6320284a22f5d19fa41de8e1dcbb/src/rt/sections_elf_shared.d#L295) also needs a mutex to synchronize access to it (same as [`handleToDSO`](https://github.com/dlang/druntime/blob/95fd6e1e395e6320284a22f5d19fa41de8e1dcbb/src/rt/sections_elf_shared.d#L307))

src/rt/sections_elf_shared.d
```diff
@@ -466,7 +468,17 @@ extern(C) void _d_dso_registry(CompilerDSOData* data)
...
 
-        if (_loadedDSOs.empty) finiLocks(); // last DSO
+        if (_loadedDSOs.empty)
+        {
+            // last DSO
+            version (Shared)
+            {
+                !pthread_mutex_lock(&_handleToDSOMutex) || assert(0);
+                _handleToDSO.reset();
+                !pthread_mutex_unlock(&_handleToDSOMutex) || assert(0);
```
MartinNowak: Druntime cannot get unloaded while being used, so locking isn't necessary here.
An assertion that the hash table is empty would be in order as well.

src/rt/sections_elf_shared.d
```diff
@@ -456,6 +456,8 @@ extern(C) void _d_dso_registry(CompilerDSOData* data)
 
             unsetDSOForHandle(pdso, pdso._handle);
             pdso._handle = null;
+
+            pdso._deps.reset();
```
MartinNowak: Better move both of those to freeDSO.

MartinNowak:
> Somewhat unrelated note: I'm not sure, but I think _loadedDSOs also needs a mutex to synchronize access to it (same as handleToDSO)

No, it's a thread-local array for the API accesses, and the runtime linker serializes any init/fini calls in shared libraries (_d_dso_registry).

Calrama:
> No, it's a thread-local array for the API accesses

Right, I don't know why I remembered seeing a `__gshared` there (as it's clearly not).
Thank you for fixing up the PR.

PR #1919, opened 2017-09-18
GC micro-optimizations

rainers:
These are a few optimizations I recently noticed while staring at the precise GC. 
Benchmarks are not very stable, but show an improvement of 1 to 5 %. The graph shows the incremental performance change from bottom to top in percent, unfortunately not in the order of the commits:

grafik

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -2021,11 +2021,11 @@ struct Gcx
                 // because it's ignored for small object pools anyhow.
                 auto offsetBase = offset & notbinsize[bin];
                 biti = offsetBase >> Pool.ShiftBy.Small;
-                base = pool.baseAddr + offsetBase;
                 //debug(PRINTF) printf("\t\tbiti = x%x\n", biti);
 
                 if (!pool.mark.set(biti) && !pool.noscan.test(biti))
                 {
+                    base = pool.baseAddr + offsetBase;
                     top = base + binsize[bin];
                     goto LaddRange;
                 }
@@ -2055,7 +2055,6 @@ struct Gcx
             else if (bin == B_PAGEPLUS)
             {
                 pn -= pool.bPageOffsets[pn];
-                base = pool.baseAddr + (pn * PAGESIZE);
                 biti = pn * (PAGESIZE >> Pool.ShiftBy.Large);
 
                 pcache = cast(size_t)p & ~cast(size_t)(PAGESIZE-1);
@@ -2064,6 +2063,7 @@ struct Gcx
 
                 if (!pool.mark.set(biti) && !pool.noscan.test(biti))
                 {
+                    base = pool.baseAddr + (pn * PAGESIZE);
                     top = base + pool.bPageOffsets[pn] * PAGESIZE;
                     goto LaddRange;
                 }
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -1935,6 +1935,8 @@ struct Gcx
     private:
         void grow()
         {
+            pragma(inline, false);
+
             enum initSize = 64 * 1024; // Windows VirtualAlloc granularity
             immutable ncap = _cap ? 2 * _cap : initSize / ScanRange.sizeof;
             auto p = cast(ScanRange*)os_mem_map(ncap * ScanRange.sizeof);
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -2113,7 +2113,7 @@ struct Gcx
         p1 = cast(void**)next.pbot;
         p2 = cast(void**)next.ptop;
         // printf("  pop [%p..%p] (%#zx)\n", p1, p2, cast(size_t)p2 - cast(size_t)p1);
-        goto Lagain;
+        goto LnextBody;
     }
 
     // collection step 1: prepare freebits and mark bits
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -1980,6 +1980,7 @@ struct Gcx
         //printf("marking range: [%p..%p] (%#zx)\n", p1, p2, cast(size_t)p2 - cast(size_t)p1);
     Lnext: for (; p1 < p2; p1++)
         {
+        LnextBody:
             auto p = *p1;
 
             //if (log) debug(PRINTF) printf("\tmark %p\n", p);
@@ -2009,6 +2010,7 @@ struct Gcx
                 size_t pn = offset / PAGESIZE;
                 Bins   bin = cast(Bins)pool.pagetable[pn];
                 void* base = void;
+                void* top = void;
 
                 //debug(PRINTF) printf("\t\tfound pool %p, base=%p, pn = %zd, bin = %d, biti = x%x\n", pool, pool.baseAddr, pn, bin, biti);
 
@@ -2022,10 +2024,10 @@ struct Gcx
                     base = pool.baseAddr + offsetBase;
                     //debug(PRINTF) printf("\t\tbiti = x%x\n", biti);
 
-                    if (!pool.mark.set(biti) && !pool.noscan.test(biti)) {
-                        stack[stackPos++] = ScanRange(base, base + binsize[bin]);
-                        if (stackPos == stack.length)
-                            break;
+                    if (!pool.mark.set(biti) && !pool.noscan.test(biti))
+                    {
+                        top = base + binsize[bin];
+                        goto LaddRange;
                     }
                 }
                 else if (bin == B_PAGE)
@@ -2044,10 +2046,10 @@ struct Gcx
                     if(!pointsToBase && pool.nointerior.nbits && pool.nointerior.test(biti))
                         continue;
 
-                    if (!pool.mark.set(biti) && !pool.noscan.test(biti)) {
-                        stack[stackPos++] = ScanRange(base, base + pool.bPageOffsets[pn] * PAGESIZE);
-                        if (stackPos == stack.length)
-                            break;
+                    if (!pool.mark.set(biti) && !pool.noscan.test(biti))
+                    {
+                        top = base + pool.bPageOffsets[pn] * PAGESIZE;
+                        goto LaddRange;
                     }
                 }
                 else if (bin == B_PAGEPLUS)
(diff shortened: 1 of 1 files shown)


ibuclaw:
Do we have a set of benchmark tests for the GC? How feasible would it be to walk through all history of the GC implementation, running these tests after every change and pushing that out to a statistics / analysis tool.

I'm not convinced that we can really accurately gauge performance of the GC in the current way these sort of PRs are done - usually by comparing the difference between A and B.

rainers:
> Do we have a set of benchmark tests for the GC? 

They are in druntime/benchmark/gcbench

> How feasible would it be to walk through all history of the GC implementation, running these tests after every change and pushing that out to a statistics / analysis tool.

That's what https://blog.thecybershadow.net/2015/05/05/is-d-slim-yet/ did, though not for the GC benchmarks. Unfortunately it is currently broken because the data set has grown too large.

MartinNowak:
> I'm not convinced that we can really accurately gauge performance of the GC in the current way these sort of PRs are done - usually by comparing the difference between A and B.

? We do run those comparisons for every GC change, so we should keep getting better, at least regarding the GC implementation.

src/gc/impl/conservative/gc.d
```diff
@@ -2022,10 +2024,10 @@ struct Gcx
...
                     //debug(PRINTF) printf("\t\tbiti = x%x\n", biti);
 
-                    if (!pool.mark.set(biti) && !pool.noscan.test(biti)) {
-                        stack[stackPos++] = ScanRange(base, base + binsize[bin]);
-                        if (stackPos == stack.length)
-                            break;
+                    if (!pool.mark.set(biti) && !pool.noscan.test(biti))
+                    {
+                        top = base + binsize[bin];
+                        goto LaddRange;
```
MartinNowak: There you go :).

src/gc/impl/conservative/gc.d
```diff
@@ -2060,35 +2062,40 @@ struct Gcx
...
+                if (p1 + 1 < p2) // *p1 already scanned
+                    toscan.push(ScanRange(p1 + 1, p2));
+                // reverse order for depth-first-order traversal
+                foreach_reverse (ref rng; stack)
+                    toscan.push(rng);
+                stackPos = 0;
+                // continue with last stack entry
+                p1 = cast(void**)base;
+                p2 = cast(void**)top;
+                goto LnextBody; // skip increment and check
```
MartinNowak: That's a second loop around/inside the loop, can we keep reusing the Lagain part for that?
Maybe by hoisting the `next` variable and setting it here `next = ScanRange(base, top)`, then jumping to the `p1 = cast(void**)next.pbot;` part.
Seems a bit cleaner to me and the case of the local stack overflowing is rare enough to not worry about 2 cycles or so.

src/gc/impl/conservative/gc.d
```diff
@@ -2113,7 +2113,7 @@ struct Gcx
         p1 = cast(void**)next.pbot;
         p2 = cast(void**)next.ptop;
         // printf("  pop [%p..%p] (%#zx)\n", p1, p2, cast(size_t)p2 - cast(size_t)p1);
-        goto Lagain;
+        goto LnextBody;
```
MartinNowak: remove Lagain label

src/gc/impl/conservative/gc.d
```diff
@@ -2629,6 +2629,8 @@ struct Pool
 
     bool isLargeObject;
 
+    enum shiftBySmall = 4;
+    enum shiftByLarge = 12;
     uint shiftBy;    // shift count for the divisor used for determining bit indices.
```
MartinNowak: Nice, how about using an enum.
```d
enum ShiftBy : ubyte
{
    small = 4,
    large = 12,
}
ShiftBy shiftBy;
```

src/gc/impl/conservative/gc.d
```diff
@@ -1984,114 +1984,114 @@ struct Gcx
...
+
+                if (low > high)
+                    continue Lnext;
+            }
+            size_t offset = cast(size_t)(p - pool.baseAddr);
+            size_t biti = void;
+            size_t pn = offset / PAGESIZE;
+            Bins   bin = cast(Bins)pool.pagetable[pn];
+            void* base = void;
+            void* top = void;
```
MartinNowak: Nesting loops is really just a bad habit ;).

src/gc/impl/conservative/gc.d
```diff
@@ -1984,114 +1984,114 @@ struct Gcx
...
-                // Adjust bit to be at start of allocated memory block
-                if (bin < B_PAGE)
+            // Adjust bit to be at start of allocated memory block
+            if (bin < B_PAGE)
+            {
+                // We don't care abou setting pointsToBase correctly
+                // because it's ignored for small object pools anyhow.
+                auto offsetBase = offset & notbinsize[bin];
+                biti = offsetBase >> pool.shiftBySmall;
+                base = pool.baseAddr + offsetBase;
```
MartinNowak: Common code, less icache/uop-cache pressure :).

MartinNowak:
Just reran the tests, haven't seen that OOM error before.
https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdruntime/detail/PR-1919/1/pipeline/175#step-488-log-45

rainers:
> Just reran the tests, haven't seen that OOM error before.

The autotester also hangs rather consistently for Linux_32. I wasn't able to reproduce both failures locally, though. No idea what could be causing this.

src/gc/impl/conservative/gc.d
```diff
@@ -2080,7 +2080,8 @@ struct Gcx
         {
             // local stack is full, push it to the global stack
             assert(stackPos == stack.length);
-            toscan.push(ScanRange(p1, p2));
+            if (p1 + 1 < p2)
```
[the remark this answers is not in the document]
rainers: > That would be better added in the above loops where we break without incrementing the pointer.

Incrementing could be done unconditionally, but I think having a simple common subexpression here might be better than writing back the value to p1.

> In fact you can just factor this out as common code between all branches.

That way you have to calculate `top` unconditionally. I guess it could be slightly better to also move the calculation of `base` into the conditional.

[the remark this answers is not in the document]
rainers: > Could help a bit to better fit the normal loop body in the µop-cache.

`toscan.push()` increases the loop size quite a bit because `ScanRange.grow()` is inlined. Disabling that with `pragma(inline, false)` didn't have any impact on performance, though.

rainers:
> Off-topic: How does GC performance compare between DMD, GDC and LDC?

According to my recent tests LDC seems about 40% faster, both in overall time and GC time. I don't have a reasonable GDC version on Windows.

rainers:
Addressed all comments but the loop changes. I'll try another simplification.

rainers:
Finally figured the cause of the failures, though I don't get why cleraing `pcache` is necessary. @MartinNowak, do you remember why it needed to be included in the Lagain loop?

The mark loop is now simplified to only use forward jumps and standard backward continuation. I also noticed a few more redundant operations.

ibuclaw:
> Finally figured the cause of the failures, though I don't get why cleraing pcache is necessary.

Just guess from me after looking at the loop. Could it happen that you have two ranges next to each other? - ie: pcache is both the end of the previous and start of the current range.

MartinNowak:
> Finally figured the cause of the failures, though I don't get why clearing pcache is necessary. 

Hihi, changing sth. w/o understanding why doesn't count as "figuring out" ;).

> @MartinNowak, do you remember why it needed to be included in the Lagain loop?

No I merely preserved the existing semantics.
https://github.com/MartinNowak/druntime/commit/85f392dba41d48368b7b00e7ac7a9a4e6beeffab#diff-6f1ab0423fff9dcd084ecf9a677dc426R2391

Actually it looks confusing that we don't make a small 16-byte bin
https://github.com/dlang/druntime/blob/4ee45696e13786d4e282876e01d5114ec83d0db1/src/gc/impl/conservative/gc.d#L2010
just because it's on the same page (4KB) as the previous element.
https://github.com/dlang/druntime/blob/4ee45696e13786d4e282876e01d5114ec83d0db1/src/gc/impl/conservative/gc.d#L1982-L1983

Am I misreading this?

MartinNowak:
Ah, a bit easier to see that `pcache` isn't set for `< B_PAGE` pages when this code wasn't so regressed as currently.
www.dsource.org/projects/tango/browser/trunk/tango/core/rt/gc/basic/gcx.d#L2243

src/gc/impl/conservative/gc.d
```diff
@@ -2012,94 +2022,106 @@ struct Gcx
...
+                    biti = offset >> Pool.ShiftBy.Large;
                     //debug(PRINTF) printf("\t\tbiti = x%x\n", biti);
 
                     pcache = cast(size_t)p & ~cast(size_t)(PAGESIZE-1);
+                    base = cast(void*)pcache;
 
                     // For the NO_INTERIOR attribute.  This tracks whether
                     // the pointer is an interior pointer or points to the
                     // base address of a block.
-                    bool pointsToBase = (base == sentinel_sub(p));
```
MartinNowak: Was it necessary to remove that name?
Even dmd should be able to optimize this and deciphering `base != sentinel_sub(p)` isn't trivial.

MartinNowak:
I don't see why `pcache` needs to be reset. What's ugly is that it's set before marking, but shouldn't be a problem.
Honestly we should stop throwing time at this crappy code. The algorithmic deficiencies are a much bigger problem and the codebase is beyond repair/maintainability.

src/gc/impl/conservative/gc.d
rainers: > Even dmd should be able to optimize

Unfortunately, it it doesn't just use the comparison in the following `if`, but stores the result to a byte with `sete` and uses that, adding a number of gratuitious instructions.

rainers:
Thanks @ibuclaw and @MartinNowak for looking into this.

> Just guess from me after looking at the loop. Could it happen that you have two ranges next to each other? - ie: pcache is both the end of the previous and start of the current range.

I don't see how `pcache` can point to two ranges at the same time.

For my own sanity: `pcache` points to the last GC managed 4k page that is found to be referenced and is inside a large pool, i.e. there is no other object on that page. As the object containing the page has been marked and its range has been added to the stack for scanning there is no need to look at pointers referencing the same page.
AFAICT that does not invalidate if the mark loop switches to another range. The test failures suggest it does, though.

rainers:
> What's ugly is that it's set before marking, but shouldn't be a problem.

I think it's not so bad, as this shortcuts tests that point to the same page even if it is already marked. Otherwise it would only work for subsequent references of the first hit.

> Honestly we should stop throwing time at this crappy code. The algorithmic deficiencies are a much bigger problem and the codebase is beyond repair/maintainability.

I'm not sure it is so bad. Last time I checked the GC was faster for manually managed memory than C's malloc/free on Windows (not considering the additional benefit of not having to call addRange/removeRange).

My actual motivation was improving the precise GC, though, so that it would pass your review ;-) Adding these seemingly simple optimizations there only would bias the benchmark results if not applied to the standard GC, too.

rainers:
Here's a graph of new benchmarks of the time spent in the GC for master, the version of the initial PR (intermediate) and the current PR (microopt):

grafik

MartinNowak:
> I'm not sure it is so bad. Last time I checked the GC was faster for manually managed memory than C's malloc/free on Windows (not considering the additional benefit of not having to call addRange/removeRange).

Are you comparing that against dmc's libc malloc/free?

MartinNowak:
BTW, I've recently started to use plots with error intervals for benchmarks, should I write a short R plot script for the runbench tool as well?

MartinNowak:
> My actual motivation was improving the precise GC, though, so that it would pass your review ;-) Adding these seemingly simple optimizations there only would bias the benchmark results if not applied to the standard GC, too.

The requirements for that remain straightforward, only very little performance decrease for existing programs, and only small API/maintainability commitments.

rainers:
> The requirements for that remain straightforward, only very little performance decrease for existing programs, and only small API/maintainability commitments.

That's why I meant I should have left these micro-optimzations for the precise GC to compensate for some performance losses ;-)

rainers:
> Are you comparing that against dmc's libc malloc/free?

That comparison is some time ago, this is the respective comment: https://github.com/dlang/druntime/pull/739#issuecomment-37757466. I'm not sure whether the dmc lib was already just a wrapper for the Windows HeapAllocate functions as with the MS libraries. The Windows functions might have improved in the meantime, too.

PR #1920, opened 2017-09-24
do one less divide in unsignedToTempString()

WalterBright:


schveiguy:
LGTM, but I see more opportunities. For instance, the while loop can just be `while(true)`. Any interest in adding these, or should I just merge as-is?

WalterBright:
Just merge, then feel free to improve it!

PR #1921, opened 2017-09-25
fix 17375 - colliding modules detected with binutils 2.28 linker ...

MartinNowak:
...and shared libraries

- fixed by removing the module collision check

  This was a too complex (and hardly maintainable) mechanism trying to catch
  rare issues in build systems. As with static libraries leave it to developers
  to use/deal with symbol interposition.

  With D's qualified mangling accidental symbol collisions between the
  executable and a shared library should only ever happen when using the same
  module/package names.

  So far haven't encountered any valid error of this collision check, but we've
  collectively spent several days to fix issues with it's fragile implementation.


wilzbach:
(Resolved the merge conflicts - the conflict arouse due to the `body` -> `do`  transition)

wilzbach (approved):

PetarKirov (approved):

PR #1922, opened 2017-09-26
Solve Issue #16641 - Infinite loop in throw

etcimon:
This solves an issue as described in https://issues.dlang.org/show_bug.cgi?id=16641

The problem happens if an invalid memory operation is followed by an assertion failure in a scope(exit). This allows the chaining to end and the application to proceed.

### DIFF
--- a/src/rt/dwarfeh.d
+++ b/src/rt/dwarfeh.d
@@ -408,7 +408,8 @@ extern (C) _Unwind_Reason_Code __dmd_personality_v0(int ver, _Unwind_Action acti
                 //printf("chain\n");
                 // Append eh's object to ehn's object chain
                 Throwable n = ehn.object;
-                while (n.next)
+                int infini_guard;
+                while (n.next && ++infini_guard < 10_000)
                     n = n.next;
                 n.next = eh.object;
 


src/rt/dwarfeh.d
```diff
@@ -408,7 +408,8 @@ extern (C) _Unwind_Reason_Code __dmd_personality_v0(int ver, _Unwind_Action acti
                 //printf("chain\n");
                 // Append eh's object to ehn's object chain
                 Throwable n = ehn.object;
-                while (n.next)
+                int infini_guard;
+                while (n.next && ++infini_guard < 10_000)
```
rainers: This looks rather hackish. It defers problems to later up the callstack, e.g. when printing the callstack indefinitely.

Maybe you could check whether the new Throwable is already in the exception chain instead. That should avoid any circles to begin with. Not sure what to actually do then, maybe issue another CyclicExceptionError.

Having a similar check for other platforms would be nice, too, but should not block merging it for DWARF.

n8sh: An alternative check could look like this:
```d
if (n.next)
{
    // Use circular list detection from list-length implementation
    // in Guy L. Steele's "Common Lisp: The Language", 2nd ed.
    Throwable slow = n;
    Throwable fast = n;
    do
    {
        fast = fast.next;
        if (!fast.next)
            break;
        if (fast is slow)
            //We've detected a cycle. If we want to do something
            //special we can do it here.
            break;
        fast = fast.next;
        slow = slow.next;
    } while (fast.next);
    n = fast;
}
```
but that may be overkill.

PR #1923, opened 2017-09-29
add plot script for runbench

MartinNowak:
- parses runbench output
- plots times for all tests as boxplots (log1p y-axis)
- also plots any GC stats if available

Also usable to compare the other benchmarks, e.g. aabench.

### DIFF
--- a/benchmark/runbench.d
+++ b/benchmark/runbench.d
@@ -87,43 +87,54 @@ void runTests(Config cfg)
         import std.datetime, std.algorithm : min;
         auto sw = StopWatch(AutoStart.yes);
         auto minDur = Duration.max;
+        string minGCProf;
 
-        stdout.writef("R %-16s", bin.baseName.stripExtension);
-        if (cfg.verbose) stdout.writeln();
-        stdout.flush();
+        immutable benchName = bin.baseName.stripExtension;
+
+        void report(string pfx, Duration dur, string gcProf)
+        {
+            auto parts = dur.split!("seconds", "msecs");
+            if (gcProf.length)
+                writefln("%s %-16s %s.%03s s, %s", pfx, benchName, parts.seconds, parts.msecs, gcProf);
+            else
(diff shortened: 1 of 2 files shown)


MartinNowak:
Examples
## times only
times
## times and GC summary
times_and_gc

rainers:
Interesting. I recently added average and max values to the output of runbench, too. Some tests have a wide range of results on my Windows systems, e.g. rand_large. I suspect this is caused by ASLR.

The graphs look a little tiny, though. Maybe it would be slightly better if the frames of the master/feature boxes are not both painted black.

benchmark/runbench.d
```diff
@@ -87,43 +87,54 @@ void runTests(Config cfg)
...
+                if (!lines.empty) gcProf = lines.front.find("GC summary:")[11..$].idup;
             }
             else
             {
                 auto lines = output.splitter(ctRegex!`\r\n|\r|\n`)
                     .find!(ln => ln.startsWith("GC summary:"));
-                if (!lines.empty) gcprof = lines.front[11..$];
+                if (!lines.empty) gcProf = lines.front[11..$];
+            }
+            report("RUN", dur, gcProf);
```
rainers: I think this should not be displayed in a standard run if you just look at the text output. Maybe print in verbose mode only?

MartinNowak:
Done @rainers
times
I'm getting very little variation usually, but I'm also on Linux (and occasionally disable ASLR for benchmarking/debugging).

rainers:
Thanks, looks better.

> I'm getting very little variation usually, but I'm also on Linux (and occasionally disable ASLR for benchmarking/debugging).

On Windows, ASLR can be disabled for DLL loading, but I have not found a way to avoid it for VirtualAlloc or stack addresses.

BTW: I have no idea why the words.d benchmark suffers a bit by the optimization. I see that on my system, too. It doesn't spend much time in the GC, though, anyway.

PR #1924, opened 2017-09-29
implement open GC registry

MartinNowak:
- allows to add new GCs at link time


src/gc/config.d
```diff
@@ -40,6 +42,8 @@ struct Config
     {
         import core.internal.traits : externDFunc;
 
+        __gshared linkDummy = &_dummy_ref_to_link_gc_register_constructor;
```
rainers: Wouldn't it be simpler to just call `_d_register_conservative_gc()` and `_d_register_manual_gc()` explicitely. You can still  override these symbols in the executable if you don't want to register them.

src/gc/impl/conservative/gc.d
```diff
@@ -332,6 +323,7 @@ class ConservativeGC : GC
             cstdlib.free(gcx);
             gcx = null;
         }
+        cstdlib.free(cast(void*) this);
```
rainers: Freeing `this` in a public method is a bit dangerous because an invariant might crash. One possible solution might be to pass the finalize function to the registry, too.

src/gc/impl/manual/gc.d
```diff
@@ -35,46 +35,41 @@ static import core.memory;
...
+}
 
-        auto instance = cast(ManualGC) gc;
-        instance.Dtor();
-        cstdlib.free(cast(void*) instance);
-    }
+class ManualGC : GC
+{
+    __gshared Array!Root roots;
+    __gshared Array!Range ranges;
```
rainers: I know it's been this way before, but it doesn't really make sense to make these members `__gshared`.

test/init_fini/src/register_custom_gc.c
```diff
@@ -0,0 +1,6 @@
+extern void register_mygc();
+
+__attribute__((constructor)) static void xxx_ctor()
+{
+    register_mygc();
+}
```
rainers: Registration still needs a solution on Windows (though something similar is possible). I'd prefer if we would not have to rely on C, though.

Some time ago, I tried implementing `pragma(dataSeg, "segment-name")`, would be pretty helpful for this, too. I think it's badly needed in a system programming language.

rainers:
The nits aside, this approach is much better for registering and selecting the GC than the current one.

src/gc/impl/conservative/gc.d
MartinNowak: Yes, not so nice.

In fact just using a static buffer instead of the C heap might work as well.

Think the best alternative is to convert Dtor to a real destructor which doesn't have the invariant problem.

test/init_fini/src/register_custom_gc.c
MartinNowak: I was hoping for the _STI/_STD mechanism to work (to be recognized by the linker), but must have misremembered that it works.
There is https://issues.dlang.org/show_bug.cgi?id=16300 and we've recently worked on constructor/destructor functions https://github.com/dlang/dmd/pull/6956 that could be added as pragma(constructor) and pragma(destructor). Not sure what gdc or ldc are doing with regards to that.

MartinNowak:
Think it would make sense to work on `pragma(crt_constructor[, priority])` first, rather than spending time on a Windows solution here.

src/gc/config.d
MartinNowak: The C object file wouldn't be referenced and doesn't get linked into the application.

MartinNowak:
The registered factory should also be `nothrow @nogc` but that's fairly invasive up to `core.thread`, should be done first anyhow.

src/gc/config.d
rainers: What C object file? Just call _d_register_conservative_gc() and _d_register_manual_gc() right here.

rainers: Just noticed yesterday that you can also reimplement gc_init in the executable registering the desired or new GCs.

MartinNowak: That wouldn't allow configuring by application users. Think of how Java GC's are often selected/optimized according to specific loads.
Wouldn't work with druntime as DLL or with hidden symbols in shared libraries.
Also many programmers hardly have an idea how linkers work.

rainers: Agreed, replacing gc_init is a bit blunt and doesn't allow adding to existing GCs. It's currently the only way to remove a GC from the binary.

The current of `_dummy_ref_to_link_gc_register_constructor` seems gratuitious, though. Just registering the GCs here should be fine.

src/gc/registry.d
```diff
@@ -0,0 +1,87 @@
+/**
+ * Contains a registry for GC factories.
```
rainers: I'm slightly confused by the terminology here: I would have expected the "registry" to be the "factory" and a GC is registered by adding a "factory method" to the factory.

MartinNowak: Well, yes it's a registry for factory methods.

andralex:
@MartinNowak I wonder if we should explore dropping the notion of GC as a class, which allows replacing it during runtime (not really used or useful). This will cause some breakage but it would be a design improvement. What do you think?

MartinNowak:
> @MartinNowak I wonder if we should explore dropping the notion of GC as a class, which allows replacing it during runtime (not really used or useful). This will cause some breakage but it would be a design improvement. What do you think?

The opposite, let's allow to implement GCs as dub packages (https://github.com/dlang/druntime/pull/1924). There is still are still a lot of improvement potentials in the GC domain, but we (as core team) will hardly invest in this area in the next 1 or 2 years.
You might be interested in https://forum.dlang.org/post/p6cq7f$2cg$1@digitalmars.com.
Opening this area for innovation outside of our busy core team could realistically get us a low-latency GC in the meantime.

MartinNowak:
> @MartinNowak I wonder if we should explore dropping the notion of GC as a class, which allows replacing it during runtime (not really used or useful).

Also I'd like this to be the last work on the GC for a while, people still get confused that our community keeps talking about GC improvements.

test/init_fini/src/register_custom_gc.c
thewilsonator: We have https://github.com/ldc-developers/druntime/blob/ldc/src/ldc/attributes.d#L229 in LDC, with the addition of `pragma(crt_constructor)` and friends, perhaps we should add that?

rainers:
Rebased and replaced the C registration with pragma(crt_constructor). Might need https://github.com/dlang/dmd/pull/9132

thewilsonator:
```
src/custom_gc.d(22): Error: class `custom_gc.MallocGC` interface function `ProfileStats profileStats() nothrow` is not implemented
```

thewilsonator:
```
/usr/bin/ld: error: generated/linux/release/64/libphobos2.a(gc.o): size of section .ctors is not multiple of address size
```

rainers:
> /usr/bin/ld: error: generated/linux/release/64/libphobos2.a(gc.o): size of section .ctors is not multiple of address size

It wasn't too funny tracking down this one: https://github.com/dlang/dmd/pull/9204

rainers:
Finally figured the spurious failures here: do not free your manually memory managed classes in the dtor when using destroy.

thewilsonator:
I'm not so sure about that last one... we have `destroy!false` for not overwriting the memory. (might need a rebase to use it)

rainers:
> I'm not so sure about that last one... we have `destroy!false` for not overwriting the memory.

Ok, let's try that.

thewilsonator:
Hmm, seems to fail on windows.

rainers:
I opted for having a tiny memory leak for now. (ProtoGC also has a small leak).

posix.mak
```diff
@@ -229,6 +229,10 @@ $(ROOT)/threadasm.o : src/core/threadasm.S
 	@mkdir -p $(dir $@)
 	$(CC) -c $(CFLAGS) $< -o$@
 
+$(ROOT)/gc/%.o : src/gc/%.c
+	@mkdir -p $(dir $@)
+	$(CC) -c $(CFLAGS) $< -o$@
```
wilzbach: Why do we need this? (I don't see any C files that have been added)

wilzbach:
FYI the master -> stable branch-off will happen in ~ 3 days. So it would be nice to have this as part of 2.085. Anything blocking this?

posix.mak
rainers: This has remained from the initial version that used C files instead of pragma(crt_constructor). I'll remove it.

rainers:
> Anything blocking this?

Good to go AFAICT. I guess a changelog entry and an update to the documentation are recommended.

thewilsonator (approved):

changelog/gc_registry.dd
```diff
@@ -0,0 +1,33 @@
...
+
+---
+extern (C) pragma(crt_constructor) void registerMyGC()
+{
+    registerGCFactory("mygc", &createMyGC);
+}
+
+GC createMyGC()
+{
+	__gshared ubyte[__traits(classInstanceSize, MyGC)] buf;
```
thewilsonator: Probably good to add an alignment here see e.g. https://github.com/dlang/druntime/pull/2411

rainers:
I've added a changelog entry, but noticed one gotcha: when writing a new GC you will have to import modules from the gc package, but that isn't available in the standard import folders.

You always needed to add the src folder to the import paths somehow, not sure we should bother now.

changelog/gc_registry.dd
rainers: Unfortunately, classInstanceAlignment doesn't seem good enough (and isn't yet available), so I opted for aligning to pointer size.
I also used core.lifetime.emplace now.

changelog/gc_registry.dd
```diff
@@ -0,0 +1,32 @@
...
+
+A GC registry has been implemented that allows to add
+garbage collector implementations by just linking them into
+the binary. To do so add a function that is executed before the
+D runtime initialization using `pragma(crt_constructor)`:
+
+---
+extern (C) pragma(crt_constructor) void registerMyGC()
+{
+    registerGCFactory("mygc", &createMyGC);
```
wilzbach: Doesn't this need an import?

wilzbach (commented):
Uhm, looks like we need to move the user-exposed bits to core.gc?

rainers:
> Uhm, looks like we need to move the user-exposed bits to core.gc?

I'd rather add the necessary modules to the import folder as is.

changelog/gc_registry.dd
rainers: It does (`import gc.registry, gc.gcinterface;`), but they are not easily accessible right now

wilzbach:
> > Uhm, looks like we need to move the user-exposed bits to core.gc?
> 
> I'd rather add the necessary modules to the import folder as is.

Wouldn't that be a potentially breaking change as we add a new top-level module?
Also, IIRC a few tools (rdmd, `dmd -i`, etc.) have hard-coded the list of top-level modules from the standard library, so it's probably not a good idea to introduce one lightly.

rainers:
> Wouldn't that be a potentially breaking change as we add a new top-level module?

It is already there, just hidden. If you create a module with the same module name, you will get symbol conflicts at link time.

> Also, IIRC a few tools (rdmd, `dmd -i`, etc.) have hard-coded the list of top-level modules from the standard library, so it's probably not a good idea to introduce one lightly.

Might be an issue, but most of the time it should be ok to just recompile the module again, the linker just picks one.

I'm also ok with not making them public. If you work on replacing the GC, adding an import path should not scare you. Maybe the changelog text is already too detailed, most of it might be better in the documentation.

rainers:
I moved most of the changelog text to the documentation https://github.com/dlang/dlang.org/pull/2577

wilzbach (approved):
We should still expose the methods to the user, s.t. dub packages have an easy way to provide a nice out-of-the-box experience for their custom GCs, but this can happen in another PR.

wilzbach:
> I'd rather add the necessary modules to the import folder as is.

Hmm, currently all the user-exposed GC-related methods are in `core.memory`:

https://dlang.org/phobos/core_memory.html#.GC

How about moving the public interfaces to this module or as a `core.memory.gc` submodule?

PR #1925, opened 2017-09-30
rt.config: Make custom mangling more robust

kinke:
`pragma(mangle, ...)` allows specifying a 'base' mangle for the symbol's linkage. So by using `extern(C) pragma(mangle, ...)`, a base C mangle can be specified, without having to take care of a platform-specific prefix; that is handled by the compiler.

The previous code relied on D symbols not getting prefixed, which isn't the case for LDC on OSX [and Win32, but that's subject to change].

### DIFF
--- a/src/rt/config.d
+++ b/src/rt/config.d
@@ -36,20 +36,18 @@ module rt.config;
 // line arguments, i.e. if command line arguments are not disabled, they can override
 // options specified through the environment or embedded in the executable.
 
-import core.demangle : cPrefix;
-
 // put each variable in its own COMDAT by making them template instances
 template rt_envvars_enabled()
 {
-    pragma(mangle, cPrefix ~ "rt_envvars_enabled") __gshared bool rt_envvars_enabled = false;
+    extern(C) pragma(mangle, "rt_envvars_enabled") __gshared bool rt_envvars_enabled = false;
 }
 template rt_cmdline_enabled()
 {
-    pragma(mangle, cPrefix ~ "rt_cmdline_enabled") __gshared bool rt_cmdline_enabled = true;
+    extern(C) pragma(mangle, "rt_cmdline_enabled") __gshared bool rt_cmdline_enabled = true;
 }
 template rt_options()
 {
-    pragma(mangle, cPrefix ~ "rt_options") __gshared string[] rt_options = [];
+    extern(C) pragma(mangle, "rt_options") __gshared string[] rt_options = [];
 }
 
 import core.stdc.ctype : toupper;


ibuclaw:
Sure,  but is that a correct interpretation of the spec? I thought that `pragma(mangle)` should be taken as is without alteration. Does marking any symbol as `extern(X)` make any sense?

kinke:
> Sure, but is that a correct interpretation of the spec?

Haven't checked; it matches `T.mangleof` though, at least for C and D symbols. DMD uses two underscores for `T.mangleof` of C++ symbols on OSX (i.e., matching the name in the object file), but no underscore for C symbols on Win32 [and OSX I guess] which I consider to be a bug.

> Does marking any symbol as extern(X) make any sense (in junction with pragma mangle)?

A curiosity I came across during testing, with DMD 2.076 on Win32:
```
extern(C++, ns1) pragma(mangle, "foo1") __gshared int global1;
pragma(mangle, "foo2") extern(C++, ns2) __gshared int global2;
pragma(msg, global1.mangleof); // foo1
pragma(msg, global2.mangleof); // ?global2@ns2@@3HA
```

:D

kinke:
> Does marking any symbol as extern(X) make any sense (in junction with pragma mangle)?

For functions definitely, in order to tweak calling convention and mangle at the same time.

ibuclaw:
> For functions definitely, in order to tweak calling convention and mangle at the same time if need be.

For changing the calling convention, yes. But the mangle remains unaltered from what the user specified.

kinke:
> But the mangle remains unaltered from what the user specified.

Clearly not, otherwise this PR wouldn't work.

ibuclaw:
And you don't find it odd that the symbol name you set is not the same symbol that ended up in the object file?

kinke:
> And you don't find it odd that the symbol name you set is not the same symbol that ended up in the object file?

At first yes, but after giving it some thought (considering `T.mangleof` behavior and portability concerns) I changed my opinion.

ibuclaw:
> To use a concrete example as a basis for discussion, how does GDC treat pragma(mangle, …) on OS X?

Now that I think about it. GDC always applies target specific mangling, regardless of what `extern` is used.  There is a trick to override target mangling by prefixing the string with a star.

So making a slight modification to the second example above:
```
extern(C) pragma(mangle, "*csym") void bar() {}
```
There was definitely a very good reason why I did it this way, but that currently escapes me for the time being.

This PR is fine, but I'd still like to question whether our use of pragma(mangle) is correct.

kinke:
> GDC always applies target specific mangling

Just to avoid confusion - on OSX, the extra underscore appears to be a platform-wide convention, and David mentioned that while the C++ docs only talk about a single leading underscore, 2 actually end up in the object file.
On Win32 on the other hand, MS uses an extra underscore for C, but not for C++ (MSVC mangling). So 'target-specific mangling' for Win32-MSVC doesn't solely depend on the platform, but also on the linkage.
So if support for 'target-specific mangling' is to be kept, we need to take linkage into account.

ibuclaw:
> There was definitely a very good reason why I did it this way, but that currently escapes me for the time being.

Having slept on it. GDC does the same thing as documented in GCC.

PR #1927, opened 2017-10-02
Add toTimeVal and toTimeSpec for Duration struct

FreeSlave:
SysTime has toTimeVal and toTimeSpec, while Duration does not have those. But timeval and timespec are used as durations too, not only as timestamps. See e.g. select and pselect.

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -1669,7 +1669,7 @@ public:
 
     /++
         Converts `Duration` to a `timeval` structure.
-    +/
+      +/
     timeval toTimeVal() const nothrow @nogc
     {
         timeval value;
@@ -1677,6 +1677,7 @@ public:
         return value;
     }
 
+    ///
     unittest
     {
         timeval value;
@@ -1685,7 +1686,7 @@ public:
         assert(dur!"msecs"(1005).toTimeVal() == value);
     }
 
-    version(StdDdoc)
+    version(CoreDdoc)
     {
         private struct timespec {}
         /++
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -1667,6 +1667,52 @@ public:
         }
     }
 
+    /++
+        Converts `Duration` to a `timeval` structure.
+    +/
+    timeval toTimeVal() const nothrow @nogc
+    {
+        timeval value;
+        split!("seconds", "usecs")(value.tv_sec, value.tv_usec);
+        return value;
+    }
+
+    unittest
+    {
+        timeval value;
+        value.tv_sec = 1;
+        value.tv_usec = 5000;
+        assert(dur!"msecs"(1005).toTimeVal() == value);
+    }
+
+    version(StdDdoc)
+    {
+        private struct timespec {}
+        /++
+            Converts `Duration` to a `timespec` structure.
+
+            $(BLUE This function is Posix-Only.)
+          +/
+        timespec toTimeSpec() const nothrow @nogc;
+    }
+    else version(Posix)
+    {
+        timespec toTimeSpec() const nothrow @nogc
+        {
+            timespec spec;
+            split!("seconds", "nsecs")(spec.tv_sec, spec.tv_nsec);
+            return spec;
+        }
+
+        ///
+        unittest
+        {
+            timespec value;
+            value.tv_sec = 1;
+            value.tv_nsec = 5005000;
+            assert(dur!"usecs"(1005005).toTimeSpec() == value);
+        }
+    }
 
 private:
 


jmdavis:
I don't know. The code looks fine, but it's so trivial to use `split` to do this that these functions really don't add much value. And in most cases, you'd be using a `timeval` or `timespec`, because you're passing them to a C function which takes them by pointer, in which case, you couldn't just call `toTimeVal` and pass the result to a function which seems like it would be the main benefit of `toTimeVal` over simply calling `split`. You'd normally be initializing a local variable to then take its address, and in that case, it would just be more efficient to call `split` directly. It would be a move rather than a copy, so the cost would be negligible, but you're still essentially doing

    timeval tv1;
    d.split!("seconds", "usecs")(tv1.tv_sec, tv1.tv_usec);
    timeval tv2 = tv1;
    cFunc(&tv2);

even if it looks like

    auto tv = d.toTimeVal();
    cFunc(&tv);

and does a move rather than a copy. But since all you'd have to do now is

    timeval tv;
    d.split!("seconds", "usecs")(tv.tv_sec, tv.tv_usec);
    cFunc(&tv);

`toTimeVal` doesn't really buy us much. Sure, the code is slightly simpler with `toTimeVal`, but the difference is trivial, and this is exactly the sort of thing that Andrei usually rejects because it's a simple wrapper.

If it weren't so straightforward to initialize a `timeval` or `timespec` using `split`, then there probably would be a good argument for these, but it's so simple to just use `split` directly that I seriously question that they add enough value to be worth adding. I fully expect that if we asked Andrei, he would say no.

FreeSlave:
Well, that's just idea.
The benefit I thought of is that a user does not need to remember weird names like tv_sec and which of structs uses microseconds and which one uses nanoseconds.

schveiguy:
I'd be in favor of this not because the implementation isn't trivial (it is), but because getting it right is not trivial.

If I have a function that takes a `timeval_t` or `timespec_t`, I don't really want to have to look up each of the fields, know how they are used, and work out how to call split in the correct way. Conversion functions between different types are useful in many places, and take all the headache away.

src/core/time.d
```diff
@@ -1667,6 +1667,52 @@ public:
         }
     }
 
+    /++
+        Converts `Duration` to a `timeval` structure.
```
WalterBright: Please use Ddoc style comments.

FreeSlave: I'm not sure what do you mean by Ddoc style comments.
For me any /++, /** or /// comment is Ddoc comment.

ZombineDev: The only improvement that I could see is that you could add a [`Returns: `](https://dlang.org/spec/ddoc.html#standard_srction) section.

src/core/time.d
```diff
@@ -1667,6 +1667,52 @@ public:
...
+    +/
+    timeval toTimeVal() const nothrow @nogc
+    {
+        timeval value;
+        split!("seconds", "usecs")(value.tv_sec, value.tv_usec);
+        return value;
+    }
+
+    unittest
```
ZombineDev: Please add `///` on the line before, to make this into a documented example (https://dlang.org/spec/ddoc.html#using_ddoc_to_generate_examples).

src/core/time.d
```diff
@@ -1667,6 +1667,52 @@ public:
...
+    {
+        timeval value;
+        value.tv_sec = 1;
+        value.tv_usec = 5000;
+        assert(dur!"msecs"(1005).toTimeVal() == value);
+    }
+
+    version(StdDdoc)
```
ZombineDev: In druntime, you need to use `CoreDdoc`. (Look at the rest of the file for examples.)

ZombineDev:
@FreeSlave You can check with the `CyberShadow/DAutoTest` CI check the documentation output, to make sure that both of the functions' examples show in the docs.

ZombineDev:
Thanks @FreeSlave. I'll wait for the CIs to pass and then I think we will be good to go.

src/core/time.d
schveiguy: Walter was asking for DDOC sections, like @ZombineDev says:

```D
/++
       Returns: A `timeval` structure representing this `Duration`.
++/
```
And do the same to the other function as well.

FreeSlave: @schveiguy I just did it the same way as in toString()
```
Converts this `Duration` to a `string`
```
And doc comments for toString() does not have ddoc sections.

jmdavis: I honestly see zero point in using `Returns` when the whole comment is telling you what the thing is returning. Sections make sense when you're breaking up the comment but not so much when the comment is simple.

jmdavis:
Well, I still think that these are too trivial to make sense to add (the documentation for each function is longer than the function), but I guess that I'm the only one reviewing this that thinks that.

CyberShadow:
+1 Interoperability/conversion functions for OS types are good to have regardless of their triviality. They carry an important semantic meaning that's better standardized than implemented in user code.

CyberShadow:
Should there not be functions to perform the conversion in the opposite direction, in the form of a constructor or Duration static method?

schveiguy:
This PR I think is OK, but is stalled due to philosophical differences. Should we just close?

PR #1933, opened 2017-10-07
Fix 'unrecognised opcode ldcl/sdcl' in threadasm.S

ibuclaw:
According to the instruction manual, neither of these instructions exist:

http://math-atlas.sourceforge.net/devel/assembly/mips-iv.pdf

However, `ldc1` and `sdc1` do, and they have the correct instruction format also, so it does appear to be a typo.

@MartinNowak - as you are the original author, could you have a look to confirm?

### DIFF
--- a/src/core/threadasm.S
+++ b/src/core/threadasm.S
@@ -322,12 +322,12 @@ fiber_switchContext:
 #ifdef __mips_hard_float
 #define ALIGN8(val) (val + (-val & 7))
 #define BELOW (ALIGN8(6 * 8 + 4))
-    sdcl $f20, (0 * 8 - BELOW)($sp)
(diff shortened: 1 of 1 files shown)


ibuclaw:
@MartinNowak had time to have a look?

ibuclaw:
I should be able to validate this weekend (albeit in an emulator).

ZombineDev:
@ibuclaw I'm ok with merging, as I'm sure that even if this turns out to be wrong, you'll send a PR to fix it :P
After we don't have that many developers that can actually test D on MIPS.

ibuclaw:
@ZombineDev the GNU assembler for mips rejects this, and the uncanny similarity between 1 and l is suspect.  But yeah, if there's problems, then this code will be rewritten anyways.

PR #1934, opened 2017-10-08
fix Issue 17876 - [REG 2.074] Internal error when comparing inout(Foo[][]) with Foo[][]

rainers:
moved bad assertion that failed for arrays of arrays with mixed modifiers to the memcmp version which should work on values anyway.

Also fixed bad comparison via memcmp only comparing first element.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3454,7 +3454,7 @@ if (__traits(isScalar, T))
 // comparisons in the semantic analysis phase of CmpExp. The ordering
 // comparison is lowered to a call to this template.
 int __cmp(T1, T2)(T1[] s1, T2[] s2)
-if (!__traits(isScalar, T1))
+if (!__traits(isScalar, T1) && !__traits(isScalar, T2))
 {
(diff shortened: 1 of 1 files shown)


schveiguy:
Checking that both are the same type was implied in the template constraint, because only T1 is checked for 'isScalar'.

To have the same behavior, I think you need to check if both are not scalar. Otherwise:

```D
struct S
{
   int x;
   int opCmp(int y) { return x < y ? -1 : x > y; }
}

auto s = [S(1)];
auto i = [0];
assert(s[0] > i[0]); // works
assert(i[0] < s[0]); // works
assert(s > i); // works
assert(i < s); // error
```

There is an opportunity here, because really `s` and `i` should be comparable. But maybe that's good to leave for an enhancement.

rainers:
> To have the same behavior, I think you need to check if both are not scalar.

The constraint so far did not check the second argument, so it would run into the internal error later on (D does not have SFINAE). I've added it nevertheless.

PR #1938, opened 2017-10-11
core/threadasm.S: PPC64 fix undefined reference to core.thread.callWithStackShell

ibuclaw:
Looks like a copy-pasto from #988.

### DIFF
--- a/src/core/threadasm.S
+++ b/src/core/threadasm.S
@@ -52,7 +52,6 @@
 #if defined( USE_ABI_2 )
     .abiversion 2
 #endif
-    .globl  _D4core6thread18callWithStackShellFNbMDFPvZvZv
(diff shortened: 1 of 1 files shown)


ibuclaw:
Ping @redstar

ZombineDev:
As mentioned in https://github.com/dlang/druntime/pull/1933#issuecomment-336381942, I trust that @ibuclaw will follow through with another PR if needed.

ibuclaw:
@ZombineDev linker fails without this patch, and it looks obviously wrong.

redstar:
Yes, it is obviously wrong. In newer versions of LDC `callWithStackShell()` is implemented as inline assembler in `thread.d`.

PR #1939, opened 2017-10-12
Deprecate core.time.FracSec.

jmdavis:
The last function in std.datetime which uses FracSec has now been
removed from Phobos (it was deprecated previously), so this now
deprecates FracSec, which probably should have been deprecated a while
ago, but oh well. All of the functionality that used it was switched to
use Duration for simplicity.

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -3514,7 +3514,13 @@ unittest
 }
 
 
+// @@@DEPRECATED_2018-10@@@
 /++
+    $(RED Everything in druntime and Phobos that was using FracSec now uses
+          Duration for greater simplicity. So, FracSec has been deprecated.
+          It will be removed from the docs in October 2018, and removed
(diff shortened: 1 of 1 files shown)


ZombineDev:
@jmdavis Jenkins and DAutoTest are failing due to network issues with code.dlang.org, so you need to force-push to restart them (Jenkins could be restarted from the UI, but unfortunately there's no way to do this for DAutoTest).

jmdavis:
> Glad to see the old core.time types finally start to be phased out.

`TickDuration` will get the axe fairly soon too now that the std.datetime benchmarking functions are finally getting deprecated, but I'm thinking of waiting until the release after the old std.datetime benchmarking functions are deprecated so that there are fewer deprecation messages.

ZombineDev:
Agreed, `TickDuration` is used in more places (AFAIR) so it's better to be extra careful with it.

PR #1940, opened 2017-10-14
Added definition of WSAOVERLAPPED

marler8997:


src/core/sys/windows/winsock2.d
```diff
@@ -740,7 +740,21 @@ struct hostent
 }
 
 // Note: These are Winsock2!!
-struct WSAOVERLAPPED;
+struct WSAOVERLAPPED
+{
+    ulong Internal;
```
ZombineDev: In the Windows API these are defined as `ULONG_PTR`, and so D analog is either `size_t`, or `core.stdc.stdint.uintptr_t` if you want to be pedantic about the use of ISO C terminology.

ZombineDev:
BTW, in the future make sure to post a link to the definition of function/type you're adding in the commit message and PR description. In this case:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms741665(v=vs.85).aspx

src/core/sys/windows/winsock2.d
ZombineDev: Or perhaps better, use the Windows typedef types directly - that way there should be no room for error, or at least only one place to fix.

rainers:
WSAOVERLAPPED should just alias OVERLAPPED, as this is what's done in the SDK/MinGW headers, too.

PR #1941, opened 2017-10-15
add core.math.toPrec intrinsic

MartinNowak:
- to force rounding to specific float precision

dmd PR: https://github.com/dlang/dmd/pull/7219


MartinNowak:
Fails on 32-bit cause the ABI returns in X87-ST0, so dmd just emits `fld*`.

We could either store values on the stack :/, teach dmd to obey casts instead of adding intrinsics, lower the intrinsics in e2ir to dmd backend ops, or throw away the dmd backend.

ibuclaw:
Retesting this again in #2864.

PR #1944, opened 2017-10-17
fix issue 17108 Associative array byKeyValue is unsafe

somzzz:


src/object.d
```diff
@@ -2089,12 +2089,15 @@ auto byKeyValue(T : V[K], K, V)(T aa) pure nothrow @nogc
 {
     import core.internal.traits : substInout;
 
+    static auto safeAARangeEmpty(AARange rr) @safe { return () @trusted { return _aaRangeEmpty(rr); } (); }
```
MetaLang: I'm not sure it's necessary to have these functions as they're only used once. I think an `@trusted` block should suffice.

src/object.d
```diff
@@ -2104,17 +2107,23 @@ auto byKeyValue(T : V[K], K, V)(T aa) pure nothrow @nogc
...
             return Pair(_aaRangeFrontKey(r),
                         _aaRangeFrontValue(r));
         }
-        void popFront() { _aaRangePopFront(r); }
+        void popFront() { return safeAARangePopFront(r); }
         @property Result save() { return this; }
     }
 
-    return Result(_aaRange(cast(void*)aa));
+    return () @trusted {return Result(_aaRange(cast(void*)aa)); } ();
```
MetaLang: Space between `{` and `return`.

src/object.d
```diff
@@ -2186,6 +2192,21 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
     return (*aa).get(key, defaultValue);
 }
 
+unittest
```
wilzbach: Why don't you simply make the unittest `@safe`?
`__compiles` is an anti-pattern and should only be used if absolutely necessary.

src/object.d
```diff
@@ -2104,17 +2104,23 @@ auto byKeyValue(T : V[K], K, V)(T aa) pure nothrow @nogc
...
             return Pair(_aaRangeFrontKey(r),
                         _aaRangeFrontValue(r));
         }
-        void popFront() { _aaRangePopFront(r); }
+        void popFront() { return () @trusted { return _aaRangePopFront(r); } (); }
         @property Result save() { return this; }
     }
 
-    return Result(_aaRange(cast(void*)aa));
+    return () @trusted { return Result(_aaRange(cast(void*)aa)); } ();
```
wilzbach: The constructor of `Result` should be safe, so you could try to move the lamda inside.

src/object.d
```diff
@@ -2094,7 +2094,7 @@ auto byKeyValue(T : V[K], K, V)(T aa) pure nothrow @nogc
         AARange r;
 
     pure nothrow @nogc:
-        @property bool empty() { return _aaRangeEmpty(r); }
+        @property bool empty() @safe { return () @trusted { return _aaRangeEmpty(r); } (); }
```
wilzbach: For DMD there is an overhead in these lamdas and in this case there is no need for the lamda if you replace `@safe` with `@trusted`

MetaLang: I thought Kenji fixed the compiler so when the compiler detects this pattern it always inlines the lambda.

schveiguy: This should simply be marked `@safe`, and the implementation of `_aaRangeEmpty` marked `@safe`.  We should avoid `@trusted` wherever we can.

src/object.d
```diff
@@ -2104,17 +2104,23 @@ auto byKeyValue(T : V[K], K, V)(T aa) pure nothrow @nogc
...
-                @property ref key() inout { return *cast(substInout!K*)keyp; }
-                @property ref value() inout { return *cast(substInout!V*)valp; }
+                @property auto key() inout
+                {
+                    return () @trusted { return *cast(substInout!K*)keyp; } ();
+                };
+                @property auto value() inout
+                {
+                    return () @trusted { return *cast(substInout!V*)valp; } ();
+                };
```
schveiguy: There is a problem here in that the AA can rehash the buckets, making any existing range point at invalid elements. I think in order to make this `@safe` we need to verify the range is still pointing at a valid element.

In particular the fetching of the value returns the pointer + key.sizeof, which even if the element pointer is null is going to be sketchy.

I still prefer to see this not being a `@trusted` escape, but instead mark the implementation as `@safe` and have the implementation use an escape.

We also have to be careful here of any kind of postblit in `K` or `V`. If you trust the whole thing, then it is going to inadvertently trust that part too.

src/object.d
```diff
@@ -2186,6 +2192,21 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
...
+            int[string] aa;
+            int a;
+            foreach (val; aa.byKeyValue)
+            {
+                ++aa[val.key];
+                a = val.value;
+            }
+        }
+    ));
+}
```
schveiguy: I'd like to also make sure that the following *doesn't* compile:

```D
struct BadValue
{
   int x;
   this(this) { *cast(ubyte*)(null + 100000) = 5; } // not @safe
   alias x this;
}

BadValue[int] aa;
auto x = aa.byKeyValue.front;
() @safe {auto y = x.value; }
```

src/object.d
andralex: @schveiguy the "abandoned" elements will be still valid memory. All we need to make sure is they are cleared properly (by means of `move` in all likelihood). That would make old ranges iterate the wrong elements but not unsafe.

schveiguy: The abandoned elements themselves are valid, and are set to null.

However, the code for fetching the value looks like this (from here: https://github.com/somzzz/druntime/blob/84af68f87d077d9632e7dc58da04670316d4219c/src/rt/aaA.d#L705):

```D
void* _aaRangeFrontValue(Range r)
    {
        return r.buckets[r.idx].entry + r.valoff;
    }
```

Here, `r.buckets` is really `r.impl.buckets` (`impl` points at the actual AA instance, so we are OK there). The `r.idx` is the range's local copy of which bucket it is currently looking at.

entry is possibly `null` (shouldn't be on a valid AA range, but if a rehash happens, it could be). So we are adding an arbitrary amount to a `null` pointer, and dereferencing that without considering it (unsafe).

Another consideration that I had figured out in the past but forgot to mention it here, it's very possible `r.idx >= r.impl.buckets.length` if the rehash has shrunk the array. Now we are definitely in unsafe territory.

All this can be fixed by returning `null` if `r.idx >= r.buckets.length` or the appropriate bucket pointer is `null` (just don't add the offset). And these aren't bounds-checked, since druntime is compiled in release mode.

src/object.d
```diff
@@ -2104,17 +2104,23 @@ auto byKeyValue(T : V[K], K, V)(T aa) pure nothrow @nogc
                 private void* keyp;
                 private void* valp;
 
-                @property ref key() inout { return *cast(substInout!K*)keyp; }
-                @property ref value() inout { return *cast(substInout!V*)valp; }
+                @property auto key() inout
+                {
+                    return () @trusted { return *cast(substInout!K*)keyp; } ();
```
andralex:
```d
auto p = @trusted { return cast(substInout!K *) keyp; }();
return *p;
```

src/object.d
```diff
@@ -2104,17 +2104,23 @@ auto byKeyValue(T : V[K], K, V)(T aa) pure nothrow @nogc
...
+                };
+                @property auto value() inout
+                {
+                    return () @trusted { return *cast(substInout!V*)valp; } ();
```
andralex: similar

src/object.d
schveiguy: Looking at this more closely, it's really the generation of the `Pair` struct that is where the safety issues may lie.

edi33416:
Sorry to be nitpicking.

- I think that `return (() @trusted => *cast(substInout!K *) keyp)();` is more pleasant to the eye than `return () @trusted { return *cast(substInout!K*)keyp; } ();`
- The style checker should complain about wanting a space between the `cast(T)` and the value
ex. `cast(substInout!K*)keyp` should be `cast(substInout!K*) keyp`; at least the `phobos` style checker does.

Aside from the nits, LGTM.

wilzbach:
> The style checker should complain about wanting a space 

FYI: atm there is no style checking in place at druntime and DMD :/
(Also Walter prefers different styles for DMD than the ones that are used for Phobos).

somzzz:
@schveiguy @andralex ping

src/rt/aaA.d
```diff
@@ -699,16 +699,25 @@ extern (C) pure nothrow @nogc
 
     void* _aaRangeFrontKey(Range r)
     {
+        if (r.idx >= r.dim)
```
rainers: Marking the function `@safe` enables bounds checking on the buckets-access, so I think the test here is unnecessary. A range error is probably better than an access violation later.

Interestingly, x64 code elides the bounds check if this condition exists, but keeps both for 32-bit code.

src/object.d
```diff
@@ -2048,21 +2048,25 @@ auto byKey(T : V[K], K, V)(T aa) pure nothrow @nogc
         AARange r;
 
     pure nothrow @nogc:
-        @property bool empty() { return _aaRangeEmpty(r); }
-        @property ref front() { return *cast(substInout!K*)_aaRangeFrontKey(r); }
-        void popFront() { _aaRangePopFront(r); }
+        @property bool empty()  @safe { return _aaRangeEmpty(r); }
+        @property auto front()
```
rainers: This changes the return value from a reference to a copy. Is this expected? Can it break code that has been using ref for iteration?
While a mutable reference is not a good idea to begin with, using a const reference might be a bit more conservative.

somzzz: An intermediate version of this PR didn't allow ref as return type so at that point I changed it to auto.
But yes, it's better to preserve the original. Thanks for pointing it out!

andralex:
@schveiguy I'll merge this since your concern has been addressed, holler if there's anything left

schveiguy:
Thanks, @andralex, I didn't notice the updates. They all look good! Thanks @somzzz

wilzbach:
Ehm just copy/pasting @klickverbot's examples of what shouldn't be allowed from https://github.com/dlang/druntime/pull/1644

```
int main() @safe {
  auto a = ["foo": 1];
  auto r = a.byValue;
  r.popFront;
  return r.front; // oops
}
```

and another one

```
struct Foo {
  int[int] aa;
  auto opCast() pure nothrow @nogc { *cast(uint*)0xdeadbeef = 0xcafebabe; return null; }
  alias aa this;
}

int main() @safe {
    Foo f;
    return !f.byKey.empty;
}
```

While 2.077.1 correctly emits warnings that this isn't `@safe`, this code now segfaults.
-> https://issues.dlang.org/show_bug.cgi?id=18071

schveiguy:
The first one, I think is expected to segfault, as you are dereferencing a null pointer (which is allowed in `@safe` code). This is equivalent to:

```D
void foo() @safe
{
   int *x = null;
   int y = *x; // oops, but it's ok.
}
```

The second, I think can be fixed by assigning to a local AA before casting. I'll look at making a PR.

Again IFTI is confusing here, as we really are trying to add a member to the AA, but instead end up adding members to items that aren't really AAs.

PR #1945, opened 2017-10-18
cleanup core.sys.*.netinet headers

MartinNowak:
- copy newly added headers to import folder
- fix core.sys.linux.sys.netinet pkg to core.sys.linux.netinet


MartinNowak:
Merging as vibe.d is depending on those for 2.077.0.

PR #1946, opened 2017-10-18
Remove extern (C) from structs, unions and classes

RazvanN7:
This currently blocks [1].   

[1] https://github.com/dlang/dmd/pull/7229

### DIFF
--- a/src/core/sys/posix/netdb.d
+++ b/src/core/sys/posix/netdb.d
@@ -30,7 +30,6 @@ else version (WatchOS)
     version = Darwin;
 
 version (Posix):
-extern (C):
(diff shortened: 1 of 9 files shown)


MartinNowak:
Whatever you're planning there will break a lot of code and is hardly acceptable as such.
Does `extern(C)` on the outside of the struct also affect (static) member methods, or why do you want to change this?

RazvanN7:
Some files start with extern(C) : ... which applies extern to all the declarations in that file, some of which are structs, unions or classes. This makes it impossible to build phobos with the patch for Issue 17748. I am keeping all the function extern (C).

PR #1949, opened 2017-10-22
add aligned_alloc

9il:
Digitalmars runtime does not implement it. But it is not a reason to do not include into C stdlib interface header.

### DIFF
--- a/src/core/stdc/stdlib.d
+++ b/src/core/stdc/stdlib.d
@@ -153,6 +153,13 @@ else
     void    srand(uint seed);
 }
 
+version (CRuntime_Glibc)
+{
+    version = AlignedAllocSupported;
+}
+// Add other C runtimes as necessary...
+else {}
+
 // We don't mark these @trusted. Given that they return a void*, one has
 // to do a pointer cast to do anything sensible with the result. Thus,
 // functions using these already have to be @trusted, allowing them to
@@ -165,8 +172,12 @@ void*   calloc(size_t nmemb, size_t size);
 void*   realloc(void* ptr, size_t size);
 ///
 void    free(void* ptr);
-/// (since C11)
-void* aligned_alloc(size_t alignment, size_t size);
+
+/// since C11
+version(AlignedAllocSupported)
+{
+    void* aligned_alloc(size_t alignment, size_t size);
+}
 
 ///
 noreturn abort() @safe;
--- a/src/core/stdc/stdlib.d
+++ b/src/core/stdc/stdlib.d
@@ -26,6 +26,10 @@ else version (TVOS)
 else version (WatchOS)
     version = Darwin;
 
+version (CRuntime_Glibc)
+    version = AlignedAllocSupported;
+else {}
+
 extern (C):
 @system:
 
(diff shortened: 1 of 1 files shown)


BBasile:
There's a version for DigitalMars in experimental allocators. Maybe you could move it in the runtime and defines it for `version(CRuntime_DigitalMars)`.

rainers:
aligned_alloc is C11 and hasn't made it into all C runtimes yet. The Microsoft libraries have _aligned_malloc instead. So it's probably best to version it depending on the C runtime and just include what's available or easily aliasable/wrappable. Reimplementations are not supposed to be part of the stdc headers, though.

RazvanN7:
As far as I could tell, only Cruntime_Glibc should include this function and it's windows counterpart. I don't know what is the procedure for adding new functions to these headers. How should we proceed with this PR?

PetarKirov:
> There's a version for DigitalMars in experimental allocators. Maybe you could move it in the runtime and defines it for version(CRuntime_DigitalMars).

In general, the goal of `core.sys.*` is to provide either direct bindings to OS / libc libraries, or if the public API is implemented as a macro, to implement it as a D function (which is should be quite rare).

So, such platform abstraction code should belong in a different module (`std.experimental.allocators` now, or perhaps in a future `core.memory.allocators` package in druntime).

> The Microsoft libraries have `_aligned_malloc` instead

> As far as I could tell, only Cruntime_Glibc should include this function and it's windows counterpart. I don't know what is the procedure for adding new functions to these headers.

The Windows counter-part was recently added by Iain here: https://github.com/dlang/druntime/blob/v2.098.0/src/core/sys/windows/stdc/malloc.d and that's the right place IMO.

> How should we proceed with this PR?

@RazvanN7  As for runtimes that actually implement the [C11](https://en.cppreference.com/w/c/memory/aligned_alloc) / [C++17](https://en.cppreference.com/w/cpp/memory/c/aligned_alloc) `aligned_alloc` we can proceed like this:

```d
version (CRuntime_Glibc)
{
    version = AlignedAllocSupported;
}
// Add other C runtimes as necessary...
else {}
    
version (AlignedAllocSupported)
{
    void* aligned_alloc(size_t alignment, size_t size);
}
```

RazvanN7:
@PetarKirov I think this is ready to go now.

RazvanN7 (approved):
Approving my own additions :-"

PetarKirov (approved):

src/core/stdc/stdlib.d
```diff
@@ -153,6 +153,13 @@ else
     void    srand(uint seed);
 }
 
+version (CRuntime_Glibc)
+{
+    version = AlignedAllocSupported;
+}
+// Add other C runtimes as necessary...
+else {}
+
```
PetarKirov, line 162: On second look, I think it's kind of awkward to have this version block sitting there middle. It should either be in the beginning of the file (e.g. after the `Darwin` version block) or right before the only use of `AlignedAllocSupported`.

src/core/stdc/stdlib.d
```diff
@@ -153,6 +153,13 @@ else
     void    srand(uint seed);
 }
 
+version (CRuntime_Glibc)
+{
+    version = AlignedAllocSupported;
+}
+// Add other C runtimes as necessary...
+else {}
```
PetarKirov, line 161: Could be a bit more elegant:

```suggestion
version (CRuntime_Glibc)
    version = AlignedAllocSupported;
else {}
```

(The comment was just for the discussion in this PR, I think it's out of place in the code.)

PR #1952, opened 2017-10-24
String switch template for lowering

somzzz:
Compiler PR to follow.

Continuing to remove dependencies between the compiler and druntime -  this time for string switches. The template here replaces the functions in rt/switch_.d (to be deleted in a different PR) which were called from the compiler.


src/object.d
```diff
@@ -3722,6 +3722,91 @@ template _arrayOp(Args...)
     alias _arrayOp = arrayOp!Args;
 }
 
+// switch template - replaces code in rt/switch_.d
+int __switch(T)(T[] condition, T[][] caseLabels...)
+{
+    for (uint i; i < caseLabels.length; i++)
+    {
+        if (__cmp(condition, caseLabels[i]) == 0){
```
andralex: brace should be on its own line, I'll fix this

src/object.d
```diff
@@ -3722,6 +3722,91 @@ template _arrayOp(Args...)
     alias _arrayOp = arrayOp!Args;
 }
 
+// switch template - replaces code in rt/switch_.d
+int __switch(T)(T[] condition, T[][] caseLabels...)
+{
+    for (uint i; i < caseLabels.length; i++)
```
rainers: Similar to _d_switch_string this should take advantage of the string array being sorted, i.e. copy the binary search. If the array size could be made a part of the template arguments, a switch with few cases could be special cased to linear search.

DmitryOlshansky: Indeed.
Also using a variation of Trie comes to mind - at least a level or two of switching on the first (second) character may help a bit.

Other tricks involve first switching on condition.length.

rainers:
Before dumping more stuff into object.d can we please consider alternative strategies? Martin used a template wrapper for array code (https://github.com/dlang/druntime/blob/master/src/object.d#L3719), I proposed public static imports in object.d at DConf, e.g.
```
    public static import core.compiler.switch_;
```
can be accessed everywhere using fully qualified names and making this import lazy should be pretty straight-forward.

andralex:
@rainers the incremental cost of parsing is virtually zero (I measured several times), so there is no need for that kind of trick. Top-level imports in object.d should be avoided at least with the current implementation because opening additional files does cause a measurable slowdown.

somzzz:
I updated the template to use the old binary search and we can work on improving that. It's performing some extra checks and I'm not sure the performance is indeed any better than a linear search.

Also, the array is sorted because of the extra work done in the compiler. Code for dmd lowering here:
https://github.com/somzzz/dmd/commit/f7f15156099ba337c917e7cc110e54a683190bd3
Quite ugly imo.

src/object.d
```diff
@@ -3722,6 +3722,186 @@ template _arrayOp(Args...)
...
+            }
+        }
+    }
+}
+body
+{
+    size_t low  = 0;
+    size_t high = caseLabels.length;
+
+    if (high &&
```
andralex: The `in` contract asserts `high` cannot be zero, so this can be eliminated.

src/object.d
```diff
@@ -3722,6 +3722,186 @@ template _arrayOp(Args...)
...
+        condition.length >= caseLabels[0].length &&
+        condition.length <= caseLabels[high - 1].length)
```
andralex: I don't know what these tests do. From what I can tell the labels are sorted lexicographically, not by length.

src/object.d
```diff
@@ -3722,6 +3722,186 @@ template _arrayOp(Args...)
...
+    {
+        // Looking for 0 length string, which would only be at the beginning
+        if (condition.length == 0)
```
andralex: This should not be special-cased, will emerge as a normal search.

src/object.d
```diff
@@ -3722,6 +3722,186 @@ template _arrayOp(Args...)
...
+            return 0;
+
+        T c1 = condition[0];
```
andralex: Who uses c1?

src/object.d
```diff
@@ -3722,6 +3722,186 @@ template _arrayOp(Args...)
...
+
+        // Do binary search
+        while (low < high)
+        {
+            auto mid = (low + high) >> 1;
+            auto pca = caseLabels[mid];
```
andralex: avoid copies for the sake of generality, just use caseLabels[mid] throughout

src/object.d
```diff
@@ -3722,6 +3722,186 @@ template _arrayOp(Args...)
...
+        // Looking for 0 length string, which would only be at the beginning
+        if (condition.length == 0)
+            return 0;
+
+        T c1 = condition[0];
+
+        // Do binary search
+        while (low < high)
+        {
+            auto mid = (low + high) >> 1;
```
andralex: use `/ 2`

src/object.d
```diff
@@ -3722,6 +3722,186 @@ template _arrayOp(Args...)
...
+            auto pca = caseLabels[mid];
+            auto c   = cast(sizediff_t)(condition.length - pca.length);
```
andralex: no need for this test - it is done in the first line of `__cmp`

src/object.d
```diff
@@ -3722,6 +3722,186 @@ template _arrayOp(Args...)
...
+            if (c == 0)
+            {
+                c = cast(ubyte)c1 - cast(ubyte)pca[0];
```
andralex: this duplicates the first comparison in `__cmp`, remove

src/object.d
```diff
@@ -3722,6 +3722,186 @@ template _arrayOp(Args...)
...
+            if (c < 0)
+            {
+                high = mid;
+            }
+            else
+            {
+                low = mid + 1;
+            }
+        }
+    }
```
andralex: This is longer than the canonical implementation. Please rewrite with an eye for efficiency, simplicity, and clarity. FWIW all you need is:

```d
size_t low = 0, high = caseLabels.length;
while (low < high)
{
    auto mid = (low + high) / 2;
    auto r = __cmp(condition, caseLabels[mid]);
    if (r == 0) return cast(int) mid;
    if (r > 0) low = mid + 1;
    else high = mid;
}
return -1;
```

Since we know `caseLabels.length > 0`, we can save one comparison:

```d
size_t low = 0, high = caseLabels.length;
do
{
    auto mid = (low + high) / 2;
    auto r = __cmp(condition, caseLabels[mid]);
    if (r == 0) return cast(int) mid;
    if (r > 0) low = mid + 1;
    else high = mid;
}
while (low < high);
return -1;
```

src/object.d
```diff
@@ -3722,6 +3722,186 @@ template _arrayOp(Args...)
...
+            case "D_Version2":      return 2;
+            case "none":            return 3;
+            case "all":             return 4;
+            default:                return 5;
+        }
+    }
+    int rc = bug5381("none"d);
+    assert(rc == 3);
+}
+
```
andralex: Looks like more unittests are needed given that the bugs involving lengths were not caught.

andralex:
After chatting with Lucia and looking at https://github.com/dlang/dmd/blob/master/src/ddmd/expression.d#L4400, it seems strings are not sorted lexicographically. Instead, they are sorted by length first, and for each equivalence class they are sorted lexicographically. So the binary search code would go something like:

```d
size_t low = 0, high = caseLabels.length;
do
{
    auto mid = (low + high) / 2;
    int r = void;
    if (condition.length == caseLabels[mid].length)
    {
        r = __cmp(condition, caseLabels[mid]);
        if (r == 0) return cast(int) mid;
    }
    else
    {
        // generates better code than "expr ? 1 : -1" on dmd and gdc, same with ldc
        r = ((condition.length > caseLabels[mid].length) << 1) - 1;
    }
    if (r > 0) low = mid + 1;
    else high = mid;
}
while (low < high);
return -1;
```

Far as I can tell this does no redundant work.

WalterBright:
Just use the original algorithm: https://github.com/dlang/druntime/blob/master/src/rt/switch_.d#L90 No need to reinvent it.

andralex:
The original implementation was rather long in the tooth, it's the initial one proposed in this PR for which I pointed out a number of inefficiencies. I'd rather go with an improved implementation along the lines proposed in the comments.

rainers:
> the incremental cost of parsing is virtually zero (I measured several times), so there is no need for that kind of trick. Top-level imports in object.d should be avoided at least with the current implementation because opening additional files does cause a measurable slowdown.

I wanted to see this in action myself, these are the results of `avgtime -r100 dmd -I. -o- test.d` with an empty test.d on my system (i7-2670QM with SSD, Win10):

 | times are ms | object.d  | empty  | import | import_lazy  | no_unittest
|-|---------|---------|---------|--------------|------------------------------------------------
Median time    | 37.8765    | 30.1845        | 37.923      | 37.058            | 37.0465
Avg time       | 38.1189    | 30.2449        | 38.1245     | 37.2388           | 37.2537
Std dev.       | 1.46114    | 0.494134       | 1.18381     | 1.02198           | 1.25786
**Minimum**        | **36.723**     | **29.34**          | **37.049**      | **36.183**            | 36.173
Maximum        | 50.823     | 32.206         | 48.234      | 45.756            | 48.294

- object.d: the file from dmd 2.076
- empty: object.d just a module statement
- import: the array template code (about 10% of object.d) moved into another file and globally imported
- import_lazy: same as import, but import wrapped in an unused template
- no_unittest: object.d stripped of all unittest code (about 40%)

Numbers will be slightly worse with a HD instead of an SSD, but after the first invocation the file cache takes over anyway. 
Is a current degredation of less than 1ms worth dismissing a better source organization that can be taken advantage of later for better performance if the code is unused (or even now depending on the mechanism used)?

andralex:
>Is a current degradation of less than 1ms worth dismissing a better source organization that can be taken advantage of later for better performance if the code is unused (or even now depending on the mechanism used)?

So the speed is a non-issue.

Currently object.d sits at 4082 lines, which makes it the seventh largest in druntime (e.g. thread.d  has 5482 and and time.d has 4725 lines). Phobos has 22 (!) files larger:

```
   12039 ../phobos/std/range/package.d
   11080 ../phobos/std/internal/unicode_tables.d
   10987 ../phobos/std/datetime/systime.d
   10581 ../phobos/std/datetime/date.d
   10446 ../phobos/std/uni.d
    9131 ../phobos/std/datetime/interval.d
    8317 ../phobos/std/math.d
    8110 ../phobos/std/typecons.d
    8089 ../phobos/std/traits.d
    6954 ../phobos/std/string.d
    6370 ../phobos/std/conv.d
    6058 ../phobos/std/format.d
    5301 ../phobos/std/internal/unicode_decomp.d
    5234 ../phobos/std/stdio.d
    5189 ../phobos/std/algorithm/iteration.d
    5109 ../phobos/std/net/curl.d
    4682 ../phobos/std/parallelism.d
    4610 ../phobos/std/algorithm/searching.d
    4468 ../phobos/std/algorithm/sorting.d
    4352 ../phobos/std/file.d
    4235 ../phobos/std/datetime/timezone.d
    4115 ../phobos/std/path.d
```

We're in good shape as far as sheer size is concerned. Worrying right now about the size of object.d, and overengineering it for the sake of minor potential future benefits, would be poor craftmanship. We should start worrying about size when it gets beyond 10KLOC.

With regard to better organizing the file, indeed things like switch and typeinfo have little in common. I'd be okay in the future with pull requests that refactor code for better organization, but at the moment I worry more about how functionality migrates in there, to the carve the appropriate pieces cleverly.

andralex:
ping @somzzz

rainers:
> I'd be okay in the future with pull requests that refactor code for better organization

Thanks, good to know. That should not be too difficult to do later with Martins pattern.

somzzz:
Compiler changes can be seen here : https://github.com/somzzz/dmd/commit/59c840255b4c1ed8a954e56e847c0fea9d8150bf

somzzz:
@andralex please take a look, thanks!

src/object.d
```diff
@@ -3722,6 +3722,200 @@ template _arrayOp(Args...)
...
+    {
+        int r = void;
+        if (condition.length == caseLabels[$ / 2].length)
+        {
+            r = __cmp(condition, caseLabels[$ / 2]);
+            if (r == 0) return cast(int) caseLabels.length / 2;
+        }
+        else
+        {
+            r = ((condition.length > caseLabels[$ / 2].length) << 1) - 1;
```
andralex: Insert: `// Equivalent to (but faster than) condition.length > caseLabels[$ / 2].length ?  1 : -1`

src/object.d
```diff
@@ -3722,6 +3722,200 @@ template _arrayOp(Args...)
...
+        return -1;
+    }
+    else static if (caseLabels.length == 1)
+    {
+        return __cmp(condition, caseLabels[0]) == 0 ? 0 : -1;
+    }
+    // To be adjusted after measurements
+    else static if (caseLabels.length < 7)
+    {
+        int r = void;
```
andralex: Insert comment: `// Compile-time inlined binary search.`

src/object.d
```diff
@@ -3722,6 +3722,200 @@ template _arrayOp(Args...)
...
+        else
+        {
+            // Search the right side
+            r = __switch!(T, caseLabels[$ / 2 + 1 .. $])(condition);
+            return r != -1 ? cast(int) (caseLabels.length / 2 + 1 + r) : -1;
+        }
+    }
+    else
+    {
+        // Binary search
```
andralex: `// Run-time binary search in a static array of labels.`

andralex:
I noticed there's a bit of churn caused by the out-of-band value `-1` returned. I think things can be made more efficient if the returned value for failure to find is `caseLabels.length`. That should cut a number of comparisons. @somzzz could you please do that on the compiler side?

src/object.d
```diff
@@ -3722,6 +3722,200 @@ template _arrayOp(Args...)
...
+    assert(rc == 3);
+
+    rc = bug5381("none");
+    assert(rc == 4);
+
+    rc = bug5381("all");
+    assert(rc == 5);
+
+    import core.stdc.stdio;
+    printf("%d\n", bug5381("nonerandom"));
```
rainers: printf left over from debugging?

somzzz: My bad, will remove.

src/object.d
```diff
@@ -3722,6 +3722,201 @@ template _arrayOp(Args...)
     alias _arrayOp = arrayOp!Args;
 }
 
+// switch template - replaces code in rt/switch_.d
+int __switch(T, caseLabels...)(T[] condition)
```
schveiguy: It would be good to have a description here in comments of how this function is used by the compiler, especially since rt/switch_.d is going to be removed.

ibuclaw: Starting with copying/pasting the current documentation from rt/switch_.d

https://github.com/dlang/druntime/blob/970b5b2a3609873c1a4df377773a0dfd4524e5ba/src/rt/switch_.d#L18-L26

It looks like you just need to replace `table` with `caseLabels` and `ca` with `condition`.

schveiguy: Right, but a good thing to note here, especially since it was confusing to most reviewers, is that the "sorted array" is sorted in an odd way (length first, then lexicographically).

andralex:
@ibuclaw @schveiguy thanks for the suggestions, don't forget the incremental work on actually performing the desired change is very low. I added the comments myself. Thx!

schveiguy:
Ugh, whitespace checker has puked on the comments. I'll see if I can fix it.

schveiguy:
fixed trailing whitespace and deduplicated unit tests. Also expanded on the description of how the case labels are sorted.

andralex:
@schveiguy thx!!

stefan-koch-sociomantic:
@somzzz this violates the style-guide.

schveiguy:
@stefan-koch-sociomantic More specificity is appreciated, thanks.

src/object.d
```diff
@@ -3722,6 +3722,133 @@ template _arrayOp(Args...)
...
+                r = __cmp(condition, cases[mid]);
+                if (r == 0) return cast(int) mid;
+            }
+            else
+            {
+                // Generates better code than "expr ? 1 : -1" on dmd and gdc, same with ldc
+                r = ((condition.length > cases[mid].length) << 1) - 1;
+            }
+
+            if (r > 0) low = mid + 1;
```
stefan-koch-sociomantic: newline please

src/object.d
```diff
@@ -3722,6 +3722,133 @@ template _arrayOp(Args...)
...
+            else high = mid;
```
stefan-koch-sociomantic: here to

andralex: According to `git grep 'else .*;' src/**/*.d` the extant style allows it, so I think we're in good shape.

stefan-koch-sociomantic: It's still different from the surrounding code.

MartinNowak:
@andralex, @rainers regarding the size issue:
Since this is a single template function it seems fine to just add it to object.d.
More complex stuff such as the vectorized array ops or e.g. an templated AA is much better off in a separate module.
At some point we could take all the small compiler/language support functions and move them to a separate module that's either always publicly imported from object or lazily loaded by stubs in object.
Since those are template functions it shouldn't have much impact on the binary size.

src/object.d
```diff
@@ -3722,6 +3722,133 @@ template _arrayOp(Args...)
...
+    else
+    {
+        // Run-time binary search in a static array of labels.
+        static const T[][caseLabels.length] cases =  [ caseLabels ];
+        size_t low = 0;
+        size_t high = cases.length;
+
+        do
+        {
+            auto mid = (low + high) / 2;
```
MartinNowak: Unfortunate that nobody saw this during the review, this binary search does not depend on the template arguments, but is duplicated in every instance of the `__switch` template.
In order to reuse template bloat, this code should be factored out into a non-templated function.

src/object.d
```diff
@@ -3722,6 +3722,133 @@ template _arrayOp(Args...)
...
+/*
+ * Support for switch statements switching on strings.
+ * Params:
+ *      caseLabels = sorted array of strings generated by compiler. Note the
+                   strings are sorted by length first, and then lexicographically.
+ *      condition = string to look up in table
+ * Returns:
+ *      index of match in caseLabels, -1 if not found
+*/
+int __switch(T, caseLabels...)(T[] condition)
```
MartinNowak: condition should be `scope` as it's not escaped

MartinNowak:
See fixup https://github.com/dlang/druntime/pull/1973

PR #1954, opened 2017-10-24
Mark GC addrOf and sizeOf as nogc

edi33416:
These functions do not produce any garbage.
`GCAllocator.resolveInternalPointer` needs this change in order to be `@nogc`

@andralex


andralex:
I think some others could be converted in the future, too. Leaving this for later. cc @MartinNowak

PR #1955, opened 2017-10-24
Make GC addrof and sizeof nogc

edi33416:
Add `@nogc` to the actual `GC` functions, as I have forgotten in #1954 

@andralex

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -595,14 +595,14 @@ struct GC
      * Returns:
      *  The base address of the memory block referenced by p or null on error.
      */
-    static inout(void)* addrOf( inout(void)* p ) nothrow /* FIXME pure */
+    static inout(void)* addrOf( inout(void)* p ) nothrow @nogc /* FIXME pure */
     {
         return cast(inout(void)*)gc_addrOf(cast(void*)p);
     }
 
 
     /// ditto
-    static void* addrOf(void* p) pure nothrow
+    static void* addrOf(void* p) pure nothrow @nogc
     {
         return gc_addrOf(p);
     }
@@ -621,14 +621,14 @@ struct GC
      * Returns:
      *  The size in bytes of the memory block referenced by p or zero on error.
      */
-    static size_t sizeOf( in void* p ) nothrow
+    static size_t sizeOf( in void* p ) nothrow @nogc /* FIXME pure */
     {
         return gc_sizeOf(cast(void*)p);
     }
 
 
     /// ditto
-    static size_t sizeOf(void* p) pure nothrow
+    static size_t sizeOf(void* p) pure nothrow @nogc
     {
         return gc_sizeOf( p );
     }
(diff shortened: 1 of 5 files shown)


src/core/memory.d
```diff
@@ -595,14 +595,14 @@ struct GC
      * Returns:
      *  The base address of the memory block referenced by p or null on error.
      */
-    static inout(void)* addrOf( inout(void)* p ) nothrow /* FIXME pure */
+    static inout(void)* addrOf( inout(void)* p ) pure nothrow @nogc
```
edi33416: Any reason why pure should still be omitted?

src/core/memory.d
```diff
@@ -621,14 +621,14 @@ struct GC
      * Returns:
      *  The size in bytes of the memory block referenced by p or zero on error.
      */
-    static size_t sizeOf( in void* p ) nothrow
+    static size_t sizeOf( in void* p ) pure nothrow @nogc
```
edi33416: Same here

edi33416:
@andralex should I revert #1954 and put everything in this one?

rainers:
You should also add these new requirements to gc.interface as well as the gc implementations. The existing ones are also missing from gc.proxy.

While adding nogc to addrOf and sizeOf is very likely to be in line with the spec ("does not allocate memory on the GC heap"), that is not so clear for addRange or addRoot as this very much depends on the GC implementation.

My impression is that everyone has a different idea what nogc means. Some think that it forbids any call into the GC so you can link without it (I believe Walter is in that group). Others consider not running a collection to be the crucial point, but even malloc might not do this for a concurrent GC. 

Is there a consensus on this?

src/core/memory.d
rainers: With the current rules for pure, the second call could be elided by the compiler:
```
    void* p = GC.malloc(10000);
    void* q = GC.addrOf(p);
    GC.free(p);
    void* r = GC.addrOf(p);
    assert(r is null);
```

edi33416: So I should keep the `/* FIXME pure */`.

edi33416:
@rainers thank you for pointing out the missing spots.

Regarding the `nogc` meaning, I don't know whether there is a consensus on this, but imho a function that does not produce garbage should be marked as `nogc`.

ZombineDev:
Since @andralex merged the other PR, I think he's on the same page w.r.t. the meaning of `@nogc`.

andralex:
>Is there a consensus on this?

nogc = "does not allocate memory that will need the garbage collector to free it"

@edi33416 we're good

rainers:
> The meaning of @nogc is tied to the run-time promise of no garbage collections

GC.disable could make code nogc, then?

> nogc = "does not allocate memory that will need the garbage collector to free it"

Thanks for clarifying, that's another interpretation ;-) A matching pair of GC.malloc and GC.free will satisfy it, too, though.

I guess a closed PR is not a good place to discuss this. I might start a forum thread instead.

andralex:
>A matching pair of GC.malloc and GC.free will satisfy it, too, though.

Affirmative, that makes a function `@nogc`. The compiler doesn't check that, so you'd need to cast the function type.

PR #1956, opened 2017-10-25
Disable stack protection page by default

nemanja-boric-sociomantic:
On platforms we're using mmap to allocate the fiber stacks we
could hit the limit of the maximum mappings of the ranges more easily
since the mprotect will create a new range (with the different
protection bits). Hence this sets the default protection page to
non-existent, so we don't regress from the previous behaviour from
v2.074.x and earlier.

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -4015,6 +4015,11 @@ private
  */
 class Fiber
 {
+    /// Default stack guard size. Historically it was PAGESIZE for Windows,
+    /// but for the mmap-enabled systems was 0.
+    version (Windows) alias DefaultGuardPageSize = PAGESIZE;
+    else enum DefaultGuardPageSize = 0;
+
     ///////////////////////////////////////////////////////////////////////////
     // Initialization
     ///////////////////////////////////////////////////////////////////////////
@@ -4028,13 +4033,18 @@ class Fiber
      *  fn = The fiber function.
      *  sz = The stack size for this fiber.
      *  guardPageSize = size of the guard page to trap fiber's stack
-     *                    overflows
+     *                    overflows. On mmap-enabled platforms it's
+     *                    off by default, one page should be
+     *                    preferred, if needed. Beware that using this
+     *                    will increase the number of mmaped regions on
+     *                    platforms using mmap, so an OS-imposed limit
+     *                    may be hit.
      *
      * In:
      *  fn must not be null.
      */
     this( void function() fn, size_t sz = PAGESIZE*4,
-          size_t guardPageSize = PAGESIZE ) nothrow
+          size_t guardPageSize = DefaultGuardPageSize ) nothrow
     in
     {
         assert( fn );
@@ -4060,7 +4070,7 @@ class Fiber
      *  dg must not be null.
      */
     this( void delegate() dg, size_t sz = PAGESIZE*4,
-          size_t guardPageSize = PAGESIZE ) nothrow
+          size_t guardPageSize = DefaultGuardPageSize ) nothrow
     in
     {
         assert( dg );
--- a/test/thread/src/fiber_guard_page.d
+++ b/test/thread/src/fiber_guard_page.d
@@ -6,6 +6,7 @@ import core.sys.posix.sys.mman;
 version = StackGrowsDown;
 
 enum stackSize = 4096;
+enum guardSize = 4096;
 
 // Simple method that causes a stack overflow
 void stackMethod()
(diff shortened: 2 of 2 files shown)


nemanja-boric-sociomantic:
I'm targeting master, since this would be the breaking change for the Windows users - which in the past always did VirtualAlloc with the guard page.

nemanja-boric-sociomantic:
Maybe we can actually pull a way so the windows gets the default guard size as before, and target the stable.

schveiguy:
Hm... although it would help the Linux problem, I think we can provide a different default for Windows, such that we are not cancelling an obvious win there.

```D
version(Windows) enum DefaultGuardPageSize = PAGESIZE;
version(Linux) enum DefaultGuardPageSize = 0;
...
this( void function() fn, size_t sz = PAGESIZE*4,
           size_t guardPageSize = DefaultGuardPageSize ) nothrow
```

nemanja-boric-sociomantic:
Yeah, I think that will work.

nemanja-boric-sociomantic:
Updated with the suggestion from @schveiguy

src/core/thread.d
```diff
@@ -4028,13 +4033,17 @@ class Fiber
      *  fn = The fiber function.
      *  sz = The stack size for this fiber.
      *  guardPageSize = size of the guard page to trap fiber's stack
-     *                    overflows
+     *                    overflows. On mmap-enabled platforms it's
+     *                    off by default, one page should be
+     *                    preferred, if needed. Beware that using this
+     *                    will increase the number of mmaped regions on
+     *                    platforms using mmap, so the limit may be hit.
```
schveiguy: "the limit" -> "An OS-imposed limit"

nemanja-boric-sociomantic:
Updated with the comment in wording and reference to bugzilla issue.

schveiguy:
thanks, LGTM. I'll let this sit for a while, as it's a fundamental change in expectation for Fibers. ping @MartinNowak

schveiguy:
RE the windows error, WUT?

How the hell did it work as a default argument then???!!!

schveiguy:
Well how about that, you learn something new every day. Apparently even non-immutable statics can be default parameters.

Try this instead:
```D
version(Windows) alias DefaultGuardPageSize = PAGESIZE;
else enum DefaultGuardPageSize = 0;
```

schveiguy:
(I also didn't know PAGESIZE was a variable, thought it was a compile-time value)

nemanja-boric-sociomantic:
Yeah, me neither - I assumed that the PAGESIZE must be known at compile time as _it is the default argument_ :-). I'll update.

nemanja-boric-sociomantic:
With the courtesy of @mathias-lang-sociomantic:

```
import core.stdc.time;

void foo (time_t t = time(null)) 
{
    import std.stdio;
    writeln(t);
}

void main()
{
    import core.thread;

    foreach (_; 0..100)
    {
        foo();
        Thread.sleep(1.seconds);
    }
}
```

I never knew.

MartinNowak:
I'm not too comfortable with just undoing this. Stack overflows are a practical issue with Fibers and really hard to debug (as they'll most likely just overwrite another Fiber's stack). There were no guard pages on Posix, simply because it wasn't implemented, not necessarily a state to conserve.

Furthermore the analysis in Issue 17914 wasn't precise, the problem is that the kernel cannot merge the mappings any longer, otherwise it would just have been a halving of allocatable Fibers.

Maybe we can properly document this and ask the users to tune this to their workloads?

Sacrificing memory safety just so that toy examples work seems like a problematic trade-off. But obviously a limitation people will run into when playing around with Fibers.

Asking a production setup to tune `sysctl -w vm.max_map_count=262144` and their Fiber stack&guard sizes seems reasonable as well.

Any other ideas for stack overflow protection?

schveiguy:
Added some suggestions on the bug report. Bottom line, I don't think we can break existing valid code this way, even though it could lead to memory corruption.

Is there a possible deprecation process we can go through?

MartinNowak:
I cannot agree with this assessment, memory corruption and the implied security issues are much more severe than hitting a resource limit.
There are 3 workarounds for that resource limit, eagerly collecting/destroying Fibers, increasing the limit, or disabling the guard page. This should be properly documented, but I disagree with disabling guard pages by default.
In particular when considering that Fibers are our foundation for asynchronous web software, which is most exposed to exploitation of such bugs.

nemanja-boric-sociomantic:
Where do you suggest we document this, Martin? In the Fiber's class constructor? I would be happy to do so, if that's the appropriate place.

MartinNowak:
Yes, at the Fiber constructor.
The guard should also be made available as Generator overloads, with a references to the Fiber documentation. https://dlang.org/library/std/concurrency/generator.this.html

schveiguy:
So close this then?

nemanja-boric-sociomantic:
Yes, let's close this and I'll reopen tomorrow as per Martin's suggestion.

PR #1957, opened 2017-10-26
Remove gcstub.

skl131313:
Has been broken for close to 2 years, nobody is missing it obviously. No point duplicating the changes that were done to the gc.

Needs https://github.com/dlang/phobos/pull/5809


MartinNowak:
It has been superseded by the manual GC, https://github.com/dlang/druntime/blob/57254581fe821bdf93a6f6e312dff3cb635cff00/src/gc/impl/manual/gc.d.

PR #1958, opened 2017-10-27
add missing lowercase sockaddr_storage name

MartinNowak:
- as with the other sockaddr structs, the lowercase name
  is also aliased as upcase name

### DIFF
--- a/src/core/sys/windows/winsock2.d
+++ b/src/core/sys/windows/winsock2.d
@@ -681,13 +681,14 @@ struct sockaddr
 alias sockaddr SOCKADDR;
 alias SOCKADDR* PSOCKADDR, LPSOCKADDR;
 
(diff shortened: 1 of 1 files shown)


src/core/sys/windows/winsock2.d
```diff
@@ -681,13 +681,14 @@ struct sockaddr
...
 
-struct SOCKADDR_STORAGE
+struct sockaddr_storage
 {
     short     ss_family;
     char[6]   __ss_pad1;
     long      __ss_align;
     char[112] __ss_pad2;
 }
+alias sockaddr_storage SOCKADDR_STORAGE;
```
nemanja-boric-sociomantic: Should be the same for `sockaddr_in6`, no?

PR #1959, opened 2017-10-27
fix winnt versions in mswsock

MartinNowak:
- most stuff supported since WinXP

### DIFF
--- a/src/core/sys/windows/mswsock.d
+++ b/src/core/sys/windows/mswsock.d
@@ -106,7 +106,7 @@ extern(Windows) {
     const GUID WSAID_CONNECTEX = {0x25a207b9,0xddf3,0x4660,[0x8e,0xe9,0x76,0xe5,0x8c,0x74,0x06,0x3e]};
 }
 
-static if (_WIN32_WINNT => 0x501) {
+static if (_WIN32_WINNT >= 0x501) {
     /*  These appear to be constants for the TRANSMIT_PACKETS_ELEMENT
      *  structure below, so I've given them the same minimum version
      */
@@ -164,7 +164,7 @@ static if (_WIN32_WINNT => 0x501) {
     }
 }
 
-static if (_WIN32_WINNT => 0x600) {
+static if (_WIN32_WINNT >= 0x600) {
     /* TODO: Standard Posix.1g macros as per RFC 2292, with WSA_uglification. */
     /* DK: MinGW doesn't define these, and neither does the MSDN docs.  Might have
        to actually look up RFC 2292... */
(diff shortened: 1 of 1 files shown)


src/core/sys/windows/mswsock.d
```diff
@@ -106,7 +106,7 @@ extern(Windows) {
     const GUID WSAID_CONNECTEX = {0x25a207b9,0xddf3,0x4660,[0x8e,0xe9,0x76,0xe5,0x8c,0x74,0x06,0x3e]};
 }
 
-static if (_WIN32_WINNT > 0x501) {
+static if (_WIN32_WINNT => 0x501) {
```
rainers: => -> >= ;-)

src/core/sys/windows/mswsock.d
```diff
@@ -175,16 +177,15 @@ static if (_WIN32_WINNT > 0x501) {
...
-
-        /*  WinXP+, Srv2k3+
-         *  ms-help://MS.MSDNQTR.2003FEB.1033/winsock/winsock/wsarecvmsg_2.htm
-         */
-        int WSARecvMsg(SOCKET, LPWSAMSG, LPDWORD, LPWSAOVERLAPPED, LPWSAOVERLAPPED_COMPLETION_ROUTINE);
-    }
+    /*  Win Vista+, Srv2k3+
+     *  ms-help://MS.MSDNQTR.2003FEB.1033/winsock/winsock/disconnectex_2.htm
+     */
+    extern(Windows) BOOL DisconnectEx(SOCKET, LPOVERLAPPED, DWORD, DWORD);
```
rainers: While I did not find DisconnectEx in any SDK header, the german MSDN page says it's available since WinXP:
https://msdn.microsoft.com/de-de/library/windows/desktop/ms737757(v=vs.85).aspx
while the US page restricts it to Vista:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms737757(v=vs.85).aspx
Maybe stay less restrictive and let it fail at link time.

ZombineDev: > Maybe stay less restrictive and let it fail at link time

As far as I can see, in [`src/core/sys/windows/w32api.d`](https://github.com/dlang/druntime/blob/v2.077.0/src/core/sys/windows/w32api.d), we're always defining `_WIN32_WINNT` to be `0x501`, since we never compile with version `Windows7` or `Windows8`, or `WindowsXXX`...
I wonder if it wouldn't be more practical to do away with the use of `_WIN32_WINNT` in druntime and unconditionally define all symbols and let them fail at link-time as you suggest for this particular case.

ZombineDev: This perhaps goes in line with @CyberShadow's suggestion to remove `version (Platform):` from the top of the header files, so you could for example use the definition of `BITMAPINFOHEADER` to read .bmp images on other platforms, instead of having to copy-paste the definition in your own project. (Though, AFAIR, a pull request to remove `version (Windows):` was blocked because it increased the size of druntime on Linux since now module info had to be emitted for more modules.)

CyberShadow (approved):
> As far as I can see, in src/core/sys/windows/w32api.d, we're always defining _WIN32_WINNT to be 0x501, since we never compile with version Windows7 or Windows8, or WindowsXXX...

Yeah, it's a hold-over from the Win32 bindings project, and is mostly ancillary now.

I guess this PR is mostly for pedantic correctness rather than fixing anything in practice, so why not.

rainers:
Ok, I pushed a commit fixing the syntax errors.

PetarKirov (approved):

PR #1960, opened 2017-10-30
core.sys.windows.winsock2: Don't implement htons() etc.

kinke:
To prevent duplicate symbols when pulling in that object file and the implementations in the ws2_32 library during linking.

Fixes https://github.com/ldc-developers/ldc/issues/468.


kinke:
Done.

[And if the reason for having them implemented in druntime was performance - these helpers are just forward-declared on Linux too.]

PR #1961, opened 2017-10-31
gc.impl.conservative.gc: Fix compiling with debug=LOGGING.

ibuclaw:


rainers:
The debug conditions SENTINEL and INVARIANT do not compile, too.

ibuclaw:
Yeah, I dare not try them all. ☺️

PR #1962, opened 2017-11-01
Mark GC free as @nogc

edi33416:
Free doesn't/mustn't produce any garbage so it should be marked `@nogc`.


edi33416:
@andralex @ZombineDev

PR #1965, opened 2017-11-02
Remove old adCmp code ( replaced by object.__cmp )

somzzz:
Cleanup.

This code is not used anymore. All array comaprisons are now done via ```object.__cmp```.
Last dmd call to this is removed here: https://github.com/dlang/dmd/pull/7279


andralex:
This seems to not be called from within dmd, do gdc or ldc need it? cc @WalterBright @ibuclaw @klickverbot @redstar

ibuclaw:
`_adCmp2` is still actively being used.  But I guess it won't be once recent switch patches have been backported.

andralex:
@ibuclaw so what's the sequence of actions? Keep this on the backburner until you are done backporting?

ZombineDev:
@kinke @JohanEngelen what's the situation like on the LDC side? Can we move forward with removing the old array comparison implementation?

andralex:
@klickverbot thx

PR #1966, opened 2017-11-02
Fix issue 17914: Document connection between fiber guard and mmap limit

nemanja-boric-sociomantic:
Since adding the fiber's stack guard pages prevents operating system
from merging the mmaped regions together, this increases the number of
mapped regions, so the OS imposed limit can be hit earlier than before.

This documents this in the Fiber's constructors, so the users are
aware of this

cc @schveiguy @MartinNowak

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -4028,7 +4028,9 @@ class Fiber
      *  fn = The fiber function.
      *  sz = The stack size for this fiber.
      *  guardPageSize = size of the guard page to trap fiber's stack
-     *                    overflows
+     *                  overflows. Beware that using this will increase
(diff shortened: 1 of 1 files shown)


schveiguy:
Hm.. this doesn't fix @Hackerpilot's problem, doesn't `Generator` need to be able to forward the guard page size to `Fiber` via it's ctors?

nemanja-boric-sociomantic:
Sorry, accompanying Phobos PR (I made the fix issues other way around): https://github.com/dlang/phobos/pull/5835#issuecomment-341419794

nemanja-boric-sociomantic:
Removed "Fix ###" from the commit message.

PR #1967, opened 2017-11-03
refactor GC option parsing into core.internal.parseoptions.d

rainers:
mostly just moving code. The biggest complication is passing a display name for errors. I opted to require a method `string errorName()`.

Supposed to make the functionality available to other parts of druntime, e.g. https://github.com/dlang/druntime/pull/1916

### DIFF
--- a/src/core/internal/parseoptions.d
+++ b/src/core/internal/parseoptions.d
@@ -3,6 +3,8 @@
 *
 * Copyright: Copyright Digital Mars 2017
 * License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+*
+* Source: $(DRUNTIMESRC src/core/internal/parseoptions.d)
 */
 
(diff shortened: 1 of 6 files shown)


src/core/internal/parseoptions.d
```diff
@@ -0,0 +1,275 @@
+/**
+* parse configuration options
+*
+* Copyright: Copyright Digital Mars 2017
+* License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+*/
```
leandro-lucarella-sociomantic: Maybe a short description on how the module is supposed to be used, with some example (that might be taken from the GC code) could be useful. At a first sight is not trivial to know how the API should be used.

rainers:
added some documentation and made methods errorName() and help() optional.

leandro-lucarella-sociomantic:
:tada:

PR #1968, opened 2017-11-05
core.sys.linux.sys.mman: Remove commit hash reference from urls

ibuclaw:
Links are still clickable, just loads the version currently in `HEAD`.


ibuclaw:
I was under the impression that the urls were just copied as each binding was added.  I certainly only look at the system headers on the current running system when I add them.

These are things that should be checked every so often anyway.

RazvanN7:
@ibuclaw I rebased this. Any objections to merging this?

RazvanN7 (approved):

PR #1970, opened 2017-11-05
Remove deprecated and undocumented feature macros

ibuclaw:
See:
https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=c941736c92fa3a319221f65f6755659b2a5e0a20

https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=498afc54dfee41d33ba519f496e96480badace8e

Current documentation lists `_DEFAULT_SOURCE` as enabling BSD and SVID features.

https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html

For us, this is a no-op, as the same features are still enabled.


ibuclaw:
Hmm, I'll deprecate the enums for now in the highly unlikely event they're being used.

PR #1971, opened 2017-11-06
Final switch on default error template

somzzz:
Compiler lowers final switch default case to this (which is a runtime error).
Old implementation is in core/exception.d

Compiler changes: https://github.com/somzzz/dmd/commit/d1bbf1e99737508b362c8f135ba2d6cefb18ab37

### DIFF
--- a/src/core/exception.d
+++ b/src/core/exception.d
@@ -312,7 +312,7 @@ unittest
  */
 class SwitchError : Error
 {
-    @safe pure nothrow this( string file = __FILE__, size_t line = __LINE__, Throwable next = null )
+    @safe pure nothrow @nogc this( string file = __FILE__, size_t line = __LINE__, Throwable next = null )
     {
         super( "No appropriate switch clause found", file, line, next );
     }
@@ -579,6 +579,12 @@ extern (C) void onSwitchError( string file = __FILE__, size_t line = __LINE__ ) 
     throw new SwitchError( file, line, null );
 }
 
+// Compiler lowers final switch default case to this (which is a runtime error)
+void __switch_errorT()(string file = __FILE__, size_t line = __LINE__) @trusted
+{
+    // Consider making this a compile time check.
+    throw staticError!SwitchError(file, line, null);
+}
 
 /**
  * A callback for unicode errors in D.  A $(LREF UnicodeException) will be thrown.
--- a/src/object.d
+++ b/src/object.d
@@ -3889,6 +3889,14 @@ unittest
     testSwitch!dchar;
 }
 
+// Compiler lowers final switch default case to this (which is a runtime error)
+// Old implementation is in core/exception.d
+void __switch_error()(string file = __FILE__, size_t line = __LINE__)
+{
+    import core.exception : __switch_errorT;
+    __switch_errorT(file, line);
+}
+
 // Helper functions
 
 private inout(TypeInfo) getElement(inout TypeInfo value) @trusted pure nothrow


ZombineDev:
This would be a breaking change. A better solution is to change [`onSwitchError`](https://github.com/dlang/druntime/blob/v2.077.0/src/core/exception.d#L577) like so:
```D
@safe pure nothrow @nogc
void onSwitchError(string file = __FILE__, size_t line = __LINE__)
{
    throw staticError!SwitchError(file, line, null);
}
``` 

That way we keep compatibility with code that catches `Error`(s) (e.g. test runners, event loop / thread schedulers, etc.) and we make the implementation `@nogc` as well.
Also, exceptions provide better error messages.

somzzz:
@ZombineDev I didn't notice ```staticError```, thanks for mentioning it!
It's not ```@nogc``` though.
```
src/core/exception.d(586): Error: @nogc function 'core.exception.__switch_errorT!().__switch_errorT' cannot call non-@nogc function 'core.exception.staticError!(SwitchError, string, ulong, typeof(null)).staticError'
```
Or after adding the ```@nogc``` tag to ```staticError```:
```
src/core/exception.d(703): Error: @nogc function 'core.exception.staticError!(SwitchError, string, ulong, typeof(null)).staticError' cannot call non-@nogc constructor 'core.exception.SwitchError.this'
```

With the current solution, an ```AssertError``` is still thrown and dmd unittests pass. Still, it looks like ```onAssertError``` isn't ```@nogc``` either. https://github.com/dlang/druntime/blob/master/src/core/exception.d#L437

It shouldn't have been possible to mark code using ```final switch``` as ```@nogc``` in the first place.

somzzz:
Made ```SwitchError``` constructor ```@nogc``` and the rest was inferred. Similarly, ```__switch_errorT``` needs to be ```@trusted```  to call ```staticError```.

src/core/exception.d
```diff
@@ -579,6 +579,12 @@ extern (C) void onSwitchError( string file = __FILE__, size_t line = __LINE__ )
     throw new SwitchError( file, line, null );
 }
 
+// Compiler lowers final switch default case to this (which is a runtime error)
+void __switch_errorT()(string file = __FILE__, size_t line = __LINE__) @trusted
+{
+    // Consider making this a compile time check.
+    throw staticError!SwitchError(file, line, null);
```
ZombineDev: Keep in mind, that we may need to add a `if (__ctfe)` branch, as I'm not sure if `staticError` would work in CTFE.
I prefer to leave the PR as it is for now, and cross the bridges when we come to them.

src/core/exception.d
```diff
@@ -579,6 +579,12 @@ extern (C) void onSwitchError( string file = __FILE__, size_t line = __LINE__ )
     throw new SwitchError( file, line, null );
 }
 
+// Compiler lowers final switch default case to this (which is a runtime error)
+void __switch_errorT()(string file = __FILE__, size_t line = __LINE__) @trusted
+{
+    // Consider making this a compile time check.
```
jpf91: We already have that compile time check for `final switch` and normal `switch` always requires a `default` branch.

I guess this runtime check is only used to verify that a value is not out-of-range, probably caused by an invalid `cast`. Ping @andralex

andralex: @jpf91 take a look at https://dpaste.dzfl.pl/a33b4b74d774

jpf91: Thanks. I didn't know `final switch` also works on non-enumeral values.

PR #1972, opened 2017-11-18
Add two Bionic declarations

joakim-noah:
These were needed to get ldc 1.6 and 1.7 to compile.  All stdlib tests pass with this pull, same for the dmd-testsuite with 1.7 on Android/ARM.

### DIFF
--- a/src/core/stdc/assert_.d
+++ b/src/core/stdc/assert_.d
@@ -64,6 +64,10 @@ else version (CRuntime_Glibc)
     ///
     void __assert_perror_fail(int errnum, const(char)* file, uint line, const(char)* func);
 }
+else version (CRuntime_Bionic)
+{
+    void __assert(const(char)* __file, int __line, const(char)* __msg);
+}
 else
 {
     static assert(0);
--- a/src/core/stdc/errno.d
+++ b/src/core/stdc/errno.d
@@ -59,23 +59,15 @@ else version (FreeBSD)
         alias errno = __error;
     }
 }
-else version (linux)
+else version (CRuntime_Bionic)
 {
     extern (C)
     {
-        ref int __errno_location();
-        alias errno = __errno_location;
+        ref int __errno();
+        alias errno = __errno;
     }
 }
 else version (Darwin)
-{
-    extern (C)
-    {
-        ref int __error();
-        alias errno = __error;
-    }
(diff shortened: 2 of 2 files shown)


src/core/stdc/assert_.d
```diff
@@ -64,6 +64,10 @@ else version (CRuntime_Glibc)
     ///
     void __assert_perror_fail(int errnum, const(char)* file, uint line, const(char)* func);
 }
+else version (CRuntime_Bionic)
+{
+    void __assert(const(char)* __file, int __line, const(char)* __msg);
```
joakim-noah: @kinke, need this for the final 1.6 release to build for Android, don't know if ldc itself needs to be patched in any way to call this `__assert` with the right arguments for `betterC`.

src/core/stdc/errno.d
```diff
@@ -59,12 +59,12 @@ else version (FreeBSD)
         alias errno = __error;
     }
 }
-else version (linux)
+else version (CRuntime_Bionic)
```
joakim-noah: `linux` refers to kernel APIs, but I don't see errno declarations in anywhere other than libc headers.

src/core/stdc/errno.d
```diff
@@ -75,14 +75,6 @@ else version (Darwin)
         alias errno = __error;
     }
 }
-else version (OSX)
```
joakim-noah: The `Darwin` block right before means this block will never be called.

src/core/stdc/assert_.d
kinke: It does need to be patched as it currently assumes Glibc on Linux.

src/core/stdc/errno.d
```diff
@@ -75,14 +75,6 @@ else version (Darwin)
...
-{
-    extern (C)
-    {
-        ref int __error();
-        alias errno = __error;
-    }
-}
 else
 {
```
joakim-noah: One thing I noticed is if I tried to go back to this old default block by making sure none of the above versions hit, I got an error in [`core.memory` here about "errno is not an lvalue"](https://github.com/dlang/druntime/blob/master/src/core/memory.d#L908), so it appears #1917 can't be used with this default block anymore.

PR #1973, opened 2017-11-19
fixup for #1952

MartinNowak:
- factor common code out of template
- add guaranteed attributes

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3771,7 +3771,7 @@ template _arrayOp(Args...)
  * Returns:
  *      index of match in caseLabels, -1 if not found
 */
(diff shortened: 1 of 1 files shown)


andralex (approved):
very nice, thx. cc @somzzz

andralex:
pls take a look at the tests

schveiguy (approved):

schveiguy:
ping @MartinNowak the code to generate the case label array triggers a failure for betterC. Is there a reason the code can't just be:

```D
static immutable T[][caseLabels.length] cases = [caseLabels];
```

See https://issues.dlang.org/show_bug.cgi?id=20613

PR #1975, opened 2017-11-20
Replace `body` keyword with its replacement: `do`

wilzbach:
Automatic replacement with

    sed -i "s/^\([ ]*\)body/\1do/" -i **/*.d

See also: https://github.com/dlang/phobos/pull/5869


ZombineDev:
LGTM, but I want to wait for dlang/phobos#5869 before merging, to make sure we don't break dscanner indirectly.

PR #1978, opened 2017-11-28
Fix DMD path to ../generated

wilzbach:
So I'm trying to move ahead with the ddmd -> dmd transition (https://github.com/dlang/dmd/pull/7135) and due to https://github.com/braddr/at-client/pull/4 we are stuck at the moment.

Hence, this PR tries to set the `DMD` binary path by default to one in the `generated` directory. As I have no Windows machine, I'm doing this more or less blindly, but it looks trivial enough.

See also: https://github.com/dlang/phobos/pull/5892

### DIFF
--- a/win32.mak
+++ b/win32.mak
@@ -2,7 +2,10 @@
 
 MODEL=32
 
-DMD=dmd
+DMD_DIR=..\dmd
+BUILD=release
+OS=windows
+DMD=$(DMD_DIR)\generated\$(OS)\$(BUILD)\$(MODEL)\dmd
 
 CC=dmc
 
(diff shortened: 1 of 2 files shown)


ZombineDev:
The Win_32_64 build failed with:
```
Error on line 6: expecting target : dependencies
```

which doesn't make sense to me. On the other hand, the phobos build went without a hitch, so the problem may go away.

rainers:
> Error on line 6: expecting target : dependencies

That's usually the error message that appears when patching the makefiles by the autotester fails. @braddr, please remove that.

PR #1981, opened 2017-12-02
Upstream some LDC patches

kinke:
Pinging @klickverbot and @ibuclaw.

### DIFF
--- a/src/core/sys/posix/pwd.d
+++ b/src/core/sys/posix/pwd.d
@@ -220,8 +220,12 @@ else version( DragonFlyBSD )
 }
 else version (Solaris)
 {
-    int getpwnam_r(in char*, passwd*, char*, size_t, passwd**);
-    int getpwuid_r(uid_t, passwd*, char*, size_t, passwd**);
+    alias getpwnam_r = __posix_getpwnam_r;
+    alias getpwuid_r = __posix_getpwuid_r;
+
+    // POSIX.1c standard version of the functions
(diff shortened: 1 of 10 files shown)


src/core/sys/posix/pwd.d
```diff
@@ -196,10 +196,14 @@ else version( OpenBSD )
     int getpwnam_r(in char*, passwd*, char*, size_t, passwd**);
     int getpwuid_r(uid_t, passwd*, char*, size_t, passwd**);
 }
-else version (Solaris)
+else version( Solaris )
```
ibuclaw: Using this coding style is discouraged. I should probably get round to mechanically changing it to the right style.

kinke:
Ping - should be a straightforward merge except for the AArch64 assembly, but that's all experimental anyway (and surely better than nothing, i.e., the status quo).

src/core/threadasm.S
```diff
@@ -449,4 +449,59 @@ fiber_switchContext:
...
+ *    9: x29 (fp)  <-- oldp / *newp save stack top
+ *    8: x30 (lr)
+ *    7: d8
+ *   ...
+ *    0: d15       <-- sp
+ */
+        .text
+        .global CSYM(fiber_switchContext)
+        .p2align  2
+CSYM(fiber_switchContext):
```
ibuclaw: @jpf91 please check this.

jpf91: AFAICS this should be correct 👍

PR #1982, opened 2017-12-02
Make deprecation warnings compile errors

wilzbach:
This is already the way we build Phobos + Tools.
See also: https://github.com/dlang/phobos/pull/5546

CC @ZombineDev @CyberShadow

### DIFF
--- a/src/core/internal/arrayop.d
+++ b/src/core/internal/arrayop.d
@@ -141,7 +141,7 @@ template typeCheck(bool fail, T, Args...)
     {
         alias UT = Args[idx - 1];
         enum op = Args[idx][1 .. $];
-        static if (is(typeof((UT a) => mixin(op ~ " a")) RT == return))
+        static if (is(typeof((UT a) => mixin(op ~ "cast(int) a")) RT == return))
             alias typeCheck = typeCheck!(fail, T, Args[0 .. idx - 1], RT, Args[idx + 1 .. $]);
         else static if (fail)
             static assert(0, "Unary `" ~ op ~ "` not supported for type `" ~ UT.stringof ~ "`.");
@@ -271,7 +271,7 @@ string scalarExp(Args...)()
         else static if (isUnaryOp(arg))
         {
             auto op = arg[0] == 'u' ? arg[1 .. $] : arg;
-            stack[$ - 1] = op ~ stack[$ - 1];
+            stack[$ - 1] = op ~ "cast(int)"~ stack[$ - 1];
         }
         else static if (arg == "=")
         {
--- a/posix.mak
+++ b/posix.mak
@@ -77,7 +77,7 @@ ifeq (solaris,$(OS))
 endif
 
 # Set DFLAGS
-UDFLAGS:=-conf= -Isrc -Iimport -w -dip1000 -preview=fieldwise $(MODEL_FLAG) $(PIC) $(OPTIONAL_COVERAGE)
+UDFLAGS:=-conf= -Isrc -Iimport -w -de -dip1000 -preview=fieldwise $(MODEL_FLAG) $(PIC) $(OPTIONAL_COVERAGE)
 ifeq ($(BUILD),debug)
 	UDFLAGS += -g -debug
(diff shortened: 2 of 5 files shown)


CyberShadow:
Sounds good, currently fails CI though.

wilzbach:
> Sounds good, currently fails CI though.

Yeah:

```
src/core/time.d(4581): Deprecation: struct core.time.FracSec is deprecated - FracSec has been deprecated in favor of just using Duration for the sake of simplicity
src/core/time.d(3594): Error: template instance core.time._assertThrown!(TimeException, FracSec) error instantiating
```

Easy to fix.

```
src/core/internal/arrayop.d(147): Error: static assert  "Unary `-` not supported for type `ubyte`."
```

That one turns out to be rather difficult as it appears deep in the `arrayop` mixin :/

wilzbach:
Updated to resolve the merge conflicts and incorporate newer deprecations.
For `delete` https://github.com/dlang/dmd/pull/8016 is required, apart from that I still haven't figured out how to fix the deprecation in `arrayop`.

src/rt/tracegc.d
```diff
@@ -40,14 +40,15 @@ version (none)
...
-        S* ps; delete ps;
+        // delete is deprecated
+        //int[] a; delete a;
+        //S[] as; delete as;
+        //C c; delete c;
+        //I i; delete i;
+        //C* pc = &c; delete *pc;
+        //I* pi = &i; delete *pi;
+        //int* pint; delete pint;
+        //S* ps; delete ps;
```
wilzbach: Note that this is in a `version(none)` block.

Geod24: Deprecation message issued at parse time ? :/

JinShil: `delete` has been deprecated for a year now.  It's time to turn it into an error.

jacob-carlborg: > Deprecation message issued at parse time ? :/

There are depreciation messages issued at lexing as well.

wilzbach: Yup, but it looks that this has been fixed already:

https://issues.dlang.org/show_bug.cgi?id=18505

src/core/internal/arrayop.d
```diff
@@ -141,7 +141,7 @@ template typeCheck(bool fail, T, Args...)
     {
         alias UT = Args[idx - 1];
         enum op = Args[idx][1 .. $];
-        static if (is(typeof((UT a) => mixin(op ~ " a")) RT == return))
+        static if (is(typeof((UT a) => mixin(op ~ "cast(int) a")) RT == return))
```
wilzbach: I could also have used `-preview=intpromote` here, but as this wouldn't be there when any user code tries to use this, I went with the "safe" approach of explicitly using the existing behavior.

wilzbach:
@thewilsonator @JinShil FYI - this is passing and ready ;-)

quickfur:
Heads up: this PR has caused a regression in my code.  Trying to reduce it to minimal case now, will post details once I reduced it.

quickfur:
Regression filed: https://issues.dlang.org/show_bug.cgi?id=19796

This PR broke array ops (`arr[] = -src[];`).

src/core/internal/arrayop.d
```diff
@@ -271,7 +271,7 @@ string scalarExp(Args...)()
         else static if (isUnaryOp(arg))
         {
             auto op = arg[0] == 'u' ? arg[1 .. $] : arg;
-            stack[$ - 1] = op ~ stack[$ - 1];
+            stack[$ - 1] = op ~ "cast(int)"~ stack[$ - 1];
```
quickfur: This is the culprit. Deleting `"cast(int)"~` fixes the problem.

I think there's a missing check that `Args` contains integral types! The cast causes problems when performing array ops on `double[]`.

wilzbach:
> Regression filed: https://issues.dlang.org/show_bug.cgi?id=19796

Thanks -> https://github.com/dlang/druntime/pull/2550

PR #1984, opened 2017-12-06
[WIP] Add system call bindings

wilzbach:
So I was just looking at @Darredevil's upcoming fiber library and realized that I had this branch lying here for quite a while (almost a year now).

One way to do system calls is by using `syscall` from libc, another is the obvious by defining the assembler directly as e.g. https://github.com/kubo39/syscall-d does.
Let me know which approach you prefer, but I think adding the syscall numbers can be done in any case.
We have used this approach for [mir-random](https://github.com/libmir/mir-random/blob/a81389bf52c04de4c692af294b51ff70e728943c/source/mir/random/engine/package.d#L548).

See also: 
- https://github.com/torvalds/linux/blob/master/arch/sh/include/uapi/asm/unistd_64.h
- https://github.com/torvalds/linux/blob/master/arch/sh/include/uapi/asm/unistd_32.h
- https://github.com/hwoarang/uClibc/blob/master-metag/include/unistd.h
- https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/unistd.h


src/core/sys/linux/unistd.d
```diff
@@ -18,3 +19,730 @@ enum {
 
 /// Prompt for a password without echoing it.
 char* getpass(const(char)* prompt);
+
+// linux system call number from Linux's asm/unistd.h
+version(X86)
```
nemanja-boric-sociomantic: Wow, this corresponds to 30x file size :-). Could we add this as a another module and publicly import it here?

src/core/sys/linux/unistd.d
```diff
@@ -18,3 +19,730 @@ enum {
...
+{
+    enum SYS_restart_syscall = 0;
```
nemanja-boric-sociomantic: `enum SYS: clong { restart_syscall = 0; ... }` is a bit too much (for the C's API)?

Darredevil:
LGTM
Looking forward to not having to define each syscall by hand when I need it.

src/core/sys/linux/unistd.d
wilzbach: Which file do you suggest?
I put them in this file, because that's where they are defined on Linux.

src/core/sys/linux/unistd.d
wilzbach: I would prefer this as well! I just wasn't sure about this and there seemed to be a couple of old precedents on `SYS_` when I did this (one year ago).

src/core/sys/linux/unistd.d
Burgos: I'm not sure, perhaps `core.sys.linux.syscalls`. I don't have a good answer, and maybe they could go here, indeed.

src/core/sys/linux/unistd.d
Burgos: I think it's fair to use enums here, but that's just my preference. It's easy to generate global enums from it at ctfe anyway, if it's required.

wilzbach:
> enum SYS: clong { restart_syscall = 0; ... } is a bit too much (for the C's API)?

Done.

> I'm not sure, perhaps core.sys.linux.syscalls. I don't have a good answer, and maybe they could go here, indeed.

Done. As `core/sys/linux/unistd.d` isn't part of the documentation build, I didn't include `syscalls.d` into the documentation build for now.

BTW I also realized that after I added the `getRandom` syscall to mir, someone opened an issue about them missing on Arm. Hence, the mir people created https://github.com/libmir/mir-linux-kernel which generates syscall numbers headers for all Linux architectures.
Is this an issue for us? Should this PR also include syscall numbers for other architectures or should we keep it light for now & move step by step?

carun:
Wow! It has been 1.5 years since this PR was raised.

What is blocking this PR at the moment? I need the syscalls for `gettid` to use in `std.experimental.logger`.

carun:
@wilzbach there are conflicts. Do you mind rebasing the branch `add-syscalls` on master?

carun:
@wilzbach just wondering, don't you need to add an entry in `mak/SRCS` as well?

carun:
ping @MartinNowak @Burgos

wilzbach:
Sorry, but I won't be working on this in the forseeable future. Feel free to adopt this PR though.

wilzbach:
For future readers, I advise them to use the [mir-linux-kernel](https://github.com/libmir/mir-linux-kernel) package.

PR #1985, opened 2017-12-07
[WIP] Aa lower lib - work in progress

somzzz:
@RazvanN7


schveiguy:
ping @quickfur as I think you attempted something like this at one point.

quickfur:
I haven't looked into this in detail yet, but from a cursory glance, I have the following comments:

1) The current usage of `AssocArray!(void,void)*` seems to be strange and questionable.  Isn't the whole point of a templated AA implementation to take advantage of compile-time type information?

2) And to that point, the continued usage of `TypeInfo` seems to indicate that this implementation isn't that different from the one we already have now.  So where does the templated part come in?  By that, I don't mean the mere fact that there are templated parameters, but how they are benefitting the code in ways the current implementation can't do. If all we're doing is to tack on template parameters for the key and value types, and then cast that away and revert to using TypeInfos, then this new implementation is, frankly, pointless. We already have a perfectly fine TypeInfo-based implementation now.  The whole thrust behind a templated AA type is to *get rid of the dependency on TypeInfos*.  The key and value types of the AA are already known at compile-time, we should be taking advantage of that, instead of waiting until runtime to inspect the TypeInfo and make decisions then.

3) If done carelessly, a templated AA type is going to introduce massive amounts of template bloat into existing code, which is Not Nice.  Much of the AA implementation doesn't actually (need to) depend on the specifics of the key/value types, e.g., the rehashing algorithm, the collision resolution algorithm, the management of the main table of buckets, etc.. Only a few places actually need to know the specifics of the type: when computing a hash value, when comparing keys, when storing a key / value in a bucket, and when returning the result of a lookup.   Ideally, these two parts should be separately implemented; the type-agnostic code should be in a core, non-templated submodule that provides the low-level primitives, and the templated part should be primarily in the user-facing side of things (probably in `object.d`).  Doing this will probably entail a radically different code organization than what is currently there, even if the underlying algorithms will pretty much be the same. (I don't recommend trying to also replace the algorithms, the scope of the code change will be too large to handle in that case, and too prone to breakage. Let's get the templated implementation working first, then we can worry about improving the algorithms.)

4) Another thing that might be worth considering in a template AA implementation is the possibility of compile-time evaluation. This may be a long shot, but in the ideal case, it would be *very nice* if AA literals that don't involve runtime values can be completely constructed at compile-time and stored directly in the object file.  Then we can have static immutable AA's that don't incur runtime initialization overhead.

andralex:
>1) The current usage of `AssocArray!(void,void)*` seems to be strange and questionable.  Isn't the whole point of a templated AA implementation to take advantage of compile-time type information?

This is a WIP on a difficult goal and we decided to first go through a proof of concepts stage whereby `AssocArray!(void,void)*` stands in for the current dynamically-typed associative array. Later on we'll get more fancy.

>2) And to that point, the continued usage of `TypeInfo` seems to indicate that this implementation isn't that different from the one we already have now. 

Indeed. That is intentional. We want to make sure the lowering generates correct calls to methods of a template type.

>So where does the templated part come in?  By that, I don't mean the mere fact that there are templated parameters, but how they are benefitting the code in ways the current implementation can't do. If all we're doing is to tack on template parameters for the key and value types, and then cast that away and revert to using TypeInfos, then this new implementation is, frankly, pointless. We already have a perfectly fine TypeInfo-based implementation now.  The whole thrust behind a templated AA type is to *get rid of the dependency on TypeInfos*.  The key and value types of the AA are already known at compile-time, we should be taking advantage of that, instead of waiting until runtime to inspect the TypeInfo and make decisions then.

Agreed. Yes, the plan is to move to a non-Typeinfo implementation.

>3) If done carelessly, a templated AA type is going to introduce massive amounts of template bloat into existing code, which is Not Nice.

We'll indeed look forward to improvements once we get the templated basics in place.

>4) Another thing that might be worth considering in a template AA implementation is the possibility of compile-time evaluation. This may be a long shot, but in the ideal case, it would be *very nice* if AA literals that don't involve runtime values can be completely constructed at compile-time and stored directly in the object file.  Then we can have static immutable AA's that don't incur runtime initialization overhead.

I think we already have that (which may cause difficulty to this project).

andralex:
@somzzz please rebase

quickfur:
@andralex Thanks for clarifying.  Using `AssocArray!(void,void)` as a transitional measure to bridge to the current state of things is not a bad idea.  Hoping to see the final results of this effort!

rainers:
You should use some git rename command for rt.aaa.d -> assoc_array.d to not lose history and make it a lot easier to review actual changes.

src/assoc_array.d
```diff
@@ -0,0 +1,1046 @@
...
+    return a < b ? a : b;
+}
+
+private T max(T)(T a, T b) pure nothrow @nogc
+{
+    return b < a ? a : b;
+}
+
+//==============================================================================
+// API AssocArrayementation
```
rainers: test replacement error

src/assoc_array.d
```diff
@@ -0,0 +1,1046 @@
...
+ *      aa = associative array opaque pointer
+ *      ti = TypeInfo for the associative array
+ *      valsz = ignored
+ *      pkey = pointer to the key value
+ * Returns:
+ *      if key was in the aa, a mutable pointer to the existing value.
+ *      If key was not in the aa, a mutable pointer to newly inserted value which
+ *      is set to all zeros
+ */
+extern (C) void* _aaGetY(AssocArray!(void, void)** aa, const TypeInfo_AssociativeArray ti, in size_t valsz,
```
rainers: maybe use `ref AssocArray!(void, void)* aa` here to avoid repeating `(*aa)` below. Or assign it to some local.

src/object.d
```diff
@@ -1959,6 +1963,37 @@ void* aaLiteral(Key, Value)(Key[] keys, Value[] values) @trusted pure
 }
 
 alias AssociativeArray(Key, Value) = Value[Key];
+/*
+public template AssocArray(K, V)
+{
+    import assoc_array;
```
rainers: Good to see the import being delayed here. I'd recommend to put assoc_array into some sub package of core or core.internal, though.

ibuclaw: I'd insist on that.

PR #1986, opened 2017-12-11
Fix issue 18063 - thread_attachThis returns dangling pointer

acehreli:
Prevent returning dangling pointer as well as call all detach varieties after locking slock. (All other calls to Thread.remove() are done while holding slock.)

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -2114,6 +2114,12 @@ extern (C) bool thread_isMainThread() nothrow @nogc
  * Registers the calling thread for use with the D Runtime.  If this routine
  * is called for a thread which is already registered, no action is performed.
  *
+ * Threads registered by this routine should normally be deregistered by $(D
+ * thread_detachThis).  Although $(D thread_detachByAddr) and $(D
+ * thread_detachInstance) can be used as well, such threads cannot be registered
+ * again by $(D thread_attachThis) unless $(D thread_setThis) is called with the
+ * $(D null) value first.
+ *
  * NOTE: This routine does not run thread-local static constructors when called.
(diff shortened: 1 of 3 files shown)


src/core/thread.d
```diff
@@ -2239,8 +2244,14 @@ version( Windows )
  */
 extern (C) void thread_detachThis() nothrow @nogc
 {
+    Thread.slock.lock_nothrow();
+    scope(exit) Thread.slock.unlock_nothrow();
```
nemanja-boric-sociomantic: Ideally, these changes should be in the separate commit.

acehreli: Agreed but I could not pass my test program without fixing these as well. Perhaps I should open two bugs and submit fixes for two.

nemanja-boric-sociomantic: Why not just add extra commit in this PR? Make the first commit to add locks, and the second commit (which includes `Fixes issue ...` entry) with `t.setThis(null)` and doc change?

test/thread/src/attach_detach.d
```diff
@@ -0,0 +1,150 @@
...
+{
+    foreach (ref t; threads)
+        start_thread(t);
+}
+
+void stop_thread(ref MyThread t)
+{
+    t.stopRequested = true;
+    while (!t.stopped)
+        Thread.sleep(1000.msecs);
```
acehreli: I hadn't realized that this sleep was so long. Reducing it to 1.mecs causes a segfault. Working on it...

acehreli:
Closing for now as I'm not sure whether SIGSEGV during pthread_detach() inside code/thread.d is a druntime or a test code bug.

PR #1988, opened 2017-12-11
Compile benchmark suite on the auto-tester to avoid a broken state

wilzbach:
Running the benchmark suite at the moment already triggers deprecation warnings. This PR intends to avoid that the suite doesn't compile in the future.
The test files of the benchmark suite are only compiled - not run (`--repeat=0`) and the more expesnive release flags are dropped too.
See the individual commits for more details.

If the increased run time of the auto-tester is a concern, we can also move this to the `circleci.sh` script.


andralex:
please check the autotester :)

wilzbach:
> please check the autotester :)

Yeah. I had to work a bit on this, but now it's finally passing :)
Also I added `-de` to `DFLAGS` string of the benchmark suite, which lead to quite a list of deprecation messages. Looks like this suite isn't used that often :/

PR #1989, opened 2017-12-11
Thread detach stability

acehreli:
(Trying again after closing PR 1986.)

Apologies for three separate commits. Two of the commits were necessary to get the test program passing.

- Prevent returning dangling pointer from thread_attachThis

- Call all detach varieties after locking slock. (All other calls to Thread.remove() are done while holding slock.)

- Do not pthread_detach non-D threads i.e. the ones that are attached by the user. After all, we don't know and have any say on the lifetimes of such threads.

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -2121,6 +2121,12 @@ extern (C) bool thread_isMainThread() nothrow @nogc
  * Posix systems, the D Runtime does not take ownership of the thread;
  * specifically, it does not call $(D pthread_detach) during cleanup.
  *
+ * Threads registered by this routine should normally be deregistered by $(D
+ * thread_detachThis).  Although $(D thread_detachByAddr) and $(D
+ * thread_detachInstance) can be used as well, such threads cannot be registered
+ * again by $(D thread_attachThis) unless $(D thread_setThis) is called with the
+ * $(D null) value first.
+ *
  * NOTE: This routine does not run thread-local static constructors when called.
  *       If full functionality as a D thread is desired, the following function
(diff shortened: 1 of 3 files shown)


jacob-carlborg:
About the tests only failing on macOS 32bit, I haven’t looked at the changes yet but note that macOS 64Bit uses native TLS implementation while 32bit uses a custom implementation.

acehreli:
I have to determine whether the new test application would fail without the changes to thread.d. Maybe it's just exposing existing issues.

Meanwhile, I'm curious about "macOS 64Bit uses native TLS implementation while 32bit uses a custom implementation". Where can I learn more about the differences?

jacob-carlborg:
> Meanwhile, I'm curious about "macOS 64Bit uses native TLS implementation while 32bit uses a custom implementation". Where can I learn more about the differences?

Mostly in the code in the compiler. I added some documentation about the 64bit implementation [1] [2]. You can read about the custom implementation here [3].

[1] https://github.com/dlang/dmd/blob/master/src/ddmd/backend/el.c#L1284-L1306
[2] https://github.com/dlang/dmd/blob/master/src/ddmd/toobj.d#L1245-L1272
[3] http://www.drdobbs.com/architecture-and-design/implementing-thread-local-storage-on-os/228701185

wilzbach:
Removed auto-merge for now to avoid clogging the priority queue of the auto-tester.

aliak00:
Maybe related to this too? https://issues.dlang.org/show_bug.cgi?id=20085

WalterBright:
@acehreli can you please fix the merge issues?

PR #1990, opened 2017-12-12
Local imports for rt.lifetime

somzzz:
Just a bit of grooming.

### DIFF
--- a/src/rt/lifetime.d
+++ b/src/rt/lifetime.d
@@ -12,17 +12,12 @@
 
 module rt.lifetime;
 
-import core.stdc.stdlib;
-import core.stdc.string;
-import core.stdc.stdarg;
-import core.bitop;
 import core.memory;
 debug(PRINTF) import core.stdc.stdio;
 static import rt.tlsgc;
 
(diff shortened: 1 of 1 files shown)


src/rt/lifetime.d
```diff
@@ -12,17 +12,12 @@
...
 
-import core.stdc.stdlib;
-import core.stdc.string;
-import core.stdc.stdarg;
-import core.bitop;
-import core.memory;
 debug(PRINTF) import core.stdc.stdio;
 static import rt.tlsgc;
 
+import core.memory;
```
nemanja-boric-sociomantic: Can this one stay at line 19?

somzzz: Sure, I can move it. What is the reason for this?

nemanja-boric-sociomantic: Oh, sorry, I meant line 19 in the previous version (so, leave it behind `debug(PRINTF) import core.stdc.stdio`). No particular reason except minimizing the diff (because there's no reason as far as I can see for this move), and if you move it back it keeps the imports sorted alphabetically.

PR #1991, opened 2017-12-13
Fix Issue 17956 - core.memory unittest failure (possibly glibc 2.26 s…

wilzbach:
…pecific)

Let's see what auto-tester thinks about this.

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -891,9 +891,9 @@ void pureFree(void* ptr) @system pure @nogc nothrow
 
     fakePureFree(y);
 
-    // subtract 2 because snn.lib adds 2 unconditionally before passing
-    //  the size to the Windows API
-    void* z = pureMalloc(size_t.max - 2); // won't affect `errno`
+    // Workaround bug in glibc 2.26
+    // See also: https://issues.dlang.org/show_bug.cgi?id=17956
+    void* z = pureMalloc(size_t.max & ~255); // won't affect `errno`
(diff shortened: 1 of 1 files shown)


ibuclaw:
It is glibc 2.26 specific (there is a CVE for it), and it should be applied to a latter 2.26 point release.

src/core/memory.d
```diff
@@ -893,7 +893,7 @@ void pureFree(void* ptr) @system pure @nogc nothrow
 
     // subtract 2 because snn.lib adds 2 unconditionally before passing
     //  the size to the Windows API
-    void* z = pureMalloc(size_t.max - 2); // won't affect `errno`
+    void* z = pureMalloc(long.max - 2); // won't affect `errno`
```
ibuclaw: This probably would be ptrdiff_t for the benefit of 32bit systems? I don't recall the signature of gc.malloc within druntime.

ibuclaw: In any case, the bug applies only when you call malloc with a size from `size_t.max-23ʼ and up.

Shifting the value by one bit would be sufficient also (`size_t.max >> 1`)

ibuclaw: Or unsetting the right most 8 bits to keep the adjustment small (`size_t.max & ~255`).

Its all just a workaround of a libc bug anyway. ☺️

wilzbach: > Or unsetting the right most 8 bits to keep the adjustment small (size_t.max & ~255).

Seems like that worked fine :)

PR #1992, opened 2017-12-13
Fix 18071 -- safe casting of AA to void * when getting ranges

schveiguy:
The unit test added in this PR should demonstrate what to avoid. If we make sure we have an actual AA, then casting to void * will not be intercepted by a malicious `opCast`.

Note, the declaration of `const(V[K])` is *necessary* since any type modifiers on the actual T are not conveyed anywhere. For example, if you passed in `inout(int[int])` to the range generating function, then `V == inout(int)` and `K == int`. If the `const` isn't there, then you can't assign to the temporary.

I'm not in love with how I'm extracting the AA itself, but I don't know of a better way. I'm going to post something in the forums to see if anyone can figure out a better way.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2035,6 +2035,17 @@ V[K] dup(T : V[K], K, V)(T* aa)
     return (*aa).dup;
 }
 
+// this should never be made public.
+private AARange _aaToRange(T: V[K], K, V)(ref T aa) pure nothrow @nogc @safe
+{
+    // ensure we are dealing with a genuine AA.
+    static if (is(const(V[K]) == const(T)))
+        alias realAA = aa;
+    else
+        const(V[K]) realAA = aa;
+    return _aaRange(() @trusted { return cast(void*)realAA; } ());
+}
+
 auto byKey(T : V[K], K, V)(T aa) pure nothrow @nogc @safe
 {
     import core.internal.traits : substInout;
@@ -2054,7 +2065,7 @@ auto byKey(T : V[K], K, V)(T aa) pure nothrow @nogc @safe
         @property Result save() { return this; }
     }
 
-    return Result(_aaRange(() @trusted { return cast(void*)aa; } ()));
+    return Result(_aaToRange(aa));
 }
 
 auto byKey(T : V[K], K, V)(T* aa) pure nothrow @nogc
@@ -2081,7 +2092,7 @@ auto byValue(T : V[K], K, V)(T aa) pure nothrow @nogc @safe
         @property Result save() { return this; }
     }
 
-    return Result(_aaRange((() @trusted => cast(void*) aa) ()));
+    return Result(_aaToRange(aa));
 }
 
 auto byValue(T : V[K], K, V)(T* aa) pure nothrow @nogc
@@ -2126,7 +2137,7 @@ auto byKeyValue(T : V[K], K, V)(T aa) pure nothrow @nogc @safe
         @property Result save() { return this; }
     }
 
-    return Result(_aaRange((() @trusted => cast(void*) aa) ()));
+    return Result(_aaToRange(aa));
 }
 
 auto byKeyValue(T : V[K], K, V)(T* aa) pure nothrow @nogc
--- a/src/rt/aaA.d
+++ b/src/rt/aaA.d
@@ -696,11 +696,12 @@ extern (C) pure nothrow @nogc @safe
 
     bool _aaRangeEmpty(Range r)
     {
-        return r.impl is null || r.idx == r.dim;
+        return r.impl is null || r.idx >= r.dim;
     }
 
     void* _aaRangeFrontKey(Range r)
     {
+        assert(!_aaRangeEmpty(r));
         if (r.idx >= r.dim)
             return null;
         return r.buckets[r.idx].entry;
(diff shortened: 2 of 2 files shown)


schveiguy:
ping @wilzbach I referenced an issue, why is the bot not seeing it?

nemanja-boric-sociomantic:
You're missing `issue` in `Fixes ####` (which is a github syntax)

Here's the regex (https://github.com/dlang-bots/dlang-bot#nerdy-details)

```
((close|fix|address)e?(s|d)? )?(ticket|bug|tracker item|issue)s?:? *([\d ,\+&#and]+)
```

schveiguy:
@nemanja-boric-sociomantic thx, I fixed it.

src/rt/aaA.d
```diff
@@ -696,18 +696,20 @@ extern (C) pure nothrow @nogc @safe
 
     bool _aaRangeEmpty(Range r)
     {
-        return r.impl is null || r.idx == r.dim;
+        return r.impl is null || r.idx >= r.dim;
     }
 
     void* _aaRangeFrontKey(Range r)
     {
+        assert(!_aaRangeEmpty(r));
```
JackStouffer: IMO

```d
if (_aaRangeEmpty(r))
    assert(0, "Trying to access the front of an empty associative array key range");
```

is safer and better for the user

schveiguy: Like I said in the bug report, assert(0) is going to simply halt the program, and effectively generate a segfault. It's no better.

JackStouffer: It gives the user an error message

schveiguy: https://run.dlang.io/is/wIMy1L

JackStouffer: Well, that sucks. It's always better to give the user an error message with the stack trace; I didn't know release just stripped that as well.

schveiguy: I have a way around that, but I'm not sure it's applicable here. At this point, it's an opaque call in druntime with a `void *`. If we are to have an error, it should be in object.d I would think.

src/object.d
```diff
@@ -2054,7 +2054,12 @@ auto byKey(T : V[K], K, V)(T aa) pure nothrow @nogc @safe
         @property Result save() { return this; }
     }
 
-    return Result(_aaRange(() @trusted { return cast(void*)aa; } ()));
+    // ensure we are dealing with a genuine AA.
+    static if(is(V[K] == T))
+        alias realAA = aa;
+    else
+        const(V[K]) realAA = aa;
+    return Result(_aaRange(() @trusted { return cast(void*)realAA; } ()));
```
ZombineDev: How about adding a template version of `_aaRange`, so you can encapsulate the logic there?

src/rt/aaA.d
```diff
@@ -1032,3 +1034,24 @@ unittest
...
     assert(aa.keys == [ 0 ]);
 }
+
+// test safety for alias-this'd AA that have unsafe opCast (issue 18071)
+unittest
+{
+    static struct Foo
+    {
+        int[int] aa;
+        auto opCast() pure nothrow @nogc {
```
ZombineDev: Brace on new line

ZombineDev:
> Regression or critical bug fixes should always target the stable branch.

@schveiguy you know the drill ;)

src/rt/aaA.d
```diff
@@ -1032,3 +1034,24 @@ unittest
...
+        }
+        alias aa this;
+    }
+    static void safetest(Foo f) @safe
+    {
+        assert(f.byKey.empty);
+    }
+
+    Foo f;
+    safetest(f);
```
wilzbach: Could have been simply;

```d
() @safe {
    Foo f;
    assert(f.byKey.empty);
}

schveiguy:
> target the stable branch

This time, I actually had thought of that before submitting 😄 

Will fix the nits. Good idea with the DRY recommendation.

schveiguy:
Updated.

src/rt/aaA.d
ZombineDev: I'm undecided on whether we should add an assert message. On one hand, I agree with Steven that currently it will not have any effect. On the other hand, assert messages do offer more information to the user about what went wrong in general, and given that Walter wants to make asserts configurable, it's not hard to imagine them implemented in terms of `core.internal.abort.abort`, when druntime is compiled in debug mode.
(If it's not already the case, it would be a good idea to make dmd use the debug version of druntime + phobos when user's code is compiled with `-debug`, I think LDC already does this.)

schveiguy: The abort function is what I was referring to. I could make it abort, but I think returning `null` here is a perfectly acceptable alternative. It keeps the control of what happens in this situation out of the low-level details. I'd rather throw the error up above where we can use the type system to make a nicer error.

schveiguy: +1 to I wish dmd would compile the non-release debug versions of druntime and phobos when you use the `-debug` switch. If that's not the case, it should be.

schveiguy:
Bot gonna merge at some point @wilzbach? Is there some sort of probation here?

wilzbach:
> Bot gonna merge at some point @wilzbach? 

The PR isn't approved, it can't merge.

schveiguy:
> The PR isn't approved, it can't merge

You sure? @ZombineDev approved it, he has commit rights. The status said all requirements passed (everything was green). Not that it matters now, but just wondering if there's a bug somewhere.

Thanks @andralex!

schveiguy:
BTW, @wilzbach thanks for all your work on the bot, it really does help a lot!

PR #1993, opened 2017-12-15
Halt on Ddoc errors & fix the existing ones

wilzbach:
Required for https://github.com/dlang/dmd/pull/7440

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -381,8 +381,8 @@ struct BitRange
      * Construct a BitRange.
      *
      * Params:
-     *   bitarr - The array of bits to iterate over
-     *   numBits - The total number of valid bits in the given bit array
+     *   bitarr = The array of bits to iterate over
+     *   numBits = The total number of valid bits in the given bit array
      */
     this(const(size_t)* bitarr, size_t numBits) @system
     {


posix.mak
```diff
@@ -147,19 +147,19 @@ endif
 doc: $(DOCS)
 
 $(DOCDIR)/object.html : src/object.d
-	$(DMD) $(DDOCFLAGS) -Df$@ project.ddoc $(DOCFMT) $<
+	$(DMD) $(DDOCFLAGS) -Df$@ project.ddoc $(DOCFMT) $< 2>&1 | grep "Warning:"; test $$? -eq 1
```
andralex: I think all you need to set the error level is `grep -v "Warning:"`.

andralex: Also a problem with this pattern is that if dmd fails with no warning the command will succeed.

wilzbach:
Okay I removed the explicit grepping for error messages and rebased the PR to just contain the Ddoc fix so that it's absolutely non-controversial and we can move on with this, because once this is merged we can merge the dlang/dmd#7440 at DMD and then `-w` will work as expected with a proper exit code and no grepping is required.

PR #1995, opened 2017-12-16
change Throwable.next to ref counted property

WalterBright:
because `next` is so generic it's too hard to find the uses of it.

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -223,17 +223,7 @@ version( Windows )
 
             void append( Throwable t )
             {
-                if (t.refcount())
-                    ++t.refcount();
-                if( obj.m_unhandled is null )
-                    obj.m_unhandled = t;
-                else
-                {
-                    Throwable last = obj.m_unhandled;
-                    while( last.next !is null )
-                        last = last.next;
-                    last.next = t;
-                }
+                obj.m_unhandled = Throwable.chainTogether(obj.m_unhandled, t);
             }
 
             version( D_InlineAsm_X86 )
@@ -382,17 +372,7 @@ else version( Posix )
 
             void append( Throwable t )
             {
-                if (t.refcount())
-                    ++t.refcount();
-                if( obj.m_unhandled is null )
-                    obj.m_unhandled = t;
-                else
-                {
-                    Throwable last = obj.m_unhandled;
-                    while( last.next !is null )
-                        last = last.next;
-                    last.next = t;
-                }
+                obj.m_unhandled = Throwable.chainTogether(obj.m_unhandled, t);
             }
 
             try
@@ -3058,14 +3038,16 @@ do
 * A callback for thread errors in D during collections. Since an allocation is not possible
 *  a preallocated ThreadError will be used as the Error instance
 *
+* Returns:
+*  never returns
 * Throws:
 *  ThreadError.
 */
-private void onThreadError(string msg = null, Throwable next = null) nothrow
+private void onThreadError(string msg) nothrow
 {
     __gshared ThreadError error = new ThreadError(null);
     error.msg = msg;
-    error.next = next;
+    error.next = null;
     import core.exception : SuppressTraceInfo;
     error.info = SuppressTraceInfo.instance;
     throw error;
--- a/src/object.d
+++ b/src/object.d
@@ -1704,11 +1704,38 @@ class Throwable : Object
      * caught $(D Exception) will be chained to the new $(D Throwable) via this
      * field.
      */
-    Throwable   next;
+    private Throwable   nextInChain;
 
     private uint _refcount;     // 0 : allocated by GC
                                 // 1 : allocated by _d_newThrowable()
                                 // 2.. : reference count + 1
+
+    /**
+     * Returns:
+     * A reference to the _next error in the list. This is used when a new
+     * $(D Throwable) is thrown from inside a $(D catch) block. The originally
+     * caught $(D Exception) will be chained to the new $(D Throwable) via this
+     * field.
+     */
+    @property inout(Throwable) next() @safe inout return scope pure nothrow @nogc { return nextInChain; }
+
+    /**
+     * Replace next in chain with `tail`.
+     * Use `chainTogether` instead if at all possible.
+     */
+    @property void next(Throwable tail) @safe scope pure nothrow @nogc
+    {
+        if (tail && tail._refcount)
+            ++tail._refcount;           // increment the replacement *first*
+
+        auto n = nextInChain;
+        nextInChain = null;             // sever the tail before deleting it
+
+        if (n && n._refcount)
+            _d_delThrowable(n);         // now delete the old tail
+
+        nextInChain = tail;             // and set the new tail
+    }
 
     /**
      * Returns:
@@ -1717,6 +1744,21 @@ class Throwable : Object
      *  and >=2 which is the reference count + 1
      */
     @system @nogc final pure nothrow ref uint refcount() return scope { return _refcount; }
+
+    /**
+     * Loop over the chain of Throwables.
+     */
+    int opApply(scope int delegate(Throwable) dg)
+    {
+        int result = 0;
+        for (Throwable t = this; t; t = t.nextInChain)
+        {
+            result = dg(t);
+            if (result)
+                break;
+        }
+        return result;
+    }
 
     /**
      * Append `e2` to chain of exceptions that starts with `e1`.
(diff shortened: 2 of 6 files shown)


andralex:
@WalterBright don't you plan to add a property returning it?

WalterBright:
> let's not?

Allowing public manipulation of this field is a bad idea, I intend to replace it with a property. Furthermore, while the idea is simple, the usage/implementation is an incomprehensible mess. The first step in figuring out what to do with it is to find where it is being used, and then figure out why.

WalterBright:
> WalterBright requested review

I did not. I added "Needs Work" for a reason. Please refrain from wasting time on this until I remove that!

schveiguy:
> WalterBright requested review

FYI, github does this because of the code owners file. It's how that works. You can just ignore that message.

wilzbach:
> FYI, github does this because of the code owners file. It's how that works. You can just ignore that message.

Yes, the idea is that if someone "owns" a file (i.e. has in-depth knowledge about it), he/she should be notified.
We can always improve the mapping here: https://github.com/dlang/druntime/blob/master/CODEOWNERS
(I added myself as a wildcard when we created this file, s.t. if no owner is found I would be notified and could help the potential submitter, but it seems that GitHub looks at code ownership per file and not per PR, which means that if there's a file not matching the list as part of the PR, I will get notified :/)

WalterBright:
Blocked by https://github.com/dlang/phobos/pull/5936

stefan-koch-sociomantic:
@WalterBright If the compiler can figure out which symbol is meant, so can anyone who queries the compiler.

leandro-lucarella-sociomantic:
Please, if you add any new public symbols mark it first as `@future`, besides regular deprecation cycles.

src/object.d
```diff
@@ -1704,7 +1704,7 @@ class Throwable : Object
      * caught $(D Exception) will be chained to the new $(D Throwable) via this
      * field.
      */
-    Throwable   next;
+    Throwable   nextInChain;
```
leandro-lucarella-sociomantic: This definitely needs a deprecation + future time. Please no more sudden breaking changes. We have a lot of mechanisms to provide smooth migration paths.

WalterBright:
ping @leandro-lucarella-sociomantic @andralex

WalterBright:
I.e. I'm trying to make the chained exceptions work with ref counted exceptions.

rainers:
> renaming .next to .nextInChain was necessary to find all the uses of it, so they can be fixed to accommodate ref counting.

Wasn't that just temporarily? A lot of names are pretty generic, we cannot start renaming them all. Instead, tooling could be improved.


> Some initial fruit is removing a bunch of repetitive code that appends to the list, and replacing it with a function call

The refactoring itself looks good.

src/core/thread.d
```diff
@@ -3058,14 +3038,15 @@ do
...
+*  never returns
 * Throws:
 *  ThreadError.
 */
-private void onThreadError(string msg = null, Throwable next = null) nothrow
+private void onThreadError(string msg) nothrow
 {
     __gshared ThreadError error = new ThreadError(null);
     error.msg = msg;
-    error.next = next;
```
rainers: While the parameter to onThreadError seems unused, resetting `next` to `null` should remain, as it might have been set in a previously caught exception.

WalterBright: good catch

WalterBright: Wait a minute, `error` is newly new'd, so it should be `null` allready

WalterBright:
> Wasn't that just temporarily?

Yes. Fortunately, it seems there weren't any other uses. So I made it `private` and turned `next` into a property.

src/core/thread.d
rainers: No, it's `__gshared`, so preallocated in the data segment.

WalterBright: The handle is put in __gshared, not the actual object instance.

rainers: Sure? Try it here: https://run.dlang.io/gist/ae0a51ac41c1e08251d7191cdf67bc13?compiler=dmd

WalterBright: yes, you're right, I forgot that it was statically initialized. Thanks for the correction.

WalterBright: This leads to a problem with the ref counting. It's getting late, I'll try to fix that tomorrow.

WalterBright: fixed

src/object.d
```diff
@@ -1718,6 +1745,21 @@ class Throwable : Object
      */
     @system @nogc final pure nothrow ref uint refcount() return scope { return _refcount; }
 
+    /**
+     * Loop over the chain of Throwables.
+     */
+    int opApply(scope int delegate(ref Throwable) dg)
```
rainers: Not sure if this isn't operator abuse. Maybe it would be more natural if an iteratable is returned from a `chain()` property?

Why did you mix this change into this PR? Why is the test for it in https://github.com/dlang/dmd/pull/7832?

WalterBright: > Not sure if this isn't operator abuse. Maybe it would be more natural if an iteratable is returned from a chain() property?

That's what opApply is for, I don't know why it would abuse. Iterables are not normally done in D.

> Why did you mix this change into this PR?

Because of the complaints I received that the PR was incomplete. Nobody was interested in doing this in stages.

> Why is the test for it in dlang/dmd#7832?

Because that's where the tests are for the chained exceptions.

andralex: The preferred iteration mode in D is by means of ranges. Uses of `opApply` should provide rationale on why internal iteration is preferred or necessary.

The delegate signature should take a `Throwable`, not a `ref Throwable`. Also it seems `scope Throwable` would be better because we don't want to escape `Throwable` objects during iteration.

All in all this seems to need careful planning.

src/object.d
```diff
@@ -1718,6 +1745,21 @@ class Throwable : Object
...
+    {
+        int result = 0;
+        for (Throwable t = this; t; t = t.nextInChain)
+        {
+            result = dg(t);
+            if (result)
+                break;
+        }
+        return result;
```
andralex: code is not covered

src/object.d
rainers: > That's what opApply is for, I don't know why it would abuse.

Without knowing all the (implementation) details of Throwable, I would not expect to be able to iterate over an exception.

> Because of the complaints I received that the PR was incomplete.

opApply is unrelated to ref counting. AFAICT reference counting is not used so far, so I actually didn't scrutinize the modified reference counting in this PR because I expect it to change anyway.

> Because that's where the tests are for the chained exceptions.

Iterating with foreach over the exception chain is syntax sugar irrelevant to the compiler implementation. I don't like all the unittests in object.d, so I sympathize with putting them elsewhere, but having them in another repo is also a complication.

src/object.d
rainers: It's tested here: https://github.com/dlang/dmd/pull/7832

WalterBright: Ranges don't work in druntime because they require things that are in Phobos. Ranges are a Phobos feature. opApply was used here to hide access to the list from users.

You're right that the delegate should be scope, but due to a bug in the compiler that doesn't work. I haven't gotten around to fixing that bug yet, but that shouldn't block this PR. The scope can always be added later.

src/object.d
WalterBright: > I would not expect to be able to iterate over an exception.

Yet existing code *does*.

> having them in another repo is also a complication.

Yes, but that's beyond the scope of this PR.

src/object.d
wilzbach: > Ranges don't work in druntime because they require things that are in Phobos. 

I don't follow. Sure, lot's of goodies are in `std.range`, but what stops your from just defining `popFront`, `front`, `save` and `empty`?

WalterBright: You're right the `ref` was unnecessary. Removed it.

andralex: We really should use the range primitives here. Let me sketch the implementation below - it's just a few lines of code:

```d
/// Provides forward iteration for the chain positioned on this `Throwable` object
auto opSlice()
{
    static struct Result
    {
        private Throwable current;
        bool empty() { return current is null; }
        Throwable front() { return current; }
        void popFront() { assert(current); current = current.nextInChain; }
        Result save() { return this; }
    }
    return Result(this);
}
```

With this we get to avoid a variety of issues that `opApply` has and make client code a whole lot more flexible.

WalterBright: The point of encapsulating the loop is I want to *completely hide* the linked list from the user, not merely provide a generic way of iterating it. This is because I want to prevent instances escaping without going through the ref counting mechanism. opApply is a better tool for that.

andralex: @WalterBright I understand. We could do it in the range by incrementing the ref counts in `this(this)` and `save()`, but granted that complicates matters. I'll approve.

andralex:
@leandro-lucarella-sociomantic will pull this, please holler if any remaning issues - thx!

PR #1996, opened 2017-12-17
add Throwable.chainTogether()

WalterBright:
The point is to encapsulate a common operation in `std.parallelism` and `core.thread` into this function.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -1718,6 +1718,31 @@ class Throwable : Object
      */
     @system @nogc final pure nothrow ref uint refcount() return scope { return _refcount; }
 
+    /**
+     * Append `e2` to chain of exceptions that starts with `e1`.
+     * Params:
+     *  e1 = start of chain (can be null)
+     *  e2 = second part of chain (can be null)
+     * Returns:
+     *  Throwable that is at the start of the chain; null if both `e1` and `e2` are null
+     */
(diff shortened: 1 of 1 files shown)


WalterBright:
Blocking https://github.com/dlang/phobos/pull/5936

WalterBright:
It might be a good idea to stay in stealth mode for the time being. There'll likely be more.

wilzbach:
> It might be a good idea to stay in stealth mode for the time being.

Then let's better [not expose this on the public documentation](http://dtest.dlang.io/artifact/website-c08728dc547a7081db9f787baa33d60b41f32b91-abffbe18a37ff2200529fbf869a356f1/web/phobos-prerelease/object.html#.Throwable.chainTogether)?

src/object.d
```diff
@@ -1718,6 +1718,31 @@ class Throwable : Object
...
 
+    /**
+     * Append `e2` to chain of exceptions that starts with `e1`.
+     * Params:
+     *  e1 = start of chain (can be null)
+     *  e2 = second part of chain (can be null)
+     * Returns:
+     *  Throwable that is at the start of the chain; null if both `e1` and `e2` are null
+     */
+    static @system @nogc pure nothrow Throwable chainTogether(return scope Throwable e1, return scope Throwable e2)
```
leandro-lucarella-sociomantic: Needs `@future` too.

WalterBright: ok

PR #1997, opened 2017-12-17
[RFC] Initial support for building druntime with musl libc

yshui:
Make druntime buildable with CRuntime_Musl defined instead of CRuntime_Glibc.

### DIFF
--- a/src/core/stdc/stdio.d
+++ b/src/core/stdc/stdio.d
@@ -121,6 +121,24 @@ else version( CRuntime_Glibc )
         L_tmpnam     = 20
     }
 }
+else version( CRuntime_Musl )
+{
+    enum
+    {
+        ///
+        BUFSIZ       = 8192,
+        ///
+        EOF          = -1,
+        ///
+        FOPEN_MAX    = 16,
+        ///
+        FILENAME_MAX = 4095,
+        ///
+        TMP_MAX      = 238328,
+        ///
+        L_tmpnam     = 20
+    }
+}
 else version( Darwin )
 {
     enum
@@ -292,6 +310,14 @@ else version( CRuntime_Bionic )
         int _size;
     }
 }
+else version( CRuntime_Musl )
+{
+    enum
+    {
+        EOF          = -1,
+    }
+
+}
 else
 {
     static assert( false, "Unsupported platform" );
@@ -381,6 +407,16 @@ else version( CRuntime_Glibc )
     ///
     alias _IO_FILE _iobuf;
     ///
+    alias shared(_IO_FILE) FILE;
+}
+else version( CRuntime_Musl )
+{
+    struct fpos_t {
+        char[16] __opaque;
+        double __align;
+    }
+    struct _IO_FILE;
+    alias _IO_FILE _iobuf;
     alias shared(_IO_FILE) FILE;
 }
 else version( Darwin )
@@ -917,6 +953,20 @@ else version( CRuntime_Bionic )
     ///
     shared stderr = &__sF[2];
 }
+else version( CRuntime_Musl )
+{
+    /// needs tail const
+    extern shared FILE* stdin;
+    ///
+    extern shared FILE* stdout;
+    ///
+    extern shared FILE* stderr;
+    enum {
+        _IOFBF = 0,
+        _IOLBF = 1,
+        _IONBF = 2,
+    }
+}
 else
 {
     static assert( false, "Unsupported platform" );
@@ -1415,6 +1465,30 @@ else version( CRuntime_Bionic )
     int  snprintf(scope char* s, size_t n, scope const char* format, ...);
     ///
     int  vsnprintf(scope char* s, size_t n, scope const char* format, va_list arg);
+}
+else version( CRuntime_Musl )
+{
+    import core.sys.posix.sys.types : off_t;
+    ///
+    int fseeko(FILE *, off_t, int);
+    @trusted
+    {
+        ///
+        void rewind(FILE* stream);
+        ///
+        pure void clearerr(FILE* stream);
+        ///
+        pure int  feof(FILE* stream);
+        ///
+        pure int  ferror(FILE* stream);
+        ///
+        int  fileno(FILE *);
+    }
+
+    ///
+    int snprintf(scope char* s, size_t n, scope const char* format, ...);
+    ///
+    int vsnprintf(scope char* s, size_t n, scope const char* format, va_list arg);
 }
 else
 {
--- a/src/core/sys/posix/signal.d
+++ b/src/core/sys/posix/signal.d
@@ -544,6 +544,28 @@ version( CRuntime_Glibc )
         void function() sa_restorer;
     }
 }
+version( CRuntime_Musl )
+{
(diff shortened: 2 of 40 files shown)


src/core/sys/posix/signal.d
```diff
@@ -544,6 +544,28 @@ version( CRuntime_Glibc )
         void function() sa_restorer;
     }
 }
+version( CRuntime_Musl )
```
rikkimax: You missed an else.

Causing: ``src/core/sys/posix/signal.d(686): Error: static assert  "Architecture not supported."``.

rikkimax: You missed an else.

Causing: ``src/core/sys/posix/signal.d(686): Error: static assert  "Architecture not supported."``.

ZombineDev:
@yshui Thanks for looking into porting adding musl libc support to druntime. It looks like there are many things that need to be changed, which is why I highly recommend splitting your work into many commits. That would allow us merge changes that are ready much faster, so we can focus on the rest.

@wilzbach If I remember correctly, you had a tool to split work into multiple commits / pull-requests automatically?

wilzbach:
This is really cool - any chance we can test this e.g. on CircleCi automatically? :)

> @wilzbach If I remember correctly, you had a tool to split work into multiple commits / pull-requests automatically?

Yes, it's a stupid bash script that will simply make a PR for every file.

https://github.com/wilzbach/splitup-prs

ZombineDev:
@yshui I'm now reviewing your changes file by file and I'll get back to you with more comments after 1-2 hours.

> This is really cool - any chance we can test this e.g. on CircleCi automatically? :)

@wilzbach very good idea, but I don't think it's easily doable in the short term, since in order to to run the dmd/druntime/phobos tests on a musl libc system, we first need to compile them for which we need either cross-compilation (telling dmd that host != target, something which only LDC and GDC support well) or a good bootstrap process (but I'm not sure how usable are the dmd-cxx branches).

wilzbach:
> good bootstrap process (but I'm not sure how usable are the dmd-cxx branches).

The latest C++ release can still compile master without any work. @ibuclaw's dmd-cxx branch contains a couple of backports, but IIRC druntime doesn't build with them yet.
Anyhow there's @CyberShadow's amazing [Digger](https://github.com/CyberShadow/Digger) which does an excellent job at bootstrapping.

ibuclaw:
> @ibuclaw's dmd-cxx branch contains a couple of backports, but IIRC druntime doesn't build with them yet.

https://github.com/dlang/dmd/blob/dmd-cxx/travis.sh

I wouldn't have uncommented it if didn't work for me.

Someone should take ownership in looking at the glue/backend in that branch anyway. I have repeatedly said that only the frontend backed by gdc has been rigorously tested, which is backed by semaphore and buildbot ci which checks 6 distinct CPU targets.

src/core/sys/posix/sys/types.d
```diff
@@ -604,6 +637,37 @@ version (CRuntime_Glibc)
 
     alias c_ulong pthread_t;
 }
+else version( CRuntime_Musl )
+{
+    union pthread_attr_t
```
ZombineDev: Again, please separate the definitions per architecture. For reference:
http://git.musl-libc.org/cgit/musl/tree/arch/i386/bits/alltypes.h.in
http://git.musl-libc.org/cgit/musl/tree/arch/x86_64/bits/alltypes.h.in
http://git.musl-libc.org/cgit/musl/tree/arch/arm/bits/alltypes.h.in
http://git.musl-libc.org/cgit/musl/tree/arch/aarch64/bits/alltypes.h.in
http://git.musl-libc.org/cgit/musl/tree/arch/mipsn32/bits/alltypes.h.in
http://git.musl-libc.org/cgit/musl/tree/arch/mips64/bits/alltypes.h.in
...

joakim-noah:
@yshui, nice work, I'll take a look.

@ZombineDev, never messed with Musl, can it be used on a regular glibc distro too?  If so, setting up CI should be easy.

@ibuclaw, I'd be happy to try testing the dmd-cxx branch, but I was under the impression you were going to update the frontend on that branch further, dlang/dmd#7227.

ZombineDev:
> @ZombineDev, never messed with Musl, can it be used on a regular glibc distro too? If so, setting up CI should be easy.

@joakim-noah I wasn't sure too, but looking at https://github.com/emk/rust-musl-builder it seems it should be doable.

wilzbach:
> @ZombineDev, never messed with Musl, can it be used on a regular glibc distro too? If so, setting up CI should be easy.

Yes, it actually looks pretty straight-forward. They even provide a gcc wrapper and just statically link it in:

https://wiki.musl-libc.org/getting-started.html#Using-the-musl-gcc-wrapper

(I haven't played with musl either)

src/core/time.d
```diff
@@ -292,7 +292,7 @@ else version(Darwin) enum ClockType
     precise = 3,
     second = 6,
 }
-else version(linux) enum ClockType
+else version(CRuntime_Glibc) enum ClockType
```
joakim-noah: This is going to break this module for Bionic and the tests below for Musl.  I see no reason to change this, even if these constants aren't defined for you.

src/core/time.d
```diff
@@ -355,9 +364,11 @@ version(Posix)
             case coarse: return CLOCK_MONOTONIC_COARSE;
             case normal: return CLOCK_MONOTONIC;
             case precise: return CLOCK_MONOTONIC;
-            case processCPUTime: return CLOCK_PROCESS_CPUTIME_ID;
             case raw: return CLOCK_MONOTONIC_RAW;
-            case threadCPUTime: return CLOCK_THREAD_CPUTIME_ID;
+            version(CRuntime_Glibc) {
+                case processCPUTime: return CLOCK_PROCESS_CPUTIME_ID;
+                case threadCPUTime: return CLOCK_THREAD_CPUTIME_ID;
+            }
```
joakim-noah: These won't be defined for Bionic anymore, and [they appear to be defined for Musl](http://git.musl-libc.org/cgit/musl/tree/include/time.h?h=v1.1.18#n87)?

src/core/stdc/stdio.d
```diff
@@ -383,6 +409,16 @@ else version( CRuntime_Glibc )
...
+else version( CRuntime_Musl )
+{
+    struct fpos_t {
+        char[16] __opaque;
+        double __align;
+    }
+    struct _IO_FILE;
+    alias _IO_FILE _iobuf;
+    alias shared(_IO_FILE) FILE;
+}
```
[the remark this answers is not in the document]
ibuclaw: Strictly speaking we don't really need to define the structures of any `FILE` type.  But for completion sake, and to match other runtimes it probably should be added.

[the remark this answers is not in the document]
yshui: @ZombineDev _iobuf is defined because it's used in phobos.

joakim-noah:
Do you plan on resurrecting the parts of this pull that haven't been submitted elsewhere? If so, maybe you can just reopen this pull with that remaining patch.

PR #1492, opened 2016-02-03
NetBSD path

nrTQgc:
It is LDC based port. My pull request merged to LDC master already. 
Druntime:
I decide try to repeat LDC team forkflow and cherry-pick my commit to ldc branch. There are a couple simple conflicts in src/core/stdc/locale.d & src/core/sys/posix/signal.d, and one absent (in master) file: src/rt/sections_ldc.d 
I successfully build and run druntime unittests on NetBSD after resolving conflicts.

PR #1999, opened 2017-12-18
Add DragonFlyBSD support for druntime

dkgroot:
Related: [DMD Pull Request](https://github.com/dlang/dmd/pull/7463)
Bootstrapped via 'dmd-cxx' branch: [See](https://github.com/dkgroot/dragonflybsd_dmd_port)

Related PR:
- https://github.com/dlang/druntime/pull/2002
- https://github.com/dlang/dmd/pull/7463

Notes:
- Fixed getAddrInfo issue in phobos/std/socket.d by giving addrinfo.ai_socktype a default value of SOCK_STREAM.
  (Without it getAddrInfo() will generate an error:'servname not supported for ai_socktype' when hints.ai_socktype is not set.)
- 1 minor dragonfly related FIXME entries in src/core/sync/mutex.d (Caused by the implementation of pthread implementation on DragonFlyBSD): trylock after destroying a mutex does not return an error. (similar to 'Use after free')


nemanja-boric-sociomantic:
It would be very nice if you could split this into several commits (as per your checklist in the PR), and amend it to say "Add DragonFlyBSD support for druntime" instead of "Port druntime to ..."

src/core/sys/posix/netdb.d
```diff
@@ -594,6 +594,107 @@ else version( OpenBSD )
...
+        char**  s_aliases;
+        int     s_port;
+        char*   s_proto;
+    }
+
+    struct addrinfo
+    {
+        int         ai_flags;
+        int         ai_family;
+        int         ai_socktype = SOCK_STREAM;           /* socktype default value required to be able to perform getAddrInfo on DragonFlyBSD
```
dkgroot: Fixed getAddrInfo issue in phobos/std/socket.d by giving addrinfo.ai_socktype a default value of SOCK_STREAM.
(Without it getAddrInfo() will generate an error:'servname not supported for ai_socktype' when hints.ai_socktype is not set.)

src/rt/util/typeinfo.d
```diff
@@ -247,7 +247,8 @@ unittest
         {
             assert(f1 == 0 + 0i);
 
-            assert(f1  == f2);
+            assert(f1 == f2);
+            version (DragonFlyBSD) {} else   // FIXME:  //core.exception.AssertError@src/rt/util/typeinfo.d(249): unittest failure
```
dkgroot: // FIXME:  
core.exception.AssertError@src/rt/util/typeinfo.d(249): unittest failure
Have not figured out what is going on here, nor what this unittest is trying to check.

src/core/memory.d
```diff
@@ -895,6 +895,8 @@ void pureFree(void* ptr) @system pure @nogc nothrow
     // See also: https://issues.dlang.org/show_bug.cgi?id=17956
     void* z = pureMalloc(size_t.max & ~255); // won't affect `errno`
     assert(errno == fakePureErrno()); // errno shouldn't change
+
+    version (DragonFlyBSD) {} else    // FIXME : z seems to be allocated and not null. Why would size_t.max - 2 not be allowed to allocate ? or is the assertion wrong
```
dkgroot: FIXME : 
z seems to be allocated and not null.
Why would size_t.max - 2 not be allowed to allocate ? or is the assertion wrong

src/core/sync/mutex.d
```diff
@@ -377,6 +377,7 @@ unittest
     // For example, Bionic doesn't appear to do so, so this test is
     // not run on Android.
     version (CRuntime_Bionic) {} else
+    version (DragonFlyBSD) {} else               // FIXME : Will have to write a test program to see what is going on here.
```
dkgroot: // FIXME : 
Will have to write a little c-program to check if DragonFlyBSD pthread does allow trylock after the mutex has been destroyed and does return 0 in this case.

posix.mak
```diff
@@ -317,18 +317,19 @@ CWS_MAKEFILES = $(filter mak/% %.mak %/Makefile,$(MANIFEST))
...
 # restrict to linux, other platforms don't have a version of grep that supports -P
-ifeq (linux,$(OS))
-	$(GREP) -n -U -P "([ \t]$$|\r)" $(CWS_MAKEFILES) ; test "$$?" -ne 0
-	$(GREP) -n -U -P "( $$|\r|\t)" $(NOT_MAKEFILES) ; test "$$?" -ne 0
-endif
+#ifeq (linux,$(OS))
+#       $(GREP) -n -U -P "([ \t]$$|\r)" $(CWS_MAKEFILES) ; test "$$?" -ne 0
+#       $(GREP) -n -U -P "( $$|\r|\t)" $(NOT_MAKEFILES) ; test "$$?" -ne 0
+#endif
+checkwhitespace: $(INSTALL_DIR)/${OS}/bin${MODEL}/dmd $(TOOLS_DIR)/checkwhitespace.d
```
dkgroot: Instead of using platform dependent grep
we should use the tools/checkwhitespace.d D-version

dkgroot:
@nemanja-boric-sociomantic 
I had a little mishap during rebasing on top of master. Before this these were seperate commits. It it helps i added some comments (See above) to clarify the locations where the issues arose).

src/core/memory.d
```diff
@@ -895,6 +895,8 @@ void pureFree(void* ptr) @system pure @nogc nothrow
...
     assert(z is null);
```
dkgroot: z seems to be allocated and not null.
Why would size_t.max - 2 not be allowed to allocate ? or is the assertion wrong

nemanja-boric-sociomantic:
> It it helps i added some comments (See above) to clarify the locations where the issues arose).

Comments do really help, thanks and thank you for the work!

dkgroot:
@nemanja-boric-sociomantic Thanks and Thanks for the review.
I could use a little help with these three open FIXME's. It is not always clear what the intent of a unittest is. Some of them could use a couple of comments as well.

posix.mak
dkgroot: This not directly related to supporting DragonFlyBSD, and should maybe have been in a separate commit.

src/core/memory.d
nemanja-boric-sociomantic: I can't find any documented reference, but I believe this is testing if the glibc's malloc would fail because of the internal block size and the additional alignment (and any other internal metadata) that should be added to the internal allocation, which while adding with the values close to `size_t.max` would lead to the overflow.

posix.mak
```diff
@@ -3,6 +3,7 @@
 #    pkg_add -r gmake
 # and then run as gmake rather than make.
 
+#QUIET:=@
```
wilzbach: What's the reason for this addition?

posix.mak
```diff
@@ -57,8 +59,6 @@ endif
 
 # build with shared library support
 # (defaults to true on supported platforms, can be overridden w/ make SHARED=0)
-SHARED=$(if $(findstring $(OS),linux freebsd),1,)
```
wilzbach: Huh?

posix.mak
wilzbach: It's not as easy as that. There's currently no logic in place that ensures that `tools` is actually checked out and there have been long discussion of tools being an option or required dependency for building DMD + Druntime + Phobos.
The current status quo is that it's okay to use use `tools` in non essential targets.
Anyhow as you said yourself, please put this in a separate PR.

posix.mak
dkgroot: Reverted all the posix.mak changes i made (They should not have been committed).

src/rt/util/typeinfo.d
nemanja-boric-sociomantic: You can see the `F` array there,it has four elements, that they all have the same value in the complex field, and first assert is checking if their value comparison is sane (they all should equal each other, as they should all be in the `(0, 0)` point). The next assert (I assume this is the failing one, you forgot braces around `else`) `f1 !is f2` checks if these elements are all different (because they are different members of the array). Not sure what's going on in here.

src/core/memory.d
dkgroot: Wrote a little test program in ansi-c
```c
#include <stdio.h>
#include <stdlib.h>

// On Linux
#if !defined(SIZE_MAX)
#define SIZE_MAX __SIZE_MAX__		// ((size_t)-1) = 9223372036854775807 = LONG_MAX
#endif
// On DragonFlyBSD
//#define SIZE_MAX UINT64_MAX  // from /usr/include/machine/int_limits.h

int main () {
  printf("SIZE_MAX=%lu\n", SIZE_MAX);
  
  unsigned long tests[] = {
    9223372036854775807,
    223372036854775808,
    23372036854775808,
    3372036854775808,
    372036854775808,
    72036854775808,
    2036854775808,
    36854775808,
    SIZE_MAX - 2,                                    // The previous value for the test // https://issues.dlang.org/show_bug.cgi?id=17956
    SIZE_MAX,
    SIZE_MAX + 'a',                                  // simulating SIZE_MAX ~ 255
    SIZE_MAX + SIZE_MAX,
    SIZE_MAX * 20};
  
  for (int test = 0; test < 12; test++) {
    printf("test%d: malloc size:%lu\n", test, tests[test]);

    int *ptr = (int *) malloc(tests[test]);
    if (ptr == NULL) {
        printf("malloc failed, ptr == NULL\n");
    } else {
        printf(" ptr == %p\n", ptr);
        free(ptr);
        ptr = NULL;
    }
  }
  return 0;
}
```
Compiled using `gcc test_malloc.c -o test_malloc`

Linux x86_64 Results:
```
SIZE_MAX=18446744073709551615
test0: malloc size:9223372036854775807
malloc failed, ptr == NULL
test1: malloc size:223372036854775808
malloc failed, ptr == NULL
test2: malloc size:23372036854775808
malloc failed, ptr == NULL
test3: malloc size:3372036854775808
malloc failed, ptr == NULL
test4: malloc size:372036854775808
malloc failed, ptr == NULL
test5: malloc size:72036854775808
malloc failed, ptr == NULL
test6: malloc size:2036854775808
malloc failed, ptr == NULL
test7: malloc size:36854775808
malloc failed, ptr == NULL
test8: malloc size:18446744073709551613
malloc failed, ptr == NULL
test9: malloc size:18446744073709551615
malloc failed, ptr == NULL
test10: malloc size:96
 ptr == 0x7fd0700008c0
test11: malloc size:18446744073709551614
malloc failed, ptr == NULL
```

DragonFlyBSD (x86_64) Resuls:
```
SIZE_MAX=18446744073709551615
test0: malloc size:9223372036854775807
malloc failed, ptr == NULL
test1: malloc size:223372036854775808
malloc failed, ptr == NULL
test2: malloc size:23372036854775808
malloc failed, ptr == NULL
test3: malloc size:3372036854775808
malloc failed, ptr == NULL
test4: malloc size:372036854775808
malloc failed, ptr == NULL
test5: malloc size:72036854775808
 ptr == 0x800800000
test6: malloc size:2036854775808
 ptr == 0x800800000
test7: malloc size:36854775808
 ptr == 0x800800000
test8: malloc size:18446744073709551613
 ptr == 0x800455000
test9: malloc size:18446744073709551615
 ptr == 0x800455000
test10: malloc size:96
 ptr == 0x8006200c0
test11: malloc size:18446744073709551614
 ptr == 0x800455000
```
On DragonFly it seems to be ok, as long as the upperbound of '9223372036854775807' is observed. Using SIZE_MAX however results faulty behaviour.

nemanja-boric-sociomantic: Perhaps is worth raising the bug report in the DragoFlyBSD tracker?

src/rt/util/typeinfo.d
```diff
@@ -247,7 +247,8 @@ unittest
...
             assert(f1 !is f2);
```
dkgroot: This is the failing assertion.

I think i might need some help from one of the 'src/rt/typeinfo/*' code owners on how to debug this.

(The {} are not missing, it's only trying to exclude a single line, without having to change it's surroundings, Wanted to keep the 'DragonFlyBSD' string and 'FIXME' on a single line, so it's easy to find using grep. I noticed this 'unittest exclusion' pattern in other commits.).

src/rt/util/typeinfo.d
dkgroot: (The {} are not missing, it's only excluding a single line, without having to change it's surroundings, Wanted to keep the 'DragonFlyBSD' string and 'FIXME' on a single line, so it's easy to find using grep. I noticed this 'unittest exclusion' pattern in other commits.).

src/core/memory.d
dkgroot: After creating the above 'proof' i think that makes sense.
Once the bug report i made, i can attach a link to the 'unittest exclusion', so it can be tracked.

src/core/stdc/errno.c
```diff
@@ -23,3 +30,4 @@ int setErrno( int val )
     errno = val;
     return val;
 }
+#endif
```
ibuclaw: Just a nit, but use an editor that adds newline at bottom of files.

src/core/stdc/fenv.d
```diff
@@ -258,6 +258,24 @@ else version ( OpenBSD )
...
+        struct _x87
+        {
+                uint control;
+                uint status;
+                uint tag;
+                uint[4] others;
+        };
+        _x87 x87;
+
+        uint mxcsr;
```
ibuclaw: Not related to this patch, but I've just noticed that the OpenBSD branch is wrong.

src/core/sys/posix/netdb.d
CyberShadow: Would it not be better to change `std.socket` instead?

dkgroot: You mean to set the default value for ai_socktype, right ?

I did try to find a good location in std.socket first (When the unittest failed and pointed out the issue). I tried this, and it worked, but i did not like this solution very much. Maybe mostly caused by the location i choose to 'fix' the issue.
```diff
diff --git a/std/socket.d b/std/socket.d
index 361ec73da..ba68e1738 100644
--- a/std/socket.d
+++ b/std/socket.d
@@ -963,6 +963,13 @@ AddressInfo[] getAddressInfo(T...)(in char[] node, T options)
         else
             static assert(0, "Unknown getAddressInfo option type: " ~ typeof(option).stringof);
     }
+    version (DragonFlyBSD)
+    {
+        if (!hints.ai_socktype)
+        {
+            hints.ai_socktype = SOCK_STREAM;
+        }
+    }
 
     return () @trusted { return getAddressInfoImpl(node, service, &hints); }();
 }
```

src/core/stdc/stdio.d
```diff
@@ -1372,6 +1451,59 @@ else version( OpenBSD )
...
+  enum __SALC = 0x4000;
+  enum __SIGN = 0x8000;
+
+  int  snprintf(scope char* s, size_t n, scope const char* format, ...);
+  int  vsnprintf(scope char* s, size_t n, scope const char* format, va_list arg);
+}
+else version (Solaris)
+{
+  // No unsafe pointer manipulation.
+  @trusted
```
n8sh: It looks like this `version(Solaris)` block has been unintentionally duplicated.

joakim-noah:
Good work.  Given the giant size of this pull, I suggest you split it up into three pulls, something like all `core.sys.posix.*` changes in one pull, `core.sys.dragonflybsd.*` in another, and leaving the rest here.  [That's what was done with the NetBSD pull that got merged before](https://github.com/dlang/druntime/pull/1492#issuecomment-179698335).

dkgroot:
@joakim-noah 
Can you give me some instructions on how to do that usng git. Creating the new pull requests would not be so difficult. But reducing the size of this one without reopening it.

joakim-noah:
Not familiar with the `git checkout -p` command, but the split appears to have worked.  I'll check to make sure everything's scoped properly and for basic formatting, which is all I was looking for from the NetBSD pull.

src/core/stdc/stdio.d
```diff
@@ -863,6 +924,24 @@ else version( OpenBSD )
...
+
+    private extern shared FILE* __stdinp;
+    private extern shared FILE* __stdoutp;
+    private extern shared FILE* __stderrp;
+
+    alias __stdinp  stdin;
+    alias __stdoutp stdout;
+    alias __stderrp stderr;
+}
+
```
joakim-noah: Take out added blank line.

src/core/time.d
```diff
@@ -386,6 +396,20 @@ version(Posix)
...
+            {
+            case coarse: return CLOCK_MONOTONIC_FAST;
+            case normal: return CLOCK_MONOTONIC;
+            case precise: return CLOCK_MONOTONIC_PRECISE;
+            case uptime: return CLOCK_UPTIME;
+            case uptimeCoarse: return CLOCK_UPTIME_FAST;
+            case uptimePrecise: return CLOCK_UPTIME_PRECISE;
+            case second: assert(0);
+            }
+        }
```
joakim-noah: You copy-and-pasted this block a second time below.

dkgroot: Thanks... Done

src/core/sys/osx/sys/mman.d
```diff
@@ -2,7 +2,7 @@
  * $(RED Deprecated. Use $(D core.sys.darwin.sys.mman) instead. This module
  *       will be removed in June 2018.)
  *
- * D header file for FreeBSD
+ * D header file for OSX
```
driusan: This change doesn't look related to DragonFly?

joakim-noah: Yep, just a random comment fix, no problem with including it.

src/core/stdc/errno.d
```diff
@@ -1822,6 +1830,109 @@ else version( OpenBSD )
...
+    enum ENOLINK            = 91;       /// Link has been severed
+    enum EPROTO             = 92;       /// Protocol error
+    enum ENOMEDIUM          = 93;
+    enum EUNUSED94          = 94;
+    enum EUNUSED95          = 95;
+    enum EUNUSED96          = 96;
+    enum EUNUSED97          = 97;
+    enum EUNUSED98          = 98;
+    enum EASYNC             = 99;
+    enum ELAST              = 99;       /// Must be equal largest errno
```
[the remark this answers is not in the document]
dkgroot: Not sure if this is revelevant. This __BSD_VISIBLE was copied (See:  src/core/sys/dragonflybsd/dlfcn.d:17) but hard coded to be 1. My interpretation of the BSD_VISIBLE 'marker' is that it is used to make header files portable and is always 1 on BSD platforms. As this section is only relevant on DragonFly, adding a static if is not going to make a difference.

src/core/stdc/math.d
```diff
@@ -1091,6 +1098,70 @@ else version( NetBSD )
...
+        FP_NAN       = 0x02,
+        FP_NORMAL    = 0x04,
+        FP_SUBNORMAL = 0x08,
+        FP_ZERO      = 0x10,
+    }
+
+    enum
+    {
+        FP_FAST_FMA  = 0,
+        FP_FAST_FMAF = 0,
```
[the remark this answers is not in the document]
dkgroot: Agreed, commented out the first two an added a reference comment about the origin.

src/core/stdc/stdio.d
```diff
@@ -551,6 +594,24 @@ else version( OpenBSD )
     ///
     alias shared(__sFILE) FILE;
 }
+else version( DragonFlyBSD )
+{
+    alias off_t fpos_t;
+
+    /// See /usr/include/stdio.h
```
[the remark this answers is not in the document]
dkgroot: Will rename `struct __sFILE` to `struct __FILE_public` as per your request.

src/core/sys/osx/sys/mman.d
dkgroot: Reverted
@joakim-noah  Woops :-)

joakim-noah: You can just remove that last revert commit.

joakim-noah:
btw, you don't need to mirror your druntime/phobos patches here to the LDC repos.  I'm going to compare your pulls here and just try to cherry-pick and merge them to the LDC ltsmaster and master branches once they're merged here, so you can just limit your ldc pulls eventually to any changes you had to make in addition to these.

For now, it doesn't hurt that you're trying to keep the two in sync, and the two separate PRs point me to potential merge conflicts, but just giving you a head's up that you don't need to spend a ton of time on the ldc pulls, as I'll likely just key off these upstream pulls.

dkgroot:
@joakim-noah Ooh that's  a real relief. Trying to juggle / sync 11 PR's was a little daunting, to be honest. Nice to see that we seem to be making real progress on all of this :-)

src/core/stdc/errno.c
```diff
@@ -11,7 +11,14 @@
...
 #include <errno.h>
 
-
+#if __DragonFly__
+#define _errno errno
+#undef errno
+int * __error1(void) {
+    return (&_errno);
+}
+#define	errno		(* __error1())
```
dkgroot: I had some real difficulty mapping the DragonFly errno.h/error() implementation to errno.d. Mostly because it already defined an __error() function. The only way i could make it to be behave was by recreating the macro, using a different __error1() function and map to that. (BTW: Trying to use the setErrno() did not work either).
If someone has a better/nicer solution, or would like me to move this 'implementation' into /src/core/sys/dragonflybsd for example. just let me know.

joakim-noah: Took a look at [the Dragonfly header](http://fxr.watson.org/fxr/source/sys/errno.h?v=DFBSD#L44): can't you just replace all this C renaming with `int * __error1(void) {return __error();}`, to turn the header-only `__error` into a compiled `__error1`, then reference it as you do from D?  What am I missing?

dkgroot: @joakim-noah I had not considered wrapping it :-) Testing your solution now.

src/core/memory.d
dkgroot: Found a potential solution to the malloc(SIZE_MAX) issue. Send a diff to upstream dragonfly issue manager (Issue [3114](https://bugs.dragonflybsd.org/issues/3114)), seems to resolve the case reported above.

dkgroot: Issue has been fixed in dragonfly master
Removing this unittest escape

src/core/sync/mutex.d
dkgroot: Have not found a good solution for this. But trying to lock after destroying the lock does not really make sense anyway. Reading up on some of the documentation regarding destroying a lock, it looks like it is underfined behaviour to try and lock the mutex after destruction. Even though it seems to "work" in some implementations.

src/rt/util/typeinfo.d
dkgroot: Actual assert Issue reraised below.

src/core/sync/mutex.d
dkgroot: Leaving it like this for now. 
Would require changing pthread implementation on dragonfly, or adding a check to "class Mutex" to verify if it has been initialized/constructed. Both don't really seem reasonable.

src/rt/util/typeinfo.d
dkgroot: @mathias-lang-sociomantic @Geod24 @WalterBright @redstar
Can one of you help me figure out why/how 'assert(f1 !is f2);' could be failing.
Could it be that comparing complex numbers is causing trouble. Does one of you have a good way of tracing ?

src/core/stdc/errno.c
```diff
@@ -11,7 +11,11 @@
 
 #include <errno.h>
 
-
+#if __DragonFly__
+int * __error1(void) {
+    return __error();
+}
+#else
```
joakim-noah: Thinking about this some more, the idea behind the block in `core.stdc.errno` below is to avoid needing this C file from druntime, as implemented in #1917.  Would something like this work to replace the D declaration and alias below, so you could drop this C patch altogether?
```
pragma(mangle, "errno") extern int __errno;
ref int errno() { return __errno;}
```

dkgroot: @joakim-noah Thanks for this much nicer solution !

joakim-noah: Does it work, ie all the stdlib tests pass?

dkgroot: Yes it does. I run all the unittests before each/any commit.

src/core/stdc/errno.c
```diff
@@ -11,7 +11,6 @@
 
 #include <errno.h>
 
-
```
ibuclaw: Unrelated change here.

dkgroot:
Hi guys, would it help if this PR were be split by file, so that it would make a little more progress ?

wilzbach:
>> Hi guys, would it help if this PR were be split by file, so that it would make a little more progress ?
> No, somebody just needs to make sure all review comments have been addressed and make a decision.

In my experience it does help. However, splitting it up by file might be a bit to granular.
Did you see @nemanja-boric-sociomantic's comment at the top:

> It would be very nice if you could split this into several commits (as per your checklist in the PR)

For example, `gc` + `rt`could one "package" PR:

image

- this is more work entirely on your side (for reviewers smaller PRs are less "mental" work)
- as @joakim-noah pointed out it's not a requirement

dkgroot:
Ping @ZombineDev @CyberShadow @klickverbot @MartinNowak 
Are there any issues with this PR ? Thanks in advance for you review !

[DragonFly Based CI](https://semaphoreci.com/dkgroot/dmd_dragonfly_ci)

dkgroot:
@wilzbach Thanks a lot !

posix.mak
```diff
@@ -204,7 +204,7 @@ $(IMPDIR)/%.d : src/%.d
 ######################## Build DMD if non-existent ##############################
 
 $(DMD):
-	make -C $(DMD_DIR)/src -f posix.mak BUILD=$(BUILD) OS=$(OS) MODEL=$(MODEL)
+	$(MAKE) -C $(DMD_DIR)/src -f posix.mak BUILD=$(BUILD) OS=$(OS) MODEL=$(MODEL)
```
joakim-noah: I see one more on line 541, don't know if you care about that one.

src/core/stdc/errno.d
```diff
@@ -1830,6 +1835,109 @@ else version( OpenBSD )
     enum ENOTSUP            = 91;       /// Not supported
     enum ELAST              = 91;       /// Must be equal largest errno
 }
+else version( DragonFlyBSD )
+{
+    enum EPERM              = 1;        /// Operation not permitted
```
joakim-noah: I'm guessing you didn't write these comments yourself but copied them from somewhere?  If so, take out all the comments that you copied, see the linux block for an example.  We don't really want to deal with copyright issues.

src/core/sync/mutex.d
```diff
@@ -378,6 +378,7 @@ unittest
     // not run on Android.
     version (CRuntime_Bionic) {} else
     version (CRuntime_Musl) {} else
+    version (DragonFlyBSD) {} else               // FIXME : Will have to write a test program to see what is going on here.
```
joakim-noah: Still need to test this?

src/rt/backtrace/dwarf.d
```diff
@@ -14,6 +14,7 @@ module rt.backtrace.dwarf;
 
 version(CRuntime_Glibc) version = linux_or_freebsd;
 else version(FreeBSD) version = linux_or_freebsd;
+else version(DragonFlyBSD) version = linux_or_freebsd;
```
joakim-noah: As long as we're here, can you change this to `glibc_or_bsd`?  This name has annoyed me for awhile.

src/rt/backtrace/elf.d
```diff
@@ -13,6 +13,7 @@ module rt.backtrace.elf;
 
 version(linux) version = linux_or_freebsd;
 else version(FreeBSD) version = linux_or_freebsd;
+else version(DragonFlyBSD) version = linux_or_freebsd;
```
joakim-noah: Same here, `linux_or_bsd`.

posix.mak
dkgroot: I just moved the posix.mak file to the other PR (#2002), because it was mostly needed to compile the dragonflybsd specific files, so a better fit (and i hope it will fix the autotester issues when compiling on win32).
I did checkout the posix.mak file,  but it only has 374 lines.

src/core/stdc/errno.d
dkgroot: WIll do !
Funny how to most 'copyright bothersome os-ses' (ie: windows and osx) do have similar comments :-) Might be a good idea, to to the same for them as well.

src/core/sync/mutex.d
dkgroot: Tested / Checked with dragonflybsd developers / Will not be corrected
FIXME: Comment removed
Comment above the test extended/updated.

posix.mak
joakim-noah: Whoops, checked the [Phobos posix.mak for `make` by mistake](https://github.com/dlang/phobos/blob/master/posix.mak#L531).

src/rt/backtrace/dwarf.d
dkgroot: Used "glibc_or_bsdlibc" instead (thought that would be a better fit, and also usable by netbsd/openbsd when needed), if that is ok with you.

src/core/sync/mutex.d
```diff
@@ -374,11 +374,11 @@ unittest
     // by checking that locking is not possible. This assumes
     // that the underlying implementation is well behaved
     // and makes the object non-lockable upon destruction.
-    // For example, Bionic doesn't appear to do so, so this test is
-    // not run on Android.
+    // BTW: Bionic/Musl and DragonFlyBSD don't (appear) to do so, so this test is
+    // not skipped on/for them.
```
joakim-noah: Change to 'The Bionic and Musl C runtimes and DragonFly don't appear to do so, so skip this test.`

src/rt/backtrace/elf.d
```diff
@@ -11,16 +11,18 @@
 
 module rt.backtrace.elf;
 
-version(linux) version = linux_or_freebsd;
-else version(FreeBSD) version = linux_or_freebsd;
+version(linux) version = glibc_or_bsdlibc;
```
joakim-noah: This won't actually enable `Glibc`, as `linux` refers to kernel APIs.  I don't know where these declarations are from, so either change the version selected to `CRuntime_Glibc` or the new version to `linux_or_bsd`.

dkgroot:
@joakim-noah Thanks for your remarks/comments/review !

posix.mak
dkgroot: Moving the make files to the other PR, seems to have been a good idea. It fixed the autotester win32 issue.
I will check the phobos posix.mak file, to see if it needs correcting (And create a PR if necessary).

PR #2001, opened 2017-12-19
 Add DragonFlyBSD support : src/core/sys/posix related changes

dkgroot:
Split PR : #1999 
Moved the src/core/sys/posix related changes into this PR

### DIFF
--- a/src/core/sys/posix/signal.d
+++ b/src/core/sys/posix/signal.d
@@ -164,6 +164,10 @@ else version (FreeBSD) {
     enum SIGRTMIN = 65;
     enum SIGRTMAX = 126;
 }
+else version (DragonFlyBSD) {
+    enum SIGRTMIN = 35;
+    enum SIGRTMAX = 126;
+}
 else version(NetBSD)
 {
     enum SIGRTMIN = 33;
@@ -523,6 +527,30 @@ else version( OpenBSD )
     enum SIGUSR2    = 31;
     enum SIGURG     = 16;
 }
+else version( DragonFlyBSD )
+{
+    //SIGABRT (defined in core.stdc.signal)
+    enum SIGALRM    = 14;
+    enum SIGBUS     = 10;
+    enum SIGCHLD    = 20;
+    enum SIGCONT    = 19;
+    //SIGFPE (defined in core.stdc.signal)
+    enum SIGHUP     = 1;
+    //SIGILL (defined in core.stdc.signal)
+    //SIGINT (defined in core.stdc.signal)
+    enum SIGKILL    = 9;
+    enum SIGPIPE    = 13;
+    enum SIGQUIT    = 3;
+    //SIGSEGV (defined in core.stdc.signal)
+    enum SIGSTOP    = 17;
+    //SIGTERM (defined in core.stdc.signal)
+    enum SIGTSTP    = 18;
+    enum SIGTTIN    = 21;
+    enum SIGTTOU    = 22;
+    enum SIGUSR1    = 30;
+    enum SIGUSR2    = 31;
+    enum SIGURG     = 16;
+}
 else version (Solaris)
 {
     enum SIGALRM = 14;
@@ -607,6 +635,19 @@ else version(OpenBSD)
         }
         sigset_t sa_mask;
         int      sa_flags;
+    }
+}
+else version( DragonFlyBSD )
+{
+    struct sigaction_t
+    {
+        union
+        {
+            sigfn_t     sa_handler;
+            sigactfn_t  sa_sigaction;
+        }
+        int      sa_flags;
+        sigset_t sa_mask;
     }
 }
 else version (Solaris)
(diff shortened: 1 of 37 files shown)


src/core/sys/posix/sys/socket.d
```diff
@@ -1089,7 +1089,6 @@ else version(NetBSD)
          SO_NOSIGPIPE    = 0x0800,          /* no SIGPIPE from EPIPE */
          SO_ACCEPTFILTER = 0x1000,          /* there is an accept filter */
          SO_TIMESTAMP    = 0x2000,          /* timestamp received dgram traffic */
-
```
joakim-noah: No reason to take this out.

joakim-noah:
Ping, @wilzbach, checked that everything was scoped to DragonFly but not for correctness.  [There was only a single comment on these headers when put up for review before](https://github.com/dlang/druntime/pull/1999#discussion_r157586930).

src/core/sys/posix/signal.d
```diff
@@ -1181,6 +1222,54 @@ else version( OpenBSD )
     int sigsuspend(in sigset_t *);
     int sigwait(in sigset_t*, int*);
 }
+else version( DragonFlyBSD )
+{
+    enum SIG_HOLD = cast(sigfn_t2) 3;
```
wilzbach: `enum sigfn_t2`

dkgroot: @wilzbach Can you clarify what you want me to do ?

joakim-noah: He's saying to just set the type of the enum, rather than casting the value.  Does that work for a function pointer like this?

dkgroot: Done... Thanks !
Note: if this is an important change, the other version(OS) entries do perform the same cast, so you might want to make a similar change to the other implementations (in a separate PR).

src/core/sys/posix/signal.d
```diff
@@ -1224,8 +1224,8 @@ else version( OpenBSD )
 }
 else version( DragonFlyBSD )
 {
-    enum SIG_CATCH = cast(sigfn_t2) 2;
-    enum SIG_HOLD = cast(sigfn_t2) 3;
+    enum SIG_CATCH = 2;
+    enum SIG_HOLD = 3;
```
CyberShadow: I think Sebastian meant to change this to `enum sigfn_t2 SIG_CATCH = 2;`, i.e. move the type from the cast to the enum declaration, but not sure if it makes any difference.

dkgroot: @CyberShadow Thanks for reviewing the code :-)
I just made that change (a couple of minutes ago).

ibuclaw:
@dkgroot rebase / fix up?

src/core/sys/posix/signal.d
CyberShadow: >  I just made that change (10 minutes ago).

Unless you're talking about something else which you haven't pushed yet, then my comment was about the code you did just push. There is no type declaration in it. I don't know if it matters or not, I just wanted to clarify on Sebastian's behalf.

To clarify even further, your code goes:

```d
enum SIG_CATCH = 2;
```

I believe the suggested change was:

```d
enum sigfn_t2 SIG_CATCH = 2;
```

Hopefully that should clear things up.

dkgroot: @CyberShadow That gives 
"Error: cannot implicitly convert expression 2 of type int to extern (C) void function(int) nothrow @nogc"

I think i need to actually revert to the original definition using the cast.

dkgroot: Reverted the previous patch..

dkgroot: Ping @wilzbach , @joakim-noah and @CyberShadow 
Reverted the previous patch... I think the requested change might require a little more deliberation/discussion.

Source fragment (to give some context):
```
nothrow @nogc
{
    private alias void function(int) sigfn_t2;
    private alias void function(int, siginfo_t*, void*) sigactfn_t2;
}
...
else version( DragonFlyBSD )
{
    enum SIG_CATCH = cast(sigfn_t2) 2;
    enum SIG_HOLD = cast(sigfn_t2) 3;
...
```

dkgroot:
Thanks for all your reviewing work guys !

src/core/sys/posix/signal.d
joakim-noah: The way you had it initially was fine.  We weren't sure if you could just apply the type to the left, but as you found, it needs to be cast.

joakim-noah:
It shouldn't matter to any of our CIs, since we don't test DragonFlyBSD and these are all scoped.  I'm not sure what's up with the single D project that fails, unit-threaded, guessing that's unrelated to this PR.

src/core/sys/posix/signal.d
dkgroot: Thanks for the update 😉. Glad I rolled it back just before ibuclaw merged it 😙

dkgroot:
I saw some issues in the Autotester on its windows slaves complaining about these prs. Not 100% what was causing the issue.

wilzbach:
Let us know about spurious failures in #ci

dkgroot:
@wilzbach will do. Looks ok at the moment

dkgroot:
Thanks /Danke schön

src/core/sys/posix/signal.d
dkgroot: @CyberShadow As mentioned in the dlang/druntime PR, that does not work (compiler error), i reverted it to the original version using the cast. The PR has already been merged now. If it does need to change (now or in the future), it would have to be done in a seperate PR and should also involve the other OS implementations.

CyberShadow: Yes, I saw the earlier messages. No objections to that, of course.

PR #2002, opened 2017-12-19
 Add DragonFlyBSD support : Split PR : src/core/sys/dragonflybsd related changes

dkgroot:
Split PR : #1999
Moved the src/core/sys/dragonflybsd related changes into this PR

Related PR:
- https://github.com/dlang/dmd/pull/7463

### DIFF
--- a/src/core/sys/dragonflybsd/dlfcn.d
+++ b/src/core/sys/dragonflybsd/dlfcn.d
@@ -1,0 +1,107 @@
+/**
+ * D header file for DragonFlyBSD
+ *
+ * Copyright: Copyright Martin Nowak 2012.
+ * License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Authors: Martin Nowak,Diederik de Groot(port:DragonFlyBSD)
+ * Copied:  From core/sys/freebsd/sys
+ */
+module core.sys.dragonflybsd.dlfcn;
+
+version (DragonFlyBSD):
+
+public import core.sys.posix.dlfcn;
+
+extern (C) nothrow @nogc @system:
+
+/*
+ * Modes and flags for dlopen().
+ */
+static assert(RTLD_LAZY   == 1);
+static assert(RTLD_NOW    == 2);
+enum RTLD_MODEMASK        =  0x3;
+static assert(RTLD_GLOBAL == 0x100);
+static assert(RTLD_LOCAL  == 0);
+enum RTLD_TRACE           =  0x200;
+enum RTLD_NODELETE        =  0x01000;
+enum RTLD_NOLOAD          =  0x02000;
+
+/*
+ * Request arguments for dlinfo().
+ */
+enum RTLD_DI_LINKMAP     = 2;    /* Obtain link map. */
+enum RTLD_DI_SERINFO     = 4;    /* Obtain search path info. */
+enum RTLD_DI_SERINFOSIZE = 5;    /*  ... query for required space. */
+enum RTLD_DI_ORIGIN      = 6;    /* Obtain object origin */
+enum RTLD_DI_MAX         = RTLD_DI_ORIGIN;
+
+/*
+ * Special handle arguments for dlsym()/dlinfo().
+ */
+enum RTLD_NEXT    = cast(void *)-1;    /* Search subsequent objects. */
+enum RTLD_DEFAULT = cast(void *)-2;    /* Use default search algorithm. */
+enum RTLD_SELF    = cast(void *)-3;    /* Search the caller itself. */
+
+/*
+ * Structure filled in by dladdr().
+ */
+struct Dl_info {
+    const(char)     *dli_fname;     /* Pathname of shared object. */
+    void            *dli_fbase;     /* Base address of shared object. */
+    const(char)     *dli_sname;     /* Name of nearest symbol. */
+    void            *dli_saddr;     /* Address of nearest symbol. */
+};
+
+
+/*
+ * Structures, returned by the RTLD_DI_SERINFO dlinfo() request.
+ */
+struct Dl_serpath {
+    char *          dls_name;       /* single search path entry */
+    uint            dls_flags;      /* path information */
+};
+
+struct Dl_serinfo {
+    size_t          dls_size;       /* total buffer size */
+    uint            dls_cnt;        /* number of path entries */
+    Dl_serpath[1]   dls_serpath;    /* there may be more than one */
+};
+
+/*-
+ * The actual type declared by this typedef is immaterial, provided that
+ * it is a function pointer.  Its purpose is to provide a return type for
+ * dlfunc() which can be cast to a function pointer type without depending
+ * on behavior undefined by the C standard, which might trigger a compiler
+ * diagnostic.  We intentionally declare a unique type signature to force
+ * a diagnostic should the application not cast the return value of dlfunc()
+ * appropriately.
+ */
+struct __dlfunc_arg {
+    int     __dlfunc_dummy;
+};
+
+alias dlfunc_t = void function(__dlfunc_arg);
+
+private template __externC(RT, P...)
+{
+    alias __externC = extern(C) RT function(P) nothrow @nogc @system;
+}
+
+/* XSI functions first. */
+static assert(is(typeof(&dlclose) == __externC!(int, void*)));
+static assert(is(typeof(&dlerror) == __externC!(char*)));
+static assert(is(typeof(&dlopen)  == __externC!(void*, const char*, int)));
+static assert(is(typeof(&dlsym)   == __externC!(void*, void*, const char*)));
+
+void*    fdlopen(int, int);
+int      dladdr(const(void)*, Dl_info*);
+dlfunc_t dlfunc(void*, const(char)*);
+int      dlinfo(void*, int, void*);
+/*void     dllockinit(void* _context,
+    void* function(void* _context) _lock_create,
+    void  function(void* _lock)    _rlock_acquire,
+    void  function(void* _lock)    _wlock_acquire,
+    void  function(void* _lock)    _lock_release,
+    void  function(void* _lock)    _lock_destroy,
+    void  function(void* _context) _context_destroy);*/
+void*    dlvsym(void*, const(char)*, const(char)*);
--- a/src/core/sys/dragonflybsd/netinet/in_.d
+++ b/src/core/sys/dragonflybsd/netinet/in_.d
@@ -1,0 +1,551 @@
+/**
+ * D header file for DragonFlyBSD's extensions to POSIX's netinet/in.h.
+ *
+ * Copyright: Copyright 2017 -
+ * License:   $(WEB www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Authors: Jonathan M Davis, Diederik de Groot(port:DragonFlyBSD)
+ * Copied:  From core/sys/freebsd/sys
+ */
+module core.sys.dragonflybsd.netinet.in_;
+
+version(DragonFlyBSD):
+
+import core.sys.dragonflybsd.sys.cdefs;
+
+public import core.sys.posix.netinet.in_;
+
+extern(C) nothrow @nogc @system:
+
+enum IPPROTO_HOPOPTS   = 0;
+
+enum IPPROTO_IPV4      = 4;
+enum IPPROTO_IPIP      = IPPROTO_IPV4;
+enum IPPROTO_ST        = 7;
+enum IPPROTO_EGP       = 8;
+enum IPPROTO_PIGP      = 9;
+enum IPPROTO_RCCMON    = 10;
+enum IPPROTO_NVPII     = 11;
+
+enum IPPROTO_ARGUS     = 13;
+enum IPPROTO_EMCON     = 14;
+enum IPPROTO_XNET      = 15;
+enum IPPROTO_CHAOS     = 16;
+enum IPPROTO_MUX       = 18;
+enum IPPROTO_MEAS      = 19;
+enum IPPROTO_HMP       = 20;
+enum IPPROTO_PRM       = 21;
+
+enum IPPROTO_TRUNK1    = 23;
+enum IPPROTO_TRUNK2    = 24;
+enum IPPROTO_LEAF1     = 25;
+enum IPPROTO_LEAF2     = 26;
+enum IPPROTO_RDP       = 27;
+enum IPPROTO_IRTP      = 28;
+enum IPPROTO_TP        = 29;
+enum IPPROTO_BLT       = 30;
+enum IPPROTO_NSP       = 31;
+enum IPPROTO_INP       = 32;
+enum IPPROTO_SEP       = 33;
+enum IPPROTO_3PC       = 34;
+enum IPPROTO_IDPR      = 35;
+enum IPPROTO_XTP       = 36;
+enum IPPROTO_DDP       = 37;
+enum IPPROTO_CMTP      = 38;
+enum IPPROTO_TPXX      = 39;
+enum IPPROTO_IL        = 40;
+enum IPPROTO_SDRP      = 42;
+enum IPPROTO_ROUTING   = 43;
+enum IPPROTO_FRAGMENT  = 44;
+enum IPPROTO_IDRP      = 45;
+enum IPPROTO_RSVP      = 46;
+enum IPPROTO_GRE       = 47;
+enum IPPROTO_MHRP      = 48;
+enum IPPROTO_BHA       = 49;
+enum IPPROTO_ESP       = 50;
+enum IPPROTO_AH        = 51;
+enum IPPROTO_INLSP     = 52;
+enum IPPROTO_SWIPE     = 53;
+enum IPPROTO_NHRP      = 54;
+enum IPPROTO_MOBILE    = 55;
+enum IPPROTO_TLSP      = 56;
+enum IPPROTO_SKIP      = 57;
+enum IPPROTO_ICMPV6    = 58;
+enum IPPROTO_NONE      = 59;
+enum IPPROTO_DSTOPTS   = 60;
+enum IPPROTO_AHIP      = 61;
+enum IPPROTO_CFTP      = 62;
+enum IPPROTO_HELLO     = 63;
+enum IPPROTO_SATEXPAK  = 64;
+enum IPPROTO_KRYPTOLAN = 65;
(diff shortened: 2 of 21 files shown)


src/core/sys/dragonflybsd/dlfcn.d
```diff
@@ -0,0 +1,116 @@
...
+ * Copied:  From core/sys/freebsd/sys
+ */
+module core.sys.dragonflybsd.dlfcn;
+
+public import core.sys.posix.dlfcn;
+
+version (DragonFlyBSD):
+extern (C):
+nothrow:
+@nogc:
```
joakim-noah: You can put all three attributes on the same line, ie `extern (C) nothrow @nogc:` as done for some other files below, probably better style.  Also, should most of these headers be `@system` too, @MartinNowak?

dkgroot: Putting the attributes on one line -> No problem.
Adding @system attribute to all of these header files, i gave that a try and all the unittests are still passed without a problem. If that is an indication that it is ok, i can commit the change.

dkgroot: Proposed
```diff
diff --git a/src/core/sys/dragonflybsd/dlfcn.d b/src/core/sys/dragonflybsd/dlfcn.d
index 9a416525..86effe89 100644
--- a/src/core/sys/dragonflybsd/dlfcn.d
+++ b/src/core/sys/dragonflybsd/dlfcn.d
@@ -11,9 +11,8 @@ module core.sys.dragonflybsd.dlfcn;
 public import core.sys.posix.dlfcn;
 
 version (DragonFlyBSD):
-extern (C):
-nothrow:
-@nogc:
+
+extern (C) nothrow @nogc @system:
 
 enum __BSD_VISIBLE = true;
 
diff --git a/src/core/sys/dragonflybsd/sys/_bitset.d b/src/core/sys/dragonflybsd/sys/_bitset.d
index 079115d7..24f9d9c1 100644
--- a/src/core/sys/dragonflybsd/sys/_bitset.d
+++ b/src/core/sys/dragonflybsd/sys/_bitset.d
@@ -7,7 +7,7 @@
 module core.sys.dragonflybsd.sys._bitset;
 
 version (DragonFlyBSD):
-extern (C) pure nothrow @nogc:
+extern (C) pure nothrow @nogc @system:
 
 import core.stdc.config : c_long;
 
diff --git a/src/core/sys/dragonflybsd/sys/elf_common.d b/src/core/sys/dragonflybsd/sys/elf_common.d
index cd94832d..4bbddf24 100644
--- a/src/core/sys/dragonflybsd/sys/elf_common.d
+++ b/src/core/sys/dragonflybsd/sys/elf_common.d
@@ -7,9 +7,8 @@
 module core.sys.dragonflybsd.sys.elf_common;
 
 version (DragonFlyBSD):
-extern (C):
-pure:
-nothrow:
+
+extern (C) pure nothrow @nogc @system:
 
 import core.stdc.stdint;
```
Etcetera for the rest of the files in this PR
Most of these could also be set to 'pure' i think, correct ?

joakim-noah: Not sure, haven't messed with `@system` much, will let someone who knows better answer.

joakim-noah:
@ibuclaw, do all the C declarations need to be marked `@system` also?

dkgroot:
@joakim-noah / @ibuclaw FYI: It will build/unittest correctly either way (with and without @system). Would be happy to add @system to them, if that would result in better / more correct code.

Question: Who will backport these types of changes to the other OS implementations, do you guys have a sort of ledger for this. If not, this would come up over and over, but never really get fixed into the other (more important) implementations.

joakim-noah:
[Looking at the `@system` definition, it would seem to be created for this](https://dlang.org/spec/function.html#system-functions), so go ahead and mark all files with C function declarations with it up top, then opt out your own local functions that you know are okay as `@safe`.

src/core/sys/dragonflybsd/dlfcn.d
```diff
@@ -14,8 +14,6 @@ version (DragonFlyBSD):
 
 extern (C) nothrow @nogc @system:
 
-enum __BSD_VISIBLE = true;
-
```
dkgroot: Note, this might look like a massive edit, but:
- it only removed checking for __BSD_VISIBLE / __POSIX_VISIBLE (which are defined in the cdefs.d file) in these subdirectories
- re-indent the lines

dkgroot:
@joakim-noah Done. Can you please check if this is what you intended me to do (revision: 965c1dc)

dkgroot:
Note revision:cf62ee3, might look like a massive edit, but:
- it only removed checking for __BSD_VISIBLE / __POSIX_VISIBLE (which are defined in the cdefs.d file) in these subdirectories
- re-indent the lines

ibuclaw:
If you wanted to add `__BSD_VISIBLE`, shouldn't that go into `core.sys.dragonflybsd.config` anyway?

Seems like FreeBSD is the exception rather than the rule for defining it elsewhere.

dkgroot:
@ibuclaw It is currently defined in core.sys.dragonflybsd.sys.cdefs.d (just like freebsd), but checking for it's setting, especially inside the dragonflybsd subdirectory is a bit superfluous.

dkgroot:
BTW: We should revisit [PR1937](https://github.com/dlang/druntime/pull/1937) regarding using statvfs instead of statfs when it is being finalized, and port the changes to dragonfly implementation (and maybe the other BSD's).
(Using this comment, to link both PR's)

ZombineDev:
@dkgroot There's no such thing as `@system`. I think you mean to change `extern (C)` to `extern (System)`.

dkgroot:
Ping @andralex / @wilzbach Is there still something keeping this PR from being merged ?

src/core/sys/dragonflybsd/sys/elf32.d
```diff
@@ -0,0 +1,187 @@
...
+ * D header file for DragonFlyBSD.
+ *
+ * Authors: Diederik de Groot(port:DragonFlyBSD)
+ * Copied:  From core/sys/freebsd/sys
+ */
+module core.sys.dragonflybsd.sys.elf32;
+
+version (DragonFlyBSD):
+
+extern (C) pure nothrow @nogc @system:
```
joakim-noah: No functions other than the few you define below with different attributes, so all attributes other than `extern (C)` can be removed.

src/core/sys/dragonflybsd/sys/elf64.d
```diff
@@ -0,0 +1,193 @@
...
+ * D header file for DragonFlyBSD.
+ *
+ * Authors: Diederik de Groot(port:DragonFlyBSD)
+ * Copied:  From core/sys/freebsd/sys
+ */
+module core.sys.dragonflybsd.sys.elf64;
+
+version (DragonFlyBSD):
+
+extern (C) pure nothrow @nogc @system:
```
joakim-noah: Same as last module.

src/core/sys/dragonflybsd/sys/elf_common.d
```diff
@@ -0,0 +1,675 @@
...
+ * D header file for DragonFlyBSD.
+ *
+ * Authors: Diederik de Groot(port:DragonFlyBSD)
+ * Copied:  From core/sys/freebsd/sys
+ */
+module core.sys.dragonflybsd.sys.elf_common;
+
+version (DragonFlyBSD):
+
+extern (C) pure nothrow @nogc @system:
```
joakim-noah: Same as last two.

src/core/sys/dragonflybsd/sys/link_elf.d
```diff
@@ -0,0 +1,81 @@
...
+    ElfW!"Phdr"*    dlpi_phdr;
+    ElfW!"Half"     dlpi_phnum;
+    uint64_t        dlpi_adds;
+    uint64_t        dlpi_subs;
+    size_t          dlpi_tls_modid;
+    void*           dlpi_tls_data;
+};
+
+
+private alias extern(C) int function(dl_phdr_info*, size_t, void *) dl_iterate_phdr_cb;
```
joakim-noah: `extern (C)` up top not enough?

src/core/sys/dragonflybsd/sys/mman.d
```diff
@@ -0,0 +1,127 @@
+/**
+ * D header file for DragonFlyBSD
+ *
+ * Authors:   Martin Nowak,Diederik de Groot(port:DragonFlyBSD)
+ * Copied:  From core/sys/freebsd/sys
+ */
+module core.sys.dragonflybsd.sys.mman;
+
+version (DragonFlyBSD):
+extern (C) nothrow @system:
```
joakim-noah: `@nogc`?

src/core/sys/dragonflybsd/time.d
```diff
@@ -0,0 +1,32 @@
...
+ * Authors: Jonathan M Davis,Diederik de Groot(port:DragonFlyBSD)
+ * Copied:  From core/sys/freebsd/sys
+ */
+module core.sys.dragonflybsd.time;
+
+public import core.sys.posix.time;
+
+version(DragonFlyBSD):
+
+extern (C) nothrow @nogc @system:
```
joakim-noah: Only enums, no need for attributes.

src/core/sys/dragonflybsd/sys/link_elf.d
ibuclaw: For aliases, maybe not.

src/core/sys/dragonflybsd/sys/_cpuset.d
```diff
@@ -0,0 +1,30 @@
+/**
+ * D header file for DragonFlyBSD.
+ *
+ * Authors: Martin Nowak, Diederik de Groot(port:DragonFlyBSD)
+ * Copied:  From core/sys/freebsd/sys
+ */
+module core.sys.dragonflybsd.sys._cpuset;
+
+version (DragonFlyBSD):
+extern (C) pure nothrow @nogc @system:
```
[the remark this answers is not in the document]
dkgroot: @joakim-noah Thanks for reviewing this !
They were already defined (e.g freebsd implementation), except for the requested addition of @system.|
Does it have a negative effect to set these attributes ? If could prevent future issues when a function or alias is added.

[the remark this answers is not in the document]
joakim-noah: Our previous modules use these attributes inconsistently, since some of them predate the attributes.  Would rather get this right for this new platform from the start, attributes do affect the mangling of D functions, ie the ABI.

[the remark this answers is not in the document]
dkgroot: Ok will work on removing them. Thanks for the explanation !

src/core/sys/dragonflybsd/sys/link_elf.d
dkgroot: @ibuclaw It seems ok (unittest wise) when i remove the extra extern (C), so going for that.

dkgroot:
Ping @wilzbach @andralex @ibuclaw 
Are there any issues with or objections against this PR in it's current state ?

src/core/sys/dragonflybsd/netinet/in_.d
```diff
@@ -0,0 +1,551 @@
...
+ * Authors: Jonathan M Davis, Diederik de Groot(port:DragonFlyBSD)
+ * Copied:  From core/sys/freebsd/sys
+ */
+module core.sys.dragonflybsd.netinet.in_;
+
+import core.sys.dragonflybsd.sys.cdefs;
+
+public import core.sys.posix.netinet.in_;
+
+version(DragonFlyBSD):
```
wilzbach: Why don't you move this five lines higher?

dkgroot: Done thanks !

dkgroot:
Thanks everyone for your reviews / input / suggestions and comments !
@wilzbach Thanks for accepting this PR

src/core/sys/dragonflybsd/sys/_cpuset.d
```diff
@@ -0,0 +1,30 @@
+/**
+ * D header file for DragonFlyBSD.
+ *
+ * Authors: Martin Nowak, Diederik de Groot(port:DragonFlyBSD)
+ * Copied:  From core/sys/freebsd/sys
+ */
+module core.sys.dragonflybsd.sys._cpuset;
+
+version (DragonFlyBSD):
+extern (C):
```
joakim-noah: Enums are not placed in an object file, so extern(C) makes no sense, think the same for aliases too?

src/core/sys/dragonflybsd/time.d
```diff
@@ -0,0 +1,32 @@
...
+ * Authors: Jonathan M Davis,Diederik de Groot(port:DragonFlyBSD)
+ * Copied:  From core/sys/freebsd/sys
+ */
+module core.sys.dragonflybsd.time;
+
+version(DragonFlyBSD):
+
+public import core.sys.posix.time;
+
+extern (C):
```
joakim-noah: Enums are like preprocessor defines in C/C++, they're pasted in and have no linkage, so this does nothing.

src/core/sys/dragonflybsd/sys/link_elf.d
```diff
@@ -0,0 +1,81 @@
...
+    ElfW!"Phdr"*    dlpi_phdr;
+    ElfW!"Half"     dlpi_phnum;
+    uint64_t        dlpi_adds;
+    uint64_t        dlpi_subs;
+    size_t          dlpi_tls_modid;
+    void*           dlpi_tls_data;
+};
+
+
+private alias int function(dl_phdr_info*, size_t, void *) dl_iterate_phdr_cb;
```
joakim-noah: When I asked about the `extern(C)` here, I honestly didn't know if it was required.  You said the tests didn't need it, but that's understandable [since you don't call the function it's used from below](https://github.com/dlang/druntime/pull/1999/files#diff-f85f6c9c8c517f0652109bd5b27466a5R814).  Thinking about it, I guess since it's a type for an `extern(C)` function, which don't mangle their types anyway, I suppose the `extern(C)` adds nothing.  I tested it on linux by removing this attribute there and the tests ran fine.

If anyone knows why it might be needed, comment here.

dkgroot: Should I remove the second 'extern', ie:
```
extern (C) nothrow @system:
...
extern int dl_iterate_phdr(dl_iterate_phdr_cb __callback, void*__data);
extern int ...
```
should become:
```
extern (C) nothrow @system:
...
int dl_iterate_phdr(dl_iterate_phdr_cb __callback, void*__data);
int ...
```
Is that what you want to tell me ?

src/core/sys/dragonflybsd/sys/_cpuset.d
dkgroot: Remarked out 'extern(C):'

src/core/sys/dragonflybsd/time.d
dkgroot: Remarked out 'extern(C):'

src/core/sys/dragonflybsd/sys/link_elf.d
joakim-noah: No, I was talking about [the `extern(C)` you already removed above](https://github.com/dlang/druntime/pull/2002#discussion_r161364963).  I don't think you need to change it again, but I'm not sure, so I left it open if anybody else wanted to comment here.

src/core/sys/dragonflybsd/sys/link_elf.d
```diff
@@ -75,7 +75,7 @@ struct dl_phdr_info
...
 private alias int function(dl_phdr_info*, size_t, void *) dl_iterate_phdr_cb;
 private alias int function(dl_phdr_info*, size_t, void *) @nogc dl_iterate_phdr_cb_ngc;
-extern int dl_iterate_phdr(dl_iterate_phdr_cb __callback, void*__data);
-extern int dl_iterate_phdr(dl_iterate_phdr_cb_ngc __callback, void*__data) @nogc;
-extern int _rtld_addr_phdr(const void*, dl_phdr_info*) @nogc;
-extern int _rtld_get_stack_prot() @nogc;
+int dl_iterate_phdr(dl_iterate_phdr_cb __callback, void*__data);
+int dl_iterate_phdr(dl_iterate_phdr_cb_ngc __callback, void*__data) @nogc;
+int _rtld_addr_phdr(const void*, dl_phdr_info*) @nogc;
+int _rtld_get_stack_prot() @nogc;
```
joakim-noah: I wasn't referring to these, but it seems to be a good catch: [the `extern` here was meaningless as applied to functions](https://dlang.org/spec/declaration.html#extern)?

src/core/sys/dragonflybsd/sys/link_elf.d
dkgroot: Removing the extern(C) at the top of link_elf.d is also possible, but i would have to add notrown to the extern's at the bottom. So this is almost more of a style issue.

src/core/sys/dragonflybsd/sys/link_elf.d
dkgroot: Moved the extern(C) to the bottom to only pertain to these fout functions.

joakim-noah: No, move it back up, I wasn't talking about the attributes up top.  You need those there.

src/core/sys/dragonflybsd/sys/link_elf.d
joakim-noah: It has nothing to do with style, these attributes normally affect linkage and functionality. However, I think they don't do anything in these cases so it's good you removed them from the bottom.  The one up top is needed though.

ibuclaw:
Don't we check for D style for new patches? I guess not.

src/core/sys/dragonflybsd/sys/link_elf.d
dkgroot: Ok Reverting the last change. Thanks for the update and review !

dkgroot:
@ibuclaw Did i miss something ? Did make some style error ? Will correct it, if you tell me where.

andralex:
Congratulations @dkgroot and many thanks to all reviewers!

dkgroot:
Thanks to all reviewers / Thanks Andrei !

PR #2005, opened 2017-12-20
[CircleCi] Make a BUILD=debug build of DRuntime first

wilzbach:
Yet another boring CI improvement. The idea here is that we use at least one CI to test `BUILD=debug`.
See https://github.com/dlang/dmd/pull/7477 for details.


wilzbach:
> N is only 2?

With CircleCi 1.0, we only had one core available. Since our recent upgrade to 2.0 we now have 2 cores and 4 GB RAM, so I tried setting this to 4. Good observation!

PR #2024, opened 2017-12-30
Fix trace profile timings when the frequency of the timer is unknown.

JohanEngelen:
On non-Windows platforms, QueryPerformanceFrequency is not available and the frequency of RDTSC is unknown (e.g. on my Macbook it apparently is the CPU clock frequency, instead of the previously hardcoded 3_579_545 Hz). So when this function is not available, the time output no longer tries to calculate microseconds (and fail), but shows the timings in megaticks instead.

This change also adds LDC-specific code to read a CPU cycle counter on non-x86 target platforms.
The x86 manual assembly is chosen in favor of the LLVM intrinsic as it performs better with optimizations turned off. The generated machinecode is otherwise identical: https://godbolt.org/g/Le7dZF

Tested on OSX.

### DIFF
--- a/src/rt/trace.d
+++ b/src/rt/trace.d
@@ -1,11 +1,11 @@
 /**
  * Contains support code for code profiling.
  *
- * Copyright: Copyright Digital Mars 1995 - 2015.
+ * Copyright: Copyright Digital Mars 1995 - 2017.
  * License: Distributed under the
  *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
  *    (See accompanying file LICENSE)
- * Authors:   Walter Bright, Sean Kelly
+ * Authors:   Walter Bright, Sean Kelly, the LDC team
  * Source: $(DRUNTIMESRC src/rt/_trace.d)
  */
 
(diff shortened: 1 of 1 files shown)


JohanEngelen:
Fixed the double-printing bug. Should be completely good now.

ZombineDev:
One a first glance the changes LGTM, but I'm wondering if we should replace QPC/QPF timing code with high-level code from `core.time` (`MonoTime.currTime` + `convClockFreq`) for the sake of portability and simplicity.
(`MonoTime.currTime` is implemented as `QueryPerformanceCounter` on Windows and with a fast equivalent on other OSes)

ZombineDev:
Looking at the generated code for `MonoTime.currTime` it looks there a noticeable amount of code bloat. Since `MonoTime.currTime` is meant to be used for time sensitive code like `std.datetime.benchmark`, I'm wondering if we should remove those checks all together or replace them with removable in `-release` mode asserts.

src/rt/trace.d
```diff
@@ -865,14 +866,14 @@ else version (D_InlineAsm_X86_64)
                 ret                     ;
             }
         }
-
-        void QueryPerformanceFrequency(timer_t* freq)
+        else version (LDC)
         {
-            *freq = 3579545;
+            import ldc.intrinsics: llvm_readcyclecounter;
+            *ctr = llvm_readcyclecounter();
```
ZombineDev: Looking at https://godbolt.org/g/Le7dZF, the codegen looks pretty good.

schveiguy:
But those checks are looking at something that is environmentally generated. It's not a programming error, but an OS error if they aren't there. I don't think you can remove the checks. Note that the original code was simply an `assert(0, msg)`. I changed it to abort because the message would be ignored.

What is unfortunate, though, is that once that data is set, it's never unset. So the check itself is wasted after the first one.

Only way around this I can think of is to factor out the check separately to see if everything is OK, and then call another function that avoids the check.

How much bloat are we talking here?

jmdavis:
Personally, I'm not all that invested in whether `assert(0, msg)` or `abort(msg)` is used (though I don't disagree that having the message in release mode is nice), but I'm not all that enthused at the idea of removing the check. If you do, you'll eventually get a division by zero with no good explanation. I think that originally, the check was in a static constructor, which would definitely be better, but we had to remove the static constructor for I don't remember what reason.

Either way, it seems off to me that we're explicitly using `QueryPerformanceCounter` elsewhere in druntime instead of just using `MonoTime`, since that's what it's there for, so if it's decided that we really can't afford the check there, I don't know what we do given that the static constructor solution wasn't acceptable. It's _really_ unlikely that that check will ever fail, so having it gone would rarely matter, but it would really suck when it did matter - though if you're using -release, and `assert(0, msg)` is used, and you don't get a core dump, then I guess that there isn't much difference between a `HLT` instruction and blowing up on division by zero. But IMHO, there's a serious problem if we think that `MonoTime` can't be used for performance reasons.

JohanEngelen:
Can we agree on making smaller steps towards improvement, and not try to fix it all in one PR? This PR basically fixes a current bug and makes the code a little more portable; further work could improve and use `MonoTime`.
Implementation of dmd-style `-profile` is stalled in LDC because of this PR.

schveiguy:
Wish I could review, but I don't really understand the actual changes here, just answering the question from @ZombineDev 

In terms of fixing `MonoTime` so it's more attractive than roll-your-own timing for performance reasons, I'm in favor. I also have no objections to this PR.

JohanEngelen:
https://issues.dlang.org/show_bug.cgi?id=18204

PR #2032, opened 2018-01-05
Fix import from rt.config

belka-ew:
src/rt isn't shipped with druntime's imports, so it can't be
imported directly from core.runtime. This PR fixes importing rt.config
from a public druntime module by replacing it with an extern
definition.

To reproduce the bug create a file that imports 'core.runtime' and
compile it with:
dmd -unittest -deps filename.d

The error is:
/usr/include/dmd/druntime/import/core/runtime.d(653): Error: module
config is in file 'rt/config.d' which cannot be read


rainers:
Thanks. Happens even without -unittest.

Looks like a regression (-deps has changed it's meaning, unfortunately, but having the import in a core module is incorrect anyway) so should be better targeting the stable branch. A bugzilla report should also be added so it appears in the changelog.

belka-ew:
> Looks like a regression (-deps has changed it's meaning, unfortunately, but having the import in a core module is incorrect anyway) so should be better targeting the stable branch. A bugzilla report should also be added so it appears in the changelog.

Should I close this PR then?

It was introduced by https://github.com/dlang/druntime/commit/d757a7d17549776e31dde6feeb7f77aa48019a4e.

rainers:
> Should I close this PR then?

No, it should be switched to target "stable" (click "Edit" for the title) and a bugzilla entry should be added at https://issues.dlang.org/

belka-ew:
It seems I messed it up. My branch is created from master. GitHub web interface allows to change the target branch only, not to rebase on another branch.

belka-ew:
Fine. That looks a bit better. @rainers and @Burgos, thanks.

wilzbach:
FYI: there are two more imports:

https://github.com/dlang/druntime/blob/7bfaa8c0755771b807a8344b8c0bfcf73aec7c82/src/core/thread.d#L713-L718

https://github.com/dlang/druntime/blob/7bfaa8c0755771b807a8344b8c0bfcf73aec7c82/src/core/thread.d#L305-L309

wilzbach:
Also this should have been accompanied by a unittest, s.t. the regression can't happen again.
@belka-ew could you quickly add a test? (see `/test`)

wilzbach:
Also @ other reviewers. This would have never been merged, because for auto-tester new users need to be approved. So if auto-tester doesn't show up in the list of CIs, head over and you will see this screen:


image

@belka-ew I just clicked on "Approve" ;-)

belka-ew:
I'll add a test this week. Can't promise to do it today.
Should I fix other imports in this PR?
What kind of test would it be, I see several directories in "test"?

wilzbach:
> Should I fix other imports in this PR?

If you ask like this, I can't really say no ;-)
But it's not a requirement for this PR.

> What kind of test would it be, I see several directories in "test"?

You probably have to add a new directory to `test`.

belka-ew:
> /home/circleci/dmd/generated/linux/release/64/dmd -m64 -fPIC -w -I../../src -I../../import -Isrc -defaultlib= -debuglib= -dip1000 -L/home/circleci/druntime/generated/linux/debug/64/libdruntime.a -g -debug -ofgenerated/linux/debug/64/fiber_guard_page src/fiber_guard_page.d

There seems to be a very generic problem with druntime tests. `-I../../src` is always passed when compiling. ../../src contains both, `core` and `rt`, it prevents such errors from being found.

It is just a note, I still need some time to experiment with the test runner.

belka-ew:
I couldn't find a better way to write the test. Is it ok? If so, I'll try to fix other two imports.

test/imports/src/core
```diff
@@ -0,0 +1 @@
+../../../src/core
```
wilzbach: Hmm, why do you need these files?
`dmd` isn't used with `-conf=` and will use the import path of druntime. If you don't trust it, you can always add `-I../../import`?

belka-ew: It fails: https://circleci.com/gh/dlang/druntime/1546
> Testing bug18193
> /home/circleci/dmd/generated/linux/release/64/dmd -o- -deps=generated/linux/debug/64/bug18193.done -Isrc -I../../import src/bug18193
> Error: cannot find source code for runtime library file 'object.d'
>       dmd might not be correctly installed. Run 'dmd -man' for installation instructions.
>       config file: /home/circleci/dmd/generated/linux/release/64/dmd.conf

belka-ew:
Fix for the second part:
https://github.com/belka-ew/druntime/compare/fix-deps-rt-import...belka-ew:fix-deps-rt-import2

belka-ew:
The fix for the second part:
https://github.com/belka-ew/druntime/compare/fix-deps-rt-import...belka-ew:fix-deps-rt-import2

rainers:
The problem seems to be that the tests are run without the import files being copied over from the source directory. That's why the other tests specify -I../../src, too.
Not sure if it is specific to the circleci script, but the default `target` should be built before running the tests.

belka-ew:
Any chance to get it merged with or without the tests before 2.080.1? As said it is a general problem with the makefiles, which prevents such bugs from being found. And this problem is not related to this regression introduced in 2.080.0 and can be fixed in another PR. I can create an issue if needed.

wilzbach:
I added 66222b6 to your PR. I think that should solve the problem. Sorry for the delay.

wilzbach:
Looks all good. @belka-ew do you mind if I squash your last three commits into one? No need to keep the experiments in the history.

belka-ew:
@wilzbach, thanks for your fast response. I squashed the commits and added the fixes for `core.thread`.

belka-ew:
I suppose that the tests fail because code.dlang.org is dead?

wilzbach:
No the DUB registry outage was only for two hours and we have mirrors:

1) Jenkins failure
---------------------

```
Sending interrupt signal to process

./travis-ci.sh: line 40: 29145 Terminated              dub build --compiler=$DC --override-config=vibe-d:core/$VIBED_DRIVER $DUB_ARGS

script returned exit code 143
```

Might be related to: https://github.com/dlang/ci/issues/118

2) DAutoTest failure
---------------------------

```
.generated/stable_dmd-2.072.2/dmd2/linux/bin64/dub build --single --compiler=.generated/stable_dmd-2.072.2/dmd2/linux/bin64/dmd assert_writeln_magic.d
Root package assert_writeln_magic references unknown package libdparse
posix.mak:818: recipe for target '.generated/assert_writeln_magic' failed
```

2.072.0 doesn't come with mirrors, will look into upgrading it.

In the mean time I will `git commit --amend` the last commit and force-push as that's the only way to restart DAutoTest.

wilzbach:
> 2.072.0 doesn't come with mirrors, will look into upgrading it.

I already did so, it's just not in `stable` yet:

https://github.com/dlang/dlang.org/pull/1926

PR #2035, opened 2018-01-10
Fix Issue 18220 - Allow `rt_trapexceptions` to be set from the CLI

wilzbach:
See the changelog or http://arsdnet.net/this-week-in-d/2016-aug-07.html for a motivation.

> Hopefully, druntime will expose this variable in a more easier way in the future. Currently, on Windows, it is automatically set if a debugger is detected, but on Linux, debugger detection is quite a bit harder and hackier, so it might not happen any time soon. Druntime could (and should!) also potentially offer it as a command line switch, like it does with GC profiling now.

> DRUNTIME DEVS: if you read this, make it happen!

CC @adamdruppe

### DIFF
--- a/changelog/exceptions-opt.dd
+++ b/changelog/exceptions-opt.dd
@@ -1,0 +1,35 @@
+Exception trapping can now be disabled via `--DRT-trapExceptions=0`
+
+Previously it was only possible to disable the trapping of exception by setting
+the global variable `rt_trapExceptions` to `false`.
+Now you can, for example, immediately open `gdb` at the uncaught exception:
+
+$(CONSOLE
+> gdb -ex run --args <my-program> --DRT-trapExceptions=0
+[Thread debugging using libthread_db enabled]
+Using host libthread_db library "/usr/lib/libthread_db.so.1".
+uncaught exception
+object.Exception@src/rt_trap_exceptions_drt.d(4): foo
+$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)
+src/rt_trap_exceptions_drt.d:4 void rt_trap_exceptions_drt.test() [0x55591026]
+src/rt_trap_exceptions_drt.d:9 _Dmain [0x55591058]
+$(P)
+Program received signal SIGABRT, Aborted.
+0x00007ffff6e7b86b in raise () from /usr/lib/libc.so.6
+(gdb) bt full
+#0  0x00007ffff6e7b86b in raise () from /usr/lib/libc.so.6
+No symbol table info available.
+#1  0x00007ffff6e6640e in abort () from /usr/lib/libc.so.6
+No symbol table info available.
+#2  0x00005555555918cc in _d_throwdwarf (o=0x7ffff7ea4000) at src/rt/dwarfeh.d:233
+        eh = 0x7ffff7fa4740
+        refcount = 0
+        r = 5
+#3  0x0000555555591027 in rt_trap_exceptions_drt.test() () at ../../src/object.d:2695
+        innerLocal = 20
+#4  0x0000555555591059 in D main (args=...) at src/rt_trap_exceptions_drt.d:9
+        myLocal = "bar"
+)
+
+$(LINK2 http://arsdnet.net/this-week-in-d/2016-aug-07.html, This Week in D) for
+an in-depth explanation of `rt_trapExceptions`
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -452,17 +452,17 @@ extern (C) int _d_run_main(int argc, char **argv, MainFunc mainFunc)
         args = argsCopy[0..j];
     }
 
-    bool trapExceptions = rt_trapExceptions;
+    auto useExceptionTrap = parseExceptionOptions();
 
     version (Windows)
     {
         if (IsDebuggerPresent())
-            trapExceptions = false;
+            useExceptionTrap = false;
     }
 
     void tryExec(scope void delegate() dg)
     {
-        if (trapExceptions)
+        if (useExceptionTrap)
         {
             try
             {
@@ -556,6 +556,18 @@ private void formatThrowable(Throwable t, scope void delegate(in char[] s) nothr
         }
         sink("=== ~Bypassed ===\n");
     }
+}
+
+private auto parseExceptionOptions()
+{
+    import rt.config : rt_configOption;
+    import core.internal.parseoptions : rt_parseOption;
+    const optName = "trapExceptions";
+    auto option = rt_configOption(optName);
+    auto trap = rt_trapExceptions;
+    if (option.length)
+        rt_parseOption(optName, option, trap, "");
+    return trap;
 }
 
 extern (C) void _d_print_throwable(Throwable t)
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -31,10 +31,11 @@ download() {
 }
 
 install_deps() {
+    sudo apt-get update
     if [ $MODEL -eq 32 ]; then
-        sudo apt-get update
-        sudo apt-get install g++-multilib
+        sudo apt-get install -y g++-multilib
     fi
+    sudo apt-get install -y gdb
 
     download "https://dlang.org/install.sh" "https://nightlies.dlang.org/install.sh" "install.sh"
 
--- a/src/core/internal/parseoptions.d
+++ b/src/core/internal/parseoptions.d
@@ -110,6 +110,28 @@ bool parseOptions(CFG)(ref CFG cfg, string opt)
     return true;
 }
 
+/**
+Parses an individual option `optname` value from a provided string `str`.
+The option type is given by the type `T` of the field `res` to which the parsed
+value will be written too.
+In case of an error, `errName` will be used to display an error message and
+the failure of the parsing will be indicated by a falsy return value.
+
+For boolean values, '0/n/N' (false) or '1/y/Y' (true) are accepted.
+
+Params:
+    optname = name of the option to parse
+    str = raw string to parse the option value from
+    res = reference to the resulting data field that the option should be parsed too
+    errName = full-text name of the option which should be displayed in case of errors
+
+Returns: `false` if a parsing error happened.
+*/
+bool rt_parseOption(T)(const(char)[] optname, ref inout(char)[] str, ref T res, const(char)[] errName)
+{
+    return parse(optname, str, res, errName);
+}
+
 private:
 
 bool optError(in char[] msg, in char[] name, const(char)[] errName)
--- a/src/rt/dwarfeh.d
+++ b/src/rt/dwarfeh.d
@@ -224,7 +224,7 @@ extern(C) void _d_throwdwarf(Throwable o)
         case _URC_END_OF_STACK:
             /* Unwound the stack without encountering a catch clause.
              * In C++, this would mean call uncaught_exception().
-             * In D, this can happen only if `rt_trapException` is cleared
+             * In D, this can happen only if `rt_trapExceptions` is cleared
              * since otherwise everything is enclosed by a top-level
              * try/catch.
              */
--- a/test/exceptions/Makefile
+++ b/test/exceptions/Makefile
@@ -1,26 +1,30 @@
 include ../common.mak
 
-TESTS:=stderr_msg unittest_assert invalid_memory_operation unknown_gc static_dtor future_message
+TESTS=stderr_msg unittest_assert invalid_memory_operation unknown_gc static_dtor future_message rt_trap_exceptions_drt
 
 ifeq ($(OS)-$(BUILD),linux-debug)
-	TESTS:=$(TESTS) line_trace rt_trap_exceptions
+	TESTS+=line_trace rt_trap_exceptions
 	LINE_TRACE_DFLAGS:=-L--export-dynamic
 endif
+ifeq ($(OS),linux)
+	TESTS+=rt_trap_exceptions_drt_gdb
+endif
 ifeq ($(OS)-$(BUILD),freebsd-debug)
-	TESTS:=$(TESTS) line_trace
+	TESTS+=line_trace
 	LINE_TRACE_DFLAGS:=-L--export-dynamic
 endif
 ifeq ($(OS)-$(BUILD),dragonflybsd-debug)
-	TESTS:=$(TESTS) line_trace
+	TESTS+=line_trace
 	LINE_TRACE_DFLAGS:=-L--export-dynamic
 endif
 ifeq ($(OS)-$(BUILD),osx-debug)
-	TESTS:=$(TESTS) line_trace
+	TESTS+=line_trace
 	LINE_TRACE_DFLAGS:=
 endif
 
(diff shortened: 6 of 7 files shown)


wilzbach:
Next on my list is to create a page on dlang.org that documents all the various options that are supported by `--DRT`

For now, this is the best page: https://dlang.org/spec/garbage.html#gc_config

adamdruppe:
yesssssss

changelog/exceptions-opt.dd
```diff
@@ -0,0 +1,26 @@
+Exception trapping can now be disabled via `--DRT-exopt=trap:0`
+
+Previously it was only possible to disable the trapping of exception by setting
+the global variable `rt_trapExceptions` to `false`.
+For example, you can now immediately open `gdc` at the uncaught exception:
```
nemanja-boric-sociomantic: Typo `gdc` ?

changelog/exceptions-opt.dd
```diff
@@ -0,0 +1,26 @@
...
+(gdb) bt
+#0  0x00005555555a65e9 in _D4core8demangle__T8DemangleTSQBcQBa7NoHooksZQBa__T10doDemangleSQCkQCi__TQCcTQBwZQCk16parseMangledNameZQCaMFNaNbNfZAa ()
+#1  0x000055555559b194 in rt.backtrace.dwarf.traceHandlerOpApplyImpl(const(void*[]), scope int(ref ulong, ref const(char[])) delegate) ()
+#2  0x0000555555590349 in core.runtime.defaultTraceHandler(void*).DefaultTraceInfo.opApply(scope int(ref ulong, ref const(char[])) delegate) const ()
+#3  0x00005555555902e2 in core.runtime.defaultTraceHandler(void*).DefaultTraceInfo.opApply(scope int(ref const(char[])) delegate) const ()
+#4  0x000055555558797c in object.Throwable.toString(scope void(const(char[])) delegate) const ()
+#5  0x0000555555589000 in rt.dmain2.formatThrowable(object.Throwable, scope void(const(char[])) nothrow delegate) ()
+#6  0x0000555555585650 in _d_print_throwable ()
+#7  0x00005555555857e9 in _d_throwdwarf ()
+#8  0x00005555555850bc in D main (args=...) at rt_opt.d:3
```
nemanja-boric-sociomantic: Something doesn't feel right with this stack trace :thinking: . This actually looks like a bug in runtime - I don't think this signal should be raised while trying to print the exception, the program should actually raise `SIGABRT`: https://github.com/dlang/druntime/blob/master/src/rt/dwarfeh.d#L232

changelog/exceptions-opt.dd
wilzbach: Oh. Yeah it should have been `gdb`. Fixed. Thanks!

changelog/exceptions-opt.dd
wilzbach: I opened an issue for this, s.t. it's not forgotten: https://issues.dlang.org/show_bug.cgi?id=18226 and we can investigate this separately.

src/rt/dmain2.d
```diff
@@ -550,6 +550,36 @@ private void formatThrowable(Throwable t, scope void delegate(in char[] s) nothr
...
+            trap = rt_trapExceptions;
+            import core.internal.parseoptions : initConfigOptions;
+            return initConfigOptions(this, this.errorName);
+        }
+
+        void help()
+        {
+            string s = "Exception options are specified as whitespace separated assignments:
+    trap:0|1      - 0 disables trapping of exceptions, 1 exceptions will be trapped (default: %d)
+";
```
rainers: This looks a bit excessive for a single option. Maybe just use rt_configOption? Or are you anticipating more exception options?

rainers:
BTW: just (re)discovered this PR: https://github.com/dlang/druntime/pull/1538

adamdruppe:
And another stack overflow question that could be trivially answered if this was merged just flew by.

What is the bloody hold up!?

WebFreak001:
yeah please make this happen, I'm sick of always pulling up that hack, I could basically already set that this week in D post as homepage in my browser

src/rt/dmain2.d
wilzbach: Ok changed to `rt_configOption`. The downside is that now it can't re-use the existing `parse` code, but OTOH the CLI is shorter and easier.

wilzbach:
Rebased this to master, addressed all comments and hopefully this should be now good to go.
Any last objections?

src/rt/dmain2.d
```diff
@@ -558,6 +558,35 @@ private void formatThrowable(Throwable t, scope void delegate(in char[] s) nothr
...
+{
+    static struct ExceptionOptions
+    {
+        bool trap;
+
+        @nogc nothrow void initialize()
+        {
+            import rt.config : rt_configOption;
+            trap = rt_trapExceptions;
+            auto res = rt_configOption("extrap");
```
wilzbach: `exceptionTrap` might be an alternative. Which one do you like better?

src/rt/dmain2.d
```diff
@@ -558,6 +558,35 @@ private void formatThrowable(Throwable t, scope void delegate(in char[] s) nothr
...
+            switch(res)
+            {
+                case "1", "y", "Y", "on", "true":
```
wilzbach: `core.internal.parseoptions` only handles `"1", "y", "Y"`, but I think "on" and "true" are too common to ignore.

rainers: Maybe just add `on`and `true` there, too? Please also note that parse only checks the first character, so it allows `yes`, too.

src/rt/dmain2.d
rainers: I'd prefer `trapException`? I think it's better to be a bit more verbose for clarity.

src/rt/dmain2.d
```diff
@@ -558,6 +558,35 @@ private void formatThrowable(Throwable t, scope void delegate(in char[] s) nothr
...
+                    break;
+                case "0", "n", "N", "off", "false":
+                    trap = true;
+                    break;
+                default:
+                    break;
+            }
+        }
+    }
+    ExceptionOptions options;
```
rainers: Still looks a little verbose to use a struct for a single bool.

changelog/exceptions-opt.dd
```diff
@@ -0,0 +1,26 @@
+Exception trapping can now be disabled via `--DRT-extrap=1`
```
Geod24: That name seems a bit contradictory to the semantic. Doesn't `--DRT-extrap=0` make more sense ?

changelog/exceptions-opt.dd
Geod24: I'd say it would be better to fix the bug than release a buggy feature.

jacob-carlborg:
This might be a useful feature, but I'm pretty sure it's possible to detect if the program is running inside a debugger and exception trapping can then automatically be turned off. Perhaps we can have both.

adamdruppe:
On Tue, Jun 26, 2018 at 06:09:59AM -0700, Jacob Carlborg wrote:
> This might be a useful feature, but I'm pretty sure it's possible to detect if the program is running inside a debugger and exception trapping can then automatically be turned off. Perhaps we can have both.

Yes, that would be ideal, but it is extremely difficult to get that
detection reliable on linux. Hence the command line arg gives the user
control to handle the ambiguities.

It is already implemented in druntime for Windows though.

wilzbach:
> Yes, that would be ideal, but it is extremely difficult to get that
detection reliable on linux. Hence the command line arg gives the user
control to handle the ambiguities.

It doesn't look [too difficult/impossible](https://stackoverflow.com/questions/3596781/how-to-detect-if-the-current-process-is-being-run-by-gdb).

Maybe we can add a best-effort auto-detection for at least for `gdb`?

(in case it fails the manual CLI/environment can always be used, like e.g. [coloring works for `git`](https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_colors_in_git)

src/rt/dmain2.d
schveiguy: Yeah, see prior work: https://github.com/dlang/druntime/blob/8c86106a56f752455d387ce9d835b95fe7192e12/src/rt/lifetime.d#L48

Also, you have both setting `trap` to `true`!

src/rt/dmain2.d
schveiguy: > but I think "on" and "true" are too common to ignore.

I think this is too clever for the value it adds. Note that "On" and "True" will be ignored as well. If you want to prevent mistakes, throw when the option is not recognized. e.g.: https://github.com/dlang/druntime/blob/522787e95062398ad96dbf35e3983c10dd370c9a/src/rt/minfo.d#L178

schveiguy:
> I don't remember the PR number

Found it: #1811

jacob-carlborg:
> I tried implementing that before and it got shot down in the PR review
because it is actually an unreliable hack and deemed expensive at
startup time - extra syscalls.

Would it be possible to always trap exceptions, then when trapping an exception detect if the program is running inside a debugger, if it is rethrow the exception. If not running inside a debugger, it would behave as it does today?

wilzbach:
>  Or will the stacktrace get messed up?

Yeah I just tried this and it looks like the stack trace gets messed up :/
Vanilla `rt_trap_exceptions_drt_gdb` today:

```
0x00007ffff6e7b86b in raise () from /usr/lib/libc.so.6
#0  0x00007ffff6e7b86b in raise () from /usr/lib/libc.so.6
No symbol table info available.
#1  0x00007ffff6e6640e in abort () from /usr/lib/libc.so.6
No symbol table info available.
#2  0x000055555558772a in _d_throwdwarf ()
No symbol table info available.
#3  0x0000555555586ec9 in D main (args=...) at ../../src/object.d:2695
        myLocal = 0x5555555874a8 <rt.dmain2._d_run_main(int, char**, extern(C) int(char[][]) function).runAll().__lambda1()+40> "H\213M\370\211\201\330\377\377\377H\213\345", <incomplete sequence \303>
```

Rethrown
-----------

```
0x00007ffff6e7b86b in raise () from /usr/lib/libc.so.6
#0  0x00007ffff6e7b86b in raise () from /usr/lib/libc.so.6
No symbol table info available.
#1  0x00007ffff6e6640e in abort () from /usr/lib/libc.so.6
No symbol table info available.
#2  0x0000555555587726 in _d_throwdwarf ()
No symbol table info available.
#3  0x000055555558734f in rt.dmain2._d_run_main(int, char**, extern(C) int(char[][]) function).tryExec(scope void() delegate)
    ()
No symbol table info available.
#4  0x0000555555587172 in _d_run_main ()
No symbol table info available.
#5  0x0000555555586ee0 in main ()
No symbol table info available.
#6  0x00007ffff6e6806b in __libc_start_main () from /usr/lib/libc.so.6
No symbol table info available.
#7  0x0000555555586d9a in _start ()
No symbol table info available.
```

jacob-carlborg:
> Yeah I just tried this and it looks like the stack trace gets messed up :/

That's disappointing. It would be nice to be able to rethrow exceptions without changing the stacktrace. Anyway, let's continue without autodetecting the debugger.

wilzbach:
> Anyway, let's continue without autodetecting the debugger.

Done (removed auto-detecting logic).

changelog/exceptions-opt.dd
```diff
@@ -0,0 +1,35 @@
...
+(gdb) bt full
+#0  0x00007ffff6e7b86b in raise () from /usr/lib/libc.so.6
+No symbol table info available.
+#1  0x00007ffff6e6640e in abort () from /usr/lib/libc.so.6
+No symbol table info available.
+#2  0x00005555555918cc in _d_throwdwarf (o=0x7ffff7ea4000) at src/rt/dwarfeh.d:233
+        eh = 0x7ffff7fa4740
+        refcount = 0
+        r = 5
+#3  0x0000555555591027 in rt_trap_exceptions_drt.test() () at ../../src/object.d:2695
```
wilzbach: I don't know why line trace points to `Exception.this` here, but that's the existing behavior.

src/core/demangle.d
```diff
@@ -1990,7 +1991,11 @@ pure @safe:
             }
             catch( Exception e )
             {
-                assert( false ); // no other exceptions thrown
+                //assert( false ); // no other exceptions thrown
+                // TODO: ugly hack around mangling error
+                dst = null;
+                dst ~= prevBuf;
+                return dst;
```
wilzbach: I still hate this, but I haven't found a better workaround so far (and this is better than just crashing ...).

rainers: Could you elaborate what exception is thrown that might be handled here?

rainers: It seems the ParseException isn't caught correctly inside a when demangling "_Dmain". As exceptions are also used for back tracking inside the demangler for valid symbols, it would be nice to actually find the root cause. Maybe _d_throwdwarf doesn't support recursive calls?

rainers: I think we don't have to block merging this PR on finding the root cause, but you should use the same code as for the ParseException to avoid calling into the GC most of the time, or just catch Exception there instead and add an appropriate comment.

src/rt/dmain2.d
```diff
@@ -558,6 +558,31 @@ private void formatThrowable(Throwable t, scope void delegate(in char[] s) nothr
...
+        case "":
+            break;
+        default:
+            fprintf(stderr, "Invalid value '%.*s' for option '%s'.\n",
+                    res.length, res.ptr,
+                    optName.ptr);
+            break;
+    }
+    return trap;
+}
```
rainers: Still quite verbose. How about exposing `parse` in core.internal.parseoptions via:
```
bool rt_parseOption(T...)(T args) { return parse(args); }
```
and then using
```
private auto parseExceptionOptions()
{
    import rt.config : rt_configOption;
    import core.internal.parseoptions : rt_parseOption;
    const optName = "trapExceptions";
    auto res = rt_configOption(optName);
    auto trap = rt_trapExceptions;
    if (res.length)
        rt_parseOption(optName, res, trap, "");
    return trap;
}
```

src/core/demangle.d
wilzbach: FYI I think @ibuclaw managed to find the root cause: https://github.com/dlang/druntime/pull/2245
I removed the mangling workaround and cherry-picked his commit to see whether the CIs agree.

jacob-carlborg:
Are we good to go with this?

wilzbach:
> Are we good to go with this?

I think so, but probably a good idea to merge https://github.com/dlang/druntime/pull/2245 first though as this only contains the cherry-picked commit from #2245, but not its test cases.

wilzbach:
Rebased. The first commit is independent to the second commit of this PR. I wanted to keep the "install gdb" separate as well for the same reason.

jacob-carlborg:
Is there a reason to have https://github.com/dlang/druntime/commit/d98d8192bf574b87ce469d6625327b11ffe15a5f as a separate commit?

wilzbach:
> Is there a reason to have d98d819 as a separate commit?

It's more or less unrelated to this PR (it just fixes a typo and removes a redundant declaration). 
Moved the commit to https://github.com/dlang/druntime/pull/2251

jacob-carlborg:
Is this feature documented somewhere except the changelog?

PR #2037, opened 2018-01-11
Add __delete for migration from delete

wilzbach:
See also:
- https://github.com/dlang/dmd/pull/7342
- https://dlang.org/deprecate.html#delete
- https://dlang.org/spec/expression.html#delete_expressions

I didn't add any tests for [Class deallocators](https://dlang.org/spec/class.html#deallocators)
as they have been deprecated since D2 and we call the existing functions.

CC @JinShil

### DIFF
--- a/changelog/core-memory-__delete.dd
+++ b/changelog/core-memory-__delete.dd
@@ -1,0 +1,35 @@
+`core.memory.__delete` has been added
+
+$(REF __delete, core, memory) allows easy migration from the deprecated `delete`.
+`__delete` behaves exactly like `delete`:
+
+---
+bool dtorCalled;
+class B
+{
+    int test;
+    ~this()
+    {
+        dtorCalled = true;
+    }
+}
+B b = new B();
+B a = b;
+b.test = 10;
+
+__delete(b);
+assert(b is null);
+assert(dtorCalled);
+// but be careful, a still points to it
+assert(a !is null);
+---
+
+For example, on a Posix platform you can simply run:
+
+$(CONSOLE
+sed "s/delete \(.*\);/__delete(\1);/" -i **/*.d
+)
+
+Users should prefer $(REF destroy, object)` to explicitly finalize objects,
+and only resort to $(REF __delete, core,memory) when $(REF destroy, object)
+would not be a feasible option.
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -919,3 +919,236 @@ extern (C) private pure @system @nogc nothrow
 
     pragma(mangle, "free") void fakePureFree(void* ptr);
 }
+
+extern(C) private @system nothrow @nogc
+{
+    pragma(mangle, "_d_delinterface") void _d_delinterface(void**);
+    pragma(mangle, "_d_delclass") void _d_delclass(Object*);
+    pragma(mangle, "_d_delstruct") void _d_delstruct(void**, TypeInfo_Struct);
+    pragma(mangle, "_d_delmemory") void _d_delmemory(void**);
+    pragma(mangle, "_d_delarray_t") void _d_delarray_t(void**, TypeInfo_Struct);
+}
+
+/**
+Destroys and then deallocates an object.
+
+In detail, `__delete(x)` returns with no effect if `x` is `null`. Otherwise, it
+performs the following actions in sequence:
+$(UL
+    $(LI
+        Calls the destructor `~this()` for the object referred to by `x`
+        (if `x` is a class or interface reference) or
+        for the object pointed to by `x` (if `x` is a pointer to a `struct`).
+        Arrays of structs call the destructor, if defined, for each element in the array.
+        If no destructor is defined, this step has no effect.
+    )
+    $(LI
+        Frees the memory allocated for `x`. If `x` is a reference to a class
+        or interface, the memory allocated for the underlying instance is freed. If `x` is
+        a pointer, the memory allocated for the pointed-to object is freed. If `x` is a
+        built-in array, the memory allocated for the array is freed.
+        If `x` does not refer to memory previously allocated with `new` (or the lower-level
+        equivalents in the GC API), the behavior is undefined.
+    )
+    $(LI
+        Lastly, `x` is set to `null`. Any attempt to read or write the freed memory via
+        other references will result in undefined behavior.
+    )
+)
+
+Note: Users should prefer $(REF destroy, object)` to explicitly finalize objects,
+and only resort to $(REF __delete, core,memory) when $(REF destroy, object)
+wouldn't be a feasible option.
+
+Params:
+    x = aggregate object that should be destroyed
+
+See_Also: $(REF destroy, object), $(REF free, core,GC)
+
+History:
+
+The `delete` keyword allowed to free GC-allocated memory.
+As this is inherently not `@safe`, it has been deprecated.
+This function has been added to provide an easy transition from `delete`.
+It performs the same functionality as the former `delete` keyword.
+*/
+void __delete(T)(ref T x) @system
+{
+    static void _destructRecurse(S)(ref S s)
+    if (is(S == struct))
+    {
+        static if (__traits(hasMember, S, "__xdtor") &&
+                   // Bugzilla 14746: Check that it's the exact member of S.
+                   __traits(isSame, S, __traits(parent, s.__xdtor)))
+            s.__xdtor();
+    }
+
+    // See also: https://github.com/dlang/dmd/blob/v2.078.0/src/dmd/e2ir.d#L3886
+    static if (is(T == interface))
+    {
+        .object.destroy(x);
+    }
+    else static if (is(T == class))
+    {
+        .object.destroy(x);
+    }
+    else static if (is(T == U*, U))
+    {
+        static if (is(U == struct))
+            _destructRecurse(*x);
+    }
+    else static if (is(T : E[], E))
+    {
+        static if (is(E == struct))
+        {
+            foreach (ref e; x)
+                _destructRecurse(e);
+        }
+    }
+    else
+    {
+        static assert(0, "It is not possible to delete: `" ~ T.stringof ~ "`");
+    }
+
+    static if (is(T == interface) || is(T == class))
+    {
+        GC.free(cast(void*) x);
+        x = null;
+    }
+    else static if (is(T == U2*, U2) || is(T : E2[], E2))
+    {
+        GC.free(&x);
+        x = null;
+    }
+}
+
+/// Deleting classes
+unittest
+{
+    bool dtorCalled;
+    class B
+    {
+        int test;
+        ~this()
+        {
+            dtorCalled = true;
+        }
+    }
+    B b = new B();
+    B a = b;
+    b.test = 10;
+
+    assert(GC.addrOf(cast(void*) b) != null);
+    __delete(b);
+    assert(b is null);
+    assert(dtorCalled);
+    assert(GC.addrOf(cast(void*) b) == null);
+    // but be careful, a still points to it
+    assert(a !is null);
+    assert(GC.addrOf(cast(void*) a) !is null);
+}
+
+/// Deleting interfaces
+unittest
+{
+    bool dtorCalled;
+    interface A
+    {
+        int quack();
+    }
+    class B : A
+    {
+        int a;
+        int quack()
+        {
+            a++;
+            return a;
+        }
+        ~this()
+        {
+            dtorCalled = true;
+        }
+    }
+    A a = new B();
+    a.quack();
+
+    assert(GC.addrOf(cast(void*) a) != null);
+    __delete(a);
+    assert(a is null);
+    assert(dtorCalled);
+    assert(GC.addrOf(cast(void*) a) == null);
+}
+
+/// Deleting structs
+unittest
+{
+    bool dtorCalled;
+    struct A
+    {
+        string test;
+        ~this()
+        {
+            dtorCalled = true;
+        }
+    }
+    auto a = new A("foo");
+
+    assert(GC.addrOf(cast(void*) a) != null);
+    __delete(a);
+    assert(a is null);
+    assert(dtorCalled);
+    assert(GC.addrOf(cast(void*) a) == null);
+}
+
+/// Deleting arrays
+unittest
+{
+    int[] a = [1, 2, 3];
+    auto b = a;
+
+    assert(GC.addrOf(b.ptr) != null);
+    __delete(b);
+    assert(b is null);
+    assert(GC.addrOf(b.ptr) == null);
+    // but be careful, a still points to it
+    assert(a !is null);
+    assert(GC.addrOf(a.ptr) !is null);
+}
+
+/// Deleting arrays of structs
+unittest
+{
+    int dtorCalled;
+    struct A
+    {
+        int a;
+        ~this()
+        {
+            dtorCalled++;
+        }
+    }
+    auto arr = [A(1), A(2), A(3)];
+
+    assert(GC.addrOf(arr.ptr) != null);
+    __delete(arr);
+    assert(dtorCalled == 3);
+    assert(GC.addrOf(arr.ptr) == null);
+}
+
+// Deleting raw memory
+unittest
+{
+    import core.memory : GC;
+    auto a = GC.malloc(5);
+    assert(GC.addrOf(cast(void*) a) != null);
+    __delete(a);
+    assert(a is null);
+    assert(GC.addrOf(cast(void*) a) == null);
+}
+
+// __delete returns with no effect if x is null
+unittest
+{
+    Object x = null;
+    __delete(x);
+}


changelog/core-memory-__delete.dd
```diff
@@ -0,0 +1,25 @@
...
+__delete(b);
+assert(b is null);
+assert(s == "exists");
+---
+
+For example, one a Posix platform you can simply run:
+
+$(CONSOLE
+sed "s/delete \(.*\);/__delete(\1);/" -i **/*.d
+)
```
JinShil: I suggest adding something to the effect that this is intended to only be used as a last resort when `destroy` is not an option.  At least that's my understanding:  We want to migrate users to `destroy`.

src/core/memory.d
```diff
@@ -919,3 +919,184 @@ extern (C) private pure @system @nogc nothrow
...
+private extern (C) void _d_delclass(Object* p);
+private extern (C) void _d_delstruct(void** p, TypeInfo_Struct inf);
+
+/**
+The `delete` keyword allowed to free GC-allocated memory.
+As this is inherently not `@safe`, it has been deprecated.
+This function has been added to provide an easy transition from `delete`.
+It performs the same functionality as `delete`.
+
+$(REF destroy, object) is `@safe` and should be used if possible.
```
JinShil: Again, I think this should have some information explaining that users should first look to `destroy` and only to `__delete` when `destroy` is not an option.

src/foo.d
```diff
@@ -0,0 +1,151 @@
+import std.stdio;
```
JinShil: What is this file (src/foo.d) doing in this commit?

changelog/core-memory-__delete.dd
JinShil: Potential verbage:  "Users should prefer `object.destroy` to explicitly finalize objects, and only resort to `__delete` when `object.destroy` would not be a feasible option."

ibuclaw:
Looks like you committed an extra file by mistake

ibuclaw:
What advantages does this have over the (poorly named) destroy?

JinShil:
> What advantages does this have over the (poorly named) destroy?

My understanding is it was mandated by @andralex.  See discussion starting [here](https://github.com/dlang/dmd/pull/7342#issuecomment-350939393)

ibuclaw:
Still, my understanding was that (years ago when delete was first put for deprecation) destroy was added as a migration path - by Andrei himself I believe.

Obviously that function has failed spectacularly if this is needed. 🙄

wilzbach:
> Still, my understanding was that (years ago when delete was first put for deprecation) destroy was added as a migration path - by Andrei himself I believe.
> Obviously that function has failed spectacularly if this is needed. :roll_eyes:

A bit of history I could dug up, but it seems that `destroy` has been there for a long time:

https://github.com/dlang/druntime/pull/231
https://github.com/dlang/druntime/pull/7

src/core/memory.d
```diff
@@ -919,3 +919,186 @@ extern (C) private pure @system @nogc nothrow
...
+        memory allocated for the class instance or struct instance.
+        If the garbage collector was not used to allocate the memory for
+        the instance, undefined behavior will result.
+    )
+    $(LI If `t` is a pointer or a dynamic array,
+        the garbage collector is called to immediately release the
+        memory.
+        If the garbage collector was not used to allocate the memory for
+        the instance, undefined behavior will result.
+    )
```
JinShil: The implementation just sets the pointer to null.  I don't think that initiates an immediate call to GC to free the memory.

src/core/memory.d
```diff
@@ -919,3 +919,186 @@ extern (C) private pure @system @nogc nothrow
...
+{
+    class A
+    {
+        string test;
+    }
+    A a = new A();
+    a.test = "exists";
+    auto b = &a;
+    auto c = &b;
+    delete c;
```
JinShil: Shouldn't this be `__delete`?

src/core/memory.d
```diff
@@ -919,3 +919,186 @@ extern (C) private pure @system @nogc nothrow
...
+}
+
+/// Deleting arrays
+unittest
+{
+    int[] a = [1, 2, 3];
+    auto b = a;
+    __delete(b);
+    assert(b is null);
+    assert(a == [1, 2, 3]);
```
JinShil: If I'm understanding correctly, you are access through `a` after `b` was `__delete`ed.  According to the documentation comments, that's undefined behavior.

src/core/memory.d
```diff
@@ -919,3 +919,186 @@ extern (C) private pure @system @nogc nothrow
...
+    {
+        t = null;
+    }
+    else static if (is(T : E[], E))
+    {
+        t = null;
+    }
+    else
+    {
+        static assert(0, "It is not possible to delete: " ~ T.stringof);
```
JinShil: "It is not possible to delete\`" ~ T.stringof ~ "\`"

ibuclaw:
> A bit of history I could dug up, but it seems that destroy has been there for a long time:

231 is titled "Rename clear to destroy".  So it could always be my memory that is incorrect. :-)

jmdavis:
`clear` was talked about in TDPL. I don't know how old it is, but I don't think that it was new then. Regardless, it's never been about freeing memory, only about destroying the object and putting it in an invalid state, so it's really not the same as what this is trying to do - though in principle, the idea behind `clear`/`destroy` and getting rid of `delete` is that no one should ever do what `delete` does if they're using GC-allocated memory. Andrei has been adamant about that being a terrible idea for years now, but we've just never gotten rid of `delete` (if nothing else, because without allocators, trying to replace code that uses with code that does something similar with non-GC-allocated memory it is a royal pain).

I am a bit surprised thought that Andrei wants this `delete`-replacement. From what I recall, his stance in the past has always been that `delete` is fundamentally a bad idea for GC-allocated memory and that it should be gone, and keeping a library function around that does the same thing doesn't fit with that - though the fact that it's not built-in does reduce its visibility and make it seem less acceptable / more dangerous, so I guess that he considers that enough now.

Personally, I think that anyone looking to do something like this should not be using GC-allocated memory, but it's not like we're completely against letting folks risk blowing their own feet off - we just don't want it to be easy to do that by default.

changelog/core-memory-__delete.dd
```diff
@@ -0,0 +1,29 @@
+`core.memory.__delete` has been added
+
+$(REF __delete, core, memory) allows easy migration from the deprecated `delete`.
+`__delete` behaves exactly like `delete` and will call the class dtor immediately
```
andralex: `delete`/`__delete` apply to structs and arrays as well, so please update the documentation

changelog/core-memory-__delete.dd
andralex: The sed command could be a bit refined: `s/\Wdelete[ \t]+\(.*\);/__delete(\1);/g`

src/core/memory.d
```diff
@@ -919,3 +919,186 @@ extern (C) private pure @system @nogc nothrow
...
+    {
+        _d_delinterface(cast(void**) &t);
+    }
+    else static if (is(T == class))
+    {
+        _d_delclass(cast(Object*) &t);
+    }
+    else static if (is(T == struct))
+    {
+        _d_delstruct(cast(void**) &t, typeid(T));
```
andralex: Wait, this is incorrect. `__delete` should be only applicable to pointers to struct. Let's not innovate here.

src/core/memory.d
```diff
@@ -919,3 +919,186 @@ extern (C) private pure @system @nogc nothrow
...
+    }
+    else static if (is(T == U*, U))
+    {
+        t = null;
```
andralex: this should free memory

src/core/memory.d
```diff
@@ -919,3 +919,186 @@ extern (C) private pure @system @nogc nothrow
...
+    }
+    else static if (is(T : E[], E))
+    {
+        t = null;
```
andralex: this should free memory

andralex:
>Still, my understanding was that (years ago when delete was first put for deprecation) destroy was added as a migration path - by Andrei himself I believe.

>Obviously that function has failed spectacularly if this is needed. 🙄

Thanks for the quip. The olden idea is to keep the safer `clear` while still allowing people to free objects. That was (and is) possible by means of two function calls. More recently our doctrine for deprecation has become to provide a simple replacement for people who have `delete` spread out through their code and just want to keep the current semantics for the time being. So both `clear` and `__delete` have their place.

wilzbach:
Alrighty I went over the implementation again and it now matches with DMD 1:1 -> https://github.com/dlang/dmd/blob/v2.078.0/src/dmd/e2ir.d#L3886

The only thing that I couldn't figure out is how to detect at compile-time whether `-profile=gc` has been set by the user (DMD uses a different function mapping for this option - see [`toTraceGC`](https://github.com/dlang/dmd/blob/v2.078.0/src/dmd/e2ir.d#L5986)), but I doubt that it's essential for this PR as `-profile=gdc` is mostly about detecting allocations.

changelog/core-memory-__delete.dd
```diff
@@ -0,0 +1,35 @@
+`core.memory.__delete` has been added
+
+$(REF __delete, core, memory) allows easy migration from the deprecated `delete`.
+`__delete` behaves exactly like $(LINK2 $(ROOT_DIR)spec/expression.html#delete_expressions, `delete`):
```
JinShil: One of the procedures we need to perform to deprecate and eventually remove `delete` is to remove all references to it in the spec (see https://github.com/dlang/dlang.org/pull/1941).  I suggest removing the link to the spec.

src/core/memory.d
```diff
@@ -919,3 +919,201 @@ extern (C) private pure @system @nogc nothrow
...
+would not be a feasible option.
+
+In detail, `__delete` does the following:
+$(UL
+
+    $(LI If `t` is a class object reference, and
+        there is a destructor for that class, the destructor
+        is called for that object instance.
+    )
+    $(LI Otherwise, the garbage collector is called to immediately free the
```
andralex: This "otherwise" is ambiguous because it's applied to a conjunction. Does it refer to the fact that t is a class object reference, to the fact there is a destructor, or to the conjunction? (It applies to the second clause only). I'll make an edit to clarify.

andralex:
@wilzbach I just sent a patch to the doc - we should reuse `destroy` both in the documentation and the implementation. Also it seems associative arrays have been forgotten.

src/core/memory.d
wilzbach: FYI: That was taken 1:1 from the spec ;-)

src/core/memory.d
```diff
@@ -919,3 +919,197 @@ extern (C) private pure @system @nogc nothrow
...
+Note: Users should prefer $(REF destroy, object)` to explicitly finalize objects,
+and only resort to $(REF __delete, core,memory) when $(REF destroy, object)
+would not be a feasible option.
+
+In detail, `__delete(x)` returns with no effect if `x` is `null`. Otherwise, it
+performs the following actions in sequence:
+$(UL
+    $(LI
+        Calls `.object.destroy(*x)` (if `x` is a pointer) or `.object.destroy(x)`
+        (otherwise) to destroy the referred entity.
```
wilzbach: That's strictly speaking not true for class/interface objects and raw memory:

Exact behavior
--------------------

- If `x` is a class or interface object -> `rt_finalize`

https://github.com/dlang/druntime/blob/098a08a9593a8e265a9b02da55bda53393f3d5c1/src/rt/lifetime.d#L163-L167

Of course, `object.destroy` does call `rt_finalize`:

https://github.com/dlang/druntime/blob/098a08a9593a8e265a9b02da55bda53393f3d5c1/src/object.d#L2887-L2896

FTR `rt_finalize` is deprecated, but `rt_finalize2` does call a few interesting things:

https://github.com/dlang/druntime/blob/098a08a9593a8e265a9b02da55bda53393f3d5c1/src/rt/lifetime.d#L1382-L1419

- If `x` is a reference to a struct -> `Typeinfo_Struct.destroy`:

https://github.com/dlang/druntime/blob/098a08a9593a8e265a9b02da55bda53393f3d5c1/src/rt/lifetime.d#L177-L183

Here it's all about calling the `dtor`:

https://github.com/dlang/druntime/blob/098a08a9593a8e265a9b02da55bda53393f3d5c1/src/object.d#L1194-L1199

- If `x` is a reference to an arrays -> for all structs within: `Typeinfo_Struct.destroy`:

https://github.com/dlang/druntime/blob/098a08a9593a8e265a9b02da55bda53393f3d5c1/src/rt/lifetime.d#L1169

https://github.com/dlang/druntime/blob/098a08a9593a8e265a9b02da55bda53393f3d5c1/src/rt/lifetime.d#L1347-L1356

- Otherwise only `GC.free` is called.

src/core/memory.d
```diff
@@ -919,3 +919,197 @@ extern (C) private pure @system @nogc nothrow
...
+    )
+    $(LI
+        Frees the memory allocated for `x`. If `x` is a reference to a class
+        or interface, the memory allocated for the underlying instance is freed. If `x` is
+        a pointer, the memory allocated for the pointed-to object is freed. If `x` is a
+        built-in array or associative array, the memory allocated for the array is freed.
```
wilzbach: Associative arrays aren't allowed: https://run.dlang.io/is/CtfP8f

src/core/memory.d
```diff
@@ -919,3 +919,197 @@ extern (C) private pure @system @nogc nothrow
...
+        equivalents in the GC API), the behavior is undefined.
+    )
+    $(LI
+        Lastly, `x` is set to `null`. Any attempt to read or write the freed memory via
+        other references will result in undefined behavior.
+    )
+)
+
+Params:
+    t = aggregate object that should be destroyed
```
wilzbach: Reminder to myself: I will rename this to `x` then.

schveiguy:
From my memory:

1. `clear`/`GC.free` was intended as the replacement for `delete`. The problem with `delete` is that it wasn't possible to destroy things deterministically without also creating a dangling pointer. The required conflation of memory management with finalization/destruction was a faulty mechanism, especially in a GC-by-default language. I think the intention was to allow a library function, and stop reserving a keyword for such a thing.
2. `clear` was renamed to destroy because `.clear` is typically the name people use for removing all elements of a container. Given how prevalent UFCS is in D, then this creates massive confusion.

IMO, the `destroy`/`GC.free` mechanism is too raw, however. Sometimes it's exactly what you want, and GC.free `*requires*` a pointer, `destroy` requires a value. You are using clunky mechanisms everywhere if you use this thing.

wilzbach:
> @wilzbach I just sent a patch to the doc

Nice (I just removed your trailing white-space in case you were wondering why I force-pushed your commit).

andralex:
@wilzbach yah I wonder if we could submit a github issue - would be great if their editor supported trailing space elimination

andralex:
@wilzbach agreed about the discrepancies - in this case we must work backwards: `__delete` IS (by fiat) "destroy then deallocate". To the extent it isn't we need to adjust either `destroy` or `__delete`.

src/core/memory.d
andralex: @wilzbach @WalterBright and I are at POPL 2018 and one thing that became abundantly clear is we must overhaul the spec making it radically more precise. More on that later.

wilzbach:
> yah I wonder if we could submit a github issue - would be great if their editor supported trailing space elimination

https://github.com/isaacs/github

> @wilzbach agreed about the discrepancies - in this case we must work backwards: __delete IS (by fiat) "destroy then deallocate". To the extent it isn't we need to adjust either destroy or __delete.

What do you suggest? Do you want to change `__delete` do call `destroy`?
`_d_delclass` still checks for deallocators as [class deallocators haven't been deprecated yet](https://dlang.org/deprecate.html#Class%20allocators%20and%20deallocators).

I would simply change the wording to sth. like:

> Calls `.object.destroy(x)` (if `x` is a class or interface object) or  `typeid(*x).destroy(x)` (if `x` is pointer to a struct) to destroy the referred entity. Arrays of structs call `typeid(*x).destroy(x)` for each element.

andralex:
Yes, `__delete` must call `destroy`.  The wording you propose is good, thanks.

src/core/memory.d
```diff
@@ -919,3 +919,218 @@ extern (C) private pure @system @nogc nothrow
...
+and only resort to $(REF __delete, core,memory) when $(REF destroy, object)
+would not be a feasible option.
+
+In detail, `__delete(x)` returns with no effect if `x` is `null`. Otherwise, it
+performs the following actions in sequence:
+$(UL
+    $(LI
+        Calls `.object.destroy(x)` (if `x` is a class or interface object) or
+        `typeid(*x).destroy(x)` (if `x` is pointer to a struct) to destroy the referred entity.
+        Arrays of structs call `typeid(*x).destroy(x)` for each element.
```
wilzbach: This could also state: Arrays of structs call `typeid(x).destroy(&x)` for each element (that's what the implementation does below).

src/core/memory.d
```diff
@@ -919,3 +919,218 @@ extern (C) private pure @system @nogc nothrow
...
+// Deleting raw memory
+unittest
+{
+    import core.memory : GC;
+    auto a = GC.malloc(5);
+    __delete(a);
+    assert(a is null);
+}
+
+// __delete returns with no effect if x is null
```
wilzbach: Is there a simple way to add a test this for "no effect"?

src/core/memory.d
```diff
@@ -919,3 +919,218 @@ extern (C) private pure @system @nogc nothrow
...
+Note: Users should prefer $(REF destroy, object)` to explicitly finalize objects,
+and only resort to $(REF __delete, core,memory) when $(REF destroy, object)
+would not be a feasible option.
+
+In detail, `__delete(x)` returns with no effect if `x` is `null`. Otherwise, it
+performs the following actions in sequence:
+$(UL
+    $(LI
+        Calls `.object.destroy(x)` (if `x` is a class or interface object) or
+        `typeid(*x).destroy(x)` (if `x` is pointer to a struct) to destroy the referred entity.
```
andralex: Here it should just be `.object.destroy(*x)`, no? It's faster and better documented.

src/core/memory.d
andralex: I think it should be `.object.destroy(e)` for each element `e` in the array.

src/core/memory.d
```diff
@@ -919,3 +919,218 @@ extern (C) private pure @system @nogc nothrow
...
+    x = aggregate object that should be destroyed
+
+See_Also: $(REF destroy, object), $(REF free, core,GC)
+*/
+void __delete(T)(ref T x) @system
+{
+    // See also: https://github.com/dlang/dmd/blob/v2.078.0/src/dmd/e2ir.d#L3886
+    static if (is(T == interface))
+    {
+        (cast(Object) x).destroy;
```
andralex: I think this should do what the spec says it does: `.object.destroy(cast(Object) x);` This is a pedantic point because `Object` does not define a method `destroy`.

src/core/memory.d
```diff
@@ -919,3 +919,218 @@ extern (C) private pure @system @nogc nothrow
...
+    }
+    else static if (is(T == class))
+    {
+        (cast(Object) x).destroy;
```
andralex: Same here, `.object.destroy(cast(Object) x);` would clarify lookup

src/core/memory.d
```diff
@@ -919,3 +919,218 @@ extern (C) private pure @system @nogc nothrow
...
+    }
+    else static if (is(T == U*, U))
+    {
+        static if (is(U == struct))
+            typeid(U).destroy(x);
```
andralex: As mentioned, I think this is less efficient - it entails an indirect call instead of simply calling the destructor (if defined) or doing nothing (if not). I suggest: `.object.destroy(*x)`

src/core/memory.d
```diff
@@ -919,3 +919,218 @@ extern (C) private pure @system @nogc nothrow
...
+    }
+    else static if (is(T : E[], E))
+    {
+        static if (is(E == struct))
+        {
+            foreach (ref e; x)
+                typeid(E).destroy(&e);
```
andralex: same here

src/core/memory.d
```diff
@@ -919,3 +919,218 @@ extern (C) private pure @system @nogc nothrow
...
+        }
+    }
+    else
+    {
+        static assert(0, "It is not possible to delete: `" ~ T.stringof ~ "`");
+    }
+
+    static if (is(T == interface) || is(T == class) || is(T == U2*, U2) || is(T : E2[], E2))
+    {
+        GC.free(&x);
```
andralex: I'm not sure about this in the case of classes and interfaces. For those shouldn't the call be `GC.free(cast(void*) x)`?

src/core/memory.d
wilzbach: Are you sure about `object.destroy` being faster?

`TypeInfo_Struct.destroy`:

https://github.com/dlang/druntime/blob/18be8bf97933c2e5c0d0948f18477268ac7c1766/src/object.d#L1192-L1201

`object.destroy`:

https://github.com/dlang/druntime/blob/18be8bf97933c2e5c0d0948f18477268ac7c1766/src/object.d#L2954-L2965

andralex: The former is an indirect call and the second is a direct call. Truth be told, destroy does more work because it reinitializes the object. In the case of `__delete` we don't care to reinitialize the object because we just wrote in the documentation that access after `__delete` is undefined behavior.

So all in all the best way to go is to write in the documentation that we call `destroy` and actually call `_destructRecurse`. That covers all defined cases and is fast.

wilzbach: `_destructRecurse` is private and in `object` - so we can't even do `package(core)`.

src/core/memory.d
wilzbach: Added tests for `GC.free`

src/core/memory.d
```diff
@@ -919,3 +919,230 @@ extern (C) private pure @system @nogc nothrow
...
+Note: Users should prefer $(REF destroy, object)` to explicitly finalize objects,
+and only resort to $(REF __delete, core,memory) when $(REF destroy, object)
+would not be a feasible option.
+
+In detail, `__delete(x)` returns with no effect if `x` is `null`. Otherwise, it
+performs the following actions in sequence:
+$(UL
+    $(LI
+        Calls `.object.destroy(x)` (if `x` is a class or interface object) or
+        `(*x).__xdtor()` (if `x` is pointer to a struct and a custom destructor exists) to destroy the referred entity.
```
wilzbach: That's exactly what we do, but I wasn't whether there's a more elegant way to put it - `__xdtor` is a bit ugly, but it's still better than "`.object.destroy(*x)`, but doesn't initialize the struct to `T.init`"

wilzbach:
So are we ready to :rocket: here?

src/core/memory.d
```diff
@@ -919,3 +919,230 @@ extern (C) private pure @system @nogc nothrow
 
     pragma(mangle, "free") void fakePureFree(void* ptr);
 }
+
+extern(C) private @system nothrow
```
andralex: they are `@nogc` as well

src/core/memory.d
```diff
@@ -919,3 +919,230 @@ extern (C) private pure @system @nogc nothrow
...
+    pragma(mangle, "_d_delstruct") void _d_delstruct(void**, TypeInfo_Struct);
+    pragma(mangle, "_d_delmemory") void _d_delmemory(void**);
+    pragma(mangle, "_d_delarray_t") void _d_delarray_t(void**, TypeInfo_Struct);
+}
+
+/**
+The `delete` keyword allowed to free GC-allocated memory.
+As this is inherently not `@safe`, it has been deprecated.
+This function has been added to provide an easy transition from `delete`.
+It performs the same functionality as `delete`.
```
andralex: Instead of assuming people know `delete` and defining `__delete` in terms of it, describe `__delete` as "destroy then deallocate" and at the very end "see also the deprecated keyword" etc.

src/core/memory.d
andralex: People cannot be expected to know what `__xdtor` does, but they do know what a destructor is. So the text should be:

"Calls the destructor `~this()` for the object referred to by `x` (if `x` is a class or interface reference) or for the object pointed to by `x` (if `x` is a pointer to a `struct`). If no destructor is defined, this step has no effect."

src/core/memory.d
```diff
@@ -919,3 +919,230 @@ extern (C) private pure @system @nogc nothrow
...
+would not be a feasible option.
+
+In detail, `__delete(x)` returns with no effect if `x` is `null`. Otherwise, it
+performs the following actions in sequence:
+$(UL
+    $(LI
+        Calls `.object.destroy(x)` (if `x` is a class or interface object) or
+        `(*x).__xdtor()` (if `x` is pointer to a struct and a custom destructor exists) to destroy the referred entity.
+        Arrays of structs with a custom destructor call `x.__xdtor()` for each element in the array.
+    )
```
andralex: "Arrays of `struct`s call the destructor, if defined, for each element in the array."

src/core/memory.d
```diff
@@ -919,3 +919,230 @@ extern (C) private pure @system @nogc nothrow
...
+
+Params:
+    x = aggregate object that should be destroyed
+
+See_Also: $(REF destroy, object), $(REF free, core,GC)
+*/
+void __delete(T)(ref T x) @system
+{
+    static void _destructRecurse(S)(ref S s)
+        if (is(S == struct))
```
andralex: unindent

src/core/memory.d
```diff
@@ -919,3 +919,230 @@ extern (C) private pure @system @nogc nothrow
...
+    {
+        static if (__traits(hasMember, S, "__xdtor") &&
+                   // Bugzilla 14746: Check that it's the exact member of S.
+                   __traits(isSame, S, __traits(parent, s.__xdtor)))
```
andralex: I see the bugzilla is fixed, why do we need this?

wilzbach: Because fixing 14746, added this to `_destructRecurse`: https://github.com/dlang/druntime/commit/1a2290b1c36f8cdf66a9c4ce388b6a4b38ff9735

wilzbach:
> ping @wilzbach

Sorry. Pong. Addressed the final nits & squashed everything into one commit.

PR #2040, opened 2018-01-13
fix issue Issue 15482 - new uuid.d forbids to link statically with other libraries

Temtaime:
All of those symbols are stored inside uuid.lib on windows.


BBasile:
Can you amend your commit, change the message to 

`fix issue Issue 15482 - new uuid.d forbids to link statically with other libraries`

and `push --force` ? This allows the bot to reference the issue and to create a changelog entry.

Temtaime:
Yes, i'd be glad
Thanks

src/core/sys/windows/uuid.d
```diff
@@ -3,4078 +3,4077 @@ version (Windows):
...
-const IID CLSID_IImageDecodeFilter = {0x607FD4E8, 0x0A03, 0x11D1, [0xAB, 0x1D, 0x00, 0xC0, 0x4F, 0xC9, 0xB3, 0x04]};
-const IID CLSID_IImgCtx = {0x3050F3D6, 0x98B5, 0x11CF, [0xBB, 0x82, 0x00, 0xAA, 0x00, 0xBD, 0xCE, 0x0B]};
-const IID CLSID_ImageList = {0x7C476BA2, 0x02B1, 0x48F4, [0x80, 0x48, 0xB2, 0x46, 0x19, 0xDD, 0xC0, 0x58]};
-const IID CLSID_ImageProperties = {0x7AB770C7, 0x0E23, 0x4D7A, [0x8A, 0xA2, 0x19, 0xBF, 0xAD, 0x47, 0x98, 0x29]};
-const IID CLSID_InProcFreeMarshaler = {0x0000033A, 0x0000, 0x0000, [0xC0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x46]};
-const IID CLSID_IntDitherer = {0x05F6FE1A, 0xECEF, 0x11D0, [0xAA, 0xE7, 0x00, 0xC0, 0x4F, 0xC9, 0xB3, 0x04]};
-const IID CLSID_IntelliForms = {0x613AB92E, 0x16BF, 0x11D2, [0xBC, 0xA5, 0x00, 0xC0, 0x4F, 0xD9, 0x29, 0xDB]};
-const IID CLSID_Internet = {0x871C5380, 0x42A0, 0x1069, [0xA2, 0xEA, 0x08, 0x00, 0x2B, 0x30, 0x30, 0x9D]};
-const IID CLSID_InternetButtons = {0x1E796980, 0x9CC5, 0x11D1, [0xA8, 0x3F, 0x00, 0xC0, 0x4F, 0xC9, 0x9D, 0x61]};
-const IID CLSID_InternetConnectionBeaconServic = {0x04DF613A, 0x5610, 0x11D4, [0x9E, 0xC8, 0x00, 0xB
```
wilzbach: What's the reason for the second `extern`?

CyberShadow:
https://issues.dlang.org/show_bug.cgi?id=15482#c8

Temtaime:
Why make our custom uuids that can somehow differ(by mistake or so) while they already here in uuid.lib ?
It's bug-prone i think.

src/core/sys/windows/uuid.d
Temtaime: Because extern behaves as one in C++

CyberShadow:
Well, an external library is an additional dependency and complicates the build process. Changing the export declarations will also avoid breaking code that relies on our `uuid.d`.

Temtaime:
It is a winapi library. It comes as well as others like kernel32, ole32 etc with every windows compiler.
It is not a dependency.

CyberShadow:
> It is not a dependency, it's a system library.

You're forgetting that libraries come in different formats.

What about OMF?

This is still a breaking change no?

Have you verified that all versions of uuid.lib contain the same or a superset of the UUIDs removed here?

rainers:
I agree with @CyberShadow here:
- the distributed OMF library is still the same as 12 years ago, it's unlikely to contain all the UUIDs.
- it's a breaking change to remove the values, because they might have been used at CTFE or for compile time initialization.

Though technically also a breaking change, I'd favor removing `export extern(C)`, too.

A current downside seems to be that all the definitions end up in a single data segment, so if one is accessed, you get all UUIDs in your executable. (Might be object file dependent).

rainers:
Here is the proposed alternative: https://github.com/dlang/druntime/pull/2063

CyberShadow:
The alternative was merged, so we can close this one. Thank you for the submission.

PR #2042, opened 2018-01-15
Build rt documentation on dlang.org

wilzbach:
I didn't see any reason why this hasn't been done before. I often missed this.
Exposing the docs, will also make it easier to find respective functions on Google.

Note: this won't be visible (part of the menu) for now. That will require https://github.com/dlang/dlang.org/pull/2084

Of course this is still a bit unpolished, but it won't change if we don't expose it.


andralex:
So nice! cc @WalterBright

PR #2046, opened 2018-01-17
[trivial] Fix ddoc source path of `rt_dmain2`

JohanEngelen:
The source path of many `rt` files is absent or broken. This fixes it for one file.

PR #2047, opened 2018-01-18
Remove complex numbers form the gcbench suite

wilzbach:
The DRuntime testsuite contains an old copy of VisualD.
However, this is preventing us to move forward with the complex deprecation:
https://github.com/dlang/phobos/pull/6014

It's save to remove this as is just the GC performance testsuite.


wilzbach:
CI failure also happens at https://github.com/dlang/druntime/pull/2046#issuecomment-358487445, I will have a look.

wilzbach:
The CircleCi failure is clearly unrelated and I am working at https://github.com/dlang/druntime/pull/2048 to fix it. However as https://github.com/dlang/druntime/pull/2048 has no clear direction except that a single line is responsible and all other CIs are happy with this - I am going to pull this.

PR #2048, opened 2018-01-18
Attempt to fix CircleCi

wilzbach:
Start to dig into the CircleCi failures seen here:
- https://github.com/dlang/druntime/pull/2046#issuecomment-358487445
- https://github.com/dlang/druntime/pull/2047

Reproducible locally with:

```
TEST_COVERAGE=1 make -f posix.mak unittest-debug
```

Failure:

```
****** FAIL debug64 core.time
core.exception.AssertError@src/core/time.d(527): unittest failure
----------------
```

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -520,7 +520,7 @@ public:
     {
         foreach(T; _TypeTuple!(Duration, const Duration, immutable Duration))
         {
-            foreach(U; _TypeTuple!(Duration, const Duration, immutable Duration))
+            foreach(U; _TypeTuple!(Duration, /*const Duration,*/ immutable Duration))
             {
                 T t = 42;
                 U u = t;
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -520,10 +520,14 @@ public:
     {
         foreach(T; _TypeTuple!(Duration, const Duration, immutable Duration))
         {
-            foreach(U; _TypeTuple!(Duration, /*const Duration,*/ immutable Duration))
+            foreach(U; _TypeTuple!(Duration, const Duration, immutable Duration))
             {
                 T t = 42;
-                U u = t;
+                // workaround https://issues.dlang.org/show_bug.cgi?id=18296
+                version (D_Coverage)
+                    U u = T(t._hnsecs);
+                else
+                    U u = t;
                 assert(t == u);
                 assert(copy(t) == u);
                 assert(t == copy(u));


src/core/time.d
```diff
@@ -520,7 +520,7 @@ public:
     {
         foreach(T; _TypeTuple!(Duration, const Duration, immutable Duration))
         {
-            foreach(U; _TypeTuple!(Duration, const Duration, immutable Duration))
+            foreach(U; _TypeTuple!(Duration, /*const Duration,*/ immutable Duration))
```
wilzbach: I don't know why this doesn't work anymore when compiled with `-cov` :/

wilzbach: This is the first build that failed: https://circleci.com/gh/dlang/druntime/1607

So probably this is due to:

https://github.com/dlang/druntime/commit/b38f88f22a264f2df5c7aabb66d1b63834db0ada

(and something else as https://github.com/dlang/druntime/pull/2032 did pass merge being merged)

schveiguy:
This failure is bizarre. Can you print out what the value is if they are not equal?

schveiguy:
I wonder if it has something to do with the `T t = 42`? Typically, you would use `42.hnsecs`, but it's calling the ctor directly since we are in the same file.

wilzbach:
> Can you print out what the value is if they are not equal?

`0 != 42` - we really need to get dmd to rewrite `assert` to be more user-friendly.

jmdavis:
I see no reason why this should be failing, especially when `Duration` and `immutable Duration` pass. And if it works without `-cov` but doesn't work with it, then that reeks of a compiler bug.

> 0 != 42 - we really need to get dmd to rewrite assert to be more user-friendly.

It's been discussed before. I would be nice, but it's too expensive.

ZombineDev:
> It's been discussed before. I would be nice, but it's too expensive.

It can be opt-in - problem solved!

rainers:
I investigated this a bit: this code fails in the second iteration:
```
            foreach(U; _TypeTuple!(Duration, const Duration))
            {
                const Duration t = 42;
                U u = t;
                assert(t == u);
            }
```
but works if either -cov or -fPIC is removed.

This disassembly shows that an instruction is omitted:
```
  ...
  90:	48 8b 0d 00 00 00 00 	mov    0x0(%rip),%rcx        # 97 <_Dmain+0x97>
			93: R_X86_64_GOTPCREL	_D4core4time8Duration6__initZ-0x4
  97:	48 8b 19             	mov    (%rcx),%rbx
  9a:	48 89 5d f0          	mov    %rbx,-0x10(%rbp)
  9e:	be 2a 00 00 00       	mov    $0x2a,%esi
  a3:	48 8d 7d f0          	lea    -0x10(%rbp),%rdi
  a7:	e8 00 00 00 00       	callq  ac <_Dmain+0xac>
			a8: R_X86_64_PLT32	_D4core4time8Duration6__ctorMFNaNbNcNiNflZSQBpQBnQBl-0x4
  ac:	48 8b 05 00 00 00 00 	mov    0x0(%rip),%rax        # b3 <_Dmain+0xb3>  //// coverage code
			af: R_X86_64_GOTPCREL	_D1x10__coverageZ-0x4
  b3:	ff 40 24             	incl   0x24(%rax)                                //// coverage code
  >>>  48 8b 4d f0          	mov    -0x10(%rbp),%rbx          <<< this instruction missing
  b6:	48 89 5d f8          	mov    %rbx,-0x8(%rbp)
  ba:	48 8b 0d 00 00 00 00 	mov    0x0(%rip),%rcx        # c1 <_Dmain+0xc1>  //// coverage code
			bd: R_X86_64_GOTPCREL	_D1x10__coverageZ-0x4
  c1:	ff 41 28             	incl   0x28(%rcx)                                //// coverage code
  c4:	48 8d 75 f0          	lea    -0x10(%rbp),%rsi
  c8:	48 8d 7d f8          	lea    -0x8(%rbp),%rdi
  cc:	b9 08 00 00 00       	mov    $0x8,%ecx
  d1:	33 c0                	xor    %eax,%eax
  d3:	f3 a6                	repz cmpsb %es:(%rdi),%ds:(%rsi)
   ...
```

MartinNowak:
Well who know what that backend does, some stale com-subs or whatever, the [IR gen for coverage](https://github.com/dlang/dmd/blob/439cc6aafab8b2fd5679472ff3efe0ae24447171/src/dmd/toir.d#L73) looks ok though.
We recently changed how the coverage symbol is emitted, so that seems to have triggered a dormant backend bug.
[18296 – [Reg2.078.1] invalid code with coverage and copy construction](https://issues.dlang.org/show_bug.cgi?id=18296)

rainers:
I suspect that the switch from static to global symbol has uncovered the bug, because it now uses an additional indirection.

wilzbach:
@MartinNowak thanks for adding your more specific workaround and I would love to approve it, but it seems like GitHub doesn't allow me to approve my own PRs :/
(auto-merge requires at least one approval)

wilzbach:
Urgh now Jenkins is failing, but the temporary fix is on the merge queue: https://github.com/dlang/ci/pull/132

So I'm force-merging this PR here.

PR #2049, opened 2018-01-18
[WIP][DONOTREVIEW] Assert Template Lowering

somzzz:
@andralex

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3982,6 +3982,10 @@ void __switch_error()(string file = __FILE__, size_t line = __LINE__)
     __switch_errorT(file, line);
 }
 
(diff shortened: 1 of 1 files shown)


wilzbach:
Are you aware of this DIP?
https://wiki.dlang.org/DIP83

Things that are really cool about this DIP:
- allows to overload opUnary and opBinary
- error diagnostics can be handled by library code

The DIP doesn't solve these cases:
- `assert([1, 2].equals([2, 3]))
- `assert(myFunctionWhichIsBuggyAndReturnsAnEmptyString)`

wilzbach:
Also here is an example of how good and helpful diagnostics look like: https://code.dlang.org/packages/fluent-asserts

andralex:
(For discussion with @somzzz and me only. I realized this doesn't need to be a PR, a branch would be enough.)

PR #2050, opened 2018-01-18
Fix paths in benchmark runner

aliak00:
On osx I was getting a “No such file” exception for dante.txt
because CWD depended on where I executed `rdmd benchmark.d` from.
And the object files were generated in a directory also depending
on that.

This fixes obj dir to same level as bin dir within the benchmarks
dir, and allows the runCmd to take a workDir parameter for running
the benchmark executables

### DIFF
--- a/benchmark/runbench.d
+++ b/benchmark/runbench.d
@@ -17,13 +17,13 @@ struct Config
     uint repeat = 10;
 }
 
-string runCmd(string cmd, bool verbose)
+string runCmd(string cmd, bool verbose, in char[] workDir = null)
 {
     import std.exception : enforce;
-    import std.process : executeShell;
+    import std.process : executeShell, Config;
 
     if (verbose) writeln(cmd);
-    auto res = executeShell(cmd);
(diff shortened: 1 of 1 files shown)


andralex:
cc @wilzbach @MartinNowak

benchmark/runbench.d
```diff
@@ -109,7 +111,7 @@ void runTests(Config cfg)
         foreach (_; 0 .. cfg.repeat)
         {
             sw.reset;
-            auto output = runCmd(cmd, cfg.verbose);
+            auto output = runCmd(cmd, cfg.verbose, cwd);
```
wilzbach: Why `cwd`? How about `__FILE_FULL_PATH__.dirName`?

wilzbach: Argh - I didn't see that `cwd` is already set to exactly this.

wilzbach:
CircleCi failure is unrelated - see https://github.com/dlang/druntime/pull/2048 for details.

aliak00:
Sweet! Thanks 😄

PR #2051, opened 2018-01-20
Add missing types for cruntime bionic.

skl131313:
Seems to be uint32_t for every platform, but I'm not too familiar with it, maybe some build config somewhere changes it or uses a different file.

https://android.googlesource.com/platform/bionic/+/fb96db14d11fa75a5a44162188d6bc441f9491e0/libc/include/sys/types.h#74

### DIFF
--- a/src/core/sys/posix/sys/types.d
+++ b/src/core/sys/posix/sys/types.d
@@ -234,6 +234,11 @@ else version( CRuntime_Bionic )
         alias ushort    mode_t;
         alias ushort    nlink_t;
     }
+    else version(X86_64)
+    {
+        alias ushort    mode_t;
+        alias uint      nlink_t;
+    }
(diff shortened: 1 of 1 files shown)


joakim-noah:
This change alone is fairly useless.  When adding new architectures, there are a fair amount of druntime modules that have to be ported.  See my pulls that added these types: #681 and #1267.  I have a patch in progress to add support for 64-bit ARM, it's an updated version of the latter pull.

wilzbach:
@joakim-noah fair enough, but that doesn't stop us from pulling this? I understand that this is just a small step towards `X86_64` for Bionic, but better a small step than no step?

joakim-noah:
I guess, I just suspect the rest will never be ported, particularly considering basically nobody uses Android/x64.

PR #2052, opened 2018-01-22
Implement a monitor system for the GC

mathias-lang-sociomantic:
Allow one to hook the start and end of a fullcollect, potentially gathering statistics / logging the event as one see fit.
This was present in the D1 runtime but got stripped for unknown reason (git history has no mention of it).
Sociomantic relies on it heavily for some real-time apps.

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -140,6 +140,12 @@ private
     extern (C) BlkInfo_ gc_query( void* p ) pure nothrow;
     extern (C) GC.Stats gc_stats ( ) nothrow @nogc;
 
+    // Alias used because DMD thinks the delegate are extern(C) as well otherwise
+    alias CollectionStartHook = void delegate() nothrow @nogc;
+    alias CollectionEndHook = void delegate(size_t, size_t) nothrow @nogc;
+    extern (C) void gc_monitor ( CollectionStartHook on_start,
+        CollectionEndHook on_end) nothrow @nogc;
+
     extern (C) void gc_addRoot( in void* p ) nothrow @nogc;
     extern (C) void gc_addRange( in void* p, size_t sz, const TypeInfo ti = null ) nothrow @nogc;
 
@@ -678,6 +684,23 @@ struct GC
     static Stats stats() nothrow
     {
         return gc_stats();
+    }
+
+    /**
+     * Provides a mean to hook the GC on the beggining and end of a collection
+     *
+     * If one of the event is of no interest, `null` can be provided.
+     *
+     * Params:
+     *   on_start = A delegate to call whenever a collection starts
+     *   on_end   = A delegate to call whenever a collection ends.
+     *              The first argument is the number of bytes freed overall,
+     *              the second the number of bytes freed within full pages.
+     */
+    static void monitor(CollectionStartHook on_start, CollectionEndHook on_end)
+        nothrow @nogc
+    {
+        gc_monitor(on_start, on_end);
     }
 
     /**
--- a/src/gc/gcinterface.d
+++ b/src/gc/gcinterface.d
@@ -37,6 +37,10 @@ struct Range
 
 interface GC
 {
+    /// Provided for implementation's convenience, not intended for public usage
+    protected alias CollectionStartHook = void delegate() nothrow @nogc;
+    /// Ditto
+    protected alias CollectionEndHook = void delegate(size_t, size_t) nothrow @nogc;
 
     /*
      *
@@ -147,6 +151,13 @@ interface GC
      * Useful for debugging and tuning.
      */
     core.memory.GC.Stats stats() nothrow;
+
+    /**
+     * Track beginning and end of allocation
+     * Useful for debugging and tuning.
+     */
+    void monitor (CollectionStartHook on_start, CollectionEndHook on_end)
+        nothrow @nogc;
 
     /**
      * add p to list of roots
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -255,6 +255,10 @@ class ConservativeGC : GC
     __gshared size_t line;
     __gshared char*  file;
 
+    /// Hook used for debugging application-side
+    __gshared CollectionStartHook on_collect_start;
+    __gshared CollectionEndHook on_collect_end;
+
     Gcx *gcx;                   // implementation
 
     import core.internal.spinlock;
@@ -1200,6 +1204,16 @@ class ConservativeGC : GC
         return ret;
     }
 
+    override void monitor (CollectionStartHook on_start, CollectionEndHook on_end)
+        nothrow @nogc
+    {
+        static void store (CollectionStartHook start, CollectionEndHook end)
+        {
+            on_collect_start = start;
+            on_collect_end = end;
+        }
+        runLocked!(store)(on_start, on_end);
+    }
 
     //
     //
@@ -2392,6 +2406,8 @@ struct Gcx
         }
 
         debug(COLLECT_PRINTF) printf("Gcx.fullcollect()\n");
+        if (ConservativeGC.on_collect_start !is null)
+            ConservativeGC.on_collect_start();
         //printf("\tpool address range = %p .. %p\n", minAddr, maxAddr);
 
         {
@@ -2456,6 +2472,8 @@ struct Gcx
 
         updateCollectThresholds();
 
+        if (ConservativeGC.on_collect_end !is null)
+            ConservativeGC.on_collect_end(freedLargePages, freedSmallPages);
         return freedLargePages + freedSmallPages;
     }
 
@@ -3443,4 +3461,3 @@ unittest
     GC.free(z);
     GC.minimize(); // release huge pool
 }
-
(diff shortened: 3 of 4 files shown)


mathias-lang-sociomantic:
CC @mihails-strasuns-sociomantic @leandro-lucarella-sociomantic

src/core/memory.d
```diff
@@ -140,6 +140,12 @@ private
     extern (C) BlkInfo_ gc_query( void* p ) pure nothrow;
     extern (C) GC.Stats gc_stats ( ) nothrow @nogc;
 
+    // Alias used because DMD thinks the delegate are extern(C) as well otherwise
+    alias CollectionStartHook = void delegate() nothrow @nogc;
+    alias CollectionEndHook = void delegate(size_t, size_t) nothrow @nogc;
```
mathias-lang-sociomantic: @mihails-strasuns-sociomantic mentioned that this could be two instances of the `GC.Stats` struct.
However, those are not so cheap to get, since we iterate over the pools (unless I misunderstood something).
Given the use cases for those hooks, I think having a no-cost approach is quite mandatory.

src/gc/impl/conservative/gc.d
```diff
@@ -255,6 +255,10 @@ class ConservativeGC : GC
     __gshared size_t line;
     __gshared char*  file;
 
+    /// Hook used for debugging application-side
+    __gshared CollectionStartHook on_collect_start;
+    __gshared CollectionEndHook on_collect_end;
```
mathias-lang-sociomantic: One point raised by @mihails-strasuns-sociomantic is that having `delegate` + `__gshared` is dangerous. Since we have the lock it should be okay, but I'm interested to know if there's something addressing this issue anywhere else.

src/core/memory.d
mihails-strasuns-sociomantic: >  I think having a no-cost approach is quite mandatory

It still has to hold GC lock which immediately makes it pretty far from no-cost.

leandro-lucarella-sociomantic: I think acquiring a lock (specially in single threaded environments, where is basically an atomic op usually) is considered no-cost compared to traverse the whole pool set. Also, there seems to be no reason to do it when the user can get the stats anyway with just one extra call, if needed.

src/gc/gcinterface.d
```diff
@@ -148,6 +152,13 @@ interface GC
      */
     core.memory.GC.Stats stats() nothrow;
 
+    /**
+     * Track beginning and end of allocation
+     * Useful for debugging and tuning.
+     */
+    void monitor (CollectionStartHook on_start, CollectionEndHook on_end)
+        nothrow @nogc;
```
leandro-lucarella-sociomantic: `@__future`?

src/gc/gcinterface.d
```diff
@@ -37,6 +37,10 @@ struct Range
 
 interface GC
 {
+    /// Provided for implementation's convenience, not intended for public usage
+    protected alias CollectionStartHook = void delegate() nothrow @nogc;
+    /// Ditto
+    protected alias CollectionEndHook = void delegate(size_t, size_t) nothrow @nogc;
```
leandro-lucarella-sociomantic: `@__future`? Although I don't think it's working for aliases.

leandro-lucarella-sociomantic:
BTW, all tests are failing, looks like a linking problem, at lease in jenkins:
```
src/core/memory.d:(.text._D4core6memory2GC7monitorFNbNiDFNbNiZvDFNbNimmZvZv[_D4core6memory2GC7monitorFNbNiDFNbNiZvDFNbNimmZvZv]+0x37): undefined reference to `gc_monitor'
```

src/gc/gcinterface.d
mathias-lang-sociomantic: I don't think that's necessary for `alias` actually. If someone defines something with the same name, it will just shadow this.

src/gc/gcinterface.d
mathias-lang-sociomantic: That makes sense.
I would be extremely surprised if someone was implementing this interface, but I suppose that's a good test anyway.

andralex:
Nice, I have a few uses already. One thing needed would be composability - a call to monitor should not overwrite the previous call. So we're looking at a singly-linked list of hooks.

There's some nice related work in the area:

* http://www.cs.tufts.edu/r/redline/papers/gcassertions-pldi-2009.pdf
* https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/notifications

leandro-lucarella-sociomantic:
I agree is good to keep the complexity out of the GC/runtime if possible. Is true that if some library want to hook to the GC monitor, then things get hairy, as user code implementing this kind of hooking mechanism won't be enough. But then this is a pretty low level construct and one might argue that probably a library shouldn't be hooking to this, but providing the user a callback in case the user want to hook it.

andralex:
This could get modularized in a number of ways. At any rate there should be a public API that allows code to add hooks without clobbering existing hooks.

One more possibility that comes to mind is what some Windows APIs do - pass the "previous" delegate as an argument to the hook. Then the hook may decide cooperatively to continue down the chain by calling the hook, or stop the chain.

src/core/memory.d
mihails-strasuns-sociomantic: I have two unrelated concerns about it:

1. It is not cheap in multi-threaded environments and while single-threaded optimization is indeed most suitable to our needs, druntime version should be more uniformly useful.

2. To avoid calculating exact stats this PR right now exposes `freedLargePages` and `freedSmallPages` as monitored values and I would argue those alone have rather limited usefulness.

What would be really cool is to accept different types of callback delegates and provide required stats based on what is desired by callback. Not sure if it can fit into API elegantly though.

src/gc/impl/conservative/gc.d
mihails-strasuns-sociomantic: Well, from pure theoretical PoV it should require `shared delegate` but I still have no idea if that is even implemented to mean something in D frontend.

andralex:
Threads are likely to make matters difficult for a get/set interface backed by globals (currently there's only means for setting).

* A thread would find it impossible to insert a new hook reliably - by the time is has loaded the current hook, another thread could install its own hook. A CAS-based approach in the implementation would solve the matter reliably and efficiently.
* In the code sketched by @mathias-lang-sociomantic, there's the matter of allocating memory for the hook itself. That's tricky given that we're talking about hooks into the GC itself; this is the kind of difficulty that shouldn't be left to user code.
* There are a few subtler issues such as a hook attempting to insert a new hook during execution (probably should be disallowed).

mihails-strasuns-sociomantic:
There is also another multi-threaded concern (referenced here https://github.com/dlang/druntime/pull/2052#discussion_r162976274) - hook list stores delegates that capture thread some context and thus executing such delegate may have a race with a thread that registered it over the context. Basically `shared(this)` problem.

andralex:
@mihails-strasuns-sociomantic wouldn't all hooks be invoked from the collection thread with all other threads frozen?

mihails-strasuns-sociomantic:
@andralex you are correct of course - and unregistering thread from runtime is `@system` right now AFAICS. False alarm.

rainers:
Useful addition. I did something similar in my early GC adventures. 

* if the callback registration gets a bit more involved, maybe it would be better if that would be in another object so that different GC implementations can call into that to invoke the hooks. That way the GC interface doesn't have to be changed for every added callback.
* callbacks could also be united by letting them pass an ID (e.g. startCollection, stopCollection) and custom data (basically a tagged union).
* would it be possible to allow calling some query functions during the callback, like GC.stats?

leandro-lucarella-sociomantic:
Just to recap, what would be the minimum requirements for this PR to be accepted?

andralex:
Once the feature is in, there will be a healthy demand for its modular use. Thinking of things like caches, memoization, weak pointers, files to close, etc. 

The list would include:

* A careful specification.
  * Are the hooks called with threads frozen/melted?
  * What limitations are imposed against the code in hooks - e.g. access to TLS, `shared`, `__gshared` data?
  * What happens when a hook calls GC query functions such as `GC.getAttr`?
  * Can hooks allocate memory?
  * Can hooks free memory?
  * When is the context pointer of the delegates considered for deallocation?
  * Is the hook persistent for the whole application, or does it go away after one invocation?
  * What if a hook tries to register another hook? (This may be needed for certain idioms and is influenced by the point above.)
* An API for atomically adding hooks; they should be chained together.
* The order of serving (FCFS vs LCFS) should be defined. In the spirit of destructors, perhaps LCFS is the order of choice. 
* I think we can get away without an API for removing hooks. (One simple solution is to give the hook the chance to remove itself upon execution, i.e. by returning false.) If a need for a more sophisticated API comes forward, we can add to it later.

An eye for applications would be great. I'm thinking assertions ("this object is no longer referenced") and caches.

mathias-lang-sociomantic:
Thanks for your feedback @andralex . This will be taken over by someone else in the future.

leandro-lucarella-sociomantic:
We kind of did a survey on our applications to see how this facility was used, and it was almost exclusively used to get how much time was spent doing collections. The [list of requirements](https://github.com/dlang/druntime/pull/2052#issuecomment-362376766) seem to be super reasonable and I think we should go that way if this gets ever implemented, but actually we are going to switch the focus to improve GC stats instead (we'll go more on the path of https://github.com/dlang/druntime/pull/1846), which is what we really need.

PR #2053, opened 2018-01-22
Fix Issue 18279 - rt.util.utf does not properly reserve buffer in toUTF16/toUTF16z

n8sh:
The offending pattern:
```d
    wchar[] r;
    size_t slen = s.length;
    r.length = slen;
    r.length = 0;
    // ... followed by code that appends to r with ~=
```

The apparent intention is to preallocate the buffer but this idiom doesn't work [as pointed out by Dmitry Olshansky](https://github.com/dlang/phobos/pull/6024#discussion_r160918451): "On append the slice is not the tail so it will reallocate."

### DIFF
--- a/src/rt/util/utf.d
+++ b/src/rt/util/utf.d
@@ -723,8 +723,14 @@ wstring toUTF16(in char[] s)
     wchar[] r;
     size_t slen = s.length;
 
-    r.length = slen;
-    r.length = 0;
+    if (!__ctfe)
+    {
+        // Reserve still does a lot if slen is zero.
+        // Return early for that case.
+        if (0 == slen)
+            return ""w;
+        r.reserve(slen);
+    }
     for (size_t i = 0; i < slen; )
     {
         dchar c = s[i];
@@ -750,8 +756,14 @@ wptr toUTF16z(in char[] s)
     wchar[] r;
     size_t slen = s.length;
 
-    r.length = slen + 1;
-    r.length = 0;
+    if (!__ctfe)
+    {
+        // Reserve still does a lot if slen is zero.
+        // Return early for that case.
+        if (0 == slen)
+            return &"\0"w[0];
+        r.reserve(slen + 1);
+    }
     for (size_t i = 0; i < slen; )
     {
         dchar c = s[i];
@@ -789,8 +801,14 @@ wstring toUTF16(in dchar[] s)
     wchar[] r;
     size_t slen = s.length;
 
-    r.length = slen;
-    r.length = 0;
+    if (!__ctfe)
+    {
+        // Reserve still does a lot if slen is zero.
+        // Return early for that case.
+        if (0 == slen)
+            return ""w;
+        r.reserve(slen);
+    }
     for (size_t i = 0; i < slen; i++)
     {
         encode(r, s[i]);


rainers:
Thanks for your PR. I guess a unittest is not really feasible, is it?

src/rt/util/utf.d
```diff
@@ -723,8 +723,8 @@ wstring toUTF16(in char[] s)
     wchar[] r;
     size_t slen = s.length;
 
-    r.length = slen;
-    r.length = 0;
+    if (!__ctfe)
+        r.reserve(slen);
```
rainers: reserve still does a lot if slen is zero. Maybe better to return early for that case. Same below.

schveiguy: Hm... reserve probably shouldn't be expensive if request is 0.

schveiguy: Probably can just return `null` in that case, or ['\0'].ptr for z versions.

schveiguy: > reserve probably shouldn't be expensive if request is 0

reserve returns the allocated block size, so unfortunately, hard to avoid looking up block info.

schveiguy:
@n8sh in the future, if you could post a link to the PR on bugzilla when you open it, so others don't [duplicate your work](https://github.com/dlang/druntime/pull/2065)

src/rt/util/utf.d
n8sh: Okay will add a check for slen == 0

src/rt/util/utf.d
```diff
@@ -750,8 +756,14 @@ wptr toUTF16z(in char[] s)
...
     size_t slen = s.length;
 
-    r.length = slen + 1;
-    r.length = 0;
+    if (!__ctfe)
+    {
+        // Reserve still does a lot if slen is zero.
+        // Return early for that case.
+        if (0 == slen)
+            return &""w[0];// Terminating null is auto-included at end of string literals.
```
schveiguy: This isn’t right. Return `””w.ptr`

n8sh: If I do that I get a complaint about the code not being `@safe`. I'll assume it's okay to mark `toUTF16z` as `@trusted` since `toUTF16` already is.

schveiguy: Well, it's not `@safe`, as the intention is to return a null-terminated array, which is inherently unsafe. Not only that, but index 0 doesn't exist, so you potentially would have a `RangeError` thrown (it should be what happens, since that is the point of making you index the first element in `@safe` code).

Where is the code that is calling this, complaining that it's not `@safe`?

schveiguy: Oh, I see, it was marked `@safe`. It shouldn't have been. ugh...

schveiguy: I think for now, even though it's not really `@safe`, the best option may be to use:

```D
return &"\0"w[0];
```

DmitryOlshansky: &""w[0] will trigger out of bounds in safe code IIRC.

schveiguy: Well, it's likely this code is *never* triggered. Who is going to toUTF16z an empty string? ;)

n8sh: We could have done `return &"\0"w[0]`

rainers: > Well, it's not @safe, as the intention is to return a null-terminated array, which is inherently unsafe.

I don't think there is a problem: a pointer to something known to exist (in the DATA segment or in GC memory) is fine in @safe code, as this compiles:
```void main() @safe
{
	static const char a = 'a';
	const char* p = &a;
}
```
You are not supposed to do any arithmetics on p and accessing *p should be fine. So "a".ptr could be accepted in @safe code, even "".ptr as literals are guaranteed to be zero-terminated. Am I missing something?

schveiguy: It's technically `@safe`, but disingenuous. The point is to return a null-terminated string, not a pointer to a single `char`.

rainers: > So "a".ptr could be accepted in @safe code

Just noticed that this was still accepted as deprecated in dmd 2.078, but no longer in master. I cannot find it  in https://dlang.org/deprecate.html, though.

wilzbach: https://dlang.org/changelog/pending.html#ptr-safe-end-of-deprecation 
When I made https://github.com/dlang/dmd/pull/7688, I didn't realize it wasn't on the deprecations page. We really need to have more structure in our deprecation process :/

schveiguy:
This one is pretty obvious, regardless of the safety debate, so I'm just going to merge it.

src/rt/util/utf.d
rainers: > It's technically @safe, but disingenuous.

I agree. The syntax `&a[0]` is not nice, especially with literals, but it makes range checks kind of explicit. Let's hope the compiler elides them if possible.

PR #2054, opened 2018-01-23
Improve destroy's documentation

JinShil:
The current documentation for `destroy` says that it sets the object to an "invalid state", but that's not what I've found in my testing, so I changed the documentation to explain what I think is happening, and added an example for illustration.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2878,12 +2878,12 @@ unittest
     assert(postblitRecurseOrder == order);
 }
 
-/++
-    Destroys the given object and puts it in an invalid state. It's used to
-    _destroy an object so that any cleanup which its destructor or finalizer
-    does is done and so that it no longer references any other objects. It does
-    $(I not) initiate a GC cycle or free any GC memory.
-  +/
+/********
+Destroys the given object and sets it back to its initial state. It's used to
+_destroy an object, calling its destructor or finalizer so it no longer
+references any other objects. It does $(I not) initiate a GC cycle or free
+any GC memory.
+*/
 void destroy(T)(T obj) if (is(T == class))
 {
     rt_finalize(cast(void*)obj);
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -2878,12 +2878,43 @@ unittest
...
+
+    // value type demonstration
+    int i;
+    assert(i == 0);           // `i`'s initial state is `0`
+    i = 1;
+    assert(i == 1);           // `i` changed to `1`
+    destroy(i);
+    assert(i == 0);           // `i` is back to its initial state `0`
+}
+----
```
wilzbach: Why don't you use a ddoc-ed unittest (or multiple)?

```
/// reference type demonstration
unittest
{

}

/// value type demonstration
unittest
{

}
```

They have the advantage of being
- checked by the testsuite -> guaranteed the code will work in the future 
- runnable by the reader

JinShil: Changed example to `unittest` blocks.  I hope that's what you meant.

src/object.d
```diff
@@ -2878,12 +2878,12 @@ unittest
     assert(postblitRecurseOrder == order);
 }
 
-/++
-    Destroys the given object and puts it in an invalid state. It's used to
-    _destroy an object so that any cleanup which its destructor or finalizer
-    does is done and so that it no longer references any other objects. It does
-    $(I not) initiate a GC cycle or free any GC memory.
-  +/
+/********
```
wilzbach: We really need to define a best practice for docstrings.
Generally I'm a fan of using only as many `*` / `+` as strictly necessary:

```
/++

+/
```

or

```
/**

*/
```

andralex: Same here. Not a fan of lots of stars or stars at the beginning of each comment line.

JinShil: I was following @andralex's prior suggestion here: https://github.com/dlang/dmd/pull/6835#discussion_r158275983

> Let's use https://github.com/dlang/phobos/blob/master/std/string.d#L2254-L2267 for new code, thanks.

Anyway, if it's important, please make a PR to the style guide with preferred dos and don'ts

wilzbach: https://github.com/dlang/dlang.org/pull/2129

PR #2056, opened 2018-01-23
Remove pointless version(unittest)

wilzbach:


schveiguy:
There's another one a few more lines above.

schveiguy:
FIFY :)

schveiguy:
Why is circleci failing with that core.time test? Seems very odd.

wilzbach:
Thanks! See https://github.com/dlang/druntime/pull/2048

schveiguy:
> See #2048

Whew! I thought maybe the `version(unittest)` removal was causing it 😄 . Still very strange.

PR #2057, opened 2018-01-24
GC Initialized on First Alloc

somzzz:
This PR removes the requirement of having the GC initialized at startup. Instead, ```gc_init``` is called when the GC is first invoked.

### DIFF
--- a/src/gc/impl/proto/gc.d
+++ b/src/gc/impl/proto/gc.d
@@ -1,0 +1,192 @@
+
+module gc.impl.proto.gc;
+
+import gc.config;
+import gc.gcinterface;
+
+import rt.util.container.array;
+
+import cstdlib = core.stdc.stdlib : calloc, free, malloc, realloc;
+static import core.memory;
+
+extern (C) void onOutOfMemoryError(void* pretend_sideffect = null) @trusted pure nothrow @nogc; /* dmd @@@BUG11461@@@ */
+
+private
+{
+    extern (C) void gc_init_nothrow() nothrow @nogc;
+    extern (C) void gc_term();
+
+    extern (C) void gc_enable() nothrow;
+    extern (C) void gc_disable() nothrow;
+
+    extern (C) void*    gc_malloc( size_t sz, uint ba = 0, const TypeInfo = null ) pure nothrow;
+    extern (C) void*    gc_calloc( size_t sz, uint ba = 0, const TypeInfo = null ) pure nothrow;
+    extern (C) BlkInfo gc_qalloc( size_t sz, uint ba = 0, const TypeInfo = null ) pure nothrow;
+    extern (C) void*    gc_realloc( void* p, size_t sz, uint ba = 0, const TypeInfo = null ) pure nothrow;
+    extern (C) size_t   gc_reserve( size_t sz ) nothrow;
+
+    extern (C) void gc_addRange( void* p, size_t sz, const TypeInfo ti = null ) nothrow @nogc;
+    extern (C) void gc_addRoot( void* p ) nothrow @nogc;
+}
+
+class ProtoGC : GC
+{
+    __gshared Array!Root roots;
+    __gshared Array!Range ranges;
+
+    this()
+    {
+    }
+
+    void Dtor()
+    {
+    }
+
+    void enable()
+    {
+        gc_init_nothrow();
+        gc_enable();
+    }
+
+    void disable()
+    {
+        gc_init_nothrow();
+        gc_disable();
+    }
+
+    void collect() nothrow
+    {
+    }
+
+    void collectNoStack() nothrow
+    {
+    }
+
+    void minimize() nothrow
+    {
+    }
+
+    uint getAttr(void* p) nothrow
+    {
+        return 0;
+    }
+
+    uint setAttr(void* p, uint mask) nothrow
+    {
+        return 0;
+    }
+
+    uint clrAttr(void* p, uint mask) nothrow
+    {
+        return 0;
+    }
+
+    void* malloc(size_t size, uint bits, const TypeInfo ti) nothrow
+    {
+        gc_init_nothrow();
+        return gc_malloc(size, bits, ti);
+    }
+
+    BlkInfo qalloc(size_t size, uint bits, const TypeInfo ti) nothrow
+    {
+        gc_init_nothrow();
+        return gc_qalloc(size, bits, ti);
+    }
+
+    void* calloc(size_t size, uint bits, const TypeInfo ti) nothrow
+    {
+        gc_init_nothrow();
+        return gc_calloc(size, bits, ti);
+    }
+
+    void* realloc(void* p, size_t size, uint bits, const TypeInfo ti) nothrow
+    {
+        gc_init_nothrow();
+        return gc_realloc(p, size, bits, ti);
+    }
+
+    size_t extend(void* p, size_t minsize, size_t maxsize, const TypeInfo ti) nothrow
+    {
+        return 0;
+    }
+
+    size_t reserve(size_t size) nothrow
+    {
+        gc_init_nothrow();
+        return reserve(size);
+    }
+
+    void free(void* p) nothrow @nogc
+    {
+        if (p) assert(false, "Invalid memory deallocation");
+    }
+
+    void* addrOf(void* p) nothrow @nogc
+    {
+        return null;
+    }
+
+    size_t sizeOf(void* p) nothrow @nogc
+    {
+        return 0;
+    }
+
+    BlkInfo query(void* p) nothrow
+    {
+        return BlkInfo.init;
+    }
+
+    core.memory.GC.Stats stats() nothrow
+    {
+        return typeof(return).init;
+    }
+
+    void addRoot(void* p) nothrow @nogc
+    {
+        gc_init_nothrow();
+        gc_addRoot(p);
+    }
+
+    void removeRoot(void* p) nothrow @nogc
+    {
+    }
+
+    @property RootIterator rootIter() return @nogc
+    {
+        return &rootsApply;
+    }
+
+    private int rootsApply(scope int delegate(ref Root) nothrow dg)
+    {
+        return 0;
+    }
+
+    void addRange(void* p, size_t sz, const TypeInfo ti = null) nothrow @nogc
+    {
+        gc_init_nothrow();
+        gc_addRange(p, sz, ti);
+    }
+
+    void removeRange(void* p) nothrow @nogc
+    {
+    }
+
+    @property RangeIterator rangeIter() return @nogc
+    {
+        return &rangesApply;
+    }
+
+    private int rangesApply(scope int delegate(ref Range) nothrow dg)
+    {
+        return 0;
+    }
+
+    void runFinalizers(in void[] segment) nothrow
+    {
+    }
+
+    bool inFinalizer() nothrow
+    {
+        return false;
+    }
+}
--- a/src/gc/impl/proto/gc.d
+++ b/src/gc/impl/proto/gc.d
@@ -31,8 +31,27 @@ private
 
 class ProtoGC : GC
 {
-    __gshared Array!Root roots;
-    __gshared Array!Range ranges;
+    Array!Root roots;
+    Array!Range ranges;
+
+    // Call this function when initializing the real GC
+    // upon ProtoGC term. This function should be called
+    // after the real GC is in place.
+    void term()
+    {
+        // Transfer all ranges
+        foreach (ref r; ranges)
+        {
+            // Range(p, p + sz, cast() ti)
+            gc_addRange(r.pbot, r.ptop - r.pbot, r.ti);
+        }
+
+        // Transfer all roots
+        foreach (ref r; roots)
+        {
+            gc_addRoot(r.proot);
+        }
+    }
 
     this()
     {
@@ -141,14 +160,24 @@ class ProtoGC : GC
         return typeof(return).init;
     }
 
+
     void addRoot(void* p) nothrow @nogc
     {
-        gc_init_nothrow();
-        gc_addRoot(p);
+        roots.insertBack(Root(p));
     }
 
     void removeRoot(void* p) nothrow @nogc
     {
+        foreach (ref r; roots)
+        {
+            if (r is p)
+            {
+                r = roots.back;
+                roots.popBack();
+                return;
+            }
+        }
+        assert(false);
     }
 
     @property RootIterator rootIter() return @nogc
@@ -158,17 +187,31 @@ class ProtoGC : GC
 
     private int rootsApply(scope int delegate(ref Root) nothrow dg)
     {
+        foreach (ref r; roots)
+        {
+            if (auto result = dg(r))
+                return result;
+        }
         return 0;
     }
 
     void addRange(void* p, size_t sz, const TypeInfo ti = null) nothrow @nogc
     {
-        gc_init_nothrow();
-        gc_addRange(p, sz, ti);
+        ranges.insertBack(Range(p, p + sz, cast() ti));
     }
 
     void removeRange(void* p) nothrow @nogc
     {
+        foreach (ref r; ranges)
+        {
+            if (r.pbot is p)
+            {
+                r = ranges.back;
+                ranges.popBack();
+                return;
+            }
+        }
+        assert(false);
     }
 
     @property RangeIterator rangeIter() return @nogc
@@ -178,6 +221,11 @@ class ProtoGC : GC
 
     private int rangesApply(scope int delegate(ref Range) nothrow dg)
     {
+        foreach (ref r; ranges)
+        {
+            if (auto result = dg(r))
+                return result;
+        }
         return 0;
     }
 
--- a/src/gc/proxy.d
+++ b/src/gc/proxy.d
@@ -15,6 +15,7 @@ module gc.proxy;
 
 import gc.impl.conservative.gc;
 import gc.impl.manual.gc;
+import gc.impl.proto.gc;
 import gc.config;
 import gc.gcinterface;
 
@@ -25,67 +26,44 @@ private
     static import core.memory;
     alias BlkInfo = core.memory.GC.BlkInfo;
 
-    extern (C) void thread_init();
-    extern (C) void thread_term();
-
     import core.internal.spinlock;
     static SpinLock instanceLock;
 
-    /*
-    * tlsInstance is a hack to ensure that only the thread initializing the GC
-    * can use GC functions before the initialization is completed. When it is completed,
-    * the GC object created is atomically stored in the __gshared instance.
-    *
-    * This is necessary because calling thead_init is part of the gc_init routine.
-    * thread_init, in turn, calls GC.enable, GC.disable and GC.malloc (through the new keyword)
-    */
-    static GC tlsInstance;
-
-    __gshared GC instance;
+    __gshared bool isInstanceInit = false;
+    __gshared GC instance = new ProtoGC();
     __gshared GC proxiedGC; // used to iterate roots of Windows DLLs
 }
 
 extern (C)
 {
-
     void gc_init()
     {
         import core.atomic : atomicLoad, atomicStore, MemoryOrder;
 
-        auto pinstance = cast(shared(GC)*) &instance;
-
-        // using double-checked locking
-        if (atomicLoad!(MemoryOrder.acq)(*pinstance) is null)
+        instanceLock.lock();
+        if (!isInstanceInit)
         {
-            instanceLock.lock();
-            scope(exit) instanceLock.unlock();
+            auto protoInstance = instance;
+            config.initialize();
+            ManualGC.initialize(instance);
+            ConservativeGC.initialize(instance);
 
-            if (atomicLoad!(MemoryOrder.raw)(*pinstance) is null)
+            if (instance == protoInstance)
             {
-                config.initialize();
-                ManualGC.initialize(tlsInstance);
-                ConservativeGC.initialize(tlsInstance);
+                import core.stdc.stdio : fprintf, stderr;
+                import core.stdc.stdlib : exit;
 
-                if (tlsInstance is null)
-                {
-                    import core.stdc.stdio : fprintf, stderr;
-                    import core.stdc.stdlib : exit;
+                fprintf(stderr, "No GC was initialized, please recheck the name of the selected GC ('%.*s').\n", cast(int)config.gc.length, config.gc.ptr);
+                instanceLock.unlock();
+                exit(1);
 
-                    fprintf(stderr, "No GC was initialized, please recheck the name of the selected GC ('%.*s').\n", cast(int)config.gc.length, config.gc.ptr);
-                    instanceLock.unlock();
-                    exit(1);
-
-                    // Shouldn't get here.
-                    assert(0);
-                }
-
-                // NOTE: The GC must initialize the thread library
-                //       before its first collection.
-                thread_init();
-
-                atomicStore!(MemoryOrder.rel)(*pinstance, cast(shared GC) tlsInstance);
+                // Shouldn't get here.
+                assert(0);
             }
+
+            isInstanceInit = true;
         }
+        instanceLock.unlock();
     }
 
     void gc_init_nothrow() nothrow
@@ -111,170 +89,130 @@ extern (C)
         // NOTE: Due to popular demand, this has been re-enabled.  It still has
         //       the problems mentioned above though, so I guess we'll see.
 
-        if (instance !is null)
-        {
-            instance.collectNoStack(); // not really a 'collect all' -- still scans
-                                        // static data area, roots, and ranges.
+        instance.collectNoStack();  // not really a 'collect all' -- still scans
+                                    // static data area, roots, and ranges.
 
-            ManualGC.finalize(instance);
-            ConservativeGC.finalize(instance);
-        }
+        ManualGC.finalize(instance);
+        ConservativeGC.finalize(instance);
     }
 
     void gc_enable()
     {
-        if (tlsInstance)
-        {
-            tlsInstance.enable();
-            return;
-        }
-        gc_init_nothrow();
         instance.enable();
     }
 
     void gc_disable()
     {
-        if (tlsInstance)
-        {
-            tlsInstance.disable();
-            return;
-        }
-        gc_init_nothrow();
         instance.disable();
     }
 
     void gc_collect() nothrow
     {
-        if (instance is null) return;
         instance.collect();
     }
 
     void gc_minimize() nothrow
     {
-        gc_init_nothrow();
         instance.minimize();
     }
 
     uint gc_getAttr( void* p ) nothrow
     {
-        if (instance is null) return 0;
         return instance.getAttr(p);
     }
 
     uint gc_setAttr( void* p, uint a ) nothrow
     {
-        assert(instance !is null);
         return instance.setAttr(p, a);
     }
 
     uint gc_clrAttr( void* p, uint a ) nothrow
     {
-        assert(instance !is null);
         return instance.clrAttr(p, a);
     }
 
     void* gc_malloc( size_t sz, uint ba = 0, const TypeInfo ti = null ) nothrow
     {
-        if (tlsInstance)
-            return tlsInstance.malloc(sz, ba, ti);
-
-        gc_init_nothrow();
         return instance.malloc(sz, ba, ti);
     }
 
     BlkInfo gc_qalloc( size_t sz, uint ba = 0, const TypeInfo ti = null ) nothrow
     {
-        gc_init_nothrow();
         return instance.qalloc( sz, ba, ti );
     }
 
     void* gc_calloc( size_t sz, uint ba = 0, const TypeInfo ti = null ) nothrow
     {
-        gc_init_nothrow();
         return instance.calloc( sz, ba, ti );
     }
 
     void* gc_realloc( void* p, size_t sz, uint ba = 0, const TypeInfo ti = null ) nothrow
     {
-        gc_init_nothrow();
         return instance.realloc( p, sz, ba, ti );
     }
 
     size_t gc_extend( void* p, size_t mx, size_t sz, const TypeInfo ti = null ) nothrow
     {
-        assert(instance !is null);
         return instance.extend( p, mx, sz,ti );
     }
 
     size_t gc_reserve( size_t sz ) nothrow
     {
-        gc_init_nothrow();
         return instance.reserve( sz );
     }
 
     void gc_free( void* p ) nothrow @nogc
     {
-        assert(instance !is null);
         return instance.free( p );
     }
 
     void* gc_addrOf( void* p ) nothrow @nogc
     {
-        if (instance is null) return null;
         return instance.addrOf( p );
     }
 
     size_t gc_sizeOf( void* p ) nothrow @nogc
     {
-        if (instance is null) return 0;
         return instance.sizeOf( p );
     }
 
     BlkInfo gc_query( void* p ) nothrow
     {
-        if (instance is null) return BlkInfo.init;
         return instance.query( p );
     }
 
     core.memory.GC.Stats gc_stats() nothrow
     {
-        gc_init_nothrow();
         return instance.stats();
     }
 
-    void gc_addRoot( void* p ) nothrow
+    void gc_addRoot( void* p ) nothrow @nogc
     {
-        gc_init_nothrow();
         return instance.addRoot( p );
     }
 
-    void gc_addRange( void* p, size_t sz, const TypeInfo ti = null ) nothrow
+    void gc_addRange( void* p, size_t sz, const TypeInfo ti = null ) nothrow @nogc
     {
-        gc_init_nothrow();
         return instance.addRange( p, sz, ti );
     }
 
     void gc_removeRoot( void* p ) nothrow
     {
-        assert(instance !is null);
         return instance.removeRoot( p );
     }
 
     void gc_removeRange( void* p ) nothrow
     {
-        assert(instance !is null);
         return instance.removeRange( p );
     }
 
     void gc_runFinalizers( in void[] segment ) nothrow
     {
-        assert(instance !is null);
         return instance.runFinalizers( segment );
     }
 
     bool gc_inFinalizer() nothrow
     {
-        if (instance is null) return false;
         return instance.inFinalizer();
     }
 
--- a/src/gc/proxy.d
+++ b/src/gc/proxy.d
@@ -38,8 +38,6 @@ extern (C)
 {
     void gc_init()
     {
-        import core.atomic : atomicLoad, atomicStore, MemoryOrder;
-
         instanceLock.lock();
         if (!isInstanceInit)
         {
@@ -48,7 +46,7 @@ extern (C)
             ManualGC.initialize(instance);
             ConservativeGC.initialize(instance);
 
-            if (instance == protoInstance)
+            if (instance is protoInstance)
             {
                 import core.stdc.stdio : fprintf, stderr;
                 import core.stdc.stdlib : exit;
@@ -61,6 +59,8 @@ extern (C)
                 assert(0);
             }
 
+            // Transfer all ranges and roots to the real GC.
+            (cast(ProtoGC) protoInstance).term();
             isInstanceInit = true;
         }
         instanceLock.unlock();
@@ -89,11 +89,14 @@ extern (C)
         // NOTE: Due to popular demand, this has been re-enabled.  It still has
         //       the problems mentioned above though, so I guess we'll see.
 
-        instance.collectNoStack();  // not really a 'collect all' -- still scans
-                                    // static data area, roots, and ranges.
+        if (isInstanceInit)
+        {
+            instance.collectNoStack();  // not really a 'collect all' -- still scans
+                                        // static data area, roots, and ranges.
 
-        ManualGC.finalize(instance);
-        ConservativeGC.finalize(instance);
+            ManualGC.finalize(instance);
+            ConservativeGC.finalize(instance);
+        }
     }
 
     void gc_enable()
--- a/src/gc/proxy.d
+++ b/src/gc/proxy.d
@@ -50,17 +50,17 @@ extern (C)
 
     void gc_init()
     {
-        import core.atomic : atomicLoad, atomicStore;
+        import core.atomic : atomicLoad, atomicStore, MemoryOrder;
 
         auto pinstance = cast(shared(GC)*) &instance;
 
         // using double-checked locking
-        if (atomicLoad(*pinstance) is null)
+        if (atomicLoad!(MemoryOrder.acq)(*pinstance) is null)
         {
             instanceLock.lock();
             scope(exit) instanceLock.unlock();
 
-            if (atomicLoad(*pinstance) is null)
+            if (atomicLoad!(MemoryOrder.raw)(*pinstance) is null)
             {
                 config.initialize();
                 ManualGC.initialize(tlsInstance);
@@ -83,7 +83,7 @@ extern (C)
                 //       before its first collection.
                 thread_init();
 
-                atomicStore(*pinstance, cast(shared GC) tlsInstance);
+                atomicStore!(MemoryOrder.rel)(*pinstance, cast(shared GC) tlsInstance);
             }
         }
     }
--- a/src/gc/proxy.d
+++ b/src/gc/proxy.d
@@ -25,28 +25,78 @@ private
     static import core.memory;
     alias BlkInfo = core.memory.GC.BlkInfo;
 
+    extern (C) void thread_init();
+    extern (C) void thread_term();
+
+    import core.internal.spinlock;
+    static SpinLock instanceLock;
+
+    /*
+    * tlsInstance is a hack to ensure that only the thread initializing the GC
+    * can use GC functions before the initialization is completed. When it is completed,
+    * the GC object created is atomically stored in the __gshared instance.
+    *
+    * This is necessary because calling thead_init is part of the gc_init routine.
+    * thread_init, in turn, calls GC.enable, GC.disable and GC.malloc (through the new keyword)
+    */
+    static GC tlsInstance;
+
     __gshared GC instance;
     __gshared GC proxiedGC; // used to iterate roots of Windows DLLs
-
 }
-
 
 extern (C)
 {
 
     void gc_init()
     {
-        config.initialize();
-        ManualGC.initialize(instance);
-        ConservativeGC.initialize(instance);
-        if (instance is null)
-        {
-            import core.stdc.stdio : fprintf, stderr;
-            import core.stdc.stdlib : exit;
+        import core.atomic : atomicLoad, atomicStore;
 
-            fprintf(stderr, "No GC was initialized, please recheck the name of the selected GC ('%.*s').\n", cast(int)config.gc.length, config.gc.ptr);
-            exit(1);
+        auto pinstance = cast(shared(GC)*) &instance;
+
+        // using double-checked locking
+        if (atomicLoad(*pinstance) is null)
+        {
+            instanceLock.lock();
+            scope(exit) instanceLock.unlock();
+
+            if (atomicLoad(*pinstance) is null)
+            {
+                config.initialize();
+                ManualGC.initialize(tlsInstance);
+                ConservativeGC.initialize(tlsInstance);
+
+                if (tlsInstance is null)
+                {
+                    import core.stdc.stdio : fprintf, stderr;
+                    import core.stdc.stdlib : exit;
+
+                    fprintf(stderr, "No GC was initialized, please recheck the name of the selected GC ('%.*s').\n", cast(int)config.gc.length, config.gc.ptr);
+                    instanceLock.unlock();
+                    exit(1);
+
+                    // Shouldn't get here.
+                    assert(0);
+                }
+
+                // NOTE: The GC must initialize the thread library
+                //       before its first collection.
+                thread_init();
+
+                atomicStore(*pinstance, cast(shared GC) tlsInstance);
+            }
         }
+    }
+
+    void gc_init_nothrow() nothrow
+    {
+        scope(failure)
+        {
+            import core.internal.abort;
+            abort("Cannot initialize the garbage collector.\n");
+            assert(0);
+        }
+        gc_init();
     }
 
     void gc_term()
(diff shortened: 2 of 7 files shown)


somzzz:
Fail seems unrelated https://github.com/dlang/druntime/pull/2048
@andralex

src/gc/proxy.d
```diff
@@ -28,32 +28,67 @@ private
...
 
+    import core.internal.spinlock;
+    static SpinLock instanceLock;
+
     __gshared GC instance;
     __gshared GC proxiedGC; // used to iterate roots of Windows DLLs
-
 }
 
+alias GCInitNoThrow = void function() nothrow @nogc;
```
marler8997: this appears to be unused

src/gc/proxy.d
```diff
@@ -28,32 +28,67 @@ private
...
+
+        auto pinstance = cast(shared(GC)*) &instance;
 
-            fprintf(stderr, "No GC was initialized, please recheck the name of the selected GC ('%.*s').\n", cast(int)config.gc.length, config.gc.ptr);
-            exit(1);
+        if (atomicLoad(*pinstance) is null)
+        {
+            instanceLock.lock();
+            scope(exit) instanceLock.unlock();
+            {
```
marler8997: unnecessary extra scope?

src/gc/proxy.d
```diff
@@ -28,32 +28,67 @@ private
...
-        {
-            import core.stdc.stdio : fprintf, stderr;
-            import core.stdc.stdlib : exit;
+        import core.atomic : atomicLoad, atomicStore;
+
+        auto pinstance = cast(shared(GC)*) &instance;
 
-            fprintf(stderr, "No GC was initialized, please recheck the name of the selected GC ('%.*s').\n", cast(int)config.gc.length, config.gc.ptr);
-            exit(1);
+        if (atomicLoad(*pinstance) is null)
```
marler8997: whenever I do double-checked locking, I like to put a comment :)
`// using double-checked locking`

src/gc/proxy.d
```diff
@@ -28,32 +28,67 @@ private
...
+                    {
+                        import core.stdc.stdio : fprintf, stderr;
+                        import core.stdc.stdlib : exit;
+
+                        fprintf(stderr, "No GC was initialized, please recheck the name of the selected GC ('%.*s').\n", cast(int)config.gc.length, config.gc.ptr);
+                        instanceLock.unlock();
+                        exit(1);
+                    }
+
+                    atomicStore(*pinstance, cast(shared GC)auxInstance);
```
marler8997: Race condition?  setting `auxInstance` to "not null" should be the last thing done correct?  Otherwise another thread could come in, see that `auxInstance` is not null and then continue before `thread_init` has finished.   However, my comment is assuming that `thread_init` should be finished before another thread can start using the GC, is my assumption incorrect?

src/gc/proxy.d
```diff
@@ -28,32 +28,67 @@ private
...
+                    ManualGC.initialize(auxInstance);
+                    ConservativeGC.initialize(auxInstance);
+
+                    if (auxInstance is null)
+                    {
+                        import core.stdc.stdio : fprintf, stderr;
+                        import core.stdc.stdlib : exit;
+
+                        fprintf(stderr, "No GC was initialized, please recheck the name of the selected GC ('%.*s').\n", cast(int)config.gc.length, config.gc.ptr);
+                        instanceLock.unlock();
```
marler8997: You've unlocked twice!  (`scope(exit) instanceLock.unlock();` above)

src/gc/proxy.d
```diff
@@ -28,32 +28,67 @@ private
...
+        auto pinstance = cast(shared(GC)*) &instance;
 
-            fprintf(stderr, "No GC was initialized, please recheck the name of the selected GC ('%.*s').\n", cast(int)config.gc.length, config.gc.ptr);
-            exit(1);
+        if (atomicLoad(*pinstance) is null)
+        {
+            instanceLock.lock();
+            scope(exit) instanceLock.unlock();
+            {
+                if (atomicLoad(*pinstance) is null)
```
marler8997: I don't think `atomicLoad` is necessary once inside the lock, since the values is guaranteed not to be modified while a thread is inside the lock.

src/gc/proxy.d
```diff
@@ -28,32 +28,67 @@ private
...
+            }
         }
+    }
 
-        // NOTE: The GC must initialize the thread library
-        //       before its first collection.
-        thread_init();
+    void gc_init_nothrow() nothrow
+    {
+        try
```
marler8997: Rather than incurring the overrhead of a try/catch for every `gc_init_nothrow`, this should probably go inside the double checked lock.  That way you only have to pay the price of a try/catch once, when you initialize the GC.

src/gc/proxy.d
```diff
@@ -68,132 +103,159 @@ extern (C)
         // NOTE: Due to popular demand, this has been re-enabled.  It still has
         //       the problems mentioned above though, so I guess we'll see.
 
-        instance.collectNoStack(); // not really a 'collect all' -- still scans
-                                    // static data area, roots, and ranges.
+        if (instance !is null)
```
marler8997: I'm not familiar with who calls `gc_term`.  Would it be better to make this synchronized like you have done in `gc_init`?  Or is this only going to be called by a single thread?

marler8997:
Thanks for making this change.  I like the idea.

src/gc/proxy.d
```diff
@@ -68,132 +103,159 @@ extern (C)
...
     void gc_enable()
     {
+        gc_init_nothrow();
         instance.enable();
     }
 
     void gc_disable()
     {
+        assert(instance !is null);
         instance.disable();
```
quickfur: Will this break code that calls `GC.disable` at the beginning of the program?  Sometimes it's useful to do this and schedule your own calls to `GC.collect`; I've done this before to optimize GC performance.

Would it be better to also initialize the GC in `gc_enable` / `gc_disable`, in addition to doing it upon the first memory allocation? IOW, initialize the GC the first time it's used in some way, whether it's a memory allocation or enable/disable, etc..

src/gc/proxy.d
```diff
@@ -28,32 +28,67 @@ private
...
-        config.initialize();
-        ManualGC.initialize(instance);
-        ConservativeGC.initialize(instance);
-        if (instance is null)
-        {
-            import core.stdc.stdio : fprintf, stderr;
-            import core.stdc.stdlib : exit;
+        import core.atomic : atomicLoad, atomicStore;
+
+        auto pinstance = cast(shared(GC)*) &instance;
```
marler8997: Note that if you remove the second `atomicLoad` (see comment below).  I believe you could just do
`if (atomicLoad(instance))`
and you don't need the `pinstance` variable.

src/gc/proxy.d
```diff
@@ -28,32 +28,67 @@ private
...
+                        fprintf(stderr, "No GC was initialized, please recheck the name of the selected GC ('%.*s').\n", cast(int)config.gc.length, config.gc.ptr);
+                        instanceLock.unlock();
+                        exit(1);
+                    }
+
+                    atomicStore(*pinstance, cast(shared GC)auxInstance);
+
+                    // NOTE: The GC must initialize the thread library
+                    //       before its first collection.
+                    thread_init();
```
MartinNowak: What do you indent to do about [`Thread.getThis`](https://dlang.org/phobos/core_thread.html#.Thread.getThis)? It is currently a GC allocated instance that is eagerly created in [`thread_init`](https://github.com/dlang/druntime/blob/a66f9fa78f2e295146fb2b3fc8e567983c9e9aed/src/core/thread.d#L2023).

src/gc/proxy.d
MartinNowak: Indeed, use sth. similar to [`initOnce`](https://github.com/dlang/phobos/blob/da00e8f3908d221008d7c550bfed97ac9628afb2/std/concurrency.d#L2569-L2589).
Instead of a separate `flag` variable, you could directly load/store the shared `instance` variable. Also seems like the spinLock would be fine to be used as mutex.

src/gc/proxy.d
MartinNowak: Depends on the startup, there could be multiple external "C" threads calling `thread_attachThis` concurrently.

src/gc/proxy.d
MartinNowak: I think the purpose here was just to trick `nothrow`.
A shorter idiom for that is the following.
```d
void foo() nothrow
{
    scope (failure) assert(0, "unexpected exception");
    stmt; // not yet nothrow annotated
}
```

marler8997: From what I understand, `try/catch` and `scope(failure)` incur some amount of overhead.  By moving this inside the lock we only have to incur it once.  I still recommend doing so.

rainers:
- This is pretty performance sensitive code, please benchmark the GC suite with benchmark/runbench

- We have removed null-checks for the instance not too long ago (though they have been used for a different purpose). Adding them back with atomics is probably not so great. An alternative would be to implement the GCInterface that switches the `instance` global to the standard GC if necessary.

- all the `assert(instance !is null)` can probably do/return something sensible knowing that no GC memory has been allocated.

- I also suspect that this change will probably not have an immediate effect because other init functions are likely to access the GC, e.g. initStaticDataGC calls GC.addRange.

src/gc/proxy.d
marler8997: Based on @MartinNowak's suggestions, `initOnce` looks like it could work here so long as you use a mutex and wrap the `init` argument in a tryCatch.

src/gc/proxy.d
wilzbach: I wonder whether you have tried the following trick of using a variable to cache and thus providing as fast, hot-path for the default case:

```
__gshared gcHasBeenInitialized;
if (!gcHasBeenInitialized)
{
 // do your normal locking here
 gcHasBeenInitialized = true;
}
```

TLS variables could be faster, but you could benchmark this for [dynamic modules too](http://david-grs.github.io/tls_performance_overhead_cost_linux/).

src/gc/proxy.d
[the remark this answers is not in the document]
andralex: (if used)

src/gc/proxy.d
andralex: @marler8997 I think that won't work because atomicLoad expects a ref to a shared object

src/gc/proxy.d
andralex: yes please delete

src/gc/proxy.d
andralex: @MartinNowak so what's the right course of action here?

src/gc/proxy.d
andralex: That may be why *pinstance had to be set just above! Seems fragile. Perhaps we can not use the GC for getThis?

src/gc/proxy.d
klickverbot: That's what that is, I believe, only fixed to be correct.

src/gc/proxy.d
klickverbot: initOnce is in Phobos, which is the reason I didn't suggest it earlier. We cut put something like it in druntime, of course, for internal use.

src/gc/proxy.d
marler8997: Would love to see a performance comparison on TLS vs atomic double-checked locking.

klickverbot: I'd expect that to be hard to do as a microbenchmark on x86 in a sensible fashion, since it depends a lot on non-local considerations. If there is no contention on the cache line, I'd expect double checked locking to be at least as fast (acq loads are just simple movs).

src/gc/proxy.d
somzzz: ```gc_term``` is called in ```dmain2.d->rt_term``` after ```thread_joinAll```.
https://github.com/dlang/druntime/blob/master/src/rt/dmain2.d#L223

src/gc/proxy.d
somzzz: For some reason, if unlock is not called before ```exit(1)```, this test hangs: ```druntime/test/exceptions/src/unknown_gc.d```

src/gc/proxy.d
somzzz: If there is a use case for it, then yes - it should be initialized.
Which of these methods do you think should initialize the GC?

src/gc/proxy.d
MartinNowak: dmd is fairly bad at optimizing initOnce, so doing it manually seems fine here.

src/gc/proxy.d
MartinNowak: Double-checked locking should be faster, TLS reads go through an extra indirection and can be very slow if TLS is emulated. Just reading a shared variable by multiple threads doesn't cost much (except maybe for the non-inlined `atomicRead!acq` overhead with dmd).

src/gc/proxy.d
MartinNowak: Definitely GC.disable as it's fairly common. Don't think anyone has explicitly called `GC.enable` yet. Even better if we could memorize GC.disable and lazily initialize and disable the GC if it's used later on. It would be weird to initialize a GC just to disable it.

src/gc/proxy.d
MartinNowak: You can currently retain a Thread object longer than the threads lives, e.g. there is even a Thread.join methods, so there isn't a clear owner of that instance. Not too easy to avoid the GC here.

src/gc/proxy.d
marler8997: uh oh, that may be a bug with D's scope(exit) implementation (anyone correct me if I'm wrong).  Given this, I would suggest not using `scope(exit)` here and instead insert `instancelock.unlock()` at every "exit point".  There should just be 2, just before calling `exit` and at the end of the block.

src/gc/proxy.d
schveiguy: Hm... just had a weird idea.

What if the instance is initialized with a GC that when first used, figures out what GC to allocate, sets the 'instance' to the new one, and calls the appropriate function on it. Then we can handle things like a bool to track the disabling without having to initialize until someone tries to call malloc, etc.

Then once the *real* instance is in place, there is no longer any checking for initialization.

You'd still need to atomically load the instance every time you use it, unfortunately.

src/gc/proxy.d
wilzbach: FWIW we have a nice language feature for exactly this:

```d
synchronized (instanceLock)
{
   // code here
}
```

src/gc/proxy.d
MartinNowak: The mutual dependency between gc_init and thread_init isn't properly resolved.
```d
if (atomicLoad(*pinstance) is null))
{
    instanceLock.lock();
    scope (exit) instanceLock.unlock();
    if (atomicLoad(*pinstance) is null))
    {
        // init
        // ...
        atomicStore(*pinstance, cast(shared GC)auxInstance);

        // thread_init <- should be moved before storing pInstance or some other thread could start to use the GC before thread_init finishes
    }
}
```
The problem is that thread_init depends on the GC, but any GC usage depends on thread_init (for suspending).
Let's figure out whether we can avoid the GC for Thread.getThis, if not then pass `auxInstance` directly to thread_init.

MartinNowak: And BTW, use weaker (and much cheaper) orderings.
```d
if (atomicLoad!(MemoryOrder.acq)(*pInstance) is null)
{
    // lock...
    if (atomicLoad!(MemoryOrder.acq)(*pInstance) is null)
    {
        // init...
        atomicStore!(MemoryOrder.rel)(*pInstance, auxInstance);
    }
}

**Comment by MartinNowak** [comment]
That's the main blocker to make this work https://github.com/dlang/druntime/pull/2057/files#r163907886.

File: src/gc/proxy.d (hunk 5 above)
Comment (marler8997) [reply src/gc/proxy.d]: If `scope(exit)` doesn't work with the `exit` function, my guess is that this might suffer from the same bug.  @somzzz can you try this out and confirm?

File: src/gc/proxy.d (hunk 4 above)
Comment (marler8997) [reply src/gc/proxy.d]: The second `atomidLoad` can be `raw` correct? (see https://github.com/dlang/phobos/pull/6063)
```D
if (atomicLoad!(MemoryOrder.acq)(*pInstance) is null)
{
    // lock...
    if (atomicLoad!(MemoryOrder.raw)(*pInstance) is null)
    {
        // init...
        atomicStore!(MemoryOrder.rel)(*pInstance, auxInstance);
    }
}
```

src/gc/proxy.d
quickfur: I like this idea.

schveiguy:
This may have been lost in the noise, so I'll link to it here:

https://github.com/dlang/druntime/pull/2057#discussion_r163907083

Might be a viable option that eliminates some of the possible performance drawbacks.

marler8997:
@schveiguy I'm actually working an an alternative that explores your idea.  Will make a PR soon if it turns out well.

src/gc/proxy.d
schveiguy: > pass `auxInstance` directly to thread_init

This seems like the correct answer.

src/gc/proxy.d
schveiguy: Wait, `exit()` doesn't return, so it can't exit the scope. This seems like the code is correct to me...

`scope(exit)` only works if the scope is exited, via stack unwinding or normal execution. Calling exit function is neither.

src/gc/proxy.d
```diff
@@ -28,32 +28,67 @@ private
...
+                    ConservativeGC.initialize(auxInstance);
+
+                    if (auxInstance is null)
+                    {
+                        import core.stdc.stdio : fprintf, stderr;
+                        import core.stdc.stdlib : exit;
+
+                        fprintf(stderr, "No GC was initialized, please recheck the name of the selected GC ('%.*s').\n", cast(int)config.gc.length, config.gc.ptr);
+                        instanceLock.unlock();
+                        exit(1);
```
schveiguy: You should add an `assert(0);` here to note that `exit(1)` shouldn't return.

src/gc/proxy.d
andralex: @somzzz let's conservatively leave atomicLoad here and review it in a future pass thx

src/gc/proxy.d
andralex: keep scope(exit) and also unlock explicitly before calling exit thx

src/gc/proxy.d
andralex: @MartinNowak let's make it most conservative and optimize in a subsequent step thx

src/gc/proxy.d
andralex: >Don't make things too complicated though, when in doubt just initialize the GC on any access.

let's do that now, refine later

src/gc/proxy.d
klickverbot: Yep, precisely. On x86_64 TLS can pretty much only be faster if the indirection is cheap (i.e. static TLS model) and the other variant is burdened down by concurrent writes to other data on the same cache line (hence my above comment). I wouldn't worry about it.

src/gc/proxy.d
klickverbot: Hmm, it does sound neat conceptually. As far as performance is concerned, the branch is followed by an indirect jump through the vtable anyway, so I wouldn't expect a massive difference by getting rid of the former.

andralex: @schveiguy it's nice - the almost-null-object pattern? :) - but a bit too cute. I'm not worried about optimizing the call to disable on an uninitialized gc. Let's move forward with the dclp, if you concurrently arrive at a better solution we can review that. Thx!

schveiguy: @andralex dclp is still required, actually. When it replaces itself, it should only do so once. In essence, it's not altering the requirements that this PR is solving. What it would do, however, is avoid any the checks for null on every call once the real GC is in place. Not much cost, but still something. I'm just thinking about what @rainers said about the performance.

I think it also would make the code look cleaner. I'm interested to see if @marler8997 can pull it off, and how it does look.

But in any case, this PR is good enough IMO. Just a thought I had.

schveiguy: > Just a thought I had.

Possible @rainers had the same thought, I just didn't see it 😄 

> An alternative would be to implement the GCInterface with an object that switches the instance global to the standard GC if necessary.

rainers:
> Possible @rainers had the same thought, I just didn't see it

Thanks for noticing ;-) I think with the switching proxy object no atomics are needed to access the instance pointer in the general case if pointer writes cannot appear as partial operations in other threads (which should cover ost architectures). The synchronization is only needed in the switching proxy, i.e. it is a one time operation and doesn't need to be very fast.

The disable state of the GC can be saved to gc.config.disable before atually initializing the GC.

BTW: how about not allocating any pool in the GC. That would have almost the same effect as not initializing it, but a few tiny allocations.

src/gc/proxy.d
somzzz: This is nastier than it looks. 
```thread_init``` calls ```thread_attachThis``` which calls ```GC.disable``` and ```GC.enable```.
https://github.com/dlang/druntime/blob/master/src/core/thread.d#L2138

While ```gc_init``` has not yet finished, the spinlock is still locked. Calling ```disable``` will redirect to ```gc_init``` (with the ```instance``` still ```null``` because ```store``` hasn't been done yet) -> deadlock.

File: src/gc/proxy.d (hunk 9 above)
Comment (MartinNowak) [reply src/gc/proxy.d]: What are the methods requiring that behavior `disable`, `addRoot, removeRoot, rootIter`, `addRange, removeRange, rangeIter`? Sounds like a reasonable approach for lazy initialization.

Comment (MartinNowak) [reply src/gc/proxy.d]: BTW, that just requires an `atomicLoad!(MemoryOrder.acq)`, and that's free on strongly-ordered CPUs.

Comment (schveiguy) [reply src/gc/proxy.d]: I would initialize a full GC on anything that requires storing an array of things to do, because that involves allocation. Just having a global bool is OK to cache until you need the whole GC.

> that's free on strongly-ordered CPUs

Right, x86 doesn't have issues with loads of a word, so those will be fine. It's the other architectures I don't know enough about how the performance might be. I would leave that decision/discussion to those who know more :)

Comment (schveiguy) [reply src/gc/proxy.d]: > I would initialize a full GC on anything that requires storing an array of things to do

Hm... I guess if you need to allocate using the C heap you can, as long as the behavior matches what the GC would actually do. You might even pass those into the GC initializer, and it would just take over ownership of the arrays.

File: src/gc/proxy.d (hunk 4 above)
Comment (somzzz) [reply src/gc/proxy.d]: That works for ```gc_disable```/```gc_enable```.
However, the code then creates an object ```new Thread``` which will become a call to ```_d_newclass``` in ```lifetime.d```. 
https://github.com/dlang/druntime/blob/master/src/rt/lifetime.d#L93

This function calls ```GC.malloc```. I can't pass the instance there and the code hangs on the spinlock (as ```GC.malloc``` tries to call ```gc_init```).

A slightly ugly workaround is to create a static ```tlsInstance``` for the thread doing the initialization and use this one, if available, in ```gc_enable```, ```gc_disable``` and ```gc_malloc```.

Comment (schveiguy) [reply src/gc/proxy.d]: `_d_newclass` should forward to a function that accepts a gc instance, and then instead of calling `new Thread`, call the function directly (with the `auxInstance` of course).

File: src/gc/proxy.d (hunk 9 above)
Comment (andralex) [reply src/gc/proxy.d]: >BTW, that just requires an atomicLoad!(MemoryOrder.acq), and that's free on strongly-ordered CPUs.

Please KISS (Keep It Simple and ConServative) for now. Then we'll take a separate PR with loads and stores tightened. If we ever have trouble with those we'll be able to track that easily. Thanks!

**Comment by wilzbach** [comment]
@somzzz there's a benchmark suite for the GC in `benchmark`. I highly recommend using it.

**Comment by somzzz** [comment]
@wilzbach this one?
https://github.com/dlang/druntime/tree/master/benchmark/gcbench

**Comment by wilzbach** [comment]
> @wilzbach this one?
> https://github.com/dlang/druntime/tree/master/benchmark/gcbench

Yeah, but I think all benchmark tests are useful here.
Copy/pasted the README:

```
# compile master dmd&druntime&phobos
> ./runbench.d -v > baseline.txt
# compile feature branch dmd&druntime&phobos
> ./runbench.d -v > feature.txt
# optionally compile variation of feature branch dmd&druntime&phobos
> ./runbench.d -v > feature_variation.txt
> Rscript --vanilla plot.R baseline.txt feature.txt feature_variation.txt
```

The resulting plots are quite nice and helpful.

**Comment by somzzz** [comment]
This is probably a silly question, but where is the README?

**Comment by wilzbach** [comment]
https://github.com/dlang/druntime/blob/master/benchmark/plot.R

**Comment by somzzz** [comment]
> https://github.com/dlang/druntime/blob/master/benchmark/plot.R

Thanks!

**Comment by somzzz** [comment]
I can't seem to be able to run ```plot.R``` for the benchmarks. There is an error about a length mismatch.
I'll drop the output files here - maybe somebody with R properly installed can run the script, please?

[baseline.txt](https://github.com/dlang/druntime/files/1669474/baseline.txt)
[feature.txt](https://github.com/dlang/druntime/files/1669475/feature.txt)

**Comment by andralex** [comment]
The files seem to be just build logs. @wilzbach can you please fetch this PR and run the benchmark? Thx!

**Comment by wilzbach** [comment]
> There is an error about a length mismatch.
> I'll drop the output files here 

Your output files contain only the compile output. Did you forget `-v`?
In interest of keeping the conversation on topic you might want to consider join Slack and asking your questions there ;-)

**Comment by wilzbach** [comment]
Looks like there's more noise in the test suite than measurable difference:

runbench

The scale is logarithmic - linear would be:

runbench

**Comment by andralex** [comment]
thanks @wilzbach! Time to merge this.

**Comment by rainers** [comment]
> Looks like there's more noise in the test suite than measurable difference:

Not sure how you got the results as they show the PR is supposed to be faster. Here is my benchmark (win64, i7-2670QM):
runbench

There is a performance loss of 0 - 9% for me (averaging to about 3%). Precise scanning has been rejected due to similar results.

Please also note that Thread.getThis() is broken as soon as there is no immediate gc_init from other initializers.

**Comment by andralex** [comment]
@rainers can you please give a bit of detail on what the benchmark is plotting? I take it the Y axis is time? Where can one see the exact percent slowdown? What OS did you test on? (TLS performance is highly variable across OSs.) Thanks!

**Comment by rainers** [comment]
As mentioned above, the benchmark is for win64 on an i7-2670QM. Tests show seconds plotted logarithmically, so the same distance is the same ratio irrespective of the absolute time. The concurrency tests (con*) are not very stable and huge_single is not really a benchmark, so I usually exclude these. Explicit results for the other tests are:

|			| master| pr2057| pr2057/master
|------------|-------| ------| -------------
|dlist		| 2.201	| 2.355	| 1.0699681963
|rand_large	| 3.055	| 3.044	| 0.9963993453
|rand_small	| 5.119	| 5.256	| 1.0267630397
|slist		| 2.297	| 2.351	| 1.0235089247
|testgc3		| 1.958	| 1.94	| 0.9908069459
|tree1		| 0.722	| 0.743	| 1.0290858726
|tree2		| 1.342	| 1.46	| 1.087928465
|vdparser	| 2.304	| 2.37	| 1.0286458333
|words		| 1.245	| 1.254	| 1.0072289157

File: src/gc/proxy.d [hunk 13]
```diff
@@ -28,32 +28,75 @@ private
...
+
+    /*
+    * tlsInstance is a hack to ensure that only the thread initializing the GC
+    * can use GC functions before the initialization is completed. When it is completed,
+    * the GC object created is atomically stored in the __gshared instance.
+    *
+    * This is necessary because calling thead_init is part of the gc_init routine.
+    * thread_init, in turn, calls GC.enable, GC.disable and GC.malloc (through the new keyword)
+    */
+    static GC tlsInstance;
```
MartinNowak: That's creates yet another implicit coupling in a quite subtle init order, why not just passing the GC to thread_init directly?
It's all just about a single `new Thread` allocation in `thread_attachThis`.
https://github.com/dlang/druntime/blob/760fd82820dc686436fccb5617a4b8f9aa406ee7/src/core/thread.d#L2138-L2143

MartinNowak:
And because ppl. always benchmark on laptops, same same on i7-6700.
runbench

test | baseline | #2057 | #2057 w/ acq,rel
-----|----------|-------|----------
slist       | 1.266 | 1.390 | 1.321
huge_single | 0.002 | 0.002 | 0.002
tree1       | 0.419 | 0.439 | 0.439
conalloc    | 0.745 | 0.743 | 0.690
rand_large  | 1.664 | 1.553 | 1.660
tree2       | 0.761 | 0.811 | 0.797
rand_small  | 3.513 | 3.645 | 3.526
conappend   | 0.741 | 0.662 | 0.735
concpu      | 0.900 | 0.941 | 0.849
dlist       | 1.259 | 1.424 | 1.317
testgc3     | 1.564 | 1.582 | 1.589
conmsg      | 0.985 | 1.049 | 1.010
words       | 0.825 | 0.829 | 0.840
vdparser    | 1.278 | 1.287 | 1.289

Given that this PR as-is, doesn't yet provide any user value, we should be wary to not introduce any slowdown.
A reasonable suggestion to not incurr any slowdown and actually delay initialization of the GC until the first *user allocation* was proposed in https://github.com/dlang/druntime/pull/2057#discussion_r163907083.

Given that we have `@nogc` containers in druntime (see [here](https://github.com/dlang/druntime/tree/760fd82820dc686436fccb5617a4b8f9aa406ee7/src/rt/util/container)), implementing such a dummy GC should be fairly simple. In fact if you look at [ManualGC](https://github.com/dlang/druntime/blob/760fd82820dc686436fccb5617a4b8f9aa406ee7/src/gc/impl/manual/gc.d#L38) which already does that.
Then it would just need to acquire a spinlock and replace itself on the first real allocation.

The main thread's `core.thread.Thread` instance could be statically allocated in a `__gshared ubyte[traits(classInstanceSize, Thread)] buf` to make thread_init `@nogc`. This could easily be done in a separate PR.

I'd suggest to go that route instead of introducing an intermediate step that doesn't provide the intended feature. Also we shouldn't merge this before 2018-02-15, as there is no point in releasing this without the finishing the user feature (lazy GC init on first *user* allocation).

src/gc/proxy.d
MartinNowak: Better yet, make `thread_init` `@nogc`, see https://github.com/dlang/druntime/pull/2079.

andralex:
>Given that we have @nogc containers in druntime (see here), implementing such a dummy GC should be fairly simple. In fact if you look at ManualGC which already does that. Then it would just need to acquire a spinlock and replace itself on the first real allocation.

I think there's a difficulty here. Consider:

```d
class ProtoGC {}
private ulong[(__traits(classInstanceSize, ProtoGC) + ulong.sizeof - 1) / ulong.sizeof] buffer;
__gshared ProtoGC instance = cast(ProtoGC) buffer.ptr;
```

The cast sadly doesn't work, but is needed if we want to use a `ProtoGC` that requires no `null`-testing. Thoughts?

Also, regarding the numbers: can we get percentages from the benchmark? Some confidence intervals would be nice too. Right now it's unclear to me how much noise there is, e.g. what is the second minimum of 5? If the difference is large, there's considerable noise.

somzzz:
@andralex 

> > Given that we have @nogc containers in druntime (see here), implementing such a dummy GC should be fairly simple. In fact if you look at ManualGC which already does that. Then it would just need to acquire a spinlock and replace itself on the first real allocation.
> 
> I think there's a difficulty here. Consider:
> 

> class ProtoGC {}
> private ulong[(__traits(classInstanceSize, ProtoGC) + ulong.sizeof - 1) / ulong.sizeof] buffer;
> __gshared ProtoGC instance = cast(ProtoGC) buffer.ptr;
> The cast sadly doesn't work, but is needed if we want to use a ProtoGC that requires no null-testing. Thoughts?
> 

This works fine:
```
    static void initialize(ref GC gc)
    {
        [...]
        auto init = typeid(ProtoGC).initializer();
        assert(init.length == __traits(classInstanceSize, ProtoGC));
        auto instance = cast(ProtoGC) memcpy(p, init.ptr, init.length);
        instance.__ctor();
        gc = instance;
    }
```

andralex:
@somzzz that still requires initialize() to be called during runtime which undoes the advantage. I have a different idea, will get back to this soon.

andralex:
@klickverbot that doens't work either, I really think we should make it work...

andralex:
[GC.pdf](https://github.com/dlang/druntime/files/1708218/GC.pdf)
@klickverbot yah that would work, however the difficulty remains in initializing the buffer appropriately during compilation.

The entire idea of `ProtoGC` as a way to avoid test against null upon each call is predicated on the ability to initialize a reference to a valid `ProtoGC` object during compilation. If we need to initialize `ProtoGC` at runtime we're back where we started from - we may as well initialize the GC proper then. I'd love to see whether that's possible, but until then that's a burden of proof on the proposers, not on this PR.

Regarding performance, I got the attached spreadsheet from @somzzz that deserves a close look. She took the best of 10 runs with the baseline and this PR, and measured the relative speedup/slowdown for each experiment. (The times are sorted.) 

According to these measurements. the average speedup over all benchmarks is 6.66%. There is a very noisy experiment, `huge_single`, which yields very very short times. I think that should be changed because right now it shows the PR improves from 0.004 s to 0.003 s, counting as a huge 25% speedup. That's most likely not realistic as it bears a lot of quantization noise. If we eliminate that experiment, we still get an average speedup of 5.25%.

I think this PR is good to go, future improvements notwithstanding.

Independently, we need to take a good look at our GC benchmarks. The difference between the best runtime and second-best runtime indicates there is a fair amount of noise, so we should increase the number of experiments. Consider e.g. `conalloc`. For the baseline, best time is 0.656 and second best time is 0.816 - a 20% difference indicating a ton of noise. Ideally the two numbers should be very close together. We should therefore increase the number of experiment until the density around the mode (which is expected to be min) is much larger. @MartinNowak do you think you have time to look into that?

[GC.pdf](https://github.com/dlang/druntime/files/1708251/GC.pdf)

wilzbach:
```
usage: runbench [-h|--help] [-v|--verbose] [-r n|--repeat=n] [<test_regex>] [<dflags>] [-- <runargs>]

   tests   - Regular expressions to select tests. Default: '.*\.d'
   dflags  - Flags passed to compiler. Default: '-O -release -inline'
   runargs - Arguments passed to each test, e.g. '--DRT-gcopt=profile=1'

Don't pass any argument to run all tests with optimized builds.
```

Note the `r-|--repeat` flag.
Martin can't do everything ...

andralex:
The PDF doesn't look good, attaching a screenshot of the sheet.
times

andralex:
cc @MartinNowak as you requested changes

andralex:
@wilzbach the idea here is not to run each benchmark a fixed number of times, but to run each as many times as necessary as to collect an accurate best time.

andralex:
@somzzz to my surprise this actuall works - at top level!

```d
___gshared GC intance = new ProtoGC();
```

Somebody implemented compile-time use of `new` at least for certain cases, so you can use it to do the trick with self-initialization. Neat!

rainers:
> If we eliminate that experiment, we still get an average speedup of 5.25%.

Can someone explain how adding null-pointer checks make the code faster? Maybe we should apply this technique elsewhere, too ;-)

andralex:
@rainers a hypothesis is the tests are quite noisy.

rainers:
> @rainers a hypothesis is the tests are quite noisy.

Yeah, that's why I wrote that I usually exclude the con* tests (these test multithreading so are noisier by design) and huge_single (which is not really a benchmark but rather a functional test).

The other tests show rather small variation as can be seen in the plots by the height of the bars (with some tiny dots showing "outliers"). Martins and my measurements are pretty consistent, but @somzzz's results are very different. Lucia, what system are you running the benchmarks on?

somzzz:
Ubuntu Linux 64bit with Intel(R) Core(TM) i7-4750HQ CPU @ 2.00GHz

rainers:
> Ubuntu Linux 64bit with Intel(R) Core(TM) i7-4750HQ CPU @ 2.00GHz

This processor has a large Turbo Boost step up to 3.2 GHz, which can make the benchmarks less predictable. I often disable this for benchmarks on my laptop. I don't know how to do this on linux, on Windows it is done by setting the minimum and maximum processor power states to 99%.

src/rt/dmain2.d
```diff
@@ -197,7 +198,7 @@ extern (C) int rt_init()
         // in other druntime systems.
         _d_initMonoTime();
         thread_init();
-        gc_init();
+        // TODO: fixme - calls GC.addRange -> Initializes GC
```
MartinNowak: Yes, that's what needs to be fixed, e.g. by storing the ranges somewhere else or in ProtoGC until we lazily initialize the real GC.

test/nogc.d
```diff
@@ -0,0 +1,5 @@
+extern(C) __gshared string[] rt_options = [ "gcopt=gc:conservative" ];
```
MartinNowak: The idea was to select a non-existing gc `"gcopt=gc:non-existing"`, so to test that the GC is indeed never initialized.

src/gc/impl/proto/gc.d
```diff
@@ -0,0 +1,192 @@
...
+
+    core.memory.GC.Stats stats() nothrow
+    {
+        return typeof(return).init;
+    }
+
+    void addRoot(void* p) nothrow @nogc
+    {
+        gc_init_nothrow();
+        gc_addRoot(p);
```
MartinNowak: Just store the root in an array and defer initialization.

src/gc/impl/proto/gc.d
```diff
@@ -0,0 +1,192 @@
...
+
+    private int rootsApply(scope int delegate(ref Root) nothrow dg)
+    {
+        return 0;
+    }
+
+    void addRange(void* p, size_t sz, const TypeInfo ti = null) nothrow @nogc
+    {
+        gc_init_nothrow();
+        gc_addRange(p, sz, ti);
```
MartinNowak: Same here, store range in an array and defer initialization.

MartinNowak:
> Some confidence intervals would be nice too. Right now it's unclear to me how much noise there is.

It's already a box plot.
grafik
GC benchmarks are particularly sensible to memory bandwidth, runbench already repeats 10x by default, but core affinity and CPU scaling need to be looked at as well.

Let's stop discussing those benchmarks and focus on the open design/implementation issue.

src/gc/impl/proto/gc.d
```diff
@@ -0,0 +1,192 @@
...
+    {
+    }
+
+    void Dtor()
+    {
+    }
+
+    void enable()
+    {
+        gc_init_nothrow();
```
MartinNowak: Better add an internal `initGC` method which must also replay all addRange/addRoot calls.

src/gc/proxy.d
```diff
@@ -25,28 +26,55 @@ private
...
+        if (!isInstanceInit)
         {
-            import core.stdc.stdio : fprintf, stderr;
-            import core.stdc.stdlib : exit;
+            auto protoInstance = instance;
+            config.initialize();
+            ManualGC.initialize(instance);
+            ConservativeGC.initialize(instance);
+
+            if (instance == protoInstance)
```
MartinNowak: `instance is protoInstance`

src/gc/proxy.d
```diff
@@ -25,28 +26,55 @@ private
...
 extern (C)
 {
-
     void gc_init()
     {
-        config.initialize();
-        ManualGC.initialize(instance);
-        ConservativeGC.initialize(instance);
-        if (instance is null)
+        import core.atomic : atomicLoad, atomicStore, MemoryOrder;
```
MartinNowak: Import no longer needed.

src/rt/dmain2.d
```diff
@@ -58,6 +58,7 @@ extern (C) void _d_monitor_staticctor();
 extern (C) void _d_monitor_staticdtor();
 extern (C) void _d_critical_init();
 extern (C) void _d_critical_term();
+extern (C) void protogc_init();
```
MartinNowak: Left-over declaration.

schveiguy:
Does the removeRange/removeRoot calls need to cache as well, or should they just initialize the GC at that point?

schveiguy:
BTW, this is really cool, and the code looks MUCH simpler!

src/rt/dmain2.d
andralex: @MartinNowak what use cases is this serving? The win of ProtoGC is to get away without ever initializing the GC. But it's unlikely someone adds a bunch of ranges of interest to the GC but ends up doing no allocation. Also: Where would the array be allocated? Also: this has gotten really long in the teeth, improvements can be discussed later.

src/gc/impl/proto/gc.d
andralex: @MartinNowak I'm opposed to this on the same considerations as above. It smacks of overengineering.

src/gc/impl/proto/gc.d
andralex: @MartinNowak please let's not

src/gc/impl/proto/gc.d
andralex: @MartinNowak please not

andralex:
Folks, let's not overengineer! Please! This is creating unbounded amounts of busywork and expands the charter of any PR forever. Next thing up... how about we test for `0` in the first line of `malloc` and `free`? How about we serve first few calls to `malloc` from a static array? The possibilities are endless, but the point of this PR is to not have the GC rely on static initialization.

MartinNowak:
> How about we serve first few calls to malloc from a static array? The possibilities are endless, but the point of this PR is to not have the GC rely on static initialization.

There are 2 simple acceptance criteria, the overhead of lazy GC must be negligible, and we must not initialize and allocate OS resources when the GC is not used (lazy initialization). The latter one still isn't met, but very simple to achieve.
If we do this change we can add to the changelog, that `@nogc main` apps no longer use the GC.

> How about we serve first few calls to malloc from a static array? The possibilities are endless, but the point of this PR is to not have the GC rely on static initialization.

This is polemic and nobody has asked for that, please refrain from strawman arguments.

MartinNowak:
> Does the removeRange/removeRoot calls need to cache as well, or should they just initialize the GC at that point?

It doesn't matter for the purpose of lazy GC initialization.
It's trivial to just remove from the arrays though, see https://github.com/dlang/druntime/blob/ee1b729d8797ab582c8850e1fb958f66f1bcac78/src/gc/impl/manual/gc.d#L237.

andralex:
>There are 2 simple acceptance criteria, the overhead of lazy GC must be negligible, and we must not initialize and allocate OS resources when the GC is not used (lazy initialization). The latter one still isn't met, but very simple to achieve.

I realized I omitted a discussion I've had with @somzzz. Next on her list to move the code that adds roots in `_Dmain` from `src/rt/sections_elf_shared.d`. The roots will be added lazily during the initialization. That eliminates the internal use of `addRoot`/`addRange`. I do agree that if one of the first things _Dmain does is to call `addRoot`/`addRange`, then this PR doesn't help a lot.

So please let's get this in without more features. Accumulating `addRoot`/`addRange` calls in an array is not trivial - where do you allocate the array? There's no GC. So then we need to use `malloc`/`free`, an additional complication. And once we remove `addRoot`/`addRange` from `_Dmain`, the clientele of that complication will be virtually nobody.

src/gc/impl/proto/gc.d
```diff
@@ -0,0 +1,192 @@
...
+    extern (C) size_t   gc_reserve( size_t sz ) nothrow;
+
+    extern (C) void gc_addRange( void* p, size_t sz, const TypeInfo ti = null ) nothrow @nogc;
+    extern (C) void gc_addRoot( void* p ) nothrow @nogc;
+}
+
+class ProtoGC : GC
+{
+    __gshared Array!Root roots;
+    __gshared Array!Range ranges;
```
MartinNowak: Do those really need to be `__gshared`?

andralex:
@CyberShadow I agree the increase of build time in unittests is unpleasant. It seems the right long-term solution to that is making the `-unittest` flag more modular. Another possibility is curate Phobos imported files upon distribution. I don't think this solution is the right response.

MartinNowak:
> So please let's get this in without more features. Accumulating addRoot/addRange calls in an array is not trivial - where do you allocate the array? There's no GC.

As said earlier, we have `@nogc` containers in druntime for those purposes.
In fact they are already added (https://github.com/dlang/druntime/pull/2057/files#diff-254a9fe518c0b5c278aeec1838d7adf8R34), but not used.

andralex:
@MartinNowak in wake of the upcoming removal of `addRange` from `_Dmain`, what would be the usefulness of your proposed added feature?

MartinNowak:
> I realized I omitted a discussion I've had with @somzzz. Next on her list to move the code that adds roots in _Dmain from src/rt/sections_elf_shared.d. The roots will be added lazily during the initialization. That eliminates the internal use of addRoot/addRange. I do agree that if one of the first things _Dmain does is to call addRoot/addRange, then this PR doesn't help a lot.

That sounds nice in theory, but you're likely underestimating the complexity in that area, because GC roots are dynamically added/removed when loading/unloading shared libraries. So instead of conflating a complex with another complex, let's just go for the straightforward solution and cache roots/ranges until lazily initializing the GC.
In fact we've prolly spent 2x more time discussing this, than would be necessary to do the change.

andralex:
@MartinNowak meet on slack?

andralex:
After discussing, the shared lib example is quite compelling. @somzzz will implement the changes with this PR.

MartinNowak:
Another nice thing that comes with *this* PR, once https://github.com/dlang/druntime/pull/1924 is implemented, applications wouldn't even need to link against any GC.

andralex:
Congrats @somzzz and many thanks to all reviewers!

MartinNowak:
> Congrats @somzzz and many thanks to all reviewers!

Yes, thanks for this important step towards a no-GC language, and thanks @schveiguy for the nice ProtoGC idea.
The amount of feedback was a bit overwhelming, but this is certainly due to the interest in the topic.

schveiguy:
> thanks @schveiguy for the nice ProtoGC idea

Thanks for the credit, I also have to mention that @rainers had the same idea (before I did), but just expressed it in a different way here: https://github.com/dlang/druntime/pull/2057#issuecomment-360257089.

src/gc/impl/proto/gc.d
```diff
@@ -0,0 +1,240 @@
...
+        {
+            if (auto result = dg(r))
+                return result;
+        }
+        return 0;
+    }
+
+    void addRange(void* p, size_t sz, const TypeInfo ti = null) nothrow @nogc
+    {
+        ranges.insertBack(Range(p, p + sz, cast() ti));
```
klickverbot: This doesn't seem thread-safe – are we sure there can never be a situation where ProtoGC sticks around long enough to be racy (e.g. after improving druntime thread startup to be nogc and using statically allocated `Thread` instances, or when attaching to multiple C threads)?

Has this been thoroughly thought through for race conditions in general, especially considering architectures where loading the global `instance` doesn't automatically have acquire semantics?

schveiguy: It's a good point. Likely you need to take the lock. I'm wondering if adding/removing ranges or roots should simply initialize the GC at that point.

I thought the instance should be atomically loaded, see Martin's suggestion [here](https://github.com/dlang/druntime/pull/2057#discussion_r164188206). But that doesn't help if you need atomic array appending.

andralex:
@klickverbot thanks for the reminder. @somzzz is off to a scholarship in Singapore so she's off the project. The concern about addRange seems legit. Can you please create a bugzilla issue? Thanks!

BBasile:
This change causes a regression in phobos Array container. `gc_init` should be made public now. [see bugzilla](https://issues.dlang.org/show_bug.cgi?id=18996#c4)

schveiguy:
> `gc_init` should be made public now.

No, the point of `ProtoGC` is to initialize properly when needed, we should not need to make `gc_init` public (nor should we require calling it).

The problem causing the issue is that `removeRoot` and `removeRange` do not behave correctly. I'll file a PR to fix this.

schveiguy:
Fixing PR: #2220

PR #2058, opened 2018-01-24
fix Issue 18252 - comparison of arrays of associative arrays broken

MartinNowak:
- add missing handling of associative arrays as array elements to __equals


MartinNowak:
Filed an issue for the druntime failure on CircleCI.
[18296 – [Reg2.078.1] invalid code with coverage and copy construction](https://issues.dlang.org/show_bug.cgi?id=18296)

PR #2059, opened 2018-01-24
workaround Issue 18296 to fix coverage tests

MartinNowak:


wilzbach:
FYI there's also https://github.com/dlang/druntime/pull/2048 with a bit of discussion.

PR #2061, opened 2018-01-24
Run CircleCi coverage for 32-bit too

wilzbach:
See also: https://github.com/dlang/dmd/pull/7718

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
     working_directory: ~/druntime
     docker:
       - image: circleci/node:4.8.2
-    parallelism: 1
+    parallelism: 2
     steps:
       - checkout
       - run:


wilzbach:
Jenkins failure is unrelated and should be gone soon (-> https://github.com/dlang/ci/pull/132)

wilzbach:
> 75.411% (+1.001%) compared to 2d45fcb

It has a measurable effect ;-)

wilzbach:
We initially planned to run the coverage on 32-bit too (hence there were no changes to `.circleci/run.sh` needed).
However, for CircleCi 1.0 the parallelism was always set to `1`. A couple of weeks ago we switched to CircleCi 2.0 and thus the switch to `parallelism: 2` is now possible.
The druntime test suite runs very quickly (~2 mins) + there aren't that many PRs compared to DMD or Phobos, so we don't have to worry about this here.
FWIW we run CircleCi with `parallelism: 3` at dlang/dmd and the test take ~18 mins to run there, so if we ever run into a bottleneck it's at dmd.

PR #2065, opened 2018-01-28
Fixes issue 18279 - rt.util.utf does not properly reserve buffer in toUTF16/toUTF16z

schveiguy:
setting length to 0 no longer works for reserving GC data. You need to use `arr.reserve`


rainers:
There is also https://github.com/dlang/druntime/pull/2053

schveiguy:
Closing in favor of #2053

PR #2069, opened 2018-01-30
Issue 18332 - rt.util.random.Rand48 remove unnecessary assert

n8sh:
There is no need for seedval to be non-zero. The assertion should be removed.


wilzbach:
Thanks. Sorry that this PR got lost. Druntime seems to be low-traffic at the moment.

PR #2070, opened 2018-01-30
[RFC] Use native C definitions

yshui:
**Don't merge!**

# Use libc definitions directory
This patch try to make it possible to automatically generate corresponding D definitions from libc header files, instead of writing them by hand.

## Motivation
Right now, adding a new platform support to druntime involves copying definitions from C header files, and then converting them to D (or maybe I'm doing it wrong?). This is a tiresome and boring process. The resulting D source files are repetive, and have a (albeit low) risk of getting out-of-sync with the C headers.

So this patch tries to solve both of these problem by automatically generating D code from libc headers.

## Method
Currently, only constants defined in macros are processed. This is done by generating a C++ file filled with `#ifdef`s, and then use that genrated file to genrate the desired D file.

Basically, the first C++ file will look like this:

```c
#ifdef C_CONSTANT
//print "enum C_CONSTANT = <value of C_CONSTANT>"
#endif
```

## Known problems
1) This will make the resulting druntime less portable. You can't copy the druntime installed on your system and use it on another platform. I don't think this is a significant problem.

2) Building a cross-compiler might become harder

## Alternatives
Use DStep to generate those D files.

## Progress
Two files are converted to demostrate the idea (and take advantage of buildbot): `core.sys.posix.fcntl` and `core.sys.posix.poll`.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -4,6 +4,7 @@
 # and then run as gmake rather than make.
 
 QUIET:=
+CXX:=g++
 
 DMD_DIR=../dmd
 
@@ -104,10 +105,13 @@ DRUNTIMESO=$(ROOT)/libdruntime.so
 DRUNTIMESOOBJ=$(ROOT)/libdruntime.so.o
 DRUNTIMESOLIB=$(ROOT)/libdruntime.so.a
 
+GEN_SRCS= src/core/sys/posix/fcntl_c.d \
+	  src/core/sys/posix/poll_c.d
+
 DOCFMT=
 
 include mak/COPY
-COPY:=$(subst \,/,$(COPY))
+COPY:=$(subst src,import,$(GEN_SRCS)) $(subst \,/,$(COPY))
 
 include mak/DOCS
 DOCS:=$(subst \,/,$(DOCS))
@@ -116,7 +120,7 @@ include mak/IMPORTS
 IMPORTS:=$(subst \,/,$(IMPORTS))
 
 include mak/SRCS
-SRCS:=$(subst \,/,$(SRCS))
+SRCS:=$(GEN_SRCS) $(subst \,/,$(SRCS))
 
 # NOTE: trace.d and cover.d are not necessary for a successful build
 #       as both are used for debugging features (profiling and coverage)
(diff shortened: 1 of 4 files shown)


CyberShadow:
Interesting idea, but I think such tools would be better used to generate .d files that are then committed to git permanently. The current approach is prohibitive to things like cross-compilation, which LDC is already experimenting with.

ibuclaw:
Over a decade ago there was a tool used in-house for gdc called x3 that did the same thing but covering a larger number of system headers.  It was frankly nuts from a maintenance perspective and I'm glad it's long gone.  Wouldn't want to see history repeat itself in druntime.

I think dstep is a better fit for the use case of getting a new platform bindings converted over quickly for testing purposes.

And I'd rather keep any tooling outside of the build tree, as adding complexity to the build is a sure way to annoy downstream.

joakim-noah:
An interesting idea, we should definitely be going in this direction of automation.  However, it'd be better if we used DStep to generate the definitions for new platforms, committed them to git, and then periodically checked them again using DStep, though certainly not on every build.

yshui:
If we are going the "Generate -> commit to git" route, then the tool should be able automatically generate `version()` blocks from headers from different platform.

Generating headers per platform and then manually combine them will be a nightmare...

yshui:
@CyberShadow Yes, cross-compilation will be tricky. But I think it's not impossible.

ibuclaw:
> Generating headers per platform and then manually combine them will be a nightmare...

I once tried packaging per platform/architecture bindings - one module per target, package module publicly imports depending on version identifiers. It never gained any traction, and that's probably for the best anyway.

joakim-noah:
>the tool should be able automatically generate version() blocks from headers from different platforms

Yeah, something like that has been possible since I used DStep to help separate out linux kernel APIs in #1010.  I'd guess the easiest way to do it is to have DStep dump a bunch of automatically translated D declarations, like I did back then, then use dlang-community/libdparse to automatically compare those D declarations with the ones committed in git, which I didn't try back then, and find any irregularities in the symbol names or various types.

Once that's working well to compare existing platforms' declarations, we could use it to generate new platforms' declarations too, by writing a small tool to merge the Dstep output into git, similar to dlang-community/dfmt.

>I once tried packaging per platform/architecture bindings - one module per target, package module publicly imports depending on version identifiers. It never gained any traction, and that's probably for the best anyway.

I thought you simply put it off for later and planned to come back to it.  I'm fine with either way of separating out platforms.

RazvanN7:
The consensus seems to be that the D header files should be checked in git and not generated. I'm going to close this. If anyone can come up with a compelling argument, please reopen this.

PR #2071, opened 2018-01-31
Some missing parts for musl

yshui:


joakim-noah:
After all your work on using the Musl libc, I thought I'd try it out, so I whipped up an Alpine VPS at my host.  I got dmd master built by bootstrapping through ldc ltsmaster and then druntime master built fine, after adding this patch and commenting out some backtrace declarations that aren't there on Alpine.

However, the druntime test runner always segfaults on me with the tests I tried, any idea what's going on?  Are you modifying dmd in some way for Musl, other than having it set `CRuntime_Musl` instead of `CRuntime_Glibc`, like I am?  Building and running a simple D program works fine for me, tried [this one from the tests](https://github.com/dlang/dmd/blob/master/test/runnable/sieve.d) after modifying it to only use `puts`/`printf` from druntime.

joakim-noah:
Alright, finally got all the druntime tests to run on Alpine with Musl, by applying this pull and making a few tweaks.  Two test blocks from [`core.thread` that](https://github.com/dlang/druntime/blob/master/src/core/thread.d#L1399) call [`GC.collect` manually](https://github.com/dlang/druntime/blob/master/src/core/thread.d#L5353) had to be disabled, and the test runner still segfaulted on exit after running and passing all other unit tests in `core.thread`.

When I made the test runner skip `core.thread`, I only had to disable [this assert in `core.sync.mutex`, just like on Bionic](https://github.com/dlang/druntime/blob/master/src/core/sync/mutex.d#L380), all the other druntime modules' tests ran fine.  When running the additional tests, I only had to disable `thread_join` from `init_fini` in release mode, everything else worked after the two changes mentioned above: `SHARED` turned off and the backtrace declarations commented out.

wilzbach:
@yshui  Once the issues that @joakim-noah has pointed out, have been addressed, how about making adding a changelog entry and announcement in the NG?

> so I whipped up an Alpine VPS at my host.

As Alpine is getting more and more popular, how about doing this in a Dockerfile which can then be used by everyone who wants to use D applications with Alpine?

The GoLang people do this: https://hub.docker.com/_/golang/
There's already [DMD in Docker](https://hub.docker.com/r/dlanguage/dmd/) and compiling an application for alpine could be as simple as:

```sh
docker run --rm -ti -v $(pwd):/src dlang-alpine/dmd dmd -run hello_alpine.d
```

Someone already missed it here: https://github.com/lindt/docker-dmd/issues/1

yshui:
@joakim-noah Yes, shared library doesn't work yet. I would need some time to look into that.

ibuclaw:
Has gdc and ldc compilers been tested with these bindings? That would separate binding problems from just dmd problems.

joakim-noah:
Why would you suspect dmd?

ibuclaw:
Experience.  GDC (and maybe LDC) just integrates into the environment if you compile it for that platform.

yshui:
@ibuclaw @joakim-noah I think it's a problem with musl.

src/core/sys/posix/sys/stat.d
```diff
@@ -1294,11 +1294,11 @@ else version (CRuntime_Musl)
         timespec st_atim;
         timespec st_mtim;
         timespec st_ctim;
-        extern(D)
+        extern(D) @safe @property
```
joakim-noah: I hit this too, wonder why `@safe` is only needed with this C runtime.

yshui: Not sure. Other C runtime doesn't use this definition. e.g. in CRuntime_Glibc this is behind a `(__USE_MISC || __USE_XOPEN2KB)` check, which is false.

yshui:
First problem I found that causes a crash: `src/rt/sections_elf_shared.d:675`, this won't work with PIE.

Still hit another segfault after disabling PIE.

yshui:
Erm, at `src/rt/sections_elf_shared.d:675`, doing the relocation seems to fix the crash.

yshui:
Looks like the glibc dynamic linker also relocates the entries in the dynamic section, while musl's doesn't. So I scoped the change to musl.

yshui:
test/shared/link assumes library constructors are always run before main constructors, which seems to not be true with musl.

yshui:
@joakim-noah Can you test again?

joakim-noah:
Ran the druntime tests again with this latest pull, which fixes most of the problems.  I still get a segfault in `init_fini/thread_join` and `shared/finalize` errors out, when running the tests in release mode.  In debug mode, asserts start tripping up in the additional tests for shared libraries.  Didn't actually need to remove the backtrace declarations, that was only needed for the bootstrap ldc, which I didn't modify to use `CRuntime_Musl`.

I went ahead and started running the Phobos tests too, most of them passed with these additions to druntime:
```
diff --git a/src/core/stdc/math.d b/src/core/stdc/math.d
index 488b9d0a..8a7000bc 100644
--- a/src/core/stdc/math.d
+++ b/src/core/stdc/math.d
@@ -626,7 +626,7 @@ else version( CRuntime_Musl )
         ///
         FP_FAST_FMAL = 0,
     }
-
+pure{
     int __fpclassifyf(float x);
     int __fpclassify(double x);
     int __fpclassifyl(real x);
@@ -730,6 +730,7 @@ else version( CRuntime_Musl )
             : __signbitl(x);
     }
   }
+    }
 }
 else version( MinGW )
 {
diff --git a/src/core/sys/posix/pthread.d b/src/core/sys/posix/pthread.d
index 23df7895..9b76e9db 100644
--- a/src/core/sys/posix/pthread.d
+++ b/src/core/sys/posix/pthread.d
@@ -384,6 +384,7 @@ else version( CRuntime_Bionic )
 else version( CRuntime_Musl )
 {
 
+    enum PTHREAD_MUTEX_INITIALIZER = pthread_mutex_t.init;
 }
 else
 {
diff --git a/src/core/sys/posix/stdlib.d b/src/core/sys/posix/stdlib.d
index 70794da8..186633c6 100644
--- a/src/core/sys/posix/stdlib.d
+++ b/src/core/sys/posix/stdlib.d
@@ -125,6 +125,10 @@ else version( CRuntime_Bionic )
     // Added since Lollipop
     int posix_memalign(void**, size_t, size_t);
 }
+else version( CRuntime_Musl )
+{
+    int posix_memalign(void**, size_t, size_t);
+}
 
 //
 // C Extension (CX)
```
That first `pure` block is needed because of a recent change in git master for druntime.  This is the patch I used to get most Phobos tests passing:
```
diff --git a/std/datetime/timezone.d b/std/datetime/timezone.d
index 61416cf9e..4de1fd7c3 100644
--- a/std/datetime/timezone.d
+++ b/std/datetime/timezone.d
@@ -758,7 +758,7 @@ public:
 
     @safe unittest
     {
-        assert(LocalTime().dstName !is null);
+        //assert(LocalTime().dstName !is null);
 
         version(Posix)
         {
diff --git a/std/experimental/allocator/building_blocks/region.d b/std/experimental/allocator/building_blocks/region.d
index 82e550e3c..dfbd6641d 100644
--- a/std/experimental/allocator/building_blocks/region.d
+++ b/std/experimental/allocator/building_blocks/region.d
@@ -830,7 +830,7 @@ version(Posix) struct SbrkRegion(uint minAlign = platformAlignment)
     }
 }
 
-version(Posix) @system unittest
+version(rosix) @system unittest
 {
     // Let's test the assumption that sbrk(n) returns the old address
     const p1 = sbrk(0);
@@ -842,7 +842,7 @@ version(Posix) @system unittest
     sbrk(-4096);
 }
 
-version(Posix) @system unittest
+version(rosix) @system unittest
 {
     import std.typecons : Ternary;
     import std.algorithm.comparison : min;
diff --git a/std/file.d b/std/file.d
index 2b042de5b..9dea2c164 100644
--- a/std/file.d
+++ b/std/file.d
@@ -1346,7 +1346,7 @@ if (isConvertibleToString!R)
     }
 
     testTimes(0);
-    version (linux)
+    version (rinux)
         testTimes(123_456_7);
 
     rmdirRecurse(newdir);
diff --git a/std/socket.d b/std/socket.d
index 6a5e5ff2b..97167ea0a 100644
--- a/std/socket.d
+++ b/std/socket.d
@@ -203,7 +203,7 @@ string formatSocketError(int err) @trusted
             else
                 return "Socket error " ~ to!string(err);
         }
-        else version (CRuntime_Bionic)
+        else version (CRuntime_Musl)
         {
             auto errs = strerror_r(err, buf.ptr, buf.length);
             if (errs == 0)
@@ -2063,7 +2063,7 @@ static if (is(sockaddr_un))
         }
     }
 
-    @safe unittest
+    version(none)@safe unittest
     {
         import core.stdc.stdio : remove;
         import std.file : deleteme;
diff --git a/std/stdio.d b/std/stdio.d
index 16d34f0e9..eab71ad10 100644
--- a/std/stdio.d
+++ b/std/stdio.d
@@ -81,6 +81,11 @@ version (CRuntime_Bionic)
     version = NO_GETDELIM;
 }
 
+version (CRuntime_Musl)
+{
+    version = GENERIC_IO;
+    version = HAS_GETDELIM;
+}
 // Character type used for operating system filesystem APIs
 version (Windows)
 {
```
One assert, three test blocks, and one linux test function were disabled.  Also, the `std.net.curl` tests don't pass, looks like some of the socket declarations in druntime are wrong and cause runtime exceptions.

Other than that, looking good so far, just more polishing to do.  I don't expect you to fix all these issues though, so whenever you're ready, we can get this pull in.

yshui:
@joakim-noah `pure` and missing definitions added.

joakim-noah:
@wilzbach, this is ready to go.  This port still needs some polishing for shared libraries and phobos to pass all their tests, but we can get that in later.

src/core/sync/mutex.d
```diff
@@ -377,6 +377,7 @@ unittest
     // For example, Bionic doesn't appear to do so, so this test is
     // not run on Android.
     version (CRuntime_Bionic) {} else
+    version (CRuntime_Musl) {} else
```
wilzbach: Weird, I thought `Musl` is promoted as drop-in replacement for glibc...

klickverbot: The compatibility only extends to the C source level, plus musl doesn't implement all features.

joakim-noah:
@yshui, any plans to finish off Phobos?  I got that linux test function in `std.file` to work with Musl by adding declarations for `utimensat` and `AT_FDCWD` to druntime, as [I had to for Bionic](https://github.com/dlang/druntime/pull/1819/files#r114472299).

yshui:
@joakim-noah Since you had it working, you could go ahead and open a PR, if you want to.

joakim-noah:
OK, just wanted to make sure we weren't duplicating effort.  I will try to fix those remaining Phobos tests for Musl, not going to do anything with shared libraries though.

PR #2072, opened 2018-01-31
Fix error due to deprecated rwm for shared variables

wilzbach:
Required for https://github.com/dlang/dmd/pull/7816

### DIFF
--- a/src/core/sync/mutex.d
+++ b/src/core/sync/mutex.d
@@ -318,7 +318,7 @@ unittest
         void useResource() shared @safe nothrow @nogc
         {
             mtx.lock_nothrow();
-            cargo++;
+            *(cast()&cargo) += 1;
             mtx.unlock_nothrow();
         }
     }


jacob-carlborg:
Are we going to disallow this? Should one use `__gshared` instead if one wants to use a mutex? What about `synchronized` blocks? Is that allowed to modify `shared`?

wilzbach:
> Are we going to disallow this?

It has triggered a deprecation warning for the past four years. Time to move on.

> Should one use __gshared instead if one wants to use a mutex?

No, `shared` works fine and should be used:

```
import core.sync.mutex;
shared mtx = new shared Mutex();
mtx.lock();
mtx.unlock();
synchronized (mtx) {}
```

https://run.dlang.io/is/3WyR5O

Though this has nothing to do with this PR. This is about the compiler preventing you from bugs when using `shared` as a read & write is inherently unsafe. Note that:

1) direct assignment `shared int foo; foo = 2` still works
2) in general: `shared` is still one of the most broken/undefined parts of the language

jacob-carlborg:
I'm referring to `cargo++;` which was guarded by a mutex. What's the solution if one wants to use mutex or a `synchronized` block? Should the declaration of `cargo` be `__gshared`? Both of these [1] examples fail with the a deprecation message.

[1] https://run.dlang.io/is/naqANn

wilzbach:
Okay changed it to use `__gshared` - seems to be the best option then.

src/core/sync/mutex.d
```diff
@@ -307,7 +307,7 @@ unittest
     class Resource
     {
         Mutex mtx;
-        int cargo;
+        __gshared int cargo;
```
klickverbot: This makes `cargo` shared between all instances.

ZombineDev:
(Changed the PR title to match the commit message, as `atomicOp` is no longer used.)

PR #2074, opened 2018-02-03
add aio.d for linux x64

Darredevil:
Added interface for posix aio in D (only for linux x64 for now).
The aiocb struct can have various implementations, for linux x64 I tested the following in C and created the equivalent in D:
```
C aiocb
Sizeof aiocb: 168 bytes
Offsets:
aiocb.aio_fildes:	 0
aiocb.aio_lio_opcode:	 4
aiocb.aio_reqprio:	 8
aiocb.aio_buf:		 16
aiocb.aio_nbytes:	 24
aiocb.aio_sigevent:	 32
aiocb.aio_offset:	 128

```

### DIFF
--- a/src/core/sys/posix/aio.d
+++ b/src/core/sys/posix/aio.d
@@ -10,8 +10,8 @@ module core.sys.posix.aio;
 
 private import core.sys.posix.signal;
 
-version (linux):
-version(X86_64):
+version (CRuntime_Glibc):
+version (X86_64):
 
 extern (C):
 @system:
@@ -30,7 +30,6 @@ struct aiocb
     ubyte[24] internal_members_padding;
     off_t aio_offset;
     ubyte[32] __glibc_reserved;
-
 }
 
 /* Return values of cancelation function.  */
--- a/src/core/sys/posix/aio.d
+++ b/src/core/sys/posix/aio.d
@@ -3,8 +3,8 @@
  * Available since Linux 2.6
  *
  * Copyright: Copyright D Language Foundation 2018.
- * License : $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
- * Authors  : Alexandru Razvan Caciulescu (github.com/darredevil)
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   Alexandru Razvan Caciulescu (github.com/darredevil)
  */
 module core.sys.posix.aio;
 
--- a/src/core/sys/posix/aio.d
+++ b/src/core/sys/posix/aio.d
@@ -1,0 +1,66 @@
+/**
+ * D header file to interface with the Linux aio API (http://man7.org/linux/man-pages/man7/aio.7.html).
+ * Available since Linux 2.6
+ *
+ * Copyright: Copyright D Language Foundation 2018.
+ * License : $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors  : Alexandru Razvan Caciulescu (github.com/darredevil)
+ */
+module core.sys.posix.aio;
+
+private import core.sys.posix.signal;
+
+version (linux):
+version(X86_64):
+
+extern (C):
+@system:
+@nogc:
+nothrow:
+
+struct aiocb
+{
+    int aio_fildes;
+    int aio_lio_opcode;
+    int aio_reqprio;
+    void *aio_buf;   //volatile
+    size_t aio_nbytes;
+    sigevent aio_sigevent;
+
+    ubyte[24] internal_members_padding;
+    off_t aio_offset;
+    ubyte[32] __glibc_reserved;
+
+}
+
+/* Return values of cancelation function.  */
+enum
+{
+    AIO_CANCELED,
+    AIO_NOTCANCELED,
+    AIO_ALLDONE
+};
+
+/* Operation codes for `aio_lio_opcode'.  */
+enum
+{
+    LIO_READ,
+    LIO_WRITE,
+    LIO_NOP
+};
+
+/* Synchronization options for `lio_listio' function.  */
+enum
+{
+    LIO_WAIT,
+    LIO_NOWAIT
+};
+
+int aio_read(aiocb *aiocbp);
+int aio_write(aiocb *aiocbp);
+int aio_fsync(int op, aiocb *aiocbp);
+int aio_error(aiocb *aiocbp);
+ssize_t aio_return(aiocb *aiocbp);
+int aio_suspend(aiocb*[] aiocb_list, int nitems, timespec *timeout);
+int aio_cancel(int fd, aiocb *aiocbp);
+int lio_listio(int mode, aiocb*[] aiocb_list, int nitems, sigevent *sevp);
--- a/src/core/sys/posix/aio.d
+++ b/src/core/sys/posix/aio.d
@@ -1,10 +1,11 @@
 /**
- * D header file to interface with the Linux aio API (http://man7.org/linux/man-pages/man7/aio.7.html).
+ * D header file to interface with the
+ * $(HTTP pubs.opengroup.org/onlinepubs/9699919799/basedefs/aio.h.html, Linux aio API).
  * Available since Linux 2.6
  *
  * Copyright: Copyright D Language Foundation 2018.
  * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
- * Authors:   Alexandru Razvan Caciulescu (github.com/darredevil)
+ * Authors:   $(HTTPS github.com/darredevil, Alexandru Razvan Caciulescu)
  */
 module core.sys.posix.aio;
 
--- a/src/core/sys/posix/aio.d
+++ b/src/core/sys/posix/aio.d
@@ -1,7 +1,6 @@
 /**
  * D header file to interface with the
- * $(HTTP pubs.opengroup.org/onlinepubs/9699919799/basedefs/aio.h.html, Linux aio API).
- * Available since Linux 2.6
+ * $(HTTP pubs.opengroup.org/onlinepubs/9699919799/basedefs/aio.h.html, Posix AIO API).
  *
  * Copyright: Copyright D Language Foundation 2018.
  * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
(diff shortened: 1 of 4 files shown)


src/core/sys/posix/aio.d
```diff
@@ -0,0 +1,66 @@
+/**
+ * D header file to interface with the Linux aio API (http://man7.org/linux/man-pages/man7/aio.7.html).
+ * Available since Linux 2.6
+ *
+ * Copyright: Copyright Alexandru Razvan Caciulescu 2018.
```
wilzbach: D Language Foundation.

Darredevil: I took `core.sys.linux.epoll` as an example for the copyright, is that file not respecting the convention?

wilzbach: That file was written long before the DFL existed. We can't easily change the copyright for existing modules, but we can for new additions.

Darredevil: fixed

src/core/sys/posix/aio.d
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Copyright: Copyright D Language Foundation 2018.
+ * License : $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors  : Alexandru Razvan Caciulescu (github.com/darredevil)
+ */
+module core.sys.posix.aio;
+
+private import core.sys.posix.signal;
+
+version (linux):
```
joakim-noah: `CRuntime_Glibc`, not `linux`.

src/core/sys/posix/aio.d
```diff
@@ -0,0 +1,66 @@
...
+    int aio_lio_opcode;
+    int aio_reqprio;
+    void *aio_buf;   //volatile
+    size_t aio_nbytes;
+    sigevent aio_sigevent;
+
+    ubyte[24] internal_members_padding;
+    off_t aio_offset;
+    ubyte[32] __glibc_reserved;
+
```
ibuclaw: Extra space

src/core/sys/posix/aio.d
```diff
@@ -0,0 +1,66 @@
+/**
+ * D header file to interface with the Linux aio API (http://man7.org/linux/man-pages/man7/aio.7.html).
+ * Available since Linux 2.6
```
ibuclaw: So this should be in core/sys/linux then, not posix.

Darredevil: Actually no, there are two instance of AIO, the POSIX aio (aio_* family) and Linux libaio.
I believe this is in the correct place.

DmitryOlshansky: This is POSIX AIO interface except for aio_init I think.

andralex:
@ibuclaw good to go?

@CyberShadow can you please advise regarding the merge error in circleci?

wilzbach:
> Can you please advise regarding the merge error in circleci?

@Darredevil's fork was pretty outdated and we upgraded to CircleCi 2.0.
tl;dr: rebasing helps. I just did so.

Darredevil:
@ibuclaw all looks good now

src/core/sys/posix/aio.d
```diff
@@ -0,0 +1,65 @@
+/**
+ * D header file to interface with the Linux aio API (http://man7.org/linux/man-pages/man7/aio.7.html).
```
ibuclaw: Any link to the posix standard that is not a Linux manpage?

Darredevil: Will this suffice?
https://www.freebsd.org/cgi/man.cgi?query=aio&sektion=4&apropos=0&manpath=FreeBSD+11.1-RELEASE+and+Ports

Darredevil: Also http://www2.fisica.unlp.edu.ar/electro/temas/p7/POSIX4.pdf , see page 277

joakim-noah: This one seems best: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/aio.h.html

andralex: I made the change myself. All, feel free to operate noncontroversial changes (spacing, better links) right on the spot - it takes only a bit longer than making the request and accelerates the cycle greatly.

src/core/sys/posix/aio.d
```diff
@@ -0,0 +1,66 @@
...
+};
+
+int aio_read(aiocb *aiocbp);
+int aio_write(aiocb *aiocbp);
+int aio_fsync(int op, aiocb *aiocbp);
+int aio_error(aiocb *aiocbp);
+ssize_t aio_return(aiocb *aiocbp);
+int aio_suspend(aiocb*[] aiocb_list, int nitems, timespec *timeout);
+int aio_cancel(int fd, aiocb *aiocbp);
+int lio_listio(int mode, aiocb*[] aiocb_list, int nitems, sigevent *sevp);
```
ibuclaw: `aiocb*[]` would be a dynamic array, so this should be instead `const(aiocb*)*`

src/core/sys/posix/aio.d
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Copyright: Copyright D Language Foundation 2018.
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   $(HTTPS github.com/darredevil, Alexandru Razvan Caciulescu)
+ */
+module core.sys.posix.aio;
+
+private import core.sys.posix.signal;
+
+version (CRuntime_Glibc):
```
ibuclaw: Top of file should be guarded with `version (Posix)`, not glibc.

src/core/sys/posix/aio.d
```diff
@@ -0,0 +1,66 @@
...
+version (X86_64):
```
ibuclaw: This should not be here, rather it should be moved to where x86_64-specific definitions are.

src/core/sys/posix/aio.d
```diff
@@ -0,0 +1,66 @@
...
+    int aio_fildes;
+    int aio_lio_opcode;
+    int aio_reqprio;
+    void *aio_buf;   //volatile
+    size_t aio_nbytes;
+    sigevent aio_sigevent;
+
+    ubyte[24] internal_members_padding;
+    off_t aio_offset;
+    ubyte[32] __glibc_reserved;
```
ibuclaw: This looks like the only platform specific part of this module here. Make a porters life easy and guard it properly with appropriate version blocks and static asserts.

i.e:
```
version (CRuntime_Glibc)
{
    version (X86_64)
    {
        struct aiocb
        {
            //... 
        }
    }
    else
        static assert(0);
}
else
    static assert(false, "Unsupported platform");
```

src/core/sys/posix/aio.d
joakim-noah: I think the idea is that this has only been tested on linux/Glibc so far, of all Posix platforms.

src/core/sys/posix/aio.d
```diff
@@ -0,0 +1,66 @@
+/**
+ * D header file to interface with the
+ * $(HTTP pubs.opengroup.org/onlinepubs/9699919799/basedefs/aio.h.html, Linux aio API).
+ * Available since Linux 2.6
```
ibuclaw: Shouldn't you have dropped the Linux references here?

Take a look any of the other `core.sys.posix` module headers, eg: https://github.com/dlang/druntime/blob/master/src/core/sys/posix/pwd.d

src/core/sys/posix/aio.d
ibuclaw: Doesn't mean we should drop the convention used in every other sys/posix module.  New modules that require porting should come with big alarm bells to notify those working on said unsupported platforms to check this code very carefully.

joakim-noah: Sure, your edit suggestions look good.

Darredevil:
@ibuclaw thanks for the suggestions, updated accordingly

src/core/sys/posix/aio.d
```diff
@@ -0,0 +1,74 @@
+/**
+ * D header file to interface with the
+ * $(HTTP pubs.opengroup.org/onlinepubs/9699919799/basedefs/aio.h.html, Linux aio API).
```
ibuclaw: I'd still say `Posix AIO` here, as I've repeatedly said, this is _not_ a linux module. ;-)

ibuclaw:
@Darredevil - I think just one last nit and then I'll shut up.

Darredevil:
@ibuclaw should be good to go now

ibuclaw:
Ping @joakim-noah @redstar @yshui @dkgroot @rracariu 

FYI this module is in dire need of porting. ALL PLATFORMS WILL FTBFS until this is fixed.

joakim-noah:
Not a big deal, none of our platforms are tested with this master branch, if at all. Once it moves downstream, we'll test and fix it.

ibuclaw:
@joakim-noah - Yeah, depends how you build druntime also.   As a litmus test, all druntime modules are built in gdc's tree regardless of target branch.  This ensures that at least `-fsyntax-only` passes on all platforms.

dkgroot:
How about adding some unittests, that would ease porting to other platforms.
example posix aio conformance tests (in c):
http://repo.or.cz/ltp-debian.git/tree/HEAD:/testcases/open_posix_testsuite/conformance/interfaces.

The linux implementation of posix_aio is (mostly) oriented toward async non-blocking (raw) file i/o. I am not 100% sure if i remember correctly, but using it with sockets was not supported on linux (?). On other platforms posix_aio can also be used to deal with sockets (though not often used as far as i know). Should the orientation toward file i/o be noted, or should using a socket FD not be permitted for example ?

ibuclaw:
> How about adding some unittests, that would ease porting to other platforms.

That's never really been done for other modules, and I can think of a few where there's good reason for that as well.  Be it math functions that return a different result depending on the platform, or format functions where format specifiers may not exist on older systems.

In any case, I don't think its our role to test the C runtime library here, and there's too many variations between them to really be useful, and these libraries have their own testsuite anyways. ;-)

Of course, if there's something we implement, then that needs a unittest.

dkgroot:
@ibuclaw i was not looking to test the c-implementation, to to check that we present the same results on all platforms. But ok, if that is not something we normally do. I assume that once there is a phobos library using this, it would be unittested there.
I will look into porting to freebsd and dragonfly when this part is merged. Should be pretty straight forward.

PR #2075, opened 2018-02-03
Fix Issue 13844 - core.stdc.config isn't listed in the docs

wilzbach:


ZombineDev:
I wonder why it's not listed in the navigation sidebar.

PR #2076, opened 2018-02-05
Fix deprecated rwm operation in core.sync.mutex

wilzbach:
See also: https://github.com/dlang/dmd/pull/7816

First attempt: https://github.com/dlang/druntime/pull/2072 (still triggers the error)

~~~The `@trusted` lambda is needed because casting away shared isn't `@safe`.~~~
edit: Apparently, the compiler doesn't mind.


wilzbach:
So can we move ahead with this?
This is still blocking the DMD PR (and I think a rather trivial change) :/

PR #2077, opened 2018-02-07
Guard unittest for atomicLoad() of int[].

redstar:
On 64 bit platforms int[] has size 128 bit. Sadly, not all of these
platforms support cas with 128 bit, which breaks the unit test.
Example is ppc64.


redstar:
Thanks!

PR #2079, opened 2018-02-07
separate thread_init/term from gc_init/term

MartinNowak:
- make thread_init/term @nogc
- move thread_init/term from GC to rt_init/term

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -1736,7 +1736,7 @@ private:
         }
     }
 
-    static void termLocks()
+    static void termLocks() @nogc
     {
         foreach (ref lock; _locks)
             (cast(Mutex)lock.ptr).__dtor();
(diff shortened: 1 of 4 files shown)


wilzbach:
> 21.052% of diff hit (target 74.653%)

Let's make it pass -> https://github.com/dlang/druntime/pull/2080

src/core/thread.d
```diff
@@ -2090,17 +2090,28 @@ extern (C) void thread_init()
...
+extern (C) void thread_term() @nogc
 {
-    destroy(Thread.sm_main);
+    assert(_mainThreadStore.ptr is cast(void*) Thread.sm_main);
+
+    // destruct manually as object.destroy is not @nogc
+    Thread.sm_main.__dtor();
+    _d_monitordelete_nogc(Thread.sm_main);
+    if (typeid(Thread).initializer.ptr)
+        _mainThreadStore[] = typeid(Thread).initializer[];
```
schveiguy: A couple comments here:

My coverage analysis is saying this code is never run.

The normal destroy will nullify the vtable, and also if the `initializer.ptr` is null, it means you should zero the data. But I don't actually think you need to check if it's null, as it should always have a vtable in the initializer.

somzzz:
Squash please.

Then I will adjust my PR. Thanks!

andralex:
Coverage needs to be looked at but I'll merge this to unblock @somzzz who is in a bit of time crunch.

src/core/thread.d
MartinNowak: Coverage messurement fails here because this is run outside of the main loop, i.e. after coverage is collected and reported.

MartinNowak: BTW, you can easily fixup or edit PRs. Much less hassle than asking for single-line changes.

MartinNowak:
No need to squash here, these are 3 separate steps (commits) that together form one particular change.
Having them as separate commits allows to review things piecewise and leaves more context in the git history.

PR #2080, opened 2018-02-07
[Codecov] Replace threshold:5 with informational:true

wilzbach:
There are still many parts of Phobos which have a low coverage. Touching them shouldn't make a PR fail.

See also: https://github.com/dlang/phobos/pull/6124

Example from https://github.com/dlang/druntime/pull/2079

> codecov/patch — 21.052% of diff hit (target 74.653%)

Phobos and DMD already use this behavior.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -37,7 +37,7 @@ coverage:
     project: off
     patch:
       default:
-        threshold: 5
+        informational: true
     changes: off
 
   notify:


MartinNowak:
Why? Patch coverage in #2079 was low due to a very technical reason (https://github.com/dlang/druntime/pull/2079#discussion_r167901714), which apparently wasn't understood by any of the reviewers. It seems that coverage rightly failed and should have been rightly overriden by a knowledagable reviewer.
I'm a bit picky here because informal usually means tests are ignored and there aren't any good reasons why a refactoring or newly added code would come without coverage.

wilzbach:
>  It seems that coverage rightly failed and should have been rightly overriden by a knowledagable reviewer.

Yes, but I opened the PR, s.t. we can use auto-merge (failing CI!).

> I'm a bit picky here because informal usually means tests are ignored and there aren't any good reasons why a refactoring or newly added code would come without coverage.

In experience with Phobos + DMD, it's very easy to get below the threshold.
The problem is that CodeCov doesn't take the coverage of the previous diff as threshold, but the overall project coverage.
Here's another example:

https://github.com/dlang/phobos/pull/6041#issuecomment-362899073

So if you touch the wrong file or function which isn't as covered as the overall project CodeCov would mark your PR as failed.

>  Please reconsider, also it should be threshold: 5 + informal, so that the result at least still reflects the tolerance correctly.

Almost all reviewers have the CodeCov extension installed and see the red lines directly on the diff.
Also as mentioned `threshold` doesn't work for diffs - seems like a  "missing feature" from CodeCov.

MartinNowak:
> Almost all reviewers have the CodeCov extension installed and see the red lines directly on the diff.

Hope they soon manage to convert their plugin to a Firefox web extension.
https://github.com/codecov/browser-extension/issues/44

> Also as mentioned threshold doesn't work for diffs - seems like a "missing feature" from CodeCov.

Mmh, that's annoying.

PR #2082, opened 2018-02-08
Use REF1 for references to object

wilzbach:
From https://github.com/dlang/dlang.org/blob/master/dlang.org.ddoc#L353

> The REF1 variants work around DDoc limitations and allow linking to
    modules not inside packages (i.e. object.)

Argh...

See also: https://github.com/dlang/dlang.org/pull/2191

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -956,14 +956,14 @@ $(UL
     )
 )
 
-Note: Users should prefer $(REF destroy, object)` to explicitly finalize objects,
+Note: Users should prefer $(REF1 destroy, object) to explicitly finalize objects,
 and only resort to $(REF __delete, core,memory) when $(REF destroy, object)
 wouldn't be a feasible option.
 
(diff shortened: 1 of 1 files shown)


changelog/core-memory-__delete.dd
```diff
@@ -30,6 +30,6 @@ $(CONSOLE
 sed "s/delete \(.*\);/__delete(\1);/" -i **/*.d
 )
 
-Users should prefer $(REF destroy, object)` to explicitly finalize objects,
-and only resort to $(REF __delete, core,memory) when $(REF destroy, object)
+Users should prefer $(REF1 destroy, object)` to explicitly finalize objects,
```
ibuclaw: There's a dangling backtick here.

wilzbach: Fixed (as the changelog file has been removed now.)

PR #2083, opened 2018-02-08
fix Issue 18396 - make demangle work with C++ mangled symbols using __cxa_demangle

timotheecour:
useful for projects that contained `extern(C++)` declarations; eg very useful for dmd and ldc that use a ton of these.

* lazy loading: uses `dlopen` to discover whether platform supports `__cxa_demangle` when demangle encounters a symbol starting with `_Z`

* uses an interesting pattern I found to make it easy for a pure function to reuse allocated data (and do lazy initializations) while at same time providing easy to user API for callers:

```
// declaration:
char[] demangleCXX(out DemangleCXXStatus status, const(char)[] buf,
        DemangleCXX* data = Singleton!DemangleCXX(0)) nothrow pure @trusted;

// caller: doesn't have to bother with auxiliary buffer
auto ret=demangleCXX(status, buf);
```
this pattern could be reused advantageously in many places, eg existing `demangle` code (eg `ddemangle` doesn't keeps reallocating on each call; using this pattern would not reallocate)

motivating example for this PR: this stacktrace would become demangled with this PR:
```
core.exception.AssertError@dmd/e2ir.d(1063): Assertion failure
----------------
4   dmd                                 0x0000000106498b05 _d_assertp + 117
5   dmd                                 0x00000001063a2e7b _ZN6toElem13ToElemVisitor5visitEP10Expression + 95
6   dmd                                 0x000000010636f402 _ZN16ParseTimeVisitorI10ASTCodegenE5visitEP14DefaultInitExp + 30
7   dmd                                 0x000000010636f88d _Z8callfuncRK3LocP7IRStateiP4TypeP4elemS5_P15FuncDeclarationS5_S7_P5ArrayIP10ExpressionES7_ + 806
...
11  dmd                                 0x00000001063abd6f _ZN6toElem13ToElemVisitor5visitEP7CallExp + 1515
12  dmd                                 0x00000001062c3dfd _ZN7CallExp6acceptEP7Visitor + 33
22  dmd                                 0x0000000106360b2d _ZN14ScopeStatement6acceptEP7Visitor + 33
34  dmd                                 0x000000010638ef3d _Z10genObjFileP6Moduleb + 1105
35  dmd                                 0x00000001063239ba int dmd.mars.tryMain(ulong, const(char)**) + 9454
36  dmd                                 0x000000010622d676 _Dmain + 38
37  dmd                                 0x00000001064aac23 void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll().__lambda1() + 39
38  dmd                                 0x00000001064aaab3 void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) + 31
39  dmd                                 0x00000001064aab8e void rt.d
```

### DIFF
--- a/mak/DOCS
+++ b/mak/DOCS
@@ -1,54 +1,55 @@
 DOCS=\
-	$(DOCDIR)\object.html \
-	$(DOCDIR)\core_atomic.html \
-	$(DOCDIR)\core_attribute.html \
-	$(DOCDIR)\core_bitop.html \
-	$(DOCDIR)\core_checkedint.html \
-	$(DOCDIR)\core_cpuid.html \
-	$(DOCDIR)\core_demangle.html \
-	$(DOCDIR)\core_exception.html \
-	$(DOCDIR)\core_math.html \
-	$(DOCDIR)\core_memory.html \
-	$(DOCDIR)\core_runtime.html \
-	$(DOCDIR)\core_simd.html \
-	$(DOCDIR)\core_thread.html \
-	$(DOCDIR)\core_time.html \
-	$(DOCDIR)\core_vararg.html \
-	\
-	$(DOCDIR)\core_stdc_assert_.html \
-	$(DOCDIR)\core_stdc_config.html \
-	$(DOCDIR)\core_stdc_complex.html \
-	$(DOCDIR)\core_stdc_ctype.html \
-	$(DOCDIR)\core_stdc_errno.html \
-	$(DOCDIR)\core_stdc_fenv.html \
-	$(DOCDIR)\core_stdc_float_.html \
-	$(DOCDIR)\core_stdc_inttypes.html \
-	$(DOCDIR)\core_stdc_limits.html \
-	$(DOCDIR)\core_stdc_locale.html \
-	$(DOCDIR)\core_stdc_math.html \
-	$(DOCDIR)\core_stdc_signal.html \
-	$(DOCDIR)\core_stdc_stdarg.html \
-	$(DOCDIR)\core_stdc_stddef.html \
-	$(DOCDIR)\core_stdc_stdint.html \
-	$(DOCDIR)\core_stdc_stdio.html \
-	$(DOCDIR)\core_stdc_stdlib.html \
-	$(DOCDIR)\core_stdc_string.html \
-	$(DOCDIR)\core_stdc_tgmath.html \
-	$(DOCDIR)\core_stdc_time.html \
-	$(DOCDIR)\core_stdc_wchar_.html \
-	$(DOCDIR)\core_stdc_wctype.html \
-	\
-	$(DOCDIR)\core_stdcpp_exception.html \
-	$(DOCDIR)\core_stdcpp_typeinfo.html \
-	\
-	$(DOCDIR)\core_sync_barrier.html \
-	$(DOCDIR)\core_sync_condition.html \
-	$(DOCDIR)\core_sync_config.html \
-	$(DOCDIR)\core_sync_exception.html \
-	$(DOCDIR)\core_sync_mutex.html \
-	$(DOCDIR)\core_sync_rwmutex.html \
-	$(DOCDIR)\core_sync_semaphore.html \
-	\
+    $(DOCDIR)\object.html \
+    $(DOCDIR)\core_atomic.html \
+    $(DOCDIR)\core_attribute.html \
+    $(DOCDIR)\core_bitop.html \
+    $(DOCDIR)\core_checkedint.html \
+    $(DOCDIR)\core_cpuid.html \
+    $(DOCDIR)\core_demangle.html \
+    $(DOCDIR)\core_demangle_cxx.html \
+    $(DOCDIR)\core_exception.html \
+    $(DOCDIR)\core_math.html \
+    $(DOCDIR)\core_memory.html \
+    $(DOCDIR)\core_runtime.html \
+    $(DOCDIR)\core_simd.html \
+    $(DOCDIR)\core_thread.html \
+    $(DOCDIR)\core_time.html \
+    $(DOCDIR)\core_vararg.html \
+    \
+    $(DOCDIR)\core_stdc_assert_.html \
+    $(DOCDIR)\core_stdc_config.html \
+    $(DOCDIR)\core_stdc_complex.html \
+    $(DOCDIR)\core_stdc_ctype.html \
+    $(DOCDIR)\core_stdc_errno.html \
+    $(DOCDIR)\core_stdc_fenv.html \
+    $(DOCDIR)\core_stdc_float_.html \
+    $(DOCDIR)\core_stdc_inttypes.html \
+    $(DOCDIR)\core_stdc_limits.html \
+    $(DOCDIR)\core_stdc_locale.html \
+    $(DOCDIR)\core_stdc_math.html \
+    $(DOCDIR)\core_stdc_signal.html \
+    $(DOCDIR)\core_stdc_stdarg.html \
+    $(DOCDIR)\core_stdc_stddef.html \
+    $(DOCDIR)\core_stdc_stdint.html \
+    $(DOCDIR)\core_stdc_stdio.html \
+    $(DOCDIR)\core_stdc_stdlib.html \
+    $(DOCDIR)\core_stdc_string.html \
+    $(DOCDIR)\core_stdc_tgmath.html \
+    $(DOCDIR)\core_stdc_time.html \
+    $(DOCDIR)\core_stdc_wchar_.html \
+    $(DOCDIR)\core_stdc_wctype.html \
+    \
+    $(DOCDIR)\core_stdcpp_exception.html \
+    $(DOCDIR)\core_stdcpp_typeinfo.html \
+    \
+    $(DOCDIR)\core_sync_barrier.html \
+    $(DOCDIR)\core_sync_condition.html \
+    $(DOCDIR)\core_sync_config.html \
+    $(DOCDIR)\core_sync_exception.html \
+    $(DOCDIR)\core_sync_mutex.html \
+    $(DOCDIR)\core_sync_rwmutex.html \
+    $(DOCDIR)\core_sync_semaphore.html \
+    \
     $(DOCDIR)\rt_aaA.html \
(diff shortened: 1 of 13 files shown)


ibuclaw:
Not particularly useful for non-gnu ABIs. Wouldn't you just use a better tool for demangling though?

timotheecour:
> Wouldn't you just use a better tool for demangling though?

such as?

ibuclaw:
c++filt?

timotheecour:
(I'm about to push an implementation based on dlopen to be more pay-as-you-go)

c++filt would require shelling out, that seems inefficient comparing to simply calling __cxa_demangle ; is there a library version of c++filt ?

`demangle` needs to work in demangling backtraces (eg for exception handlers), where performance is important; among other use cases

timotheecour:
PTAL, now avoiding all the drawbacks of previous implementation:
* no need to hardcode linker flags (eg `-lc++abi`)
* instead, pay-as-needed with lazy symbol loading upon 1st encounter of "_Z" prefix
* `demangleCXX` now reuses buffers so no reallocations needed after each call
* the trick to make it work with `pure` was to use this signature:
```
char[] demangleCXX(out DemangleCXXStatus status, const(char)[] buf,
        DemangleCXX* data = Singleton!DemangleCXX(0)) nothrow pure @trusted
```

mak/DOCS
```diff
@@ -5,12 +5,14 @@ DOCS=\
...
 	$(DOCDIR)\core_cpuid.html \
-	$(DOCDIR)\core_demangle.html \
+    $(DOCDIR)\core_demangle.html \
+	$(DOCDIR)\core_demangle_cxx.html \
 	$(DOCDIR)\core_exception.html \
 	$(DOCDIR)\core_math.html \
 	$(DOCDIR)\core_memory.html \
 	$(DOCDIR)\core_runtime.html \
-	$(DOCDIR)\core_simd.html \
+    $(DOCDIR)\core_simd.html \
```
wilzbach: Indentation

src/core/sharedlib.d
```diff
@@ -0,0 +1,43 @@
+module core.sharedlib;
```
wilzbach: Shouldn't this reside in the `utils` package or being `private` until we are sure on the API exposed?

src/core/demangle_cxx.d
```diff
@@ -0,0 +1,152 @@
...
+    data.buffer2 = ret[0 .. len];
+
+    switch (status2)
+    {
+    case 0:
+        {
+            status = DemangleCXXStatus.success;
+            assert(len > 0 && ret[len - 1] == '\0');
+            return ret[0 .. len - 1];
+        }
```
wilzbach: You are aware that you don't need the extra scope here?

src/core/demangle_cxx.d
```diff
@@ -0,0 +1,152 @@
+module core.demangle_cxx;
+
+version (OSX)
+    enum libCXXABI = "libc++abi.dylib";
```
jacob-carlborg: I'm not sure when this library became available. We support macOS back to Lion (10.7).

jacob-carlborg:
What about all other platforms?

src/core/demangle_cxx.d
timotheecour: how about i try these in order until a match is found?
```
"libc++abi.dylib"
"libc++.dylib"
"libstdc++.dylib"
```
?

ibuclaw:
`__cxa_demangle` just forwards to libiberty anyway, so you'll be pulling in a lot more libraries along the way.

I'm not sure if adding a new dependency to druntime is really desirable.

ibuclaw:
> is there a library version of c++filt ?

Yes, it uses libiberty. The demangler in the library supports many languages, and has a better D demangler compared to core.demangle when I last checked also.

src/core/demangle_cxx.d
jacob-carlborg: Perhaps,  I don't know what is the best solution.

mak/DOCS
timotheecour: done

src/core/demangle_cxx.d
timotheecour: done

src/core/sharedlib.d
timotheecour: done, after slack discussion, moved to core.internal.sharedlib ; and also moved other reusable utilities under core.internal

src/core/demangle_cxx.d
timotheecour: done, improved logic to make it work on a best effort basis on osx and linux; trivial to add new search locations in new design

timotheecour:
> What about all other platforms?

currently support on OSX and linux; other platforms can be added in future PR's; should be a 1-liner change in this new design; of course for windows it'll be more complex but out of scope for this PR

timotheecour:
@ibuclaw 
> __cxa_demangle just forwards to libiberty anyway, so you'll be pulling in a lot more libraries along the way.
on OSX (and IIRC) `__cxa_demangle` should be always available (via libc++abi, libc++ or libstdc++); I don't think that's the case for `libiberty` at least on OSX it doesn't seem to be there by default.

Support for more languages (eg objective C) that libiberty supports could be added in future PR's; (I wrote it in a way that allows re-usability).

> I'm not sure if adding a new dependency to druntime is really desirable.

* I'm using a lazy symbol loading upon 1st encounter of `_Z` 
* also, it's not a required dependency; if `__cxa_demangle` is not found, nothing will break

jacob-carlborg:
Do the symbols in the `internal` package need to be `public`, or can they be `package`?

timotheecour:
> Do the symbols in the internal package need to be public, or can they be package?

they are public (this is pre-existing to this PR) and I think that's a practical thing: allows client code to reuse it (keep things DRY) with caveat that API in core.internal may change. I made it clear by adding comment in `core.internal.package.d`

jacob-carlborg:
> they are public (this is pre-existing to this PR) and I think that's a practical thing: allows client code to reuse it (keep things DRY) with caveat that API in core.internal may change. I made it clear by adding comment in core.internal.package.d

The following files are completely new in this PR:

```
core/internal/adapted/package.d
core/internal/adapted/string.d
core/internal/package.d
core/internal/sharedlib.d
core/internal/util.d
```

Just because the existing files are `public` doesn't mean that was the correct decision. Keep in mind that the `package` protection recently got enhanced to allow specifying which package.

src/core/demangle.d
```diff
@@ -2009,20 +2011,24 @@ pure @safe:
...
- *  The demangled name or the original string if the name is not a mangled D
- *  name.
+ *  The demangled name or the original string if the name is not a recognized mangled name.
  */
 char[] demangle( const(char)[] buf, char[] dst = null ) nothrow pure @safe
 {
-    //return Demangle(buf, dst)();
+    // TODO: also report which type: cpp,D,unrecognized
+    // TODO: use same technique as used in demangleCXX for `dst`
+    DemangleCXXStatus status;
```
rainers: I'd recommend not mixing this into the D demangler. Instead keep it in it's own module, and add a wrapper that forwards to the appropriate demangler.

src/core/demangle.d
```diff
@@ -2560,6 +2566,7 @@ version(unittest)
     foreach( i; staticIota!(table.length) )
     {
         enum r = demangle( table[i][0] );
+        // NOTE: can't be run at compile time anymore
```
rainers: That would probably be a nogo, but would also be avoided by a better separation.

src/core/internal/adapted/string.d
```diff
@@ -0,0 +1,70 @@
...
+     * strings, and the storage allocator will put a 0 past the end
+     * of newly allocated char[]'s.
+     */
+    immutable p = s.ptr + s.length;
+    // Is p dereferenceable? A simple test: if the p points to an
+    // address multiple of 4, then conservatively assume the pointer
+    // might be pointing to a new block of memory, which might be
+    // unreadable. Otherwise, it's definitely pointing to valid
+    // memory.
+    if ((cast(size_t) p & 3) && *p == 0)
```
rainers: This looks fishy, e.g. with memory views that have odd sizes or if the trailing 0 is referenced elsewhere and still mutable.

src/core/demangle.d
timotheecour: actually this comment was a leftover, i had fixed that issue in the meantime; removed comment.

src/core/demangle.d
timotheecour: I thought about this but the advantage of doing it in demangle is that all existing tools will automatically benefit from multi-language demangling. I can however add a demangleD (analog to demangleCXX) if needed to only work on D demangling. That would be more symmetrical.

src/core/internal/adapted/string.d
timotheecour: as noted in `core.internal.adapted`, anything I put under `core.internal.adapted` is straight up copied from phobos equivalent (and adapted to avoid pulling all dependencies)

ibuclaw:
> I'd recommend not mixing this into the D demangler.

I'm not convinced that druntime should support anything more than our own language. At least everything that is part of `core.*` that isn't a C/C++ binding.

> I'm using a lazy symbol loading upon 1st encounter of _Z

Maybe at best it should be part of `etc.c` then.

Maybe something more agreeable could be turning `core.demangle` into an interface that allows the user to pass their own type to handle demangling if they want to extend or use some external library for demangling symbols or stack traces.

timotheecour:
like `traceHandler` ?
what do u have in mind so that it's trivial for all the tools that use `demangle` to just work, including for 3rd party libraries over which u have no control (including static libraries)
eg:
* backtrace
* ddemangle

timotheecour:
> Yes, it uses libiberty. The demangler in the library supports many languages, and has a better D demangler compared to core.demangle when I last checked also.

cc @ibuclaw how up to date is libiberty's support for D demangling?
the patches you mentioned in https://forum.dlang.org/post/lbhooq$apv$1@digitalmars.com seem to be 4 y old and IIRC D demangling has been updated since then

ibuclaw:
I haven't implemented back references, but there have been a number of updates since then, there haven't been so many new additions to the language in that time.

PR #2085, opened 2018-02-11
posix.mak: Add check for trailing whitespace from phobos

ibuclaw:
Begin copying all the `style_lint` stuff from phobos.  It's better to apply this one at a time, as there are _many_ lint failures that get flagged up.


wilzbach:
> as there are many lint failures that get flagged up.

Yeah, DRuntime is super-inconsistent :/
BTW @timotheecour suggested to use `dfmt` for the git diff to do an incremental update like `git clang-format`: https://github.com/dlang-community/dfmt/issues/159, but the way I see it still requires a bit of work to get `dfmt` to be able to handle all special cases.
FWIW even the LDC people use this `clang-format`.

PR #2086, opened 2018-02-13
Duration examples fail to run on dlang.org without import.

smandy:
The issue... 

go to ...

https://dlang.org/phobos/core_time.html

Search for 'edit' ( i.e. find the first runnable example ). Run the example, for me anyway example is broken.

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -475,6 +475,8 @@ unittest
 
     Examples:
 --------------------
+import std.datetime;
+
 assert(dur!"days"(12) == dur!"hnsecs"(10_368_000_000_000L));
 assert(dur!"hnsecs"(27) == dur!"hnsecs"(27));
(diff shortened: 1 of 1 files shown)


jmdavis:
The example was written long before we put any import statements in examples, and because it involves std.datetime, and it's in druntime, it can't be a ddoc-ed unittest, which is why the circleci stuff has never complained about it.

wilzbach:
> which is why the circleci stuff has never complained about it.

I just realized that the publictests checking is only done for Phobos:

https://github.com/dlang/druntime/pull/2087

jmdavis:
Well, it wouldn't have helped in this case. It also doesn't affect druntime all that much, since most of druntime's public modules don't depend on each other much. But it doesn't hurt to add a check.

PR #2087, opened 2018-02-13
Add publictests test checker to druntime

wilzbach:
I'm not sure whether this is worth the effort for DRuntime as the examples are pretty simple (compared to Phobos) and we have never had major problems here.


MoonlightSentinel:
Superseeded by #3069

PR #2088, opened 2018-02-13
Solaris: fix errno.

nykytenko:
Add Solaris errno correct interface.


schveiguy:
I have no idea what solaris requires, but I'm inclined to pull this just because I can't see how one could mess this up.

PR #2090, opened 2018-02-13
Solaris: add idtype_t visibility.

nykytenko:


wilzbach:
Don't you have to import it in `core.thread` too?

There has been a silent leakage of selective symbols, but this has been fixed with the upcoming release:

https://dlang.org/changelog/pending.html#fix18219

nykytenko:
Thank, wilzbach. Done.

PR #2091, opened 2018-02-13
zero class storage when initializer is null

MartinNowak:
- all zero initializers are not stored in the typeinfo although
  classes currently always come with a non-zero vtable

fixup for #2079

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -2104,6 +2104,8 @@ extern (C) void thread_term() @nogc
     _d_monitordelete_nogc(Thread.sm_main);
     if (typeid(Thread).initializer.ptr)
         _mainThreadStore[] = typeid(Thread).initializer[];
+    else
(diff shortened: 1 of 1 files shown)


schveiguy:
> although classes currently always come with a non-zero vtable

Do we even need the null check then? That was part of my point also. In any case, this is at least more correct.

src/core/thread.d
```diff
@@ -2112,6 +2112,8 @@ extern (C) void thread_term() @nogc
     _d_monitordelete_nogc(Thread.sm_main);
     if (typeid(Thread).initializer.ptr)
         _mainThreadStore[] = typeid(Thread).initializer[];
+    else
+        _mainThreadStore[] = 0;
```
ZombineDev: You need a cast here as `_mainThreadStore` is `void[]`.

PR #2093, opened 2018-02-14
[WIP] core.array.staticArray static array litteral: `staticArray(1, 2, 3)`

timotheecour:
## usage
```
auto a = staticArray(1, 2, 3);
float[2] = staticArray!float(1, 2);
```

## motivation
places like here: https://github.com/dlang/dub/pull/1377#discussion_r168262426

## notes
* NOTE: in core.runtime so it can be used in more settings including dmd compiler which can't use phobos

* NOTE: other (stalled) PR's have proposed something like `staticArray(T)(T[] a)` ; I suggest this could be done instead (in another PR) via (something like) `toStaticArray(T)(T[]a)` to avoid name ambiguity

## links
https://issues.dlang.org/show_bug.cgi?id=8008 Issue 8008 - Syntax for fixed size array literals like [1,2,3]s (edit)
https://issues.dlang.org/show_bug.cgi?id=481 Issue 481 - Fixed-length arrays with automatically computed length (edit)
https://github.com/dlang/phobos/pull/4936 Fix Issue 16745 - Add template helper for creating static arrays with the size inferred #4936
https://github.com/dlang/phobos/pull/4090 Better static array support in std.array #4090

## not related
https://issues.dlang.org/show_bug.cgi?id=9165 Issue 9165 - Auto conversion from dynamic array to fixed size array at return (edit)

## TODO
* update the dreadful `mak/*`
* decide whether to put in std.array or core.array
* decide whether both staticArray and asStatic are needed or whether asStatic is sufficient
* probably need to add asStaticCast as well for symmetry (is there a better way to avoid the code dup?)
* should i make rename `staticArrayCast` as `staticArray` so that there's no user-looking difference bw case where an explicit cast is needed vs not.

### DIFF
--- a/src/core/array.d
+++ b/src/core/array.d
@@ -2,7 +2,10 @@ module core.array;
 
 import core.internal.traits : Unqual, CommonType;
 
-@safe nothrow pure:
+debug import std.stdio;
+
+// @nogc: https://issues.dlang.org/show_bug.cgi?id=18439
+nothrow @safe pure:
 /++
     Returns a static array constructed from `a`. The type of elements can be
     specified implicitly (`int[2] a = staticArray(1,2);`) or explicitly
@@ -15,6 +18,17 @@ pragma(inline, true) U[T.length] staticArray(U = CommonType!T, T...)(T a) @nogc
     return [a];
 }
 
+// D20180214T185602: Workaround https://issues.dlang.org/show_bug.cgi?id=16779 (make alias to staticArray once fixed)
+pragma(inline, true) U[T.length] staticArrayCast(U = CommonType!T, T...)(T a) @nogc
+{
+    enum n = T.length;
+    U[n] ret = void;
+    static foreach (i; 0 .. n)
+        cast(Unqual!U)(ret[i]) = cast(U)(a[i]);
+    return ret;
+}
+
+///
 unittest
 {
     {
@@ -34,6 +48,18 @@ unittest
         auto a = staticArray!float(1, 2);
         assert(is(typeof(a) == float[2]));
         assert(a == [1, 2]);
+
+    }
+    {
+        // see D20180214T185602
+        // auto a = staticArray!byte(1, 2);
+        auto a = staticArrayCast!byte(1, 2);
+        assert(is(typeof(a) == byte[2]));
+        assert(a == [1, 2]);
+    }
+    {
+        auto a = staticArrayCast!byte(1, 129);
+        assert(a == [1, -127]);
     }
 
     assert(is(typeof(staticArray([1])) == int[][1]));
@@ -54,7 +80,8 @@ pragma(inline, true) T[n] asStatic(T, size_t n)(auto ref T[n] arr) @nogc
     return arr;
 }
 
-U[n] asStatic(U, T, size_t n)(auto ref T[n] arr) @nogc if (!is(U == T) && is(T : U))
+//U[n] asStatic(U, T, size_t n)(auto ref T[n] arr) @nogc if (!is(U == T) && is(T : U))
+U[n] asStatic(U, T, size_t n)(auto ref T[n] arr) @nogc if (!is(U == T))
 {
     U[n] ret = void;
     static foreach (i; 0 .. n)
@@ -83,6 +110,7 @@ auto asStatic(alias arr)() @nogc
     return ret;
 }
 
+///
 unittest
 {
     {
@@ -103,6 +131,9 @@ unittest
 
     {
         auto a = [1, 2, 3].asStatic!int;
+        // https://issues.dlang.org/show_bug.cgi?id=16779
+        //auto a2 = [1, 2, 3].asStatic!byte;
+        //auto a3 = [1, 2, 3].asStatic!ubyte;
     }
     {
         auto a1 = [1, 2, 3].asStatic!(const(int));
--- a/src/core/array.d
+++ b/src/core/array.d
@@ -24,17 +24,65 @@ unittest
     }
     assert(!__traits(compiles, staticArray(1, "")));
     assert(is(typeof(staticArray()) == void[0]));
-    // NOTE: `int[] temp=staticArray(1,2)` correctly issues a deprecation
 
     {
         auto a = staticArray!float(1, 2);
         assert(is(typeof(a) == float[2]));
         assert(a == [1, 2]);
     }
+
+    assert(is(typeof(staticArray([1])) == int[][1]));
+
+    // NOTE: correctly issues a deprecation
+    //int[] a2 = staticArray(1,2);
+}
+
+pragma(inline, true) T[n] asStatic(T, size_t n)(auto ref T[n] arr) @nogc
+{
+    return arr;
+}
+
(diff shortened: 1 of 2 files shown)


src/core/array.d
```diff
@@ -0,0 +1,61 @@
...
+    Returns a static array constructed from `a`. The type of elements can be
+    specified implicitly (`int[2] a = staticArray(1,2)`) or explicitly
+    (`float[2] a = staticArray!float(1,2)`).
++/
+pragma(inline, true) U[T.length] staticArray(U = CommonType!T, T...)(T a)
+{
+    return [a];
+}
+
+unittest
```
wilzbach: Make the unittest public, s.t. it's visible on the docs.

src/core/array.d
```diff
@@ -0,0 +1,61 @@
+module core.array;
+
+/++
+    Returns a static array constructed from `a`. The type of elements can be
+    specified implicitly (`int[2] a = staticArray(1,2)`) or explicitly
+    (`float[2] a = staticArray!float(1,2)`).
```
wilzbach: Params + Returns

src/core/array.d
```diff
@@ -0,0 +1,61 @@
...
+    {
+        auto a = staticArray(1, 2, 3);
+        assert(is(typeof(a) == int[3]));
+        assert(a == [1, 2, 3]);
+    }
+    {
+        auto a = staticArray(1, 2.0);
+        assert(is(typeof(a) == double[2]));
+        assert(a == [1, 2.0]);
+    }
```
wilzbach: Simply create new unittest blocks instead of scoping. It doesn't hurt and allows the user of the documentation to hit "run" and "edit" for an individual block. Plus you can give a short summary title in `/// Amazing example`

src/core/array.d
```diff
@@ -0,0 +1,61 @@
...
+
+    {
+        auto a = staticArray!float(1, 2);
+        assert(is(typeof(a) == float[2]));
+        assert(a == [1, 2]);
+    }
+}
+
+package:
+// copied from std.traits ; TODO: move to core.internal.adapted?
```
wilzbach: Well, imho `std.traits` and `std.meta` should have been part of druntime, but yeah for nor `core.internal.traits` is the place to dump traits from Phobos.

src/core/array.d
wilzbach: Also add `@nogc @safe nothrow and pure`. A static array should be usable in the full set.

wilzbach:
Missing comments from Andrei:

- [ ] Add an overload that takes a range and is meant to be evaluated during compilation
- [ ] Add documentation explaining the tradeoffs of the rvalue return
- [ ] add constraint if (is(T : TargetElemT))

jmdavis:
Well, this implementation is going to have problems with VRP, but until https://issues.dlang.org/show_bug.cgi?id=16779 is fixed, we don't have a way around that.

Regardless though, this isn't the sort of thing that normally goes in druntime. Typically, only the stuff that needs to go in druntime goes in druntime, and this isn't needed in druntime. It belongs in Phobos.

timotheecour:
done:Add an overload that takes a range and is meant to be evaluated during compilation
done:Add documentation explaining the tradeoffs of the rvalue return
done:add constraint if (is(T : TargetElemT))

> Regardless though, this isn't the sort of thing that normally goes in druntime. Typically, only the stuff that needs to go in druntime goes in druntime, and this isn't needed in druntime. It belongs in Phobos.

I was hesitationg as well. Would be useful to have access to it in compiler, which uses druntime and not phobos. Let's delay that decision until rest of PR is good so i don't keep moving files around

skl131313:
> I was hesitationg as well. Would be useful to have access to it in compiler, which uses druntime and not phobos. Let's delay that decision until rest of PR is good so i don't keep moving files around

All of phobos would be useful to have access in the compiler. Should everything be backported to druntime then? If the answer is no, then this shouldn't be included either. If you want access to phobos-like features then an effort should be be made into getting phobos running in the compiler. Don't know DMD's structure that well but if it isn't possible then a separate library that only DMD uses should be created instead.

src/core/array.d
timotheecour: > Make the unittest public, s.t. it's visible on the docs.
done

wilzbach:
>  If you want access to phobos-like features then an effort should be be made into getting phobos running in the compiler.

There's no effort required for this, it's a deliberate decision from Walter to explicitly avoid using Phobos. Phobos could already be used at `dmd` -  in fact the `HOST_DMD` even links with it, but the linker strips it out as it's not used.

skl131313:
Oh I thought it wasn't used cause it wasn't possible. Then this definitely shouldn't go into druntime. If features like these are desired in DMD then someone needs to convince for phobos to be used in DMD.

timotheecour:
ok @skl131313 u convinced me, will reopen a PR after all these issues are cleared

>someone needs to convince for phobos to be used in DMD

yes! I completely disagree with the decision for dmd not being allowed to use phobos btw, that just creates all this ugly guesswork and copies in core.internal

src/core/array.d
```diff
@@ -0,0 +1,161 @@
...
+debug import std.stdio;
+
+// @nogc: https://issues.dlang.org/show_bug.cgi?id=18439
+nothrow @safe pure:
+/++
+    Returns a static array constructed from `a`. The type of elements can be
+    specified implicitly (`int[2] a = staticArray(1,2);`) or explicitly
+    (`float[2] a = staticArray!float(1,2)`).
+
+    The result is an rvalue, therefore uses like staticArray(1, 2, 3).find(x) may be inefficient.
```
skl131313: Maybe providing some links to what an rvalue is or explaining the inefficiency a bit better. Just from the stand point of someone that isn't familiar with what rvalues are and why this might be inefficient. Using `find(staticArray(...), x)` might better illustrate that point as well.

timotheecour:
@jmdavis 
> Well, this implementation is going to have problems with VRP, but until https://issues.dlang.org/show_bug.cgi?id=16779 is fixed, we don't have a way around that.

added `staticArrayCast` which allows casting

src/core/array.d
timotheecour: > Maybe providing some links to what an rvalue is 

i think that's out of scope; other functions mentioning rvalue don't explain either

> explaining the inefficiency a bit better. 
done

> find(staticArray(...), x) 

changed it to fun(staticArray(...))

src/core/array.d
timotheecour: i think this is too much boilerplate and doesn't add anything in this case; looking at https://dlang.org/phobos/std_array.html#array , functions where it's obvious from rest of ddoc don't do it either.

src/core/array.d
timotheecour: pushing back on this request as well, but happy to revisit if you have good arguments
* creates less visual clutter to group (and takes less vertical space both in rendered doc and in source code); the inline comments will still be visible in rendered docs
* (maybe this one isn't an important point) potentially less executable / compile time / runtime bloat

timotheecour:
after feedback from reviewers, moved to a new PR under phobos: https://github.com/dlang/phobos/pull/6178

PR #2094, opened 2018-02-15
Base for Switch Template Search

somzzz:
This adds a base template argument for the binary search in the switch template.
Removes an extra comparison against -1 on this branch: ```else static if (caseLabels.length < 7)```

Needs to be pulled at the same time with the dmd patch or it's going to break everything.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3879,6 +3879,13 @@ template _arrayOp(Args...)
  * Returns:
  *      index of match in caseLabels, -1 if not found
 */
+
+// Hack to see tests passing before merging => remove me asap
+int __switch(T, caseLabels...)(/*in*/ const scope T[] condition) pure nothrow @safe @nogc
+{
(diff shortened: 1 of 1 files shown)


schveiguy:
Question: what is the actual test in the resulting code? I'm not a compiler savvy dev, so I don't know exactly what code is being generated. Is it checking specifically for -1, or checking for just a negative value?

If it's just checking for a negative value, why not return `int.min` instead of -1?  And then you can add the base without having to check anything, and not have to bloat the template with the "base" parameter.

schveiguy:
OK, looking at the generated AST, it simply turns it into a switch on the integer results, with each string mapping to the index of the item in the sorted list.

So my mechanism will work just fine.

schveiguy:
> why not return int.min instead of -1

Put it in a PR: #2095

schveiguy:
#2095 was merged, so this is no longer necessary.

PR #2096, opened 2018-02-17
Last remaining Musl declarations needed for Phobos

joakim-noah:
The first and last are needed for higher-precision linux-only tests of `std.file.setTimes`, while `std.socket` and `std.net.curl` tests broke because of the wrong type in `sockaddr`.

@yshui, let me know if you need anything else in.

### DIFF
--- a/src/core/sys/posix/sys/socket.d
+++ b/src/core/sys/posix/sys/socket.d
@@ -1829,7 +1829,7 @@ else version( CRuntime_Bionic )
 else version( CRuntime_Musl )
 {
     alias uint socklen_t;
-    alias ubyte sa_family_t;
+    alias ushort sa_family_t;
 
     struct sockaddr
     {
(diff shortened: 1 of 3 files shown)


src/core/sys/posix/sys/socket.d
```diff
@@ -1829,7 +1829,7 @@ else version( CRuntime_Bionic )
 else version( CRuntime_Musl )
 {
     alias uint socklen_t;
-    alias ubyte sa_family_t;
+    alias ushort sa_family_t;
```
joakim-noah: See the header upstream: https://git.musl-libc.org/cgit/musl/tree/include/alltypes.h.in#n71

src/core/sys/posix/sys/stat.d
```diff
@@ -1317,6 +1317,9 @@ else version (CRuntime_Musl)
     extern (D) bool S_ISREG( mode_t mode )  { return S_ISTYPE( mode, S_IFREG );  }
     extern (D) bool S_ISLNK( mode_t mode )  { return S_ISTYPE( mode, S_IFLNK );  }
     extern (D) bool S_ISSOCK( mode_t mode ) { return S_ISTYPE( mode, S_IFSOCK ); }
+
+    int utimensat(int dirfd, const char *pathname,
+        ref const(timespec)[2] times, int flags);
```
wilzbach: Probably `out` would be better suited for this, but I doubt that it matters.

PR #2100, opened 2018-02-19
Add rest of core.*, rt.* and gc.os of uClibc definitions.

rracariu:
This is the final patch with uCLibc definitions for druntime.

This will get the druntime test runner working on an ARMv7 machine running OpenWRT. All this using a hacked LDC 1.7 compiler.

Phobos tests run also, except some missing math functions and related formatting issues re. real.

This means that you can hack your *Open*WRT ARM router and run D on it!

### DIFF
--- a/src/rt/backtrace/dwarf.d
+++ b/src/rt/backtrace/dwarf.d
@@ -12,11 +12,12 @@
 
 module rt.backtrace.dwarf;
 
-version(CRuntime_Glibc) version = glibc_or_bsdlibc;
-else version(FreeBSD) version = glibc_or_bsdlibc;
-else version(DragonFlyBSD) version = glibc_or_bsdlibc;
+version(CRuntime_Glibc) version = has_backtrace;
+else version(FreeBSD) version = has_backtrace;
+else version(DragonFlyBSD) version = has_backtrace;
+else version(CRuntime_UClibc) version = has_backtrace;
 
-version(glibc_or_bsdlibc):
+version(has_backtrace):
 
 import rt.util.container.array;
 import rt.backtrace.elf;
--- a/src/rt/qsort.d
+++ b/src/rt/qsort.d
@@ -87,6 +87,21 @@ else version (Darwin)
         return a;
     }
 }
+else version (CRuntime_UClibc)
+{
+    alias extern (C) int function(scope const void *, scope const void *, scope void *) __compar_d_fn_t;
+    extern (C) void qsort_r(scope void *base, size_t nmemb, size_t size, __compar_d_fn_t cmp, scope void *arg);
+
+    extern (C) void[] _adSort(return scope void[] a, TypeInfo ti)
+    {
+        extern (C) int cmp(scope const void* p1, scope const void* p2, scope void* ti)
+        {
+            return (cast(TypeInfo)ti).compare(p1, p2);
+        }
+        qsort_r(a.ptr, a.length, ti.tsize, &cmp, cast(void*)ti);
+        return a;
+    }
+}
 else
 {
     private TypeInfo tiglobal;
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -15,6 +15,7 @@ else version (CRuntime_Musl) enum SharedELF = true;
 else version (FreeBSD) enum SharedELF = true;
 else version (NetBSD) enum SharedELF = true;
 else version (DragonFlyBSD) enum SharedELF = true;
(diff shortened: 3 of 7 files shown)


src/rt/qsort.d
```diff
@@ -87,6 +87,21 @@ else version (Darwin)
...
+    extern (C) void[] _adSort(return scope void[] a, TypeInfo ti)
+    {
+        extern (C) int cmp(scope const void* p1, scope const void* p2, scope void* ti)
+        {
+            return (cast(TypeInfo)ti).compare(p1, p2);
+        }
+        qsort_r(a.ptr, a.length, ti.tsize, &cmp, cast(void*)ti);
+        return a;
+    }
+}
```
wilzbach: That's a lot of duplication - to me it looks like all this special casing isn't needed. A follow-up PR could try to remove it and see if the CIs break...

src/rt/sections_elf_shared.d
```diff
@@ -106,6 +107,7 @@ private:
     invariant()
     {
         assert(_moduleGroup.modules.length);
+        version (CRuntime_UClibc) {} else
```
wilzbach: Ehm that looks wrong. What's the motivation for doing this? Does uclibc have issues with TLS?

wilzbach:
> This means that you can hack your OpenWRT ARM router and run D on it!

Awesome. I'm looking forward to a blog article! For the DBlog, you can give Mike (https://github.com/mdparker) a quick ping. He's excellent in guiding you towards a kickass article.

src/rt/backtrace/dwarf.d
```diff
@@ -15,6 +15,7 @@ module rt.backtrace.dwarf;
 version(CRuntime_Glibc) version = glibc_or_bsdlibc;
 else version(FreeBSD) version = glibc_or_bsdlibc;
 else version(DragonFlyBSD) version = glibc_or_bsdlibc;
+else version(CRuntime_UClibc) version = glibc_or_bsdlibc;
```
joakim-noah: Think about renaming this to something like `has_backtrace`.

joakim-noah:
And yeah, a blog post would be great, pinging @mdparker.

src/rt/sections_elf_shared.d
rracariu: The debug version of the druntime test fails there as the `_tlsMod` is 0 for that particular case.

TLS is working otherwise, but I haven't spent much time on this particular issue. The issue is linked to shared objects as they are not working at all (shared builds of druntime don't work.)

src/rt/backtrace/dwarf.d
rracariu: Could be done in another PR. Like to keep this simple and focused on adding the new defs.

rracariu:
Blog post will sure be nice, but I need to push this all the way, like have a working compiler so people can test whatever the blog post says.

This means pushing this through so it lands in LDC and push the required changes for LDC and make it work smoothly.

ibuclaw:
And GDC. 🙂

mdparker:
@rracariu I'm ready for a blog post whenever you are. Ping me any time :-)

src/rt/backtrace/dwarf.d
joakim-noah: Would rather you change it now, nobody ever comes back and changes this stuff later. Renaming a single variable hardly makes this pull complex. :smile:

src/rt/sections_elf_shared.d
joakim-noah: Yeah, this test simply checks to make sure shared libraries register their TLS data in a certain way, which may be different for UClibc.  I suppose if shared libraries are not working yet, it's okay to just skip this for now.

joakim-noah:
>This means pushing this through so it lands in LDC and push the required changes for LDC and make it work smoothly.

Did you have to change ldc much?  Once these changes for the stdlib are merged here, it's pretty straightforward to get them into ldc, so any compiler changes will really be all that you need.

src/rt/backtrace/dwarf.d
rracariu: OK, renamed it.

rracariu:
After getting this merged here and then merged in LDC , getting a working compiler *should* be fairly easy. I did changed the driver part to get the uclibc env from the triple, so not much.

Dunno about GDC? Haven't followed progress with it in a while.

ibuclaw:
> Dunno about GDC? Haven't followed progress with it in a while.

All you need to do is run configure with the target you want (uclibc in this case) and make.

PR #2101, opened 2018-02-20
Fix recursive reserve call in ProtoGC

schveiguy:
Also added global scope operator to all `gc_*` calls from `ProtoGC` to clarify that they are the global functions and not members (would have potentially helped catch the original issue).

See https://github.com/dlang/druntime/pull/2057#discussion_r168266752

Ping @somzzz @andralex @klickverbot @MartinNowak 

I did not do anything about the race condition on the root/range lists, as I don't know the correct direction to choose there.

### DIFF
--- a/src/gc/impl/proto/gc.d
+++ b/src/gc/impl/proto/gc.d
@@ -63,14 +63,14 @@ class ProtoGC : GC
 
     void enable()
     {
-        gc_init_nothrow();
-        gc_enable();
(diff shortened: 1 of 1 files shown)


schveiguy:
Dammit, I just realized this should have targeted stable, I didn't realize Martin had forked after the ProtoGC thing was added. @MartinNowak can you cherry pick this?

PR #2102, opened 2018-02-20
Fix recursive reserve call in ProtoGC. Also added global scope operator to all gc_* calls from ProtoGC to clarify that they are the global functions.

wilzbach:
> Dammit, I just realized this should have targeted stable, I didn't realize Martin had forked after the ProtoGC thing was added. @MartinNowak can you cherry pick this

CC @schveiguy

Well, cherry-picking isn't ideal as it leads to more merge conflicts, but the master/stable branch-off already happened.

### DIFF
--- a/src/gc/impl/proto/gc.d
+++ b/src/gc/impl/proto/gc.d
@@ -63,14 +63,14 @@ class ProtoGC : GC
 
     void enable()
     {
-        gc_init_nothrow();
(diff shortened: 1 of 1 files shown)


schveiguy:
> cherry-picking isn't ideal

I know, I wish I had seen Martin's post about the beta before making this.

PR #2104, opened 2018-02-21
[Trivial] Fix deprecated use of delete(x) in favor of x.destroy()

dkgroot:


schveiguy:
Hm... this isn't the same. `destroy` is going to do nothing to most of those things, and it certainly isn't going to "exercise each function" in the file.

dkgroot:
@schveiguy any suggestion on how to get rid of:
```
src/rt/tracegc.d(43): Deprecation: The delete keyword has been deprecated.  Use object.destroy() instead.
src/rt/tracegc.d(44): Deprecation: The delete keyword has been deprecated.  Use object.destroy() instead.
src/rt/tracegc.d(45): Deprecation: The delete keyword has been deprecated.  Use object.destroy() instead.
src/rt/tracegc.d(46): Deprecation: The delete keyword has been deprecated.  Use object.destroy() instead.
src/rt/tracegc.d(47): Deprecation: The delete keyword has been deprecated.  Use object.destroy() instead.
src/rt/tracegc.d(48): Deprecation: The delete keyword has been deprecated.  Use object.destroy() instead.
src/rt/tracegc.d(49): Deprecation: The delete keyword has been deprecated.  Use object.destroy() instead.
src/rt/tracegc.d(50): Deprecation: The delete keyword has been deprecated.  Use object.destroy() instead.
```
During compilation of druntime ?

schveiguy:
Sorry for the lack of response. Indeed, that message is not enough. What it really should say is "you should use destroy, and don't worry about freeing the memory, the GC will do it for you", as that is really the safe route. But often you want to free the memory. So really the answer is `destroy(*ptr); GC.free(ptr)`.

However, in this case, there is no allocation, so you just want to free the pointers. Destroying them does nothing. You will have the same effect as using delete (the functions will be called), and the warning should go away.

dkgroot:
@schveiguy  Thanks for the clarification !

schveiguy:
Filed an issue about the diagnostic message: https://issues.dlang.org/show_bug.cgi?id=18505

schveiguy:
As you may notice there, I forgot there is actually a function that does the same as delete: https://dlang.org/phobos-prerelease/core_memory.html#.__delete,

But I think at this point, just calling GC.free should be enough here.

dkgroot:
@schveiguy Are you going to fix/update druntime to use either one of these two solutions ?

schveiguy:
Possibly, but you could do it as well.

JinShil:
Please note that `__delete` was created to be used as a last resort when there is no other alternatives.  See https://dlang.org/changelog/pending.html#deprecate_delete

Please use `destroy()` if possible.

JinShil:
> Also, all instances of `delete` in tracegc.d are in a `version(none)` block, so how is it being called?

OK, I remember now.  It's because the `delete`, the keyword, has been deprecated, so the deprecation message is emitted by the parser.

dkgroot:
@JinShil Good question, don't have an answer for it though. Please compile druntime manually to see them popping up. Maybe deprecation does not care about version(none) ??

schveiguy:
@JinShil `destroy` does nothing here. You can call it, sure, but none of the targets have been allocated.

> Also, all instances of delete in tracegc.d are in a version(none) block,

That concerns me a bit as well, as any changes we make will not actually be tested. But at this point, I'm not worried about this code, we should just get rid of the deprecation warning.

JinShil:
>  `destroy` does nothing here

Does `delete` do anything here either?  It looks like replacing `delete` with `destroy` would do the same thing...nothing.

Also, it looks like that block of code is just a test to exercise each function.  Due to the fact that it is in a `version(none)` block, we could comment it out and we'd still be keeping the status quo.

schveiguy:
Actually, looking at the disassembly, `delete` calls the appropriate runtime functions (e.g. `_d_delarray_t`, `_d_delclass`). `__delete` doesn't do this (I get a segfault), and `GC.free` isn't going to do this properly.

So I would recommend, we just comment this whole thing out. The only other alternative is to directly call the runtime functions, which is such a low level detail that I don't think it's reasonable to assume they will not change in the future. The `version(none)` makes me just think, let's comment this whole thing out, and not worry about it right now.

ping @WalterBright as he seems to have committed this code 3 years ago.

BTW, since the commit has gone away, there is no real reference to what we are talking about. Here are the offending lines: https://github.com/dlang/druntime/blob/2d87bb9679602a0a108e06f0f16c12e69bb9c8e8/src/rt/tracegc.d#L43

wilzbach:
BTW you might also be interested in https://github.com/dlang/druntime/pull/1982 which adds `-de` to druntime, but I had a rather hard-time fixing the deprecations in the new vectorized array operations.

wilzbach:
> Actually, looking at the disassembly, delete calls the appropriate runtime functions (e.g. _d_delarray_t, _d_delclass). __delete doesn't do this (I get a segfault)

Interesting. When I first implemented `__delete`, I went with calling the runtime functions, but then changed it in course of the review at https://github.com/dlang/druntime/pull/2037

I assumed that `GC.free` checks for null pointers, but apparently it doesn't.
We should fix that -> https://github.com/dlang/druntime/pull/2108

> So I would recommend, we just comment this whole thing out. 

:+1:

PR #2105, opened 2018-02-23
Mark monitor functions as nothrow

FeepingCreature:
Synchronized is such a basic feature that you should really use errors instead of exceptions on failure.

All the default functions do this already.

This change will allow synchronized(this) to be used in nothrow functions.

### DIFF
--- a/src/core/sync/condition.d
+++ b/src/core/sync/condition.d
@@ -148,7 +148,7 @@ class Condition
      * Throws:
      *  SyncError on error.
      */
-    void wait()
+    void wait() nothrow
     {
         version( Windows )
         {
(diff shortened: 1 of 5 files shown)


FeepingCreature:
Ah, my mistake. The DMD change in https://github.com/dlang/dmd/pull/7942 needs to come first, apparently?

They should be treated as one really.

FeepingCreature:
The DMD side of the change breaks existing code, apparently, because people have decided to implement monitors that throw on lock. So the trivial solution is out. Darn.

FeepingCreature:
Looks like the issue is only in one library...?

RazvanN7:
Is this change needed anymore?

FeepingCreature:
I genuinely don't know. I haven't run into a related issue again.

Close for now?

RazvanN7:
I will close it. Feel free to reopen if needed.

PR #2106, opened 2018-02-23
[Trivial] Add dragonfly implementation for shared memory

dkgroot:
This one was overlooked (and noticed by @joakim-noah in backporting to ldc-ltsmaster)

### DIFF
--- a/src/core/sys/posix/sys/shm.d
+++ b/src/core/sys/posix/sys/shm.d
@@ -156,6 +156,32 @@ else version(NetBSD)
     int   shmdt(in void*);
     int   shmget(key_t, size_t, int);
 }
+else version(DragonFlyBSD)
+{
+    enum SHM_RDONLY     = 0x01000; // 010000
+    enum SHM_RND        = 0x02000; // 020000
+    enum SHMLBA         = 1 << 12; // PAGE_SIZE = (1<<PAGE_SHIFT)
+
+    alias c_ulong   shmatt_t;
+
+    struct shmid_ds
+    {
+         ipc_perm       shm_perm;
+         int            shm_segsz;
+         pid_t          shm_lpid;
+         pid_t          shm_cpid;
+         short          shm_nattch;
+         time_t         shm_atime;
+         time_t         shm_dtime;
+         time_t         shm_ctime;
+         private void*  shm_internal;
+    }
+
+    void* shmat(int, in void*, int);
+    int   shmctl(int, int, shmid_ds*);
+    int   shmdt(in void*);
+    int   shmget(key_t, size_t, int);
+}
 else version( Darwin )
 {
 


src/core/sys/posix/sys/shm.d
```diff
@@ -156,6 +156,32 @@ else version(NetBSD)
...
+    {
+         ipc_perm    shm_perm;
+         int         shm_segsz;
+         pid_t       shm_lpid;
+         pid_t       shm_cpid;
+         short       shm_nattch;
+         time_t      shm_atime;
+         time_t      shm_dtime;
+         time_t      shm_ctime;
+         void *      __shm_internal;
```
wilzbach: DStyle: `void*` (also do you have any resource where it's `__shm_internal`. I did a quick search, but could only find `shm_internal` like it's for FreeBSD.

dkgroot:
@wilzbach > Don't you want to target stable?
Sorry i don't understand, what do you mean ?

wilzbach:
> Sorry i don't understand, what do you mean ?

The master -> stable branch-off already happened. If you target stable, it will be part of the 2.079 tag.

src/core/sys/posix/sys/shm.d
dkgroot: On Dfly, /usr/include/sys/shm.h looks like:
```
struct shmid_ds {
    struct ipc_perm shm_perm;   /* operation permission structure */
    size_t          shm_segsz;  /* size of segment in bytes */
    pid_t           shm_lpid;   /* process ID of last shared memory op */
    pid_t           shm_cpid;   /* process ID of creator */
    shmatt_t    shm_nattch; /* number of current attaches */
    time_t          shm_atime;  /* time of last shmat() */
    time_t          shm_dtime;  /* time of last shmdt() */
    time_t          shm_ctime;  /* time of last change by shmctl() */
    void           *shm_internal;   /* sysv stupidity */
};
```
Have not looked at the use or implementation of this header file. I did have to gloss over the comment for this pointer (Logically it would be the opaque pointer to an internal representation/expansion for this struct,). But i thought size might matter (doesn't it always).

wilzbach: Thanks.
I would suggest to keep the `shm_internal` name then. You could set it to private though?

> But i thought size might matter 

Yes.

wilzbach:
For reference, when retargeting a PR to `stable` it needs to be rebased locally too:

```
git rebase --onto upstream/stable upstream/master
```

Also if you depend on this in master, you can simply run [`merge_stable`](https://github.com/wilzbach/git-tools/blob/master/merge_stable) yourself and get the changes back into master.

dkgroot:
Thanks @wilzbach / @joakim-noah

PR #1060, opened 2014-12-06
add core.sys.freebsd.pthread_np

MartinNowak:

PR #2107, opened 2018-02-23
[Trivial] Match the module name to the file name (pthread_np)

dkgroot:
This caused a unittest issue in ldc-ltsmaster. Would like to keep these files in sync.

Note: The file does not seem to be used at this moment, and was copied over from the FreeBSD implementation (which does not seem to use it's version either). Maybe instead of renaming the module name, we should opt to remove this file. Not sure what the reason for it's creation is or was.

### DIFF
--- a/src/core/sys/dragonflybsd/pthread_np.d
+++ b/src/core/sys/dragonflybsd/pthread_np.d
@@ -4,7 +4,7 @@
  * Authors: Martin Nowak,Diederik de Groot(port:DragonFlyBSD)
  * Copied:  From core/sys/freebsd/sys
  */
-module core.sys.dragonflybsd.pthread;
+module core.sys.dragonflybsd.pthread_np;
 
 version (DragonFlyBSD):
 


dkgroot:
Note: The file does not seem to be used at this moment, and was copied over from the FreeBSD implementation (which does not seem to use it's version either). Maybe instead of renaming the module name, we should opt to remove this file. Not sure what the reason for it's creation is or was.

Please let me know what you would like to happen !

wilzbach:
> Not sure what the reason for it's creation is or was.

https://github.com/dlang/druntime/pull/1060

>> Just tried to use pthread_suspend_np to fix Issue 13416. Didn't work out, but the header might still be useful to someone else.

dkgroot:
@wilzbach I don't mind if the file stays, i just want to match the filename to the module name (got some errors when running unittests in ldc  because of the name mismatch).

wilzbach:
I assume that LDC doesn't support FreeBSD (or at least no one builds it regularly for it), hence it didn't come up earlier?

dkgroot:
Thanks guys !

PR #2108, opened 2018-02-23
__delete shouldn't segfault for null

wilzbach:
See also: https://github.com/dlang/druntime/pull/2104#issuecomment-368101722

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -1010,14 +1010,16 @@ void __delete(T)(ref T x) @system
         static assert(0, "It is not possible to delete: `" ~ T.stringof ~ "`");
     }
 
-    static if (is(T == interface) || is(T == class))
+    static if (is(T == interface) ||
+              is(T == class) ||
+              is(T == U2*, U2))
     {
         GC.free(cast(void*) x);
         x = null;
     }
-    else static if (is(T == U2*, U2) || is(T : E2[], E2))
+    else static if (is(T : E2[], E2))
     {
-        GC.free(&x);
+        GC.free(x.ptr);
         x = null;
     }
 }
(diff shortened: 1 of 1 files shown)


schveiguy:
I think there is a different problem here!

This does not segfault:

```D
void main()
{
   int *x = null;
   GC.free(x);
}
```

src/core/memory.d
```diff
@@ -1012,13 +1012,19 @@ void __delete(T)(ref T x) @system
...
+            x = null;
+        }
     }
     else static if (is(T == U2*, U2) || is(T : E2[], E2))
     {
-        GC.free(&x);
-        x = null;
+        if (x !is null)
+        {
+            GC.free(&x);
```
schveiguy: This seems wrong, shouldn't it be `GC.free(x)` ?

wilzbach: Yeah, looks like I was a bit too fast. Thanks!

schveiguy:
OK, I see you fixed that problem. But now I'm slightly confused!

Because extracting out the problem:
```D
void main()
{
    int *ptr = null;
    GC.free(&ptr);
}
```

This segfaults, but this:

```D
void main()
{
   int *ptr = new int(1);
   GC.free(&ptr);
}
```

DOESN'T segfault. Neither is a valid heap pointer, so why does it segfault at all? The first thing `GC.free` should do is to find the pool the pointer is in, and it should fail in both cases.

In any case, a bug for another day. Your changes LGTM.

schveiguy:
Hah, I think I know why it was segfaulting. Because of the new ProxyGC! Attempting to free a pointer before actually allocating any pointer will do this: https://github.com/dlang/druntime/blob/master/src/gc/impl/proto/gc.d#L140, which equates to a segfault when compiled in release mode.

schveiguy:
This does fix a bug, though (`__delete` was not actually freeing the memory), so I'm pulling.

PR #2110, opened 2018-02-24
add FreeBSD github URLs

WalterBright:
Convenient when checking for correctness


wilzbach:
FreeBSD 32-bit is broken due to https://github.com/braddr/d-tester/issues/70 - merging manually...

PR #2111, opened 2018-02-24
Musl: Add missing grp.h declarations and remove kernel constants

joakim-noah:
With this patch and cherry picking all of @yshui's Musl port, everything builds and runs with the ldc 1.8 beta.  This was apparently missed before because [ldc builds all of `core.sys.posix` for its testrunner for non-Windows platforms](https://github.com/ldc-developers/ldc/blob/master/runtime/CMakeLists.txt#L161), but the default dmd makefiles build very little of it. Nothing actually there to be built of course, but this means static assert blocks and redundancies are not checked, like they are for ldc.

Might be worth doing so for dmd too, at least for the tests.

### DIFF
--- a/src/core/sys/posix/time.d
+++ b/src/core/sys/posix/time.d
@@ -408,9 +408,7 @@ else version( CRuntime_Musl )
     enum CLOCK_REALTIME = 0;
     enum CLOCK_PROCESS_CPUTIME_ID = 2;
     enum CLOCK_THREAD_CPUTIME_ID = 3;
-    enum CLOCK_MONOTONIC_RAW = 4;
     enum CLOCK_REALTIME_COARSE = 5;
-    enum CLOCK_MONOTONIC_COARSE = 6;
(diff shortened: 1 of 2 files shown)


src/core/sys/posix/time.d
```diff
@@ -408,9 +408,7 @@ else version( CRuntime_Musl )
     enum CLOCK_REALTIME = 0;
     enum CLOCK_PROCESS_CPUTIME_ID = 2;
     enum CLOCK_THREAD_CPUTIME_ID = 3;
-    enum CLOCK_MONOTONIC_RAW = 4;
     enum CLOCK_REALTIME_COARSE = 5;
-    enum CLOCK_MONOTONIC_COARSE = 6;
```
joakim-noah: Oh, these are removed because they're already declared in a `version(linux)` block at the top.

wilzbach:
> Might be worth doing so for dmd too, at least for the tests.

:+1:

PR #2112, opened 2018-02-25
Add clarification about using destroy with pointers to value types

JinShil:
See discussion here (https://github.com/dlang/dmd/pull/7947#issuecomment-368219079) for context.

Also see http://ddili.org/ders/d.en/memory.html#ix_memory.destroy

cc @schveiguy @wilzbach

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2971,6 +2971,30 @@ unittest
     assert(i == 0);           // `i` is back to its initial state `0`
 }
 
+/********
+To destroy a value type through a pointer, the pointer must be dereferenced.
+Otherwise `destroy` will only destroy the pointer itself, not the entity being
+referenced.
+*/
+nothrow @safe unittest
+{
+    static struct S
+    {
+        static int dtorCount;
+        ~this() { dtorCount++; }
+    }
+
+    S* s1 = new S();
+    destroy(s1);         // destroys only the pointer `s1`, not what it references,
+    assert(s1 is null);  // and sets the pointer to `null`
+    assert(S.dtorCount == 0);
+
+    S* s2 = new S();
+    destroy(*s2);        // destroys the struct referenced by `s2`
+    assert(s2 !is null); // the pointer itself is not affected
+    assert(S.dtorCount == 1);
+}
+
 unittest
 {
    interface I { }


JinShil:
Or do we need an overload...something like this?

```D
void destroy(T)(T* obj) if (is(T == struct))
{
	object.destroy(*obj);   
}
```

src/object.d
```diff
@@ -2971,6 +2971,31 @@ unittest
     assert(i == 0);           // `i` is back to its initial state `0`
 }
 
+/********
+To destroy a value type through a pointer, the pointer must be dereferenced.
+Otherwise `destroy` will only destroy the pointer itself, not the entity being
+pointed to.
```
schveiguy: pointed to -> referenced

src/object.d
```diff
@@ -2971,6 +2971,31 @@ unittest
...
+*/
+unittest
+{
+    struct S
```
schveiguy: static struct

src/object.d
```diff
@@ -2971,6 +2971,31 @@ unittest
...
+    {
+        static int dtorCount;
+        ~this() { dtorCount++; }
+    }
+
+    void main()
```
schveiguy: Leave this out, it's a unit test. Otherwise, the test doesn't actually get run.

src/object.d
```diff
@@ -2971,6 +2971,31 @@ unittest
     assert(i == 0);           // `i` is back to its initial state `0`
 }
 
+/********
+To destroy a value type through a pointer, the pointer must be dereferenced.
+Otherwise `destroy` will only destroy the pointer itself, not the entity being
+pointed to.
+*/
+unittest
```
schveiguy: `@safe` or `@system`? Not sure which one applies, but it needs to be tagged.

src/object.d
```diff
@@ -2971,6 +2971,31 @@ unittest
...
+    }
+
+    void main()
+    {
+        S* s1 = new S();
+        destroy(s1);      // destroys only the pointer `s1`, not what it points to.
+        assert(S.dtorCount == 0);
+
+        S* s2 = new S();
+        destroy(*s2);     // destroys the struct pointed to by `s2`
```
schveiguy: referenced by

src/object.d
```diff
@@ -2971,6 +2971,31 @@ unittest
...
+    struct S
+    {
+        static int dtorCount;
+        ~this() { dtorCount++; }
+    }
+
+    void main()
+    {
+        S* s1 = new S();
+        destroy(s1);      // destroys only the pointer `s1`, not what it points to.
```
schveiguy: clarify that destroying a pointer simply just sets it to null.

schveiguy:
> Or do we need an overload...something like this?

If we can verify that this isn't going to interfere with other things like recursive destruction, I'd be all for this. It's what people would expect I would think.

JinShil:
Added alternative `destroy` overload implementation at #2115 for your consideration.

JinShil:
> though I wonder if would be better to do the dereferencing automatically

@ZombineDev see my last comment in this PR.  I added such an overload in PR #2115

JinShil:
After the discussion in #2115, it became apparent that there are a few problems with the current implementation of `destroy`.  I will be submitting pull requests to individually address them.  See https://github.com/dlang/druntime/pull/2115#issuecomment-369081319

Depending on how that goes, this PR will either not be needed, or will need modification.  For now, there's no use for it being in the PR queue consuming resources, until the implementation of `destroy` is improved.

wilzbach:
There are no resources consumed and its easy to lose track of work if it's closed, so I strongly suggest keeping this open until it's really not needed anymore. Just set Blocked as title or label and no one will look at it casually.

PR #2116, opened 2018-02-27
fixes issue 18208 -- demangle RangeError@src/core/demangle.d(230)

timotheecour:


rainers:
See also https://github.com/dlang/druntime/pull/2062

wilzbach:
I think we should go with https://github.com/dlang/druntime/pull/2062, but we can add the testcase nonetheless after https://github.com/dlang/druntime/pull/2062 has been merged.

PR #2117, opened 2018-02-27
Pr fix 18531 demangle

timotheecour:
NOTE: depends on https://github.com/dlang/druntime/pull/2116


src/core/demangle.d
```diff
@@ -207,8 +207,9 @@ pure @safe:
         if( val.length )
         {
             assert( contains( dst[0 .. len], val ) );
+            if (len + val.length > dst.length)
+                overflow();
```
rainers: Why would removing something have to cause an overflow?

src/core/demangle.d
```diff
@@ -207,8 +207,9 @@ pure @safe:
...
             debug(info) printf( "removing (%.*s)\n", cast(int) val.length, val.ptr );
-
             size_t v = &val[0] - &dst[0];
             for (size_t p = v; p < len; p++)
```
rainers: This loop is the culprit: `len` is wrong, it should be `len-val.length`, i.e. the change to len below could just be moved up.

src/core/demangle.d
```diff
@@ -2563,6 +2573,19 @@ version(unittest)
         static assert( r == table[i][1],
                 "demangled \"" ~ table[i][0] ~ "\" as \"" ~ r ~ "\" but expected \"" ~ table[i][1] ~ "\"");
     }
+
+    {
+        auto dst = new char[50];
+        auto example = table[$-2];
```
rainers: `$-2` is not so good to use here, it will silently test something else if other symbols are added to the table.

src/core/demangle.d
timotheecour: updated

src/core/demangle.d
timotheecour: done

src/core/demangle.d
timotheecour: updated

timotheecour:
PTAL

joakim-noah:
You need to rebase.

timotheecour:
oups, done

src/core/demangle.d
```diff
@@ -208,11 +208,11 @@ pure @safe:
         {
             assert( contains( dst[0 .. len], val ) );
             debug(info) printf( "removing (%.*s)\n", cast(int) val.length, val.ptr );
-
             size_t v = &val[0] - &dst[0];
+            assert(len >=val.length && len <= dst.length);
```
rainers: Missing space in `>=val`.

rainers:
@wilzbach Do you know about the jenkins failures? Can these safely be ignored ATM?

src/core/demangle.d
timotheecour: done

rainers: Not sure if this assert is necessary, it is also implied by `contains` above.

timotheecour: it's a bit more straightforward with the assert ; shouldn't affect performance even with asserts enabled given the rest of the code in this function; if you insist I'll remove this assert

timotheecour:
> @wilzbach Do you know about the jenkins failures? Can these safely be ignored ATM?
@rainers @wilzbach 

continuous-integration/jenkins/pr-merge passes, and the only failure (auto-tester) is a known bug: https://issues.dlang.org/show_bug.cgi?id=18519

wilzbach:
> @wilzbach Do you know about the jenkins failures? Can these safely be ignored ATM?

While Jenkins does has spurious issues sometime, this one wasn't.
I fixed it with https://github.com/dlang/ci/pull/182

wilzbach:
(rebased to `stable`)

timotheecour:
@wilzbach after you rebased to stable it gives errors, eg: 
../../../../.dub/packages/eventcore-0.8.28/eventcore/source/eventcore/internal/utils.d(3,8): Error: module `taggedalgebraic` is in file 'taggedalgebraic.d' which cannot be read

import path[0] = ../../inet

wilzbach:
That looks spurious (restarted), but the auto-tester failure doesn't:

```
failed to execute '\"Program Files (x86)"\"Microsoft Visual Studio 10.0"\VC\bin\amd64\cl.exe /c /nologo runnable\extra-files\cpp_abi_tests.cpp /Fotest_results\runnable\cpp_abi_tests.cpp.obj'
Makefile:220: recipe for target 'test_results/runnable/cpp_abi_tests.d.out' failed
gmake[1]: *** [test_results/runnable/cpp_abi_tests.d.out] Error 1
gmake[1]: *** Waiting for unfinished jobs....
```

The same one appeared also on Windows before I rebased though it seems unrelated to your test ...

rainers:
> failed to execute '\"Program Files (x86)"\"Microsoft Visual Studio 10.0"\VC\bin\amd64\cl.exe

Looks like Brad is in the middle of getting rid of patching d_do_test, as this is on the patched line.

PR #2119, opened 2018-02-28
fix issue 18536 - Bad stack traces when building with -m32mscoff

rainers:
dmdDecodeString is harmful for non-mangled names, disable it for anything but OMF

### DIFF
--- a/src/core/sys/windows/stacktrace.d
+++ b/src/core/sys/windows/stacktrace.d
@@ -293,9 +293,9 @@ private:
 
         auto res = formatStackFrame(pc);
         res ~= " in ";
-const(char)[] tempSymName = symName[0 .. strlen(symName)];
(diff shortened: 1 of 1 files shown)


wilzbach:
@rainers is there a chance to add a test for this, s.t. it doesn't regress?
Maybe something similar to https://github.com/dlang/druntime/blob/master/test/exceptions/Makefile, but for Windows?

rainers:
It would have been caught if we'd run the dmd test site for -m32mscoff, too. Maybe AppVeyor should be switched to that platform instead of Win64 which is already covered by the auto tester.

wilzbach:
> It would have been caught if we'd run the dmd test suite for -m32mscoff, too. Maybe AppVeyor should be switched to that platform instead of Win64 which is already covered by the auto tester.

Can't we run both? At the the average build time is ~18 minutes which isn't great, but still has some leeway for more.

PR #2121, opened 2018-02-28
Issue 17964 - [CTFE] If array is large enough it hits __simd at CTFE

rainers:
skip executing simd instructions at compile time


rainers:
> I'll tag as auto-merge, but the auto-tester seems to be choking on something.

Windows seems to be transitioning to not being patched by the auto-tester, and FreeBSD is broken permanently atm. Only few people seem to be able to override the required checks, though, e.g. @wilzbach.

wilzbach:
Yeah I really don't like the current situation either and I strongly urged @WalterBright and @braddr to disable FreeBSD until it's fixed.

Anyhow, I got too tired too being apparently the only active person who can override the checks and remove the required checks from auto-tester. From now on everyone should be able to manually merge a PR if the auto-tester fails.

PR #2122, opened 2018-02-28
fix Issue 17188 - qsort predicate requires scope parameters

MartinNowak:
- remove scope as it's not checked for `@system` functions anyhow
- adding overloads would be an alternative approach, but it's 32
  variants to support all of `scope`, `@safe`, `pure`, `nothrow`, `@nogc`


wilzbach:
Shouldn't we add the code from the bugreport as a unittest, s.t. no further regression can occur?

PR #2123, opened 2018-03-02
[RFC] Add the core.sentinel module

marler8997:
REQUEST FOR COMMENT

In D we have APIs that accept both D arrays and pointers with sentinel values (https://en.wikipedia.org/wiki/Sentinel_value).  The classic example is the string, where some functions accept D's `string` type and others accept null-terminated strings (i.e. the C standard library functions).  This change proposes the addition of the `SentinelPtr` and `SentinelArray` types.
```D
// A pointer to an array with elements of type `T` terminated by `sentinelValue`
template SentinelPtr(T, T sentinelValue = defaultSentinel!T) { ... }

// An standard D array that, along with a length is also terminated by `sentinelValue`.
// This type can safely be used as a standard D array and its `ptr` field is also
// a valid SentinelPtr.
template SentinelArray(T, T sentinelValue = defaultSentinel!T) { ... }

// C string
alias cstring = SentinelPtr!(const(char));
```

A `SentinelPtr` is a pointer to an array that contains a sentinel value to terminate its elements (i.e. a C string).  A `SentinelArray` is a D array whose pointer is a `SentinelPtr`.  D's "string literal" is an example of a `SentinelArray` since it contains both a length and guarantees null-termination.  Note that the new `cstring` type (an alias to `SentinelPtr!(const(char))` is an exact definition of "c-style" strings.  This allows the semantics of c-strings to exist within the D type system, which comes with alot of advantages (see below).

Adding a `SentinelPtr` and `SentinelArray` type would allow an API to explicitly declare when a sentinel value is required.  This allows the type system to enforce this requirement rather than relying on documentation/trust that the user only passes sentinel pointers when they are required, i.e.
```D
// Params:
//     str = a null-terminated string
void foo1(const(char)* str);

// Params:
//     str = a null-terminated string
void foo2(cstring str);

// NOTE: cstring == SentinelPtr!(const(char))

void bar(string s)
{
    foo1(s.ptr); // Even though this is invalid, the type system doesn't catch it
    foo2(s.ptr); // Error: `immutable(char)*` is not convertible to SentinelPtr!(const(char))
}
```

Expanding on the example a bit further, the correct thing to do there would have been:
```D
// Copies str to a temporary buffer so it can add a sentinel value and return it
const(char)* tempCString(const(char)[] str);

void bar(string s)
{
    foo1(s.tempCString); // OK
    foo2(s.tempCString); // OK
}
````

This pattern is seen all over phobos. The problem here is that if `bar` had been called with a string literal in the first place, then the copy in `tempCString` would have been completely unnecessary, but the `bar` function has no way of knowing whether or not `s` already has a sentinel value.  This is why `SentinelArray` comes in handy because now `bar` can know this, i.e.
```D
// If `T` is a SentinelArray this is a no-op that simply returns `str`. Otherwise,
// it copies str to a temporary buffer so it can add a sentinel value and return it
auto tempCString(T)(T str);

void bar(T)(T s)
{
    foo1(s.tempCString); // OK
    foo2(s.tempCString); // OK
}

bar("hello"); // Assuming the string literal type is now `SentinelString`,
              // this means `tempCString` will be a no op!
```

It's important to note that `SentinelArray` allows the type to indicate that the array not only "owns" all the elements between `array[0..$]`, it also owns the sentinel element `array[$]`.  The array type itself has no way to indicate this.

One reason for adding this to druntime (instead of phobos) is so that it could take advantage of these types in its C bindings.  In cases where functions currently use pointers to accept sentinel arrays, they can be replaced with SentinelPtr to explicitly declare this requirement.

The overall impact should be an increase in safety, better template introspection and provides the opportunity to avoid unnecessary copy/allocations of arrays in certain situations.  Safety would be increased because functions can declare when sentinel values are required allowing compile-time and runtime checks to enforce this.  This also allows templates to take advantage of a common type to know when an array has a sentinel value.  This can result in selecting more optimized algorithms when they apply (see https://www.youtube.com/watch?v=AxnotgLql0k&t=14m41s).  It also allows functions to omit an extra copy/allocation when a sentinel array is required.  The classic example of this is usage of the `tempCString` function which becomes a no-op for `SentinelArray!char`.

Note that with this change the string literal type could be redefined as `SentinelArray!(immutable(char))`, (whose alias is `SentinelString`) which would explicitly guarantee it is null-terminated allowing functions to know this as well.  This prevents the need to copy the string to a temporary buffer when it needs to be passed to a function accepting "c-style" strings.

### DIFF
--- a/src/core/sentinel.d
+++ b/src/core/sentinel.d
@@ -1,0 +1,620 @@
+/**
+Contains types to differentiate arrays with sentinel values.
+*/
+module core.sentinel;
+
+/**
+Selects the default sentinel value for a type `T`.
+
+It has a special case for the char types, and also allows
+the type to define its own default sentinel value if it
+has the member `defaultSentinel`. Otherwise, it uses `T.init`.
+*/
+private template defaultSentinel(T)
+{
+         static if (is(Unqual!T ==  char)) enum defaultSentinel = '\0';
+    else static if (is(Unqual!T == wchar)) enum defaultSentinel = cast(wchar)'\0';
+    else static if (is(Unqual!T == dchar)) enum defaultSentinel = cast(dchar)'\0';
+    else static if (__traits(hasMember, T, "defaultSentinel")) enum defaultSentinel = T.defaultSentinel;
+    else                                   enum defaultSentinel = T.init;
+}
+
+// NOTE: T should be unqalified (not const/immutable etc)
+//       This "unqualification" of T is done by the `SentinelPtr` and `SentinelArray` templates.
+private template SentinelTemplate(T, immutable T sentinelValue)
+{
+    private enum CommonPtrMembers = q{
+        /**
+        Converts the ptr to an array by "walking" it for the sentinel value to determine its length.
+
+        Returns:
+            the ptr as a SentinelArray
+        */
+        inout(ArrayType) toArray() inout
+        {
+            ArrayType array = void;
+            array.array = cast(typeof(array.array))ptr[0 .. walkLength()];
+            return cast(inout(ArrayType))array;
+        }
+
+        /**
+        Return the current value pointed to by `ptr`.
+        */
+        auto front() inout { return *ptr; }
+    };
+    struct MutablePtr
+    {
+        alias ArrayType = MutableArray;
+
+        union
+        {
+            T* ptr;
+            ConstPtr asConst;
+        }
+        alias asConst this; // facilitates implicit conversion to const type
+        // alias ptr this; // NEED MULTIPLE ALIAS THIS!!!
+        mixin(CommonPtrMembers);
+
+        /**
+        Coerce the given `array` to a `SentinelPtr`. It checks and asserts
+        if the given array does not contain the sentinel value at `array.ptr[array.length]`.
+        */
+        this(T[] array) @system
+        in { assert(array.ptr[array.length] == sentinelValue,
+            "array does not end with sentinel value"); } do
+        {
+            this.ptr = array.ptr;
+        }
+    }
+    struct ImmutablePtr
+    {
+        alias ArrayType = ImmutableArray;
+
+        union
+        {
+            immutable(T)* ptr;
+            ConstPtr asConst;
+        }
+        alias asConst this; // facilitates implicit conversion to const type
+        // alias ptr this; // NEED MULTIPLE ALIAS THIS!!!
+        mixin(CommonPtrMembers);
+
+        /**
+        Coerce the given `array` to a `SentinelPtr`. It checks and asserts
+        if the given array does not contain the sentinel value at `array.ptr[array.length]`.
+        */
+        this(immutable(T)[] array) @system
+        in { assert(array.ptr[array.length] == sentinelValue,
+            "array does not end with sentinel value"); } do
+        {
+            this.ptr = array.ptr;
+        }
+    }
+    struct ConstPtr
+    {
+        alias ArrayType = ConstArray;
+
+        const(T)* ptr;
+        alias ptr this;
+        mixin(CommonPtrMembers);
+
+        /**
+        Coerce the given `array` to a `SentinelPtr`. It checks and asserts
+        if the given array does not contain the sentinel value at `array.ptr[array.length]`.
+        */
+        this(const(T)[] array) @system
+        in { assert(array.ptr[array.length] == sentinelValue,
+            "array does not end with sentinel value"); } do
+        {
+            this.ptr = array.ptr;
+        }
+
+        /**
+        Returns true if `ptr` is pointing at the sentinel value.
+        */
+        @property bool empty() const { return *this == sentinelValue; }
+
+        /**
+        Move ptr to the next value.
+        */
+        void popFront() { ptr++; }
+
+        /**
+        Walks the array to determine its length.
+
+        Returns:
+            the length of the array
+        */
+        size_t walkLength() const
+        {
+            for(size_t i = 0; ; i++)
+            {
+                if (ptr[i] == sentinelValue)
+                {
+                    return i;
+                }
+            }
+        }
+    }
+
+    private enum CommonArrayMembers = q{
+        /**
+        A no-op that just returns the array as is.  This is to be useful for templates that can accept
+        normal arrays an sentinel arrays. The function is marked as `@system` not because it is unsafe
+        but because it should only be called in unsafe code, mirroring the interface of the free function
+        version of asSentinelArray.
+
+        Returns:
+            this
+        */
+        pragma(inline) auto asSentinelArray() @system inout { return this; }
+        /// ditto
+        pragma(inline) auto asSentinelArrayUnchecked() @system inout { return this; }
+    };
+    struct MutableArray
+    {
+        union
+        {
+            // TODO: assert and make sure the length/ptr has the same alignment as array
+            union
+            {
+                struct
+                {
+                    size_t length;
+                    MutablePtr ptr;
+                }
+                T[] array;
+            }
+            ConstArray asConst;
+        }
+        alias asConst this; // facilitates implicit conversion to const type
+        // alias array this; // NEED MULTIPLE ALIAS THIS!!!
+        mixin(CommonArrayMembers);
+
+        /**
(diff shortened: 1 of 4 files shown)


jmdavis:
> The reason for adding this to druntime (instead of phobos) is so that it could take advantage of these types in its C bindings. In cases where functions currently use pointers to accept sentinel arrays, they can be replaced with SentinelPtr to explicitly declare this requirement.

Aside from whether these are a good idea in general, that won't work for C bindings. The C bindings must match the C declarations. And we don't do wrappers for the C bindings in druntime.

marler8997:
> Aside from whether these are a good idea in general, that won't work for C bindings. The C bindings must match the C declarations. And we don't do wrappers for the C bindings in druntime.

C doesn't mangle their function names...so the bindings should just work :)  I also verified this using the sentinel library.

wilzbach:
Q: Why can't this be a dub package? Or: what's the motivation for putting this into druntime?

marler8997:
> The reason for adding this to druntime (instead of phobos [or dub]) is so that it could take advantage of these types in its C bindings. In cases where functions currently use pointers to accept sentinel arrays, they can be replaced with SentinelPtr to explicitly declare this requirement.

> ...This also allows templates to take advantage of a common type to know when an array has a sentinel value. This can result in selecting more optimized algorithms when they apply

marler8997:
Also if this was implemented, we would want to change the type of string literals from `string` to `SentinelString`.  If this was the case, then these sentinel types would need to be "built-in" as much as the `string` type currently is (hence why I'm putting it in druntime for now).

Note this is a HUGE change and likely requires a DIP.  This PR is meant as a demonstration of the advantages of adding these sentinel types.

Also note that if the string literal type was changed to `SentinelString` then we would first need to have multiple alias this so that `SentinalArray!(immutable(char))` could be implicitly converted to both `SentinalArray!(const(char))` and `immutable(char)[]`.

src/core/sentinel.d
```diff
@@ -0,0 +1,385 @@
...
+        alias asConst this; // facilitates implicit conversion to const type
+        // alias ptr this; // NEED MULTIPLE ALIAS THIS!!!
+        mixin(CommonPtrMembers);
+
+        /**
+        Coerce the given `array` to a `SentinelPtr`. It checks and asserts
+        if the given array does not contain the sentinel value at `array.ptr[array.length]`.
+        */
+        this(T[] array)
+        in { assert(array.ptr[array.length] == sentinelValue,
```
aG0aep6G: 1) Using a contract/assert means this is only enforced in debug builds. This does not improve the safety of the program in terms of the `@safe` attribute.
2) You're going out of bounds here. You can't assume that `array.ptr[array.length]` is supposed to be a sentinel, even if it happens to have the right value.

Same throughout.

marler8997: > 1. Using a contract/assert means this is only enforced in debug builds. This does not improve the safety of the program in terms of the @safe attribute.

I'm not sure, you could either go with `assert` or `enforce`.  It's synonymous with array bounds checking, you may only want it turned on in `debug` mode.  Maybe this check should be "tied into" the same setting, only enabled when array bounds checking is enabled.

> 2. You're going out of bounds here. You can't assume that array.ptr[array.length] is supposed to be a sentinel, even if it happens to have the right value.

Right that's the point of this type.  You can't go outside of the bounds of an array (or a "slice") unless you know the array also owns the elements you are checking.  The point of `SentinalArray` is that if you have one, you know it also "owns" the sentinel element at `array.ptr[array.length]`.  Now this particular constructor is "coercing" a normal array to a `SentinalArray`, and whenever you convert from one type to another, it is up to the developer to make sure the conversion is valid, there's nothing the compiler can do about that.  However, if `SentinalArray` is builtin, meaning that string literals will already be of type `SentinalArray`, then this coercion will only be used as much as normal casting.

aG0aep6G: > It's synonymous with array bounds checking, you may only want it turned on in debug mode.

Note that `@safe` needs bounds checking to work, which is why `-release` keeps bounds checking on in `@safe` code.

> The point of SentinalArray is that you are explicitly declaring it "owns" the sentinel element at array.ptr[array.length].
[...]
> it is up to the developer to make sure the conversion is valid

So getting anything `@safe` is not the plan here? It's just about adding sanity checks to `@system` code?

How about expecting the sentinel in `arr[$ - 1]` (or maybe anywhere in `arr`)? With that and with a run-time check that's always there, you could maybe make it impossible to construct an invalid `SentinelArray`. Then it could be `@safe`:

```d
extern(C) cstring getenv(cstring name) @safe;
extern(C) int puts(cstring s) @safe;
    /* No idea if the char*s are really the only thing keeping these functions from being @safe. */

void main() @safe
{
    cstring p = getenv("PATH\0".asSentinelPtr);
    puts(p);
}
```

(And if `"PATH"` were already a `SentinelArray` as you're planning, it wouldn't need the ugly explicit terminator.)

marler8997: > So getting anything @safe is not the plan here? It's just about adding sanity checks to @system code?

the grand plan would be to change the string literal type to `SentinalString` so you wouldn't have to "coerce" them.  So it should all work in `@safe` code.  The only code that would not be safe would be converting normal D arrays to `SentinalArray` since there's no way to know if an array owns the value `array[$]`.  Actually I also added the `StringLiteral` template which is `@trusted` so you could also use that in `@safe` code, i.e.
```D
puts(StringLiteral!"hello"); // safe
```

> How about expecting the sentinel in arr[$ - 1] (or maybe anywhere in arr)? With that and with a run-time check that's always there, you could maybe make it impossible to construct an invalid SentinelArray. Then it could be @safe:

Good idea.  This is an additional use case that could be used in `@safe`.  I'll see if I can find a good way to add this in.

I know I already answered this, but your getenv example could be written as:
```D
    cstring p = getenv(StringLiteral!"PATH");
```

marler8997: Ok here it is:
```D
/**
This function takes an array whose last element MUST BE the sentinel value and
returns it as a `SentinelArray`.  This differs from the `asSentinelArray` function
which requires not only that the element at `array[$]` is the sentinel value, but also
that the array "owns" the value as well.  This function only operates within the bounds
of the given array, so no "ownership" outside of the array is required.
*/
@property auto reduceToSentinelArray(T)(T[] array) @trusted
in {
    assert(array.length > 0);
    assert(array[$ - 1] == defaultSentinel!T);
   } do
{
    return asSentinelArrayUnchecked(array[0 .. $-1]);
}

///
@safe unittest
{
    auto s = "abc\0".reduceToSentinelArray;
}
```

sentinelPerfTest.d
```diff
@@ -0,0 +1,109 @@
+/*
```
wilzbach: Shouldn't this file be part of the benchmark suite?

marler8997: I wasn't planning on keeping it in the finalized change.  Where is the benchmark suite?  Maybe it would make sense to add this to it.

wilzbach: In `/benchmark`

wilzbach: (though that one is specifically about comparing different druntime versions, but I think just moving the file into this folder should be good enough as it will be a lot easier to find when it's checked in the source code.)

src/core/sentinel.d
```diff
@@ -0,0 +1,620 @@
...
+    {
+        alias ArrayType = MutableArray;
+
+        union
+        {
+            T* ptr;
+            ConstPtr asConst;
+        }
+        alias asConst this; // facilitates implicit conversion to const type
+        // alias ptr this; // NEED MULTIPLE ALIAS THIS!!!
```
wilzbach: https://github.com/dlang/dmd/pull/3998 though it will require quite an effort to revive this.

src/core/sentinel.d
```diff
@@ -0,0 +1,620 @@
...
+        */
+        void popFront() { ptr++; }
+
+        /**
+        Find the length of the array by scanning it for the sentinel value.
+
+        Returns:
+            the length of the array
+        */
+        size_t findLength() const
```
wilzbach: In phobos this is called walkLength

PR #2124, opened 2018-03-02
New `destroy` implementation for value types and instinsic types

JinShil:
See discussion at #2115 for context.

This is my attempt to implement @andralex's bidding as he attempted to describe in the comments of #2115.  Honestly, I don't understand his rationale and motivation for this implementation; it has a couple of special cases that seem arbitrary to me.  It's also possible that I have not implemented this the way he envisioned.

But, I fear @andralex  is losing his patience trying to describe what he wants and why, so I'm just going with it.  Perhaps by having this implementation in front of us we can converge on understanding more quickly.

cc @schveiguy, @jmdavis

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2620,6 +2620,12 @@ unittest
     static assert(is(typeof(caa.byValue().front) == const int));
 }
 
+void _destructRecurse(T)(ref T obj)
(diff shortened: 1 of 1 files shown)


JinShil:
Closing in favor of #2126

PR #2125, opened 2018-03-02
Musl: Move IPV6 constant that breaks Bionic

joakim-noah:
This constant causes a collision for Bionic, but not for Glibc or UClibc because they have declaration blocks earlier.  Technically, this constant comes from the kernel headers, but it's usually also defined by the C runtime headers, so probably easier to put it there.

### DIFF
--- a/src/core/sys/posix/netinet/in_.d
+++ b/src/core/sys/posix/netinet/in_.d
@@ -397,8 +397,7 @@ else version( linux )
         IPPROTO_TCP  = 6,
         IPPROTO_PUP  = 12,
         IPPROTO_UDP  = 17,
-        IPPROTO_IDP  = 22,
-        IPPROTO_IPV6 = 41
+        IPPROTO_IDP  = 22
     }
 
     enum : c_ulong
@@ -1406,6 +1405,8 @@ else version( CRuntime_Musl )
 
     enum : uint
     {
+        IPPROTO_IPV6 = 41,
+
         IPV6_UNICAST_HOPS   = 16,
         IPV6_MULTICAST_IF   = 17,
         IPV6_MULTICAST_HOPS = 18,


joakim-noah:
That's a lot of repetition for two sets of APIs that should be mostly orthogonal, ie platform vs runtime. I like it the way we have now, even though there wil be some slight overlap like this, especially because it'll be much easier when some new combo like kFreeBSD comes along.  This organization was also why the recent `Musl` and `uClibc` ports were fairly easy to add, unlike when I first added `Bionic` and had to separate out the `linux` and `Glibc` APIs.

ibuclaw:
Well, it should probably be `CRuntime_FreeBSD`, `CRuntime_OSX`, etc...  Then have any platform specific stuff inside the CRuntime blocks.  The majority of it will be common to all though - just the data structures that will differ.

joakim-noah:
It's done a little nicer for linux platforms right now: `linux` refers to kernel headers and APIs, while `CRuntime_Glibc|Bionic|Musl|uClibc` refer to the different libc APIs.  Therefore, if you ever want to port D to kFreeBSD, which runs Glibc with the FreeBSD kernel, the Glibc APIs are already separated out for you, though you're right that you'd still have to split the FreeBSD kernel from its libc APIs.

I don't think those libc APIs vary much by platform, which is when your mooted organization would make the most sense, so separating by kernel and libc APIs seems the best approach to me, though many kernel/libc combos are still clubbed together for now.

ibuclaw:
@joakim-noah - Yep.  Here's one example that was recently highlighted to me.

https://github.com/dlang/druntime/blob/2d87bb9679602a0a108e06f0f16c12e69bb9c8e8/src/core/sys/posix/dirent.d#L52-L92

`struct dirent` differs on kFreeBSD.

PR #2126, opened 2018-03-02
Improved semantics for destroy()

andralex:
This supersedes https://github.com/dlang/druntime/pull/2124 and previous attempts. Apologies for the override but it seemed easier to simply write this than discuss what to do over several sessions. cc @JinShil

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2623,9 +2623,7 @@ unittest
 private void _destructRecurse(S)(ref S s)
     if (is(S == struct))
 {
-    static if (__traits(hasMember, S, "__xdtor") &&
-               // Bugzilla 14746: Check that it's the exact member of S.
-               __traits(isSame, S, __traits(parent, s.__xdtor)))
+    static if (__traits(hasMember, S, "__xdtor"))
         s.__xdtor();
 }
 
@@ -2644,9 +2642,7 @@ private void _destructRecurse(E, size_t n)(ref E[n] arr)
 void _postblitRecurse(S)(ref S s)
     if (is(S == struct))
 {
-    static if (__traits(hasMember, S, "__xpostblit") &&
-               // Bugzilla 14746: Check that it's the exact member of S.
-               __traits(isSame, S, __traits(parent, s.__xpostblit)))
+    static if (__traits(hasMember, S, "__xpostblit"))
         s.__xpostblit();
 }
 
--- a/src/object.d
+++ b/src/object.d
@@ -2918,21 +2918,53 @@ unittest
     assert(postblitRecurseOrder == order);
 }
 
+// destroy
 /********
-Destroys the given object and sets it back to its initial state. It's used to
-_destroy an object, calling its destructor or finalizer so it no longer
-references any other objects. It does $(I not) initiate a GC cycle or free
-any GC memory.
-*/
-void destroy(T)(T obj) if (is(T == class))
-{
-    rt_finalize(cast(void*)obj);
-}
+Destroys an object by means of calling its destructor, if present. For
+types `T` that define a destructor, `destroy` then also fills the
+object with `T.init` to ensure the object's destructor can be called
+again (e.g. when the object goes out of scope). Unlike $(LREF
+__delete), `destroy` does not release memory or other resources by
+itself (though it is possible that destructors it calls do so).
 
-/// ditto
-void destroy(T)(T obj) if (is(T == interface))
+Note: A `struct` type may have a destructor even if it does not
+directly define `~this()`, if at least one field has a destructor
+(transitively). A `class` type may have a destructor even if it does
+not directly define `~this()`, if at least one field has a destructor
+(transitively), or if the destructor is inherited.
+
+Detailed behavior of `destroy(obj)` depends on the type `T` as
+follows:
+
+$(UL $(LI If `T` is a primitive type, a `struct` types that does not
+have a destructor, or a `class` type that does not have a destructor,
+the call to has no effect.)
+
+$(LI For `class` and `interface` types, the call `destroy(obj)` has no
+effect if `obj` is `null`.)
+
+$(LI For `struct` and `class` types that have a destructor (either
+defined explicitly or synthesized by the compiler), the destructor is
+called. Then the object is initialized using
+`T.init`. Reinitialization occurs even if the type disables default
+construction with `@disable this()`.)
+
+$(LI For `interface` types, the dynamic `class` object is fetched from
+the interface object, and then `destroy` is applied to it.)
+
+$(LI For static and dynamic arrays, `destroy` is applied right-to-left
+to all elements. This means multidimensional arrays are destroyed in
+depth, transitively (behavior different from $(LREF __delete), which
+is by necessity shallow). If one or more destructors throw an
+exception, the calls are completed and then the first exception
+encountered is rethrown.))
+*/
+void destroy(T)(T obj) if (is(T == class) || (is(T == interface)))
 {
-    destroy(cast(Object)obj);
+    static if (is(T == interface))
+        destroy(cast(Object) obj);
+    else
+        rt_finalize(cast(void*) obj);
 }
 
 /// Reference type demonstration
@@ -2964,7 +2996,7 @@ unittest
     i = 1;
     assert(i == 1);           // `i` changed to `1`
     destroy(i);
-    assert(i == 0);           // `i` is back to its initial state `0`
+    assert(i == 1);           // `i` stays unchanged
 }
 
 unittest
@@ -3022,18 +3054,35 @@ unittest
    }
 }
 
+// Does calling destroy against an object of type T do anything?
+private template hasElaborateDestroy(T)
+{
+    // Note: __xdtor includes fields, __dtor doesn't.
+    static if (__traits(hasMember, T, "__xdtor"))
+        enum bool hasElaborateDestroy = true;
+    else static if (is(T : U[], U))
+        enum bool hasElaborateDestroy = hasElaborateDestroy!U;
+    else static if (is(T : U[n], U, size_t n))
+        enum bool hasElaborateDestroy = hasElaborateDestroy!U && n > 0;
+    else
+        enum bool hasElaborateDestroy = false;
+}
+
 /// ditto
 void destroy(T)(ref T obj) if (is(T == struct))
 {
-    _destructRecurse(obj);
-    () @trusted {
-        auto buf = (cast(ubyte*) &obj)[0 .. T.sizeof];
-        auto init = cast(ubyte[])typeid(T).initializer();
-        if (init.ptr is null) // null ptr means initialize to 0s
-            buf[] = 0;
-        else
-            buf[] = init[];
-    } ();
+    static if (hasElaborateDestroy!T)
+    {
+        obj.__xdtor();
+        () @trusted {
+            auto buf = (cast(ubyte*) &obj)[0 .. T.sizeof];
+            auto init = cast(ubyte[])typeid(T).initializer();
+            if (init.ptr is null) // null ptr means initialize to 0s
+                buf[] = 0;
+            else
+                buf[] = init[];
+        } ();
+    }
 }
 
 nothrow @safe @nogc unittest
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -2922,21 +2918,53 @@ unittest
...
+Note: A `struct` type may have a destructor even if it does not
+directly define `~this()`, if at least one field has a destructor
+(transitively). A `class` type may have a destructor even if it does
+not directly define `~this()`, if at least one field has a destructor
+(transitively), or if the destructor is inherited.
+
+Detailed behavior of `destroy(obj)` depends on the type `T` as
+follows:
+
+$(UL $(LI If `T` is a primitive type, a `struct` types that does not
```
JinShil: \`struct\` type (singular)

src/object.d
```diff
@@ -2922,21 +2918,53 @@ unittest
...
 }
 
+// destroy
 /********
-Destroys the given object and sets it back to its initial state. It's used to
-_destroy an object, calling its destructor or finalizer so it no longer
-references any other objects. It does $(I not) initiate a GC cycle or free
-any GC memory.
+Destroys an object by means of calling its destructor, if present. For
+types `T` that define a destructor, `destroy` then also fills the
```
JinShil: probably best to use "... that have a  destructor" here since that's the verbage used in the following "NOTE".

JinShil:
This is better than my prior understanding of @andralex's specification as, at least, reference types and value types are treated uniformly.  However, I still don't understand why having a destructor or not having a destructor should result in different behavior.

sbstp:
I'm not the author of Higgs, you should probably cc @maximecb.

maximecb:
Oha @JinShil.

So, it's been a long time since I've looked at this code, but this might be the source of your problem: https://github.com/higgsjs/Higgs/blob/master/source/runtime/gc.d#L914

12345swordy:
no attempts of fixing bugs regarding the class deconstructor when it comes to attributes?

JinShil:
Breaking Higgs
---------------
I removed one destructor from Higgs in an attempt to make it compatible with this PR, but there are still 3 more that need attention:

https://github.com/higgsjs/Higgs/blob/dbe1fb993addfef98e900df6fe2e1e3afe8ecae4/source/runtime/object.d#L512
https://github.com/higgsjs/Higgs/blob/dbe1fb993addfef98e900df6fe2e1e3afe8ecae4/source/runtime/vm.d#L805
https://github.com/higgsjs/Higgs/blob/dbe1fb993addfef98e900df6fe2e1e3afe8ecae4/source/runtime/gc.d#L87

Alternate Proposal
-------------------
I propose decoupling destruction from re-initialization for all cases, and delegating the decision and method of re-initialization to the user.  That puts the burden on the user to understand what they need to do and why, but I believe that's responsibility the user opts into when they choose to implement a destructor.  Consider the following contrived examples.

Method 1:  Use a boolean flag to indicate whether or not the object has already been destroyed
```D
struct S 
{
    void* p;
    bool destroyed = false;
    this(size_t size)
    {
        p = allocateResource(size);  // may not have the same semantics as `malloc`
    }
    ~this() 
    {
        if (!destroyed)
        {
            deallocateResource(p);   // may not have the same semantics as `free`
            destroyed = true;
        }
    }
}
```

Method 2:  Use a special state to indicate whether or not the object has already been destroyed
```D
struct S 
{
    void* p;
    this(size_t size)
    {
        p = allocateResource(size);   // may not have the same semantics as `malloc`
    }
    ~this() 
    {
        if (p == null)
        {
            deallocateResource(p);   // may not have the same semantics as `free`
            p = null;
        }
    }
}
```

Method 3:  Re-destuction is benign due to the semantics of the resource deallocation
```D
struct S 
{
    void* p;
    this(size_t size)
    {
        p = allocateResource(size);  // same semantics as `malloc`
    }
    ~this() 
    {
        deallocateResource(p);     // same semantics as `free`
        p = null;
    }
}
```

Learning from C#
------------------
There may also be some wisdom to be found in C#'s dispose pattern documented here:  https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern.  It is Microsoft's solution to basically the same problem.

JinShil:
> no attempts of fixing bugs regarding the class deconstructor when it comes to attributes?

That's out of scope for now, and probably needs changes to be made in DMD.

JinShil:
> In what way are they broken?

I believe it is their mere existence, causing the object to be re-initialized, that cause problems for Higg's test suite.  But, I need to look into it in more detail to be sure.

JinShil:
@andralex How confident are you in this implementation?  For some reason, noone from team DRuntime  seems very interested in it (or maybe you're just too scary).  Are you willing to play the dictator with this, or are you looking for feedback?

andralex:
@JinShil thanks for asking! :) I think the intent is what we need; of course it can be improved, and also the implementation may have issues. So feedback is welcome.

Generally `destroy` needs to rely on what we already have in the language instead of requiring new paraphernalia or asking the user to rewrite their destructors in a certain way. 

The mechanisms at our disposal are the following:

* Destructors get rid of an object's extra resources.
* Destructors leave the object in an invalid state; adding a requirement here is not tenable.
* Every object (even those with `@disable this();`) has a well-defined `init` state
* At least for objects without  `@disable this();`, an object in its initial state must be destroyable, otherwise trivial code such as `{ T t; }` would be incorrect.
* We can use introspection to figure out whether an object has a destructor or not.

The fundamental approach of this PR is simple - calling `destroy` should call the destructor, if any, and leave the object in a valid destroyable state. So if the object has no destructor there's no need to do anything at all - leave the object as is, it's as destroyable as it was before calling `destroy`. If the object does have a destructor we should call it and then repaint the object with its initial state - the one that we can reasonably assume it's safely destroyable. If we remove from this behavior we allow for unsafe code (people call `destroy` and then the destructor gets automatically invoked). If we add to this behavior we do work for naught (as was the case before - why should destroy on an int set it to zero? Every int has as much resources (none) and as many rights as others). If we add requirements to the destructor we break all existing destructors. If we add compiler support we incur overhead. There's really not a lot of wiggle room here.

I'm unclear on how Higgs got broken, and probably we need to get to the bottom of that. Where can be failures be seen and how do they manifest themselves?

wilzbach:
> I'm unclear on how Higgs got broken, and probably we need to get to the bottom of that. Where can be failures be seen and how do they manifest themselves?

See Mike's comment above. They can be seen on the Project Tester:

https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdruntime/detail/PR-2126/2/pipeline

For local reproduction, do sth. like:

```sh
git clone https://github.com/higgsjs/Higgs
cd Higgs
dub --compiler=/home/seb/dlang/dmd/generated/linux/release/64/dmd
```

JinShil:
> I'm unclear on how Higgs got broken, and probably we need to get to the bottom of that. Where can be failures be seen and how do they manifest themselves?

The symptom appears in the Higgs test suite where the unitttests are specifically testing if an object is in a certain state:

> Assertion failed (runtime/gc.d@472): gcForward: object not in from-space heap

https://github.com/higgsjs/Higgs/blob/dbe1fb993addfef98e900df6fe2e1e3afe8ecae4/source/runtime/gc.d#L472

I believe this is caused simply by the existence of some of destructors in the Higgs code base.  [I already removed 1 empty constructor](https://github.com/higgsjs/Higgs/pull/214) which solved one error, but another appeared.  There are current 2 more empty constructors in Higgs:

https://github.com/higgsjs/Higgs/blob/dbe1fb993addfef98e900df6fe2e1e3afe8ecae4/source/runtime/object.d#L512
https://github.com/higgsjs/Higgs/blob/dbe1fb993addfef98e900df6fe2e1e3afe8ecae4/source/runtime/vm.d#L805

I tried to remove those as well, but it didn't solve the problem we're currently experiencing.  

The current error appears to be due to the following destructor, which is not empty:

https://github.com/higgsjs/Higgs/blob/dbe1fb993addfef98e900df6fe2e1e3afe8ecae4/source/runtime/gc.d#L87

I tried to fix that also to make it compatible with this PR, but ultimately failed.  I don't understand the code or the test suite logic well enough to make the change.

If you are convinced that this PR is the way to go, and are willing to exercise your authority, given the lack of reviews, I propose the following procedure:

1.  Remove Higgs from the Jenkins test suite
2.  File an issue in the Higgs repository and leave it to the authors there to fix it
3.  Merge this PR
4.  If and when Higgs is fixed to make it compatible with this PR, we can re-add it to the Jenkins test suite.

Note that I am not a member of team DRuntime, so if you don't want to merge your own PR, you'll still have to convince someone to review this PR.  I'm not crazy about these semantics, but I also don't have any other solution.  I'd be hesitant putting my name next to the merge notice even if I were on team DRuntime, but I'd do it if enough time passed without anyone else stepping up with an alternate solution.

Pinging @klickverbot  @schveiguy @MartinNowak @rainers @ZombineDev @DmitryOlshansky @maximecb for help.

12345swordy:
!ping

ntrel:
Optimization: If a class/struct aggregate does not *define* a destructor itself, but one of its fields defines a destructor, we need to do `field = field.init` so the field destructor can be called again later. The aggregate's *generated* destructor will not access any fields that don't have a destructor, so we do not need to reinit those. This pull seems to always reinit *all* fields of the aggregate (which is fine, but overkill).

Also, I like the new semantics of `destroy`, but I expect people have relied on it always doing the reinit. I read the discussion in #2115, and even if the doc description was vague [until recently?], the unittests were not - they clearly show reinit, implying that users can rely on it. I suggest this pull either uses a different function name or somehow does a sophisticated deprecation dance to make users aware of the change.

andralex:
@ntrel we won't cater to people relying on undocumented behavior. This not only creates difficulties for this particular PR, but creates a dangerous precedent whereby we dig ourselves in bureaucracy.

src/object.d
```diff
@@ -2922,21 +2918,53 @@ unittest
...
+(transitively). A `class` type may have a destructor even if it does
+not directly define `~this()`, if at least one field has a destructor
+(transitively), or if the destructor is inherited.
+
+Detailed behavior of `destroy(obj)` depends on the type `T` as
+follows:
+
+$(UL $(LI If `T` is a primitive type, a `struct` types that does not
+have a destructor, or a `class` type that does not have a destructor,
+the call to has no effect.)
```
schveiguy: "call to has no effect" Maybe missing "destroy" in there?

Note, this is technically incorrect for a class type with no destructor, as destroying a class instance that has no destructor is still going to call `rt_finalize`, which at least sets the vtable to null.

src/object.d
```diff
@@ -2922,21 +2918,53 @@ unittest
...
+
+$(LI For static and dynamic arrays, `destroy` is applied right-to-left
+to all elements. This means multidimensional arrays are destroyed in
+depth, transitively (behavior different from $(LREF __delete), which
+is by necessity shallow). If one or more destructors throw an
+exception, the calls are completed and then the first exception
+encountered is rethrown.))
 */
-void destroy(T)(T obj) if (is(T == class))
+void destroy(T)(T obj) if (is(T == class) || (is(T == interface)))
```
schveiguy: I think there is a pre-existing bug in here, as I think the static test will pass for `extern(C++)` classes, which is definitely NOT correct to call `rt_finalize` on. Something to think about as we move towards betterC.

src/object.d
```diff
@@ -3145,9 +3231,8 @@ unittest
 
 /// ditto
 void destroy(T)(ref T obj)
-    if (!is(T == struct) && !is(T == interface) && !is(T == class) && !_isStaticArray!T)
+if (!is(T == struct) && !is(T == interface) && !is(T == class) && !_isStaticArray!T)
```
schveiguy: This is going to accept an lvalue dynamic array, which is potentially going to override the array code above. I'm not 100% sure, because it's a specialization.

This may fail:

```D
static struct S
{
    int x;
    ~this() {}
}

S[] arr = new S[5];
arr[] = S(42);
destroy(arr);
assert(arr[0].x == 0); // will this fail?
```

src/object.d
jacob-carlborg: And Objective-C classes. There’s a recent change where `__traits(getLinkage)` supports classes.

ntrel:
> people relying on undocumented behavior

@andralex I suggest we revert #2054 ASAP unless this is merged now. #2054 was merged on 27 Jan (it also says you approved those changes).

Observation: The `hasElaborateDestroy` changes and the `destroy(T[])` changes are orthogonal and could have been separated.

PR #2129, opened 2018-03-04
fix issue 18547 - Win32: throwing exception in fiber crashes application

rainers:
increase the default stack size because exception handling might need up to 16k. The actually used stack can depend on the version of DbgHelp.dll, the existence of debug information and possibly other conditions.

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -4071,6 +4071,14 @@ class Fiber
     // Initialization
     ///////////////////////////////////////////////////////////////////////////
 
+    version(Windows)
+        // exception handling walks the stack, invoking DbgHelp.dll which
+        // needs up to 16k of stack space depending on the version of DbgHelp.dll,
+        // the existence of debug symbols and other conditions. Avoid causing
+        // stack overflows by defaulting to a larger stack size
+        enum defaultStackPages = 8;
+    else
+        enum defaultStackPages = 4;
 
     /**
      * Initializes a fiber object which is associated with a static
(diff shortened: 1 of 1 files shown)


src/core/thread.d
```diff
@@ -4087,7 +4086,7 @@ class Fiber
      * In:
      *  fn must not be null.
      */
-    this( void function() fn, size_t sz = PAGESIZE*4,
+    this( void function() fn, size_t sz = size_t.max,
```
Burgos: Could we remove the stack size as the default value for this parameter, and add set of the constructors `this (void function()/*delegate*/ fn)` which will default to  8 pages or 4 pages, depending on platform. If you keep the existing approach, perhaps the documentation of these constructors should indicate default stack size?

rainers: Could be possible, but would duplicate the number of necessary ctors and the documentation.
I was also wondering whether the default 4/8*PAGESIZE makes much sense if the system uses large pages. (I started using `enum defaultStackSize` just to notice that PAGESIZE is runtime-initialized).

rainers: On second thought, maybe it is better to change the default-argument to defaultStackPages*PAGESIZE, as it is more self-documenting and does not need a magic value.

rainers: > it is better to change the default-argument to defaultStackPages*PAGESIZE

Done.

ZombineDev: Agreed, tnx

PR #2133, opened 2018-03-06
Added aio.h implementations for FreeBSD, NetBSD and DragonFly

dkgroot:
Consists of two seperate commits:
1. Fix the previous implementation (#2074).
Note: ```ssize_t aio_return(const(aiocb)* aiocbp);``` should have been ```ssize_t aio_return(aiocb* aiocbp);``` (According to Linux, BSD's and [OpenGroup-Posix](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/aio.h.html) )
2. Add implementations for FreeBSD, NetBSD and DragonFly

Port Requested by @ibuclaw : https://github.com/dlang/druntime/pull/2074#issuecomment-370154627
Ping @Darredevil

### DIFF
--- a/src/core/sys/posix/aio.d
+++ b/src/core/sys/posix/aio.d
@@ -26,7 +26,7 @@ version (CRuntime_Glibc)
             int aio_fildes;
             int aio_lio_opcode;
             int aio_reqprio;
-            void *aio_buf;   //volatile
+            void* aio_buf;   //volatile
             size_t aio_nbytes;
             sigevent aio_sigevent;
 
@@ -34,9 +34,68 @@ version (CRuntime_Glibc)
             off_t aio_offset;
             ubyte[32] __glibc_reserved;
         }
+    } else
+        static assert(false, "Unsupported CPU Type");
+}
+else version (FreeBSD)
+{
+    struct __aiocb_private
+    {
+        long status;
+        long error;
+        void* kernelinfo;
     }
-    else
-        static assert(0);
+
+    struct aiocb
+    {
+        int aio_fildes;
+        off_t aio_offset;
+        void* aio_buf;   // volatile
+        size_t aio_nbytes;
+        private int[2] __spare;
+        private void* _spare2__;
+        int aio_lio_opcode;
+        int aio_reqprio;
+        private __aiocb_private _aiocb_private;
+        sigevent aio_sigevent;
+    }
+
+    version = bsd_posix;
+}
+else version (NetBSD)
+{
+    struct aiocb
+    {
+        off_t aio_offset;
+        void* aio_buf;   // volatile
+        size_t aio_nbytes;
+        int aio_fildes;
+        int aio_lio_opcode;
+        int aio_reqprio;
+        sigevent aio_sigevent;
+        private int _state;
+        private int _errno;
+        private ssize_t _retval;
+    }
+
+    version = bsd_posix;
+}
+else version (DragonFlyBSD)
+{
+    struct aiocb
+    {
+        int aio_fildes;
+        off_t aio_offset;
+        void* aio_buf;   // volatile
+        size_t aio_nbytes;
+        sigevent aio_sigevent;
+        int aio_lio_opcode;
+        int aio_reqprio;
+        private int _aio_val;
+        private int _aio_err;
+    }
+
+    version = bsd_posix;
 }
 else
     static assert(false, "Unsupported platform");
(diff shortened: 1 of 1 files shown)


dkgroot:
@ibuclaw As per your request: #2074 (comment)

src/core/sys/posix/aio.d
```diff
@@ -34,6 +34,66 @@ version (CRuntime_Glibc)
...
             ubyte[32] __glibc_reserved;
         }
+
+    }
+    else version (FreeBSD)
+    {
+        struct __aiocb_private {
+            long status;
+            long error;
+            void *kernelinfo;
```
ibuclaw: Should probably use D style type here, and in other places (`void* foo`, not `void *foo`).

src/core/sys/posix/aio.d
```diff
@@ -34,6 +34,66 @@ version (CRuntime_Glibc)
...
+            int aio_fildes;
+            int aio_lio_opcode;
+            int aio_reqprio;
+            sigevent aio_sigevent;
+            private int _state;
+            private int _errno;
+            private ssize_t _retval;
+        }
+
+        version = bsd_posix;
```
ibuclaw: Not sure about this, does it really make things easier to have this instead of specific versions?

Don't mind either way.

ibuclaw: Given how small this module is, I mean.

src/core/sys/posix/aio.d
dkgroot: Q:Would it make sense to to extend checkwhitespace.d to also check for this ?

src/core/sys/posix/aio.d
dkgroot: I think it focusses the attention on the difference between the enums ordering.

Would be happy to reverse it though, let's see if somebody would like it changed as well.

dkgroot:
BTW: Would it be helpfull to copy the comments/documentation over from the linux/bsd header files. I noticed that some of the other sys/posix/*.d files did have this ?

ZombineDev:
@dkgroot AFAIK we avoid copying comments since they are covered by copyright, unlike pure API bindings.

src/core/sys/posix/aio.d
ZombineDev: We should have a much more comprehensive lint used for phobos, so we should probably replace checkwhitespace with that and spend effort on something that may be replaced in the future.

src/core/sys/posix/aio.d
```diff
@@ -26,14 +26,74 @@ version (CRuntime_Glibc)
...
 
             ubyte[24] internal_members_padding;
             off_t aio_offset;
             ubyte[32] __glibc_reserved;
         }
+
+    }
+    else version (FreeBSD)
+    {
+        struct __aiocb_private {
```
ZombineDev: Put brace on next line.

src/core/sys/posix/aio.d
```diff
@@ -50,25 +110,59 @@ enum
...
-    LIO_READ,
-    LIO_WRITE,
-    LIO_NOP
-};
+    enum
+    {
+        LIO_READ,
+        LIO_WRITE,
+        LIO_NOP
+    };
```
ZombineDev: Remove semicolon.

src/core/sys/posix/aio.d
```diff
@@ -50,25 +110,59 @@ enum
...
 int aio_suspend(const(aiocb*)* aiocb_list, int nitems, const(timespec)* timeout);
-int aio_cancel(int fd, aiocb *aiocbp);
-int lio_listio(int mode, const(aiocb*)* aiocb_list, int nitems, sigevent *sevp);
+int aio_cancel(int fd, aiocb* aiocbp);
+int lio_listio(int mode, const(aiocb*)* aiocb_list, int nitems, sigevent* sevp);
+
+/* functions outside/extending posix requirement */
+version (FreeBSD)
+{
+    int aio_waitcomplete(aiocb ** aiocb_list, const(timespec)* timeout);
```
ZombineDev: `aiocb**`

dkgroot:
Ping @MartinNowak @CyberShadow

dkgroot:
Ping @MartinNowak @CyberShadow

CyberShadow:
I don't use these operating systems so I can't really vouch for this, but the diff looks good.

Would you mind squashing the changes from the last commit into the first two?

dkgroot:
@CyberShadow Thanks for checking. 
Don't mind squashing, but wanted to preserve the two first commits to show the distinction (The first one is fixing the previously merged implementation, the second is extending it by adding the BSD implementation). The squashing can be done when someone has been able to review it.

Do you know anybody who does use one of the BSD's for checking/testing ? If so can you add them as a reviewer ?

dkgroot:
Does one of you know why auto-tester fails on linux_64_32 for this change, i don't see how 'std/regex/internal/tests.d' could be related to these changes. Maybe it's just an unrelated regression.

wilzbach:
Spurious failure because std.regex's CTFE tests consumes so much memory (see e.g. https://github.com/dlang/phobos/pull/6164)
auto-tester automatically restarts itself.

dkgroot:
@wilzbach Thanks for the heads-up !
Ping @MartinNowak 
Anything holding this PR up ?

src/core/sys/posix/aio.d
ibuclaw: Can we proceed with removing `version = bsd_posix`?  Just to speed up pulling this in, as this is the only part I'm hesitant about here.

wilzbach:
version (X86_64)
{
...	
}
else version (DragonFlyBSD)
{

}
```

I thought DragonFlyBSD is x86_64 only?

dkgroot:
@wilzbach Not sure how this came to be (It has been a couple of weeks). It looks like the BSD versions are at the wrong scope.

PR #2134, opened 2018-03-08
[uClibc] extend `core.sys.linux.netinet.in_` definitions

rracariu:
`Glibc` definitions cover `uClibc` also.

Probably `Musl` needs to be included also?

This fixes build and run of `vibe.d`. See https://github.com/lindt/docker-dmd/issues/1 and https://forum.dlang.org/post/oyrmgpvyncuanzokkxba@forum.dlang.org

### DIFF
--- a/src/core/sys/linux/netinet/in_.d
+++ b/src/core/sys/linux/netinet/in_.d
@@ -13,7 +13,15 @@ import core.sys.posix.config;
 
 public import core.sys.posix.netinet.in_;
 
-version(CRuntime_Glibc)
+version(CRuntime_Glibc)         version = linux_libc;
+else version(CRuntime_Musl)     version = linux_libc;
+else version(CRuntime_Bionic)   version = linux_libc;
+else version(CRuntime_UClibc)   version = linux_libc;
+
+version(CRuntime_Glibc)         version = gnu_libc;
+else version(CRuntime_UClibc)   version = gnu_libc;
+
+version(linux_libc)
 {
     extern(C) nothrow @nogc:
 
@@ -107,7 +115,7 @@ version(CRuntime_Glibc)
     enum IN6ADDR_ANY_INIT      = in6_addr.init;
     enum IN6ADDR_LOOPBACK_INIT = in6_addr([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1]);
 
-    static if (__USE_MISC)
+    version(gnu_libc) static if (__USE_MISC)
     {
         struct ip_mreq
         {
@@ -166,13 +174,13 @@ version(CRuntime_Glibc)
 
     extern(D) bool IN6_ARE_ADDR_EQUAL(in6_addr* a, in6_addr* b) pure @safe { return *a == *b; }
 
-    static if (__USE_MISC)
+    version(gnu_libc) static if (__USE_MISC)
     {
         int bindresvport(int __sockfd, sockaddr_in* __sock_in);
         int bindresvport6(int __sockfd, sockaddr_in6* _);
     }
 
-    static if (__USE_GNU)
+    version(gnu_libc) static if (__USE_GNU)
     {
         struct in6_pktinfo
         {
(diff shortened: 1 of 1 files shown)


rracariu:
@joakim-noah probably Musl is just as easy?

joakim-noah:
I'll look into it.

joakim-noah:
It appears vibe-d/eventcore#22 broke building vibe.d with these non-Glibc runtimes, as it pulled in these constants, since I was able to build it using dub _on_ my Android device fine before, except for one unrelated issue vibe-d/vibe.d#1885.  Now I hit some of these compilation errors on Android too, as these constants were pulled in after that.

I will look into a better way to get this working for all the non-Glibc libc runtimes and get back to you later today.

rracariu:
It looks that most of these definitions (all?) are actually generated/derived from the Linux headers. Bionic explicitly says so in the headers (https://github.com/rapid7/meterpreter/blob/master/source/bionic/libc/kernel/common/linux/in.h)

I wonder if the real deal is to move all of them under `version (linux)`, I might missed some but my compare at least showed identical values on enums at least for Glibc/Musl/uClibc and Bionic.

joakim-noah:
Yeah, that's what I was thinking, just been putting off checking everything.

joakim-noah:
Looks like most of these declarations are from the linux kernel headers, with the exception of a few like `getsourcefilter`, but all the Musl, Glibc, and Bionic libc's cut-and-paste or `#include` the kernel APIs in their own `netinet/in.h` also.  Since most C programmers will include the libc headers normally, I suggest you simply add Bionic and Musl to your list of libc's where these are defined, though something will also have to be done about `__USE_GNU` and `__USE_MISC`, which are undefined in non-Glibc right now.

That simple approach means some libc mixing may happen, but barring someone sitting down and writing a DStep/libdparse-based tool to automate separating the declarations out (as discussed in #2070), I don't think it's worth manually checking all these.

rracariu:
Included all libcs supported for `Linux` atm in a global version block. I separated `Glibc` and `uClibc` in a separate version to scope those static ifs that make sense only for them.

src/core/sys/linux/netinet/in_.d
```diff
@@ -13,7 +13,15 @@ import core.sys.posix.config;
 
 public import core.sys.posix.netinet.in_;
 
-version(CRuntime_Glibc)
+version(CRuntime_Glibc)     version = linux_libc;
+version(CRuntime_Musl)      version = linux_libc;
```
joakim-noah: Make these all `else version`, as we don't support using multiple libc's at once.

src/core/sys/linux/netinet/in_.d
```diff
@@ -107,7 +115,7 @@ version(CRuntime_Glibc)
     enum IN6ADDR_ANY_INIT      = in6_addr.init;
     enum IN6ADDR_LOOPBACK_INIT = in6_addr([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1]);
 
-    static if (__USE_MISC)
+    version(gnu_libc) static if (__USE_MISC)
```
joakim-noah: Not sure why you chose this version name, plus don't think this will work for uClibc, as `__USE_MISC` is undefined there.  You may have seen it in `core.sys.linux.config` but this module invokes `core.sys.posix.config`, which doesn't have it.

rracariu: uClibc generally follows Glibc, figured that they both got the GNU roots so that's the best name I could come up with...

'__USE_MISC' is defined in `core.sys.posix.config` for both Glibc and uClibc.

src/core/sys/linux/netinet/in_.d
rracariu: I noticed this pattern in other parts of druntime. I can change it however here.

src/core/sys/linux/netinet/in_.d
joakim-noah: OK, name is fine, but [I don't see that enum defined for uClibc](https://github.com/dlang/druntime/blob/master/src/core/sys/posix/config.d#L80).

src/core/sys/linux/netinet/in_.d
joakim-noah: [I've been cleaning those up where I can](https://github.com/dlang/phobos/pull/6193#discussion_r168917047).

src/core/sys/linux/netinet/in_.d
rracariu: OK, they were on my setup... fixed in https://github.com/dlang/druntime/pull/2143

PR #2136, opened 2018-03-10
Add website link to author name for Jonathan M Davis.

jmdavis:


jmdavis:
> Wouldn't this be what a contributors file/table is used for?

I don't know anything about such a file. I assume that you're talking about whatever it is Seb did to put together a list of the contributors so that it could be listed with each release? Regardless, other authors (Walter and Andrei in particular) have their names in the Author sections link to their websites. I'm just adding my website to my name now that I have it set up.

wilzbach:
> Seb did to put together a list of the contributors so that it could be listed with each release

There are actually two variants of this:

https://dlang.org/changelog/2.079.0.html#contributors
https://dlang.org/foundation/contributors.html (overall)

I also used to run a [small service](https://github.com/wilzbach/phobos-contribs) that displays the GitHub contributors at bottom of every module, but the server is broken at the moment and I haven't found the time to migrate it to another box.

There are also the CODEOWNERS files but that's just about getting notified automatically whenever a PR touches a file of your interest area.

jmdavis:
> Yeah imho ideally we should auto-generate this as there are way too many people who don't add their name to the authors list.

Well, for better or worse, historically, the Authors list has only included folks who have contributed significantly to a module and not folks who just fixed a bug here and there, though there are almost certainly folks who have contributed significantly to some modules and never added their names to the Authors list (e.g. std.algorithm currently only lists Andrei, whereas there are plenty of other folks who could be in the list even if we were restricting that list to those with major contributions). I think that for the most part, the ones with my name either were written by me or had large portions rewritten by me at some point. Regardless, if someone is looking for the full list of folks who have contributed at all to a module, the Authors list isn't it.

ibuclaw:
Yeah, just a suggestion that rather than having to put your link everywhere, there's a macro which takes care of that for you, i.e `$(AUTHORREF Jonathan M Davis)`. Then we just have a map file of all authors in the format:
```
Name Link
```
Whether we'd be able to do that sort of chicanery is another matter. :-)

wilzbach:
> here's a macro which takes care of that for you

Well, DDoc just supports expansion. No conditionals or loops.
However, since a month we use a preprocessor for all ddoc files, so in theory we could some crude replacements to https://github.com/dlang/dlang.org/blob/master/ddoc/source/preprocessor.d, but I'm not sure whether it's worth the effort.

jmdavis:
In the case of Authors links, I don't think that anything like that is worth the effort. What I've done here is simple ddoc which won't have to be maintained much, and most authors listed on druntime and Phobos' modules either don't have websites or haven't listed them, so unless that changes, it wouldn't save much effort. It might make sense to have the information listed somewhere so that the auto-generated contributors list for releases and whatnot could have that information, but not much changes with the Authors list very often.

PR #2139, opened 2018-03-12
TypeInfo.getHash(): use const scope

WalterBright:
Start on making TypeInfo methods safe. Should be no problem since users won't be deriving from it.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -252,7 +252,7 @@ class TypeInfo
      * Bugs:
      *    fix https://issues.dlang.org/show_bug.cgi?id=12516 e.g. by changing this to a truly safe interface.
      */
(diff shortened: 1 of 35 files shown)


src/object.d
```diff
@@ -1463,7 +1463,7 @@ const:
         assert(flag >= MItlsctor && flag <= MIname);
         assert(!(flag & (flag - 1)) && !(flag & ~(flag - 1) << 1));
     }
-    do
+    body
```
wilzbach: Why? Use of `body` has been deprecated in favor of `do` with https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1003.md

WalterBright:
I don't think we can fix `in`. It may break an astonishing amount of code, since it was never enforced.

PR #2141, opened 2018-03-13
add SafeObject to object.d

WalterBright:
Since `Object` has resisted all attempts to add attributes, I created a derived class `SafeObject` that adds in *all* the attributes. By deriving from `SafeObject` instead of `Object`, one can create objects that are pure, nothrow, const, etc.

Note that derived classes will not be able to remove the pure, nothrow, @nogc, etc. attributes. For classes that cannot abide by this, they can still derive from `Object`.

Having `SafeObject` enables Phobos code to have separate overloads for `SafeObject`, so that they can operate safely.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -137,6 +137,34 @@ class Object
     }
 }
 
+/**
+ * @safe version of Object
+ */
+class SafeObject
+{
+  @safe pure nothrow @nogc:
+    override string toString() const return scope
+    {
+        return typeid(this).name;
+    }
+
+    override size_t toHash() const scope nothrow @trusted
+    {
+        // BUG: this prevents a compacting GC from working, needs to be fixed
+        return cast(size_t)cast(void*)this;
+    }
+
+    override int opCmp(scope const Object o) const scope
+    {
+        assert(0);
+    }
+
+    override bool opEquals(scope const Object o) const scope
+    {
+        return this is o;
+    }
+}
+
 auto opEquals(Object lhs, Object rhs)
 {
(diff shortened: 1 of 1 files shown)


jacob-carlborg:
I recommend putting the PR description in the commit message. Next time, if you have only one commit, GitHub will use the commit message as the PR title and description. Just separate the title and description with an empty newline.

wilzbach:
I thought @andralex wanted to go the other way and define an `RootObject` without opEquals and friends, s.t. it's easier to overwrite it.

See: https://github.com/dlang/druntime/pull/1439#issuecomment-369604503 (the PR itself was a solution to this problem too)

WalterBright:
It's nice to have multiple ideas put forth to solving it. I have been thinking about this for a while, and decided to just put it out there to push some progress.

12345swordy:
Couple of questions regarding this: 
What if I want only some attributes and not all of them? Can I create a safe only class that allows gc to run?
Does calling the SafeObject finalize member function have the attributes  pure, nothrow, nogc, etc? If it doesn't have a finalize member function what is the equivalent for this?
Wouldn't implementing system default metaclasses(I.E. SafeClass, NoGCClass, PureClass) better suited for this?

andralex:
@WalterBright and I discussed and agreed we'll pursue the `ProtoObject` angle instead of this. This is a good catalyst nevertheless to get the DIP started!

WalterBright:
Actually what we're going to do is write a DIP and see if we can't think of all the angles.

PR #2142, opened 2018-03-14
object: fix escaping p

WalterBright:
This is to correct the error:
```
src/object.d(4450): Error: returning `S(& p)` escapes a reference to local variable `p`
```
which is blocking https://github.com/dlang/dmd/pull/8030

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -4447,6 +4447,6 @@ unittest
     }
 
     int p;
-    scope arr = [S(&p)];
+    scope S[1] arr = [S(&p)];
     auto a = arr.dup; // dup does escape
 }


src/object.d
```diff
@@ -4447,6 +4447,6 @@ unittest
     }
 
     int p;
-    scope arr = [S(&p)];
+    scope S[1] arr = [S(&p)];
```
wilzbach: Isn't `scope` a bit useless if the data is on the stack?

WalterBright: `scope` only applies to data on the stack.

andralex: Would `S[1] arr = [S(&p)];` suffice?

WalterBright: No. `S` needs to be stack allocated, not gc allocated. Hence the change.

PR #2145, opened 2018-03-21
[WIP] Fix Issue 17592 - destroy isn't marked @nogc when the class destructor is marked as @nogc

JinShil:


src/rt/lifetime.d
```diff
@@ -133,7 +133,7 @@ extern (C) void _d_delinterface(void** p)
 
 
 // used for deletion
-private extern (D) alias void function (Object) fp_t;
+private extern (D) alias void function (Object) @nogc fp_t;
```
JinShil: This is the type of the class's destructor.  `ClassInfo.destructor` is cast to this type in `rt_finalize2`.  But, what if the class's destructor is *not* attributed with `@nogc`?

JinShil:
Yeah, this isn't going to work.  See https://issues.dlang.org/show_bug.cgi?id=15246

Closing.

PR #2147, opened 2018-03-22
Move DMD's object.d members to druntime

wilzbach:
Needed for https://github.com/dlang/dmd/pull/8067

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -4449,3 +4449,30 @@ unittest
     scope S[1] arr = [S(&p)];
     auto a = arr.dup; // dup does escape
 }
+
+// compiler frontend lowers dynamic array comparison to this
+bool __ArrayEq(T1, T2)(T1[] a, T2[] b)
+{
+    if (a.length != b.length)
+        return false;
+    foreach (size_t i; 0 .. a.length)
+    {
+        if (a[i] != b[i])
+            return false;
+    }
+    return true;
+}
+
+// compiler frontend lowers struct array postblitting to this
+void __ArrayPostblit(T)(T[] a)
+{
+    foreach (ref T e; a)
+        e.__xpostblit();
+}
+
+// compiler frontend lowers dynamic array deconstruction to this
+void __ArrayDtor(T)(T[] a)
+{
+    foreach_reverse (ref T e; a)
+        e.__xdtor();
+}


wilzbach:
So anyone objecting on this?

rainers:
IMO having these definitions in the compiler is strange. Moving them to druntime slightly reduces what you can do with an empty object.d, though. So I'm on the fence whether this or rewriting on the AST level is better.

JinShil:
Ping!

ibuclaw:
> IMO having these definitions in the compiler is strange. Moving them to druntime slightly reduces what you can do with an empty object.d, though. So I'm on the fence whether this or rewriting on the AST level is better.

Same here, but no objections either way.

JinShil:
> So I'm on the fence...

I totally understand this as there are consequences to these decisions, but please be aware that being on the fence is the worst place for PR to be.  We need to find a way to move this PR one way or the other.

> Moving them to druntime slightly reduces what you can do with an empty object.d

I did some further testing with this.  It appears these implementations are only used by dynamic arrays.  If a user is employing dynamic arrays, there is no way they can use an empty object.d, as there is much more runtime implementation code required to use dynamic arrays.  So I don't think it's an issue.

src/object.d
```diff
@@ -4449,3 +4449,30 @@ unittest
     scope S[1] arr = [S(&p)];
     auto a = arr.dup; // dup does escape
 }
+
+// internal function needed by the compiler
```
klickverbot: This is a fairly meaningless comment; by definition `__*` symbols are internal and used/emitted by the compiler. "Needed by the compiler" is also slightly ambiguous – the templates are needed by the user programs to be compiled, as opposed to the compiler implementation.

Perhaps just substitute a one-liner saying what the frontend lowers to this?

wilzbach: Good idea.

JinShil:
@wilzbach Would you mind adding proper DDoc headers with `params`, `returns`, etc..  I know you probably didn't sign up for that initially, but it would be nice to have.  I think I'd be willing to merge this then.

wilzbach:
I wouldn't mind, but I just started a one week long hiking trip and won't have much internet access nor a laptop, so I will get to this at the end of next week. Sorry.

PR #2148, opened 2018-03-22
Bionic bindings for AArch64 and a Glibc/AArch64 fix

joakim-noah:
This is WIP, as right now it assumes llvm's brand of emulated TLS, which cannot be passed to the GC, so that part will have to change. Just putting this up early in case @ZombineDev or someone else wants to help with the Android/AArch64 port.

### DIFF
--- a/src/core/stdc/stdlib.d
+++ b/src/core/stdc/stdlib.d
@@ -151,21 +151,10 @@ else
 // No unsafe pointer manipulation.
 @trusted
 {
-    version(CRuntime_Bionic)
-    {
-       import core.sys.posix.stdlib: lrand48, srand48;
-       ///
-       alias core.sys.posix.stdlib.lrand48 rand;
-       ///
-       alias core.sys.posix.stdlib.srand48 srand;
-    }
-    else
-    {
-        ///
-       int     rand();
-       ///
-       void    srand(uint seed);
-    }
+    /// These two were added to Bionic in Lollipop.
+    int     rand();
+    ///
+    void    srand(uint seed);
 }
 
 // We don't mark these @trusted. Given that they return a void*, one has
--- a/src/core/sys/posix/fcntl.d
+++ b/src/core/sys/posix/fcntl.d
@@ -98,6 +98,12 @@ version( CRuntime_Glibc )
     enum F_SETLK        = 6;
     enum F_SETLKW       = 7;
   }
+  else version(AArch64)
+  {
+    enum F_GETLK        = 5;
+    enum F_SETLK        = 6;
+    enum F_SETLKW       = 7;
+  }
   else
   static if( __USE_FILE_OFFSET64 )
   {
--- a/src/core/sys/posix/fcntl.d
+++ b/src/core/sys/posix/fcntl.d
@@ -760,6 +760,17 @@ else version( CRuntime_Bionic )
         enum O_NONBLOCK     = 0x800;    // octal    04000
         enum O_SYNC         = 0x1000;   // octal   010000
     }
+    else version (AArch64)
+    {
+        enum O_CREAT        = 0x40;     // octal     0100
+        enum O_EXCL         = 0x80;     // octal     0200
+        enum O_NOCTTY       = 0x100;    // octal     0400
+        enum O_TRUNC        = 0x200;    // octal    01000
+
+        enum O_APPEND       = 0x400;    // octal    02000
+        enum O_NONBLOCK     = 0x800;    // octal    04000
+        enum O_SYNC         = 0x101000; // octal 04010000
+    }
     else
     {
         static assert(false, "Architecture not supported.");
--- a/src/core/sys/posix/signal.d
+++ b/src/core/sys/posix/signal.d
@@ -175,15 +175,28 @@ else version(NetBSD)
 }
 else version (CRuntime_Bionic)
 {
-    enum SIGRTMIN = 32;
-    version(ARM)
-        enum SIGRTMAX = 64;
-    else version(X86)
-        enum SIGRTMAX = 64;
-    else version(MIPS32)
-        enum SIGRTMAX = 128;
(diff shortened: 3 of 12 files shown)


src/core/sys/posix/sys/types.d
```diff
@@ -418,7 +423,7 @@ else version( CRuntime_Bionic )
     alias uint     id_t;
     alias int      key_t;
     alias c_long   suseconds_t;
-    alias c_long   useconds_t;
+    alias uint     useconds_t; // Updated in Lollipop
```
ibuclaw: So the bindings will only ever support one version of Android?

src/rt/sections_android.d
```diff
@@ -155,14 +158,14 @@ ref void[] getTLSBlock() nothrow @nogc
...
-    if (!pary.length)
+    /*if (!pary.length)
     {
         auto pbeg = cast(void*)&_tlsstart;
         auto pend = cast(void*)&_tlsend;
         auto p = .malloc(pend - pbeg);
         memcpy(p, pbeg, pend - pbeg);
         *pary = p[0 .. pend - pbeg];
-    }
+    }*/
```
ibuclaw: Just remove it if it is no longer used.

src/rt/sections_android.d
```diff
@@ -180,9 +183,11 @@ extern(C)
         void* __start_minfo;
         void* __stop_minfo;
 
+        void* _tlsend;
         size_t __bss_end__;
 
+    }extern
```
ibuclaw: What's going on here?

src/core/sys/posix/signal.d
```diff
@@ -175,13 +175,35 @@ else version(NetBSD)
 }
 else version (CRuntime_Bionic)
 {
-    enum SIGRTMIN = 32;
+    //enum SIGRTMIN = 32;
```
ibuclaw: Remove rather than comment out?

src/core/stdc/stdlib.d
```diff
@@ -151,7 +151,7 @@ else
 // No unsafe pointer manipulation.
 @trusted
 {
-    version(CRuntime_Bionic)
+    version(none)
```
ibuclaw: Don't use `version(none)`

src/core/sys/posix/sys/types.d
joakim-noah: No, I believe every version since uses this now, just noting the version where the switch was made.

src/rt/sections_android.d
joakim-noah: This and all subsequent comments are because this pull is a Work-In-Progess, as noted in OP. I'll clean it up once the functionality is set.

src/rt/sections_android.d
joakim-noah: Avoids linker issues related to emulated TLS, just a hack for now.

src/core/sys/posix/fcntl.d
```diff
@@ -98,6 +98,12 @@ version( CRuntime_Glibc )
     enum F_SETLK        = 6;
     enum F_SETLKW       = 7;
   }
+  else version(AArch64)
+  {
+    enum F_GETLK        = 5;
+    enum F_SETLK        = 6;
+    enum F_SETLKW       = 7;
+  }
```
joakim-noah: This turned up when trying to run [this Phobos test on linux/AArch64](https://github.com/dlang/phobos/blob/master/std/stdio.d#L1416), dug this fix out of the Glibc headers. It appears Glibc uses these constants on AArch64 regardless of whether `__USE_FILE_OFFSET64` is set or not, just like on `x64`.

I don't want to go digging through the rat's nest of platform-specific Glibc headers to figure out why this is, so I just added this exception. If someone wants to do that digging and suggest a better way to normalize this, I'd be glad to make that change.

src/core/sys/posix/signal.d
```diff
@@ -760,7 +773,7 @@ else version( CRuntime_UClibc )
         static assert(false, "Architecture not supported.");
     }
 }
-else version (linux)
+else version (CRuntime_Bionic)
```
joakim-noah: [I switched this to `linux` a couple years ago](https://github.com/dlang/druntime/commit/aeecc9#diff-951b22b328337c533b375267d2f927c3R476), maybe thinking it matched the kernel APIs, but now see that it doesn't. All other C runtimes that usually run on linux have their own version blocks above, so nobody other than Bionic users were likely using this.

joakim-noah:
ping @ibuclaw, no longer WIP, this pull is ready to go.

src/core/sys/posix/signal.d
```diff
@@ -175,15 +175,28 @@ else version(NetBSD)
...
-    else
-        static assert(false, "Architecture not supported.");
+    // Switched to calling these functions since Lollipop
+    private extern (C) nothrow @nogc
+    {
+        int __libc_current_sigrtmin();
+        int __libc_current_sigrtmax();
+    }
+
+    @property int SIGRTMIN() nothrow @nogc {
```
ibuclaw: Could this be a template?

joakim-noah: It's a straight copy-and-paste of the Glibc block above, as Bionic switched to using the same function-calling scheme as Glibc. If you'd prefer that both blocks use template functions, I can modify them to that, don't care.

joakim-noah:
I don't know what's up with CircleCI, it can't build dmd? Would closing and re-opening this pull get it to kick off again and start working?

rainers:
> Would closing and re-opening this pull get it to kick off again and start working?

Should work now as the compiler has been upgraded by https://github.com/dlang/druntime/pull/2205

joakim-noah:
Thanks, now to cherry pick to ldc, which actually has an AArch64 backend.

PR #2149, opened 2018-03-23
add scope to abort()

WalterBright:


src/core/internal/abort.d
```diff
@@ -4,14 +4,14 @@ module core.internal.abort;
  * Use instead of assert(0, msg), since this does not print a message for -release compiled
  * code, and druntime is -release compiled.
  */
-void abort(string msg, string filename = __FILE__, size_t line = __LINE__) @nogc nothrow @safe
+void abort(scope string msg, string filename = __FILE__, size_t line = __LINE__) @nogc nothrow @safe
```
schveiguy: Why not scope on the `filename` parameter? True most of the time you use the default, but if you didn't want to, then this makes it slightly more difficult.

wilzbach:
Windows failure is unrelated -> braddr/d-tester#71

PR #2150, opened 2018-03-25
Build docs with -transition=markdown

dgileadi:
This is in support of the [PR to add Markdown support to Ddoc in dmd](https://github.com/dlang/dmd/pull/8043). This PR should not be merged until that PR is accepted.


dgileadi:
The meaning of `-transition=markdown` isn't what I previously understood, and this PR is unnecessary.

PR #2151, opened 2018-03-25
2 fixes for rt.backtrace.dwarf

kinke:


src/rt/backtrace/dwarf.d
```diff
@@ -71,23 +71,35 @@ int traceHandlerOpApplyImpl(const void*[] callstack, scope int delegate(ref size
...
-            snprintf(addressBuffer.ptr, addressBuffer.length, "%.*s:%d ", cast(int) locations[i].file.length, locations[i].file.ptr, locations[i].line);
+        {
+            appendToBuffer("%.*s:%d ", cast(int) locations[i].file.length, locations[i].file.ptr, locations[i].line);
+        }
         else
-            addressBuffer[] = "??:? \0";
+        {
+            buffer[0 .. 5] = "??:? ";
+            bufferLength = 5;
+        }
```
kinke: Fix 1:
```D
addressBuffer[] = "??:? \0";
```
=> `object.Error@(0): Array lengths don't match for copy` https://run.dlang.io/is/4RQZWh

src/rt/backtrace/dwarf.d
```diff
@@ -71,23 +71,35 @@ int traceHandlerOpApplyImpl(const void*[] callstack, scope int delegate(ref size
...
-            bufferLength = snprintf(buffer.ptr, buffer.length, "%s%.*s [0x%x]", addressBuffer.ptr, cast(int) symbol.length, symbol.ptr, callstack[i]);
-        else
-            bufferLength = snprintf(buffer.ptr, buffer.length, "%s[0x%x]", addressBuffer.ptr, callstack[i]);
+            appendToBuffer("%.*s ", cast(int) symbol.length, symbol.ptr);
 
-        assert(bufferLength >= 0);
+        static if (size_t.sizeof == 8)
+            appendToBuffer("[0x%llx]", callstack[i]);
+        else
+            appendToBuffer("[0x%x]", callstack[i]);
```
kinke: Fix 2: don't restrict to lower 32 bits of address for 64-bit platforms.

JinShil:
Does this fix https://issues.dlang.org/show_bug.cgi?id=18068?

kinke:
Nope, but I discovered bug #1 because of missing line infos too, or rather because the code here doesn't handle Dwarf v4+ debuginfos.

src/rt/backtrace/dwarf.d
wilzbach: Any chance we can add a test for this to prevent future regressions?

kinke: It's pretty bad with `-release`, causing a segfault as long as the destination buffer is sufficiently large: https://run.dlang.io/is/NEqZOB. Without `-release` an exception is thrown during stacktrace generation, which is also pretty bad.

The latter is how I stumbled on this. LLVM 6.0 led to DWARF v4 debuginfos, rt.backtrace.dwarf doesn't support that version, so no file/line infos => "??:? \0" + exception (debug druntime only; release druntime compiled with `-release` happily reads beyond the string constant, causing no segfault on x86 apparently). The exception with debug-druntime is swallowed in [my test case](https://github.com/ldc-developers/ldc/blob/1572c91fc0dea24d9aa642c07300b1ca5ba82ea6/tests/codegen/exception_stack_trace.d), it only manifests itself in an incomplete stack trace (single entry).

kinke: @wilzbach: There's no debug druntime (i.e., compiled with enabled bounds checks) available for CI, right? Then it's untestable.

kinke:
I extended the previous trivial fix by a revised version of `rt.backtrace.dwarf`, tackling:

* Support for DWARF v4.
* Support for 64-bit DWARF.
* Alignment issues on non-x86 architectures (DWARF data is packed).
* [Issue #18068](https://issues.dlang.org/show_bug.cgi?id=18068); DSO-relative DWARF addresses vs. absolute addresses returned by `backtrace()`.

Tests are missing, I'm unsure how to tackle that in a robust fashion (for DMD, I'd know how for LDC ;)).

kinke:
Rebased and obsolete 18068 fix removed. FWIW, the first commit is part of LDC 1.9, the second in 1.10.

kinke:
I restricted this to the first trivial fixes again and moved the 2nd commit to #2559. @thewilsonator: Please push.

PR #2152, opened 2018-03-26
Issue 18652 - hashOf example doesn't compile

WalterBright:
https://issues.dlang.org/show_bug.cgi?id=18652


wilzbach:
Oh and there was also something wrong with your git commit. It seems like you did amend to the last commit (a merge commit). I fixed that too.

PR #2154, opened 2018-03-27
[uClibc] core.stdc.fenv add mips32 architecture

rracariu:


rracariu:
Sure, this PR can be referenced to https://github.com/hwoarang/uClibc/blob/master-metag/libc/sysdeps/linux/mips/bits/fenv.h#L59

This missing definition prevented compiling on a MIPS32 target ([LinkIt™ Smart 7688](https://labs.mediatek.com/en/platform/linkit-smart-7688)) that I'm working on.

Thanks!

rracariu:
@ibuclaw I haven't found any references to MIPS64 in uClibc. I think that architecture is not supported.

ibuclaw:
Maybe because they both use the same path? Just speculating though. I would imagine that you won't see any indication other than an `#if` condition for a word size of 64.

PR #2155, opened 2018-03-28
Add access to GC runtime profile stats

marcioapm:
Would be very helpful to get this in the next public release.


wilzbach:
This will need a changelog entry - see the `changelog` folder.

wilzbach:
@marcioapm FYI as this is a new symbol it requires approval from @andralex.
I'm also CCing @rainers and @DmitryOlshansky who are our GC experts.

src/core/memory.d
```diff
@@ -139,6 +139,7 @@ private
 
     extern (C) BlkInfo_ gc_query( void* p ) pure nothrow;
     extern (C) GC.Stats gc_stats ( ) nothrow @nogc;
+    extern (C) GC.ProfileStats gc_profileStats ( ) nothrow @nogc;
```
andralex: `@safe` too please

src/core/memory.d
```diff
@@ -170,6 +171,18 @@ struct GC
...
+    /**
+      * Aggregation of current profile information, requires profiling to be on
+     */
+    static struct ProfileStats {
+        /// total number of GC cycles
+        size_t numCollections;
+        /// total time spent doing GC in hnsecs
+        long totalCollectionTime;
+        // largest time spent doing one GC cycle in hnsecs
+        long maxCollectionTime;
```
andralex: `ulong`

src/core/memory.d
```diff
@@ -170,6 +171,18 @@ struct GC
...
     }
 
+    /**
+      * Aggregation of current profile information, requires profiling to be on
+     */
+    static struct ProfileStats {
+        /// total number of GC cycles
+        size_t numCollections;
+        /// total time spent doing GC in hnsecs
+        long totalCollectionTime;
```
andralex: `ulong`

src/core/memory.d
```diff
@@ -680,6 +693,16 @@ struct GC
         return gc_stats();
     }
 
+    /**
+     * Returns runtime profile stats for currently active GC implementation
+     * See `core.memory.GC.ProfileStats` for list of available metrics.
+     * Requires GC profiling to be on
+     */
+    static ProfileStats profileStats() nothrow
```
andralex: safe nogc

marcioapm:
Will this be included in the next public release?

wilzbach:
@marcioapm We can't promise this as the next master-stable branch-off is in two weeks, but as Andrei has already approved this, it looks pretty good.
This just needs to be looked at by one of our GC experts - I don't see anything else that's blocking this.

rainers:
I'm not a huge fan of adding interface members for every stat function which can be specific to a single GC. I think there should be a more generic way to request information from the GC, e.g. using a tagged union. Please also note that there are other PRs adding similar functions: https://github.com/dlang/druntime/pull/2052, https://github.com/dlang/druntime/pull/1846.

Maybe we should also switch GC from an interface to a base class so that it can have (empty) default implementation. That way it might be easier to add functions without having to update every GC. In addition it could result in a tiny performance improvement because it doesn't need the interface thunks that adjusts the class instance offset.

mihails-strasuns-sociomantic:
I would expect this to be nested under regular `GC.Stats`

marcioapm:
Filling in `GC.Stats` is not O(1), requires a lock, and doesn't depend on GC profiling being on.
They are somewhat related, but `GC.Stats` seems to be about memory statistics, at least in it's current form, whilst `GC.ProfileStats` is about performance.

mihails-strasuns-sociomantic:
You don't expect to call `ProfileStats` in a hot loop, do you? Because implementation of actual GC profiling is very far from being usable for that (I tweak that a bit in https://github.com/dlang/druntime/pull/1846 but that is still different league of performance).

The fact that it doesn't depend on GC profiling being on is no different for actual GC profiling data - it is always compiled as part of runtime and available to be used. So if you link it with some modules that are compiled with `-profile=gc`, it will get populated, and otherwise it will stay zero. But it is still there all the time - and in presence of shared libraries it is not even known if it will _stay_ zero if nothing uses GC profiling right now. This is exactly why I expect it to be an inherent part of GC stats.

mihails-strasuns-sociomantic:
Oh, stupid me. I have read PR title and assumed it is also about `-profile=gc` stats - and only now paid attention to actual content - which seems to be for `--DRT-opt` profile thing.

Nevermind.

marcioapm:
This is failing because of a shell script.
```
[libdparse] Running shell script
+ cd test
+ ./run_tests.sh
find: ‘../stdx-allocator/source’: No such file or directory
Compiling unit tests...../src/dparse/parser.d(9): Error: module `mallocator` is in file 'stdx/allocator/mallocator.d' which cannot be read
import path[0] = ../src/
import path[1] = ../stdx-allocator/source
import path[2] = /var/lib/jenkins/dlang_projects@3/distribution/bin/../imports
+ find test/coverage -type f -exec mv {} . ;
find: ‘test/coverage’: No such file or directory
script returned exit code 1```

**Comment by wilzbach** [comment]
Don't worry about Jenkins - it's currently broken and known -> https://github.com/dlang/ci/issues/190 (and here's yours: https://github.com/dlang/ci/issues/193)

File: src/core/memory.d [hunk 5]
```diff
@@ -170,6 +171,18 @@ struct GC
...
     }
 
+    /**
+      * Aggregation of current profile information, requires profiling to be on
+     */
+    static struct ProfileStats {
+        /// total number of GC cycles
+        size_t numCollections;
+        /// total time spent doing GC in hnsecs
+        ulong totalCollectionTime;
```
rainers: I think it would be better to keep Duration as the time type.

src/gc/impl/conservative/gc.d
```diff
@@ -1201,6 +1201,17 @@ class ConservativeGC : GC
     }
 
 
+    core.memory.GC.ProfileStats profileStats() nothrow
+    {
+        typeof(return) ret;
+
+        ret.numCollections = numCollections;
+        ret.totalCollectionTime = (prepTime + markTime + sweepTime + recoverTime).total!"hnsecs";
+        ret.maxCollectionTime = maxPauseTime.total!"hnsecs";
```
rainers: collection time and pause time are not the same. The latter is the time that all other threads are suspended aka the world is stopped.

MartinNowak:
Please only use milestones for things that MUST be in the next release, e.g. because it completes an already merged PR. Please do not use them for changes you want in the next release or for general planning. We're releasing often enough so that nobody has to run for the release train.
Milestones are used as a communication tool so we don't miss to release essential PRs.

jacob-carlborg:
What is the status of this?

marcioapm:
@rainers I think I got all your concerns addressed.

rainers:
LGTM but the `Requires GC profiling to be active.` hints are no longer valid.

The CI failures seem unrelated. core.sys.posix.stdio is not compiled into the druntime lib for some reason...

src/core/memory.d
```diff
@@ -170,6 +171,23 @@ struct GC
...
+     */
+    static struct ProfileStats {
+        import core.time : Duration;
+        /// total number of GC cycles
+        size_t numCollections;
+        /// total time spent doing GC
+        Duration totalCollectionTime;
+        /// total time threads were paused doing GC
+        Duration totalPauseTime;
+        // largest time threads were paused during one GC cycle
```
rainers: ddoc `///` here and below, too.

src/core/memory.d
```diff
@@ -680,6 +698,16 @@ struct GC
         return gc_stats();
     }
 
+    /**
+     * Returns runtime profile stats for currently active GC implementation
+     * See `core.memory.GC.ProfileStats` for list of available metrics.
+     * Requires GC profiling to be on
```
rainers: Still mentions GC profiling needs to be on.

src/gc/gcinterface.d
```diff
@@ -148,6 +148,12 @@ interface GC
      */
     core.memory.GC.Stats stats() nothrow;
 
+    /**
+     * Retrieve profile statistics about garbage collection.
+     * Useful for debugging and tuning. Requires profiling to be on.
```
rainers: Here, too.

rainers:
Closing and reopening to retrigger CIs.

ZombineDev:
Rebased in #2332.

ZombineDev:
I didn't intend to close this PR, GitHub somehow automatically did. I rebased @marcioapm's branch locally, but I couldn't force push it to his branch as the PR got closed, so I opened #2332 - let's continue the discussion there.

PR #2157, opened 2018-04-03
[uClibc] fix TLS segment mod

rracariu:
uClibc doesn't provide `dlpi_tls_modid` in the `dl_phdr_info` struct definition, see https://github.com/kraj/uclibc-ng/blob/1.0/include/link.h where the definition is disabled.

Testing on MIPS32 revealed this bug trough a SEG FAULT.


src/rt/sections_elf_shared.d
```diff
@@ -753,8 +753,15 @@ void scanSegments(in ref dl_phdr_info info, DSO* pdso) nothrow @nogc
 
         case PT_TLS: // TLS segment
             assert(!pdso._tlsSize); // is unique per DSO
-            pdso._tlsMod = info.dlpi_tls_modid;
-            pdso._tlsSize = phdr.p_memsz;
+            version (CRuntime_UClibc)
+            {
+              pdso._tlsSize = phdr.p_memsz;
```
joakim-noah: Set in both so put it outside. Best way to do this change is to put a `version(uclibc) {} else` on the line before the `_tlsMod` statement you want omitted.

rracariu: done. thanks!

ibuclaw: I'm not so sure if I agree with this style though.

src/rt/sections_elf_shared.d
```diff
@@ -753,6 +753,7 @@ void scanSegments(in ref dl_phdr_info info, DSO* pdso) nothrow @nogc
 
         case PT_TLS: // TLS segment
             assert(!pdso._tlsSize); // is unique per DSO
+            version (CRuntime_UClibc) {} else
```
ibuclaw: To follow up on @joakim-noah's suggestion, I would prefer the following style here.
```
version (CRuntime_UClibc)
{
    // comment explaining why.
} 
else
    pdso._tlsMod = info.dlpi_tls_modid;
```
As it makes the version block stand out in the code, rather than hidden away.

rracariu: Changed as suggested. Thanks!

ibuclaw:
@joakim-noah - Looks ok?

@rracariu - Is the `_tlsMod` field present in the strict?  Maybe it could be removed entirely to save on the storage.  But maybe later.

jpf91:
@ibuclaw @MartinNowak this looks dangerous to me. This means `getTLSRange` will always return `[]`, but doesn't this mean, the GC will not be able to scan the TLS segments on uclibc? This leads to hard to debug errors, in general the testsuite will not catch it. I think we have code (IIRC references to per-thread default RNG in std.random?) which will break after a GC collection. Or am I missing something here?

jacob-carlborg:
> uClibc doesn't provide dlpi_tls_modid in the dl_phdr_info struct definition, see https://github.com/kraj/uclibc-ng/blob/1.0/include/link.h#L185 where the definition is disabled.

According to the above link the field might be available:

"Following members were introduced after the first version of this structure was available.  Check the SIZE argument passed to the dl_iterate_phdr callback to determine whether or not each later member is available".

rracariu:
@jpf91 - My understanding is this only affects shared ELF - static linking should not be affected, an this is the only way I ran the unit-tests.  Can you think of a test-case where this GC issue can be observed? I have hardware and tool-chains for both MIPS32 and ARM, so I can ran any tests required.

@jacob-carlborg On my toolchain dlpi_tls_modid gives bad values, which means it is not present in the struct layout, this showed in the debugger when porting druntime to uClibc.

jacob-carlborg:
> On my toolchain dlpi_tls_modid gives bad values, which means it is not present in the struct layout, this showed in the debugger when porting druntime to uClibc.

According to the documentation you linked, you're supposed to check the size of the struct to determine if the `dlpi_tls_modid` field is present.

rracariu:
How can I effectively do that from the D side?

jacob-carlborg:
The size is apparently passed to the `dl_iterate_phdr` callback:

"Check the SIZE argument passed to the dl_iterate_phdr callback to determine whether or not each later member is available".

rracariu:
Something like
```diff
 src/rt/sections_elf_shared.d | 19 +++++++++++++------
 1 file changed, 13 insertions(+), 6 deletions(-)

diff --git a/src/rt/sections_elf_shared.d b/src/rt/sections_elf_shared.d
index 473e5346..0dcb3d67 100644
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -403,10 +403,11 @@ extern(C) void _d_dso_registry(CompilerDSOData* data)
         pdso._moduleGroup = ModuleGroup(toRange(data._minfo_beg, data._minfo_end));
 
         dl_phdr_info info = void;
-        const headerFound = findDSOInfoForAddr(data._slot, &info);
+        size_t dl_phdr_info_size;
+        const headerFound = findDSOInfoForAddr(data._slot, &info, &dl_phdr_info_size);
         safeAssert(headerFound, "Failed to find image header.");
 
-        scanSegments(info, pdso);
+        scanSegments(info, pdso, dl_phdr_info_size);
 
         version (Shared)
         {
@@ -753,7 +754,7 @@ version (Shared)
  * Scan segments in Linux dl_phdr_info struct and store
  * the TLS and writeable data segments in *pdso.
  */
-void scanSegments(in ref dl_phdr_info info, DSO* pdso) nothrow @nogc
+void scanSegments(in ref dl_phdr_info info, DSO* pdso, size_t dl_phdr_info_size) nothrow @nogc
 {
     foreach (ref phdr; info.dlpi_phdr[0 .. info.dlpi_phnum])
     {
@@ -776,7 +777,12 @@ void scanSegments(in ref dl_phdr_info info, DSO* pdso) nothrow @nogc
             safeAssert(!pdso._tlsSize, "Multiple TLS segments in image header.");
             version (CRuntime_UClibc)
             {
-                // uClibc doesn't provide a 'dlpi_tls_modid' definition
+                // uClibc 'dlpi_tls_modid' definition is not always defined
+                // need to check the size to determine its presence
+                if (dl_phdr_info_size == dl_phdr_info.sizeof)
+                {
+                    pdso._tlsMod = info.dlpi_tls_modid;
+                }
             }
             else
                 pdso._tlsMod = info.dlpi_tls_modid;
@@ -797,7 +803,7 @@ void scanSegments(in ref dl_phdr_info info, DSO* pdso) nothrow @nogc
  * References:
  *      http://linux.die.net/man/3/dl_iterate_phdr
  */
-bool findDSOInfoForAddr(in void* addr, dl_phdr_info* result=null) nothrow @nogc
+bool findDSOInfoForAddr(in void* addr, dl_phdr_info* result=null, size_t* dl_phdr_info_size=null) nothrow @nogc
 {
     version (linux)       enum IterateManually = true;
     else version (NetBSD) enum IterateManually = true;
@@ -805,7 +811,7 @@ bool findDSOInfoForAddr(in void* addr, dl_phdr_info* result=null) nothrow @nogc
 
     static if (IterateManually)
     {
-        static struct DG { const(void)* addr; dl_phdr_info* result; }
+        static struct DG { const(void)* addr; dl_phdr_info* result; size_t* dl_phdr_info_size}
 
         extern(C) int callback(dl_phdr_info* info, size_t sz, void* arg) nothrow @nogc
         {
@@ -813,6 +819,7 @@ bool findDSOInfoForAddr(in void* addr, dl_phdr_info* result=null) nothrow @nogc
             if (findSegmentForAddr(*info, p.addr))
             {
                 if (p.result !is null) *p.result = *info;
+                *p.dl_phdr_info_size = sz;
                 return 1; // break;
             }
             return 0; // continue iteration
```

jacob-carlborg:
Yes, possibly.

jpf91:
@rracariu Something like this should trigger the problem:
```d
import core.memory, core.thread, core.bitop;

class TestTLS
{
    uint a;
    void addNumber()
    {
        auto val = volatileLoad(&a);
        val++;
        volatileStore(&a, val);
    }
}

TestTLS tlsPtr;

static this()
{
    tlsPtr = new TestTLS();
}

void main()
{
    void runThread()
    {
        for(size_t i = 0; i < 100; i++)
        {
            Thread.sleep(10.msecs);
            tlsPtr.addNumber();
            GC.collect();
        }
    }

    Thread[] threads;
    for(size_t i = 0; i < 20; i++)
    {
        auto t = new Thread(&runThread);
        threads ~= t;
        t.start();
    }
    runThread();

    foreach(thread; threads)
        thread.join();
}
```
The `volatilePart` is necessary to avoid any compiler optimizations. I think this code should reliably trigger the bug. If you can reproduce a problem using that code it would be great if you could add this as a test case to the test suite in addition to the updated uclibc code.

rracariu:
@jpf91 - Thanks for the sample. It works (albeit using an older LDC because of a regression in recent ones (https://github.com/ldc-developers/ldc/issues/2994)). Tested with a static link build.

The patch above is not effective, as the size of `dl_phdr_info_size` is always less than `dl_phdr_info.sizeof` - the fields are not present, at least on the toolchains I use.

PR #2158, opened 2018-04-06
[uClibc] core.sys.posix.signal add mips32 defs

rracariu:
Adds missing MIPS32 definitions. For reference consult C headers (common/mips):

https://github.com/kraj/uclibc-ng/blob/57f820bab9108550907f03828d61525c833004a4/libc/sysdeps/linux/mips/bits/sigaction.h#L25
https://github.com/kraj/uclibc-ng/blob/57f820bab9108550907f03828d61525c833004a4/libc/sysdeps/linux/common/bits/sigaction.h#L28

https://github.com/kraj/uclibc-ng/blob/57f820bab9108550907f03828d61525c833004a4/libc/sysdeps/linux/mips/bits/siginfo.h#L52
https://github.com/kraj/uclibc-ng/blob/57f820bab9108550907f03828d61525c833004a4/libc/sysdeps/linux/common/bits/siginfo.h#L50

### DIFF
--- a/src/core/sys/posix/signal.d
+++ b/src/core/sys/posix/signal.d
@@ -710,23 +710,54 @@ else version (Solaris)
 }
 else version( CRuntime_UClibc )
 {
-    struct sigaction_t
+    version (ARM)           version = sigaction_common;
+    else version (X86_64)   version = sigaction_common;
+
+    version (sigaction_common)
     {
-        static if( true /* __USE_POSIX199309 */ )
+        struct sigaction_t
         {
-            union
+            static if( true /* __USE_POSIX199309 */ )
+            {
+                union
+                {
+                    sigfn_t     sa_handler;
+                    sigactfn_t  sa_sigaction;
+                }
+            }
+            else
             {
                 sigfn_t     sa_handler;
-                sigactfn_t  sa_sigaction;
             }
+            c_ulong     sa_flags;
+            void function() sa_restorer;
+            sigset_t    sa_mask;
         }
-        else
+    }
+    else version (MIPS32)
+    {
+        struct sigaction_t
         {
-            sigfn_t     sa_handler;
+            uint     sa_flags;
+            static if( true /* __USE_POSIX199309 */ )
+            {
+                union
+                {
+                    sigfn_t     sa_handler;
+                    sigactfn_t  sa_sigaction;
+                }
+            }
+            else
+            {
+                sigfn_t     sa_handler;
+            }
+            sigset_t    sa_mask;
+            void function() sa_restorer;
         }
-        c_ulong         sa_flags;
-        void function() sa_restorer;
-        sigset_t        sa_mask;
+    }
+    else
+    {
+        static assert(false, "Architecture not supported.");
     }
 }
 else version (linux)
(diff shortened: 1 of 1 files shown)


src/core/sys/posix/signal.d
```diff
@@ -710,23 +710,47 @@ else version (Solaris)
...
-                sigactfn_t  sa_sigaction;
             }
+            sigset_t    sa_mask;
+            void function() sa_restorer;
         }
-        else
+    }
+    else
+    {
+        struct sigaction_t
```
ibuclaw: Preferably, if there's a type definition that differs between architectures, then each supported architecture should be explicitly defined.

The else branch then becomes a `static assert(false)` that is intended to alert porters of something that needs checking.

src/core/sys/posix/signal.d
```diff
@@ -1687,6 +1711,8 @@ else version( CRuntime_UClibc )
                             // this signal, as defined in <errno.h>
         int si_code;        // Signal code
 
+        version ( MIPS32 ) int[__SI_MAX_SIZE / int.sizeof - __SI_PAD_SIZE - 3] __pad0;
```
ibuclaw: I'd prefer not to have version conditions nested in as per previous comment.

src/core/sys/posix/signal.d
rracariu: uClibc has this common definition, ARM and x86 uses it, mips doesn't (probably others, haven't checked). I guess I can add a "sigaction_common" version for ARM and x86 and then separate the MIPS and add the static assert in the end.

src/core/sys/posix/signal.d
rracariu: Didn't wanted to duplicate that huge structure...

rracariu:
ping

rracariu:
ping :)

src/core/sys/posix/signal.d
```diff
@@ -1680,90 +1711,191 @@ else version( CRuntime_UClibc )
...
+            @property ref clock_t si_stime() return { return _sifields._sigchld.si_stime; }
+            @property ref sigval si_value() return { return _sifields._rt.si_sigval; }
+            @property ref int si_int() return { return _sifields._rt.si_sigval.sival_int; }
+            @property ref void* si_ptr() return { return _sifields._rt.si_sigval.sival_ptr; }
+            @property ref void* si_addr() return { return _sifields._sigfault.si_addr; }
+            @property ref c_long si_band() return { return _sifields._sigpoll.si_band; }
+            @property ref int  si_fd() return { return _sifields._sigpoll.si_fd; }
+            @property ref void*  si_call_addr() return { return _sifields._sigsys._call_addr; }
+            @property ref int  si_syscall() return { return _sifields._sigsys._syscall; }
+            @property ref uint  si_arch() return { return _sifields._sigsys._arch; }
```
ibuclaw: Can these be templates for the benefit of those who use the C bindings, but don't link to druntime?

rracariu: changed to templates, removed `return` attribute as it should be inferred for templates.

ibuclaw:
Yeah, I raised this here: https://github.com/dlang/druntime/pull/2158#discussion_r179910200

On the one hand, the use of `sigaction_common` is the same as what we do with `X86_Any` in standard math.  I would prefer separation, despite duplication, but at least there's a static assert for platforms not tested.

rracariu:
On the uClibc side there is that common definition header included in the PR message, then it comes in various flavors for divergent platforms. At least the organization is similar for anybody looking at this and the headers.

I'm not a big fan of duplicating huge structures, I see them in druntime in multiple places,

PR #2160, opened 2018-04-09
config.d: use enums instead of structs for __c_long, etc.

WalterBright:
needs https://github.com/dlang/dmd/pull/8152

### DIFF
--- a/src/core/stdc/config.d
+++ b/src/core/stdc/config.d
@@ -1,5 +1,6 @@
-/**
- * D header file for C99.
+/***
+ * D compatible types that correspond to various basic types in associated
+ * C and C++ compilers.
  *
  * Copyright: Copyright Sean Kelly 2005 - 2009.
  * License: Distributed under the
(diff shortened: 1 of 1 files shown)


WalterBright:
Fix https://issues.dlang.org/show_bug.cgi?id=17890

src/core/stdc/config.d
```diff
@@ -21,40 +69,10 @@ else version (TVOS)
...
-     * 1. Implicit casting of an int to __c_long doesn't happen, because D doesn't
-     *    allow constructor calls in implicit conversions.
-     * 2. long lng;
-     *    cast(__c_long)lng;
-     *    does not work because lng has to be implicitly cast to an int in the constructor,
-     *    and since that truncates it is not done.
-     * Both of these break existing code, so until we find a resolution the types are named
-     * cpp_xxxx.
-     */
+    enum __c_long  : int;
```
jmdavis: I had no clue that it was even legal to declare an `enum` with no members. That seems very weird. Is the point of doing this instead of using an `alias` so that it's a distinct type from `int`? I guess that that would mean that this technique is a `typedef` of sorts.

jacob-carlborg: > Is the point of doing this instead of using an alias so that it's a distinct type from int?

Yes, so that the compiler can do special mangling for that type. Although I don't see how it's any better than a struct. I proposed to add support for adding `pragma(mangle)` to alias declarations but Walter doesn't seem to like that.

WalterBright:
This is a Vision issue, as it resolves a long-running sore in compatibility with C++.

src/core/stdc/config.d
```diff
@@ -150,3 +155,15 @@ else version( SDC )
 }
 
 static assert(is(c_long_double), "c_long_double needs to be declared for this platform/architecture.");
+
+version (Darwin)
+{
+    alias cpp_size_t = cpp_ulong;
```
jacob-carlborg: What happens if we do the other way around, i.e. `alias size_t = cpp_ulong` in `object.d`? This would be to avoid having to use `cpp_size_t` instead of `size_t`.

jacob-carlborg:
I think there should be a test somewhere (in druntime or in DMD) that verifies that the correct mangling is used.

src/core/stdc/config.d
jacob-carlborg: @jmdavis BTW, one use case of declaring an enum without any members is to use it as a UDA that doesn't contain any data.

WalterBright:
> I think there should be a test somewhere (in druntime or in DMD) that verifies that the correct mangling is used.

https://github.com/dlang/dmd/pull/8152

WalterBright:
@CyberShadow I don't understand this error message:
```
Test the D Language specification
../dmd/generated/linux/release/64/dmd -run tools/dspec_tester.d --compiler=../dmd/generated/linux/release/64/dmd
---
void main() {
import std.stdio;
import std.algorithm;
int[] s = [1, 2, 3, 4];

copy(s[1..3], s[0..2]);
assert(s == [2, 3, 3, 4]);

}
---
gcc: error: __stdin.o: No such file or directory
struct.dd: 35 examples found
traits.dd: 1 examples found
template.dd: 1 examples found
type.dd: 4 examples found
arrays.dd: 31 examples found
function.dd: 36 examples found
posix.mak:811: recipe for target 'test_dspec' failed
```

WalterBright:
DAutoTest appears to fail the same way with other PRs, so is not caused by this PR.

CyberShadow:
The problem is in DMD's handling of programs from stdin:
https://issues.dlang.org/show_bug.cgi?id=18762#c1

PR #2162, opened 2018-04-14
This adds an Associative Array function getOrAdd

GilesBathgate:
The function provides a means to get a value corresponding to the key, but if the value doesn't exist it will evaluate the lazy argument to create a new value, add this to the associative array and then return it.

```d
auto p = lookup.getOrAdd("giles", new Person);
p.eyeColor = Color.Brown;
//...
assert("giles" in lookup);
```


src/rt/aaA.d
```diff
@@ -376,8 +396,10 @@ extern (C) void* _aaGetY(AA* aa, const TypeInfo_AssociativeArray ti, in size_t v
...
 
     // found a value => return it
-    if (auto p = aa.findSlotLookup(hash, pkey, ti.key))
+    if (auto p = aa.findSlotLookup(hash, pkey, ti.key)) {
+        found = true;
         return p.entry + aa.valoff;
+    }
 
     auto p = aa.findSlotInsert(hash);
     if (p.deleted)
```
GilesBathgate: I think I should be able to avoid the `p.entry =  allocEntry...`  when doing getOrAdd, but not sure how yet.

DmitryOlshansky:
Save for the name the idea is awesome and it was a blocker for eg use of builtin AAs in DMD. I never seen add mean insert in D language except maybe in my codepoint set stuff in std.uni.

Bikesheding the name can be elsewhere though;)

wilzbach:
The NG (aka forum.dlang.org - general)

GilesBathgate:
@wilzbach @DmitryOlshansky : ok we can discuss here then: https://forum.dlang.org/post/ejdiweskylrprbculbna@forum.dlang.org

src/object.d
```diff
@@ -2309,6 +2310,85 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
...
+    assert(b is a);
+    auto c = aa.get("bar", new T);
+    assert(T.count == 2);
+    assert(c !is a);
+
+    //Obviously get doesn't add.
+    assert("bar" !in aa);
+}
+
+auto getOrAdd(K, V)(ref V[K] aa, K key, lazy V value)
```
aG0aep6G: Return by `ref`, please. Could also use `V.init` as a default for `value`. Then we can write code like this:
```d
struct S { int value; }
S[string] aa;
aa.getOrAdd("foo").value = 42;
assert(aa == ["foo": S(42)]);
```

GilesBathgate: I wondered whether it should be `ref` (which is why I added the basic value unit test), I hadn't considered this scenario, I will add a test for it and update.

GilesBathgate: The default value is also a nice idea, so I will add that too (I didn't know lazy args could have a default value)

schveiguy: Hm... I just suggested a name of `getPtr` on the newsgroup, as I assumed that's what you were doing, but it seems you actually are returning by value. I think returning by ref is preferable to either (maybe `getRef`?). Note that if you wanted to create a pointer to the value, you have to go through a *triple* lookup today:

```D
Value *p = void;
if(!(p = key in aa)) // one
{
    aa[key] = initializeValue(); // two
    p = key in aa;  // three
}
```
Having this function return by ref makes this very easy to do in one lookup.

Note that a class value type is going to be confusing here unless you do a default value of `new Value`.

GilesBathgate: @schveiguy  `getPtr`and `getRef` are certainly both terse which I like. My only criticism is that they don't as clearly communicate that the value will be added/inserted into the associative array.

I am a little bit confused by the triple lookup example. Can I not just do `p = (aa[key] = initializeValue())`

That said this PR will make this much more simple: `Value* p = aa.getOrAdd(key, initializeValue());` Is there a unittest you can suggest to ensure this works as you expected?

schveiguy: > Can I not just do `p = (aa[key] = initializeValue())`

You are still thinking in terms of classes. You may want a pointer to a value type (think `int[int]`), where you need to use `key in aa` to avoid further lookups.

GilesBathgate: >In all honesty, before looking closer, I thought the get function already did this.
Ha, yeah I made the same mistake...that lead to a few hours of head scratching ;)

GilesBathgate: @schveiguy I have added a unittest which I think covers your example. @aG0aep6G I've made the updates as you've suggested too.

src/object.d
```diff
@@ -2309,6 +2310,94 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
...
+
+    aa["bar"] = S(2);
+    auto a = aa.getOrAdd("bar", S(3));
+    assert(a == S(2));
+
+    auto b = aa["bar"];
+    assert(b == S(2));
+
+    S* c = &aa.getOrAdd("baz", S(4));
+    assert(c is &aa["baz"]);
```
GilesBathgate: @schveiguy This is the unittest I am referring to.

src/object.d
```diff
@@ -2309,6 +2310,94 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
...
+    auto e = aa.getOrAdd("baz", new T);
+    assert(T.count == 2);
+    assert(e !is a);
+
+    assert("baz" in aa);
+}
+
+unittest
+{
+    struct S
```
schveiguy: static struct here, otherwise there is a connection between the unittest stack frame and every S instance.

src/object.d
```diff
@@ -2309,6 +2310,145 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
...
+    assert(b is a);
+    auto c = aa.get("bar", new T);
+    assert(T.count == 2);
+    assert(c !is a);
+
+    //Obviously get doesn't add.
+    assert("bar" !in aa);
+}
+
+ref V require(K, V)(ref V[K] aa, K key, lazy V value = V.init) pure
```
JinShil: new public symbols need a DDoc header with appropriate summary, `params`, and `returns`.  See https://dlang.org/spec/ddoc.html#standard_sections

src/object.d
```diff
@@ -2309,6 +2310,145 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
...
+    assert(c is &aa["baz"]);
+    assert(*c == S(4));
+
+    assert("baz" in aa);
+
+    auto d = aa["baz"];
+    assert(d == S(4));
+}
+
+void update(K, V, C, U)(ref V[K] aa, K key, C create, U update)
```
JinShil: New public symbols need a DDoc header with appropriate summary, `params`, and `returns`.  See https://dlang.org/spec/ddoc.html#standard_sections

JinShil:
Looks really nice.  Why didn't it go anywhere?

@WalterBright and @andralex:  This is modifying the public API.  Please chime in on the API change.

JinShil:
BTW, the lack of this feature prevented a change in DMD relatively recently (https://github.com/dlang/dmd/pull/7696#discussion_r161407629)  cc @ibuclaw

src/object.d
GilesBathgate: No problem, I can add that. There is some documentation added in dlang/dlang.org#2343 if you could review and make comments.

GilesBathgate:
>This is modifying the public API.

@JinShil A more conservative approach would be to add these functions to a phobos module (`std.array` ?).

For me `object.d` feels like the right place for this, which is why I wanted to be absolutely sure everyone would be happy with the symbol name, we had quite an in-depth bikeshed discussion: https://forum.dlang.org/post/ejdiweskylrprbculbna@forum.dlang.org

>The lack of this feature prevented a change in DMD relatively recently

Its certainly useful for those cases, which is why equivalent features exist in other languages/libraries. I think at the very least the change to druntime `aaA.d` (as was discussed in https://github.com/dlang/dmd/pull/7696#discussion_r161407629) should be considered, and then high level APIs can at least be built on top of that.

JinShil:
> A more conservative approach would be to add these functions to a phobos module (std.array ?).

If that were done we wouldn't be able to utilize this in DMD as, as a matter of policy, Phobos dependencies in DMD are discouraged.  If associative arrays are to be utilized in DMD or in the DRuntime itself, it is best if they reside in DRuntime.

src/object.d
JinShil: The DDoc header is still required for library documentation at https://dlang.org/phobos/object.html

GilesBathgate: >No problem, I can add that.

src/object.d
```diff
@@ -2309,6 +2310,145 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
...
+    static class C{}
+    C[string] aa;
+
+    C orig = new C;
+    aa["foo"] = orig;
+
+    C newer;
+    C older;
+
+    void test(string key) {
```
jacob-carlborg: Curly braces on their own lines.

src/rt/aaA.d
```diff
@@ -376,8 +397,10 @@ extern (C) void* _aaGetY(AA* aa, const TypeInfo_AssociativeArray ti, in size_t v
     immutable hash = calcHash(pkey, ti.key);
 
     // found a value => return it
-    if (auto p = aa.findSlotLookup(hash, pkey, ti.key))
+    if (auto p = aa.findSlotLookup(hash, pkey, ti.key)) {
```
jacob-carlborg: Curly braces on their own lines.

schveiguy:
> BTW, the lack of this feature prevented a change in DMD relatively recently

We have to be careful with depending on certain features of the compiler to build the compiler. At this point, we are still on 2.068.2 as the base compiler. When we fix certain issues, I think we will be at 2.079. The more we use newer compiler features, the more stages you need to bootstrap the compiler from C++.

That being said, I'd love to see this change get in. IIRC, I was waiting for the return value to be `ref` (got sidetracked, sorry @GilesBathgate), and we also need @andralex's approval for the added names.

Will go through again, and do a proper review.

src/rt/aaA.d
GilesBathgate: @jacob-carlborg Fixed

src/object.d
GilesBathgate: @jacob-carlborg Fixed

src/object.d
```diff
@@ -2309,6 +2387,164 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
...
+ * Looks up key; if it exists applies the update delegate else evaluates the
+ * create delegate and adds it to the associative array
+ * Params:
+ *      aa =     The associative array.
+ *      key =    The key.
+ *      create = The delegate to apply on create.
+ *      update = The delegate to apply on update.
+ */
+void update(K, V, C, U)(ref V[K] aa, K key, C create, U update)
+if (is(C : V delegate()) || is(C : V function()) && (is(U : V delegate(ref V)) || is(U : V function(ref V))))
```
klickverbot: Missing parens around `||`.

JinShil:
I don't think @andralex is interested in this.  I think we should keep this open for 1 more week in a "last call" for objections.  If no show-stoppers are raised, I think we should merge this.

JinShil:
I do think this could benefit from a changelog entry, however.  It's quite an important change, I think it would be good to advertise it's availability.  See https://github.com/dlang/druntime/blob/master/changelog/README.md

JinShil:
@GilesBathgate I'm just waiting on a changelog entry.

GilesBathgate:
@JinShil Sure, I will try to find the time to get the changelog updated as soon as I can.

JinShil:
Disabled auto-merge so I can manually squash commits when all tests pass.

src/object.d
```diff
@@ -3127,6 +3205,164 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
...
+ * Looks up key; if it exists applies the update delegate else evaluates the
+ * create delegate and adds it to the associative array
+ * Params:
+ *      aa =     The associative array.
+ *      key =    The key.
+ *      create = The delegate to apply on create.
+ *      update = The delegate to apply on update.
+ */
+void update(K, V, C, U)(ref V[K] aa, K key, C create, U update)
+if ((is(C : V delegate()) || is(C : V function())) && (is(U : V delegate(ref V)) || is(U : V function(ref V))))
```
Geod24: I'm not sure this is the right approach. It forbids `opCall` and restrict some conversion which would be legal under the typesystem.
Maybe use `is(typeof({ V* ptr; *ptr = create(); })` ?
Also shouldn't the arguments `create` and `update` be `scope` to avoid pointless allocation when passing delegate literals ?

src/object.d
```diff
@@ -3127,6 +3205,164 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
...
+    auto p = cast(V*) _aaGetX(cast(void**)&aa, typeid(V[K]), V.sizeof, &key, found);
+    if (!found)
+        *p = create();
+    else
+        *p = update(*p);
+}
+
+unittest
+{
+    static class C{}
```
Geod24: Nit: add a space

src/object.d
```diff
@@ -3127,6 +3205,164 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
...
+ * Looks up key; if it exists returns corresponding value else evaluates
+ * value, adds it to the associative array and returns it.
+ * Params:
+ *      aa =     The associative array.
+ *      key =    The key.
+ *      value =  The required value.
+ * Returns:
+ *      The value.
+ */
+ref V require(K, V)(ref V[K] aa, K key, lazy V value = V.init) pure
```
Geod24: I would remove the `pure` annotation and make the template works its magic, or is there a specific reason you put it ?

JinShil:
I'll refrain from merging this until @Geod24's comments are addressed.

src/object.d
GilesBathgate: @Geod24 I Just copied the function attributes from the `get` function.

src/object.d
GilesBathgate: @Geod24 Between `C` and `{}`?

Geod24: Yes

src/object.d
GilesBathgate: @Geod24 Just so I am clear this is the 'magic' you are referring to: http://klickverbot.at/blog/2012/05/purity-in-d/#templates-and-purity i.e templates infer pure?
Presumably this wasn't the case when 'get' was implemented.

Geod24: Template infers attributes, so `@nogc`, `@safe`/`@system`, and `pure`
In this case since only `pure` and `nothrow` are put on the `extern (C)` declaration, that's as much as could get infered. Maybe there were some inference bug at the time ?

You can make sure this works by making a `pure unittest`

JinShil:
@GilesBathgate Nice Work! And, thank you for sticking with it.

@Geod24 Could you please give this one last look.  I'm itching to merge this.

JinShil:
Disabled auto-merge so I can manually squash commits when all tests pass.

CyberShadow:
I'm proposing a small modification (addition) to the API:
https://github.com/dlang/druntime/pull/3012

Thank you @GilesBathgate for originally adding these, I got a lot of use out of them! :)

PR #2163, opened 2018-04-15
Add missing declarations to Mach-O interface

jacob-carlborg:
Add missing declarations to the `core.sys.darwin.mach.getsect` and `loader` modules.

### DIFF
--- a/src/core/sys/darwin/mach/getsect.d
+++ b/src/core/sys/darwin/mach/getsect.d
@@ -1,18 +1,347 @@
 /**
- * Copyright: Copyright Digital Mars 2010.
+ * D header file for $(LINK2 https://opensource.apple.com/source/cctools/cctools-895/include/mach-o/getsect.h.auto.html, mach-o/getsect.h).
+ *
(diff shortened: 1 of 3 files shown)


src/core/sys/darwin/mach/getsect.d
```diff
@@ -22,12 +351,121 @@ else version (WatchOS)
...
+    c_ulong* size
+);
+
+const(section)* getsectbynamefromheader(
+    in mach_header* mhp,
+    in char* segname,
+    in char* sectname
+);
+
+const(section)* getsectbynamefromheaderwithswap_64(
```
kinke: @jacob-carlborg: I take it the `_64` suffix is a typo? LDC doesn't like `extern(C)` overloads.

jacob-carlborg: Yes, that's a typo.

jacob-carlborg: https://github.com/dlang/druntime/pull/2229

PR #2164, opened 2018-04-16
Improve performance of `GC.stats`

mihails-strasuns-sociomantic:
Turns `GC.stats()` from O(n) to O(1) call, calculating used and free memory on the fly as part of regular GC operation. That makes a very big performance difference for `GC.stats()` usage in apps that allocate lot of GC memory.

Right now it I only confirmed it to produce same stats as previous implementation in synthetic examples. Going to test on some real-world projects in a day or two. For now general feedback from those who know about GC internals would be appreciated.

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -42,6 +42,7 @@ import gc.bits;
 import gc.os;
 import gc.config;
 import gc.gcinterface;
+import gc.stats;
 
 import rt.util.container.treap;
 
@@ -109,6 +110,8 @@ __gshared Duration recoverTime;
 __gshared Duration maxPauseTime;
 __gshared size_t numCollections;
 __gshared size_t maxPoolMemory;
+
+__gshared GCStatsTracker statsTracker;
 
 __gshared long numMallocs;
 __gshared long numFrees;
@@ -596,7 +599,8 @@ class ConservativeGC : GC
     {
         if (!size)
         {   if (p)
-            {   freeNoSync(p);
+            {
+                freeNoSync(p);
                 p = null;
             }
             alloc_size = 0;
@@ -693,6 +697,8 @@ class ConservativeGC : GC
                         pool.setBits(biti, bits);
                     }
                     alloc_size = newsz * PAGESIZE;
+                    statsTracker.freed(psize);
+                    statsTracker.allocated(alloc_size);
                     return p;
                 }
 
@@ -721,6 +727,9 @@ class ConservativeGC : GC
                     //debug(PRINTF) printf("\tcopying %d bytes\n",size);
                     memcpy(p2, p, size);
                     p = p2;
+
+                    statsTracker.freed(psize);
+                    statsTracker.allocated(size);
                 }
                 else
                     alloc_size = psize;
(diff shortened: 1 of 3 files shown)


src/gc/stats.d
```diff
@@ -0,0 +1,54 @@
+/**
+ * Utility to simplify calculating `core.memory.GC.Stats`
+ *
+ * Copyright: Copyright Digital Mars 2018.
```
WalterBright: Should be copyright the D Language Foundation

src/gc/stats.d
```diff
@@ -0,0 +1,54 @@
...
+ * License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ */
```
WalterBright: Please add:
```
Source: link to github source code
```

src/gc/stats.d
```diff
@@ -0,0 +1,54 @@
...
+    to maniplate it while doing some sanity checks.
+ */
+package(gc) struct GCStatsTracker
+{
+    @nogc nothrow:
+
+    import core.memory;
+    GC.Stats stats;
+
+    /// Record new memory added to the GC pool from OS
```
WalterBright: The style for Ddoc functions is:
```
/*********
 * Record new memory added to the GC pool from the OS.
 * Params:
 *   bytes = amount of new memory added
 */
```
Yes, it looks rather redundant, but tools expect this format.

src/gc/stats.d
```diff
@@ -0,0 +1,54 @@
...
+    void added(size_t bytes)
```
WalterBright: looks like @safe and pure can be added, too

src/gc/impl/conservative/gc.d
```diff
@@ -1673,8 +1654,12 @@ struct Gcx
 
     void* alloc(size_t size, ref size_t alloc_size, uint bits) nothrow
     {
-        return size <= 2048 ? smallAlloc(binTable[size], alloc_size, bits)
+        auto p = size <= 2048 ? smallAlloc(binTable[size], alloc_size, bits)
                             : bigAlloc(size, alloc_size, bits);
+
+        statsTracker.allocated(alloc_size);
```
WalterBright: Since this function is performance critical, being about to turn this statement into a no-op is important.

mihails-strasuns-sociomantic: What do you mean, `pragma(inline)`?

WalterBright: By no-op I mean no code is generated for the statement.

mihails-strasuns-sociomantic: I am afraid it is not possible - `GC.stats` is supposed to work for every program at any moment of time, it is not controlled by version or runtime flag.

Though considering how extremely expensive call to `alloc` is, I am not sure I understand concern about 2 extra integer additions either.

jacob-carlborg:
@mihails-strasuns-sociomantic can we close this while figuring out a better solution?

mihails-strasuns-sociomantic:
Yes, I was only keeping it as a reference for internal discussion.

For the record, I consider two orthogonal approaches to address this problem:

- Modify `GC.stats` interface to allow caller to request specific stats to calculate (to fix bloating of GC API with new methods when someone wants to add new stats but doesn't need rest)
- Refactor GC implementation to make all pool management happen through fixed API so that some form of this PR becomes viable again (right now it is too fragile against any GC changes to consider seriously).

PR #2170, opened 2018-04-29
Use regular import for dstrcmp

thaven:
An import of `dstrcmp` already happens on [line 3680](https://github.com/thaven/druntime/blob/93269b4d820b55495b4cf7a99d63b797298e2a48/src/object.d#L3680), though that one is inside a template, and this one isn't.

Usage of `externDFunc` in this case looks like a leftover of when it was `rt.util.string.dstrcmp`.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -219,7 +219,6 @@ class TypeInfo
 
     override int opCmp(Object o)
     {
-        import core.internal.traits : externDFunc;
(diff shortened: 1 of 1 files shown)


WalterBright:
The idea was to avoid as much as possible having object.d cause the compiler to read and process lots of other files, which slows down compilation of everything. Please revert this.

PR #2172, opened 2018-05-03
Fix Issue 18068 - No file names and line numbers in stack trace

wilzbach:
Turns out all that changed was PIE and the addresses with ASLR.

Example:

```d
import std.exception;
void main() {
    enforce(0);
}
```

So we finally get our line number back in the stack trace on Linux:

```
> dmddev -g hello.d && ./hello
/home/seb/dlang/dmd/generated/linux/release/64/../../../../../phobos/std/exception.d:515 pure @safe void std.exception.bailOut!(Exception).bailOut(immutable(char)[], ulong, const(char[])) [0x555869fe]
/home/seb/dlang/dmd/generated/linux/release/64/../../../../../phobos/std/exception.d:436 pure @safe int std.exception.enforce!().enforce!(int).enforce(int, lazy const(char)[], immutable(char)[], ulong) [0x55586979]
hello.d:3 _Dmain [0x555868f7]
```

CC @JinShil @ZombineDev

This is a PoC because I'm not sure on the best way to get the base offset of the binary.

### DIFF
--- a/src/rt/backtrace/elf.d
+++ b/src/rt/backtrace/elf.d
@@ -58,6 +58,62 @@ struct Image
 
         return null;
     }
+
+    @property size_t baseAddress()
+    {
+        version(linux)
+        {
+            import core.sys.linux.link;
+            import core.sys.linux.elf;
+        }
+        else version(FreeBSD)
+        {
+            import core.sys.freebsd.sys.link_elf;
+            import core.sys.freebsd.sys.elf;
+        }
+        else version(DragonFlyBSD)
+        {
+            import core.sys.dragonflybsd.sys.link_elf;
+            import core.sys.dragonflybsd.sys.elf;
+        }
+
+        static struct ElfAddress
+        {
+            size_t begin;
+            bool set;
+        }
+        ElfAddress elfAddress;
+
+        // the DWARF addresses for DSOs are relative
+        const isDynamicSharedObject = (file.ehdr.e_type == ET_DYN);
+        if (!isDynamicSharedObject)
+            return 0;
+
+        extern(C) int dl_iterate_phdr_cb_ngc_tracehandler(dl_phdr_info* info, size_t, void* elfObj) @nogc
+        {
+            auto obj = cast(ElfAddress*) elfObj;
+            // only take the first address as this will be the main binary
+            if (obj.set)
+                return 0;
+
+            obj.set = true;
+            // search for the executable code segment
+            foreach (const ref phdr; info.dlpi_phdr[0 .. info.dlpi_phnum])
+            {
+                if (phdr.p_type == PT_LOAD && phdr.p_flags & PF_X)
+                {
+                    obj.begin = info.dlpi_addr + phdr.p_vaddr;
+                    //obj.length = phdr.p_memsz;
+                }
+            }
+            // temporarily fall back to this
+            obj.begin = info.dlpi_addr;
+            //obj.length = info.dlpi_phdr.p_memsz;
+            return 0;
+        }
+        dl_iterate_phdr(&dl_iterate_phdr_cb_ngc_tracehandler, &elfAddress);
+        return elfAddress.begin;
+    }
 }
 
 private:
--- a/src/rt/backtrace/dwarf.d
+++ b/src/rt/backtrace/dwarf.d
@@ -72,7 +72,7 @@ int traceHandlerOpApplyImpl(const void*[] callstack, scope int delegate(ref size
             foreach(size_t i; 0 .. callstack.length)
                 locations[i].address = cast(size_t) callstack[i];
 
-            resolveAddresses(debugLineSectionData, locations[]);
+            resolveAddresses(debugLineSectionData, locations[], image.baseAddress);
         }
     }
 
@@ -108,7 +108,7 @@ int traceHandlerOpApplyImpl(const void*[] callstack, scope int delegate(ref size
 private:
 
 // the lifetime of the Location data is the lifetime of the mmapped ElfSection
-void resolveAddresses(const(ubyte)[] debugLineSectionData, Location[] locations) @nogc nothrow
+void resolveAddresses(const(ubyte)[] debugLineSectionData, Location[] locations, size_t baseAddress) @nogc nothrow
 {
     debug(DwarfDebugMachine) import core.stdc.stdio;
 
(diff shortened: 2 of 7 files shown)


JinShil:
Nice work! I think you can use https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/baselib-dladdr-3.html

It's already bound in the druntime at https://github.com/dlang/druntime/blob/05944a29e2dd1d839c9a919892e2e2c14519c398/src/core/sys/linux/dlfcn.d#L293

wilzbach:
> The behavior of dladdr() is only specified in dynamically linked programs.

However, using the first result from [`dl_iterate_phdr`](https://linux.die.net/man/3/dl_iterate_phdr) seems to work too.

JinShil:
Tested locally.  DMD test17559 finally passes.  In fact for the first time, the entire DMD test suite passes when run locally on my PC.  Great!

JinShil:
Did a little more testing locally.  This is working great.

kinke:
I had a go at this too as one of the fixes in https://github.com/dlang/druntime/pull/2151. I was unsure about how to obtain the base address too and came up with cumbersome parsing of `/proc/self/maps`. Other differences:
* As only the executable's DWARF lineinfo data is used, I check the ELF header whether it's a DSO, and only then use its base address.
* I add the base address to the relative DWARF address, making sure the virtual address is unique, which may not be the case in this version here, where the base offset is subtracted from each frame address (incl. frames in other DSOs).

wilzbach:
> I was unsure about how to obtain the base address too 

FYI @MartinNowak and I looked at the implementation in FreeBSD and Linux and `dl_iterate_phdr` will always iterate over the main binary for the first iteration.

FreeBSD:

https://github.com/freebsd/freebsd/blob/d9ac9c210e73d5141195a5497d3e96dfc755d1dc/libexec/rtld-elf/rtld.c#L572
https://github.com/freebsd/freebsd/blob/d9ac9c210e73d5141195a5497d3e96dfc755d1dc/libexec/rtld-elf/rtld.c#L637
https://github.com/freebsd/freebsd/blob/d9ac9c210e73d5141195a5497d3e96dfc755d1dc/libexec/rtld-elf/rtld.c#L3723

Linux:

https://github.com/bminor/glibc/blob/15c19147a8361b7281519b9784b3b823c447cf7a/elf/dl-iteratephdr.c#L32
https://github.com/bminor/glibc/blob/master/elf/rtld.c#L1114

(`LM_ID_BASE` is always 0 - see https://github.com/bminor/glibc/blob/09533208febe923479261a27b7691abef297d604/dlfcn/dlfcn.h#L47)

---------------

> I had a go at this too as one of the fixes in #2151 

That's a shame to have this work duplicated :/

> As only the executable's DWARF lineinfo data is used, I check the ELF header whether it's a DSO, and only then use its base address.

@MartinNowak recommended to only to use the executable code segment (and not the entire DSO), but apparently that currently fails on the CIs. I'm not sure that's actually necessary. Can the binary segment order really be dynamically changed by the loader?

kinke:
> FYI MartinNowak and I looked at the implementation in FreeBSD and Linux and dl_iterate_phdr will always iterate over the main binary for the first iteration.

Good, my main point was that the fix shouldn't be Linux-only.

> That's a shame to have this work duplicated

I expected the bot to add something to the dlang issue. - Glad the maps-file parsing can be avoided, so it did pay off. :)

> @MartinNowak recommended to only to use the executable code segment (and not the entire DSO), but apparently that currently fails on the CIs. I'm not sure that's actually necessary. Can the binary segment order really be dynamically changed by the loader?

No idea, but I'd have considered it unnecessary as well. My concern was that offsetting the addresses for non-relocatable executables may likely be wrong, but I see that you added the check now. - CI fails because of missing `.sys` in BSD imports.

I'd suggest applying the offset to the DWARF addresses in `resolveAddresses()` as I do. You don't have t to touch the frame addresses this way, which are later used in the stack trace string, so you don't end up with invalid relative addresses for frames outside the executable & keep all frame addresses absolute.
You'll also have to implement (dummy) `baseAddress()` for new `rt.backtrace.macho`. And I recommend returning a void* or size_t instead of the platform-dependent `ElfBaseAddress` struct.

wilzbach:
> No idea, but I'd have considered it unnecessary as well. My concern was that offsetting the addresses for non-relocatable executables may likely be wrong, but I see that you added the check now.

I actually removed the check for now as it didn't seem to make a difference.

BTW I also added a test for shared libraries and it seems like debug symbols from shared libraries don't work at all (and have never worked). I still added the test, s.t. it's easier to fix this in a follow-up.

src/rt/backtrace/elf.d
```diff
@@ -58,6 +58,62 @@ struct Image
 
         return null;
     }
+
+    @property size_t baseAddress()
```
JinShil: This is an awful lot of logic for a property, and would be run needlessly if called multiple times by a caller that isn't aware of the implementation.  The baseAddress shouldn't change during the lifetime of an `Image`, so wouldn't it be better to put this in the constructor, and store the result as a field?

kinke:
> BTW I also added a test for shared libraries and it seems like debug symbols from shared libraries don't work at all (and have never worked).

I see 2 remaining problems with the current backtrace code:
* As mentioned above, only the executable's DWARF lineinfo section is scanned, so no file/line infos for frames in other DSOs.
* The function names are derived from the dynamic symbols table (via `backtrace_symbols()`), requiring the ld flag `-export-dynamic` (added implicitly by DMD, not by LDC), which may significantly increase the binary file size. I guess DWARF debuginfos could be used here as well (if available via `-g`, falling back to `backtrace_symbols()` for release binaries).

jacob-carlborg:
What's the status of this?

wilzbach:
Sorry. I haven't had much time for this lately, but I will get to this once I am back from my hiking trip (~end of next week).
IIRC not much was missing here, shared libraries have never been supported for getting the debug info.

jacob-carlborg:
> Sorry. I haven't had much time for this lately, but I will get to this once I am back from my hiking trip (~end of next week).

Why are you replying here during your hike :smiley:.

wilzbach:
(Because it takes about 12h to get there and luckily there's free roaming in Europe nowadays :) )

JinShil:
Ping!  Is this ready to go?  It'd be awesome to get this in the upcoming release.

wilzbach:
Superseded by https://github.com/dlang/druntime/pull/2230

PR #2175, opened 2018-05-06
nullptr_t not defined.

TurkeyMan:
Added alias for `nullptr_t`, which is defined in `stddef.h`.

Note: DMD still doesn't mangle this properly with `extern(C++)`!

### DIFF
--- a/src/core/stdc/stddef.d
+++ b/src/core/stdc/stddef.d
@@ -19,6 +19,9 @@ extern (C):
 nothrow:
 @nogc:
 
+///
+alias nullptr_t = typeof(null);
+
 // size_t and ptrdiff_t are defined in the object module.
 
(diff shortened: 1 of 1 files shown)


TurkeyMan:
Surely this is uncontroversial...?

jmdavis:
> Surely this is uncontroversial...?

If that were why it hadn't been merged, there would be comments about it. Rather, I think that it's pretty clear that it's simply the case that it hasn't been looked at yet, because it's so new. It looks trivial enough though that I don't see any reason for me to not just merge it.

WalterBright:
> because

... some of us are still traveling! I'm at Code Europe in Poland.

TurkeyMan:
Poland roxors!

PR #2177, opened 2018-05-07
Add 'lib' Makefile target

wilzbach:
Analogous to Phobos: https://github.com/dlang/phobos/blob/6c11b77dbc07a2de293cb9f9c14a55890773331d/posix.mak#L307

BTW I wonder whether we should include the [DRuntime development page](https://wiki.dlang.org/DRuntime_development) that I wrote just within this repo (e.g. README.md or CONTRIBUTING.md). Thoughts?


JinShil:
The DAutoTest failure is the same as that mentioned at https://github.com/dlang/dmd/pull/8241#issuecomment-390069007  A new push may re-initiate a build and cause it to pass.

PR #2178, opened 2018-05-09
Fix Issue 18828 - [-betterC] helpless error in object.d

JinShil:
I don't really like this fix, but I'm not sure how else to do it, and maybe this PR will elicit some ideas from others.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -241,7 +241,7 @@ UT_MODULES:=$(patsubst src/%.d,$(ROOT)/unittest/%,$(SRCS))
 HAS_ADDITIONAL_TESTS:=$(shell test -d test && echo 1)
 ifeq ($(HAS_ADDITIONAL_TESTS),1)
 	ADDITIONAL_TESTS:=test/init_fini test/exceptions test/coverage test/profile test/cycles test/allocations test/typeinfo \
-	    test/thread test/unittest test/imports
+	    test/thread test/unittest test/imports test/betterc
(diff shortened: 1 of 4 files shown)


JinShil:
I'm not sure how to test -betterC in the runtime, so, instead, I added a test at https://github.com/dlang/dmd/pull/8235

jacob-carlborg:
Is this testable?

JinShil:
> Is this testable? 

Yes, at https://github.com/dlang/dmd/pull/8235.  I didn't know how else to do it.

jacob-carlborg:
Not sure about how much work it would be, but what about adding a new directory here: https://github.com/dlang/druntime/tree/master/test ?

JinShil:
I added the directory and test to the best of my ability, but I don't even understand what most of the syntax in the Makefile is for and I'm not even sure if the test is running.

RazvanN7:
Can we close : https://github.com/dlang/dmd/pull/8235 now ?

schveiguy:
Haha, I just went into bugzilla to make a new issue about the inefficiency of destroy, to find that the code I *swore* was using `typeid` is no longer there. And lo and behold, it's changed, but for another reason.

I'm still going to file an issue, because I feel like smaller structs can be done much more efficiently than casting to ubyte arrays. But this change helps significantly.

See: https://forum.dlang.org/post/pe3pcp$13t4$1@digitalmars.com

schveiguy:
> I'm still going to file an issue

See https://issues.dlang.org/show_bug.cgi?id=18899

PR #2179, opened 2018-05-10
BetterC: use templates in core.stdc.stdio instead of functions

WalterBright:
Using a template means that druntime will not need to be linked in if these functions are used.

### DIFF
--- a/src/core/stdc/stdio.d
+++ b/src/core/stdc/stdio.d
@@ -1,5 +1,5 @@
 /**
- * D header file for C99.
+ * D header file for C99 <stdio.h>
  *
  * $(C_HEADER_DESCRIPTION pubs.opengroup.org/onlinepubs/009695399/basedefs/_stdio.h.html, _stdio.h)
  *
@@ -9,7 +9,7 @@
  *    (See accompanying file LICENSE)
  * Authors:   Sean Kelly,
  *            Alex Rønne Petersen
- * Source:    $(DRUNTIMESRC core/stdc/_stdio.d)
+ * Source:    https://github.com/dlang/druntime/blob/master/src/core/stdc/stdio.d
  * Standards: ISO/IEC 9899:1999 (E)
  */
 
@@ -996,9 +996,9 @@ else version( NetBSD )
     }
 
     private extern __gshared FILE[3] __sF;
-    @property auto __stdin() { return &__sF[0]; }
-    @property auto __stdout() { return &__sF[1]; }
-    @property auto __stderr() { return &__sF[2]; }
+    @property auto __stdin()() { return &__sF[0]; }
+    @property auto __stdout()() { return &__sF[1]; }
+    @property auto __stderr()() { return &__sF[2]; }
     ///
     alias __stdin stdin;
     ///
@@ -1284,13 +1284,13 @@ int   puts(scope const char* s);
 extern (D) @trusted
 {
     ///
-    int getchar()                 { return getc(stdin);     }
+    int getchar()()                 { return getc(stdin);     }
     ///
-    int putchar(int c)            { return putc(c,stdout);  }
+    int putchar()(int c)            { return putc(c,stdout);  }
     ///
-    int getc(FILE* stream)        { return fgetc(stream);   }
+    int getc()(FILE* stream)        { return fgetc(stream);   }
     ///
-    int putc(int c, FILE* stream) { return fputc(c,stream); }
+    int putc()(int c, FILE* stream) { return fputc(c,stream); }
 }
 
 ///
@@ -1320,14 +1320,14 @@ version( MinGW )
   // No unsafe pointer manipulation.
   extern (D) @trusted
   {
-      ///
-    void rewind(FILE* stream)   { fseek(stream,0L,SEEK_SET); stream._flag = stream._flag & ~_IOERR; }
     ///
-    pure void clearerr(FILE* stream) { stream._flag = stream._flag & ~(_IOERR|_IOEOF);                 }
+    void rewind()(FILE* stream)   { fseek(stream,0L,SEEK_SET); stream._flag = stream._flag & ~_IOERR; }
     ///
-    pure int  feof(FILE* stream)     { return stream._flag&_IOEOF;                       }
+    pure void clearerr()(FILE* stream) { stream._flag = stream._flag & ~(_IOERR|_IOEOF); }
     ///
-    pure int  ferror(FILE* stream)   { return stream._flag&_IOERR;                       }
+    pure int  feof()(FILE* stream)     { return stream._flag&_IOEOF; }
+    ///
+    pure int  ferror()(FILE* stream)   { return stream._flag&_IOERR; }
   }
   ///
     int   __mingw_snprintf(scope char* s, size_t n, scope const char* fmt, ...);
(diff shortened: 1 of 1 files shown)


ibuclaw:
Not related to betterC. This is just common sense.

Can you remove stdio from the makefile so that the module it isn't compiled in to druntime?

WalterBright:
> Not related to betterC.

It is, as these functions were not usable with BetterC.

JinShil:
I'm not sure how to test betterC code from within the D runtime, but perhaps you could add some tests to compilable/betterc.d and runnable/betterc.d in the DMD repository.

schveiguy:
It's unfortunate we can't force inlining here, as is technically done with the C preprocessor.

WalterBright:
> I guess we aren't too worried about code that takes function pointers to these (which should be rare)?

On the C side they were macros, so never had function pointers to them. On the D side, you can still take the address of a template function. So no, not worried about it :-)

> I guess we aren't too worried about code that takes function pointers to these (which should be rare)?

D has `pragma(inline, true)` available for such cases.

> add some tests

I don't much value to be added by doing that. Should we have a test case for every one of those enum values in all those files, too? All I did was add `()` to them, I didn't change any of the function bodies or any semantics.

schveiguy:
> D has pragma(inline, true) available for such cases.

I mean "always inline this function", even when -inline isn't specified. This would be equivalent to the C macro. The template solution is the next best thing, but means that it's still going to needlessly generate functions for these one-liners.

WalterBright:
> it's still going to needlessly generate functions for these one-liners.

Functions go into their own sections, and the linker is supposed to remove unreferenced sections.

JinShil:
> I don't much value to be added by doing that.

The value is to ensure changes in the future don't break working code.  You really only need to ensure a program compiles and links, so it should be a simple test to create.  Whether or not to do it is up to you; it was just a suggestion.

JinShil:
The DAutoTest failure is a spurious one that I've seen before.  I believe a re-push will initiate a new build and cause it to pass.  I don't yet know what the root cause is as I don't have enough information to reproduce it.

JinShil:
I know this ship has probably sailed, but this is another reason why the C standard library and C++ standard library bindings need to be in their [own library in Deimos](https://github.com/D-Programming-Deimos/libc).  Then the -betterC folks can get what they need from there without having to worry about whether DRuntime is linked or not, and we don't have to do these kinds of workarounds.  At the moment, I don't approve. :angry:

JinShil:
> Druntime depends on libc.

This is an artificial dependency.  The fact that druntime uses libc is matter of convenience, not one of necessity.  All calls made to libc could be replaced by pure D code.  Furthermore, druntime has many "ports" (see https://issues.dlang.org/show_bug.cgi?id=11666 for what I mean).  If one or more of those ports chooses to depend on libc, then they should do so privately, and not leak that implementation as a dependency for all others to bear.

Consider those porting druntime to a new bare-metal platform.  Especially in that domain there is absolutely no reason to depend on libc.

jacob-carlborg:
It's quite a lot of work to reimplement everything that's needed from libc and other system libraries.

JinShil:
> It's quite a lot of work to reimplement everything that's needed from libc and other system libraries.

That's well understood.  One can still import libc privately in the platform's port if they want the convenience.

schveiguy:
> Furthermore, druntime has many "ports" (see https://issues.dlang.org/show_bug.cgi?id=11666 for what I mean)

I perused through that bug report and the PRs. I wish I had been involved with that, because when I read @WalterBright's comment [here](https://github.com/dlang/druntime/pull/732#issuecomment-36531064), I thought "yep, that's exactly what it should be".

I think a project to have a framework that implements that idea makes a ton of sense. Note that druntime doesn't have to do it, it can be separate. There is no requirement to depend on druntime for betterC. If it gets to the point that it is better implemented (very likely), then we can consider making druntime depend on it.

I think a hierarchy of phobos->druntime->cbindings would be fine.

> This is an artificial dependency. The fact that druntime uses libc is matter of convenience, not one of necessity.

Well, it's not just a matter of convenience -- much of the OS interface is part of libc, and most of the ecosystem depends on libc. If you need malloc/free for example. We can't do everything our own way, it would cut out the possibility of using D with other libraries.

But your own bare-bones port of Druntime, I don't see a problem carving out libc dependencies from that if you can do it.

FWIW, Tango did not depend on the "C" part of libc. It only used the OS bindings.

JinShil:
> much of the OS interface is part of libc

That's certainly true for the *nix OSs, but I think less so for Windows;  If we want D to have features currently implemented in libc or through the OS's APIs, we should expose those features through a well-engineered D API, not through the legacy API of libc.  For example D users should be using `byte[]` instead of `void*` and `string` instead of `char*.  And the D API should have D guarantees like `pure` , @trusted`, and `@safe`.  It may resemble libc and the OS APIs, or it may not.  The D API could use libc and the OS APIs internally, but that's an implementation detail.  Basically, by exposing the libc and OS APIs publicly, we're egregiously violating encapsulation, and not providing a proper platform abstraction.

> I think a hierarchy of phobos->druntime->cbindings would be fine.

Exactly what I'm proposing (assuming you mean Phobos wraps druntime and druntime wraps libc).

> Tango did not depend on the "C" part of libc. It only used the OS bindings.

Sound about right, but even the OS bindings should not be publicly exposed except through a well-designed D API, as I described above.

If a user wanted to write code to utilize a specific platform feature not provided by the above-mentioned D API, they could do so by importing libc or OS bindings from Deimos or DUB. It'd be no different than how D users currently interface with DirectX, XWindows, etc.

JinShil:
> But I want to stress that I don't think we want druntime depending on projects outside the dlang set of repositories, i.e. deimos.

Actually, the druntime ports don't need all of libc; just a small subset.  Some libc functions could even be replaced with D code.  When all that's done you'd probably only have a few libc functions remaining, so those bindings could be added locally, privately, and internally in the platform's druntime port.  I would *much* prefer that approach over this PR.

Users (-betterC or otherwise) that want direct bindings to libc should then look to Deimos or dub.  We'd then deprecate (but not remove) the `core.stdc` and `core.stdcpp` bindings and eventually wash our hands of maintaining them.  Documentation would let users know it has been deprecated and direct them elsewhere.  I think I'd actually prefer it be a dub project maintained by the users that need it over us having to maintain it in Deimos.  And all of this would apply equally to the bindings in `core.sys`, but one step at a time.  It'd make everything just that much simpler for the maintenance druntime.

JinShil:
> Unless I misunderstand, the discussion here is about future related work rather than this PR itself, and this PR would not make those proposals more difficult.

Yes, but those proposals would make this PR unnecessary.  This PR is a workaround, not a proper solution.  The proper solution would be to separate libc bindings from druntime so they can still be used in -betterC without encapsulating them in templates, and without sacrificing any legitimate use cases, like function pointers, no matter how rare they may be.  

`core.stdc` should be deprecated and a new, isolated `stdc` package should be created.  Druntime implementations that rely on libc bindings should import `stdc` privately.

n8sh:
>without sacrificing any legitimate use cases, like function pointers, no matter how rare they may be.

I thought from this discussion that templatization wouldn't interfere with function pointers. Not so?

n8sh:
>I thought from this discussion that templatization wouldn't interfere with function pointers. Not so?

Ah, to answer my own question, it is possible but requires alterations.

https://run.dlang.io/is/GCyPdU
```d
auto f1(string s) { return s.length; }
auto f2()(string s) { return s.length; }

void main()
{
    auto fp1 = &f1;
    //auto fp2 = &f2; // oops doesn't work
    auto fp2 = &(f2!()); // have to use different syntax
    static assert(is(typeof(fp1) == typeof(fp2)));
}
```

PR #2180, opened 2018-05-11
Make errno work with nothrow, @nogc, and @safe code.

jmdavis:
For whatever reason, errno wasn't marked with any of those attributes,
and I don't see any reason why it can't be, so this fixes that.

I haven't messed with the functions in the else block, because I don't
understand why it's even there instead of doing what we normally do with
static assert(false, ...) to indicate that the declaration is missing.

### DIFF
--- a/src/core/stdc/errno.d
+++ b/src/core/stdc/errno.d
@@ -31,7 +31,7 @@ version (CRuntime_DigitalMars)
 {
     extern (C)
     {
-        ref int _errno();
+        ref int _errno() @nogc nothrow @trusted;
         alias errno = _errno;
     }
 }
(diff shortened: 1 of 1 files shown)


schveiguy:
Are you sure this isn't already done at the top here?

https://github.com/jmdavis/druntime/blob/d1970d585fc8d1afe29e650cc74f026dd539fed0/src/core/stdc/errno.d#L26

I don't see why this wouldn't permeate to all the calls...

jmdavis:
> Are you sure this isn't already done at the top here?

> https://github.com/jmdavis/druntime/blob/d1970d585fc8d1afe29e650cc74f026dd539fed0/src/core/stdc/errno.d#L26

> I don't see why this wouldn't permeate to all the calls...

Oh gag, it does. I hate mass-applying of attributes like this. They're way too easy to miss. And then of course, because I never use them, I never remember that they're a thing. :|

These changes are indeed unnecessary. Sorry for the noise.

PR #2181, opened 2018-05-14
[WIP] Add object.__arrayLiteral template

JinShil:
@andralex @WalterBright et al.

This is my attempt at picking up where @somzzz left off, trying to replace runtime hooks that depend on runtime `TypeInfo` with templates.

For now `__arrayLiteral` is just a replacement for `_d_arrayLiteralTX` to verify the compiler/runtime interface is correct.  `__arrayLiteral` needs to be refactored to no longer rely on `TypeInfo`, but I want to get the compler/runtime interface implemented first and passing all tests.  Then I can refactor `__arrayLiteral` to no longer rely on runtime information.

I could use some direction on this.  I'm not sure about naming conventions, where these templates should reside in the source tree, etc...

This requires https://github.com/dlang/dmd/pull/8245

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -4461,3 +4461,21 @@ unittest
     scope S[1] arr = [S(&p)];
     auto a = arr.dup; // dup does escape
 }
+
+private extern (C) void* _d_arrayliteralTX(const TypeInfo ti, size_t length);
+public T[] __arrayLiteral(T)(size_t length)
+{
+    // FIXME: This function template needs refactoring so it only relies on information
+    // known at compile time.  i.e. It should not rely on runtime calls to `typeid`
+    // for `TypeInfo`.  It may be possible to completely eliminate `_d_arrayLiteralTX`
+    // and any compiler-generated calls to it.
+    T[] result;
+    struct Arr
+    {
+        size_t l;
+        void* p;
+    }
+    Arr* a = cast(Arr*)(&result);
+    a.p = _d_arrayliteralTX(typeid(T[]), length);
+    return result;
+}


UplinkCoder:
Please think about the impact of this change.

UplinkCoder:
Is there no other way to get the desired functionality?

schveiguy:
@UplinkCoder we need to find a way to make the library handle more of the responsibilities of the compiler, to reduce the dependence on "compiler magic", and to make it much easier to affect how the language operates (druntime is MUCH more approachable than the compiler).

Yes, this bloats the code some as it adds an extra helper template function. We *also* need to find a way to mitigate this. In other words, there needs to be a way to implement some of the compiler via lowering in the library in a way that *doesn't* impact the compiled code. In a perfect world, this results in the same compiled code as before.

There is also the impact on the compile time. I hope there's a way to mitigate this as well, but I don't have any clue how this can be done.

JinShil:
Please review https://www.youtube.com/watch?v=endKC3fDxqs and https://forum.dlang.org/post/mailman.385.1440156134.13986.digitalmars-d@puremagic.com for context.

The template bloat can be mitigated when `__arrayLiteral` is refactored; it can make a runtime call to an allocator after the information it needs is extracted from the type.

schveiguy:
I want to caution here -- there are zero instantiations of this template, so the fact that it "passes" simply means that it is valid syntax AFAIK. I'm not sure if it's even possible to unittest this, as the template will not deliver a completed array literal (just the memory for one).

I suppose at least you could `__traits(compiles, ...)` some various array types (including modifiers).

JinShil:
I screwed up the DMD PR creating a false positive.  The implementation is much too naive. I need to spend more time on it.  Don't waste your time looking at it.

PR #2182, opened 2018-05-15
Android: update core.stdc.math with real declarations

joakim-noah:
Most of the Android port was done against API 9, when Android only supported 32-bit platforms and defined `long double` to always be 64-bit, so they didn't bother with even providing aliases for several `long double` math functions. However, they started adding those with the following APIs, culminating in API 21, Lollipop, which also added 64-bit platforms with different native `long double` sizes.

This pull updates the math function declarations for Bionic and notes when they were added with Lollipop. Some functions were added in between API 9 and 21, though not for 64-bit till API 21 obviously, so those were left uncommented, since they were added at different APIs for different CPU arches.

This pull is especially needed for the dmd 2.080 frontend, which switches to using them for CTFE math, which is why I had to use [this patch to compile the latest ldc beta with the 2.080 frontend natively for Android/ARM](https://github.com/joakim-noah/termux-packages/blob/beta/packages/ldc-beta/ldc-druntime-math.patch.beforehostbuild).

### DIFF
--- a/src/core/stdc/math.d
+++ b/src/core/stdc/math.d
@@ -3578,118 +3578,131 @@ else version(DragonFlyBSD)
 }
 else version(CRuntime_Bionic)
 {
-    // Bionic defines long double as 64 bits, same as double, so several long
-    // double functions are missing.  nexttoward was modified to reflect this.
     ///
     double  acos(double x);
     ///
     float   acosf(float x);
-    //real    acosl(real x);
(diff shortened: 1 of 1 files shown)


joakim-noah:
ping, like to get this into the next ldc release.

joakim-noah:
Thanks, now to cherrypick into LDC master.

PR #2183, opened 2018-05-16
set core.stdc.stdint types so mangling matches C++

WalterBright:
Provides better interoperability with C++, hence betterC.

### DIFF
--- a/src/core/stdc/config.d
+++ b/src/core/stdc/config.d
@@ -15,17 +15,35 @@ module core.stdc.config;
 
 version (StdDdoc)
 {
+    private
+    {
+        version (Posix)
+            enum isPosix = true;
+        else
+            enum isPosix = false;
+        static if (isPosix && (void*).sizeof > int.sizeof)
+        {
+            alias ddoc_long = long;
+            alias ddoc_ulong = ulong;
+        }
+        else
+        {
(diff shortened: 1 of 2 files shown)


src/core/stdc/config.d
```diff
@@ -15,17 +15,35 @@ module core.stdc.config;
...
 {
+    private
+    {
+        version (Posix)
+            enum isPosix = true;
+        else
+            enum isPosix = false;
+        static if (isPosix && (void*).sizeof > int.sizeof)
+        {
+            alias ddoc_long = long;
```
JinShil: What the heck is a `ddoc_` type?

JinShil: I don't see why we need these itermediate `ddoc_` aliases here.  Can't we just declare `c_long` and `c_ulong` here.

JinShil:
There's an awful lot of redundancy in this patch and it makes review a little difficult.  Can't we just version out the differences?

WalterBright:
> There's an awful lot of redundancy in this patch and it makes review a little difficult. Can't we just version out the differences?

I tried that initially. The result was incomprehensible.

WalterBright:
I put you on the team. Druntime can use your help.

PR #2184, opened 2018-05-16
Fix Issue 18493 - [betterC] Can't use aggregated type with postblit

JinShil:
This PR is required for https://github.com/dlang/dmd/pull/8253.  Please read the motivation there for context.

~Despite the enabling features this provides for https://github.com/dlang/dmd/pull/8253, this PR is still quite useful and forward-looking.~

~Although -betterC does not link with druntime or phobos, it still imports object.d.  There is a large amount of code in object.d that not utilized when compiling with -betterC, but it is not `version`ed out like it should be, and that causes the compiler to emit errors for code that isn't even being utilized.  This is not only necessary for https://github.com/dlang/dmd/pull/8253, but also for other planned -betterC features like enabling the use of classes with only `static` members (See https://github.com/dlang/dmd/pull/8204 for an already merged PR enabling this for minimal runtime builds).~

~It is likely future PRs will need to be made to hoist certain templates out of the `version(Not_BetterC)` area up to the `version(D_BetterC`)` area as users utilize -betterC more and file issues identifying certain language features that don't work in a -betterC environment.  It will also be beneficial to acquire those issue reports to identify lapses in test coverage for -betterC.~

EDIT: #2194 took care of the excessive importing.  This PR is now only needed to support https://github.com/dlang/dmd/pull/8253.

### DIFF
--- a/src/core/stdc/stdlib.d
+++ b/src/core/stdc/stdlib.d
@@ -29,10 +29,14 @@ else version (WatchOS)
 extern (C):
 @system:
 
-/* Placed outside @nogc in order to not constrain what the callback does.
+/* Placed outside `nothrow` and `@nogc` in order to not constrain what the callback does.
  */
 ///
-alias int function(const void*, const void*) _compare_fp_t;
+alias _compare_fp_t = int function(const void*, const void*);
+
+nothrow:
+@nogc:
+
 ///
 inout(void)* bsearch(const void* key, inout(void)* base, size_t nmemb, size_t size, _compare_fp_t compar);
 ///
@@ -50,9 +54,6 @@ void    qsort(void* base, size_t nmemb, size_t size, _compare_fp_t compar);
     int key;
     bsearch(&key, arr.ptr, arr[0].sizeof, arr.length, &S.cmp);
 }
-
-nothrow:
-@nogc:
 
 ///
 struct div_t
--- a/src/rt/profilegc.d
+++ b/src/rt/profilegc.d
@@ -113,7 +113,7 @@ shared static ~this()
         Entry entry;
 
         // qsort() comparator to sort by count field
-        extern (C) static int qsort_cmp(scope const void *r1, scope const void *r2)
+        extern (C) static int qsort_cmp(scope const void *r1, scope const void *r2) @nogc nothrow
         {
             auto result1 = cast(Result*)r1;
             auto result2 = cast(Result*)r2;
--- a/src/rt/trace.d
+++ b/src/rt/trace.d
@@ -156,7 +156,7 @@ private void stack_free(Stack *s)
 //////////////////////////////////////
 // Qsort() comparison routine for array of pointers to SymPair's.
 
-private int sympair_cmp(scope const void* e1, scope const void* e2)
+private int sympair_cmp(scope const void* e1, scope const void* e2) nothrow @nogc
 {
     auto count1 = (*cast(SymPair**)e1).count;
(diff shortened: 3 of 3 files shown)


src/object.d
```diff
@@ -10,13 +10,14 @@
 
 module object;
 
-private
+version(D_BetterC)
 {
-    extern (C) Object _d_newclass(const TypeInfo_Class ci);
-    extern (C) void rt_finalize(void *data, bool det=true);
+    nothrow:
```
Geod24: This doesn't work (attributes in a `version` block don't apply to the outer scope).

JinShil: Thanks.  What's the proper idiom, then?

RazvanN7: I don think that there is one. I tried something along the lines of : 

```d
import std.stdio: writeln;

version(V1)
{
    enum versionl = true;
}
else
{
}

static if (versionl)
{
    notrow:
}

void f() 
{
    writeln("bla");
}
```

But it doesn't work since it looks like `nothrow:` cannot be placed inside a `static if` body. If static if doesn't work and `nothrow:` inside a version body does not work for the outer scope I see no other solution then to actually declare the code as a string and mixin it in each branch. I hope that someone has a better idea, cause the one I proposed is awful.

WalterBright: You could put the optionally nothrow code into a template mixin, and mixin the template on each leaf of the version statement.

JinShil: > You could put the optionally nothrow code into a template mixin, and mixin the template on each leaf of the version statement.

Oh, man, that would be great if it worked.  Please illustrate with an example, as I don't quite follow  You can see my prior attempts here:  https://forum.dlang.org/post/uhgzgmowqcczczrdtsol@forum.dlang.org

JinShil:
@WalterBright, I'm sorry for interrupting the workflow.  I got a little confused since you objected to the DMD PR, but approved this.  I see now that this PR has value regardless of the fate of the DMD PR.

I also implemented your elegant solution to the conditional attribute problem.  It turns out, however, it wasn't needed because the templated code has its attributes inferred by the caller.  Nevertheless, since this idiom was hard-won, I left the boiler-plate in place so, if we find code that needs this conditional attribution, we know where to put it.  I predict this code will need to be modified as we continue to implement -betterC.

Please give it another look.  Once this is merged, I should be able to address https://issues.dlang.org/show_bug.cgi?id=18905 with more predicatable results.

JinShil:
https://github.com/dlang/druntime/pull/2194 Should be merged first.  It has the same changes as this PR, but removes the controversy around the `nothrow` issue.  After that is merged, the fate of this PR can be decided based on the merits of the `nothrow` semantics of -betterC in isolation.

JinShil:
#2194 was merged, and the proposed fix for dlang/dmd#8253 is proving to be controversial.  So, the context and motivation for this has changed.  It has been re-submitted without the conflation of the -betterC versioning, and with a more relevant motivation as https://github.com/dlang/druntime/pull/2203.

JinShil:
#2194 took care of the excessive importing. This PR is now only needed to support dlang/dmd#8253.

rainers:
> I'm seem some strange undefined reference errors in the Circle CI tests. I think they're related to backend conversion to D. Any ideas?

You'll need to upgrade the D host compiler (or remove -betterC from the command line for dvec.d).

JinShil:
> You'll need to upgrade the D host compiler

Done.  https://github.com/dlang/druntime/pull/2205

JinShil:
Ping! Why hasn't this merged?

PR #2188, opened 2018-05-19
core.stdc.stdint: fix mistake with intmax_t on OSX

WalterBright:


WalterBright:
blocking https://github.com/dlang/dmd/pull/8252

PR #2189, opened 2018-05-20
Important to clean up aggregates too!

TurkeyMan:
Got tripped up by the `__xdtor` thing.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2930,7 +2930,7 @@ void destroy(T)(T obj) if (is(T == class))
 {
     static if(__traits(getLinkage, T) == "C++")
     {
-        obj.__dtor();
+        obj.__xdtor();
 
         enum classSize = __traits(classInstanceSize, T);
         (cast(void*)obj)[0 .. classSize] = typeid(T).initializer[];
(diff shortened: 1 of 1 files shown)


TurkeyMan:
This is an amendment on a patch that was already merged. Everything was in that prior PR.

TurkeyMan:
Also, I can't produce a working test case for aggregate destruction, because it's broken (a separate issue).

kinke:
AFAICT, we'd need yet something similar to `__xdtor` for `extern(C++)` classes, mangled as C++ dtor and calling what `__xdtor` would alias (calling its own extern(C++) dtor implemented in D and/or destructing its own fields in reversed order) but then also calling its base class' C++ dtor (cleaning up itself and its bases recursively this way).

Only if the dtor is implemented in D that is; if it's just forward declared and the implementation in C++, it forward-declares the external real C++ dtor, and there should be no need for any D destructors (__dtor, __fieldDtor, __aggrDtor, __xdtor).

TurkeyMan:
Yes, I wrote all those things on the forum.
But we don't have that yet... so for the time being, just pull this and fix the bug ;)

TurkeyMan:
I just pushed https://github.com/dlang/dmd/pull/8274 to fix the destructor bug, which means this is now required to do the right thing.

TurkeyMan:
dlang/dmd#8274 is merged now, so `destroy()` is now broken until this merges...

TurkeyMan:
Gah. Okay, I added the aggregate checks.

TurkeyMan:
If you want to bless the other one, then the tests can attempt to run ;)

TurkeyMan:
Do I need to do anything to rebuild against new DMD master?

TurkeyMan:
I notice that there is a button to trigger a circle-ci rebuild if you have permission...

TurkeyMan:
Huzzah!

PR #2190, opened 2018-05-21
opEquals was made auto so it appeared in the now defunct generated .di file.

schveiguy:
See https://github.com/dlang/druntime/commit/8ff270cf86f86c9848728238c9981173e0ddfa2f

This makes the ddoc generation return auto, but it should really specify the return type. There's also no reason to semantic the function every compilation to figure this out.


jacob-carlborg:
It seems like it was changed to `auto` for a reason. Don't we care about that now?  What about making it a dummy template, or is the compiler invoking this function somewhere?

rainers:
According to git-blame it was me who changed that. I don't remember what the exact reason was, but I suspect that was to keep the .d file in sync with the .di file which was manually edited back then. Now that we actually only copy object.d to the import folder, that should be unnecessary now.

Please note that there is a second `auto opEquals` function right below.

schveiguy:
> it was changed to auto for a reason

Yes. At some point in the past, object_.d and object.di were two separate files that were *manually* synchronized. In other words, whenever you added something to object_.d, you had to add it to object.di. In PR #1222, this was migrated to *generating* object.di from object_.d, which in case of this function, would REMOVE the function declaration completely (I think).

Fast forward to today, we have no object.di file, we have no reason to keep this as auto. Not a big deal, except the docs show it as returning auto, when it returns a bool (and should be listed this way).

schveiguy:
> Please note that there is a second auto opEquals function right below.

drat, didn't notice that. Will update.

schveiguy:
Updated, now got both of them.

rainers:
I still only see one.

jacob-carlborg:
Same here.

schveiguy:
Sorry, I'm learning new git techniques 😆. Should be fixed now.

PR #2191, opened 2018-05-23
[Trivial] OS X: Make core.stdc.stdint independent of D `long` C++-mangling

kinke:
C++ `int64_t` is defined as `long long` on both 32/64-bit OS X, so use the appropriate alias defined in `core.stdc.config`.

### DIFF
--- a/src/core/stdc/stdint.d
+++ b/src/core/stdc/stdint.d
@@ -111,8 +111,8 @@ else version (OSX)
     alias uint16_t = ushort; ///
     alias int32_t  = int;    ///
     alias uint32_t = uint;   ///
-    alias int64_t  = long;   ///
-    alias uint64_t = ulong;  ///
+    alias int64_t  = cpp_longlong;  ///
+    alias uint64_t = cpp_ulonglong; ///
 
     alias int_least8_t   = byte;   ///
     alias uint_least8_t  = ubyte;  ///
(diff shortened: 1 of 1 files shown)


schveiguy:
It looks reasonable. Is there a way to know that it's correct? There are no unittests for this kind of stuff it seems.

kinke:
There are DMD mangling interop tests (or at least will be); I'm about to create a DMD PR reverting to the previous mangling for `long` on 64-bit OS X; this is a prerequisite to get it working.

schveiguy:
I have a mac, is there something I can compile to ensure this is correct?

kinke:
For testing, you could compile
```D
import core.stdc.stdint;
extern(C++) void fooCpp(int64_t, uint64_t);
void main() { foo(123, 456); }
```
and check the function mangle. It should be `__Z6fooCpplm` according to https://godbolt.org/g/iaqA2y.

schveiguy:
I'm confused -- I wrote the following C++ file:
```C++
#include <stdint.h>

void fooCpp(int64_t, uint64_t)
{
}
```
And it mangles as `__Z6fooCppxy`

If I write your code (changing foo to fooCpp of course), it links fine against the C++ object.

I don't know what the plan is here, but it seems like we shouldn't be ultimately changing the mangling at all, no?

kinke:
So it's working (and this stuff is in LDC 1.10). ;) - This tiny PR here doesn't change the mangling, it just prepares for a D `long` mangled as C++ `long` again without breaking `core.stdc.stdint.int64_t` and friends. It's just a more robust way of defining them as C++ `long long` on Mac.

schveiguy:
Sorry, wish I could have helped better, but my understanding of the OSX mangling issues is not good...

PR #2193, opened 2018-05-25
Modify `GC.stats` to allow requesting specific stats

mihails-strasuns-sociomantic:
Relevant to https://github.com/dlang/druntime/pull/2155 and https://github.com/dlang/druntime/pull/1846 and potentially more future PRs. This should be a mostly backwards-compatible change
that enables adding more stats with less performance concerns.

Natural drawback is of course that it will only work until `Stats` has 64+ fields :) Alternative I have in mind is to define `Optional` type in druntime and use that for `Stats` fields - but that is quite more intrusive change.


mihails-strasuns-sociomantic:
FYI @leandro-lucarella-sociomantic

mihails-strasuns-sociomantic:
(in the meanwhile I am trying to implement alternative to resurrect https://github.com/dlang/druntime/pull/2164)

leandro-lucarella-sociomantic:
As I said in https://github.com/dlang/druntime/pull/1846#issuecomment-386563504, I think the right approach is to enable/disable stats calculation globally instead of doing it at query-time, as we should aim to completely remove the heavy calculations when using `gc_query()`. Ideally it should just copy some pre-calculated stats and that's it.

Unfortunately I didn't have the time yet to try to move forward with this...

src/core/memory.d
```diff
@@ -674,10 +680,16 @@ struct GC
...
+     *    fields = optional bit flag argument which specifies which stats need
+     *      to be calculated. By default equals to "all fields". If some field
+     *      was not requested via bit flag, its value in returned `Stats` struct
+     *      will be undefined.
      */
-    static Stats stats() nothrow
+    static Stats stats(ulong fields = ulong.max) nothrow
     {
-        return gc_stats();
+        return gc_stats(ulong.max);
```
n8sh: Should this be `gc_stats(fields)`?

src/core/memory.d
```diff
@@ -1168,6 +1180,53 @@ unittest
...
+ */
+private string generateFieldEnum(alias S)()
+{
+    import core.internal.string;
+
+    string code = "enum " ~ __traits(identifier, S) ~ "Fields\n{\n";
+    ulong shift = 0;
+    char[3] buf;
+
+    foreach (idx, _; S.init.tupleof)
```
thewilsonator: `static foreach`

PR #2196, opened 2018-05-30
Set the semaphore handle `protected` instead of `private`

BBasile:
For example now, if i want to create a `NamedSemaphore` class, i have to copy and paste most of the code while i would only need to write a custom `this()` after the change proposed in this PR.

### DIFF
--- a/src/core/sync/semaphore.d
+++ b/src/core/sync/semaphore.d
@@ -337,19 +337,17 @@ class Semaphore
     }
 
 
-private:
-    version( Windows )
-    {
-        HANDLE  m_hndl;
-    }
-    else version( Darwin )
-    {
-        semaphore_t m_hndl;
-    }
-    else version( Posix )
-    {
-        sem_t   m_hndl;
-    }
+protected:
+
+    /// Aliases the operating-system-specific semaphore type.
+    version(Windows)        alias Handle = HANDLE;
+    /// ditto
+    else version(Darwin)    alias Handle = semaphore_t;
+    /// ditto
+    else version(Posix)     alias Handle = sem_t;
+
+    /// Handle to the system-specific semaphore.
+    Handle m_hndl;
 }
 
 


jacob-carlborg:
@andralex ping, this changes the public API.

BBasile:
> Everything that is protected needs to have Ddoc documentation. Should also have a changelog entry.

Okay done. For the changelog i've simply created an issue. Most simple solution.

src/core/sync/semaphore.d
```diff
@@ -50,6 +50,13 @@ else
     static assert(false, "Platform not supported");
 }
 
+/// Aliases the operating system-specific semaphore type.
+version(Windows)        alias SemaphoreHandle =  HANDLE;
+/// ditto
+else version(Darwin)    alias SemaphoreHandle = semaphore_t;
+/// ditto
+else version(Posix)     alias SemaphoreHandle = sem_t;
```
Geod24: I'll move this `alias` within the scope of the class, instead of prefixing it with the class name.

BBasile: done. Initially i just changed and move this because it was better for ddoc.

jacob-carlborg:
> Okay done. For the changelog i've simply created an issue. Most simple solution.

Then you need to change the commit message to start with "Fix issue XXX" where "XXX" is the bugzilla issue number. That's how the bugzilla issue is connected to a PR.

BBasile:
> Then you need to change the commit message

Done too. You might not be used to be in contact w/ me but i'm not a first time contributor, so i know all these stuff, although it's true that i'm usually more making Phobos bug fixes.

jacob-carlborg:
It's difficult to keep track of which ones that know how the process work and which don't.

BBasile:
Can't the common sense prevail here ?

andralex:
All: I'm traveling with scant connectivity through the end of June. I'll approve this although it would be so much better if it were motivated by an example.

PR #2198, opened 2018-05-30
Fix Issue 18921 - make core.internal.hash cater to memberwise hash chaining

n8sh:
For use in memberwise hashing of structs (as in issue #18918) or in hashing arrays, speed up the hashing of individual scalars.

Before this PR, for a `struct` where each field is `uint`, memberwise chained hashing has a +250% speed penalty with LDC2 and a +350% speed penalty with DMD compared to treating the struct as an array of bytes. This seems fairly constant regardless of the size of the struct. After this patch chained memberwise hashing has no speed penalty with LDC2 and only a +60% speed penalty with DMD.


n8sh:
>generated/freebsd/debug/64/unittest/libdruntime-ut.so: undefined reference to >`_D4core3sys5posix5stdio12__ModuleInfoZ'
>cc: error: linker command failed with exit code 1 (use -v to see invocation)
>Error: linker exited with status 1

>Undefined symbols for architecture i386:
>  "__D4core3sys5posix5stdio12__ModuleInfoZ", referenced from:
>      __D2rt5cover12__ModuleInfoZ in test_runner.o
>ld: symbol(s) not found for architecture i386
>clang: error: linker command failed with exit code 1 (use -v to see invocation)

I assume these errors are unrelated to this PR. I tried resetting but the errors persist.

n8sh:
It looks like removing uses of `core.bitop.rol` caused the linker problems to vanish.

n8sh:
This also allows the compiler to infer that `hashOf` is `@nogc` for the scalars to which it applies (all except complex numbers and extended-precision `real`), and as a consequence also for raw pointers.

src/core/internal/hash.d
```diff
@@ -79,14 +79,62 @@ size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && __traits
...
+            h = (h << r1) | (h >>> (typeof(h).sizeof * 8 - r1));
+            h = (h * c2) ^ seed;
+            h = (h << r2) | (h >>> (typeof(h).sizeof * 8 - r2));
+            return h * 5 + c3;
+        }
+        else static if (T.sizeof > size_t.sizeof && __traits(isIntegral, T))
+        {
+            seed = hashOf(cast(size_t) val, seed);
+            static foreach (i; 1 .. T.sizeof / size_t.sizeof)
+                seed = hashOf(cast(size_t) (val >>> (size_t.sizeof * 8 * i)), seed);
```
wilzbach: Why not moving the first line into the loop and starting at `0`?
`val >>> 0` is legal.

src/core/internal/hash.d
```diff
@@ -79,14 +79,62 @@ size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && __traits
...
+                enum ulong c2 = 0x4cf5ad432745937f;
+                enum ulong c3 = 0x52dce729;
+                enum uint r1 = 31;
+                enum uint r2 = 27;
+            }
+            auto h = c1 * val;
+            h = (h << r1) | (h >>> (typeof(h).sizeof * 8 - r1));
+            h = (h * c2) ^ seed;
+            h = (h << r2) | (h >>> (typeof(h).sizeof * 8 - r2));
+            return h * 5 + c3;
```
wilzbach: @ other reviewers: a reference implementation can be found here https://github.com/aappleby/smhasher/blob/61a0530f28277f2e850bfc39600ce61d02b518de/src/MurmurHash3.cpp#L94

src/core/internal/hash.d
```diff
@@ -79,14 +79,62 @@ size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && __traits
...
+                enum ulong c1 = 0x87c37b91114253d5;
+                enum ulong c2 = 0x4cf5ad432745937f;
+                enum ulong c3 = 0x52dce729;
+                enum uint r1 = 31;
+                enum uint r2 = 27;
+            }
+            auto h = c1 * val;
+            h = (h << r1) | (h >>> (typeof(h).sizeof * 8 - r1));
+            h = (h * c2) ^ seed;
+            h = (h << r2) | (h >>> (typeof(h).sizeof * 8 - r2));
```
wilzbach: As rotl32 is used by `bytesHash`, it might have made sense to move rotl32 / rotl64 in separate functions.

n8sh:
Rebased.

src/core/internal/hash.d
```diff
@@ -84,12 +84,59 @@ size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && __traits
 {
     static if(__traits(isFloating, val))
     {
-        T data = (val != val) ? T.nan : val;
-        return bytesHash(toUbyte(data), seed);
+        static if (T.mant_dig == float.mant_dig && T.sizeof == uint.sizeof)
```
WalterBright: This is subtly altering the behavior. NaNs aren't just one bit pattern, they are a whole set of values. The former implementation would factor this out, and treat all NaN values identically. The revised implementation has them hashing to different values.

Different NaN values should have the same hash.

n8sh: The current `hashOf` doesn't regularize NaNs when hashing an array of floats or hashing a struct that has float members. Removing the special case here made the treatment of NaNs consistent.

>Different NaN values should have the same hash.

What is the motivation for ensuring NaNs have the same hash despite not being `==`? I am thinking here about keys in hash tables.

n8sh:
Added this bit to the top:

```d
// If true ensure that positive zero and negative zero have the same hash.
// typeid(float) does this but historically hashOf(float) did not.
private enum floatCoalesceZeroes = false;
// If true ensure that all NaNs of the same floating point type have the same hash.
// typeid(float) does not do this but historically hashOf(float) did.
private enum floatCoalesceNaNs = true;
private enum floatRequiresMemberwiseHash = floatCoalesceZeroes || floatCoalesceNaNs;
```

src/core/internal/hash.d
n8sh: Good idea.

n8sh:
```
// If either of the above are true, then no struct or array that contains the
// representation of a floating point number can be hashed with `bytesHash`.
private enum floatRequiresMemberwiseHash = floatCoalesceZeroes || floatCoalesceNaNs;
```

In the interest of keeping this PR easily digestible I've removed the code for this. Enforcing consistency in the treatment of floats can be a future PR. Now this PR is back to changing only a single function.

As @WalterBright's request that `hashOf(nan)` continue to work the same way has been met, are there any other obstacles to merging?

n8sh:
Rebased.

PR #2199, opened 2018-05-31
Fix Issue 18924 - Use addition rather than XOR for order-independent hash combination

n8sh:
(for associative arrays)

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -173,7 +173,7 @@ size_t hashOf(T)(auto ref T aa, size_t seed = 0) if (!is(T == enum) && __traits(
         size_t[2] hpair;
         hpair[0] = key.hashOf();
         hpair[1] = val.hashOf();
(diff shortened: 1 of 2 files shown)


JinShil:
Why? What's motivating this?

JinShil:
I don't want to blindly approve this.  Can you cite a technical whitepaper, or demonstrate an improvement with a benchmark?

n8sh:
Here's a "benchmark". Let's say we're hashing sets of elements and each element has the same hash, with the lowest bit set to 1 and all other bits 0. If you combine the hashes with XOR, you get a parity bit. Every set will be hashed to either 0 or 1, which is bad news if we hope to minimize hash collisions. If you combine the hashes with addition the result is the length of the set, the low bit of which is identical to the result of combining the hashes with XOR.

JinShil:
I'll defer to other reviewers who have more expertise in this area than I.

PR #2200, opened 2018-05-31
Fix Issue 18925 - core.internal.hash auto-hash for structs of scalar fields should be `@safe`

n8sh:
This patch enables:
```d
@safe unittest
{
    import core.internal.hash : hashOf;
    static struct S { int a; int b; }
    auto h = hashOf(S.init);
}
```

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -57,8 +57,7 @@ if (!is(T == enum) && !is(T : typeof(null)) && is(T S: S[]) && !__traits(isStati
     else static if (is(typeof(toUbyte(val)) == const(ubyte)[]))
     //ubyteble array (arithmetic types and structs without toHash) CTFE ready for arithmetic types and structs without reference fields
     {
-        auto bytes = toUbyte(val);
-        return (() @trusted => bytesHash(bytes.ptr, bytes.length, seed))();
+        return bytesHash(toUbyte(val), seed);
     }
     else //Other types. CTFE unsupported
     {
@@ -73,6 +72,12 @@ if (!is(T == enum) && !is(T : typeof(null)) && is(T S: S[]) && !__traits(isStati
     const _ = hashOf("abc");
 }
 
+nothrow pure @system unittest
+{
+    void*[] val;
+    const _ = hashOf(val); // Check a PR doesn't break this.
+}
+
 //arithmetic type hash
 @trusted nothrow pure
 size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && __traits(isArithmetic, T))
@@ -80,13 +85,11 @@ size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && __traits
     static if(__traits(isFloating, val))
     {
         T data = (val != val) ? T.nan : val;
-        auto bytes = toUbyte(data);
-        return bytesHash(bytes.ptr, bytes.length, seed);
+        return bytesHash(toUbyte(data), seed);
     }
     else
     {
-        auto bytes = toUbyte(val);
-        return bytesHash(bytes.ptr, bytes.length, seed);
+        return bytesHash(toUbyte(val), seed);
     }
 }
 
@@ -134,16 +137,21 @@ size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && (is(T ==
 
         static if (is(typeof(toUbyte(val)) == const(ubyte)[]))//CTFE ready for structs without reference fields
         {
-            auto bytes = toUbyte(val);
-            return bytesHash(bytes.ptr, bytes.length, seed);
+            return bytesHash(toUbyte(val), seed);
         }
         else // CTFE unsupproreted for structs with reference fields
         {
             assert(!__ctfe, "unable to compute hash of "~T.stringof);
-            const(ubyte)[] bytes = (cast(const(ubyte)*)&val)[0 .. T.sizeof];
-            return bytesHash(bytes.ptr, bytes.length, seed);
+            return bytesHash(toUbyte(val), seed);
         }
     }
+}
+
+nothrow pure @safe unittest // issue 18925
+{
+    // Check hashOf struct of scalar fields is usable in @safe code.
+    static struct S { int a; int b; }
+    auto h = hashOf(S.init);
 }
 
 //delegate hash. CTFE unsupported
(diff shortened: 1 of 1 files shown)


src/core/internal/hash.d
```diff
@@ -135,7 +135,7 @@ size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && (is(T ==
         static if (is(typeof(toUbyte(val)) == const(ubyte)[]))//CTFE ready for structs without reference fields
         {
             auto bytes = toUbyte(val);
-            return bytesHash(bytes.ptr, bytes.length, seed);
+            return (() @trusted => bytesHash(bytes.ptr, bytes.length, seed))();
```
jacob-carlborg: Can `bytesHash` be made safe instead?

JinShil: If it can't be made safe, and we want to use it in `@safe` context, we should try to certify it as safe and attribute the entire function as @trusted.

jacob-carlborg: Perhaps add an overload to take an array instead of pointer and length pair as well.

jacob-carlborg: It's doing pointer arithmetic in a new places, it's explicitly marked as `@system`.

n8sh: >Perhaps add an overload to take an array instead of pointer and length pair as well.

I could add a `@trusted` overload that calls `bytesHash` if you think that buys us anything over the `@trusted` lambda that calls `bytesHash`. My sole concern is that it might result in worse machine code, since the compiler might treat inline lambdas more intelligently. I'll take a look.

JinShil: This PR appears to be an attempt to silence the compiler rather than heeding its advice.  Essentially this PR is just casting away `byteHash`'s unsafeness.

When we attribute anything as `@trusted`, we are saying to our users that we have thoroughly analyzed the implementation and certified it as memory safe, even though the implementation utilizes unsafe language features.  I'm not seeing any evidence that that due diligence has been done.

IMO, we need to rewrite `bytesHash` using the language's memory-safe guarantees and attribute it with `@safe`, or we need to do a thorough analysis of the `bytesHash` function to certify it as memory-safe, ad attribute it with `@trusted`.

n8sh: >This PR appears to be an attempt to silence the compiler rather than heeding its advice. Essentially this PR is just casting away byteHash's unsafeness.

Yes. Although the function is unsafe in general due to the lack of language-level enforcement that the first two arguments describe a valid array slice, we know the call is safe here because we are passing it the pointer and length of an actual array. I believe this is an appropriate use of `@trusted`.

JinShil: @n8sh Ok.  I'll need to look a little closer at this PR then, but at a minimum that explanation should accompany the `@trusted` attribution in the form of a comment.

JinShil:
I looked into this a little more.  This PR does appear safe given that the byte array's length is being used to constrain the memory access.  However, I'd prefer to do a little more due diligence on this.

There is safe and unsafe code in `bytesHash`.  I'd like to see the 32-bit block hashing code factored out into it's own `@system` function.  Then have the `bytesHash` declared `@safe` with an `@trusted` call to that unsafe 32-bit block hashing function.  I believe this will make the safety easier to audit, and then all calls to `bytesHash` will be safe.

n8sh:
>Gah! Scratch that. It looks like `bytesHash` will need to be audited as whole

It's very nearly a copy-paste of the reference implementation of 32 bit MurmurHash3, so the audit should be quick.

https://github.com/aappleby/smhasher/blob/61a0530f28277f2e850bfc39600ce61d02b518de/src/MurmurHash3.cpp#L94-L146

JinShil:
> the audit should be quick.

I don't think `bytesHash` can be safe because it depends on how its called.  Only the callers know whether or not the length is within the bounds of the memory.  So, the current implementation isn't bad, but I think it would be best if we could make something like this:

```D
size_t safeBytesHash(ubyte[] byte, size_t seed) @safe
{
    return (() @trusted => bytesHash(bytes.ptr, bytes.length, seed))();
}
```

I realize this seems no different than what you've done, but it provides a safe interface for all callers, not just the on in this PR.  Users won't have to audit their own calls individually.  By using this interface, they'll know their call is safe.

jacob-carlborg:
Is there a need for the pointer and length to be separate, except for backwards compatibility in the API?

n8sh:
>I realize that seems no different than what you've done, but it provides a safe interface for all callers, not just the one in this PR. Users won't have to audit their own calls individually. By using this interface, they'll know their call is safe.

If you named that function `hashOf(const ubyte[] val, size_t seed=0)`, that would be the same as this PR!

n8sh:
>If you named that function `hashOf(const ubyte[] val, size_t seed=0)`, that would be the same as this PR!

Oops - I was mistaking this PR with the already-merged PR https://github.com/dlang/druntime/pull/2197 which is like this PR except for arrays. Because now that's in, I could use it if you like.

jacob-carlborg:
```d
size_t safeBytesHash(ubyte[] byte, size_t seed) @safe
{
    return (() @trusted => bytesHash(bytes.ptr, bytes.length, seed))();
}
```

Can we do the opposite? Make `bytesHash` `@safe` by taking an array and have an overload that have the pointer and length separately?

```d
size_t bytesHash(ubyte[] byte, size_t seed) @safe
{
    // implementation
}

size_t bytesHash(ubyte* ptr, size_length, size_t seed) @system
{
    bytesHash(ptr[0 .. length]);
}
```

n8sh:
If you replace `@safe` with `@trusted` then sure.

JinShil:
Yes it will need to be @trusted because the implementation uses unsafe language features.  But I would support that design as well.

n8sh:
Implemented suggested change.

jacob-carlborg:
Please squash the commits when done.

n8sh:
Squashed.

n8sh:
>I'm not sure if this is performance critical code

Is the speed of a hashtable performance critical?

n8sh:
Rebased.

wilzbach:
This introduced a regression: https://issues.dlang.org/show_bug.cgi?id=19005

n8sh:
The fix is to replace
```d
        else // CTFE unsupproreted for structs with reference fields
        {
            assert(!__ctfe, "unable to compute hash of "~T.stringof);
            return bytesHash(toUbyte(val), seed);
        }
```
with
```d
        else // CTFE unsupproreted for structs with reference fields
        {
            assert(!__ctfe, "unable to compute hash of "~T.stringof);
            const(ubyte)[] bytes = (cast(const(ubyte)*)&val)[0 .. T.sizeof];
            return bytesHash(bytes, seed);
        }
```

n8sh:
Will make a PR with an appropriate test.

n8sh:
PR: https://github.com/dlang/druntime/pull/2223

PR #2202, opened 2018-06-01
Fix Issue 18932 - core.internal.hash.hashOf(val, seed) ignores `seed` when val is a raw pointer

n8sh:
This PR makes this pass:
```d
unittest
{
    import core.internal.hash : hashOf;
    assert(hashOf(null, 0) != hashOf(null, 123456789));
}
```


wilzbach:
Once https://github.com/dlang/druntime/pull/2205 has been merged, CircleCi can be restarted and this PR can finally get merged.

PR #2203, opened 2018-06-01
Make -betterC code nothrow and @nogc

JinShil:
This is a rework of https://github.com/dlang/druntime/pull/2184, but since https://github.com/dlang/druntime/pull/2194 has already been merged, the context and motivations for it have changed.

-betterC does not support throwing exceptions or garbage collection.  Users should, therefore, always be attributing their -betterC code with `nothrow` and `@betterC`.  To ensure -betterC builds receive D's compile-time correctness guarantees, runtime implementations, when compiled with -betterC, should also be attributed `nothrow` and `@nogc`.

This PR is still required for https://github.com/dlang/dmd/pull/8253, but is still valuable (and fundamentally correct) despite the fate of that PR.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -44,6 +44,12 @@ alias dstring = immutable(dchar)[];
 
 version (D_ObjectiveC) public import core.attribute : selector;
 
+/*======================= Attribute-inferred code ==============================
+  The following code consists primarily of templates whose attributes (e.g.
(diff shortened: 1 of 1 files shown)


JinShil:
Gah.  My mistake.  _destructRecurse is a template.  Closing.

PR #2204, opened 2018-06-02
Fix linker error in destroy under -betterC

JinShil:
This is a followup to https://github.com/dlang/druntime/pull/2178

My fix was insufficient.  Although it fixed the compiler error, there was still a linker error.  Because we were copying two arrays, it was making a runtime call to `_d_arraycopy`.  Since the runtime is not linked in -betterC, it resulted in an unresolved reference error at link time.

See the implementation of `_d_arraycopy` at https://github.com/dlang/druntime/blob/4a1d176fb9f26441d16aa8ae12bb8beacc6e1972/src/rt/arraycat.d#L22.  It basically just forwards to `memcpy`, so the array copy code was simply replaced with a direct call to `memcpy`.

The test case was also updated to compile and link.

Sorry for the mistake.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -509,9 +509,10 @@ void destroy(T)(ref T obj) if (is(T == struct))
 
     _destructRecurse(obj);
     () @trusted {
+        import core.stdc.string : memcpy;
         auto dest = (cast(ubyte*) &obj)[0 .. T.sizeof];
         auto src = (cast(ubyte*) &init)[0 .. T.sizeof];
-        dest[] = src[];
+        memcpy(dest.ptr, src.ptr, T.sizeof);
     } ();
 }
 
--- a/test/betterc/Makefile
+++ b/test/betterc/Makefile
@@ -6,7 +6,7 @@ TESTS:=test18828
 all: $(addprefix $(ROOT)/,$(addsuffix ,$(TESTS)))
 
 $(ROOT)/%: $(SRC)/%.d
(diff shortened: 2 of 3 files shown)


n8sh:
If `dest[] = src[]` doesn't work in `betterC`, that should be fixed rather than applying a bandaid here.

JinShil:
It is possible to make `dest[] = src[]` work in -betterC, but that is an enormous undertaking.  It requires rewriting the array lowerings in the compiler to templates (similar to what was done for __equals).  I've been working on it, but is extremely difficult and filled with all kinds of problems.

For example, did you know you can set the `length` property of an array in `@safe nothrow pure` code, but it gets lowered to a runtime hook which is neither `@safe`, `nothrow`, nor `pure`?  The compiler is lying to us.  I attempted it already, but I discovered it required all the in https://github.com/dlang/druntime/blob/098a08a9593a8e265a9b02da55bda53393f3d5c1/src/rt/lifetime.d to be rewritten.  That is a long-term project and one, which if not done carefully, could cause quite a few problems.  It's going to take years to get that done, but in the meantime users are expecting features like this to work.

n8sh:
@JinShil Of course you're right.

I was also writing under the mistaken assumption that slice copying not working in `betterC` was a recent regression rather than a longstanding limitation, since [the documentation claims that array slicing is supported in -betterC](https://dlang.org/spec/betterc.html#retained).

PR #2209, opened 2018-06-05
Fix Issue 18942 - core.internal.hash can take advantage of alignment info on non-x86

n8sh:
On machines that do not support unaligned memory access, `core.internal.hash` can use statically available alignment information to determine when `uint` reads can be used.

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -57,12 +57,12 @@ if (!is(T == enum) && !is(T : typeof(null)) && is(T S: S[]) && !__traits(isStati
     else static if (is(typeof(toUbyte(val)) == const(ubyte)[]))
     //ubyteble array (arithmetic types and structs without toHash) CTFE ready for arithmetic types and structs without reference fields
     {
-        return bytesHash(toUbyte(val), seed);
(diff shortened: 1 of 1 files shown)


n8sh:
This code that disables unaligned reads with DMD is from the original 2013 commit. Maybe the original reason (whatever it was) no longer applies.

```d
version(AnyX86)
{
    version(DigitalMars)
    {
    }
    else
    {
        version = HasUnalignedOps;
    }
}
```

This code and the comment about some kind of inlining bug also dates back to the original 2013 commit.
```d
        //Compiler can optimize this code to simple *cast(uint*)x if it possible.
        version(HasUnalignedOps)
        {
            if (!__ctfe)
                return *cast(uint*)x; //BUG: Can't be inlined by DMD
        }
```

n8sh:
The above-mentioned code indeed was outdated, so I've made PR https://github.com/dlang/druntime/pull/2210 to remove it.

n8sh:
Revised to remove unnecessary uses of `version(HasUnalignedOps)` which would be removed anyway by PR https://github.com/dlang/druntime/pull/2209.

wilzbach:
@n8sh FYI - rebased this for you.

PR #2210, opened 2018-06-05
Fix Issue 18943 - core.internal.hash remove outdated special case for DMD unaligned reads

n8sh:
`core.internal.hash` has lines dating back to the [initial commit](https://github.com/dlang/druntime/commit/8df094c4d5e0c3db52c4f372fe91d686f3163ad8#diff-1cb07fafd18644ca9974aea7a74483e9) on November 16, 2013 that disable unaligned reads when using DMD and a comment that a DMD bug would prevent inlining. That is not true in the present day so the comment and alternative code branch have no purpose except to confuse.

With optimization enabled DMD produces the same code for
```d
return *cast(uint*)x;
```
and
```d
return ((cast(uint) x[3]) << 24) | ((cast(uint) x[2]) << 16) | ((cast(uint) x[1]) << 8) | (cast(uint) x[0]);
```
when `x` is `ubyte*`, so this patch is not expected to improve performance except for non-optimized builds.

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -424,23 +424,6 @@ unittest // issue 15111
 // MurmurHash3 was written by Austin Appleby, and is placed in the public
 // domain. The author hereby disclaims copyright to this source code.
 
-version(X86)
-    version = AnyX86;
-version(X86_64)
-    version = AnyX86;
-
-version(AnyX86)
-{
-    version(DigitalMars)
-    {
-    }
-    else
-    {
-        version = HasUnalignedOps;
-    }
-}
-
-
 @system pure nothrow @nogc
 size_t bytesHash(const(void)* buf, size_t len, size_t seed)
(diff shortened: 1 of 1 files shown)


n8sh:
That's a difference between llvm and gcc and so probably not applicable to GDC, so I'll just replace the `version(DigitalMars)` with `version(LDC)`.

n8sh:
BTW, this PR adds a unit test that passes an unaligned byte range to `bytesHash` and ensures the result is as expected.

n8sh:
>and using the array syntax, we could even make the whole function @safe.

Doesn't the specification require that `@safe` methods always perform bounds checks on array access, even in release mode, unless bounds checking is disabled program-wide? I don't think that's a good thing.

n8sh:
Pushed revised version. `HasUnalignedOps` branch entirely removed.

PR #2214, opened 2018-06-14
fix issue 18981: SIGSEGV during backtrace when debug info is compressed

yshui:
Detect compressed debug line section, do not try to resolve line numbers if
the debug info is compressed.

Side note: seems I could not get line number no matter the debug info is compressed or not, is that a bug?


yshui:
@JinShil What is needed to be done for #2172 ?

JinShil:
> Side note: seems I could not get line number no matter the debug info is compressed or not, is that a bug?

Probably https://issues.dlang.org/show_bug.cgi?id=18068

src/rt/backtrace/elf.d
```diff
@@ -52,6 +52,10 @@ struct Image
         if (dbgSectionIndex != -1)
         {
             auto dbgSectionHeader = ElfSectionHeader(&file, dbgSectionIndex);
+            // we don't support compressed debug sections
+            static if (__traits(compiles, SHF_COMPRESSED))
```
wilzbach: AFAICT there's no reason for this special casing here. FreeBSD and DragonflyBSD also use the same ELF format and thus imho it's better to add the `SHF_COMPRESSED` constant to the respective files.

yshui:
@wilzbach Fixed

PR #2219, opened 2018-06-18
Allow to start the GC eagerly

BBasile:
In some very rare cases (so far verified for string literals used in custom array) it's necessary to initialize the GC by hand. See https://issues.dlang.org/show_bug.cgi?id=18996#c4.

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -160,6 +160,15 @@ struct GC
     @disable this();
 
     /**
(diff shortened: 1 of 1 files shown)


Geod24:
I think we should fix the bug rather than relying on the developper's virtue. If you put a module ctor in Phobos that calls this, what's the point of it being lazy ?

PR #2220, opened 2018-06-18
Fix issue 18996 - ProtoGC should support removing roots and ranges that were not originally added

schveiguy:
The original behavior was to assert(false), but this is inconsistent with the standard GC, which ignores removal of ranges it didn't already know about.

The code in question, `Array!string`, whenever it would realloc, it would removeRange on the previous array, and addRange on the new one. Since it starts out with a null array, `removeRange(null)` was called.

Arguably, this is not behavior we want (`Array` could easily be changed to not try and `removeRange(null)`), but the `ProtoGC` shouldn't behave differently from the actual GC while it is in charge of things.

Added a simple test to make sure this doesn't happen again.

### DIFF
--- a/src/gc/impl/proto/gc.d
+++ b/src/gc/impl/proto/gc.d
@@ -177,7 +177,6 @@ class ProtoGC : GC
                 return;
             }
         }
-        assert(false);
     }
 
     @property RootIterator rootIter() return @nogc
@@ -211,7 +210,6 @@ class ProtoGC : GC
                 return;
             }
         }
-        assert(false);
     }
 
     @property RangeIterator rangeIter() return @nogc
--- a/test/init_fini/Makefile
+++ b/test/init_fini/Makefile
@@ -1,6 +1,6 @@
 include ../common.mak
 
-TESTS:=thread_join runtime_args
+TESTS:=thread_join runtime_args test18996
 
 .PHONY: all clean
 all: $(addprefix $(ROOT)/,$(addsuffix .done,$(TESTS)))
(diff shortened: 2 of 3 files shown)


schveiguy:
ping @MartinNowak this should go into 2.081.0, it's a regression.

wilzbach:
@schveiguy as long as it's targeted on stable (which it is) and merged before the end of this month, it will be part of the final 2.081

schveiguy:
> ProtoGC remembers identical roots/ranges multiple times, so might keep them alive longer.

That's a simple fix as well, just don't return :)

schveiguy:
I think jenkins wasn't coming back with a test, apparently you have to close and reopen.

wilzbach:
@schveiguy ci.dlang.io is currently down. It will never reply.

schveiguy:
Hm... so I guess I'll use the auto-tester to merge. Thanks for the update.

schveiguy:
Auto-merge toggled on

wilzbach:
> Hm... so I guess I'll use the auto-tester to merge. Thanks for the update.

That won't help you, because it can't merge when there are required CI checks that haven't passed yet.
I do have admin access and can still merge this pull request (exactly for these cases).
I will do this for this PR now.
I hope that ci.dlang.io gets back online soonish, but I haven't heard back from Martin.

schveiguy:
Ah thanks, I was confusing this case with the one where the bot won't merge when non-required CIs haven't reported.

rainers:
> That's a simple fix as well, just don't return :)

Yes, but I think the returning API is much saner. If you have two independent sources that happen to add the same root/range, it is very likely that the root/range is still in use after one source has released it.

schveiguy:
> If you have two independent sources that happen to add the same root/range, it is very likely that the root/range is still in use after one source has released it.

Agreed. So maybe it's the conservative GC that needs updating.

PR #2223, opened 2018-06-19

n8sh:


n8sh:
The case that revealed this regression also revealed a bug in `core.internal.convert.toUbyte`. It can't convert an enum E if `E.sizeof == 1` because it matches multiple templates.

schveiguy:
Would it not be better to fix `toUbyte`?

n8sh:
@schveiguy The bug here is real. It just exposed a second bug in another module. I'm working now on a fix for `toUbyte` of enums.

schveiguy:
OK, so help me understand -- I see the only real line in any of the flavors of `toUbyte` functions if it's not `__ctfe` is this:

```D
        return (cast(const(ubyte)*)&val)[0 .. T.sizeof];
```

Which actually is what you have done here locally. I get that there's a bug, but I'm having a hard time seeing that what you are doing is different. Can you explain?

n8sh:
@schveiguy Give me a second, it will be in the patch. The change will explain the problem.

n8sh:
Here: PR https://github.com/dlang/druntime/pull/2224

n8sh:
>I'm having a hard time seeing that what you are doing is different. Can you explain?

Oh, I think I get your question. `toUbyte` is CTFE-able and needs a CTFE branch for every argument. If that isn't available then `toUbyte` can't be called. So that's why there exists a runtime-only branch.

schveiguy:
So I agree about the toUbyte bug, but I still have the question, do we need this PR if we fix the toUbyte bug? If so, please target stable for this PR as well. This should go into 2.081.

schveiguy:
Huh? I brought up CTFE because you have right in your code `assert(!__ctfe)`, so it must be calling the non `__ctfe` branch.

n8sh:
This doesn't need to target stable because it's fixing a regression that didn't make it there yet.

schveiguy:
> This doesn't need to target stable because it's fixing a regression that didn't make it there yet.

The bug says 2.081-b1, which means it's on the stable branch already. It hasn't been released, but if this patch goes into master instead of stable, the bug will get released as 2.081.

n8sh:
I see the mistake I made

nordlow:
Thanks!

PR #2225, opened 2018-06-19

n8sh:
Resubmission of https://github.com/dlang/druntime/pull/2223 to target stable.


schveiguy:
In order to target stable, you have to rebase onto stable. I always have to look at these instructions when I'm doing it:
https://wiki.dlang.org/Starting_as_a_Contributor#Stable_Branch

n8sh:
It turns out this was unnecessary because the change this is fixing isn't in stable yet.

PR #2226, opened 2018-06-19
Fix Issue 19008 - core.internal.convert.toUbyte doesn't work on enums

n8sh:
Resubmission of https://github.com/dlang/druntime/pull/2224 to target stable.

PR #2227, opened 2018-06-19

n8sh:
Resubmission of #2223 to target stable, again.

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -139,10 +139,11 @@ size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && (is(T ==
         {
             return bytesHash(toUbyte(val), seed);
         }
-        else // CTFE unsupproreted for structs with reference fields
+        else // CTFE unsupported
         {
             assert(!__ctfe, "unable to compute hash of "~T.stringof);
-            return bytesHash(toUbyte(val), seed);
+            const(ubyte)[] bytes = (() @trusted => (cast(const(ubyte)*)&val)[0 .. T.sizeof])();
+            return bytesHash(bytes, seed);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


schveiguy:
I tested locally, the changes in #2226 fix the issue, so I think we should go that route.

n8sh:
> I think we should go that route.

@schveiguy We need to go both routes. There are two bugs. The bug in `hash.d` is that `toUbyte` is called in a branch where we have determined that `toUbyte` cannot be called. The bug in `convert.d` is that `toUbyte` should be callable on enums but isn't.

schveiguy:
OK, *now* I see what you are talking about. I didn't see the static if above (it was helpfully cut off by github when doing the review). However, let's take a step back...

PR #2226 is *also* fixing 19005 with `Date.hashOf`, because now it *can* call toUbyte on the enum, and the second branch isn't even used, the first one is. I've verified your test case here does pass with that change alone. So the unittest can go in either PR, we can just leave it here.

You are right that this second branch is nonsense. Looking at the history, I can see that what you are writing here re-implements what was there before, but I'm curious if it will ever be triggered with your enum fix. I wonder if it was a "I can't figure out why this is working, I'll just stick this here" branch. It seems rather odd to reproduce the `toUbyte` code locally.

Looks like the branch was originally in this when first committed. I wonder if the kinks have all been worked out in `toUbyte` that would make this work? Maybe the enum fix is what would do it?

BTW, can you fix the typo "unsupproreted" to "unsupported"?

schveiguy:
OK, now I'm starting to understand this better. I didn't realize that the CTFE version must *still* compile, even if you only use the function at runtime. So that is why the repeat of the non-ctfe code is necessary.

I think this can be simplified. But probably not in the stable branch, I'll look into it once these are merged.

n8sh:
>BTW, can you fix the typo "unsupproreted" to "unsupported"?

The comment seems to be outdated, so I'll just remove it.

PR #2228, opened 2018-06-20
Fix Issue 19009 - core.internal.hash.hashOf default hash (absent `toHash`) should be `@nogc`

n8sh:
The default `hashOf` implementation used for types that do not have `toHash` uses `toUbyte` to get the byte representation which during CTFE requires allocation but during execution does not. The compiler currently takes CTFE-only code into account when inferring attributes but we can give it some help.

This PR is based on the premise that it is correct to label a function as `@nogc` if it only uses GC for CTFE interpretation. If that's not true then this PR should be rejected.

### DIFF
--- a/src/core/internal/convert.d
+++ b/src/core/internal/convert.d
@@ -10,13 +10,36 @@
 module core.internal.convert;
 import core.internal.traits : Unqual;
 
-@trusted pure nothrow
+/+
+A @nogc function can allocate memory during CTFE.
++/
+@nogc nothrow pure @trusted
+private ubyte[] ctfe_alloc()(size_t n)
+{
+    if (!__ctfe)
+    {
+        assert(0, "CTFE only");
+    }
+    else
+    {
+        static ubyte[] alloc(size_t x) nothrow pure
+        {
+            if (__ctfe) // Needed to prevent _d_newarray from appearing in compiled prorgam.
+                return new ubyte[x];
+            else
+                assert(0);
+        }
+        return (cast(ubyte[] function(size_t) @nogc nothrow pure) &alloc)(n);
+    }
+}
+
+@trusted pure nothrow @nogc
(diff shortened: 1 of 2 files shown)


src/core/internal/convert.d
```diff
@@ -10,13 +10,33 @@
...
+    if (!__ctfe)
+    {
+        assert(0, "CTFE only");
+    }
+    else
+    {
+        static ubyte[] alloc(size_t x)
+        {
+            return new ubyte[x];
+        }
```
schveiguy: So it's clear we want to add `@nogc` to the allocation artificially. But since you don't have the `pure` and `nothrow` attributes on here, it's not required that they are inferred pure/nothrow. But it's definitely possible for someone to come along and put in impure or throwing functions in here, and all of a sudden you have casted that away below.

2 options I see: a) attribute the function (can that even be done? I can't remember), b) static assert the type of the function before casting.

src/core/internal/convert.d
```diff
@@ -10,13 +10,33 @@
 module core.internal.convert;
 import core.internal.traits : Unqual;
 
-@trusted pure nothrow
+/+
+A @nogc function can allocate memory during CTFE.
++/
+@nogc nothrow pure @trusted
+private ubyte[] ctfe_alloc()(size_t n)
```
schveiguy: I suppose this is a template to avoid having it actually get stored in the object file?

n8sh: Yes, if it's unused. It doesn't prevent the function from appearing in the object file if it's used during CTFE, alas.

src/core/internal/convert.d
n8sh: Added attributes.

schveiguy:
Pretty funny that 0% of the code has coverage (as it's all only ever used in CTFE!). LGTM

n8sh:
Unfortunately this still can't be used with betterC. With the -betterC flag `new ubyte[x]` cannot be used even in CTFE.

PR #2238, opened 2018-07-02
Fix Issue 19048 - In core.internal.hash.hashOf reduce template bloat: remove `auto ref` where unneeded and add `const` where possible

n8sh:
This PR reduces template proliferation by not using `auto ref` unnecessarily. Scalars, dynamic arrays, raw pointers, delegates, objects, and associative arrays don't need to be passed by reference to `hashOf`. Additionally by adding `const` where we know it's legal this PR lets IFTI produce fewer distinct instantiations. For instance we can avoid having separate template instantiations for `core.internal.hash.hashOf(const int(1))` and `core.internal.hash.hashOf(int(1))`. While I was doing this I also added `scope` when references are known not to escape.

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -56,7 +56,7 @@ size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && __traits
 }
 
 //dynamic array hash
-size_t hashOf(T)(auto ref T val, size_t seed = 0)
+size_t hashOf(T)(T val, size_t seed = 0)
 if (!is(T == enum) && !is(T : typeof(null)) && is(T S: S[]) && !__traits(isStaticArray, T)
     && !is(T == struct) && !is(T == class) && !is(T == union))
 {
@@ -101,13 +101,14 @@ if (!is(T == enum) && !is(T : typeof(null)) && is(T S: S[]) && !__traits(isStati
 
 //arithmetic type hash
 @trusted @nogc nothrow pure
-size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && __traits(isArithmetic, T))
+size_t hashOf(T)(scope const T val, size_t seed = 0) if (!is(T == enum) && __traits(isArithmetic, T))
 {
     static if(__traits(isFloating, val))
     {
         static if (floatCoalesceZeroes || floatCoalesceNaNs)
         {
-            auto data = val;
+            import core.internal.traits : Unqual;
+            Unqual!T data = val;
             // Zero coalescing not supported for deprecated complex types.
             static if (floatCoalesceZeroes && is(typeof(data = 0)))
                 if (data == 0) data = 0; // +0.0 and -0.0 become the same.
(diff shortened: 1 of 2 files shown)


src/core/internal/hash.d
```diff
@@ -173,21 +173,21 @@ nothrow pure @safe unittest // issue 19005
...
 //class or interface hash. CTFE depends on toHash
-size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && is(T == interface) || is(T == class))
+size_t hashOf(T)(T val, size_t seed = 0) if (!is(T == enum) && is(T == interface) || is(T == class))
 {
     return hashOf(val ? (cast(Object)val).toHash() : 0, seed);
 }
 
 //associative array hash. CTFE depends on base types
-size_t hashOf(T)(auto ref T aa, size_t seed = 0) if (!is(T == enum) && __traits(isAssociativeArray, T))
+size_t hashOf(T)(T aa, size_t seed = 0) if (!is(T == enum) && __traits(isAssociativeArray, T))
```
schveiguy: It's not as straightforward, but you can foreach on e.g. a `const int[int]`, is it worth adding const here if possible? You'd have to check if you can call `key.hashOf` and `val.hashOf`.

n8sh: IMO it's worth doing but since it involves more than just making trivial attribute changes I'd prefer to do it in a future PR.

>You'd have to check if you can call `key.hashOf` and `val.hashOf`.

It's a bit more involved than that. No matter what the type is I'll be able to call `hashOf(key)` and `hashOf(cast(const typeof(key)) key)` but they might not have the same result, for instance if the key has a non-const `toHash` function.

n8sh:
Rebased.

schveiguy:
This is simple enough to merge without other reviews.

src/core/internal/hash.d
schveiguy: I understand. It's concerning to have `hashOf` return different results on const and non-const structs regardless of whether it's in an AA.

src/core/internal/hash.d
```diff
@@ -256,21 +257,21 @@ nothrow pure @safe unittest // issue 19005
...
+size_t hashOf(T)(scope const T val, size_t seed = 0) if (!is(T == enum) && is(T == delegate))
 {
     assert(!__ctfe, "unable to compute hash of "~T.stringof);
     const(ubyte)[] bytes = (cast(const(ubyte)*)&val)[0 .. T.sizeof];
     return bytesHashAlignedBy!T(bytes, seed);
 }
 
 //class or interface hash. CTFE depends on toHash
-size_t hashOf(T)(auto ref T val, size_t seed = 0) if (!is(T == enum) && is(T == interface) || is(T == class))
+size_t hashOf(T)(T val, size_t seed = 0) if (!is(T == enum) && is(T == interface) || is(T == class))
```
ZombineDev: Why not `scope const T`? Even if the `T`'s `toHash` is not qualified appropriately it doesn't matter since we're casting to `Object`. At most we'll need a trusted cast.

schveiguy: This whole function should probably be split up into multiple pieces. Note that this will be called if the class is a C++ class, in which case the cast to Object will return null and cause a segfault.

PR #2239, opened 2018-07-02
Fix Issue 19049 - object.hashOf - don't wrap a public function with an identical public function

n8sh:
This is hashOf in object.d:

```d
{
    import core.internal.hash;
    return core.internal.hash.hashOf(arg, seed);
}
```

This serves no purpose because `core.internal.hash.hashOf` is public and also has an `auto ref` argument. So `core.internal.hash.hashOf` should either be publicly imported or exposed through an alias.

This issue has synergy with PR https://github.com/dlang/druntime/pull/2238.


src/object.d
```diff
@@ -4236,10 +4236,18 @@ Params:
...
-size_t hashOf(T)(auto ref T arg, size_t seed = 0)
+version (D_Ddoc)
 {
-    import core.internal.hash;
-    return core.internal.hash.hashOf(arg, seed);
+    // This lets DDoc produce better documentation.
+    size_t hashOf(T)(auto ref T arg, size_t seed = 0)
+    {
+        static import core.internal.hash;
+        return core.internal.hash.hashOf(arg, seed);
```
wilzbach: Do we even need a body if it's for ddoc only? An empty body should work too?

jacob-carlborg: Should work.

n8sh: When I replace it with `size_t hashOf(T)(auto ref T arg, size_t seed = 0);` I get an error:
```
../dmd/generated/osx/release/64/dmd -conf= -c -w -o- -Isrc -Iimport -version=CoreDdoc -Dfdoc/object.html project.ddoc  src/object.d
src/object.d(1605): Error: function object.hashOf!(void delegate()).hashOf is not nothrow
src/object.d(1603): Error: nothrow function object.TypeInfo_Delegate.getHash may throw
```

JinShil:
This PR may have caused a regression: https://issues.dlang.org/show_bug.cgi?id=19701

PR #2242, opened 2018-07-05
Add fixes to get Android/AArch64 working

joakim-noah:
This patch brings Bionic/AArch64 almost to parity with Glibc/AArch64 in terms of stdlib tests passing, see ldc-developers/ldc#2153.

### DIFF
--- a/src/core/sys/posix/sys/types.d
+++ b/src/core/sys/posix/sys/types.d
@@ -1056,12 +1056,15 @@ else version( CRuntime_Bionic )
         size_t  guard_size;
         int     sched_policy;
         int     sched_priority;
-        version(AArch64) char[16] __reserved;
+        version(D_LP64) char[16] __reserved;
     }
 
     struct pthread_cond_t
     {
-        int value; //volatile
+        version(D_LP64)
+            int[12] __private;
+        else
+            int[1] __private;
     }
 
     alias c_long pthread_condattr_t;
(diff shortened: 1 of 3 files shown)


src/rt/sections_android.d
```diff
@@ -69,6 +69,8 @@ void initSections() nothrow @nogc
 
     auto pbeg = cast(void*)&_tlsend;
     auto pend = cast(void*)&__bss_end__;
+    // _tlsend is a 32-bit int and may not be 64-bit void*-aligned, so align pbeg.
+    version(D_LP64) pbeg = cast(void*)(cast(size_t)(pbeg + 7) & ~cast(size_t)7);
```
joakim-noah: `_tlsend` is an integer delimiter added by ldc at the end of TLS data in a compiled object file. The next ELF section after `.tbss` is always 64-bit aligned from all the executables I've seen, so align to the next 64-bit word here. Without aligning, this caused a bug because the GC assumed it was aligned and would iterate over this range in 64-bit steps when scanning and not see any GC pointers. Not a problem for 32-bit because `void*` is 32-bit there, just like `_tlsend`.

src/core/sys/posix/sys/types.d
```diff
@@ -1056,37 +1056,40 @@ else version( CRuntime_Bionic )
...
 
     struct pthread_rwlock_t
     {
-        pthread_mutex_t  lock;
-        pthread_cond_t   cond;
-        int              numLocks;
-        int              writerThreadId;
-        int              pendingReaders;
-        int              pendingWriters;
-        void*[4]         reserved;
```
wilzbach: Isn't this a breaking change as people might have been accessing these fields?

joakim-noah: Yep, [they chose to break it 3 years back](https://android.googlesource.com/platform/bionic/+/2fabea47ac9475bcc52aff0715819d18aa5bdf1d%5E%21/#F1), just updating this file to reflect that change.

jacob-carlborg: Could we add forwarding properties and deprecate those?

joakim-noah: Eh, I doubt anyone is using this Bionic declaration, and given that it's broken upstream in their libc, what's the point?

joakim-noah:
Forgot to add one last change to this pull, don't merge, will update it tomorrow.

joakim-noah:
Alright, got the last change in, function declaration in `core.stdc.stdlib`, this pull is ready to go.

src/core/sys/posix/sys/types.d
ZombineDev: According to the Posix standard, the fields of those structs are implementation defined, so the user has no business in acessing them.

ZombineDev: For libc implementations we only need to ensure that they have the same ABI (size and perhaps alignment).

PR #2243, opened 2018-07-05
use core.internal.hash for TypeInfo.getHash

IgorStepanov:
Use the new `core.internal.hash.hashOf` in all `TypeInfo.getHash` instead of the old `rt.util.hash.hashOf`.
Also make `typeid(T).getHash(&val)` to get the same result with `hashOf(val)`.


IgorStepanov:
Please, note, tests in another projects, which fails on `continuous-integration/jenkins/pr-merge` depends on  associative array order:
for example, in `ae` xml build expects result string 
`<svg xmlns="http://www.w3.org/2000/svg" version="1.1"><text fill="red" x="0" y="15">I love SVG</text></svg>`
or 
`<svg xmlns="http://www.w3.org/2000/svg" version="1.1"><text x="0" y="15" fill="red">I love SVG</text></svg>`

After this PR, result string is
`<svg xmlns="http://www.w3.org/2000/svg" version="1.1"><text fill="red" y="15" x="0">I love SVG</text></svg>
`
Obviously this fails is a test issue and doesn't threaten real user code.

wilzbach:
> Please, note, tests in another projects, which fails on continuous-integration/jenkins/pr-merge

FYI before this PR can be merged, the Project Tester needs to be green.
In other words: `ae` needs to pass on the Project Tester.
CC @CyberShadow 

> Obviously this fails is a test issue and doesn't threaten real user code.

Yes, but we still want our CIs to be green ;-)

CyberShadow:
Yeah, that's a bad test. Fixed & pushed tag.

src/core/internal/hash.d
```diff
@@ -64,14 +59,15 @@ if (!is(T == enum) && !is(T : typeof(null)) && is(T S: S[]) && !__traits(isStati
...
+                       && is(typeof(val[0].toHash()) == size_t)) ||
+                   ((floatCoalesceZeroes || floatCoalesceNaNs) && __traits(isFloating, ElementType)))
     //class or interface array or struct array with toHash(); CTFE depend on toHash() method
     //also do this for arrays of structs whose hashes would be calculated in a memberwise fashion
     {
         size_t hash = seed;
         foreach (o; val)
         {
-            hash = hashOf(o, hash);
+            hash = hashOf(hashOf(o), hash); //double hashing because TypeInfo.getHash doesn't allow to pass seed value
```
n8sh: Please explain why this is necessary. Does this patch make `hashOf(o, hash)` not compile?

src/rt/util/container/hashtab.d
```diff
@@ -146,11 +146,10 @@ private:
 
     static hash_t hashOf(in ref Key key) @trusted
     {
-        import rt.util.hash : hashOf;
         static if (is(Key U : U[]))
-            return hashOf(key, 0);
+            return .hashOf(key, 0);
         else
-            return hashOf((&key)[0 .. 1], 0);
+            return .hashOf((&key)[0 .. 1], 0);
```
n8sh: This changes the meaning! Is that intentional? The old version was hashing the key as bytes.

src/object.d
```diff
@@ -1703,7 +1703,7 @@ class TypeInfo_Class : TypeInfo
     override size_t getHash(scope const void* p) @trusted const
     {
         auto o = *cast(Object*)p;
-        return o ? o.toHash() : 0;
+        return hashOf(o ? o.toHash() : 0);
```
n8sh: Why is this necessary instead of
```d
o ? o.toHash() : 0
```
?
The reason that the version in `core.internal.hash.hashOf` is different is that it (currently) doesn't distinguish between the chaining and non-chaining cases, but that isn't the case here.

src/core/internal/hash.d
n8sh: Ah nevermind, I see
>Also make `typeid(T).getHash(&val)` to get the same result with `hashOf(val)`.

src/object.d
n8sh: Instead I'd write just `return hashOf(o)` if the goal is to ensure that the result is the same as `core.internal.hashOf`.

src/rt/aaA.d
```diff
@@ -672,25 +672,28 @@ extern (C) int _aaEqual(in TypeInfo tiRaw, in AA aa1, in AA aa2)
 extern (C) hash_t _aaGetHash(in AA* aa, in TypeInfo tiRaw) nothrow
 {
     if (aa.empty)
-        return 0;
+        return hashOf(0);
```
n8sh: Suggestion: revert this to `return 0` and change `core.internal.hash` to match.

src/rt/typeinfo/ti_cent.d
```diff
@@ -30,7 +28,7 @@ class TypeInfo_zi : TypeInfo
 
     override size_t getHash(scope const void* p)
     {
-        return rt.util.hash.hashOf(p[0 .. cent.sizeof], 0);
+        return hashOf(p[0 .. cent.sizeof], 0);
```
n8sh: To match what's in `core.internal.hash` you should instead write:
```d
return hashOf(*cast(const ucent*) p);
```

src/rt/typeinfo/ti_delegate.d
```diff
@@ -28,7 +27,7 @@ class TypeInfo_D : TypeInfo
 
     override size_t getHash(scope const void* p)
     {
-        return rt.util.hash.hashOf(p[0 .. dg.sizeof], 0);
+        return hashOf(*cast(dg*)p, 0);
```
n8sh: No need for `0` second argument. Currently it will be inferred, and in the future an explicit chaining argument might result in different behavior (https://github.com/dlang/druntime/pull/2246).

src/rt/typeinfo/ti_long.d
```diff
@@ -28,7 +26,7 @@ class TypeInfo_l : TypeInfo
 
     override size_t getHash(scope const void* p)
     {
-        return rt.util.hash.hashOf(p[0 .. long.sizeof], 0);
+        return hashOf(*cast(long*)p, 0);
```
n8sh: No need for `0` second argument. Currently it will be inferred, and in the future an explicit chaining argument might result in different behavior (https://github.com/dlang/druntime/pull/2246).

src/rt/typeinfo/ti_ucent.d
```diff
@@ -30,7 +28,7 @@ class TypeInfo_zk : TypeInfo
 
     override size_t getHash(scope const void* p)
     {
-        return rt.util.hash.hashOf(p[0 .. ucent.sizeof], 0);
+        return hashOf(p[0 .. ucent.sizeof], 0);
```
n8sh: To match what's in `core.internal.hash` you should instead write:
```d
return hashOf(*cast(const ucent*) p);
```

src/object.d
n8sh: If https://github.com/dlang/druntime/pull/2246 is accepted you will be able to revert this to `o ? o.toHash() : 0`.

src/object.d
```diff
@@ -1930,14 +1934,11 @@ class TypeInfo_Struct : TypeInfo
         assert(p);
         if (xtoHash)
         {
-            return (*xtoHash)(p);
+            return hashOf((*xtoHash)(p));
```
n8sh: If https://github.com/dlang/druntime/pull/2246 is accepted you will be able to revert this to `return (*xtoHash)(p);`

src/object.d
```diff
@@ -1857,10 +1857,14 @@ class TypeInfo_Interface : TypeInfo
...
     {
+        if (!*cast(void**)p)
+        {
+            return hashOf(null);
+        }
         Interface* pi = **cast(Interface ***)*cast(void**)p;
         Object o = cast(Object)(*cast(void**)p - pi.offset);
         assert(o);
-        return o.toHash();
+        return hashOf(o.toHash());
```
n8sh: If https://github.com/dlang/druntime/pull/2246 is accepted you will be able to revert this to `return o.toHash();`

src/rt/typeinfo/ti_n.d
```diff
@@ -21,7 +21,7 @@ class TypeInfo_n : TypeInfo
 
     override size_t getHash(scope const void* p) const
     {
-        return 0;
+        return hashOf(cast(void*)null);
```
n8sh: Post https://github.com/dlang/druntime/pull/2246 this should be the same as `return 0`.

src/rt/typeinfo/ti_ulong.d
```diff
@@ -28,7 +27,7 @@ class TypeInfo_m : TypeInfo
 
     override size_t getHash(scope const void* p)
     {
-        return rt.util.hash.hashOf(p[0 .. ulong.sizeof], 0);
+        return hashOf(*cast(ulong*)p, 0);
```
n8sh: No need for `0` second argument. Currently it will be inferred, and in the future an explicit chaining argument might result in different behavior (https://github.com/dlang/druntime/pull/2246).

src/rt/util/container/hashtab.d
IgorStepanov: I use `object.hashOf` instead of `rt.util.hash : hashOf`, because `rt.util.hash` has been removed.
I may force use `object.hashOf(void[])` if we strongly want bytes hash, but I think it is not needed.

src/rt/aaA.d
IgorStepanov: `core.internal.hash.hashOf` may use seed value.
The current function looks like
```
size_t hashOf(T)(T aa, size_t seed = 0) if (!is(T == enum) && __traits(isAssociativeArray, T))
{
    if (!aa.length) return hashOf(0, seed);
}
```

If I rewrite it as...
```
size_t hashOf(T)(T aa, size_t seed = 0) if (!is(T == enum) && __traits(isAssociativeArray, T))
{
    if (!aa.length) return seed;
}
```
... we will disable way to chain several hashes:
```
a = hashOf(cast(int[int])null, hashOf(cast(int[int])null));
b = hashOf(cast(int[int])null, hashOf(cast(int[int])null, hashOf(cast(int[int])null)));
// a == b but a should differs b
```
as result:
hashes of two arrays with different count of null AA will be same:
```
auto arr = [int[int].init, int[int].init];
auto arr2 = [int[int].init, int[int].init, int[int].init];
assert(hashOf(arr) != hashOf(arr2)); // will be failed
```

src/rt/typeinfo/ti_cent.d
IgorStepanov: done

src/rt/typeinfo/ti_delegate.d
IgorStepanov: done

src/rt/typeinfo/ti_long.d
IgorStepanov: done

src/rt/typeinfo/ti_ucent.d
IgorStepanov: done

src/rt/typeinfo/ti_ulong.d
IgorStepanov: done

src/rt/typeinfo/ti_n.d
IgorStepanov: Same above:
`hashOf([null, null])` and `hashOf([null, null, null])` should have different values, thus we need chaining by seed, thus we need to rehash seed 0 each time.

IgorStepanov:
About failed tests: I've created some PRs for the another projects which fixes tests.
Looks like this PR can be merged without breaking real code (except failing wrong unittests).
The single exception is `vibe.d`. 
My [PR](https://github.com/vibe-d/vibe.d/pull/2183) should be merged before this, because current `Bson` comparing is wrong and failed with new hashes (and should sometimes fails with old hashes, but do it less often).

mak/SRCS
```diff
@@ -347,7 +347,6 @@ SRCS=\
 	src\rt\backtrace\macho.d \
 	\
 	src\rt\util\array.d \
-	src\rt\util\hash.d \
```
wilzbach:
```
/dlang.org/.generated/modlist-prerelease.ddoc /dev/shm/dtest/work/repo/dlang.org/nodatetime.ddoc src/rt/util/container/treap.d
make[1]: *** No rule to make target '/dev/shm/dtest/work/repo/dlang.org/web/phobos-prerelease/rt_util_hash.html', needed by 'doc'.  Stop.
```

I think it needs to be removed from `mak/DOCS` too.

wilzbach:
> About failed tests: I've created some PRs for the another projects which fixes tests.

Thanks a lot!
I think all PRs have been merged. Vibe.d releases take a while, hence https://github.com/dlang/ci/pull/241 is needed for now.

IgorStepanov:
Any idea why does `ci/circleci` fail?

wilzbach:
> Any idea why does ci/circleci fail?

Yes. We recently added `gdb` to the druntime testsuite and unfortunately this was part of the CircleCi file that doesn't automatically upgrade itself :/
A rebase helps though.

IgorStepanov:
@wilzbach ping

wilzbach:
>  I think we should add a small changelog entry stating that the default toHash method has been changed to use hashOf and thus as an resulted e.g. tests which fasly assume that the order of built-in associative arrays is fixed, may fail.

Would be nice to still have this as this should raise awareness in case someone runs into similar issues. It's simple about adding a file to the `changelog` folder.
See https://github.com/dlang/druntime/tree/master/changelog

IgorStepanov:
@wilzbach Thanks!

>Would be nice to still have this as this should raise awareness in case someone runs into similar issues. It's simple about adding a file to the changelog folder.

Something like #2256 ?

PR #2244, opened 2018-07-08
core.sys.posix.signal: Fix size of sigevent and siginfo_t on 64bit.

ibuclaw:
Because the static if condition determining `__SI_PAD_SIZE` and `__SIGEV_PAD_SIZE` being default false, the overall size of `sigevent` and `siginfo_t` was 8 bytes too much, this affects all users of these types also, such as `core.sys.posix.aio`.

### DIFF
--- a/src/core/sys/posix/signal.d
+++ b/src/core/sys/posix/signal.d
@@ -941,7 +941,7 @@ version( CRuntime_Glibc )
 
     private enum __SI_MAX_SIZE = 128;
 
-    static if( false /* __WORDSIZE == 64 */ )
+    static if( __WORDSIZE == 64 )
     {
         private enum __SI_PAD_SIZE = ((__SI_MAX_SIZE / int.sizeof) - 4);
     }
@@ -1740,7 +1740,7 @@ else version( CRuntime_UClibc )
 
     private enum __SI_MAX_SIZE = 128;
 
-    static if( false /* __WORDSIZE == 64 */ )
+    static if( __WORDSIZE == 64 )
     {
         private enum __SI_PAD_SIZE = ((__SI_MAX_SIZE / int.sizeof) - 4);
     }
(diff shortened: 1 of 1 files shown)


ibuclaw:
@rracariu - Is `CRuntime_UClibc` also wrong?

rracariu:
Most likely. That logic is exactly the same in uclibc and I copied the glibc version for determining the pad size.

ibuclaw:
@rracariu - I don't suppose you could confirm with `printf("%d\n", sizeof(struct sigevent));` ?

The size is 168 from glibc.

ibuclaw:
@rracariu - ok. I didn't have any reason to believe that any 32bit platforms are problematic in the first place.

I'll uncomment it then as I expect 64bit to report the wrong size same as glibc path.

n8sh:
I would ask if there initially was a reason for `static if( false /* __WORDSIZE == 64 */ )` but since that line is at least 10 years old and predates this git repository that would probably be futile.

ibuclaw:
@n8sh legacy days of 32bit only.

n8sh:
[The jenkins failure is for `mecca.reactor.io.signals` so it might not be spurious](https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdruntime/detail/PR-2244/2/pipeline/240).

wilzbach:
Restarted their test. Let's hope it really was a spurious failure :/

wilzbach:
The new Mecca failure is known (https://github.com/weka-io/mecca/issues/7), though it shouldn't happen as they explicitly disable this test for Jenkins :/

jacob-carlborg:
vibe.d is failing as well.

ibuclaw:
> vibe.d is failing as well.

```
Running ../../../../../../tmp/.dub/build/tests-~master/application-debug-linux.posix-x86_64-dmd_2081-69416B10FDA08D03A91B76292C7FB277/tests 
Not an executable file
```

ibuclaw:
> The jenkins failure is for mecca.reactor.io.signals so it might not be spurious.

If there was anything wrong with this (the struct is now too small), then the result would be a segmentation fault in libc.

PR #2246, opened 2018-07-09
Fix Issue 19071 - core.internal.hash should have non-chained toHash overloads

n8sh:
Each hash function in `core.internal.hash` currently has the form:

```d
size_t hashOf(T)(auto ref T val, size_t seed = 0);
```

This PR changes it to:
```d
size_t hashOf(T)(auto ref T val, size_T seed);
size_t hashOf(T)(auto ref T val);
```

This allows the non-"chained" form to be distinguished from the chained form, which avoids unnecessary work for types for which `hashOf` is not a wrapper around `bytesHash`.

The impetus for this was seeing the needless rehashing @IgorStepanov had to perform in https://github.com/dlang/druntime/pull/2243 to make `typeid(T).getHash(&val)` produce the same result as `hashOf(val)`.

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -147,7 +147,9 @@ private template useScopeConstPassByValue(T)
     else static if (is(T == struct) || is(T == union))
     {
         // Overly restrictive for simplicity.
-        enum useScopeConstPassByValue = false;
+        enum useScopeConstPassByValue = T.sizeof <= (int[]).sizeof &&
+            __traits(isPOD, T) && // "isPOD" just to check there's no dtor or postblit.
+            canBitwiseHash!T; // We can't verify toHash doesn't leak.
     }
     else static if (is(T : E[], E))
     {
@@ -157,8 +159,8 @@ private template useScopeConstPassByValue(T)
         else static if (T.length == 0)
             enum useScopeConstPassByValue = true;
         else
-            enum useScopeConstPassByValue = T.sizeof <= size_t.sizeof
-                && .useScopeConstPassByValue!(typeof(T[0]));
+            enum useScopeConstPassByValue = T.sizeof <= (uint[]).sizeof
+                && .useScopeConstPassByValue!(typeof(T.init[0]));
     }
     else static if (is(T : V[K], K, V))
     {
@@ -189,7 +191,21 @@ private template useScopeConstPassByValue(T)
 }
 
 //enum hash. CTFE depends on base type
-size_t hashOf(T)(scope const T val, size_t seed = 0)
+size_t hashOf(T)(scope const T val)
+if (is(T EType == enum) && useScopeConstPassByValue!EType)
+{
+    static if (is(T EType == enum)) //for EType
+    {
+        return hashOf(cast(const EType) val);
+    }
+    else
+    {
+        static assert(0);
+    }
+}
+
+//enum hash. CTFE depends on base type
+size_t hashOf(T)(scope const T val, size_t seed)
 if (is(T EType == enum) && useScopeConstPassByValue!EType)
 {
     static if (is(T EType == enum)) //for EType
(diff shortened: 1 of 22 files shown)


src/core/internal/hash.d
```diff
@@ -30,8 +30,20 @@ private enum floatCoalesceNaNs = true;
         assert(hashOf(double.nan) == hashOf(-double.nan)); // Same hash for different NaN.
 }
 
+private enum hasCallableToHash(T) = __traits(compiles,
+    {
+        size_t x = ((T* x) => (*x).toHash())(null);
+    });
```
n8sh: This PR also contains a relaxation of the required signature of `toHash` for structs & unions. If the type implicitly converts to `size_t` it's fine. This originally was in https://github.com/dlang/druntime/pull/2240 but that PR has not been pulled yet.

ZombineDev:
Let's first merge #2240 and then this one, as I think it would be easier.

n8sh:
Bump.

To resummarize, this PR makes it so that `typeid(x).getHash(&x)` can be `x.toHash()` instead of `hashOf(x.toHash(), 0)`.

src/core/internal/hash.d
```diff
@@ -351,56 +407,52 @@ size_t hashOf(T)(scope const T val, size_t seed = 0) if (!is(T == enum) && __tra
...
-        }
-        else static if (T.sizeof > size_t.sizeof && __traits(isIntegral, T))
-        {
-            static foreach (i; 0 .. T.sizeof / size_t.sizeof)
-                seed = hashOf(cast(size_t) (val >>> (size_t.sizeof * 8 * i)), seed);
-            return seed;
-        }
-        else
-        {
-            return bytesHashAlignedBy!T(toUbyte(val), seed);
```
wilzbach: Where did this go?

src/core/internal/hash.d
```diff
@@ -157,7 +159,7 @@ private template useScopeConstPassByValue(T)
         else static if (T.length == 0)
             enum useScopeConstPassByValue = true;
         else
-            enum useScopeConstPassByValue = T.sizeof <= size_t.sizeof
+            enum useScopeConstPassByValue = T.sizeof <= (uint[]).sizeof
```
wilzbach: Why double this from 8 to 16 (on x86_64)?

src/core/internal/hash.d
```diff
@@ -147,7 +147,9 @@ private template useScopeConstPassByValue(T)
     else static if (is(T == struct) || is(T == union))
     {
         // Overly restrictive for simplicity.
-        enum useScopeConstPassByValue = false;
+        enum useScopeConstPassByValue = T.sizeof <= (int[]).sizeof &&
```
wilzbach: Why do you do this check?

n8sh: If `T.sizeof` is too large pass by value could lead to excessive data copying.

src/core/internal/hash.d
n8sh: This branch was only reached when `__traits(isArithmetic, T) && !__traits(isFloating, T) && !__traits(isIntegral, T)` which doesn't describe any `T`.

wilzbach:
Added "72h no objection -> merge" - anyone with objections should raise their concerns within the next three days.

src/rt/typeinfo/ti_ulong.d
```diff
@@ -27,7 +27,10 @@ class TypeInfo_m : TypeInfo
 
     override size_t getHash(scope const void* p)
     {
-        return hashOf(*cast(ulong*)p);
+        static if (ulong.sizeof <= size_t.sizeof)
+            return *cast(const ulong*)p;
+        else
+            return hashOf(*cast(const ulong*)p);
```
ZombineDev: Could this kind of logic be encapsulated in `hashOf` itself?

ZombineDev: It's not a big deal, but I think that we'll end up with instances of `hashOf` for most (all?) basic types anywa, so I'm unsure if this optimization is worth doing.

wilzbach:
I'm going to make an exception of the 72h merge label as I think we should include this into the 2.082 beta as this release already contains a few other improvements to `hashOf` and thus users will only need to look at their code once in the unlikely chance that they abused some non-supported things. Also no one has raised any concern since this PR was opened (more than a month ago).

PR #2247, opened 2018-07-09
Fix Issue 19072 - Object.toHash and typeid(void*).getHash(&ptr) should be more varied in their low bits

n8sh:
The low bits of `Object.toHash` are insufficiently varied. Depending on the platform the bottom 4, 3, or 2 bits of the result will always be zero. This is bad because the low bits of a hash code are typically the most significant for hashtable implementations. D's builtin AA deals with this and other potential defects by rehashing the hash codes it receives. Some 3rd party hashtable implementations do the same, but others assume that keys with a defined `toHash` have reasonable hashcodes that can be used without need for further mixing.

Fixing this is not hard and not computationally expensive. This PR also does a similar thing for raw pointers. Although it will not always be necessary for them the cost is low and it will frequently be a benefit.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -871,7 +871,15 @@ class Object
     size_t toHash() @trusted nothrow
     {
         // BUG: this prevents a compacting GC from working, needs to be fixed
-        return cast(size_t)cast(void*)this;
+        size_t addr = cast(size_t) cast(void*) this;
+        // The bottom log2((void*).alignof) bits of the address will always
+        // be 0. Moreover it is likely that each Object is allocated with a
+        // separate call to malloc. The alignment of malloc differs from
+        // platform to platform, but rather than having special cases for
+        // each platform it is safe to use a shift of 4. To minimize
+        // collisions in the low bits it is more important for the shift to
+        // not be too small than for the shift to not be too big.
+        return addr ^ (addr >>> 4);
     }
 
     /**
@@ -1183,7 +1191,12 @@ class TypeInfo_Pointer : TypeInfo
 
     override size_t getHash(scope const void* p) @trusted const
     {
-        return cast(size_t)*cast(void**)p;
+        // If the target of `p` is n-byte aligned then the
+        // bottom log2(n) bits of `p` will always be 0.
+        // Performing a constant magnitude xor-shift is
+        // inexpensive and will generally pay off.
+        size_t addr = cast(size_t) *cast(const void**) p;
+        return addr ^ (addr >>> 4);
     }
 
     override bool equals(in void* p1, in void* p2) const


jacob-carlborg:
It would be nice to have the PR description in the commit message as well. If a PR consists of a single commit, the first line (up to an empty newline) of the commit message will be used as the title for the PR, the rest will be used as the description for the PR.

n8sh:
Added PR description to the commit message.

n8sh:
BTW the original version of this PR used different shift sizes for different known malloc alignments. I removed this because it seemed to me like overengineering, and taking a step back it wasn't clear if in practice -- aside from the toy problem of placing into a tiny hash table keys that all have minimum `classInstanceSize` and are allocated in sequence -- there would be a benefit compared to a constant xor-shift of 4.

<details>
<summary>Discarded version</summary>

```d
size_t toHash() @trusted nothrow
{
    const addr = cast(size_t) cast(void*) this;
    // The bottom log2((void*).alignof) bits of the address will
    // always be 0. Moreover it is likely that each Object is
    // allocated with a separate call to malloc.
    enum log2PtrAlign = (size_t.sizeof >= 8 || (void*).alignof >= 8) ? 3 : 2;
    version(ARM) { enum isARM32 = true; } else { enum isARM32 = false; }
    version(MIPS32) { enum isMIPS32 = true; } else { enum isMIPS32 = false; }

    // 16 byte aligned malloc:
    version (OSX)
        return addr ^ (addr >>> 4);
    else version (CRuntime_Musl)
        return addr ^ (addr >>> 4);
    else version (OpenBSD)
        return addr ^ (addr >>> 4);
    else version (DragonFlyBSD)
        return addr ^ (addr >>> 4);
    // (size_t.sizeof * 2) aligned malloc:
    else version (Windows)
        return addr ^ (addr >>> (log2PtrAlign + 1));
    else version (CRuntime_Glibc)
        return addr ^ (addr >>> (log2PtrAlign + 1));
    // (size_t.sizeof) aligned malloc:
    else version (CRuntime_UClibc)
        return addr ^ (addr >>> log2PtrAlign);
    // NetBSD & FreeBSD:
    else version (NetBSD)
        return addr ^ (addr >>> (4 - uint(isARM32)));
    else version (FreeBSD)
        return addr ^ (addr >>> (4 - uint(isARM32 || isMIPS32)));
    // Otherwise we only know that the log2(size_t.sizeof) low bits
    // are always zero. We shift 1 more than this because shifting
    // by too little will cause every hash to be even, while shifting
    // by "too much" is not particularly bad.
    else
        return addr ^ (addr >>> (log2PtrAlign + 1));
}
```
</details>

wilzbach:
Should we do something about the 16 zero top bytes on 64-bit architectures?
Maybe sth. like `k << 16 & (size_t.max << 48)`?

n8sh:
@wilzbach I don't think so. An AA that used power-of-2 sized tables and didn't rehash keys would need to have a capacity of more than `2^^48` before the top 16 bits mattered at all.

src/object.d
```diff
@@ -1183,7 +1191,12 @@ class TypeInfo_Pointer : TypeInfo
 
     override size_t getHash(scope const void* p) @trusted const
     {
-        return cast(size_t)*cast(void**)p;
+        // If the target of `p` is n-byte aligned then the
+        // bottom log2(n) bits of `p` will always be 0.
+        // Performing a constant magnitude xor-shift is
+        // inexpensive and will generally pay off.
+        size_t addr = cast(size_t) *cast(const void**) p;
+        return addr ^ (addr >>> 4);
```
ZombineDev: This changes sits at odds with https://github.com/dlang/druntime/pull/2243/files#diff-a68e58fcf0de5aa198fcaceafe4e8cf9R1186. Would hashing the pointer address produce better distribution?

n8sh: >Would hashing the pointer address produce better distribution

Possibly. But let's remember the only goal in improving the distribution is to increase AA speed by reducing hash collisions and the speed penalty they incur. Rehashing adds to the time of every hashtable operation so if rehashing takes too long it becomes a net detriment. [In a recent D hash table implementation comparison](https://forum.dlang.org/post/zwkpeeduwrxugoevpmoa@forum.dlang.org) `tanya.container.hashtable` with default settings was severely hurt by the rehashing it performs on integer keys.

ZombineDev: Makes sense. Let's go with your approach.

PR #2248, opened 2018-07-11
Fix Issue 19075 - rt.util.random.Rand48.defaultSeed should prefer RDTSC or mach_absolute_time to ctime.time

n8sh:
`rt.util.random.Rand48.defaultSeed` generates a seed from the clock using `ctime.time`. With `D_InlineAsm_X86_64` it would be better to use the RDTSC instruction because it is much faster as well as being finer precision. Finer precision is a benefit not because we care about accurate time, but because the differences between seeds generated this way are potentially greater. On macOS `mach_absolute_time` is also faster than `ctime.time`. The Microsoft Windows equivalent is `QueryPerformanceCounter` but I haven't benchmarked it to verify that it is faster than `ctime.time`.

This PR doesn't use RDTSC with `D_InlineAsm_X86` because some x86 processors don't support RDTSC and because on x86 (but not x86-64) Linux `prctl` can disable a process's ability to use RDTSC.


jacob-carlborg:
FYI I don’t think there are any supported Macs that aren’t 64bit.

n8sh:
I was thinking about macOS on AArch64, which might not be officially supported but is a plausible target.

jacob-carlborg:
> I was thinking about macOS on AArch64, which might not be officially supported but is a plausible target.

Well, the comment mentions x86 :wink:.

n8sh:
Ah - that comment wasn't supposed to be about the following `version(OSX)` block, but to document a non-existent `version(D_InlineAsm_X86)` block. But since that block doesn't exist of course the comment appears to pertain to the next block that actually exists. I'll reformat it to avoid this confusion.

jacob-carlborg:
Makes more sense.

src/rt/util/random.d
```diff
@@ -12,10 +12,47 @@ struct Rand48
...
+        else version(OSX)
+        {
+            // mach_absolute_time is much faster than ctime.time.
+            import core.time : mach_absolute_time;
+            rng_state = mach_absolute_time();
+            popFront();
+        }
+        else
+        {
+            // Old version using timestamp in seconds:
```
ibuclaw: Old version?

n8sh: Current version.

ibuclaw: Perhaps you mean libc or fallback version then. Old version sounds like you intend to remove it.

n8sh: Changed comment as suggested.

n8sh:
Confirmed `QueryPerformanceCounter` is faster so the PR now uses that on Microsoft Windows.

wilzbach:
(rebased to upstream/master to fix the CircleCi failure)

wilzbach:
It only merges if all CI pass (that behavior was changed a few months ago). And unfortunately we currently are still having problems with Jenkins, but at least when Buildkite is passing you know that the failure is spurious.

PR #2249, opened 2018-07-11
Handle page protection errors as D errors on Linux with --DRT-memoryError=1

wilzbach:
See also: https://forum.dlang.org/post/pi31ab$me3$1@digitalmars.com

### DIFF
--- a/changelog/trap_memory_error.dd
+++ b/changelog/trap_memory_error.dd
@@ -1,0 +1,38 @@
+Page protection error handling can now be registered via `--DRT-memoryError=1`
+
+In environments where attaching a debugger or retrieving a core dump isn't possible or hard,
+on Linux x86_64 one has always been able to attach druntime's  memory handler:
+
+---
+shared static this()
+{
+    import etc.linux.memoryerror : registerMemoryErrorHandler;
+    registerMemoryErrorHandler;
+}
+
+void main() {
+    int* ptr = null;
+    (*ptr)++;
+}
+---
+
+And thus retrieved the full segfault as D `Error` on the command-line:
+
+$(CONSOLE dmd -g -run test.d
+etc.linux.memoryerror.NullPointerError@src/etc/linux/memoryerror.d(325)
+$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)
+??:? void etc.linux.memoryerror.sigsegvUserspaceProcess(void*) [0x8353df09]
+??:? void etc.linux.memoryerror.sigsegvDataHandler() [0x8353de4a]
+test.d:9 _Dmain [0x8353c2b5]
+)
+
+With this release, instead of recompiling the binary and adding the `registerMemoryError`,
+one can use the runtime flag `--DRT-memoryError=1` to activate druntime's memory error handling.
+
+$(CONSOLE dmd -g test.d && ./test --DRT-memoryError=1
+etc.linux.memoryerror.NullPointerError@src/etc/linux/memoryerror.d(325)
+$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)
+??:? void etc.linux.memoryerror.sigsegvUserspaceProcess(void*) [0x8353df09]
+??:? void etc.linux.memoryerror.sigsegvDataHandler() [0x8353de4a]
+test.d:2 _Dmain [0x8353c2b5]
+)
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -441,6 +441,14 @@ private extern (C) int _d_run_main2(char[][] args, size_t totalArgsLength, MainF
 
     auto useExceptionTrap = parseExceptionOptions();
 
+    // Handle page protection errors using D errors (exceptions) when supported
+    version (linux)
+    {
+        import etc.linux.memoryerror : registerMemoryErrorHandler;
+        if (parseMemoryErrorOptions())
+            registerMemoryErrorHandler();
+    }
+
     version (Windows)
     {
         if (IsDebuggerPresent())
@@ -552,6 +560,18 @@ private auto parseExceptionOptions()
     const optName = "trapExceptions";
     auto option = rt_configOption(optName);
     auto trap = rt_trapExceptions;
+    if (option.length)
+        rt_parseOption(optName, option, trap, "");
+    return trap;
+}
+
+private auto parseMemoryErrorOptions()
+{
+    import rt.config : rt_configOption;
+    import core.internal.parseoptions : rt_parseOption;
+    enum optName = "memoryError";
+    auto option = rt_configOption(optName);
+    bool trap;
     if (option.length)
         rt_parseOption(optName, option, trap, "");
     return trap;
--- a/test/exceptions/Makefile
+++ b/test/exceptions/Makefile
@@ -4,7 +4,7 @@ TESTS=stderr_msg unittest_assert invalid_memory_operation unknown_gc static_dtor
 	  future_message refcounted rt_trap_exceptions_drt catch_in_finally
 
 ifeq ($(OS)-$(BUILD),linux-debug)
-	TESTS+=line_trace long_backtrace_trunc rt_trap_exceptions
+	TESTS+=line_trace long_backtrace_trunc rt_trap_exceptions rt_trap_memory_error
 	LINE_TRACE_DFLAGS:=-L--export-dynamic
 endif
 ifeq ($(OS),linux)
@@ -67,6 +67,9 @@ $(ROOT)/catch_in_finally.done: STDERR_EXP="success."
 $(ROOT)/rt_trap_exceptions.done: STDERR_EXP="object.Exception@src/rt_trap_exceptions.d(12): this will abort"
 $(ROOT)/rt_trap_exceptions.done: STDERR_EXP2="src/rt_trap_exceptions.d:8 main"
 $(ROOT)/assert_fail.done: STDERR_EXP="success."
+$(ROOT)/rt_trap_memory_error.done: RUN_ARGS="--DRT-memoryError=1"
+$(ROOT)/rt_trap_memory_error.done: STDERR_EXP="etc.linux.memoryerror.NullPointerError@src/etc/linux/memoryerror.d"
+
 $(ROOT)/%.done: $(ROOT)/%
 	@echo Testing $*
 	$(QUIET)$(TIMELIMIT)$(ROOT)/$* $(RUN_ARGS) 2>$(ROOT)/$*.stderr || true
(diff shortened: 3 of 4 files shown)


Geod24:
This change could use a bit more rationale. Is there a way to disable it ? Do we really want that ?

adamdruppe:
Null pointers on win32 also throw errors btw.

adamdruppe:
Though while I'm not voting no, I do object to calling segfaults bad behavior. They do exactly what they are supposed to do - and exactly what a D error is allowed to do according to our spec - and have superior debugging capabilities. (I'd say superior handling capabilities too, you can handle-and-resume or retry a segfault, not so much a D exception, but we don't really lose that since you can just register your own signal handler instead.)

The exception isn't bad, but neither is the segfault, you just need to know how to actually use it and then you see the advantages.

nemanja-boric-sociomantic:
One good thing about segmentation fault is that you can generate core dump which you could inspect to see what caused the segmentation fault very easily. Is one still able to do that with this approach? If I'm not mistaken, the position of the segmentation fault is removed from the backtrace.

nemanja-boric-sociomantic:
Sorry, I read the stacktrace in a wrong way :man_facepalming:

JohanEngelen:
This feels very much rushed after minimal discussion on the NG. There is even a post that mentions that in the past this was debated, so at least I expect some arguments both ways and with a conclusion of why it now should come out different than before.

Did you test how this works with debuggers and ASan and other tools?

JohanEngelen:
My red flag here is that with this PR _per default_ after a page error, more code is executed that even allocates (`new`), uses D's exception mechanism, calls constructors of class objects, etc. ...... That sounds problematic to me.

Shachar:
I second the detractors.
The Linux segmentation fault behavior is a well known capabilities with established ways of handling it. Turning it into an exception makes zero sense to me.

At the very least, it should be possible to turn this capability off.

adamdruppe:
A few notes:

* you can turn this off by calling a function (just like right now, you can turn it on by calling the function). Maybe we should make it a DRT command line argument too. The implementation btw is it is just registers a signal handler with `sigaction` http://dpldocs.info/experimental-docs/source/etc.linux.memoryerror.d.html#L33 ). If running in a debugger, just like how we want uncaught exceptions to go through the debugger, these should too.

* Executing more code after it doesn't actually bother me since the segfault interrupts the illegal operation... it is perfectly normal to handle and resume with a page fault.


This is why I could go either way on this. I like things the way they are now, but I could live with this change too. (or a compromise: print the stack trace to the console, then abort the program, but I think that has the same downsides to this basically anyway)

Shachar:
After having slept on this, I have to raise the level of my objection to this change.

Before I begin, I'll point out that I've tried to follow the code for `registerMemoryErrorHandler`, but I'm not sure what the playing around with the variables called `RIP` actually do. As such, it might be that this objection is incorrect.

With that in mind, please keep in mind that the program might have been doing *anything* when it segfaulted. It might have been in the middle of a `throw`. It might have been in the middle of a context switch. It might have been mucking around with variables that outside `scope(failure)`s might be relying on. It might have been in the middle of performing five separate program lines, interleaved by the compiler, all partially completed.

Letting the program flow continue, even as a an exception, is just too likely to cause further problems. These will mask the original problem, or might even drive the program into an infinite loop of crashing, trying to throw, crashing, trying to throw.

As such, I must re-iterate how bad an idea I think it is to turn this on by default.

wilzbach:
Alright, I was just trying to follow up on the NG thread.
Thanks a lot for all your input and I will close this soon. Really appreciated!

However, we definitely should improve the documentation and I will keep it open until then, s.t. I don't forget this.

jacob-carlborg:
What about turning it on by default when the `-debug` flag is used?

ibuclaw:
-1 on the grounds that the Linux memory error module is both dmd and x86 centric, and should be improved by some measure first so that it be more useable before making it a druntime dependency.

Though that is a purely technical reason only.

Shachar:
I think we should leave it as is. It should be documented, along with the possible pitfalls it might bring. If anyone wants it, they can add it to their code.

I don't think it should be on by default. Ever.

schveiguy:
I don't know the original reasons why this isn't the default, or why it's not supported on other platforms. What I do know, is that I enable it on Linux in any server program that I write, and it works swimmingly well.

Yes, a segfault can be useful in certain circumstances. But a stack trace is eminently easier to use for diagnostics, especially when you aren't in control of the environment. I can tell just by reading the stack trace where the problem has occurred. It may require more in-depth analysis, and perhaps using a debugger, or dumping a core file. But it also could be a "oops! I forgot to initialize that thing!" type of bug, in which case, I need nothing more than an editor to fix, and I'm done in 2 minutes. Instead of digging out the exact binary (hopefully you saved all those), loading into a debugger, hoping that the customer has core dumps turned on, hoping that they didn't just delete the file, hoping their disk has enough space to store it, etc., etc.

In regards to "different behavior on different platforms", my understanding is that Windows ALWAYS generated a stack trace for seg faults. There is plenty of precedent here.

It definitely makes sense to provide runtime switches to turn it off or on (given that this would enable it before e.g. any static ctors are run), but I still believe the default should be on. I'm not sure of the harm here, we should be exploiting the most we can from the platform in terms of providing diagnostic information.

schveiguy:
BTW, here is the original PR, you can read a LOT of comments in there: https://github.com/dlang/druntime/pull/187

wilzbach:
So I came back to this PR and it looks like I already did the required work here, i.e. adding a flag to druntime (`--DRT-memoryError=1`) which allows users to enable this option _without_ needing to recompile their binaries (see the changelog entry of this PR).

Anything else that we want to do here?

CC @thewilsonator

thewilsonator:
I think its fine, and a lot of the previous criticisms were about it being a default, but I still like to get some more feedback on this (but don't let me forget about it!).

schveiguy:
Since my comments on this, I have found one place where this really sucks. If you encounter a segfault inside a destructor being called by the GC, then the attempt to `new` an `Error` causes an invalid memory operation. It took me a while to track that down, as the information printed there is totally useless.

Before making this the default (I think reading the code, this is not the default, if it was before), I think we should make the memory error not depend on the GC. In fact, if there is a way to print the stack trace and abort without unwinding anything, that would be ideal, and much much safer.

schveiguy:
> Don't we have a buffer somewhere for statically allocated exceptions?

It's done elsewhere, but not for this extension. Would be a useful change, even if this PR isn't accepted.

wilzbach:
> Before making this the default (I think reading the code, this is not the default, if it was before),

Yep no longer the default. This PR was down-graded to just adding the CLI-equivalent of `registerMemoryErrorHandler`.

> Don't we have a buffer somewhere for statically allocated exceptions?

See e.g. https://github.com/dlang/druntime/pull/1710

changelog/trap_memory_error.dd
```diff
@@ -0,0 +1,38 @@
+Page protection error handling can now be registered via `--DRT-memoryError=1`
+
+In environments where attaching a debugger or retrieving a core dump isn't possible or hard,
+on Linux x86_64 one has always been able to attach druntime's  memory handler:
```
jacob-carlborg: The code in the `etc.linux.memoryerror` module is defined for both 32 bit and 64 bit.

src/rt/dmain2.d
```diff
@@ -579,6 +587,18 @@ private auto parseExceptionOptions()
     return trap;
 }
 
+private auto parseMemoryErrorOptions()
+{
+    import rt.config : rt_configOption;
+    import core.internal.parseoptions : rt_parseOption;
+    const optName = "memoryError";
```
jacob-carlborg: `enum`

src/rt/dmain2.d
```diff
@@ -579,6 +587,18 @@ private auto parseExceptionOptions()
...
+    auto option = rt_configOption(optName);
```
jacob-carlborg: `const`?

jacob-carlborg:
There's no test for what happens when this flag is passed on a platform that doesn't support it.

schveiguy:
Ugh, looking again at `staticError`, it appears that it suppresses the stack trace. Which is completely useless. Same issue as I had with `InvalidMemoryOperation` that I mentioned earlier.

Somehow, we need to fix this problem. However, that shouldn't hold up this PR. All it is doing is exposing the linux memory error registration as a DRT flag.

schveiguy:
> There's no test for what happens when this flag is passed on a platform that doesn't support it.

In general, unsupported or unknown DRT flags are going to be ignored. For example, if you passed this flag on an older version of druntime, nothing will happen.

jacob-carlborg:
> In general, unsupported or unknown DRT flags are going to be ignored. For example, if you passed this flag on an older version of druntime, nothing will happen.

This is slightly different in that it's only available on some platforms. But it might not be worth testing.

jacob-carlborg:
> Ugh, looking again at staticError, it appears that it suppresses the stack trace. 

Technically it's not the `staticError` function but rather all Errors that use `staticError` that sets the `info` instance variable to an object that will suppress the stack trace.

> Somehow, we need to fix this problem

Looking at `core.runtime.defaultTraceHandler`, there are two things stopping this from being `@nogc`: calling `core.memory.gc_inFinalizer` and creating an instance of `DefaultTraceInfo`. The latter can be fixed with again with a static allocation, not sure what to do about `gc_inFinalizer`.

schveiguy:
`gc_inFinalizer` actually doesn't require GC since it's now lazily initialized: https://github.com/dlang/druntime/blob/01daddfa7a49a5e86347c08e741ff9095978ae57/src/gc/impl/proto/gc.d#L234

schveiguy:
We'd have to mark `inFinalizer` as `@nogc` in the interface, but that shouldn't be a problem, as the implementation even with a GC doesn't require doing any GC-ish things.

jacob-carlborg:
> We'd have to mark inFinalizer as @nogc in the interface, but that shouldn't be a problem, as the implementation even with a GC doesn't require doing any GC-ish things.

Yes, exactly.

I was looking into the allocation of `DefaultTraceInfo`, the question is: does it need to return a new instance every time or can it return the same pre-allocated instance?

schveiguy:
> does it need to return a new instance every time or can it return the same pre-allocated instance?

I think you need a unique instance per exception. But if we are preallocating the exception, we can preallocate the trace info as well. Looks like the big part of the trace info is the stack frame array. The way the default trace info is stored will have to change. Looks like it's an inner class, but it doesn't need to be.

jacob-carlborg:
> I think you need a unique instance per exception.

The way `staticError` currently you can only have one exception anyway. It will always overwrite the buffer.

src/rt/dmain2.d
wilzbach: Can't pass that to `rt_parseOption` then.

RazvanN7:
@schveiguy @wilzbach @thewilsonator I have rebased this. Should we merge it?

ibuclaw:
> -1 on the grounds that the Linux memory error module is both dmd and x86 centric, and should be improved by some measure first so that it be more useable before making it a druntime dependency.
> 
> Though that is a purely technical reason only.

Still a -1, this time because it would add a circular dependency between druntime and phobos.

PR #2250, opened 2018-07-11
Fix test/exceptions test: the exceptions test wasn't checking for the exception.

wilzbach:


wilzbach:
Included as part of https://github.com/dlang/druntime/pull/2245

PR #2252, opened 2018-07-17
Fix Issue 19090 - core.internal.hash.bytesHash unit test uses incorrect test vector on BigEndian machines

n8sh:


src/core/internal/hash.d
```diff
@@ -663,7 +663,7 @@ pure nothrow @system @nogc unittest
     version(BigEndian)
     {
         const ubyte[7] a = [99, 4, 3, 2, 1, 5, 88];
-        const uint[2] b = [0x01_02_03_04, 0x05_ff_ff_ff];
+        const uint[2] b = [0x04_03_02_01, 0x05_ff_ff_ff];
```
wilzbach: Shouldn't `0x05_ff_ff_ff` be swapped then too?

n8sh: 5 bytes are fed into `bytesHash`. `a` is read skipping the first byte, and `b` is read from its start. Both 5-byte sequences should be the same.

The second integer _is_ reversed compared to the little-endian version, which is:
```d
const uint[2] b = [0x04_03_02_01, 0xff_ff_ff_05];
```
The error was that the first integer _shouldn't_ have been reversed, due to a detail of our MurmurHash3 implementation. It reads 4-integer blocks as integers interpreted according to native word order, after which it reads the remaining 0 to 3 bytes in a machine-independent fashion.

wilzbach:
As there's no CI which tests for BigEndian, I think it would be best to get someone who has access to a big endian CI on the review list for this.
Do GDC (@ibuclaw) or LDC (@kinke) support backend with BigEndian?

kinke:
We support big-endian platforms, but I don't have access to a test machine. Pinging @redstar as he may be able to test this on PowerPC.

ibuclaw:
Yes, gdc does. Library support is another thing though.

thewilsonator:
In the interest of moving things forward and @n8sh comment about our MurmurHash3 implementation I'm going to merge this. If it still isn't correct the issue can be reopened.

PR #2253, opened 2018-07-18
Fix Issue 19092 - __delete doesn't work with immutable

wilzbach:


src/core/memory.d
```diff
@@ -1024,6 +1025,26 @@ void __delete(T)(ref T x) @system
     }
 }
 
+/// ditto
+void __delete(T)(ref const(T)[] x) @system
+if (is(const(T) : T))
+{
+    import core.internal.traits : Unqual;
+    auto x2 = cast(Unqual!T[]) x;
+    __delete(x2);
```
wilzbach: We could also do `GC.free(cast(void*) x);` directly here.

src/core/memory.d
```diff
@@ -1024,6 +1025,26 @@ void __delete(T)(ref T x) @system
     }
 }
 
+/// ditto
+void __delete(T)(ref const(T)[] x) @system
+if (is(const(T) : T))
+{
+    import core.internal.traits : Unqual;
+    auto x2 = cast(Unqual!T[]) x;
```
jacob-carlborg: Why not doing this in the mutable overload?

wilzbach: Fair point!

jacob-carlborg:
Although the issue doesn't mention this, but it doesn't work for const/immutable classes, interfaces or pointers as well.

wilzbach:
> Although the issue doesn't mention this, but it doesn't work for const/immutable classes, interfaces or pointers as well.

Yes, but `delete` didn't support this either: https://run.dlang.io/is/gKbcrq

wilzbach:
Failures are unrelated :/

PR #2255, opened 2018-07-27
-betterC: Add support for types when comparing arrays

JinShil:
Implementation for https://github.com/dlang/dmd/pull/8521

This makes `dstrcmp` a template so it can be used from `object.__cmp` when compiling with -betterC.

See https://github.com/dlang/druntime/blob/cb5efa9854775c5a72acd6870083b16e5ebba369/src/object.d#L63

We don't have good testing infrastructure for -betterC in druntime.  I'm hoping reviewers can test locally by pulling down the druntime and dmd PRs and running `test/run.d test/runnable/betterc.d` from the dmd repository.

@wilzbach This PR pair is a good example of the kinds of things that we don't have good test infrastructure for.

### DIFF
--- a/src/core/internal/string.d
+++ b/src/core/internal/string.d
@@ -212,7 +212,7 @@ unittest
     static assert(!__traits(compiles, 100.numDigits!37()));
 }
 
-int dstrcmp( scope const char[] s1, scope const char[] s2 ) @trusted
+int dstrcmp()( scope const char[] s1, scope const char[] s2 ) @trusted
 {
     immutable len = s1.length <= s2.length ? s1.length : s2.length;
     if (__ctfe)


wilzbach:
> @wilzbach This PR pair is a good example of the kinds of things that we don't have good test infrastructure for.

Aware, as discussed on Slack in theory you could create a branch at dmd with the tests, then create a branch here and open a PR targeting that branch, and after approval and merge, eventually merge it back to master, but that's a lot of overhead :/

Thus, my best suggestion is to move druntime into DMD, the compiler depends on it anyhow and it just makes life for developers a lot harder. Thinking about it, it shouldn't be too hard to keep the git history information and lock druntime.
A potential migration could look like this:
- lock druntime temporarily (e.g. by setting it in "Archive" mode or only allowing admins to merge)
- move source code to dmd
- update all CIs and build scripts
- set druntime to "Archive" mode permanently

This would mean that for a week or so, we couldn't merge things into druntime, but looking at the current velocity with which druntime is moving, this shouldn't be a problem.
I will start a thread to check for other people's opinions.

wilzbach:
> I will start a thread to check for other people's opinions.

https://forum.dlang.org/post/wiqhgzmotcmvgclxxvmw@forum.dlang.org

ZombineDev:
Regardless of the outcome of the discussion this change shouldn't be blocked as it's easy enough to review.

ZombineDev:
The libmir and mecca failures on Buildkite appear on other PRs, so I don't think they are related.

PR #2256, opened 2018-07-28
add changelog for PR #2243

IgorStepanov:
add changelog for PR #2243


IgorStepanov:
>A changelog entry should have a short header, an empty line and then the description.

Is it ok now?

PR #2257, opened 2018-07-29
core.internal.convert: add IEEE Quadruple support

joakim-noah:
This is a WIP pull, as I'm going to add another commit that reworks how druntime initializes TLS data on Android before passing it to the GC, based on a suggestion of Martin from the ldc team. I wanted to get this first math commit up now to run it through the CI and review, as this one should be done.

I've tested this commit with ldc 0.17 natively on Android/AArch64 and linux/AArch64, and by cross-compiling with ldc 1.11 master for Android/AArch64 (all tests build and pass, but only because they're all truncated at 80-bit precision), along with making sure it didn't break anything on linux/x64. With this patch, all the druntime unit tests finally compile and pass on AArch64.

### DIFF
--- a/src/core/internal/convert.d
+++ b/src/core/internal/convert.d
@@ -57,14 +57,32 @@ const(ubyte)[] toUbyte(T)(const ref T val) if(is(Unqual!T == float) || is(Unqual
         ubyte[] buff = ctfe_alloc(T.sizeof);
         size_t off_bytes = 0;
         size_t off_bits  = 0;
+        // Quadruples won't fit in one ulong, so check for that.
+        enum mantissaMax = FloatTraits!T.MANTISSA < ulong.sizeof*8 ?
+                           FloatTraits!T.MANTISSA : ulong.sizeof*8;
 
-        for(; off_bytes < FloatTraits!T.MANTISSA/8; ++off_bytes)
+        for(; off_bytes < mantissaMax/8; ++off_bytes)
         {
             buff[off_bytes] = cast(ubyte)mantissa;
             mantissa >>= 8;
         }
-        off_bits = FloatTraits!T.MANTISSA%8;
-        buff[off_bytes] = cast(ubyte)mantissa;
+
+        static if (floatFormat!T == FloatFormat.Quadruple)
+        {
+            ulong mantissa2 = parsed.mantissa2;
+            off_bytes--; // go back one, since mantissa only stored data in 56
+                         // bits, ie 7 bytes
+            for(; off_bytes < FloatTraits!T.MANTISSA/8; ++off_bytes)
+            {
+                buff[off_bytes] = cast(ubyte)mantissa2;
+                mantissa2 >>= 8;
+            }
+        }
+        else
+        {
+            off_bits = FloatTraits!T.MANTISSA%8;
+            buff[off_bytes] = cast(ubyte)mantissa;
+        }
 
         for(size_t i=0; i<FloatTraits!T.EXPONENT/8; ++i)
         {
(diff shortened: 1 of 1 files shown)


src/core/internal/convert.d
```diff
@@ -106,8 +124,8 @@ private Float parse(bool is_denormalized = false, T)(T x) if(is(Unqual!T == iflo
 private Float parse(bool is_denormalized = false, T:real)(T x_) if(floatFormat!T != FloatFormat.Real80)
 {
     Unqual!T x = x_;
-    assert(floatFormat!T != FloatFormat.DoubleDouble && floatFormat!T != FloatFormat.Quadruple,
-           "doubledouble and quadruple float formats are not supported in CTFE");
+    static assert(floatFormat!T != FloatFormat.DoubleDouble,
```
joakim-noah: This `assert` was never actually triggered at compile-time.

src/core/internal/convert.d
```diff
@@ -238,14 +281,14 @@ private template FloatTraits(T) if(floatFormat!T == FloatFormat.DoubleDouble) //
...
+private template FloatTraits(T) if(floatFormat!T == FloatFormat.Quadruple)
 {
     enum EXPONENT = 15;
     enum MANTISSA = 112;
     enum ZERO     = Float(0, 0, 0);
     enum NZERO    = Float(0, 0, 1);
-    enum NAN      = Float(-1, 0x7fff, 0);
-    enum NNAN     = Float(-1, 0x7fff, 1);
+    enum NAN      = Float(0, 0x7fff, 0, 0x80000000000000UL);
+    enum NNAN     = Float(0, 0x7fff, 1, 0x80000000000000UL);
```
joakim-noah: I didn't bother adding `0`'s everywhere else the `Float` constructor was called, as this is the only 112-bit mantissa where the most significant bits in `mantissa2` needed to be set.

ibuclaw:
This is a poorly named pr. As this isn't android specific. You infact fix all platforms that have 128bit IEEE reals.

src/core/internal/convert.d
```diff
@@ -323,14 +366,42 @@ private ulong denormalizedMantissa(T)(T x) if(floatFormat!T == FloatFormat.Real8
...
+    if(floatFormat!T == FloatFormat.Float || floatFormat!T == FloatFormat.Double)
 {
     x *= 2.0L^^FloatTraits!T.MANTISSA;
     auto fl = parse!true(x);
     ulong mant = fl.mantissa >> (FloatTraits!T.MANTISSA - fl.exponent);
     return shiftrRound(mant);
 }
 
+@safe pure nothrow @nogc
+private Float denormalizedMantissa(T)(T x, uint sign) if(floatFormat!T == FloatFormat.Quadruple)
```
ibuclaw: This doesn't overload the above functions and so should have a distinct name, ie: `floatDenormalizedMantissa`.

Having a look at all uses of the other function, they are all `Float(denormalizedMantissa(x), 0, sign)`, so they all could be refactored to return a Float as well as sharing the same signature.

ibuclaw:
@joakim-noah ping

src/core/internal/convert.d
joakim-noah: Good point, I'll fix it.

joakim-noah:
I want to get back to testing that second commit, but busy with other stuff first. Will get back to the second commit next week probably, and once that's up, this pull should be ready to go. If you've any further feedback on this commit, we can discuss it in the meantime.

joakim-noah:
Btw, is you pinging me now in any way related to the current gcc submission? If so, feel free to use and modify this patch in any way you want to there: I'm making it available under the same Boost license as the rest of druntime.

ibuclaw:
This one patch is actually blocking all ports that have 128-bit reals from being buildable. So this a high priority fix IMO. Your other patch can wait.

joakim-noah:
Fine with me, added another commit with the change you asked for. Feel free to squash the two and merge once the CI passes.

joakim-noah:
Looks like the FreeBSD64 part of the auto-tester hung, so squashed the commits to get it to restart.

joakim-noah:
CI passed, ready to go.

src/core/internal/convert.d
```diff
@@ -314,21 +354,49 @@ private uint binLog2(T)(const T x)
 }
 
 @safe pure nothrow @nogc
-private ulong denormalizedMantissa(T)(T x) if(floatFormat!T == FloatFormat.Real80)
+private Float denormalizedMantissa(T)(T x, uint sign) if(floatFormat!T == FloatFormat.Real80)
```
ibuclaw: I don't suppose these could be named `floatDenormalizedMantissa` or `floatWithDenormalizedMantissa` to make it more clear from the caller what they return.

Not too fussed either way.

ibuclaw:
Bikeshed nit, but otherwise LGTM.

src/core/internal/convert.d
joakim-noah: Private function only used in this module at CTFE, don't think it's worth kicking off the CI again for this trivial rename.

PR #2258, opened 2018-07-30
Fix issue 19128 - argument to alloca may be too large

belka-ew:
In rt.arrayassign._d_arraysetassign alloca is called if the buffer should be allocated is larger than 16 bytes:

```d
void[16] buf = void;
void[] tmp;
if (element_size > buf.sizeof)
{
    tmp = alloca(element_size)[0 .. element_size];
}
else
    tmp = buf[];
```

It is dangerous since alloca unavailable to allocate causes undefined behaviour, so the alloca man page states:

> The alloca() function returns a pointer to the beginning of the allocated space.  If the allocation causes stack overflow, program behavior is undefined.

See related discussion:
https://github.com/D-Programming-GDC/GDC/pull/699

### DIFF
--- a/src/rt/arrayassign.d
+++ b/src/rt/arrayassign.d
@@ -30,8 +30,25 @@ extern (C) void[] _d_arrayassign(TypeInfo ti, void[] from, void[] to)
 
     // Need a temporary buffer tmp[] big enough to hold one element
     void[16] buf = void;
-    void* ptmp = (elementSize > buf.sizeof) ? alloca(elementSize) : buf.ptr;
-    return _d_arrayassign_l(ti, from, to, ptmp);
+    void* ptmp;
+    if (elementSize <= buf.sizeof)
+    {
+        ptmp = buf.ptr;
+    }
+    else if (elementSize < 1024)
+    {
+        ptmp = alloca(elementSize);
+    }
+    else
+    {
+        ptmp = malloc(elementSize);
+    }
+    auto dst = _d_arrayassign_l(ti, from, to, ptmp);
+    if (elementSize >= 1024)
+    {
+        free(ptmp);
+    }
+    return dst;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/rt/arrayassign.d
```diff
@@ -30,8 +30,25 @@ extern (C) void[] _d_arrayassign(TypeInfo ti, void[] from, void[] to)
 
     // Need a temporary buffer tmp[] big enough to hold one element
     void[16] buf = void;
-    void* ptmp = (elementSize > buf.sizeof) ? alloca(elementSize) : buf.ptr;
-    return _d_arrayassign_l(ti, from, to, ptmp);
+    void* ptmp;
+    if (elementSize <= buf.sizeof)
+    {
+        ptmp = buf.ptr;
```
iain-buclaw-sociomantic: I would argue here that if we are going to fall back to malloc. Then having a pre-allocated buffer is pointless.

This should be:
```
if (elementSize <= maxAllocaSize)
  ptmp = alloca(elementSize);
else
  ptmp = malloc(elementSize);
```

JinShil:
This is an excellent candidate for migration to a template.  @belka-ew, how confident are you in modifying DMD to lower assignment expressions to something like `void __ArrayAssign(TFrom, TTo)(const TFrom[] from, TTo[] to)`?  All this size checking could be done at compile-time.

belka-ew:
> how confident are you in modifying DMD to lower assignment expressions to something like `void __ArrayAssign(TFrom, TTo)(const TFrom[] from, TTo[] to)`?

At least I could try. This PR has to wait then.

src/rt/arrayassign.d
```diff
@@ -30,8 +30,25 @@ extern (C) void[] _d_arrayassign(TypeInfo ti, void[] from, void[] to)
...
+    }
+    else if (elementSize < 1024)
```
JinShil: How was 1024 chosen?

belka-ew: Pretty random. Should be enough for a lot of data types, and not too much.

JinShil:
I recently noticed that DMD's `__alloca` implementation [has a stack overflow check](https://github.com/dlang/druntime/blob/9a8edfb48e4842180c706ee26ebd8edb10be53f4/src/rt/alloca.d#L149).  Can that be utilized in this implementation, and is it consistent with other compiler implementations?

ibuclaw:
Moving PR to #2409

PR #2259, opened 2018-07-30
Merge all cpp sources

TurkeyMan:
Merged cpp source into one file as per @WalterBright's recommendations for correct usage of extern(C++).


Geod24:
Uh, that is obviously going to scale very well /s
Thanks for the context @JinShil

jmdavis:
*sigh* I don't see how stuffing all of C++'s std in a single file is going to be anything other than an unmaintainable mess. There isn't a lot of code here right now, but if we ever end up with much of the STL and whatnot in druntime, then this file will become so huge that it would be insane. If the "correct" usage of `extern(C++)` involves shoving entire namespaces in a single file, then we're doing something very wrong.

TurkeyMan:
> sigh I don't see how stuffing all of C++'s std in a single file is going to be anything other than an unmaintainable mess. [...]

I know, right...

PR #2263, opened 2018-08-01
Move common parts of win{32,64} into mak

wilzbach:
An opportunity to reduce quite a lot of redundancy. Let's see whether the CIs are happy too.


wilzbach:
(resolved the merge conflicts)

PR #2264, opened 2018-08-02
Convert _d_arraycast to template

JinShil:
This is all dead code (well, except for the unittest).  The idea is to verify this implementation first, merge it, then merge https://github.com/dlang/dmd/pull/8531 (assuming it passes all tests).

### DIFF
--- a/src/core/internal/string.d
+++ b/src/core/internal/string.d
@@ -15,7 +15,20 @@ nothrow:
 
 alias UnsignedStringBuf = char[20];
 
-char[] unsignedToTempString(ulong value, return scope char[] buf, uint radix = 10) @safe
+/**
+Converts an unsigned integer value to a string of characters.
+
+This implementation is a template so it can be used when compiling with -betterC.
+
+Params:
+    value = the unsigned integer value to convert
+    buf   = the pre-allocated buffer used to store the result
+    radix = the numeric base to use in the conversion (defaults to 10)
+
+Returns:
+    The unsigned integer value as a string of characters
+*/
+char[] unsignedToTempString()(ulong value, return scope char[] buf, uint radix = 10) @safe
 {
     if (radix < 2)
         // not a valid radix, just return an empty string
@@ -52,7 +65,19 @@ private struct TempStringNoAlloc
     alias get this;
 }
 
-auto unsignedToTempString(ulong value, uint radix = 10) @safe
+/**
+Converts an unsigned integer value to a string of characters.
+
+This implementation is a template so it can be used when compiling with -betterC.
+
+Params:
+    value = the unsigned integer value to convert
+    radix = the numeric base to use in the conversion (defaults to 10)
+
+Returns:
+    The unsigned integer value as a string of characters
+*/
+auto unsignedToTempString()(ulong value, uint radix = 10) @safe
 {
     TempStringNoAlloc result = void;
     result._len = unsignedToTempString(value, result._buf, radix).length & 0xff;
--- a/src/object.d
+++ b/src/object.d
@@ -4881,3 +4881,101 @@ void __ArrayDtor(T)(T[] a)
     foreach_reverse (ref T e; a)
         e.__xdtor();
 }
+
+/**
+Used by `__ArrayCast` to emit a descriptive error message.
+
+It is a template so it can be used by `__ArrayCast` in -betterC
+builds.  It is separate from `__ArrayCast` to minimize code
+bloat.
+
+Params:
+    fromType = name of the type being cast from
+    fromSize = total size in bytes of the array being cast from
+    toType   = name of the type being cast o
+    toSize   = total size in bytes of the array being cast to
+ */
+private void onArrayCastError()(string fromType, size_t fromSize, string toType, size_t toSize) @trusted
+{
+    import core.internal.string : unsignedToTempString;
+    import core.stdc.stdlib : alloca;
+
+    const(char)[][8] msgComponents =
+    [
+        "Cannot cast `"
+        , fromType
+        , "` to `"
+        , toType
+        , "`; an array of size "
+        , unsignedToTempString(fromSize)
+        , " does not align on an array of size "
+        , unsignedToTempString(toSize)
+    ];
+
+    // convert discontiguous `msgComponents` to contiguous string on the stack
+    size_t length = 0;
+    foreach (m ; msgComponents)
+        length += m.length;
+
+    auto msg = (cast(char*)alloca(length))[0 .. length];
+
+    size_t index = 0;
+    foreach (m ; msgComponents)
+        foreach (c; m)
+            msg[index++] = c;
+
+    // first argument must evaluate to `false` at compile-time to maintain memory safety in release builds
+    assert(false, msg);
+}
+
+/**
+The compiler lowers expressions of `cast(TTo[])TFrom[]` to
+this implementation.
+
+Params:
+    from = the array to reinterpret-cast
+
+Returns:
+    `from` reinterpreted as `TTo[]`
+ */
+TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @nogc pure @trusted
+{
+    const fromSize = from.length * TFrom.sizeof;
+    const toLength = fromSize / TTo.sizeof;
+
+    if ((fromSize % TTo.sizeof) != 0)
+    {
+        onArrayCastError(TFrom.stringof, fromSize, TTo.stringof, toLength * TTo.sizeof);
+    }
+
+    struct Array
+    {
+        size_t length;
+        void* ptr;
+    }
+    auto a = cast(Array*)&from;
+    a.length = toLength; // jam new length
+    return *cast(TTo[]*)a;
+}
+
+@safe @nogc pure nothrow unittest
+{
+    byte[int.sizeof * 3] b = cast(byte) 0xab;
+    int[] i;
+    short[] s;
+
+    i = __ArrayCast!(byte, int)(b);
+    assert(i.length == 3);
+    foreach (v; i)
+        assert(v == cast(int) 0xabab_abab);
+
+    s = __ArrayCast!(byte, short)(b);
+    assert(s.length == 6);
+    foreach (v; s)
+        assert(v == cast(short) 0xabab);
+
+    s = __ArrayCast!(int, short)(i);
+    assert(s.length == 6);
+    foreach (v; s)
+        assert(v == cast(short) 0xabab);
+}


src/object.d
```diff
@@ -4881,3 +4881,45 @@ void __ArrayDtor(T)(T[] a)
...
+ * this implementation.
+ * Params:
+ *  from = the array to reinterpret-cast
+ * Returns:
+ *  `from` reinterpreted as `TTo[]`
+ */
+TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @trusted
+{
+    auto nbytes = from.length * TFrom.sizeof;
+    assert((nbytes % TTo.sizeof) == 0, "array cast misalignment");
```
Geod24: This probably should be a proper error. Because if someone compiles in `-release`, this check is gone.
Also the error message should be more descriptive.

src/object.d
```diff
@@ -4881,3 +4881,45 @@ void __ArrayDtor(T)(T[] a)
...
+ * The compiler lowers expressions of `cast(TTo[])TFrom[]` to
+ * this implementation.
+ * Params:
+ *  from = the array to reinterpret-cast
+ * Returns:
+ *  `from` reinterpreted as `TTo[]`
+ */
+TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @trusted
+{
+    auto nbytes = from.length * TFrom.sizeof;
```
Geod24: `const nbytes = ...`

src/object.d
```diff
@@ -4881,3 +4881,45 @@ void __ArrayDtor(T)(T[] a)
...
+        size_t length;
+        void* ptr;
+    }
+
+    auto a = cast(Array*)&from;
+    a.length = nbytes / TTo.sizeof; // jam new length
+    return *cast(TTo[]*)a;
+}
+
+unittest
```
Geod24: `@safe @nogc pure nothrow`

src/object.d
```diff
@@ -4881,3 +4881,45 @@ void __ArrayDtor(T)(T[] a)
...
+    b.length = 12;
+
+    i = __ArrayCast!(byte, int)(b);
+    assert(i.length == 3);
+
+    s = __ArrayCast!(byte, short)(b);
+    assert(s.length == 6);
+
+    s = __ArrayCast!(int, short)(i);
+    assert(s.length == 6);
```
Geod24: I'd also do an array comparison, or at the very least, a first-element comparison in addition to this length check

src/object.d
JinShil: I can't make this an error, or it won't work in `@nogc` code.  See [this comment](https://github.com/dlang/druntime/pull/2234#issuecomment-409777211) for a more detailed explanation.

JinShil: > Because if someone compiles in -release, this check is gone.

That's a good point that I hadn't considered.  Note that if the `assert` is removed, this code can no longer be considered `@trusted`.

JinShil: @wilzbach said:

> Catching Error is defined to result in undefined behavior. We don't need to cater for such code. So you can use assert?

In an e-mail to me, @WalterBright said:

> I agree with Seb. It should have been implemented that way the first time around.

So, now that I've been directed to [`staticError`](https://github.com/dlang/druntime/blob/d8a671dac21ac0a6a8b2f10ef61b50e59ba4c2c1/src/core/exception.d#L687) (Thanks @Geod24) I can throw `Error`s in `@nogc` code, but based on @WalterBright's response, I'm not sure what I should do.

src/object.d
```diff
@@ -4881,3 +4881,61 @@ void __ArrayDtor(T)(T[] a)
...
+ * Params:
+ *  from = the array to reinterpret-cast
+ * Returns:
+ *  `from` reinterpreted as `TTo[]`
+ */
+TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @trusted
+{
+    const auto nbytes = from.length * TFrom.sizeof;
+
+    if ((nbytes % TTo.sizeof) != 0)
```
Geod24: I would additionally nest this in:
```D
version (D_NoBoundsChecks) {} // Oh, the irony
else
{
    if ((nbytes...))
        ...
}
```

but that might be up for discussion

src/object.d
```diff
@@ -4881,3 +4881,61 @@ void __ArrayDtor(T)(T[] a)
...
+ * The compiler lowers expressions of `cast(TTo[])TFrom[]` to
+ * this implementation.
+ * Params:
+ *  from = the array to reinterpret-cast
+ * Returns:
+ *  `from` reinterpreted as `TTo[]`
+ */
+TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @trusted
+{
+    const auto nbytes = from.length * TFrom.sizeof;
```
Geod24: A storage class is enough, so `auto` is redundant

src/object.d
```diff
@@ -4881,3 +4881,61 @@ void __ArrayDtor(T)(T[] a)
...
+
+@safe @nogc pure nothrow unittest
+{
+    byte[int.sizeof * 3] b;
+    int[] i;
+    short[] s;
+
+    i = __ArrayCast!(byte, int)(b);
+    assert(i.length == 3);
+    foreach(v; i)
```
Geod24: Here and later instances are missing a space: `foreach (v; i)`

src/object.d
```diff
@@ -4881,3 +4881,61 @@ void __ArrayDtor(T)(T[] a)
...
+TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @trusted
+{
+    const auto nbytes = from.length * TFrom.sizeof;
+
+    if ((nbytes % TTo.sizeof) != 0)
+    {
+        version(D_Exceptions)
+        {
+            import core.exception : onArrayMisalignmentError;
+            onArrayMisalignmentError();
```
Geod24: I think we should pass more informations to this function.

I'd like to see a message similar to this:
```
Error: Cannot cast `int[5]` to `ulong[]`, as size `20` is not aligned on `8` bytes.
```

It does not need to allocate. If you pass the elements separately, you can then provide a `toString` overload [which accepts a sink](https://github.com/dlang/druntime/blob/251cab752c25b9dcaf5bacb1344d2a26b49ed80c/src/object.d#L2635-L2642).

Let's not repeat the mistake that is array bounds mismatch's error message.

src/object.d
```diff
@@ -4881,3 +4881,61 @@ void __ArrayDtor(T)(T[] a)
...
+ * Returns:
+ *  `from` reinterpreted as `TTo[]`
+ */
+TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @trusted
+{
+    const nbytes = from.length * TFrom.sizeof;
+
+    if ((nbytes % TTo.sizeof) != 0)
+    {
+        version(D_Exceptions)
```
Geod24: Also missing space: `version (`

src/core/exception.d
```diff
@@ -371,6 +371,36 @@ unittest
     }
 }
 
+/**
+ * Thrown on an array misalignment error, that may occur during an array reinterpreting case.
+ */
+class ArrayMisalignmentError : Error
+{
+    @safe pure nothrow @nogc this(string file = __FILE__, size_t line = __LINE__, Throwable next = null )
```
Geod24: attributes should go on the RHS. Also, trailing space between `null` and closing `)`

src/core/exception.d
```diff
@@ -371,6 +371,36 @@ unittest
...
+    {
+        super( "Array misalignment", file, line, next );
```
Geod24: Extra space: `super(args);`

src/object.d
JinShil: Is your suggestion to have the `Error`'s `msg` say "Array cast misalignment", and then have the overloaded `toString` function return the more elaborate error message?

Geod24: I forgot people might be using `msg` directly... Although for `Error` it's unlikely. I think your suggestion makes sense.

src/core/exception.d
```diff
@@ -371,6 +371,70 @@ unittest
...
+ */
+class ArrayCastMisalignmentError : Error
+{
+    this(string fromType, size_t fromSize, string toType, size_t toSize,
+        string file = __FILE__, size_t line = __LINE__, Throwable next = null) @trusted pure nothrow @nogc
+    {
+        import core.internal.string : unsignedToTempString;
+        import core.stdc.stdlib : alloca;
+
+        const(char)[][8] msgComponents =
```
JinShil: How about this for formulating a more detailed error message.  I'm trying to avoid using the C standard library ([`alloca` actually isn't part of the C standard library](https://github.com/dlang/druntime/blob/9a8edfb48e4842180c706ee26ebd8edb10be53f4/src/rt/alloca.d#L58)).

Geod24: > I'm trying to avoid using the C standard library 

Why ?
Also, I would suggest to just store those values as field and do the formatting in `toString`

JinShil: > Why ?

Because eventually I'd like to jettison it completely from druntime.  I also need to consider using the same code for the `assert` message in `__ArrayCast`.

JinShil: I need to come up with a utility for creating a dynamic string on the stack.

JinShil: I ended up merging the message generation code into the `__ArrayCast` implementation itself so I could use it for the `assert` as well.

src/core/exception.d
```diff
@@ -601,6 +600,22 @@ extern (C) void onUnicodeError( string msg, size_t idx, string file = __FILE__,
...
+ *
+ * Params:
+ *  msg  = Information about the error.
+ *  file = The name of the file that signaled this error.
+ *  line = The line number on which this error occurred.
+ *
+ * Throws:
+ *  $(LREF Error).
+ */
+void throwStaticError(string msg, string file = __FILE__, size_t line = __LINE__) @trusted pure nothrow @nogc
```
JinShil: This just throws a generic `Error`.  I don't know what the policy is on strongly typed exceptions.

jacob-carlborg: What's the reason for this function?

src/object.d
```diff
@@ -4881,3 +4881,90 @@ void __ArrayDtor(T)(T[] a)
...
+                msg[index++] = c;
+
+        version (D_Exceptions)
+        {
+            import core.exception : throwStaticError;
+            throwStaticError(cast(string)msg);
+        }
+        else
+        {
+            assert(false, cast(string)msg);
```
jacob-carlborg: I don't know how `assert(false)` behaves but `assert(0)` will halt the program when assertions are turned off.

src/core/exception.d
JinShil: See the body of the function `throw staticError!Error(msg, file, line, null);`.  That template instantiates a static `Error` so that it can be used in `@nogc` code.  However, it is `private`.  I created the `throwStaticError` function to keep it private and still allow throwing static `Error`s.

wilzbach: To be able to throw an error in `@nogc`.

src/object.d
wilzbach: `assert(false` will do the same because it's the same: https://run.dlang.io/is/AVSUFK

@JinShil we typically write `assert(0, ...`

src/object.d
```diff
@@ -4881,3 +4881,90 @@ void __ArrayDtor(T)(T[] a)
...
+        foreach(m ; msgComponents)
+            foreach(c; m)
+                msg[index++] = c;
+
+        version (D_Exceptions)
+        {
+            import core.exception : throwStaticError;
+            throwStaticError(cast(string)msg);
+        }
+        else
```
wilzbach: It looks like D is moving in C++ land here:

> Divergent error handling has fractured the C++ community into incompatible dialects:
(1) C++ projects often ban even turning on compiler support for exception handling, but this means they are
not using Standard C++. Exceptions are required to use central C++ standard language features (e.g., con-
structors) and the C++ standard library. Yet in [SC++F 2018], __over half of C++ developers__ report that __excep-
tions are banned__ in part (32%) or all (20%) of their code, which means they are using a__ divergent language
dialect with different idioms__ (e.g., two-phase construction) and either a nonconforming standard library
dialect or none at all. We must make it possible for all C++ projects to at least turn on exception handling
support and use the standard language and library.

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0709r0.pdf

src/object.d
JinShil: >  we typically write assert(0, ...

IMO we shot stop doing that.  It's a condition, not a number.

src/object.d
JinShil: Indeed.  That's what happens when a language is not designed from the beginning with freestanding, resource-constrained systems and scalability in mind.  The emergence of things like `@nogc`, betterC, etc. shows that D is backtracking, trying to undo some mistakes of the past.  GC should be an opt-in library feature and exceptions should not be rooted in `Object` or Java-style `class`-like types.  But that's what we have.  If we want D to scale to systems other than resource-abundant PCs and servers running full-featured operating systems, then we have to offer alternatives.  I hope future language decisions will keep principles of minimalism in mind.  Given [what's possible in D](https://forum.dlang.org/post/qjowkuwstewnmdunepiq@forum.dlang.org), I'd prefer if classes were just deprecated, and structs were give a few more features.

src/object.d
JinShil: @jacob-carlborg I tested this out and you're right.  If the first argument is `0` or `false` it still runs when asserts are disabled.  However, I can find no mention of this behavior in the spec.  If this is by design, it'd great if you could document it, and then I could potentially remove throwing the static `Error` altogether.  

Kindly asking for advise from all reviewers on how to proceed.

JinShil: Found this in [the spec](https://dlang.org/spec/expression.html#assert_expressions).

> The implementation may handle the case of the first AssignExpression evaluating at compile time to false differently in that in release mode it may simply generate a HLT instruction or equivalent. 

I guess I'l just remove the static `Error` and rely on `assert(false)` for all implementations, then.  It should keep the functions `@trusted` even in release mode.

JinShil: > Please don't.

I'd be happy to revert back to throwing a static `Error`, but I need consensus from reviewers or decree from @WalterBright

src/core/exception.d
jacob-carlborg: > I created the throwStaticError function to keep the staticError template private and still allow throwing static

What's the advantage of this wrapper function compared to using `staticError` directly? I don't see how introducing a new public symbol to keep an existing symbol private is an advantage. You can make the existing `staticError` `package` to keep it internal to druntime.

src/object.d
```diff
@@ -4881,3 +4881,108 @@ void __ArrayDtor(T)(T[] a)
...
+
+/***********************************************************
+ * The compiler lowers expressions of `cast(TTo[])TFrom[]` to
+ * this implementation.
+ * Params:
+ *  from = the array to reinterpret-cast
+ * Returns:
+ *  `from` reinterpreted as `TTo[]`
+ */
+TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @trusted
```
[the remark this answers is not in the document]
jacob-carlborg: Perhaps use the same order as the `cast` expression.

src/object.d
jacob-carlborg: I know there's been some disagreement about the behavior of `assert(0/false)` because not all CPUs have the `HLT` instruction, but perhaps that covered with the "or equivalent" in the spec.

src/core/exception.d
JinShil: I can't make it `package` because that would only apply to modules within `core`.

src/object.d
```diff
@@ -4881,3 +4881,108 @@ void __ArrayDtor(T)(T[] a)
...
+{
+    const fromSize = from.length * TFrom.sizeof;
+    const toLength = fromSize / TTo.sizeof;
+
+    if ((fromSize % TTo.sizeof) != 0)
+    {
+        const errorMessage = generateArrayCastErrorMessage(TFrom.stringof, fromSize, TTo.stringof, toLength * TTo.sizeof);
+
+        // must be `false` or `0` to ensure the assertion remains
+        // even when compiling with `-release`
```
wilzbach: That comment is a bit misleading to future readers as the `assert(0)` expression gets replaced by a HLT instruction in `-release` ...)

src/core/exception.d
jacob-carlborg: Hmm, right. This is in the `object` module which is not inside any package. But again, is this any better than making the existing `staticError` `public`?

JinShil: No, it's not any better IMO.  If we decide to use static `Error`s, I think I'll just follow the existing pattern for other static `Error`s in `core.exception`.  That is, create a strongly typed `Error` and throw it with an `on{whatever}` function.

src/object.d
```diff
@@ -4881,3 +4881,99 @@ void __ArrayDtor(T)(T[] a)
...
+{
+    const fromSize = from.length * TFrom.sizeof;
+    const toLength = fromSize / TTo.sizeof;
+
+    if ((fromSize % TTo.sizeof) != 0)
+    {
+        onArrayCastError(TFrom.stringof, fromSize, TTo.stringof, toLength * TTo.sizeof);
+    }
+
+    struct Array
```
n8sh: Does this need `static`?

src/object.d
```diff
@@ -4881,3 +4881,99 @@ void __ArrayDtor(T)(T[] a)
...
+ *  from = the array to reinterpret-cast
+ * Returns:
+ *  `from` reinterpreted as `TTo[]`
+ */
+TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @trusted
+{
+    const fromSize = from.length * TFrom.sizeof;
+    const toLength = fromSize / TTo.sizeof;
+
+    if ((fromSize % TTo.sizeof) != 0)
```
n8sh: Making sure that @Geod24's suggestion in https://github.com/dlang/druntime/pull/2264#discussion_r207113286 isn't lost:
```d
version (D_NoBoundsChecks) {}
else if ((fromSize % TTo.sizeof) != 0)
```

src/object.d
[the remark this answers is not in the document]
Geod24: The only rule I know of is to put arguments that needs to be explicitly specified first, otherwise follow the order or arguments. Additionally, in very rare cases, you might have 2 templates argument depending on each others (e.g. via a constraint) in which case the order is self-evident.

In this case, this is never called directly by the user so that order doesn't matter much.
Personally, I find `TFrom, TTo` (as currently specified) to be the clearer alternative.

src/object.d
Geod24: Since there is no functions, it does not

src/object.d
JinShil: I'm waiting to find out what happens with https://github.com/dlang/dmd/pull/8532.  I don't think it should hold up this PR, and would even be better if it were done in a followup PR.

src/object.d
```diff
@@ -4881,3 +4881,99 @@ void __ArrayDtor(T)(T[] a)
...
+@safe @nogc pure nothrow unittest
+{
+    byte[int.sizeof * 3] b;
+    int[] i;
+    short[] s;
+
+    i = __ArrayCast!(byte, int)(b);
+    assert(i.length == 3);
+    foreach (v; i)
+        assert(v == 0);
```
wilzbach: Maybe use something more interesting as values than zeros?

src/object.d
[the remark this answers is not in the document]
wilzbach: Shouldn't array casting be also `pure` and `@nogc`?

src/object.d
n8sh: Fair.

src/object.d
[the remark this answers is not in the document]
n8sh: Ah, the style policy on template attributes has changed. Good!
https://github.com/dlang/dlang.org/commit/f95cb98ea5f8a8314bf8e41193a10a8180255dc8#diff-2ceb1eaea6947a0a0ad321e635cec2e0
```diff
-    $(LI $(I Templated) functions should $(B not) be annotated with attributes
-         as the compiler can infer them.)
+    $(LI If the template arguments for a $(I templated) function affect whether
+         an attribute is appropriate, then the function should $(B not) be
+         annotated with that attribute so that the compiler can infer it.
+         However, if the attribute is not affected by the template arguments
+         (and thus would always be inferred), then the function should be
+         explicitly annotated with that attribute just like a non-templated
+         function would be.)
```

src/object.d
```diff
@@ -4881,3 +4881,99 @@ void __ArrayDtor(T)(T[] a)
...
+
+/***********************************************************
+ * The compiler lowers expressions of `cast(TTo[])TFrom[]` to
+ * this implementation.
+ * Params:
+ *  from = the array to reinterpret-cast
+ * Returns:
+ *  `from` reinterpreted as `TTo[]`
+ */
+TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @trusted
```
n8sh: Can `inout` be used here to reduce template bloat?

src/object.d
```diff
@@ -4881,3 +4881,99 @@ void __ArrayDtor(T)(T[] a)
...
+ * builds.  It is separate from `__ArrayCast` to minimize code
+ * bloat.
+ * Params:
+ *  fromType = name of the type being cast from
+ *  fromSize = total size in bytes of the array being cast from
+ *  toType   = name of the type being cast o
+ *  toSize   = total size in bytes of the array being cast to
+ * Returns:
+ *  a descriptive error message describing the cast failure
+ */
```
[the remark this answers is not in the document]
n8sh: It looks like the rest of `object.d` doesn't follow this convention. Is it more important to keep a consistent style within a document?

n8sh:
To speed things along I've edited the commit to incorporate minor changes requested by @wilzbach: explicit `@nogc pure` and not using all zeroes in the unittest.

src/object.d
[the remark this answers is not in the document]
JinShil: @n8sh, you are right, and I doubt anyone would disagree with you, but @WalterBright has placed restrictions on large-scale changes to the code base, forcing us to resort to incremental changes to achieve consistent coding conventions.  So, unfortunately, this is how we're going to get there.

src/object.d
```diff
@@ -4881,3 +4881,97 @@ void __ArrayDtor(T)(T[] a)
...
+        , toType
+        , "`; an array of size "
+        , unsignedToTempString(fromSize)
+        , " does not align on an array of size "
+        , unsignedToTempString(toSize)
+    ];
+
+    // convert discontiguous `msgComponents` to contiguous string on the stack
+    size_t length = 0;
+    foreach(m ; msgComponents)
```
Geod24: `foreach (`

src/core/internal/string.d
```diff
@@ -15,7 +15,17 @@ nothrow:
 
 alias UnsignedStringBuf = char[20];
 
-char[] unsignedToTempString(ulong value, return scope char[] buf, uint radix = 10) @safe
+/**
+Converts an unsigned integer value to a string of characters. This
+implemenation is a template so it can be used when compiling with -betterC
```
Geod24: Regarding DDOC:

- The first line is a "title", so it should usually be a one liner/sentence
- I **think** the preferred style is to have 1 blank line between sections (e.g. before `Params:` / `Returns:`)
- `implemenation`

Would `pragma(inline, true)` help in this case ?

JinShil: > Would pragma(inline, true) help in this case ?

Yeah, maybe.  The source code would be available in -betterC builds.  GDC doesn't currently implement `pragma(inline, true)`, though [I think it's coming](https://github.com/D-Programming-GDC/GDC/pull/587).  🤔 Yeah, let's go with that for now.  I can always change it to a template later if it doesn't work out.

jacob-carlborg: > The first line is a "title", so it should usually be a one liner/sentence

Yes everything up to the first blank line is considered the title and therefore should be short. After the blank line the description follows which can be long. 

> I think the preferred style is to have 1 blank line between sections (e.g. before Params: / Returns:)

I prefer that.

PR #2265, opened 2018-08-02
Fix FreeBSD's statvfs

jmdavis:
For some reason, FreeBSD's declarations for core.sys.posix.sys.statvfs are statfs declarations that come from sys/mount.h rather than sys/statvfs.h. This adds core.sys.freebsd.sys.mount and moves the statfs stuff to there and then adds the correct declarations to core.sys.posix.sys.statvfs.

### DIFF
--- a/src/core/sys/freebsd/sys/mount.d
+++ b/src/core/sys/freebsd/sys/mount.d
@@ -1,0 +1,303 @@
+//Written in the D programming language
+
+/++
+    D header file for FreeBSD's sys/mount.h.
+
+    Copyright: Copyright 2018 -
+    License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+    Authors:   $(HTTP jmdavisprog.com, Jonathan M Davis)
+ +/
+module core.sys.freebsd.sys.mount;
+
+version(FreeBSD):
+
+import core.sys.posix.sys.stat : stat_t;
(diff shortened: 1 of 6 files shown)


src/core/sys/freebsd/sys/mount.d
```diff
@@ -0,0 +1,303 @@
...
+    uid_t f_owner;
+    fsid_t f_fsid;
+    char[80] f_charspare;
+    char[MFSNAMELEN] f_fstypename;
+    char[MNAMELEN] f_mntfromname;
+    char[MNAMELEN] f_mntonname;
+}
+
+
+enum ulong MNT_RDONLY = 0x0000000000000001;
```
ZombineDev: I would have grouped the constants in an `enum MNT : ulong { /+ ... +/ }` aggregate and then "exported" out the individual members as aliases. Easily automatable via `static foreach` + `mixin`.

jmdavis: Except that we haven't been doing enums like that in druntime. This file does with `FFlag`, but it's the only one, and I'm actually planning to create a PR after this to remove it so that they're just enums directly in the module like we've done everywhere else when porting constants from C in druntime. While it's not how I'd do it if I were writing proper D code, it's closer to how the C code actually is and thus makes using the C bindings more like using the actual C functions and constants in C.

I also don't see much benefit in having all of the values in there twice, even if the generation of the second set of them is automated.

PR #2267, opened 2018-08-12
Convert _d_arraycast to template - Take 2

JinShil:
`alloca` doesn't work in -betterC.  See https://issues.dlang.org/show_bug.cgi?id=19159.  Therefore, I resorted to `malloc`.


jacob-carlborg:
I thought we considered `malloc` to be pure just as we do with `new`.

PR #2269, opened 2018-08-12
Add setjmp/ucontext definitions for FreeBSD/aarch64

myfreeweb:
Tested with ldc ltsmaster.

Relevant headers:
https://github.com/freebsd/freebsd/blob/master/sys/arm64/include/setjmp.h
https://github.com/freebsd/freebsd/blob/master/sys/arm64/include/ucontext.h

### DIFF
--- a/src/core/sys/posix/ucontext.d
+++ b/src/core/sys/posix/ucontext.d
@@ -765,6 +765,38 @@ else version( FreeBSD )
             int[6]          mc_spare2;
         }
     }
+    else version( AArch64 )
+    {
+        alias __register_t = long;
+
+        struct gpregs
+        {
+            __register_t[30] gp_x;
+            __register_t     gp_lr;
+            __register_t     gp_sp;
+            __register_t     gp_elr;
+            uint             gp_spsr;
+            int              gp_pad;
+        }
+
+        struct fpregs
+        {
+            ulong[2][32]    fp_q; // __uint128_t
+            uint            fp_sr;
+            uint            fp_cr;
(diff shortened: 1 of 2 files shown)


joakim-noah:
Looks like this new submitter needs to be approved on the auto-tester before it will run.

wilzbach:
Approved (:

src/core/sys/posix/ucontext.d
```diff
@@ -765,6 +765,38 @@ else version( FreeBSD )
             int[6]          mc_spare2;
         }
     }
+    else version( AArch64 )
+    {
+        alias long __register_t;
```
wilzbach: Can we please use the new alias syntax? (will correct this quickly myself)

wilzbach: > ERROR: Permission to myfreeweb/druntime.git denied to wilzbach.

Oh well, it looks like you need to do it yourself or enable maintainer access to your PRs.

myfreeweb: Oh the one with the `=` sign is new? I'm not familiar with D, this is the first D code I'm editing :) I assumed this is the new one since it's more common in that file.

wilzbach: yes, `alias __register_t = long` is the "new alias syntax".
Unfortunately the old one was never deprecated and as there's still some code that uses it (druntime being a prime example).
We try to avoid using it for new code at least ;-)
Thanks for the fix-up!

PR #2274, opened 2018-08-15
[RFC] Issue 19087 - final switch cannot be used in -betterC

wilzbach:
Not fully sure about this as the `SwitchError` does contain a lot more information.
Maybe we should version this and only use `assert` in betterC and otherwise throw the exception?

### DIFF
--- a/src/core/exception.d
+++ b/src/core/exception.d
@@ -574,14 +574,20 @@ extern (C) void onInvalidMemoryOperationError(void* pretend_sideffect = null) @t
  */
 extern (C) void onSwitchError( string file = __FILE__, size_t line = __LINE__ ) @safe pure nothrow
 {
-    throw new SwitchError( file, line, null );
+    version (D_Exceptions)
+        throw new SwitchError( file, line, null );
+    else
+        assert(0, "No appropriate switch clause found");
 }
 
 // Compiler lowers final switch default case to this (which is a runtime error)
 void __switch_errorT()(string file = __FILE__, size_t line = __LINE__) @trusted
 {
     // Consider making this a compile time check.
-    throw staticError!SwitchError(file, line, null);
+    version (D_Exceptions)
+        throw staticError!SwitchError(file, line, null);
+    else
+        assert(0, "No appropriate switch clause found");
 }
 
 /**


JinShil:
@WalterBright has already affirmed to me (twice actually) in private e-mail exchange that he is OK replacing `Error`s with `assert`s even in non-betterC code.

Me:   I could use `version(D_BetterC)` to generate a runtime `assert` instead of an `Error` for -betterC.  Would that be OK, or do you have something else in mind?
Walter: Seems like just having it assert() would solve the problem.

@wilzbach: Catching Error is defined to result in undefined behavior. We don't need to cater for such code. So you can use assert?
Me: Is that right, can I just make the `Error` an `assert`?  I think that would solve all problems.
Walter: I thought I already suggested to just use assert(), but perhaps I sent the email awry. Anyhow, I agree with Seb. It should have been implemented that way the first time around.

IMO the decision has been made in direction for the future established.

src/core/exception.d
```diff
@@ -563,25 +532,22 @@ extern (C) void onInvalidMemoryOperationError(void* pretend_sideffect = null) @t
...
  *  file = The name of the file that signaled this error.
  *  line = The line number on which this error occurred.
- *
- * Throws:
- *  $(LREF SwitchError).
  */
 extern (C) void onSwitchError( string file = __FILE__, size_t line = __LINE__ ) @safe pure nothrow
 {
-    throw new SwitchError( file, line, null );
+    assert(0, "No appropriate switch clause found");
```
JinShil: One of the things I'd really like to do here, however, is make use of the `__FILE__` and `__LINE__` parameters in the error message.  That was the only reason I hadn't fixed this PR yet.  Given that such a trivial thing is actually quite difficult given the limitations discussed in https://github.com/dlang/druntime/pull/2268, however, I'd be willing to merge this PR as is, as long as there is a bugzilla issue filed to improve the error message.

src/core/exception.d
```diff
@@ -563,25 +532,22 @@ extern (C) void onInvalidMemoryOperationError(void* pretend_sideffect = null) @t
...
 }
 
 // Compiler lowers final switch default case to this (which is a runtime error)
 void __switch_errorT()(string file = __FILE__, size_t line = __LINE__) @trusted
 {
     // Consider making this a compile time check.
-    throw staticError!SwitchError(file, line, null);
+    assert(0, "No appropriate switch clause found");
```
JinShil: Another thing we need is the ability to test runtime assertions in `unittest`s.  Any ideas?

jmdavis: It generally works just fine to catch an `AssertError` in a unit test. That being said, I don't really see anything to test here. The stuff that was in the `SwitchError` tests all had to do with testing that the members of `SwitchError` were initialized correctly. I guess that you could do the same with an `AssertError`, but unless `-unittest` makes `assert(0, ...)` still throw an `AssertError` when the assertion isn't directly in a `unittest` block when compiling with -release, then you couldn't test it in -release. And honestly, I don't think that testing that the exception is initialized with all of the right values is all that valuable. What you generally need to test is that it's thrown when it's supposed to and not thrown when it's not supposed to, and that would involve testing code that actually triggered `__switch_errorT()`. All around though, it does get pretty weird to try to test anything that involves `assert(0, ...)`, since pretty much by definition, that code is never supposed to be reached, and it's a serious bug in your program if it is.

jacob-carlborg: From the spec

“Unlike AssertExpressions used elsewhere, the assert is not assumed to hold, and upon assert failure the program is still in a defined state.”

“AssertExpression has different semantics if it is in a unittest or in contract.”

jacob-carlborg: I don't know if it's in the spec but it's been pointed out that `Error`s should not be caught. Since `assert` throws an `AssertError` it's not possible to create a unit test runner that continues after a failed test.

wilzbach:
> @WalterBright has already affirmed to me (twice actually) in private e-mail exchange that he is OK replacing Errors with asserts even in non-betterC code.
> ...
> One of the things I'd really like to do here, however, is make use of the __FILE__ and __LINE__ parameters in the error message. That was the only reason I hadn't fixed this bug myself. Given that such a trivial thing is actually quite difficult given the limitations discussed in #2268

I was thinking a bit more about this and I now changed it too:

```d
version (D_Exceptions)
    throw new SwitchError( file, line, null );
else
    assert(0, "No appropriate switch clause found");
```

So that until we figure out a nice way to do error formatting in `-betterC`, normal D code will see the more informative error.

src/core/exception.d
wilzbach: Haha spec != implementation. It works quite well to catch an `AssertError` and lots of code in Phobos depends on this, e.g.

```
ag "collectException.*Error"
std/exception.d
168:    assert(collectExceptionMsg!AssertError(assertNotThrown!StringException(
176:    assert(collectExceptionMsg!AssertError(assertNotThrown!StringException(
180:    assert(collectExceptionMsg!AssertError(assertNotThrown!StringException(
184:    assert(collectExceptionMsg!AssertError(assertNotThrown!StringException(
311:    assert(collectExceptionMsg!AssertError(assertThrown!StringException(
706:        auto e = collectException!Error(enforceEx!OutOfMemoryError(false));
714:        auto e = collectException!Error(enforceEx!OutOfMemoryError(false, "file", 42));
770:    assert(collectException!RangeError(a[4], b));
1843:    assert(throwRangeError().ifThrown(0).collectException!RangeError() !is null);
1844:    assert(throwRangeError().ifThrown(e=>0).collectException!RangeError() !is null);
```

```
ag "catch.*Error"
std/typecons.d
4302:        catch (Error e) {}

std/bitmanip.d
708:    catch (AssertError ae)
712:    catch (AssertError ae)

std/datetime/package.d
293:        catch (AssertError e)
320:        catch (AssertError e)

std/range/package.d
720:    catch (AssertError unused)

std/exception.d
200:    catch (AssertError) assert(0);
206:    catch (AssertError) assert(0);
212:    catch (AssertError) assert(0);
218:    catch (AssertError) assert(0);
227:        catch (AssertError) thrown = true;
238:        catch (AssertError) thrown = true;
249:        catch (AssertError) thrown = true;
261:        catch (AssertError) thrown = true;
327:    catch (AssertError) assert(0);
334:    catch (AssertError) assert(0);
341:    catch (AssertError) assert(0);
349:    catch (AssertError) assert(0);
356:        catch (AssertError)
366:        catch (AssertError)
376:        catch (AssertError)
386:        catch (AssertError)

std/experimental/allocator/gc_allocator.d
69:        catch (OutOfMemoryError)

std/parallelism.d
4830:    catch (ParallelForeachError e) {}
```

jacob-carlborg: > Haha spec != implementation. It works quite well to catch an AssertError and lots of code in Phobos depends on this, e.g

Yeah, I know. I do that as well. But I would feel better if it was an Exception instead of an Error.

JinShil:
> I was thinking a bit more about this and I now changed it to.

I'm OK with the change or leaving it as an `assert` with a vague error message for both -betterC and non-betterC builds, as long as there is an issue filed in bugzilla that the error message needs updating.

wilzbach:
> But, either way, please file a bugzilla issue that the error message needs updating.

https://issues.dlang.org/show_bug.cgi?id=19172

n8sh:
Aren't `assert` messages omitted anyway when compiling in release mode?

JinShil:
> Aren't assert messages omitted anyway when compiling in release mode?

No, See item #2 at https://dlang.org/spec/expression.html#assert_expressions

jacob-carlborg:
> Aren't assert messages omitted anyway when compiling in release mode?
No, See item #2 at https://dlang.org/spec/expression.html#assert_expressions

There's no message printed in release mode:

```
$ cat main.d
void main()
{
    assert(false);
}
$ dmd -release main.d
$ ./main
Illegal instruction: 4
```

https://run.dlang.io/is/pNZtUA

jmdavis:
Specifically, if the condition is known to be `false`, then the compiler turns the assertion into a `HLT` instruction with -release. So, reaching that line will kill your program, and there will be no message.

JinShil:
Ah Right!  That's actually #11 at https://dlang.org/spec/expression.html#assert_expressions

jacob-carlborg:
#2 is for when indicating some code path is unreachable, i.e.

```d
int foo(int a)
{
    if (a == 3)
        return 1;

    assert(false);
}
```

In the above example the compiler will not complain that there's no `return` after the `if` since `assert(false)` is unreachable code.

src/core/exception.d
```diff
@@ -574,14 +574,20 @@ extern (C) void onInvalidMemoryOperationError(void* pretend_sideffect = null) @t
  */
 extern (C) void onSwitchError( string file = __FILE__, size_t line = __LINE__ ) @safe pure nothrow
 {
-    throw new SwitchError( file, line, null );
+    version (D_Exceptions)
```
JinShil: I suggest adding a comment here that this is temporary, and that once we get either a `pure` malloc, or an `alloca` for -betterC that these should be changed to an `assert` for both cases.  Either that, or file a bugzilla issue with the same.

PR #2275, opened 2018-08-16
Fix signature of getfsstat to use c_long instead of long.

jmdavis:


src/core/sys/freebsd/sys/mount.d
```diff
@@ -292,7 +292,7 @@ int fhstat(const fhandle_t*, stat_t*);
 int fhstatfs(const fhandle_t*, statfs_t*);
 int fstatfs(int, statfs_t*);
 int getfh(const char*, fhandle_t*);
-int getfsstat(statfs_t*, long, int);
+int getfsstat(statfs_t*, c_long, int);
```
ZombineDev: You're missing an import.

jmdavis: Whoops. It should be fixed now.

PR #2276, opened 2018-08-16
[RFC] Add pure fail-fast assertMalloc, assertCalloc, assertRealloc

n8sh:
These functions achieve purity by aborting the program on failure so errno is never observed to change. This PR also changes `pureMalloc` etc. to templates so they work in betterC.

The idea was suggested by @ZombineDev at https://github.com/dlang/phobos/pull/6660#issuecomment-412838596 and  https://github.com/dlang/druntime/pull/2268#issuecomment-412845205. The sole difference is that instead of adding a boolean template parameter to `pureMalloc` I instead named the new variant differently. I felt that `assertMalloc(100)` has a more obvious meaning than `pureMalloc!true(100)` which gives no hint as to whether it's the one that aborts on failure or the one that saves and restores errno.

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -812,17 +812,120 @@ struct GC
 
 /**
  * Pure variants of C's memory allocation functions `malloc`, `calloc`, and
+ * `realloc` that achieve purity by aborting the program on failure so
+ * they never visibly change errno.
+ *
+ * The functions may terminate the program using `onOutOfMemoryError` or
+ * `assert(0)`. These functions' purity guarantees no longer hold if
+ * the program continues execution after catching AssertError or
+ * OutOfMemoryError.
+ *
+ * A similar fail-fast variant of C's memory deallocation function `free`
+ * is not provided due to the lack of a portable and efficient way to test
+ * for failure. If purity is required `pureFree` may be used.
+ *
+ * See_Also:
+ *     $(LREF pureMalloc), $(LREF pureCalloc), $(LREF pureRealloc),
+ *     and $(LREF pureFree) which achieve purity by saving and restoring
+ *     errno.
+ *     $(LINK2 https://dlang.org/library/object/error.html, Error class documentation)
+ *     on the danger of continued execution after catching an Error.
+ */
+void* assertMalloc()(size_t size) @nogc nothrow pure @trusted
+{
+    void* ret = fakePureMalloc(size);
+    if (!ret && size)
+    {
+        version (D_Exceptions)
+        {
+            import core.exception : onOutOfMemoryError;
+            onOutOfMemoryError();
+        }
+        else assert(0, "Memory allocation failed");
+    }
+    return ret;
+}
+/// ditto
+void* assertCalloc()(size_t nmemb, size_t size) @nogc nothrow pure @trusted
+{
+    void* ret = fakePureCalloc(nmemb, size);
+    if (!ret && nmemb && size)
+    {
+        version (D_Exceptions)
+        {
+            import core.exception : onOutOfMemoryError;
+            onOutOfMemoryError();
+        }
+        else assert(0, "Memory allocation failed");
+    }
+    return ret;
+}
+/// ditto
+void* assertRealloc()(void* ptr, size_t size) @nogc nothrow pure @system
+{
+    void* ret = fakePureRealloc(ptr, size);
+    if (!ret && size)
+    {
+        version (D_Exceptions)
+        {
+            import core.exception : onOutOfMemoryError;
+            onOutOfMemoryError();
+        }
+        else assert(0, "Memory allocation failed");
+    }
+    return ret;
+}
+
+///
+@nogc nothrow pure @system unittest
+{
+    void[] allocate(size_t n) pure
+    {
+        return assertMalloc(n)[0 .. n];
+    }
+
+    auto buf = allocate(100);
+    assert(buf.length == 100);
+    pureFree(buf.ptr);
+}
+
+@nogc nothrow pure @system unittest
+{
+    const int errno = fakePureErrno();
+
+    void* x = assertMalloc(10);       // normal allocation
+    assert(errno == fakePureErrno()); // errno shouldn't change
+    assert(x !is null);               // allocation should succeed
+
+    x = assertRealloc(x, 20);         // normal reallocation
+    assert(errno == fakePureErrno()); // errno shouldn't change
+    assert(x !is null);               // allocation should succeed
+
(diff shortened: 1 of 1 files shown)


src/core/memory.d
```diff
@@ -810,6 +810,106 @@ struct GC
...
+ *     errno.
+ *     $(LINK2 https://dlang.org/library/object/error.html, Error class documentation)
+ *     on the danger of continued execution after catching an Error.
+ */
+void* assertMalloc()(size_t size) @nogc nothrow pure @trusted
+{
+    void* ret = fakePureMalloc(size);
+    if (!ret && size)
+    {
+        version (D_BetterC) assert(0, "Memory allocation failed");
```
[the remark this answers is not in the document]
n8sh: Got it.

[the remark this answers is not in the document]
wilzbach: Also it would be better to use `version (D_Exceptions)` to check whether exceptions can be thrown as LDC/GDC allow to disable them selectively (and e.g. WebAssembly might not defined `-betterC`, but still not support exceptions.)

src/core/memory.d
```diff
@@ -810,6 +810,91 @@ struct GC
...
+    }
+
+    auto buf = allocate(100);
+    assert(buf.length == 100);
+    pureFree(buf.ptr);
+}
+
+@nogc nothrow pure @system unittest
+{
+    const int errno = fakePureErrno();
```
JinShil: `fakePureErrno` is mangled to `fakePureErrnoImpl` which is not a template.  Therefore this won't work until `fakePureErrnoImpl` is made into a template, or some other solution is devised.

src/core/memory.d
```diff
@@ -810,6 +810,91 @@ struct GC
...
 
+/**
+ * Pure variants of C's memory allocation functions `malloc`, `calloc`, and
+ * `realloc` that achieve purity by aborting the program on failure so
+ * they never visibly change errno.
+ *
+ * The functions may terminate the program using `onOutOfMemoryError` or
+ * `assert(0)`. These functions' purity guarantees no longer hold if
+ * the program continues execution after catching AssertError or
+ * OutOfMemoryError.
```
JinShil: I think this comment needs updating.

n8sh: I kept the mention of the possibility so if we later change to using `onOutOfMemoryError` no one will be blindsided.  Do you think it's better to remove it because that's too unlikely?

JinShil: No need to change anything immediately, IMO.  We can wait to see how the review goes.  It just eventually needs to reflect the final implementation.

src/core/memory.d
n8sh: Should be fixed now.

src/core/memory.d
```diff
@@ -899,20 +987,51 @@ void pureFree(void* ptr) @system pure @nogc nothrow
 }
 
 // locally purified for internal use here only
+static import core.stdc.errno;
+
+private enum canLinkToErrno = __traits(getOverloads, core.stdc.errno, "errno").length == 1
+    && __traits(getLinkage, core.stdc.errno.errno) == "C";
 
+version (D_BetterC) @nogc nothrow private @system
```
JinShil: Will this implementation work in non-betterC builds?  Can't we use this implementation for both -betterC and non-betterC builds?

n8sh:
It would work. The purpose of the version branch is so outside code that assumes `fakePureErroImpl` exists and tries to link to it will not break.

src/core/memory.d
n8sh: It would work. The purpose of the version branch is so outside code that assumes `fakePureErroImpl` exists and tries to link to it will not break.

ZombineDev:
I would argue that OOM is not a logical error on the programmer side, rather it is an exception from the program environment, albeit a non-recoverable from one. IMO, we should continue calling `onOutOfMemoryError` as it is provides for a bit more systematic way of handling this type of exception, just make sure it has a fallback for `version (D_Exceptions) { } { /* ... */ }` .

n8sh:
@ZombineDev That makes sense.

JinShil:
> I would argue that OOM is not a logical error on the programmer side, rather it is an exception from the program environment, albeit a non-recoverable from one. IMO, we should continue calling onOutOfMemoryError as it is provides for a bit more systematic way of handling this type of exception

I'm not terribly partial to either one, but @WalterBright [has spoken](https://github.com/dlang/druntime/pull/2274#issuecomment-413375758).  We need to come to some consensus about this and apply it consistently.

JinShil:
Sorry! Wrong button.

n8sh:
>I'm not terribly partial to either one, but @WalterBright has spoken. 

I took your summary to mean that he was okay with using assert but not that he was mandating it. Did I get the wrong idea?

>@WalterBright has already affirmed to me (twice actually) in private e-mail exchange that he is OK replacing Errors with asserts even in non-betterC code.

JinShil:
> I took your summary to mean that he was okay with using assert but not that he was mandating it. Did I get the wrong idea?

@WalterBright will have to speak for himself, but my interpretation is that he was preferred us to use `assert`s.  It's very unlike him to mandate anything, from my experience.

JinShil:
We still need to come to some consensus about `Error`s vs `assert`s.  Once that decision is made, this should be updated to reflect it.

src/core/memory.d
```diff
@@ -810,6 +810,106 @@ struct GC
...
+        {
+            import core.exception : onOutOfMemoryError;
+            onOutOfMemoryError();
+        }
+        else assert(0, "Memory allocation failed");
+    }
+    return ret;
+}
+/// ditto
+void* assertRealloc()(void* ptr, size_t size) @nogc nothrow pure @system
```
WalterBright: Since realloc() free's memory, it cannot ever be considered pure.

aG0aep6G: > Since realloc() free's memory, it cannot ever be considered pure.

We have a [`pureFree`](https://dlang.org/phobos/core_memory.html#.pureFree), and [`GC.free`](https://dlang.org/phobos/core_memory.html#.GC.free) is also marked as `pure`.

`pure` functions are allowed to modify their parameters. Freeing is never `@safe`, but in what way is it less pure than other forms of mutation?

n8sh: There is also already a [`pureRealloc`](https://dlang.org/phobos/core_memory.html#.pureRealloc).

https://github.com/dlang/druntime/blob/db2910ee7425f91dc6fe38f149f4b8eaa3b88f7f/src/core/memory.d#L840-L847

PR #2280, opened 2018-08-18
Issue 17596: Version bindings so that they work for both FreeBSD 11 and 12.

jmdavis:
I also have an alternate version of the changes which just use `version(FreeBSD11)` and treat 12 as `version(FreeBSD)`, which is slightly shorter, and is more future-proof if those symbols stay the same for FreeBSD 13, but what's here is more in line with how we usually do versioning. I don't know which is ultimately better in this case. Regardless, the changes in https://github.com/dlang/dmd/pull/8567 are required, since we don't currently have version identifiers for any particular version of FreeBSD.

I doubt that this is all of the symbols that need to be changed for full compatibility, but it's enough to pass the druntime and Phobos unit tests and the dmd test suite on my local FreeBSD 11 and 12 boxes.

### DIFF
--- a/src/core/sys/freebsd/sys/event.d
+++ b/src/core/sys/freebsd/sys/event.d
@@ -43,15 +43,33 @@ extern(D) void EV_SET(kevent_t* kevp, typeof(kevent_t.tupleof) args)
     *kevp = kevent_t(args);
 }
 
-struct kevent_t
+version(FreeBSD12)
 {
-    uintptr_t    ident; /* identifier for this event */
-    short       filter; /* filter for event */
-    ushort       flags;
-    uint        fflags;
-    intptr_t      data;
-    void        *udata; /* opaque user data identifier */
+    struct kevent_t
+    {
+        uintptr_t ident;
+        short     filter;
+        ushort    flags;
+        uint      fflags;
+        long      data;
+        void*     udata;
+        ulong[4]  ext;
+    }
 }
+else version(FreeBSD11)
+{
+    struct kevent_t
+    {
+        uintptr_t    ident; /* identifier for this event */
+        short       filter; /* filter for event */
+        ushort       flags;
+        uint        fflags;
+        intptr_t      data;
+        void        *udata; /* opaque user data identifier */
+    }
+}
+else
+    static assert(0, "Unsupported version of FreeBSD");
 
 enum
 {
--- a/src/core/sys/freebsd/sys/mount.d
+++ b/src/core/sys/freebsd/sys/mount.d
@@ -31,8 +31,19 @@ struct fid
 }
 
 enum MFSNAMELEN = 16;
-enum MNAMELEN   = 88;
-enum STATFS_VERSION = 0x20030518;
+
+version(FreeBSD12)
+{
+    enum MNAMELEN   = 1024;
+    enum STATFS_VERSION = 0x20140518;
+}
+else version(FreeBSD11)
+{
+    enum MNAMELEN   = 88;
+    enum STATFS_VERSION = 0x20030518;
+}
+else
+    static assert(0, "Unsupported version of FreeBSD");
 
 struct statfs_t
 {
--- a/src/core/sys/posix/dirent.d
+++ b/src/core/sys/posix/dirent.d
@@ -149,15 +149,33 @@ else version( FreeBSD )
         DT_WHT      = 14
     }
 
-    align(4)
-    struct dirent
+    version(FreeBSD12)
     {
-        uint      d_fileno;
-        ushort    d_reclen;
-        ubyte     d_type;
-        ubyte     d_namlen;
-        char[256] d_name;
+        struct dirent
+        {
+            ino_t     d_fileno;
+            off_t     d_off;
+            ushort    d_reclen;
+            ubyte     d_type;
+            ushort    d_namlen;
+            ushort    d_pad1;
+            char[256] d_name;
+        }
     }
+    else version(FreeBSD11)
+    {
+        align(4)
+        struct dirent
+        {
+            uint      d_fileno;
+            ushort    d_reclen;
+            ubyte     d_type;
+            ubyte     d_namlen;
+            char[256] d_name;
+        }
+    }
+    else
+        static assert(0, "Unsupported version of FreeBSD");
 
     alias void* DIR;
 
--- a/src/core/sys/posix/sys/stat.d
+++ b/src/core/sys/posix/sys/stat.d
@@ -806,35 +806,82 @@ else version( FreeBSD )
 {
     // https://github.com/freebsd/freebsd/blob/master/sys/sys/stat.h
 
-    struct stat_t
+    version(FreeBSD12)
     {
-        dev_t       st_dev;
-        ino_t       st_ino;
-        mode_t      st_mode;
-        nlink_t     st_nlink;
-        uid_t       st_uid;
-        gid_t       st_gid;
-        dev_t       st_rdev;
+        struct stat_t
+        {
+            dev_t     st_dev;
+            ino_t     st_ino;
+            nlink_t   st_nlink;
+            mode_t    st_mode;
+            short st_padding0;
+            uid_t     st_uid;
+            gid_t     st_gid;
+            int st_padding1;
+            dev_t     st_rdev;
 
-        time_t      st_atime;
-        c_long      __st_atimensec;
-        time_t      st_mtime;
-        c_long      __st_mtimensec;
-        time_t      st_ctime;
-        c_long      __st_ctimensec;
+            version(X86) int st_atim_ext;
 
-        off_t       st_size;
-        blkcnt_t    st_blocks;
-        blksize_t   st_blksize;
-        fflags_t    st_flags;
-        uint        st_gen;
-        int         st_lspare;
+            time_t      st_atime;
+            c_long      __st_atimensec;
 
-        time_t      st_birthtime;
-        c_long      st_birthtimensec;
+            version(X86) int st_mtim_ext;
 
-        ubyte[16 - timespec.sizeof] padding;
+            time_t      st_mtime;
+            c_long      __st_mtimensec;
+
+            version(X86) int st_ctim_ext;
+
+            time_t      st_ctime;
+            c_long      __st_ctimensec;
+
+            version(X86) int st_btim_ext;
+
+            time_t      st_birthtime;
+            c_long      st_birthtimensec;
+
+            off_t     st_size;
+            blkcnt_t st_blocks;
+            blksize_t st_blksize;
+            fflags_t  st_flags;
+            ulong st_gen;
+            ulong[10] st_spare;
+        }
     }
+    else version(FreeBSD11)
+    {
+        struct stat_t
+        {
+            uint        st_dev;
+            uint        st_ino;
+            mode_t      st_mode;
+            ushort      st_nlink;
+            uid_t       st_uid;
+            gid_t       st_gid;
+            uint        st_rdev;
+
+            time_t      st_atime;
+            c_long      __st_atimensec;
+            time_t      st_mtime;
+            c_long      __st_mtimensec;
+            time_t      st_ctime;
+            c_long      __st_ctimensec;
+
+            off_t       st_size;
+            blkcnt_t    st_blocks;
+            blksize_t   st_blksize;
+            fflags_t    st_flags;
+            uint        st_gen;
+            int         st_lspare;
+
+            time_t      st_birthtime;
+            c_long      st_birthtimensec;
+
+            ubyte[16 - timespec.sizeof] padding;
+        }
+    }
+    else
+        static assert(0, "Unsupported version of FreeBSD");
 
     enum S_IRUSR    = 0x100; // octal 0000400
     enum S_IWUSR    = 0x080; // octal 0000200
--- a/src/core/sys/posix/sys/types.d
+++ b/src/core/sys/posix/sys/types.d
@@ -147,12 +147,26 @@ else version( FreeBSD )
 {
     // https://github.com/freebsd/freebsd/blob/master/sys/sys/_types.h
     alias long      blkcnt_t;
-    alias uint      blksize_t;
-    alias uint      dev_t;
+
+    version(FreeBSD12)
+    {
+        alias ulong blksize_t;
+        alias ulong dev_t;
+        alias ulong ino_t;
+        alias ulong nlink_t;
+    }
+    else version(FreeBSD11)
+    {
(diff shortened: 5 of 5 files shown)


src/core/sys/freebsd/sys/event.d
```diff
@@ -43,15 +43,33 @@ extern(D) void EV_SET(kevent_t* kevp, typeof(kevent_t.tupleof) args)
     *kevp = kevent_t(args);
 }
 
-struct kevent_t
+version(FreeBSD12)
```
MartinNowak: Let's go with `static if (__FreeBSD_version >= INO64_FIRST)` to avoid this becoming a maintenance task for future FBSD releases. Also see https://github.com/dlang/dmd/pull/8567/files#r211100860.

ibuclaw:
Can we make this `FreeBSD_11` and `FreeBSD_12`?  So that it follows the convention in https://dlang.org/spec/version.html#predefined-versions (if the compiler enforced, these should be automatically reserved).

ibuclaw:
What's the hold-up on this?  I've already implemented `version (FreeBSD_%u)` in gdc.

jmdavis:
> What's the hold-up on this? I've already implemented `version (FreeBSD_%u)` in gdc, but missing druntime bindings makes it difficult to test on FreeBSD.

The main issue is that I haven't been able to work on any programming stuff lately, so I haven't been able to finish sorting this out. The secondary issue is that after the initial version of this PR, changes were made to the build scripts for dmd which make it build programs and run them as part of the build process. Prior to that, it was possible to use the FreeBSD 11 build of dmd to build dmd on FreeBSD 12, and it then built a fully functional version of dmd, druntime, and Phobos with the FreeBSD 12 bindings. After the build stuff was changed, that didn't work anymore, because the build stuff built a program and ran it using the existing version of dmd on the system, and that crashed, because it was generating a FreeBSD 11 binary, since it hadn't built the FreeBSD 12 version of dmd yet.

That in and of itself doesn't necessarily make what this PR is doing wrong, but it does mean that you can't just build dmd on FreeBSD 12, which complicates the upgrade process considerably. The correct way to handle that needs to be figured out, but I haven't gotten around to dealing with it like I should have by now, since I've been battling headache issues that get in the way of me doing much programming. I'm hoping to get that sorted out soon, but in the interim, it's meant that this PR has languished when it shouldn't have.

ibuclaw:
> The main issue is that I haven't been able to work on any programming stuff lately, so I haven't been able to finish sorting this out.

As I have a working FreeBSD 11.x and 12.x on both i686 and x86_64 with gdc, I guess there'd be no objection to rebasing and pushing all changes independently?  (I've already started on the 53-bit rounded real support in [druntime](https://github.com/dlang/druntime/pull/3266) and [phobos](https://github.com/dlang/phobos/pull/7689)).

> The secondary issue is that after the initial version of this PR, changes were made to the build scripts for dmd which make it build programs and run them as part of the build process. Prior to that, it was possible to use the FreeBSD 11 build of dmd to build dmd on FreeBSD 12, and it then built a fully functional version of dmd, druntime, and Phobos with the FreeBSD 12 bindings. After the build stuff was changed, that didn't work anymore, because the build stuff built a program and ran it using the existing version of dmd on the system, and that crashed, because it was generating a FreeBSD 11 binary, since it hadn't built the FreeBSD 12 version of dmd yet.

Right, so dmd is not good as a cross-compiler.  I guess this is not surprising, though I'm not so sure about the extent of the limitation.  By starting off with a FBSD11 compiler on FBSD11, you could build a FBSD12 compiler and compatible library by passing DFLAGS on the build command-line to request targeting a specific version of FreeBSD.

I've already said that gcc can only target one major version of FreeBSD per compiler (so you have one separate compiler and runtime for each version of FreeBSD you want to target).  How dmd wants to achieve FreeBSD 12 support can be ratified later without blocking druntime support.

jmdavis:
@ibuclaw The main hiccup with making druntime changes without having dmd properly supporting FreeBSD 12 is that someone (I forget who) committed changes to druntime which forces the mangling to match the FreeBSD 11 versions of the symbols for at least some of the critical functions. This allows the current dmd and druntime to build and work on FreeBSD 12 by using the FreeBSD 11 symbols but won't work in the long term, since those will be going away with FreeBSD 13 IIRC. Changing druntime to have the correct FreeBSD 12 symbols would break that unless the versioning were such that dmd always grabbed the FreeBSD 11 versions until it could properly handle the FreeBSD 12 versions. Of course, that doesn't mean that changes can't be made to allow gdc to properly handle the FreeBSD 12 symbols with druntime, but it complicates things if we don't want to break it for the people currently using dmd on FreeBSD 12 with the old symbols. That being said, we may need gdc or ldc to be able to cross-compile with the correct FreeBSD 12 symbols in order to get dmd working with the updated symbols on FreeBSD 12 anyway given that dmd's build process is now building programs other than dmd as part of building dmd.

ibuclaw:
> @ibuclaw The main hiccup with making druntime changes without having dmd properly supporting FreeBSD 12 is that someone (I forget who) committed changes to druntime which forces the mangling to match the FreeBSD 11 versions of the symbols for at least some of the critical functions. 

IIRC, it's not mangling that's different, but some other ABI feature. So `pragma(mangle)` won't work with gdc and ldc to get the correct resolved symbol - we need a new `pragma(symver, "foo@VER_1")` to be added.

jmdavis:
@ibuclaw What's different is the structs like `stat_t` (IIRC, the main difference is that they increased the inode size to 64-bit from 32-bit). So, if you call a function like `statfs` on FreeBSD 12, you have to pass it the newer version of the struct, and if you pass it the old version, you'll get a crash. However, as part of the migration, the old function symbols exist for FreeBSD 11 binaries to use on FreeBSD 12 (making it possible to run a FreeBSD 11 binary on FreeBSD 12 in many cases). You won't link to those functions normally (since it's just supposed to be for binaries built on FreeBSD 11), but it's possible if you mess with the mangling to make it match the FreeBSD 11 mangling. e.g. core.sys.freebsd.sys.mount currently has

    pragma(mangle, "statfs@FBSD_1.0")     int statfs(const char*, statfs_t*);

IIRC, internally, those libc functions convert the struct to the correct FreeBSD 12 struct before making the actual system call, but it's just a workaround for the migration. Normally, when you link to `statfs` on FreeBSD 12, you get the FreeBSD 12 version, and you need the FreeBSD 12 version of `stat_t`. So, adding the pragmas to druntime allowed druntime to work on FreeBSD 12 with the FreeBSD 11 versions of the structs, but what we really need is for druntime to use the FreeBSD 11 versions of the structs on FreeBSD 11 and the FreeBSD 12 versions of the structs on FreeBSD 12. And that of course requires us to be able to version the bindings based on the version of FreeBSD that you're building for, which is the point of this PR. The fact that you can run FreeBSD 11 binaries on FreeBSD 12 even initially allowed this PR to work without truly cross-compiling, but having dmd now build other stuff using the existing Phobos library as part of the build instead of waiting to build other programs until it had built the new dmd and the new Phobos library screws things up. The newly built programs end up using the FreeBSD 12 bindings with the Phobos built on FreeBSD 11, so boom. A proper cross-compiler obviously wouldn't have that problem.

kinke:
> So pragma(mangle) won't work with gdc and ldc to get the correct resolved symbol - we need a new pragma(symver, "foo@VER_1") to be added.

AFAIK, this works just fine with LDC; see the `__sym_compat` macro in https://people.freebsd.org/~deischen/symver/library_versioning.txt. A recent LDC built on FreeBSD 12 was successfully tested to be working on FreeBSD 13 ([ref](https://github.com/ldc-developers/ldc/pull/3464#issuecomment-643600364)), and I expect the binaries to work on FreeBSD 11 too - exactly because of using older version of C symbols which have changed with recent FreeBSD versions, and thus no hassle with shipping different binaries for each FreeBSD version.

> This allows the current dmd and druntime to build and work on FreeBSD 12 by using the FreeBSD 11 symbols but won't work in the long term, since those will be going away with FreeBSD 13 IIRC.

I doubt these are supposed to just vanish soon. I don't use FreeBSD, but there's probably not much point in having such a versioning scheme when the plan is to quickly drop older symbols and so backwards compatibility.

jmdavis:
> I doubt these are supposed to just vanish soon. I don't use FreeBSD, but there's probably not much point in having such a versioning scheme if the plan is to quickly drop older symbols and so backwards compatibility.

In general, FreeBSD doesn't promise backwards compatibility as far as built binaries go. You can usually build the same software on two different versions of FreeBSD and have that software work on the version that it's built on, but they do not promise binary compatibility across major versions. You typically have to build software for the version of FreeBSD that you're running it on. There are times that it works (especially if you're not relying on much more than libc, since the core functions and types don't change often), but it's not guaranteed to do so, because they do make breaking API and ABI changes across major versions. They approached the situation with 64-bit inodes a bit differently, because it's a core part of libc, and they wanted to make the transition as smooth as possible. If nothing else, it made it much easier to build software while FreeBSD 12 was being worked on prior to release.

But regardless of how long the old symbols get left in, you're stuck with the smaller inode size if you're not using the current symbols, and the intention is that any software built on or for FreeBSD 12 or later use the new symbols. It's a non-issue for C/C++ programs, because they use the libc headers, and it all just works, but since we use our own bindings for D, it gets messier - especially since our bindings currently assume that the libc APIs are the same for all versions, which is not necessarily true. We've mostly gotten away with it, because they don't change often, and POSIX OSes tend to avoid changing them unless they have to, but they do sometimes change.

kinke:
> But regardless of how long the old symbols get left in, you're stuck with the smaller inode size if you're not using the current symbols, and the intention is that any software built on or for FreeBSD 12 or later use the new symbols. It's a non-issue for C/C++ programs, because they use the libc headers, and it all just works, but since we use our own bindings for D, it gets messier - especially since our bindings currently assume that the libc APIs are the same for all versions, which is not necessarily true. We've mostly gotten away with it, because they don't change often, and POSIX OSes tend to avoid changing them unless they have to, but they do sometimes change.

I'd usually advocate a simple support for non-EOL OS versions and so the FreeBSD 11 symbols as common denominator in this case, and waiting one more year for the FreeBSD 12 symbols incl. inode64 support, when FreeBSD 11 reaches EOL. On Windows, we support every version since Vista IIRC, and for macOS, the min OS version the binaries will run on is trivially set via the `MACOSX_DEPLOYMENT_TARGET` env var (currently `10.9` for LDC).

I take it we wouldn't have to expect a lot of `version` cluttering for FreeBSD's libc bindings due to not that many differing symbols, but as versions cannot be integer values, there would probably still be ugliness in the form of
```D
version (FreeBSD_11) enum minFreeBSD = 11;
else version (FreeBSD_12) enum minFreeBSD = 12;
else version (FreeBSD_13) enum minFreeBSD = 13;
...

static if (minFreeBSD >= 12)
    // yay, inode64
```

jmdavis:
@kinke dmd officially doesn't support anything older than Windows 7 (since IIRC, that's what the auto-tester runs), and technically even that's EOL now.

In any case, regardless of whether we wait for FreeBSD 11 to go EOL before updating to the new API, there's still the fundamental issue that the actual OS is versioned and has header changes, but we only have a single version of the bindings, and the issue of how to move from the old bindings to the new is more or less the same.

> I take it we wouldn't have to expect a lot of version cluttering for FreeBSD's libc bindings due to not that many differing symbols, but as versions cannot be integer values, there would probably still be ugliness in the form of

Well, at Martin's request, I created a version of this PR which adds the symbol `__FreeBSD_version` which grabs its value from `getosreldate` so that instead of using the OS major version, you use the ABI version number (which is the same throughout the life of the OS major version but changes during its development). I thought that I'd updated this PR to do that, but it looks like that's when I hit the problem with the dmd build not working, because it started building programs that used Phobos, so I hadn't fully tested it. It has the advantage of making it so that an OS based off of FreeBSD CURRENT (e.g. TrueOS) could work when there was a libc change like with 64-bit inodes, whereas that becomes more problematic when you go off of the OS major version numbers. I'm not married to either approach, but going off of the ABI version number is arguably more correct.

kinke:
> there's still the fundamental issue that the actual OS is versioned and has header changes, but we only have a single version of the bindings

Yes, but that's not unique for FreeBSD - we also only have a single binding for all MSVC versions, which has seen major breaking changes too. Some functions missing in older versions are substituted by weak implementations in druntime, and struct declarations like `FILE` simply assume a runtime 2015+.

The current MSVC solution ported to FreeBSD wrt. inode64 here would be to use the new structs and provide custom weak implementations for functions like `statfs@FBSD_1.5`, translating to `statfs@FBSD_1.0`. Then the user doesn't have to have 3 precompiled libphobos for FreeBSD[11-13]; the OS version (for MSVC: Visual C++ SDK version) when linking a program determines the referenced libc symbols and thus the minimum runnable version. A program linked on FreeBSD 11 should also run on 12, but inode64 will only be used when linking on 12 or newer.

That would be a greater effort, and if such binary compatibility isn't expected in the FreeBSD world anyway, probably not really worth it.

ibuclaw:
> > So pragma(mangle) won't work with gdc and ldc to get the correct resolved symbol - we need a new pragma(symver, "foo@VER_1") to be added.
> 
> AFAIK, this works just fine with LDC; see the `__sym_compat` macro in https://people.freebsd.org/~deischen/symver/library_versioning.txt. A recent LDC built on FreeBSD 12 was successfully tested to be working on FreeBSD 13 ([ref](https://github.com/ldc-developers/ldc/pull/3464#issuecomment-643600364)), and I expect the binaries to work on FreeBSD 11 too - exactly because of using older version of C symbols which have changed with recent FreeBSD versions, and thus no hassle with shipping different binaries for each FreeBSD version.

I've just checked again, and the assembler still rejects it.
```
/tmp//cc8dAC7m.s: Assembler messages:
/tmp//cc8dAC7m.s:19: Error: junk `@@FBSD_1.0' after expression
/tmp//cc8dAC7m.s:23: Error: junk `@FBSD_1.0' after expression
```

kinke:
Seems gcc specific, see https://d.godbolt.org/z/PsdPGT - with LDC, the (Linux) linker complains about `undefined reference to 'statfs', version 'FBSD_1.0'`.

ibuclaw:
> Seems gcc specific, see https://d.godbolt.org/z/PsdPGT - with LDC, the (Linux `ld.gold`) linker complains about `undefined reference to 'statfs', version 'FBSD_1.0'`.

Not gcc, but binutils.  Having a closer look at gcc's support for attribute(symver), it requires binutils 2.35, which is the most currrent release.  Even then, gas-2.35 rejects the asm call to the versioned symbol directly.

ibuclaw:
#3271 has been merged.

PR #2281, opened 2018-08-18
[stable] core.stdc.config: Don't define cpp_[u]long as special enum on 64-bit Posix

kinke:
That was only necessary while DMD was mangling long as C++ `long long` on 64-bit OSX, i.e., in 2.079-2.081.

This is required for 2.082-based LDC to build itself successfully on 64-bit Linux (conflicting `Array!ulong` vs. `Array!__c_ulong` template instantiations with identical C++ mangling...).

### DIFF
--- a/src/core/stdc/config.d
+++ b/src/core/stdc/config.d
@@ -104,11 +104,11 @@ version( Windows )
     enum __c_long  : int;
     enum __c_ulong : uint;
 
-    alias __c_long   cpp_long;
-    alias __c_ulong  cpp_ulong;
-
     alias int   c_long;
     alias uint  c_ulong;
+
+    alias __c_long   cpp_long;
+    alias __c_ulong  cpp_ulong;
 
     alias long  cpp_longlong;
     alias ulong cpp_ulonglong;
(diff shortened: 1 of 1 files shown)


src/core/stdc/config.d
```diff
@@ -104,31 +104,28 @@ version( Windows )
...
 
-    alias __c_long   cpp_long;
-    alias __c_ulong  cpp_ulong;
-
     alias int   c_long;
     alias uint  c_ulong;
 
+    alias __c_long   cpp_long;
+    alias __c_ulong  cpp_ulong;
+
```
schveiguy: Why is this just moved a few lines?

src/core/stdc/config.d
```diff
@@ -137,12 +134,12 @@ else version( Posix )
...
```
schveiguy: Same thing here

src/core/stdc/config.d
```diff
@@ -104,31 +104,28 @@ version( Windows )
...
 
-    alias __c_long   cpp_long;
-    alias __c_ulong  cpp_ulong;
-
     alias long  c_long;
     alias ulong c_ulong;
 
+    alias long   cpp_long;
+    alias ulong  cpp_ulong;
+
```
schveiguy: Are `__c_long` and `__c_ulong` used anywhere else? These are no longer defined, so any usage will break.

src/core/stdc/config.d
kinke: In order to comply with the DDoc order, which is the intuitive one: C long, C++ long, C++ longlong.

src/core/stdc/config.d
kinke: Nope, and they aren't supposed to be. `__c_longlong` is also only defined where needed.

kinke:
> I don't really have understanding for the purpose of these changes

Preventing redundant template instantiations. `__c_long` represents a long mangled as `C++ long`. That's D `long` on all 64-bit Posix platforms now, so e.g. no need to have 2 redundant frontend `Array!T` instantiations (`Array!size_t` => `Array!ulong` vs. `Array!d_uns64` => `Array!__c_ulong`).

PR #2286, opened 2018-08-27
Added std::array<>

TurkeyMan:
First and simplest of the bunch. I'd like to get this in to establish layout and styling for the others.

### DIFF
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -46,6 +46,8 @@ SRCS=\
 	src\core\stdc\time.d \
 	src\core\stdc\wchar_.d \
 	\
+	src\core\stdcpp\xutility.d \
+	\
 	src\core\sync\barrier.d \
 	src\core\sync\condition.d \
 	src\core\sync\config.d \
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -46,6 +46,7 @@ SRCS=\
 	src\core\stdc\time.d \
 	src\core\stdc\wchar_.d \
 	\
+	src\core\stdcpp\array.d \
 	src\core\stdcpp\xutility.d \
 	\
 	src\core\sync\barrier.d \
--- a/posix.mak
+++ b/posix.mak
@@ -244,7 +244,7 @@ HAS_ADDITIONAL_TESTS:=$(shell test -d test && echo 1)
 ifeq ($(HAS_ADDITIONAL_TESTS),1)
 	ADDITIONAL_TESTS:=test/init_fini test/exceptions test/coverage test/profile test/cycles test/allocations test/typeinfo \
 	    test/aa test/hash \
-	    test/thread test/unittest test/imports test/betterc
+	    test/thread test/unittest test/imports test/betterc test/stdcpp
 	ADDITIONAL_TESTS+=$(if $(SHARED),test/shared,)
 endif
 
--- a/src/core/stdcpp/array.d
+++ b/src/core/stdcpp/array.d
@@ -1,0 +1,145 @@
+/**
+ * D header file for interaction with C++ std::array.
+ *
+ * Copyright: Copyright (c) 2018 D Language Foundation
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors:   Manu Evans
+ * Source:    $(DRUNTIMESRC core/stdcpp/array.d)
+ */
+
+module core.stdcpp.array;
+
+// hacks to support DMD on Win32
+version (CppRuntime_Microsoft)
+{
+    version = CppRuntime_Windows; // use the MS runtime ABI for win32
+}
+else version (CppRuntime_DigitalMars)
+{
+    version = CppRuntime_Windows; // use the MS runtime ABI for win32
+    pragma(msg, "std::array not supported by DMC");
+}
+version (CppRuntime_Clang)
+{
+    private alias AliasSeq(Args...) = Args;
+    private enum StdNamespace = AliasSeq!("std", "__1");
+}
+else
+{
+    private enum StdNamespace = "std";
+}
+
+extern(C++, (StdNamespace)):
+
+/**
+ * D language counterpart to C++ std::array.
+ *
+ * C++ reference: $(LINK2 https://en.cppreference.com/w/cpp/container/array)
+ */
+extern(C++, class) struct array(T, size_t N)
+{
+extern(D):
+pragma(inline, true):
+
+    ///
+    alias size_type = size_t;
+    ///
+    alias difference_type = ptrdiff_t;
+    ///
+    alias value_type = T;
+    ///
+    alias pointer = T*;
+    ///
+    alias const_pointer = const(T)*;
+
+    ///
+    alias as_array this;
+
+    /// Variadic constructor
+    this(T[N] args ...)                                             { this[] = args[]; }
+
+    ///
+    size_type size() const nothrow @safe @nogc                      { return N; }
+    ///
+    alias length = size;
+    ///
+    alias opDollar = length;
+    ///
+    size_type max_size() const nothrow @safe @nogc                  { return N; }
+    ///
+    bool empty() const nothrow @safe @nogc                          { return N == 0; }
+
+    ///
+    ref inout(T) front() inout nothrow @safe @nogc                  { static if (N > 0) { return this[0]; } else { return as_array()[][0]; /* HACK: force OOB */ } }
+    ///
+    ref inout(T) back() inout nothrow @safe @nogc                   { static if (N > 0) { return this[N-1]; } else { return as_array()[][0]; /* HACK: force OOB */ } }
+
+    ///
+    void fill()(auto ref const(T) value)                            { this[] = value; }
+
+    version (CppRuntime_Windows)
+    {
+        ///
+        inout(T)* data() inout nothrow @safe @nogc                  { return &_Elems[0]; }
+        ///
+        ref inout(T)[N] as_array() const inout @safe @nogc          { return _Elems[0 .. N]; }
+        ///
+        ref inout(T) at(size_type i) inout nothrow @safe @nogc      { return _Elems[0 .. N][i]; }
+
+        version (CppRuntime_Microsoft)
+        {
+            import core.stdcpp.xutility : MSVCLinkDirectives;
+            mixin MSVCLinkDirectives!false;
+        }
+
+    private:
+        T[N ? N : 1] _Elems;
+    }
+    else version (CppRuntime_Gcc)
+    {
+        ///
+        inout(T)* data() inout nothrow @safe @nogc                  { static if (N > 0) { return &_M_elems[0]; } else { return null; } }
+        ///
+        ref inout(T)[N] as_array() inout nothrow @trusted @nogc     { return data()[0 .. N]; }
+        ///
+        ref inout(T) at(size_type i) inout nothrow @trusted @nogc   { return data()[0 .. N][i]; }
+
+    private:
+        static if (N > 0)
+        {
+            T[N] _M_elems;
+        }
+        else
+        {
+            struct _Placeholder {}
+            _Placeholder _M_placeholder;
+        }
+    }
+    else version (CppRuntime_Clang)
+    {
+        ///
+        inout(T)* data() inout nothrow @trusted @nogc               { static if (N > 0) { return &__elems_[0]; } else { return cast(inout(T)*)__elems_.ptr; } }
+        ///
+        ref inout(T)[N] as_array() inout nothrow @trusted @nogc     { return data()[0 .. N]; }
+        ///
+        ref inout(T) at(size_type i) inout nothrow @trusted @nogc   { return data()[0 .. N][i]; }
+
+    private:
+        static if (N > 0)
+        {
+            T[N] __elems_;
+        }
+        else
+        {
+            struct _ArrayInStructT { T[1] __data_; }
+            align(_ArrayInStructT.alignof)
+            byte[_ArrayInStructT.sizeof] __elems_ = void;
+        }
+    }
+    else
+    {
+        static assert(false, "C++ runtime not supported");
+    }
+}
--- a/src/core/stdcpp/xutility.d
+++ b/src/core/stdcpp/xutility.d
@@ -1,0 +1,60 @@
+/**
+ * D header file for interaction with Microsoft C++ <xutility>
+ *
+ * Copyright: Copyright (c) 2018 D Language Foundation
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors:   Manu Evans
+ * Source:    $(DRUNTIMESRC core/stdcpp/xutility.d)
+ */
+
+module core.stdcpp.xutility;
+
+extern(C++, "std"):
+
+version (CppRuntime_Microsoft)
+{
+    // Client code can mixin the set of MSVC linker directives
+    mixin template MSVCLinkDirectives(bool failMismatch = false)
+    {
+        import core.stdcpp.xutility : __CXXLIB__;
+
+        static if (__CXXLIB__ == "libcmtd")
+        {
+            pragma(lib, "libcpmtd");
+            static if (failMismatch)
+                pragma(linkerDirective, "/FAILIFMISMATCH:RuntimeLibrary=MTd_StaticDebug");
+        }
+        else static if (__CXXLIB__ == "msvcrtd")
+        {
+            pragma(lib, "msvcprtd");
+            static if (failMismatch)
+                pragma(linkerDirective, "/FAILIFMISMATCH:RuntimeLibrary=MDd_DynamicDebug");
+        }
+        else static if (__CXXLIB__ == "libcmt")
+        {
+            pragma(lib, "libcpmt");
+            static if (failMismatch)
+                pragma(linkerDirective, "/FAILIFMISMATCH:RuntimeLibrary=MT_StaticRelease");
+        }
+        else static if (__CXXLIB__ == "msvcrt")
+        {
+            pragma(lib, "msvcprt");
+            static if (failMismatch)
+                pragma(linkerDirective, "/FAILIFMISMATCH:RuntimeLibrary=MD_DynamicRelease");
+        }
+    }
+
+    // convenient alias for the C++ std library name
+    enum __CXXLIB__ = __traits(getTargetInfo, "cppRuntimeLibrary");
+
+package:
+    // these are all [[noreturn]]
+    void _Xbad() nothrow @trusted @nogc;
+    void _Xinvalid_argument(const(char)* message) nothrow @trusted @nogc;
+    void _Xlength_error(const(char)* message) nothrow @trusted @nogc;
+    void _Xout_of_range(const(char)* message) nothrow @trusted @nogc;
+    void _Xoverflow_error(const(char)* message) nothrow @trusted @nogc;
+    void _Xruntime_error(const(char)* message) nothrow @trusted @nogc;
+}
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -5,6 +5,8 @@ TESTS:=array
 .PHONY: all clean
 all: $(addprefix $(ROOT)/,$(addsuffix .done,$(TESTS)))
 
+# osx32 does not link properly, nobody uses it anyway...
+ifneq ($(OS)$(MODEL), osx32)
 $(ROOT)/%.done : $(ROOT)/%
 	@echo Testing $*
 	$(QUIET)$(TIMELIMIT)$(ROOT)/$* $(RUN_ARGS)
@@ -12,11 +14,10 @@ $(ROOT)/%.done : $(ROOT)/%
 
 $(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
 	mkdir -p $(dir $@)
-	# osx32 does not link properly, nobody uses it anyway...
-	if [ $(OS)$(MODEL) = "osx32" ]; then
+	
 	$(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
 	$(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
-	fi
+endif
 
 clean:
 	rm -rf $(GENERATED)
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -14,7 +14,6 @@ $(ROOT)/%.done : $(ROOT)/%
 
 $(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
 	mkdir -p $(dir $@)
-	
 	$(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
 	$(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
 endif
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -13,9 +13,9 @@ $(ROOT)/%.done : $(ROOT)/%
 $(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
 	mkdir -p $(dir $@)
 	# osx32 does not link properly, nobody uses it anyway...
-	ifneq ($(OS)$(MODEL),osx32)
+	ifneq ($(OS)$(MODEL), osx32)
             $(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
-	    $(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
+            $(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
         endif
 
 clean:
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -13,10 +13,10 @@ $(ROOT)/%.done : $(ROOT)/%
 $(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
 	mkdir -p $(dir $@)
 	# osx32 does not link properly, nobody uses it anyway...
-	ifneq ($(OS)$(MODEL), osx32)
+	if [($(OS)$(MODEL) = "osx32")
 	$(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
 	$(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
-	endif
+	fi
 
 clean:
 	rm -rf $(GENERATED)
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -3,10 +3,17 @@ include ../common.mak
 TESTS:=array
 
 .PHONY: all clean
-all: $(addprefix $(ROOT)/,$(addsuffix .done,$(TESTS)))
 
 # osx32 does not link properly, nobody uses it anyway...
-ifneq ($(OS)$(MODEL), osx32)
+ifeq ($(OS)$(MODEL), osx32)
+
+all:
+
+clean:
+
+else
+all: $(addprefix $(ROOT)/,$(addsuffix .done,$(TESTS)))
+
 $(ROOT)/%.done : $(ROOT)/%
 	@echo Testing $*
 	$(QUIET)$(TIMELIMIT)$(ROOT)/$* $(RUN_ARGS)
@@ -16,7 +23,7 @@ $(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
 	mkdir -p $(dir $@)
 	$(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
 	$(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
-endif
 
 clean:
 	rm -rf $(GENERATED)
+endif
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -13,7 +13,7 @@ $(ROOT)/%.done : $(ROOT)/%
 $(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
 	mkdir -p $(dir $@)
 	# osx32 does not link properly, nobody uses it anyway...
-	if [($(OS)$(MODEL) = "osx32")
+	if [$(OS)$(MODEL) = "osx32"]
 	$(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
 	$(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
 	fi
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -1,0 +1,19 @@
+include ../common.mak
+
+TESTS:=array
+
+.PHONY: all clean
+all: $(addprefix $(ROOT)/,$(addsuffix .done,$(TESTS)))
+
+$(ROOT)/%.done : $(ROOT)/%
+	@echo Testing $*
+	$(QUIET)$(TIMELIMIT)$(ROOT)/$* $(RUN_ARGS)
+	@touch $@
+
+$(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
+	mkdir -p $(dir $@)
+	$(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
+	$(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
+
+clean:
+	rm -rf $(GENERATED)
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -12,8 +12,11 @@ $(ROOT)/%.done : $(ROOT)/%
 
 $(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
 	mkdir -p $(dir $@)
-	$(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
-	$(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
+	# osx32 does not link properly, nobody uses it anyway...
+	ifneq ($(OS)$(MODEL),osx32)
+            $(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
+	    $(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
+        endif
 
 clean:
 	rm -rf $(GENERATED)
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -14,9 +14,9 @@ $(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
 	mkdir -p $(dir $@)
 	# osx32 does not link properly, nobody uses it anyway...
 	ifneq ($(OS)$(MODEL), osx32)
-            $(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
-            $(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
-        endif
+	$(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
+	$(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
+	endif
 
 clean:
 	rm -rf $(GENERATED)
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -13,7 +13,7 @@ $(ROOT)/%.done : $(ROOT)/%
 $(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
 	mkdir -p $(dir $@)
 	# osx32 does not link properly, nobody uses it anyway...
-	if [$(OS)$(MODEL) = "osx32"]
+	if [ $(OS)$(MODEL) = "osx32" ]; then
 	$(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
 	$(QUIET)$(DMD) $(DFLAGS) -main -unittest -of$@ $< $(ROOT)/$*_cpp.o
 	fi
--- a/test/stdcpp/src/array.cpp
+++ b/test/stdcpp/src/array.cpp
@@ -1,0 +1,20 @@
+#include <array>
+
+std::array<int, 5> fromC_val(std::array<int, 5>);
+std::array<int, 5>& fromC_ref(std::array<int, 5>&);
+
+std::array<int, 5>& sumOfElements_ref(std::array<int, 5>& arr)
+{
+    int r = 0;
+    for (size_t i = 0; i < arr.size(); ++i)
+        r += arr[i];
+    arr.fill(r);
+    return arr;
+}
+
+std::array<int, 5> sumOfElements_val(std::array<int, 5> arr)
+{
+    int r = sumOfElements_ref(arr)[0] + fromC_ref(arr)[0] + fromC_val(arr)[0];
+    arr.fill(r);
+    return arr;
+}
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,6 @@
 root = true
 
-[*.{c,h,d,di,dd}]
+[*.{c,cpp,h,d,di,dd}]
 insert_final_newline = true
 indent_style = space
 indent_size = 4
--- a/changelog/array.dd
+++ b/changelog/array.dd
@@ -1,0 +1,3 @@
+Added `core.stdcpp.array`.
+
+Added `core.stdcpp.array`, which links against C++ `std::array`
--- a/mak/COPY
+++ b/mak/COPY
@@ -48,6 +48,7 @@ COPY=\
 	$(IMPDIR)\core\stdc\wchar_.d \
 	$(IMPDIR)\core\stdc\wctype.d \
 	\
+	$(IMPDIR)\core\stdcpp\array.d \
 	$(IMPDIR)\core\stdcpp\exception.d \
 	$(IMPDIR)\core\stdcpp\typeinfo.d \
 	$(IMPDIR)\core\stdcpp\xutility.d \
--- a/mak/COPY
+++ b/mak/COPY
@@ -48,8 +48,9 @@ COPY=\
 	$(IMPDIR)\core\stdc\wchar_.d \
 	$(IMPDIR)\core\stdc\wctype.d \
 	\
-	$(IMPDIR)\core\stdcpp\typeinfo.d \
 	$(IMPDIR)\core\stdcpp\exception.d \
+	$(IMPDIR)\core\stdcpp\typeinfo.d \
+	$(IMPDIR)\core\stdcpp\xutility.d \
 	\
 	$(IMPDIR)\core\sys\darwin\execinfo.d \
 	$(IMPDIR)\core\sys\darwin\pthread.d \
--- a/mak/DOCS
+++ b/mak/DOCS
@@ -38,6 +38,7 @@ DOCS=\
 	$(DOCDIR)\core_stdc_wchar_.html \
 	$(DOCDIR)\core_stdc_wctype.html \
 	\
+	$(DOCDIR)\core_stdcpp_array.html \
 	$(DOCDIR)\core_stdcpp_exception.html \
 	$(DOCDIR)\core_stdcpp_typeinfo.html \
 	\
--- a/mak/WINDOWS
+++ b/mak/WINDOWS
@@ -190,6 +190,9 @@ $(IMPDIR)\core\stdcpp\exception.d : src\core\stdcpp\exception.d
 $(IMPDIR)\core\stdcpp\typeinfo.d : src\core\stdcpp\typeinfo.d
 	copy $** $@
 
+$(IMPDIR)\core\stdcpp\xutility.d : src\core\stdcpp\xutility.d
+	copy $** $@
+
 $(IMPDIR)\core\sys\darwin\execinfo.d : src\core\sys\darwin\execinfo.d
 	copy $** $@
 
--- a/mak/WINDOWS
+++ b/mak/WINDOWS
@@ -187,6 +187,9 @@ $(IMPDIR)\core\stdc\wctype.d : src\core\stdc\wctype.d
 $(IMPDIR)\core\stdcpp\exception.d : src\core\stdcpp\exception.d
 	copy $** $@
 
+$(IMPDIR)\core\stdcpp\array.d : src\core\stdcpp\array.d
+	copy $** $@
+
 $(IMPDIR)\core\stdcpp\typeinfo.d : src\core\stdcpp\typeinfo.d
 	copy $** $@
 
--- a/test/common.mak
+++ b/test/common.mak
@@ -28,3 +28,7 @@ else
 	DFLAGS += -O -release
 	CFLAGS += -O3
 endif
+CXXFLAGS:=$(CFLAGS) -std=c++11
+ifeq (osx,$(OS))
+	CXXFLAGS+=-stdlib=libc++
+endif
--- a/test/stdcpp/src/array_test.d
+++ b/test/stdcpp/src/array_test.d
@@ -1,0 +1,62 @@
+import core.stdcpp.array;
+
+extern (C++) int test_array()
+{
+    array!(int, 5) arr;
+    arr[] = [0, 2, 3, 4, 5];
+    ++arr.front;
+
+    assert(arr.size == 5);
+    assert(arr.length == 5);
+    assert(arr.max_size == 5);
+    assert(arr.empty == false);
+    assert(arr.front == 1);
+
+    assert(sumOfElements_val(arr)[0] == 160);
+    assert(sumOfElements_ref(arr)[0] == 15);
+
+    array!(int, 0) arr2;
+    assert(arr2.size == 0);
+    assert(arr2.length == 0);
+    assert(arr2.max_size == 0);
+    assert(arr2.empty == true);
+    assert(arr2[] == []);
+
+    return 0;
+}
+
+
+extern(C++):
+
+// test the ABI for calls to C++
+array!(int, 5) sumOfElements_val(array!(int, 5) arr);
+ref array!(int, 5) sumOfElements_ref(return ref array!(int, 5) arr);
+
+// test the ABI for calls from C++
+array!(int, 5) fromC_val(array!(int, 5) arr)
+{
+    assert(arr[] == [1, 2, 3, 4, 5]);
+    assert(arr.front == 1);
+    assert(arr.back == 5);
+    assert(arr.at(2) == 3);
+
+    arr.fill(2);
+
+    int r;
+    foreach (e; arr)
+        r += e;
+
+    assert(r == 10);
+
+    arr[] = r;
+    return arr;
+}
+
+ref array!(int, 5) fromC_ref(return ref array!(int, 5) arr)
+{
+    int r;
+    foreach (e; arr)
+        r += e;
+    arr[] = r;
+    return arr;
+}
--- a/test/stdcpp/win64.mak
+++ b/test/stdcpp/win64.mak
@@ -1,0 +1,12 @@
+# built from the druntime top-level folder
+# to be overwritten by caller
+DMD=dmd
+MODEL=64
+DRUNTIMELIB=druntime64.lib
+CC=cl
+
+test:
+	"$(CC)" -c /Foarray_cpp.obj test\stdcpp\src\array.cpp /EHsc
+	"$(DMD)" -of=test.exe -m$(MODEL) -conf= -Isrc -defaultlib=$(DRUNTIMELIB) -main -unittest test\stdcpp\src\array_test.d array_cpp.obj
+	test.exe
+	del test.exe test.obj array_cpp.obj
--- a/win64.mak
+++ b/win64.mak
@@ -102,6 +102,11 @@ test_aa:
 test_hash:
 	$(DMD) -m$(MODEL) -conf= -Isrc -defaultlib=$(DRUNTIME) -run test\hash\src\test_hash.d
 
+test_stdcpp:
+	$(MAKE) -f test\stdcpp\win64.mak "DMD=$(DMD)" MODEL=$(MODEL) "VCDIR=$(VCDIR)" DRUNTIMELIB=$(DRUNTIME) "CC=$(CC)" test
+
+test_all:  test_uuid test_aa test_hash test_stdcpp
+
 ################### zip/install/clean ##########################
 
 zip: druntime.zip
@@ -121,4 +126,4 @@ clean:
 
 auto-tester-build: target
 
-auto-tester-test: unittest test_uuid test_aa test_hash
+auto-tester-test: unittest test_all


mak/SRCS
```diff
@@ -46,6 +46,8 @@ SRCS=\
 	src\core\stdc\time.d \
 	src\core\stdc\wchar_.d \
 	\
+	src\core\stdcpp\array.d \
```
TurkeyMan: Why aren't any other the other existing files in here?

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,91 @@
...
+    alias difference_type = ptrdiff_t;
+    alias value_type = T;
+    alias reference = ref T;
+    alias const_reference = ref const(T);
+    alias pointer = T*;
+    alias const_pointer = const(T)*;
+
+    alias as_array this;
+
+    extern(D) size_type size() const nothrow @safe @nogc                    { return N; }
```
Geod24: You can just stick an `extern(D):` at the top. Also I would explain in the doc why those are `extern(D)`, as it might be confusing to users/reviewers.

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,91 @@
...
+ * Source:    $(DRUNTIMESRC core/stdcpp/array.d)
+ */
+
+module core.stdcpp.array;
+
+alias array = std.array;
+
+extern(C++, std):
+
+extern(C++, class) struct array(T, size_t N)
```
Geod24: Documentation

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,91 @@
...
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors:   Manu Evans
+ * Source:    $(DRUNTIMESRC core/stdcpp/array.d)
+ */
+
+module core.stdcpp.array;
+
+alias array = std.array;
```
Geod24: That's not necessary

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,91 @@
+﻿/**
+ * D header file for interaction with C++ std::array.
+ *
+ * Copyright: Manu Evans 2014 - 2018.
```
Geod24: That'd be problematic I think

Geod24:
Can you add ABI tests ?

src/core/stdcpp/array.d
TurkeyMan: It really is!

src/core/stdcpp/array.d
TurkeyMan: True.. what's the proper copyright assignment?
I just copied what I found in neighbouring code.

src/core/stdcpp/array.d
TurkeyMan: Yeah, I thought about that... but then I thought it might be too easy to accidentally mis-attribute any proper extern's.

Geod24: Any proper `extern(C++)` will have it in front of the prototype, won't it ?
I remember doing this for a project of my own and not having to define any `extern(C++)` anyway because everything gets inlined.

src/core/stdcpp/array.d
Geod24: Can you elaborate ?

src/core/stdcpp/array.d
Geod24: To the "D Language Foundation"

TurkeyMan:
I don't really know how to test this. It involves interaction with all the possible C compilers, and the build/CI scripts are chaos.
Perhaps someone (you?) could amend this patch to include a framework for testing these STL patches, and then I can populate with tests and proliferate to vector/string/etc.

src/core/stdcpp/array.d
TurkeyMan: https://forum.dlang.org/thread/lppkkcsyvlvhebmlapat@forum.dlang.org
Knock yourself out!

TurkeyMan: TL;DR, Walter says.

TurkeyMan: TL;DR, slightly longer: you can't import 2 such modules because the namespace scopes conflict with eachother. Also, the namespace called 'std' conflicts with phobos. Basically, we just need to alias all these things out into the module scope (where they should be anyway) so that anything is accessible.

JinShil:
Why are we putting C++ bindings in druntime?  Can't these exist in Dub package that doesn't require the D Language Foundation to take ownership of it, and assume the burden of maintaining it?

TurkeyMan:
This was a D Language Foundation project from a while back. There was a hired employee working on this.
I think he quit, this was incomplete, and I'm resurrecting the effort.

I guess it feels natural that they exist beside `core.stdc`??

There's no way a dub package will maintain correct sync, and these files, possibly more than any other files in the distribution, definitely need to run their CI and tests together with the releases to guarantee ABI match.

TurkeyMan:
Anything less will be chaos. These bindings are of critical importance, and really need to 'just work'. An external repo will never keep sync. I don't think they can live anywhere else.

JinShil:
I realize someone decided long ago that they should be here in druntime;  I'm challenging that decision.

I'm not convinced that a Dub repository would not work.  Even if the D Language Foundation chose to take ownership of these bindings, they could exist in a different repository or https://github.com/D-Programming-Deimos/

I think the only reason why the C standard library bindings are here is because they are used by some of the druntime internals.  If druntime has no need for these bindings internally, I don't think they belong here.  Even if druntime did use them internally, there are other ways to go about that.

I ask we strive to keep the scope of druntime to be just what is necessary to implement the D language, and nothing more.

TurkeyMan:
They're also not like other 'libraries', these aren't subject to active development. They're not 'hot', they'll never change once they're correct, so it doesn't suffer dev characteristics like in-development phobos modules that are being spun up in dub packages.

TurkeyMan:
C++ is definitely not a 'deimos' binding. This isn't 'just some lib' we're making available to D users for convenience... as long as `extern(C++)` exists in the language, we're making a commitment to get the ABI right, and the stdc++ ABI is a critical part of the compiler vendor's responsibility, and therefore a part of the core language's responsibility to maintain compatibility. There's nowhere the C++ library is distributed separately to the toolchain, and the same applies here.

Consider GDC updates to the next version of GCC, or we upgrade the win builds to VS2019, we need tests to fail, and to correct the ABI in lockstep with toolchain development and release. druntime is the only place for that.

Point is, stdc++ is not a 'lib', it is effectively core language.

src/core/stdcpp/array.d
jacob-carlborg: Yes, `extern(D)` is confusing to me.

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,125 @@
...
+ *    (See accompanying file LICENSE)
+ * Authors:   Manu Evans
+ * Source:    $(DRUNTIMESRC core/stdcpp/array.d)
+ */
+
+module core.stdcpp.array;
+
+// accepted solution to the C++ namespace problem is to alias symbols outside of the namespace
+// https://forum.dlang.org/thread/lppkkcsyvlvhebmlapat@forum.dlang.org
+alias array = std.array;
```
jacob-carlborg: Missing Ddoc comment.

JinShil:
A properly implemented systems programming language should not depend on C compilers, C++ compilers, or any other language's toolchain; it should only need the linker.  The fact that D has a dependency on the C toolchain (at least for now), is a unfortunate drag; please, let's not make it worse.

C++ ABI testing is the responsibility of our CIs.  Our CI infrastructure can incorporate a C++ toolchain to verify D's implementation.  The CIs could even import the stdc++ bindings from another repository for testing, just as we when testing Dub packages with each change to DMD.

Furthermore, we don't have the resources for this.  @TurkeyMan, if this is important to you, then we need someone like you to take ownership of it, and not just leave it for the rest of us to deal with.  We're already way understaffed and the issues and PRs just keep coming.  If it's yours you won't need to wait on or negotiate with anyone to get things done; and you'll be much more productive.  Once you have a repository created, we can incorporate it into our CIs like we do the other Dub repositories to ensure DMD remains compatible.

src/core/stdcpp/array.d
TurkeyMan: If these functions are extern(C++), they may conflict with actual C++ implementations of these functions. Locally defined functions should be D-mangled.

But what does it matter, you guys are rejecting it all anyway :/

TurkeyMan:
I really can't see it that way... I would call the opposite; a properly implemented systems language absolutely does have intimate access to the c runtime, it's like the core native lib for any system.

What resources does this require? I'm already carrying this.
If you're talking about ongoing maintenance, that's exactly the point, it is legitimate baggage bourne by `extern(C++)`, they're inseparable. If any language or build/tooling change causes the STL tests to fail, then it's effectively a 'badcodegen' regression, and should be fixed urgently, and it should hopefully not be possible that a toolchain release can happen without a fix.

I expect these are very low-maintenance once they work. This is system code that almost never changes, and in the rare event they do, that's the whole point.

JinShil:
@TurkeyMan, I'm not trying to discourage you or anyone else from doing this work; I would love to see this work done, I just don't believe it belongs in the druntime repository.  Also I think progress on these bindings would be MUCH more accelerated if they were not here.  In fact, putting them here will likely be a hindrance (as you are currently experiencing, and have experienced with your other PRs).

Alternatives include, but are not limited to:

- A Dub package owned by you or some other group of individual D programmers (This is my current preference).
- A Deimos repository (better than druntime, but still requires allocating resources from dlang)
- A separate repository with the dlang organization (though I don't think that is much different than a Deimos repository)

Please note that even if these bindings are in a different repository, they can still be included in the D compiler distribution, if someone believes that would be appropriate for some reason.

> What resources does this require?

The maintenance for this work requires creating the bindings (yes, the majority of the work does happen up front, but it doesn't stop there), updating them as needed, responding to questions from users (on the forum, in bugzilla issues, and in PRs), debugging issues, devising solutions to any issues, implementing those solutions as PRs, reviewing those PRs for coding conventions, tests, coverage, documentation, and correct implementation, and negotiating with contributors.  All that while trying to keep up with the backlog of issues and PRs we already have competing for our attention.

Furthermore, these bindings require expertise and experience with C++ and stdc++.  While most of us probably have *some* experience there, it's not a prerequisite for maintaining druntime.  To assume ownership of this, we require **active** team members that are very familiar with the C++ and the stdc++.  I can't even count on one hand enough team members that would qualify as **active** let alone having stdc++ experience.  Perhaps we could add you to the team as code owner, @TurkeyMan.  Are you willing to assume that responsibility, answer questions, handling issues, reviewing PRs, negotiating solutions with contributors, etc.?

Even if we did have you, or a team with the right expertise and resources to create and maintain these bindings, I still don't believe the bindings belong here.  Simply put, druntime has a different charter.  There are  many things in druntime that don't actually belong here.  We should be trying to get rid of those things, not add to them.

And where does it stop?  We've recently had PRs for adding bindings and `version` identifiers for the musl C library.  Will we eventually need to maintain bindings ulibc++ too?  IMO, none of this stuff belongs in druntime.  Team druntime should be focused on implementing the D language, and that's all.

> If you're talking about ongoing maintenance, that's exactly the point, it is legitimate baggage bourne by extern(C++), they're inseparable. If any language or build/tooling change causes the STL tests to fail, then it's effectively a 'badcodegen' regression, and should be fixed urgently, and it should hopefully not be possible that a toolchain release can happen without a fix.

`extern(C++)` is useful for any C++ bindings, not just stdc++.  I don't believe by having `extern(C++)` in the language we're obligated to have stdc++ bindings.  Even if we were, by someone's subjective judgement, they can still exist in a different repository.  We can ensure quality and compatibility by incorporating the external repository in our CI.

"Perfection is attained not when there is nothing left to add, but when there is nothing left to take away." -- Antoine de Saint-Exupery

With all that said, note that I am just one person on this team expressing my opinion and exercising whatever influence I might have.  I do not speak for the whole team; each will have to voice their own opinions.  I don't expect anyone to agree with me, but that doesn't mean I shouldn't articulate my point of view.  I've already had my say, and maybe even said to much.  I defer to those more senior than I for a decision.

jacob-carlborg:
If we are going to add bindings for the C++ standard library, does that mean we're going to add bindings for the Objective-C frameworks?

Geod24:
I have nothing to add that @JinShil haven't said yet.

It's an awesome initiative and I'm looking forward to have it available. I think it should live somewhere special (dlang-community?) and be a `dub` package, and quickly added to the CI.

@TurkeyMan : Your main concern seems to be about codegen regressions. It is my understanding Buildkite would solve it as efficiently as putting the code in druntime would. Am I missing something ?

TurkeyMan:
I don't, and don't intend to ever use dub :/

If it's added to the CI, then what's the practical difference from placing it here? It will have the same effect in terms of maintenance.
How do I guarantee that GDC and LDC's test environments also match the DMD configuration? It just sounds like duplicating that work 3 times rather than slotting it in where it's already performed correctly, and requiring no additional work to invoke the CI.
It also sounds like a lot of extra work to bundle a separate repo into releases than to just slot it in here, where it would be automatic.

I still can't really understand the argument. It's about maintenance, but initial development of 3rd party interaction is a lot of additional scripting, and ongoing maintenance are all only harder if it's in some 3rd party location...

> We can ensure quality and compatibility by incorporating the external repository in our CI.

If you accept this, then I don't understand how integrating another library parallel to druntime with the same build, CI, and distribution semantics as druntime isn't more work... it just... is more work, for the same outcome no?

Anyway. I yield.
Why did Andrei and Walter put it here int he first place?
Should we remove the existing modules?

JinShil:
> If you accept this, then I don't understand how integrating another library parallel to druntime with the same build, CI, and distribution semantics as druntime isn't more work... it just... is more work, for the same outcome no?

Each repository has different team members.  IMO the goals of druntime and the skills required to maintain druntime do not have significant overlap with stdc++ bindings.  stdc++ bindings should be in a repository that has team members with stake in their success and focus on stdc++ bindings' goals.

> Why did Andrei and Walter put it here in the first place?

I suspect that since the C standard library bindings are also in druntime (unfortunately), they just assumed this would be a good place for stdc++ bindings too.  I'm not sure they put much consideration into the decision.  But, that's just how I see it; they will have to speak for themselves.

> Should we remove the existing modules?

IMO, most definitely YES!

> Anyway. I yield.

No reason to yield just yet.  I'm just voicing my opinion.  Others may disagree, and @WalterBright and @andralex may choose to overrule me.

@WalterBright and @andralex, I am challenging the decision to put stdc++ bindings here in the druntime repository.

Please see my arguments in the comments above, but in summary...
- I don't think druntime's charter includes the maintenance of such bindings, and even if it did, I think it should be reconsidered.
- druntime does not have any intrinsic dependency on stdc++
- There are already too many things in druntime that shouldn't really be there; we should be looking for ways to remove dependencies and liabilities, not add to them.
- I don't believe the skill set required to develop and maintain druntime has significant overlap with the skill set required to develop and maintain stdc++ bindings.  We should have a different team for stdc++ bidning.
- We're spread too thin as it is, and I don't think we have the resources for this.  I think those that are interested in these bindings should take ownership of them and develop them independently, not pass the burden onto team druntime.
- I think it will be a hindrance to the success of stdc++ bindings to place them here.  Progress would be much more accelerated if they could be developed independently of druntime by a different team that has stake in their success.
- We can still incorporate stdc++ bindings, from their external repository, into our CIs to ensure quality and compatibility.  We already do this several Dub packages.
- stdc++ bindings can still be shipped with the the compiler's distribution package from their external repository, if someone believes that would be beneficial.
- I would prefer that the dlang organization not assume the burden of developing and maintaining stdc++ bindings.  Instead, I suggest the community take ownership of them and maintain them in a Dub package.  If however, you decide the the dlang organization should take ownership of them, I suggest putting them in Deimos or in a separate dlang repository with a team suited to their development and success.

If you feel you have already made this decision, please reconsider.  I will abide by whatever you decide.

TurkeyMan:
Oh... give me a break! ;) .. You've made your point.
I'll do something else. Although I'll never work out how to script up the build solution, in all DMD, GDC, LDC processes...

TurkeyMan:
Fwiw, I do buy into your philosophy that "druntime" is not the right place for a lot of this stuff... That's just the only library that lives so close to the compiler. I wonder if there's room for a 'core' lib that's not "druntime"...?

andralex:
The matter is being overblown. Technically (and sadly not found by anyone), the problem is that the code currently is not very useful because it won't work with C++ types with copy constructors and destructors. So at the very least on our side the template should be limited to plain old data types.

The second is organizational. On the face of it it's not terribly controversial to place C++ bindings in the runtime support library. The code depends on minute compiler implementation details (such as function call ABIs, mangling etc) and it makes sense to sync the release of foreign language bindings with releases of the compiler. That is the case, I imagine, with most other languages offering a foreign interface. (E.g. I don't think Python delegates C API bindings to a library etc.) But then placing the C++ bindings in an external library is ALSO not a bad idea, especially because it offers us the possibility to toy with it for a while until we get the kinks out. Things may be a tad more difficult (match it with a specific dmd/cc pair) but hey it's a matter of engineering. So either way, it's not a huge deal. We can work it out - if we only wanted to.

This brings up the third matter - managerial. Picture Walter and I were both managing @TurkeyMan and @JinShil. Two very capable and well intended senior engineers. Both Walter and I have a history of suggesting them high-impact projects, suggestions that were declined. So we gave up on that and decided to take whatever they'd want to work on. Now @JinShil 's arguments in this PR, however amplified, are sensible, and so the normal managerial course of action would be "cool, @TurkeyMan can you please take it to dub for now and we can always move it to our repo after we accumulate some experience with it". But we have an unexpected problem - @TurkeyMan flat out refuses to use dub, which suddenly puts us in an impasse. This we cannot solve with traditional managerial techniques (sit down and discuss the matter, remind him of basic employment obligations etc) because this is not a traditional work situation.

So we have an instance of a pattern we often witness:

* We can't ask people to do important things.
* Any initiative is met with endless hand-wringing upon the least consequential aspects
* If Walter and I try to just lay down the law everybody digs in for an unending trench warfare
* Either way, a complete standstill is soon to ensue
* Essential technical matters (as is here with copy ctors and dtors) are completely forgotten in favor of the Bollywood drama
* No work is getting done

If anyone has an idea on how to fix this, I'd be in their debt. Because I'm sure out of tricks.

Geod24:
In this situation, the "dub package" does not require dub to be used. Personally I'm using Makefiles + `git submodules` and just have to configure the include path.
@TurkeyMan : What build system do you use ? How do you usually deal with libraries ?

jacob-carlborg:
> I don't know how to create a library whose CI is run by, and is bundled with GDC, LDC, DMD... it'll just sit in a repo somewhere for people that want it, and it'll be maintained on-demand.

You can add your project here [1]. It will be tested for each pull request to the dlang (DMD, Phobos, druntime, etc.) repositories, at least for DMD.

> It might be nice if dub had a package fetch feature which would fetch packages to some shared location and make them all automatically available to the compilers... like /usr/include works. That'd be useful to all build environments.

Dub provides the `fetch` command that will download a package to a specified directory:

```
dub fetch <package> --cache=local --root=/usr
```

Although that will create a new directory in `/usr` name `<package-x.y.z>`, which might not be exactly what you want.

[1] https://github.com/dlang/ci/blob/21cb22242e0bd807c6428f9fc76e31b418280612/buildkite.sh#L94-L99

andralex:
@TurkeyMan this is a confusion - the drama is in the review stage, and the "no work getting done" would be better put as "no progress getting done in spite of the work investment". We have many cases in which work is invested and then wasted in review. Apologies for not making that clearer.

I continue to think that C++ stdlib interface is well placed in druntime, based upon on @TurkeyMan and my arguments. Experience with the C stdlib suggests the maintenance costs tend to be low.

(Nevertheless: @TurkeyMan if you want to work on something else, finishing SIMD support or your DIP with rvalue binding may be good things to look at.)

Regarding the technical issue: `array` is meant to pass arrays of objects across a C++/D interface. So the element type of the array must be representable in both C++ and D. Currently we don't have a solution for representing (on the D side) C++ objects that are not PODs, hence the limitation.

TurkeyMan:
> if you want to work on something else

Sorry, this is another confusion; I mean "do something else" as in, not merge it to druntime... I'll 'do' [host/test/CI] something else.

TurkeyMan:
> You can add your project here [...] at least for DMD

DMD covers the smallest surface area by far.

> Dub provides the `fetch` command

That's not really different than just cloning or submodule-ing the repo... it doesn't register the include or link paths with the compiler anywhere. You're left to manually hook up the paths and perform any dependent builds manually.

apt-get would be more useful.

TurkeyMan:
> your DIP with rvalue binding

I think it's done. Is there something I'm meant to be doing with it?
I don't know what more I can do; I feel it's over-specified as it is. Other dips are a quarter of the size, but I expect this one will be criticised far more harshly than others.

> Currently we don't have a solution for representing (on the D side) C++ objects that are not PODs, hence the limitation.

You mean while doing zero-work on the D side? Ie, a pure-extern interface? Copy-ctor is the only missing piece I think, although elaborate move can be a problem. Certainly, bining C++ objects to D remains non-trivial.

That doesn't mean this STL work isn't immediately useful, and a massive enabler; in my experience to date, I just expect I must do some work on the D side. That usually looks like adding a post-blit to the D-side definition of the object. Not the end of the world, and much better than recent history, where construction and destruction were completely broken, and assignment (part of the rule-of-3/5) and other operators weren't working.

Elaborate move, and classes by-value are a problem that I don't have a suggestion for, but that's parallel to this work.

Certainly, your copy-ctor DIP is extremely welcome! That'll unlock no-manual-intervention bindings for a large number of objects.

andralex:
@TurkeyMan all I'm thinking is a stopgap:

```
extern(C++, class) struct array(T, size_t N)
if (...........)
```

thus limiting the types with which array can be instantiated. That way we limit invalid/undocumented/dangerous uses.

TurkeyMan:
That's not a terrible idea... but I have no idea how to define `.............`?
`is(T == class) || isPrimitiveType || isPod || (needsPostblit && hasPostblit)`
That last part seems hard to define...

ZombineDev:
@JinShil I imagine that if we had a monorepo (where all top-level folders are part of the build and subsequently distribution), this PR would be a much smaller point of contention:
```sh
dmd/
sys/      # system specific bindings like Linux uapi, Win32, macOS/iOS, WebAssembly
libc/     # Bindings to the different libc implementations
druntime/
libcxx/ # Bindings to different C++ `std::` implementations
phobos/
tools/
dlang.org/
installer/
```
(The repos are listed in their order of dependence.)

Note that we're already using [`CODEOWNERS`](
https://github.com/dlang/druntime/blob/master/CODEOWNERS), so maintainers can scope their participation to the areas of expertise/that they're interested in.

So in the absence of a monorepo hosting all core D projects, how about we add a `libcxx/` to the dlang organization. The point against hosting the C++ `std` bindings outside the organization is that they need to be part of the distribution and would be much harder to maintain as part of the existing dlang infrastructure. Also AFAIK, LDC and GDC are already using druntime and phobos as git submodules,  so another git submodule we easily slot into place.

JinShil:
> The point against hosting the C++ std bindings outside the organization is that they need to be part of the distribution

Do they though?  Why can't users just get these bindings from Dub?  I think they are useful for verifying compatibility with DMD's ABI and mangling implementations, but that can be done with the CI and Dub.  I don't think the dlang organization needs to assume ownership of them.

> Note that we're already using CODEOWNERS, so maintainers can scope their participation to the areas of expertise/that they're interested in.

I don't think the CODEOWNERS idea is working out well.  It reminds me of the old adage "When everyone is responsible, noone is".  Or "You can't kick a collective butt".  Also, I definitely don't think the dlang organization should assume ownership of the stdc++ bindings unless 3 or 4 people step up and volunteer to take responsibility for their development and maintenance.

> how about we add a libcxx/ to the dlang organization

I'd still prefer that the dlang organization not assume ownership of stdc++ bindings, and would even encourage the dlang organization to release ownership of a few things it currently has (e.g. [https://dlang.org/phobos/index.html](The Phobos Data Formats).

However, if the dlang organization does choose to assume ownership, your suggestion to have a separate `libcxx`/`libc++`/`stdc++`/`Insert-your-favorite-name` repository is my preference.

> Also AFAIK, LDC and GDC are already using druntime and phobos as git submodules, so another git submodule we easily slot into place.

While I'd prefer druntime not have anything to do with stdc++, if I'm overruled, the suggestion to include it as a submodule is my preference.

ZombineDev:
> Do they though? [..]

Isn't this what you suggested earlier:

> stdc++ bindings can still be shipped with the the compiler's distribution package from their external repository, if someone believes that would be beneficial.

As long as have a "distribution" and not give the Dub the role of a toolchain manager, e.g. fetching phobos and other libraries, I think it would be highly beneficial for D's success.

> I'd still prefer that the dlang organization not assume ownership of stdc++ bindings, and would even encourage the dlang organization to release ownership of a few things it currently has

No. While I'm all for better modularity, ownership of code critical for D's success needs to be a burden on the whole organization. Not taking ownership of such critical things implies not taking ownership in our users' success.

Druntime may not have anything to do with stdc++, except for the exception handling (and that's why I put it as something independent of drt and phobos), but the compiler has everything to do with it. The repos and CI infrastructure of dmd, gdc and ldc needs to work in tight synchronization in order for `extern (C++)` interop to work.

Reusing the existing CI infra is what's number one priority here and while Dub + Buildkite is great, it covers a much smaller portion of our targets (i386/x86_64 Windows, macOS, FreeBSD, plus everything else that LDC and GDC support) and unless you offer to work on extending Buildkite with more platforms, it is simply not an option for something as platform specific as stdc++.

JinShil:
> Isn't this what you suggested earlier:

I qualified my statement with 2 important phrases:  "can be" and "if someone believes that would be beneficial".  What I'm implying is that that is up the the package distributor or release manager.  If the dlang organization wants to distribute stdc++ bindings with their release, I think they should add logic to https://github.com/dlang/installer to pull it from its (preferable, IMO) external repository.  Though we already release Dub with the distribution, so why can't users just run a `dub` command to get it?

JinShil:
> Not taking ownership of such critical things implies not taking ownership in our users' success.

While I agree with the sentiment, and I do want the dlang organization to act professionally on behalf of its users, users' success is not entirely the responsibility of the dlang organization.  It's a partnership, with responsibilities on both sides.

> but the compiler has everything to do with it. The repos and CI infrastructure of dmd, gdc and ldc needs to work in tight synchronization in order for extern (C++) interop to work.

The compiler definitely has support for the C++ ABI, mangling, etc..., but that applies equally to _all_ C++ libraries.  As long as those features are properly unit tested in our CIs, I think the dlang organization has fulfilled its obligation.  It would certainly be wise to add stdc++ bindings to our CIs for quality assurance, but we should also consider Qt, Boost, or other C++ bindings for the same purpose.  Certainly stdc++ is more prominent than those libraries, but I don't think that justifies the dlang organization assuming ownership of it.  And our first line of defense should be DMD's test suite, not the stdc++ bindings.  We should have enough coverage in our test suite to cover all of DMD's C++ features to where the stdc++ tests are unnecessary.  That is probably easier said than done, so adding stdc++ tests to our CIs is a good idea, but they should not supplant our need test DMD's C++ features in its test suite.

> Reusing the existing CI infra is what's number one priority here and while Dub + Buildkite is great, it covers a much smaller portion of our targets (i386/x86_64 Windows, macOS, FreeBSD, plus everything else that LDC and GDC support) and unless you offer to work on extending Buildkite with more platforms, it is simply not an option for something as platform specific as stdc++.

The dlang organization does own the C++ features in DMD, so we do have a responsibility to sufficiently test those features, but those features should largely be tested by the DMD test suite.  It's no different than what we are doing now:  We unit test individual bugs and features with the DMD test suite (auto-tester and friends), and then we do an _ecosystem test_ in Buildkite with many of the prominent Dub packages as supplemental testing.  stdc++ belongs in that supplemental _ecosystem test_,  We should already have tests in the DMD test suite to verify all of DMD's C++ features with or without stdc++.

The dlang organzization should test DMD's C++ features, but testing the stdc++ bindings is the responsibility of their developers.  The dlang organization is only responsible for the compiler implementation, and we should already have 100% test coverage for those features in DMD's test suite.  The stdc++ tests are just extra credit.

TurkeyMan:
> Qt, Boost, or other C++ libs

The thing is, none of those libs have back-doors into the name mangling, they don't have intrinsic data structures, they don't have builtin compiler logic that populates the object files with additional bonus linker directives. stdc++ is special and unique; it's only part library, but it's also a meaningful part actual core language specified by nothing other than the respective compiler's implementation.

Have a browse from here on down: https://github.com/dlang/dmd/blob/master/src/dmd/cppmangle.d#L490

There's lots of magic like that. No other C++ library can possibly have that affect on the core language, because a library can't inject code into the compiler.
stdc++ is **not a library**, it is effectively core language.

Geod24:
@JinShil : In fact, the stdc++ binding *are* a bit special.

The ABI used on POSIX define a special mangling for the `std` namespace, as well as some primitives (see [the blue text at the end of this section](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling-compression)).

To test that this works properly with the presence of other C++ mangling features (compression, substitution) requires some work which would be better put directly in the binding.

Regarding ownership, I think popular packages are critical to the success of D and was very glad https://github.com/dlang-community/ was started, which is where I was hoping this new repository could be put. But putting it directly under `dlang` would not be a stretch, for the reasons mentioned earlier.

Having it under `dlang` does not mean it will put the burden on our contributors more than having Phobos under dlang puts a burden on you.

andralex:
>In fact, the stdc++ binding are a bit special.

That is confirmed by @WalterBright 's and my experience looking into it. There are any number of compiler-intimate hacks related to C++, including classes and namespaces having special names in the mangler, and features using backdoor intrinsics. It does seem glomming onto those is the charter of the D compiler distro.

TurkeyMan:
@ZombineDev `... if (__traits(isPOD, T))`

But, that doesn't work for not-pod's... that's the part that I don't know how to define.

JinShil:
> Does that process exist, across DMD, GDC, LDC?

From what I can tell DMD's C++ features are part of the compiler's frontend (though I welcome someone to correct me on that, if I'm wrong).  GDC and LDC are both using copies of DMD's frontend, so they will automatically inherit tested DMD implementations. They also employ copies of Phobos, and I don't see why the same procedure can't be used for stdc++.  

@wilzbach and @MartinNowak seem to be our CI wizards, so they would have to comment with their own perspective on the matter, but we already do a lot of testing to ensure compatibility between the 3 compilers and that upstream changes don't break things downstream. (e.g. https://semaphoreci.com/dlang/dmd-2/branches/pull-request-8629/builds/3).  I think team-stdc++ should shoulder most the burden for developing the testing infrastructure of the bindings themselves, but I recognize that they will require team-dmd and team-ci to play a supporting role and to serve as integrators.

JinShil:
Keep in mind, too, that if stdc++ were it's own repository, it will likely have it's own set of dedicatd CIs.

TurkeyMan:
> [...] so they will automatically inherit tested DMD implementations [...]

Tests are build-specific. It's possible that every single build of every single compiler for every single arch produce different results. They all need to run the complete tests.

JinShil:
OK, so we can see that there is cooperation between multiple teams.  We already do that today, and I don't see why we can't cooperate on stdc++ as well.

Getting back to the topic that started this conversation, I don't think stdc++ belongs in druntime.  I believe it can exist as its own repository inside or outside of the dlang organization.  If the dlang organization can recruit a small team, I have no problem with the dlang organization assuming ownership.

There are still details to be worked out with regard to testing and integration.  If stdc++ is its own repository, much of the infrastructure for testing the bindings themselves can be built by the stdc++ team in their repository.  With that is in place, I think the path forward for integration testing with the various compilers will become more apparent and the different teams can begin working together much as they do now to ensure success.

DMD, LDC, and GDC currently manage their release distributions independently and I suspect they will want to decide for themselves how to best incorporate stdc++ into their toolchain's distribution.

For now, I recommend getting a stdc++ repository started, recruit a team, get the repository's CIs started, and when the first significant milestone is reached, begin discussions for integration.  It will be easier to work out the details when there is something concrete to work with.

TurkeyMan:
>much of the infrastructure for testing the bindings themselves can be built by the stdc++ team in their repository

Yeah... see, I won't do that. I'm not trying to be obtuse, I'm just telling you what's true. I'd rather iterate on this work than re-implement the druntime build and test environment. I can't think of anything more tedious; we have devops people do that sort of thing.
100% of that infrastructure is already in place, exactly where it needs to be, right here in druntime. That's why the PR is here.

But I've already conceded on Monday, I'll commit to some other repo, and whoever cares enough can do something from there. I'm satisfied by my work, I wrote it for myself... but it certainly won't be tested, and it probably won't work for others.... those linux or android or apple users will need to contribute fixes. If CI fails, I'll fix it, but if CI doesn't fail, I won't fix it... and there's no CI ;)

Tell me the module namespace to use... I guess `core.stdcpp` is inappropriate.

Shall I close this PR?
I'll close this PR...

JinShil:
> Tell me the module namespace to use

I would prefer `stdcpp` for the root package name.  But that's just me.

jpf91:
I guess I'm late to this discussion, but there are some points I want to make:

> There is no other place which will build and test against GDC6, 7, 8, on all architectures, on LDC against its myriad of architectures, on DMD which proves MSVC, and DMC... It sounds like a truly gigantic scripting task to reproduce druntime's build environment, where druntime is already in exactly the right place at the right time to tick all the boxes.

> Experience with the C stdlib suggests the maintenance costs tend to be low.

Both of these statements are probably true from an upstream perspective. But as @klickverbot already said:
> "Downstream", i.e. the GDC and LDC people, currently do all that work for "you".

Whenever people change anything in `core.stdc` or anything else architecture specific, this won't be noticed in DMD. Instead the process is as following:
* Commit code to upstream druntime/phobos which breaks ARM
* Code passes on DMD test infrastructure
* Wait for next code merge into GDC/LDC
* Broken code breaks GDC/LDC merge
* Merge is delayed, complain to upstream people to fix
* a) Somebody fixes the issue. This is the best possible case.
* b) Person who commited initial broken commit is not reachable, or can not reproduce/test ARM
* Code merge is further delayed
* GDC / LDC people have to fix the code. This is a real problem for us as our manpower is way more limited than upstream and we don't want to spend it fixing upstream bugs.
* We submit bug fix for upstream
* We probably early-merge the bugfix into GDC, even if this introduces a deviation from upstream, simply to get our builds passing again and avoid further stalls in merging. 

As @ibuclaw now merges the master branch weekly, such problems will at least get noticed earlier and it's more likely to reach the original commiter. Nevertheless I'm obviously opposed in further expanding druntime's complexity by introduction of code which is highly likely to break on GDC/LDC targets but not on upstream. This is effectively shifting the maintenance burden for non-upstream architectures to GDC/LDC developers.

If instead, you have a library which is tested externally, you can do testing with gdc/ldc and spot problems immediately, before even merging a pull request. This is a huge benefit imho.

For library such as `core.stdcpp` which are tightly coupled to the compiler and which are likely to break because of compiler bugs, we obviously have to test the other way round as well. I.e. DMD/GDC/LDC test infrastructure should still test stdcpp on every compiler commit. But keeping sources external makes sure we only have to fix the compiler bugs and library bugs are fixed by original authors.

Now, to be fair, I have to admit that thorough GDC testing is not easily available for external projects. Even our GDC CI is mostly work in progress. What I want to do for GDC is publish nightly build packages for all architectures. Although we don't have much of a build infrastructure for all architectures, my plan is to integrate with [build.opensuse.org](https://build.opensuse.org/project/show/home:jpf91) to build rpm and deb nightlies for many architectures. This will make it very easy for CIs to use the latest gdc.

However, for non-X86 architectures you still need machines to test on. For GDC and simple smoke-testing I'll probably use qemu VMs with buildkite (https://bierbaumer.net/qemu/) but this is probably too slow for real tests. Cross-compilers are another option, but you can't do runtime testing. Probably cross-compiler+qemu is fast enough?

Either way, we should try to work on a many-architectures CI setup, whatever it may look like. Even for 'well-supported' architectures such as ARM our test coverage across D code (dub packages) is probably minimal. And right now, even ARM druntime doesn't compile in GDC master, even though the fix is probably trivial (https://build.opensuse.org/package/live_build_log/home:jpf91/gdc/Fedora_28/armv7l) ;-)

WalterBright:
There are sound arguments on both sides. But I'm in the camp of it belongs in druntime:

1. core.stdc is already in druntime, and there are no complaints about it
2. C++ interoperability is a hugely significant feature for D
3. It requires custom compiler support to work
4. Users will clearly expect it to "just work" without requiring a separate install
5. No worries about the compiler getting out of sync with the C++ support
6. Being builtin has been enormously successful with Ddoc and unittests, it's a game changer for adoption. Even though those should logically be separate tools.
7. Being builtin greatly increases the motivation to get it and keep it working
8. It's not going to break anything

I'm excited that Manu has picked up the flag on this. He's highly motivated, has the technical chops, writes code at a furious pace, and will eat his own dogfood. Therefore I am motivated to defer to his judgement as much as I can.

Note: I oppose any other part of druntime importing the C++ modules, because if the user code is not connecting to C++ code, the C++ runtime library should never need to be linked in. D should never have a creeping dependency on C++ libraries.

WalterBright:
@andralex let me know he agrees. This matter is resolved and let's let @TurkeyMan get on with it, and help him be successful.

Ahead warp factor 4, Mr. Evans!

ibuclaw:
> However, for non-X86 architectures you still need machines to test on. For GDC and simple smoke-testing I'll probably use qemu VMs with buildkite (https://bierbaumer.net/qemu/) but this is probably too slow for real tests. Cross-compilers are another option, but you can't do runtime testing. Probably cross-compiler+qemu is fast enough?

Problems I've found with qemu is finding an emulator that actually works for the platform you're targetting.  ARM works well, even for user emulation mode, others seem hopelessly outdated.

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,125 @@
...
+        extern(D) ref inout(T)[N] as_array() const inout @safe @nogc        { return _Elems[0 .. N]; }
+
+    private:
+        import core.stdcpp.xutility : _Xout_of_range;
+
+        T[N ? N : 1] _Elems;
+
+        extern(D) void _Xran() const @safe @nogc                            { _Xout_of_range("invalid array<T, N> subscript"); }
+    }
+    else version(CRuntime_Glibc)
```
ibuclaw: Nit, you shouldn't be using this version.  C++ bindings are neither CRuntime nor Glibc!

TurkeyMan: Well... yeah, true I guess... what would you suggest?
I mean, libstdc++ is maintained by the same GNU lot as glibc isn't it? Are they separable? I think libstdc++ is just as intrinsically connected to the toolchain as glibc is no?

TurkeyMan:
> I'm obviously opposed in further expanding druntime's complexity by introduction of code which is highly likely to break on GDC/LDC targets but not on upstream. 

@jpf91 Well, on the plus side, I'm probably one of relatively few people who is genuinely sensitive to GDC/LDC compatibility. As long I have access to a toolchain build that is broken, and CI works (ie, can reproduce broken code) I'll fix it. At least, I can fix problems with the lib... if the lib exposes problems with `extern(C++)` codegen, then they're already legit bugs, and precisely the problems I want to make sure we're catching in the first place.

src/core/stdcpp/array.d
jpf91: `libstdc++` is part of GCC, but `glibc` is an external project with different maintainers. It is perfectly possible to use `musl` C library instead of `glibc` and link GCC and `libstdc++` against it. Then you have `musl+libstdc++`. Same should be true for other C libraries (dietlibc, uclibc).

jpf91:
> Well, on the plus side, I'm probably one of relatively few people who is genuinely sensitive to GDC/LDC compatibility. 

@TurkeyMan That's probably true, so let's just keep this in druntime for now as agreed upon by Walter.

src/core/stdcpp/array.d
TurkeyMan: So, you're saying `libstdc++` is actually closer to the compiler than glibc?
So, should the versions check toolchains rather than cruntimes?
Should we make new versions?

jpf91: > Should we make new versions?

I think so. We also can't use the DMD/LDC/GDC versions as some compilers may use different c++ libraries (I think GCC always uses libstdc++, even on windows. clang however uses libstdc++ for compatibility reasons on linux, but libc++ on OSX and probably the microsoft c++ library for windows). So I think there's no way around new versions.

TurkeyMan: Okay. What's the complete set of likely C++ implementations, and what naming scheme should we roll with?

TurkeyMan:
@WalterBright I'm sure there's 100 more hurdles... this one was surprisingly epic.
FWIW; I can't really work around this: https://issues.dlang.org/show_bug.cgi?id=19179

andralex:
@TurkeyMan word

TurkeyMan:
Wow. Thanks.

TurkeyMan:
I don't know how to structure tests for this that prove all platforms and DMD/MSVC.

TurkeyMan:
Also, this should be amended to use the pending C++ versions in review.

TurkeyMan:
Don't merge this... I need to update it against all the recent tech.

TurkeyMan:
I've added a patch that inserts some test skeletons, but I have no idea at all how to rig up the makefiles to build and execute the tests appropriately for all targets.

Can someone who has good familiarity with the test ecosystem please draft up the required machinery (just a makefile? I dunno) and push it to this branch so I can take that as example?
Once I have such an example, I can take it and work it from there...

TurkeyMan:
I literally can't even work out how to RUN the tests locally, let alone add support for C++ building and having it work on all test targets :/

andralex:
@RazvanN7 @edi33416 can you please help @TurkeyMan?

wilzbach:
> I literally can't even work out how to RUN the tests locally

```
make -f posix.mak unittest
```

(or just `unittest-debug` for faster builds)

> let alone add support for C++ building and having it work on all test targets :/

Yeah that's pretty tricky with druntime's outdated testsuite. Probably the easiest for now would be to limit it for Linux-only.

To add a new test folder, you will need to:
- add the folder (e.g. `test/cpp`)
- add a Makefile (see e.g. https://github.com/dlang/druntime/blob/master/test/betterc/Makefile for a simple one). The targets `all` and `clean` are run by the global Makefile
- add e.g. `test/cpp` to the global Makefile:

https://github.com/dlang/druntime/blob/4f86f98ccfb7e7e3a5ea33cbe36fe88d7a461161/posix.mak#L245-L246

- the testsuite will stop and fail if any of the targets that are called by `all` fails

Does this helps a bit?

TurkeyMan:
1. I don't run linux. Windows makefile to run tests doesn't seem to work... at least, I have no idea how to configure my environment to run it properly.

2. I got that far with trying to make a makefile, but from there, I have no idea how to rig up C++ compile jobs. Every target platform builds with different options, and different compilers, and I can't see where/how to rig up the selective logic to drive that stuff.
I tried to lift from DMD's tests as inspiration, but they're structured totally differently.

The whole thing where I can't even work out how to run the tests at all doesn't help, let alone write new and more complex tests ;)

TurkeyMan:
Why does the documentation generate nothing:
http://dtest.dlang.io/artifact/website-826ea94809d870328bd98d9aeac2472b657d0c49-92ce4b6ec9882abaab02a12f1f17e8fe/web/library-prerelease/core/stdcpp/array.html

??

rainers:
>     1. Windows makefile to run tests doesn't seem to work... at least, I have no idea how to configure my environment to run it properly.

The only windows test run for druntime is test/uuid invoked by target test_uuid in druntime/win64.mak.

TurkeyMan:
Why does the documentation generate nothing:
http://dtest.dlang.io/artifact/website-826ea94809d870328bd98d9aeac2472b657d0c49-92ce4b6ec9882abaab02a12f1f17e8fe/web/library-prerelease/core/stdcpp/array.html

??

TurkeyMan:
> The only windows test run for druntime is test/uuid invoked by target test_uuid in druntime/win64.mak.

But... the auto-tester seems to run something...
Is it that the druntime tests should be in DMD? It looks like it runs the DMD tests after building druntime. How do I add a test do DMD that depends on code in druntime? the DMD PR and druntime PR would be mutually exclusive.

rainers:
> But... the auto-tester seems to run something...

It runs the unittests of the druntime library, but not the tests in the test-folder (but test/uuid on win64).

> Is it that the druntime tests should be in DMD? It looks like it runs the DMD tests after building druntime. How do I add a test do DMD that depends on code in druntime? the DMD PR and druntime PR would be mutually exclusive.

For the auto-tester, all tests are run for any PR no matter what repo is targeted. So a test in dmd/test is run for druntime PRs, too. But making changes in both repos simultaneously is not well supported.

For this PR I'd suggest adding druntime/test/stdcpp and adding it to posix.mak (see ADDITIONAL_TESTS) and win64.mak (see auto-tester-test).

rainers:
The windows failure is caused by missing copy rules in mak/WINDOWS.

wilzbach:
From DAutotest:

```
object.Exception@../tools/changed.d(251): Changelog entries should consist of one title line, a blank separator line, and a description.
```

posix.mak
```diff
@@ -312,6 +312,13 @@ test/%/.run: test/%/Makefile $(DMD)
 		DRUNTIME=$(abspath $(DRUNTIME)) DRUNTIMESO=$(abspath $(DRUNTIMESO)) LINKDL=$(LINKDL) \
 		QUIET=$(QUIET) TIMELIMIT='$(TIMELIMIT)' PIC=$(PIC)
 
+################### tests ######################################
```
rainers: The existing posix tests are listed in variable ADDITIONAL_TESTS and expect just a Makefile in the test folder.

TurkeyMan: Oh right. Wow.
This setup is ridiculous, that's no consistency at all. This is **sooooo** time consuming >_<

TurkeyMan:
Holy crap! It built and ran everywhere! O_O

posix.mak
rainers: Agreed. The legacy is using underpowered DigitalMars make on Windows, though. I think there shouldn't be two ways to run tests in posix.mak.

rainers:
> Holy crap! It built and ran everywhere! O_O

Congrats.

For testing against dmc, I guess @braddr would have add stlport to the compiler.

TurkeyMan:
I'd prefer to see the DMC test added as a follow-up.
Is there anything holding this up? I think it's finally there...

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,158 @@
...
+pragma(inline, true):
+
+    ///
+    alias size_type = size_t;
+    ///
+    alias difference_type = ptrdiff_t;
+    ///
+    alias value_type = T;
+    ///
+    alias reference = ref T;
```
ZombineDev: Unfortunately this doesn't work (`ref` is not a type constructor, so it's ignored here). Perhaps it's better to remove it, than to keep the wrong definition.

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,158 @@
...
+    ///
+    alias const_reference = ref const(T);
```
ZombineDev: ditto

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,158 @@
...
+    ///
+    reference back() @safe @nogc                                    { static if (N > 0) { return as_array()[N-1]; } else { return as_array()[][0]; } }
+    ///
+    const_reference back() const @safe @nogc                        { static if (N > 0) { return as_array()[N-1]; } else { return as_array()[][0]; } }
+
+    ///
+    void fill()(auto ref const(T) value) @safe @nogc                { this[] = value; }
+
+    ///
+    size_type opDollar(size_t pos)() const nothrow @safe @nogc      { static assert(pos == 0, "std::array is one-dimensional"); return N; }
```
ZombineDev: Can be simplified to:
```D
alias opDollar = length;
```

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,158 @@
...
+    size_type size() const nothrow @safe @nogc                      { return N; }
+    ///
+    alias length = size;
+    ///
+    size_type max_size() const nothrow @safe @nogc                  { return N; }
+    ///
+    bool empty() const nothrow @safe @nogc                          { return N == 0; }
+
+    ///
+    reference front() @safe @nogc                                   { static if (N > 0) { return as_array()[0]; } else { return as_array()[][0]; } }
```
ZombineDev: Replace `reference` with `ref`.

test/stdcpp/src/array.d
```diff
@@ -0,0 +1,57 @@
+module test.stdcpp.array;
+
+import core.stdcpp.array;
+
+unittest
+{
+    array!(int, 5) arr;
+    arr[] = [1, 2, 3, 4, 5];
+
```
ZombineDev: Add `arr.front++; assert(arr.front == 2);` to verify that return by `ref` works as expected.

TurkeyMan:
Fixed the things.

TurkeyMan:
Can someone who knows the platform check out what that link error in the auto-tester is about?
I can't reproduce that locally, and circleci is working (which is linux/gcc I think?).
Is it mangling?

TurkeyMan:
FWIW, if anyone feels charitable, I could really use some help kicking this across the line; just making sure it's compatible with all targets.
I don't have 20-30 minutes to wait for the auto-tester to see if I fixed it each time I try something :/

TurkeyMan:
Oh no... I can't give a tuple to extern(C++)... which means I can't drive the namespace by version >_<
text mixin doesn't work either. I need to reinstate `extern(C++, "std::__1");`

jacob-carlborg:
> I think the string literal should accept expressions that evaluate to strings...?

How do you determine if it's a namespace using the old/identifier syntax or the new string syntax but where the string is a variable?

TurkeyMan:
```d
enum NS = 10;
struct NS {}
```
Multiple defined symbol is a normal error when a declaration is already defined.
Either is a string expression, or it's not...?

TurkeyMan:
True. Well, not that ;)
Is kinda tricky to resolve... 
If it worked under the mixin, that would do it. Why does it fail by mixin?

TurkeyMan:
@jacob-carlborg Okay, I have it; just the same way C++ (and others) uses parens to disambiguate, we can do that here.
```d
extern(C++, (StringExp)): // <- not ambiguous
```
If/when the named scope is lifted into its own feature (as many have called), it's possible to relax the parens.
That's the best I've got!

jacob-carlborg:
Looks a bit ugly but I guess it will work.

TurkeyMan:
It's fine. It's like:
```c++
void (name)(int, float); // <- parens around `name` ;)
```
or
```c++
int (&array)[3]; // ref to an array
```
etc...

TurkeyMan:
@Geod24 Please present a superior option...

Geod24:
@TurkeyMan : The fact that a new feature is already insufficient for its first usage should be a very clear indicator that it wasn't properly thought through.

You've been pretty vocal about the slowness of the process, and I completely understand the frustration (been there), but OTOH, it is has been pretty efficient at preventing those rushed up design decisions.

Back to the subject, I don't think using parenthesis to disambiguate a syntax is remotely sensible.

At this point I suggest to just duplicate the definitions, which can be done solely in this PR, without depending on another compiler change. This allows to unblock this PR and will give us more time to gather experience about how to bind C++.


Note that the problem looks similar to the one we'd have if we tried to provide different attributes for a class' member functions based on a version. IOW trying to write a class where `version (My_Module_Nogc)` makes all functions in the module `@nogc` will have the same issue.

TurkeyMan:
I agree it wasn't finished baking. I think every person discussing the PR was extremely surprised when Walter merged it.

That said, I certainly don't think it's preferable that we had gone nowhere in the meantime, I'd just be frustrated and not working on the next hurdle. There's actually velocity on this, which is immeasurably better than half a decade of stagnation.

That said, I don't think it was ever intended that it wouldn't accept an Expression... at least not in my head. It had to support an expression, that was a fundamental part of the goal.

> This allows to unblock this PR and will give us more time to gather experience about how to bind C++.

Well I think accepting an Expression should have been part of the original patch... so I'm fairly motivated to address that as priority. I'm trying to write that patch...

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,149 @@
...
+
+// hacks to support DMD on Win32
+version (CppRuntime_Microsoft)
+{
+    version = CppRuntime_Windows; // use the MS runtime ABI for win32
+}
+else version (CppRuntime_DigitalMars)
+{
+    version = CppRuntime_Windows; // use the MS runtime ABI for win32
+    pragma(msg, "std::array not supported by DMC");
```
Geod24: Shouldn't it be a `static assert(false, "...")` ?

TurkeyMan: Then autotesters fail...

Geod24: That means it's running on Win32 even though it should not. Can you just disable the test there ?

Geod24: You got them running on Windows in the first place, didn't you ?

TurkeyMan: I don't use OMF. I wish it would die already!

TurkeyMan: DMD doesn't have `<array>`

posix.mak
```diff
@@ -378,7 +378,7 @@ style_lint:
 auto-tester-build: target checkwhitespace
 
 .PHONY : auto-tester-test
-auto-tester-test: unittest benchmark-compile-only
+auto-tester-test: unittest test_all benchmark-compile-only
```
rainers: Tests fail because there is no `test_all` target in posix.mak. The "additional tests" are run as dependency of `unittest`.

src/core/stdcpp/xutility.d
```diff
@@ -0,0 +1,60 @@
...
+    enum __CXXLIB__ = __traits(getTargetInfo, "cppRuntimeLibrary");
+
+package:
+    // these are all [[noreturn]]
+    void _Xbad() nothrow @trusted @nogc;
+    void _Xinvalid_argument(const(char)* message) nothrow @trusted @nogc;
+    void _Xlength_error(const(char)* message) nothrow @trusted @nogc;
+    void _Xout_of_range(const(char)* message) nothrow @trusted @nogc;
+    void _Xoverflow_error(const(char)* message) nothrow @trusted @nogc;
+    void _Xruntime_error(const(char)* message) nothrow @trusted @nogc;
```
Geod24: I don't think they should be `@trusted`, because we cannot audit them. And it's likely that passing a non-`\0` terminated string will create errors, so they don't expose a `@safe` interface.

src/core/stdcpp/xutility.d
```diff
@@ -0,0 +1,60 @@
...
+                pragma(linkerDirective, "/FAILIFMISMATCH:RuntimeLibrary=MD_DynamicRelease");
+        }
+    }
+
+    // convenient alias for the C++ std library name
+    enum __CXXLIB__ = __traits(getTargetInfo, "cppRuntimeLibrary");
+
+package:
+    // these are all [[noreturn]]
+    void _Xbad() nothrow @trusted @nogc;
```
Geod24: This one can be marked as `@safe` tho

test/stdcpp/src/array.cpp
```diff
@@ -0,0 +1,17 @@
...
+    int r = 0;
+    for (size_t i = 0; i < arr.size(); ++i)
+        r += arr[i];
+    return r;
+}
+
+int sumOfElements_val(std::array<int, 5> arr)
+{
+    return sumOfElements_ref(arr) + fromC_ref(arr) + fromC_val(arr);
+}
```
Geod24: I'd also like to see a function which returns an `std::array` by value and by `ref`.

test/stdcpp/src/array.d
```diff
@@ -0,0 +1,59 @@
+module test.stdcpp.array;
```
Geod24: Module name does not match file path, something we should avoid.

src/core/stdcpp/array.d
Geod24: Okay, that one makes sense

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,149 @@
...
+
+            ///
+            ref inout(T) at(size_type i) inout nothrow @trusted @nogc   { static if (N > 0) { if (N <= i) _Xran(); return _Elems.ptr[i]; } else { _Xran(); return _Elems[0]; } }
+        }
+        else
+        {
+            import core.exception : RangeError;
+
+            ///
+            ref inout(T) at(size_type i) inout @trusted             { if (i >= N) throw new RangeError("Index out of range"); return _Elems[i]; }
```
Geod24: Since this does an array access, is that check necessary ?
If you remove it, it's also more consistent with the `-boundscheck` switch.

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,149 @@
...
+    ///
+    bool empty() const nothrow @safe @nogc                          { return N == 0; }
+
+    ///
+    ref inout(T) front() inout nothrow @safe @nogc                  { static if (N > 0) { return this[0]; } else { return as_array()[][0]; /* HACK: force OOB */ } }
+    ///
+    ref inout(T) back() inout nothrow @safe @nogc                   { static if (N > 0) { return this[N-1]; } else { return as_array()[][0]; /* HACK: force OOB */ } }
+
+    ///
+    void fill()(auto ref const(T) value) @safe @nogc                { this[] = value; }
```
Geod24: From what I can see (if you follow my suggestion about `RangeError` you can just slap `nothrow @nogc @safe:` at the top of this `struct`.

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,149 @@
...
+
+    version (CppRuntime_Windows)
+    {
+        ///
+        inout(T)* data() inout nothrow @nogc                        { return &_Elems[0]; }
```
Geod24: This can be `@safe`, can't it ?

src/core/stdcpp/array.d
TurkeyMan: Yeah, I wasn't sure if this operation should match C++ error semantics, or D error semantics.
I've been leaning towards D semantics. C++ error semantics as called from D code isn't actually useful. I'll change all of these style functions to strictly use D bounds checking.

TurkeyMan:
Thanks for the review, I'll make those corrections.

src/core/stdcpp/array.d
TurkeyMan: Actually, thinking on this just now; this function is neither `@safe` nor `@nogc`; T's assignment doesn't promise those properties...
I guess I should remove them here and allow it to infer from the assignments.

src/core/stdcpp/array.d
Geod24: That's not really relevant. The function itself is `@safe`.

TurkeyMan:
Rebased, and comments addressed.

TurkeyMan:
Great, now it doesn't work... and with a gloriously helpful error message!

rainers:
auto-tester doesn't like your Makefile:
```
gmake[2]: *** No rule to make target 'generated/freebsd/release/32/array.done', needed by 'all'.  Stop.
```
Maybe, it is because `$(ROOT)/%: $(SRC)/%.d $(SRC)/%.cpp` doesn't match the name of the D file array_test.d.

TurkeyMan:
Oh yeah! I renamed a file so the automatic module name didn't conflict with the import.

src/core/stdcpp/array.d
```diff
@@ -0,0 +1,145 @@
...
+            mixin MSVCLinkDirectives!false;
+        }
+
+    private:
+        T[N ? N : 1] _Elems;
+    }
+    else version (CppRuntime_Gcc)
+    {
+        ///
+        inout(T)* data() inout nothrow @safe @nogc                  { static if (N > 0) { return &_M_elems[0]; } else { return null; } }
```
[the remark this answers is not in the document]
TurkeyMan: The whole file is inline one-liners at the same indentation level.
I think @WalterBright addresses this issue in his lecture: https://youtu.be/lbp6vwdnE0k?t=3672

thewilsonator:
OSX 32-bit fails with 
```
src/core/stdcpp/array.d(34): Error: namespace `core.stdcpp.array.std.__1` is forward referenced when looking for `size_t`
src/core/stdcpp/array.d(34): Error: namespace `core.stdcpp.array.std.__1` is forward referenced when looking for `size_t`
```
and win64 fails with
```
Error: module `array` is in file 'test\stdcpp\src\array.d' which cannot be read
import path[0] = src
```
Why, I have no idea.

TurkeyMan:
You merged the Clang fix in DMD.

It's Win32, not Win64, and it fails because Win32 builds with DMD/OMF against DMC, which doesn't support C++11...

TurkeyMan:
Oh sorry, there is a Win64 issue! I didn't see it creep in there, I think I introduced it when tweaking the tests... should be trivial.

thewilsonator:
Good!

thewilsonator:
I guess the OSX32 will be fixed by dmd#8921

TurkeyMan:
I mean, it's cool how the DMD build machines take an hour to build ;)

TurkeyMan:
> I guess the OSX32 will be fixed by dmd#8921

Yes.

thewilsonator:
You're telling me? I've merged what, two dozen DMD prs over last 4 days. I know all about that.

TurkeyMan:
Yay, FreeBSD is green!

Geod24:
But not darwin:
```
cc -m64 -fPIC -Wall -g -std=c++11 -c -o ./generated/osx/debug/64/array_cpp.o src/array.cpp
src/array.cpp:1:10: fatal error: 'array' file not found
#include <array>
         ^
1 error generated.
```

Are you using the C compiler to compile C++ by any chance ?

test/stdcpp/Makefile
```diff
@@ -0,0 +1,19 @@
...
+all: $(addprefix $(ROOT)/,$(addsuffix .done,$(TESTS)))
+
+$(ROOT)/%.done : $(ROOT)/%
+	@echo Testing $*
+	$(QUIET)$(TIMELIMIT)$(ROOT)/$* $(RUN_ARGS)
+	@touch $@
+
+$(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
+	mkdir -p $(dir $@)
+	$(QUIET)$(CC) $(CFLAGS) -std=c++11 -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
```
thewilsonator: It would seem so.

thewilsonator: TBH I'm surprised it didn't barf at the `-std=c++11`, though it is still clang hiding underneath anyway.

TurkeyMan:
Oh FFS, now Apple is just trolling me!
This is clearly the build machine's fault. It probably hasn't been updated for 10 years.

thewilsonator:
Thats odd, my `<array>` is in `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/array`

TurkeyMan:
It must be a build machine configuration fail... :/

jacob-carlborg:
I recommend adding `-v` to print out the include paths the compiler is looking in. Using Xcode 10, I got this:

```
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1
 /usr/local/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/10.0.0/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include
 /usr/include
 /System/Library/Frameworks (framework directory)
 /Library/Frameworks (framework directory)
```

jacob-carlborg:
> Thats odd, my <array> is in /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/array

Are you expecting the to be inside `/usr/include`? By default, unless you install the Xcode command line tools, all include files and libraries will be located inside the Xcode app bundle, i.e. `/Applications/Xcode.app`. This is for Xcode to be self contained and better match App Store apps.

thewilsonator:
No, I'm expecting it to exist, which it does for me. I'm wondering why it appears not to on the test machine.

jacob-carlborg:
I think the first step is to know where the compiler is looking.

thewilsonator:
True dat.

TurkeyMan:
I've never had to tell a compiler where to look for the standard library... :/

TurkeyMan:
https://stackoverflow.com/questions/12516830/mac-osx-10-7-4-xcode-4-4-1-no-array-header-file
Gotta supply `-stdlib=libc++` in the makefile for OSX builds... how am I supposed to supply platform-specific configuration in the makefiles?

Can someone with a mac try this and submit a PR please?

thewilsonator:
Yay for Stackoverflow-driven-development!
```make
ifeq ($(OS),Windows_NT)
else
    OS := $(shell uname)
    ifeq ($(OS),Darwin)
        CXX_FLAGS += -stdlib=libc++
    endif
endif
```

TurkeyMan:
`CXXFLAGS` doesn't seem to be defined anywhere... and it should definite include `CFLAGS` along with other C++ options... you have a mac right? Can you try and tweak the makefile appropriately? I will blindly attempt changed and push them, and wait an hour for a darwin CI build repeatedly until I randomly stumble upon the proper incantation...

test/stdcpp/Makefile
```diff
@@ -0,0 +1,19 @@
...
+all: $(addprefix $(ROOT)/,$(addsuffix .done,$(TESTS)))
+
+$(ROOT)/%.done : $(ROOT)/%
+	@echo Testing $*
+	$(QUIET)$(TIMELIMIT)$(ROOT)/$* $(RUN_ARGS)
+	@touch $@
+
+$(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
+	mkdir -p $(dir $@)
+	$(QUIET)$(CXX) $(CXXFLAGS) -std=c++11 -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
```
thewilsonator: Sorry as in 
```make
OS := $(shell uname)
CXXFLAGS :=  -std=c++11
ifeq ($(OS),Darwin)
    CXX_FLAGS += -stdlib=libc++
endif
...
$(ROOT)/%: $(SRC)/%_test.d $(SRC)/%.cpp
 	mkdir -p $(dir $@)
        $(QUIET)$(CXX) $(CXXFLAGS) -c -o $(ROOT)/$*_cpp.o $(SRC)/$*.cpp
```

TurkeyMan: I think `CXXFLAGS` should probably be initially defined in the root makefile. It seems weird to put it here in a leaf...

TurkeyMan: it also needs to aggregate `CFLAGS`

TurkeyMan: Other makefiles use `ifeq ($(OS),osx)` instead of `Darwin`... I can't test this. It's fiddly to get right.

TurkeyMan:
Okay, trying before bed.
The CI infrastructure is seriously under-powered. It's just painfully unproductive to spend an hour finding out if anything you try works.
I have never had a PR not require days of my time to satisfy the build machines... that's just madness.
I only have about 10,000 days, and if I accumulate 100 days where I spend the evening trying to appease the build machine (probably ~20-30 PR's), that's 1% of my life time that I'm **not** satisfied to lose in that way.

thewilsonator:
Looks like its passed on OSX :)

TurkeyMan:
OMFG, Apple doesn't even have a 32bit platform! That test is just wasting all of our time!

TurkeyMan:
I'm going to bed... I'll waste another day of my life tomorrow!

jacob-carlborg:
> OMFG, Apple doesn't even have a 32bit platform! That test is just wasting all of our time!

I've asked if we can remove the 32bit support. Nothing yet 😞.

I've created a Docker image for cross-compiling to macOS using LDC [1]. You cannot run the code but you can compile it. Not sure if that would help you in this case.

[1] https://github.com/jacob-carlborg/docker-ldc-darwin

jacob-carlborg:
It's also possible to run macOS in a virtual machine.

thewilsonator:
May I tactfully suggest you just say bugger it to OSX32bit since no-one uses it anyway?

```make
ifneq ($(OS)$(MODEL),osx32)
#do the test
endif
```

jacob-carlborg:
> May I tactfully suggest you just say bugger it to OSX32bit since no-one uses it anyway?

Fine by me.

jacob-carlborg:
It's also possible to abuse CircleCI. You can run a test with SSH enable, which will give you access to the machine if the test failed, for one hour, IIRC.

andralex:
@TurkeyMan 

>I'm going to bed... I'll waste another day of my life tomorrow!

"I'm going to bed... I'll spend one more day spearheading the important effort of C++ stdlib integration tomorrow!"

Fixed that for ya.

andralex:
>I've asked if we can remove the 32bit support.

I'm in favor. The only recollection I have is that platform helped us find a couple of portability bugs in the past. @WalterBright ?

thewilsonator:
Test is no longer run on OSX32, if clean is run `rm -rf $GENERATED` may fail if its expecting something to have been generated, but that is also easy fixed by the same approach.

ibuclaw:
Generally speaking it shouldn't be that difficult to maintain old ports. Solaris 10 and OSX 10.6 will be baselines of gcc 9 for those respective platforms.

There should be better way to manage transitions though. For instance, only supporting one version of FreeBSD is a pain, not having a clear `-std=c++XX` old platform support should never have a reason for it to stop working.

thewilsonator:
I agree about the `-std=c++XX`, but this just disables a test that won't work.

thewilsonator:
@TurkeyMan please squash this after I'm done with it.

TurkeyMan:
> @TurkeyMan
> 
> > I'm going to bed... I'll waste another day of my life tomorrow!
> 
> "I'm going to bed... I'll spend one more day spearheading the important effort of C++ stdlib integration tomorrow!"
> 
> Fixed that for ya.

@andralex Thanks man!
Might I suggest taking a serious interest in the productivity issues associated with D contribution from a foundation level? Consider employing a devops professional, and decomissioning/upgrading any build machine that takes >10 minuets. This whole process needs a serious streamlining effort.

andralex:
Considering we are. No funds yet.

jacob-carlborg:
I'm currently working as a DevOps, professionally.

TurkeyMan:
Woah! That was sudden... I would have liked to squash Nick's stream of consciousness ;)

TurkeyMan:
But thanks! That was a grueling marathon!
Now the real fun starts; string/vector.

wilzbach:
Fwiw Buildkite is intended to replace the auto-tester, but we currently don't have Windows nor OSX agents connected to it :/

TurkeyMan:
Buildkite... that doesn't even let you see the state of the builds!

thewilsonator:
See slack.

jacob-carlborg:
> Fwiw Buildkite is intended to replace the auto-tester, but we currently don't have Windows nor OSX agents connected to it :/

What about the existing machines used  for the auto-tester?

wilzbach:
We don't have access to them.

jacob-carlborg:
> We don't have access to them.

If we know how owns them we could ask.

wilzbach:
No because we can't install the buildkite agent on them because we don't have access. That was just a link to a previous attempt at improving the hardware setup.

PR #2289, opened 2018-08-29
Fix Issue 19204 - hashOf doesn't accept SIMD vectors

n8sh:


n8sh:
In trying to get CTFE to work ran into https://issues.dlang.org/show_bug.cgi?id=19223.

n8sh:
This needs to be merged before https://github.com/dlang/dmd/pull/8676 can be merged, or else any program that has a struct that has a vector field and for any reason needs to generate `__xtoHash` will fail to compile.

n8sh:
CTFE now works for float vectors.

test/hash/src/test_hash.d
```diff
@@ -87,6 +88,35 @@ void issue19005() @nogc nothrow pure @safe
...
+            assert(ctfeHash == rtHash);
+        }}
+        static if (is(simd.void16)) // non __traits(isArithmetic)
+        {{
+            auto h = hashOf(simd.void16.init);
+        }}
+        static if (is(simd.float4)) // __traits(isArithmetic) and __traits(isFloating)
+        {{
+            enum simd.float4 val = [1.1f, 2.2f, 3.3f, 4.4f];
+            enum ctfeHash = hashOf!(simd.float4)(val);
```
ZombineDev: Why do you need to explicitly instantiate `hashOf` here?

src/core/internal/hash.d
```diff
@@ -423,6 +423,31 @@ size_t hashOf(T)(scope const T val, size_t seed = 0) if (!is(T == enum) && __tra
...
+            size_t result = seed;
+            alias E = typeof(val[0]);
+            foreach (i; 0 .. T.sizeof / E.sizeof)
+                result = hashOf(val[i], result);
+            return result;
+        }
+        size_t result = seed;
+        foreach (x; val.array)
+            result = hashOf(x, result);
+        return result;
```
ZombineDev: Why not keep only the CTFE version?

n8sh: It seemed to produce better code with DMD. It looks like LDC2 compiles them both the same though.

test/hash/src/test_hash.d
n8sh: Didn't need to.

n8sh: Actually, I did need to. It was long enough ago that I forgot. Without the explicit instantiation the type is inferred as a floating point array.

n8sh:
The explicit instantiation in the CTFE test was needed because at compile time the type was inferred as `float[4]` rather than `__vector(float[4])`. To work around this I'm making the SIMD hash behave the same as for a static array, although for floating-point types this is currently slightly slower.

PR #2294, opened 2018-09-05
Decrease template bloat in core.internal.hash.hashOf

n8sh:
This PR causes fewer distinct instantiations of `hashOf` with function template parameter inference.

* Integral types up to `size_t`'s width are hashed as `size_t`. This changes the hash code of signed integral types narrower than `size_t`.
* Non-shared pointers are all hashed as `void*`. Shared pointers aren't all hashed as `shared void*` because in combination with the non-shared case this caused a function template parameter inference problem with `typeof(null)`.

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -340,16 +340,16 @@ if (!is(T == enum) && !is(T : typeof(null)) && is(T S: S[]) && !__traits(isStati
 }
 
 //arithmetic type hash
-@trusted @nogc nothrow pure
-size_t hashOf(T)(scope const T val) if (!is(T == enum) && __traits(isArithmetic, T)
+@safe @nogc nothrow pure
+size_t hashOf(T = size_t)(scope const size_t val) if (!is(T == enum) && __traits(isArithmetic, T)
     && __traits(isIntegral, T) && T.sizeof <= size_t.sizeof)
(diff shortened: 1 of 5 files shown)


n8sh:
Tests were initially failing in 32 bit mode due to `hashOf(42L)` using `hashOf!uint` instead of `hashOf!long` when `42L` is a compile-time constant.

n8sh:
Unfortunately this can cause `toHash` to be ignored for types with `alias this`. With this PR the following fails:
```d
    static struct S
    {
        int a;
        size_t toHash() const @nogc nothrow pure @safe { return 33; }
        alias a this;
    }
    static assert(hashOf(S.init) == 33);
```

Because I don't believe this problem can be solved, I am closing this PR.

jacob-carlborg:
Can’t you explicitly check for `toHash`?

n8sh:
I can't because this trick relies on implicit conversion to `size_t`. By that point it's too late: it's already `size_t`.

jacob-carlborg:
If you only one function and skip all overloads and handle everything inside the function?

n8sh:
Could do that, but it wouldn't do anything to decrease/consolidate template instantiations.

jacob-carlborg:
Ok.

PR #2295, opened 2018-09-08
core.internal.hash prefer FNV to MurmurHash3 for short sequences of statically known length

n8sh:
MurmurHash3 is optimized for long inputs. For short inputs a byte-oriented hash function can be faster. In my informal benchmarking using `ldc2 -O2 -release` FNV is faster than MurmurHash3 for inputs of around 10 bytes and fewer which is [similar to what others have reported](https://aras-p.info/blog/2016/08/09/More-Hash-Function-Tests/#conclusions).

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -257,7 +257,7 @@ if (!is(T == enum) && __traits(isStaticArray, T) && canBitwiseHash!T)
     //else static if (T.length == 1)
     //    return hashOf(val[0], seed);
     //else
-    //    /+ hash like a dynamic array +/
+    //    return bytesHashWithExactSizeAndAlignment!T(toUbyte(val), seed);
     //
     // ... but that's inefficient when using a runtime TypeInfo (introduces a branch)
     // and PR #2243 wants typeid(T).getHash(&val) to produce the same result as
@@ -412,7 +412,7 @@ size_t hashOf(T)(scope const T val, size_t seed = 0) if (!is(T == enum) && __tra
         else static if (T.mant_dig == double.mant_dig && T.sizeof == ulong.sizeof)
             return hashOf(*cast(const ulong*) &data, seed);
         else
-            return bytesHashAlignedBy!T(toUbyte(data), seed);
+            return bytesHashWithExactSizeAndAlignment!T(toUbyte(data), seed);
     }
     else
     {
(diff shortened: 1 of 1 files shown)


n8sh:
>Add a branch inside `TypeInfo_Struct` that dynamically determines which hash function to use.

Going with this for now.

src/core/internal/hash.d
```diff
@@ -1010,6 +1010,41 @@ private template bytesHashAlignedBy(AlignType)
...
+        alias bytesHashWithExactSizeAndAlignment = bytesHashAlignedBy!SizeAndAlignType;
+}
+
+// Size at & below which to prefer `smallBytesHash` over `bytesHash`.
+// Internal. Not to be used other than by druntime.
+enum smallBytesHashThreshold = 10;
+
+// Internal. Not to be used other than by druntime.
+@nogc nothrow pure @trusted
+size_t smallBytesHash()(scope const(ubyte)[] bytes, size_t seed)
```
jacob-carlborg: Does this and the above symbols need to be public?

n8sh: I would prefer not, but I needed to make it public to call it from `object.TypeInfo_struct`.

src/object.d
```diff
@@ -1947,7 +1947,12 @@ class TypeInfo_Struct : TypeInfo
         }
         else
         {
-            return hashOf(p[0 .. initializer().length]);
+            static import core.internal.hash;
+            const length = initializer().length;
+            if (length <= core.internal.hash.smallBytesHashThreshold)
+                return core.internal.hash.smallBytesHash((cast(const ubyte*)p)[0 .. length], 0);
+            else
+                return hashOf(p[0 .. initializer().length]);
```
jacob-carlborg: Can't `hashOf` handle this internally? They you don't need to expose two new symbols.

n8sh: It could. Updated the PR to do that.

n8sh: Oh. I thought that's what you meant, expose 1 symbol instead of 2.

jacob-carlborg: No, previously `hashOf` was called. Why not still calling that and handle everything inside that function?

n8sh: If I put that logic in `hashOf(ubyte[])` it would apply to more than just structs. I am cautious about that.

n8sh:
Changed this so it no longer applies to `hashOf(struct)` but does apply when performing a bitwise hash of a field of a struct. This means that `TypeInfo_Struct.getHash` doesn't need to know about the change and there aren't any new public symbols.

thewilsonator:
@n8sh is this good to go?

n8sh:
@thewilsonator I believe so.

n8sh:
Updated a comment.

PR #2301, opened 2018-09-18
Flags for send and recv on socket fd.

ivanroubicek:


CyberShadow:
What of the ones in core.sys.posix.sys.socket?

ivanroubicek:
I did not notice that one :) I've moved it to a socket.d and squashed the commits.

src/core/sys/posix/sys/socket.d
```diff
@@ -594,14 +594,24 @@ version( CRuntime_Glibc )
 
     enum : uint
```
ZombineDev: Pedantically speaking, those values are operating system-specific, as well as libc-specifc. In this case they're the correct for linux/glibc, but not for hurd/glibc. Can you add `version (linux)` before the enum definition?

(I know that you didn't write the original code, but we try to improve things incrementally whenever the opportunity arises.)

src/core/sys/posix/sys/socket.d
```diff
@@ -594,14 +594,24 @@ version( CRuntime_Glibc )
...
-        MSG_NOSIGNAL    = 0x4000
+        MSG_FIN         = 0x200,
+        MSG_SYN         = 0x400,
+        MSG_CONFIRM     = 0x800,
+        MSG_RST         = 0x1000,
+        MSG_ERRQUEUE    = 0x2000,
+        MSG_NOSIGNAL    = 0x4000,
+        MSG_MORE        = 0x8000,
+        MSG_WAITFORONE  = 0x10000,
+        MSG_FASTOPEN    = 0x20000000
```
ZombineDev: According to https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/bits/socket.h;h=c3fbb2110296273c1275f9a0c37ae753428dff7c;hb=HEAD, the complete definition would be:
```D
enum : uint
{
    MSG_OOB          = 0x01,
    MSG_PEEK         = 0x02,
    MSG_DONTROUTE    = 0x04,
    MSG_TRYHARD      = MSG_DONTROUTE,
    MSG_CTRUNC       = 0x08,
    MSG_PROXY        = 0x10,
    MSG_TRUNC        = 0x20,
    MSG_DONTWAIT     = 0x40,
    MSG_EOR          = 0x80
    MSG_WAITALL      = 0x100
    MSG_FIN          = 0x200,
    MSG_SYN          = 0x400,
    MSG_CONFIRM      = 0x800,
    MSG_RST          = 0x1000,
    MSG_ERRQUEUE     = 0x2000,
    MSG_NOSIGNAL     = 0x4000,
    MSG_MORE         = 0x8000,
    MSG_WAITFORONE   = 0x10000,
    MSG_BATCH        = 0x40000,
    MSG_ZEROCOPY     = 0x4000000,
    MSG_FASTOPEN     = 0x20000000,
    MSG_CMSG_CLOEXEC = 0x40000000
}
```

src/core/sys/posix/sys/socket.d
jmdavis: If they're specific to Linux, they're supposed to go in core.sys.linux.sys.socket, not core.sys.posix.sys.socket. Nothing should go in the posix modules unless they're standard POSIX.

ZombineDev: AFAICS, the latest edition (2016) of the Posix standard defines only the following constants:
```
MSG_CTRUNC
MSG_DONTROUTE
MSG_EOR
MSG_OOB
MSG_NOSIGNAL
MSG_PEEK
MSG_TRUNC
MSG_WAITALL
```
The values for those constants are implementation defined.
You may argue that, the rest of the values that are not part of the standard should be outside of the `core.sys.posix.` headers, but I think splitting the enum definition would be confusing and will do more harm than good.

jmdavis: They're just constants. It's not a named enum, and the way that druntime's bindings are laid out, the Linux-specific stuff _must_ go under core.sys.linux and not core.sys.posix. That's how the bindings are organized. It's just plain wrong to put system-specific bindings in core.sys.posix. You do end up with version blocks in the modules in core.sys.posix, because stuff like the actual values of the constants change, but only stuff that's actually part of POSIX goes in core.sys.posix. That way, we don't have someone on Linux importing core.sys.posix and ending up with Linux-specific code (or someone on FreeBSD importing core.sys.posix and ending up with FreeBSD-specific code). The core.sys.linux.* modules publicly import their core.sys.posix counterparts so that someone on Linux can can just import the linux one and get everything if they don't care about portability, but the posix modules must not contain _anything_ that isn't POSIX.

jacob-carlborg: > unless CRuntime_Glibc always implies Linux.

It does:

https://github.com/dlang/dmd/blob/4c5bb35ffcfdcd29c9726c00f46c1f6a81c80a05/src/dmd/mars.d#L1212-L1217

ivanroubicek:
Ok, so then I move linux specific flags to src/core/sys/linux/sys/socket.d and we should be good to go.

ZombineDev:
@ivanroubicek correct.

src/core/sys/posix/sys/socket.d
ZombineDev: @jacob-carlborg the logic in dmd's code is the other way around: linux implies glibc. It's also not correct as there are other libc implementations for linux even for x86[_64] that dmd can support, notably musl libc.

jacob-carlborg: But `CRuntime_Glibc` is not set on any other platforms so what’s the difference.

ZombineDev: @jacob-carlborg obviously that's because we haven't suppoerted such platforms *yet* ;)

MoonlightSentinel:
@Geod24 auto-merge didn't work

thewilsonator:
Thats because there are still changes requested on the PR.

Geod24:
> Thats because there are still changes requested on the PR.

Thanks, I checked that it was addressed, but forgot to dismiss them,

thewilsonator:
Auto-merge toggled on

PR #2302, opened 2018-09-20
[stable] core.stdc.stdint: Fix some aliases for Windows/MSVC

kinke:
In VS 2017's `stdint.h`, `int32_t` is a typedef for `int`, not `long`; also verifiable on [MSDN](https://msdn.microsoft.com/en-us/library/323b6b3k.aspx) (incl. VS 2015). I haven't checked DMC for conformance.

This also fixes one part of [issue 19249](https://issues.dlang.org/show_bug.cgi?id=19249).

### DIFF
--- a/src/core/stdc/stdint.d
+++ b/src/core/stdc/stdint.d
@@ -147,74 +147,37 @@ else version (OSX)
 }
 else version (Posix)
 {
-    version (D_LP64)
-    {
-        alias int8_t   = byte;      ///
-        alias int16_t  = short;     ///
-        alias uint8_t  = ubyte;     ///
-        alias uint16_t = ushort;    ///
-        alias int32_t  = int;       ///
-        alias uint32_t = uint;      ///
-        alias int64_t  = cpp_long;  ///
-        alias uint64_t = cpp_ulong; ///
+    alias int8_t   = byte;   ///
+    alias int16_t  = short;  ///
+    alias uint8_t  = ubyte;  ///
+    alias uint16_t = ushort; ///
+    alias int32_t  = int;    ///
+    alias uint32_t = uint;   ///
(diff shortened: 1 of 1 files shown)


kinke:
This was changed to `long` with v2.081 and used to be `int` before as well.

src/core/stdc/stdint.d
```diff
@@ -37,32 +37,32 @@ static if (is(ucent))
 
 version (Win32)
 {
-    alias int8_t   = byte;      ///
-    alias int16_t  = short;     ///
-    alias uint8_t  = ubyte;     ///
-    alias uint16_t = ushort;    ///
-    alias int32_t  = cpp_long;  ///
-    alias uint32_t = cpp_ulong; ///
```
TurkeyMan: O_o ... wow, wild ;)

ZombineDev:
Well this fails on Win32...

ZombineDev:
I think the correct fix would be to split Win32 on DMC vs MSVC.

TurkeyMan:
I'm surprise it's not already that way..

kinke:
> I'm surprise it's not already that way..

Yeah, me too, but no wonder - I hate these crappy mangling tests using a hardcoded mangle:

`static assert(int32.mangleof        == "?int32@@YAXJ@Z");`

Proper tests consist in a .cpp and a .d file and really make sure they can be linked...

kinke:
I have 2 follow-up fixes for `int_fast8_t` on Win64 and `[u]int_fast{8,16,32}_t` on FreeBSD, which would currently break `compilable/teststdint.d` and require https://github.com/dlang/dmd/pull/8729, which in turn requires those druntime fixes to get green.
@ZombineDev: How do you guys usually handle that chicken-and-egg problem?

PR #2303, opened 2018-09-21
Refactor core.stdcpp to use extern(C++, string) and CppRuntime_* versions

Geod24:
Look at the list of commits for a full description of what is being done.
This does not pass currently, because of what I suspect to be a DMD bug, but will check this weekend.
@TurkeyMan you might be interested in it.
We really need to document this test-suite better.

### DIFF
--- a/src/core/stdcpp/exception.d
+++ b/src/core/stdcpp/exception.d
@@ -11,95 +11,88 @@
 
 module core.stdcpp.exception;
 
+extern (C++, "std"):
+
 version (CRuntime_DigitalMars)
 {
     import core.stdcpp.typeinfo;
 
-    extern (C++, std)
+    alias void function() unexpected_handler;
+    unexpected_handler set_unexpected(unexpected_handler f) nothrow;
+    void unexpected();
+
+    alias void function() terminate_handler;
+    terminate_handler set_terminate(terminate_handler f) nothrow;
+    void terminate();
+
+    bool uncaught_exception();
+
+    class exception
     {
-        alias void function() unexpected_handler;
-        unexpected_handler set_unexpected(unexpected_handler f) nothrow;
-        void unexpected();
+        this() nothrow { }
+        this(const exception) nothrow { }
+        //exception operator=(const exception) nothrow { return this; }
+        //virtual ~this() nothrow;
+        void dtor() { }
+        const(char)* what() const nothrow;
+    }
 
-        alias void function() terminate_handler;
-        terminate_handler set_terminate(terminate_handler f) nothrow;
-        void terminate();
-
-        bool uncaught_exception();
-
-        class exception
-        {
-            this() nothrow { }
-            this(const exception) nothrow { }
-            //exception operator=(const exception) nothrow { return this; }
-            //virtual ~this() nothrow;
-            void dtor() { }
-            const(char)* what() const nothrow;
-        }
-
-        class bad_exception : exception
-        {
-            this() nothrow { }
-            this(const bad_exception) nothrow { }
-            //bad_exception operator=(const bad_exception) nothrow { return this; }
-            //virtual ~this() nothrow;
-            override const(char)* what() const nothrow;
-        }
+    class bad_exception : exception
+    {
+        this() nothrow { }
+        this(const bad_exception) nothrow { }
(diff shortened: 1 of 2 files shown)


Geod24:
Note: DMD has [tests for catching C++ exceptions](https://github.com/dlang/dmd/blob/a6dad1d2e8ae99afcce5235297407567f96c2c26/test/runnable/cppa.d#L968-L1057).
Those tests check that a C++ exception can be caught, and that RAII works on the D side.

They do not check that a (C++) exception can be thrown from D and caught from C++ or that RAII works on the C++ side.
Currently one cannot `throw` types that aren't derived from `Throwable`. I'm not familiar with that part of the compiler, so I don't know if it makes more sense to fix the compiler to allow it, or ask people to call a C++ function to do the `throw`.

test/c++/Makefile
```diff
@@ -0,0 +1,24 @@
...
+	rm -rf $(GENERATED)
+
+# Generic testing rule
+$(ROOT)/%.done: $(ROOT)/%
+	@echo Testing $*
+	$(QUIET)$(TIMELIMIT)$(ROOT)/$* $(RUN_ARGS)
+	@touch $@
+
+$(ROOT)/exceptions:
+	$(QUIET)$(CXX) exceptions/exception.cpp -c -o exceptions/cpp_exception_test.o
```
TurkeyMan: How do you manipulate this makefile to build with MSVC?

Geod24: Ergh... I don't think this testsuite works on Windows at all.

TurkeyMan: The auto tester runs on most platforms that DMD targets... how do those tests run?

Geod24: https://auto-tester.puremagic.com/pull-history.ghtml?projectid=1&repoid=2&pullid=2303

The tests pass on Windows

TurkeyMan: yeah, i spotted that... I have no idea how; that's not even remotely a valid compile command line :/

Geod24: When I said "works on Windows", I meant "is run". That's why the test pass, because it just does not run this.

src/core/stdcpp/exception.d
```diff
@@ -11,95 +11,109 @@
...
-            this(const(char)* msg = "bad exception");
-            //virtual ~this();
-        }
+    class bad_exception : exception
+    {
+        this(const(char)* msg = "bad exception");
+        //virtual ~this();
+    }
+}
+else version (CppRuntime_Clang)
```
jacob-carlborg: Not sure if `CppRuntime_Clang` is the best name since Clang can use both libc++ and stdc++.

TurkeyMan: We could remove those and use `__traits(getTargetInfo, "something")`...

Geod24: @jacob-carlborg : Currently compiling on OSX == `CppRuntime_Clang` defined.
Also linking with libstdc++ is deprecated.

jacob-carlborg: I’m pretty sure it’s only deprecated on macOS and Clang works on other platforms as well.

TurkeyMan: Let's not bike-shed this... I can't move forwards on anything until C++ unit-tests work >_<

jacob-carlborg: Fine.

Geod24:
The tests are somehow blocked by https://github.com/dlang/dmd/pull/8743

test/c++/exceptions/dexception.d
```diff
@@ -0,0 +1,48 @@
...
+    try
+    {
+        func1();
+        assert(0, "Exception not thrown");
+    }
+    catch (exception e)
+    {
+        assert(strcmp(e.what(), "C++ sends his regards") == 0);
+    }
+    assert(0);
```
thewilsonator: This assert is unconditionally tripped, did you mean to return in the catch?

Geod24:
Okay I removed the addition and turned this into a simple refactor for the time being, so people at least won't get tripped by the missing definitions.

Geod24:
@n8sh : Not at the moment, no. Why ?

wilzbach:
Buildkite failure is unrelated.

PR #2304, opened 2018-09-21
Fix use of std.json in gcbench.conmsg

FeepingCreature:
Avoid issues with https://github.com/dlang/phobos/pull/6716 : preemptively make std.json const correct.

See also https://issues.dlang.org/show_bug.cgi?id=19256

### DIFF
--- a/benchmark/gcbench/conmsg.d
+++ b/benchmark/gcbench/conmsg.d
@@ -7,7 +7,7 @@
  */
 import std.algorithm, std.concurrency, std.conv, std.file, std.json, std.range;
 
-JSONValue buildVal(in dchar[] word)
+JSONValue buildVal(in dchar[] word) pure
 {
     JSONValue[string] res;
     res["word"] = word.to!string;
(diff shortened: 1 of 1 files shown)


rainers:
> Should have a changelog entry or referencing a bugzilla issue.

Maybe for the phobos change, but not for this change to the benchmarks.

jacob-carlborg:
Ah, didn’t look close enough.

benchmark/gcbench/conmsg.d
```diff
@@ -16,7 +16,7 @@ JSONValue buildVal(in dchar[] word)
     foreach (i; 0 .. pos.length)
         pos[i] = i;
     res["array"] = pos;
-    return JSONValue(res);
+    return cast(immutable) JSONValue(res);
```
jacob-carlborg: Can the function be made `pure` instead of having to cast to immutable?

FeepingCreature: I didn't know that would do that. Great! So done.

jacob-carlborg: If it's a `pure` function that compiler can infer that a value is unique and therefore implicitly convertible to `immutable`.

PR #2305, opened 2018-09-22
[refactor] posix.mak: Enforce whitespace before opening parenthesis

ibuclaw:
Style rule taken from phobos posix.mak.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -370,6 +370,9 @@ style_lint:
 	@echo "Check for trailing whitespace"
 	$(GREP) -nr '[[:blank:]]$$' $(MANIFEST) ; test $$? -eq 1
 
+	@echo "Enforce whitespace before opening parenthesis"
+	grep -nrE "\<(for|foreach|foreach_reverse|if|while|switch|catch)\(" $$(find src -name '*.d') ; test $$? -eq 1
+
 .PHONY : auto-tester-build
 auto-tester-build: target checkwhitespace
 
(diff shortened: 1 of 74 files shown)


posix.mak
```diff
@@ -370,6 +370,9 @@ style_lint:
 	@echo "Check for trailing whitespace"
 	$(GREP) -nr '[[:blank:]]$$' $(MANIFEST) ; test $$? -eq 1
 
+	@echo "Enforce whitespace before opening parenthesis"
+	grep -nrE "\<(for|foreach|foreach_reverse|if|while|switch|catch)\(" $$(find src -name '*.d') ; test $$? -eq 1
```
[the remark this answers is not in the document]
ibuclaw: Not since object.di.  They should all be auto-generated.

posix.mak
```diff
@@ -370,6 +370,9 @@ style_lint:
...
+
```
ZombineDev: Can you add a check for ~`version ( abc )`~ -> `version (abc)`

ibuclaw: I have it in a separate branch, do you really want it all squashed together?  This change alone is big enough as it is.

ibuclaw: There's loads more styling checks that would be nice to copy from phobos as well.  But I'm not going to spend too much time on that.

jacob-carlborg: > do you really want it all squashed together?

No.

PR #2308, opened 2018-09-23
[AArch64] [GDC] AArch64: Fix backtraces in Fibers

jpf91:
Currently in GDC, when throwing an Exception in the Fiber the backtrace generation crashes. GDB backtrace generation also tries to generate some more function frames as it does not find the stack bottom. `Using .cfi_undefined x30` tells the debug info that the value in the lr is unknown, which seems to be the nicest way to stop the unwinder. Setting `x30` to 0 is another option, however it still creates one invalid frame in gdb, so the `.cfi` variant is used here.

We need to introduce an extra function for this, as we can't do it in fiber_entryPoint directly and it cannot be in fiber_switchContext either.

The root problem is that fiber_switchContext effectively always generates a function return, i.e.:
* Set link register lr to destination
* Jump to destination

Which is correct most of the times, but not for the initial fiber setup: Here we actually want to call `fiber_entryPoint`, as we jump to the start of the function. Our current code effectively tells the debugger that `fiber_entryPoint` was called by `fiber_entryPoint`, as the lr register points to that in the initial call. Even generating a call wouldn't really suffice though, you still need to set lr to 0 or use the cfi undefined somewhere.

Link: https://github.com/D-Programming-GDC/GDC/pull/744

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -3678,7 +3678,11 @@ private
 
   // Look above the definition of 'class Fiber' for some information about the implementation of this routine
   version (AsmExternal)
-    extern (C) void fiber_switchContext( void** oldp, void* newp ) nothrow @nogc;
+  {
+      extern (C) void fiber_switchContext( void** oldp, void* newp ) nothrow @nogc;
+      version (AArch64)
+          extern (C) void fiber_trampoline() nothrow;
+  }
   else
     extern (C) void fiber_switchContext( void** oldp, void* newp ) nothrow @nogc
     {
@@ -4908,7 +4912,7 @@ private:
             // Only need to set return address (lr).  Everything else is fine
             // zero initialized.
             pstack -= size_t.sizeof * 11;    // skip past x19-x29
-            push(cast(size_t) &fiber_entryPoint);
+            push(cast(size_t) &fiber_trampoline); // see threadasm.S for docs
             pstack += size_t.sizeof;         // adjust sp (newp) above lr
         }
         else version (AsmARM_Posix)
--- a/src/core/threadasm.S
+++ b/src/core/threadasm.S
@@ -475,6 +475,7 @@ fiber_switchContext:
         .text
         .global CSYM(fiber_switchContext)
         .p2align  2
+        .type   fiber_switchContext, %function
 CSYM(fiber_switchContext):
         stp     d15, d14, [sp, #-20*8]!
         stp     d13, d12, [sp, #2*8]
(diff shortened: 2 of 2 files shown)


Burgos:
Should we also do the same thing with the ARM's version of `fiber_SwitchContext` to make code and behavior uniform across these architectures? The current code there sets lr to 0, instead of using cfi_undefined, which seems to be a common practice:

https://github.com/dlang/druntime/blob/b435b5491b914b5bc4434db6cf25b15529ae2904/src/core/threadasm.S#L437-L448

jpf91:
@ibuclaw yes, but that should be done in an extra pull request.
@Burgos I agree that we should change, however I don't want to change it without testing first and I don't have time to test the ARM port now, maybe next weekend. I propose we review the AArch64 only part (which is required for GDC and needs to be backported) and I'll file an ARM pull request later (it's less critical, as that code is already working, no need for an immediate backport here).

ibuclaw:
> it's less critical, as that code is already working, no need for an immediate backport here

And we vet that by having ARM servers in our CI.  We also have AArch64 servers, but they can't build druntime yet - hence this patch.

ibuclaw:
Interestingly, I've come across similar things for x86 (`.cfi_undefined rip`) and PPC (`.cfi_undefined lr`), so I wonder if your debugging woes are reproducible on other architectures - particularly x86.

nemanja-boric-sociomantic:
> Interestingly, I've come across similar things for x86 (.cfi_undefined rip) and PPC (.cfi_undefined lr), so I wonder if your debugging woes are reproducible on other architectures - particularly x86.

Interestingly I never saw this issue in fibers on x64, but I did see it in the D1 runtime's exception handling on Xenial where glibc was compiled without frame pointers and I fixed it with equivalent of setting lr to 0.

> Adding a comment as to why this is here would be good (e.g. describe the directive as stop unwinding), as this is fairly arcane.

> Maybe something like fiber_trampoline would be a better name though, and definitely a comment which an explanation for the why it exists.

I agree with both points.

>  @Burgos I agree that we should change, however I don't want to change it without testing first and I don't have time to test the ARM port now, maybe next weekend. I propose we review the AArch64 only part (which is required for GDC and needs to be backported) and I'll file an ARM pull request later (it's less critical, as that code is already working, no need for an immediate backport here).

Fine with me, thanks!

jpf91:
I've addressed the comments and pushed a new version.

In hindsight, the ARM approach of clearing lr in `fiber_switchContext` doesn't seem like a bad solution either (I didn't remember that ARM code when I first wrote this AArch64 solution). In the end, this trampoline function here allows us to remove the instruction in `fiber_switchContext` clearing lr, at the expensive of a slightly more inefficient startup of Fibers.

> Interestingly, I've come across similar things for x86 (.cfi_undefined rip) and PPC (.cfi_undefined lr), so I wonder if your debugging woes are reproducible on other architectures - particularly x86.

Well, it's easy to test this:
```d
import std.stdio, core.thread;
void foo() {
    writeln();
}
void main() { auto f = new Fiber(&foo); f.call(); }
```
X86:
```
(gdb) bt
#0  test.foo() () at test.d:5
#1  0x0000000000448c54 in core.thread.Fiber.run() ()
#2  0x0000000000448b33 in fiber_entryPoint ()
#3  0x0000000000000000 in ?? ()
Backtrace stopped: Cannot access memory at address 0x7ffff7ff7000
```

So it might make sense to use this trampoline for all supported architectures.

However ARM and AARCH64 even segfault when we use libbacktrace to print the call stack, so it's more of an issue there. (I guess this would also happen when unwinding to find the exception handler, however we always have a top-level catch handler, so unwinding stops early and we do not see the effect).

ibuclaw:
LGTM, anyone else?

src/core/thread.d
```diff
@@ -3677,8 +3677,12 @@ private
     }
 
   // Look above the definition of 'class Fiber' for some information about the implementation of this routine
-  version (AsmExternal)
-    extern (C) void fiber_switchContext( void** oldp, void* newp ) nothrow @nogc;
+  version(AsmExternal)
```
ibuclaw: `version (AsmExternal)`

jpf91: Thanks, fixed.

ibuclaw:
I'll hit merge on Monday. @klickverbot - any last words?

PR #2310, opened 2018-09-23
Added std::string

TurkeyMan:
This is still WIP, but the MS implementation is fully working and usable.

This is blocked on:

* https://issues.dlang.org/show_bug.cgi?id=19179 and 
* implementation of DIP1014

### DIFF
--- a/changelog/std_string.dd
+++ b/changelog/std_string.dd
@@ -1,0 +1,3 @@
+Added `core.stdcpp.string`.
+
+Added `core.stdcpp.string`, which links against C++ `std::string`
--- a/mak/COPY
+++ b/mak/COPY
@@ -71,6 +71,7 @@ COPY=\
 	$(IMPDIR)\core\stdcpp\array.d \
 	$(IMPDIR)\core\stdcpp\exception.d \
 	$(IMPDIR)\core\stdcpp\new_.d \
+	$(IMPDIR)\core\stdcpp\string.d \
 	$(IMPDIR)\core\stdcpp\string_view.d \
 	$(IMPDIR)\core\stdcpp\typeinfo.d \
 	$(IMPDIR)\core\stdcpp\type_traits.d \
--- a/src/core/stdcpp/allocator.d
+++ b/src/core/stdcpp/allocator.d
@@ -48,10 +48,10 @@ extern(D):
                 import core.stdcpp.xutility : _Xbad_alloc;
                 if (count == 0)
                     return null;
-                T* mem;
-                if ((size_t.max / T.sizeof < count) || (mem = __cpp_new(count * T.sizeof)) == 0)
+                void* mem;
+                if ((size_t.max / T.sizeof < count) || (mem = __cpp_new(count * T.sizeof)) is null)
                     _Xbad_alloc();
-                return mem;
+                return cast(T*)mem;
             }
             else
             {
@@ -67,7 +67,10 @@ extern(D):
                         return _Max_possible < _Count ? size_t.max : _Count * T.sizeof;
                     }
                 }
+
                 const size_t _Bytes = _Get_size_of_n!T(count);
+                if (_Bytes == 0)
+                    return null;
 
                 static if (!__cpp_aligned_new || _Align <= __STDCPP_DEFAULT_NEW_ALIGNMENT__)
                 {
@@ -76,12 +79,10 @@ extern(D):
                         if (_Bytes >= _Big_allocation_threshold)
                             return cast(T*)_Allocate_manually_vector_aligned(_Bytes);
                     }
-                    return _Bytes ? cast(T*)__cpp_new(_Bytes) : null;
+                    return cast(T*)__cpp_new(_Bytes);
                 }
                 else
                 {
-                    if (_Bytes == 0)
-                        return null;
                     size_t _Passed_align = _Align;
                     version (INTEL_ARCH)
                     {
--- a/src/core/stdcpp/allocator.d
+++ b/src/core/stdcpp/allocator.d
@@ -14,11 +14,6 @@ module core.stdcpp.allocator;
 import core.stdcpp.new_;
 import core.stdcpp.xutility : __cpp_sized_deallocation, __cpp_aligned_new;
 
-// some versions of VS require a `* const` pointer mangling hack
-// we need a way to supply the target VS version to the compile
-version (CppRuntime_Microsoft)
-    version = NeedsMangleHack;
-
 /**
  * Allocators are classes that define memory models to be used by some parts of
  * the C++ Standard Library, and most specifically, by STL containers.
@@ -30,35 +25,114 @@ struct allocator(T)
     static assert(!is(T == const), "The C++ Standard forbids containers of const elements because allocator!(const T) is ill-formed.");
     static assert(!is(T == immutable), "immutable is not representable in C++");
     static assert(!is(T == class), "Instantiation with `class` is not supported; D can't mangle the base (non-pointer) type of a class. Use `extern (C++, class) struct T { ... }` instead.");
+extern(D):
+
+    ///
+    this(U)(ref allocator!U) {}
 
     ///
     alias value_type = T;
 
+    ///
+    alias rebind(U) = allocator!U;
+
     version (CppRuntime_Microsoft)
     {
+        import core.stdcpp.xutility : _MSC_VER;
+
         ///
-        T* allocate(size_t count) @nogc;
+        T* allocate(size_t count) @nogc
+        {
+            static if (_MSC_VER <= 1800)
+            {
+                import core.stdcpp.xutility : _Xbad_alloc;
+                if (count == 0)
+                    return null;
+                T* mem;
+                if ((size_t.max / T.sizeof < count) || (mem = __cpp_new(count * T.sizeof)) == 0)
+                    _Xbad_alloc();
+                return mem;
+            }
+            else
+            {
+                enum _Align = _New_alignof!T;
+
+                static size_t _Get_size_of_n(T)(const size_t _Count)
+                {
+                    static if (T.sizeof == 1)
+                        return _Count;
+                    else
+                    {
+                        enum size_t _Max_possible = size_t.max / T.sizeof;
+                        return _Max_possible < _Count ? size_t.max : _Count * T.sizeof;
+                    }
+                }
+                const size_t _Bytes = _Get_size_of_n!T(count);
+
+                static if (!__cpp_aligned_new || _Align <= __STDCPP_DEFAULT_NEW_ALIGNMENT__)
+                {
+                    version (INTEL_ARCH)
+                    {
+                        if (_Bytes >= _Big_allocation_threshold)
+                            return cast(T*)_Allocate_manually_vector_aligned(_Bytes);
+                    }
+                    return _Bytes ? cast(T*)__cpp_new(_Bytes) : null;
+                }
+                else
+                {
+                    if (_Bytes == 0)
+                        return null;
+                    size_t _Passed_align = _Align;
+                    version (INTEL_ARCH)
+                    {
+                        if (_Bytes >= _Big_allocation_threshold)
+                            _Passed_align = _Align < _Big_allocation_alignment ? _Big_allocation_alignment : _Align;
+                    }
+                    return cast(T*)__cpp_new_aligned(_Bytes, cast(align_val_t)_Passed_align);
+                }
+            }
+        }
         ///
-        void deallocate(T* ptr, size_t count) @nogc;
+        void deallocate(T* ptr, size_t count) @nogc
+        {
+            static if (_MSC_VER <= 1800)
+            {
+                __cpp_delete(ptr);
+            }
+            else
+            {
+                // this is observed from VS2017
+                void* _Ptr = ptr;
+                size_t _Bytes = T.sizeof * count;
+
+                enum _Align = _New_alignof!T;
+                static if (!__cpp_aligned_new || _Align <= __STDCPP_DEFAULT_NEW_ALIGNMENT__)
+                {
+                    version (INTEL_ARCH)
+                    {
+                        if (_Bytes >= _Big_allocation_threshold)
+                            _Adjust_manually_vector_aligned(_Ptr, _Bytes);
+                    }
+                    static if (_MSC_VER <= 1900)
+                        __cpp_delete(ptr);
+                    else
+                        __cpp_delete_size(_Ptr, _Bytes);
+                }
+                else
+                {
+                    size_t _Passed_align = _Align;
+                    version (INTEL_ARCH)
+                    {
+                        if (_Bytes >= _Big_allocation_threshold)
+                            _Passed_align = _Align < _Big_allocation_alignment ? _Big_allocation_alignment : _Align;
+                    }
+                    __cpp_delete_size_aligned(_Ptr, _Bytes, cast(align_val_t)_Passed_align);
+                }
+            }
+        }
 
         ///
         enum size_t max_size = size_t.max / T.sizeof;
-
-        version (NeedsMangleHack)
-        {
-            // HACK: workaround to make `deallocate` link as a `T * const`
-            private extern (D) enum string constHack(string name) = (){
-                version (Win64)
-                    enum sub = "AAXPE";
-                else
-                    enum sub = "AEXPA";
-                foreach (i; 0 .. name.length - sub.length)
-                    if (name[i .. i + sub.length] == sub[])
-                        return name[0 .. i + 3] ~ 'Q' ~ name[i + 4 .. $];
-                assert(false, "substitution string not found!");
-            }();
-            pragma(linkerDirective, "/alternatename:" ~ deallocate.mangleof ~ "=" ~ constHack!(deallocate.mangleof));
-        }
     }
     else version (CppRuntime_Gcc)
     {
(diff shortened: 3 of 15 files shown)


src/core/stdcpp/allocator.d
```diff
@@ -0,0 +1,50 @@
...
+
+extern(C++, "std"):
+
+/**
+ * Allocators are classes that define memory models to be used by some parts of
+ * the C++ Standard Library, and most specifically, by STL containers.
+ */
+extern(C++, class) struct allocator(T)
+{
+    static assert(!is(T == const), "The C++ Standard forbids containers of const elements because allocator!(const T) is ill-formed.");
```
Geod24: Should also check for `immutable` no ?

TurkeyMan: Sure. What I really care about though, is working out how to run those tests in this pr to prove it works on all those build machines...

Geod24: I'm pretty sure it does not work on POSIX, because [template mangling is broken](https://github.com/dlang/dmd/pull/8455).
But having those tests running on Windows would be awesome.

TurkeyMan: We need to run it on every target, then I'd know that template mangling is broken and fix it...
I'm not aware of problems with template mangling. This needs to work everywhere. Can't know that unless ci executes the tests.

thewilsonator: `immutable` should already be caught by the mangling code, because there is no equivalent.

TurkeyMan: I don't actually think there's any reason to believe these C++ containers will exclusively interact with `extern(C++)`... once you interface C++ and these are floating around, they'll probably be used generally by D code too.

Geod24: That's right, however the additional check is a small price to pay here. A casual reader might not realize that `immutable` is disallowed by the compiler. Also, we have no guarantee this won't change in the future. Lastly, it will provide a better error message.

TurkeyMan: These are all great ideas, but it's all worthless unless we find a way to test it ;)

rainers: See https://github.com/TurkeyMan/druntime/pull/1. I suspect it will conflict with master due to other added tests, though.

TurkeyMan: @rainers Thanks! I can't see how the `.cpp` file gets built in there though, that's the trouble point, since it needs to compile for all targets, and that includes DMC and MSVC on windows...

rainers: > I can't see how the `.cpp` file gets built in there though, that's the trouble point, since it needs to compile for all targets, and that includes DMC and MSVC on windows...

The amended changes to the makefile now contain the cpp-compilation, but as string.d doesn't compile on anything but win64, the auto-tester stops before it schedules anything to the Windows machines. I'd suggest adding `version(CRuntime_Microsoft):` at the top for now.

When this works to some extend, adding makefiles for other targets to the test should be straight-forward.

mak/COPY
```diff
@@ -48,8 +48,10 @@ COPY=\
 	$(IMPDIR)\core\stdc\wchar_.d \
 	$(IMPDIR)\core\stdc\wctype.d \
 	\
+	$(IMPDIR)\core\stdcpp\string.d \
```
rainers: type_traits.d missing here and all the other makefile locations.

test/stdcpp/win64.mak
```diff
@@ -0,0 +1,13 @@
+# built from the druntime top-level folder
+# to be overwritten by caller
+DMD=dmd
+MODEL=64
+DRUNTIMELIB=druntime64.lib
+CC=cl
+
+test:
+	$(CC) -c /Fostring_cpp.obj test\stdcpp\src\string.cpp
```
rainers: `$(CC)` should better be in quotes as the default location contains spaces.

src/core/stdcpp/string.d
```diff
@@ -0,0 +1,428 @@
...
+        }
+
+        static void _Xran() @trusted @nogc { _Xout_of_range("invalid string position"); }
+        static void _Xlen() @trusted @nogc { _Xlength_error("string too long"); }
+
+        _String_alloc!(_String_base_types!(T, Alloc)) _Base;
+    }
+    else
+    {
+        static assert(false, "C++ runtime not supported");
```
thewilsonator: @TurkeyMan  you don't have to actually implement the `extern(C++)` functions, a declaration should suffice. In fact only functions that you want inlinable by DMD (that excludes platforms that are not of interest to you, if someone else wants that performance, let them do it) need implementations. LTO will solve perf problems on other compilers.
Things that are `extern(D)` e.g. constructors that assign `const(T)[]` need to be implemented in D but can be done later as they are a convenience. 

What _is_ required are the data members. Doesn't matter what the layout is if you never use them, anything that adds up to 12 (for 32bit0 or 24 (for 64 bit) bytes will do provided it is aligned to 4 or 8 bytes.

TurkeyMan: I thought that at first. But reality is, link errors errors are inevitable. I've come to the conclusion that only hard-functions can extern (allocation, etc). Template functions must mirror. And that's fine... They're pretty simple really.

thewilsonator: Ah, I forgot about template functions. 

> But reality is, link errors errors are inevitable

Ouch, I thought the whole purpose of `extern(C++)` was link compatibility.

> They're pretty simple really.

Good luck then ;)

TurkeyMan: Basically, user experience is terrible unless I write at least the functions you see here.
They match the C++ runtimes exactly... but they don't need to. I could consolidate them if we didn't care about having a perfect match.
It's really just the allocator functions that have to work right, and the data structures.

thewilsonator: Fair enough. Good luck!

TurkeyMan: Luck indeed! The hardest part is getting tests to work... And the brutal agonising level of patience required to move anything forward here ;)

src/core/stdcpp/string.d
```diff
@@ -328,6 +328,121 @@ extern(D):
 
         _String_alloc!(_String_base_types!(T, Alloc)) _Base;
     }
+    else version (CppRuntime_Gcc)
+    {
+        ///
+        this(DefaultConstruct)                                              { _M_p = _M_local_data(); _M_set_length(0); }
```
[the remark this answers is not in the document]
thewilsonator: If _you_ don't need to support it, I'd say defer adding the `CppRuntime_Gcc` implementation. Note that because there is no instantiations of the template in druntime, you can probably just `static assert(0, "CppRuntime_Gcc not yet supported");`

 If you do... you'll need to wait until DIP1014 is implemented.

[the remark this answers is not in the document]
TurkeyMan: I tickled my own needs months ago... I don't need to do all this work to make the tests work and hit all platforms ;)
I just want to put the lid on this whole thing... it's gone on way too long.

TurkeyMan:
Is mangling issues?
`./generated/linux/debug/64/string_cpp.o` should surely have those symbols...

src/core/stdcpp/string.d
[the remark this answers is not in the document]
andralex: This is a major problem. AFAIK nobody is working on DIP 1014.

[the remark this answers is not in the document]
TurkeyMan: Well, it's sitting on your review queue... is there something wrong with it? Are you planning on rejecting it?
It's literally the thing on your desk right now isn't it? (according to the DIP schedule)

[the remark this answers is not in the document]
thewilsonator: I assume he means implementation, but it is ambiguous.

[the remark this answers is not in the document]
TurkeyMan: DIP 1014 looks simple. I'm sure it'll appear promptly if it's approved.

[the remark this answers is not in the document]
andralex: I meant internal pointers in std::string are a major problem, and that nobody is working on DIP 1014.

thewilsonator:
@TurkeyMan it would if you link the C++ std lib to it ;)

TurkeyMan:
Those symbols are all templates... they should have been instantiated by `string.cpp`, and contained in the `.o` file? How can template instantiations live in the std lib?

src/core/stdcpp/string.d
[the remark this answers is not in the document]
TurkeyMan: Right, but DIP 1014 is resting with you and Walter is it not? You are the only one that can move that along.

thewilsonator:
You're missing a makefile for non-windows platforms so I guess the tester is just recursing into `test/stdcpp/src/` and building `.d` files, ignorant of the fact that  the `.cpp` need to be built as well.

TurkeyMan:
`test/stdcpp/Makefile` is in the patch. That triggers the build. You can see the `.cpp` compile in the build log.

thewilsonator:
Oh was looking at freeBSD.

That does seem odd, can you reduce it?

src/core/stdcpp/string.d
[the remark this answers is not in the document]
12345swordy: @TurkeyMan DIP 1014 have been recently approved

TurkeyMan:
@andralex This will want to make use of the accepted `opMove` DIP, is there a PR prepared for that? Work in progress?

andralex:
@TurkeyMan there is no work on this at the moment beyond the approval. cc @WalterBright

andralex:
@TurkeyMan and yes this PR was a major motivator for the approval.

TurkeyMan:
Do you know if anyone is working on it, or intends to?

andralex:
@TurkeyMan no, it's all very recent.

thewilsonator:
@TurkeyMan please rebase.

TurkeyMan:
If anyone that runs posix systems can clone this PR and have a go... that'd be really helpful!
I think there's some minor issues remaining, but should be real close.

TurkeyMan:
Okay, this is officially blocked on `opMove` and perhaps even the copy ctor DIP.
GCC's `std::string` has an interior pointer, and I can't move forward on this without addressing that :(

Geod24:
Do you have an unittest ? I tried to reproduce it with the following:
```d
extern(C++, `foo`) void str(T)(T arg);

extern(C++, `std`)
{
    void str(T)(T arg);

    struct char_traits(CharT) {}
    struct allocator(T) {}

    extern (C++, `__cxx11`) struct basic_string(T, Traits = char_traits!T, Alloc = allocator!T) {}
}

void main()
{
    basic_string!char bs;
    str(bs);
}
```

Result:
```console
% nm bug.o                                                                                                                                                                                                              0000000000000050 s EH_frame0
0000000000000238 S __D104TypeInfo_S3bug3std7__cxx11__T12basic_stringTaTSQBlQBk__T11char_traitsTaZQqTSQCoQCn__T9allocatorTaZQnZQCu6__initZ
                 U __D15TypeInfo_Struct6__vtblZ
0000000000000178 S __D36TypeInfo_S3bug3std__T9allocatorTaZQn6__initZ
00000000000000c0 S __D39TypeInfo_S3bug3std__T11char_traitsTaZQq6__initZ
0000000000000040 S __D3bug12__ModuleInfoZ
0000000000000320 S __D3bug3std7__cxx11__T12basic_stringTaTS3bugQBl__T11char_traitsTaZQqTSQBeQCo__T9allocatorTaZQnZQCv6__initZ
0000000000000170 S __D3bug3std__T11char_traitsTaZQq6__initZ
0000000000000230 S __D3bug3std__T9allocatorTaZQn6__initZ
0000000000000330 S __D6object__T10RTInfoImplVAmA2i1i0Z4datayG2m
0000000000000340 S __D6object__T10RTInfoImplVAmA2i1i0ZQwyPm
0000000000000000 T __Dmain
                 U __Dmain
0000000000000068 S __Dmain.eh
                 U __ZN3foo3strINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEEvS6_
                 U __d_run_main
                 U _main
0000000000000028 T _main
0000000000000090 S _main.eh
```

So the mangled name is `__ZN3foo3strINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEEvS6`.
There is however a bug when `foo` is part of `std`, as the first `std` does not get substituted. But I could not reproduce a case where substitution happens when it should not, only missing substitution.

Geod24:
Changing `test` definition to be `extern(C++)` and adding a pragma (because it does not end up in the object file with this test):
```d
extern(C++) void test(basic_string!char arg);
pragma(msg, test.mangleof);
```

```console
% /Users/geod24/projects/install/osx/bin/dmd -c bug2.d
_Z4testNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
% /Users/geod24/projects/install/osx/bin/dmd | head -n 1
DMD64 D Compiler v2.084.0-236-g747d605ff
```

Tried on a Linux64, same result

TurkeyMan:
:/ ... Test it with `core.stdcpp.string` from my branch?

TurkeyMan:
This is the exact declaration in `stdcpp.string`
```d
private alias AliasSeq(Args...) = Args;
private enum StdNamespace = AliasSeq!("std", "__cxx11");

extern(C++, "std") struct char_traits(CharT) {}

extern(C++, class)
extern(C++, (StdNamespace))
struct basic_string(T, Traits = char_traits!T, Alloc = allocator!T) { ... }
```

12345swordy:
DIP1014 has now been recently implemented.

thewilsonator:
I thought it was just the runtime section was implemented.

12345swordy:
Phobos side of DIP1014 has been implemented as well.

thewilsonator:
I should have worded that as: has the compiler section been done.

12345swordy:
No, I don't see any pr regarding dip 1014 on the dmd side of things. How difficult is it to implement it in compiler section?

TurkeyMan:
Windows passes comprehensive tests, and I think we should make the tests work and merge it. At least that will make a working reference with tests running on at least one platform that other platforms can dev against. The test ecosystem makes it SOOO HARD to get this working in one big lump.

thewilsonator:
`unittest.obj : error LNK2001: unresolved external symbol "public: __cdecl std::_Lockit::~_Lockit(void)" (??1_Lockit@std@@QEAA@XZ)
950`

TurkeyMan:
Visual Studio version is old. MS like to fiddle with that iterator debug struct between VS versions.
It's been stable since 2015 AFAIK, so is our VS compiler super-old?

We need the makefile to check the MSC version installed, and then translate that into a version token like `_MSCVER140`. I guess we can run the MSC version query command, parse the result, and assign it to an environment variable or something?

TurkeyMan:
As much as I'd like to just say we support the 'current' versions of respective C compilers, that's not realistic, so I can support a reasonable back-list, just need to know which one to target when compiling...

thewilsonator:
Should be able to use the target info for that? In fact we really just need the `-D` switch from C/C++ compilers.

TurkeyMan:
I'm not sure what you mean?
We need to know what version of msc the cpp code was compiled with... That's not a thing target info knows, it's just a thing that the user has to inform via the command line. If the user omits an `-version=_MSCVER1xx`, then it will assume latest VS version, but since we have a crazy test farm, I think we need to get the msc version and translate it to a version since it'll be different on every build box.

TurkeyMan:
Can we also echo the msc version to the build log somewhere?

thewilsonator:
That would also work. I think you'd just need to add a command to do it to https://github.com/dlang/druntime/blob/master/appveyor.yml

TurkeyMan:
I'm fatigued, and it's almost 3am, I'm going to bed. Gnite.

thewilsonator:
`posix.mak:359: recipe for target 'checkwhitespace' failed`

TurkeyMan:
I'm iterating on the test problems. Check what's going on in the diff...

TurkeyMan:
It's getting pretty close...

TurkeyMan:
I think this is it... it's gonna pass!

TurkeyMan:
Oh FFS! Auto-tester seems to be completely different than appveyor >_<

TurkeyMan:
And it's always great how there's a half-hour turn-around time between trying to tweak the build script and finding out if you fixed it or not!

TurkeyMan:
Typical, there's exactly one machine in the farm running VS2013, which is very old... so I need to detect and support it.

thewilsonator:
Lets give this a little bit of testing priority...

TurkeyMan:
I'm rebasing, since the last merge collides.

TurkeyMan:
This is working, the test just depends on getting an environment variable into make >_<

win64.mak
```diff
@@ -108,7 +108,8 @@ test_hash:
 	"$(DMD)" -m$(MODEL) -conf= -Isrc -defaultlib=$(DRUNTIME) -run test\hash\src\test_hash.d
 
 test_stdcpp:
-	"$(MAKE)" -f test\stdcpp\win64.mak "DMD=$(DMD)" MODEL=$(MODEL) "VCDIR=$(VCDIR)" DRUNTIMELIB=$(DRUNTIME) "CC=$(CC)" test
+	setmscver.bat
+	"$(MAKE)" -f test\stdcpp\win64.mak "DMD=$(DMD)" MODEL=$(MODEL) "VCDIR=$(VCDIR)" DRUNTIMELIB=$(DRUNTIME) "CC=$(CC)" "_MSC_VER=$(_MSC_VER)" test
```
TurkeyMan: This is the CI issue:
`setmscver.bat` sets the `_MSC_VER` variable, but then when make executes the next line (make), it seems like the variable is forgotten...

TurkeyMan:
ffs, something else took the win64 runner; so we'll have to wait an hour again >_<

win64.mak
wilzbach: It's not forgotten. You simply execute different processes that fork from the parent environment. Try e.g. `&&`

TurkeyMan: child processes should inherit the parent environment?

thewilsonator:
It didn't it just took a while to come online.

TurkeyMan:
while our lives slowly drift away...

TurkeyMan:
slowly.

win64.mak
TurkeyMan: I can't find an incantation that works... how are you supposed to do this? I need to set a make variable from a thing that a script determines?

rainers: `setmscver.bat && "$(MAKE)" "_MSC_VER=%_MSC_VER%"` should work.

TurkeyMan: it doesn't... i tried every conceivable combination of stuff like that I could think of.
I found a solution though.

But there's another problem; I just realised that windows make isn't actually make... It's an incomplete make clone!

It won't let me do `ifeq` in the makefile... what's that about? How do I set a variable conditionally?

src/core/stdcpp/allocator.d
```diff
@@ -30,35 +25,114 @@ struct allocator(T)
...
-        enum size_t max_size = size_t.max / T.sizeof;
+        T* allocate(size_t count) @nogc
+        {
+            static if (_MSC_VER <= 1800)
+            {
+                import core.stdcpp.xutility : _Xbad_alloc;
+                if (count == 0)
+                    return null;
+                T* mem;
+                if ((size_t.max / T.sizeof < count) || (mem = __cpp_new(count * T.sizeof)) == 0)
```
thewilsonator:
```
src\core\stdcpp\allocator.d(52): Error: cannot implicitly convert expression `__cpp_new(count * 4LU)` of type `void*` to `int*`
test\stdcpp\src\allocator_test.d(17): Error: template instance `core.stdcpp.allocator.allocator!int` error instantiating
src\core\stdcpp\allocator.d(52): Error: cannot implicitly convert expression `__cpp_new(count * 8LU)` of type `void*` to `double*`
test\stdcpp\src\allocator_test.d(18): Error: template instance `core.stdcpp.allocator.allocator!double` error instantiating
src\core\stdcpp\allocator.d(52): Error: cannot implicitly convert expression `__cpp_new(count * 24LU)` of type `void*` to `MyStruct*`
test\stdcpp\src\allocator_test.d(19): Error: template instance `core.stdcpp.allocator.allocator!(MyStruct)` error instantiating
```

```suggestion
-                 if ((size_t.max / T.sizeof < count) || (mem = __cpp_new(count * T.sizeof)) == 0)
+                if ((size_t.max / T.sizeof < count) || (mem = cast(T*)__cpp_new(count * T.sizeof)) is null)
```

win64.mak
rainers: It's Digital Mars make, a very basic make with some surprising behavior. (see https://dlang.org/blog/2018/06/11/dasbetterc-converting-make-c-to-d/)

> How do I set a variable conditionally?

I don't think that is possible directly. Depending on the problem invoking make recursively might be part of a solution.

src/core/stdcpp/allocator.d
TurkeyMan: I can see that ;) ... i just had to wait 2 hours to get the error message!

rainers: > I can see that ;) ... i just had to wait 2 hours to get the error message!

If you cannot test locally, maybe it would be more productive to setup your own CI on appveyor, you can select an image with VS2013 installed, too (see top of appveyor.xml). You can probably set it up before next iteration here is done...

TurkeyMan: I pushed before going to work. if it fails again, i'll do that tonight.
I had no idea appveyor had testers for 2013 still. we should just use it for all windows builds and get rid of the win32 auto-tester machines.

Look at the auto-tester logs; most machines take 5-10 minutes, except the windows machines which take 30-50 miunutes...

TurkeyMan:
Done!

I don't think those buildkite issues are related to this PR...

TurkeyMan:
Almost 1 year in the making!
What a fucking marathon! >_<

TurkeyMan:
Hooolyyy smokes. That happened.
Now let's never speak of it again!

changelog/std_string.dd
```diff
@@ -0,0 +1,3 @@
+Added `core.stdcpp.string`.
+
+Added `core.stdcpp.string`, which links against C++ `std::string`
```
wilzbach: Should have stated that it's only for Windows :/

TurkeyMan: Good point. I'll update that, or I'll just make the other platforms CI work. (Well, Linux still has the interior point issue...)

PR #2311, opened 2018-09-24
Move hashOf non-documentation unittests to tests/hash

n8sh:
Tests like these can be run in CI and don't need to be repeated in every user's unittest builds. This can be done for other tests too. I am dividing the work among several PRs to make the review easier.


test/hash/src/test_hash.d
```diff
@@ -0,0 +1,430 @@
...
+    issue_18918();
+    issue_18925();
+    issue_19005();
+    test_TypeInfo_Array_getHash_1();
+    test_TypeInfo_Array_getHash_2();
+    pr_2243();
+}
+
+/// Check that `hashOf` can be called on an array of void pointers.
+void hash_of_void_ptr_array() @nogc nothrow pure @system
```
jacob-carlborg: Prefer to use camel casing.

n8sh: Fixed.

n8sh:
Rebased.

PR #2314, opened 2018-09-29
std::array<> posix tests

TurkeyMan:
Trying to jig the posix tests in the normal way.


TurkeyMan:
@rainers [or anyone else that knows] I tried to copy all the other posix tests, but I'm not sure why this doesn't work.
I setup a linux environment, and tried to run it locally, but, naturally, it's impossible to build DMD/druntime/whatever locally by following the instructions (https://wiki.dlang.org/Building_under_Posix).

rainers:
> I setup a linux environment, and tried to run it locally, but, naturally, it's impossible to build DMD/druntime/whatever locally by following the instructions (https://wiki.dlang.org/Building_under_Posix).

I mount my dmd/druntime/phobos source folder in a Ubuntu12 VM (not good with autcrlf, though ;-)), create a link for `*/generated/linux` to the linux-drive so links work and use this script for building:
```
MODEL=64
export HOST_DMD=~/dlang/dmd-2.078.1/linux/bin64/dmd

cd dmd/src
make -f posix.mak
make -f posix.mak MODEL=$MODEL  || exit 1
cd ../..

BUILD=release
cd druntime
make -f posix.mak MODEL=$MODEL BUILD=$BUILD || exit 1
cd ..

cd phobos
make -f posix.mak MODEL=$MODEL BUILD=$BUILD || exit 1
cd .. 
```

test/stdcpp/Makefile
```diff
@@ -0,0 +1,18 @@
+include ../common.mak
+
+TESTS:=array
+
+.PHONY: all clean
+all: $(addprefix $(ROOT)/,$(addsuffix .done,$(TESTS)))
+
+$(ROOT)/test_stdcpp.done: $(ROOT)/%.done : $(ROOT)/%
```
rainers: This should just be
```
$(ROOT)/%.done : $(ROOT)/%
```

test/stdcpp/Makefile
```diff
@@ -0,0 +1,18 @@
...
+	@echo Testing $*
+	$(QUIET)$(TIMELIMIT)$(ROOT)/$* $(RUN_ARGS)
+	@touch $@
+
+$(ROOT)/%: $(SRC)/%.d $(SRC)/%.cpp
+	$(QUIET)$(CC) $(CFLAGS) -c -o $(ROOT)/%_cpp.o $(SRC)/%.cpp
```
rainers: use `$*` instead of `%` inside the rules.

rainers: The dated C++ compiler installed in my VM doesn't create the target directory, so you have to prefix this with
```
	mkdir -p $(dir $@)
```

test/stdcpp/Makefile
TurkeyMan: I copied this from neighbour makefiles, which are exactly like this...

rainers: But they'd have a test target `test_stdcpp` in that case.

PR #2315, opened 2018-10-01
Fix Issue 19280 - Remove unnecessary error checks in core.time.currSystemTick

n8sh:
`QueryPerformanceCounter` doesn't fail on Windows XP or later.
https://msdn.microsoft.com/en-us/library/windows/desktop/ms644904(v=vs.85).aspx

`gettimeofday` called with a valid timeval address and a null second parameter doesn't fail.
http://pubs.opengroup.org/onlinepubs/9699919799/functions/gettimeofday.html

`clock_gettime` called with a valid clock_id and a valid timespec address is in principle allowed to fail if the number of seconds doesn't fit in time_t, so even though no known implementation does this it is probably best to retain the error check at this time.
http://pubs.opengroup.org/onlinepubs/9699919799/functions/clock_getres.html

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -3344,10 +3344,8 @@ struct TickDuration
         import core.internal.abort : abort;
         version (Windows)
         {
-            ulong ticks;
-            if (QueryPerformanceCounter(cast(long*)&ticks) == 0)
-                abort("Failed in QueryPerformanceCounter().");
-
+            ulong ticks = void;
+            QueryPerformanceCounter(cast(long*)&ticks);
             return TickDuration(ticks);
         }
         else version (Darwin)
@@ -3356,10 +3354,8 @@ struct TickDuration
                 return TickDuration(cast(long)mach_absolute_time());
             else
             {
-                timeval tv;
-                if (gettimeofday(&tv, null) != 0)
-                    abort("Failed in gettimeofday().");
-
+                timeval tv = void;
+                gettimeofday(&tv, null);
                 return TickDuration(tv.tv_sec * TickDuration.ticksPerSec +
                                     tv.tv_usec * TickDuration.ticksPerSec / 1000 / 1000);
             }
@@ -3377,10 +3373,8 @@ struct TickDuration
             }
             else
             {
-                timeval tv;
-                if (gettimeofday(&tv, null) != 0)
-                    abort("Failed in gettimeofday().");
-
+                timeval tv = void;
+                gettimeofday(&tv, null);
                 return TickDuration(tv.tv_sec * TickDuration.ticksPerSec +
                                     tv.tv_usec * TickDuration.ticksPerSec / 1000 / 1000);
             }


CyberShadow:
I don't think this is a good change. We should be writing code against the interface, not the current observed behavior of current implementations. If a function returns an error code, we should check it.

CyberShadow:
@ZombineDev I think it's better to leave non-trivial/obviously-correct PRs open for at least 24 hours, so that everyone around the globe has a chance to review them.

jmdavis:
I don't know if the changes are good or not (I haven't had a chance to look at it in detail yet), but I second that in general, non-trivial changes should not be merged in just a few hours. And since the date/time stuff is kind of my thing, in general, I'd like the opportunity to review it before it gets merged.

n8sh:
>I don't think this is a good change. We should be writing code against the interface, not the current observed behavior of current implementations.

I agree and we are doing precisely that.

[QueryPerformanceCounter function](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644904%28v=vs.85%29.aspx): "On systems that run Windows XP or later, the function will always succeed and will thus never return zero." _(D doesn't support any versions of Windows earlier than Windows XP.)_

[gettimeofday](http://pubs.opengroup.org/onlinepubs/9699919799/functions/gettimeofday.html): "The gettimeofday() function shall return 0 and no value shall be reserved to indicate an error." _(If the second parameter is non-null the behavior is undefined by this specification.)_

CyberShadow:
I see, thanks for the clarification.

Personally, I would prefer to check the error code even if it is documented as never needing it; unsupported or non-conforming third-party implementations (consider e.g. Wine or asm.js-based libc implementations) might have reasons to let these functions fail despite the documentation of the interface they are implementing. But I would agree that these goals are in conflict with more tangible ones, like reducing code bloat or making the D code using it `nothrow`.

CyberShadow:
I see, thanks for the clarification.

Personally, I would prefer to check the error code even if it is documented as never needing it; unsupported or non-conforming third-party implementations (consider e.g. Wine or asm.js-based libc implementations) might have reasons to let these functions fail despite the documentation of the interface they are implementing. But I would agree that these goals are in conflict with more tangible ones, like reducing code bloat or making the D code using it `nothrow`.

jmdavis:
I'm semi-tempted to argue that the reason for the lack of error-checking should be documented with comments in each case (especially if the other case is still checking the return value), but the API documentation is quite clear that they're never going to fail. As for wine, well, at the end of the day, it really needs to be conforming. I know that it frequently isn't (far, far too frequently, sadly), but there's only so much that we can do there, and we don't officially support wine. If a D program fails on wine because of this, it's clearly a wine bug, and honestly, these aren't functions that I ever would have expected to fail in practice anyway. If they do, something is seriously wrong.

CyberShadow:
Yeah. I guess I'm mostly bugged by that we're discarding the functions' return value; and, we have every reason to expect what the return value is, but we don't *know* that it will, because the implementation is out of our control.

PR #2317, opened 2018-10-01
Fix Issue 19280 - Remove unnecessary error checks in core.time, Part 2

n8sh:
`clock_gettime` called with a valid `clock_id` and a valid `timespec` address is in principle allowed to fail if the number of seconds doesn't fit in `time_t`. If `time_t` is `long` or `ulong` we don't need to check for this since the current time won't overflow before 292 billion years A.D.

This PR also extends PR #2315 's improvments to `currSystemTick` to also apply to `currTime`.

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -2108,26 +2108,26 @@ struct MonoTimeImpl(ClockType clockType)
 
         version (Windows)
         {
-            long ticks;
-            if (QueryPerformanceCounter(&ticks) == 0)
-            {
-                // This probably cannot happen on Windows 95 or later
-                import core.internal.abort : abort;
-                abort("Call to QueryPerformanceCounter failed.");
-            }
+            long ticks = void;
+            QueryPerformanceCounter(&ticks);
             return MonoTimeImpl(ticks);
         }
         else version (Darwin)
             return MonoTimeImpl(mach_absolute_time());
         else version (Posix)
         {
-            timespec ts;
-            if (clock_gettime(clockArg, &ts) != 0)
+            timespec ts = void;
+            immutable error = clock_gettime(clockArg, &ts);
+            // clockArg is supported and if tv_sec is long or larger
+            // overflow won't happen before 292 billion years A.D.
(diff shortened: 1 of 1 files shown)


CyberShadow:
As in #2315 I'm not entirely convinced this is a direction we should put much effort into. It seems to be going against the principles of defensive programming considering the API implementations are beyond our control. The benefits of this change seem to be very small (removal of one conditional jump per function), and two of the hunks in this PR increase the number of lines of code rather than simplifying it.

jmdavis:
I'm not sure that I'm completely comfortable with the changes either. However, as far as I can tell, they _are_ technically correct

src/core/time.d
```diff
@@ -2108,21 +2108,19 @@ struct MonoTimeImpl(ClockType clockType)
...
             return MonoTimeImpl(mach_absolute_time());
         else version (Posix)
         {
-            timespec ts;
-            if (clock_gettime(clockArg, &ts) != 0)
+            timespec ts = void;
+            static if (ts.tv_sec.max >= long.max)
+                // clockArg is supported and overflow won't happen before 2.82e11 A.D.
+                clock_gettime(clockArg, &ts);
+            else if (clock_gettime(clockArg, &ts) != 0)
```
jmdavis: Please adjust both of the sections so that there's only one call to `clock_gettime` e.g. something like

            timespec ts = void;
            immutable result = clock_gettime(clockArg, &ts);
            // clockArg is supported, and if tv_sec is long or larger, overflow
            // won't happen before 2.82e11 A.D.
            static if (ts.tv_sec.max < long.max)
            {
                if (result != 0)
                {
                    import core.internal.abort : abort;
                    abort("Call to clock_gettime failed.");
                }
            }

Repeating the calls like that just seems needlessly risky with regards to long term maintenance when the only difference that's really needed is whether the return value is actually read. There shouldn't be any difference in the resulting binary.

n8sh: Done.

PR #2318, opened 2018-10-01
Fix Issue 19282 - hashOf segfaults for non-null C++ objects

n8sh:
Without this PR the following code crashes:

```d
extern(C++) class SomeCppClass {}

void main()
{
    SomeCppClass c = new SomeCppClass();
    size_t h = .hashOf(c);
}
```

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -57,6 +57,18 @@ private enum isFinalClassWithAddressBasedHash(T) = __traits(isFinalClass, T)
     static assert(!isFinalClassWithAddressBasedHash!C3);
 }
 
+private template isCppClassWithoutHash(T)
+{
+    static if (!is(T == class) && !is(T == interface))
+        enum isCppClassWithoutHash = false;
+    else
+    {
+        import core.internal.traits : Unqual;
+        enum bool isCppClassWithoutHash = __traits(getLinkage, T) == "C++"
+            && !is(Unqual!T : Object) && !hasCallableToHash!T;
+    }
+}
+
 /+
 Is it valid to calculate a hash code for T based on the bits of its
 representation? Always false for interfaces, dynamic arrays, and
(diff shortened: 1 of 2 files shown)


kinke:
I don't think any `extern(C++)` class can derive from `extern(D) Object` (vtbl layout, monitor field...).

n8sh:
>I don't think any extern(C++) class can be derived from extern(D) Object (or any other D base class, due to vtbl layout, monitor field...).

I tried it out and this runs:

https://run.dlang.io/is/ibGnDk
```d
extern(C++) class SomeCppClass : Object {}

void main()
{
    SomeCppClass c = new SomeCppClass();
    size_t h = .hashOf(c);
}
```

kinke:
Surprising/shocking/whatever. :]

Adding `c.toString()` yields a crash, as expected.

schveiguy:
> I tried it out and this runs:

That shouldn't compile. @kinke is right. Either a class has a C++ layout or a D layout. It can't have both, they are conflicting.

src/core/internal/hash.d
```diff
@@ -553,7 +565,8 @@ size_t hashOf(T)(scope const T val, size_t seed = 0) if (!is(T == enum) && is(T
 @nogc nothrow pure @trusted
 size_t hashOf(T)(scope const T val)
 if (!is(T == enum) && (is(T == interface) || is(T == class))
-    && isFinalClassWithAddressBasedHash!T)
+    && isFinalClassWithAddressBasedHash!T
+    || isCppClassWithoutHash!T)
```
wilzbach: As this is used three times it might make sense to define `canBitwiseHash` as a separate template.

n8sh:
@wilzbach I've edited the PR description to not refer to C++ classes inheriting from Object, so I think there's nothing remaining to be addressed.

PR #2319, opened 2018-10-02
Add a gc that can be used to devel a gc using delegates

BBasile:
Ping @nordlow . This is not as nice as i expected but this can be used to experiment without recompiling everything after each change. Working example:

```d
module runnable;

extern(C) __gshared const string[] rt_options = [ "gcopt=initReserve:2 gc:handlers" ];

import gc.impl.handlers.gc;
import gc.gcinterface;
import core.memory;

struct Handler
{
    void assignHandlers()
    {
        assert(hgc);
        if (hgc)
            hgc.mallocHandler = &malloc;
    }

    void* malloc(size_t s, uint, const TypeInfo) nothrow
    {
        import core.stdc.stdio;
        printf("some allocation happening...\n");
        import core.stdc.stdlib : m = malloc;
        return m(s);
    }
}

void main(string[] args)
{
    auto c = new class {int[32] b;}; // lazy init
    Handler h;
    h.assignHandlers;
    auto d = new class {int[32] b;}; // this once calls the handler
} 
```

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -16,6 +16,12 @@ COPY=\
 	$(IMPDIR)\core\time.d \
 	$(IMPDIR)\core\vararg.d \
 	\
+	$(IMPDIR)\gc\gcinterface.d \
+    $(IMPDIR)\gc\impl\handlers\gc.d \
+    $(IMPDIR)\gc\config.d \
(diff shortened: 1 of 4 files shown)


BBasile:
sorry i wanted to open this in my fork

PR #2322, opened 2018-10-05
[WIP] rt.sections_elf_shared: Add support for OSX (prepare for shared Phobos lib)

kinke:
This is almost exactly LDC's version, which uses this module for 32/64-bit OSX too, incl. support for Phobos as shared library, and doesn't use `rt.sections_osx_x86[_64]` at all.


kinke:
For this to work on OSX, DMD needs to be adapted. After a quick glance, [this block](https://github.com/dlang/dmd/blob/aa391954552990ee8895d74fcfc42df3ac2d1386/src/dmd/backend/elfobj.c#L3384-L3719) would be needed for Mach-O objects too, in addition to
* a dummy TLS 'anchor' variable for the DSO,
* a little helper function (`void* getTlsAnchor() { return &tlsAnchor; }`), returning its address for the executing thread, and
* adding the pointer to that function to the compiler-generated DSO data struct.

src/rt/sections_elf_shared.d
```diff
@@ -36,6 +40,15 @@ else version (FreeBSD)
     import core.sys.freebsd.sys.elf;
     import core.sys.freebsd.sys.link_elf;
 }
+else version (OSX)
```
jacob-carlborg: Not sure I understand when you're using `SharedDarwin` and when you're using `OSX`.

src/rt/sections_elf_shared.d
```diff
@@ -894,42 +965,58 @@ struct tls_index
...
-else version (MIPS32)
-    enum TLS_DTV_OFFSET = 0x8000;
-else version (MIPS64)
-    enum TLS_DTV_OFFSET = 0x8000;
+    void[] getTLSRange(void *tlsSymbol) nothrow @nogc
+    {
+        void* start = null;
+        size_t size = 0;
+        _d_dyld_getTLSRange(tlsSymbol, &start, &size);
+        assert(start && size, "Could not determine TLS range.");
```
jacob-carlborg: This check will be removed if asserts are removed, is that intentional?

src/rt/sections_elf_shared.d
```diff
@@ -352,6 +388,12 @@ else
 // Compiler to runtime interface.
 ///////////////////////////////////////////////////////////////////////////////
 
+version (OSX)
+    private alias ImageHeader = mach_header*;
```
jacob-carlborg: Will this work for both 32bit and 64bit? There's a `mach_header_64` type as well.

jacob-carlborg:
> For this to work on OSX, DMD needs to be adapted. After a quick glance, this block would be needed for Mach-O objects too

Yeah, that's the big task 😃.

src/rt/sections_elf_shared.d
kinke: Yeah I noticed the inconsistency too. It's not my code, and as you know, I have no interest in & knowledge of the different Mac OS flavours (desktop, iOS, watchOS, whatever).

src/rt/sections_elf_shared.d
kinke: Most likely not => #2324.

src/rt/sections_elf_shared.d
jacob-carlborg: It is your code, you added `SharedDarwin`.

src/rt/sections_elf_shared.d
kinke: It does work for LDC, apparently for the reason you mentioned (used for `_dyld_get_image_header_containing_address()`, `_dyld_get_image_slide()` and LDC-specific hacky [getSection()](https://github.com/ldc-developers/druntime/blob/ldc/src/rt/mach_utils.d#L31); the latter is available in the `sections_osx_*.d` files here and would need to be moved.]

jacob-carlborg:
> It should be doable. My intention was to prepare the druntime side of it, so that DMD is the only blocker. I was hoping you specifically would be interested and knowledgeable enough to finally make DMD catch up with LDC in this regard ;) 

I'm definitely interested. If I'm knowledgeable enough that's a different question 😃. I would think I could handle the druntime side of it, i.e. basically what you have in this PR. But the compiler side, I'm not so sure. If the assembly is the same as for ELF, I could probably get the data into the correct sections for Mach-O but if the assembly is different then it would be much more difficult for me. Also it's the standard issue with the lack of time.

> Btw, the ModuleInfos section range currently required for the compiler-generated DSO data is the reason for the linking errors you got with LLD (magic linker symbols); deriving them from the image headers would be nicer indeed. That should be pretty simple for OSX via getsectbyname(); more work for ELF, although all the code to get an ELF section by name would already be in rt.backtrace.elf. I'll see if I can refactor it, so that we can use it here too (and @joakim-noah could use it for Android too - as well as user code inspecting binary images).

Coo. BTW, I know that DMD had problems in the past with bracketed sections and the linker messing them up. That's why DMD stopped using them when someone showed how easy it was to get access to a section. But I guess LDC doesn't have those problems since it's using LLVM.

jacob-carlborg:
FYI, DMD only uses the native TLS implementation on macOS 64bit. On 32bit it still uses the old emulated TLS implementation.

Geod24:
So @kinke , what's the status with this ? Still blocked on the DMD backend I suppose ?

kinke:
Yes.

ibuclaw:
Incase it hasn't been said already, this module is in desperate need of a rename if you add Darwin support to it.

src/rt/sections_elf_shared.d
```diff
@@ -809,6 +872,12 @@ bool findDSOInfoForAddr(in void* addr, dl_phdr_info* result=null) nothrow @nogc
          */
         return dl_iterate_phdr(&callback, &dg) != 0;
     }
+    else version (OSX)
+    {
+        auto header = _dyld_get_image_header_containing_address(addr);
+        if (result) *result = header;
+        return !!header;
```
ibuclaw, line 879: Without knowing this existed, I independently came up with this implementation for findDSOInfoForAddr:
```
foreach (i; 0 .. _dyld_image_count())
{
    const header = _dyld_get_image_header(i);
    const slide = _dyld_get_image_vmaddr_slide(i);
    const section = getSection(header, slide, SEG_DATA, SECT_DATA);

    // less than base address of section means quick reject
    if (!section.length || addr < section.ptr)
       continue;

    if (addr < section.ptr + section.length)
    {
        result.header = header;
        result.slide = slide;
        return true;
    }
}
return false;
```
Probably could align ourselves up, mixing what you have here with `_dyld_get_image_vmaddr_slide` for getting the slide value.  So you don't have to rely on a private function later.

ibuclaw: On darwin9, and older versions of darwin10, will have to use `dladdr()` instead.

kinke: This PR is more than 2 years old - this now looks like this: https://github.com/ldc-developers/druntime/blob/a261a84e4c6b208c65ae39da0ae844ae3dd06744/src/rt/sections_elf_shared.d#L903-L907

jacob-carlborg: The slide can be calculated like this [1]. Also note this [2]:

"The following functions allow you to iterate through all loaded images. This is not a thread safe operation. Another thread can add or remove an image during the iteration.  
 
Many uses of these routines can be replace by a call to dladdr() which  will return the mach_header and name of an image, given an address in  the image. dladdr() is thread safe."

[1] https://github.com/ldc-developers/druntime/blob/a261a84e4c6b208c65ae39da0ae844ae3dd06744/src/rt/sections_darwin_64.d#L134

[2] https://opensource.apple.com/source/dyld/dyld-750.6/include/mach-o/dyld.h.auto.html

ibuclaw: @jacob-carlborg good to know, though you're still using [_dyld_image_count and get_image_header](https://github.com/ldc-developers/druntime/blob/a261a84e4c6b208c65ae39da0ae844ae3dd06744/src/rt/sections_darwin_64.d#L40-L42)

jacob-carlborg: Yeah, IIRC I couldn't get that to work with `dladdr`.

Geod24:

ibuclaw:
FYI I ended up naming these files after their object format (sections.macho, sections.elf, sections.pecoff), rather than trying to shoehorn it all into one place.  All inevitable duplication then was moved over to a sections.common module.

PR #2324, opened 2018-10-09
rt.sections_elf_shared: Replace (almost) all runtime assertions by abort()

kinke:
An `assert(0)` in the release libs yields an "invalid instruction" error, which isn't very helpful for troubleshooting.

Normal assertions (throwing `AssertError`s) aren't really usable as the GC may not be initialized (yet or anymore).

### DIFF
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -59,6 +59,18 @@ import rt.minfo;
 import rt.util.container.array;
 import rt.util.container.hashtab;
 
+/****
+ * Asserts the specified condition, independent from -release, by abort()ing.
+ * Regular assertions throw an AssertError and thus require an initialized
+ * GC, which isn't the case (yet or anymore) for the startup/shutdown code in
+ * this module (called by CRT ctors/dtors etc.).
+ */
+private void safeAssert(bool condition, scope string msg, size_t line = __LINE__) @nogc nothrow @safe
+{
+    import core.internal.abort;
+    condition || abort(msg, __FILE__, line);
+}
+
 alias DSO SectionGroup;
 struct DSO
 {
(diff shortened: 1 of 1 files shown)


src/rt/sections_elf_shared.d
```diff
@@ -522,8 +531,8 @@ version (Shared)
     void decThreadRef(DSO* pdso, bool decAdd)
     {
         auto tdata = findThreadDSO(pdso);
-        tdata !is null || assert(0);
-        !decAdd || tdata._addCnt > 0 || assert(0, "Mismatching rt_unloadLibrary call.");
```
n8sh: What do you think of using the style `condition || abort("message")`? `abort` already does the trick of auto-filling the file name and line number.

src/rt/sections_elf_shared.d
```diff
@@ -59,6 +59,12 @@ import rt.minfo;
 import rt.util.container.array;
 import rt.util.container.hashtab;
 
+void safeAssert(bool condition, scope string msg, scope string filename = __FILE__, size_t line = __LINE__) @nogc nothrow @safe
```
n8sh: This would be a sensible place to include your comment
>Normal assertions (throwing AssertErrors) aren't really usable as the GC may not be initialized (yet or anymore).

Geod24:
@kinke : That looks like a perfect use case for https://github.com/dlang/dmd/pull/8718

ZombineDev:
@Geod24 Actually no, because the switch affects all compiled modules and we need something specific to code that runs before `rt_init()` returns.

src/rt/sections_elf_shared.d
Geod24: `private` + DDOC would be perfect

n8sh:
Will amend to include niggling suggestions then automerge.

src/rt/sections_elf_shared.d
kinke: I don't like it too much (and haven't seen it much in druntime/Phobos), that's why I decided to wrap it with an extra helper function.

n8sh:
For whatever reason I can't squash this. As soon as that's done this should be merged.

kinke:
Squashed manually.

ZombineDev:
Auto-merge toggled on

PR #2325, opened 2018-10-10
Bionic: Add x64 declarations, switch to single TLS function, and define __USE_GNU

joakim-noah:
This ~is a WIP~ pull is ready, ~as~ I'm going to add another commit (__Update:__ in a later pull) that reworks how druntime initializes TLS data on Android before passing it to the GC, based on a suggestion of Martin from the ldc team.

This first commit adding Bionic/x64 declarations and consolidating where possible is pretty much done, so putting it up for review. All druntime tests for Android/x86 and x64 pass when cross-compiled from linux/x64 to Termux running in [an Anbox Android/x64 container](https://anbox.io).

Pretty much the same Phobos tests pass as other Android platforms, ie only some math-related modules like `std.math` or `std.internal.math.gammafunction` assert, with the only difference that ~`std.random` asserts or segfaults in the massive last test block on both x86/x64 and~ `std.variant` segfaults in various tests on Android/x64. ~I'll look into those more and update this pull if needed.~

### DIFF
--- a/src/core/sys/posix/setjmp.d
+++ b/src/core/sys/posix/setjmp.d
@@ -269,6 +269,10 @@ else version (CRuntime_Bionic)
     {
         enum _JBLEN = 32;
     }
+    else version (X86_64)
+    {
+        enum _JBLEN = 11;
+    }
     else
     {
         static assert(false, "Architecture not supported.");
--- a/src/core/sys/posix/sys/types.d
+++ b/src/core/sys/posix/sys/types.d
@@ -247,34 +247,15 @@ else version (CRuntime_Bionic)
     alias c_long    time_t;
     alias uint      uid_t;
 
-    version (X86)
-    {
-        alias ushort    mode_t;
-        alias ushort    nlink_t;
-    }
-    else version (X86_64)
-    {
-        alias ushort    mode_t;
-        alias uint      nlink_t;
-    }
-    else version (ARM)
-    {
-        alias ushort    mode_t;
-        alias ushort    nlink_t;
-    }
-    else version (AArch64)
-    {
-        alias uint      mode_t;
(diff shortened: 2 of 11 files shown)


src/core/sys/posix/sys/types.d
```diff
@@ -247,34 +247,15 @@ else version (CRuntime_Bionic)
...
-    {
-        alias ushort    mode_t;
-        alias ushort    nlink_t;
-    }
-    else version (AArch64)
-    {
-        alias uint      mode_t;
-        alias uint      nlink_t;
-    }
-    else version (MIPS32)
```
joakim-noah: [Android NDK 17, released earlier this year, dropped support for MIPS](https://github.com/android-ndk/ndk/wiki/Changelog-r17), allowing this simplification. Nobody was using Android/MIPS anyway.

The `x64` `mode_t` declaration above was wrong.

src/core/sys/posix/setjmp.d
```diff
@@ -269,6 +269,10 @@ else version (CRuntime_Bionic)
     {
         enum _JBLEN = 32;
     }
+    else version( X86_64 )
```
ibuclaw: Wrong style.

thewilsonator:
@joakim-noah whats the status of this?

joakim-noah:
Still working on removing those TLS section delimiters, switched over to trying to use @kinke's #2330, why?

thewilsonator:
This was green, thats all.

joakim-noah:
OK, yeah, go ahead and merge, I'll get that last TLS commit in separately.

joakim-noah:
I think it needs an approved review first, merge won't go through otherwise.

PR #2326, opened 2018-10-11
Move non-standard DragonflyBSD AF and PF constants to core.sys.dragonflybsd

ibuclaw:
Concessions are made for IPX and APPLETALK, but they shouldn't really be there either (for all platforms, not just dragonfly).

FYI @dkgroot

### DIFF
--- a/src/core/sys/dragonflybsd/sys/socket.d
+++ b/src/core/sys/dragonflybsd/sys/socket.d
@@ -1,0 +1,86 @@
+/**
+ * D header file for DragonFlyBSD
+ *
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ */
+module core.sys.dragonflybsd.sys.socket;
+
+public import core.sys.posix.sys.socket;
+
+version (DragonFlyBSD):
+extern(C):
+@nogc:
+nothrow:
+
+enum
+{
+    AF_IMPLINK          = 3,
+    AF_PUP              = 4,
+    AF_CHAOS            = 5,
+    AF_NETBIOS          = 6,
+    AF_ISO              = 7,
+    AF_OSI              = AF_ISO,
+    AF_ECMA             = 8,
+    AF_DATAKIT          = 9,
+    AF_CCITT            = 10,
+    AF_SNA              = 11,
+    AF_DECnet           = 12,
+    AF_DLI              = 13,
+    AF_LAT              = 14,
+    AF_HYLINK           = 15,
+    AF_ROUTE            = 17,
+    AF_LINK             = 18,
+    pseudo_AF_XTP       = 19,
+    AF_COIP             = 20,
+    AF_CNT              = 21,
+    pseudo_AF_RTIP      = 22,
+    AF_SIP              = 24,
+    pseudo_AF_PIP       = 25,
+    AF_ISDN             = 26,
+    AF_E164             = AF_ISDN,
+    pseudo_AF_KEY       = 27,
+    AF_NATM             = 29,
+    AF_ATM              = 30,
+    pseudo_AF_HDRCMPLT  = 31,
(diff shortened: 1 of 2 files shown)


jacob-carlborg:
@dkgroot

n8sh:
Isn't it a step backwards in usability if instead of getting everything from `import core.sys.posix.sys.socket` we have to explicitly include `core.sys.<name here>.sys.socket` for each system, considering that there is significant overlap between available non-standard options?

ZombineDev:
@n8sh the division between `core.sys.posix.*` and `core.sys.<posix_compatible_os` is not one that optimizes for convenience (e.g. like `package.d` modules), but for portability. The idea is to push the user in right direction by presenting him with an interface that has the highest chance of just working, regardless of which posix-compatible OS he is targeting. Of course we provide access to platform specific interfaces, but we must make sure that this is a conscious decision, not something that happened by accident.

n8sh:
Why are all the `PF_...` constants being moved even if the corresponding `AF_...` is not being moved?

ibuclaw:
`PF` is not posix compliant.

n8sh:
Ensuring code is cross-platform except if explicitly using platform-specific features is a good idea. Is there any posix-compatible OS that doesn't have `PF` that anyone has even contemplated supporting?

* [Haiku has PF](https://github.com/haiku/haiku/blob/master/headers/posix/sys/socket.h) (and Haiku generally is a stickler for not going one inch beyond what posix requires).
* [Minix too](https://github.com/Stichting-MINIX-Research-Foundation/minix/blob/master/sys/sys/socket.h).
* [Seemingly AIX too](https://chromium.googlesource.com/chromiumos/third_party/glibc/+/d16463090097b7257b382bceee74410b55316559/sysdeps/unix/sysv/aix/bits/socket.h).
* [Solaris too](https://github.com/illumos/illumos-gate/blob/master/usr/src/uts/common/sys/socket.h)
* Cygwin too.
* Hurd too.
* And of course linux and macOS and every BSD.

n8sh:
Okay, LynxOS might not have `PF`. I don't have the header file but `PF_*` doesn't appear in the reference manual.

ZombineDev:
@n8sh De jure a Posix-compatible OS only needs to implement what is defined in the Posix standard and nothing else. De facto you may of course find a larger set of common APIs, but there's no guarantee that this larger set will remain portable, although it most likely will.

n8sh:
Thinking about this more I would urge to reverse changes of this type. Splitting definitions to different modules defeats compile-time introspection, which is one of D's greatest strengths. Here's a real example of something I wanted to be able to write in Phobos but was prevented by this type of separation:

```d
static if (clockType == ClockType.second)
{
    import t = core.sys.posix.time;
    static if (is(typeof(t.clock_gettime(t.CLOCK_SECOND, cast(t.timespec*) null))))
    {
        t.timespec ts = void;
        t.clock_gettime(t.CLOCK_SECOND, &ts);
        return unixTimeToStdTime(ts.tv_sec);
    }
    else
    {
        return unixTimeToStdTime(t.time(null));
    }
}
```

ibuclaw:
@n8sh - we could always publicly import platform specific modules from the posix equivalent.

ibuclaw:
(However one might have good reason to argue that as library maintainers, you should not be afraid to just manage the importing of platform specific modules yourself).

thewilsonator:
Whats the status of this?

dkgroot:
I am perfectly happy with the move of PF. (Sorry it took me a while to notice/respond).

(Isn't the posix standard mostly dead now adays. Most OS's seem to follow / mimic the two big open source OS'es)
The 'publicly import platform specific modules from the posix equivalent' or package-like solution suggested above  sounds like a good alternative. Then again most end users will mostly depend on phobos not directly on druntime and use the d-interface instead of going to the druntime low-level stuff.

PR #2327, opened 2018-10-11
Fix Issue 11174 - Define non-standard address family and socket option values for Linux

ibuclaw:
https://issues.dlang.org/show_bug.cgi?id=11174


src/core/sys/linux/sys/socket.d
```diff
@@ -13,7 +13,146 @@ extern(C):
...
+    PF_PHONET     = 35,
+    PF_IEEE802154 = 36,
+    PF_CAIF       = 37,
+    PF_ALG        = 38,
+    PF_NFC        = 39,
+    PF_VSOCK      = 40,
+    PF_KCM        = 41,
+    PF_QIPCRTR    = 42,
+    PF_SMC        = 43,
+    PF_MAX        = 44,
```
n8sh: At https://github.com/torvalds/linux/blob/master/include/linux/socket.h `PF_MAX` is 45. Where is your list coming from?

ibuclaw: Linux 4.15, which is not the latest release, but still fairly new.

ibuclaw: https://github.com/torvalds/linux/blob/v4.17/include/linux/socket.h#L211

So the new macro was introduced in 4.18, which is the current release.

ibuclaw: > Neither 4.15 nor 4.18 socket.h seem to have AF_FILE/PF_FILE. Where is it coming from? AF_IPX/PF_IPX and AF_APPLETALK/PF_APPLETALK are present in 4.15 & 4.18 socket.h but are not in this PR. Why?

You're looking at the wrong header. It's in glibc / musl / insert your C library here headers.

n8sh: Is it a problem that `PF_NETLINK` and `PF_ROUTE` are defined as 16 here and in linux/musl/uClibc's headers but glibc's socket.h has no `PF_NETLINK` (instead having `#define	PF_APPLETALK	16	/* Don't use this.  */`) and defines `PF_ROUTE` as 17?

ibuclaw: > AF_IPX/PF_IPX and AF_APPLETALK/PF_APPLETALK are present in 4.15 & 4.18 socket.h but are not in this PR. Why?

Check the posix.sys.socket module. At least APPLETALK is present there, although it perhaps shouldn't be as it's nonstandard.

ibuclaw: > Is it a problem that PF_NETLINK and PF_ROUTE are defined as 16 here and in linux/musl/uClibc's headers but glibc's socket.h has no PF_NETLINK (instead having #define PF_APPLETALK 16 /* Don't use this. */) and defines PF_ROUTE as 17?

I don't think so.  Can check what value ROUTE would be on glibc at runtime. If its already in Linux specific headers, then it shouldn't be overridden by libc.

PR #2330, opened 2018-10-14
Refactor ELF utils from rt.backtrace.elf and rt.sections_elf_shared to core.internal.elf.{dl,io}

kinke:
A common place for working with ELF binaries, accessible by user code as well and providing a bit more convenience.

### DIFF
--- a/src/core/elf.d
+++ b/src/core/elf.d
@@ -144,113 +144,173 @@ struct SharedObject
     }
 }
 
-// -------------------------------
-//  File-based memory-mapped I/O:
-// -------------------------------
-
-struct ElfFile
+/**
+ * File-based memory-mapped I/O (read-only).
+ * Only supports ELF files with a byte-order matching the target platform's.
+ * Params:
+ *     Elf_Ehdr = Expected type of the ELF header (Elf{32,64}_Ehdr)
+ *     Elf_Shdr = Expected type of the ELF section header (Elf{32,64}_Shdr)
+ *     ELFCLASS = Expected ELF class (ELFCLASS{32,64})
+ */
+template ElfIO(Elf_Ehdr, Elf_Shdr, ubyte ELFCLASS)
 {
-@nogc nothrow:
-    static bool open(const(char)* path, out ElfFile file)
+    struct ElfFile
     {
-        file = ElfFile(.open(path, O_RDONLY));
-        return file.isValid();
-    }
-
-    this(int fd)
-    {
-        this.fd = fd;
-        if (fd != -1)
+    @nogc nothrow:
+        static bool open(const(char)* path, out ElfFile file)
         {
-            // memory map header
-            this.ehdr = MMapRegion!Elf_Ehdr(fd, 0);
+            file = ElfFile(.open(path, O_RDONLY));
+            return file.isValid();
+        }
+
+        this(int fd)
+        {
+            this.fd = fd;
+            if (fd != -1)
+            {
+                // memory map header
+                this.ehdr = MMapRegion!Elf_Ehdr(fd, 0);
+            }
+        }
+
+        @disable this(this);
+
+        ~this()
+        {
+            if (fd != -1)
+                close(fd);
+        }
+
+        private int fd = -1;
+        MMapRegion!Elf_Ehdr ehdr;
+
+        bool isValid() const
+        {
+            enum EI_MAG0 = 0;
+            enum EI_MAG1 = 1;
+            enum EI_MAG2 = 2;
+            enum EI_MAG3 = 3;
+            enum EI_CLASS = 4;
+            enum EI_DATA = 5;
+
+            enum ELFMAG0 = 0x7f;
+            enum ELFMAG1 = 'E';
+            enum ELFMAG2 = 'L';
+            enum ELFMAG3 = 'F';
+
+            enum ELFCLASS32 = 1;
+            enum ELFCLASS64 = 2;
+
+            enum ELFDATA2LSB = 1;
+            enum ELFDATA2MSB = 2;
+
+            version (LittleEndian)   alias ELFDATA = ELFDATA2LSB;
+            else version (BigEndian) alias ELFDATA = ELFDATA2MSB;
+            else static assert(0, "unsupported byte order");
+
+            if (fd == -1)
+                return false;
+
+            const ident = ehdr.e_ident;
+
+            if (!(ident[EI_MAG0] == ELFMAG0 &&
+                  ident[EI_MAG1] == ELFMAG1 &&
+                  ident[EI_MAG2] == ELFMAG2 &&
+                  ident[EI_MAG3] == ELFMAG3))
+                return false;
+
+            if (ident[EI_CLASS] != ELFCLASS)
+                return false;
+
+            // the file's byte order must match the target's
+            if (ident[EI_DATA] != ELFDATA)
+                return false;
+
+            return true;
+        }
+
+        bool findSectionHeaderByName(const(char)[] sectionName, out ElfSectionHeader header) const
+        {
+            const index = findSectionIndexByName(sectionName);
+            if (index == -1)
+                return false;
+            header = ElfSectionHeader(this, index);
+            return true;
+        }
+
+        size_t findSectionIndexByName(const(char)[] sectionName) const
+        {
+            const stringSectionHeader = ElfSectionHeader(this, ehdr.e_shstrndx);
+            const stringSection = ElfSection(this, stringSectionHeader);
+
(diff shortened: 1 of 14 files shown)


src/core/elf.d
```diff
@@ -0,0 +1,255 @@
...
+ *
+ * Copyright: Copyright Digital Mars 2015 - 2015.
+ * License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Authors:   Yazan Dabain
+ * Source: $(DRUNTIMESRC core/elf.d)
+ */
+
+module core.elf;
+
+version (linux) version = linux_or_bsd;
```
jacob-carlborg: `linux_or_bsd` does not follow standard D naming conventions.

src/core/elf.d
```diff
@@ -0,0 +1,255 @@
...
+
+    @disable this(this);
+
+    ~this() @nogc nothrow
+    {
+        if (fd != -1) close(fd);
+    }
+
+    int fd = -1;
+    MMapRegion!Elf_Ehdr ehdr;
```
jacob-carlborg: Do we want these two to be directly exposed?

src/core/elf.d
```diff
@@ -0,0 +1,255 @@
...
+ *
+ * Reference: http://www.dwarfstd.org/
+ *
+ * Copyright: Copyright Digital Mars 2015 - 2015.
+ * License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Authors:   Yazan Dabain
+ * Source: $(DRUNTIMESRC core/elf.d)
+ */
+
+module core.elf;
```
jacob-carlborg: Perhaps this file should be located in a package so we can later add modules for Mach-O, COFF and OMF if we want to.

src/core/elf.d
```diff
@@ -0,0 +1,341 @@
...
+            }
+            return false;
+        }
+        else
+        {
+            return !!_rtld_addr_phdr(address, &result.info);
+        }
+    }
+
+    dl_phdr_info info;
```
jacob-carlborg: Do we want this to be directly exposed?

src/rt/backtrace/elf.d
```diff
@@ -45,330 +51,39 @@ struct Image
...
+            return null;
+
+        // we don't support compressed debug sections
+        if ((dbgSectionHeader.shdr.sh_flags & SHF_COMPRESSED) != 0)
+            return null;
+
+        auto dbgSection = ElfSection(file, dbgSectionHeader);
+        const sectionData = cast(const(ubyte)[]) dbgSection.get();
+        // do not munmap() the section data to be returned
+        import core.stdc.string;
```
jacob-carlborg: I think it's a but unusual to have an import in the middle like this.

src/core/elf.d
kinke: I think so; `rt.sections_elf_backtrace` currently needs it for the TLS stuff, and I don't want to abstract it fully away, particularly since the struct is OS-dependent.

src/core/elf.d
kinke: `fd` perhaps not, but `ehdr` is essential.

src/core/elf.d
kinke: Just taken from `rt.backtrace.elf`. Feel free to edit the files directly, that's way more efficient.

kinke:
That was my main goal, letting the user have access to this functionality too. E.g., we could presumably use it in LDC to parse the dependencies of an .so to be linked; if it depends on shared druntime/Phobos, LDC would use `-link-defaultlib-shared` to link the binary against the shared default libs too, preventing the runtime error 'recompile with -link-defaultlib-shared' for executables compiled without it and linked with shared D libs.

kinke:
Well, because we need the stuff in druntime too.

iain-buclaw-sociomantic:
I don't think it's wise to have a platform-specific module in a common place.

kinke:
I'm totally flexible wrt. final place of this; it could live in `core.sys.elf` or so too, it's just that I haven't found any suited existing dir yet (afaik not suited for Posix due to Apple's MachO).

joakim-noah:
Needs rebase and fix merge.

src/core/elf.d
```diff
@@ -0,0 +1,341 @@
...
+        return -1;
+    }
+}
+
+struct ElfSectionHeader
+{
+@nogc nothrow:
+    this(ref const ElfFile file, size_t index)
+    {
+        assert(Elf_Shdr.sizeof == file.ehdr.e_shentsize);
```
joakim-noah: Tried my upcoming Android commit for #2325 with this pull, but it segfaults here: will look into why. In the meantime, turn this into an `abort` message, just as you did with #2324, so it exits immediately rather than repeatedly trying to throw an exception and then segfaulting. I'm doing the same in my Android pull, replacing asserts with abort messages.

joakim-noah:
@kinke, any chance we can get this into the final 1.13 release? I'd like to use it for Android too.

kinke:
I don't think so; I'm not too keen on having this pretty big diff wrt. upstream in LDC.

joakim-noah:
No, I mean do you think you can get it merged here in time for a cherrypick before the 1.13 release?

kinke:
Lol, no way! :)

src/core/elf.d
kinke: This is probably because this module currently only supports *native* ELF files, i.e., corresponding with the host, which is quite a serious limitation.

I don't think aborting here is a good idea; #2324 was different, as the chances for the GC not being initialized yet/already shut down are very high for that code, so throwing in that code is to be avoided.

joakim-noah:
Heh, OK, I still need to look into why this pull doesn't work on Android anyway, maybe I'll just try to get it working for the native Android builds of LDC.

src/rt/backtrace/elf.d
kinke: This is a hack anyway and leaks memory (it already is or at least was when I wrote this; there's a bugzilla issue about it).

This `getDebugLineSectionData()` should probably be converted to a `processDebugLineSectionData(void delegate(const(ubyte)[] data) processor)` or so, so that the method can free the data itself afterwards.

kinke:
Reworked; now split into a tiny 'package' (no `package.d`):

* `core.elf.dl`: dynamic loader stuff to inspect the loaded shared ELF objects for the current process. IMO also useful for the user, e.g., for data to be attached to error/crash reports etc.
* `core.elf.io`: read-only file I/O for ELF files, usable on all Posix targets; e.g., allowing to inspect 32/64-bit ELF files on macOS too - if you provide the matching ELF header and section header types; the types in `core.sys.*` are only available for the target, so those won't do.

I also added a fix for the aforementioned bugzilla issue (memory leak during backtrace generation with `-g`); no test though.

ibuclaw:
> Perhaps this file should be located in a package so we can later add modules for Mach-O, COFF and OMF if we want to.

Indeed, something should go in-between core and elf.  `core.objfmt.elf`, `core.binfmt.elf`, `core.utils.elf`, ...

kinke:
Then please settle for one such infix, I'll adapt it then. My favourite would be `core.bin.elf.{dl,io}`.

jacob-carlborg:
Some cleanup among the commits would be nice.

src/core/elf.d
joakim-noah: I'm building on Android itself, but it's likely that some of the ELF definitions are different there.

As for aborting here, I will be using this on Android before the GC is initialized, but I didn't check how you're using it.

kinke: It doesn't matter how I'm using it; if this code is in `core`, the user is supposed to be able to use it too, so I don't want to abort in case the ELF file doesn't match; it might as well just be an attempt of opening the file as e.g. 32-bit ELF file. So I should actually change the assertion to a normal throw.

WalterBright:
@kinke Should this be rebased and put back in action?

kinke:
It'd be good to establish the final location first. If there's still doubt this is suited for a regular `core` package, I can also move it to `core.internal.elf.{dl,io}`.

kinke:
Rebased and moved to `core.internal.elf.{dl,io}`.

PetarKirov:
Let's stick to `core.internal.elf.{dl,io}` for now. We could always change the color of the bikeshed if/when we promote the module to a non-interal core package.

kinke:
Ping ding dong, another month has passed by.

CyberShadow:
@kinke This pull request introduced a regression:
https://issues.dlang.org/show_bug.cgi?id=21656

PR #2331, opened 2018-10-16
Merge remote-tracking branch 'upstream/stable' into merge_stable

MartinNowak:


ibuclaw:
@MartinNowak - There's no changes here, is this PR really necessary?

ZombineDev:
@ibuclaw the process is automatic, so I'll merge in case some script relies on this for unknown to us reason.

PR #2332, opened 2018-10-18
Add access to GC runtime profile stats

ZombineDev:
A rebase of #2155

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -139,6 +139,7 @@ private
 
     extern (C) BlkInfo_ gc_query( void* p ) pure nothrow;
     extern (C) GC.Stats gc_stats ( ) nothrow @nogc;
+    extern (C) GC.ProfileStats gc_profileStats ( ) nothrow @nogc @safe;
 
     extern (C) void gc_addRoot( in void* p ) nothrow @nogc;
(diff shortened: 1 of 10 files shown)


src/core/memory.d
```diff
@@ -170,6 +171,23 @@ struct GC
         size_t freeSize;
     }
 
+    /**
+     * Aggregation of current profile information
+     */
+    static struct ProfileStats {
```
jacob-carlborg: Curly braces on their own lines.

thewilsonator:
@ZombineDev ping.

ZombineDev:
@thewilsonator In the short term, I don't think I'll be able to dedicate the time this PR deserves to get through the finish line. Can you adopt it?

thewilsonator:
rebased, we'll see what happens.

rainers:
I've pushed an update to mak/SRC and mak/COPY for missing files. It seems the compiler adds a ModuleInfo reference to a module that is not on the command line.

rainers:
Rebased, squashed and added a micro-test.

BTW: the original PR was already approved by Andrei: https://github.com/dlang/druntime/pull/2155#pullrequestreview-108505370

PR #2333, opened 2018-10-18
Fix crash on attaching external threads

mihails-strasuns:
Part of https://issues.dlang.org/show_bug.cgi?id=19288

Issue comes from combination of two factors:

1) Calling `thread_attachThis` does `new Thread` as part of its
   implementation (because any thread context has to be wrapped into
   Thread class for registration in runtime).
2) Current GC implementation is not prepared to be called from external
   thread context and will crash in several places if `Thread.getThis`
   returns `null`.

There are different possible fixes but it seems that the least intrusive
one is to simply skip collection when called from the context of
unregistered thread. Worst thing it can possibly do for any other code
not affected by the issue is to delay garbage collection a bit.

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -2384,6 +2384,14 @@ struct Gcx
      */
     size_t fullcollect(bool nostack = false) nothrow
     {
+        // It is possible that `fullcollect` will be called from a thread which
+        // is not yet registered in runtime (because allocating `new Thread` is
+        // part of `thread_attachThis` implementation). In that case it is
+        // better not to try actually collecting anything
+
+        if (Thread.getThis() is null)
+            return 0;
+
         MonoTime start, stop, begin;
 
         if (config.profile)
--- a/test/thread/Makefile
+++ b/test/thread/Makefile
@@ -1,6 +1,6 @@
 include ../common.mak
 
-TESTS:=fiber_guard_page
+TESTS:=fiber_guard_page external_threads
 
 .PHONY: all clean
 all: $(addprefix $(ROOT)/,$(addsuffix .done,$(TESTS)))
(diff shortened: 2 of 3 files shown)


mihails-strasuns:
Also question: is there currently any testing infrastructure for mixing C/C++ and D code? I can't find one.

Geod24:
> Also question: is there currently any testing infrastructure for mixing C/C++ and D code? I can't find one.

It's being worked on. See Manu's various PRs, e.g. https://github.com/dlang/druntime/pull/2309

jacob-carlborg:
> Also question: is there currently any testing infrastructure for mixing C/C++ and D code? I can't find one.

Do you need this for this PR? Should be possible to create a new thread with the Posix/Windows APIs to test this.

mihails-strasuns:
It will be a bit different because of presence of `d_main` but should expose similar behaviour. I will give it a try.

On an unrelated note - how does one check report from buildkite?

jacob-carlborg:
> On an unrelated note - how does one check report from buildkite?

For now one needs to have an account and be invited by an admin, like @wilzbach.

jacob-carlborg:
The error is from testing Dub:

```
[INFO] Running /var/lib/buildkite-agent/builds/buildkite-agent-02-1/dlang/druntime/build/dlang-dub/test/interactive-remove.sh...
Package dub not found for registry at https://code.dlang.org/ (fallback ["registry at http://code.dlang.org/", "registry at https://code-mirror.dlang.io/", "registry at https://code-mirror2.dlang.io/", "registry at https://dub-registry.herokuapp.com/"]): No package candidate found for dub 0.9.20
No package dub was found matching the dependency 0.9.20
Fetching dub 0.9.21...
[ERROR] /var/lib/buildkite-agent/builds/buildkite-agent-02-1/dlang/druntime/build/dlang-dub/test/interactive-remove.sh:10 command failed
[ERROR] Script failure.
```

The same error has occurred in other PRs so it's not related to this PR.

ZombineDev:
@mihails-strasuns Try using this account to view the logs in Buildkite:

```
user: dummy@dlang.io
password: dlangrocks
```

mihails-strasuns:
Updated:
- added test case which crashes before this fix
- created dedicated bugzilla sub-issue to refer with "Fixes"

PR #2334, opened 2018-10-18
Fix crash in destructor of detached thread

mihails-strasuns:
Part of https://issues.dlang.org/show_bug.cgi?id=19288

Expectation is that after thread is detached from druntime, it becomes
responsibility of external environment to terminate it properly (for
example, by calling `join` from C/C++ code).

However, `Thread` object for that thread will still remain and will
eventually be collected by GC. Destructor tries to call `pthread_detach`
unconditionally but it is undefined behaviour if thread was alread
joined before (see `man pthread_detach`).

This fix invalidates thread object when it is detached so that
destructor can't conflict with external thread management code.

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -596,7 +596,10 @@ class Thread
      */
     ~this() nothrow @nogc
     {
-        if ( m_addr == m_addr.init )
+        bool no_context = m_addr == m_addr.init;
+        bool not_registered = !next && !prev && (sm_tbeg !is this);
+
+        if (no_context || not_registered)
         {
             return;
         }
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -1859,8 +1859,9 @@ private:
     do
     {
         // Thread was already removed earlier, might happen b/c of thread_detachInstance
-        if (!t.next && !t.prev)
+        if (!t.next && !t.prev && (sm_tbeg !is t))
             return;
+
         slock.lock_nothrow();
         {
             // NOTE: When a thread is removed from the global thread list its
--- a/test/thread/src/external_threads.d
+++ b/test/thread/src/external_threads.d
@@ -1,12 +1,31 @@
 import core.sys.posix.pthread;
 import core.memory;
+import core.thread;
+
+extern (C) void  rt_moduleTlsCtor();
+extern (C) void  rt_moduleTlsDtor();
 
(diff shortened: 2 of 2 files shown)


mihails-strasuns:
Looks like this clashes with internal thread attaching/detaching. Will investigate more.

mihails-strasuns:
Right, so regular `Thread.join` still should work after detaching as it is not only used externally. Chosing between few alternatives, will update PR soon.

Laeeth:
Blocking deployment of analytic tools across the firm.

src/core/thread.d
```diff
@@ -1860,6 +1868,8 @@ private:
             return;
```
nemanja-boric-sociomantic: Seems that this was used previously to see if the thread was detached. You can either replace this
with `if (t.m_detached == false)` or use this in the destructor directly.

mihails-strasuns:
Yup, was waiting for the other PR to get merged. Will update in a moment.

mihails-strasuns:
Updated

mihails-strasuns:
@jacob-carlborg `Fixes https://issues.dlang.org/show_bug.cgi?id=19314` ?

jacob-carlborg:
The commit message needs to start with "Fix issue XXX" where "XXX" is the bugzilla issue. The dlang-bot will post a message in the PR when it has been properly connected to the bugzilla issue.

mihails-strasuns:
Such a choosy bot :X

wilzbach:
> Such a choosy bot :X

This is needed for parsing the git commit messages of all repos for a release.

See also: https://github.com/dlang/dlang-bot#automated-references

mihails-strasuns:
My remark was about how limited are supported formats of mention :) I doubt that supporting few more common patterns would result in lot of false positives.

src/core/thread.d
```diff
@@ -596,7 +596,10 @@ class Thread
      */
     ~this() nothrow @nogc
     {
-        if ( m_addr == m_addr.init )
+        bool no_context = m_addr == m_addr.init;
+        bool not_registered = !next && !prev;
```
n8sh: If there's only one registered thread its `next` and `prev` would also be null, but it appears to me that we don't need to worry about this because it could only happen in the destructor of the main thread.

PR #2337, opened 2018-10-25
hashOf fails to compile for const struct that has non-const toHash & has all fields bitwise-hashable

n8sh:
Example of non-compiling code that this PR fixes:
https://run.dlang.io/is/46J93Z
```d
struct NotOkay
{
    int canBitwiseHashInts;
    size_t toHash() /* not const */ { return 0; }
}

struct Okay
{
    float cantBitwiseHashFloats;
    size_t toHash() /* not const */ { return 0; }
}

void main()
{
    const okay = Okay.init;
    hashOf(okay); // Compiles but writes a message about the non-callable toHash.
    
    const notOkay = NotOkay.init;
    hashOf(notOkay); // Compilation fails!
}
```

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -10,7 +10,7 @@
 module core.internal.hash;
 
 import core.internal.convert;
-import core.internal.traits : allSatisfy;
+import core.internal.traits : allSatisfy, Unconst;
 
 // If true ensure that positive zero and negative zero have the same hash.
 // Historically typeid(float).getHash did this but hashOf(float) did not.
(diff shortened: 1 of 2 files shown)


n8sh:
Added comment:
>TODO: in the future maybe this should be changed to a static assert(0),
>because if there's a `toHash` the programmer probably expected it to be called
>and a compilation failure here will expose a bug in his code.
>  In the future we also might want to disallow non-const toHash altogether.

Also put comments on test cases that in the future support for this behavior might be removed.

wilzbach:
Thanks! Putting the 72h objection label on this in case someone wants to raise his/her concern.

ZombineDev:
vibe-d has consistent runtime failure, not sure if it's directly cause by this PR.

wilzbach:
Unrelated and due to https://github.com/vibe-d/eventcore/pull/86

thewilsonator:
CyberShadow/DAutoTest failure is also unrelated, unfortunately that is required for merging.

thewilsonator:
@n8sh please rebase this

n8sh:
Rebased.

thewilsonator:
@ZombineDev 72h is up, please merge.

PR #2340, opened 2018-10-27
Remove deprecated core.sys.osx package

ibuclaw:
> $(RED Deprecated. Use $(D core.sys.darwin.xxx) instead. This module
>       will be removed in June 2018.)

It is now October 2018...


jacob-carlborg:
What about a changelog entry?

thewilsonator:
I'll add the changelog in another PR

MartinNowak:
That broke building dmd with 2.084.0, see https://buildkite.com/dlang/build-release/builds/117#75f5e6c4-9f15-4a4c-bdcf-36b42c17ad5d/6-12147. Also it would be nice to think about communicating things beforehand instead of forgetting changelog entries afterwards ;).

PR #2341, opened 2018-10-28
Dwarf EH: explicitly remove our reference to caught Throwable

JohanEngelen:
We should explicitly remove our reference to the caught Throwable, such that there are no lingering references to it. 
This enables the code in line 71 to work (where we want to reuse the preallocated ehstorage). Without this patch, only the very first exception thrown is using `ehstorage` and all other exceptions cannot use that struct any more because of the dangling ehstorage.object reference.


s-ludwig:
This was (very likely) fixed in DUB 1.11.0 (included with 2.082.0). Any way to quickly upgrade that version on the autotester? The bandaid fix would be to pin the dependencies in dpl-docs' dub.sdl to the latest versions.

PR #2342, opened 2018-10-31
core.sys.posix.aio: Add OSX bindings

ibuclaw:
Builds are failing on OSX unless these are present.

### DIFF
--- a/src/core/sys/posix/aio.d
+++ b/src/core/sys/posix/aio.d
@@ -63,6 +63,19 @@ version (CRuntime_Glibc)
         }
     }
 }
+else version (OSX)
+{
+    struct aiocb
+    {
+        int aio_filedes;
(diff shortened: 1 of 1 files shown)


ibuclaw:
Apparently sigevent was missing also.

Geod24:
Out of curiosity, what uncovered this ?

src/core/sys/posix/aio.d
```diff
@@ -63,6 +63,19 @@ version (CRuntime_Glibc)
         }
     }
 }
+else version (OSX)
+{
+    struct aiocb
+    {
+        int aio_filedes;
+        off_t aio_offset;
+        void* aio_buf;
```
n8sh: The other branches have comments noting `aio_buf` is volatile. This is `volatile` on OSX too.

PR #2343, opened 2018-10-31
core.sys.posix.aio: Add musl bindings

n8sh:
https://git.musl-libc.org/cgit/musl/tree/include/aio.h

### DIFF
--- a/src/core/sys/posix/aio.d
+++ b/src/core/sys/posix/aio.d
@@ -63,6 +63,27 @@ version (CRuntime_Glibc)
         }
     }
 }
+else version (CRuntime_Musl)
+{
+    // https://git.musl-libc.org/cgit/musl/tree/include/aio.h
+    struct aiocb
+    {
+        int aio_fildes;
+        int aio_lio_opcode;
+        int aio_reqprio;
+        void* aio_buf;   //volatile
+        size_t aio_nbytes;
+        sigevent aio_sigevent;
+        void* __td;
+        int[2] __lock;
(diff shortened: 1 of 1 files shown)


n8sh:
Should I define these aliases unconditionally for CRuntime_Musl? Right now I don't define them at all.
```c
#if defined(_LARGEFILE64_SOURCE) || defined(_GNU_SOURCE)
#define aiocb64 aiocb
#define aio_read64 aio_read
#define aio_write64 aio_write
#define aio_error64 aio_error
#define aio_return64 aio_return
#define aio_cancel64 aio_cancel
#define aio_suspend64 aio_suspend
#define aio_fsync64 aio_fsync
#define lio_listio64 lio_listio
#define off64_t off_t
#endif
```

jacob-carlborg:
> Should I define these aliases unconditionally for CRuntime_Musl? Right now I don't define them at all.

I *think* what's usually done is to have a condition, like a static if.

n8sh:
What should the condition be?

jacob-carlborg:
Same as in the header file:

```d
enum _LARGEFILE64_SOURCE = // whatever is the default value
enum _GNU_SOURCE = // whatever is the default value

static if (_LARGEFILE64_SOURCE || _GNU_SOURCE)
```

thewilsonator:
> Should I define these aliases unconditionally for CRuntime_Musl

In the interest of moving forward, I don't think so. They do not (at least superficially) look to provide much value. They can always be added later if needed.

PR #2344, opened 2018-11-01
Add druntime bindings for Solaris/SPARC

ibuclaw:
Posted and reviewed on gcc-patches, I suspect it's only enough support needed in order to have an initial build, it's a start at least.

### DIFF
--- a/src/core/sys/posix/aio.d
+++ b/src/core/sys/posix/aio.d
@@ -136,6 +136,28 @@ else version (DragonFlyBSD)
 
     version = BSD_Posix;
 }
+else version (Solaris)
+{
+    struct aio_result_t
+    {
+        ssize_t aio_return;
+        int aio_errno;
+    }
+
+    struct aiocb
+    {
+        int aio_fildes;
+        void* aio_buf;   // volatile
+        size_t aio_nbytes;
+        off_t aio_offset;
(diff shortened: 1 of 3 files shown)


src/core/sys/posix/aio.d
```diff
@@ -136,6 +136,32 @@ else version (DragonFlyBSD)
...
 }
+else version (Solaris)
+{
+    struct aio_result_t
+    {
+        ssize_t aio_return;
+        int aio_errno;
+    }
+
+    struct aiocb
```
n8sh: This doesn't match current illumos aiocb ([head/aio.h](https://github.com/illumos/illumos-gate/blob/master/usr/src/head/aio.h), [uts/common/sys/aiocb.h](https://github.com/illumos/illumos-gate/blob/master/usr/src/uts/common/sys/aiocb.h)), which might very well be correct but makes me at least pause to question it. @ibuclaw [when you reviewed this](https://gcc.gnu.org/ml/gcc-patches/2018-10/msg02066.html) did you check Solaris 11 aio.h yourself to verify aiocb was correct?

ibuclaw: I can only assume they did the absolute minimum just to get build passing.

I've corrected it as per headers.

PR #2346, opened 2018-11-09
std::string_view windows tests

TurkeyMan:


TurkeyMan:
Our windows machines don't run VS2017...

TurkeyMan:
Huzzah!

thewilsonator:
You will need to make it not run on old VS installations. I'd probably split the tests into C++17/non-C++17 (or even 98/11/17).

PR #2347, opened 2018-11-09
std::string_view posix tests

TurkeyMan:


TurkeyMan:
I don't run these systems... if people can help me get these across the line, that'd be real great!

jacob-carlborg:
> I don't run these systems... if people can help me get these across the line, that'd be real great!

You need to start 😉.

thewilsonator:
I remember something similar to that output not too long ago, can I suggest you rebase? If nothing else it will rule that out.

jacob-carlborg:
> I would see Apple burn before I infect any piece of hardware in my presence! ;)

The PR title says "posix". There are other Posix systems than those created by Apple 😉.

TurkeyMan:
> > I would see Apple burn before I infect any piece of hardware in my presence! ;)
> 
> The PR title says "posix". There are other Posix systems than those created by Apple 😉.

True, but Apple has caused the the most trouble, with having a slightly weird ecosystem; gotta explicitly request to link the C++ runtime (even when supplying `-std=`, that whole `std::__1` thing, 32bit darwin not working, etc.

> Beside the smiley I added above, it would gain you a lot to test on non Windows platforms locally. To avoid having to wait for the CI systems to pass or for someone else to fix your PR.

I mean, I could just stop doing this altogether... I'm not getting anything at all for my time from this at this point! My stuff already works fine.

atilaneves:
Tests are broken due to `-std=c++17` - whatever gcc version is on the autotester, it's too old. Unfortunately, and unlike Travis, it's impossible AFAIK to install anything on the autotester machines.

wilzbach:
> Tests are broken due to `-std=c++17` - whatever gcc version is on the autotester, it's too old. Unfortunately, and unlike Travis, it's impossible AFAIK to install anything on the autotester machines.

You could just not run the test for older compilers (or on the auto-tester). Since a few days, buildkite runs the full druntime test suite too (see e.g. https://github.com/dlang/ci/pull/301)

PR #2348, opened 2018-11-11
[WIP] Add @nogc array

edi33416:
This PR adds a `@nogc nothrow pure @safe` implementation of a generic array that mimics the behaviour of the built-in array (`T[]`).

This array implementation makes use of the Copy Constructor [PR](https://github.com/dlang/dmd/pull/8688). By doing so, it both tests the implementation and proves it's usefulness.

I know this is a lengthy PR, but please bear with me, as there are a lot of unittests.
And now, as Andrei says "destroy()"

### DIFF
--- a/src/core/experimental/array.d
+++ b/src/core/experimental/array.d
@@ -1125,8 +1125,7 @@ struct rcarray(T)
      */
     immutable(rcarray!T) idup(this Q)()
     {
-        auto r = immutable rcarray!T(this);
-        return r;
+        return immutable rcarray!T(this);
     }
 
     ///
--- a/src/core/experimental/array.d
+++ b/src/core/experimental/array.d
@@ -682,10 +682,12 @@ version(unittest)
 ///
 struct Array(T)
 {
+private:
+
     import core.atomic : atomicOp;
 
-    package T[] _payload;
-    package Unqual!T[] _support;
+    T[] _payload;
+    Unqual!T[] _support;
 
     version(unittest)
     {
@@ -696,11 +698,8 @@ struct Array(T)
         alias _sallocator = shared PrefixAllocator.instance;
     }
 
-private:
-
     static enum double capacityFactor = 3.0 / 2;
     static enum initCapacity = 3;
-
     bool _isShared;
 
     @trusted
@@ -809,10 +808,7 @@ private:
         size_t i = 0;
         foreach (ref item; } ~ stuff ~ q{)
         {
-            //writefln("the type is %s %s %s", T.stringof, typeof(_support).stringof, typeof(_payload).stringof);
             alias TT = ElementType!(typeof(_payload));
-            //pragma(msg, typeof(item).stringof, " TT is ", TT.stringof);
-
             size_t s = i * stateSize!TT;
             size_t e = (i + 1) * stateSize!TT;
             void[] tmp = tmpSupport[s .. e];
@@ -916,7 +912,7 @@ public:
     // Begin Copy Ctors
     // {
 
-    private static enum copyCtorIncRef = q{
+    private enum copyCtorIncRef = q{
         _payload = rhs._payload;
         _support = rhs._support;
         _isShared = rhs._isShared;
@@ -1064,19 +1060,24 @@ public:
     }
 
     /**
-     * Set the length of the array to `len`. `len` must be less than or equal
-     * to the `capacity` of the array.
+     * Set the length of the array to `len`. If `len` exceeds the available
+     * `capacity` of the array, an attempt to extend the array in place is made.
+     * If extending is not possible, a reallocation will occur; if the new
+     * length of the array is longer than `len`, the remainder will be default
+     * initialized.
      *
      * Params:
      *      len = a positive integer
      *
-     * Complexity: $(BIGOH 1).
+     * Complexity: $(BIGOH n).
      */
-    @nogc nothrow pure @trusted
-    void forceLength(size_t len)
+    void length(size_t len)
     {
-        assert(len <= capacity);
-        _payload = cast(T[])(_support[slackFront .. len]);
+        if (capacity < len)
+        {
+            reserve(len);
+        }
+        _payload = (() @trusted => cast(T[])(_support[slackFront .. len]))();
     }
 
     ///
@@ -1084,8 +1085,16 @@ public:
     @safe unittest
     {
         auto a = Array!int(1, 2, 3);
-        a.forceLength(2);
+        a.length = 2;
         assert(a.length == 2);
+
+        auto b = a;
+        assert(a.capacity < 10);
+        a.length = 10; // will trigger a reallocation
+        assert(a.length == 10);
+        assert(b.length == 2);
+        a[0] = 20;
+        assert(a[0] != b[0]);
     }
 
     /**
@@ -1178,7 +1187,6 @@ public:
             if (i < _payload.length)
             {
                 emplace(&tmpSupport[i], _payload[i]);
-                //pragma(msg, typeof(&tmpSupport[i]).stringof);
             }
             else
             {
--- a/src/core/experimental/array.d
+++ b/src/core/experimental/array.d
@@ -278,7 +278,7 @@ private struct PrefixAllocator
     enum uint alignment = size_t.alignof;
     static enum prefixSize = size_t.sizeof;
 
-    version (unittest)
+    version (CoreUnittest)
     {
         // During unittesting, we are keeping a count of the number of bytes allocated
         size_t bytesUsed;
@@ -296,7 +296,7 @@ private struct PrefixAllocator
         // Init reference count to 0
         *(cast(size_t *) p) = 0;
 
-        version (unittest)
+        version (CoreUnittest)
         {
             static if (is(typeof(this) == shared))
             {
@@ -318,7 +318,7 @@ private struct PrefixAllocator
         import core.memory : pureFree;
         assert(b !is null);
 
-        version (unittest)
+        version (CoreUnittest)
         {
             static if (is(typeof(this) == shared))
             {
@@ -366,6 +366,7 @@ private struct PrefixAllocator
     static shared PrefixAllocator instance;
 }
 
+version (CoreUnittest)
 @safe pure nothrow @nogc unittest
 {
     shared PrefixAllocator a;
@@ -382,7 +383,7 @@ private struct PrefixAllocator
 }
 
 
-version (unittest)
+version (CoreUnittest)
 {
     private alias SCAlloc = shared PrefixAllocator;
     private alias SSCAlloc = shared PrefixAllocator;
@@ -438,9 +439,7 @@ struct rcarray(T)
     private T[] payload;
     private Unqual!T[] support;
 
-    version (unittest)
-    {
-    }
+    version (CoreUnittest) { }
     else
     {
         alias localAllocator = shared PrefixAllocator.instance;
@@ -508,7 +507,7 @@ struct rcarray(T)
         if (opPrefix!("-=")(_support, 1) == 0)
         {
             () @trusted {
-                version (unittest)
+                version (CoreUnittest)
                 {
                     pureDispose(isShared, _support);
                 }
@@ -527,7 +526,7 @@ struct rcarray(T)
 
         return stuffLengthStr ~ q{
 
-        version (unittest)
+        version (CoreUnittest)
         {
             void[] tmpSupport = (() @trusted => pureAllocate(isShared, stuffLength * stateSize!T))();
         }
@@ -853,7 +852,7 @@ struct rcarray(T)
         if (support && opCmpPrefix!"=="(support, 1))
         {
             void[] buf = support;
-            version (unittest)
+            version (CoreUnittest)
             {
                 auto successfulExpand = pureExpand(isShared, buf, (n - capacity) * stateSize!T);
             }
@@ -880,7 +879,7 @@ struct rcarray(T)
             }
         }
 
-        version (unittest)
+        version (CoreUnittest)
         {
             auto tmpSupport = (() @trusted  => cast(Unqual!T[])(pureAllocate(isShared, n * stateSize!T)))();
         }
@@ -1694,7 +1693,10 @@ struct rcarray(T)
     }
 }
 
-version (unittest) private nothrow pure @safe
+version (CoreUnittest)
+{ // Begin CoreUnittest - conditional compilation so we can check for mem leaks
+
+private nothrow pure @safe
 void testConcatAndAppend()
 {
     auto a = rcarray!(int)(1, 2, 3);
@@ -1748,7 +1750,7 @@ void testConcatAndAppend()
     assert(sharedAllocator.bytesUsed == 0, "Array ref count leaks memory");
 }
 
-version (unittest) private nothrow pure @safe
+private nothrow pure @safe
 void testSimple()
 {
     auto a = rcarray!int();
@@ -1794,7 +1796,7 @@ void testSimple()
     assert(sharedAllocator.bytesUsed == 0, "Array ref count leaks memory");
 }
 
-version (unittest) private nothrow pure @safe
+private nothrow pure @safe
 void testSimpleImmutable()
 {
     auto a = rcarray!(immutable int)();
@@ -1833,7 +1835,7 @@ void testSimpleImmutable()
     assert(sharedAllocator.bytesUsed == 0, "Array ref count leaks memory");
 }
 
-version (unittest) private nothrow pure @safe
+private nothrow pure @safe
 void testCopyAndRef()
 {
     auto aFromList = rcarray!int(1, 2, 3);
@@ -1874,7 +1876,7 @@ void testCopyAndRef()
     assert(sharedAllocator.bytesUsed == 0, "Array ref count leaks memory");
 }
 
-version (unittest) private nothrow pure @safe
+private nothrow pure @safe
 void testImmutability()
 {
     auto a = immutable rcarray!(int)(1, 2, 3);
@@ -1921,7 +1923,7 @@ void testImmutability()
     }
 }
 
-version (unittest) private nothrow pure @safe
+private nothrow pure @safe
 void testConstness()
 {
     auto a = const rcarray!(int)(1, 2, 3);
@@ -1950,7 +1952,7 @@ void testConstness()
     assert(sharedAllocator.bytesUsed == 0, "Array ref count leaks memory");
 }
 
-version (unittest) private nothrow pure @safe
+private nothrow pure @safe
 void testWithStruct()
 {
     auto array = rcarray!int(1, 2, 3);
@@ -1990,7 +1992,7 @@ void testWithStruct()
     assert(sharedAllocator.bytesUsed == 0, "Array ref count leaks memory");
 }
 
-version (unittest) private nothrow pure @safe
+private nothrow pure @safe
 void testWithClass()
 {
     class MyClass
@@ -2022,7 +2024,7 @@ void testWithClass()
     assert(sharedAllocator.bytesUsed == 0, "Array ref count leaks memory");
 }
 
-version (unittest) private @nogc nothrow pure @safe
+private @nogc nothrow pure @safe
 void testOpOverloads()
 {
     auto a = rcarray!int(1, 2, 3, 4);
@@ -2062,7 +2064,7 @@ void testOpOverloads()
     assert(sharedAllocator.bytesUsed == 0, "Array ref count leaks memory");
 }
 
-version (unittest) private nothrow pure @safe
+private nothrow pure @safe
 void testSlice()
 {
     auto a = rcarray!int(1, 2, 3, 4);
@@ -2093,3 +2095,5 @@ void testSlice()
     assert(localAllocator.bytesUsed == 0, "Array ref count leaks memory");
     assert(sharedAllocator.bytesUsed == 0, "Array ref count leaks memory");
 }
+
+} // End CoreUnittest
--- a/src/core/experimental/array.d
+++ b/src/core/experimental/array.d
@@ -1263,11 +1263,6 @@ struct rcarray(T)
      */
     Qualified opSlice(this Qualified)(size_t start, size_t end)
     {
-        version (D_NoBoundsChecks) { } else
-        {
-            assert(start <= end && end <= length,
-                   "Array.opSlice(s, e): Invalid bounds: Ensure start <= end <= length");
-        }
         return typeof(this)(support, payload[start .. end], isShared);
     }
 
@@ -1295,10 +1290,6 @@ struct rcarray(T)
      */
     ref auto opIndex(this _)(size_t idx)
     {
-        version (D_NoBoundsChecks) { } else
-        {
-            assert(idx < length, "Array.opIndex: Index out of bounds");
-        }
         return payload[idx];
     }
 
@@ -1324,10 +1315,6 @@ struct rcarray(T)
      */
     ref auto opIndexUnary(string op)(size_t idx)
     {
-        version (D_NoBoundsChecks) { } else
-        {
-            assert(idx < length, "Array.opIndexUnary!" ~ op ~ ": Index out of bounds");
-        }
         mixin("return " ~ op ~ "payload[idx];");
     }
 
@@ -1357,10 +1344,6 @@ struct rcarray(T)
     ref auto opIndexAssign(U)(U elem, size_t idx)
     if (is(U : T))
     {
-        version (D_NoBoundsChecks) { } else
-        {
-            assert(idx < length, "Array.opIndexAssign: Index out of bounds");
-        }
         return payload[idx] = elem;
     }
 
@@ -1419,11 +1402,6 @@ struct rcarray(T)
     auto opSliceAssign(U)(U elem, size_t start, size_t end)
     if (is(U : T))
     {
-        version (D_NoBoundsChecks) { } else
-        {
-            assert(start <= end, "Array.opSliceAssign: Index out of bounds");
-            assert(end < length, "Array.opSliceAssign: Index out of bounds");
-        }
         return payload[start .. end] = elem;
     }
 
@@ -1453,10 +1431,6 @@ struct rcarray(T)
     ref auto opIndexOpAssign(string op, U)(U elem, size_t idx)
     if (is(U : T))
     {
-        version (D_NoBoundsChecks) { } else
-        {
-            assert(idx < length, "Array.opIndexOpAssign!" ~ op ~ ": Index out of bounds");
-        }
         mixin("return payload[idx]" ~ op ~ "= elem;");
     }
 
--- a/src/core/experimental/array.d
+++ b/src/core/experimental/array.d
@@ -431,7 +431,7 @@ version (unittest)
 }
 
 ///
-struct Array(T)
+struct rcarray(T)
 {
     import core.atomic : atomicOp;
 
@@ -600,18 +600,18 @@ struct Array(T)
     {
         // Create a list from a list of ints
         {
-            auto a = Array!int(1, 2, 3);
+            auto a = rcarray!int(1, 2, 3);
             assert(a == [1, 2, 3]);
         }
         // Create a list from an array of ints
         {
-            auto a = Array!int([1, 2, 3]);
+            auto a = rcarray!int([1, 2, 3]);
             assert(a == [1, 2, 3]);
         }
         // Create a list from a list from an input range
         {
-            auto a = Array!int(1, 2, 3);
-            auto a2 = Array!int(a);
+            auto a = rcarray!int(1, 2, 3);
+            auto a2 = rcarray!int(a);
             assert(a2 == [1, 2, 3]);
         }
     }
@@ -698,34 +698,34 @@ struct Array(T)
     static if (is(T == int))
     @nogc nothrow pure @safe unittest
     {
-        auto a = Array!int(1, 2, 3);
+        auto a = rcarray!int(1, 2, 3);
 
         // Infer safety
-        static assert(!__traits(compiles, () @safe { Array!Unsafe(Unsafe(1)); }));
-        static assert(!__traits(compiles, () @safe { auto a = const Array!Unsafe(Unsafe(1)); }));
-        static assert(!__traits(compiles, () @safe { auto a = immutable Array!Unsafe(Unsafe(1)); }));
+        static assert(!__traits(compiles, () @safe { rcarray!Unsafe(Unsafe(1)); }));
+        static assert(!__traits(compiles, () @safe { auto a = const rcarray!Unsafe(Unsafe(1)); }));
+        static assert(!__traits(compiles, () @safe { auto a = immutable rcarray!Unsafe(Unsafe(1)); }));
 
-        static assert(!__traits(compiles, () @safe { Array!UnsafeDtor(UnsafeDtor(1)); }));
-        static assert(!__traits(compiles, () @safe { auto s = const Array!UnsafeDtor(UnsafeDtor(1)); }));
-        static assert(!__traits(compiles, () @safe { auto s = immutable Array!UnsafeDtor(UnsafeDtor(1)); }));
+        static assert(!__traits(compiles, () @safe { rcarray!UnsafeDtor(UnsafeDtor(1)); }));
+        static assert(!__traits(compiles, () @safe { auto s = const rcarray!UnsafeDtor(UnsafeDtor(1)); }));
+        static assert(!__traits(compiles, () @safe { auto s = immutable rcarray!UnsafeDtor(UnsafeDtor(1)); }));
 
         // Infer purity
-        static assert(!__traits(compiles, () pure { Array!Impure(Impure(1)); }));
-        static assert(!__traits(compiles, () pure { auto a = const Array!Impure(Impure(1)); }));
-        static assert(!__traits(compiles, () pure { auto a = immutable Array!Impure(Impure(1)); }));
+        static assert(!__traits(compiles, () pure { rcarray!Impure(Impure(1)); }));
+        static assert(!__traits(compiles, () pure { auto a = const rcarray!Impure(Impure(1)); }));
+        static assert(!__traits(compiles, () pure { auto a = immutable rcarray!Impure(Impure(1)); }));
 
-        static assert(!__traits(compiles, () pure { Array!ImpureDtor(ImpureDtor(1)); }));
-        static assert(!__traits(compiles, () pure { auto s = const Array!ImpureDtor(ImpureDtor(1)); }));
-        static assert(!__traits(compiles, () pure { auto s = immutable Array!ImpureDtor(ImpureDtor(1)); }));
+        static assert(!__traits(compiles, () pure { rcarray!ImpureDtor(ImpureDtor(1)); }));
+        static assert(!__traits(compiles, () pure { auto s = const rcarray!ImpureDtor(ImpureDtor(1)); }));
+        static assert(!__traits(compiles, () pure { auto s = immutable rcarray!ImpureDtor(ImpureDtor(1)); }));
 
         // Infer throwability
-        static assert(!__traits(compiles, () nothrow { Array!Throws(Throws(1)); }));
-        static assert(!__traits(compiles, () nothrow { auto a = const Array!Throws(Throws(1)); }));
-        static assert(!__traits(compiles, () nothrow { auto a = immutable Array!Throws(Throws(1)); }));
+        static assert(!__traits(compiles, () nothrow { rcarray!Throws(Throws(1)); }));
+        static assert(!__traits(compiles, () nothrow { auto a = const rcarray!Throws(Throws(1)); }));
+        static assert(!__traits(compiles, () nothrow { auto a = immutable rcarray!Throws(Throws(1)); }));
 
-        static assert(!__traits(compiles, () nothrow { Array!ThrowsDtor(ThrowsDtor(1)); }));
-        static assert(!__traits(compiles, () nothrow { auto s = const Array!ThrowsDtor(ThrowsDtor(1)); }));
-        static assert(!__traits(compiles, () nothrow { auto s = immutable Array!ThrowsDtor(ThrowsDtor(1)); }));
+        static assert(!__traits(compiles, () nothrow { rcarray!ThrowsDtor(ThrowsDtor(1)); }));
+        static assert(!__traits(compiles, () nothrow { auto s = const rcarray!ThrowsDtor(ThrowsDtor(1)); }));
+        static assert(!__traits(compiles, () nothrow { auto s = immutable rcarray!ThrowsDtor(ThrowsDtor(1)); }));
     }
 
     private @nogc nothrow pure @trusted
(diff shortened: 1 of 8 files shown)


src/core/experimental/array.d
```diff
@@ -0,0 +1,2559 @@
...
+    static if (is(typeof(T.outer)))
+        enum isInnerClass = __traits(isSame, typeof(T.outer), __traits(parent, T));
+    else
+        enum isInnerClass = false;
+}
+
+enum classInstanceAlignment(T) = size_t.alignof >= T.alignof ? size_t.alignof : T.alignof;
+
+T emplace(T, Args...)(T chunk, auto ref Args args)
+if (is(T == class))
```
jacob-carlborg: Needs to handle non D classes.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2559 @@
...
+    {
+        destroy(*p);
+    }
+    alloc.deallocate((cast(void*) p)[0 .. T.sizeof]);
+    static if (__traits(isRef, p))
+        p = null;
+}
+
+void dispose(A, T)(auto ref A alloc, auto ref T p)
+if (is(T == class) || is(T == interface))
```
jacob-carlborg: Needs to handle non D classes and interfaces.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2559 @@
...
+     * Set the length of the array to `len`. `len` must be less than or equal
+     * to the `capacity` of the array.
+     *
+     * Params:
+     *      len = a positive integer
+     *
+     * Complexity: $(BIGOH 1).
+     */
+    @nogc nothrow pure @trusted
+    void forceLength(size_t len)
```
jacob-carlborg: Why isn’t this called `length`?

src/core/experimental/array.d
```diff
@@ -0,0 +1,2559 @@
...
+        else
+        {
+            insert(0, stuff);
+        }
+    }
+
+    // Begin Copy Ctors
+    // {
+
+    private static enum copyCtorIncRef = q{
```
jacob-carlborg: `static` is redundant.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2559 @@
...
+    package T[] _payload;
+    package Unqual!T[] _support;
+
+    version(unittest)
+    {
+    }
+    else
+    {
+        alias _allocator = shared PrefixAllocator.instance;
+        alias _sallocator = shared PrefixAllocator.instance;
```
jacob-carlborg: Should these be public?

src/core/experimental/array.d
```diff
@@ -0,0 +1,2559 @@
...
+    /**
+     * Provide access to the first element in the array. The user must check
+     * that the array isn't `empty`, prior to calling this function.
+     *
+     * Returns:
+     *      a reference to the first element.
+     *
+     * Complexity: $(BIGOH 1).
+     */
+    ref auto front(this _)()
```
jacob-carlborg: Why the this template parameter?

src/core/experimental/array.d
```diff
@@ -0,0 +1,2559 @@
...
+     *
+     * Params:
+     *      idx = a positive integer
+     *
+     * Complexity: $(BIGOH 1).
+     */
+    ref auto opIndex(this _)(size_t idx)
+    in
+    {
+        assert(idx < length, "Array.opIndex: Index out of bounds");
```
jacob-carlborg: This does not behave the same as a built in array.

thewilsonator: This should also respect `version(D_NoBoundsCheck)`.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2559 @@
...
+    assert(equal(aInsFromRange, [2, 3]));
+
+    Array!int aInsBackFromRange = Array!int();
+    aInsBackFromRange.insert(pos, aFromList);
+    aFromList.popFront();
+    assert(aFromList.empty);
+    assert(equal(aInsBackFromRange, [3]));
+
+    auto aFromRef = aInsFromRange;
+    auto aFromDup = aInsFromRange.du
```
jacob-carlborg: Should have tests with non D classes as well.

jacob-carlborg:
* There’s a lot of code related to traits, do we want that in the same module?
* Why are some range functions included and why are they in the Array struct?

src/core/experimental/array.d
edi33416: Got it. What non D classes are supported? C++ ones?

src/core/experimental/array.d
edi33416: Same question: can you please provide me with an example?

src/core/experimental/array.d
edi33416: It was't doing exactly as built-in array `length`. I'll update the PR with the updated name and implementation.

src/core/experimental/array.d
edi33416: > Should these be public?

No, they shouldn't. Good catch.

> What's the difference between these two?

Currently, none. They are there in case we decide to use a different allocator for the the `shared` types, so we can just drop the replacement there. I will remove `_sallocator` if we'll not pursue this idea.

src/core/experimental/array.d
edi33416: So all the `Array!T` types, regardless of the qualifiers can call front. It can't be `ref auto front() const` because mutable arrays must be able to change the element using front, like `a.front = 42;`

src/core/experimental/array.d
jacob-carlborg: If you keep it, please rename it to “sharedAllocator”, that would have saved me this question.

src/core/experimental/array.d
jacob-carlborg: Replied to the other comment.

src/core/experimental/array.d
jacob-carlborg: So it’s just for attribute inference? In that case you can use an empty template parameter list instead. Perhaps add a comment as well.

src/core/experimental/array.d
jacob-carlborg: > Will change. According to the docs, this should throw a RangeError. Am I right?

Yes, but it depends on which compiler setting is used. Have a look at the `boundscheck` flag https://dlang.org/dmd-osx.html#switch-boundscheck

edi33416:
>     * There’s a lot of code related to traits, do we want that in the same module?

There is, and I want to remove as much as I can. I didn't know about `core.internal.traits` until like 10mins ago.
Regarding `core.internal.traits`, I see there is a duplicate of the Phobos `Unqual` there. I'll open a PR in Phobos that will alias to `core.internal.traits.Unqual`, so we won't have to maintain the same code twice.

>     * Why are some range functions included and why are they in the Array struct?

I needed those for some of the template constraints and other in order to consume `T[]`

src/core/experimental/array.d
```diff
@@ -0,0 +1,2559 @@
...
+    }
+}
+
+///
+struct Array(T)
+{
+    import core.atomic : atomicOp;
+
+    package T[] _payload;
+    package Unqual!T[] _support;
```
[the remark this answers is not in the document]
jacob-carlborg: Some developers have that habit, especially C++ developers. I recommend against it. It should only be used if it would otherwise conflict with a keyword or a method, but the. It should be a trailing underscore.

jacob-carlborg:
> I needed those for some of the template constraints and other in order to consume T[]

If they’re only for implementation they should be private.

src/core/experimental/array.d
andralex: @jacob-carlborg that was copied from the phobos implementation (later to be moved), can you explain what needs to be done here? Thanks! Also: would rather add that as an additional PR after this.

andralex:
TODO: rename module to rcslice.d and type name with rcslice.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+
+// Range functions
+
+/**
+Implements the range interface primitive `front` for built-in arrays. Due to the
+fact that nonmember functions can be called with the first argument using the dot
+notation, `array.front` is equivalent to `front(array)`. For $(GLOSSARY narrow strings),
+`front` automatically returns the first $(GLOSSARY code point) as _a `dchar`.
+*/
+@property ref T front(T)(T[] a) @safe pure nothrow @nogc
```
andralex: remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+{
+    assert(a.length, "Attempting to fetch the front of an empty array of " ~ T.stringof);
+    return a[0];
+}
+
+/**
+Implements the range interface primitive `empty` for types that obey `hasLength`
+property and for narrow strings.
+ */
+@property bool empty(T)(auto ref scope const(T) a)
```
andralex: remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+{
+    return !a.length;
+}
+
+/**
+Implements the range interface primitive `popFront` for built-in arrays. For
+$(GLOSSARY narrow strings), `popFront` automatically advances to the next
+$(GLOSSARY code point).
+*/
+void popFront(T)(ref T[] a) @safe pure nothrow @nogc
```
andralex: remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+if (!is(T[] == void[]))
+{
+    assert(a.length, "Attempting to popFront() past the end of an array of " ~ T.stringof);
+    a = a[1 .. $];
+}
+
+/**
+This is a best-effort implementation of `length` for any kind of range.
+*/
+auto walkLength(Range)(Range range)
```
andralex: remove?

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+        size_t result;
+        for ( ; !range.empty ; range.popFront() )
+            ++result;
+        return result;
+    }
+}
+
+/**
+*/
+enum bool isInputRange(R) =
```
andralex: remove (no range operations at this level)

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+    is(typeof(R.init) == R)
+    && is(typeof(R.init.empty()) == bool)
+    && is(typeof((return ref R r) => r.front))
+    && !is(typeof(R.init.front()) == void)
+    && is(typeof((R r) => r.popFront));
+
+/**
+*/
+template isInfinite(R)
```
andralex: remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+    else
+        enum bool isInfinite = false;
+}
+
+// End Range functions
+
+/**
+Yields `true` if and only if `T` is an aggregate that defines a symbol called `name`.
+ */
+enum hasMember(T, string name) = __traits(hasMember, T, name);
```
andralex: remove?

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+
+/**
+Detect whether type `T` is an array (static or dynamic.
+*/
+enum bool isArray(T) = isStaticArray!T || isDynamicArray!T;
```
andralex: probably remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+
+/**
+ * Detect whether type `T` is a static array.
+ */
+enum bool isStaticArray(T) = __traits(isStaticArray, T);
```
andralex: probably remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+
+/**
+ * Detect whether type `T` is a dynamic array.
+ */
+enum bool isDynamicArray(T) = is(DynamicArrayTypeOf!T) && !isAggregateType!T;
```
andralex: probably remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+
+/*
+ */
+template DynamicArrayTypeOf(T)
```
andralex: probably remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+    static if (is(Unqual!X : E[], E) && !is(typeof({ enum n = X.length; })))
+    {
+        alias DynamicArrayTypeOf = X;
+    }
+    else
+        static assert(0, T.stringof~" is not a dynamic array");
+}
+
+// SomethingTypeOf
+private template AliasThisTypeOf(T)
```
andralex: probably remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+
+    static if (members.length == 1)
+    {
+        alias AliasThisTypeOf = typeof(__traits(getMember, T.init, members[0]));
+    }
+    else
+        static assert(0, T.stringof~" does not have alias this type");
+}
+
+template AliasSeq(TList...)
```
andralex: probably remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+{
+    alias AliasSeq = TList;
+}
+
+/**
+ * Strips off all `enum`s from type `T`.
+ */
+template OriginalType(T)
```
andralex: probably remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+    {
+        static if (is(T U == enum)) alias Impl = OriginalType!U;
+        else                        alias Impl =              T;
+    }
+
+    alias OriginalType = ModifyTypePreservingTQ!(Impl, T);
+}
+
+// [For internal use]
+private template ModifyTypePreservingTQ(alias Modifier, T)
```
andralex: probably remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+    else static if (is(T U ==        inout const U)) alias ModifyTypePreservingTQ =        inout const Modifier!U;
+    else static if (is(T U ==        inout       U)) alias ModifyTypePreservingTQ =              inout Modifier!U;
+    else static if (is(T U ==              const U)) alias ModifyTypePreservingTQ =              const Modifier!U;
+    else                                             alias ModifyTypePreservingTQ =                    Modifier!T;
+}
+
+/**
+ * Detect whether type `T` is an aggregate type.
+ */
+enum bool isAggregateType(T) = is(T == struct) || is(T == union) ||
```
andralex: probably remove

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+                               is(T == class) || is(T == interface);
+
+/**
+Is `From` implicitly convertible to `To`?
+ */
+enum bool isImplicitlyConvertible(From, To) = is(From : To);
```
andralex: maybe doesn't even deserve its own abstraction

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+
+/**
+The element type of `R`. `R` does not have to be a range. The element type is
+determined as the type yielded by `r.front` for an object `r` of type `R`. For
+example, `ElementType!(T[])` is `T` if `T[]` isn't a narrow string; if it is, the
+element type is `dchar`. If `R` doesn't have `front`, `ElementType!R` is `void`.
+ */
+template ElementType(R)
+{
+    static if (is(typeof(R.init.front.init) T))
```
andralex: `static if (is(typeof((*R.init.ptr).init) T))`

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+    else
+        alias ElementType = void;
+}
+
+/**
+Yields `true` if `R` has a `length` member that returns a value of `size_t`
+type. `R` does not have to be a range. If `R` is a range, algorithms in the
+standard library are only guaranteed to support `length` with type `size_t`.
+*/
+template hasLength(R)
```
andralex: remove

src/core/experimental/array.d
jacob-carlborg: TypeInfo is somewhat broken for non D classes. But `emplace` might not be affected.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2542 @@
...
+    if (!p) return;
+    static if (is(T == interface))
+    {
+        version (Windows)
+        {
+            import core.sys.windows.unknwn : IUnknown;
+            static assert(!is(T: IUnknown), "COM interfaces can't be destroyed in "
+                ~ __PRETTY_FUNCTION__);
+        }
+        auto ob = cast(Object) p;
```
jacob-carlborg: This is certainly broken for non D interface since they a non D class doesn't need to inherit from `Object.

edi33416:
Just to bump question

> TODO: rename module to rcslice.d and type name with rcslice.

> All: would rcarray be preferable to rcslice?

thewilsonator:
>would rcarray be preferable to rcslice?

I think so, slice has connotations of being unowned, whereas array is more container like.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+{
+    static if (is(T == class) || is(T == interface))
+        enum stateSize = __traits(classInstanceSize, T);
+    else static if (is(T == void))
+        enum size_t stateSize = 0;
+    else
+        enum stateSize = T.sizeof;
+}
+
+template isAbstractClass(T...)
```
wilzbach: private

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+if (T.length == 1)
+{
+    enum bool isAbstractClass = __traits(isAbstractClass, T[0]);
+}
+
+template isInnerClass(T)
```
wilzbach: private

src/core/experimental/array.d
wilzbach: Should be private then.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+
+private @nogc pure nothrow @safe
+void testEmplaceChunk(void[] chunk, size_t typeSize, size_t typeAlignment)
+{
+    assert(chunk.length >= typeSize, "emplace: Chunk size too small.");
+    assert((cast(size_t) chunk.ptr) % typeAlignment == 0, "emplace: Chunk is not aligned.");
+}
+
+enum hasElaborateDestructor(T) = __traits(compiles, { T t; t.__dtor(); })
+                                 ||  __traits(compiles, { T t; t.__xdtor(); });
```
wilzbach: private

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+template ElementType(R)
+{
+    static if (is(typeof(R.init[0].init) T))
+        alias ElementType = T;
+    else
+        alias ElementType = void;
+}
+
+auto tail(Collection)(Collection collection)
+if (is(typeof(collection.popFront())))
```
wilzbach: This was rejected for Phobos:

https://github.com/dlang/phobos/pull/4010

I would still love to see it happen though.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+{
+    collection.popFront();
+    return collection;
+}
+
+auto equal(T, U)(T a, U b)
```
wilzbach: Shouldn't this be `opEquals`

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+    Unqual!T[] support;
+
+    version(unittest)
+    {
+    }
+    else
+    {
+        alias localAllocator = shared PrefixAllocator.instance;
+        alias sharedAllocator = shared PrefixAllocator.instance;
+    }
```
wilzbach: Uhm this looks very dangerous.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+
+    @nogc nothrow
+    void _dispose(T)(bool isShared, T[] b)
+    {
+        return isShared ?  sharedAllocator.dispose(b) : localAllocator.dispose(b);
+    }
+}
+
+///
```
wilzbach: Needs documentation

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+        {
+            () @trusted {
+                version(unittest)
+                {
+                    pureDispose(isShared, _support);
+                }
+                else
+                {
+                    localAllocator.dispose(_support);
+                }
```
wilzbach: What's the motivation for this hack?

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+    {
+        auto stuffLengthStr = q{
+            size_t stuffLength = } ~ stuff ~ ".length;";
+
+        return stuffLengthStr ~ q{
+
+        version(unittest)
+        {
+            void[] tmpSupport = (() @trusted => pureAllocate(isShared, stuffLength * stateSize!T))();
+        }
```
wilzbach: No unittest-versioning in the production code.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+        if (n <= capacity) { return; }
+
+        static if (__traits(hasMember, typeof(localAllocator), "expand"))
+        if (support && opCmpPrefix!"=="(support, 1))
+        {
+            void[] buf = support;
+            version(unittest)
+            {
+                auto successfulExpand = pureExpand(isShared, buf, (n - capacity) * stateSize!T);
+            }
```
wilzbach: No versioining in production code.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+        {
+            result.reserve(length);
+            foreach(i; 0 .. length)
+            {
+                result ~= this[i];
+            }
+        }
+        else
+        {
+            result.insert(0, this);
```
wilzbach: Need to reserve here too. Why not call the constructor directly here?

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+        assert(arr3 > arr1);
+        assert(arr4 < arr1);
+        assert(arr4 < arr3);
+        assert(arr3 > arr4);
+    }
+
+    ///
+    auto toHash()
+    {
+        // will be safe with 2.082
```
wilzbach: It's 2.082 now.

rracariu:
Will this be usable with `-betterC` ?

src/core/experimental/array.d
edi33416: What do you guys think about moving `emplace` to `core.internal.traits` ?

src/core/experimental/array.d
edi33416: Good catch. Will remove obsolete comment.

src/core/experimental/array.d
edi33416: The `version(unittest)` is needed in order to use an allocator that can collect information about the number of bytes used, during testing. I know this is not the ideal approach, but I couldn't figure something better.
We need the unittest version of the allocator, as this has caught a couple of implementation bugs in the past. I would strongly advise against removing it.

src/core/experimental/array.d
jacob-carlborg: It's not really a trait is it?

edi33416:
> Versioning based on `version(unittest)` is very evil and has lead to many tricky bugs which only happened in production. Let's try hard to avoid this.

I know this is not ideal, but this is needed in order to check if there were any memory leaks.

> Also I think almost all unittests should have `@nogc` as this is the primary motivation for this PR.

I tried to add this to as many tests as I could, but some of them might allocate using the GC. Code such as `Array!int([1, 2, 3])` falls in this case, as the array literal might call the GC.

edi33416:
> Will this be usable with `-betterC` ?

I haven't checked this. I will do and come back with an answer.

src/core/experimental/array.d
edi33416: You are right. My bad, sorry :).
The question remains: should we move it into a `core.internal.conv` ?

I think `emplace` is especially usefull when you want to construct an immutable object into a predefined memory area.

src/core/experimental/array.d
wilzbach: Well, it was my comment after all ;-)
You should add a unittest to ensure that it's actually @safe though.

wilzbach:
> I tried to add this to as many tests as I could, but some of them might allocate using the GC. Code such as Array!int([1, 2, 3]) falls in this case, as the array literal might call the GC.

Yup, I'm aware, but my point was to have at least one small unittest per function which is `@nogc` to ensure that it really is, by e.g. `Array!int(1, 2, 3)`, `static immutable` or maybe even moving a private copy of (parts of) `staticArray` to druntime (checkout https://dlang.org/changelog/2.082.0.html#std-array-asStatic).

src/core/experimental/array.d
wilzbach: Let's make it `private` for now until we can convince Andrei to add this to `std.range`

wilzbach:
>> Versioning based on version(unittest) is very evil and has lead to many tricky bugs which only happened in production. Let's try hard to avoid this.

> I know this is not ideal, but this is needed in order to check if there were any memory leaks.

Can we at least use sth. like `CoreUnittest`, s.t. people will use the production version of the allocator in their own tests?

See e.g. https://github.com/dlang/phobos/pull/5927 for a similar attempt for Phobos, which was reverted due to a regression with symbols, but IIRC we just need to be careful not to slap it on entire functions.

src/core/experimental/array.d
andralex: Let's make it private for now and postpone that decision. Fight one fight at a time.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+{
+    static if (is(typeof(T.outer)))
+        enum isInnerClass = __traits(isSame, typeof(T.outer), __traits(parent, T));
+    else
+        enum isInnerClass = false;
+}
+
+enum classInstanceAlignment(T) = size_t.alignof >= T.alignof ? size_t.alignof : T.alignof;
+
+T emplace(T, Args...)(T chunk, auto ref Args args)
```
andralex: Please do add a (non-ddoc) comment explaining this function, perhaps with a ref to the phobos version.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+template isInnerClass(T)
+if (is(T == class))
+{
+    static if (is(typeof(T.outer)))
+        enum isInnerClass = __traits(isSame, typeof(T.outer), __traits(parent, T));
+    else
+        enum isInnerClass = false;
+}
+
+enum classInstanceAlignment(T) = size_t.alignof >= T.alignof ? size_t.alignof : T.alignof;
```
andralex: Add a comment.

src/core/experimental/array.d
andralex: add comment

src/core/experimental/array.d
andralex: add a lil comment here, too

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
+///
+module core.experimental.array;
+
+import core.internal.traits : Unqual;
+
+// { "Imports" from Phobos
+
+// { Allocators
+
+/**
```
andralex: make it non-ddoc if private

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+}
+
+private @nogc pure nothrow @safe
+void testEmplaceChunk(void[] chunk, size_t typeSize, size_t typeAlignment)
+{
+    assert(chunk.length >= typeSize, "emplace: Chunk size too small.");
+    assert((cast(size_t) chunk.ptr) % typeAlignment == 0, "emplace: Chunk is not aligned.");
+}
+
+enum hasElaborateDestructor(T) = __traits(compiles, { T t; t.__dtor(); })
```
andralex: and add comment

src/core/experimental/array.d
andralex: get rid or move

src/core/experimental/array.d
andralex: yah, "do as T[] does" -> T[] does not have some magic `equal` function.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+
+    assert(equal(a, a));
+    assert(equal(a, b));
+    assert(equal(b, a));
+    assert(equal(b, b));
+    a ~= 1;
+    assert(!equal(a, b));
+}
+
+struct PrefixAllocator
```
andralex: private

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+    void _dispose(T)(bool isShared, T[] b)
+    {
+        return isShared ?  sharedAllocator.dispose(b) : localAllocator.dispose(b);
+    }
+}
+
+///
+struct Array(T)
+{
+private:
```
andralex: recommend repeated private instead of label (clearer without context)

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+
+    @nogc nothrow
+    void _dispose(T)(bool isShared, T[] b)
+    {
+        return isShared ?  sharedAllocator.dispose(b) : localAllocator.dispose(b);
+    }
+}
+
+///
+struct Array(T)
```
andralex: rcarray

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+    }
+    else
+    {
+        alias localAllocator = shared PrefixAllocator.instance;
+        alias sharedAllocator = shared PrefixAllocator.instance;
+    }
+
+    static enum double capacityFactor = 3.0 / 2;
+    static enum initCapacity = 3;
+    bool isShared;
```
andralex: eliminate

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+     * If `Stuff` defines `length`, `Array` will use it to reserve the
+     * necessary amount of memory.
+     *
+     * Params:
+     *      stuff = an input range of elements that are implitictly convertible
+     *              to `T`
+     *
+     * Complexity: $(BIGOH m), where `m` is the number of elements in the range.
+     */
+    version(none)
```
andralex: delete this ctor

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+
+    /**
+     * Check if the array is empty.
+     *
+     * Returns:
+     *      `true` if there are no elements in the array; `false` otherwise.
+     *
+     * Complexity: $(BIGOH 1).
+     */
+    @nogc nothrow pure @safe
```
andralex: delete

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+    /**
+     * Provide access to the first element in the array. The user must check
+     * that the array isn't `empty`, prior to calling this function.
+     *
+     * Returns:
+     *      a reference to the first element.
+     *
+     * Complexity: $(BIGOH 1).
+     */
+    ref auto front(this _)()
```
andralex: delete

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+    }
+
+    /**
+     * Advance to the next element in the array. The user must check
+     * that the array isn't `empty`, prior to calling this function.
+     *
+     * Complexity: $(BIGOH 1).
+     */
+    @nogc nothrow pure @safe
+    void popFront()
```
andralex: delete

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+     *
+     * This must be used in order to iterate through a `const` or `immutable`
+     * array For a mutable array this is equivalent to calling `popFront`.
+     *
+     * Returns:
+     *      an array that starts with the next element in the original array.
+     *
+     * Complexity: $(BIGOH 1).
+     */
+    Qualified tail(this Qualified)()
```
andralex: delete

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+            //alias fn = unaryFun!fun;
+            alias fn = fun;
+
+            // Iterate through the underlying payload
+            // The array is kept alive (rc > 0) from the caller scope
+            foreach (ref e; this.payload)
+            {
+                static if (!is(typeof(fn(T.init)) == int))
+                {
+                    cast(void) fn(e);
```
andralex: For integration with `Flag!"each"`, I suggest:

```
static if (is(typeof(fn(T.init)) == int) Result)
{
    static if (is(typeof(Result.init == Result.init)) { if (fn(e) == Result.init) return false; }
    cast(void) fn(E);
}
```

src/core/experimental/array.d
```diff
@@ -0,0 +1,2299 @@
...
+    enum uint alignment = size_t.alignof;
+    static enum prefixSize = size_t.sizeof;
+
+    version(unittest)
+    {
+        // During unittesting, we are keeping a count of the number of bytes allocated
+        size_t bytesUsed;
+    }
+
+    @trusted @nogc nothrow pure
```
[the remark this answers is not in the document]
edi33416: The entire function was made `@trusted` for esthetic reasons; marking it safe would add some ugly `@trusted` lambdas in the implementation without providing any other gain.

src/core/experimental/array.d
edi33416: Not sure if it can be eliminated as it's still needed in order to figure out if the prefix is a `shared size_t` or just a `size_t`.

src/core/experimental/array.d
edi33416: done

src/core/experimental/array.d
edi33416: done

src/core/experimental/array.d
edi33416: done

src/core/experimental/array.d
edi33416: This is now gone.

src/core/experimental/array.d
edi33416: done

src/core/experimental/array.d
ZombineDev: You can make it a tempalte parameter and require users to use helper functions to instatiate the type and make it's constructors private to the module.

edi33416: I thought about that, but I'd rather leave it as it is instead of having some wrappers. I was hoping I could get the compile time information about how the struct was instantiated.
I was hoping I can do something along the lines
```
static if (is(typeof(this) == immutable))
{
   // note that it started as immutable;
}
else
{
  // mutable reference
}
```

Such that
```
auto a = Array!int(1, 2, 3); // will hit the else branch
auto a2 = immutable Array!int(1, 2, 3); // will take the if branch
```

ZombineDev: I know, as I have found myself reaching for the same technique previously, but I also haven't found a reliable and clean way to do it. Let's hope that @andralex can offer an alternative.

src/core/experimental/array.d
edi33416: Not sure if this is correct, though, as a function that returns `bool` will now match the `Flag!each` pattern

src/core/experimental/array.d
edi33416: > > Will change. According to the docs, this should throw a RangeError. Am I right?
> 
> Yes, but it depends on which compiler setting is used and if it’s a `@safe` function or not. Have a look at the `boundscheck` flag https://dlang.org/dmd-osx.html#switch-boundscheck

@jacob-carlborg do you happen to know how can I check if `boundscheck` is set to `safeonly`? The only conditional compilation I could find was `version (D_NoBoundsCheck)`, so I can distinguish the `on/off` behaviour, but not the `safeonly` one.

Also, RangeError is not `@nogc`, so I've kept the `assert(...)` instead of throwing a RangeError. Any suggestions here?

edi33416:
> > > Versioning based on version(unittest) is very evil and has lead to many tricky bugs which only happened in production. Let's try hard to avoid this.
> 
> > I know this is not ideal, but this is needed in order to check if there were any memory leaks.
> 
> Can we at least use sth. like `CoreUnittest`, s.t. people will use the production version of the allocator in their own tests?
> 
> See e.g. [dlang/phobos#5927](https://github.com/dlang/phobos/pull/5927) for a similar attempt for Phobos, which was reverted due to a regression with symbols, but IIRC we just need to be careful not to slap it on entire functions.

@wilzbach is there a standard name, like `CoreUnittest` defined for [conditional compilation](https://dlang.org/spec/version.html) ? Somewhere in the discussion of the PR you pointed, it was being discussed if a `Std*` should be reserved.

Thanks

src/core/experimental/array.d
edi33416: Can you elaborate, please?

src/core/experimental/array.d
edi33416: This is solved now, as it's delegating to the underlying built-in array type.

edi33416:
> > Will this be usable with `-betterC` ?
> 
> I haven't checked this. I will do and come back with an answer.

@rracariu @wilzbach `rcarray` is now usable in `-betterC` code

edi33416:
@rracariu @wilzbach `rcarray` is now usable in `-betterC` code

edi33416:
@rracariu @wilzbach
`rcarray` is now usable in `-betterC` code

RazvanN7:
Please rebase this and let's see if works as expected.

src/core/experimental/array.d
```diff
@@ -0,0 +1,2175 @@
...
+
+    this(const ref typeof(this) rhs) immutable
+    {
+        bool _isShared = true;
+        mixin(immutableInsert!(typeof(rhs.payload), true, "rhs.payload")());
+        //isShared = true;
+        setIsShared(support, true);
+    }
+
+    this(immutable ref typeof(this) rhs) immutable
```
edi33416: Changing to
`    this(scope immutable ref typeof(this) rhs) immutable` fix the `scope` error, but will make the ctor `@system`.
Adding a `@trusted` (For testing purposes only) to quickly mute all the complaints will pass all the tests. Why is the `scope` required and why is it making the function `@system` when used?

edi33416: Another way of bypassing the error, without any `@trusted` "magic" is to change the `idup` implementation from
```
immutable(rcarray!T) idup(this Q)()
{
    return immutable rcarray!T(this);
}
```
to
```
immutable(rcarray!T) idup(this Q)()
{
    static if (is(Q == immutable))
    {
        return this;
    }
    else
    {
        return immutable rcarray!T(this);
    }
}
```

RazvanN7: The generated copy constructor has the signature "this(ref scope return inout typeof(this) ths) inout". The error appears if the rcarray is a member of a struct for which a copy constructor would be implicitly defined: I assume that the generated copy constructor calls your defined copy constructor.

anon17:
The `is` condition in PrefixAllocator is always true.

```d
alias localAllocator = shared PrefixAllocator.instance;
alias sharedAllocator = shared PrefixAllocator.instance;
```
`shared` looks out of place here.

PR #2349, opened 2018-11-12
Use `__c_wchar_t`.

thewilsonator:


thewilsonator:
Hmm
```
src/core/stdc/wchar_.d(86): Error: cannot implicitly convert expression `65535` of type `int` to `__c_wchar_t`
src/core/stdc/config.d(159): Error: enum `core.stdc.config.__c_wchar_t` is forward referenced when looking for `min`
```

I thought it was supposed to build with the latest dmd master, that hack was extended to `__c_wchar_t`.

src/core/stdc/config.d
```diff
@@ -31,6 +31,10 @@ version (StdDdoc)
             alias ddoc_long = int;
             alias ddoc_ulong = uint;
         }
+        version (Posix)
+            alias ddoc_wchar_t = dchar;
+        else
+            alias ddoc_wchar_t = wchar;
```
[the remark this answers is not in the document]
jacob-carlborg: Ok

src/core/stdc/config.d
```diff
@@ -112,6 +122,9 @@ version (Windows)
 
     alias long  cpp_longlong;
     alias ulong cpp_ulonglong;
+
+    enum __c_wchar_t : wchar;
+    alias wchar_t = __c_wchar_t;
```
kinke: I strongly suggest aliasing directly - `alias wchar_t = wchar`. `__c_[u]long[long]` are reserved identifiers; these magic enums (previously: structs) are solely needed to recognize these types (as opposed to a regular `[u]{int,long}`) in the frontend for C++ mangling purposes, and not needed here.

thewilsonator: They are magic https://github.com/dlang/dmd/commit/4ba6c2a46f02d3662c9e289ef5454fc261ecd8d3

kinke: Why?! Can't `{w,d}char` be simply mangled appropriately for `extern(C++)`?

thewilsonator: they need to mangled as `char[16|32]_t`, see https://github.com/dlang/dmd/pull/8342

thewilsonator: Hence `wchar_t` needs its own.

kinke: Ah, thx for the link. C++ is just such a mess.

thewilsonator: True dat.

ibuclaw:
```
src\rt\dmain2.d(115): Error: function `core.sys.windows.winbase.LoadLibraryW(const(wchar)*)` is not callable using argument types `(const(__c_wchar_t*))`
src\rt\dmain2.d(115):        cannot pass argument `name` of type `const(__c_wchar_t*)` to parameter `const(wchar)*`
```
These function signatures need to be updated to use `wchar_t` instead I guess.
```
winnt.d: alias const(wchar)* LPCWCH, PCWCH, LPCWSTR, PCWSTR;
```

thewilsonator:
Hmm, looks like CircleCI builds with 2.079.1, which obviously is not going to work as this requires DMD maser.

thewilsonator:
@ibuclaw should be ready to go.

src/core/sys/windows/winnt.d
```diff
@@ -383,43 +384,43 @@ enum : WORD {
     SECURITY_MANDATORY_PROTECTED_PROCESS_RID = 0x5000,
     SECURITY_MANDATORY_MAXIMUM_USER_RID      = SECURITY_MANDATORY_SYSTEM_RID
 }
-
+package enum wchar_t[] TEXT(string a) = cast(wchar_t[])a;
```
rainers: Shouldn't this keep immutable/const?

Maybe better as a regular CTFE function to avoid the template bloat in memory and object file?

thewilsonator: You can cast a literal to wstring at compile time fine, you'd have to decode it if you wanted that as a function wouldn't you? Anyway feel free to make any changes you think are appropriate. I don't do windows stuff, I'm just trying to make this pass.

rainers:
> The only thing I'm worried about is implicit conversions from `string` to `immutable(wchar_t)[]` failing to happen breaking code. This requires people to actually use `""w` and `""w.ptr` properly.

We had this with c_long when it was a struct, too. I suspect this will cause a lot of breakage. I haven't tried but `""w.ptr` is probably not good enough, you still have to cast it with `cast(immutable(wchar_t)*)""w.ptr`. Or do they convert implicitly?

I'm not sure the windows definitions have to change to begin with, these are C APIs, not C++.

thewilsonator:
Removed most of the windows diff, yeah pointer conversions don't work yet, need to fix that in DMD.

thewilsonator:
this needs 
https://github.com/dlang/dmd/pull/8958

WalterBright:
Here's the trouble:
```
src\rt\dmain2.d(115): Error: function `core.sys.windows.winbase.LoadLibraryW(const(wchar)*)` is not callable using argument types `(const(__c_wchar_t*))`
src\rt\dmain2.d(115):        cannot pass argument `name` of type `const(__c_wchar_t*)` to parameter `const(wchar)*`
```
This is brought on by the following line in `core.sys.windows.winnt`:
```
alias wchar WCHAR;
```
The interesting thing is that the Windows API functions do not use name mangling. So the name mangling isn't an issue. The problem is really in `src\rt\dmain2.d`, i.e. declare `rt_loadLibraryW(const wchar_t* name)` instead as `rt_loadLibraryW(const WCHAR* name)` instead, and all the other uses of `wchar_t` in the file.

This should be future proof, as it will follow whatever Microsoft does with `WCHAR`.

WalterBright:
Now that https://github.com/dlang/druntime/pull/2389 has merged, this should be ready to rock.

WalterBright:
```
src\rt\dmain2.d(386): Error: function `core.stdc.wchar_.wcslen(scope const(__c_wchar_t*) s)` is not callable using argument types `(const(wchar*))`
src\rt\dmain2.d(386):        cannot pass argument `wCommandLine` of type `const(wchar*)` to parameter `scope const(__c_wchar_t*) s`
src\rt\dmain2.d(403): Error: function `core.stdc.wchar_.wcslen(scope const(__c_wchar_t*) s)` is not callable using argument types `(wchar*)`
src\rt\dmain2.d(403):        cannot pass argument `wargs[cast(ulong)i]` of type `wchar*` to parameter `scope const(__c_wchar_t*) s`
```
A cast is needed to call `wcslen()`.

thewilsonator:
That is exactly what https://github.com/dlang/dmd/pull/8958 was supposed to fix.

thewilsonator:
I mean, I can add it, but it doesn't fix the underlying issue and the same thing will affect users as well.

WalterBright:
An alternative is to change the declaration of `WCHAR` from `wchar` to `wchar_t`. This would match what Microsoft is doing.

Argh. There are no good solutions.

thewilsonator:
https://docs.microsoft.com/en-us/windows/desktop/extensiblestorageengine/gg269344(v%3Dexchg.10)

```c
#if !defined(_NATIVE_WCHAR_T_DEFINED)
typedef unsigned short WCHAR;
#else
typedef wchar_t WCHAR;
#endif
```

`_NATIVE_WCHAR_T_DEFINED` is defined iff  `/Zc:wchar_t-` is passed which is apparently a [bad thing to do](https://docs.microsoft.com/en-au/cpp/build/reference/zc-wchar-t-wchar-t-is-native-type?view=vs-2017).

> change the declaration of WCHAR from wchar to wchar_t

We should probably do that anyway, but it won't fix the problem, only move it elsewhere.

WalterBright:
This suggests a solution for us.

1. define `WCHAR` as `wchar_t`
2. for `version(Cpp98)`, define `wchar_t` the old way
3. otherwise define `wchar_t` the new way
4. uses casts to ensure druntime/Phobos compiles either way
5. people whose code breaks can use `-transition=cpp98` until their code is transitioned

WalterBright:
See https://github.com/dlang/dmd/pull/9029 for addition of switch.

Geod24:
Closing as it hasn't seen any activity for a while and is outdated (also cleaning up branches in the repository).

PR #2350, opened 2018-11-14
Fix Issue 19398 - Document meaning of core.atomic.MemoryOrder

n8sh:
https://forum.dlang.org/thread/vzzitsnvsxljscrzbcnm@forum.dlang.org
"Hoist-load", "hoist-store", "sink-load", and "sink-store" are used in the documentation but they're not formally defined. A full description would be nice but minimally linking to an external site makes the documentation more useful.


n8sh:
Good to go!

PR #2351, opened 2018-11-14
[RFC] Add core.memory.onOutOfMemoryError usable in betterC

n8sh:
See https://github.com/dlang/phobos/pull/6764. This PR puts a new function in `core.memory` rather than altering the existing function in `core.exception` because most of `core.exception` will not compile in betterC. When exceptions are enabled this is just an alias of `core.exception.onOutOfMemoryError`.

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -1184,4 +1184,24 @@ unittest
     assert(GC.addrOf(y.ptr) == null);
 }
 
-
+/**
+ * A callback for out of memory errors in D. An $(REF OutOfMemoryError, core,exception)
+ * will be thrown if exceptions are enabled. Otherwise the program will abort.
+ * When exceptions are enabled this function is a wrapper around
+ * $(REF onOutOfMemoryError, core,exception).
+ *
+ * Throws:
+ *  $(REF OutOfMemoryError, core,exception).
+ */
+void onOutOfMemoryError()(void* pretend_sideffect = null) @nogc nothrow pure @trusted
+{
+    version (D_Exceptions)
+    {
+        static import core.exception;
+        core.exception.onOutOfMemoryError();
+    }
+    else
+    {
+        assert(0, "Memory allocation failed");
+    }
+}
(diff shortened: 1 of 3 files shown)


jacob-carlborg:
How does the test work?

ZombineDev:
I'd prefer making `core.exception.onOutOfMemoryError` a template, but this is the next best thing. So on the current implementation, why don't just call `core.exception.onOutOfMemoryError` internally in the new function, instead of aliasing it? That way would be cleaner, IMO.

n8sh:
@ZombineDev I want the new function to be a template during `betterC` but not otherwise, so there would still need to be 2 different top-level definitions of onOutOfMemoryError. At that point there is nothing to be gained from a wrapper function.

@jacob-carlborg The test aims to verify that code using `core.memory.onOutOfMemoryError` compiles with the `-betterC` flag.

ZombineDev:
> I want the new function to be a template during betterC but not otherwise

Why? This the part to which I was referring that the code would be cleaner without, as I think it's better to keep things uniform w.r.t. build settings.

n8sh:
The aim is to avoid any reduction in capabilities in non-betterC builds that use the D Runtime as a shared library. If no one cares about being able to replace a shared library's `onOutOfMemoryError` and have it affect already-compiled programs, then it could be a template.

n8sh:
Wait a minute. `onOutOfMemoryError` was under `core`, not `rt`, so maybe what I'm describing was never possible in the first place.

thewilsonator:
> maybe what I'm describing was never possible in the first place.

Ping me when you figure that out.

ZombineDev:
> [..]  make `core.exception.onOutOfMemoryError` work in `betterC`

Yes, I would highly prefer this as well. I'm just wondering whether `extern (C)` is a necessary attribute for `core.exception.onOutOfMemoryError`. If it turns out that it's not, I think the best solution is to turn it into a template.

n8sh:
Closing in favor of https://github.com/dlang/druntime/pull/2361

PR #2354, opened 2018-11-17
rt.sections_elf_shared: Fix struct tls_index definition on x32

ibuclaw:
Running all druntime unittests, this is the only place where an X32 regression was found, the effect seen was that live TLS data was being collected and freed at runtime.

### DIFF
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -21,6 +21,7 @@ static if (SharedELF):
 
 // debug = PRINTF;
 import core.memory;
+import core.stdc.config;
 import core.stdc.stdio;
 import core.stdc.stdlib : calloc, exit, free, malloc, EXIT_FAILURE;
 import core.stdc.string : strlen;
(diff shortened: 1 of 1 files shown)


src/rt/sections_elf_shared.d
```diff
@@ -906,8 +907,27 @@ version (Shared) void* handleForAddr(void* addr) nothrow @nogc
...
+        // See https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/x86_64/dl-tls.h
+        version (X86_64)
+        {
+            ulong ti_module;
+            ulong ti_offset;
+        }
+        else
+        {
+            c_ulong ti_module;
+            c_ulong ti_offset;
```
n8sh: I've verified that in other cases `ti_module` & `ti_offset` are `unsigned long int`.

ibuclaw: Oh, just for the discussion thread, I left `size_t` in as the default branch.  At least musl defines this as a `size_t[2]` type.

(Grep'ing musl 1.16 for __tls__get_addr):
```
void *__tls_get_addr(size_t *);
...
return __tls_get_addr((size_t []){def.dso->tls_id, def.sym->st_value});
```

n8sh: By "other cases" I meant only by grepping the source of glibc. I didn't mean for musl etc.

PR #2356, opened 2018-11-18
[GDC] Do not include padding when hashing floating point types

jpf91:
The exact contents of the padding is not specified. Usually it's 0, but some optimizations in GDC sometimes lead to different padding content.

* For floating point types without padding, the code is unchanged.
* For real or imaginary types with padding, just ignore the trailing padding.
* For complex types where the real/imaginary types have padding, we can't hash the value as one contiguous memory block, as there will be a padding 'hole'. Here, hash real and imaginary part
individually. Complex types without padding are still hashed as one memory block.

ping @ibuclaw

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -385,6 +385,8 @@ size_t hashOf(T)(scope const T val, size_t seed = 0) if (!is(T == enum) && __tra
 {
     static if (__traits(isFloating, val))
     {
+        import core.internal.convert : floatSize;
+
         static if (floatCoalesceZeroes || floatCoalesceNaNs)
         {
             import core.internal.traits : Unqual;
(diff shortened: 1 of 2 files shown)


thewilsonator:
Looks fine to me. cc resident hashing expert @n8sh

src/core/internal/hash.d
```diff
@@ -412,7 +414,23 @@ size_t hashOf(T)(scope const T val, size_t seed = 0) if (!is(T == enum) && __tra
...
+        {
+            static if (is(T : creal) && T.sizeof != 2 * floatSize!(typeof(T.re)))
+            {
+                auto h1 = hashOf(data.re);
+                return hashOf(data.im, h1);
+            }
+            else static if (is(T : real) || is(T : ireal))
+            {
+                // Ignore trailing padding
+                auto bytes = toUbyte(data)[0 .. floatSize!T];
```
ibuclaw: One consideration is float endianess, but we don't currently expose that in the language with either version condition nor traits (yes, a platform could have one endian for floats, and another for integers).

n8sh:
LGTM.

PR #2361, opened 2018-11-20
Make core.exception.onOutOfMemoryError work in betterC

n8sh:


src/core/exception.d
```diff
@@ -9,6 +9,39 @@
  */
 module core.exception;
 
+// Compiler lowers final switch default case to this (which is a runtime error)
+void __switch_errorT()(string file = __FILE__, size_t line = __LINE__) @trusted
```
thewilsonator: Please leave this in its original spot.

n8sh: Then it wouldn't be enable in BetterC.

thewilsonator: Ohh, I see this is for 
```version (D_BetterC)
{
    //...
}
else:
```

PR #2362, opened 2018-11-22
[refactor] core.sys.posix.msg: Genericize msqid_ds definition

ibuclaw:
This more closely matches how glibc arranges all port headers (all the arch/bits/msq.h sources are now removed).

### DIFF
--- a/src/core/sys/posix/sys/msg.d
+++ b/src/core/sys/posix/sys/msg.d
@@ -171,13 +171,6 @@ version (linux)
             ulong_t __glibc_reserved5;
         }
     }
-
-
-    public enum MSG_MEM_SCALE =  32;
-    public enum MSGMNI =     16;
-    public enum MSGMAX =   8192;
-    public enum MSGMNB =  16384;
-
 }
 else
 {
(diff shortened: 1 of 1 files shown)


thewilsonator:
Is there any reason for 8 commits?

ibuclaw:
@thewilsonator - I hope it makes it easier to follow.  And so you don't have to ask me "Why was this removed?" etc...

ibuclaw:
I mean:
- core.sys.posix.msg: Refactor common GENERICMSQ version (Begins the introduction of a common version condition).
- core.sys.posix.msg: Use GENERICMSQ for more architectures (Copies glibc refactoring commit, removes references to now removed headers).
- core.sys.posix.msg: Factor out linux and non-linux definitions (Obvious, this is a posix module that only implements linux-specific `bits/msq.h` - added `version(linux)` around it all, then added an `else` branch which mirrors glibc's non-linux `bits/msq.h`).
- core.sys.posix.msg: Use common msqid_ds for all architectures (Copies glibc refactoring commit)
- core.sys.posix.msg: Merge duplicate multilib definition blocks (Refactoring that only makes sense after the previous commit.
- core.sys.posix.msg: Use ulong_t instead of c_ulong in msgq_ids (Copies glibc commit, needed for supporting X32 ABI)
- core.sys.posix.msg: Remove unneeded use of public import and enum (Local refactoring after all previous changes)
- core.sys.posix.msg: Remove abitrary MSG constants (Independent change / final clean-up).

PR #2363, opened 2018-11-22
Fix Issue 19370 - AA require() and update() can't be used in @safe code

radcapricorn:


src/object.d
```diff
@@ -3163,7 +3163,18 @@ inout(V) get(K, V)(inout(V[K])* aa, K key, lazy inout(V) defaultValue)
 ref V require(K, V)(ref V[K] aa, K key, lazy V value = V.init)
 {
     bool found;
-    auto p = cast(V*) _aaGetX(cast(void**)&aa, typeid(V[K]), V.sizeof, &key, found);
+    // if key and value are @safe-ly copyable, `require` can infer @safe
+    static if (isSafeCopyable!K && isSafeCopyable!V)
```
jacob-carlborg: How does this condition help, why not always use the `@trusted` wrapper?

radcapricorn: Maybe that wasn't that clear of an answer: if we just wrap the whole call into a @trusted, it's going to allow un-@safe key postblits. The unittest where S is key will fail.

jacob-carlborg: I still don't understand. Your running exactly the same code regardless if it's `isSafeCopyable` or not.

radcapricorn: Yes, the code is the same, but the inferred attributes for update()/require() will be different. _aaGetX internally calls the key's postblit. That postblit may be user-defined. If that postblit is `@system`, the whole require()/update() call should be `@system`. But if I were to wrap the call to _aaGetX into a `@trusted` unconditionally, I'd in effect silently promote `@system` user code to `@safe`.

jacob-carlborg: Ah, ok, I see. Thanks for explaining. Wasn't every obvious to me at least.

radcapricorn: No problem.

thewilsonator: This good to go?

radcapricorn: By me? Yes :)

thewilsonator: Please squash.

src/object.d
```diff
@@ -3205,13 +3219,54 @@ void update(K, V, C, U)(ref V[K] aa, K key, scope C create, scope U update)
...
         *p = update(*p);
 }
 
+unittest
+{
+    static struct S
+    {
+        int x;
+    @nogc nothrow pure:
+        this(this) /* not @safe */ {}
```
jacob-carlborg: Perhaps say `@system` here to make it extra clear.

radcapricorn: Done.

PR #2364, opened 2018-11-22
Fix Issue 19421: Make pureMalloc, etc. usable in BetterC

n8sh:


Geod24:
Honestly, this is not an enhancement of the code. If we have to apply those kinds of hacks to anywhere where we want to use `-betterC`, we're going to end up with a giant mess.

n8sh:
The "hack" for betterC was just changing `void* pureMalloc(size_t)` to `void* pureMalloc()(size_t)` and getting rid of the unnecessary `fakePureErrno` wrapper function in most cases. I'll disagree that his is a giant mess.

The rest is of the PR -- the ugly parts -- was fixing `fakePureErrno` for the fallback case for unrecognized C runtimes where `core.stdc.errno.errno` is two separate property functions.

PR #2365, opened 2018-11-22
Make TempStringNoAlloc zero-init + avoid unnecessary init

n8sh:


src/core/internal/string.d
```diff
@@ -56,7 +56,7 @@ char[] unsignedToTempString()(ulong value, return scope char[] buf, uint radix =
 private struct TempStringNoAlloc
 {
     // need to handle 65 bytes for radix of 2 with negative sign.
-    private char[65] _buf;
+    private char[65] _buf = 0;
```
rainers: Maybe use `= void` here, too. Since recently, it has the same effect as zero-initialization regarding the .init-value handling, but could allow some optiizations. `= void` also hints that there is no enforced zero-termination.

n8sh: Done.

PR #2366, opened 2018-11-23
Fix Issue 11294 - Object destruction with alias this

radcapricorn:


thewilsonator:
cc @RazvanN7 whats your take on this? Should this be fixed in the compiler rather than the runtime?

Geod24:
I'd say compiler, but in the meantime, having a no-cost/trivial workaround in the runtime can't harm.

RazvanN7:
The bug is another manifestation of : https://issues.dlang.org/show_bug.cgi?id=6777 . This should definitely be fixed in the compiler.

thewilsonator:
@Geod24 that seems reasonable. @RazvanN7 how long do you think a fix will take?

radcapricorn:
Maybe still add it until Issue 6777 is fixed, but add a `@@@BUG@@@` comment?

thewilsonator:
Sounds good!

RazvanN7:
> @RazvanN7 how long do you think a fix will take?
 
I will submit a patch today.

thewilsonator:
Even better! Thanks!

radcapricorn:
Oh, ok, will wait and test then.

RazvanN7:
I've posted my patch for 6777 : https://github.com/dlang/dmd/pull/8712 .

Unfortunately, that patch alone does not suffice to fix this issue. There is another problem in the semantic of the cast expression, which I already have a solution for, but it needs the fix for 6777 to work.

RazvanN7:
> So, if I understand correctly, the course of action now is to first get that fix in, then finish it off, and then this PR wouldn't be needed?

That is correct.

> Or shall I still leave it, to be reverted later?

This PR should stay open for the moment, because I am not sure if my PR is going to get in. I am confident that it will, but you never know if Walter might have some objections. Once Issue 11294 is completely fixed in dmd, we can close this PR

radcapricorn:
Alright, thank you.

RazvanN7:
I was thinking the same thing.

Geod24:
The unittest makes sense, but the workaround is not necessary anymore.

radcapricorn:
Maybe revert, then I'll remove the workaround but leave the unittest?

Geod24:
> Maybe revert, then I'll remove the workaround but leave the unittest?

`git checkout -p 3b9c093d^ -- src/object.d` is enough.
https://github.com/dlang/druntime/pull/2378

PR #2367, opened 2018-11-23
Remove static initializer from struct `destroy`.

radcapricorn:
This change removes a shared static initializer from a `destroy`
implementation for structs, and with it a dependency on `memcpy`.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -506,20 +506,21 @@ any GC memory.
 */
 void destroy(T)(ref T obj) if (is(T == struct))
 {
-    // We need to re-initialize `obj`.  Previously, the code
-    // `auto init = cast(ubyte[])typeid(T).initializer()` was used, but
-    // `typeid` is a runtime call and requires the `TypeInfo` object which is
-    // not usable when compiling with -betterC.  If we do `obj = T.init` then we
-    // end up needlessly calling postblits and destructors.  So, we create a
-    // static immutable lvalue that can be re-used with subsequent calls to `destroy`
-    shared static immutable T init = T.init;
+    // We need to re-initialize `obj`.  Previously, an immutable static
+    // and memcpy were used to hold an initializer. With improved unions, this is no longer
+    // needed.
+    union UntypedInit
+    {
+        T dummy;
+    }
+    static struct UntypedStorage
+    {
+        align(T.alignof) void[T.sizeof] dummy;
+    }
 
     _destructRecurse(obj);
     () @trusted {
-        import core.stdc.string : memcpy;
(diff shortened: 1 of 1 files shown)


thewilsonator:
Hmm, this appears to have broken a few things:
```
ae 0.0.2292: building configuration "ae-test-library"...
--
  | /var/lib/buildkite-agent/builds/build2-1/dlang/druntime/build/distribution/bin/../imports/object.d(516,9): Error: cannot modify struct instance `*cast(Untyped*)&obj` of type `Untyped` because it contains `const` or `immutable` members
  | /var/lib/buildkite-agent/builds/build2-1/dlang/druntime/build/distribution/bin/../imports/std/typecons.d(6240,17): Error: template instance `object.destroy!(NamedPipeImpl)` error instantiating
```
Not sure if this is also related:
```
Generating test runner configuration 'sumtype-test-library' for 'library' (library).
--
  | Performing "unittest" build using dmd for x86_64.
  | sumtype 0.7.0: building configuration "sumtype-test-library"...
  | src/sumtype.d(855,8): Error: static assert:  "No matching handler for type `Evil`"
  | src/sumtype.d(693,14):        instantiated from here: `matchImpl!(SumType!(Evil, int))`
  | src/sumtype.d(280,9):        instantiated from here: `match!(SumType!(Evil, int))`
  | src/sumtype.d(560,2):        instantiated from here: `SumType!(Evil, int)`
```

thewilsonator:
Perhaps just create a temporary on the stack and bitblast it and let the optimiser do its job?

radcapricorn:
Ouch, that first one's bad. Could squash that with an `Unqual`, I guess. I really want a `memcpy` to **not** be there. Input welcome.
Not sure about the second one either, I'll take a look into SumType later, unless @pbackus could relate.

thewilsonator:
I'm not sure `Unqual` would help because `struct Foo { immutable int i; }` is not going to change unless you manually strip out the immutable be reconstruction from the `.tupleof` which seems like overkill. 

You're going to end up with a `memcpy` there anyway because assignment of `.init` is going to copy from where `T.init` lives in the executable anyway.

radcapricorn:
Letting the optimizer do the job is exactly why an explicit `memcpy` should not be there. Whether it ends up inserting a `memcpy` or just blasting values directly in there should be up to the compiler. Let's see how the updated version fares on those tests.

thewilsonator:
Thats some nice wizardry!

thewilsonator:
Appears to have fixed 'em both.

radcapricorn:
:tada:

PR #2368, opened 2018-11-23
In core.stdc.errno directly link __errno on OpenBSD & NetBSD

n8sh:


n8sh:
Moved https://github.com/dlang/druntime/pull/2369 into this PR.

PR #2372, opened 2018-11-24
Fix Issue 11168: attributes for functions in core.stdc.time.

radcapricorn:
Preferably this needs to be looked at by @WalterBright. May break some third-party code. Phobos tests passed though.

### DIFF
--- a/src/core/stdc/time.d
+++ b/src/core/stdc/time.d
@@ -148,21 +148,22 @@ else
 }
 
 ///
-double  difftime(time_t time1, time_t time0);
+pure double  difftime(time_t time1, time_t time0); // MT-Safe
 ///
-time_t  mktime(tm* timeptr);
+@system time_t  mktime(scope tm* timeptr); // @system: MT-Safe env locale
 ///
-time_t  time(time_t* timer);
+time_t  time(scope time_t* timer);
+
 ///
-char*   asctime(in tm* timeptr);
+@system char*   asctime(const scope tm* timeptr); // @system: MT-Unsafe race:asctime locale
 ///
-char*   ctime(in time_t* timer);
+@system char*   ctime(const scope time_t* timer); // @system: MT-Unsafe race:tmbuf race:asctime env locale
 ///
-tm*     gmtime(in time_t* timer);
+@system tm*     gmtime(const scope time_t* timer); // @system: MT-Unsafe race:tmbuf env locale
 ///
-tm*     localtime(in time_t* timer);
+@system tm*     localtime(const scope time_t* timer); // @system: MT-Unsafe race:tmbuf env locale
 ///
-@system size_t  strftime(char* s, size_t maxsize, in char* format, in tm* timeptr);
+@system size_t  strftime(scope char* s, size_t maxsize, const scope char* format, const scope tm* timeptr); // @system: MT-Safe env locale
 
 version (Windows)
 {
@@ -171,9 +172,9 @@ version (Windows)
     ///
     void  _tzset();                          // non-standard
     ///
-    @system char* _strdate(char* s);                 // non-standard
+    @system char* _strdate(return scope char* s);                 // non-standard
     ///
-    @system char* _strtime(char* s);                 // non-standard
+    @system char* _strtime(return scope char* s);                 // non-standard
 
     ///
     extern __gshared const(char)*[2] tzname; // non-standard


src/core/stdc/time.d
```diff
@@ -148,21 +148,22 @@ else
 }
 
 ///
-double  difftime(time_t time1, time_t time0);
+@safe pure double  difftime(time_t time1, time_t time0); // MT-Safe
 ///
-time_t  mktime(tm* timeptr);
+@system time_t  mktime(scope tm* timeptr); // @system: MT-Safe env locale
```
n8sh: Why does "MT-Safe env locale" imply `@system`? `@safe` doesn't require `pure`, does it?

radcapricorn: Because in a nutshell only unqualified "MT-Safe" is, well, safe. `mktime` is reading environment and locale unsynchronized. Nothing is stopping other threads from stomping over global mutable state while this function is being executed.

src/core/stdc/time.d
```diff
@@ -148,21 +148,22 @@ else
...
+@system char*   ctime(const scope time_t* timer); // @system: MT-Unsafe race:tmbuf race:asctime env locale
 ///
-tm*     gmtime(in time_t* timer);
+@system tm*     gmtime(const scope time_t* timer); // @system: MT-Unsafe race:tmbuf env locale
 ///
-tm*     localtime(in time_t* timer);
+@system tm*     localtime(const scope time_t* timer); // @system: MT-Unsafe race:tmbuf env locale
 ///
-@system size_t  strftime(char* s, size_t maxsize, in char* format, in tm* timeptr);
+@system size_t  strftime(scope char* s, size_t maxsize, const scope char* format, const scope tm* timeptr); // @system: MT-Safe env locale
```
n8sh: The reasons it was correct for this to be `@system` are (1) nothing mechanically compels `maxsize` to have any relation to `s` and (2) nothing mechanically compels `format` to be a null-terminated C string. Locale doesn't enter into it.

radcapricorn: It does :( Take a look at `man 7 attributes`. `strftime` *may* produce undefined output in `s`, unless a certain convention is followed. Sure, a garbled string isn't exactly process-rending, but corruption is corruption.

src/core/stdc/time.d
n8sh: >locale
>Functions annotated with locale as an MT-Safety issue read
>from the locale object without any form of synchronization.
>Functions annotated with locale called concurrently with
>locale changes may behave in ways that do not correspond to
>any of the locales active during their execution, but an
>unpredictable mix thereof.

>We do not mark these functions as MT-Unsafe, however, because
>functions that modify the locale object are marked with
>const:locale and regarded as unsafe.  Being unsafe, the latter
>are not to be called when multiple threads are running or
>asynchronous signals are enabled, and so the locale can be
>considered effectively constant in these contexts, which makes
>the former safe.

Do you agree with this reasoning ("it can be considered `safe` because corruption can only occur if some other explicitly un`safe` function is called")?

radcapricorn: I read that as "it's safe because convention, but really it isn't".

ibuclaw: FYI, `@safe` means memory safety. Thread safety is `shared`.

radcapricorn: Thanks, Iain, I didn't know that :) Now, seriously, you're suggesting that a function that may store undefined garbage in `*timeptr` can be `@trusted`?

radcapricorn: @schveiguy well, it shouldn't.

> Safe functions are functions that are statically checked to exhibit no possibility of undefined behavior.

> Undefined behavior happens when an illegal code construct is executed. Undefined behavior can include **random, erratic results**, crashes, faulting, etc. A buffer overflow is an example of undefined behavior.

Documentation of these C functions:
> ...may behave in ways that do not correspond to any of the locales active during their execution, but an unpredictable mix thereof.

Unpredictable time value, unpredictable formatted time string, etc. fit the definition of "random, erratic results" quite unambiguously. Yes, it's not buffer overflow. But receiving a malformed date in a HTTP response, would you trust the server you got that from?

schveiguy: I view `@safe` as memory-safety. That is, you can create incorrect or wacky unpredictable values, but you can't cause exploits. I know it says all UB, but I think that's nearly impossible to implement, you'd have to turn off all optimizations for `@safe` code. I think it's more of a lofty goal than a reality.

But by all means, bring up the issue with Walter and Andrei, maybe I'm wrong.

Also note that memory safety is like a brick wall. If you build safety on top of garbage, the wall is going to fall down, even if you check everything. So in that sense, the fact that you can only create corrupted non-pointer data when you call unsafe functions is perfectly acceptable IMO. It's why `@trusted` code is so hard to get right.

radcapricorn: Unpredictable integer is literally the same thing as unpredictable pointer. Literally. You very much can cause exploits with them as well. Oops, you can't login anymore because apparently your account expired last century. Oops, we accidentally signed your mortgage duration for thirty thousand years. Oops, I can't download a security update because my date doesn't match...

schveiguy: `@safe` != bug free.

radcapricorn: Indeed. But
```
int iDoNotGuaranteeIReturnValidData();
```
is not `@safe`. The appropriate way to call it is:
```d
iMadeSureYouWillReturnValidData();
return () @trusted { return iDoNotGuaranteeIReturnValidData(); }
```
**not** slapping `@trusted` onto it's signature. Those C docs say as much.

schveiguy: The difference between a memory corruption bug and this, is that a memory corruption bug can invalidate ANY state. This only invalidates the state you give it. The warnings can be observed, and taken into account by any users of the function just like any other functions that could cause issues.

Again, a potential problem or bug does not mean something is not `@safe` (as long as it's not a memory safety bug). I can read the docs for this function, and deal with the fallout of the potential problem locally. Some function that returns a dangling pointer can not be handled the same way -- I can't even use the result in any reasonable way.

BTW, perhaps we should never mark C library functions as `@safe`, but rather `@trusted` as this is more accurate (there is no compiler verifying safety, we are assuming it).

radcapricorn: That's funny. Memory corruption is memory corruption. What is unspecified state then?
You **must** read the docs for this function and deal with the preconditions. That's what the snippet above illustrates. You won't do that if your "`@safe`" code happily compiles. Using results of that function cannot be safe without understanding it's implications. You assert that understanding by deliberately documenting your call with an explicit `@trusted`.

> BTW, perhaps we should never mark C library functions as @safe, but rather @trusted as this is more accurate (there is no compiler verifying safety, we are assuming it).

You can't be serious. Let's just mark everything `@safe` then, why bother?..

> Trusted functions are guaranteed by the programmer to not exhibit any undefined behavior if called by a safe function...

These C functions do not satisfy that rule. At all.

schveiguy: > That's funny. Memory corruption is memory corruption

When I say a memory corruption bug I mean modifying state that is not supposed to be accessible to the code in question.

If I use a function that multiplies two numbers together, and I state in the docs that in certain cases it may return an incorrect answer, I want to know what those cases are and avoid them. But it's not going to corrupt memory without me knowing about it.

Change that result to a corrupted pointer, and now you have access to memory that you shouldn't have access to. You can affect totally unrelated parts of the code. In those cases I may STILL want to call the function, but I want to know what the risks are, and I certainly only want to call it from `@system` code.

> You can't be serious. Let's just mark everything `@safe` then, why bother?

Maybe I didn't express myself properly. I meant that marking a function written in and compiled with C as `@safe` is not honest. A `@safe` function implies the compiler has checked it for safety violations and approved it. We really should use `@trusted` because no compiler has checked those functions for safety violations.

> > Trusted functions are guaranteed by the programmer to not exhibit any undefined behavior if called by a safe function...
> These C functions do not satisfy that rule. At all.

Again, `@safe` is not the same as bug free. This function has bugs in certain situations. It does not cause memory corruption. I wouldn't call it undefined behavior, just buggy behavior. And it's perfectly avoided by not changing the locale mid-program in a separate thread.

n8sh: It seems that thread-safety issues are intended to be not entirely disjoint from `@safe`. For instance the following does not compile:

```d
__gshared int count;

int currentCount() @safe
{
    return count;
}
```

radcapricorn: If you call a function that gives you unpredictable date and then use the result to modify a database selection you corrupt your database. Problem with memory corruption is that it's usually separated from the cause, and the cause is usually unpredictable values gotten elsewhere. Just like your "safe" return-uninitialized-int function.

> A @safe function implies the compiler has checked it for safety violations and approved it.

Not unless you rewrite the language spec. A @safe function is a function that doesn't produce undefined results, doesn't have unspecified side effects, i.e. does not exert UB. There are plenty of C functions that are perfectly `@safe`. There are way more that are `@system`. What we actually shouldn't have all that much is `@trusted` C functions.

> Again, @safe is not the same as bug free. This function has bugs in certain situations. It does not cause memory corruption. I wouldn't call it undefined behavior, just buggy behavior.

This function doesn't have bugs. It may produce undefined results. That's literally UB.

> And it's perfectly avoided by not changing the locale mid-program in a separate thread.

Which the compiler cannot ascertain in any way, shape, or form. This is exactly why this function has to be `@system`, and calling it from `@safe` should go through an explicit `@trusted`.

I'm really astonished that you're defending it, or that you really think unpredictable behavior can be `@trusted`. It can't, not without the programmer making sure of that, which is what `@trusted` actually is per its own definition, for crying out loud.

radcapricorn: @n8sh, yes, that's one of the rules of `@safe`. Which is one more reason that these functions simply cannot be implicitly `@trusted`.

schveiguy: @radcapricorn You are completely misunderstanding that marking a C function as `@safe` is exactly the same as marking it as `@trusted`. It doesn't mean anything different, because the compiler *didn't check it*. A C function compiled by a C compiler, but marked as `@safe` can call any function, including ones that would have to be marked `@system`. The `@safe` attribute is dishonest, it should be `@trusted` (if appropriate).

The rule about `__gshared` is probably the most convincing that these shouldn't be marked `@trusted`, as essentially that *is* the problem that the rule is trying to avoid. Not that it's a good logical argument, but simply that it should follow the rules that have been established.

I still view `@safe` as focused on memory safety, and not in general thread or race issues. I don't see the access of `__gshared` data as being critical to the promise of `@safe`-ty. There are many more ways `@safe` allows unpredictable data or races.

Trivially, there's an easy way for @n8sh's code to be compilable, and still have the same problems as it would with `__gshared`:

```D
shared int count;

int currentCount() @safe
{
    return count;
}
```

radcapricorn: I'm sorry Steven, but it's you who are misunderstanding here. Interfacing to C, you will either have `@safe` functions, which you **know** can't cause UB, like the one in this very module (`difftime`), or you will have `@system` functions. Marking arbitraty C functions `@trusted` is just a hack to allow calling them from `@safe` D functions, nothing more. Which is further illustrated by you now resorting to "how to make this compile" instead of "how to make this `@safe`", which I find just unnecessary hand waving.

If you have your own opinions about `@safe`, and as a "code owner" for this module, why, exactly, did you not close the original bug report? Why are we having this debate here? I could care less and just close this PR and forget about it, if that's what you want. But then please do something about that bug report, it's five years old.

aG0aep6G: > If you call a function that gives you unpredictable date and then use the result to modify a database selection you corrupt your database. Problem with memory corruption is that it's usually separated from the cause, and the cause is usually unpredictable values gotten elsewhere. Just like your "safe" return-uninitialized-int function.

Bad example. A program that exhibits undefined behavior might write garbage to the database, but there are also `@safe` ways of doing that. Preventing wrong data is not the purpose of `@safe`.

> A `@safe` function is a function that doesn't produce undefined results, doesn't have unspecified side effects, i.e. does not exert UB.

It's not entirely clear yet whether an undefined value implies undefined behavior in D.

[Issue 18016](https://issues.dlang.org/show_bug.cgi?id=18016) is relevant here. The spec currently says that using the (undefined) value of a void initialized variable is invalid (has undefined behavior), but DMD disagrees. Walter's solution is to change the spec, making it valid to use uninitialized values (https://github.com/dlang/dlang.org/pull/2260).

> There are plenty of C functions that are perfectly `@safe`. There are way more that are `@system`. What we actually shouldn't have all that much is `@trusted` C functions.

On a function signature without a body, `@safe` and `@trusted` mean the same thing: verified manually as safe.

radcapricorn: > Bad example. A program that exhibits undefined behavior might write garbage to the database, but there are also @safe ways of doing that. Preventing wrong data is not the purpose of @safe.

I get a strong feeling that what I'm not communicating clearly enough. Why do you guys keep bringing up conflation of bugs and `@safe`?!? Yes there are ways to write bugs in `@safe` code. That's not the issue here. The issue is that a function that can produce unpredictable results should not be allowed to be called by `@safe` code without the programmer's knowing about it. Especially a function that the D compiler has no authority over.

That, or the spec is completely outdated now, in which case we have much bigger problems than this PR.

aG0aep6G: > Marking arbitraty C functions `@trusted` is just a hack to allow calling them from `@safe` D functions, nothing more.

Marking them `@safe` does the same thing. You seem to have a slightly wrong understanding of `@trusted`. `@trusted` functions are supposed to be 100% safe, just like `@safe` functions.

An example you made earlier:

    iMadeSureYouWillReturnValidData();
    return () @trusted { return iDoNotGuaranteeIReturnValidData(); } ();

The `@trusted` part is not actually safe. It needs the call to `iMadeSureYouWillReturnValidData` to be safe, so that call must be in the `@trusted` enclosure:

    return () @trusted {
        iMadeSureYouWillReturnValidData();
        return iDoNotGuaranteeIReturnValidData();
    } ();

radcapricorn: I'm sorry, what? I have a slightly wrong understanding of trusted?

> Trusted functions are guaranteed by the programmer to not exhibit any undefined behavior if called by a safe function.

Are we talking about the same `@trusted` here? This is getting just silly now.

aG0aep6G: > I'm sorry, what? I have a slightly wrong understanding of trusted?

Yes, I think you might have. You seem to see a difference between `@safe` and `@trusted` on function prototypes. But there's isn't one. They have the same effect and the same meaning. The difference between `@safe` and `@trusted` is whether the compiler checks the body of the function. Without a body, the difference disappears. 

> > Trusted functions are guaranteed by the programmer to not exhibit any undefined behavior if called by a safe function.

That's right. But you break that rule in your iDoNotGuaranteeIReturnValidData example.

radcapricorn: How is @WalterBright's response in 18016 relevant here? He wants to make void initialization in D code "implementation defined", whatever that means. These C functions

>  may behave in ways that do not correspond to any of the locales active during their execution, but an unpredictable mix thereof

IOTW, their results may be undefined. There, I'v quoted that one more time.

schveiguy: Yes, `@trusted` functions are manually checked for safety by the programmer, `@safe` functions are mechanically checked by the compiler. As @aG0aep6G says, marking a prototype as `@safe` or `@trusted` is identical in terms of where it can be called, but the `@safe` attribute annotates that it was mechanically checked, when obviously it wasn't (as a C compiler doesn't have this feature).

The reason you need `@trusted` is specifically to call `@system` functions in a way that is safe, basically to bridge the barrier between safe and system code.

> The issue is that a function that can produce unpredictable results should not be allowed to be called by `@safe` code without the programmer's knowing about it.

I'm satisfied that this shouldn't be `@trusted` for the simple fact that it's using `__gshared` data in a way that would be rejected in `@safe` code. But strictly speaking, unpredictable results does NOT rule out code being `@trusted` or `@safe`. As I demonstrated with the `= void` sample. And according to the linked issue (thanks for bringing that up), [Walter agrees](https://issues.dlang.org/show_bug.cgi?id=18016#c4).

radcapricorn: I see what you're saying now, yes, indeed, a function that would be `@safe` in D should be `@trusted` in C. I guess I didn't think that one through, thanks.

@aG0aep6G there's no difference whether I wrap both calls into a `@trusted` delegate, or leave the first call out of it. Precondition is satisfied in both cases. That `@trusted` delegate cannot exhibit undedined behavior because I manually verified that. In that case, by calling another function.
In fact, leaving the first call **out** of that `@trusted` delegate is safer in the general case. If it's safety is inferred (i.e. it's a template), later modifications may silently demote it to `@system` without the programmer realizing it.
Thing is, with this specific set of C functions, the only way to "satisfy" the requirement is to, well:
1) Don't modify environment/locale concurrently with these calls.
2) Document that fact at the call site.
As in, there's no magical `iMadeSureYouWillReturnValidData()` here.

schveiguy: > IOTW, their results may be undefined

I disagree with that assessment. Unpredictable doesn't mean undefined behavior. Random numbers (true random numbers) are unpredictable, yet the compiler can't take liberty to do anything it wants because of that (undefined behavior does give it that leeway).

> If you have your own opinions about `@safe`, and as a "code owner" for this module, why, exactly, did you not close the original bug report?

I didn't comment here to review the PR exactly, I wanted to clarify the meaning of `@safe`, as it seems to be misunderstood in the prior comments. The code owner thing is nebulous, several people are code owners of all the stdc stuff, we don't all know everything about all the modules. I had no idea about the bug or the possible issues until you brought them up.

In terms of closing the bug report, I think this PR will do that just fine, and the only thing I'd suggest is to remove the `@safe` attribute on the one function (`@trusted` at the top is good enough). I think it's fine the way it is (let all the others become `@system`), and it's good to add all the scopes and return attributes.

radcapricorn: This could've been done 2 days ago. Let's in the future leave the side debates to the forums, pretty please.

aG0aep6G: > In fact, leaving the first call **out** of that `@trusted` delegate is safer in the general case. If it's safety is inferred (i.e. it's a template), later modifications may silently demote it to `@system` without the programmer realizing it.

I see where you're coming from. I've had this discussion before (e.g. [here](https://forum.dlang.org/post/pqit2k$1k96$1@digitalmars.com)) and your understanding of `@trusted` is not uncommon. I still think it's not the intented interpretation and that it's ultimately less useful than a stronger one where functions may only be marked as `@trusted` when any possible call to them is safe.

The problem with your `@trusted` is that later modifications to the `@safe` part of the code may break memory safety. One might accidentally skip over the `iMadeSureYouWillReturnValidData` call, for example. That's exactly what's not supposed to be happen. Worrying about memory safety is the one thing you shouldn't have to do when editing `@safe` code.

Apple's ["goto fail"](https://en.wikipedia.org/wiki/Unreachable_code#goto_fail_bug) comes to mind. Applied to your example, there could be similar bug in the `@safe` part of the code:

    if (cond)
        iMadeSureYouWillReturnValidData();
        return () @trusted { return iDoNotGuaranteeIReturnValidData(); } ();

Now, if `cond` is false, memory safety is broken. No `@trusted` or `@system` code was touched. Whoops.

schveiguy:
I re-triggered the build, not sure why the style check failed, I don't see any white space problems.

thewilsonator:
For reference, so is `dlang/ci`.

thewilsonator:
Hmm, linux32 OOM'd @radcapricorn you'll need to give this a kick.

wilzbach:
The auto-tester will automatically restart itself...

wilzbach:
Besides if it passed before on the auto-tester on the same machine and that's the only failure and it's obviously unrelated, feel free to force merge (though do this with care).

schveiguy:
@thewilsonator I saw the rebuild button, I pushed it. Sorry ;)

thewilsonator:
NP, you know for next time ;)

PR #2374, opened 2018-11-24
Tentative fix for Issue 7270 - needless qualifiers in `TypeInfo.toString`

radcapricorn:
I don't like this approach too too much, but something needs to be done, so it's a starting point. That issue is louder than it appears. If `typeid(X) == `typeid(Y)`, then it better be that `typeid(X).toString() == typeid(Y).toString()`.
I'm open to suggestions on other approaches.
Need more tests. And feedback!

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -1028,6 +1028,14 @@ struct OffsetTypeInfo
     TypeInfo ti;        /// TypeInfo for this member
 }
 
+enum ToStringContext
+{
+    none    = 0x00,
+    const_  = 0x01,
+    shared_ = 0x02,
+    inout_  = 0x04,
+}
+
 /**
  * Runtime type information about a type.
  * Can be retrieved for any type using a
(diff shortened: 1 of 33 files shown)


n8sh:
A possible problem with fixing 7270 is that base TypeInfo.opEquals uses string equality.

https://github.com/dlang/druntime/blob/e7981859137ad5b0c8b6314d04823b8746481859/src/object.d#L1060-L1070

radcapricorn:
That's actually a requirement for fixing it :)

src/object.d
```diff
@@ -1028,18 +1028,31 @@ struct OffsetTypeInfo
     TypeInfo ti;        /// TypeInfo for this member
 }
 
+enum ToStringContext
```
thewilsonator: DDOC comments

radcapricorn: Yup, will do.

PR #2375, opened 2018-11-25
Fix Issue 19433 - Don't consume --DRT-* options if rt_cmdline_enabled is false

CyberShadow:


rainers:
Maybe update the documentation in rt.config, too: "Command line options starting with --DRT- are filtered out before calling main, so the program will not see them."

Geod24:
@CyberShadow : I was looking at fixing dlang/dub#1280 but I don't see how this can be used to fix the issue.
I can see 3 possible cases (assuming we're running `dub` from a package):
- `dub --DRT-gcopt:whatever -- ProgramArguments` (Configure dub)
- `dub -- --DRT-gcopt:whatever ProgramArguments` (Configure the package)
- `dub --DRT-gcopt:whatever -- --DRT-gcopt:whatever ProgramArguments` (Configure both)

Currently we're only able to configure `dub` (option 1).
Using the functionality of this PR, we're able to configure the package (option 2).
However, I don't see how this allow achieving all 3 options.

Correct me if I'm missing something, but using the end of option (`--`) POSIX recommendation would allow us to achieve this. It also has the nice advantage of being composable, e.g.:
``` console
# Pseudo code, but this should work
dub --DRT-gcopt=OptionForDUB run rdmd -- --DRT-gcopt=OptionForRDMD main.d -- --DRT-gcopt=OptionForProgram arg1 arg2...
```

CyberShadow:
Stopping command-line processing in Druntime at `--` sounds like a good idea to me and would be orthogonal to this change. Though, I can't think of many use cases when you would want to separately configure both D programs (where one runs the other using its command-line arguments).

Geod24:
Submitted as https://issues.dlang.org/show_bug.cgi?id=20459

PR #2377, opened 2018-11-26
Add length, lower and upper bounds to RangeError/_d_arraybounds

thewilsonator:


src/core/exception.d
```diff
@@ -47,24 +47,34 @@ else:
  */
 class RangeError : Error
 {
-    @safe pure nothrow this( string file = __FILE__, size_t line = __LINE__, Throwable next = null )
+    @safe pure nothrow this( string file = __FILE__, size_t line = __LINE__, Throwable next = null)
+    {
+        this( 0, 0, 0, file, line, next);
+    }
+    @safe pure nothrow this( size_t lwr, size_t upr, size_t len,
```
jacob-carlborg: `lwr` is not used.

src/core/exception.d
```diff
@@ -47,24 +47,34 @@ else:
...
+                             string file = __FILE__, size_t line = __LINE__, Throwable next = null)
     {
         super( "Range violation", file, line, next );
+        lower = len;
```
jacob-carlborg: `len` is used twice.

src/core/exception.d
```diff
@@ -47,24 +47,34 @@ else:
...
+        upper = upr;
+        length = len;
     }
+    size_t lower, upper; // lower/upper bound passed, if slice (array[lower .. upper]), if not a slice (array[upper])
+    size_t length;       //length of the indexed array
```
jacob-carlborg: * Instance variables should go before an methods/constructors
* `const`/`immutable` as well?

thewilsonator:
Hmm, test is being compiled with `-release` which is suppressing the call to `_d_array_bounds`.

jmdavis:
> Hmm, test is being compiled with -release which is suppressing the call to _d_array_bounds.
e: Darn, looks like the only way around this is to change the makefiles, urgh.

Mark the `unittest` as `@safe`.

thewilsonator:
Thanks!

thewilsonator:
Hmm, nope: `Error: can only catch class objects derived from Exception in @safe code, not core.exception.RangeError`

thewilsonator:
Ahh, `@safe` lambdas work.

jmdavis:
> Hmm, nope: Error: can only catch class objects derived from Exception in @safe code, not core.exception.RangeError

That's nice. I wasn't expecting that to be the case, but it really should be.

> Ahh, @safe lambdas work.

Yeah, if you can't make the whole thing `@safe`, then making the pieces you need `@safe` via lambdas should get around it. Either way, since -release should never be stripping out array bounds checking in `@safe` code, as long as the portions of the tests indexing arrays out-of-bounds are `@safe`, the checks should still be in place. If not, then it's a compiler bug.

thewilsonator:
This doubles as a nice test for that then.

src/core/exception.d
```diff
@@ -47,29 +47,81 @@ else:
  */
 class RangeError : Error
 {
-    @safe pure nothrow this( string file = __FILE__, size_t line = __LINE__, Throwable next = null )
+    // lower/upper bound passed, if slice (array[lower .. upper]), if not a slice (array[upper])
```
jacob-carlborg: Is there a reason these are not Ddoc comments?

src/core/exception.d
```diff
@@ -47,29 +47,81 @@ else:
...
+    size_t lower, upper;
+    //length of the indexed array
+    size_t length;
```
jacob-carlborg: I don't see a reason why these need to be mutable.

src/core/exception.d
thewilsonator: Because I suck at DDOC and this is still a WIP ;)

jacob-carlborg: It doesn't have the `wip` label 😉.

src/core/exception.d
thewilsonator: I don't see any benefit of them being const or immutable, neither file, line nor message are.

src/core/exception.d
thewilsonator: OK good to know.

src/core/exception.d
jacob-carlborg: They probably should have been, but that's probably too late to change now. That doesn't justify that we cannot improve the other fields.

wilzbach: It shows intent and will make debugging easier.

thewilsonator:
(I've altered the testes to figure out what indices are passed to `_d_arraybounds`.)

thewilsonator:
`core.exception.RangeError@src/core/exception.d(137): Range violation: indicies [6250512 .. 6250512] exceeds array length 6250512`
Yeah that looks like garbage arguments.

src/core/exception.d
```diff
@@ -47,29 +47,121 @@ else:
...
+        sink("@"); sink(file);
+        sink("("); sink(unsignedToTempString(line, tmpBuff, 10)); sink(")");
+
+        sink(": "); sink(msg);
+        if (upper)
+        {
+            sink(lower ? ": indicies [" : ": index [");
+            if (lower)
+            {
+                sink(unsignedToTempString(length, tmpBuff, 10)); sink(" .. ");
```
adamdruppe: why are you printing length there? shouldn't it be lower?

src/core/exception.d
```diff
@@ -47,29 +47,121 @@ else:
...
+            }
+            sink(unsignedToTempString(length, tmpBuff, 10));
```
adamdruppe: and this should be upper

adamdruppe:
I think you were over-zealous with the copy/paste of your print code rather than the arguments being garbage.

That length does seem a bit odd, but it being repeated is just it printing the same variable three times.

thewilsonator:
Hmm, 
```d
int[] a = [1, 2, 3];
() @safe
{
      int[] bad = a[1 .. 7]; // __LINE__ == 137
      assert(0);
}();
```

`core.exception.RangeError@src/core/exception.d(137): Range violation: index [140723719909984] exceeds array length 140723719909984
` so yeah lower index is zero (wrong), upper index is garbage.

adamdruppe:
Even with your druntime code, it works for me:

```
core.exception.RangeError@test.d(5): Range violation: indicies [1 .. 7] exceeds array length 3
----------------
??:? _d_arrayboundsp [0x427dc9]
??:? pure nothrow @nogc @safe void test.main().__lambda1() [0x427c5e]
??:? _Dmain [0x427c26]
```

I have tried a bunch of combinations, -m32, -m64, -inline, -O, -release, -unittest, and it consistently works. Have you tried it locally?

Are you sure the test runner is getting the correct dmd branch? Looking at one of these links: https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3442742&dataid=24442103&isPull=true it looks like it pulled master of all three, then only merged this druntime branch. I see no evidence the dmd branch is actually correct.

thewilsonator:
No I haven't tested it locally, but you're right, it does seem to be pulling dmd `master` rather than `fix_range_error`, which is weird because I had problems of it accidentally using matching branches when I didn't want that.

adamdruppe:
So, how do we convince the autotester to do this right?

wilzbach:
I am not sure there's a way to modify the Auto-Tester. I could fix Buildkite to detect the branches properly though.

wilzbach:
For other readers's convenience: https://github.com/dlang/dmd/pull/9015

thewilsonator:
This needs a review.

wilzbach:
(and a rebase to restart the CIs which I just did)

src/core/exception.d
```diff
@@ -47,29 +47,121 @@ else:
...
         assert(re.line == 42);
         assert(re.next !is null);
         assert(re.msg == "Range violation");
     }
+
+    {
+        int[] a = [1, 2, 3];
+        //try
+        //{
+            // This is necessary becase the unittests are compiled with -release
```
wilzbach: Nope, we have have `unittest-debug` and `unittest-release` targets (both are run when you run `unittest`). Simply version these test out for release builds.

thewilsonator:
That'd be because I forgot to git add.

thewilsonator:
This should be good to go. Please review :)

thewilsonator:
Gah, WTF is circle doing?

jacob-carlborg:
Where do we have the changelog entry for this?

wilzbach:
> Gah, WTF is circle doing?

Read the error log:

```
changelog/range-error.dd:3:Errors resulting from bad indexing will now contain the length of 
changelog/range-error.dd:4:the array that was sliced as well as the offending index for `arr[bad]` 
changelog/range-error.dd:5:or for bad slices `arr[bad1 .. bad2]`. The associated runtime functions 
posix.mak:345: recipe for target 'checkwhitespace' failed
```

src/core/exception.d
```diff
@@ -47,29 +47,121 @@ else:
...
+            {
+                sink("\n----------------");
+                foreach (t; info)
+                {
+                    sink("\n"); sink(t);
+                }
+            }
+            catch (Throwable)
+            {
+                // ignore more errors
```
wilzbach: Very bad idea to be a black hole. This will be very hard to debug.

changelog/range-error.dd
```diff
@@ -0,0 +1,6 @@
+RangeError now has bounds information
+
+Errors resulting from bad indexing will now contain the length of 
+the array that was sliced as well as the offending index for `arr[bad]` 
+or for bad slices `arr[bad1 .. bad2]`. The associated runtime functions 
+`onRangeError` and `_d_arraybounds[p]` have changed signature to match.
```
wilzbach: Have a look at https://github.com/dlang/druntime/pull/2413 for an example of nice error messages.

I would do the following:
- simple example of mis-indexing
- use the `$(CONSOLE)` macro to show this new feature of

thewilsonator:
>Read the error log:

I meant before I added the changelog. It was failing in object.d unit tests for destroy or something.

thewilsonator:
>Very bad idea to be a black hole. This will be very hard to debug.

Take that up with [object.d](https://github.com/dlang/druntime/blob/424640864c2aa001731467e96f637bd3e704e481/src/object.d#L2804) where this was taken from. If you have suggestions, I'm all ears but I think that should be for a later PR.

changelog/range-error.dd
```diff
@@ -0,0 +1,37 @@
...
+
+For example, currently the following file:
+
+---
+void main()
+{
+    int[] a = [1,2,3];
+    int b = a[7];
+    assert(a == b);
```
wilzbach: I don't think this assert would compile.
Also, do we even need it?

thewilsonator:
CircleCI fails with 
```
****** FAIL debug64 object
core.exception.AssertError@src/object.d(737): unittest failure
```

which is [here](https://github.com/dlang/druntime/blob/master/src/object.d#L737), have absolutely no idea why.

thewilsonator:
I think CircleCI runs some things that the autotester doesn't but I'm pretty sure that both of the run the unto tests for druntime, which is why I'm very confused as to why it is failing in some completely unrelated code.

jacob-carlborg:
Compatibility with visual studio?

thewilsonator:
CircleCI reports its triple as `x86_64-pc-linux-gnu`.

src/core/exception.d
```diff
@@ -690,15 +785,15 @@ extern (C)
...
-    void _d_arrayboundsp(immutable(char*) file, uint line)
+    void _d_arrayboundsp(immutable(char*) file, uint line, size_t lower, size_t upper, size_t length)
     {
         import core.stdc.string : strlen;
-        onRangeError(file[0 .. strlen(file)], line);
+        onRangeError(lower, upper, length, file[0 .. strlen(file)], line);
     }
 
-    void _d_arraybounds(string file, uint line)
+    void _d_arraybounds(string file, uint line, size_t lower, size_t upper, size_t length)
```
ibuclaw: Rather than changing the signatures of the public extern C interface, new functions should be introduced.

jacob-carlborg:
> CircleCI reports its triple as x86_64-pc-linux-gnu.

Right, not sure what I was thinking.

thewilsonator:
@ibuclaw signature of public functions is now compatible with default arguments. The functions you commented in are compiler generated.

ibuclaw:
They will still silently break with applications linked to previous libraries. Just questioning whether this warrants a backwards incompatible ABI version bump.

jacob-carlborg:
We don't guarantee any ABI compatibility as far as I know.

thewilsonator:
Hey, it passes on CircleCI! Yay! I've no idea why that was failing before.

src/core/exception.d
```diff
@@ -47,29 +47,121 @@ else:
...
+
+        if (info)
+        {
+            try
+            {
+                sink("\n----------------");
+                foreach (t; info)
+                {
+                    sink("\n"); sink(t);
+                }
```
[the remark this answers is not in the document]
thewilsonator: I'm not sure how to inject just a few `sink` calls when the function is overridden.

changelog/range-error.dd
```diff
@@ -0,0 +1,36 @@
...
+    int b = a[7];
+}
+---
+
+would yield this error when compiled and run:
+
+$(CONSOLE
+> dmd -run main.d
+core.exception.RangeError@onlineapp.d(4): Range violation
+$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)
```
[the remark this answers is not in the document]
thewilsonator:
```
core.exception.RangeError@onlineapp.d(4): Range violation
----------------
??:? _d_arrayboundsp [0xd1f573cd]
```

16 `-`s, but who's counting?

src/core/exception.d
```diff
@@ -47,29 +47,121 @@ else:
...
+            }();
+        }
+        catch (RangeError re)
+        {
+            assert(re.line   == __LINE__ - 6);
+            assert(re.lower  == 0);
+            assert(re.upper  == 9);
+            assert(re.length == 3);
+        }
+    }
```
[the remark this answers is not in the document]
thewilsonator: Test added.

src/core/exception.d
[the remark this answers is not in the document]
CyberShadow: What do you think about a free or private final function, like

```d
void formatThrowable(Sink sink, void delegate(Sink) formatMessage)
{
    // put into sink first common part of throwable (file name etc.)...
    formatMessage(sink);
    // put into sink last common part of throwable (stack trace)...
```
?

src/core/exception.d
thewilsonator: For reference the DMD patch is already in, hence why I want to get this in. Also as noted by Jacob, we don't guarantee binary compatibility across releases.

ibuclaw: See `_aaGet`, `_aaGetX`, `_aaGetY` for examples of doing this in the past.  Infact pick any function that ends in X (`_d_arrayliteralX`)

Even `_d_arrayboundsp` was a new name given to replace `_d_arrayboundsm`. :-)

wilzbach: But what's the point of doing so and creating more legacy cruft if we don't we don't guarantee binary compatibility across releases anyhow?

src/core/exception.d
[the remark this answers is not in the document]
thewilsonator: Done.

CyberShadow:
Suggested fix:

- Introduce new C functions as suggested by @ibuclaw 
- Store separately whether we have zero (old API), one (index) or two (slice) indices
- Format accordingly to number of indices we have

src/core/exception.d
ibuclaw: - Gives compiler vendors time to adjust their codegen and test the new api.
- Doesn't break CI (druntime functions should have been applied first before compiler changes).
- Doesn't break git bisect (Digger) infrastructure.

wilzbach: > Gives compiler vendors time to adjust their codegen and test the new api.

Fair point, but wouldn't this apply to new changes?

> Doesn't break CI (druntime functions should have been applied first before compiler changes).

It passed on all CIs.

> Doesn't break git bisect (Digger) infrastructure

Digger is used to build DAutoTest, so that's tested too.

CyberShadow: > It passed on all CIs.

Question: do you know *why* it passes? Looking at this and the compiler change, I do not.

thewilsonator: Call parameter stack layout: adding args to the end doesn't affect previous ones. I.e. passing too many will ignore the later ones, whereas passing too few will lead to garbage/corruption.

thewilsonator:
a,b,c and e are edge cases, I will fix those soon. 

d while potentially misleading,  cross-referencing the line number will show a slice.

CyberShadow:
> a,b,c and e are edge cases, I will fix those soon.

Not sure what you mean by fix - a and c look correct, they are there for comparison.

I don't think there's a reasonable way to fix this in general without additional state / transmitted information. There are no free bits to hold information on whether the indices are actually indices, and the two indices don't necessarily have a relation to each other.

> d while potentially misleading, cross-referencing the line number will show a slice.

That doesn't sound great. It's the worst case, and it might not be the only indexing operation on the line.

thewilsonator:
> Not sure what you mean by fix - a and c look correct, they are there for comparison.
 
My bad about a, the rest were wrong and I assumed that was too. c, unless you mispasted, is missing the the length of the array at the end.

CyberShadow:
Yes, that was a mispaste, sorry.

src/core/exception.d
ibuclaw: x86 is the only target that pushes arguments in reverse.  All others are left to right.

I have an ARM and Aarch64 server that I can run a test on. Regressions on those targets could warrant a revert... Did the druntime changes end up in a release?

ibuclaw: Assuming that it's a dmd backend change only though, it's not so critical.  If array bounds checks were being generated by the frontend, then it's another matter.

thewilsonator: It is a backend thing.

dkorpel:
I would love to get this in. Are you still going to work on this @thewilsonator?

thewilsonator:
not with any sense of urgency.

12345swordy:
How much left is needed for this to be done?

Geod24:
Closing for lack of activity (and to clean up upstream branches). The enhancement would be amazing to have, though.

PR #2379, opened 2018-11-28
Use monotonic clock if available

kubo39:
This changes the behaivior that thread waiting on condition variable
is woken up too late when system is modified backwards.


wilzbach:
This should appear in the changelog, so please either reference an issue or add a changelog entry. See the bot's message for details.

Geod24:
It doesn't change anything since it's in a `static if (false && ....)`

PR #2380, opened 2018-11-28
[DO-NOT-MERGE] debug buildkite style failure

wilzbach:
Looks like `$(CWS_MAKEFILES)` is undefined.


wilzbach:
Okay I'm getting closer and it looks like `git ls-tree --name-only -r HEAD` is not returning any files on Buildkite.

wilzbach:
So the `style` failure is due to no `.git` directory existing on Buildkite.
Fix: https://github.com/dlang/ci/pull/346

PR #2384, opened 2018-11-30
fix issue 19455 - GC wastes too much memory

rainers:
add bins with about 3/2 of existing sizes

As @kinke noticed when enabling the GC for dmd (see https://github.com/ldc-developers/ldc/pull/2916), about 40% more memory is used than with the bump-pointer-allocation. This PR should drop the overhead to approximately 15% (similar to the overhead when using C malloc).

The example in the bugzilla report drops from 247 MB to 176 MB.

Unfortunately, std.experimental.gc_allocator makes some assumptions about the GC that are not guaranteed and are only valid for the existing GC (and might also break if you use another replacement GC).

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -2284,14 +2284,14 @@ struct Gcx
                     {
                         immutable size = binsize[bin];
                         void *p = pool.baseAddr + pn * PAGESIZE;
-                        void *ptop = p + PAGESIZE;
+                        void *ptop = p + PAGESIZE - size;
                         immutable base = pn * (PAGESIZE/16);
                         immutable bitstride = size / 16;
 
                         bool freeBits;
                         PageBits toFree;
 
-                        for (size_t i; p < ptop; p += size, i += bitstride)
+                        for (size_t i; p <= ptop; p += size, i += bitstride)
                         {
                             immutable biti = base + i;
 
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -942,25 +942,11 @@ class ConservativeGC : GC
         {
             p = sentinel_sub(p);
             size_t size = gcx.findSize(p);
-
-            // Check for interior pointer
-            // This depends on:
-            // 1) size is a power of 2 for less than PAGESIZE values
-            // 2) base of memory pool is aligned on PAGESIZE boundary
-            if (cast(size_t)p & (size - 1) & (PAGESIZE - 1))
-                size = 0;
             return size ? size - SENTINEL_EXTRA : 0;
         }
         else
         {
             size_t size = gcx.findSize(p);
-
-            // Check for interior pointer
-            // This depends on:
-            // 1) size is a power of 2 for less than PAGESIZE values
-            // 2) base of memory pool is aligned on PAGESIZE boundary
-            if (cast(size_t)p & (size - 1) & (PAGESIZE - 1))
-                return 0;
             return size;
         }
     }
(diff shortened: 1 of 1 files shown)


wilzbach:
> Unfortunately, std.experimental.gc_allocator makes some assumptions about the GC that are not guaranteed

It's experimental and can be changed without notice at any time if you want to.

rainers:
> It's experimental and can be changed without notice at any time if you want to.

As the GC can be changed when linking or via a runtime switch, gc_allocator.goodAllocSize needs to forward the request to the GC. That probably means another method added to the GC interface :-/

thewilsonator:
@rainers please give this a kick.

rainers:
Added [WIP] as I noticed some misbehaviour that has to be checked.

src/gc/impl/conservative/gc.d
```diff
@@ -2265,14 +2284,14 @@ struct Gcx
...
-                        void *ptop = p + PAGESIZE;
+                        void *ptop = p + PAGESIZE - size;
                         immutable base = pn * (PAGESIZE/16);
                         immutable bitstride = size / 16;
 
                         bool freeBits;
                         PageBits toFree;
 
-                        for (size_t i; p < ptop; p += size, i += bitstride)
+                        for (size_t i; p <= ptop; p += size, i += bitstride)
```
thewilsonator: Assuming that the `<=` is intentional. Please add a comment saying it is, I treat `<=` used as a loop bound as very suspicious.

src/gc/impl/conservative/gc.d
```diff
@@ -2356,7 +2375,7 @@ struct Gcx
 
                 Lnotfree:
                     p = pool.baseAddr + pn * PAGESIZE;
-                    for (u = 0; u < PAGESIZE; u += size)
+                    for (u = 0; u <= PAGESIZE - size; u += size)
```
thewilsonator: Ditto

src/gc/impl/conservative/gc.d
```diff
@@ -3228,10 +3255,10 @@ struct SmallObjectPool
         // Convert page to free list
         size_t size = binsize[bin];
         void* p = baseAddr + pn * PAGESIZE;
-        void* ptop = p + PAGESIZE - size;
+        void* ptop = p + PAGESIZE - 2 * size;
         auto first = cast(List*) p;
 
-        for (; p < ptop; p += size)
+        for (; p <= ptop; p += size)
```
thewilsonator: Ditto

BBasile:
dparse use a fork that has derived since phobos 2.077. Maybe some backports are necessary if the phobos gc_allocator was modified.

rainers:
@BBasile Thanks for quick action.

BBasile:
One more PR is needed actually. This time it's slower because DUB website has to detect the stdx alloc tag.

BBasile:
problem solved for good now.

BBasile:
I think this is good for merge now @rainers, isn't it ? Since you removed the WIP label.

rainers:
> I think this is good for merge now @rainers, isn't it ?

Yes, I think so. Let's hope it doesn't break anything...

ZombineDev:
Great work, @rainers! Can you add a changelog entry detailing the memory usage improvements?

rainers:
> Let's hope it doesn't break anything...

Here's the first one: https://github.com/dlang/druntime/pull/2417

rainers:
> Can you add a changelog entry detailing the memory usage improvements?

I'm hoping to get a couple more GC related PRs merged. That might warrant a summary of the changes.

ZombineDev:
Ok, sounds good.

nordlow:
Was wasn't 24 include in the size set?

kinke:
24 isn't a multiple of 16 bytes, so that would have possibly reduced the guaranteed alignment from 16 to 8.

nordlow:
Ok, thanks.

PR #2385, opened 2018-12-01
add appveyor test for VS2017

rainers:


appveyor.yml
```diff
@@ -0,0 +1,52 @@
+
+os: Visual Studio 2017
+
+environment:
+  matrix:
+    - ARCH:                x64
```
jacob-carlborg: * This does not seem to be used
* If you want to keep this, it seems like `platform` would be a more suitable than a custom environment variable

rainers: I just copied it from dmd's appveyor.xml, I'll remove it.

thewilsonator:
You'll need to copy across `appveyor.sh` for `gmake` as well and add the cache to `appveyor.yml`.

rainers:
> You'll need to copy across `appveyor.sh` for `gmake` as well and add the cache to `appveyor.yml`.

gmake is already preinstalled in appveyor, it's actually not necessary to build it. 

I was currently not planning to run the dmd-tests here, but only druntime (maybe phobos, too).

As it turned out, msvc-dmc doesn't work with VS2017 (assumes old directory structure), so I switched to Visual D as the build system.

thewilsonator:
Thanks!

PR #2386, opened 2018-12-01
POSIX `free` purity

radcapricorn:


thewilsonator:
Hmm, not sure if the FreeBSD is related or not, looks like it could be.

n8sh:
If a `pure` function is explicitly `@system` is it acceptable for it to violate purity if used improperly (violating conditions explained in the documentation that the compiler does not check and the programmer is solely responsible for ensuring)? The answer might reasonably be "yes" since a `@system` function might arbitrarily corrupt memory if called with improper arguments. If yes this PR should go ahead, otherwise it should be closed.

PR #2390, opened 2018-12-02
set WCHAR to wchar_t for future

WalterBright:
Part of transition to C++11 types. Blocked by https://github.com/dlang/dmd/pull/9029

### DIFF
--- a/src/core/sys/windows/winnt.d
+++ b/src/core/sys/windows/winnt.d
@@ -21,7 +21,22 @@ FIELD_OFFSET(t,f), CONTAINING_RECORD(address, type, field)
 
 alias void   VOID;
 alias char   CHAR, CCHAR;
-alias wchar  WCHAR;
+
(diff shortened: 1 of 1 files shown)


WalterBright:
> What does C++11 compatibility have to do with the Windows API?

Microsoft changed the definition of WCHAR to wchar_t. Meaning D cannot connect to C++ code that uses WCHAR. It'll get worse with increasing use of char16_t.

TurkeyMan:
They didn't 'change' the definition of `WCHAR` to `wchar_t`, that's just how it is for C++. For C, it's defined as `unsigned short`. It's always been this way.

thewilsonator:
Right, and for C it doesn't matter anyway because there is no mangling.

TurkeyMan:
And for MS code, it just so happens that `wchar_t` matches `unsigned short`, and also that `WINAPI` is `extern(C)`, not C++... but the type definitions and value of the macros has never changed.

PR #2391, opened 2018-12-03
prepare std.wchar_ for new definition of wchar_t

WalterBright:
By adding overloads for the old definition.
Blocked by https://github.com/dlang/dmd/pull/9029

PR #2392, opened 2018-12-03
core/runtime.d: use WCHAR to interface to Windows

WalterBright:


CyberShadow:
I don't understand what this PR changes at all.

Is `wchar_t*` no longer implicitly convertible to `WCHAR*`? If so, why?

WalterBright:
The trouble is that, for Windows, wchar_t and wchar are used interchangeably. Unfortunately, going forward, they will mangle differently, because C++11 unhelpfully mangles wchar_t and char16_t differently.

This is not directly a problem with the Windows API, as that is a C API that does not depend on mangling. But once you connect with anything else, things break. Hence use WCHAR for all types that interact with Windows rather than wchar or wchar_t.

WalterBright:
For example, it becomes a problem for the "w" functions in core.stdc.wchar_. A wchar* is no longer the same thing as a wchar_t*.

It's best to just keep WCHAR, wchar, and wchar_t each in their own sandbox and don't conflate them.

CyberShadow:
> But once you connect with anything else, things break.

Could you please be more specific here?

I understand how this affects code that links with C++, but I don't see why this needs to affect anything D-related at all.

Just make `wchar` and `wchar_t` (and pointers to them) cast to each other implicitly (when they have the same size).

WalterBright:
> Just make wchar and wchar_t (and pointers to them) cast to each other implicitly (when they have the same size).

That's @thewilsonator 's idea. Problems:

1. C++ distinguishes char16_t* and wchar_t*. They don't implicitly convert. They have different mangling.
2. If they were the same only different, then you couldn't overload `f(char16_t*)` against `f(wchar_t*)`.

Besides, every time I try to slide on C++ compatibility with this nonsense, I get a plethora of complaints about it, like when VC++ mangles a struct differently from a class. People say it's blockers. They say they cannot modify the C++ code. The only way forward is to be compatible with the C++ stuff. If I was in charge of C++, `wchar_t` would be an alias, not a distinct type. But nobody on the C++ committee wants to hear my opinions :-)

The good news is this only matters when interfacing with C++ code. There's no reason to *ever* use `wchar_t` in strictly D code.

WalterBright:
I also cannot count how many hours I've lost trying to deal with `size_t` on OSX and how it mangles. There's just no way to get things to work cross-platform without making it a unique type, because of C++.

adamdruppe:
Could D's wchar mangle the same way as C++'s wchar_t? Since D's wchar doesn't exist in C++, it has nothing to break for mangling compatibility, and they are the same on a binary level. Then wchar, wchar_t, and WCHAR would all be interchangeable.

thewilsonator:
No, it needs to mangle to `char16_t`.

thewilsonator:
>  Problems:

Gah!

1. Just because C++ does something stupid does't mean we should follow suit.
2. bummer. Although how likely is that to a) occur in the wild, and b) actually matter, surely an overload working on a thing that is essentially the same type would do something compatible... then again it is C++. I think its much more likely to come up with `wstring` vs. `u16string` but those are structs (in D) so incompatibility is par for the course.

CyberShadow:
>     1. C++ distinguishes char16_t* and wchar_t*. They don't implicitly convert. They have different mangling.

I feel like this is C++'s problem, and doesn't necessarily have to be D's problem too.

> If they were the same only different, then you couldn't overload `f(char16_t*)` against `f(wchar_t*)`.

It doesn't sound like we've ran out of options here. What about things like...

- Make `wchar` and `wchar_t` be different types, but just implicitly convert to each other.
- Make the distinction only effective in `extern(C++)` code
- Make the distinction only effective in overload resolution
etc.

What bugs me here is that I don't think we've estimated how much user code this will affect. @alphaglosined pointed out on IRC that we are going to have a lot of fun with https://github.com/dlang/phobos/blob/master/std/stdio.d#L93, for one example.

Perhaps consider one of the above compiler hacks temporarily, for the sole purpose of emitting meaningful deprecation warnings. Once everyone has transitioned to `WCHAR`, the compiler hacks can be removed.

thewilsonator:
> Make wchar and wchar_t be different types, but just implicitly convert to each other.

Thats not the problem and is easy to do, the prob is that pointers to/ arrays of do not. I suppose for the overload case if the pointers implicitly convert you'd just declare only one of the overloads.

CyberShadow:
> I suppose for the overload case if the pointers implicitly convert you'd just declare only one of the overloads.

Then you can't link to the overloads in C++ code.

thewilsonator:
I don't follow. given two overloads in C++

```c++
void f(wchar_t *) {} // Both doing something equivalent
void f(char16_t *) {} // ditto
```

in d you'd declare 
```d
extern(C++) void f(wchar *);
```

and use one of the two.

CyberShadow:
Yes, but what if they *don't* do something equivalent :)

Well, I guess that's even rarer than having to link to `char16_t` overloads.

thewilsonator:
> Yes, but what if they don't do something equivalent :)

Then you probably have bigger problems to deal with ;)

WalterBright:
> Just because C++ does something stupid does't mean we should follow suit.

If we want to interface with C++, yes we do. This has been shown over and over, and I'm tired of doing it your way and getting grief over it :-(

CyberShadow:
Good luck breaking the code of D users who don't care one bit about C++ and not getting grief over it :)

WalterBright:
> we are going to have a lot of fun with https://github.com/dlang/phobos/blob/master/std/stdio.d#L93, for one example.

See https://github.com/dlang/druntime/pull/2391

WalterBright:
> You have said may times that if you your overloads do something radically different you're doing it wrong

That's right. Doesn't seem to help matters, though. And consider:
```
void abc(char16_t);
void abc(wchar_t);
void abc(char32_t);
```
Now what?

> Besides as mentioned above that can be worked around in C++ by adding a function that disambiguates the overload.

Every time I suggest that I'm told they can't change one character in the C++ code.

rikkimax:
After having some time to think about it, I'm concerned that this set of changes may cause regressions with i.e. COM classes. We may want to be opinionated with anything that is ``extern(Windows)`` or ``extern(C++)`` for Windows and throw some warnings out, in the short term till we can confirm that this is not breaking anything.

thewilsonator:
> And consider:

Thats no different to the case above, because `wchar_t` will conflict with only one of the other two overloads. Frankly I think we should consider that if the C++ code we're trying to link to is modern enough to use `char16_t` then they shouldn't be using `wchar_t` (again for C code/the Windows API it doesn't matter).

TurkeyMan:
> Good luck breaking the code of D users who don't care one bit about C++ and not getting grief over it :)

Why is any D programmer or any D program using `wchar_t` in the first place? What use is that type in D beyond C++ compatibility? (which includes the WIN32API)

thewilsonator:
Nil, but this affects users that using the APIs subject to the change i.e. users of the Windows API. They _will_ be affected if we don't make pointers cast implicitly.

TurkeyMan:
Oh. Well, hard to help the mistake that was made up-front where the API binding didn't use WCHAR when it should... correcting broken code is just something that needs to happen.

thewilsonator:
Indeed, but without pointer casting, this affects users who get no benefit, i.e. those who use the windows API, but not c++

CyberShadow:
> Why is any D programmer or any D program using `wchar_t` in the first place?

Keep in mind that `WCHAR` is now defined as `wchar_t`. Anything interacting with the Windows API will now need casts etc., and the usage of `WCHAR` might propagate deeper into D code due to conventions or such.

CyberShadow:
> mistake that was made up-front where the API binding didn't use WCHAR

That's a problem, but not the biggest problem here. 

Aside the Windows API, there's also the "wide" C APIs, see e.g.:
https://github.com/dlang/druntime/pull/2391/files
and the other open PRs to the same tune.

thewilsonator:
I think that particular PR just adds overloads for the different types, e.g. https://run.dlang.io/is/zKTRKe which will help not hinder.

thewilsonator:
Oh right, I though you were against that PR, not the need for compatibility. `wchar`/`wchar_t` is not a problem as it works, `wchar*`/`wchar_t*` is still a problem.

CyberShadow:
> I though you were against that PR

I am against the PR.

> `wchar*`/`wchar_t*` is still a problem.

I meant those. Sorry, I thought it was obvious. None of those functions take just a `wchar_t` (and not a `wchar_t*`).

anon17:
```
void abc(char16_t);
void abc(wchar_t);
void abc(char32_t);
```
```
void abc("1".to_cpp_char16_tz);
void abc("2");
void abc("3".to_cpp_char32_tz);
```
Can be invoked like this.

PR #2393, opened 2018-12-03
stdlib.d - compatibility with new wchar_t

WalterBright:


n8sh:
Do the incompatible function signatures cause problems for LDC?

PR #2394, opened 2018-12-03
posix stdlib, inttypes - compatibility with new wchar_t

WalterBright:


CyberShadow:
These are pretty terrible.

It's even worse thinking that D users will need to do the same for similar code in their local bindings.

n8sh:
Does this cause problems for LDC2 or can this just be merged?

ibuclaw:
This might cause issues with gdc, as C types are specially recognised by the compiler with own version path in core.stdc.config, and may not follow this logic.

Static if would be better, making sure that wchar_t is a distinct type, and not an alias.

rainers:
FWIW: I agree with most that these changes are pretty bad, the cure seems worse than the disease (AFAICT mangling of C++ functions that happen to use wchar_t, too). I think we should further explore a way to define aliases that only affect C++ mangling of the symbol, but not the function type, as if pragma(mangle) is being used.

jacob-carlborg:
> I think we should further explore a way to define aliases that only affect C++ mangling of the symbol, but not the function type, as if pragma(mangle) is being used.

https://github.com/dlang/dmd/pull/7458

My conclusion of that PR:

The issue is, as Rainer pointed out, that the mangling is used for type comparison as well. An alias with pragma(mangle) is not intended to introduce a new type.

jacob-carlborg:
> Maybe your PR can work if it is only applied to C++ symbol mangling. Type comparison happens for D mangling. It should probably use another pragma then.

Possibly.

RazvanN7:
Closing as per @rainers comment and lack of activity.

PR #2395, opened 2018-12-04
Make onRangeError nogc

edi33416:


CyberShadow:
https://issues.dlang.org/show_bug.cgi?id=20857

PR #2396, opened 2018-12-04
[Trivial] Fix core.stdc.errno for dragonflybsd

dkgroot:
[Trivial] Fix core.stdc.errno for dragonflybsd

Old Code would result in:
```
src/core/memory.d(914): Deprecation: core.stdc.errno.getErrno is not visible from module memory
src/core/memory.d(917): Deprecation: core.stdc.errno.setErrno is not visible from module memory
```

### DIFF
--- a/src/core/stdc/errno.d
+++ b/src/core/stdc/errno.d
@@ -102,8 +102,14 @@ else version (FreeBSD)
 }
 else version (DragonFlyBSD)
 {
-    pragma(mangle, "errno") extern int __errno;
-    ref int errno() { return __errno;}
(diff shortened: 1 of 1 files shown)


dkgroot:
@thewilsonator Thanks for setting the auto-merge so quickly. I just ran into a little snag though and had to make a small change (and force pushed it). I was using extern inside and extern(C) block, which caused some issue when -betterC is enabled.

thewilsonator:
Np.

thewilsonator:
Readding auto merge.

PR #2397, opened 2018-12-06
Fix issue 19468: improve cyclic dependency error message.

quickfur:
Make it clear that the dependency cycle is between module ctors/dtors, otherwise users may be misled to fix the non-existent problem of cyclic imports.

### DIFF
--- a/src/rt/minfo.d
+++ b/src/rt/minfo.d
@@ -287,7 +287,7 @@ struct ModuleGroup
             else
                 enum EOL = "\n";
 
-            sink("Cyclic dependency between module ");
+            sink("Cyclic dependency between module constructors/destructors of ");
             sink(_modules[sourceIdx].name);
             sink(" and ");
             sink(_modules[cycleIdx].name);


schveiguy:
@quickfur thx!

src/rt/minfo.d
```diff
@@ -287,7 +287,7 @@ struct ModuleGroup
             else
                 enum EOL = "\n";
 
-            sink("Cyclic dependency between module ");
+            sink("Cyclic dependency between module ctors/dtors of ");
```
jacob-carlborg: Can we please spell out the full words.

quickfur: Done, thanks!

jacob-carlborg: Well, now we got two commits for no good reason :(

quickfur:
Ahhh oops.  Didn't realize @schveiguy already did it. Sorry for overwriting your changes!

jacob-carlborg:
> Ahhh oops. Didn't realize schveiguy already did it. Sorry for overwriting your changes!

No, it's better, now there's only one commit 😃.

src/rt/minfo.d
schveiguy: Haha, @quickfur fixed it :) I just edited it in github. I guess I was a tad quicker.

schveiguy:
> Sorry for overwriting your changes!

No worries, they were exactly the same changes 😄

schveiguy:
Hmmm interesting how github now records force-pushes...

PR #2398, opened 2018-12-07
Move `emplace` to druntime for symmetry with `destroy`.

TurkeyMan:
`destroy()` is in object.d, which makes sense, because one of the most primitive operations you can perform is to cause destruction of an object.
Strangely, we can't construct objects though... because `emplace()` is in phobos, which isn't helpful for projects that don't link phobos. (it's huge, and it's not `nothrow @nogc`)

I think `emplace()` really needs to live right next to `destroy()`; it's a direct counterpart, so here it is.
If this is accepted, we'll make phobos emplace alias this one moved here.

### DIFF
--- a/changelog/emplace.dd
+++ b/changelog/emplace.dd
@@ -1,0 +1,3 @@
+Moved `std.conv.emplace` to core/lifetime.d
+
+`emplace` is the counterpart to `destroy`, so it has been moved to also live in druntime (core/lifetime.d) where it is accessible by projects that use a shallow runtime library stack.
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -17,9 +17,9 @@ template TypeTuple(TList...)
 template FieldTypeTuple(T)
 {
     static if (is(T == struct) || is(T == union))
-    alias FieldTypeTuple = typeof(T.tupleof[0 .. $ - __traits(isNested, T)]);
+        alias FieldTypeTuple = typeof(T.tupleof[0 .. $ - __traits(isNested, T)]);
     else static if (is(T == class))
-    alias FieldTypeTuple = typeof(T.tupleof);
+        alias FieldTypeTuple = typeof(T.tupleof);
     else
     {
         alias FieldTypeTuple = TypeTuple!T;
@@ -153,14 +153,14 @@ template isInnerClass(T) if (is(T == class))
         template hasOuterMember(T...)
         {
             static if (T.length == 0)
-            enum hasOuterMember = false;
+                enum hasOuterMember = false;
             else
-            enum hasOuterMember = T[0] == "outer" || hasOuterMember!(T[1 .. $]);
+                enum hasOuterMember = T[0] == "outer" || hasOuterMember!(T[1 .. $]);
         }
         enum isInnerClass = __traits(isSame, typeof(T.outer), __traits(parent, T)) && !hasOuterMember!(__traits(allMembers, T));
     }
     else
-    enum isInnerClass = false;
+        enum isInnerClass = false;
 }
 
 template dtorIsNothrow(T)
@@ -206,11 +206,11 @@ template anySatisfy(alias F, T...)
 private template maxAlignment(U...)
 {
     static if (U.length == 0)
-    static assert(0);
+        static assert(0);
     else static if (U.length == 1)
-    enum maxAlignment = U[0].alignof;
+        enum maxAlignment = U[0].alignof;
     else static if (U.length == 2)
-    enum maxAlignment = U[0].alignof > U[1].alignof ? U[0].alignof : U[1].alignof;
+        enum maxAlignment = U[0].alignof > U[1].alignof ? U[0].alignof : U[1].alignof;
     else
     {
         enum a = maxAlignment!(U[0 .. ($+1)/2]);
@@ -274,8 +274,8 @@ template hasElaborateAssign(S)
     else static if (is(S == struct))
     {
         enum hasElaborateAssign = is(typeof(S.init.opAssign(rvalueOf!S))) ||
-        is(typeof(S.init.opAssign(lvalueOf!S))) ||
-        anySatisfy!(.hasElaborateAssign, FieldTypeTuple!S);
+                                  is(typeof(S.init.opAssign(lvalueOf!S))) ||
+                                  anySatisfy!(.hasElaborateAssign, FieldTypeTuple!S);
     }
     else
     {
--- a/src/core/lifetime.d
+++ b/src/core/lifetime.d
@@ -1,0 +1,1294 @@
+module core.lifetime;
+
+/+
+emplaceRef is a package function for druntime internal use. It works like
+emplace, but takes its argument by ref (as opposed to "by pointer").
+This makes it easier to use, easier to be safe, and faster in a non-inline
+build.
+Furthermore, emplaceRef optionally takes a type parameter, which specifies
+the type we want to build. This helps to build qualified objects on mutable
+buffer, without breaking the type system with unsafe casts.
++/
+private void emplaceRef(T, UT, Args...)(ref UT chunk, auto ref Args args)
+{
+    static if (args.length == 0)
+    {
+        static assert(is(typeof({static T i;})),
+            "Cannot emplace a " ~ T.stringof ~ " because " ~ T.stringof ~
+            ".this() is annotated with @disable.");
+        static if (is(T == class)) static assert(!__traits(isAbstractClass, T),
+            T.stringof ~ " is abstract and it can't be emplaced");
+        emplaceInitializer(chunk);
+    }
+    else static if (
+        !is(T == struct) && Args.length == 1 /* primitives, enums, arrays */
+        ||
+        Args.length == 1 && is(typeof({T t = args[0];})) /* conversions */
+        ||
+        is(typeof(T(args))) /* general constructors */)
+    {
+        static struct S
+        {
+            T payload;
+            this(ref Args x)
+            {
+                static if (Args.length == 1)
+                    static if (is(typeof(payload = x[0])))
+                        payload = x[0];
+                    else
+                        payload = T(x[0]);
+                else
+                    payload = T(x);
+            }
+        }
+        if (__ctfe)
+        {
+            static if (is(typeof(chunk = T(args))))
+                chunk = T(args);
+            else static if (args.length == 1 && is(typeof(chunk = args[0])))
+                chunk = args[0];
+            else assert(0, "CTFE emplace doesn't support "
+                ~ T.stringof ~ " from " ~ Args.stringof);
+        }
+        else
+        {
+            S* p = () @trusted { return cast(S*) &chunk; }();
+            static if (UT.sizeof > 0)
+                emplaceInitializer(*p);
+            p.__ctor(args);
+        }
+    }
+    else static if (is(typeof(chunk.__ctor(args))))
+    {
+        // This catches the rare case of local types that keep a frame pointer
+        emplaceInitializer(chunk);
+        chunk.__ctor(args);
+    }
+    else
+    {
+        //We can't emplace. Try to diagnose a disabled postblit.
+        static assert(!(Args.length == 1 && is(Args[0] : T)),
+            "Cannot emplace a " ~ T.stringof ~ " because " ~ T.stringof ~
+            ".this(this) is annotated with @disable.");
+
+        //We can't emplace.
+        static assert(false,
+            T.stringof ~ " cannot be emplaced from " ~ Args[].stringof ~ ".");
+    }
+}
+// ditto
+static import core.internal.traits;
+private void emplaceRef(UT, Args...)(ref UT chunk, auto ref Args args)
+    if (is(UT == core.internal.traits.Unqual!UT))
+{
+    emplaceRef!(UT, UT)(chunk, args);
+}
+
+//emplace helper functions
+private void emplaceInitializer(T)(scope ref T chunk) @trusted pure nothrow
+{
+    import core.internal.traits : hasElaborateAssign, isAssignable;
+    static if (!hasElaborateAssign!T && isAssignable!T)
+        chunk = T.init;
+    else
+    {
+        static if (__traits(isZeroInit, T))
+        {
+            import core.stdc.string : memset;
+            memset(&chunk, 0, T.sizeof);
+        }
+        else
+        {
+            import core.stdc.string : memcpy;
+            static immutable T init = T.init;
+            memcpy(&chunk, &init, T.sizeof);
+        }
+    }
+}
+
+// emplace
+/**
+Given a pointer `chunk` to uninitialized memory (but already typed
+as `T`), constructs an object of non-`class` type `T` at that
+address. If `T` is a class, initializes the class reference to null.
+Returns: A pointer to the newly constructed object (which is the same
+as `chunk`).
+ */
+T* emplace(T)(T* chunk) @safe pure nothrow
+{
+    emplaceRef!T(*chunk);
+    return chunk;
+}
+
+///
+@system unittest
+{
+    static struct S
+    {
+        int i = 42;
+    }
+    S[2] s2 = void;
+    emplace(&s2);
+    assert(s2[0].i == 42 && s2[1].i == 42);
+}
+
+///
+@system unittest
+{
+    interface I {}
+    class K : I {}
+
+    K k = void;
+    emplace(&k);
+    assert(k is null);
+
+    I i = void;
+    emplace(&i);
+    assert(i is null);
+}
+
+/**
+Given a pointer `chunk` to uninitialized memory (but already typed
+as a non-class type `T`), constructs an object of type `T` at
+that address from arguments `args`. If `T` is a class, initializes
+the class reference to `args[0]`.
+This function can be `@trusted` if the corresponding constructor of
+`T` is `@safe`.
+Returns: A pointer to the newly constructed object (which is the same
+as `chunk`).
+ */
+T* emplace(T, Args...)(T* chunk, auto ref Args args)
+    if (is(T == struct) || Args.length == 1)
+{
+    emplaceRef!T(*chunk, args);
+    return chunk;
+}
+
+///
+@system unittest
+{
+    int a;
+    int b = 42;
+    assert(*emplace!int(&a, b) == 42);
+}
+
+@system unittest
+{
+    shared int i;
+    emplace(&i, 42);
+    assert(i == 42);
+}
+
+private @nogc pure nothrow @safe
+void testEmplaceChunk(void[] chunk, size_t typeSize, size_t typeAlignment)
+{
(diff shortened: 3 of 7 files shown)


jacob-carlborg:
I don't think this is a good idea. There's already way too much stuff in object.d.

jacob-carlborg:
I don't think this is a good idea. There's already way too much stuff in object.d.

JinShil:
> There's already way too much stuff in object.d.

I tried to do something about that in #2222, but then I ran into https://issues.dlang.org/show_bug.cgi?id=19014  Maybe that can move forward now since time has passed.

TurkeyMan:
> There's already way too much stuff in object.d.

Then `destroy` needs to move somewhere else... but either way, `emplace` needs to go there too. It's not okay that `emplace` and `destroy` aren't together.

`emplace` is harder to implement than destroy, and no user should ever attempt to get it right themselves. It's not reasonable that it be in phobos, it's too low-level and foundational.

TurkeyMan:
@JinShil If you wanna move `destroy`, then please either merge this first, then move them both, or move `emplace` at the same time...

JinShil:
I'm not going to do anything. I just wanted to point out that there is a way forward to reduce the size of object.d.  I have no opinion on whether this should be merged or not, and I am no longer working on the team.

TurkeyMan:
Okay. Well while I sympathise that object.d is kinda big, and it could be smaller, I find that issue microscopic compared to the fact that `emplace` isn't together with `destroy`, which is catastrophic for anyone that doesn't link phobos... you can't build anything meaningful without construction. You can't write any container object without construction. I can't move on std.vector or any of those without construction, etc. I've been annoyed about this for years... just gotta fix it.

schveiguy:
Would it be a reasonable compromise to put `emplace` into it's own module in `core`?

The truth is, D supplies 2 forms of construction -- one on the stack, and one on the heap. The reason `destroy` is in druntime is because it's the complement to heap construction for deterministic destruction. It's needed to replace `delete`, which is deprecated.

I agree it would be nice to have `emplace` in druntime, that's really where it belongs. But not sure if it needs to be in `object.d` There's lots of code that might use `destroy` but not `emplace`, and `emplace` is way more hairy, and pulls in more things. Having it in `core` means you don't need to import phobos, but that you don't have to import it for every module.

TurkeyMan:
I have no issue with it being in another module, but I do think it should be in the same place as `destroy`...
If someone wants to move `destroy` somewhere else, I'll put `emplace` there.

TurkeyMan:
> D supplies 2 forms of construction

I'd argue that the language supplies no forms of _construction_, only allocation (with implicit construction). The language has no construction syntax (ie, C++ placement new), and so we write it awkwardly in a library (`emplace`), which is chaos to implement!

You can't write a whole lot of tooling and container objects without being able to construct.

schveiguy:
> If someone wants to move `destroy` somewhere else

`destroy` is not coupled with `emplace`. You can `destroy` objects that were not created with `emplace`. Why does it need to be in the same module?

In any case, `destroy` can't be moved, as then literally all code that uses `destroy` would break.

TurkeyMan:
Which is exactly why I put emplacement here ;)

JinShil:
> In any case, destroy can't be moved, as then literally all code that uses destroy would break.

I believe it can be moved to another module and then publicly imported by object.d.  That, I think, will help clean up the object.d mess and hopefully make this PR more palatable.  See #2222 for a first attempt.

JinShil:
> moving destroy to another module, and then publicly importing that module, is not much different than having it here -- every module will import it.

Understood, but that's not the motivation for moving it.  object.d has too much stuff in it which causes problems when creating pull requests against it;  see the diff for #2194.  All that PR did was moving some implementations behind a `version` block and the diff algorithm marked the entire file with changes. It's unmaintainable.  @jacob-carlborg's concern if right that this PR only makes matters worse, and that, unfortunately, subtracts from the merit of this PR.  Another attempt at #2222 would probably significantly help this PR.

TurkeyMan:
The thing that's bullshit about this whole community is that people will readily complain and bikeshed when someone tried to get something done, but they wouldn't do it themselves.
Strong technical rejection is always proper, but trivial bikeshedding should be banned by anyone who isn't prepared to make their proposed amendments themselves.

This is holding me up. Let me fix it, or fix it yourself... but don't be a barrier to both.

JinShil:
In my attempt with #2222 I had to solve and fix 2 other bugs:  https://github.com/dlang/dmd/pull/8519 and https://github.com/dlang/dmd/pull/8519 and eventually got stalled on https://issues.dlang.org/show_bug.cgi?id=19014.  I didn't get any shortcuts; I did what needed to be done and even made time to help you, @TurkeyMan , with a bug I could care less about:  https://github.com/dlang/dmd/pull/8631 and a number of other bugs for other people.

When working on D (or any other software for that matter) this is how it often goes: https://www.youtube.com/watch?v=AbSehcT19u0

The size of object.d is a legitimate technical concern, and I already admitted that it is a distraction from the merits of this PR.  This PR could be merged regardless (I defer to others' judgement on that), but someone, someday will eventually need to deal the the object.d monstrosity and ignoring it now (actually adding to it) is just passing the buck.

JinShil:
> Is destroy definitely going to move?

I don't know.  Let's wait to see what happens with the PR.  Perhaps my work and our concerns were all for naught.  There are competing opinions here, and I can't predict what wins.  I have no opinion about whether `emplace` and `destroy` belong together.  I just want to remove the distraction of object.d so we can have that discussion.

jacob-carlborg:
> I don't think any anyone could argue that 'malloc' and 'free' belong together, 'new' and 'delete', etc. They shouldn't be separated, that wouldn't make sense to anyone.

One could argue that the opposite of `destroy` is `new`, i.e. the keyword.

schveiguy:
> people will readily complain and bikeshed

This is NOT bikeshedding. Once you add a symbol to object.d, it has to stay there forever (or at least always be imported automatically). If you put it in a different module inside druntime, then it's opt-in to import. This is a huge difference.

Think of it this way, if you import `core.emplace`, you will get `emplace` and `destroy`. Why is that a problem?

If it's determined that it would be better suited in object.d, we can move it there later. If we put it in object.d now, and decide later it shouldn't be there, *the decision cannot be reversed*.

I've said my piece. I'm not the final decision maker anyway, but it's frustrating not to be able to bring up *technical problems* with a PR without getting yelled at.

TurkeyMan:
> One could argue that the opposite of destroy is new, i.e. the keyword.

You could argue, but I think you'd fail.
`new` allocates memory, `destroy` performs destruction, which is the inverse of performing construction. Neither `emplace` or `destroy` have anything to do with memory allocation, and that's the whole point.

TurkeyMan:
> Once you add a symbol to object.d, it has to stay there forever

Why? We've talked about getting `destroy` out of `object.d`... which sounds sensible to me too.
Move it somewhere else, then deprecate the object.d reference (recommending people update their code to point at the new location?)

TurkeyMan:
`emplace` is an unbelievably complicated function... which is weird, because `new` additionally performs an `emplace` without any library support at all!
Clearly the language has all this complexity baked into it, but there's no language mechanism to invoke construction, and we have this horrible library implementation in its place.

Is there a way where we just expose the in-language construction machinery rather than try and reproduce it in-library?

schveiguy:
> `emplace` is an unbelievably complicated function... which is weird, because `new` additionally performs an `emplace` without any library support at all!

Looking at the differences -- the actual construction of the object is done by the compiler inline in the case of `new` (the runtime simply initializes the memory block), while `emplace` cannot operate inside the calling function. This means emplace has to deal with copying parameters, etc. and everything that's needed for actual construction, whereas the compiler doesn't have to deal with all that. That's why it's so complicated.

> Is there a way where we just expose the in-language construction machinery rather than try and reproduce it in-library?

I think this is a really good idea! It also would cut down considerably on compile time and template bloat. This would necessarily mean that whatever mechanism is used (whether it was an intrinsic or whether it's some new syntax invented) is compiler-recognized and supported.

Getting emplace into druntime is probably a good first step in this direction.

TurkeyMan:
Every time I push an iteration, it just leads to some other deps that I missed... it has a truly immense dependency stack!

It does also reveal that an awful lot of really fundamental `std.traits` should really be in druntime too.
They don't emit code, wouldn't increase the size of druntime, just a bunch of hard-to-wrangle compile-time tools, and they are all concepts that wouldn't change if you decided to write an alternative std library or runtime for instance.

TurkeyMan:
Like `AliasSeq`, why wouldn't that be at the bottom level of the stack? That level of utility shouldn't be in phobos, and there are a lot of things like that.
`-betterc` or embedded apps which don't use phobos are quite common, and getting more common.

TurkeyMan:
And like, `max`... it evaluates which thing is bigger. `a > b ? a : b` should be about the simplest function, but it's insanely complicated for its task, and even imports a deep dependency chain! I don't know where it's reasonable to draw lines.

My feeling is, I should just start pulling some critical core machinery into druntime proper, rather than just hiding it in an internal module somewhere as is the existing standard.

jacob-carlborg:
> Looking at the differences -- the actual construction of the object is done by the compiler inline in the case of new (the runtime simply initializes the memory block), while emplace cannot operate inside the calling function.

AST macros hello 😃.

schveiguy:
> And like, `max`... it evaluates which thing is bigger. `a > b ? a : b`

Can we replace instances of max with the latter expression? Seems reasonable to me...

I haven't looked at max's implementation, but I didn't realize all the dependencies.

jacob-carlborg:
>> And like, max... it evaluates which thing is bigger. a > b ? a : b

> Can we replace instances of max with the latter expression? Seems reasonable to me...
> I haven't looked at max's implementation, but I didn't realize all the dependencies.

`max` in Phobos is a bit more complicated than the above expression. It supports a variadic number of arguments and of different types. Then it needs to figure out which is the smallest type of those passed in which can hold any of the given types. That is, you can pass in a `short` and a `byte` and it will, probably, return a `short`.

TurkeyMan:
@schveiguy I have indeed factored out `max` (and other stuff)

schveiguy:
> `max` in Phobos is a bit more complicated than the above expression. It supports a variadic number of arguments and of different types.

My understanding is that `max` is equivalent to `(a > b ? a : b)` but only evaluating a or b once. That is, if any of those expressions are not simple values, they are evaluated only once.

The language takes care of the common type. In fact, if you look at `std.traits.CommonType`, the key clause is

https://github.com/dlang/phobos/blob/19f72cfe7f325ec27e763795a1db9bc7dc8b27df/std/traits.d#L4815-L4818

Which actually uses the `?:` expression to figure out the common type.

But there are a lot more symbols in that dependency tree than I would have expected.

I agree with @TurkeyMan that dependency list seems crazy excessive.

TurkeyMan:
I don't understand the CI errors...

src/object.d
```diff
@@ -498,6 +498,1321 @@ unittest
...
+            "Cannot emplace a " ~ T.stringof ~ " because " ~ T.stringof ~
+            ".this(this) is annotated with @disable.");
+
+        //We can't emplace.
+        static assert(false,
+            T.stringof ~ " cannot be emplaced from " ~ Args[].stringof ~ ".");
+    }
+}
+// ditto
+static import core.internal.traits;
```
thewilsonator: I think this import is taking precedence over the one in the test causing a different message, is this needed here?

TurkeyMan: `if (is(UT == core.internal.traits.Unqual!UT))` just 2 lines down...
I tried to `static import` so that it didn't pollute the local symbol table.
This is because constraints can't have the imports attached to the statement... I think there'd a DIP about this somewhere?

thewilsonator: `private static import`?

TurkeyMan: O_o ... import is private by default?

thewilsonator: Ya, but the problem is with the a test that deals with an old bug of import not being private.

wilzbach:
> Think of it this way, if you import core.emplace, you will get emplace and destroy. Why is that a problem?

Can we just move it in `core` somewhere, s.t. it can be used in druntime and postpone the discussion on whether it should be in `object.d`?

ZombineDev:
Yes, please! Let's use either `core.lifetime` or `core.internal.lifetime`.

src/core/internal/traits.d
```diff
@@ -14,6 +14,18 @@ template TypeTuple(TList...)
...
+    static if (is(T == struct) || is(T == union))
+        alias FieldTypeTuple = typeof(T.tupleof[0 .. $ - __traits(isNested, T)]);
+    else static if (is(T == class))
+        alias FieldTypeTuple = typeof(T.tupleof);
+    else
+    {
+        alias FieldTypeTuple = TypeTuple!T;
+    }
+}
+
```
ZombineDev: Those traits were already added by @thewilsonstor - please rebase.

thewilsonator:
Moved to `core/lifetime.d`.

changelog/emplace.dd
```diff
@@ -0,0 +1,3 @@
+Moved `std.conv.emplace` to object.d
```
jacob-carlborg: This is no incorrect.

src/core/lifetime.d
```diff
@@ -0,0 +1,1294 @@
...
+}
+
+// emplace
+/**
+Given a pointer `chunk` to uninitialized memory (but already typed
+as `T`), constructs an object of non-`class` type `T` at that
+address. If `T` is a class, initializes the class reference to null.
+Returns: A pointer to the newly constructed object (which is the same
+as `chunk`).
+ */
```
jacob-carlborg: * Missing `Params` section
* Please add a blank line between the sections

src/core/lifetime.d
```diff
@@ -0,0 +1,1294 @@
...
+Given a pointer `chunk` to uninitialized memory (but already typed
+as a non-class type `T`), constructs an object of type `T` at
+that address from arguments `args`. If `T` is a class, initializes
+the class reference to `args[0]`.
+This function can be `@trusted` if the corresponding constructor of
+`T` is `@safe`.
+Returns: A pointer to the newly constructed object (which is the same
+as `chunk`).
+ */
+T* emplace(T, Args...)(T* chunk, auto ref Args args)
```
jacob-carlborg: * Missing `Params` section
* Please add a blank line between the sections

src/core/lifetime.d
```diff
@@ -0,0 +1,1294 @@
...
+is passed the arguments `Args`.
+If `T` is an inner class whose `outer` field can be used to access an instance
+of the enclosing class, then `Args` must not be empty, and the first member of it
+must be a valid initializer for that `outer` field. Correct initialization of
+this field is essential to access members of the outer class inside `T` methods.
+Note:
+This function is `@safe` if the corresponding constructor of `T` is `@safe`.
+Returns: The newly constructed object.
+ */
+T emplace(T, Args...)(T chunk, auto ref Args args)
```
jacob-carlborg: * Missing `Params` section
* Please add a blank line between the sections

src/core/lifetime.d
```diff
@@ -0,0 +1,1294 @@
...
+            "Don't know how to initialize an object of type "
+            ~ T.stringof ~ " with arguments " ~ typeof(args1).stringof);
+    }
+    return chunk;
+}
+
+///
+@safe unittest
+{
+    () @safe {
```
jacob-carlborg: Why is this necessary when the whole unit test block is `@safe`?

src/core/lifetime.d
```diff
@@ -0,0 +1,1294 @@
...
+this field is essential to access members of the outer class inside `T` methods.
+Preconditions:
+`chunk` must be at least as large as `T` needs and should have an alignment
+multiple of `T`'s alignment. (The size of a `class` instance is obtained by using
+$(D __traits(classInstanceSize, T))).
+Note:
+This function can be `@trusted` if the corresponding constructor of `T` is `@safe`.
+Returns: The newly constructed object.
+ */
+T emplace(T, Args...)(void[] chunk, auto ref Args args)
```
jacob-carlborg: * Missing `Params` section
* Please add a blank line between the sections

src/core/lifetime.d
```diff
@@ -0,0 +1,1294 @@
...
+Preconditions:
+`chunk` must be at least as large
+as `T` needs and should have an alignment multiple of `T`'s
+alignment.
+Note:
+This function can be `@trusted` if the corresponding constructor of
+`T` is `@safe`.
+Returns: A pointer to the newly constructed object.
+ */
+T* emplace(T, Args...)(void[] chunk, auto ref Args args)
```
jacob-carlborg: * Missing `Params` section
* Please add a blank line between the sections

src/core/lifetime.d
```diff
@@ -0,0 +1,1294 @@
...
+    class K : I {}
+
+    K k = void;
+    emplace(&k);
+    assert(k is null);
+
+    I i = void;
+    emplace(&i);
+    assert(i is null);
+}
```
jacob-carlborg: Should there be a test that verifies that the constructor has been called?

TurkeyMan:
@thewilsonator Just an FYI, I think it's considered extremely poor form to rebrand someone's git commit metadata when you amend. I've seen this occur on a few other PR's, I'd suggest that when you amend someone's PR's, take extra care not to erase the original authors attribution and commit metadata.

TurkeyMan:
@jacob-carlborg I don't want to change anything about the implementation in this PR. It's a perfect clone of phobos to the extent we know it works. If we tidy the things you say, I'd like to see that in a separate PR where it's not conflated with moving it to druntime.

jacob-carlborg:
> I don't want to change anything about the implementation in this PR. It's a perfect clone of phobos to the extent we know it works

Fine.

WalterBright:
`src/core/internal/traits.d` As far as I can tell, all the changes in this file are reformatting. Please put reformatting in a separate file, as this diff is way way too large to start adding irrelevant changes.

TurkeyMan:
I see what's happened, this PR was actually the original patch to introduce those traits, but nick extracted that code into a separate PR and merged, but reformatted it in the process. When I rebased, my (original) formatting changes were the only differences that remained.
I moved those traits over from phobos verbatim, the formatting in this patch is identical to the phobos formatting. I'm actually a bit upset that it was disturbed when it was merged separately.

TurkeyMan:
The idea here was to have ZERO delta from the phobos code, and that's what's in this patch ;)

thewilsonator:
>  but reformatted it in the process.

Sorry, my editor occasionally does funny things with formatting.

TurkeyMan:
Thx!

Sadly, I just noticed that it doesn't end here :(
I also need to relocate `std.algorithm.mutation.move` to druntime. Does anyone wanna have a go at that?
Druntime needs all the allocation primitives; `emplace`, `destroy`, `move`... also `forward`?

ZombineDev:
+1 for `move` and `forward`.

PR #2399, opened 2018-12-09
Refactor object.d: Move `destroy` and friends to new file, lifetime.d

JinShil:
Reboot of #2222 in an attempt to help with #2398


thewilsonator:
```
/var/lib/buildkite-agent/builds/buildkite-agent-04-1/dlang/druntime/build/distribution/bin/../imports/object_/lifetime.d(446): Deprecation: `object.isnan` is not visible from module `lifetime`
--
  | /var/lib/buildkite-agent/builds/buildkite-agent-04-1/dlang/druntime/build/distribution/bin/../imports/object_/lifetime.d(446): Error: function `object.isnan` is not accessible from module `lifetime`
```

JinShil:
Thanks, @thewilsonator

thewilsonator:
No problems.

JinShil:
The auto-tester was failing, so I tried to modify some of the makefiles.  Now circleci is failing.  I don't know why.  If someone has an idea, please let me know.

JinShil:
What is this all about, and why is it affecting this PR?

```compilable/json.d(185): Deprecation: `extern(Pascal)` is deprecated. You might want to use `extern(Windows)` instead.
```

JinShil:
WARNING:  This PR causes a huge change in the documentation; see DAutoTest, but I can't see the usual diffs to determine what the changes are.  I browsed the documentation and couldn't find any issues, but that doesn't mean there aren't any.

JinShil:
I don't understand what's failing in DAutoTest.  @CyberShadow could you take a quick look?

JinShil:
Ok, this is looking good now.  Please notice the changes in the documentation.  There's a new `object_` package in the left navigation and that's why there are so many changed files.

There was [a concern](https://github.com/dlang/druntime/pull/2222#issuecomment-398390889) in #2222 about this negatively affecting build times.  I couldn't see any measurable difference when building phobos (at least with all the statistical noise).  Reviewers are advised to independently verify.

src/object_/lifetime.d
```diff
@@ -0,0 +1,749 @@
+/**
+ * Contains utilities for managing lifetimes.
+ *
+ * This module is not intended to be imported or used directly.  It is
+ * publicly imported by object.d, and should be accessed through there.
```
CyberShadow: What does "should be accessed through there" mean? Would it be more accurate to say that the declarations in this module are automatically available in all D modules, like those in `object`?

JinShil: > This module is not intended to be imported or used directly.  It...

"It" refers to "This module". 

> It is publicly imported by object.d, and should be accessed through there.

"there" refers to "object.d"

This is natural English for me (using pronouns to refer to previously mentioned things).  If you'd like me to change it, please send me copy-pastable text.

CyberShadow: What I'm trying to say is that it might not be obvious to users that they don't have to import anything in order to use the symbols from that module. E.g. someone googles for "dlang destroy object", finds this page on dlang.org, sees that it's in a module, and tries importing it. Heck, at that point the user might not even know that even `object` doesn't need to be imported.

"accessed through there" also can be interpreted that something needs to be done by the user to achieve this, specifically something involving what "there" refers to (the `object` module).

JinShil: I gave it another go to address your concerns.  I don't know if it's good enough.  Please send me something better it's not.

CyberShadow: It's better?

mak/DOCS
```diff
@@ -1,5 +1,8 @@
 DOCS=\
 	$(DOCDIR)\object.html \
+    \
+    $(DOCDIR)\object__lifetime.html \
+    \
```
TurkeyMan: Spaces where should be tabs?

TurkeyMan:
Why `object_/lifetime.d`, and not just `lifetime.d`? That underscore looks pretty janky for such a fundamental package?

mak/DOCS
TurkeyMan: I'm not sure about this location, but make can be fussy about tabs vs spaces.

JinShil: Done

JinShil:
> Why object_/lifetime.d, and not just lifetime.d? 

There were [concerns](https://github.com/dlang/druntime/pull/2222#issuecomment-399089348) in #2222 about placing lifetime.d in the root and preventing users from being able to use lifetime.d as their own top leve package.  I'm actually not sure about that.  @wilzbach, care to clarify?

> That underscore looks pretty janky for such a fundamental package?

Actually, I just realized what @schveiguy [was saying](https://github.com/dlang/druntime/pull/2222#issuecomment-399108536) in #2222.  I think I should have done `core.internal.object` all along.  I think I'll try that.  It will prevent a lot of the other changes I made as well.

TurkeyMan:
> preventing users from being able to use lifetime.d as their own top level package

So, `core.lifetime`?

JinShil:
> So, core.lifetime? What's what I would have assumed.

I think we could make that work, but I'd like to keep all the publicly imported parts of object.d together as a group and segregated from the other `core` files.  I'd be ok with it but I'm going with `core.internal.object.lifetime.d` until I find a reason not to.

TurkeyMan:
So we're gonna publicly import `emplace` too then? Because it's an impenetrable module name otherwise.

JinShil:
> So we're gonna publicly import emplace too then?

If `emplace` is placed alongside `destroy` in `lifetime.d` it would be imported publicly through `object.d`

JinShil:
> why does destroy need to be implicitly available?

I don't know, but now that it is, it'll take a deprecation cycle and a lot of persuasions to change.

> is it possible to deprecate that public import with time?

I suppose, if you can convince the right people.  There are a lot of architecture problems with druntime.

JinShil:
Damn!  If you see the autotester, it's building with 2.079 which doesn't yet have the fix from https://issues.dlang.org/show_bug.cgi?id=19014.  I'm sure what to do about that yet.  2.083 (the version on my local machine) also seems to still have the problem. When I build from HEAD everything works fine.  @RazvanN7 @wilzbach  wasn't that fix supposed to be released already?

n8sh:
Currently `core.internal.*` modules don't appear in the documentation. We could keep it that way and just add this in `object.d`:
```d
version(D_Doc)
{
    /**
    Destroys the given object and optionally resets to initial state. It's used to
    _destroy an object, calling its destructor or finalizer so it no longer
    references any other objects. It does $(I not) initiate a GC cycle or free
    any GC memory.
    If `initialize` is supplied `false`, the object is considered invalid after
    destruction, and should not be referenced.
    */
    void destroy(bool initialize = true, T)(ref T obj)
    if (!is(T == class) && !is(T == interface))
    {
        import lifetime = core.internal.object.lifetime;
        lifetime.destroy!(initialize)(obj);
    }
    /// ditto
    void destroy(bool initialize = true, T)(T obj)
    if (is(T == class) || is(T == interface))
    {
        import lifetime = core.internal.object.lifetime;
        lifetime.destroy!(initialize)(obj);
    }
}
else
{
    public import core.internal.object.lifetime : destroy;
}
```

src/core/internal/object/lifetime.d
```diff
@@ -0,0 +1,752 @@
+/**
+ * Contains utilities for managing lifetimes.
+ *
+ * This module is not intended to be imported or used directly.  It is
```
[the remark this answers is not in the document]
JinShil: If I make them `package` then they won't be visible outside of `core.internal.object`, right?  If that's the case, then I won't be able to publicly import them into `object.d`.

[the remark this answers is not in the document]
JinShil: This goes along with my [original idea](https://github.com/dlang/druntime/pull/2222#issuecomment-399092667) in #2222 to put them all in the root, rename `object.d` to `package.d` and then modify the compiler to look for `package.d` (or the root of the runtime) instead of `object.d`  There were concerns about not being able to use `lifetime.d` in one's own file hierarchy.  But, honestly, I don't undersand that, because they would be under a different folder and users should never import the runtime's `lifetime.d` anyway.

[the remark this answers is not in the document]
jacob-carlborg: It's possible to specify a symbol along with package, i.e. `package(core)` which would make it accessible to the whole `core` package and all its sub packages. But since `object` lives outside any package, that might not work anyway.

JinShil:
Due to [Issue 19014](https://issues.dlang.org/show_bug.cgi?id=19014) This needs the autotester upgraded to 2.084 before it can move forward.  Closing until that becomes reality.

PR #2400, opened 2018-12-09
Reduce uses of windows.d b/c it's slow (see PR #2336)

n8sh:
In general avoided removing `core.sys.windows.windows` imports from other `core.sys.windows.*` header files so imports-of-imports from them will not break. The exception is `core.sys.windows.threadaux` because it is imported by `core.thread`.


n8sh:
I can also do this in `core.sys.windows.snmp` and `core.sys.windows.httpext` because their imports of `core.sys.windows.window` are private.

rainers:
> In general avoided removing non-private `core.sys.windows.windows` imports from other core.sys.windows.* header files so imports-of-imports from them will not break. 

import is private by default, so only "public imports" might be affected.

Selective imports do create public symbols, though (albeit deprecated). So this could be a breaking change causing ambiguities before the deprecated behaviour is removed from the compiler. It might be less invasive to not list imported symbols (while mostly desirable in general some lists look a bit excessive).

Do you have some example numbers what the compile time improvements are?

n8sh:
>Selective imports do create public symbols, though (albeit deprecated). So this could be a breaking change causing ambiguities before the deprecated behaviour is removed from the compiler.

Didn't know that, thanks. This PR should not be merged before I fix this.

>Do you have some example numbers what the compile time improvements are?

I assumed there would only be a noticeable speed difference with uncached reads from a non-SSD drive (I'm assuming the main speed penalty from importing windows.d is opening ~32 other files), but benchmarking would still be a good idea.

thewilsonator:
Ping @n8sh

n8sh:
@thewilsonator removed top-level selective imports as @rainers recommended. I also did an informal benchmark with a non-optimized DMD development build. Compiling the following program went from 640 milliseconds (best of 40 tries) to 240 milliseconds (best of 40 tries):
```d
import core.internal.abort;
import core.runtime;
import core.sync.condition;
import core.sync.mutex;
import core.sync.semaphore;
import core.sys.windows.httpext;
import core.sys.windows.snmp;
import core.sys.windows.threadaux;
import core.thread;
import core.time;

void main(string[] args) {}
```

n8sh:
That was a more noticeable difference than I expected. I'll remove the remaining imports of `windows.d` from other `core.sys.windows.*` files.

PR #2403, opened 2018-12-10
Make rt utf publicly available

edi33416:
This is the first step in reusing some of druntime's `utf.d` implementations in Phobos.

Please see [forum](https://forum.dlang.org/post/ahieisavseagdkzvexps@forum.dlang.org) discussion.

PR #2404, opened 2018-12-11
Make encode reusable in Phobos

edi33416:
This PR adds two overloads to `encode` so those can be called from `std.utf.encode` (see [here](https://github.com/dlang/phobos/blob/master/std/utf.d#L2253))

This PR depends on #2403


src/core/internal/utf.d
```diff
@@ -425,61 +425,98 @@ dchar decode(in dchar[] s, ref size_t idx)
...
+ * Encodes `c` into the static array `buf`.
+ *
+ * Params:
+ *     buf = destination of encoded character
+ *     c = character to encode
+ *
+ * Returns:
+ *     The length of the encoded character (a number between `1` and `4` for
+ *     `char[4]` buffers and a number between `1` and `2` for `wchar[2]` buffers)
+ *     or `0` in case of failure.
```
andralex: Great idea. Let's define this function for all inputs (no contracts, no `assert(isValidDChar)`) and have it return 0 upon bad dchar. Then higher-level functions can throw exceptions, use replacement char etc.

Geod24: Additionally, instead of returning a length, return a slice to the buffer, or `null` on bad `dchar`, because the next thing people will need to do is slice that buffer.
Returning `null` and *not* an empty slice is important, because otherwise `if (auto s = encode(...))` will always be `true`.
Lastly, since this piece of code is highly performance sensitive, did you inspect the assembly to make sure `out` does not get default-initialized on function entry ?

jacob-carlborg: Isn't it better to pass `buf` as a regular `char[]` to push the allocation choice to the caller?

src/core/internal/utf.d
```diff
@@ -425,61 +425,98 @@ dchar decode(in dchar[] s, ref size_t idx)
...
+ */
+nothrow pure @safe
+void encode(ref char[] s, dchar c)
+in
+{
+    assert(isValidDchar(c));
+}
+do
+{
+    char[4] buf;
```
WalterBright: `= void;`

src/core/internal/utf.d
```diff
@@ -499,43 +536,32 @@ unittest
...
-        }
-        s = r;
-    }
+in
+{
+    assert(isValidDchar(c));
+}
+do
+{
+    wchar[2] buf;
```
WalterBright: `= void;`

src/core/internal/utf.d
```diff
@@ -425,61 +425,98 @@ dchar decode(in dchar[] s, ref size_t idx)
...
+        return 2;
+    }
+    return 0;
+}
 
+/**
+ * Encodes character c and appends it to array s[].
+ */
+nothrow pure @safe
+void encode(ref char[] s, dchar c)
```
WalterBright: Not a fan of the following functions because:
1. They are an overload of the previous `encode` functions, but do something fundamentally different
2. They rely on the GC. We should be crafting functions that do not allocate, but instead send the characters to a sink or an output range.

jacob-carlborg: Exactly: https://github.com/dlang/druntime/pull/2404#discussion_r244679517.

jacob-carlborg: Also, is this the right order of arguments for UFCS? Would it be better to pass `c` first?

WalterBright: The sink normally is the first argument, as in:
```
sink.put(c);
```

RazvanN7:
@edi33416 Are you willing to take this to the finish line?

PR #2406, opened 2018-12-11
Add a few public examples to object.d

wilzbach:
These are all core functions. When I enter them in Google, I should be directed to at least one example.

Oh and yes, lots of stuff is still `@system`.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3661,6 +3661,7 @@ private
 {
     return _d_arraysetcapacity(typeid(T[]), 0, cast(void[]*)&arr);
 }
+
 ///
 @safe unittest
 {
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -3309,6 +3522,28 @@ if (isCreateOperation!(C, V) && isUpdateOperation!(U, V))
...
+///
+@system unittest
+{
+    auto aa = ["k1": 1];
+
+    aa.update("k1", {
+        assert(0, "k1 already exists.");
+        return 0;
+    }, (ref int v) {
+        return v + 1; // update
```
wilzbach: NB: we can't use `v++` here as `update` calls the function with the dereferenced `value`.
However, it also requires that the function accepts the parameter by reference. Can we please agree to fix this?

PR #2407, opened 2018-12-12
core/threadasm.S: Split up per target and move to core/internal/threadasm

ibuclaw:
GDC is able to compile the correct source based on what's determined at configure-time.

Can LDC do the same? Ping @kinke / @redstar


kinke:
Sure; it'll require some CMake fiddling though.

joakim-noah:
Is it possible to move these to D? It'd be nice to get rid of the C compiler requirement to build the stdlib, at least as an option. `errno.c` is not even used by druntime anymore, by default. If we make `etc.c.zlib` optional, we could try to have a build option that only uses the D compiler.

ibuclaw:
@klickverbot - I removed one layer of ifdef's, so you'll still need e.g:
```
#if defined( __PPC64__ )
#include "core/internal/threadasm/powerpc64.S"
#endif
```
But yes, in principle you could do that.

Actually, there's a further refactoring to be done, move the header definitions into a common asm file.

ibuclaw:
> Actually, there's a further refactoring to be done, move the header definitions into a common asm file.

Done and rebased.

thewilsonator:
Needs another rebase but otherwise looks fine to me.

Geod24:

PR #2409, opened 2018-12-16
Fix issue 19128 - argument to alloca may be too large

ibuclaw:
Replaces #2258.

### DIFF
--- a/src/rt/arrayassign.d
+++ b/src/rt/arrayassign.d
@@ -30,7 +30,12 @@ extern (C) void[] _d_arrayassign(TypeInfo ti, void[] from, void[] to)
 
     // Need a temporary buffer tmp[] big enough to hold one element
     void[16] buf = void;
-    void* ptmp = (elementSize > buf.sizeof) ? alloca(elementSize) : buf.ptr;
+    void* ptmp = (elementSize > buf.sizeof) ? malloc(elementSize) : buf.ptr;
+    scope (exit)
+    {
+        if (ptmp != buf.ptr)
+            free(ptmp);
+    }
     return _d_arrayassign_l(ti, from, to, ptmp);
 }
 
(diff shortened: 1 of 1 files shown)


ibuclaw:
Dammit windows, doesn't support basic functionality.
```
Error: cannot mix `core.std.stdlib.alloca()` and exception handling in `_d_arrayassign()`
--- errorlevel 1
```

ibuclaw:
Only fixed `_d_arraysetassign` to not use `scope(exit)`.

For `_d_arrayassign` partially reverted back to use `void[16]` if the element fits, otherwise always use `malloc()`.  This is fine IMO, as this function is deprecated, and the compiler should be calling `_d_arrayassign_l` directly itself.

src/rt/arrayassign.d
```diff
@@ -206,24 +211,20 @@ extern (C) void* _d_arraysetassign(void* p, void* value, int count, TypeInfo ti)
...
-    //Need a temporary buffer tmp[] big enough to hold one element
-    void[16] buf = void;
-    void[] tmp;
-    if (element_size > buf.sizeof)
-    {
-        tmp = alloca(element_size)[0 .. element_size];
-    }
-    else
-        tmp = buf[];
+    // Need a temporary buffer tmp[] big enough to hold one element
```
ZombineDev:
```suggestion
    // Need a temporary buffer big enough to hold one element.
    // Stack allocate if element_size < 512.
```

thewilsonator:
Please give the a kick, FreeBSD seems to be getting a lot of those socket errors.

wilzbach:
> Please give the a kick, FreeBSD seems to be getting a lot of those socket errors.

Auto-tester's agents will automatically restart.

wilzbach:
Though I merged this manually as currently some Buildkite agents are running into timeouts.

PR #2410, opened 2018-12-16
 core/thread: Support StackGrowsUp in getStackBottom()

ibuclaw:
The only targets I'm aware of where version != StackGrowsDown are PA-Risc and Sanyo's Stormy16 (we don't care about the latter, however).

Regarding rationale of fix, see documentation of `pthread_attr_getstack`:
http://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_attr_getstack.html

> The base (lowest addressable byte) of the storage shall be stackaddr, and the size of the storage shall be stacksize bytes.

So the implementation of this function should look something like the following:
```
int pthread_attr_getstack (in pthread_attr_t* attr, void** stackaddr, size_t* stacksize)
{
  auto iattr = cast(pthread_internal_attr *)attr;

  version (StackGrowsDown)
    *stackaddr = iattr.stackaddr - iattr.stacksize;
  else
    *stackaddr = iattr.stackaddr;

  *stacksize = iattr.stacksize;

  return 0;
}
```

With this, all druntime unittests bar core.thread pass, as it allows the GC to work correctly.

Tested on hppa-linux-gnu.

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -3256,7 +3256,9 @@ private void* getStackBottom() nothrow @nogc
         pthread_getattr_np(pthread_self(), &attr);
         pthread_attr_getstack(&attr, &addr, &size);
         pthread_attr_destroy(&attr);
-        return addr + size;
+        version (StackGrowsDown)
+            addr += size;
+        return addr;
     }
     else version (FreeBSD)
     {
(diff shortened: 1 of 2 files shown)


ZombineDev:
Auto-merge toggled on

PR #2411, opened 2018-12-16
Fix Issue 19481: mutex.d(95) Error: pthread_mutex_init failed

ibuclaw:
As seen on at least Solaris, maybe SPARC/Linux as well.

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -240,6 +240,12 @@ template maxAlignment(U...)
     }
 }
 
+template classInstanceAlignment(T)
+if (is(T == class))
+{
+    alias classInstanceAlignment = maxAlignment!(void*, typeof(T.tupleof));
+}
+
 /// See $(REF hasElaborateMove, std,traits)
 template hasElaborateMove(S)
 {
--- a/src/core/thread/osthread.d
+++ b/src/core/thread/osthread.d
@@ -190,6 +190,13 @@ private
 {
     import core.atomic, core.memory, core.sync.mutex;
 
+    // Handling unaligned mutexes are not supported on all platforms, so we must
+    // ensure that the address of all shared data are appropriately aligned.
+    import core.internal.traits : classInstanceAlignment;
+
+    enum mutexAlign = classInstanceAlignment!Mutex;
+    enum mutexClassInstanceSize = __traits(classInstanceSize, Mutex);
+
     //
     // exposed by compiler runtime
     //
@@ -1790,29 +1797,30 @@ package(core.thread):
     // lock order inversion.
     @property static Mutex slock() nothrow @nogc
     {
-        return cast(Mutex)_locks[0].ptr;
+        return cast(Mutex)_slock.ptr;
     }
 
     @property static Mutex criticalRegionLock() nothrow @nogc
     {
-        return cast(Mutex)_locks[1].ptr;
+        return cast(Mutex)_criticalRegionLock.ptr;
     }
 
-    __gshared align(Mutex.alignof) void[__traits(classInstanceSize, Mutex)][2] _locks;
+    __gshared align(mutexAlign) void[mutexClassInstanceSize] _slock;
+    __gshared align(mutexAlign) void[mutexClassInstanceSize] _criticalRegionLock;
 
     static void initLocks() @nogc
     {
-        foreach (ref lock; _locks)
-        {
-            lock[] = typeid(Mutex).initializer[];
-            (cast(Mutex)lock.ptr).__ctor();
-        }
+        _slock[] = typeid(Mutex).initializer[];
+        (cast(Mutex)_slock.ptr).__ctor();
+
+        _criticalRegionLock[] = typeid(Mutex).initializer[];
+        (cast(Mutex)_criticalRegionLock.ptr).__ctor();
     }
 
     static void termLocks() @nogc
     {
-        foreach (ref lock; _locks)
-            (cast(Mutex)lock.ptr).__dtor();
+        (cast(Mutex)_slock.ptr).__dtor();
+        (cast(Mutex)_criticalRegionLock.ptr).__dtor();
     }
 
     __gshared Context*  sm_cbeg;
@@ -3726,7 +3734,7 @@ private
     __gshared size_t ll_nThreads;
     __gshared ll_ThreadData* ll_pThreads;
 
-    __gshared align(Mutex.alignof) void[__traits(classInstanceSize, Mutex)] ll_lock;
+    __gshared align(mutexAlign) void[mutexClassInstanceSize] ll_lock;
 
     @property Mutex lowlevelLock() nothrow @nogc
     {


thewilsonator:
@TurkeyMan is about to add something like maxAlignment soon.

ibuclaw:
But in the meantime, Sparc and i386 Solaris is failing _now_, and this fixes it.

thewilsonator:
See https://github.com/dlang/druntime/pull/2412 for the traits from @TurkeyMan's PR.

src/core/thread.d
```diff
@@ -1709,13 +1709,20 @@ private:
...
+
+    __gshared align(mutexAlign) void[alignedMutexInstanceSize][2] _locks;
 
     static void initLocks() @nogc
     {
+        const init = typeid(Mutex).initializer[];
         foreach (ref lock; _locks)
         {
-            lock[] = typeid(Mutex).initializer[];
+            lock[0 .. init.length] = init[];
```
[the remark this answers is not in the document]
ZombineDev: `lock.length == typeid(Mutex).initializer.length` and then `lock[] == lock[0 .. $] == lock[0 .. init.length]`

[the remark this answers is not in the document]
ibuclaw: > lock.length == typeid(Mutex).initializer.length 

That's where you're wrong (read the [issue](https://issues.dlang.org/show_bug.cgi?id=19481#c3) on alignment of pthread_mutex_t).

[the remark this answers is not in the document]
ibuclaw:
```
assert(typeid(Mutex).initializer.length == 44);
assert(lock.length == 48); // Aligned up so that locks[1] is _also_ on an 8-byte boundary.
```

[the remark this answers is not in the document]
ZombineDev: Ah ok, thanks! There was a difference between how I thought it should work and what the proposed changes do. Which leads me to making the suggestion above:

src/core/thread.d
```diff
@@ -1709,13 +1709,22 @@ private:
...
-    __gshared align(Mutex.alignof) void[__traits(classInstanceSize, Mutex)][2] _locks;
+    // Handling unaligned mutexes are not supported on all platforms, so we
+    // must round up the statically allocated size of the _locks array to
+    // ensure that the address of all elements are appropriately aligned.
+    import core.internal.traits : classInstanceAlignment;
+
+    enum mutexAlign = classInstanceAlignment!Mutex;
+    enum alignedMutexInstanceSize = (__traits(classInstanceSize, Mutex) + mutexAlign - 1) & ~(mutexAlign - 1);
+
+    __gshared align(mutexAlign) void[alignedMutexInstanceSize][2] _locks;
```
ZombineDev: This can be simplified like this:
```D
struct RawStorage(size_t size, ubyte alignment)
{ align (alignment) void[size] storage; }

alias MutexStorage = RawStorage!(
    __traits(classInstanceSize, Mutex),
    classInstanceAlignment!Mutex
);

__gshared MutexStorage[2] _locks;
```
...

```D
static void initLocks() @nogc
{
    foreach (ref lock; _locks)
    {
        lock.storage[] = typeid(Mutex).initializer[];
        (cast(Mutex)lock.storage.ptr).__ctor();
    }
}
```

src/core/internal/traits.d
```diff
@@ -250,3 +250,25 @@ template staticMap(alias F, T...)
                 staticMap!(F, T[$/2 ..  $ ]));
     }
 }
+
+// std.traits.classInstanceAlignment
+private template maxAlignment(U...)
```
thewilsonator: @ibuclaw please rebase and remove this.

schveiguy:
Would it be better/easier to avoid using an array to define `_locks`? In other words, just `_lock0`, and `_lock1` to allow more control over the alignment of the second lock?

ibuclaw:
> Would it be better/easier to avoid using an array to define _locks?

Granted that there's only two locks, there's nothing lost having two separate static variables instead.

ibuclaw:
> This can be simplified like this:

I think that's just over engineering the problem. There's no need to wrap it up in a struct.

ZombineDev:
I'm strongly against repeating the compiler's alignment calculation logic. Repeating it is over-engineering.

I agree with with @schveiguy that code could be even simpler and less error prone if we do away with the array entirely.

ibuclaw:
> I'm strongly against repeating the compiler's alignment calculation logic. Repeating it is over-engineering.

The small pattern that rounds a number up to an alignment boundary? It's hardly repeating the compiler (we're doing something that the compiler didn't - as it doesn't make sense to add padding to the end of classes), and is used throughout in various other places from varargs to sections (perhaps there should be an `alignTo!(X)` template).

kinke:
> perhaps there should be an alignTo!(X) template

I think such a helper would be really useful, I missed it countless times too.

ibuclaw:
Rebased.

thewilsonator:
Hmm
```
src/mecca/lib/memory.d:279: error: undefined reference to '_D4core6thread6Thread6_locksG2G72v'
--
  | src/mecca/lib/memory.d:291: error: undefined reference to '_D4core6thread6Thread6_locksG2G72v'
```

thewilsonator:
Urgh: https://github.com/weka-io/mecca/blob/master/src/mecca/lib/memory.d#L443-L451

ibuclaw:
This is what happens when `private` does not have any effect on symbol visibility in the object file. :-)

thewilsonator:
cc @EyalIO whats the best way forward? Commit this and then fix up your end?

Also @ibuclaw this should target stable and then merge stable into master straight after, that way the DMD releases can be  used to build mecca.

ibuclaw:
@thewilsonator - Shall we just move forward?  The fix for Mecca will not work until this gets merged anyway...

@klickverbot - I know, but I still like to complain about long compile times caused by inability to discard instantiated symbols. ;-)

ibuclaw:
Meanwhile, Solaris port is broken...

jacob-carlborg:
> The fix for Mecca will not work until this gets merged anyway...

Mecca can be patched to check for the `_locks` member. Something like this:

```d
static if (__traits(hasMember, Thread, "_locks"))
{
    pragma(mangle, "_D4core6thread6Thread6_locksG2G72v") extern __gshared static
            void[__traits(classInstanceSize, Mutex)][2] _locks;
}

else
{
    pragma(mangle, ...) extern __gshared 
        align(mutexAlign) void[mutexClassInstanceSize] _slock;
}
```

thewilsonator:
https://github.com/weka-io/mecca/pull/14

thewilsonator:
This is ready to go as soon as a new mecca version is tagged.

src/core/internal/traits.d
```diff
@@ -220,6 +220,12 @@ template maxAlignment(U...)
     }
 }
 
+template classInstanceAlignment(T)
+if (is(T == class))
+{
+    alias classInstanceAlignment = maxAlignment!(void*, typeof(T.tupleof));
```
rainers: This doesn't seem to work for fields with explicit alignment:
```
class C { align(16) int x; }
pragma(msg, classInstanceAlignment!C); // 4u on 32-bit
```

ibuclaw: Neither does the phobos template of the same name.  It's a library limitation.

ibuclaw:
Rebased, this also fixes protection fault on HPPA due to unaligned access as well.

Any hold up?

jacob-carlborg:
I'm not so sure that the patch for Mecca is correct: 
https://github.com/weka-io/mecca/pull/14/files#r254973896.

ibuclaw:
To me, it looks like buildkite is testing the wrong version still anyway.

jacob-carlborg:
> To me, it looks like buildkite is testing the wrong version still anyway.

Yes.

src/core/internal/traits.d
ZombineDev: @rainers I can't find a way to get the explicit alignment of class fields:
```D
class C { align(16) int x; }

void main()
{
    auto c = new C();    
    pragma (msg, c.x.alignof);
    pragma (msg, C.x.alignof);
    pragma (msg, C.tupleof[0].alignof);
}
```

```
4LU
4LU
4LU
```

ZombineDev:
@EyalIO @JohanEngelen can we get a new tag for mecca? You may also want to recheck whether @thewilsonator's PR is correct, given the latest version of this PR.

src/core/internal/traits.d
rainers: The problem (maybe a bug) seems to be that the compiler only looks at the type, not the symbol.

ibuclaw:
@ZombineDev, what about adding a compatibility stub? Though this is private at language level, it's still a public ABI change in the resultant binary.

ibuclaw:
There should be some impending deadline or high priority label, as gcc-9 release is only 4 weeks away and this _needs_ to be in before then to unblock the completion of porting of a few archs.

wilzbach:
We can always temporarily remove mecca if the maintainers are unresponsive or switch to master if creating a new tag is not possible. We have done that in the past too.

thewilsonator:
The maintainers are reasonably responsive, the breakage is due to them needing internals of `core.Thread.Context` because `core.thread` is poorly designed. The easiest way forward is to not affect the mangling of `Thread._locks`.

jacob-carlborg:
> it's still a public ABI change in the resultant binary

We don't offer any ABI compatibility.

jacob-carlborg:
> The easiest way forward is to not affect the mangling of Thread._locks.

Using `pragma(mangle)` here to match mecca? The mangling in this case if platform dependent since it contains the size of `Mutex`, IIRC.

ibuclaw:
> The mangling in this case if platform dependent since it contains the size of Mutex, IIRC. But that can be fixed as well.

The Mecca lib has locked itself out of being platform independent anyway, due to this assert.
```
static assert (__traits(classInstanceSize, Mutex) == 72); // This size is part of the mangle
```

jacob-carlborg:
> The Mecca lib has locked itself out of being platform independent anyway, due to this assert.
>```
>static assert (__traits(classInstanceSize, Mutex) == 72); // This size is part of the mangle
>```

Yes, but if we’re going to use `pragma(mangle)` in druntime doesn’t it need to be correct on all platforms? There’s also this PR that adds support for macOS to mecca: https://github.com/weka-io/mecca/pull/12. Which handles the issue by using `__traits(classInstanceSize, Mutex)` as part of the mangled name instead of completely hard coding it.

thewilsonator:
Thats not the problem:

error: undefined reference to '_D4core6thread6Thread6_locksG2G72v'

```
struct DRuntimeStackDescriptor {
    static if (__traits(hasMember, Thread, "_locks")) { // not the case anymore
       //...
    }
    else {
        pragma(mangle,"_D4core6thread6Thread6_slockG72v") extern __gshared static
            void[__traits(classInstanceSize, Mutex)] _slock;
   }
}
```

`Thread` doesn't have a member `_locks` anymore (this PR changes it) so I don't know why its taking the wrong static branch.

jacob-carlborg:
It's still using the `v0.0.8` tag because the new tag is missing the `v` prefix: https://github.com/weka-io/mecca/releases/tag/0.0.9.

thewilsonator:
Ahh.

thewilsonator:
https://github.com/weka-io/mecca/issues/16

jacob-carlborg:
@thewilsonator this still looks suspicious to me: https://github.com/weka-io/mecca/blob/1f6909ec02d19ef2115b7c7402cf4c0327bd316a/src/mecca/lib/memory.d#L453-L454. Having two symbols with the same name in the same scope, as far as I can see.

thewilsonator:
Probably, but thats not the reason it is failing at the moment.

jacob-carlborg:
> Probably, but thats not the reason it is failing at the moment.

No, but it still needs to be fixed. When they fix/add a new tag it's going to fail for another reason.

ibuclaw:
>  When they fix/add a new tag it's going to fail for another reason.

Indeed.  https://explore.dgnu.org/z/dNR7_q

EyalIO:
Sorry about the missing `v` in the tag, the `v0.0.9` tag is now equal to `0.0.9`.

ibuclaw:
Rebased, interestingly still get undefined reference.

jacob-carlborg:
Using the correct tag now at least.

ibuclaw:
Nope, the tag was never correct in the first place.

https://github.com/weka-io/mecca/compare/v0.0.9...master

EyalIO:
Created a v0.0.10 tag with the new commit: https://github.com/weka-io/mecca/releases/tag/v0.0.10

ibuclaw:
And now it is failing as we expected. No point creating more tags until the problem in mecca is fixed.

JinShil:
It appears mecca has already been disabled:  https://github.com/dlang/ci/blob/9d820657ea140e7128ba9daf4b60f68bfd54afb8/buildkite.sh#L145  Please rebase.

ibuclaw:
Rebased.

thewilsonator:
Auto-merge toggled on

PR #2412, opened 2018-12-17
Add some traits

thewilsonator:
cc @ibuclaw

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -14,6 +14,18 @@ template TypeTuple(TList...)
     alias TypeTuple = TList;
 }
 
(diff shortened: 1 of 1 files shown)


ZombineDev:
If we're going to add those traits to druntime, can we move them completely, by aliasing them from phobos?

thewilsonator:
They're in `core.internal.traits` so no, not unless we move `traits` out of `core.internal`.

wilzbach:
> They're in core.internal.traits so no, not unless we move traits out of core.internal.

Why not?
We can do everything with .internal modules - even use public symbols there. And we have done so in the past too.

PR #2413, opened 2018-12-19
_d_assert_fail: Add formatting for Assertion errors

wilzbach:
Requires https://github.com/dlang/dmd/pull/8517

This adds `_d_assert_fail` to `object.d` which allows for nicely formatted and informative error messages on an assertion failure.

```
int a = 1, b = 2;
assert(a == b); // 1 != 2 (and no longer) 
```

Compare at the moment, it would yield sth. like `core.exception.AssertError@onlineapp.d(4): Assertion failure`. But I think we all agree that we can do better.

The printing supported added here is pretty minimalistic as
- this is still behind a version flag
- the program will terminate after this formatting (so saving a few bytes isn't a huge concern)
- Phobos shouldn't be used to reduce the compilation overhead of this feature

tl;dr: yes, the formatting can be improved (and should be before this feature gets activated by default), but this PR adds enough formatting to bootstrap https://github.com/dlang/dmd/pull/8517 as an existing `_d_assert_fail` is required. Also, once this is merged, it will be a lot easier for everyone to improve upon this.

### DIFF
--- a/changelog/assert.dd
+++ b/changelog/assert.dd
@@ -53,7 +53,7 @@ assert expression:
 void main()
 {
     int a, b = 2;
-    assert(a == b);
+    assert(a && a == b);
 }
 ---
 
--- a/changelog/assert.dd
+++ b/changelog/assert.dd
@@ -53,7 +53,7 @@ assert expression:
 void main()
 {
     int a, b = 2;
-    assert(a && a == b);
+    assert(a && (a == b));
 }
 ---
 
--- a/changelog/assert.dd
+++ b/changelog/assert.dd
@@ -1,0 +1,71 @@
+Context-aware assertion error messages
+
+With this release DMD supports generating context-aware assertion error messages
+when no error message has been provided by the user.
+For example, currently the following file:
+
+---
+void main()
+{
+    int a, b = 2;
+    assert(a == b);
+}
+---
+
+would yield this error when compiled and run:
+
+$(CONSOLE
+> dmd -run main.d
+core.exception.AssertError@main.d(4): Assertion failure
+$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)
+??:? _d_assertp [0x1c4eae48]
+onlineapp.d:4 _Dmain [0x1c4ead85]
+)
+
+However, with the new experimental compiler switch `-checkaction=context` it yields:
+
+$(CONSOLE
+> dmd -checkaction=context -run main.d
+core.exception.AssertError@main.d(4): 0 != 2
+$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)
+??:? _d_assert_msg [0x4a3f9cf0]
+??:? _Dmain [0x4a3f8fc4]
+)
+
+The new switch already supports a variety of assertion messages:
+
+---
+string dlang = "d2";
+assert(dlang != dlang); // ERROR: "dlang" == "dlang"
+
+struct S { int s; }
+assert(S(0) == S(1)); // ERROR: "S(0) !is S(1)"
+
+int a = 1, b = 2);
+assert(a > b); // ERROR: 1 <= 2
+---
+
+Also if no error message can be generated, it will now fallback to displaying
+the text of the `assert` expression. For example, for this more complicated
+assert expression:
+
+---
+void main()
+{
+    int a, b = 2;
+    assert(a == b);
+}
+---
+
+Compiling and running with `-checkaction=context` will now result in:
+
+$(CONSOLE
+> dmd -checkaction=context -run main.d
+core.exception.AssertError@main.d(4): assert(a && (a == b)) failed
+$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)$(NDASH)
+??:? _d_assert_msg [0xb7e5fdfc]
+??:? _Dmain [0xb7e5fd40]
+)
+
+This switch for context-aware assertion error messages is still experimental
+and feedback is welcome.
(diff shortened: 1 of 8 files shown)


wilzbach:
Oh and as this depends on the DMD PR, we will have to do a druntime -> dmd -> druntime PR series.
So I disabled the added tests from running by removing `assert_fail` from `TESTS`. Once this is merged and the respective DMD PR, we can actually enable the test. It sucks that the repos are so intertwined.

src/object.d
```diff
@@ -4795,3 +4795,152 @@ TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @nogc pure @trusted
...
+private template getPrintfFormat(T)
+{
+    static if (is(T == long))
+    {
+        enum getPrintfFormat = "%lld";
+    }
+    else static if (is(T == ulong))
+    {
+        enum getPrintfFormat = "%llu";
+    }
```
jacob-carlborg: According to Clang on macOS:

* `c_long`  -> `%ld` 
* `c_ulong` -> `%lu`
* `size_t` on 64bit-> `%lu` 
* `size_t` on 32bit-> `%zu`

src/object.d
```diff
@@ -4795,3 +4795,152 @@ TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @nogc pure @trusted
...
+    import core.stdc.stdio : sprintf;
+    import core.stdc.string : strlen;
+    static if (__traits(isIntegral, V))
+    {
+        enum printfFormat = getPrintfFormat!V;
+        char[20] val;
+        sprintf(val.ptr, printfFormat, v);
+        return val.idup[0 .. strlen(val.ptr)];
+    }
+    else static if (__traits(isFloating, V))
```
jacob-carlborg: Why a separate case for floats, why not handle this in `getPrintfFormat`?

src/object.d
```diff
@@ -4795,3 +4795,152 @@ TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @nogc pure @trusted
...
+    {
+        char[60] val;
+        sprintf(val.ptr, "%g", v);
+        return val.idup[0 .. strlen(val.ptr)];
+    }
+    else static if (__traits(compiles, { string s = V.init.toString(); }))
+    {
+        return v.toString();
+    }
+    else static if (is(V : string))
```
jacob-carlborg: Perhaps any kind of a character array should be printed as a string.

src/object.d
wilzbach: Because I kept the formatting very naive for floating points numbers and always use `%g` and a very wide buffer for the reasons mentioned above (first version of this, happens before program termination).

src/object.d
wilzbach: Done.

src/object.d
wilzbach: We don't need to worry about `size_t` as it will be either `ulong` or `uint`, but I changed the formatting for (u)longs to `%lu` and `%ld`.

jacob-carlborg: No, it was correct before. `long long` in C is `%lld` and `unsigned long long is `%llu`. It uses different format specifiers for `long` and `long long`. I was thinking of adding additional cases for `c_long`.

src/object.d
jacob-carlborg: So you're thinking this will be extended with a more complicated formatting of floats?

src/object.d
jacob-carlborg: Regarding `size_t`, it's supposed to use `%zu` on 32bit macOS. Not sure what the difference is compared to `%u`.

src/object.d
```diff
@@ -4795,3 +4795,153 @@ TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @nogc pure @trusted
     foreach (v; s)
         assert(v == cast(short) 0xabab);
 }
+
+// Yields the appropriate printf format token for a type T
+// Indended to be used by miniFormat
+private template getPrintfFormat(T)
```
ZombineDev: Can we move the new code to `core.internal.assert`?

wilzbach: Can't name it `assert` as it's a keyword, thus I went for `dassert`.

ZombineDev: Sure, thanks!

thewilsonator:
@wilzbach please give this a kick.

wilzbach:
> @wilzbach please give this a kick.

I changed the DMD PR, s.t. it passes without the druntime PR.
So we can first merge the DMD PR and this PR will pass (that's why I re-enabled the tests here).

thewilsonator:
Sounds good.

wilzbach:
> So we can first merge the DMD PR and this PR will pass (that's why I re-enabled the tests here).

Rebased & added a changelog entry.

wilzbach:
Changelog preview:

image

changelog/assert.dd
```diff
@@ -0,0 +1,71 @@
...
+struct S { int s; }
+assert(S(0) == S(1)); // ERROR: "S(0) !is S(1)"
+
+int a = 1, b = 2);
+assert(a > b); // ERROR: 1 <= 2
+---
+
+Also if no error message can be generated, it will now fallback to displaying
+the text of the `assert` expression. For example, for this more complicated
+assert expression:
```
rainers: It's the same expression as above.

src/core/internal/dassert.d
```diff
@@ -0,0 +1,147 @@
...
+        case ">":
+            return "<=";
+        case ">=":
+            return "<";
+        case "is":
+            return "!is";
+        case "!is":
+            return "is";
+        default:
+            return "!=";
```
rainers: I wouldn't default to `!=`, better fall back to something using the original operator name.

wilzbach: Ah actually we whitelist in DMD which tokens are allowed (see https://github.com/dlang/dmd/blob/1ffe034e743f5f7833ce5ed457b1f5eb2012d718/src/dmd/expressionsem.d#L5546), so I think the better course of action is to use an `assert(0)`.

changelog/assert.dd
wilzbach: Fixed thanks!

wilzbach:
Any other requirements that this initial implementation needs to fulfill?
It's not intended to be perfect, but at least better than just a non-informational `AssertError`.

thewilsonator:
No, I think this is a good start and we'll see what the feedback is like.

PR #2414, opened 2018-12-20
Fix Issue 19498 - undefined identifier rt_loadLibraryW

edi33416:


edi33416:
@WalterBright

src/core/runtime.d
```diff
@@ -21,6 +21,8 @@ else version (WatchOS)
 
 /// C interface for Runtime.loadLibrary
 extern (C) void* rt_loadLibrary(const char* name);
+/// ditto
+version (Windows) extern (C) void* rt_loadLibrary(const WCHAR* name);
```
kinke: Duplicate declarations for `version (Windows)` like this are rejected by LDC, and rightly so IMO.

I don't think it's a good idea to have a C interface which diverges for Windows. The different param type isn't part of the mangled name, so you won't get link errors, but ugly crashes for external (C[++]...) code using that interface.

thewilsonator: @edi33416 per https://github.com/dlang/druntime/pull/2392#pullrequestreview-186042151 that should be added back as `rt_loadLibraryW` that uses `WCHAR*`

edi33416: @thewilsonator I had the same fix in mind, but please see @WalterBright [comments](https://issues.dlang.org/show_bug.cgi?id=19498#c1) on the issue.

edi33416: I can't move the `core.runtime.loadLibrary()` implementation to `rt.dmain2.rt_loadLibrary()`, as `rt_loadLibrary()` takes a `const char*` instead of `const char[]`, so it's losing the length information (unless it's safe to call `strlen(name)` )

src/core/runtime.d
```diff
@@ -227,27 +227,8 @@ struct Runtime
...
-
-            len = MultiByteToWideChar(
-                CP_UTF8, 0, name.ptr, cast(int)name.length, buf, len);
-            if (len == 0)
-                return null;
-
-            buf[len] = '\0';
-
-            return rt_loadLibraryW(buf);
+            name[$ - 1] = '\0';
```
thewilsonator: Is this actually tested?  Given `in char[] name` I would expect mutation to fail.

rainers: Yes, fails to compile. Time to add a test similar to test/uuid?

rainers:
Pedantically, this and https://github.com/dlang/druntime/pull/2392 are breaking changes:
- rt_loadLibraryW was a public symbol before
- rt_loadLibrary now assumes a different encoding (UTF8 instead of ANSI)

rainers:
My suggestion: just restore the declaration: https://github.com/dlang/druntime/pull/2425

ZombineDev:
#2425 was merged -> closing.

PR #2415, opened 2018-12-25
win32 runtime patch

despair86:
this fixes a few corner cases on 32-bit windows where `RtlCaptureContext(2)` may not be present


src/rt/minit.asm
```diff
@@ -109,6 +109,58 @@ __minit:
     mov dword ptr __moduleinfo_array,EAX
     ret
 
+; RTLCaptureContext - Windows 2000 does not have this!
```
jacob-carlborg: I’m not sure if we’re supporting that old version of Windows.

PR #1022, opened 2014-11-14
[WIP] precise GC as presented at DConf 2013

rainers:
This is the precise GC as presented at DConf 2013. It generates pointer bitmaps for each type T using RTInfo!T and stores a bitmap alongside pool memory to use for following only pointer references.

RTInfo generation isn't working properly at the moment (needs https://github.com/D-Programming-Language/dmd/pull/3958 or https://github.com/D-Programming-Language/dmd/pull/2480), but using the compiler default will just scan some types conservatively instead of precisely.

Some notes:
- only the heap is scanned precisely, not stacks or the data/TLS segments
- compilation is slower: I measured 3:46min for the "quick" test suite on Win32 instead of 3:26 for master (we might need a special trait to avoid a lot of CTFE)
- execution is slower: the allocation heavy druntime benchmarks take 58sec instead of 51sec (unfortunately there are 2 list-benchmark  which don't give reliable timings).
- associative arrays need a dirty hack to support precise scanning of the key/value pairs in Entry: the TypeInfo for that is geneerated at runtime and stored in the Impl struct of each instance.
- dynamic arrays need to "re-emplace" their type info pointer bitmaps for large arrays, because data is "sometimes" placed at an offset of 16 bytes

I'm opening this as a pull request to verify building on other platforms aswell as making it more publically visible and open for review/discussion.

I plan to create another version that stores type info per allocation block and use this during scanning. This can avoid the troubles for the arrays, but will not allow reassigning new types to partial memory blocks (e.g. when using std.emplace on fields of a class).

PR #2418, opened 2018-12-27
Another precise GC implementation

rainers:
This is a variation of https://github.com/dlang/druntime/pull/1603, but with less changes to the GC interface.

It works without adding an `emplace` function and new `BlkAttr` flags. Instead, it assumes knowledge about the implementation of dynamic arrays. The `APPENDABLE` attribute tells the GC to repeat the pointer bitmap given by the elements type info. In addition, a memory block in the large object pool with APPENDABLE set is assumed to contain data starting at offset 16.

There is no longer a bit reserved for each possible memory location for data in the large object pool. Instead the rtInfo of its type is saved for every page, so allocation has almost no overhead in that case. This also allows restricting scanning to the actual size of the type.

In case you are running the conservative GC, the overhead of having the option to switch to the precise GC is hardly noticable:
- a bool check when creating an AA (it has to build the (key,value)-pair type info at runtime for precise information (this already happens if either has a destructor)
- a bool check per allocation
- a bool check during collection
- some additional bytes for each type info containing the pointer bitmap (identical bitmaps merged).

The checks in the GC could be omitted by making the GC a template with preciseness as a parameter. That duplicates a lot of code, though. I'd reconsider that after #1924 is available so you can easily opt-in/out at link time.

This PR needs #2417 to successfully run the benchmarks. In general, the precise version still has results 0-10% slower than the conservative GC on Win64, the difference is smaller for Win64. The tests don't produce many false pointers, though.


src/gc/config.d
```diff
@@ -16,7 +16,7 @@ struct Config
 {
     bool disable;            // start disabled
     ubyte profile;           // enable profiling with summary when terminating program
-    string gc = "conservative"; // select gc implementation conservative|manual
+    string gc = "precise"; // select gc implementation conservative|precise|manual
```
thewilsonator: Do we really want to change that default?

src/gc/bits.d
```diff
@@ -19,14 +19,21 @@ import core.stdc.string;
 import core.stdc.stdlib;
 import core.exception : onOutOfMemoryError;
 
+
+// use version bitwise to disable optimizations that use word operands
+// on bulk operation copyRange, setRange, clrRange, etc.
+// version = bitwise;
```
thewilsonator: That version should probably be namespaced and more descriptive.

src/rt/typeinfo/ti_cdouble.d
```diff
@@ -71,4 +71,6 @@ class TypeInfo_r : TypeInfo
         arg2 = typeid(double);
         return 0;
     }
+
+    override @property immutable(void)* rtInfo() nothrow pure const @safe { return rtinfoNoPointers; }
```
thewilsonator: Can this stuff be split off into another PR?

src/rt/dmain2.d
```diff
@@ -65,11 +65,12 @@ extern (C) void _d_monitor_staticctor();
 extern (C) void _d_monitor_staticdtor();
 extern (C) void _d_critical_init();
 extern (C) void _d_critical_term();
-extern (C) void gc_init();
+extern (C) void gc_config() @nogc;
```
thewilsonator: Does this really need to change?

src/gc/config.d
rainers: No, but I want the tests to run with the precise version. It should be reverted to "conservative" before merging.

src/gc/bits.d
rainers: Not sure what you mean. `version = gc_bits_bitwise;`?

src/rt/dmain2.d
rainers: The AA initialization reads the configuration once, but as the GC is initialized lazily, its configuration needs to be updated early. I don't like checking for "precise" in aaa_init(), though. It should be something other implementations can also configure.

rainers:
With respect to splitting it up: seperate PRs don't make much sense without the GC, so I'm a bit hesitant. I can make them seperate commits, though.

schveiguy:
> seperate PRs don't make much sense without the GC

The bitrange changes can stand alone.

rainers:
rebased and added a test case to actually check skipping false pointers. I hope it's not going to be a cause for spurious failures as there is no guarantee no other false pointers exist.

src/gc/impl/conservative/gc.d
```diff
@@ -243,6 +247,39 @@ debug (LOGGING)
 
 /* ============================ GC =============================== */
 
+
+debug(PRINTF)
+void printGCBits(GCBits* bits)
```
rainers: I think these and other debug functions (LOGGING/PRINTF) should go to the bottom of the file, but that's for another PR.

src/rt/dmain2.d
rainers: As an alternative we could also opt to generate the RTInfo of the key/value pair unconditionally (maybe with some optimization for no pointers), in the hope that the compiler will generate the proper typeinfo eventually.

src/gc/config.d
```diff
@@ -36,7 +36,7 @@ struct Config
         string s = "GC options are specified as whitespace separated assignments:
     disable:0|1    - start disabled (%d)
     profile:0|1|2  - enable profiling with summary when terminating program (%d)
-    gc:conservative|manual - select gc implementation (default = conservative)
+    gc:conservative|precise|manual - select gc implementation (default = conservative)
```
wilzbach: Just a reminder to update the spec too (or at least open up a PR, s.t. it doesn't get lost once this PR has been merged): https://github.com/dlang/dlang.org/blob/master/spec/garbage.dd#L386

marcioapm:
What is the impact? Would be nice to get some impact estimates on the changelog.

wilzbach:
> What is the impact? Would be nice to get some impact estimates on the changelog.

From @rainers's initial comment:

> In general, the precise version still has druntime/benchmark results 0-10% slower than the conservative GC on Win64, the difference is smaller for Win32. The tests don't produce many false pointers, though.

src/gc/config.d
rainers: https://github.com/dlang/dlang.org/pull/2553

rainers:
BTW: the failure on buildkite is that the ocean library needs a tiny patch https://github.com/sociomantic-tsunami/ocean/pull/663 that is already merged, but not yet available in a tagged version. As the default GC will be reverted to 'conservative' that shouldn't be a blocker.

The precisegc test can be a source of spurious errors though due to false pointers on the stack or in registers. Maybe it should be changed to just not crash?

rainers:
> The precisegc test can be a source of spurious errors though due to false pointers on the stack or in registers. Maybe it should be changed to just not crash?

The test now only produces some output insted of failing. It still verifies that not more objects are collected than expected.

I've also restored the conservative GC as the default now.

thewilsonator:
Let me know when this is good to go.

rainers:
I've made a (hopefully) last change: the dependency of aaa.d on `gc.config.gc == "precise"` was a bit ugly, so I changed `fakeEntryTI` to always generate the proper RTInfo for the key/value pair struct:
- it is now available for other GC implementations, too
- there is a lot happening when `Impl` is being created, the TypeInfo entry only adds a bit extra. It might be created a bit more often with the conservative GC, but is now optimized to be created less often for the precise GC
- it is no longer limited to small key/value pairs (512 bytes for 64-bit process, 128 bytes for 32-bit process), but generates proper RTInfo for any size.
- it has some direct unittests for proper RTInfo generation
- long term, the TypeInfo for the key/value pair struct should be generated by the compiler and added to TypeInfoAssocArray, making all this runtime generation obsolete.

For testing purposes, the default GC is back to "precise", I'll revert it once tests have passed.

rainers:
> Let me know when this is good to go.

Tests passed (but ocean as expected), so I reverted to "conservative". I squashed commits to remove temporary changes. Good to go IMO.

rainers:
Hooray! It's just been about 6 years since its first incarnation... ;-) The first PR was a bit later, though: https://github.com/dlang/druntime/pull/1022

PR #2426, opened 2018-12-30
Fix Issue 19524 - Make core.checkedint work in betterC

n8sh:
Do this by making everything in checkedint a zero-argument template function. The risk is that this changes the function mangling and may cause `mulu` etc. to cease being intrinsic functions if a compiler had been handling them specially, although DMD does not appear to have been doing this.

### DIFF
--- a/src/core/checkedint.d
+++ b/src/core/checkedint.d
@@ -47,7 +47,7 @@ pure:
  */
 
 pragma(inline, true)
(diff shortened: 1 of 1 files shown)


ibuclaw:
> cease being intrinsic functions if a compiler had been handling them specially, although DMD does not appear to have been doing this.

And you didn't bother checking other compilers? Are they not important enough to give due notice?

PR #2428, opened 2018-12-31
Add C++ allocator and type traits

thewilsonator:
cc @TurkeyMan 

I need this for dealing with stuff like. 

```C++
template<typename T,
         typename = typename std::enable_if<std::is_base_of<NodeRef, T>::value>::type >
class Array : public NodeRef {
...
}
```

### DIFF
--- a/src/core/stdcpp/allocator.d
+++ b/src/core/stdcpp/allocator.d
@@ -1,0 +1,54 @@
+/**
+ * D header file for interaction with C++ std::allocator.
+ *
+ * Copyright: Copyright (c) 2018 D Language Foundation
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors:   Guillaume Chatelet
+ *            Manu Evans
+ * Source:    $(DRUNTIMESRC core/stdcpp/allocator.d)
+ */
+
+module core.stdcpp.allocator;
+
+extern(C++, "std"):
+
+/**
+ * Allocators are classes that define memory models to be used by some parts of
+ * the C++ Standard Library, and most specifically, by STL containers.
+ */
+extern(C++, class) struct allocator(T)
+{
+    static assert(!is(T == const), "The C++ Standard forbids containers of const elements because allocator!(const T) is ill-formed.");
+    static assert(!is(T == immutable), "immutable is not representable in C++");
+
+    alias value_type = T;
+
+    version (CppRuntime_Microsoft)
+    {
+        T* allocate(size_t count) nothrow @nogc;
+        // HACK: workaround to make `deallocate` link as a `T * const`
+        extern (D) private static string constHack(string name)
+        {
+            version (Win64)
+                enum sub = "AAXPE";
+            else
+                enum sub = "AEXPA";
+            foreach (i; 0 .. name.length - sub.length)
+                if (name[i .. i + sub.length] == sub[])
+                    return name[0 .. i + 3] ~ 'Q' ~ name[i + 4 .. $];
+            assert(false, "substitution string not found!");
+        }
+        pragma(mangle, constHack(deallocate.mangleof))
+        void deallocate(T* ptr, size_t count) nothrow @nogc;
+    }
+    else
+    {
+        T* allocate(size_t count, const(void)* = null) nothrow @nogc;
+        void deallocate(T* ptr, size_t count) nothrow @nogc;
+    }
+
+    //    extern (D) T[] allocArray(size_t count) nothrow @nogc   { return allocate(count)[0 .. count]; }
+    //    extern (D) void deallocArray(ref T[] array) nothrow @nogc   { deallocate(array.ptr, array.length); }
+}
--- a/src/core/stdcpp/allocator.d
+++ b/src/core/stdcpp/allocator.d
@@ -49,6 +49,6 @@ extern(C++, class) struct allocator(T)
         void deallocate(T* ptr, size_t count) nothrow @nogc;
     }
 
-    //    extern (D) T[] allocArray(size_t count) nothrow @nogc   { return allocate(count)[0 .. count]; }
-    //    extern (D) void deallocArray(ref T[] array) nothrow @nogc   { deallocate(array.ptr, array.length); }
+    extern (D) T[] allocArray(size_t count) nothrow @nogc   { return allocate(count)[0 .. count]; }
+    extern (D) void deallocArray(ref T[] array) nothrow @nogc   { deallocate(array.ptr, array.length); }
 }
--- a/src/core/stdcpp/type_traits.d
+++ b/src/core/stdcpp/type_traits.d
@@ -1,0 +1,34 @@
+/**
+ * D header file for interaction with C++ std::type_traits.
+ *
+ * Copyright: Copyright Digital Mars 2018.
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors:   Manu Evans
+ * Source:    $(DRUNTIMESRC core/stdcpp/type_traits.d)
+ */
+
+module core.stdcpp.type_traits;
+
+extern(C++, "std"):
+
+struct integral_constant(T, T Val)
+{
+    enum T value = Val;
+    alias value_type = T;
+    alias type = integral_constant!(value_type, value);
+}
+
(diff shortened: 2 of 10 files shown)


src/core/stdcpp/type_traits.d
```diff
@@ -0,0 +1,34 @@
...
+
+module core.stdcpp.type_traits;
+
+extern(C++, "std"):
+
+struct integral_constant(T, T Val)
+{
+    enum T value = Val;
+    alias value_type = T;
+    alias type = integral_constant!(value_type, value);
```
rainers: Can this be different from `typeof(this)`?

rainers:
Is it possible to add a test case in https://github.com/dlang/druntime/tree/master/test/stdcpp, especially for the allocate/deallocate functions?

thewilsonator:
Will do.

src/core/stdcpp/type_traits.d
thewilsonator: No idea, will check.

thewilsonator: Looks like it always is.

thewilsonator:
Done.

thewilsonator:
```
/home/circleci/druntime/test/stdcpp/src/allocator_test.d:11: undefined reference to `std::allocator<int>::allocate(unsigned long, void const*)'
```

Maybe it's not linking the c++ stdlib? How do we do that?

test/stdcpp/win64.mak
```diff
@@ -5,8 +5,12 @@ MODEL=64
 DRUNTIMELIB=druntime64.lib
 CC=cl
 
-test:
-	"$(CC)" -c /Foarray_cpp.obj test\stdcpp\src\array.cpp /EHsc
-	"$(DMD)" -of=test.exe -m$(MODEL) -conf= -Isrc -defaultlib=$(DRUNTIMELIB) -main -unittest test\stdcpp\src\array_test.d array_cpp.obj
-	test.exe
-	del test.exe test.obj array_cpp.obj
+TESTS= allcator array
```
Geod24: typo

thewilsonator: Thanks.

src/core/stdcpp/allocator.d
```diff
@@ -0,0 +1,54 @@
...
+                enum sub = "AAXPE";
+            else
+                enum sub = "AEXPA";
+            foreach (i; 0 .. name.length - sub.length)
+                if (name[i .. i + sub.length] == sub[])
+                    return name[0 .. i + 3] ~ 'Q' ~ name[i + 4 .. $];
+            assert(false, "substitution string not found!");
+        }
+        pragma(mangle, constHack(deallocate.mangleof))
+        void deallocate(T* ptr, size_t count) nothrow @nogc;
```
rainers: This is interesting: deallocate sets the mangling by retrieving and modifying its own mangling. It seems to works, but looks brittle.

jacob-carlborg: Yeah, I did not expect this to work.

rainers: Actually, in my tests (by default with VS2013) the const hack is wrong. Linking succeeds without it.

test/stdcpp/win64.mak
```diff
@@ -5,8 +5,12 @@ MODEL=64
...
-test:
-	"$(CC)" -c /Foarray_cpp.obj test\stdcpp\src\array.cpp /EHsc
-	"$(DMD)" -of=test.exe -m$(MODEL) -conf= -Isrc -defaultlib=$(DRUNTIMELIB) -main -unittest test\stdcpp\src\array_test.d array_cpp.obj
-	test.exe
-	del test.exe test.obj array_cpp.obj
+TESTS= allocator array
+
+test: $(TESTS)
+
+$(TESTS): %.obj %.d %.cpp
```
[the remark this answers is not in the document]
thewilsonator: > %-rules are not recognized by DM make.

*laments* Why do we have such a terrible build system?

> The output of C++ compilation should be `/Fo$@_cpp.obj`.

Thanks.

> Stilll causes link errors, though.

I think the C++ standard library needs to be linked in. What is it called on windows?

src/core/stdcpp/allocator.d
thewilsonator: Wonderful! Do we have a version or `__tratits(getTargetInfo, "blah")` to detect what version of VS we are linking against?

test/stdcpp/win64.mak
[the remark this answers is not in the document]
jacob-carlborg: We do have `build.d` as an alternative for make.

src/core/stdcpp/allocator.d
rainers: I retried with VS2015 and VS2017, these also pass without complaining about `deallocate` (without the const hack), but a missing generic delete in the array-test:
```
"c:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023/bin/Hostx64/x64/cl.exe" -c /Foallocator_c
pp.obj test\stdcpp\src\allocator.cpp /EHsc
Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27023.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

allocator.cpp

"c:\s\d\rainers\windows\bin\dmd.exe" -of=allocator.exe -m64 -conf= -Isrc -defaultlib=lib\druntime64.lib -main -unittest test\stdcp
p\src\allocator_test.d allocator_cpp.obj

allocator.exe

del allocator.exe allocator.obj allocator_cpp.obj
"c:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023/bin/Hostx64/x64/cl.exe" -c /Foarray_cpp.o
bj test\stdcpp\src\array.cpp /EHsc
Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27023.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

array.cpp

"c:\s\d\rainers\windows\bin\dmd.exe" -of=array.exe -m64 -conf= -Isrc -defaultlib=lib\druntime64.lib -main -unittest test\stdcpp\sr
c\array_test.d array_cpp.obj
array_cpp.obj : error LNK2019: unresolved external symbol "void __cdecl operator delete(void *,unsigned __int64)" (??3@YAXPEAX_K@Z
) referenced in function "public: virtual void * __cdecl std::_Generic_error_category::`scalar deleting destructor'(unsigned int)"
 (??_G_Generic_error_category@std@@UEAAPEAXI@Z)
array.exe : fatal error LNK1120: 1 unresolved externals
Error: linker exited with status 1120

--- errorlevel 1
```

TurkeyMan: Yes this line is sensitive to VS version. I wrote this against VS2017, which is 'current', but we may need to be able to specify the target VS version (similar problem to the recent `-cppstd` debate.

TurkeyMan:
Sorry, I've been AFK since before Christmas.

So what's going on here... you've extracted `allocator` and `type_traits` from my `std::string` PR into a distinct PR?
I'm not sure I understand what this is for?

thewilsonator:
@TurkeyMan this is self contained, easier to review, necessary for more things than just `std::string`, e.g. `std::vector` and dealing with `enable_if` BS.

Geod24:
👍 for the allocator split, I agree getting digestable diffs will increase review/merge speed.

Not so found of the `type_traits` part, as from my experience in C++ binding, type traits were always much better dealt with using D's superior reflection abilities.
On the plus side it might allow to keep the code closer to the C++ one, but I think most if not all usage can be replaced with a trivial D construct. Do you have an example of where it shows its true usefulness ?

Geod24:
Oh, that makes perfect sense now. Thanks.

thewilsonator:
> Incidentally, here's another case of re-attributing all the commit metadata.

I ain't got the git-fu for that.

> I might further split this into 2 separate PR's, and we can get the type_traits one in eagerly.

Sure, sounds good.

jacob-carlborg:
> I ain't got the git-fu for that.

A commit can have a separate author and a separate committer. You can change the author using:

```
git commit --amend --author="Author Name <email@address.com>"
```

wilzbach:
> I ain't got the git-fu for that.

It's not that hard. It's actually more effort to misattribute it. If you checkout a PR branch locally, you have the commit(s) from the original author. You might want to squash them before you amend work, but why would you change the metadata. Are you not using git to checkout a PR, but downloading the files individually from GitHub's web interface?

TurkeyMan:
See #2439, and we'll work out the allocator stuff further... it depends on some other things in the works to define the extern's properly.

TurkeyMan:
#2444 should conclude the work in this patch. This should be closed.

PR #2430, opened 2018-12-31
GC: keep track of the size of contiguous free pages to avoid loops

rainers:
This is an optimization inspired by a report on the forum: https://forum.dlang.org/thread/zrouxhbkssrncyinhvdz@forum.dlang.org

The idea is to keep track of the size of contiguous blocks of free pages in the large object pool, so it is unnecessary to ever scan it until the next used block. The necessary memory for this is already available as the bPageOffsets array is used for a similar purpose for allocated pages.

On my system, the rand_large benchmark is about 8% faster for Win32, and about 15% faster for Win64. Most of the time is saved during the sweep phase. The other benchmarks are not really affected because they don't use a lot of large allocations.


thewilsonator:
Nice!

thewilsonator:
@rainers This good to go?

rainers:
> @rainers This good to go?

Yes, I think so.

PR #2431, opened 2018-12-31
Add CPU_COUNT

kubo39:


src/core/sys/linux/sched.d
```diff
@@ -86,6 +95,11 @@ bool CPU_ISSET(size_t cpu, cpu_set_t* cpusetp) pure
     return __CPU_ISSET_S(cpu, cpu_set_t.sizeof, cpusetp);
 }
 
+int CPU_COUNT(cpu_set_t* cpusetp) pure
+{
+    return __CPU_COUNT_S(cpu_set_t.sizeof, cpusetp);
```
n8sh: What is the point of making `__CPU_COUNT_S` a private helper function? It's only called here.

kubo39: I just ported it from glibc.

PR #2432, opened 2019-01-01
Zero-init or void-init most char arrays

n8sh:
* char fields in C interop structs: zero-init (void-init if they are padding, although for a struct there is currently no difference between zero-init and void-init). This allows the struct's initializer to be entirely zero, obviating the need to store it and perhaps enabling some optimizations.
* Fixed-length char arrays used as buffers: void-initialize if large or if in heavily-used functions.

This PR changes some stuff that is commented out or in `version (none)` blocks in case it is ever re-enabled.


n8sh:
Added changelog entry.

ZombineDev:
It looks like there's some problematic whitespace in the changelog.

thewilsonator:
> It looks like there's some problematic whitespace in the changelog.

Its windows line endings, fixed now.

PR #2433, opened 2019-01-02
[dmd-cxx] Synchronize dmd-cxx with gdc-stable

ibuclaw:
GDC's github repo is dead (all active development is done in gcc/svn).  Until GCC 10 development starts next April, I'll be using this branch as the "upstream" to sync from for platform and bug fixes in GCC 9.  Starting with synchronizing the druntime fork (minus the gdc-specific bits).

The druntime master branch is still being twice weekly tested with gdc/ddmd.

Expect this branch to be a bit noisy, as system headers will be routinely ported from master as I set-up, fix and test more architectures.


ibuclaw:
> I trust you know what you're doing here.

In which case, i'll jut go ahead and merge. :-)

ibuclaw:
This is tested on gdc's CI - and I of course bootstrapped and regression test on x86_64-linux-gnu.

PR #2436, opened 2019-01-02
fix Issue 11393 - [GC] GC realloc and free don't ignore interior pointers

rainers:
fix Issue 11446 - [GC] GC realloc doesn't ignore non-GC owned pointers

changed specification of GC.realloc to something more robust:

- return null for failure (returning the original pointer is no indication of failure as it might also be returned on success. This can easily lead to overwriting memory)
- allow freeing old memory immediately (previous implementation did this for large objects when shrinking)
- only set attribute bits on existing block if reused (previous implementation didn't set bits if it reused a small block)

refactor realloc to match that spec

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -435,45 +435,54 @@ struct GC
 
 
     /**
-     * If sz is zero, the memory referenced by p will be deallocated as if
-     * by a call to free.  A new memory block of size sz will then be
-     * allocated as if by a call to malloc, or the implementation may instead
-     * resize the memory block in place.  The contents of the new memory block
-     * will be the same as the contents of the old memory block, up to the
-     * lesser of the new and old sizes.  Note that existing memory will only
-     * be freed by realloc if sz is equal to zero.  The garbage collector is
-     * otherwise expected to later reclaim the memory block if it is unused.
-     * If allocation fails, this function will call onOutOfMemory which is
-     * expected to throw an OutOfMemoryError.  If p references memory not
-     * originally allocated by this garbage collector, or if it points to the
-     * interior of a memory block, no action will be taken.  If ba is zero
-     * (the default) and p references the head of a valid, known memory block
-     * then any bits set on the current block will be set on the new block if a
-     * reallocation is required.  If ba is not zero and p references the head
-     * of a valid, known memory block then the bits in ba will replace those on
-     * the current memory block and will also be set on the new block if a
-     * reallocation is required.
+     * Extend, shrink or allocate a new block of memory keeping the contents of
+     * an existing block
+     *
+     * If `sz` is zero, the memory referenced by p will be deallocated as if
+     * by a call to `free`.
+     * If `p` is `null`, new memory will be allocated via `malloc`.
+     * If `p` is pointing to memory not allocated from the GC or to the interior
+     * of an allocated memory block, no operation is performed and null is returned.
+     *
+     * Otherwise, a new memory block of size `sz` will be allocated as if by a
+     * call to `malloc`, or the implementation may instead resize or shrink the memory
+     * block in place.
+     * The contents of the new memory block will be the same as the contents
+     * of the old memory block, up to the lesser of the new and old sizes.
+     *
+     * The caller guarantees that there are no other live pointers to the
+     * passed memory block, still it might not be freed immediately by `realloc`.
+     * The garbage collector can reclaim the memory block in a later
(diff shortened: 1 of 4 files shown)


thewilsonator:
```
changelog/gc_realloc.dd:7:   is no good indication of failure as it might also be returned on success. It can 
```
I'm glad I'm not the only one to get bitten by this beast. Please also squash.

rainers:
> Needs a chanelog entry?

I want to create a summary of the GC related changes, but a starting it here sounds reasonable. I've also fixed the example that failed to run online.

Being a change to the public API I guess this PR needs approval by @andralex .

thewilsonator:
> I want to create a summary of the GC related changes, 

Yeah, one aggregate entry would be good.

> I've also fixed the example that failed to run online.

Nice.

> Being a change to the public API I guess this PR needs approval by @andralex .

I know that is needed for phobos, not sure about druntime, @andralex you got 72h.

rainers:
> Please also squash.

This PR also contains https://github.com/dlang/druntime/pull/2437 as a seperate commit, I guess it will be slightly easier to rebase if the other PR is merged first.

thewilsonator:
Done.

rainers:
> Done.

Thanks. Squashed this PR now.

src/core/memory.d
```diff
@@ -435,35 +435,43 @@ struct GC
...
-     * originally allocated by this garbage collector, or if it points to the
-     * interior of a memory block, no action will be taken.  If ba is zero
-     * (the default) and p references the head of a valid, known memory block
-     * then any bits set on the current block will be set on the new block if a
-     * reallocation is required.  If ba is not zero and p references the head
-     * of a valid, known memory block then the bits in ba will replace those on
-     * the current memory block and will also be set on the new block if a
-     * reallocation is required.
+     * by a call to free.
+     * If p is null, new memory will be allocated via malloc.
```
andralex: p, null, malloc etc should be in code backquotes

src/gc/impl/conservative/gc.d
```diff
@@ -597,146 +597,123 @@ class ConservativeGC : GC
...
+            return mallocNoSync(size, bits, alloc_size, ti);
+
+        debug(PRINTF) printf("GC::realloc(p = %p, size = %llu)\n", p, cast(ulong)size);
+
+        Pool *pool = gcx.findPool(p);
+        if (!pool)
+            return null;
+
+        size_t psize;
+        size_t biti;
```
andralex: I figure this anachronism (variable declaration way before initialization) was kept as a sort of a style background compatibility. Perhaps it's worth dropping that and moving the declarations as close to their initialization as possible.

src/gc/impl/conservative/gc.d
```diff
@@ -597,146 +597,123 @@ class ConservativeGC : GC
...
-                    }
-                    p2 = mallocNoSync(size, bits, alloc_size, ti);
-                    if (psize < size)
-                        size = psize;
-                    //debug(PRINTF) printf("\tcopying %d bytes\n",size);
-                    memcpy(p2, p, size);
-                    p = p2;
-                }
+            if (size <= PAGESIZE / 2 || alwaysMalloc)
+                goto Lmalloc; // switching from large object pool to small object pool
```
andralex: Replace this style with local functions?

thewilsonator:
Great, no need for waiting then.

WalterBright:
@rainers :

There appears to be a problem with this:
```
src\core\memory.d(1487): [unittest] 0x556080 != `null`
```
The line in question is:
```
assert(q == null);
```
and the realloc is `p + 16` on memory allocated as `16`. So, the reference is just beyond the allocated memory. But memory could be allocated there! I think this test is invalid.

rainers:
@WalterBright Indeed, see https://github.com/dlang/druntime/pull/3149

schveiguy:
Hm... this seems like a breaking change -- freeing the memory immediately for realloc would leave the possibility of dangling pointers. Though I don't know how much code this might affect, I don't know who uses GC.realloc. And of those, who uses it without guaranteeing it's the only reference.

rainers:
@schveiguy Please note that the actual change was merged 18 months ago. realloc is just as unsafe as free, but had some pretty non-deterministic behavior.

schveiguy:
Yikes, I thought it said January of *this* year.

So sure, if it's been in that long, then it's probably fine.

PR #2437, opened 2019-01-02
fix spurious crashes/freezes due to corruption in GC.realloc caused by #2430

rainers:
while debugging #2436 I noticed an issue introduced by #2430.


rainers:
Maybe this bug has frozen buildkite as an assertion triggered in the GC causes the application to freeze in a recursive lock.

wilzbach:
> Maybe this bug has frozen buildkite

Jobs have a default timeout of an hour and agents are constantly freshly provisioned and decommissioned. However, the provisioning takes a bit, so it always takes about 15 minutes for new agents to start.

rainers:
> Jobs have a default timeout of an hour

This one's running for 5 hours now: https://buildkite.com/dlang/druntime/builds/969

thewilsonator:
All green now.

PR #2440, opened 2019-01-03
GC cleanup can now be configured as a DRT GC option

rainers:
This is another change I meant to do for a while, see changelog entry for details.

The druntime benchmarks dlist, slist and vdparser benefit from this by running 10-20% faster on Win32, hardly noticable on Win64, though. This happens because Win32 runs fewer collection but uses more memory (probably due to false pointers), with the last collection taking up to half a second for the dlist test.

Please note that the final collection is currently also run if the GC is disabled.

### DIFF
--- a/changelog/gc_cleanup.dd
+++ b/changelog/gc_cleanup.dd
@@ -1,0 +1,26 @@
+GC cleanup can now be configured as a DRT GC option
+
+The default cleanup method for the GC is to unconditionally run a
+collection before runtime termination to finalize objects
+that are still alive and hold resources that affect system state outside
+the current process. This combines the worst of possible alternatives:
+it can cause a considerable delay and does not guarantee finalization
+of all objects as roots might still exist.
+
+The cleanup behaviour can now be configured by a DRT option to the
+$(LINK2 $(ROOT_DIR)spec/garbage.html, GC configuration),
+e.g. by passing `--DRT-gcopt=cleanup:none` on the command
+line. Three options are provided:
+
+$(UL
+  $(LI collect: run a collection (the default for backward compatibility))
+  $(LI none: do nothing)
+  $(LI finalize: all live objects are finalized unconditionally)
+)
+
+As usual, you can also embed the configuration into the application by
+redefining `rt_options`, e.g.
+
+-------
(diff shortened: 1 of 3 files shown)


thewilsonator:
Whitespace fixes needed.

changelog/gc_cleanup.dd
```diff
@@ -0,0 +1,26 @@
...
+it can cause a considerable delay and does not guarantee finalization
+of all objects as roots might still exist.
+
+The cleanup behaviour can now be configured by a DRT option to the
+$(LINK2 $(ROOT_DIR)spec/garbage.html, GC configuration),
+e.g. by passing `--DRT-gcopt=cleanup:none` on the command
+line. Three options are provided:
+
+$(UL
+  $(LI collect: run a collection (the default fo backward compatibility))
```
wilzbach:
```suggestion
  $(LI collect: run a collection (the default for backward compatibility))
```

wilzbach:
Though I assume the specification (https://github.com/dlang/dlang.org/blob/master/spec/garbage.dd#L386) also needs an update PR

changelog/gc_cleanup.dd
rainers: Thanks, amended the change. Is it ok to use githubs "Commit suggestion" or do I have to squash afterwards anyway?

rainers:
> Though I assume the specification (https://github.com/dlang/dlang.org/blob/master/spec/garbage.dd#L386) also needs an update PR

Will do.

changelog/gc_cleanup.dd
wilzbach: Yes it is. Reviewers can use the `auto-merge-squash` option or select it manually in the GitHub merge select menu:

image

Though of course, it doesn't hurt to squash yourself.

thewilsonator:
Auto-merge toggled on

wilzbach:
> > Though I assume the specification (https://github.com/dlang/dlang.org/blob/master/spec/garbage.dd#L386) also needs an update PR
> 
> Will do.

@thewilsonator we typically only merge things when there's a documentation PR so that doc updates don't get forgotten or lost as they too easily do ...

rainers:
Soory, I've been a bit late with this: https://github.com/dlang/dlang.org/pull/2550

PR #2443, opened 2019-01-05
[dmd-cxx] Backport a couple windows makefile changes

ibuclaw:
Just mildly interested if this is what's causing windows builds to fail...

I'll just close and move on if it still fails.


ibuclaw:
```
make -f mak/WINDOWS import DMD="" IMPDIR="import"
conf= -c -o- -Isrc -Iimport -Hfimport\core\sync\barrier.di src\core\sync\barrier.d 
Error: 'conf=' not found
```
Oh well, I have no idea why the dmd variable is undefined...

win32.mak
```diff
@@ -2,7 +2,10 @@
 
 MODEL=32
 
-DMD=dmd
+DMD_DIR=..\dmd
+BUILD=release
+OS=windows
+DMD=$(DMD_DIR)\generated\$(OS)\$(BUILD)\$(MODEL)\dmd
```
wilzbach: I think we changed this around 2.079/2.080
Before it, the built executable was placed into `src`:

```
dmc -odmd.exe mars.obj frontend.lib glue.lib backend.lib root.lib -cpp -mn -Ar -L/STACK:8388608 -L/delexe/la
```

(from the auto-tester)

wilzbach: Looking more at at-client, it looks like it's already passing in `DMD` to the scripts:


https://github.com/braddr/at-client/blob/master/src/do_build_druntime.sh

```
DMD_PATH=`ls -1 ../dmd/generated/*/release/$COMPILER_MODEL/dmd$EXE`

$makecmd DMD=$DMD_PATH MODEL=$OUTPUT_MODEL $EXTRA_ARGS -f $makefile auto-tester-build >> ../druntime-build.log
```

So I guess one workaround would be to rename the DMD variable.

wilzbach: (committed this experiment for you, let's see whether it helps)

ibuclaw: Yep, though https://github.com/dlang/dmd/pull/9208 seems to make that change redundant, so I'll merge the other first.

ibuclaw:
@wilzbach - I assume I can just proceed with this anyway.

PR #2444, opened 2019-01-05
std::allocator

TurkeyMan:
Final piece of foundation before `std::string` and `std::vector` can move in.

### DIFF
--- a/src/core/stdcpp/allocator.d
+++ b/src/core/stdcpp/allocator.d
@@ -1,0 +1,126 @@
+/**
+ * D binding to C++ std::allocator.
+ *
+ * Copyright: Copyright (c) 2019 D Language Foundation
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors:   Manu Evans
+ * Source:    $(DRUNTIMESRC core/stdcpp/allocator.d)
+ */
+
+module core.stdcpp.allocator;
+
+import core.stdcpp.new_;
+import core.stdcpp.xutility : __cpp_sized_deallocation, __cpp_aligned_new;
+
+// some versions of VS require a `* const` pointer mangling hack
+// we need a way to supply the target VS version to the compile
+version (CppRuntime_Microsoft)
+    version = NeedsMangleHack;
+
+/**
+ * Allocators are classes that define memory models to be used by some parts of
+ * the C++ Standard Library, and most specifically, by STL containers.
+ */
+extern(C++, class)
+extern(C++, "std")
+struct allocator(T)
+{
+    static assert(!is(T == const), "The C++ Standard forbids containers of const elements because allocator!(const T) is ill-formed.");
+    static assert(!is(T == immutable), "immutable is not representable in C++");
+    static assert(!is(T == class), "Instantiation with `class` is not supported; D can't mangle the base (non-pointer) type of a class. Use `extern (C++, class) struct T { ... }` instead.");
+
+    ///
+    alias value_type = T;
+
+    version (CppRuntime_Microsoft)
+    {
+        ///
+        T* allocate(size_t count) @nogc;
+        ///
+        void deallocate(T* ptr, size_t count) @nogc;
+
+        ///
+        enum size_t max_size = size_t.max / T.sizeof;
+
+        version (NeedsMangleHack)
+        {
+            // HACK: workaround to make `deallocate` link as a `T * const`
+            private extern (D) enum string constHack(string name) = (){
+                version (Win64)
+                    enum sub = "AAXPE";
+                else
+                    enum sub = "AEXPA";
+                foreach (i; 0 .. name.length - sub.length)
+                    if (name[i .. i + sub.length] == sub[])
+                        return name[0 .. i + 3] ~ 'Q' ~ name[i + 4 .. $];
+                assert(false, "substitution string not found!");
+            }();
+            pragma(linkerDirective, "/alternatename:" ~ deallocate.mangleof ~ "=" ~ constHack!(deallocate.mangleof));
+        }
+    }
+    else version (CppRuntime_Gcc)
+    {
+        ///
+        T* allocate(size_t count, const(void)* = null) @nogc
+        {
+//            if (count > max_size)
+//                std::__throw_bad_alloc();
+
+            static if (__cpp_aligned_new && T.alignof > __STDCPP_DEFAULT_NEW_ALIGNMENT__)
+                return cast(T*)__cpp_new_aligned(count * T.sizeof, cast(align_val_t)T.alignof);
+            else
+                return cast(T*)__cpp_new(count * T.sizeof);
+        }
+        ///
+        void deallocate(T* ptr, size_t count) @nogc
+        {
+            // NOTE: GCC doesn't seem to use the sized delete when it's available...
+
+            static if (__cpp_aligned_new && T.alignof > __STDCPP_DEFAULT_NEW_ALIGNMENT__)
+                __cpp_delete_aligned(cast(void*)ptr, cast(align_val_t)T.alignof);
+            else
+                __cpp_delete(cast(void*)ptr);
+        }
+
+        ///
+        enum size_t max_size = (ptrdiff_t.max < size_t.max ? cast(size_t)ptrdiff_t.max : size_t.max) / T.sizeof;
+    }
+    else version (CppRuntime_Clang)
+    {
+        ///
+        T* allocate(size_t count, const(void)* = null) @nogc
+        {
+//            if (count > max_size)
+//                __throw_length_error("allocator!T.allocate(size_t n) 'n' exceeds maximum supported size");
+
+            static if (__cpp_aligned_new && T.alignof > __STDCPP_DEFAULT_NEW_ALIGNMENT__)
+                return cast(T*)__cpp_new_aligned(count * T.sizeof, cast(align_val_t)T.alignof);
+            else
+                return cast(T*)__cpp_new(count * T.sizeof);
+        }
+        ///
+        void deallocate(T* ptr, size_t count) @nogc
+        {
+            static if (__cpp_aligned_new && T.alignof > __STDCPP_DEFAULT_NEW_ALIGNMENT__)
+            {
+                static if (__cpp_sized_deallocation)
+                    return __cpp_delete_size_aligned(cast(void*)ptr, count * T.sizeof, cast(align_val_t)T.alignof);
+                else
+                    return __cpp_delete_aligned(cast(void*)ptr, cast(align_val_t)T.alignof);
+            }
+            else static if (__cpp_sized_deallocation)
+                return __cpp_delete_size(cast(void*)ptr, count * T.sizeof);
+            else
+                return __cpp_delete(cast(void*)ptr);
+        }
+
+        ///
+        enum size_t max_size = size_t.max / T.sizeof;
+    }
+    else
+    {
+        static assert(false, "C++ runtime not supported");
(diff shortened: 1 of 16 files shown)


src/core/stdcpp/allocator.d
```diff
@@ -0,0 +1,54 @@
+/**
+ * D header file for interaction with C++ std::allocator.
+ *
+ * Copyright: Copyright (c) 2018 D Language Foundation
```
wilzbach: 2019 ;-)

thewilsonator:
I think you need to make sure the platforms C++ standard library is linked in here, e.g.
```
/usr/include/c++/4.9/ext/new_allocator.h:110: undefined reference to `operator delete(void*)'
```

test/stdcpp/win64.mak
```diff
@@ -5,8 +5,12 @@ MODEL=64
...
-test:
-	"$(CC)" -c /Foarray_cpp.obj test\stdcpp\src\array.cpp /EHsc
-	"$(DMD)" -of=test.exe -m$(MODEL) -conf= -Isrc -defaultlib=$(DRUNTIMELIB) -main -unittest test\stdcpp\src\array_test.d array_cpp.obj
-	test.exe
-	del test.exe test.obj array_cpp.obj
+TESTS= allocator array
+
+test: $(TESTS)
+
+$(TESTS): %.obj %.d %.cpp
```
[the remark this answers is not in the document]
rainers: Just remove everything on the right side of `:`.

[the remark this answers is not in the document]
TurkeyMan: Oh yeah! :)

test/stdcpp/src/allocator.cpp
```diff
@@ -0,0 +1,12 @@
+#include <memory>
+
+//Anything that will instantiate an allocator
+#include <vector>
+
+struct Mystruct { int a; };
+void foo()
+{
+    std::vector<int> a = { 10 };
```
rainers: this doesn't seem to really instantiate the full allocator, `template std::allocator<int>` does.

test/stdcpp/src/allocator_test.d
```diff
@@ -0,0 +1,14 @@
+import core.stdcpp.allocator;
+
+extern(C++) struct Mystruct { int a; }
+
+alias AliasSeq(T...) = T;
+unittest
+{
+    static foreach(T; AliasSeq!(int,float, Mystruct))
```
jacob-carlborg: Space after comma.

src/core/stdcpp/allocator.d
```diff
@@ -0,0 +1,66 @@
...
+            pragma(linkerDirective, "/alternatename:" ~ deallocate.mangleof ~ "=" ~ constHack(deallocate.mangleof));
+        }
+    }
+    else
+    {
+        T* allocate(size_t count, const(void)* = null) nothrow @nogc;
+        void deallocate(T* ptr, size_t count) nothrow @nogc;
+    }
+
+    // TODO: should we have D helpers like these?
```
jacob-carlborg: Do you expect there will be a lot of allocations from the D side?

thewilsonator:
> Yeah, but GCC wants -lstdc++, and Clang wants -lc++... how do I distinguish the build environment and supply the right runtime lib?

Thats the same problem I was having.

test/stdcpp/src/allocator.cpp
TurkeyMan: It worked for me ;) ... i'll change it.

src/core/stdcpp/allocator.d
TurkeyMan: hard to know. If someone is linking against C++, and they're doing their memory management in C++, then I can imagine that the D code might also want to use the same allocator locally. Perhaps it would be wrapped by a `std.experimental.allocator`, but also perhaps not...?

jacob-carlborg: How about we wait with adding these helpers to see if there's a need? It's easier to add symbols than to remove.

TurkeyMan: Well that's why they're commented out ;)

src/core/stdcpp/allocator.d
```diff
@@ -0,0 +1,66 @@
...
+
+    version (CppRuntime_Microsoft)
+    {
+        T* allocate(size_t count) nothrow @nogc;
+        void deallocate(T* ptr, size_t count) nothrow @nogc;
+
+        version (NeedsMangleHack)
+        {
+            // HACK: workaround to make `deallocate` link as a `T * const`
+            extern (D) private static string constHack(string name)
```
ZombineDev: Use:
```suggestion
            private enum constHack(string name) = ()
```

as we don't want `constHack` to get codegen-ed.

test/stdcpp/src/allocator.cpp
Geod24: Yeah we had [this problem before](https://github.com/dlang/dmd/pull/9136)

test/stdcpp/src/allocator_test.d
```diff
@@ -0,0 +1,14 @@
+import core.stdcpp.allocator;
+
+extern(C++) struct Mystruct { int a; }
+
+alias AliasSeq(T...) = T;
+unittest
+{
+    static foreach(T; AliasSeq!(int, float, Mystruct))
```
Geod24: Should there also be a test for:
- `extern(C++) class`
- `Object` / `extern(D) class`

TurkeyMan: `std::allocator` doesn't work with D classes... it would just allocate a class pointer. should I wrap it up so that it does?

src/core/stdcpp/allocator.d
```diff
@@ -0,0 +1,57 @@
...
+extern(C++, "std"):
+
+/**
+ * Allocators are classes that define memory models to be used by some parts of
+ * the C++ Standard Library, and most specifically, by STL containers.
+ */
+extern(C++, class) struct allocator(T)
+{
+    static assert(!is(T == const), "The C++ Standard forbids containers of const elements because allocator!(const T) is ill-formed.");
+    static assert(!is(T == immutable), "immutable is not representable in C++");
```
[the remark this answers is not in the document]
TurkeyMan: That's a problem for the assignment/copy/move/etc operations. The asserts here are to inhibit malformed instantiations. If a type has an immutable member, then it's valid to allocate it, but it's an external problem that it won't be assignable.

src/core/stdcpp/allocator.d
```diff
@@ -0,0 +1,57 @@
+/**
+ * D header file for interaction with C++ std::allocator.
```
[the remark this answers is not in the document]
TurkeyMan: mmm

test/stdcpp/src/allocator_test.d
TurkeyMan: Actually, no, it's impossible to instantiate with a class. D has no syntax to refer to the classes actual type, only the pointer to it.

TurkeyMan: It assert's a helpful message.

TurkeyMan:
@klickverbot I tried linking with `$(CXX)`, but that just flips the problem, now G++ doesn't automatically link druntime...

thewilsonator:
Try using the makefile variable `$(DRUNTIMELIB)`

TurkeyMan:
I did.

TurkeyMan:
This whole stupid turnaround thing would just be infinitely easier if it was possible to build and run locally! >_<

TurkeyMan:
I really really do ***HATE*** contributing to dlang. I wish I never did, and just kept complaining on the forum instead.
This is the worst dev environment I've suffered any time in my life, and I just don't understand WHY.

It takes an obscene amount of my life away from me, and just makes be resent the project in general.

TurkeyMan:
Okay, I think it's there... That's in the running for the worst 2 days of my life!

TurkeyMan:
Aaaand, OSX is a massive arsehole as usual!

Can anyone suggest what is wrong? Does anyone have a mac? The error output is not helpful.

jacob-carlborg:
@TurkeyMan  I'll have a look.

jacob-carlborg:
Works for me locally.

jacob-carlborg:
The error is this?

```
Assertion failed: (linked_addr_pos != line_table_map.end()), function FixReferences, file /SourceCache/dwarf_utilities/dwarf_utilities-119/source/DWARFdSYM.cpp, line 3749.
```

jacob-carlborg:
Most likely due to old versions of the developer tools.

jacob-carlborg:
> So.... what does that mean? Is this a real issue?

Not sure. I'm guessing it's an issue with the CI environment. This might be caused by Clang being invoked with the `-g` flag (which will then call dsymutil). Can you configured Clang to not be invoked with `-g`?

wilzbach:
> @braddr: speculation is that build tooling is out of date on this particular mac?

Brad is hard to reach via GitHub. Try emailing him.

TurkeyMan:
@thewilsonator This one's passed already... if we merge this, it'll pull the other one along with it.

thewilsonator:
Auto-merge toggled on

thewilsonator:
Auto-merge toggled off

thewilsonator:
Auto-merge toggled on

Geod24:
Erg, I tried to replace my custom `allocator` with this, but hit a small issue:
```
/Users/geod24/projects/install/osx/bin/../../src/druntime/import/core/stdcpp/allocator.d(120,45): Error: struct `scpd.common.xdr.Stellar_SCP.stellar.SCPQuorumSet` no size because of forward reference
source/scpd/common/xdr/types.d(47,50): Error: template instance `core.stdcpp.allocator.std.allocator!(SCPQuorumSet)` error instantiating
source/scpd/common/xdr/types.d(215,12):        instantiated from here: vector!(SCPQuorumSet)
source/scpd/common/xdr/Stellar_SCP.d(269,5):        instantiated from here: xvector!(SCPQuorumSet, 4294967292u)
```

Here's the definition of `SCPQuorumSet`:
```d
struct SCPQuorumSet {
    uint32_t threshold;
    xvector!(PublicKey) validators;
    xvector!(SCPQuorumSet) innerSets;
}
```

TurkeyMan:
I'm not sure what to make of that. Can you boil it to a unit test? Apparently the unit test for allocator is not enough.

src/core/stdcpp/allocator.d
```diff
@@ -0,0 +1,126 @@
...
+        ///
+        void deallocate(T* ptr, size_t count) @nogc;
+
+        ///
+        enum size_t max_size = size_t.max / T.sizeof;
+
+        version (NeedsMangleHack)
+        {
+            // HACK: workaround to make `deallocate` link as a `T * const`
+            private extern (D) enum string constHack(string name) = (){
```
look-at-me: @TurkeyMan 
This is always a pain point when trying to link to C++ libraries. I've been thinking of it and possible adding a special "cpp_const" symbol that can only be used in extern(C++) functions would be worth implementing?

```D
extern(C++) void test( cpp_const char* a, cpp_const char* cpp_const b, char* cpp_const c )
{
    pragma(msg, typeof(a)); // D type: const(char)*
    pragma(msg, typeof(b)); // D type: const(char*)
    pragma(msg, typeof(c)); // D type: char* 
                            // Note:    ^ don't actually get const here, would only be for mangling
}
```

Thoughts ?

look-at-me: Yah same here, the best option would be to have const behave the same way in C++ so we can do the same mangling as C++. It'd be the simplest cleanest solution. The problem is D's const behaves differently and probably should stay the way it is. So I was thinking a new keyword could be used but it obviously won't be as nice as just using const.

PR #2446, opened 2019-01-07
C++ operator new/delete bindings

TurkeyMan:
`std::allocator` depends on this.

### DIFF
--- a/src/core/stdcpp/new_.d
+++ b/src/core/stdcpp/new_.d
@@ -1,0 +1,98 @@
+/**
+ * D binding to C++ <new>
+ *
+ * Copyright: Copyright (c) 2019 D Language Foundation
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors:   Manu Evans
+ * Source:    $(DRUNTIMESRC core/stdcpp/new_.d)
+ */
+
+module core.stdcpp.new_;
+
+import core.stdcpp.xutility : __cpp_sized_deallocation, __cpp_aligned_new;
+
+// TODO: this really should come from __traits(getTargetInfo, "defaultNewAlignment")
+enum size_t __STDCPP_DEFAULT_NEW_ALIGNMENT__ = 16;
+
+
+extern (C++):
+
+///
+extern (C++, "std") enum align_val_t : size_t { defaultAlignment = __STDCPP_DEFAULT_NEW_ALIGNMENT__ };
+
+/// Binding for ::operator new(std::size_t count)
+pragma(mangle, __new_mangle)
+void* __cpp_new(size_t count) @nogc;
+
+/// Binding for ::operator delete(void* ptr)
+pragma(mangle, __delete_mangle)
+void __cpp_delete(void* ptr) @nogc;
+
+static if (__cpp_sized_deallocation)
+{
+    /// Binding for ::operator delete(void* ptr, size_t size)
+    pragma(mangle, __delete_size_mangle)
+    void __cpp_delete_size(void* ptr, size_t size) @nogc;
+}
+static if (__cpp_aligned_new)
+{
+    /// Binding for ::operator new(std::size_t count, std::align_val_t al)
+    pragma(mangle, __new_align_mangle)
+    void* __cpp_new_aligned(size_t count, align_val_t alignment) @nogc;
+
+    /// Binding for ::operator delete(void* ptr, std::align_val_t al)
+    pragma(mangle, __delete_align_mangle)
+    void __cpp_delete_aligned(void* ptr, align_val_t alignment) @nogc;
+
+    /// Binding for ::operator delete(void* ptr, size_t size, std::align_val_t al)
+    pragma(mangle, __delete_size_align_mangle)
+    void __cpp_delete_size_aligned(void* ptr, size_t size, align_val_t alignment) @nogc;
+}
+
+private:
+
+// we have to hard-code the mangling for the global new/delete operators
+version (CppRuntime_Microsoft)
+{
+    version (D_LP64)
+    {
+        enum __new_mangle               = "??2@YAPEAX_K@Z";
+        enum __delete_mangle            = "??3@YAXPEAX@Z";
+        enum __delete_size_mangle       = "??3@YAXPEAX_K@Z";
+        enum __new_align_mangle         = "??2@YAPEAX_KW4align_val_t@std@@@Z";
+        enum __delete_align_mangle      = "??3@YAXPEAXW4align_val_t@std@@@Z";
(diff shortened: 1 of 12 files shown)


TurkeyMan:
I don't know what to do about that OSX fail. @jacob-carlborg?
Is it spurious? I'm not even sure what it wants from me...

This PR is holding all my work up, so I'd like to get it through.

jacob-carlborg:
> I don't know what to do about that OSX fail. @jacob-carlborg?
Is it spurious? I'm not even sure what it wants from me...

Seems to be the same problem as the other PR. I don't think it's something wrong with your PR, I think it's the machine the CI is running on. Try removing `-g` from `CFLAGS` here:

https://github.com/dlang/druntime/blob/227a925a1892deaab7a64a1bda576864d04dd528/test/common.mak#L26

But since it's a shared makefile it might cause other tests to fail. Alternatively don't use the `CXXFLAGS` variable here to avoid `-g` being added:

https://github.com/dlang/druntime/blob/227a925a1892deaab7a64a1bda576864d04dd528/test/stdcpp/Makefile

thewilsonator:
Can confirm, removal of `-g` now results in 
```
c++ -m64 -fPIC -Wall -std=c++11 -stdlib=libc++ -o generated/osx/debug/64/array src/array.cpp ./generated/osx/debug/64/array_d.o ~/d/druntime/generated/osx/debug/64/libdruntime.a -lpthread
ld: warning: could not create compact unwind for __D2rt9backtrace5dwarf23traceHandlerOpApplyImplFxAPvMDFKmKxAaZiZi: offset of saved registers too far to encode
```
and the execrable run successfully with no output.

thewilsonator:
```diff
diff --git a/test/common.mak b/test/common.mak
index fbd72ae..617375d 100644
--- a/test/common.mak
+++ b/test/common.mak
@@ -17,18 +17,20 @@ ROOT:=$(GENERATED)/$(OS)/$(BUILD)/$(MODEL)
 ifneq (default,$(MODEL))
 	MODEL_FLAG:=-m$(MODEL)
 endif
-CFLAGS:=$(MODEL_FLAG) $(PIC) -Wall
+CFLAGS_BASE:= $(MODEL_FLAG) $(PIC) -Wall
 DFLAGS:=$(MODEL_FLAG) $(PIC) -w -I../../src -I../../import -I$(SRC) -defaultlib= -debuglib= -dip1000
 # LINK_SHARED may be set by importing makefile
 DFLAGS+=$(if $(LINK_SHARED),-L$(DRUNTIMESO),-L$(DRUNTIME))
 ifeq ($(BUILD),debug)
 	DFLAGS += -g -debug
-	CFLAGS += -g
+    CFLAGS := $(CFLAGS_BASE) -g
 else
 	DFLAGS += -O -release
-	CFLAGS += -O3
+	CFLAGS := $(CFLAGS_BASE) -O3
 endif
+CXXFLAGS_BASE := $(CFLAGS_BASE) -std=c++11
 CXXFLAGS:=$(CFLAGS) -std=c++11
 ifeq (osx,$(OS))
 	CXXFLAGS+=-stdlib=libc++
+    CXXFLAGS_BASE+=-stdlib=libc++
 endif
diff --git a/test/stdcpp/Makefile b/test/stdcpp/Makefile
index 4497e45..1677a51 100644
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -22,7 +22,7 @@ $(ROOT)/%.done : $(ROOT)/%
 $(ROOT)/%: $(SRC)/%.cpp $(SRC)/%_test.d
 	mkdir -p $(dir $@)
 	$(QUIET)$(DMD) $(DFLAGS) -main -unittest -c -of=$(ROOT)/$*_d.o $(SRC)/$*_test.d
-	$(QUIET)$(CXX) $(CXXFLAGS) -o $@ $< $(ROOT)/$*_d.o $(DRUNTIME) -lpthread
+	$(QUIET)$(CXX) $(CXXFLAGS_BASE) -o $@ $< $(ROOT)/$*_d.o $(DRUNTIME) -lpthread

 clean:
 	rm -rf $(GENERATED)
```
Should do the job.

TurkeyMan:
I mean... yeah.
It feels kinda brutal to hack the makefiles that way; as if they weren't already complicated enough.

TurkeyMan:
Well, I pushed it... are we happy to commit that patch?

thewilsonator:
> are we happy to commit that patch?

Yes. I'd rather you not get stuck on things that, while nice, are not necessary for moving things forward. We can fix these things up later.

TurkeyMan:
And #2444, which is basically this, just wrapped in `std::allocator` API.

thewilsonator:
I know, I know. I can can only go as fast as the CI pipeline ;)

src/core/stdcpp/new_.d
```diff
@@ -0,0 +1,98 @@
...
+ * Authors:   Manu Evans
+ * Source:    $(DRUNTIMESRC core/stdcpp/new_.d)
+ */
+
+module core.stdcpp.new_;
+
+import core.stdcpp.xutility : __cpp_sized_deallocation, __cpp_aligned_new;
+
+// TODO: this really should come from __traits(getTargetInfo, "defaultNewAlignment")
+enum size_t __STDCPP_DEFAULT_NEW_ALIGNMENT__ = 16;
```
jacob-carlborg: Is this necessary to be public?

src/core/stdcpp/xutility.d
```diff
@@ -11,6 +11,22 @@
...
+    cpp11 = 201103,
+    cpp14 = 201402,
+    cpp17 = 201703
+}
+
+enum __cplusplus = __traits(getTargetInfo, "cppStd");
+
+// wrangle C++ features
+enum __cpp_sized_deallocation = __cplusplus >= CppStdRevision.cpp14 ? 201309 : 0;
+enum __cpp_aligned_new = __cplusplus >= CppStdRevision.cpp17 ? 201606 : 0;
```
jacob-carlborg: Is it necessary to all these symbols are public?

TurkeyMan:
I dunno... They are in C++...

src/core/stdcpp/new_.d
TurkeyMan: Is it conceivable that someone binds their own containers that respond to this value?
It does need to be a `getTargetInfo` though; this value is wrong.

src/core/stdcpp/xutility.d
TurkeyMan: I dunno... They are in C++...?

src/core/stdcpp/new_.d
jacob-carlborg: I don't know enough about this but it looks like implementation details. BTW, they miss Ddoc comments.

TurkeyMan: Yeah, I didn't put ddoc because I wasn't sure if public or not.

jacob-carlborg: Either have them as public and documented or not public.

PR #2442, opened 2019-01-05
Moved `move`, `moveEmplace`, and `forward` to `core.lifetime`

TurkeyMan:
WIP, I still need to wrangle phobos dependencies...

PR #2447, opened 2019-01-08
Move doesPointTo

TurkeyMan:
As a bonus on top of #2442, this re-enables the message that complains if you try and `move` a type with an interior pointer.

I'm not even convinced the assert should exist; if I want to move something, I want to bloody move it!
If I was stupid enough to create a thing with an interior pointer; then shame on me, but that shouldn't cause a compile error when I move it... perhaps the interior pointer is a tiny subset of the thing, which I intend to handle manually. If I *need* to move it, there's no escape hatch here. It feels like a higher-level concern.

Anyway, aside from the question of whether that assert should even be there at all, this PR pulls ~2000 new lines of code... to emit that error message.
I don't know if I'm comfortable with the idea of introducing 2000 lines to druntime that do nothing except emit an error message when the user made a possible mistake.
I'm not convinced the value justifies the cost...

### DIFF
--- a/changelog/emplace.dd
+++ b/changelog/emplace.dd
@@ -1,3 +1,4 @@
-Moved `std.conv.emplace` to core/lifetime.d
+Moved `std.conv.emplace`, `std.algorithm.mutation.move`, `std.algorithm.mutation.moveEmplace`, and `std.functional.forward` to core/lifetime.d
 
 `emplace` is the counterpart to `destroy`, so it has been moved to also live in druntime (core/lifetime.d) where it is accessible by projects that use a shallow runtime library stack.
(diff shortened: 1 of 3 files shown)


wilzbach:
Note that without this PR we can't get rid of the now duplicate implementation of `move,{Emplace}` in Phobos as there are tests for the `doesPointTo` check.
See also: https://github.com/dlang/phobos/pull/6848

TurkeyMan:
Yeah see, I didn't move that function, because it's obscene. The fact that function is like 2 thousand lines suggests at some serious deficiency somewhere.
Whatever the hell it's for, it should be worked-around ;)

PR #2448, opened 2019-01-08
std::vector

TurkeyMan:
This is WIP, gotta wrangle CI.

### DIFF
--- a/src/core/stdcpp/vector.d
+++ b/src/core/stdcpp/vector.d
@@ -1,0 +1,816 @@
+/**
+ * D header file for interaction with C++ std::vector.
+ *
+ * Copyright: Copyright (c) 2018 D Language Foundation
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors:   Guillaume Chatelet
+ *            Manu Evans
+ * Source:    $(DRUNTIMESRC core/stdcpp/vector.d)
+ */
+
+module core.stdcpp.vector;
+
+///////////////////////////////////////////////////////////////////////////////
+// std::vector declaration.
+//
+// Current caveats :
+// - missing noexcept
+// - nothrow @trusted @nogc for most functions depend on knowledge
+//   of T's construction/destruction/assignment semantics
+///////////////////////////////////////////////////////////////////////////////
+
+import core.stdcpp.allocator;
+
+enum DefaultConstruct { value }
+
+/// Constructor argument for default construction
+enum Default = DefaultConstruct();
+
+extern(C++, "std"):
+
+extern(C++, class) struct vector(T, Alloc = allocator!T)
+{
+    import core.lifetime : forward, move, moveEmplace, core_emplace = emplace;
+
+    static assert(!is(T == bool), "vector!bool not supported!");
+extern(D):
+
+    ///
+    alias size_type = size_t;
+    ///
+    alias difference_type = ptrdiff_t;
+    ///
+    alias value_type = T;
+    ///
+    alias allocator_type = Alloc;
+    ///
+    alias pointer = T*;
+    ///
+    alias const_pointer = const(T)*;
+
+    ///
+    alias as_array this;
+
+    /// MSVC allocates on default initialisation in debug, which can't be modelled by D `struct`
+    @disable this();
+
+    ///
+    alias length = size;
+    ///
+    alias opDollar = length;
+
+    ///
+    ref inout(T) front() inout pure nothrow @safe @nogc                     { return this[0]; }
+    ///
+    ref inout(T) back() inout pure nothrow @safe @nogc                      { return this[$-1]; }
+
+
+    // WIP...
+
+//    this(size_type count);
+//    this(size_type count, ref const(value_type) val);
+//    this(size_type count, ref const(value_type) val, ref const(allocator_type) al);
+//    this(ref const(vector) x);
+//    this(iterator first, iterator last);
+//    this(iterator first, iterator last, ref const(allocator_type) al = defaultAlloc);
+//    this(const_iterator first, const_iterator last);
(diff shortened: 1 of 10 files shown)


src/core/stdcpp/vector.d
```diff
@@ -0,0 +1,248 @@
...
+    alias const_pointer = const(T)*;
+    alias difference_type = ptrdiff_t;
+    alias size_type = size_t;
+
+//    alias iterator = pointer;
+//    alias const_iterator = const_pointer;
+//    alias reverse_iterator
+//    alias const_reverse_iterator
+
+    alias as_array this;
```
Geod24: I see why you'd go this way - I did the same in my local port. But I'm not sure it's suitable for druntime.

src/core/stdcpp/vector.d
```diff
@@ -0,0 +1,248 @@
...
+    // MSVC allocates on default initialisation in debug, which can't be modelled by D `struct`
+    @disable this();
+
+    // Capacity
+    size_type max_size() const nothrow @trusted @nogc;
+
+    void clear() nothrow;
+    void resize(size_type n);
+    void resize(size_type n, T c);
+    void reserve(size_type n = 0) @trusted @nogc;
```
Geod24: No point on putting `@trusted` on a definition. It's either `@safe` or `@system`

src/core/stdcpp/vector.d
```diff
@@ -0,0 +1,248 @@
...
+//    const_iterator end() const @trusted @nogc;
+//    const_iterator cend() const @trusted @nogc;
+
+    // no reverse iterator for now.
+
+    // MSVC allocates on default initialisation in debug, which can't be modelled by D `struct`
+    @disable this();
+
+    // Capacity
+    size_type max_size() const nothrow @trusted @nogc;
```
Geod24: Re `@trusted`. And arguably we could stick `@nogc:` at the top of this `struct`.

TurkeyMan:
No angle, iterators don't belong in D. Use `std.range` and `std.algorithm`.

src/core/stdcpp/vector.d
TurkeyMan: They're all like this. It makes these types interact with D in any meaningful way.
What would you suggest?

TurkeyMan:
Also, no need to review this yet; it's not baked.

thewilsonator:
FreeBSD fails with
```
../../src/core/stdcpp/vector.d(90): Error: destructor `core.stdcpp.vector.std.vector!(int, allocator!int).vector.~this` conflicts with destructor `core.stdcpp.vector.std.vector!(int, allocator!int).vector.~this` at ../../src/core/stdcpp/vector.d(58)
../../src/core/stdcpp/vector.d(52): Error: undefined identifier `as_array`
../../src/core/stdcpp/vector.d(61): Error: undefined identifier `size`, did you mean function `resize`?
src/vector_test.d(28): Error: template instance `core.stdcpp.vector.std.vector!(int, allocator!int)` error instantiating
```

Linux fails with 
```
../../src/core/stdcpp/vector.d(53): Error: undefined identifier `as_array`
../../src/core/stdcpp/vector.d(71): Error: undefined identifier `size`, did you mean function `resize`?
src/vector_test.d(28): Error: template instance `core.stdcpp.vector.std.vector!(int, allocator!int)` error instantiating
```

TurkeyMan:
Yes... this is not baked... **at all**.

src/core/stdcpp/vector.d
Geod24: To let users call `as_array()` themselves at first, see if the convenience is worth it. It will be easy to add, it's painful to remove.

TurkeyMan:
Windows passes comprehensive tests, and I think we should make the unit tests work and merge it. At least that will make a working reference with tests running on at least one platform that other platforms can dev against. The test ecosystem makes it SOOO HARD to get this working in one big lump.

TurkeyMan:
It's gonna happen!

src/core/stdcpp/vector.d
```diff
@@ -0,0 +1,816 @@
...
+    ///
+    alias opDollar = length;
+
+    ///
+    ref inout(T) front() inout pure nothrow @safe @nogc                     { return this[0]; }
+    ///
+    ref inout(T) back() inout pure nothrow @safe @nogc                      { return this[$-1]; }
+
+
+    // WIP...
```
thewilsonator: You going to leave this for later?

TurkeyMan: I'll be continuing on other platforms (clang next) and fleshing out the features as I go.
Some of those are implemented for windows. I probably won't do the iterator ones; there's no reason to mirror the C++ iterators in D; we have ranges.

TurkeyMan: But this is such a mammoth thing; it'd be much easier if I could do it a little more incrementally, and this is a good sync-point, because it's working on windows at least.
Windows users can try it out, and we can start to get bug reports.

TurkeyMan: Also, linux is still blocked on move construction (there's an interior pointer).

TurkeyMan:
Actually, hold off on this guy... I just found an important unit test I missed locally that reveals a copy construction bug. I'll fix that before we merge this.

TurkeyMan:
If `T` has a `@disable` constructor, there are build errors. I have some fixes locally and need to think of more unit tests of similar nature.

Geod24:
@thewilsonator :
> Actually, hold off on this guy... I just found an important unit test I missed locally that reveals a copy construction bug. I'll fix that before we merge this.

It would be great if you could hold off on some merge, especially when the CI is red or when the author asked you to...

thewilsonator:
Sorry I missed that, also CI was green, `aliak00/ooptional:Failed to download package silly from https://code.dlang.org/packages/silly/0.8.2.zip`

Geod24:
Ci was not green. If there's a spurious failure, just restart the build. You can do it on a job basis.

TurkeyMan:
It's fine. Build is green, it might be useful for someone to test out, but it won't be long before people report the issue ;) .. I'll fix it tonight.

PR #2450, opened 2019-01-08
Remove hashOf wrappers for enums

n8sh:
The wrappers in most cases appear to have no benefit. For `int` enums a wrapper that calls `hashOf!int` is larger than `hashOf!int` itself. For `string` enums the wrapper doesn't avoid code duplication because `hashOf!string` itself just calls `bytesHash`. Removing special handling of enums reduces the number of `hashOf` overrides and simplifies the template constraints.

I'd like to leave this PR open for at least 72 hours after approval in case there is a reason for special-casing enums that is not obvious to me.

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -84,12 +84,13 @@ assumption is yet to be verified.
 +/
 private template canBitwiseHash(T)
 {
-    static if (is(T EType == enum))
-        enum canBitwiseHash = .canBitwiseHash!EType;
-    else static if (__traits(isFloating, T))
-        enum canBitwiseHash = !(floatCoalesceZeroes || floatCoalesceNaNs);
-    else static if (__traits(isScalar, T))
-        enum canBitwiseHash = true;
+    static if (__traits(isScalar, T))
+    {
+        static if (__traits(isFloating, T))
+            enum canBitwiseHash = !(floatCoalesceZeroes || floatCoalesceNaNs);
+        else
+            enum canBitwiseHash = true;
+    }
     else static if (is(T == class))
     {
         enum canBitwiseHash = isFinalClassWithAddressBasedHash!T || isCppClassWithoutHash!T;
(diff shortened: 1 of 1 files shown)


ibuclaw:
Maybe add tests that checks all combinations of enums basetypes from `enum : int` through to `enum : interface`?

If there's no change in before and after, then this should be fine to go.

n8sh:
@ibuclaw is this what you have in mind?
```d
void testEnumHash()
{
    static void test(T)() if (is(T == enum))
    {
        import std.traits : EnumMembers, OriginalType;
        foreach (alias e; EnumMembers!T)
            assert(hashOf(cast(OriginalType!T) e) == hashOf(e));
    }

    enum IntEnum : int { a = 123456789, }
    test!IntEnum();

    enum DoubleEnum : double { a = 0x1.921fb54442d18p+1, }
    test!DoubleEnum();

    enum StringEnum : string { a = null, b = "some value", }
    test!StringEnum();

    static struct S { int value; }
    enum StructEnum : S { a = S(1), b = S(2), c = S(3), }
    test!StructEnum;

    interface I {}
    static class C1 : I {}
    static class C2 : I
    {
        immutable size_t x;
        this(size_t x) { this.x = x; }
        override size_t toHash() { return x; }
    }
    enum InterfaceEnum : I { a = null, b = new C1(), c = new C1(), d = new C2(123456789), }
    test!InterfaceEnum();
}
```

ibuclaw:
Yes, plus a few more distinct types: bool, char, complex, dynamic array, static array, typeof(null), __vector, function, delegate, associative array, and of course enum itself - both regular and specially handled (__c_long).

n8sh:
>and of course enum itself - both regular and specially handled (__c_long).

I'm not sure what you mean by enum itself.

ibuclaw:
```
enum Foo : int {...}
enum Bar : Foo {...}
```
I have no idea why someone would do that, but it's legal, and so something that needs to be backed by tests.

n8sh:
It seems like it's not even possible to have an enum of `__c_longlong`. I haven't yet checked `__c_long` since it's not defined on 32-bit posix.

https://run.dlang.io/is/nNQgJh
```d
import core.stdc.config;
enum E : __c_longlong { a = 0, }

void main()
{
}
```

>**onlineapp.d(2):** Error: enum `onlineapp.E` unable to resolve forward reference in definition

n8sh:
There is also a problem with interface enums.

https://run.dlang.io/is/wRMKad
```d
void main()
{
    interface I {}
    static class C : I {}
    enum E : I // If you change "E : I" to "E : C" the program doesn't crash.
    {
        a = new C(),
    }
    size_t h = hashOf(E.a); // Segfault.
}
```

ibuclaw:
> It seems like it's not even possible to have an enum of __c_longlong.

Wow, that means this would be broken as well on OSX.
```
import core.stdc.stdint;
enum E : int64_t
{
    a = 0,
}
```
Worth noting in a new issue, but not relevant here.

ibuclaw:
> There is also a problem with interface enums.

Looks like data layout is potentially mismatched between run-time and compile-time.
```
I a = E.a;      // {__vptr = 0x5555555dfa08 <vtable for test.main().C>}
I b = new C();  // {__vptr = 0x5555555e5478 <ClassInfo for test.main().C+184>}
```

n8sh:
Delegate enums also seem not to work.

https://run.dlang.io/is/hMACDw
```d
size_t doNothing(scope const int delegate(int) dg)
{
    return 0;
}

void main()
{
    enum E : int delegate(int)
    {
        identity = (x) => x,
        increment = (x) => x + 1,
    }
    doNothing(E.identity); // Error: delegate `onlineapp.main.E.__lambda3` is a nested function and cannot be accessed from D main
}
```
> **onlineapp.d(13):** Error: delegate `onlineapp.main.E.__lambda3` is a nested function and cannot be accessed from **D main**

ibuclaw:
> Delegate enums also seem not to work.

That's interesting, OK to skip that then.

n8sh:
Closing this PR. It needs an update to support enums of basetype `struct` and `union`, and if enums of basetype `interface`/`class`/`delegate` are also allowed then this PR would not result in decreased complexity in its current conception.

PR #2452, opened 2019-01-09
dmangle.d: add 'return' to constructor args

WalterBright:
Normally `-dip1000` infers the `return`, but when compiling code that links to druntime without `-dip1000` and with `-inline`, it doesn't get inferred so we add it manually. In particular, runnable/testconst.d.

This is blocking https://github.com/dlang/dmd/pull/9220

### DIFF
--- a/src/core/demangle.d
+++ b/src/core/demangle.d
@@ -54,13 +54,13 @@ pure @safe:
     enum AddType { no, yes }
 
 
-    this( const(char)[] buf_, char[] dst_ = null )
+    this( return const(char)[] buf_, return char[] dst_ = null )
     {
         this( buf_, AddType.yes, dst_ );
     }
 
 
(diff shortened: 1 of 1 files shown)


thewilsonator:
This seems exactly like the worries in https://github.com/dlang/dmd/pull/8504#issuecomment-452553620

WalterBright:
Compiling parts of the code with -dip1000 and parts without can produce link errors. druntime is compiled with -dip1000. The easiest fix is to add the annotations to druntime, and then you can link it to code compiled with or without -dip1000. That's what I did here.

rainers:
Shouldn't `return` be treated like `scope`? i.e. don't mangle it for parameters if it is inferred IIRC.

WalterBright:
`return scope` is indeed redundant with `return`, and it should mangle the same.

PR #2455, opened 2019-01-09
Overhaul of std::exception

TurkeyMan:
This had old-style virtual destructors. Also, there's a lot more commonality in the code than the original makes it appear.

### DIFF
--- a/src/core/stdcpp/exception.d
+++ b/src/core/stdcpp/exception.d
@@ -6,95 +6,111 @@
  * Copyright: Copyright (c) 2016 D Language Foundation
  * License:   $(HTTP boost.org/LICENSE_1_0.txt, Boost License 1.0).
  * Authors:   $(HTTP digitalmars.com, Walter Bright)
+ *            Manu Evans
  * Source:    $(DRUNTIMESRC core/stdcpp/_exception.d)
  */
 
 module core.stdcpp.exception;
 
-extern (C++, "std"):
+import core.stdcpp.xutility : __cplusplus, CppStdRevision;
 
 version (CppRuntime_DigitalMars)
+    version = GenericBaseException;
+version (CppRuntime_Gcc)
+    version = GenericBaseException;
+version (CppRuntime_Clang)
+    version = GenericBaseException;
+
+extern (C++, "std"):
+@nogc:
+
+///
+alias terminate_handler = void function() nothrow;
+///
+terminate_handler set_terminate(terminate_handler f) nothrow;
+///
+terminate_handler get_terminate() nothrow;
+///
+void terminate() nothrow;
+
+static if (__cplusplus < CppStdRevision.cpp17)
 {
-    import core.stdcpp.typeinfo;
-
-    alias void function() unexpected_handler;
-    unexpected_handler set_unexpected(unexpected_handler f) nothrow;
-    void unexpected();
-
-    alias void function() terminate_handler;
-    terminate_handler set_terminate(terminate_handler f) nothrow;
-    void terminate();
-
-    bool uncaught_exception();
-
-    class exception
-    {
-        this() nothrow { }
-        this(const exception) nothrow { }
-        //exception operator=(const exception) nothrow { return this; }
-        //virtual ~this() nothrow;
-        void dtor() { }
-        const(char)* what() const nothrow;
-    }
-
-    class bad_exception : exception
-    {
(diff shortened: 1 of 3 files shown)


TurkeyMan:
Can anyone explain these errors?
```
libdruntime.so: undefined reference to `std::exception::exception(char const*, int)'
libdruntime.so: undefined reference to `std::exception::~exception()'
libdruntime.so: undefined reference to `std::exception::what() const'
Etc...
```
I don't understand why it's trying to extern to those functions; there's no prototypes in this patch... those functions are fully defined :/

src/core/stdcpp/exception.d
```diff
@@ -6,95 +6,111 @@
...
-        //virtual ~this();
-        void dtor1();
-        void dtor2();
-        const(char)* what() const;
-    }
+    @nogc:
+        ///
+        this() nothrow {}
+        ///
+        ~this() nothrow {} // HACK: this should extern, but then we have link errors!
```
Geod24: How is that supposed to work?

src/core/stdcpp/exception.d
```diff
@@ -6,95 +6,111 @@
...
 
-    class bad_exception : exception
-    {
-        this();
-        //virtual ~this();
-        override const(char)* what() const;
+        ///
+        const(char)* what() const nothrow {} // HACK: this should extern, but then we have link errors!
```
Geod24: That won't compile

JinShil:
> Can anyone explain these errors?

I believe it's due to the `nothrow` attribution.

src/core/stdcpp/exception.d
```diff
@@ -6,95 +6,111 @@
...
+static if (__cplusplus < CppStdRevision.cpp17)
+{
+    ///
+    alias unexpected_handler = void function();
+    ///
+    deprecated unexpected_handler set_unexpected(unexpected_handler f) nothrow;
+    ///
+    deprecated unexpected_handler get_unexpected() nothrow;
+    ///
+    deprecated void unexpected();
```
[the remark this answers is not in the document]
TurkeyMan: they are.

src/core/stdcpp/exception.d
TurkeyMan: it reserves the appropriate position in the vtable.

TurkeyMan: D can't throw C++ exceptions, and also doesn't seem to be able to catch C++ exceptions either.

src/core/stdcpp/exception.d
TurkeyMan: haha, sotrue... why wasn't that the error I got!?

TurkeyMan:
C++ doesn't mangle nothrow... maybe it's the thing @Geod24 said, and I just didn't see it in the log...

TurkeyMan:
Okay, didn't help... I can't work out what I've changed compared to how it used to be. Why didn't the old code equally have unresolved extern's?

rainers:
Try adding core.stdcpp.exception to the SRC files :-)

TurkeyMan:
Really? How was this working before?

rainers:
> Really? How was this working before?

By not compiling or linking the file.

TurkeyMan:
It must be because I put `bad_alloc` in `new_.d` that caused the file to be pulled into a test somewhere...

src/core/stdcpp/exception.d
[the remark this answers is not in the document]
thewilsonator: http://www.cplusplus.com/reference/exception/set_unexpected/ under Compatibility: 
>The use of dynamic-exception-specifiers is deprecated (since C++11).

Geod24:
> C++ doesn't mangle nothrow...

POSIX: It does on function type.

> Try adding core.stdcpp.exception to the SRC files :-)

Wait wait wait, the reason it was not in the `SRCS` is because then, you tie up the symbols with the `CppRuntime` used to compile druntime as a static library. That's one of the reason having it in druntime was a bad idea, and the roadblock I hit when I wanted to throw C++ exceptions from D.

TurkeyMan:
Say that again? I don't know what you just said :)

Geod24:
For reference, this is what I'm refering to: https://github.com/dlang/dmd/pull/8743#issue-218409619

Let's continue this discussion on slack (#cpp-interop), for the sake of efficiency.

PR #2456, opened 2019-01-09
D `new` wrapper

TurkeyMan:


TurkeyMan:
@WalterBright Perhaps you can help me understand what's going on here?
Tests error with unresolved extern to the C++ ::new/delete functions because it doesn't link the C++ runtime.
...but why? There's no calls to these functions anywhere, so they should be stripped, rather than trying to link their extern dependencies.

Now, that issue aside, these functions have no reason to exist in physical form in the first place!
I tried to wrap them in `pragma(inline, true)` to cause the compiler to not generate code for them... but it seems that doesn't work. Why not? Isn't that exactly what `pragma(inline)` is meant to do?

I tried an alternative strategy making a dummy template argument, so that I could guarantee no codegen without instantiation; except that lead to this issue: https://issues.dlang.org/show_bug.cgi?id=19569

I'm out of ideas... I've run into this issue a lot with the C++ stuff... I need to write a function that should not gen code unless it's called, and it doesn't always need to be a template. I thought that's what `pragma(inline)` was meant to address, no?

jacob-carlborg:
On Posix all symbols are visible outside of the object files. There's nothing stopping code in other object files from using that symbol. Therefore code most be generated.

TurkeyMan:
An inline shouldn't be code until it's called... it should be like a template, but without the `T`.

rainers:
Whether `pragma(inline,true)` actually causes an error if the function is not inlined is currently "implementation defined". If it is made an error the function never has to be generated, but I don't think this is currently done (it should also be an error to take the address the function then).

The build errors here are from building the shared library that exports all symbols, so none can be removed by the linker.

TurkeyMan:
This is blocked on:
https://issues.dlang.org/show_bug.cgi?id=19570
https://github.com/dlang/dmd/pull/9235

look-at-me:
Seems like a mistake to allow overloads that only differ in that they are `nothrow`. In C++ you can't do this and it seems D makes no effort to try to resolve which function should be called based on the current scope. Even if it did, it would be an ambiguous call in functions that aren't marked nothrow. Using a different name for the nothrow overload instead will solve the problem, or pass a nothrow object like the C++ version.

PR #2461, opened 2019-01-14
Make core.internal.convert.toUbyte in CTFE for arrays work with reference type elements and not call postblits/dtors

n8sh:


n8sh:
Don't merge yet. Adding fix for enums.

n8sh:
Changes done. Please re-review.

test/hash/src/test_hash.d
```diff
@@ -195,6 +196,30 @@ void issue19568()
...
+        {
+            import core.stdc.stdio : puts;
+            if (x) puts("impure");
+        }
+    }
+    enum b2 = () {
+            S[10] a;
+            return ((const S[] a) @nogc nothrow pure @safe => toUbyte(a))(a);
+        }();
+}
```
ZombineDev: As far the compiler is concerned, `b1` and `b2` are not used, so there's not need to do any semantic analysis on the values they refer to. You should assert the values of `b1` and `b2` are what you expect them to be.

n8sh: I think the test is right since the test doesn't compile without this PR: https://run.dlang.io/is/v2dXCy

PR #2463, opened 2019-01-14
Add support for the in operator in dassert

wilzbach:
Requires https://github.com/dlang/dmd/pull/9264


src/core/internal/dassert.d
```diff
@@ -140,6 +140,10 @@ string invertCompToken(string comp)
             return "!is";
         case "!is":
             return "is";
+        case "in":
+            return "!in";
+        case "!in":
+            return "in";
```
wilzbach: Technically this isn't an operator (and the DMD PR doesn't include this change), but if we ever end up supporting this, it would very likely use `!in`.

12345swordy:
PR needs rebased

PR #2466, opened 2019-01-16
Replace module-level data with Meyers singletons

RazvanN7:
These destructors are superfluous. In addition, they block [1] because some structs that are used to instantiate array and hashtab are destroyed manually (_loadedDSO[3]/_handleToDSO[4]). For more information see: [1], [2].

[1] https://github.com/dlang/dmd/pull/8688
[2] https://forum.dlang.org/thread/lykddibnholrmtkotzxq@forum.dlang.org
[3] https://github.com/dlang/druntime/blob/master/src/rt/sections_elf_shared.d#L328
[4] https://github.com/dlang/druntime/blob/master/src/rt/sections_elf_shared.d#L340

### DIFF
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -183,7 +183,7 @@ version (Shared)
      */
     Array!(ThreadDSO)* initTLSRanges() @nogc nothrow
     {
-        return &_loadedDSOs;
+        return &_loadedDSOs();
     }
 
     void finiTLSRanges(Array!(ThreadDSO)* tdsos) @nogc nothrow
@@ -274,7 +274,7 @@ else
      */
     Array!(void[])* initTLSRanges() nothrow @nogc
     {
-        return &_tlsRanges;
+        return &_tlsRanges();
     }
 
     void finiTLSRanges(Array!(void[])* rngs) nothrow @nogc
@@ -325,7 +325,8 @@ version (Shared)
             _tlsRange = _pdso.tlsRange();
         }
     }
-    Array!(ThreadDSO) _loadedDSOs;
+    @property ref Array!(ThreadDSO) _loadedDSOs() @nogc nothrow { static Array!(ThreadDSO) x; return x; }
+    //Array!(ThreadDSO) _loadedDSOs;
 
     /*
      * Set to true during rt_loadLibrary/rt_unloadLibrary calls.
@@ -337,7 +338,8 @@ version (Shared)
      * The hash table is protected by a Mutex.
      */
     __gshared pthread_mutex_t _handleToDSOMutex;
-    __gshared HashTab!(void*, DSO*) _handleToDSO;
+    @property ref HashTab!(void*, DSO*) _handleToDSO() @nogc nothrow { __gshared HashTab!(void*, DSO*) x; return x; }
+    //__gshared HashTab!(void*, DSO*) _handleToDSO;
 
     /*
      * Section in executable that contains copy relocations.
@@ -351,13 +353,15 @@ else
      * Static DSOs loaded by the runtime linker. This includes the
      * executable. These can't be unloaded.
      */
-    __gshared Array!(DSO*) _loadedDSOs;
+    @property ref Array!(DSO*) _loadedDSOs() @nogc nothrow { __gshared Array!(DSO*) x; return x; }
+    //__gshared Array!(DSO*) _loadedDSOs;
 
     /*
      * Thread local array that contains TLS memory ranges for each
      * library initialized in this thread.
      */
-    Array!(void[]) _tlsRanges;
+    @property ref Array!(void[]) _tlsRanges() @nogc nothrow { static Array!(void[]) x; return x; }
+    //Array!(void[]) _tlsRanges;
 
     enum _rtLoading = false;
 }


Geod24:
> These destructors are superfluous

They call `free`. How are they superfluous ?

RazvanN7:
The one in array isn't and those variables wrapped in hashtab are manually destroyed by calling reset.

Geod24:
> The one in array doesn't call free

It most certainly do.
It sets `length` to 0, which is a call to `@property void length(size_t nlength)` with `nlength = 0`, which will then call `_ptr = cast(T*)common.xrealloc(_ptr, reqsize);` (Line 44) with `reqsize = 0`, and guess what `xrealloc` do [when you call it with a size = 0 argument](https://github.com/dlang/druntime/blob/94e58d38b8d86076c3c448ab1520cd355f452a4a/src/rt/util/container/common.d#L19)

> those variables wrapped in hashtab are manually destroyed by calling reset.

[I don't see a call to `reset` here](https://github.com/dlang/druntime/blob/954c7216fbc918ce68126027c778a4520327871b/src/rt/minfo.d#L241-L272)

RazvanN7:
This is embarrassing, but at first I thought that the `length = 0` line was simply setting a length var to 0. After a debugging session with @edi33416 (thanks!) we came with this solution, where the destructor is present only if you want it there. This, I think is the best compromise at the moment. Thoughts?

RazvanN7:
Even better, now the globals are wrapped in a @property method that returns a static instance of the variable. This will have to be changed when destructors are called for static variables.

dnadlinger:
Toggled off auto-merge because there are still leftover comments in there.

andralex:
>@andralex: I'm not sure I follow your rationale. This is a hack relying on the fact that apparently static TLS globals aren't destructed by Razvan's proposed DMD implementation, while module-level ones are. You mention that "[the] change is good because it replaces uncontrolled data", but what the patch actually does is to ensure that data _stays_ uncontrolled even after [dlang/dmd#9151](https://github.com/dlang/dmd/pull/9151).

>It might be okay to put in a hack to allow work on DMD to progress, but at least it should be properly documented as such.

Yes, if this were a hack it should be documented as such. Here, switching from globals to Meyers singletons is a step forward regardless and should be recommended for any project. Assuming the right infrastructure is in place, plain globals have destruction generated upon thread/process exit and there is no way to control the order in which globals in distinct modules are destroyed. Meyers singletons, in contrast, queue destruction upon the first use so they are destroyed in reverse order of first use. This makes dynamic dependencies more tractable and also allows code to easily detect whether a destroyed object is being used illegally (per Phoenix Singleton).

Please approve, thanks!

andralex:
@WalterBright btw what is our current mechanism for destroying thread-local statics in functions? Thx!

andralex:
I'm disappointed that `@property` does not make the function entirely transparent - `&_loadedDSOs` is still interpreted as taking the address of the function, not the result of the function. Wasn't this literally the only job `@property` had?

dnadlinger:
> there is no way to control the order in which globals in distinct modules are destroyed

The same rules as for static destructors should probably be used, because that's what module-level  destructors are. Of course, this makes cyclic dependencies more brittle, but pointing out the danger is a feature, not a bug.

andralex:
@dnadlinger thanks I'll put this on 72h approval just in case more ideas come forward.

dnadlinger:
> @WalterBright btw what is our current mechanism for destroying thread-local statics in functions?

There currently is none. Hence my above comment – your argument applies to a sensible design (although I'm not sure how that would interact with shared library unloading, off the top of my head), my argument about it being a hack applies to the status quo.

dnadlinger:
@andralex: Also, it's not obvious to me how encouraging a dynamically defined initialisation order (init on first use) in D avoids the [static (de)initialization order fiasco](https://isocpp.org/wiki/faq/ctors#construct-on-first-use-v2).

dnadlinger:
(Module cycle detection being our brittle and pessimistic, if correct, response.)

jacob-carlborg:
> I'm disappointed that `@property` does not make the function entirely transparent - `&_loadedDSOs` is still interpreted as taking the address of the function, not the result of the function. Wasn't this literally the only job `@property` had?

Fixing that doesn't make it completely transparent. There's still the issue with a property needs to emulate variable chaining, i.e. `a = b = c`.

andralex:
@dnadlinger ok I've added https://issues.dlang.org/show_bug.cgi?id=19671 as a major issue. I'm honestly quite bummed about it.

Dynamically intialization is usually better for the simple intuition that "the most necessary objects come to life first because they are used during initialization of other objects". For example an allocator would be needed to allocate data and would need to be still alive as that data is deallocated.

It's not infallible because e.g maybe you want to log a message during destruction but not during construction. For those cases a detection can be written easily:

```D
ref T meyersSingleton() 
{
    static bool destroyed; 
    static struct Sentry { ~this() { destroyed = true; }
    static T payload;
    static Sentry guard;
    if (destroyed) 
    {
        printf("Access after destruction. You are a bad, bad person.\n");
        abort();
    }
    return payload;
}
```

The working of this (again assuming the right infrastructure is in place) is somewhat subtle but the problem is arguably also subtle. The function above can be templated and reused.

dnadlinger:
> For those cases [where objects are accessed out of order] a detection can be written easily 

We might need to emit this for every static field by default, then, to guarantee `@safe`-ty.

andralex:
@dnadlinger we'd then need to define under what circumstances access after destruction would cause memory safety issues. E.g. if the static field has no destructor - no need. It also seems to be the case that blitting .init over the data after destruction would be a good thing to do for safety.

thewilsonator:
72h is up, dunno what the dlang bot is doing.

PR #2467, opened 2019-01-18
Fix Issue 19593 - use memcmp instead of dstrcmp

kubo39:
Using dstrcmp inside trace_addsym causes stack overflow, because dmd
 inserts trace_pro function to top of dstrcmp when -profile option is added.


kubo39:
> I assume you are not recompiling the runtime with -profile, but that the issue is that rt.trace and your code reference the same symbol dstrcmp, but also generate different implementations. The linker picks the one with profiling hooks causing the stack overflow in the profiler code.

Ah, yes, I found this at `FuncDeclaration__toObjFile` in dmd/src/dmd/glue.d (sorry for my poor english..)

PR #2472, opened 2019-01-22
Basic NetBSD support needed to build GDC

coypoop:
Uncomment dlinfo, provide assert and mbstate_t

The one patch I have left for is that I had trouble using src/core/sys/netbsd/execinfo.d with GDC, it didn't seem to provide `D_InlineAsm_X86_64` or `D_InlineAsm_X86` (I am building on x86-64)

I worked around it by providing a declaring matching [libexecinfo](https://man-k.org/man/NetBSD-current/3/backtrace?r=1&q=backtrace) but that required supplying -lexecinfo.

One remaining issue I have to resolve is that time_t (src/core/sys/posix/sys/types.d:177) is considered c_long, it is actually 64bit on all archs, but only if you use the newer functions.

NetBSD has a funky way of updating libc functions, where it relies on the inclusion of headers to use the new functions (as opposed to relying on the linker with symbol versioning), which is a little problematic for languages that can't use those headers. The old references for 32bit work though.

### DIFF
--- a/src/core/stdc/assert_.d
+++ b/src/core/stdc/assert_.d
@@ -53,6 +53,13 @@ else version (FreeBSD)
      */
     void __assert(const(char)* exp, const(char)* file, uint line);
 }
+else version (NetBSD)
+{
+    /***
+     * Assert failure function in the NetBSD C library.
+     */
+    void __assert(const(char)* file, int line, const(char)* exp);
+}
 else version (DragonFlyBSD)
 {
     /***
--- a/src/core/stdc/wchar_.d
+++ b/src/core/stdc/wchar_.d
@@ -40,6 +40,15 @@ version (CRuntime_Glibc)
         ___value __value;
     }
 }
+else version (NetBSD)
+{
+    union __mbstate_t
+    {
+        int64_t   __mbstateL;
+        char[128] __mbstate8;
+    }
+    alias mbstate_t = __mbstate_t;
+}
 else version (OpenBSD)
 {
     union __mbstate_t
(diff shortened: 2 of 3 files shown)


thewilsonator:
cc @ibuclaw

ibuclaw:
> D_InlineAsm_X86_64 or D_InlineAsm_X86 (I am building on x86-64)

These are for dmd syntax inline assembler.

rmboggs:
sorry to comment on a merged pull request but is there any reason why __mbstate_t is defined in both [wchar_d ](https://github.com/dlang/druntime/blob/badbc184b2493da7a41c26d768976e73d2364cfa/src/core/stdc/wchar_.d#L43)and [stdio.d](https://github.com/dlang/druntime/blob/badbc184b2493da7a41c26d768976e73d2364cfa/src/core/stdc/stdio.d#L232) files?  Or am I misreading something?  This can also be said for OpenBSD.

ZombineDev:
@rmboggs you're right, `*mbstate_t` should be defined only in `core.stdc.wchar_` and imported in `core.stdc.stdio`. My guess is that someone needed the definition for *bsd's stdio, but since it wasn't in wchar_, they defined it inline. Feel free to open a PR that (re)moves `*mbstate_t` from stdio.

rmboggs:
Thank you for confirming.  Currently trying to work on OpenBSD support so it is good to know where it belongs.  I'll see about creating a pull request to organize this for Net/OpenBSD when I can.

rmboggs:
https://github.com/dlang/druntime/pull/2473

PR #2473, opened 2019-01-22
Removing __mbstate_t union definition from core/stdc/stdio.d

rmboggs:
Both NetBSD and OpenBSD have __mbstate_t union definitions in both stdio.d and wchar_.d files in src/core/std.  Since __mbstate_t should be defined in wchar_.d and __mbstate_t isn't used by Net/OpenBSD in stdio.d anyway, just remove the defs from stdio.d and it should be ok.

Please let me know if anything else needs to be modified.

### DIFF
--- a/src/core/stdc/stdio.d
+++ b/src/core/stdc/stdio.d
@@ -228,12 +228,6 @@ else version (NetBSD)
         ubyte *_base;
         int _size;
     }
-
-    union __mbstate_t // <sys/_types.h>
-    {
-        char[128]   _mbstate8 = 0;
-        long        _mbstateL;
-    }
 }
 else version (OpenBSD)
 {
@@ -257,12 +251,6 @@ else version (OpenBSD)
     {
         ubyte *_base;
         int _size;
-    }
-
-    union __mbstate_t // <sys/_types.h>
-    {
-        char[128]   __mbstate8 = 0;
-        long        __mbstateL;
     }
 }
 else version (DragonFlyBSD)


ZombineDev:
Thanks! Can you move the FreeBSD definition to wchar_.d as well?

ibuclaw:
LGTM, just noted comments to address.

rmboggs:
I was wondering about the other BSDs, actually.  The only reason why I left them as is was because I didn't know what else relied on mbstate being in stdio.d, for FreeBSD at least since it is a supported platform.

ibuclaw:
Perhaps it should go in its own, but given a choice between the two, wchar_.d looks to be the better place to have the definition.

In stdio.d, just be sure to have `import core.stdc.wchar_ : mbstate_t` for the BSDs that use it.

ZombineDev:
As far as I could see, in FreeBSD's libc implementation `__mbstate_t` is definited in `<sys/_typses.h>`  and `typedef`-ed to `mbstate_t` in `<wchar.h>` More importantly, the C standard says that `mbstate_t` is definited by `<wchar.h>`, so the definition should be there. As far as `core.sys.stdc.stdio` is concerned, it should only privately import `core.stdc.wchar_ : mbstate_t` and no users of stdio should assume that `mbstate_t` is exposed from there.

rmboggs:
Ok, makes sense. I'll push an update as soon as I can. Thanks.

rmboggs:
Ok, well. All the other BSDS are updated accordingly now. Please take a good look at the FreeBSD and DragonFlyBSD updates as I can't test those at home currently.

rmboggs:
Thanks, glad to help.

Do you still need me to refactor mbstate_t for solaris? Looking at the code, I thought that one was fine as-is.

ibuclaw:
#2474

PR #2474, opened 2019-01-23
[Solaris] core/stdc/wchar_.d: Move alias for mbstate_t here

ibuclaw:
It's the one remaining platform that still has an alias in core.stdc.stdio.

### DIFF
--- a/src/core/stdc/stdio.d
+++ b/src/core/stdc/stdio.d
@@ -643,10 +643,7 @@ else version (DragonFlyBSD)
 }
 else version (Solaris)
 {
-    import core.stdc.wchar_ : __mbstate_t;
-
-    ///
-    alias mbstate_t = __mbstate_t;
+    import core.stdc.wchar_ : mbstate_t;
 
     ///
     alias c_long fpos_t;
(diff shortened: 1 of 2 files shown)


src/core/stdc/stdio.d
```diff
@@ -643,10 +643,7 @@ else version (DragonFlyBSD)
 }
 else version (Solaris)
 {
-    import core.stdc.wchar_ : __mbstate_t;
-
-    ///
-    alias mbstate_t = __mbstate_t;
```
jacob-carlborg: Do we have a policy on the C bindings and breaking headers?

ibuclaw: It was already broken in #2473, this makes it consistent with the rest.

ibuclaw: It would be nice to have more strict documentation in each binding module that tells the porter what do define here.

Currently, people just put whatever they want wherever they want / require it.

jacob-carlborg: Yeah, that would be nice.

ibuclaw: This is a good example:
https://github.com/dlang/druntime/blob/3d4d4376cebaeb9d624834d334d3c287e010dccf/src/core/sys/posix/fcntl.d#L53-L102

PR #2475, opened 2019-01-24
Remove leading whitespaces in object.d

edi33416:
`object.d` had some weird indentation (leading whitespace) in some places.
While cleaning this up, I also noticed that not all the `unittest`s had `@safe`/`@system` so I added those as well.


wilzbach:
image


# CAN WE PLEASE STOP MERGING "qf" comments!!??

I added `auto-merge-squash` back again and while it comes with the downside that it will just merge everything into one commit, it's often exactly what you want.

PR #2476, opened 2019-01-24
Add opEquals to gc.gcinterface.Range

RazvanN7:
This is currently blocking [1]. A few explanations: before this patch, whenever 2 Range objects were compared for equality (`==`/`!=`), the comparison would have been made directly between the aliased this members (in this case `lhs.pbot == rhs.pbot`). That comparison is a bitwise one so it's nothrow. After the 
dmd patch, the comparison is going to be made using the generated opEquals (which bitwise compares the basic types and uses the opEquals of subfields); since Range has a TypeInfo member, its opEquals is going to get call which may throw. That will make this line [2] to fail compilation, because all the opApply functions expect a `nothrow` delegate, while the foreach body contains a call to a throwing opEquals [3]

[1] https://github.com/dlang/dmd/pull/9289
[2] https://github.com/dlang/druntime/blob/master/src/rt/util/container/treap.d#L59
[3] https://github.com/dlang/druntime/blob/master/src/rt/util/container/treap.d#L63

### DIFF
--- a/src/gc/gcinterface.d
+++ b/src/gc/gcinterface.d
@@ -33,6 +33,7 @@ struct Range
     void* ptop;
     TypeInfo ti; // should be tail const, but doesn't exist for references
     alias pbot this; // only consider pbot for relative ordering (opCmp)
+    bool opEquals(in Range rhs) nothrow const { return pbot == rhs.pbot; }
 }
 
 interface GC


src/gc/gcinterface.d
```diff
@@ -33,6 +33,7 @@ struct Range
     void* ptop;
     TypeInfo ti; // should be tail const, but doesn't exist for references
     alias pbot this; // only consider pbot for relative ordering (opCmp)
+    bool opEquals(in Range rhs) nothrow const { return pbot == rhs.pbot; }
```
rainers: treap uses both `==` and `<`. If the order of `alias this` and operator overload lookup for `opEquals` changes, shouldn't this also happen for `opCmp`?

RazvanN7: Not really, because opCmp is treated differently:

struct A
{
    int x;
    bool opCmp(ref const A q)
    {
        return this.x < q.x;                                                                                                                  
    }
}  
   
struct B
{
    A a, b;
}

struct C
{
    A a, b;
    alias a this;
}

static assert(C(A(1), A(1)) < C(A(1), A(2))); // error: need opCmp for struct C;
static assert(C(A(1), A(1)) != C(A(1), A(2))); // fails because A.opEquals is called

The dmd patch makes it so that the last assert passes (C.opEquals has a higher priority).

rainers: >  static assert(C(A(1), A(1)) < C(A(1), A(2))); // error: need opCmp for struct C;

That just happens because your opCmp overload must not `ref` the rvalue parameter. With `    bool opCmp(A q)` I get: "Error: static assert:  C(A(1), A(1)) < C(A(1), A(2)) is false". see https://run.dlang.io/is/ZdfrfR

andralex:
I'll pull this regardless because it's good style.

PR #2479, opened 2019-01-31
Fix Issue 19635 - -checkaction=context not working with attributes

wilzbach:
The program will be terminated after the assertion error message has been printed and its not considered part of the "main" program. Also, catching an AssertError is Undefined Behavior.

Hence, we can fake purity and @nogc-ness here.

An alternative would be to directly ignore the `assert` message expression directly in DMD's type checker (see e.g. https://issues.dlang.org/show_bug.cgi?id=15100 for this).

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -37,7 +37,7 @@ private template getPrintfFormat(T)
 Minimalistic formatting for use in _d_assert_fail to keep the compilation
 overhead small and avoid the use of Phobos.
 */
-auto miniFormat(V)(auto ref V v)
+auto miniFormat(V)(V v)
 {
     import core.stdc.stdio : sprintf;
     import core.stdc.string : strlen;
@@ -45,14 +45,14 @@ auto miniFormat(V)(auto ref V v)
     {
         enum printfFormat = getPrintfFormat!V;
         char[20] val;
-        sprintf(val.ptr, printfFormat, v);
-        return val.idup[0 .. strlen(val.ptr)];
+        const len = sprintf(&val[0], printfFormat, v);
+        return val.idup[0 .. len];
     }
     else static if (__traits(isFloating, V))
     {
         char[60] val;
-        sprintf(val.ptr, "%g", v);
-        return val.idup[0 .. strlen(val.ptr)];
+        const len = sprintf(&val[0], "%g", v);
+        return val.idup[0 .. len];
     }
     else static if (__traits(compiles, { string s = V.init.toString(); }))
     {
@@ -145,3 +145,26 @@ string invertCompToken(string comp)
     }
 }
 
+private auto assumeFakeAttributes(T)(T t) @trusted
+{
+    import core.internal.traits : Parameters, ReturnType;
+    alias RT = ReturnType!T;
+    alias P = Parameters!T;
+    alias type = RT function(P) nothrow @nogc @safe pure;
+    return cast(type) t;
+}
+
+auto miniFormatFakeAttributes(T)(T t)
+{
+    alias miniT = miniFormat!T;
+    return assumeFakeAttributes(&miniT)(t);
+}
+
+auto pureAlloc(size_t t)
+{
+    static auto alloc(size_t len)
+    {
+        return new ubyte[len];
+    }
+    return assumeFakeAttributes(&alloc)(t);
+}
--- a/src/object.d
+++ b/src/object.d
@@ -4819,11 +4819,29 @@ TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @nogc pure @trusted
 }
 
 // Allows customized assert error messages
-string _d_assert_fail(string comp, A, B)(A a, B b)
+string _d_assert_fail(string comp, A, B)(A a, B b) @nogc @safe nothrow pure
 {
-    import core.internal.dassert : invertCompToken, miniFormat;
-    auto valA = miniFormat(a);
-    auto valB = miniFormat(b);
+    import core.internal.dassert : invertCompToken, miniFormatFakeAttributes, pureAlloc;
+    /*
+    The program will be terminated after the assertion error message has
+    been printed and its not considered part of the "main" program.
+    Also, catching an AssertError is Undefined Behavior
+    Hence, we can fake purity and @nogc-ness here.
+    */
+
+    auto valA = miniFormatFakeAttributes(a);
+    auto valB = miniFormatFakeAttributes(b);
     enum token = invertCompToken(comp);
-    return valA ~ " " ~ token ~ " " ~ valB;
+
+    const totalLen = valA.length + token.length + valB.length + 2;
+    char[] buffer = cast(char[]) pureAlloc(totalLen)[0 .. totalLen];
+    // @nogc-concat of "<valA> <comp> <valB>"
+    auto n = valA.length;
+    buffer[0 .. n] = valA;
+    buffer[n++] = ' ';
+    buffer[n .. n + token.length] = token;
+    n += token.length;
+    buffer[n++] = ' ';
+    buffer[n .. n + valB.length] = valB;
+    return buffer;
 }
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -334,3 +334,217 @@ void assertCTFEable(alias dg)()
     static assert({ cast(void) dg(); return true; }());
     cast(void) dg();
 }
+
+// std.traits.FunctionTypeOf
+/*
+Get the function type from a callable object `func`.
+
+Using builtin `typeof` on a property function yields the types of the
+property value, not of the property function itself.  Still,
+`FunctionTypeOf` is able to obtain function types of properties.
+
+Note:
+Do not confuse function types with function pointer types; function types are
+usually used for compile-time reflection purposes.
+ */
+template FunctionTypeOf(func...)
+if (func.length == 1 /*&& isCallable!func*/)
+{
+    static if (is(typeof(& func[0]) Fsym : Fsym*) && is(Fsym == function) || is(typeof(& func[0]) Fsym == delegate))
+    {
+        alias FunctionTypeOf = Fsym; // HIT: (nested) function symbol
+    }
+    else static if (is(typeof(& func[0].opCall) Fobj == delegate))
+    {
+        alias FunctionTypeOf = Fobj; // HIT: callable object
+    }
+    else static if (is(typeof(& func[0].opCall) Ftyp : Ftyp*) && is(Ftyp == function))
+    {
+        alias FunctionTypeOf = Ftyp; // HIT: callable type
+    }
+    else static if (is(func[0] T) || is(typeof(func[0]) T))
+    {
+        static if (is(T == function))
+            alias FunctionTypeOf = T;    // HIT: function
+        else static if (is(T Fptr : Fptr*) && is(Fptr == function))
+            alias FunctionTypeOf = Fptr; // HIT: function pointer
+        else static if (is(T Fdlg == delegate))
+            alias FunctionTypeOf = Fdlg; // HIT: delegate
+        else
+            static assert(0);
+    }
+    else
+        static assert(0);
+}
+
+@safe unittest
+{
+    class C
+    {
+        int value() @property { return 0; }
+    }
+    static assert(is( typeof(C.value) == int ));
+    static assert(is( FunctionTypeOf!(C.value) == function ));
+}
+
+@system unittest
+{
+    int test(int a);
+    int propGet() @property;
+    int propSet(int a) @property;
+    int function(int) test_fp;
+    int delegate(int) test_dg;
+    static assert(is( typeof(test) == FunctionTypeOf!(typeof(test)) ));
+    static assert(is( typeof(test) == FunctionTypeOf!test ));
+    static assert(is( typeof(test) == FunctionTypeOf!test_fp ));
+    static assert(is( typeof(test) == FunctionTypeOf!test_dg ));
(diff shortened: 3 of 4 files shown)


wilzbach:
Hmm but making this work nicely with attributes leads to a lot of problems:
- can we use assume that `sprintf` is `pure` (or maybe store at floating point register and reset them afterwards)?
- allocate with sth. smarter than `pureMalloc`, s.t. it's still `@nogc` (though technically never freeing memory could be considered a "valid state" :thinking:)
- custom, user-defined `toString` methods will always be problematic (?)

> I'm not sure about catching AssertError

https://dlang.org/library/object/exception.html

> In principle, only thrown objects derived from this class [Exception] are safe to catch inside a catch block. Thrown objects not derived from Exception represent runtime errors that should not be caught, as certain runtime guarantees may not hold, making it unsafe to continue program execution.

jacob-carlborg:
> https://dlang.org/library/object/exception.html
>> In principle, only thrown objects derived from this class [Exception] are safe to catch inside a catch >> block. Thrown objects not derived from Exception represent runtime errors that should not be >> >> caught, as certain runtime guarantees may not hold, making it unsafe to continue program execution.

I know, at the same time the spec says this [1]:

"Implementation Defined:
5. Whether the program stops on the first unit test failure, or continues running the unit tests."

Currently this is only possible to implemented by catching `AssertError`. Also druntime is catching all `Throwable`, so there are currently some contradictions:

https://github.com/dlang/druntime/blob/ca1be3fc715a45ae8b34baef5e272dfe2cda6cad/src/rt/dmain2.d#L480-L484

[1] https://dlang.org/spec/unittest.html

jacob-carlborg:
Does `invertCompToken` need to have attributes added?

jacob-carlborg:
> can we use assume that sprintf is pure (or maybe store at floating point register and reset them afterwards)?

I'm no expert at this but it sounds like it could be pure.

> allocate with sth. smarter than pureMalloc, s.t. it's still @nogc (though technically never freeing memory could be considered a "valid state" 🤔)

Alternatives would be a stack allocated buffer, either using `alloca` or a static buffer. If the result doesn't fit in the buffer it could fallback to the behavior when `-checkaction=context` is not used.

> custom, user-defined toString methods will always be problematic (?)

Yeah, that sounds a bit problematic.

wilzbach:
> Alternatives would be a stack allocated buffer, either using alloca or a static buffer.

Stack allocated buffer won't work here because  `_d_assert_fail` needs to return a string.
A static bufffer is problematic because if catching `AssertError` is valid then we would overwrite that user-code message.

So imho this trick is the best we got for now until rcstring becomes a thing (and even then we might not be able to use it because a `string` is expected for `Error.message` here :/

wilzbach:
> Does invertCompToken need to have attributes added?

No, it's called at compile-time (`enum token = invertCompToken(comp);`) (and there even was a bit of discussion on the NG about not inverting the token, so we might end up removing it entirely).

anon17:
```
void f() @nogc
{
    assert(false,"");
}
```
This compiles and throws a GC allocated exception. I believe the general sentiment is that assert failure can bypass some language constraints, notably `nothrow`.

src/core/internal/dassert.d
```diff
@@ -37,21 +37,21 @@ private template getPrintfFormat(T)
...
 {
     import core.stdc.stdio : sprintf;
     import core.stdc.string : strlen;
     static if (__traits(isIntegral, V))
     {
         enum printfFormat = getPrintfFormat!V;
         char[20] val;
-        sprintf(val.ptr, printfFormat, v);
+        sprintf(&val[0], printfFormat, v);
         return val.idup[0 .. strlen(val.ptr)];
```
edi33416: `val.ptr` -> `&val[0]`

src/core/internal/dassert.d
```diff
@@ -37,21 +37,21 @@ private template getPrintfFormat(T)
...
     }
     else static if (__traits(isFloating, V))
     {
         char[60] val;
-        sprintf(val.ptr, "%g", v);
+        sprintf(&val[0], "%g", v);
         return val.idup[0 .. strlen(val.ptr)];
```
edi33416: ditto

edi33416:
> > Stack allocated buffer won't work here because _d_assert_fail needs to return a string.
> 
> Not sure if it helps, but if `alloca` is used as the default value of parameter it's allocated in the caller's stack, i.e.:
> 
> ```d
> void* foo(void* buffer = alloca(1))
> {
>     return buffer;
> }
> ```

This might be helpful, but the caller needs to be aware of this, or he'll get bitten
```
void *bar()
{
    return foo(); // oops.. returns ptr to local alloca buffer
}
```

jacob-carlborg:
> The current implementation of Errors in core.exception use a static buffer. IIRC, this was accepted as ok because an error will be thrown when your program is in an unrecoverable state.

As I mentioned previously:

"Unlike AssertExpressions used elsewhere, the assert is not assumed to hold, and upon assert failure the program is still in a defined state." [1].

[1] https://dlang.org/spec/unittest.html

src/core/internal/dassert.d
wilzbach: This has to be `@trusted` code anyhow :/

jacob-carlborg:
> it wouldn't fix e.g. user-generated toString:

What about the sink version of `toString`. If we only support that version would that help?

wilzbach:
> What about the sink version of `toString`. If we only support that version would that help?

Yeah that would help, though who uses a sink version of `toString` these days?
And as soon as the author does sth. with strings it wouldn't e.g. be nothrow [1] anymore

```
import std.stdio, std.range;
struct MyStruct {
    enum mystring = "test";
    auto toString(W)(W w) { return w.put(mystring.front); }
}
void main() @nogc nothrow {
    struct Sink {
        auto put(D)(D d) {}   
    }
    MyStruct().toString(Sink());
}
```

https://run.dlang.io/is/xQeV33

[1] And, of course, as exceptions still require the GC also not `@nogc`.

Thus, I'm more leaning forward to this "fake it" approach as long as we can guarantee that it plays nicely with `unittest`. So maybe it's better to allocate the memory with the GC (and just mark it as `@nogc`) as this would at least prevent memory leaks?

jacob-carlborg:
> So maybe it's better to allocate the memory with the GC (and just mark it as `@nogc`) as this would at least prevent memory leaks?

What happens then when someone using betterC and doesn't link druntime?

wilzbach:
> What happens then when someone using betterC and doesn't link druntime?

Then they wouldn't be able to use D's exceptions + errors (e.g. AssertError) and even `assert` automatically gets rewritten to C's `assert`.

jacob-carlborg:
> `assert` automatically gets rewritten to C's `assert`

Yes, but what happens then with the assert message?

anon17:
I use C `__assert` function to print message from D assert :)

wilzbach:
> So maybe it's better to allocate the memory with the GC (and just mark it as @nogc) as this would at least prevent memory leaks?

Changed the PR to do this now.

anon17:
```d
private auto assumeFakeAttributes(T)(T t)
{
	import core.internal.traits : Parameters, ReturnType;
	alias RT = ReturnType!T;
	alias P = Parameters!T;
	mixin("alias type = RT function(P) nothrow @nogc pure;");
	return cast(type) t;
}

auto miniFormatFakeAttributes(T)(T t) @trusted
{
	alias miniT = miniFormat!T;
	return assumeFakeAttributes(&miniT)(t);
}
```
Apparently you don't need `@safe` for pureAlloc. I think, it can be just blanket marked as `@trusted` (what can possibly go wrong?), so that mixin is not needed either.

wilzbach:
> Apparently you don't need @safe for pureAlloc. 

I'm not sure I understand you correctly, it gets automatically inferred because of `auto` and also it's used an `@trusted`  function in `_dassert`. The `@safe` in the function is required too (even though there's an `@trusted`), because we return the function declaration which then gets passed on.

Try to remove it and then run:

```
make -f posix.mak unittest-debug
```

You can use [`pr_checkout 2479`](https://github.com/wilzbach/git-tools#pr_checkout---checkout-a-pr-locally) for checking out a PR locally directly in the repo (or of course `dub run digger -- build "master + druntime#2479"`).

> mixin is not needed either.

Fair point! Updated. Thanks!

anon17:
Oh, btw, casting `byte[]` to `char[]` is safe, so you can remove that trusted block.

nordlow:
Keep up the good work, @wilzbach and i'll disturb @andralex if this gets stalled ;)

wilzbach:
> Keep up the good work, @wilzbach and I'll disturb @andralex if this gets stalled ;)

Hehe as far as I can tell, this already got stalled. At least, I don't see anything else I can do to help this PR.

andralex:
Isn't this more of a @WalterBright thing?

nordlow:
You are right.

src/core/internal/dassert.d
```diff
@@ -49,14 +49,14 @@ auto miniFormat(V)(auto ref V v)
...
+        sprintf(&val[0], printfFormat, v);
+        return val.idup[0 .. strlen(&val[0])];
     }
     else static if (__traits(isFloating, V))
     {
         char[60] val;
-        sprintf(val.ptr, "%g", v);
-        return val.idup[0 .. strlen(val.ptr)];
+        sprintf(&val[0], "%g", v);
+        return val.idup[0 .. strlen(&val[0])];
```
thewilsonator: sprintf return the number of characters written, no need to call strlen.

wilzbach:
Rebased to `stable` as without this PR `-checkaction=assert` is barely usable. Also, added the 72h objection label as no one really has provided a better way and so far it looks like this is the only option we have for now.

src/core/internal/dassert.d
```diff
@@ -37,22 +37,22 @@ private template getPrintfFormat(T)
...
 {
     import core.stdc.stdio : sprintf;
     import core.stdc.string : strlen;
     static if (__traits(isIntegral, V))
     {
         enum printfFormat = getPrintfFormat!V;
         char[20] val;
-        sprintf(val.ptr, printfFormat, v);
-        return val.idup[0 .. strlen(val.ptr)];
+        auto len = sprintf(&val[0], printfFormat, v);
```
jacob-carlborg: `const`.

src/core/internal/dassert.d
```diff
@@ -37,22 +37,22 @@ private template getPrintfFormat(T)
...
+        return val.idup[0 .. len];
     }
     else static if (__traits(isFloating, V))
     {
         char[60] val;
-        sprintf(val.ptr, "%g", v);
-        return val.idup[0 .. strlen(val.ptr)];
+        auto len = sprintf(&val[0], "%g", v);
```
jacob-carlborg: `const`.

src/object.d
```diff
@@ -4819,11 +4819,29 @@ TTo[] __ArrayCast(TFrom, TTo)(TFrom[] from) @nogc pure @trusted
...
-    import core.internal.dassert : invertCompToken, miniFormat;
-    auto valA = miniFormat(a);
-    auto valB = miniFormat(b);
+    import core.internal.dassert : invertCompToken, miniFormatFakeAttributes, pureAlloc;
+    /*
+    The program will be terminated after the assertion error message has
+    been printed and its not considered part of the "main" program.
+    Also, catching an AssertError is Undefined Behavior
+    Hence, we can fake purity and @nogc-ness here.
+    */
```
jacob-carlborg: I disagree with this comment.

veelo:
> Rebased to `stable` as without this PR `-checkaction=assert` is barely usable.

At the risk of going OT: apropos barely usable, are you aware of https://issues.dlang.org/show_bug.cgi?id=19711?

wilzbach:
> > Rebased to `stable` as without this PR `-checkaction=assert` is barely usable.
> 
> At the risk of going OT: apropos barely usable, are you aware of https://issues.dlang.org/show_bug.cgi?id=19711?

Now I'm, but the issue is very unrelated.

> This is just a run.dlang.io issue as it just matches for `-c`.
See e.g. https://run.dlang.io/is/8eUzXP (by adding `-run` yourself you workaround the run.dlang.io issue)).

anon17:
Yeah, it's that thing, "don't catch Errors" is a rule for business logic, which one might know how to break, druntime code that catches Throwable has code tailored to print a message without much relying on the rest of application code.

jacob-carlborg:
> druntime code that catches Throwable has code tailored to print a message without much relying on the rest of application code.

The default unit test runner in druntime will catch all `Throwable` and continue executing remaining tests. One of the test runners in DMD will do the same.

anon17:
It won't execute the application after the tests, will it? If a failing test corrupted memory, then following tests can experience heisenbugs.

wilzbach:
I'm not sure where the discussion is going about. The runtime _already_ uses the GC to allocate other parts of the error handling and formatting. This just makes the new context-aware switch work similar.
And so far no one has come up with a better solution.

jacob-carlborg:
> I'm not sure where the discussion is going about. The runtime already uses the GC to allocate other parts of the error handling and formatting. This just makes the new context-aware switch work similar.
And so far no one has come up with a better solution.

I disagree with the following reasoning, which you have in a comment:

The program will be terminated after the assertion error message has been printed and its not considered part of the "main" program. Also, catching an AssertError is Undefined Behavior. Hence, we can fake purity and @nogc-ness here.

jacob-carlborg:
> It won't execute the application after the tests, will it?

No.

> If a failing test corrupted memory, then following tests can experience heisenbugs.

Yes, that might happen. But I assume you would rerun all tests once the failing test is fixed.

PR #2480, opened 2019-02-01
GCC bug 88127: qsort_r is not available in old glibc

jpf91:
In GCC, we'll simply use the configure script to check if `qsort_r` is available and set `Have_Qsort_R` in `gcc.config`. Upstream druntime doesn't have an advanced build system which supports such checks, so just continue to assume the glibc supports `qsort_r`.

Can't use a version here, cause defining it with static if will lead to version defined after use errors.

Original bug report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88127

ping @ibuclaw

### DIFF
--- a/src/rt/qsort.d
+++ b/src/rt/qsort.d
@@ -27,7 +27,25 @@ else version (TVOS)
 else version (WatchOS)
     version = Darwin;
 
+// qsort_r was added in glibc in 2.8. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88127
 version (CRuntime_Glibc)
+{
+    version (GNU)
+    {
+        import gcc.config : Have_Qsort_R;
+        enum Glibc_Qsort_R = Have_Qsort_R;
+    }
+    else
+    {
+        enum Glibc_Qsort_R = true;
+    }
+}
+else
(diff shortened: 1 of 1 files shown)


jpf91:
Thanks!

ibuclaw:
I suspect that other platforms may require similar treatment depending on how far back we want to support things.

jpf91:
Hmm, well I could wrap everything in a static if, if you prefer that?
```d
// qsort_r may not be available
version (GNU)
    import gcc.config : Have_Qsort_R;
else
    enum Have_Qsort_R = true;

static if (Have_Qsort_R)
{
    version (CRuntime_Glibc)
    {
        enum Use_Qsort_R = true;
        ...
    }
    else version (FreeBSD)
    {
        enum Use_Qsort_R = true;
        ...
    }
    else version (DragonFlyBSD)
    {
        enum Use_Qsort_R = true;
        ...
    }
    else
    {
        enum Use_Qsort_R = false;
    }
}

static if (!Use_Qsort_R)
{
    //qsort implementation
}
```

However, I don't know how long qsort_r has been available on the other platforms.

PR #2482, opened 2019-02-04
FIX Issue 19643 - add missing Objective-C function calling convention

kubo39:


thewilsonator (approved):
LGTM, cc @jacob-carlborg

src/core/demangle.d
```diff
@@ -2389,21 +2403,24 @@ private template hasPlainMangling(FT) if (is(FT == function))
     static extern(C) void fooC();
     static extern(Windows) void fooW();
     static extern(C++) void fooCPP();
+    static extern(Objective-C) void fooObjC();
```
jacob-carlborg: Not sure if this works, but `extern(Objective-C)` only makes sense on classes.

rainers:
The mangling for Objective-C has an ambiguity if used with the grammar from the website, see https://issues.dlang.org/show_bug.cgi?id=17310. IIRC the demangler should be able to handle these nevertheless, but you might want to add some symbols from that bug report to the tests, too.

src/core/demangle.d
kubo39: I think `extern (Objective-C)` with classes works only if `macOS && 64bit`.
Should I use `version(OSX) { extern(Objective-C) class Foo { ... ` ?

jacob-carlborg: Yes, but use `version(D_ObjectiveC)` instead.

PR #2486, opened 2019-02-09
Add rusage struct for musl

hatf0:
The rusage struct for musl is missing, causing the build for certain packages (botan, in particular) to fail. This fixes the issue, and adds in the rusage struct taken from the musl headers.

### DIFF
--- a/src/core/sys/posix/sys/resource.d
+++ b/src/core/sys/posix/sys/resource.d
@@ -466,6 +466,40 @@ else version (CRuntime_Bionic)
 else version (CRuntime_Musl)
 {
     alias ulong rlim_t;
+
+    int getrlimit(int, rlimit*);
+    int setrlimit(int, in rlimit*);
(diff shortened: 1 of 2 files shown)


src/core/sys/posix/sys/resource.d
```diff
@@ -476,10 +483,31 @@ else version (CRuntime_Musl)
         RLIMIT_NOFILE = 7,
         RLIMIT_AS     = 9,
     }
+
     int getrlimit(int, rlimit*);
     int setrlimit(int, in rlimit*);
     alias getrlimit getrlimit64;
     alias setrlimit setrlimit64;
+
+    struct rusage {
```
jacob-carlborg: Curly braces on their own lines.

hatf0:
Squashed all commits, and fixed the style issue

hatf0:
Looks like I managed to squash an unrelated commit into mine. Oops, reverting.

PR #2488, opened 2019-02-11
[uClibc][core.sys.linux.sys.eventfd] Fix definitions for mips and sparc

rracariu:
uClibc uses some different definitions see: https://github.com/kraj/uClibc/search?q=EFD_NONBLOCK&unscoped_q=EFD_NONBLOCK


rracariu:
ping.

PR #2491, opened 2019-02-15
move modules necessary to implement a GC to core.gc

rainers:
This allows compiling new GC implementations without requiring adding an import path to druntime/src.

Not sure if this is better than just adding the 3 gc-files to the import folder as is. Comments?

### DIFF
--- a/src/core/gc/config.d
+++ b/src/core/gc/config.d
@@ -5,7 +5,7 @@
 * License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
 */
 
-module gc.config;
+module core.gc.config;
 
 import core.stdc.stdio;
 import core.internal.parseoptions;
@@ -34,7 +34,7 @@ struct Config
 
     void help() @nogc nothrow
     {
-        import gc.registry : registeredGCFactories;
+        import core.gc.registry : registeredGCFactories;
 
         printf("GC options are specified as white space separated assignments:
     disable:0|1    - start disabled (%d)
--- a/src/gc/impl/proto/gc.d
+++ b/src/gc/impl/proto/gc.d
@@ -1,8 +1,7 @@
 
 module gc.impl.proto.gc;
 
-import gc.config;
-import gc.gcinterface;
+import core.gc.gcinterface;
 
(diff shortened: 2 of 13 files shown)


src/gc/impl/proto/gc.d
```diff
@@ -1,8 +1,7 @@
 
 module gc.impl.proto.gc;
 
-import gc.config;
-import gc.gcinterface;
+import core.gc.gcinterface;
```
wilzbach: Couldn't this file import the declarations from `gc.proxy` instead of defining the headers again?

src/core/gc/config.d
```diff
@@ -5,7 +5,7 @@
 * License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
 */
 
-module gc.config;
+module core.gc.config;
```
wilzbach: Do we want to expose this to the user and docs? Maybe make everything `package` for now?

src/gc/impl/proto/gc.d
rainers: Not sure what you mean. Do you prefer a cyclic dependency and a public import in gc.proxy over a simple import of the public module?

src/core/gc/config.d
rainers: I think a GC implementation should be able to use the configuration even if it does not support all the values in there.

rainers: > Maybe make everything `package` for now?

That would disallow accessing these modules from package gc AFAICT.

wilzbach: > That would disallow accessing these modules from package gc AFAICT.

Fair enough.

src/gc/impl/proto/gc.d
wilzbach: > Do you prefer a cyclic dependency

Ah sorry. Nevermind then.

rainers: Ah, I didn't see the gc_* declarations in the diff. I think we have these declarations elsewhere too, e.g. core.memory. IIRC this is done to fake some attributes.

PR #2494, opened 2019-02-19
rt: Remove some unused compiler helpers

ibuclaw:


wilzbach:
```
make: *** No rule to make target 'src/rt/obj.d', needed by 'generated/linux/release/32/libdruntime.so.a'.  Stop.
make: *** Waiting for unfinished jobs....
```

Looks like you need adjust `posix.mak` and the Windows Makefiles...

ibuclaw:
I was far too tired to follow through. :-)

wilzbach:
*** No rule to make target '/dev/shm/dtest/work/repo/dlang.org/web/phobos-prerelease/rt_obj.html'

Looks like target this also needs be removed from the doc list.

ibuclaw:
Rebasing as @JinShil got round to removing some already.

src/object.d
```diff
@@ -332,6 +332,22 @@ if (!__traits(isScalar, T1) && !__traits(isScalar, T2))
     assert(__cmp([c2, c2], [c2, c1]) > 0);
 }
 
+@safe unittest
```
JinShil: Oops!  Sorry.  But thank you.

PR #2495, opened 2019-02-23
[DO NOT PULL] OpenBSD support - WIP

rmboggs:
Hi,

I still have work to do to get this finished but it would be good to get more eyes on my work in the meantime.  Please DO NOT PULL/MERGE until I can get all the kinks out but please run this through the finest tooth comb you guys can find to make sure it follows the coding standards of druntime/dlang.  And please let me know what needs modifications and I'll get to it as time becomes available.

### DIFF
--- a/src/core/sys/openbsd/execinfo.d
+++ b/src/core/sys/openbsd/execinfo.d
@@ -1,0 +1,133 @@
+/**
+ * NetBSD implementation of glibc's $(LINK2 http://www.gnu.org/software/libc/manual/html_node/Backtraces.html backtrace) facility.
+ *
+ * Copyright: Copyright Martin Nowak 2012.
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   Martin Nowak
+ * Source:    $(DRUNTIMESRC core/sys/netbsd/_execinfo.d)
+ */
+module core.sys.openbsd.execinfo;
+
+version (OpenBSD):
+extern (C):
+nothrow:
+
+import core.sys.openbsd.dlfcn;
+
+// Use extern (D) so that these functions don't collide with libexecinfo.
+
+extern (D) int backtrace(void** buffer, int size)
+{
+    import core.thread : thread_stackBottom;
+
+    void** p, pend=cast(void**)thread_stackBottom();
+    version (D_InlineAsm_X86)
+        asm nothrow @trusted { mov p[EBP], EBP; }
+    else version (D_InlineAsm_X86_64)
+        asm nothrow @trusted { mov p[RBP], RBP; }
+    else
+        static assert(false, "Architecture not supported.");
+
+    int i;
+    for (; i < size && p < pend; ++i)
+    {
+        buffer[i] = *(p + 1);
+        auto pnext = cast(void**)*p;
+        if (pnext <= p) break;
+        p = pnext;
+    }
+    return i;
+}
+
+
+extern (D) char** backtrace_symbols(const(void*)* buffer, int size)
+{
+    static void* realloc(void* p, size_t len) nothrow
+    {
+        static import cstdlib=core.stdc.stdlib;
+        auto res = cstdlib.realloc(p, len);
+        if (res is null) cstdlib.free(p);
+        return res;
+    }
+
+    if (size <= 0) return null;
+
+    size_t pos = size * (char*).sizeof;
+    char** p = cast(char**)realloc(null, pos);
+    if (p is null) return null;
+
+    Dl_info info;
+    foreach (i, addr; buffer[0 .. size])
+    {
+        if (dladdr(addr, &info) == 0)
+            (cast(ubyte*)&info)[0 .. info.sizeof] = 0;
+        fixupDLInfo(addr, info);
+
+        immutable len = formatStackFrame(null, 0, addr, info);
+        assert(len > 0);
+
+        p = cast(char**)realloc(p, pos + len);
+        if (p is null) return null;
+
+        formatStackFrame(cast(char*)p + pos, len, addr, info) == len || assert(0);
+
+        p[i] = cast(char*)pos;
+        pos += len;
+    }
+    foreach (i; 0 .. size)
+    {
+        pos = cast(size_t)p[i];
+        p[i] = cast(char*)p + pos;
+    }
+    return p;
+}
+
+
+extern (D) void backtrace_symbols_fd(const(void*)* buffer, int size, int fd)
+{
+    import core.sys.posix.unistd : write;
+    import core.stdc.stdlib : alloca;
+
+    if (size <= 0) return;
+
+    Dl_info info;
+    foreach (i, addr; buffer[0 .. size])
+    {
+        if (dladdr(addr, &info) == 0)
+            (cast(ubyte*)&info)[0 .. info.sizeof] = 0;
+        fixupDLInfo(addr, info);
+
+        enum maxAlloca = 1024;
+        enum min = (size_t a, size_t b) => a <= b ? a : b;
+        immutable len = min(formatStackFrame(null, 0, addr, info), maxAlloca);
+        assert(len > 0);
+
+        auto p = cast(char*)alloca(len);
+        if (p is null) return;
+
+        formatStackFrame(p, len, addr, info) >= len || assert(0);
+        p[len - 1] = '\n';
+        write(fd, p, len);
+    }
+}
+
+
+private void fixupDLInfo(const(void)* addr, ref Dl_info info)
+{
+    if (info.dli_fname is null) info.dli_fname = "???";
+    if (info.dli_fbase is null) info.dli_fbase = null;
+    if (info.dli_sname is null) info.dli_sname = "???";
+    if (info.dli_saddr is null) info.dli_saddr = cast(void*)addr;
+}
+
+
+private size_t formatStackFrame(char* p, size_t plen, const(void)* addr, const ref Dl_info info)
+{
+    import core.stdc.stdio : snprintf;
+
+    immutable off = addr - info.dli_saddr;
+    immutable len = snprintf(p, plen, "%p <%s+%zd> at %s",
+                             addr, info.dli_sname, off, info.dli_fname);
+    assert(len > 0);
+    return cast(size_t)len + 1; // + '\0'
+}
--- a/src/core/sys/openbsd/sys/link_elf.d
+++ b/src/core/sys/openbsd/sys/link_elf.d
@@ -1,7 +1,7 @@
 /** 
   * D header file for OpenBSD
   *
-  * Path to online source file reference goes here
+  * https://github.com/openbsd/src/blob/master/include/link_elf.h
   */
 module core.sys.openbsd.sys.link_elf;
 
--- a/src/core/sys/openbsd/sys/link_elf.d
+++ b/src/core/sys/openbsd/sys/link_elf.d
@@ -20,7 +20,7 @@ version (OpenBSD):
 
     template ElfW(string type)
     {
-        mixin("alias Elf"~__ELF_NATIVE_CLASS.stringof~"_"~type~" Elfw;");
+        mixin("alias Elf"~__ELF_NATIVE_CLASS.stringof~"_"~type~" ElfW;");
     }
 
     struct link_map
--- a/src/core/sys/openbsd/sys/link_elf.d
+++ b/src/core/sys/openbsd/sys/link_elf.d
@@ -6,49 +6,49 @@
 module core.sys.openbsd.sys.link_elf;
 
 version (OpenBSD):
-    extern (C):
-    nothrow:
+extern (C):
+nothrow:
 
-    import core.stdc.stdint : uint64_t;
-    import core.sys.posix.sys.types : caddr_t;
-    import core.sys.openbsd.sys.elf;
+import core.stdc.stdint : uint64_t;
+import core.sys.posix.sys.types : caddr_t;
+import core.sys.openbsd.sys.elf;
 
-    version (D_LP64)
-        enum __ELF_NATIVE_CLASS = 64;
-    else
-        enum __ELF_NATIVE_CLASS = 32;
+version (D_LP64)
+    enum __ELF_NATIVE_CLASS = 64;
+else
+    enum __ELF_NATIVE_CLASS = 32;
 
-    template ElfW(string type)
-    {
-        mixin("alias Elf"~__ELF_NATIVE_CLASS.stringof~"_"~type~" ElfW;");
-    }
+template ElfW(string type)
+{
+    mixin("alias Elf"~__ELF_NATIVE_CLASS.stringof~"_"~type~" ElfW;");
+}
 
-    struct link_map
-    {
-        caddr_t     l_addr;     // Base address of library
-        char*       l_name;     // Absolute path to library
-        void*       l_ld;       // pointer to _DYNAMIC
-        link_map*   l_next;
-        link_map*   l_prev;
-    }
-    alias link_map Link_map;
+struct link_map
+{
+    caddr_t     l_addr;     // Base address of library
+    char*       l_name;     // Absolute path to library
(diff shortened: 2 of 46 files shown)


jacob-carlborg:
A draft pull request would have been suitable: https://github.blog/2019-02-14-introducing-draft-pull-requests/

src/core/sys/openbsd/execinfo.d
```diff
@@ -0,0 +1,133 @@
...
+
+import core.sys.openbsd.dlfcn;
+
+// Use extern (D) so that these functions don't collide with libexecinfo.
+
+extern (D) int backtrace(void** buffer, int size)
+{
+    import core.thread : thread_stackBottom;
+
+    void** p, pend=cast(void**)thread_stackBottom();
```
jacob-carlborg: One declaration per line.

src/core/sys/openbsd/sys/link_elf.d
```diff
@@ -0,0 +1,54 @@
...
+        ElfW!"Addr"     dlpi_addr;
+        char*           dlpi_name;
+        ElfW!"Phdr"*    dlpi_phdr;
+        ElfW!"Half"     dlpi_phnum;
+    };
+
+    private alias extern(C) int function(dl_phdr_info*, size_t, void *) dl_iterate_phdr_cb;
+    private alias extern(C) int function(dl_phdr_info*, size_t, void *) @nogc dl_iterate_phdr_cb_ngc;
+    extern int dl_iterate_phdr(dl_iterate_phdr_cb __callback, void*__data);
+    extern int dl_iterate_phdr(dl_iterate_phdr_cb_ngc __callback, void*__data) @nogc;
```
jacob-carlborg: Unnecessary indentation in this file.

src/core/sys/openbsd/sys/mman.d
```diff
@@ -0,0 +1,45 @@
...
+        alias MADV_DONTNEED = POSIX_MADV_DONTNEED;
+
+        enum MADV_SPACAVAIL = 5;    // insure that resources are reserved
+        enum MADV_FREE      = 6;    // pages are empty, free them
+
+        int madvise(void *, size_t, int);
+        int mincore(const(void) *, size_t, char *);
+        int minherit(void *, size_t, int);
+    }
+
```
jacob-carlborg: Unnecessary indentation in this file.

src/core/sys/openbsd/sys/types.d
```diff
@@ -0,0 +1,98 @@
...
+    {
+        enum _ALIGNBYTES        = 0xf;
+        alias _STACKALIGNBYTES  = _ALIGNBYTES;
+    }
+    else
+    {
+        // Not sure what to use for default since these values are
+        // not available across all OpenBSD supported platforms.
+    }
+
```
jacob-carlborg: Unnecessary indentation in this file.

src/core/sys/openbsd/time.d
```diff
@@ -0,0 +1,19 @@
...
+version (OpenBSD):
+
+    // Definitions Located in src/sys/sys/_time.h file
+    // in OpenBSD source code
+    enum CLOCK_REALTIME             = 0;
+    enum CLOCK_PROCESS_CPUTIME_ID   = 2;
+    enum CLOCK_MONOTONIC            = 3;
+    enum CLOCK_THREAD_CPUTIME_ID    = 4;
+    enum CLOCK_UPTIME               = 5;
+    enum CLOCK_BOOTTIME             = 6;
```
jacob-carlborg: Unnecessary indentation in this file.

src/core/sys/posix/arpa/inet.d
```diff
@@ -158,6 +158,31 @@ else version (NetBSD)
...
+    }
+
+    enum INET_ADDRSTRLEN    = 16;
+
+    @trusted pure
+    {
+    uint32_t htonl(uint32_t);
+    uint16_t htons(uint16_t);
+    uint32_t ntohl(uint32_t);
+    uint16_t ntohs(uint16_t);
```
jacob-carlborg: Should be indented.

src/core/sys/posix/arpa/inet.d
```diff
@@ -158,6 +158,31 @@ else version (NetBSD)
     const(char)*    inet_ntop(int, in void*, char*, socklen_t);
     int             inet_pton(int, in char*, void*);
 }
+else version (OpenBSD)
+{
+    alias uint16_t in_port_t;
+    alias uint32_t in_addr_t;
```
jacob-carlborg: Please use the assignment style syntax, i.e. `alias in_port_t = uint16_t`.

src/core/sys/posix/sys/socket.d
```diff
@@ -1119,6 +1119,222 @@ else version (NetBSD)
...
+		AF_NATM		        = 27,		/* native ATM access */
+		AF_ENCAP	        = 28,
+		AF_SIP		        = 29,		/* Simple Internet Protocol */
+		AF_KEY		        = 30,
+		pseudo_AF_HDRCMPLT  = 31,		/* Used by BPF to not rewrite headers in interface output routine */
+		AF_BLUETOOTH	    = 32,		/* Bluetooth */
+		AF_MPLS             = 33,       /* MPLS */
+		pseudo_AF_PFLOW     = 34,		/* pflow */
+		pseudo_AF_PIPEX     = 35,		/* PIPEX */
+		AF_MAX              = 36,
```
jacob-carlborg: Too much indentation.

src/core/sys/posix/sys/socket.d
```diff
@@ -1119,6 +1119,222 @@ else version (NetBSD)
...
+		AF_OSI		        = AF_ISO,
+		AF_ECMA		        = 8,		/* european computer manufacturers */
+		AF_DATAKIT	        = 9,		/* datakit protocols */
+		AF_CCITT	        = 10,		/* CCITT protocols, X.25 etc */
+		AF_SNA		        = 11,		/* IBM SNA */
+		AF_DECnet	        = 12,		/* DECnet */
+		AF_DLI		        = 13,		/* DEC Direct data link interface */
+		AF_LAT		        = 14,		/* LAT */
+		AF_HYLINK	        = 15,		/* NSC Hyperchannel */
+		AF_APPLETALK	    = 16,		/* Apple Talk */
```
jacob-carlborg: All the values are not aligned.

src/core/sys/posix/sys/ttycom.d
```diff
@@ -108,3 +108,113 @@ version (OSX)
     enum uint SLIPDISC = 4;       // serial IP discipline
     enum uint PPPDISC  = 5;       // PPP discipline
 }
+else version (OpenBSD)
+{
+    struct winsize {
```
jacob-carlborg: Curly braces on their own lines.

src/core/thread.d
```diff
@@ -3660,8 +3668,17 @@ shared static this()
...
     else version (Posix)
     {
-        PAGESIZE = cast(size_t)sysconf(_SC_PAGESIZE);
-        PTHREAD_STACK_MIN = cast(size_t)sysconf(_SC_THREAD_STACK_MIN);
+        version (OpenBSD)
+        {
+            static import core.sys.openbsd.sys.types;
+            PAGESIZE = cast(size_t)core.sys.openbsd.sys.types.PAGE_SIZE;
+            PTHREAD_STACK_MIN = cast(size_t)(1U << core.sys.openbsd.sys.types._MAX_PAGE_SHIFT);
+        }
```
jacob-carlborg: I suggest putting this `OpenBSD` block outside the `Posix` block. Will reduce indentation.

jacob-carlborg:
Have you looked at the backtrace implementation: https://github.com/dlang/druntime/tree/5a32892fd72c34b822201569ab2362ac4c3edcae/src/rt/backtrace?

src/core/sys/openbsd/execinfo.d
```diff
@@ -0,0 +1,133 @@
...
+    version (D_InlineAsm_X86)
+        asm nothrow @trusted { mov p[EBP], EBP; }
+    else version (D_InlineAsm_X86_64)
+        asm nothrow @trusted { mov p[RBP], RBP; }
+    else
+        static assert(false, "Architecture not supported.");
```
ibuclaw: Assuming that this is a copy of NetBSD, it's a pity that this non portable code got accepted.

src/core/sys/posix/arpa/inet.d
rmboggs: do you want me to fix the indentation for the code in the other versions sections as well (i.e.: lines 75-78, 100-103, 125-128, etc)?  I ask because their indentations are the same as this section.

src/core/sys/posix/arpa/inet.d
rmboggs: again, do you want me to correct the other version sections as well? (i.e.: lines 63, 64, 88, 89, 113, 114, etc.)?

src/core/sys/posix/sys/ttycom.d
rmboggs: You want me to fix the curly braces in the OSX section as well?

src/core/sys/posix/arpa/inet.d
jacob-carlborg: That should probably be fixed in a separate PR.

src/core/sys/posix/arpa/inet.d
jacob-carlborg: See above.

src/core/sys/posix/sys/ttycom.d
jacob-carlborg: No, only for the code you've added.

src/core/sys/openbsd/sys/link_elf.d
rmboggs: Fixed.

src/core/sys/openbsd/sys/mman.d
rmboggs: fixed.

src/core/sys/openbsd/sys/types.d
rmboggs: Fixed.

src/core/sys/openbsd/time.d
rmboggs: fixed.

src/core/sys/posix/arpa/inet.d
rmboggs: fixed.

src/core/sys/posix/arpa/inet.d
rmboggs: fixed.

src/core/sys/posix/sys/socket.d
rmboggs: fixed.

src/core/sys/posix/sys/socket.d
rmboggs: fixed.

src/core/sys/posix/sys/ttycom.d
rmboggs: fixed.

src/core/thread.d
rmboggs: Fixed.

src/core/sys/openbsd/execinfo.d
rmboggs: I'll need to look at the backtrace stuff again.

rmboggs:
It looks like the backtrace implementation needs a way of locating the path of the current executing file and that is what I am currently stuck on.  OpenBSD doesn't have a set way of getting this information like Linux or DragonflyBSD, it's more like FreeBSD's but with more steps needed.  I was directed to the link below for guidance on this matter but haven't gotten to writing a dlang version of this yet.

https://github.com/ponylang/ponyc/commit/3fb555fc050f307d4acfa16cc71afb75ff19f879#diff-d994394a35a330e93cc7d95e9186a884R231

At this point, I'm thinking to just close this request until i can get this resolved.  Or make a new pull request with just the OpenBSD header code to just get those bits in for the time being.

CyberShadow:
>  I was directed to the link below for guidance on this matter but haven't gotten to writing a dlang version of this yet.

In absence of a better way, that looks like a sensible "heuristic". Some modifications I'd do:

- Relative paths are handled weirdly. A leading `.` is considered relative, but `foo/bar` is not. Unless OpenBSD is doing something weird that looks like a bug in ponylang's code.

- I would avoid using `realpath` (assuming it does the same thing as the utility / Glibc function), as it does unnecessary canonicalization of symlinks. I would suggest simply concatenating `getcwd` with the argv0 path. Keep in mind that the current directory can change since program start, so it would have to be done at initialization. 

A more "correct" approach would be to open the program as a file at initialization, which would ensure we are looking at the right file, even if it was moved / deleted / recompiled, but that does use up one file descriptor.

- `execvp` does `PATH` expansion/lookup but I don't think there's anything in POSIX for just that part and not the `exec`. Might be worth checking the OpenBSD implementation of `execvp` to see if it perhaps exposes the `PATH` expansion as its own function.

jacob-carlborg:
> A more "correct" approach would be to open the program as a file at initialization, which would ensure we are looking at the right file, even if it was moved / deleted / recompiled, but that does use up one file descriptor.


All the existing platforms, except macOS, has this problem.

CyberShadow:
> All the existing platforms, except macOS, has this problem.

Not quite:
- Windows locks the file of mapped modules, so you can't delete it (or sometimes rename it?)
- I believe Linux allows accessing even deleted files via `/proc/self/exe` (it is represented as a symlink, but actually opening it allows reading said file, i.e. it is not equivalent to open(readlink(...))`.

jacob-carlborg:
> Windows locks the file of mapped modules, so you can't delete it (or sometimes rename it?)

Right, but there isn't an implementation for Windows, as far as I can see.

> I believe Linux allows accessing even deleted files via /proc/self/exe (it is represented as a symlink, but actually opening it allows reading said file, i.e. it is not equivalent to open(readlink(...))`.

Ah, interesting. What happens if it's recompiled?

CyberShadow:
> Right, but there isn't an implementation for Windows, as far as I can see.

For the purpose of reading debug information for stack traces, yeah, that's not necessary on Windows, but more generally we're essentially talking about [`thisExePath`](https://dlang.org/phobos/std_file.html#thisExePath).

> Ah, interesting. What happens if it's recompiled?

You get the original file!

I tested it with this program:

```d
import std.stdio;

void main()
{
	writeln("magic-string-1");
	readln();
}
```

Ran it, opened a new terminal, changed the 1 to 2, rebuilt it, then ran `strings /proc/$(pgrep exetest)/exe | grep magic-string-`. Still shows 1.

I guess it depends "how" the compiler/linker overwrites the file - I assume that if it were to do the equivalent of `dd of=exetest` then it would clobber the original file, but the typical `fopen(fn, "w")` will create a new file.

jacob-carlborg:
> For the purpose of reading debug information for stack traces, yeah, that's not necessary on Windows, but more generally we're essentially talking about thisExePath.

This is druntime 😉.

> You get the original file!

👍.

jacob-carlborg:
@rmboggs what does `getprogname` [1] return? Just the name and not the full path?

[1] https://man.openbsd.org/getprogname.3

CyberShadow:
> The setprogname() function sets the name of the program to be **the last path component** of the name argument.

Looks like it can't have a path.

jacob-carlborg:
> Looks like it can't have a path.

😞.

rmboggs:
Was going to look at getprogname (3) next but it looks like either route I take, I need to fish for the correct directly by searching current path and every direction listen in the $PATH environment variable. 

To add to this, I have the same problem in phobos with the thisExePath function as well. So whatever the solution is, it feels like there will be code duplication between druntime and phobos, which I am not too thrilled about.

jacob-carlborg:
> To add to this, I have the same problem in phobos with the thisExePath function as well. So whatever the solution is, it feels like there will be code duplication between druntime and phobos, which I am not too thrilled about.

Perhaps `thisExePath` can be moved to druntime. At least as an internal function.

PR #2498, opened 2019-02-27
Move emplaceRef to core/internal

edi33416:
Move `emplaceRef` to `core.internal.lifetime` so we can remove the code duplication from Phobos.
`emplaceRef` is intentionally non-ddoc'ed

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -27,6 +27,7 @@ COPY=\
 	$(IMPDIR)\core\internal\string.d \
 	$(IMPDIR)\core\internal\traits.d \
 	$(IMPDIR)\core\internal\utf.d \
+	$(IMPDIR)\core\internal\lifetime.d \
 	\
 	$(IMPDIR)\core\stdc\assert_.d \
 	$(IMPDIR)\core\stdc\complex.d \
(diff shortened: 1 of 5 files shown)


thewilsonator:
Why does this need to be moved for phobos deduplication?

thewilsonator:
At the very least it should be imported by `core.lifetime`

cc'ing @TurkeyMan

edi33416:
> Why does this need to be moved for phobos deduplication?

`emplaceRef` is used internally in various parts of Phobos. See @wilzbach 's comment [here](https://github.com/dlang/druntime/pull/2442#issuecomment-459377410)

wilzbach:
> I don't think we should advertise `emplaceRef` as part of the public API, or at least not yet

Imho it's not going to go anywhere anyhow (like `emplace`), so we could just start to officially lock us in ;-)
Though, of course `core.internal` is safer and solves the duplication problem.

edi33416:
> Why does this need to be moved for phobos deduplication?



> At the very least it should be imported by `core.lifetime`
> 
> cc'ing @TurkeyMan

I chose local imports against top level ones so a user won't pay the import cost if he doesn't use `emplace`.

PR #2499, opened 2019-03-01
Quote variables in win64.mak

wilzbach:
The escaping is required for https://github.com/dlang/dmd/pull/9398.

### DIFF
--- a/win64.mak
+++ b/win64.mak
@@ -8,7 +8,7 @@ SDKDIR=\Program Files (x86)\Microsoft SDKs\Windows\v7.0A
 DMD_DIR=..\dmd
 BUILD=release
 OS=windows
-DMD=$(DMD_DIR)\generated\$(OS)\$(BUILD)\$(MODEL)\dmd
+DMD="$(DMD_DIR)\generated\$(OS)\$(BUILD)\$(MODEL)\dmd"
 
 CC="$(VCDIR)\bin\amd64\cl"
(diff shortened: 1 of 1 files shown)


thewilsonator:
Read auto-merge when you're done, I'll be away for a bit.

win64.mak
```diff
@@ -8,7 +8,7 @@ SDKDIR=\Program Files (x86)\Microsoft SDKs\Windows\v7.0A
 DMD_DIR=..\dmd
 BUILD=release
 OS=windows
-DMD=$(DMD_DIR)\generated\$(OS)\$(BUILD)\$(MODEL)\dmd
+DMD="$(DMD_DIR)\generated\$(OS)\$(BUILD)\$(MODEL)\dmd"
```
rainers: I think we should never quote when assigning to a variable, see also https://github.com/dlang/druntime/pull/2438

That also avoids the horrible quoting when invoking a sub make command, e.g. https://github.com/dlang/druntime/pull/2499/files#diff-e35cbd05831954a48feae8130a5df3f4R89 (even broken in this case).

rainers:
I suspect this PR broke the build on master and other PRs. AFAICT win-farm2 uses a different VC path and fails because it contains spaces.

wilzbach:
Argh sorry about this. Here's the revert -> https://github.com/dlang/druntime/pull/2500

We really need to move away from the auto-tester or at least make all its clients behave the same :scream:

PR #2500, opened 2019-03-01
Revert "Quote variables in win64.mak"

wilzbach:
Reverts dlang/druntime#2499 (as it appears to have broken win-farm2)


Basile-z:
Manual merge needed maybe ? The fix seems to be affected by the problem it's supposed to solve.

PR #2502, opened 2019-03-01
[stable] gc.impl.conservative.gc: Fix fwd declaration signature

kinke:
LDC is picky about this: https://github.com/dlang/druntime/blob/f2276846d53306cb8e2832fc0319c1ac89eddd71/src/core/exception.d#L590


wilzbach:
DAutoTest failure is unrelated. It's because 2.085 has been tagged at some places, but not fully.

kinke:
I revised all 4 signatures and added some attributes.

PR #2504, opened 2019-03-02
Fix Issue 10535 - Add initialize function to allocate an empty AA

ntrel:
See changelog/docs for details. I'll make a dlang.org pull for this soon.

This is adapted from a patch by Ketmar Dark: https://issues.dlang.org/show_bug.cgi?id=16269#c4.
CC'ing @jmdavis @schveiguy.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3024,6 +3024,7 @@ extern (C)
     inout(void)[] _aaKeys(inout void* p, in size_t keysize, const TypeInfo tiKeyArray) pure nothrow;
     void* _aaRehash(void** pp, in TypeInfo keyti) pure nothrow;
     void _aaClear(void* p) pure nothrow;
+    void _aaInitialize(void* p, const TypeInfo_AssociativeArray ti);
 
     // alias _dg_t = extern(D) int delegate(void*);
     // int _aaApply(void* aa, size_t keysize, _dg_t dg);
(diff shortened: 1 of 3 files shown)


wilzbach:
You might want to have a look at a very similar attempt: `reserve` -> https://github.com/dlang/druntime/pull/1929

src/object.d
```diff
@@ -3057,10 +3058,47 @@ void* aaLiteral(Key, Value)(Key[] keys, Value[] values) @trusted pure
...
+ * Allocate memory for an empty associative array.
+ * This is useful to avoid copying a null reference when each reference
+ * should point to the same associative array data.
+ * Without using this function, an associative array is initialized only
+ * when the first element is added.
+ * Params:
+ *      aa = The associative array.
+ * See_Also: $(LREF clear)
+ */
+void initialize(T : Value[Key], Value, Key)(ref T aa) @trusted
```
jacob-carlborg: I don't like filling up the `object` module with a lot of symbols. This is a pretty common name as well.

rainers:
I guess it has been suggested before, but maybe it would be better to make
```
    int[int] aa = [];
```
allocate the internal data struct. It currently yields a nonsense `Error: cannot have associative array key of void`.

ntrel:
@wilzbach `reserve(0)` could work for this, but it looks like we don't have an implementation for non-zero sizes ATM - would that be a blocker? (I think `reserve` is less intuitive).

@jacob-carlborg Andrei suggested adding a function to initialize an empty AA, instead of adding built-in syntax, in 2013:
https://forum.dlang.org/post/kr2vhl$2jf6$1@digitalmars.com

@rainers `[]` could work, but it then means a silent allocation in an AA context, and no allocation in a slice context, and which of these contexts it is is sometimes not visible at the use-site. I think users would also do things like `aa == []`, causing unnecessary allocations and possible confusion. I suggested `aa = new AA` as being clearer syntax in the issue tracker.

rainers:
`new AA` sounds good to me too. With `alias AA = int[string]` it currently errors out with `new can only create structs, dynamic arrays or class objects, not int[string]'s` and `new int[string]` yields `Error: cannot pass type string as a function argument`. So code breakage is unlikely.

jacob-carlborg:
> jacob-carlborg Andrei suggested adding a function to initialize an empty AA, instead of adding built-in syntax, in 2013:
https://forum.dlang.org/post/kr2vhl$2jf6$1@digitalmars.com

I still don't like adding a new function for it. Using `new AA` works for me too. It's not really like we're adding new built-in syntax. It's already there, we're just adding a meaning to it.

PR #2505, opened 2019-03-03
[GDC] Update MinGW code

jpf91:
For https://github.com/D-Programming-GDC/gcc/pull/1 . Basically removes most of the code, adds some slight changes to fiber code.


src/core/stdc/stdio.d
```diff
@@ -1146,97 +1146,30 @@ void setbuf(FILE* stream, char* buf);
...
-    int vprintf(scope const char* format, va_list arg);
-    ///
-    int vscanf(scope const char* format, va_list arg);
-    ///
-    int printf(scope const char* format, ...);
-    ///
-    int scanf(scope const char* format, ...);
-}
+///
+int fprintf(FILE* stream, scope const char* format, ...);
```
kinke: Have you verified that those work with 80-bit `real`? Last time I checked, these are the MSVC functions, where `long double` is 64-bit, so that `printf("%Lf", myReal)` prints garbage.

jpf91: No really, I only checked the unittests which seem to be OK but might not test this at all. How do LDC/DMD handle this when targeting MSVC? Do you use 64bit reals instead? We could do that in GDC, just set `-mlong-double-64` by default.

To be honest, MinGW is a low-priority target for me. It's not well supported, developing for windows is frustrating (even running the GDC testsuite is a hacky process, as it uses symlinks..., there's no well-working SSH server, IO and building GCC is way too slow, ...). People will also always find something to complain about (uses emulated TLS, uses dwarf debug info, c++ abi is not compatible with MSVC, use .a import libraries) so I want to keep as little MinGW specific code as possible. Preferably, windows users should just use LDC, but we also need something to compile GDC with the D frontend for windows hosts and I'm not sure if GCC can easily be built with clang.

kinke: > How do LDC/DMD handle this when targeting MSVC? Do you use 64bit reals instead?

LDC, unlike DMD, uses 64-bit `real`, but for MSVC targets only, i.e., (only partially supported) MinGW targets use 80-bit extended precision.

> People will also always find something to complain about (uses emulated TLS, uses dwarf debug info, c++ abi is not compatible with MSVC, use .a import libraries) so I want to keep as little MinGW specific code as possible.

Same here. :]

> Preferably, windows users should just use LDC

Heh, I always thought MinGW users should just use GDC. ;)

thewilsonator: The real irony here is that as DMD is moving away from DMC/optlink they are considering lld+mingw libraries.

jpf91: Oh, I didn't know LDC supports a MinGW target. So I'll revert these changes here.

> Heh, I always thought MinGW users should just use GDC. ;)
This makes me wonder: Do you actually use native TLS when targeting MinGW or the libgcc emutls? If you use emutls, how did you manage to integrate it with the GC? I think in the history of GDC, we never really had reliable MinGW builds because of that, the GC has always been broken. The only exception might be Daniels old builds which patched GCC and binutils to support native windows TLS ;-)

jpf91: @kinke Do you need any of the other MinGW definitions here? `fenv.d`, `math.d` and the rest of the changed files?

jpf91: @thewilsonator 
> The real irony here is that as DMD is moving away from DMC/optlink they are considering lld+mingw libraries.

I think when using LLD+MinGW import libraries only you're not affected by most of the problems: DWARF debug info / object file size / missing TLS support / C++ ABI are GCC and binutils problems, using LLD/clang you're probably fine.

thewilsonator: Ahh, good to know.

kinke: > Oh, I didn't know LDC supports a MinGW target. So I'll revert these changes here.

LDC used to support MinGW (but not 100% I guess - especially wrt. C++ ABI/name-mangling), but doesn't anymore - MSVC is our primary/exclusive Windows target. There shouldn't be that many changes to get it working again (mainly druntime/Phobos), but it's been like that for years now, so I don't think proper MinGW support is coming back anytime soon (I have 0 interest in it).

Wrt. TLS, according to https://run.dlang.io/is/PR7PuP we use native TLS. clang has the option to produce MSVC objects with dwarf debuginfo; apparently some Windows guys prefer gdb over the Visual Studio debugger. I'd also expect LLD to be usable as linker for MinGW in general (support for .a, dwarf, huge objects).

dnadlinger: > LDC used to support MinGW (but not 100% I guess - especially wrt. C++ ABI/name-mangling & EH)

Yes, LDC/MinGW was the fully working first non-DMD compiler targeting Windows. The C++ ABI/EH interface wasn't a thing back then.

jpf91: OK, reverted this change.

src/core/stdc/stdlib.d
```diff
@@ -128,13 +128,6 @@ version (CRuntime_Microsoft)
         return strtod(nptr, endptr);
     }
 }
-else version (MinGW)
-{
-    ///
-    real __mingw_strtold(scope inout(char)* nptr, scope inout(char)** endptr);
-    ///
-    alias __mingw_strtold strtold;
-}
```
kinke: Is this intentional? I.e., leaving the specialized MinGW printf functions to support full x87 precision, but crippling strtold to double precision?

jpf91: That was an oversight, thanks.

Many MinGW specific code paths in phobos (especially in unit tests) are caused by different formatting of reals caused by the `_mingw_printf` functions. I'd like to get rid of this. OTOH we probably can't default to 64 bit reals, as people expect MinGW gcc to be compatible with MinGW gdc with default settings. Long-term I guess I'll track down where in phobos we call the C snprintfs, convert real to double there instead and make it call the standard MSVC snprintfs. Not sure what to do for these C bindings then though, ideally users should be able to call either the extended mingw function or the default msvc version. 

But all that is probably for GCC 10, I'd be happy if we get some sort of even basic MinGW support into GCC9, we're actually already quite late for that.

kinke: > Many MinGW specific code paths in phobos (especially in unit tests) are caused by different formatting of reals caused by the _mingw_printf functions.

I know, and I don't like it either, but it's just due to the weird hybrid nature of the MinGW runtime - MSVC runtime + 80-bit long-double + some C99/POSIX stuff which used to be missing in earlier MSVC versions (but C99 support is thankfully nearly complete nowadays).

The real shame to be fixed IMO is that Phobos still depends on C formatting functions.

jacob-carlborg: > The real shame to be fixed IMO is that Phobos still depends on C formatting functions.

I agree.

src/core/stdc/fenv.d
```diff
@@ -44,8 +44,6 @@ version (SystemZ) version = IBMZ_Any;
 version (X86)     version = X86_Any;
 version (X86_64)  version = X86_Any;
 
-version (MinGW)
-    version = GNUFP;
```
kinke: Looking at MinGW-w64 `fenv.h`, this should stay as-is.

src/core/stdc/math.d
```diff
@@ -893,100 +893,6 @@ else version (CRuntime_UClibc)
...
-    int signbit(double x)    { return __signbit(x);  }
-    ///
-    int signbit(real x)
-    {
-        return (real.sizeof == double.sizeof)
-            ? __signbit(x)
-            : __signbitl(x);
-    }
-  }
-}
```
kinke: These are macros, so using the `CRuntime_Microsoft` block would probably be okay-ish (although the constants don't match etc.). But the `CRuntime_Microsoft` block doesn't handle MinGW's 80-bit `real`, i.e., all of these functions will assert, the only exception being `signbit()`.

jpf91:
Thanks for all the review comments, seems like I didn't think this through to the end.

kinke:
The thread/fiber changes looked good though. :)

PR #2506, opened 2019-03-05
Refactor TypeInfo_Array

edi33416:
This is a first step towards refactoring TypeInfo_Array. Ideally, imho, we would replace all the inheritance chain with templates.


src/rt/typeinfo/ti_Afloat.d
```diff
@@ -19,43 +19,12 @@ private import rt.util.typeinfo;
...
-    }
-
-    override @property inout(TypeInfo) next() inout
-    {
-        return cast(inout)typeid(F);
-    }
+    mixin TypeInfo_A_T!float;
 }
 
 // ifloat[]
```
edi33416: I feel that we can now delete those comments, as the type is now stated with the `mixin` instantiation.
Ex. `mixin TypeInfo_A_T!ifloat;`

jacob-carlborg:
If you want this to be replaced with templates, is there any reason to do this refactoring? Will more of the code now be reusable if/when this code is replaced with templates?

JinShil:
Since I had, at one time, an interest in reducing druntime's dependency on `TypeInfo`, I was asked to comment on this PR, so here are my thoughts...

I'm not sure what' motivating this PR. It appears to be simply reducing some boilerplate and increasing code re-use.  That's fine, but that does not address the fundamental issue with `TypeInfo` and why we want to use more templates in druntime.

The fundamental problem with druntime's use of `TypeInfo` has to do with the [runtime hooks](https://wiki.dlang.org/Runtime_Hooks) which are basically just compiler lowerings.  For example, consider the following code:

```D
void main()
{
    auto bArray = new byte[5];
    bArray.length = 8;  // gets lowered to `_d_arraysetlengthT`
}
```

`bArray.length = 8;` gets lowered by the compiler to the runtime hook [`_d_arraysetlengthT`](https://github.com/dlang/druntime/blob/c86e4a0e541dc63f9e9f2ee09c7efd325c0be2d5/src/rt/lifetime.d#L1443).  If you look at the signature for `_d_arraysetlengthT`, you'll see that it takes a runtime `TypeInfo` object.  That is unfortunate because everything the compiler needs for the implementation is known at compile-time.  Therefore, `_d_arraysetlengthT` should be converted to a template, and the compiler should be changed to lower `bArray.length = 8;` to that template implementation instead.  There are a number of benefits, some of which should be self-evident IMO, so I won't elaborate at this time.

Although it was before my time, my understanding for why things are currently implemented with runtime `TypeInfo` when everything is known at compile-time is because the runtime hooks were written before D had templates.  Now that D has templates, it seems, at least to me, that templates is the obvious choice.

There is another problem, however, when converting these runtime hooks to templates.  In the compiler, the lowerings are performed after the semantic phase in [e2ir.d](https://github.com/dlang/dmd/blob/f4f29c3e47ee30b2646cede35ed0f9bd823f0add/src/dmd/e2ir.d#L2611).  This means that currently, for example, [setting the length of a dynamic array can be done in a `@safe`, `pure`, `nothrow` context](https://run.dlang.io/is/2aefNu), but [`_d_arraysetlengthT` is neither `@safe`, `pure`, nor `nothrow`](https://github.com/dlang/druntime/blob/c86e4a0e541dc63f9e9f2ee09c7efd325c0be2d5/src/rt/lifetime.d#L1443).  The compiler is lying to us.  Therefore, if you convert these runtime hooks to templates, [the lowering will be done in the semantic phase](https://github.com/dlang/dmd/pull/8531), and suddenly, you'll be forced to be honest.  After a conversation with @WalterBright, I was told to make use of [`pureMalloc`](https://github.com/dlang/druntime/blob/29f495b5b3571484bcf4d0af2fe211d6b7d86830/src/core/memory.d#L862) to work around the `pure` constraint.  `trusted` can be used to work around the `@safe` constraint, and that instead of throwing exceptions in `nothrow` code, to use `assert`ions.

So, I don't know what's motivating this PR, but it doesn't seem to address the fundamental problem with `TypeInfo` in the runtime.  It may still be an improvement to the code, but I'll leave that for others to decide.

src/rt/util/typeinfo.d
```diff
@@ -104,6 +109,133 @@ if (is(T ==  float) || is(T ==  double) || is(T ==  real) ||
     public alias hashOf = core.internal.hash.hashOf;
 }
 
+enum isIntegralTypeInfoT(T) = is(T == int) || is(T == uint) || is(T == dchar) ||
+        is(T == long) || is(T == ulong) ||
+        is(T == short) || is(T == ushort) || is(T == wchar) ||
+        is(T == byte) || is(T == ubyte) || is(T == void) || is(T == bool) ||
+        is(T == char) || is(T == immutable(char)) || is(T == const(char));
```
andralex: This is long and also misses a variety of types, such as const int etc. How about using https://dlang.org/spec/traits.html#isIntegral?

src/rt/util/typeinfo.d
```diff
@@ -104,6 +109,133 @@ if (is(T ==  float) || is(T ==  double) || is(T ==  real) ||
...
+
+private template Array(T)
+if (isIntegralTypeInfoT!T)
+{
+    bool equals(T[] s1, T[] s2)
```
andralex: Ideally there'd be some `@trusted` here and stuff.

PR #2508, opened 2019-03-08
[WIP] Lower newExp

edi33416:
Add a templated version of `_d_newitemT` that drops `TypeInfo` and uses the type known at compile time.

This is required by this dmd [pr](https://github.com/dlang/dmd/pull/9433)

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -85,18 +85,17 @@ private T* ___d_newitemU(T)()
 T* ___d_newitemT(T)()
 {
     import core.stdc.string;
-    auto p = ___d_newitemU!T();
-    memset(p, 0, T.sizeof);
-    return p;
-}
-
-// Same as above, for item with non-zero initializer.
-T* ___d_newitemiT(T)()
-{
-    import core.stdc.string;
-    auto p = ___d_newitemU!T();
-    auto init = T.init;
-    memcpy(p, &init, T.sizeof);
+    import core.internal.traits : Unqual;
+    auto p = (() @trusted => ___d_newitemU!T())();
+    static if (__traits(isZeroInit, T))
+    {
+        () @trusted { memset(cast(Unqual!T *) p, 0, T.sizeof); }();
+    }
+    else
+    {
+        auto init = T.init;
+        () @trusted { memcpy(cast(Unqual!T *) p, &init, T.sizeof); }();
+    }
     return p;
 }
 
--- a/src/object.d
+++ b/src/object.d
@@ -38,6 +38,68 @@ alias dstring = immutable(dchar)[];
 
 version (D_ObjectiveC) public import core.attribute : selector;
 
+// Size used to store the TypeInfo at the end of an allocation for structs that
+// have a destructor
+private size_t structTypeInfoSize(T)() pure nothrow @nogc
+{
+    // can't use because it's in rt.lifetime
+    //if (!callStructDtorsDuringGC)
+        //return 0;
+
+    static if (is(T == struct) && is(typeof({ T t; t.__xdtor; })))
+        return size_t.sizeof;
+    return 0;
+}
+
+/*
+ * Allocate an uninitialized non-array item.
+ * This is an optimization to avoid things needed for arrays like the __arrayPad(size).
+ */
+private T* ___d_newitemU(T)()
+{
+    import core.memory;
+    import core.internal.traits : Unqual;
+
+    auto ti = typeid(Unqual!T);
+    auto flags = !(ti.flags & 1) ? GC.BlkAttr.NO_SCAN : 0;
+    enum tiSize = structTypeInfoSize!T();
+    enum size = T.sizeof + tiSize;
+
+    static if (is(T == struct))
+    {
+        if (tiSize)
+            flags |= GC.BlkAttr.STRUCTFINAL | GC.BlkAttr.FINALIZE;
+    }
+    auto blkInf = GC.qalloc(size, flags, ti);
+    auto p = blkInf.base;
+
+    static if (is(T == struct))
+    {
+        if (tiSize)
+            *cast(TypeInfo*)(p + blkInf.size - tiSize) = cast() ti;
+    }
+    return cast(T*) p;
+}
+
+// Same as above, for item with non-zero initializer.
+T* ___d_newitemT(T)()
+{
+    import core.stdc.string;
+    auto p = ___d_newitemU!T();
+    memset(p, 0, T.sizeof);
+    return p;
+}
+
+// Same as above, for item with non-zero initializer.
+T* ___d_newitemiT(T)()
+{
+    import core.stdc.string;
+    auto p = ___d_newitemU!T();
+    auto init = T.init;
+    memcpy(p, &init, T.sizeof);
+    return p;
+}
+
 int __cmp(T)(scope const T[] lhs, scope const T[] rhs) @trusted
     if (__traits(isScalar, T))
 {


src/object.d
```diff
@@ -38,6 +38,68 @@ alias dstring = immutable(dchar)[];
...
+// have a destructor
+private size_t structTypeInfoSize(T)() pure nothrow @nogc
+{
+    // can't use because it's in rt.lifetime
+    //if (!callStructDtorsDuringGC)
+        //return 0;
+
+    static if (is(T == struct) && is(typeof({ T t; t.__xdtor; })))
+        return size_t.sizeof;
+    return 0;
```
thewilsonator: This will complain about unreachable statement if the above static if is true.

src/object.d
```diff
@@ -38,6 +38,68 @@ alias dstring = immutable(dchar)[];
...
+ * Allocate an uninitialized non-array item.
+ * This is an optimization to avoid things needed for arrays like the __arrayPad(size).
+ */
+private T* ___d_newitemU(T)()
+{
+    import core.memory;
+    import core.internal.traits : Unqual;
+
+    auto ti = typeid(Unqual!T);
+    auto flags = !(ti.flags & 1) ? GC.BlkAttr.NO_SCAN : 0;
```
thewilsonator: I thought the whole point was to avoid typeid? Surely you should be able to use a compiler trait for `ti.flags & 1`.

jacob-carlborg: Otherwise one can be added.

src/object.d
```diff
@@ -38,6 +38,68 @@ alias dstring = immutable(dchar)[];
...
+    memset(p, 0, T.sizeof);
+    return p;
+}
+
+// Same as above, for item with non-zero initializer.
+T* ___d_newitemiT(T)()
+{
+    import core.stdc.string;
+    auto p = ___d_newitemU!T();
+    auto init = T.init;
```
jacob-carlborg: `const`.

src/object.d
```diff
@@ -38,6 +38,68 @@ alias dstring = immutable(dchar)[];
...
+    }
+    return cast(T*) p;
+}
+
+// Same as above, for item with non-zero initializer.
+T* ___d_newitemT(T)()
+{
+    import core.stdc.string;
+    auto p = ___d_newitemU!T();
+    memset(p, 0, T.sizeof);
```
jacob-carlborg: The comment says "for item with non-zero initializer". But doesn't this initialize to zero?

jacob-carlborg: Is there a problem of using `T.init` when it's zero? Why can't the function below be used for all types?

edi33416: Maybe `memset` is faster than `memcpy`? Not sure; I'll do some benchmarking

dnadlinger: Zero initialisers don't actually get emitted as such; the _pointer_ then is zero.

src/object.d
edi33416: Maybe `static const` even?

src/object.d
edi33416: I haven't found a compiler trait. Maybe I missed it. Anyone know about one?

src/object.d
```diff
@@ -38,6 +38,67 @@ alias dstring = immutable(dchar)[];
 
 version (D_ObjectiveC) public import core.attribute : selector;
 
+// Size used to store the TypeInfo at the end of an allocation for structs that
+// have a destructor
+private size_t structTypeInfoSize(T)() pure nothrow @nogc
+{
+    // can't use because it's in rt.lifetime
```
edi33416: What about this? Any solutions?

src/object.d
```diff
@@ -38,6 +38,67 @@ alias dstring = immutable(dchar)[];
...
+    //if (!callStructDtorsDuringGC)
+        //return 0;
+
+    static if (is(T == struct) && is(typeof({ T t; t.__xdtor; })))
```
edi33416: Is this the same as `TypeInfo_Struct.xdtor()`?

src/object.d
```diff
@@ -38,6 +38,67 @@ alias dstring = immutable(dchar)[];
...
+    import core.stdc.string;
+    import core.internal.traits : Unqual;
+    auto p = (() @trusted => ___d_newitemU!T())();
+    static if (__traits(isZeroInit, T))
+    {
+        () @trusted { memset(cast(Unqual!T *) p, 0, T.sizeof); }();
+    }
+    else
+    {
+        auto init = T.init;
```
andralex: `immutable init = T.init;`

src/object.d
jacob-carlborg: The simplest would be to make `callStructDtorsDuringGC` `package` and import `rt.lifetime`.

src/object.d
jacob-carlborg: No, not that I can see. But add one. I think it's currently handled here: https://github.com/dlang/dmd/blob/bf26fb4ffc173f016eae9c241bcfd5a88da92957/src/dmd/toobj.d#L1291-L1327.

src/object.d
jacob-carlborg: Isn't `__gshared` redundant when it's `immutable`?

src/object.d
```diff
@@ -38,6 +38,67 @@ alias dstring = immutable(dchar)[];
...
+
+    static if (is(T == struct))
+    {
+        if (tiSize)
+            *cast(TypeInfo*)(p + blkInf.size - tiSize) = cast() ti;
+    }
+    return cast(T*) p;
+}
+
+// Same as above, for item with non-zero initializer.
```
jacob-carlborg: This is a public symbol, should it has a Ddoc comment?

src/object.d
```diff
@@ -38,6 +38,67 @@ alias dstring = immutable(dchar)[];
...
+    static if (is(T == struct) && is(typeof({ T t; t.__xdtor; })))
+        return size_t.sizeof;
+    return 0;
+}
+
+/*
+ * Allocate an uninitialized non-array item.
+ * This is an optimization to avoid things needed for arrays like the __arrayPad(size).
+ */
+private T* ___d_newitemU(T)()
```
jacob-carlborg: I don't think these names, `___d_newitemU` and `___d_newitemT`, are very descriptive.  It's difficult to know what's the difference just by looking at their names. Perhaps `___d_newitemU` instead should contain the word "allocate".

src/object.d
```diff
@@ -38,6 +38,67 @@ alias dstring = immutable(dchar)[];
...
+    //if (!callStructDtorsDuringGC)
+        //return 0;
+
+    static if (is(T == struct) && is(typeof({ T t; t.__xdtor; })))
+        return size_t.sizeof;
+    return 0;
+}
+
+/*
+ * Allocate an uninitialized non-array item.
```
jacob-carlborg: What's an `item`? It's not very descriptive.

src/object.d
```diff
@@ -38,6 +38,67 @@ alias dstring = immutable(dchar)[];
...
+ * This is an optimization to avoid things needed for arrays like the __arrayPad(size).
+ */
+private T* ___d_newitemU(T)()
+{
+    import core.memory;
+    import core.internal.traits : Unqual;
+
+    auto ti = typeid(Unqual!T);
+    auto flags = !(ti.flags & 1) ? GC.BlkAttr.NO_SCAN : 0;
```
jacob-carlborg: Avoid using magic constants like this `1`.

src/object.d
edi33416: Can't import `rt.lifetime`. You can't import anything from `rt/*` since it's not in `mak/COPY`'s `$(IMPDIR)`.

Maybe I'm missing something, as the GC is able [to import](https://github.com/dlang/druntime/blob/master/src/gc/impl/conservative/gc.d#L41) from `rt/util/`

jacob-carlborg: Then you can do a hack, declare an `extern` variable and use `pragma(mangle)` to get the same mangling as for `callStructDtorsDuringGC` declared in `rt.lifetime`.

wilzbach: We can do even better. Here's an example from the existing code:

```d
import core.internal.traits : externDFunc;
alias rt_tlsgc_init = externDFunc!("rt.tlsgc.init", void* function() nothrow @nogc);
```

jacob-carlborg: Does that work with non-function types?

kinke:
> Add a templated version of _d_newitemT that drops TypeInfo

It clearly doesn't, so what's the benefit of this? Right now it's just bloating code if I understand correctly, and replaces 1-2 runtime-ifs with static ifs.

kinke:
To reduce bloat and still allow for future revamping without TypeInfo or whatever, and including some potential performance improvements due to static size for initialization, we could also have something like

```D
T* _d_newValueOnHeap(T, bool initialize)()
{
    import core.internal.traits : Unqual;
    return cast(T*) newValueOnHeap!(Unqual!T, initialize)();
}

private T* newValueOnHeap(T, bool initialize)() @trusted
{
    enum hasDtor = is(T == struct) && is(typeof({ T t; t.__xdtor; }));
    auto ti = typeid(T);
    auto p = allocateValueOnHeap!(hasDtor)(ti);

    static if (initialize)
    {
        import core.stdc.string;

        static if (__traits(isZeroInit, T))
        {
            memset(p, 0, T.sizeof);
        }
        else
        {
            const init = ti.initializer();
            memcpy(p, init.ptr, T.sizeof);
        }
    }

    return p;
}

private T* allocateValueOnHeap(bool hasDtor)(TypeInfo ti)
{
    import core.memory;

    const fullSize = ti.tsize() + (hasDtor ? size_t.sizeof : 0);
    auto flags = !(ti.flags & 1) ? GC.BlkAttr.NO_SCAN : 0;
    static if (hasDtor)
        flags |= GC.BlkAttr.STRUCTFINAL | GC.BlkAttr.FINALIZE;
    auto blkInf = GC.qalloc(fullSize, flags, ti);
    auto p = blkInf.base;

    static if (hasDtor)
        *cast(TypeInfo*)(p + blkInf.size - size_t.sizeof) = cast() ti;

    return cast(T*) p;
}
```

src/object.d
JinShil: I didn't know what `(ti.flags & 1)` meant, so I had to go spelunking in DMD source code to see what it means.  Assuming I'm reading the DMD source code correctly, it's `isCOMclass`.  At a minimum maybe you could do something like `enum COMClass = 1` to save the next poor soul the trouble.

Second, according to https://dlang.org/spec/interface.html#com-interfaces does that flag mean that it inherits from `core.stdc.windows.com.IUnknown`?  Is there an existing trait that can check for that?  Or can you use `__traits(compiles,...)` to achieve the desired result.

JinShil:
I'm ok with a direct, imperfect translation.  Once DMD is modified to call the template correctly, the implementation of the template can be improved with followup PRs and additional unittests.

src/object.d
kinke: It's really bad: https://github.com/dlang/druntime/blob/4ad638f61a9b4a98d8ed6eb9f9429c0ef6afc8e3/src/object.d#L1313-L1315

Especially as there is this easily confusing beauty (used for `TypeInfo_Class.m_flags`) you referred to: https://github.com/dlang/druntime/blob/4ad638f61a9b4a98d8ed6eb9f9429c0ef6afc8e3/src/object.d#L1969-L1980

kinke:
I was disappointed by the lack of reasoning and the reviews just tackling the surface, boring stylistic things. And then no response for 20 days. I'd be okay with following up myself (although the 2nd template param would require a DMD change too, so doing it properly the first time would be less hassle).

JinShil:
To aid this review, I think this PR could use a couple of unittests that call the `__newItem` directly, just to test the implementation.  At least I found that useful when working on https://github.com/dlang/druntime/pull/2264.

src/object.d
JinShil: You could add an `extern (C)` function in lifetime.d to return `callStructDtorsDuringGC` and then just declare that prototype in object.d.  You can already find precedent for that in object.d (e.g. https://github.com/dlang/druntime/blob/4ad638f61a9b4a98d8ed6eb9f9429c0ef6afc8e3/src/object.d#L3909-L3910)

src/object.d
JinShil: I'm afraid I don't know at this time.  I'm sorry.

src/object.d
JinShil: https://github.com/dlang/druntime/blob/4ad638f61a9b4a98d8ed6eb9f9429c0ef6afc8e3/src/object.d#L2194

Now I'm thoroughly confused.

rainers: This one is for structs, not classes. TypeInfo_Class.flags() always returns 1 as TypeInfo_Class tries to describe both the reference and the instance of a class at the same time (which is causing trouble elsewhere, too, e.g. when trying to specify the type of memory for the precise GC).

jacob-carlborg: > Is there an existing trait that can check for that? Or can you use __traits(compiles,...) to achieve the desired result?

I guess `is(T : IUnknown)` can be used.

src/object.d
```diff
@@ -38,6 +38,67 @@ alias dstring = immutable(dchar)[];
...
+
+// Same as above, for item with non-zero initializer.
+T* ___d_newitemT(T)()
+{
+    import core.stdc.string;
+    import core.internal.traits : Unqual;
+    auto p = (() @trusted => ___d_newitemU!T())();
+    static if (__traits(isZeroInit, T))
+    {
+        () @trusted { memset(cast(Unqual!T *) p, 0, T.sizeof); }();
```
n8sh: `cast(void*)`, no need for `Unqual!T*`.

n8sh:
>Once DMD is modified to call the template correctly, the implementation of the template can be improved

I agree with kinke's comment that this PR as it stands is a pessimization. Making it not worse than what it is replacing should be done before it enters use, IMO.

PR #864, opened 2014-07-04
Fix Issue 2834 - The GC will now call destructors on heap allocated structs

Orvid:
~~This PR is not quite ready to merge, as I need to add a set  of tests for it, and also need to get it to work on arrays of structures as well. I am creating it now to make sure that I've not gone and broken druntime terribly with my approach to fixing the issue, as well as to take feedback on my method of solving the issue.~~ And the link to the issue this solves is [here](https://issues.dlang.org/show_bug.cgi?id=2834).

In the case where a struct is a member of a class, the struct's destructor will be called after the body of the class's destructor.

This PR is now ready to be reviewed and hopefully merged. It is not feasible to implement calling destructors of structs in heap allocated arrays for various reasons, including the inability to know which elements of the array are actually initialized, as well as which elements are valid.

This requires D-Programming-Language/dmd#3727

https://issues.dlang.org/show_bug.cgi?id=2834

PR #2509, opened 2019-03-14
Remove callStructDtorsDuringGC rt flag

edi33416:
Struct dtors should always be called; we shouldn't allow configuration files to change the meaning of the language.

Edit: Anyone knows when and why was this introduced? Any projects using it?

### DIFF
--- a/src/rt/lifetime.d
+++ b/src/rt/lifetime.d
@@ -39,17 +39,8 @@ private
     }
 }
 
-private immutable bool callStructDtorsDuringGC;
-
 extern (C) void lifetime_init()
 {
-    // this is run before static ctors, so it is safe to modify immutables
-    import rt.config;
-    string s = rt_configOption("callStructDtorsDuringGC");
-    if (s != null)
-        cast() callStructDtorsDuringGC = s[0] == '1' || s[0] == 'y' || s[0] == 'Y';
-    else
-        cast() callStructDtorsDuringGC = true;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


thewilsonator:
This i separate from the not calling the GC at all on shutdown, right?

edi33416:
> This i separate from the not calling the GC at all on shutdown, right?

As far as I can tell, this is only used by `structTypeInfoSize` to determine if `_d_newitem` needs to store a ptr to the `dtor`. I don't think this is related with the GC shutdown.

andralex:
I wonder how to deprecate this - issue a message at runtime to stderr if the flag is false?

Geod24:
> Edit: Anyone knows when and why was this introduced? Any projects using it?

A long time ago, the GC would not call destructors on heap allocated struct. When it was fixed, I think this was put as a transitional step. https://github.com/dlang/druntime/pull/864

src/rt/lifetime.d
```diff
@@ -2590,7 +2578,6 @@ unittest
     delete arr1;
     assert(dtorCount == 7);
 
-    if (callStructDtorsDuringGC)
```
rainers: Please remove scoping/indentation, too. Later it will just confuse people why it is there sometimes, and sometimes not.

jacob-carlborg:
> I wonder how to deprecate this - issue a message at runtime to stderr if the flag is false?

I would print the message as long as the flag has been passed to the application, regardless of the value of the flag.

jacob-carlborg:
> AFAICT it has never been documented so we might consider it internal and skip a deprecation, or make that a short one.

Are any of the druntime flags documented?

wilzbach:
> > AFAICT it has never been documented so we might consider it internal and skip a deprecation, or make that a short one.
> 
> Are any of the druntime flags documented?

No. "Luckily", druntime is chronically under-documented.
The only documentation of `--DRT` is here: https://dlang.org/spec/garbage.html#gc_config (for the GC bits).

jacob-carlborg:
I think it should be deprecated anyway. It's not difficult to do. For me it's enough for it to be deprecated for one release.

JohanEngelen:
'// this is run before static ctors, so it is safe to modify immutables'

Wow, incredible. Very nice to get rid of this kind of UB in druntime code.

dnadlinger:
> Wow, incredible. Very nice to get rid of this kind of UB in druntime code.

Immutable globals can be initialised in static constructors, so whether this is UB or not is questionable.

andralex:
@edi33416 ?

edi33416:
@andralex @jacob-carlborg I opened #2513 for the deprecation of the feature

PR #2513, opened 2019-03-19
Deprecate callStructDtorsDuringGC

edi33416:
Please see #2509 for related discussion.

### DIFF
--- a/src/rt/lifetime.d
+++ b/src/rt/lifetime.d
@@ -41,13 +41,19 @@ private
 
 private immutable bool callStructDtorsDuringGC;
 
+deprecated ("The callStructDtorsDuringGC feature will be removed in the next release cycle.")
+private pragma(inline, true) void setCallStructDtorsDuringGC(bool flag)
+{
+    cast() callStructDtorsDuringGC = flag;
+}
+
 extern (C) void lifetime_init()
 {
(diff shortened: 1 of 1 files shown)


wilzbach:
Changlog entry...

rainers:
It doesn't help to print the deprecation when compiling druntime, it must be printed when the option is used, i.e. here: https://github.com/dlang/druntime/pull/2513/files#diff-50840eb79e3292d148af41006e3af980R51

andralex:
I think this message will be printed while druntime is built, i.e. it is not client-visible at all. Its helpfulness is limited.

A better way to go about this is to detect the situation when user code is attempting to set that variable to `false`, and emit a deprecation message at that time (i.e. at run time).

andralex:
@rainers saw yours after I typed in mine...

JinShil:
> Message still printed at the wrong time (when compiling druntime rather than when it is used).

But how does one print a deprecation warning at runtime?  `fprintf(stderr, ...)`?

rainers:
> `fprintf(stderr, ...)`?

Yes.

JinShil:
https://github.com/dlang/druntime/pull/2743 was merged so I believe this can now be closed.

thewilsonator:
Please reopen if this is still required.

PR #2514, opened 2019-03-22
add parallel marking to the GC

rainers:
I can make the commits separate PRs if it is desired, but this is why events and unmanaged threads are needed.

### DIFF
--- a/src/gc/bits.d
+++ b/src/gc/bits.d
@@ -88,28 +88,33 @@ struct GCBits
     // return non-zero if bit already set
     size_t setLocked(size_t i) nothrow
     {
-        version (D_InlineAsm_X86)
+        version(Windows)
         {
-            asm @nogc nothrow {
-                naked; // assume RAX=this, [esp+4]=i
-                mov ECX, data[EAX];
-                mov EDX,[ESP+4];
-                lock;
-                bts dword ptr[ECX], EDX;
-                sbb EAX,EAX;
-                ret 4;
+            version (D_InlineAsm_X86)
+            {
+                asm @nogc nothrow {
+                    naked; // assume RAX=this, [esp+4]=i
+                    mov ECX, data[EAX];
+                    mov EDX,[ESP+4];
+                    lock;
+                    bts dword ptr[ECX], EDX;
+                    sbb EAX,EAX;
+                    ret 4;
+                }
             }
-        }
-        else version (D_InlineAsm_X86_64)
-        {
-            asm @nogc nothrow {
-                naked; // assume RCX=this, RDX=i
-                mov RAX, data[RCX];
-                lock;
-                bts qword ptr[RAX], RDX;
-                sbb RAX,RAX;
-                ret;
+            else version (D_InlineAsm_X86_64)
+            {
+                asm @nogc nothrow {
+                    naked; // assume RCX=this, RDX=i
+                    mov RAX, data[RCX];
+                    lock;
+                    bts qword ptr[RAX], RDX;
+                    sbb RAX,RAX;
+                    ret;
+                }
             }
+            else
+                static assert(false, "unexpected Windows architecture");
         }
         else
         {
--- a/src/gc/bits.d
+++ b/src/gc/bits.d
@@ -85,6 +85,63 @@ struct GCBits
         return core.bitop.btr(data, i);
     }
 
+    // return non-zero if bit already set
+    size_t setLocked(size_t i) nothrow
+    {
+        version (D_InlineAsm_X86)
+        {
+            asm @nogc nothrow {
+                naked; // assume RAX=this, [esp+4]=i
+                mov ECX, data[EAX];
+                mov EDX,[ESP+4];
+                lock;
+                bts dword ptr[ECX], EDX;
+                sbb EAX,EAX;
+                ret 4;
+            }
+        }
+        else version (D_InlineAsm_X86_64)
+        {
+            asm @nogc nothrow {
+                naked; // assume RCX=this, RDX=i
+                mov RAX, data[RCX];
+                lock;
+                bts qword ptr[RAX], RDX;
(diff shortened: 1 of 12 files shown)


thewilsonator:
I think it would be good to split out adding `@nogc` to the windows functions into a separate PR as that is trivial. Ditto for fixing the abort message in `src/core/sync/mutex.d`.

rainers:
> I think it would be good to split out adding `@nogc` to the windows functions into a separate PR as that is trivial. Ditto for fixing the abort message in `src/core/sync/mutex.d`.

Done: https://github.com/dlang/druntime/pull/2515, https://github.com/dlang/druntime/pull/2516

kinke:
> This reduces pause times for a collection considerably.

Do you have any hard numbers?

rainers:
> Do you have any hard numbers?

Unfortunately most druntime benchmarks do not use much memory so GC activity is only a fraction of the overall runtime, so it doesn't show much difference there.
With a modified gcbench/vdparser-test (more iterations and keeping every second pass in memory) the sequential version prints:
```
win32: GC summary: 1099 MB,   11 GC 1672 ms, Pauses 1366 ms <  572 ms
win64: GC summary: 1355 MB,   13 GC 1386 ms, Pauses 1044 ms <  394 ms
```
and the parallel version:
```
win32: GC summary: 1099 MB,   11 GC  676 ms, Pauses  371 ms <  145 ms
win64: GC summary: 1355 MB,   13 GC  631 ms, Pauses  289 ms <  100 ms
```
Thats on a i7-4790K (quad-core + HT) @ 4GHz.

src/gc/bits.d
```diff
@@ -85,6 +85,68 @@ struct GCBits
...
+            }
+            else version (D_InlineAsm_X86_64)
+            {
+                asm @nogc nothrow {
+                    naked; // assume RCX=this, RDX=i
+                    mov RAX, data[RCX];
+                    lock;
+                    bts qword ptr[RAX], RDX;
+                    sbb RAX,RAX;
+                    ret;
```
ibuclaw: Is it possible to not use inline asm here?

I can see that both inline and non-inline branches are semantically the same, however I'm surprised by the difference in complexity.

I would have expected instead:
```
atomicFence();
return bts(data, i);
```

rainers: > I would have expected that using mixed atomicLoad / bts / atomicFence also be doable.

No, a fence is not good enough. it must be an atomic operation, antoher thread might read/write the same word concurrently.

rainers: > Is it possible to not use inline asm here?

I'd love to use an intrinsic for `lock bts`. Does gdc have it?

rainers:
I've added WIP as there is one known scenario that I don't have a solution yet: if running as a DLL, the GC is terminated during PROCESS_DLL_DETACH, but won't be able to stop the GC threads as they block on the Windows loader lock for DllMain.

Maybe the GC should default to disable parallel marking to avoid breakage.

src/gc/bits.d
ibuclaw: I can only think of `__atomic_fetch_or`, but that's a `lock cmpxchg`.

rainers: `__atomic_fetch_or` could work, too. It's basically the same loop I implemented for the generic version.

ibuclaw: I managed to coerce a `lock bts` out of gcc.  All I needed was a `cast(size_t)`. :-)

https://explore.dgnu.org/z/G-rB79

rainers: Pretty amazing. LDC has a similar intrinsic, but doesn't seem to be able to optimize it to `lock bts`. I'll add both nonetheless.

ibuclaw: Could the static assert be removed still anyway?  No point blocking out compilers that don't implement dmd-style asm.

Why isn't the asm good enough for the posix targets?

rainers:
Rebased on top of #2533 to avoid the freeze in the Windows DLL tests.

src/gc/bits.d
rainers: > Could the static assert be removed still anyway?

You mean for GDC on Windows? I've added a version for the GDC builtin now.

> Why isn't the asm good enough for the posix targets?

Calling convention. I didn't want to go through all the possible architectures for naked assembly. But it's probably ok to just use a regular function. The profiler shows the atomic operation is the culprit, anything else in that function is almost never sampled.

rainers:
Rebased on top of https://github.com/dlang/druntime/pull/2570 and added workaround for missing module dependency (core.cpuid not initialized yet).

thewilsonator:
I this good to go?

rainers:
> I this good to go?

I think so, but I'm not 100% sure if parallel marking should be on by default. There is a chance that some programs that use little memory might get a little slower, mainly due to the possibly only (and dumb) default collection at the very end that might start threads and then has to wait for these to terminate. Any opinions?

thewilsonator:
I suggest making sure that the feature is sufficiently advertised, have it on by default for a release and see what the reaction is. 

Having it on by default would mean any kinks would get reported quickly.

The final collection is configurable too, right? I guess the user can adapt the configuration to suit their needs. Does the GC know if the collection it is doing is to shutdown? It shouldn't be too hard to not spawn a bunch of thread if that is the case. We can fix that later if it is needed.

rainers:
> The final collection is configurable too, right?

Yes, it is --DRT-gcopt=cleanup:none.

> Does the GC know if the collection it is doing is to shutdown?

No, only during DLL shutdown ATM. Thread creation fails in that case anyway.


> Having it on by default would mean any kinks would get reported quickly.

Good. I'll make a PR for the documentation...

thewilsonator:
> Good. I'll make a PR for the documentation...

Thanks.

PR #2517, opened 2019-03-22
add core.sync.event to be used as Windows events

rainers:
extracted from https://github.com/dlang/druntime/pull/2514

### DIFF
--- a/changelog/core_sync_event.dd
+++ b/changelog/core_sync_event.dd
@@ -2,32 +2,3 @@ New module core.sync.event
 
 This module provides a cross-platform interface for lightweight signaling of other threads.
 It can be used to start execution of multiple waiting threads simultaneously.
----
-import core.sync.event, core.thread, std.file;
-
-struct ProcessFile
-{
-    ThreadGroup group;
-    Event event;
-    void[] buffer;
-
-    void doProcess()
-    {
-        event.wait();
-        // process buffer
-    }
-
-    void process(string filename)
-    {
-        event.initialize(true, false);
-        group = new ThreadGroup;
-        for (int i = 0; i < 10; ++i)
-            group.create(&doProcess);
-
-        buffer = std.file.read(filename);
-        event.set();
-        group.joinAll();
-        event.terminate();
-    }
-}
----
--- a/changelog/core_sync_event.dd
+++ b/changelog/core_sync_event.dd
@@ -1,0 +1,33 @@
+New module core.sync.event
+
+This module provides a cross-platform interface for lightweight signaling of other threads.
+It can be used to start execution of multiple waiting threads simultaneously.
+---
+import core.sync.event, core.thread, std.file;
+
+struct ProcessFile
+{
+    ThreadGroup group;
+    Event event;
+    void[] buffer;
+
+    void doProcess()
+    {
+        event.wait();
+        // process buffer
+    }
+
+    void process(string filename)
+    {
+        event.initialize(true, false);
+        group = new ThreadGroup;
+        for (int i = 0; i < 10; ++i)
+            group.create(&doProcess);
+
+        buffer = std.file.read(filename);
+        event.set();
+        group.joinAll();
+        event.terminate();
+    }
+}
+---
--- a/src/core/sync/event.d
+++ b/src/core/sync/event.d
@@ -40,7 +40,6 @@ import core.internal.abort : abort;
  */
 struct Event
 {
-    // Posix version inspired by http://www.it.uu.se/katalog/larme597/win32eoposix
 nothrow @nogc:
     /**
      * Creates an event object.
@@ -74,8 +73,10 @@ nothrow @nogc:
         {
             if (m_initalized)
                 return;
-            !pthread_mutex_init(cast(pthread_mutex_t*) &m_mutex, null) ||
+            pthread_mutex_init(cast(pthread_mutex_t*) &m_mutex, null) == 0 ||
                 abort("Error: pthread_mutex_init failed.");
+            pthread_cond_init(&m_cond, null) == 0 ||
+                abort("Error: pthread_cond_init failed.");
             m_state = initialState;
             m_manualReset = manualReset;
             m_initalized = true;
@@ -107,9 +108,10 @@ nothrow @nogc:
         {
             if (m_initalized)
             {
-                assert(!m_waiter);
                 pthread_mutex_destroy(&m_mutex) == 0 ||
                     abort("Error: pthread_mutex_destroy failed.");
+                pthread_cond_destroy(&m_cond) == 0 ||
+                    abort("Error: pthread_cond_destroy failed.");
                 m_initalized = false;
             }
         }
@@ -130,12 +132,7 @@ nothrow @nogc:
             {
                 pthread_mutex_lock(&m_mutex);
                 m_state = true;
-                for (auto waiter = m_waiter; waiter != null; waiter = waiter.next)
-                {
-                    pthread_mutex_lock(&waiter.mutex);
-                    pthread_cond_signal(&waiter.cond);
-                    pthread_mutex_unlock(&waiter.mutex);
-                }
+                pthread_cond_broadcast(&m_cond);
                 pthread_mutex_unlock(&m_mutex);
             }
         }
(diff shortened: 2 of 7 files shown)


src/core/sync/event.d
```diff
@@ -0,0 +1,318 @@
...
+ *    $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors: Rainer Schuetze
+ * Source:    $(DRUNTIMESRC core/sync/event.d)
+ */
+module core.sync.event;
+
+version (Windows)
+{
+    import core.sys.windows.basetsd /+: HANDLE +/;
```
jacob-carlborg: Why these comments?

src/core/sync/event.d
```diff
@@ -0,0 +1,318 @@
...
+        {
+            pthread_mutex_lock(&m_mutex);
+            m_state = true;
+            for (auto waiter = m_waiter; waiter != null; waiter = waiter.next)
+            {
+                pthread_mutex_lock(&waiter.mutex);
+                pthread_cond_signal(&waiter.cond);
+                pthread_mutex_unlock(&waiter.mutex);
+            }
+            pthread_mutex_unlock(&m_mutex);
```
jacob-carlborg: `scope(exit)`?

src/core/sync/event.d
```diff
@@ -0,0 +1,318 @@
...
+            assert(!m_waiter);
+            pthread_mutex_destroy(&m_mutex) == 0 ||
+                abort("Error: pthread_mutex_destroy failed.");
+        }
+    }
+
+
+    /**
+     * Set the event to "signaled", so that waiting clients are resumed
+     */
```
jacob-carlborg: Using `///` would save two lines.

src/core/sync/event.d
```diff
@@ -0,0 +1,318 @@
+/**
+ * The event module provides a primitive for lightweight signalling of other threads
+ * (aka Windows Events)
```
jacob-carlborg: Reading this, I though the implementation was for Windows only.

src/core/sync/event.d
```diff
@@ -0,0 +1,318 @@
...
+else
+{
+    static assert(false, "Platform not supported");
+}
+
+import core.time;
+import core.internal.abort : abort;
+
+/**
+ * This class represents an event. Clients of an event are suspended while waiting
```
jacob-carlborg: This is a struct.

src/core/sync/event.d
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    /**
+     * Initializes an event object.
+     *
+     * Params:
+     *  manualReset  = the state of the event is not reset automatically after resuming waiting clients
+     *  initialState = initial state of the signal
+     */
+    void initialize(bool manualReset, bool initialState)
```
jacob-carlborg: Why is this a separate method?

src/core/sync/event.d
rainers: This follows the example of https://github.com/dlang/druntime/pull/2400/files (selective imports create a public alias).

src/core/sync/event.d
rainers: The function is nothrow anyway, so it doesn't make a difference in  code generation. With a short function like this, having the symmetry show up in the function is slightly better IMO.

src/core/sync/event.d
rainers: Ok.

src/core/sync/event.d
rainers: Ok. It started out as a class, but no longer.

src/core/sync/event.d
rainers: You cannot pass arguments to the constructor if it is part of another struct, e.g. https://github.com/dlang/druntime/pull/2514/files#diff-6615611ba51118215a271e3500c61122R2605. You have to use an explicit `initialize`-call then. `__ctor` could be an alternative, but I think it should not be part of the API.

src/core/sync/event.d
wilzbach: Hasn't this been fixed a year ago or so?

src/core/sync/event.d
rainers: reworded slightly.

src/core/sync/event.d
wilzbach: Oh well, looks like it's really still in deprecation, but there's a PR to remove them for good:

https://github.com/dlang/dmd/pull/9393

src/core/sync/event.d
jacob-carlborg: Is there a downside to do `evStart = Event(...)` in `Gcx.initialize`?

rainers: I've explicitly added `@disable this(this)` as this could duplicate the handle reference.

jacob-carlborg: Does not seem to prevent assignment: https://run.dlang.io/is/iefkBG.

rainers: Hmm, seems like opAssign should be disabled, too.

jacob-carlborg: If that's required then I'm fine with using `initialize` instead of a constructor.

src/core/sync/event.d
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    @disable this(this);
+
+    ~this()
+    {
+        terminate();
+    }
+
+    void terminate()
```
jacob-carlborg: No documentation.

jacob-carlborg:
No changelog.

src/core/sync/event.d
thewilsonator: That will go in very soon.

rainers:
Added changelog and a bit more documentation. I've also added a bool to keep track of the initialization state of the posix version.

>     * uncommented selective imports

I think this should be done for all the imports once the dmd deprecation is gone (https://github.com/dlang/dmd/pull/9393 doesn't help for this case yet).

thewilsonator:
`posix.mak:348: recipe for target 'checkwhitespace' failed`

changelog/core_sync_event.dd
```diff
@@ -0,0 +1,34 @@
+New module core.sync.event
+
+This module provides a primitive for lightweight signaling of other threads
+by emulating Windows events on posix. It can be used to start execution of
```
jacob-carlborg: * Perhaps I'm picky, but now it sounds like it's for Posix only. I would write something like: "... a cross-platform interface  for lightweight signaling of other threads. "
* `posix` -> `Posix`

rainers: No problem ;-) Amended.

changelog/core_sync_event.dd
```diff
@@ -0,0 +1,33 @@
+New module core.sync.event
+
+This module provides a cross-platform interface for lightweight signaling of other threads.
+It can be used to start execution of multiple waiting threads simultaniously.
```
kinke: simultan**e**ously

src/core/sync/event.d
jacob-carlborg: @rainers this has been merged now.

rainers: Not good  enough:
```
test.d(3): Deprecation: core.sync.event.HANDLE is not visible from module test
```

jacob-carlborg: 😞

kinke:
Rainer, have you explored whether [pthread_cond_broadcast](http://pubs.opengroup.org/onlinepubs/009696699/functions/pthread_cond_broadcast.html) and a single condition variable could work too?

rainers:
> Rainer, have you explored whether [pthread_cond_broadcast](http://pubs.opengroup.org/onlinepubs/009696699/functions/pthread_cond_broadcast.html) and a single condition variable could work too?

Didn't know about it, now replaced the manual implementation.

src/core/sync/event.d
```diff
@@ -0,0 +1,301 @@
...
+    {
+        pthread_mutex_t m_mutex;
+        pthread_cond_t m_cond;
+        bool m_initalized;
+        bool m_state;
+        bool m_manualReset;
+    }
+}
+
+// Test single-thread (non-shared) use.
```
wilzbach: Maybe make this example public, s.t. there's at least one user-facing example.

rainers: I think this example is a pretty bad one as it doesn't involve multiple threads. Maybe better to move the example from the changelog into the documentation?

wilzbach: Sure. I just saw that the module is currently without any public example.

src/core/sync/event.d
```diff
@@ -37,6 +37,35 @@ import core.internal.abort : abort;
...
+        for (int i = 0; i < 10; ++i)
+            group.create(&doProcess);
+
+        buffer = std.file.read(filename);
+        event.set();
+        group.joinAll();
+        event.terminate();
+    }
+}
+---
```
Geod24: Why isn't it a documented unittest instead ?

rainers: I think its already a bit too long for a really good example, and making it actually runnable would make it even longer.

PR #2518, opened 2019-03-22
[stable] Get rid of superfluous 2nd global variable in object.RTInfoImpl

kinke:


kinke:
The 2nd commit fails because `__traits(getPointerBitmap, Object)` returns `[N, 0]` (as for any empty class), i.e., no pointers, which doesn't agree with `typeid(Object).flags & 1` (which is 1, i.e., has pointers).

The 1st slot of the class payload, TypeInfo+vtbl, should be of no interest to the GC; I'm not sure about the 2nd slot (monitor).

ZombineDev:
`core.sync.mutex.Mutex` has a constructor which attaches the newly created `Mutex` to an existing object (an `extern (D)` class, derived from `object.Object`). Which means that the monitor field should be checked for GC references.

rainers:
> The 2nd commit fails because `__traits(getPointerBitmap, Object)` returns `[N, 0]` (as for any empty class), i.e., no pointers, which doesn't agree with `typeid(Object).flags & 1` (which is 1, i.e., has pointers).

ClassFlags.noPointers is 2, not 1. dmd ignores the monitor field when determining this, see https://github.com/dlang/dmd/blob/master/src/dmd/toobj.d#L1310. Maybe this is different in LDC?

rainers:
> `core.sync.mutex.Mutex` has a constructor which attaches the newly created `Mutex` to an existing object (an `extern (D)` class instance, derived from `object.Object`). Which means that the monitor field should be checked for GC references.

Is this the crash I've just seen [here](https://circleci.com/gh/dlang/druntime/3979#tests/containers/1)? Given that it is not considered for "having pointers", that seems like abuse (normal monitors are manually managed).

src/object.d
```diff
@@ -4156,15 +4156,19 @@ void __ctfeWrite(scope const(char)[] s) @nogc @safe pure nothrow {}
...
-    immutable RTInfoImpl = data.ptr;
+    immutable size_t[pointerBitmap.length] RTInfoImpl = pointerBitmap[];
 }
 
 template RTInfo(T)
 {
-    enum RTInfo = RTInfoImpl!(__traits(getPointerBitmap, T));
+    enum pointerBitmap = __traits(getPointerBitmap, T);
+    enum size_t[pointerBitmap.length - 1] allZeros = 0;
+    static if (pointerBitmap[1 .. $] == allZeros)
```
rainers: I also considered this optimization, but decided against for simplicity of implementation, with less impact on compile times at the cost of a few more bytes emitted. There are not so many likely instantiations of this, though there can also be some extreme examples of gratuitious output.
Not sure if it has a notable impact, but would be good to not create many additional symbols `allZeroes`, e.g. by making it a seperate template.

src/object.d
```diff
@@ -4156,15 +4156,19 @@ void __ctfeWrite(scope const(char)[] s) @nogc @safe pure nothrow {}
  * Create RTInfo for type T
  */
 
-template RTInfoImpl(size_t[] pointers)
+template RTInfoImpl(size_t[] pointerBitmap)
 {
-    immutable size_t[pointers.length] data = pointers[];
-    immutable RTInfoImpl = data.ptr;
+    immutable size_t[pointerBitmap.length] RTInfoImpl = pointerBitmap[];
```
rainers: I don't remember exactly, but there might have been some troubles with this version when first implemented 2015. Looks good now.

kinke:
> ClassFlags.noPointers is 2, not 1.

The AA implementation doesn't [use](https://github.com/dlang/druntime/blob/bd0f9111530cf878b09bd782471ea745a9e25e7b/src/rt/aaA.d#L63-L64) these `ClassFlags` though, but `TypeInfo.flags()` ([here](https://github.com/dlang/druntime/blob/bd0f9111530cf878b09bd782471ea745a9e25e7b/src/object.d#L1283-L1285)), where a `1` means 'GC should scan for pointers'. `TypeInfo_Class` overrides it and always returns 1 ([here](https://github.com/dlang/druntime/blob/bd0f9111530cf878b09bd782471ea745a9e25e7b/src/object.d#L1920)].

> I also considered this optimization, but decided against for simplicity of implementation, with less impact on compile times at the cost of a few more bytes emitted.

I thought the assumption was that the compiler somehow magically skips it for appropriate types already, according to the comment [here](https://github.com/dlang/druntime/blob/bd0f9111530cf878b09bd782471ea745a9e25e7b/src/object.d#L4170-L4173). If you think it may have a noticeable effect on compilation speed, I can move the 2nd commit to another PR targeting master.

rainers:
> `TypeInfo_Class` overrides it and always returns 1:

This is the missing separation between class reference and class instance at work. TypeInfo_Class is a bad mixture of both. The noPointer flag is used in _d_newclass instead.

> I thought the assumption was that the compiler somehow magically skips it for appropriate types already, according to the comment:

The compiler actually only emits these values if it fails to analyze RTInfo properly, see https://github.com/dlang/dmd/pull/2480. They are used by the type infos predefined in the runtime, though. Using rtinfoNoPointers usually does not have a real effect on the precise GC as the BlkAttr.NOSCAN flag is passed to malloc aswell.

rainers:
> If you think it may have a noticeable effect on compilation speed, I can move the 2nd commit to another PR targeting master.

Yes, maybe better.

kinke:
> This is the missing separation between class reference and class instance at work.

Ah, understood. So the unittest in `rt/aaA.d` is to be fixed.

I already removed the 2nd commit; I'll open a PR for that as soon as this is merged into master (as long as I don't forget about it ;)).

rainers:
> So the unittest in `rt/aaA.d` is to be fixed.

I guess supporting a mixture of rtinfo-magics and actual pointerBitmaps here?
https://github.com/dlang/druntime/blob/bd0f9111530cf878b09bd782471ea745a9e25e7b/src/rt/aaA.d#L389
Sounds good.

kinke:
Yeah, accounting for the AA's `hasPointers` flag always being set if key and/or value are objects, independent from their pointer bitmaps.

kinke:
I guess I'll have to dummy-amend to make buildkite retrigger and auto-merge work...

wilzbach:
> retriggering manually with the dummy account works, nice.

You can also send me an email with your address if you want to be added to the Buildkite team as now that buildkite has public logs we might remove that dummy account soon.

kinke:
> now that buildkite has public logs

Yeah, much appreciated.

> if you want to be added to the Buildkite team

Thx for the offer, I might do so if I need it again - this was the first time I needed it (and only because I pushed the previous 2nd commit before the 1st one was fully tested).

PR #2519, opened 2019-03-24
add lowlevel thread support

rainers:
another extract of https://github.com/dlang/druntime/pull/2514

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -196,7 +196,7 @@ version (Windows)
         import core.sys.windows.winnt /+: CONTEXT, CONTEXT_CONTROL, CONTEXT_INTEGER+/;
 
         extern (Windows) alias btex_fptr = uint function(void*);
-        extern (C) uintptr_t _beginthreadex(void*, uint, btex_fptr, void*, uint, uint*) nothrow;
+        extern (C) uintptr_t _beginthreadex(void*, uint, btex_fptr, void*, uint, uint*) nothrow @nogc;
 
         //
         // Entry point for Windows threads
@@ -2020,7 +2020,9 @@ extern (C) void thread_init() @nogc
     //       exist to be scanned at this point, it is sufficient for these
     //       functions to detect the condition and return immediately.
 
+    initLowlevelThreads();
     Thread.initLocks();
+
     // The Android VM runtime intercepts SIGUSR1 and apparently doesn't allow
     // its signal handler to run, so swap the two signals on Android, since
     // thread_resumeHandler does nothing.
@@ -2116,6 +2118,7 @@ extern (C) void thread_term() @nogc
         Thread.pAboutToStart = null;
     }
     Thread.termLocks();
+    termLowlevelThreads();
 }
 
 
@@ -3070,7 +3073,7 @@ do
 * Throws:
 *  ThreadError.
 */
-private void onThreadError(string msg) nothrow
+private void onThreadError(string msg) nothrow @nogc
 {
     __gshared ThreadError error = new ThreadError(null);
     error.msg = msg;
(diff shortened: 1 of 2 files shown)


CyberShadow:
Could the high-level thread code be refactored on top of the low-level code added here?

rainers:
> Could the high-level thread code be refactored on top of the low-level code added here?

Possible, but with a couple of complications, because some high level variables are changed inbetween (e.g. m_hndl, m_isRunning). That would probably need a greater refactoring.

CyberShadow:
If the symbols added here are to be public, then I think that would make this a situation that @andralex asked to be notified about.

rainers:
I wouldn't mind keeping the symbols "private", i.e. undocumented, but I want to call them from the GC.

jacob-carlborg:
> I wouldn't mind keeping the symbols "private", i.e. undocumented, but I want to call them from the GC.

The GC is not in the `core` top level package?

rainers:
> The GC is not in the `core` top level package?

No, it has its own top level package `gc`.

thewilsonator:
This needs another patch to weka-io/mecca

thewilsonator:
That patch for weka is blocked on https://issues.dlang.org/show_bug.cgi?id=19761

src/core/thread.d
```diff
@@ -1716,7 +1716,12 @@ private:
         return cast(Mutex)_locks[1].ptr;
     }
 
-    __gshared align(Mutex.alignof) void[__traits(classInstanceSize, Mutex)][2] _locks;
+    @property static Mutex lowLevelThreadslock() nothrow @nogc
+    {
+        return cast(Mutex)_locks[2].ptr;
+    }
+
+    __gshared align(Mutex.alignof) void[__traits(classInstanceSize, Mutex)][3] _locks;
```
ibuclaw: Can this be its own private field to not interfere with #2411 (accessing this array causes crash with SPARC, SystemZ, PA-RISC, or any Solaris platforms).

ibuclaw: From what I can tell from other comments, you will also avoid being blocked by weka too.

rainers: Thanks for the hint, now using a separate lock.

src/core/thread.d
```diff
@@ -5628,3 +5631,187 @@ version (Windows)
...
+
+    void initLowlevelThreads() nothrow @nogc
+    {
+        ll_lock[] = typeid(Mutex).initializer[];
+        lowlevelLock.__ctor();
+    }
+
+    void termLowlevelThreads() nothrow @nogc
+    {
+        lowlevelLock.__dtor();
```
ibuclaw:
```
core/thread.d(5657): Error: destructor `core.sync.mutex.Mutex.~this` is not `nothrow`
core/thread.d(5655): Error: `nothrow` function `core.thread.termLowlevelThreads` may throw
```

thewilsonator:
Why does CircleCI fail to install DMD?

thewilsonator:
This otherwise looks go to me but CircleCI is required.

src/core/thread.d
```diff
@@ -5628,3 +5631,187 @@ version (Windows)
...
     alias ThreadID = pthread_t;
+
+///////////////////////////////////////////////////////////////////////////////
+// lowlovel threading support
+private
+{
+    __gshared size_t ll_nThreads;
+    __gshared ThreadID* ll_pThreads;
+
+    __gshared align(Mutex.alignof) void[__traits(classInstanceSize, Mutex)] ll_lock;
```
ibuclaw: Note: this is alignment is insufficient for targets with strict alignment requirements.

rainers: Yes, will need to be updated in lock-step with Thread._locks.

PR #2524, opened 2019-03-26
GC: optimize alloc of small object by keeping track of unused pages

rainers:
This simple change is what yields the larger time difference of https://github.com/dlang/druntime/pull/2520#issue-263920143, i.e. reduces execution time of the vdparser-test from approx. 10.5 s to 7.5 s.

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -1300,10 +1300,8 @@ struct Gcx
             for (size_t p = 0; p < pooltable.length; p++)
                 if (pooltable.pools[p].isLargeObject)
                     (cast(LargeObjectPool*)(pooltable.pools[p])).Invariant();
-
-            for (size_t p = 0; p < pooltable.length; p++)
-                if (pooltable.pools[p].isLargeObject)
-                    (cast(LargeObjectPool*)(pooltable.pools[p])).Invariant();
+                else
+                    (cast(SmallObjectPool*)(pooltable.pools[p])).Invariant();
 
             if (!inCollection)
                 (cast()rangesLock).lock();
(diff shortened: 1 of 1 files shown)


src/gc/impl/conservative/gc.d
```diff
@@ -2538,6 +2538,10 @@ struct Pool
     //  only for the first and the last page.
     uint* bPageOffsets;
 
+    // The small object pool uses the same array to keep a chain of pages with the same
+    // bin size and free pages (searchStart is first free page)
```
rainers: This comment is a bit premature, but keeping track of pages with the same bin size will be needed for deferring incremental `recover` into later allocations.

kinke:
Rainer on fire ;) - the GC improvements are impressive (and at the same time kinda shocking that so much potential hasn't been exploited yet).

ZombineDev:
@rainers rebasing to master should fix the CircleCI build.

rainers:
> reduces execution time of the vdparser-test from approx. 10.5 s to 7.5 s.

BTW: time drops to about 5 seconds when compiling with LDC!

kinke:
Is that so surprising? 30-40% runtime reduction with LDC seems like a rough rule of thumb for general code (probably more for number-crunching code), e.g., DMD itself.

rainers:
Not really surprising, though profiling suggested that marking is pretty much memory bandwidth limited (most of the samples were taken at the instruction reading the scanned memory). LDC compiled marking code was better similar to the overall ratio which hints that there is still more leeway for improvement.

kinke:
[@rainers: Just as a reminder, `-disable-gc2stack` may be useful for comparisons with LDC.]

rainers:
> [@rainers: Just as a reminder, `-disable-gc2stack` may be useful for comparisons with LDC.]

Doesn't make a notable difference. The parser doesn't create many temporary objects AFAICT.

PR #2525, opened 2019-03-26
Windows: not all import folders created

rainers:
Noticed in some logs, e.g. https://ci.appveyor.com/project/greenify/druntime/builds/23381487#L163

Not sure why the copy failure doesn't cause the build to fail.


ZombineDev:
Auto-merge toggled on

PR #2526, opened 2019-03-27
[CircleCi] Fix build by using Ubuntu 18.04 as base

wilzbach:


wilzbach:
Similar as in https://github.com/dlang/dmd/pull/9499, I had to disable the 32-bit build due to https://issues.dlang.org/show_bug.cgi?id=19116.
Moreover, I had to disable building DMD with `debug` because that's broken too :/

PR #2528, opened 2019-03-27
Add missing cast(string) in _d_assert_fail

nordlow:
See https://forum.dlang.org/post/eijbfurdbbnhvepfaatb@forum.dlang.org

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -4842,5 +4842,5 @@ string _d_assert_fail(string comp, A, B)(A a, B b) @nogc @safe nothrow pure
     n += token.length;
     buffer[n++] = ' ';
     buffer[n .. n + valB.length] = valB;
-    return buffer;
+    return (() @trusted => cast(string) buffer)();
 }


src/object.d
```diff
@@ -4842,5 +4842,5 @@ string _d_assert_fail(string comp, A, B)(A a, B b) @nogc @safe nothrow pure
     n += token.length;
     buffer[n++] = ' ';
     buffer[n .. n + valB.length] = valB;
-    return buffer;
+    return cast(string)buffer;
```
MoonlightSentinel: Local trusted?

```suggestion
    return (() @trusted => cast(string) buffer)();
```

kinke:
Please target stable, so that this makes it into 2.085.1.

nordlow:
Sorry for asking but...how do I target stable `instead` of `master`?

src/object.d
nordlow: Done.

wilzbach:
Changed the target branch for you on GitHub. However, you still need to run sth. like this to update your branch:

```
git rebase --onto upstream/stable upstream/master
```

(then the other commits will disappear).
I'm currently travelling. Sorry.

nordlow:
Done, @wilzbach :)

wilzbach:
No tests??

PR #2530, opened 2019-03-30
GC: keep list of small pages to recover, defer into malloc

rainers:
This removes recovering the free lists from the actual garbage collection and does it lazily during allocation. The advantage is a shorter collection time in the allocating thread (the world isn't stopped for the recovering phase) and memory only being written to when being reused anyway after allocation. Benchmarks don't show a notable difference either way, though.

I've also added the extended vdparser test I've been using for benchmarking larger memory usage.

### DIFF
--- a/benchmark/gcbench/vdparser.d
+++ b/benchmark/gcbench/vdparser.d
@@ -11,24 +11,40 @@ FalsePointers* ptrs;
 
 void main(string[] argv)
 {
-    size_t nIter = 20;
+    version(EXTENDEND)
+        size_t nIter = 100;
+    else
+        size_t nIter = 20;
     if(argv.length > 2)
         nIter = to!size_t(argv[2]);
 
-    // create some random data as false pointer simulation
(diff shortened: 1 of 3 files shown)


thewilsonator:
Need to rebase to get rid of circle.

rainers:
> Need to rebase to get rid of circle.

That's the style checker complaining. BTW: any opinions on whether having this behind "version(DEFER_RECOVER)" makes sense? My original version also called the finalizers lazily, but with this out of the way, there should not be an observable effect. I'd be in favor of removing the version.

This version might be a bit slower in `Gc.stats()`, but that should not be a critical function in an application to begin with.

rainers:
>  I'd be in favor of removing the version.

Done.
 
> This version might be a bit slower in `Gc.stats()`,

Oops, that code was not pudhed.

thewilsonator:
src\core\thread.d(5777): Error: undefined identifier `_thread_dll_process_detaching`

PR #2531, opened 2019-03-30
Fix object.__ArrayCast error message generation so it works with CTFE

JinShil:
This is a followup to https://github.com/dlang/druntime/pull/2264

It is intended to resolve an error currently being encountered on the DMD side in https://github.com/dlang/dmd/pull/9516

```
Error: `alloca` cannot be interpreted at compile time, because it has no available source code
```

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -4735,31 +4735,38 @@ Params:
 private void onArrayCastError()(string fromType, size_t fromSize, string toType, size_t toSize) @trusted
 {
     import core.internal.string : unsignedToTempString;
-    import core.stdc.stdlib : alloca;
 
-    const(char)[][8] msgComponents =
+    const(char)[][9] msgComponents =
     [
-        "Cannot cast `"
-        , fromType
-        , "` to `"
-        , toType
-        , "`; an array of size "
+        "An array of size "
         , unsignedToTempString(fromSize)
         , " does not align on an array of size "
         , unsignedToTempString(toSize)
+        , ", so `"
+        , fromType
+        , "` cannot be cast to `"
+        , toType
+        , "`"
     ];
 
     // convert discontiguous `msgComponents` to contiguous string on the stack
-    size_t length = 0;
-    foreach (m ; msgComponents)
-        length += m.length;
-
-    auto msg = (cast(char*)alloca(length))[0 .. length];
+    enum msgLength = 2048;
+    char[msgLength] msg;
 
     size_t index = 0;
-    foreach (m ; msgComponents)
+    foreach (m; msgComponents)
+    {
         foreach (c; m)
+        {
             msg[index++] = c;
+            if (index >= (msgLength - 1))
+                break;
+        }
+
+        if (index >= (msgLength - 1))
+            break;
+    }
+    msg[index] = '\0'; // null-termination
 
     // first argument must evaluate to `false` at compile-time to maintain memory safety in release builds
     assert(false, msg);


src/object.d
```diff
@@ -4750,16 +4749,22 @@ private void onArrayCastError()(string fromType, size_t fromSize, string toType,
     ];
 
     // convert discontiguous `msgComponents` to contiguous string on the stack
-    size_t length = 0;
-    foreach (m ; msgComponents)
-        length += m.length;
-
-    auto msg = (cast(char*)alloca(length))[0 .. length];
```
thewilsonator: Can't you just do 
```d
char[] msg;
if (__ctfe)
    msg = new char[length];
else
    msg = (cast(char*)alloca(length))[0 .. length];
```

or are you trying to get rid of the `alloca`?

JinShil: hmm.  I wonder if that would work.  Would `length` be determined at compile-time?

JinShil: I am trying to get rid of `alloca` in a way.  This was a response I received from @WalterBright  in a private e-mail exchange when I was trying to figure out how to create error messages dynamically in pure code.  

Walter's response to me.
> I understand the need to use alloca() to create error messages purely. But
there's another way:
>
> Use char[100]tmp; and simply truncate the message on overflow. It's not perfect,
one can't have a 5000 character error message, but it works for now, and gets
you unblocked.

I chose 2048 arbitrarily to ensure there was plenty of room for even the largest error message with templated types.

JinShil: Getting rid of `alloca` would also permit this to work in -betterC.  Currently `alloca` is compiled into druntime, so it's not linked in when compiling with -betterC.

thewilsonator: That seems fine.

>I chose 2048 arbitrarily to ensure there was plenty of room for even the largest error message with templated types.

In which case please reorder the message s.t. the array lengths are given first as they are bounded in length.

JinShil: Sorry, but I don't understand.  Reordering the strings would result in a grammatically incorrect sentence.

thewilsonator: Cannot cast array:  size 42, does not align on an array of size 21, of type foo, to type bar

JinShil: I went with "An array of size 42 does not align on an array of size 21, so `foo` cannot be cast to `bar`.  I hope that will work.

thewilsonator: Nice!

jacob-carlborg:
`alloca` cannot be made into a template to have the source code available. Or won’t that help?

thewilsonator:
Its a compiler intrinsic most of the time, I've no idea why or how its part of druntime.

jacob-carlborg:
https://github.com/dlang/druntime/blob/master/src/rt/alloca.d

I guess inline assembly doesn’t work at compile time.

JinShil:
See https://github.com/dlang/druntime/blob/fbc76f7dc8282f8933bffddb9cb51fc63ad27b00/src/rt/alloca.d#L47

It's a magic function, given special treatment by the compiler.  It is unlike any other and requires compiler assistance to work properly.

JinShil:
Test suite is printing something like this in DMD...

```
/home/mike/github/JinShil/druntime/import/object.d(4772): Error: `['A', 'n', ' ', 'a', 'r', 'r', 'a', 'y', ' ', 'o', 'f', ' ', 's', 'i', 'z', 'e', ' ', '8', ' ', 'd', 'o', 'e', 's', ' ', 'n', 'o', 't', ' ', 'a', 'l', 'i', 'g', 'n', ' ', 'o', 'n', ' ', 'a', 'n', ' ', 'a', 'r', 'r', 'a', 'y', ' ', 'o', 'f', ' ', 's', 'i', 'z', 'e', ' ', '0', ',', ' ', 's', 'o', ' ', '`', 't', 'y', 'p', 'e', 'o', 'f', '(', 'n', 'u', 'l', 'l', ')', '`', ' ', 'c', 'a', 'n', 'n', 'o', 't', ' ', 'b', 'e', ' ', 'c', 'a', 's', 't', ' ', 't', 'o', ' ', '`', 's', 't', 'r', 'i', 'n', 'g', '`', '\x00', '\xff', '\xff', '\xff',
```
It seems to work find at run.dlang.org:  https://run.dlang.io/is/5a6tje

How do I get it to render as a normal string.  Beware, we are already casting an array, when we're in this function, so I don't thing `cast(string)msg` would be wise.

jacob-carlborg:
I’ve seen the same thing with failing static asserts

JinShil:
Yep, looks like CTFE string rendering has a problem:  https://run.dlang.io/  (Sigh, I guess I need to figure out how to fix that too).

jacob-carlborg:
Regression?

JinShil:
I think this is ready to go, but I think the DMD side requires some more work.  At least it will allow me to continue with the DMD PR.  If additional fixes are needed, I can submit them in followup PRs.

PR #2533, opened 2019-03-31
allow lowlevel thread creation to fail

rainers:
joinLowLevelThread: terminate thread when called from DllMain on windows

Addition to https://github.com/dlang/druntime/pull/2519 necessary for DLL support in https://github.com/dlang/druntime/pull/2514

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -5676,6 +5676,9 @@ private
     }
 }
 
+version (Windows)
+    package(core) bool thread_DLLProcessDetaching;
+
 /**
  * Create a thread not under control of the runtime, i.e. TLS module constructors are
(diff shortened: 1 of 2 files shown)


src/core/thread.d
```diff
@@ -5676,6 +5676,9 @@ private
     }
 }
 
+version (Windows)
+package(core) bool _thread_dll_process_detaching;
```
jacob-carlborg: Why snake case instead of camel case and why the leading underscore?

rainers: I had this variable in core.sys.windows.dll first which uses this coding style. Now changed it to `thread_DLLProcessDetaching` following the style of `thread_attachThis`.

PR #2534, opened 2019-03-31
[dmd-cxx] Fix Issue 19481: mutex.d(95) Error: pthread_mutex_init failed

ibuclaw:
Committing the backport before it's in master... should give you a couple months to fix failing mecca in #2411 before gdc switches to ddmd/master when stage1 opens for gcc-10.

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -170,6 +170,29 @@ template anySatisfy(alias F, T...)
     }
 }
 
+// simplified from std.traits.maxAlignment
+template maxAlignment(U...)
+{
+    static if (U.length == 0)
+        static assert(0);
+    else static if (U.length == 1)
+        enum maxAlignment = U[0].alignof;
+    else static if (U.length == 2)
+        enum maxAlignment = U[0].alignof > U[1].alignof ? U[0].alignof : U[1].alignof;
+    else
+    {
+        enum a = maxAlignment!(U[0 .. ($+1)/2]);
+        enum b = maxAlignment!(U[($+1)/2 .. $]);
+        enum maxAlignment = a > b ? a : b;
+    }
+}
+
+template classInstanceAlignment(T)
+if (is(T == class))
+{
+    alias classInstanceAlignment = maxAlignment!(void*, typeof(T.tupleof));
+}
+
 // Somehow fails for non-static nested structs without support for aliases
 template hasElaborateDestructor(T...)
 {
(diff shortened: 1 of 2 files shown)


ibuclaw:
There's nothing wrong with the alignment of Mutex. If you're lucky, it's the size of N in void[N][2] that's the issue, as only the first index is suitably aligned.  (You're unlucky if the alignment of pointers is smaller than the alignment of the OS mutex type).

There is no one constant alignment either, what type of and what alignment of the mutex is varies between platforms.

ibuclaw:
Committing, if an amendment is needed before May, then so be it.

rainers:
> There's nothing wrong with the alignment of Mutex.

Mutex.alignof might be unusable here because it is the alignment of the class reference (the unlucky case):
```
class C
{
	double x;
}

struct S
{
	align(16) int x;
}

// for -m32
pragma(msg, Object.alignof); // 4
pragma(msg, C.alignof);      // 4
pragma(msg, C.x.alignof);    // 8
pragma(msg, S.alignof);      // 16
pragma(msg, S.x.alignof);    // 4
```

Agreed, the aligned size of the object is another issue.

> There is no one constant alignment either, what type of and what alignment of the mutex is varies between platforms.

I meant something like
```
version(mySpecialOS)
    enum mutexAlign = 16; // due to explicit alignment on field of mutex
else
    enum mutexAlign = classInstanceAlignment!Mutex;
```
so this declaration doesn't have to be repeated everytime something similar is done.

But probably the better solution would be for `alignof(C.x)` to return the explicit alignment for symbols, not the alignment of the type.

ibuclaw:
> Mutex.alignof might be unusable here because it is the alignment of the class reference (the unlucky case):

Yes, you can't use Mutex.alignof, which why it is being changed from.  The underlying record is suitably aligned, hence I re-iterate there is no problem with Mutex (i.e: when used as `new Mutex()`).

> I meant something like

Only OS_mutex_type.alignof is necessary (Currently: `CRITICAL_SECTION` on Windows, `pthread_mutex_t` on Posix, etc...).

I had considered something similar to core.sync.semaphore.
```
public:
    /// Aliases the operating-system-specific mutex type.
    version (Windows)        alias Handle = CRITICAL_SECTION;
    /// ditto
    else version (Posix)     alias Handle = pthread_mutex_t;

private:
    /// Handle to the system-specific mutex.
    Handle m_hndl;
```

But ultimately went for this instead over `align(Mutex.Handle.alignof) xxx`, as I'd rather not assume that the OS mutex has the greater alignment requirement of the class.

PR #2538, opened 2019-04-02
Add explicit cast in stc/core/stdcpp/array.d

RazvanN7:
Part 2 for [1] blockade. 

[1] https://github.com/dlang/dmd/pull/9505

### DIFF
--- a/src/core/stdcpp/array.d
+++ b/src/core/stdcpp/array.d
@@ -97,7 +97,7 @@ pure nothrow @nogc:
         ///
         inout(T)* data() inout @safe                    { static if (N > 0) { return &_M_elems[0]; } else { return null; } }
         ///
-        ref inout(T[N]) as_array() inout @trusted       { return data()[0 .. N]; }
+        ref inout(T)[N] as_array() inout @trusted       { return cast(inout(T)[N])data()[0 .. N]; }
         ///
         ref inout(T) at(size_type i) inout @trusted     { return data()[0 .. N][i]; }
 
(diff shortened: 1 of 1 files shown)


src/core/stdcpp/array.d
```diff
@@ -97,7 +97,7 @@ pure nothrow @nogc:
         ///
         inout(T)* data() inout @safe                    { static if (N > 0) { return &_M_elems[0]; } else { return null; } }
         ///
-        ref inout(T[N]) as_array() inout @trusted       { return data()[0 .. N]; }
+        ref inout(T)[N] as_array() inout @trusted       { return cast(inout(T)[N])data()[0 .. N]; }
```
Geod24: I feel like that cast should never be necessary. Could you explain why it is ?

RazvanN7: data returns an inout(T)* which cannot be implicitly converted to ref inout(T)[N] because you do not have a size.  Note that the function is @trusted because the compiler introduced the cast implicitly before the patch in https://github.com/dlang/dmd/pull/9505 .

jacob-carlborg: Why don't you access `_M_elems` directly, which already has the correct type?

RazvanN7: I wanted the change to be minimally invasive.

jacob-carlborg: I think it's better for it to be correct and avoiding casts.

PR #2539, opened 2019-04-02
Slight simplification in core.internal.hash

n8sh:
Reduce the number of `hashOf` templates and reduce the number of lines of code. Also now floats and doubles no longer force hash chaining (the hash code of a float is now just the float's bits, after possibly regularizing NaN or -0.0).

EDIT: The initial motivation was simplifying the code regarding floats, which had become a bit tangled due to handling of imaginary and complex types.

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -10,7 +10,7 @@
 module core.internal.hash;
 
 import core.internal.convert;
-import core.internal.traits : allSatisfy, Unconst;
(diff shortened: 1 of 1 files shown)


thewilsonator:
```
src/core/internal/hash.d:347:    else static if(__traits(isFloating, T))
posix.mak:383: recipe for target 'style_lint' failed
```

thewilsonator:
src/core/internal/hash.d(340): Error: unrecognized trait `isComplex`

PR #2541, opened 2019-04-04
Resolve deprecation warning for 32 byte vectors

JinShil:
This will remove the deprecation warning

```
src/core/simd.d(53): Deprecation: 32 byte vector types are only supported with -mcpu=avx
```
currently polluting druntime build output and facilitate advancement of the deprecation to an error.


wilzbach:
Awesome! We should soon follow with fixing this to use `-de` like we do for Phobos, so that we don't end up in this state for years again.

JinShil:
> We should soon follow with fixing this to use -de

I would if I knew immediately where to put it.

wilzbach:
https://github.com/dlang/druntime/pull/1982 (i.e. adding `-de` to DFLAGS)

PR #2548, opened 2019-04-07
Suppress object.RTInfoImpl global if aggregate contains no pointers

kinke:
Reducing bloat introduced with v2.085.


kinke:
@thewilsonator: Please wait a bit, Nicholas, and let @rainers have a look at it - he was worried about potential compile-time pessimization. I'm also a bit unsure about the AA stuff.

thewilsonator:
OK

PR #2550, opened 2019-04-09
Fix Issue 19796  -  wrong assumption of integral element type

wilzbach:
I didn't know any better way to do this, so better ideas are very welcome!


wilzbach:
> Regression or critical bug fixes should always target the stable branch. 

Regression is in `master` only.

quickfur:
Thanks!

PR #2554, opened 2019-04-13
object: Declare private struct AA and fix aaA function signatures

ibuclaw:
All AA tests currently SEGV on 32-bit Solaris/SPARC,

Closer investigation reveals object.d calls _aaValues (and _aaKeys) incorrectly:

While the implementation expects a struct AA

libdruntime/rt/aaA.d:
```
extern (C) inout(void[]) _aaValues(inout AA aa, in size_t keysz, in size_t valsz,
```

the call passes a void * instead in libdruntime/object.d:
```
inout(void)[] _aaValues(inout void* p, in size_t keysize, in size_t valuesize, const TypeInfo tiValArray) pure nothrow;
...
auto a = cast(void[])_aaValues(cast(inout(void)*)aa, Key.sizeof, Value.sizeof, typeid(Value[]));
```

The problem is that on 32-bit SPARC small structs are passed by reference, so there's a mismatch between caller and callee.

On 64-bit SPARC instead, they are passed by value, so the mismatch doesn't matter and the tests PASS.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -1695,12 +1695,12 @@ class TypeInfo_AssociativeArray : TypeInfo
 
     override bool equals(in void* p1, in void* p2) @trusted const
     {
-        return !!_aaEqual(this, *cast(const void**) p1, *cast(const void**) p2);
+        return !!_aaEqual(this, *cast(const AA*) p1, *cast(const AA*) p2);
     }
 
     override hash_t getHash(scope const void* p) nothrow @trusted const
     {
-        return _aaGetHash(cast(void*)p, this);
+        return _aaGetHash(cast(AA*)p, this);
     }
 
     // BUG: need to add the rest of the functions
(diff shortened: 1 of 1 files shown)


ibuclaw:
Alternatively, this could be fixed the other way around.  Have the signatures in _aaA.d changed to match object.d.

wilzbach:
CC @CyberShadow as are fails with this.

CyberShadow:
Looking at it, I don't see any reason to believe this is anything but a genuine regression... @ibuclaw Maybe you have some ideas. I will run a reduction, but it will take a while.

CyberShadow:
@ibuclaw I reduced it to the following:

```d
struct S
{
	string fileName;
	void[][string] cachedData;
	alias cachedData this;
}

unittest
{
	auto s = S("a");
	assert(s.keys.length == 0);
}
```

This segfaults for me with your patch.

ibuclaw:
@CyberShadow - thanks, I'll add that to the druntime unittests.

ibuclaw:
I think the problem is _aaRange needs updating too.

ibuclaw:
@CyberShadow 's bug has been addressed.

PR #2564, opened 2019-04-17
fix destroy for C++ classes without destructors

9il:


thewilsonator:
Does this have a bugzilla number? If not please create one so that the changelog picks up on it.

thewilsonator:
Ping me when this is good to go.

PR #2565, opened 2019-04-17
moveEmplace: optimise and make it betterC friendly

9il:


thewilsonator:
Please also update [the Phobos copy](https://github.com/dlang/phobos/blob/2f030ebbd947c3fed874b3884ca40c793f766dfe/std/algorithm/mutation.d#L1414) as wells they don't get out of sync.

9il:
btw, why Phobos just does not imports core.lifetime?

9il:
https://github.com/dlang/phobos/pull/6964

thewilsonator:
Phobos has a check for no internal pointers or something that can't be in druntime because it depends on too much of phobos.

9il:
Is it this because of optimization reasons?

thewilsonator:
No, its because druntime cannot depend on phobos.

9il:
Yes, but phobos can depend on druntime. Why not to use public imports?

jacob-carlborg:
> Yes, but phobos can depend on druntime. Why not to use public imports?

Because the implementation is not the same.

wilzbach:
Druntime would need this to match the Phobos implementation: https://github.com/dlang/druntime/pull/2447

9il:
I closed it because of the same change in Phobos fails.

PR #2570, opened 2019-04-21
createLowLevelThread: add an optional DLL unload callback

rainers:
... that is called if the DLL is supposed to be unloaded.

Finally figured a way to deal with the DLL issues in #2514. See comment in the source for explanation: https://github.com/dlang/druntime/compare/master...rainers:dll_thread?expand=1#diff-8bb12ed976acf0a5132e877ec5a01ea8R5696


rainers:
Added another special case: do not start thread during DLL_PROCESS_DETACH, it will crash later.

thewilsonator:
Is this good to go?

rainers:
> Is this good to go?

I think so. I have rebased #2514 to this and it also passes now.

PR #2573, opened 2019-04-22
[Trivial] Change suggestion for delete to destroy in core.memory

JinShil:
`delete` has been deprecated in favor of `destroy`.  See https://dlang.org/deprecate.html#delete

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -597,7 +597,7 @@ struct GC
      * collector, if p points to the interior of a memory block, or if this
      * method is called from a finalizer, no action will be taken.  The block
      * will not be finalized regardless of whether the FINALIZE attribute is
-     * set.  If finalization is desired, use delete instead.
+     * set.  If finalization is desired, call $(REF1 destroy, object) prior to `GC.free`.
      *
      * Params:
      *  p = A pointer to the root of a valid memory block or to null.


src/core/memory.d
```diff
@@ -597,7 +597,7 @@ struct GC
      * collector, if p points to the interior of a memory block, or if this
      * method is called from a finalizer, no action will be taken.  The block
      * will not be finalized regardless of whether the FINALIZE attribute is
-     * set.  If finalization is desired, use delete instead.
+     * set.  If finalization is desired, use $(REF1 destroy, object) instead.
```
rainers: `destroy` should be called in addition before `GC.free`, not instead.

src/core/memory.d
```diff
@@ -597,7 +597,7 @@ struct GC
      * collector, if p points to the interior of a memory block, or if this
      * method is called from a finalizer, no action will be taken.  The block
      * will not be finalized regardless of whether the FINALIZE attribute is
-     * set.  If finalization is desired, use delete instead.
+     * set.  If finalization is desired, call $(REF1 destroy, object) prior to `free`.
```
thewilsonator: I know this in in `struct GC` but you still might want to emphasise that it is `GC.free` that you should call, not C's `free`.

PR #2576, opened 2019-04-23
[OpenBSD] Updated extern(C) bindings for minimal compilation passes.

ibuclaw:
The platform is quite unfriendly for native D runtime however.  There's a complete lack of any TLS support, but there's always faking it such as what used to be done on OSX, or currently on Android.

### DIFF
--- a/src/core/stdc/stdio.d
+++ b/src/core/stdc/stdio.d
@@ -621,6 +621,8 @@ else version (OpenBSD)
     }
 
     ///
+    alias __sFILE _iobuf;
+    ///
     alias shared(__sFILE) FILE;
 }
 else version (DragonFlyBSD)
@@ -994,14 +996,16 @@ else version (OpenBSD)
         _IONBF = 2,
     }
 
-    private extern shared FILE[] __sF;
-
+    private extern __gshared FILE[3] __sF;
+    @property auto __stdin()() { return &__sF[0]; }
+    @property auto __stdout()() { return &__sF[1]; }
+    @property auto __stderr()() { return &__sF[2]; }
     ///
-    shared stdin  = &__sF[0];
+    alias __stdin stdin;
     ///
-    shared stdout = &__sF[1];
+    alias __stdout stdout;
     ///
-    shared stderr = &__sF[2];
+    alias __stderr stderr;
 }
 else version (DragonFlyBSD)
 {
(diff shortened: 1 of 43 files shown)


src/core/sys/openbsd/sys/elf32.d
```diff
@@ -0,0 +1,187 @@
...
+}
+
+extern (D) pure @safe
+{
+    auto ELF32_R_SYM(V)(V val) { return val >> 8; }
+    auto ELF32_R_TYPE(V)(V val) { return val & 0xff; }
+    auto ELF32_R_INFO(S, T)(S sym, T type) { return (sym << 8) + (type & 0xff); }
+}
+
+alias Elf_Note Elf32_Nhdr;
```
carun: It would be good to have this written in the form `alias newName = oldName;`

src/core/sys/posix/sys/stat.d
```diff
@@ -1124,6 +1124,83 @@ else version (NetBSD)
...
+      static if (__POSIX_VISIBLE >= 200809 || __BSD_VISIBLE)
+      {
+        timespec __st_birthtim;
+      }
+      else
+      {
+        time_t  __st_birthtime;
+        long    __st_birthtimensec;
+      }
+    }
```
jacob-carlborg: The content of this block doesn't look properly indented.

ibuclaw: static ifs are two space indented rather than four so as to not break the indentation alignment of all fields.

src/core/stdc/stdio.d
```diff
@@ -620,6 +620,8 @@ else version (OpenBSD)
         fpos_t          _offset;
     }
 
+    ///
+    alias __sFILE _iobuf;
```
thewilsonator: Ditto to what @carun said

src/core/stdc/stdio.d
```diff
@@ -994,14 +996,16 @@ else version (OpenBSD)
...
+    @property auto __stderr()() { return &__sF[2]; }
     ///
-    shared stdin  = &__sF[0];
+    alias __stdin stdin;
     ///
-    shared stdout = &__sF[1];
+    alias __stdout stdout;
     ///
-    shared stderr = &__sF[2];
+    alias __stderr stderr;
```
thewilsonator: Ditto.

src/core/stdc/stdio.d
```diff
@@ -1516,9 +1520,9 @@ else version (OpenBSD)
     enum __SALC = 0x4000;
     enum __SIGN = 0x8000;
 
-    extern int __isthreaded;
+    extern immutable __gshared int __isthreaded;
 
-    extern (D)
+    extern (D) @trusted
```
thewilsonator: Why would these not be `@safe` already?

src/core/sys/openbsd/sys/elf32.d
```diff
@@ -0,0 +1,187 @@
...
+module core.sys.openbsd.sys.elf32;
+
+version (OpenBSD):
+
+extern (C):
+
+import core.stdc.stdint;
+public import core.sys.openbsd.sys.elf_common;
+
+alias uint16_t Elf32_Half;
```
thewilsonator: Ditto.

src/core/stdc/stdio.d
ibuclaw: It's a copy-pasto from another block. Ideally they'd all be commonized and there be only one place for the definition/signature.

PR #2582, opened 2019-04-25
Convert FreeBSD's sys/ttycom.h into core.sys.posix.sys.ttycom

oikumene:
In FreeBSD, as in OSX, the tty part of ioctl is defined in sys/ttycom.h which is included by sys/ioctl.h.
In druntime, there were OSX part of definitions in ttycom.d, but there were no FreeBSD equivalent.
This change implements FreeBSD part of ttycom.d and ioccom.d, and public import core.sys.posix.sys.ttycom in ioctl.d.
The OSX users and FreeBSD users can now use full ioctls related to tty by importing core.sys.posix.sys.ioctl, like including sys/ioctl.h in C language. (For example, TIOCGWINSZ ioctl was
not provided in FreeBSD.)
Since there are only version(OSX) and version(FreeBSD) part in ttycom.d, public import'ing core.sys.posix.sys.ttycom from ioctl.d will make no harm to other platforms.

### DIFF
--- a/src/core/sys/posix/sys/ioccom.d
+++ b/src/core/sys/posix/sys/ioccom.d
@@ -66,7 +66,7 @@ version (OSX)
         return _IOC!(T)(IOC_INOUT, cast(uint)g, cast(uint)n, T.sizeof);
     }
 }
-else version(FreeBSD)
+else version (FreeBSD)
 {
     /* FreeBSD ioctl's encode the command in the lower 16-bits
      * and the size of any in/out parameters in the lower 13 bits of the upper
--- a/src/core/sys/posix/sys/ioccom.d
+++ b/src/core/sys/posix/sys/ioccom.d
@@ -97,7 +97,7 @@ else version (FreeBSD)
     enum uint IOC_DIRMASK = (IOC_VOID|IOC_OUT|IOC_IN);
 
     // encode the ioctl info into 32 bits
-    uint _IOC(T=typeof(null))(uint inorout, uint group, uint num, size_t len)
+    uint _IOC(uint inorout, uint group, uint num, size_t len)
     {
         return (inorout | ((len & IOCPARM_MASK) << 16) | (group << 8) | num);
     }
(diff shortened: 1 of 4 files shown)


src/core/sys/posix/sys/ioccom.d
```diff
@@ -66,3 +66,64 @@ version (OSX)
...
+    enum uint IOCPARM_MAX = (1 << IOCPARM_SHIFT); // max size of ioctl args
+
+    enum uint IOC_VOID = 0x20000000; // no parameters
+    enum uint IOC_OUT = 0x40000000; // copy parameters back
+    enum uint IOC_IN = 0x80000000; // copy parameters into
+    enum uint IOC_INOUT = (IOC_IN | IOC_OUT);
+    enum uint IOC_DIRMASK = (IOC_VOID|IOC_OUT|IOC_IN);
+
+    // encode the ioctl info into 32 bits
+    uint _IOC(T=typeof(null))(uint inorout, uint group, uint num, size_t len)
```
thewilsonator: T is unused.

src/core/sys/posix/sys/ioccom.d
```diff
@@ -66,3 +66,64 @@ version (OSX)
...
+    }
+
+    // encode a command with no parameters
+    uint _IO(char g, int n)
+    {
+        return _IOC(IOC_VOID, cast(uint)g, cast(uint)n, cast(size_t)0);
+    }
+    uint _IOWINT(T=int)(char g, int n)
+    {
+        return _IOC(IOC_VOID, cast(uint)g, cast(uint)n, int.sizeof);
```
thewilsonator: T is unused, did you mean to use T.sizeof?

src/core/sys/posix/sys/ioccom.d
oikumene: Removed template parameter and make _IOC a function.

src/core/sys/posix/sys/ioccom.d
oikumene: Same as above. Removed not used template parameter.

jacob-carlborg: I had problems in the past with functions in the bindings not being available causing linker errors. This is most likely due to the bindings not being compiled. Can someone confirm this?  What I’m saying it that this might need to be a template

thewilsonator: This can be fixed later if it is an issue.

PR #2585, opened 2019-04-26
Fix Issue 19830 - core.memory.__delete destructs arrays of structs in the wrong order

JinShil:
This issue was discovered while working on https://github.com/dlang/dmd/pull/9666

This test https://github.com/dlang/dmd/blob/b0007c1193dfb72bcfa8cef3f420fee68564ce49/test/runnable/sdtor.d#L186-L208 does not pass when `delete` is replaced with `__delete`.  This PR fixes that.

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -1067,7 +1067,7 @@ void __delete(T)(ref T x) @system
     {
         static if (is(E == struct))
         {
-            foreach (ref e; x)
+            foreach_reverse (ref e; x)
                 _destructRecurse(e);
         }
     }
@@ -1192,10 +1192,14 @@ unittest
         int a;
         ~this()
         {
+            assert(dtorCalled == a);
             dtorCalled++;
         }
     }
     auto arr = [A(1), A(2), A(3)];
+    arr[0].a = 2;
+    arr[1].a = 1;
+    arr[2].a = 0;
 
     assert(GC.addrOf(arr.ptr) != null);
     __delete(arr);


thewilsonator:
Thanks!

src/core/memory.d
```diff
@@ -1067,7 +1067,7 @@ void __delete(T)(ref T x) @system
     {
         static if (is(E == struct))
         {
-            foreach (ref e; x)
+            foreach_reverse (ref e; x)
```
rainers: See https://github.com/dlang/druntime/blob/master/src/rt/lifetime.d#L1157 for currently happens with delete: finalizers are only called if it is a runtime allocated array (debatable), pointer offset is corrected for large arrays (GC.free is wrong otherwise), block cache is invalidated (not doing so might make alter operations unsafe).

JinShil: That seems to be an existing issue with `__delete`'s impementation, but out of scope of this PR.  Can you just file a bug report, so it can be dealt with properly?

JinShil: Probably we should just have `__delete` call `_d_delarray_t`.

wilzbach: Can't do that as these C functions are supposed to go once the deprecation is over.

rainers: > That seems to be an existing issue with `__delete`'s impementation, but out of scope of this PR. 

Fine with me.

> Can you just file a bug report, so it can be dealt with properly?

https://issues.dlang.org/show_bug.cgi?id=13558

> Probably we should just have `__delete` call `_d_delarray_t`.

That's what I thought, too.

JinShil: I think the implementation is still needed though; it just doesn't get called directly by the compiler.  We either need to move the implementation into `__delete` or have `__delete` call it.  I suppose we could also convert `_d_delarray_t` into a template to get rid of the dependency on `TypeInfo`.

wilzbach: I think Andrei would want us to move the implementation into __delete as the general idea is to move into a D without any runtime.

JinShil: Ok, I'll look into that soon.

PR #2588, opened 2019-04-28
Fix broken link in README

Flying-Toast:
Attempt #<span></span>2. (This time I won't delete the fork until *after* the PR is merged :laughing:)


thewilsonator:
Removing automerge because DAutoTest fails, cc @CyberShadow 

@Flying-Toast you'll need to force push again once whatever is wrong is fixed.

CyberShadow:
Looks like a temporary error.

PR #2594, opened 2019-04-29
[netbsd] core.stdc.fenv: Fix error undefined uint32_t

ibuclaw:


src/core/sys/posix/signal.d
```diff
@@ -1136,55 +1136,57 @@ else version (NetBSD)
...
+                int   _trap;
+                int   _trap2;
+                int   _trap3;
+            } fault_t fault;
+            struct poll_t
+            {
+                c_long _band;
+                int  _fd;
+            } poll_t _poll;
+        }
```
krytarowski: For your interest, NetBSD 9.0 will ship with one new struct here.

ibuclaw: Wonderful. I'm slowly coming round to thinking about that auto-generating these binding is likely the best way forward.

krytarowski: On the other hand these things change rarely.

ibuclaw: Yes, but currently you can't reliably support multiple versions of the same platform, and assuming the latest version is not a good way to go.

PR #2596, opened 2019-05-05
[GDC] Support GDC in core.atomic

belka-ew:
This is the implementation currently used by GDC.

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -24,6 +24,12 @@ else version (D_InlineAsm_X86_64)
     enum has64BitCAS = true;
     enum has128BitCAS = true;
 }
+else version (GNU)
+{
+    import gcc.config;
+    enum has64BitCAS = GNU_Have_64Bit_Atomics;
+    enum has128BitCAS = GNU_Have_LibAtomic;
+}
 else
 {
     enum has64BitCAS = false;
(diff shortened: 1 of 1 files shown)


ibuclaw:
Actually, there's a modified version in trunk/9.1 that supports targets like riscv that have libatomic, but no compiler intrinsics.  Likewise there's a new fallback path to accommodate the absence of any atomics at all.

Both of which will need to be included in this.

src/core/atomic.d
```diff
@@ -24,6 +24,12 @@ else version (D_InlineAsm_X86_64)
     enum has64BitCAS = true;
     enum has128BitCAS = true;
 }
+else version (GNU)
+{
+    import gcc.config;
```
n8sh: Does GDC treat the modules `gcc.config` and `gcc.builtins` specially?

ibuclaw: The former is a template `config.d.in` that's generated at configure-time. The latter is an empty module that has all builtins injected into it at compile-time, its contents may vary depending on compiler flags.

belka-ew:
@ibuclaw I updated the code from trunk. Is this what you meant?

ibuclaw:
@belka-ew, yes, PR looks correct now.

src/core/atomic.d
```diff
@@ -1404,6 +1410,328 @@ else version (AsmX86_64)
...
+        }
+
+        // Implements lock/unlock operations.
+        private struct AtomicMutex
+        {
+            int lock() pure @trusted @nogc nothrow
+            {
+                static if (GNU_Thread_Model == ThreadModel.Posix)
+                {
+                    if (!_inited)
```
rainers: Not sure if it is critical but this looks a bit dangerous if initialization is requested from two threads concurrently.
What platforms match the condition (!GNU_Have_Atomics && !GNU_Have_LibAtomic)? I guess Windows will always have atomics.

ibuclaw: I think the only way that could happen is if a C program calls rt_init in two threads concurrently.  Using pragma(crt_constructor) is another option, though I don't recall adding that feature to the dmd-cxx branch.

rainers: Because there is an atomic operation somewhere called from rt_init? A comment to that respect would be nice in that case.

ibuclaw: Hmm, I did write this at the 11th hour, so wasn't strictly thinking too clear about this.  Also this was committed before @jpf91 added gcc.gthreads, so I guess we could remove some duplication here.

ibuclaw: I think something like this would be more succinct, and doesn't have the same pitfall.
```
static if (!GNU_Have_Atomics && !GNU_Have_LibAtomic)
{
    import gcc.gthread;

    // Internal static mutex reference.
    private __gthread_mutex_t* _getAtomicMutex() @trusted @nogc nothrow @property
    {
        __gshared static __gthread_mutex_t mutex = GTHREAD_MUTEX_INIT;
        return &mutex;
    }

    // Implements lock/unlock operations.
    private int _atomicMutexLock(__gthread_mutex_t* mutex) @trusted @nogc nothrow
    {
        return __gthread_mutex_lock(mutex);
    }

    private int _atomicMutexUnlock(__gthread_mutex_t* mutex) @trusted @nogc nothrow
    {
        return __gthread_mutex_unlock(mutex);
    }

    // Fake the purity of the functions so that they can be used in pure/nothrow/@safe code.
    pragma(mangle, _getAtomicMutex.mangleof)
    private __gthread_mutex_t* getAtomicMutex() pure @trusted @nogc nothrow @property;

    pragma(mangle, _atomicMutexLock.mangleof)
    private int atomicMutexLock(__gthread_mutex_t*) pure @trusted @nogc nothrow;

    pragma(mangle, _atomicMutexUnlock.mangleof)
    private int atomicMutexUnlock(__gthread_mutex_t*) pure @trusted @nogc nothrow;
}
```
Once __gthread_mutex_t and functions have been added to gcc.gthread, that is.

n8sh:
BTW what is the motivation for this PR? Is the goal to upstream all druntime & phobos changes in LDC & GDC?

src/core/atomic.d
```diff
@@ -1404,6 +1410,328 @@ else version (AsmX86_64)
...
+                   op == "%=" || op == "^^=" || op == "&="  || op == "|=" ||
+                   op == "^=" || op == "<<=" || op == ">>=" || op == ">>>=" ) // skip "~="
+        {
+            TailShared!T get, set;
+
+            do
+            {
+                get = set = atomicLoad!(MemoryOrder.raw)( val );
+                mixin( "set " ~ op ~ " mod;" );
+            } while ( !cas( &val, get, set ) );
```
n8sh: Can GDC use `__atomic_add_fetch` instead of a compare and swap loop for "+=" or is that not available?

belka-ew:
> BTW what is the motivation for this PR? Is the goal to upstream all druntime & phobos changes in LDC & GDC?

Yes. druntime contains already compiler and platform dependent code. To be honest I don't see a better way to make druntime more extendable, so that some functionality can be implemented in the compiler and made available for the drutnime.
Currently updating the druntime is a pain, I have to look what files have been changed, apply changes, resolve conflicts, look if everything has been updated...

ibuclaw:
> BTW what is the motivation for this PR? Is the goal to upstream all druntime & phobos changes in LDC & GDC?

Druntime has always been lenient when it comes to compiler specific things in common/shared modules (core.cpuid, core.stdc.stdlib, rt.qsort).

Phobos should not have any compiler specific code. The only exception to the rule is std.math, but I think that the right way to go there is to turn it into a package, similar to how bigint is split into asm and non-asm parts.

belka-ew:
> Phobos should not have any compiler specific code. The only exception to the rule is std.math, but I think that the right way to go there is to turn it into a package, similar to how bigint is split into asm and non-asm parts.

Or move asm parts as simple functions into druntime and call them from phobos.

belka-ew:
> Druntime has always been lenient when it comes to compiler specific things in common/shared modules (core.cpuid, core.stdc.stdlib, rt.qsort).

`stdc.stdarg` is the next candidate.

src/core/atomic.d
jpf91: This should work. However, I don't really like that we duplicate `core.sync.mutex` here. Is this only required because `core.sync.mutex` is not `pure`? If so, can we change it to be pure?

TurkeyMan:
@ibuclaw This PR will completely change your solution here: #2739
I'm trying to separate the platform specific parts from the rest of the code, the functions int `core.internal.atomic` should be a whole lot easier to substitute with intrinsics.

PR #2599, opened 2019-05-05
GC: fix wrong profile result if more than 4GB used

rainers:


thewilsonator:
Does this have a bugzilla? Also this is so trivial it should go to stable.

rainers:
Added bug reort and rebased to  stable.

thewilsonator:
Thanks!

jacob-carlborg:
Is it possible to make an artificial test, by somehow manually setting `mappedPages` to a big value?

rainers:
> Is it possible to make an artificial test, by somehow manually setting `mappedPages` to a big value?

Possible, but I think that would be rather invasive for such a trivial fix. It would be nice if the compiler would warn about cases like these, i.e. when the result of a calculation might overflow an int, but the assigned lvalue has larger size (similar for comparisons).

src/gc/impl/conservative/gc.d
```diff
@@ -1596,7 +1596,7 @@ struct Gcx
 
     private @property bool lowMem() const nothrow
     {
-        return isLowOnMem(mappedPages * PAGESIZE);
+        return isLowOnMem(cast(size_t)mappedPages * PAGESIZE);
```
ZombineDev: I wonder if it would be better to instead change the type of the `PAGESIZE` enum to `size_t`. That way any operation where one of the operands is `PAGESIZE` would promote the other operand as necessary.

PR #2602, opened 2019-05-10
Try to fix issue 19831 without understanding by copying the behavior from __dmd_personality_v0

FeepingCreature:
NOT feeling good about this. Seems to work though???

Mostly opening this PR so tests get run.

This was not arrived at through any kind of insight, but mostly by flailing around.

Ping @donc . Help us Don Clugston, you're our only hope.

edit: @kinke ping? ("No. There is another.")

### DIFF
--- a/src/rt/dwarfeh.d
+++ b/src/rt/dwarfeh.d
@@ -472,13 +472,25 @@ ClassInfo getClassInfo(_Unwind_Exception* exceptionObject)
 {
     ExceptionHeader* eh = ExceptionHeader.toExceptionHeader(exceptionObject);
     Throwable ehobject = eh.object;
+    auto currentLsd = eh.languageSpecificData;
     //printf("start: %p '%.*s'\n", ehobject, ehobject.classinfo.info.name.length, ehobject.classinfo.info.name.ptr);
     for (ExceptionHeader* ehn = eh.next; ehn; ehn = ehn.next)
     {
+        // like __dmd_personality_v0, don't combine when the exceptions are from different functions
+        // (fixes issue 19831, exception thrown and caught while inside finally block)
+        if (currentLsd != ehn.languageSpecificData)
+        {
+            // printf("break: %p %p\n", currentLsd, ehn.languageSpecificData);
+            break;
+        }
+
         //printf("ehn =   %p '%.*s'\n", ehn.object, cast(int)ehn.object.classinfo.info.name.length, ehn.object.classinfo.info.name.ptr);
         Error e = cast(Error)ehobject;
         if (e is null || (cast(Error)ehn.object) !is null)
+        {
+            currentLsd = ehn.languageSpecificData;
             ehobject = ehn.object;
+        }
     }
     //printf("end  : %p\n", ehobject);
     return ehobject.classinfo;
--- a/test/exceptions/Makefile
+++ b/test/exceptions/Makefile
@@ -1,7 +1,7 @@
 include ../common.mak
 
 TESTS=stderr_msg unittest_assert invalid_memory_operation unknown_gc static_dtor \
-	  future_message rt_trap_exceptions_drt
+	  future_message rt_trap_exceptions_drt catch_in_finally
 
 ifeq ($(OS)-$(BUILD),linux-debug)
 	TESTS+=line_trace long_backtrace_trunc rt_trap_exceptions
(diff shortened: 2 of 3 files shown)


FeepingCreature:
oh no they're passing

FeepingCreature:
Note from IRL: @ibuclaw looked it over and thinks it's basically sound, he might have some time tomorrow at the Hackathon to get it documented a bit better.

donc:
Note that I only wrote the original Windows exception chaining, I don't know anything about how the Linux exception handler works. Does this bug apply to Windows as well?
BTW despite having implemented it, I think that exception chaining is a really, really bad idea. I don't think anybody would ever want to look at the exception chain, and it creates a host of problems.

FeepingCreature:
Walter also said that exception chaining should be removed and we should go back to C++ and crashing on throw-in-finally.

I must strenuously object to at least that last part - `scope(exit)` is a core, heavily advertised feature of D and it'll look terrible (and basically make it unusable) if a basic feature like exceptions causes crashes inside it.

FeepingCreature:
Re Windows: I tried with wine+DMDwin32 and at least in that configuration, the problem does not occur.

ibuclaw:
I think there needs to be a few more tests. See eh.d exception chaining tests, but adapt as you do here to trigger the bug.

FeepingCreature:
I don't understand what you mean. `eh.d` exception chaining tests clearly did not trigger the bug. Should I just rewrite `eh.d` to be `eh.d, but every `finally` throws and catches an exception?

You already can't `throw`/`catch` in finally directly. I'm not sure how else you could order these elements to produce a test.

FeepingCreature:
ping @ibuclaw

ibuclaw:
@FeepingCreature there is a slight distinction between those tests and this test.

In eh.d you have:
```
void collideone()
{
    try
    {
        throw new Exception("x");
    }
    finally
    {
        throw new Exception("y");
    }
}
```
This test is:
```
void collideone()
{
    try
    {
        throw new Exception("x");
    }
    finally
    {
        funcThatThrowsAndCatches();
    }
}
```

I think that those tests could be copied to your new test file and amended as needed to check that the various weird chain cases and this mix well.

FeepingCreature:
I'll take a look tomorrow.

FeepingCreature:
@ibuclaw Is this better? It's hard to write comments and tests for code that I don't understand how it works.

src/rt/dwarfeh.d
```diff
@@ -472,13 +472,25 @@ ClassInfo getClassInfo(_Unwind_Exception* exceptionObject)
 {
     ExceptionHeader* eh = ExceptionHeader.toExceptionHeader(exceptionObject);
     Throwable ehobject = eh.object;
+    auto currentLsd = eh.languageSpecificData;
     //printf("start: %p '%.*s'\n", ehobject, ehobject.classinfo.info.name.length, ehobject.classinfo.info.name.ptr);
     for (ExceptionHeader* ehn = eh.next; ehn; ehn = ehn.next)
     {
+        // like __dmd_personality_v0, don't combine when the exceptions are from different functions
+        // (fixes issue 19831, exception thrown and caught while inside finally block)
```
ibuclaw: Note, if a function has been partitioned, this will also not match.

However this will never happen with dmd, and gdc turns off partitioning unless the optimization is explicitly requested on the command line.

FeepingCreature: ....

[_throws up hands_ I GUESS--](https://i.imgur.com/Kp6RN00.png)

donc:
I would like to see a test where the exception is thrown in one function and caught in another. If this works, your PR has my blessing.
```
void thrower()
{
   throw new Exception("a");
}

void functhatthrowsandslowlycatches()
{
     try {
         thrower();
    }
    catch(Exception e){}
}

void test5()
{
   try {
      throw new Exception("x");
   }
   finally {  
      functhatthrowsandslowlycatches();
  }
}
````

FeepingCreature:
@donc See `test5`. Seems to work.

donc:
LGTM.

thewilsonator:
As this fixes a severity major bug it should go in stable, unless you don't feel confident about doing so.

src/rt/dwarfeh.d
kinke: This broke `runnable/eh.d` for LDC [with its slight modifications in this file]. The testsuite passes again when NOT breaking here as long as `currentLsd` is null, but I have no idea whether that's correct. @ibuclaw: any thoughts?

ibuclaw: Why would it be null?

kinke: AFAICT, each `ExceptionHeader.languageSpecificData` starts out with null and is only set **after** finding a matching handler/catch in the search phase.

ibuclaw: Well the lsda is known since the beginning of the personality function, you could either eagerly set it, or pass it onto getClassInfo to initialize cuurentLsd with.

kinke: Yeah, the question is rather 'why does DMD have no problem with it'. I'll pass it through; pls ping me if GDC has problems too, so that we can come up with something here upstream.

PR #2603, opened 2019-05-11
fix ldap bindings, fix issue 17248

WebFreak001:
Looked up and corrected most LDAP methods:

changed `char*` (`PCHAR`) arguments to `const(char)*` (`PCSTR`) where documentation says `const`.

Renamed some types to be the same as in the documentation. (purely aesthetic, fit in with adding the const to the parameters well)

Fixed a PLDAPControlW actually being PLDAPControlA in an ASCII method.

Changed `T[]` arguments to `T*` (fixes the referenced issue)

Did not test this with an actual connection yet. The definition is more like in my `ldap` package on dub now, which works for me.

### DIFF
--- a/src/core/sys/windows/winldap.d
+++ b/src/core/sys/windows/winldap.d
@@ -672,11 +672,11 @@ extern (C) {
     LDAP ldap_conn_from_msg(LDAP*, LDAPMessage*);
     INT LdapUnicodeToUTF8(LPCWSTR, int, LPSTR, int);
     INT LdapUTF8ToUnicode(LPCSTR, int, LPWSTR, int);
-    deprecated {
-        ULONG ldap_bindA(LDAP*, PSTR, PCHAR, ULONG);
-        ULONG ldap_bindW(LDAP*, PWSTR, PWCHAR, ULONG);
-        ULONG ldap_bind_sA(LDAP*, PSTR, PCHAR, ULONG);
-        ULONG ldap_bind_sW(LDAP*, PWSTR, PWCHAR, ULONG);
+    ULONG ldap_bindA(LDAP*, PSTR, PCHAR, ULONG);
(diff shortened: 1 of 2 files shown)


src/core/sys/windows/winldap.d
```diff
@@ -491,111 +491,111 @@ extern (C) {
...
-    ULONG ldap_simple_bind_sW(LDAP*, PWCHAR, PWCHAR);
+    ULONG ldap_simple_bindA(LDAP*, PSTR, PSTR);
+    ULONG ldap_simple_bindW(LDAP*, PWSTR, PWSTR);
+    ULONG ldap_simple_bind_sA(LDAP*, PSTR, PSTR);
+    ULONG ldap_simple_bind_sW(LDAP*, PWSTR, PWSTR);
     ULONG ldap_unbind(LDAP*);
     ULONG ldap_unbind_s(LDAP*);
-    ULONG ldap_search_extA(LDAP*, PCHAR, ULONG, PCHAR, PCHAR[], ULONG,
+    ULONG ldap_search_extA(LDAP*, PCSTR, ULONG, PCSTR, PZPSTR, ULONG,
+      PLDAPControlA*, PLDAPControlA*, ULONG, ULONG, ULONG*);
```
WebFreak001: this was `PLDAPControlW*` before even though this is an `A` method (corrected it to what the docs said this type is)

thewilsonator:
Please merge when you have had a chance to test.

WebFreak001:
I undeprecated the `ldap_bind` functions (they aren't obsolte or deprecated in the documentation) and added a deprecation message to the `ldap_modrdn` functions to make it clear why they are deprecated.

Otherwise usage works fine now, (connection works, etc.) the `[]` methods were the only broken methods before and this PR just fixes those and makes the methods allow more inputs (using const)

PR #2604, opened 2019-05-11
SAOC: fork() based garbage collector

FraMecca:
Full benchmarks here:
https://gist.github.com/FraMecca/ee50c9f4d1dc16d115dcddce8000cf9f
The other benchmark programs from druntime are not indicative in the same way they weren't for the threaded version.

Benchmark results for vdparser_extended:

fork enabled, average of 20 runs
```
Elapsed (wall clock) time (h:mm:ss or m:ss)      12.6355 seconds
GC summary                                     1454.2 MB
Grand total GC time                             186.5 ms
Max Pause Time                                   65.45 ms
Number of collections                             9 
Total GC prep time                               20.550 ms
Total mark time                                 106.25 ms
Total sweep time                                 58.80 ms
```
standard, average of 20 runs:
```
Elapsed (wall clock) time (h:mm:ss or m:ss)      14.363 seconds
GC summary                                     1355.0 MB
Grand total GC time                            2629.30 ms
Max Pause Time                                  969.85 ms
Number of collections                            14 
Total GC prep time                               24.950 ms
Total mark time                                2560.05 ms
Total sweep time                                 43.350 ms
```

Benchmark results for `Dustmite --force ../../druntime/src/ 'grep -nR HAVE_FORK'`:

fork enabled, average of 20 runs:
```
Elapsed (wall clock) time (h:mm:ss or m:ss)     18.04 seconds
GC summary                                     477 MB
Grand total GC time                             69.60 ms
Max Pause Time                                  24.50 ms
Number of collections                            6
Total GC prep time                               2.05 ms
Total mark time                                 38.85 ms
Total sweep time                                27.350 ms
```
standard, average of 20 runs:
```
Elapsed (wall clock) time (h:mm:ss or m:ss)     18.1285  seconds
GC summary                                     477 MB
Grand total GC time                            475.45 ms
Max Pause Time                                 260.75 ms
Number of collections                            9
Total GC prep time                               2 ms
Total mark time                                452.35 ms
Total sweep time                                19.85 ms
```


src/gc/impl/conservative/gc.d
```diff
@@ -2581,9 +2604,54 @@ struct Gcx
         begin = start = currTime;
 
         debug(COLLECT_PRINTF) printf("Gcx.fullcollect()\n");
+        bool doFork = config.fork;
+        version (COLLECT_PARALLEL)
+            bool doParallel = config.parallel > 0;
+        else
+            enum doParallel = false;
+
+        assert(!(doFork && doParallel), "No reason in mixing threads and fork");
```
FraMecca: this assert is dangerous because it will deadlock if the user runs with DRT-gcopt=fork:1 and doesn't disable threaded collection.

Every assert related to GC code deadlocks given that assert allocates. This is a problem related to code outside of this PR as well

wilzbach:
Remove the trailing whitespace for the CI to pass:

```
src/gc/impl/conservative/gc.d:2700:						break; // bogus
src/gc/impl/conservative/gc.d:2701:					default: // the parent
src/gc/impl/conservative/gc.d:2723:							disableFork();
posix.mak:348: recipe for target 'checkwhitespace' failed
```

src/gc/bits.d
```diff
@@ -33,29 +34,37 @@ struct GCBits
...
         }
     }
 
-    void alloc(size_t nbits) nothrow
+    void alloc(size_t nbits, bool mmap = false) nothrow
     {
         this.nbits = nbits;
-        data = cast(typeof(data[0])*)calloc(nwords, data[0].sizeof);
+        if (mmap)
+          data = cast(typeof(data[0])*)os_mem_map(nwords * data[0].sizeof, true); // Allocate as MAP_SHARED
```
jacob-carlborg: Indentation is wrong.

src/gc/os.d
```diff
@@ -49,6 +49,48 @@ else version (Posix)
...
 
+
+    /**
+     * Possible results for the wait_pid() function.
+     */
+    enum WRes
+    {
+     DONE, /// The process has finished successfully
+     RUNNING, /// The process is still running
+     ERROR /// There was an error waiting for the process
```
jacob-carlborg: * Indentation is wrong
* The standard D style is to use lower case for enum members.

src/gc/os.d
```diff
@@ -49,6 +49,48 @@ else version (Posix)
     version (CRuntime_UClibc) import core.sys.linux.sys.mman : MAP_ANON;
     import core.stdc.stdlib;
 
+
+    /**
+     * Possible results for the wait_pid() function.
+     */
+    enum WRes
```
jacob-carlborg: Please avoid shortening names.

src/gc/os.d
```diff
@@ -49,6 +49,48 @@ else version (Posix)
     version (CRuntime_UClibc) import core.sys.linux.sys.mman : MAP_ANON;
     import core.stdc.stdlib;
 
+
+    /**
+     * Possible results for the wait_pid() function.
+     */
```
jacob-carlborg: This comment could be one line by using `///`.

src/gc/os.d
```diff
@@ -69,13 +111,22 @@ else static if (is(typeof(malloc)))
...
+ * Indicates if an implementation support fork().
+ *
+ * The value shown here is just demostrative, the real value is defined based
+ * on the OS it's being compiled in.
+ * enum HAVE_FORK = true;
+*/
 
 static if (is(typeof(VirtualAlloc))) // version (GC_Use_Alloc_Win32)
 {
+    enum { HAVE_FORK = false }
```
jacob-carlborg: * Use lower case letters and camel casing
* You can write the above with the following syntax: `enum haveFork = false`.

src/gc/bits.d
```diff
@@ -33,29 +34,37 @@ struct GCBits
     enum BITS_MASK = (BITS_PER_WORD - 1);
     enum BITS_0 = cast(wordtype)0;
     enum BITS_1 = cast(wordtype)1;
+    bool mmap;
```
thewilsonator: Please move this down a couple of lines so that it is with the variables `data` and `nbits`

src/gc/impl/conservative/gc.d
```diff
@@ -1642,7 +1656,7 @@ struct Gcx
                 if (!newPool(1, false))
                 {
                     // out of memory => try to free some memory
-                    fullcollect();
+                    fullcollect(false, true); // stop the worlds
```
thewilsonator: `// stop the world` (singular)

src/core/gc/config.d
```diff
@@ -15,6 +15,7 @@ __gshared Config config;
 struct Config
 {
     bool disable;            // start disabled
+    bool fork = false;            // optional concurrent behaviour
```
rainers: please align comments.

src/core/gc/config.d
```diff
@@ -39,6 +40,7 @@ struct Config
 
         printf("GC options are specified as white space separated assignments:
     disable:0|1    - start disabled (%d)
+    fork:0|1    - set fork behaviour (disabled by default) (%d)
```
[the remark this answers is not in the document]
rainers: The user will probably not care how it's implemented (i.e. that is using fork), but selects the GC to be `concurrent`. So maybe the option should be the latter. Alternatively, it could also just be another GC named `concurrent`, but how to combine that with `precise`?

src/gc/impl/conservative/gc.d
```diff
@@ -43,6 +43,7 @@ import core.gc.gcinterface;
 import rt.util.container.treap;
```
rainers: I think you should define a version (e.g. `COLLECT_FORKED`) that is set only for supporting platforms. Most code should only be enabled with that version.

src/gc/impl/conservative/gc.d
```diff
@@ -56,6 +57,11 @@ else                   import core.stdc.stdio : sprintf, printf; // needed to ou
 import core.time;
 alias currTime = MonoTime.currTime;
 
+extern (C) nothrow @nogc
+{
+    void _Exit(int);
+}
+
```
rainers: You don't need braces for a single definition. Doesn't `_Exit` already exist in some core.sys.... module? You should better import that where used then.

src/gc/impl/conservative/gc.d
```diff
@@ -997,6 +1003,7 @@ class ConservativeGC : GC
 
     /**
      * Do full garbage collection.
+     The collection is done concurrently only if block is false.
```
rainers: missing `*` at the beginning of comment.

src/gc/impl/conservative/gc.d
```diff
@@ -1665,8 +1679,10 @@ struct Gcx
         // Return next item from free list
         bucket[bin] = (cast(List*)p).next;
         auto pool = (cast(List*)p).pool;
+
         auto biti = (p - pool.baseAddr) >> pool.shiftBy;
         assert(pool.freebits.test(biti));
+        pool.mark.set(biti); // be sure that the child is aware of the page being used
```
rainers: `page`->`bin`. This only needs to be executed while a collection is running in the background.

If mark is shared memory between processes, this needs to be an atomic operation (setLocked). Otherwise this might clear bits set concurrently by the other process. But it will also require the background scanning to use atomic operations.

src/gc/impl/conservative/gc.d
```diff
@@ -1746,7 +1762,12 @@ struct Gcx
         assert(pool);
 
         debug(PRINTF) printFreeInfo(&pool.base);
+        pool.pagetable[pn] = B_PAGE;
```
rainers: This looks like a merge conflict (but pool.mark.set where the same applies as in smallAlloc). Updating the pool (pagetable, freepages) already happens in LargeObjectPool.allocPages().

src/gc/impl/conservative/gc.d
rainers: The assert will go away in a release build, emit an error instead.

src/gc/impl/conservative/gc.d
```diff
@@ -2557,11 +2580,17 @@ struct Gcx
         return recoverPool[bin] = poolIndex < npools ? cast(SmallObjectPool*)pool : null;
     }
 
+    void disableFork() nothrow
+    {
+        config.fork = false;
```
rainers: Maybe better to not modify the configuration, but remember the failure in a member variable. Printing a warning might also be useful.

src/gc/impl/conservative/gc.d
```diff
@@ -2575,9 +2604,54 @@ struct Gcx
...
+
+        assert(!(doFork && doParallel), "No reason in mixing threads and fork");
         //printf("\tpool address range = %p .. %p\n", minAddr, maxAddr);
 
+        // If there is a mark process running, check if it already finished.
+        // If that is the case, we move to the sweep phase.
+        // If it's still running, either we block until the mark phase is
+        // done (and then sweep to finish the collection), or in case of error
+        // we redo the mark phase without forking.
+        if (collectInProgress && doFork)
```
rainers: As discussed move this section to a function if possible.

src/gc/impl/conservative/gc.d
```diff
@@ -2596,12 +2670,65 @@ struct Gcx
...
+            // in the child. If the collection should not block, the parent process
+            // tells the caller no memory could be recycled immediately (if this collection
+            // was triggered by an allocation, the caller should allocate more memory
+            // to fulfill the request).
+            // If the collection should block, the parent will wait for the mark phase
+            // to finish before returning control to the mutator,
+            // but other threads are restarted and may run in parallel with the mark phase
+            // (unless they allocate or use the GC themselves, in which case
+            // the global GC lock will stop them).
+            if (doFork)
```
rainers: This section, too.

src/gc/impl/conservative/gc.d
```diff
@@ -2634,6 +2770,14 @@ struct Gcx
             ConservativeGC._inFinalizer = false;
         }
 
+        // minimize() should be called only after a call to fullcollect
+        // terminates with a sweep
+        if (doMinimize || lowMem)
```
rainers: Can `doMinimize` be converted to a function parameter instead of a field? It's unlikely to actually find completely unused pool, so it's probably not too bad if it is skipped sometimes.

src/gc/impl/conservative/gc.d
```diff
@@ -3009,6 +3152,7 @@ struct Pool
             freebits.alloc(nbits);
             freebits.setRange(0, nbits);
         }
+        freebits.setAll();
```
rainers: This is already done with the `freebits.setRange(0, nbits);` call above.

src/gc/impl/conservative/gc.d
```diff
@@ -3819,6 +3963,9 @@ struct SmallObjectPool
         // Convert page to free list
         size_t size = binsize[bin];
         void* p = baseAddr + pn * PAGESIZE;
+        size_t biti = pn * (PAGESIZE / 16);
+        base.freebits.set(biti);
```
rainers: Merge error? BTW: freebits are already maintained correctly during malloc/free (this changed recently). They are supposed to be all set for an unused page.

src/gc/bits.d
thewilsonator: That also works.

src/gc/impl/conservative/gc.d
FraMecca: I get a sigsegv if I replace that call with  freebits.setRange(0, nbits);
setAll does this: `memset(data, 0xFF, nwords * wordtype.sizeof) `

src/gc/impl/conservative/gc.d
FraMecca: The GC can use a bool to remember if it should fork or not.
In any case it would be useful to give the user an option to disable the forking behaviour in certain section of user code.
If you think that is needed, I will add a enableFork function

src/gc/impl/conservative/gc.d
rainers: freebits is only allocated for SmallObjectPool, but setAll is called unconditionally. I suspect setRange doesn't work as expected with nbits == 0.

src/gc/impl/conservative/gc.d
rainers: There is currently no way to add a custom member function to the GC interface, so it would be a bit problematic for the user to call it.
Instead core.gc.config can be imported and config.fork can be modified, so the GC can adapt accordingly, but it should not modify the user setting.
I think you should just remember whether forking failed in a previous attempt and then ignore `config.fork`.

src/gc/impl/conservative/gc.d
FraMecca: Done, see commit 
https://github.com/dlang/druntime/pull/2604/commits/d3fe6ad574d09c8facc70172f9162b1d3c256fc6

src/gc/impl/conservative/gc.d
FraMecca: Done, see commit 
https://github.com/dlang/druntime/pull/2604/commits/d3fe6ad574d09c8facc70172f9162b1d3c256fc6

src/gc/impl/conservative/gc.d
```diff
@@ -165,6 +165,8 @@ class ConservativeGC : GC
     this()
     {
         //config is assumed to have already been initialized
+        if (config.fork && config.parallel > 0)
+            throw new Error("No reason in mixing threads and fork");
```
FraMecca: @rainers this still results in a deadlock if the user does not specify 'fork:1 parallel:0'

Where do I throw this error avoiding the recursive lock?

FraMecca:
I don't know if I am doing the right thing by pushing small commits every time. Should I rebase before every push?

thewilsonator:
That is fine, you can combine them all at the end.

src/gc/bits.d
```diff
@@ -38,24 +39,30 @@ struct GCBits
     wordtype* data;
     size_t nbits;
 
-    void Dtor() nothrow
+    void Dtor(bool mmap) nothrow @nogc
```
rainers: Could use the same default (mmap = false) as the alloc function.

src/gc/impl/conservative/gc.d
rainers: Or use `static if (HaveFork)`.

src/gc/impl/conservative/gc.d
```diff
@@ -1746,7 +1759,11 @@ struct Gcx
         assert(pool);
 
         debug(PRINTF) printFreeInfo(&pool.base);
+        if (npages > 1)
+            memset(&pool.pagetable[pn + 1], B_PAGEPLUS, npages - 1);
```
rainers: The memset and the update to freepages is also already done in pool.allocPages

src/gc/impl/conservative/gc.d
```diff
@@ -2557,11 +2576,109 @@ struct Gcx
...
+            case ChildStatus.error:
+                debug(COLLECT_PRINTF) printf("\t\tmark proc ERROR\n");
+                // Try to keep going without forking
+                // and do the marking in this thread
+                break;
+        }
+        return rc;
+    }
+
+    ChildStatus markFork(bool nostack, bool block, const MonoTime begin, ref MonoTime start, ref MonoTime stop, ref Duration markTime) nothrow
```
rainers: I guess you can avoid passing all/most of these MonoTime variables here. Some are temporaries, markTime is a global.

src/gc/impl/conservative/gc.d
```diff
@@ -2985,7 +3155,7 @@ struct Pool
         topAddr = baseAddr + poolsize;
         auto nbits = cast(size_t)poolsize >> shiftBy;
 
-        mark.alloc(nbits);
+        mark.alloc(nbits, true);
```
rainers: Using mmap should only be done if the forking version is used.

src/gc/os.d
```diff
@@ -49,6 +49,46 @@ else version (Posix)
     version (CRuntime_UClibc) import core.sys.linux.sys.mman : MAP_ANON;
     import core.stdc.stdlib;
 
+
+    /// Possible results for the wait_pid() function.
+    enum ChildStatus
+    {
+     done, /// The process has finished successfully
+     running, /// The process is still running
+     error /// There was an error waiting for the process
```
rainers: bad indentation.

src/gc/os.d
```diff
@@ -49,6 +49,46 @@ else version (Posix)
...
+    }
+
+    /**
+     * Wait for a process with PID pid to finish.
+     *
+     * If block is false, this function will not block, and return ChildStatus.RUNNING if
```
rainers: The comment still uses upper case enum values.

src/gc/os.d
```diff
@@ -69,13 +109,22 @@ else static if (is(typeof(malloc)))
     version = GC_Use_Alloc_Malloc;
 else static assert(false, "No supported allocation methods available.");
 +/
+/**
+ * Indicates if an implementation support fork().
```
rainers: s/support/supports/

src/gc/os.d
```diff
@@ -69,13 +109,22 @@ else static if (is(typeof(malloc)))
...
+ *
+ * The value shown here is just demostrative, the real value is defined based
```
rainers: I don't think this will end up in any official documentation, so just move the description above to the first definition.

src/gc/impl/conservative/gc.d
rainers: You can use a statically allocated error:
```
__gshared Error err = new Error("No reason in mixing threads and fork");
throw err;
```

thewilsonator: we have `throw staticError!Error` for that.

src/gc/bits.d
FraMecca: Should it be implicit or explicit?
I worry that someone could forget to explicit the parameter in new code

FraMecca:
> Thanks.
> 
> I wonder what @leandro-lucarella-sociomantic 's original implementation did regarding the shared memory accesses or are these new to this version. Is it a valid concern that atomics have to be used?

The implementation is very similar to old Sociomantic that didn't use atomics as well.

jacob-carlborg:
> I don't know if I am doing the right thing by pushing small commits every time. Should I rebase before every push?

In general the final merged PR should not have any more commits than it had when it was first created. When changing the PR due to a request in the review the best is to amend the commit (using the `--amend` flag) that introduced the problem and then force push the branch. Alternatively you can create fix up commits (using the `--fixup`) flag. This will create new commits but it's clear that they are fix up commits intended to be squashed at the end before merging the PR.

src/gc/impl/conservative/gc.d
```diff
@@ -1820,6 +1835,8 @@ struct Gcx
         if (pool)
         {
             pool.initialize(npages, isLargeObject);
+            if (collectInProgress)
+                pool.mark.setAll();
```
leandro-lucarella-sociomantic: This is definitely not a problem for the concurrency/bit sharing, as this memory will not ever even be seen by the forked mark process as it has an old snapshot of the memory. This pool will not be marked until the next collection.

src/gc/impl/conservative/gc.d
leandro-lucarella-sociomantic: Right, I only now see the code commented here is in `smallAlloc()` and just the case bits could be updated while the mark process is also updating them :-/
I wonder if the set operation isn't just always atomic on x86. That would explain why it just works...

leandro-lucarella-sociomantic: Another way to deal with this that will decrease concurrency is to just wait for the forked process in this case and on allow allocating new pools while the mark is running (which sucks a bit), or fulfil allocations when the concurrent mark is running with new pools (marking somehow that a pool is new and hidden to the running mark process).

src/gc/impl/conservative/gc.d
```diff
@@ -2563,11 +2580,17 @@ struct Gcx
         return recoverPool[bin] = poolIndex < npools ? cast(SmallObjectPool*)pool : null;
     }
 
+    void disableFork() nothrow
+    {
+        config.fork = false;
+        markProcPid = 0;
```
leandro-lucarella-sociomantic: Thinking about the case where fork can be enabled and disabled freely at runtime, here it should be checked if a concurrent mark is running. I do not recommend allowing that though, not at least without a deep review of the code as the original implementation didn't consider this option.

src/gc/bits.d
leandro-lucarella-sociomantic: I think is a valid concern, but in any case it should be consistent (all with default or all without it).

src/gc/impl/conservative/gc.d
rainers: IIRC parallel marking didn't work without the atomic operations. These are also a bit more expensive, running the mark function with single threaded, but with atomics, is slower than the non-atomic version.

An alternative might be to keep track of what changed with respect to the freebits between the snapshot and its state when sweep is called. Maybe mark bits don't have to be set at all during allocation.

src/gc/impl/conservative/gc.d
rainers: If config is not modified by the GC but keeps its own flag, `config.fork` can just be checked before starting the next collection. That won't cause any racing conditions.

src/gc/impl/conservative/gc.d
leandro-lucarella-sociomantic: I'm still trying to get around why this have never been a problem in tango runtime with CDGC...

leandro-lucarella-sociomantic: > An alternative might be to keep track of what changed with respect to the freebits between the snapshot and its state when sweep is called. Maybe mark bits don't have to be set at all during allocation.

Yeah, that sounds like a good idea, freebits shouldn't be changed until the sweep is done and by then the mark process is already gone.

rainers: > I'm still trying to get around why this have never been a problem in tango runtime with CDGC...

Allocating during the forked scan is quite different from the parallel marking. If the collection is triggered by a small allocation, all free pages in SmallObjectPools are used and there are probably not that many objects on the free lists anymore (though that might differ with the application). That means that new allocations most often go into new pools which are never scanned by the fork.

IIRC your long running applications also don't allocate that much after the startup phase, so an allocation hitting the same mark-bits as whats currently modifed by the fork is pretty rare.

rainers: > Yeah, that sounds like a good idea, freebits shouldn't be changed until the sweep is done and by then the mark process is already gone.

freebits are now updated by GC.free and GC.malloc, so I guess a combination of its state when sweep is executing and the copy in the forked memory needs to be used to determine whether an object must be collected.

src/gc/impl/conservative/gc.d
FraMecca: staticError creates a deadlock in any case.
As you can see my `onConfigurationError` called during gc_init calls the traceHandler that needs the GC again. We can work around this issue with @nogc exceptions I believe or by aborting.

```
(gdb) bt 
#0  0x00007ffff7bc7c31 in __GI___nanosleep (requested_time=0x7fffffffdb00, remaining=0x7fffffffdb10) at ../sysdeps/unix/sysv/linux/nanosleep.c:28
#1  0x00005555557da1f4 in _D4core6thread6Thread5sleepFNbNiSQBf4time8DurationZv (val=...) at src/core/thread.d:1311
#2  0x000055555586653e in core.internal.spinlock.SpinLock.yield(ulong) shared (this=..., k=1015) at src/core/internal/spinlock.d:57
#3  0x00005555558664ba in core.internal.spinlock.SpinLock.lock() shared (this=...) at src/core/internal/spinlock.d:39
#4  0x000055555586187c in gc_init () at src/gc/proxy.d:55
#5  0x000055555584d121 in gc_init_nothrow () at src/gc/proxy.d:90
#6  0x000055555581e028 in _D2gc4impl5protoQo7ProtoGC6mallocMFNbmkxC8TypeInfoZPv (this=0x555555b3c5e0 <internal>, 
    ti=0x555555b49a30 <ClassInfo for core.runtime.defaultTraceHandler(void*).DefaultTraceInfo>, bits=0, size=1064) at src/gc/impl/proto/gc.d:104
#7  0x00005555557dcf7b in gc_malloc (sz=1064, ba=0, ti=0x555555b49a30 <ClassInfo for core.runtime.defaultTraceHandler(void*).DefaultTraceInfo>) at src/gc/proxy.d:165
#8  0x00005555557d92fb in core.memory.GC.malloc(ulong, uint, const(TypeInfo)) (ti=0x555555b49a30 <ClassInfo for core.runtime.defaultTraceHandler(void*).DefaultTraceInfo>, 
    ba=0, sz=1064) at src/core/memory.d:399
#9  0x00005555557dfbca in _d_newclass (ci=0x555555b49a30 <ClassInfo for core.runtime.defaultTraceHandler(void*).DefaultTraceInfo>) at src/rt/lifetime.d:94
#10 0x000055555584b0d9 in core.runtime.defaultTraceHandler(void*) (ptr=0x0) at src/core/runtime.d:983
#11 0x000055555582022f in _d_traceContext (ptr=0x0) at src/rt/dmain2.d:295
#12 0x000055555581ffeb in _d_createTrace (o=0x7ffff7fec410, context=0x0) at src/rt/deh.d:24
#13 0x00005555557de91d in _d_throwdwarf (o=0x7ffff7fec410) at src/rt/dwarfeh.d:219
#14 0x0000555555874e97 in onConfigurationError (msg=..., file=..., line=170) at src/core/exception.d:625
#15 0x00005555558679f0 in _D2gc4impl12conservativeQw14ConservativeGC6__ctorMFZCQBzQBzQBxQCiQBn (this=0x555555b54910) at src/gc/impl/conservative/gc.d:170
#16 0x0000555555867783 in _D2gc4impl12conservativeQw10initializeFZC4coreQBs11gcinterface2GC () at src/gc/impl/conservative/gc.d:133
#17 0x0000555555865546 in _D4core2gc8registry16createGCInstanceFAyaZCQBpQBn11gcinterface2GC (name=...) at src/core/gc/registry.d:64
#18 0x00005555558618c4 in gc_init () at src/gc/proxy.d:61
#19 0x000055555584d121 in gc_init_nothrow () at src/gc/proxy.d:90
#20 0x000055555581e028 in _D2gc4impl5protoQo7ProtoGC6mallocMFNbmkxC8TypeInfoZPv (this=0x555555b3c5e0 <internal>, ti=0x0, bits=0, size=48) at src/gc/impl/proto/gc.d:104
#21 0x00005555557dcf7b in gc_malloc (sz=48, ba=0, ti=0x0) at src/gc/proxy.d:165
#22 0x00005555557d92fb in core.memory.GC.malloc(ulong, uint, const(TypeInfo)) (ti=0x0, ba=0, sz=48) at src/core/memory.d:399
#23 0x00005555557dfb2d in _d_allocmemory (sz=48) at src/rt/lifetime.d:60
#24 0x000055555573a310 in D main (args=...) at dustmite.d:99
```

src/gc/impl/conservative/gc.d
FraMecca: I don;t think it can because it needs to be set when:
* lowMem
* tryAllocNewPool fails

we could avoid minimizing after a failed tryAllocNewPool  and call minimize() only on lowMem (at least for the forking GC only)

src/core/exception.d
```diff
@@ -19,6 +19,31 @@ void __switch_errorT()(string file = __FILE__, size_t line = __LINE__) @trusted
         assert(0, "No appropriate switch clause found");
 }
 
+
+/**
+ * Thrown on a configuration error.
+ */
+class ConfigurationError : Error
```
rainers: I'm not sure `ConfigurationError` and `ForkError` have to go into this generic module. For example, `ThreadErrror` is defined locally in `core.thread`. Do you expect someone catching this specific type so a definition has to be public?

src/gc/impl/conservative/gc.d
```diff
@@ -34,6 +34,9 @@ module gc.impl.conservative.gc;
 
 /***************************************************/
 version = COLLECT_PARALLEL;  // parallel scanning
+version (Posix)
+    version = COLLECT_FORK;
+version = COLLECT_FORK;
```
rainers: Unconditional version?

src/gc/impl/conservative/gc.d
```diff
@@ -163,7 +167,10 @@ class ConservativeGC : GC
 
     this()
     {
+        import core.exception : onConfigurationError;
         //config is assumed to have already been initialized
+        if (config.fork && config.parallel > 0)
+            onConfigurationError("Can't mix threads and fork strategies for GC marking");
```
rainers: Is this actually an existing limitation? I can think of use cases where a limited number of threads in the marking process can be helpful.

src/gc/impl/conservative/gc.d
```diff
@@ -1746,7 +1773,9 @@ struct Gcx
         assert(pool);
 
         debug(PRINTF) printFreeInfo(&pool.base);
+        pool.mark.set(pn);
```
rainers: add `if (collectInProgress)`.

src/gc/impl/conservative/gc.d
```diff
@@ -1746,7 +1773,9 @@ struct Gcx
...
         usedLargePages += npages;
+        pool.freepages -= npages;
```
rainers: This already happens elsewhere.

src/gc/impl/conservative/gc.d
```diff
@@ -2557,17 +2588,133 @@ struct Gcx
...
+        // was triggered by an allocation, the caller should allocate more memory
+        // to fulfill the request).
+        // If the collection should block, the parent will wait for the mark phase
+        // to finish before returning control to the mutator,
+        // but other threads are restarted and may run in parallel with the mark phase
+        // (unless they allocate or use the GC themselves, in which case
+        // the global GC lock will stop them).
+        // fork now and sweep later
+        import core.stdc.stdio;
+        version (COLLECT_FORK)
```
rainers: function already in `version(COLLECT_FORK)`

src/gc/impl/conservative/gc.d
```diff
@@ -2557,17 +2588,133 @@ struct Gcx
...
+                    markAll!markConservative(nostack);
+                _Exit(0);
+                return ChildStatus.done; // bogus
+            default: // the parent
+                thread_resumeAll();
+                if (!block)
+                {
+                    markProcPid = pid;
+                    // update profiling informations
+                    stop = currTime;
```
rainers: You can move the timing functions into the switch after the call to `markFork`, so the MonoTimes don't have to be passed around.

src/gc/impl/conservative/gc.d
rainers: Maybe it just needs a better name, e.g. `minimizeAfterNextCollection`.

src/gc/impl/conservative/gc.d
```diff
@@ -2985,7 +3183,10 @@ struct Pool
         topAddr = baseAddr + poolsize;
         auto nbits = cast(size_t)poolsize >> shiftBy;
 
-        mark.alloc(nbits);
+        version (COLLECT_FORK)
+            mark.alloc(nbits, true);
```
rainers: shared memory only needed if `config.fork`.

src/gc/impl/conservative/gc.d
rainers: Remove this line or replace the freebits.setRange call above with it.

src/gc/os.d
```diff
@@ -58,24 +101,33 @@ else
...
-else static if (is(typeof(valloc)))
-    version = GC_Use_Alloc_Valloc;
-else static if (is(typeof(malloc)))
-    version = GC_Use_Alloc_Malloc;
-else static assert(false, "No supported allocation methods available.");
-+/
 
 static if (is(typeof(VirtualAlloc))) // version (GC_Use_Alloc_Win32)
 {
+    /+
```
rainers: This /++/ comment should not be moved.

src/gc/impl/conservative/gc.d
```diff
@@ -3090,10 +3292,10 @@ struct Pool
         {
             freebits.Dtor();
         }
-        finals.Dtor();
-        structFinals.Dtor();
-        noscan.Dtor();
-        appendable.Dtor();
+        finals.Dtor(false);
```
rainers: if allocations don't explicitly set the default parameter, Dtor shouldn't do either.

src/core/gc/config.d
```diff
@@ -15,6 +15,7 @@ __gshared Config config;
 struct Config
 {
     bool disable;            // start disabled
+    bool fork = false;       // optional concurrent behaviour
```
rainers: I think you should enable it temporarily so the CI tests run with the concurrent GC. It should be restored to `false` before merging, though.

src/core/exception.d
FraMecca: The problem is that throw staticError is private

src/gc/impl/conservative/gc.d
rainers: > As you can see my `onConfigurationError` called during gc_init calls the traceHandler that needs the GC again.

That totally defeats the purpose of staticError ;-( It is only tricking the compiler...

src/gc/impl/conservative/gc.d
FraMecca: Why do you want me to remove that line? 
It leaks without it while setRange generates a sigsegv because it does a different thing.
I can put it behind an else statement so that it gets called for smallPools only.

src/gc/impl/conservative/gc.d
rainers: > That totally defeats the purpose of staticError

...and of "`@nogc` exceptions". In the case of the configuration check, you can just abort similar to https://github.com/dlang/druntime/blob/master/src/gc/proxy.d#L67

src/gc/impl/conservative/gc.d
rainers: freebits is empty in the else case, so it does the same as setRange for small object pools.

src/gc/impl/conservative/gc.d
FraMecca: I am not convinced, but we will let the user profile and decide. https://github.com/dlang/druntime/pull/2604/commits/c734c8ef8f5f311f893111ace6a2523f099ea403  allows mixed forked  + threaded marking

FraMecca:
I don't understand this failure: https://ci.appveyor.com/project/greenify/druntime/builds/25353377

gcbits:58 should be called only if HaveFork == true but that should not be the case in version (GC_Use_Alloc_Valloc)

rainers:
> gcbits:58 should be called only if HaveFork == true but that should not be the case in version (GC_Use_Alloc_Valloc)

You have to check `HaveFork` at compile time with some sort of `static if`, ut that doesn't work without some code duplication. e.g.
```
      static if (!HaveFork)
            free(data);
      else if (share)
            os_mem_unmap(data, nwords * data[0].sizeof);
      else
            free(data);
```

FraMecca:
Can someone help me understand how to interpret CI output?

wilzbach:
```
Test runnable/testthread2.d failed.  The logged output:
/home/braddr/sandbox/at-client/pull-3681891-FreeBSD_32/dmd/generated/freebsd/release/32/dmd -conf= -m32 -Irunnable    -odgenerated/runnable -ofgenerated/runnable/testthread2_0  runnable/testthread2.d 
generated/runnable/testthread2_0
[138371652, 1346320920, 2711460388, 3487484534, 3286390291, 92039947, 2088217342, 73546287, 1618245274, 953686636, 207639124, 3307404891, 1507434922, 3663444359, 1462875166, 3551546083, 290356273, 4211913830, 449149495, 1325090547, 2439770995, 3122431783, 561555678, 4264589494, 2795522631, 2388022856, 304027360, 845144253, 3667492552, 2175687532, 1381129205, 3886292630, 2880536606, 3299431039, 2632444204, 3554571102, 2652372728, 3673550711, 3964278397, 3246761200, 3250794532, 2618754634, 2902747942, 251039008, 2380700020, 2633673947, 2482211443, 2781636899, 2534048199, 390867864, 4226847282, 1925798820, 2454294092, 1562007011, 2827464829, 2942554616, 2938194178, 3624392628, 2023283336, 468101438, 86919302, 3554389615, 2486928834, 297791070, 397928802, 1069166792, 795525543, 1834111481, 3868221356, 2881111267, 1713175229, 3567727569, 2921585279, 2672259882, 2922407810, 1120680078, 728843246, 3314812051, 4068035906, 881341291, 139849402, 3149775520, 1470710775, 2639826168, 1931055016, 3208506906, 2835955837, 3108246182, 2487557185, 1529896354, 3505295511, 189565634, 895898068, 2595472399, 4158931534, 2672980030, 3556660285, 4075346136, 3988973329, 2912291374, 1281133677, 3272304413, 3807509134, 3023101326, 3252687490, 3773514963, 1531440219, 2686650854, 1950686686, 3298881277, 386548307, 2024031000, 3519963296, 4092377959, 2969908817, 67067764, 1402930284, 1706636187, 3630683154, 2006833032, 666781489, 926893934, 1432646510, 1122988325, 229123000, 505818918, 3744275748, 4271777893, 931746827, 3731408382, 3752075790, 3519997775, 2128419623, 2605565214, 987285610, 3166739277, 962199865, 628212802, 1198274761, 3341010049, 1236323061, 4076024248, 723346167, 2318056145, 2558939330, 877755931, 347590327, 651248204, 2738143870, 566253835, 808717022, 1704494110, 2724086208, 1383111892, 2545157784, 2969030469, 3093499121, 1018833547, 1641560931, 1573344145, 3796862986, 2931132537, 2645186826, 2099279920, 1935855396, 2523171194, 2531727621, 2111627168, 2404143132, 2212134324, 1779365892, 3934121997, 1880250711, 2477320402, 3568960867, 3722652327, 1068246321, 939121957, 216211719, 2019232731, 1989571165, 2289064, 2588869441, 1821388472, 3076339253, 3357545356, 3392592226, 2467997465, 526625089, 1229494826, 763908786, 1562026143, 925583598, 3446742677, 2700308686, 2060843518, 479542310, 1271936953, 3017562763, 2852171024, 687696792, 808487611, 3048662860, 3402511995, 1765279671, 3716218750, 2713299618, 3060152145, 2043624707, 3542896948, 1934287324, 538988951, 2607244805, 2499588261, 4048111334, 477936648, 375689955, 1687988612, 728739076, 1434284427, 2002609671, 3960189991, 356948306, 552631789, 4102980339, 408460523, 2445402931, 3937727917, 2855958506, 3441378969, 1496499422, 1274792669, 1684637588, 1493399738, 4032298329, 1176446082, 494147158, 982679050, 1029076740, 3245240440, 2378264753, 1304974433, 1054558331, 3795777477, 4033319801, 2810842589, 1925620401, 745049852, 2328365291, 2119135830, 993284020, 983377592, 3270679176, 3427379203, 2214815055, 1399932619, 1396114677, 4164731899, 166152439, 2759837449, 3177064104, 2042371371, 3456129216, 4020877772, 3806746928, 852466788, 138107246, 1016344745, 2820371070, 2069232264, 3773279933, 2094504813, 1351435408, 2498313613, 2586235284, 3679300225, 1249401323, 2534067741, 3610589784, 3771175447, 1604226920, 155995627, 1374642183, 1403892753, 50340233, 650886879, 480597540, 3139352910, 2365212515, 2270181884, 1771253780, 1049804426, 1248791447, 3791546838, 4128940376, 3695705754, 3058942890, 4115584004, 2281066530, 2342111575, 4159714909, 3756443772, 2052902799, 2398762798, 4132756931, 1749251849, 288206272, 3139750337, 1452280478, 4249248955, 3642730401, 4263907933, 300497152, 244678442, 2164026793, 3368015349, 1747765291, 1544449919, 2323540608, 3449351061, 3318249449, 1235299827, 1236869516, 534102031, 2150952037, 1946608821, 3846947004, 156796149, 3664332868, 3622749249, 3548690471, 3447540347, 1216344119, 3240960213, 2431686059, 3147836326, 3689849998, 2615977993, 3933841258, 3086013446, 3162176977, 1050392402, 3259747682, 3402079606, 453260489, 1724385317, 3726922254, 3561158417, 2331657149, 903543699, 1955992812, 3350213799, 3129599431, 1204806503, 2318278125, 970369480, 4280033002, 3398549591, 3740035835, 2506961611, 1228803346, 3011503365, 2319111903, 3601980161, 3899242283, 3187878251, 3873898064, 1226401497, 743252024, 3349838751, 487509351, 1509188604, 739764339, 1841367190, 3813769138, 1199623489, 2072678384, 2782883444, 1204391930, 3109726012, 3208552641, 2459543794, 1646113327, 1056555477, 1104545259, 2111877489, 3918236631, 304690649, 345631574, 1256922725, 3885663848, 2488965730, 1332210908, 361612355, 2046395415, 211693939, 770707719, 1655708829, 2939655947, 2058678862, 666037554, 1335844298, 3186209461, 353436614, 2614104782, 2404501637, 2826248280, 3435756378, 3051786868, 1488398128, 2815568694, 1150186075, 3767727797, 1565209903, 3312443516, 2078515121, 3092356233, 2687736714, 2206272388, 3451095967, 771942997, 3191390545, 1385032477, 1512092073, 568039569, 1040438580, 3553643745, 694161997, 230588118, 240157320, 1255571122, 3911534029, 4004003616, 959441669, 3191915453, 2571924956, 3488499032, 1933228649, 402672349, 3668971997, 2283776492, 1246690548, 169179620, 3866413803, 2573910932, 3840264714, 4009425696, 1868670313, 3409289008, 1386176620, 300022949, 3683089620, 4261074772, 1156161787, 2190620814, 2164480297, 2085716726, 2731644417, 679186770, 3891109066, 1929618794, 887791588, 318184067, 240559328, 618001685, 682463979, 1251454967, 3861828502, 193496807, 618628155, 4175882665, 2556933782, 3036566456, 3882610844, 39546331, 3131935229, 2046057013, 1048592423, 2803950363, 934998461, 2854526632, 1419349482, 2456988835, 1675528853, 2822440635, 1371577468, 2676016229, 4189043309, 1670990688, 1419956678, 3060311380, 143461375, 615505769, 1465427520, 1671751610, 409718796, 825394119, 2462281580, 790192379, 3273680885, 3237370927, 610916032, 3211022031, 89771975, 3281782510, 1107045942, 3000395867, 3892460028, 804166183, 4146579820, 1092175526, 916767101, 3093960408, 3179254196, 1566948118, 2179231536, 1362771296, 4229271054, 1779222407, 865527432, 3274443983, 916936933, 3362323891, 2522511626, 2134515831, 2544623182, 667599532, 939607456, 3006332225, 1216328503, 1409552225, 2615679319, 2587970146, 939756251, 364100779, 3627278440, 2992425029, 3665263057, 837728816, 3769940557, 2666448889, 3238911891, 2408516594, 2666695598, 163478341, 3541305025, 1283441703, 3511082549, 285174722, 2503982968, 3741120720, 3888236707, 558722433, 1207933700, 1339519109, 4148502007, 562448803, 3441908486, 240033264, 1463071700, 1249868764, 3412014390, 1513170806, 399538901, 287156469, 1032861713, 2310599217, 2609554753, 596067147, 1604365668, 722282794, 2181896486, 2504987025, 582798431, 3215159013, 1943389115, 1538775668, 572130555, 3074024543, 4092446552, 2266864712, 1920584866, 857471990, 2941150571, 2521740888, 212362386, 4160701822, 1031162481, 1994394857, 193945768, 2671763257, 680372839, 2109149454, 3011421043, 1738436180, 1159419791, 1939004532, 334072677, 3544069687, 747965543, 1905884981, 497966970, 4108807250, 609584731, 1273230708, 2584171923, 39388224, 821275485, 822026658, 2823686094, 344252759, 3982112233, 3413971639, 1529294133, 3152003976, 1134032003, 680308773, 3679394010, 1101495907, 3104952029, 2476986534, 422404660, 1636043437, 479465007]
0
core.exception.AssertError@runnable/testthread2.d(94): Assertion failure
----------------
??:? _d_assertp [0x286287ae]
??:? ??? [0x80583a8]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll().__lambda1() [0x28658e92]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x28658cd5]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll() [0x28658df0]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x28658cd5]
??:? _d_run_main [0x28658c36]
??:? pure nothrow immutable(char)[] std.conv.toImpl!(immutable(char)[], int).toImpl(int, uint, std.ascii.LetterCase).toStringRadixConvert!(24u).toStringRadixConvert(uint) [0x80621af]
??:? ??? [0x8058038]
??:? ??? [0x8057f17]
??:? ??? [0x0]
```

You can reproduce this locally by building DMD and then using `./test/run.d runnable/testthread2.d`

ZombineDev:
One relatively simple to reproduce issue is this failure in one the popular projects in the ecosystem:
https://github.com/CyberShadow/ae/blob/6913c6757192f7a3d7145e72dbbc3ecc42533f29/sys/memory.d#L105

rainers:
> One relatively simple to reproduce issue is this failure in one the popular projects in the ecosystem:
> https://github.com/CyberShadow/ae/blob/6913c6757192f7a3d7145e72dbbc3ecc42533f29/sys/memory.d#L105

This looks as if the explicit collection doesn't wait for the termination of the collection started with the same call?

src/gc/impl/conservative/gc.d
```diff
@@ -1224,6 +1228,12 @@ struct Gcx
     auto rangesLock = shared(AlignedSpinLock)(SpinLock.Contention.brief);
     Treap!Root roots;
     Treap!Range ranges;
+    private bool minimizeAfterNextCollection = false;
```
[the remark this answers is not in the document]
FraMecca: sorry, that was a missus. I removed "private"

src/core/gc/config.d
```diff
@@ -39,6 +40,7 @@ struct Config
 
         printf("GC options are specified as white space separated assignments:
     disable:0|1    - start disabled (%d)
+    fork:0|1       - set fork behaviour (disabled by default) (%d)
```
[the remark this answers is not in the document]
FraMecca: Honestly I don't have a preference, if someone comes up with a better description I would change it.

FraMecca:
Can someone help me understand how to interpret CI output?

> > One relatively simple to reproduce issue is this failure in one the popular projects in the ecosystem:
> > https://github.com/CyberShadow/ae/blob/6913c6757192f7a3d7145e72dbbc3ecc42533f29/sys/memory.d#L105
> 
> This looks as if the explicit collection doesn't wait for the termination of the collection started with the same call?

I had this discussion with Leandro already:
should GC.collect block when called explicitely? Or should we provide a default block=true parameter on GC.collect?
I would go with the first.

thewilsonator:
> Can someone help me understand how to interpret CI output?

[This line](https://github.com/dlang/druntime/blob/000b0e8862a21d2754ae1389da4134dc6fea12c2/test/exceptions/Makefile#L72) fails.

FraMecca:
> > Can someone help me understand how to interpret CI output?
> 
> [This line](https://github.com/dlang/druntime/blob/000b0e8862a21d2754ae1389da4134dc6fea12c2/test/exceptions/Makefile#L72) fails.
> 
> Presumably the output doesn't contain `InvalidMemoryOperationError`.
> 
> I'd suggest running `generated/$OS/release/64/invalid_memory_operation 2>&1 1>/dev/null` and seeing what it prints and ensuring that it fails.

I'll try to explain the reason for this fail.

Standard stop the world version:
0. the program enters main
1. the gc allocates 
2. a collections gets issued
3. nothing gets marked
4. the program returns from main
5. the RT issues a collection
6. the object that would throw on the destructor gets marked
7. RT throws InvalidMemoryOperation 

fork() based version:
0. the program enters main
1. the gc allocates 
2. a collections gets issued
3. the program returns from main
4. nothing gets marked
5. the RT issues a collection, it sees that the child has just terminated
6. the GC collects using mark bits from the first collection, the object was not marked and the destructor is not called.
7. RT does not throw

The intended behaviour of the fork() based GC is to check if a child is running the mark phase and proceed with a collection if the child has terminated. This means that there could be some allocations that could be freed but wont because the outlived the child mark phase and died just before the gc goes into collect() with the previous mark bits.

This behaviour however is not optimal at termination and I could either:
1. ignore the child (disableFork()) but I think that is not desiderable because it wastes the results from the child and we could have a zombie
2. wait for the child (block=true, as it currently is) and issue a second and final collection. This is doable with a variable that signals that this is the last collection.

FraMecca:
> ```
> Test runnable/testthread2.d failed.  The logged output:
> /home/braddr/sandbox/at-client/pull-3681891-FreeBSD_32/dmd/generated/freebsd/release/32/dmd -conf= -m32 -Irunnable    -odgenerated/runnable -ofgenerated/runnable/testthread2_0  runnable/testthread2.d 
> generated/runnable/testthread2_0
> [138371652, 1346320920, 2711460388, 3487484534, 3286390291, 92039947, 2088217342, 73546287, 1618245274, 953686636, 207639124, 3307404891, 1507434922, 3663444359, 1462875166, 3551546083, 290356273, 4211913830, 449149495, 1325090547, 2439770995, 3122431783, 561555678, 4264589494, 2795522631, 2388022856, 304027360, 845144253, 3667492552, 2175687532, 1381129205, 3886292630, 2880536606, 3299431039, 2632444204, 3554571102, 2652372728, 3673550711, 3964278397, 3246761200, 3250794532, 2618754634, 2902747942, 251039008, 2380700020, 2633673947, 2482211443, 2781636899, 2534048199, 390867864, 4226847282, 1925798820, 2454294092, 1562007011, 2827464829, 2942554616, 2938194178, 3624392628, 2023283336, 468101438, 86919302, 3554389615, 2486928834, 297791070, 397928802, 1069166792, 795525543, 1834111481, 3868221356, 2881111267, 1713175229, 3567727569, 2921585279, 2672259882, 2922407810, 1120680078, 728843246, 3314812051, 4068035906, 881341291, 139849402, 3149775520, 1470710775, 2639826168, 1931055016, 3208506906, 2835955837, 3108246182, 2487557185, 1529896354, 3505295511, 189565634, 895898068, 2595472399, 4158931534, 2672980030, 3556660285, 4075346136, 3988973329, 2912291374, 1281133677, 3272304413, 3807509134, 3023101326, 3252687490, 3773514963, 1531440219, 2686650854, 1950686686, 3298881277, 386548307, 2024031000, 3519963296, 4092377959, 2969908817, 67067764, 1402930284, 1706636187, 3630683154, 2006833032, 666781489, 926893934, 1432646510, 1122988325, 229123000, 505818918, 3744275748, 4271777893, 931746827, 3731408382, 3752075790, 3519997775, 2128419623, 2605565214, 987285610, 3166739277, 962199865, 628212802, 1198274761, 3341010049, 1236323061, 4076024248, 723346167, 2318056145, 2558939330, 877755931, 347590327, 651248204, 2738143870, 566253835, 808717022, 1704494110, 2724086208, 1383111892, 2545157784, 2969030469, 3093499121, 1018833547, 1641560931, 1573344145, 3796862986, 2931132537, 2645186826, 2099279920, 1935855396, 2523171194, 2531727621, 2111627168, 2404143132, 2212134324, 1779365892, 3934121997, 1880250711, 2477320402, 3568960867, 3722652327, 1068246321, 939121957, 216211719, 2019232731, 1989571165, 2289064, 2588869441, 1821388472, 3076339253, 3357545356, 3392592226, 2467997465, 526625089, 1229494826, 763908786, 1562026143, 925583598, 3446742677, 2700308686, 2060843518, 479542310, 1271936953, 3017562763, 2852171024, 687696792, 808487611, 3048662860, 3402511995, 1765279671, 3716218750, 2713299618, 3060152145, 2043624707, 3542896948, 1934287324, 538988951, 2607244805, 2499588261, 4048111334, 477936648, 375689955, 1687988612, 728739076, 1434284427, 2002609671, 3960189991, 356948306, 552631789, 4102980339, 408460523, 2445402931, 3937727917, 2855958506, 3441378969, 1496499422, 1274792669, 1684637588, 1493399738, 4032298329, 1176446082, 494147158, 982679050, 1029076740, 3245240440, 2378264753, 1304974433, 1054558331, 3795777477, 4033319801, 2810842589, 1925620401, 745049852, 2328365291, 2119135830, 993284020, 983377592, 3270679176, 3427379203, 2214815055, 1399932619, 1396114677, 4164731899, 166152439, 2759837449, 3177064104, 2042371371, 3456129216, 4020877772, 3806746928, 852466788, 138107246, 1016344745, 2820371070, 2069232264, 3773279933, 2094504813, 1351435408, 2498313613, 2586235284, 3679300225, 1249401323, 2534067741, 3610589784, 3771175447, 1604226920, 155995627, 1374642183, 1403892753, 50340233, 650886879, 480597540, 3139352910, 2365212515, 2270181884, 1771253780, 1049804426, 1248791447, 3791546838, 4128940376, 3695705754, 3058942890, 4115584004, 2281066530, 2342111575, 4159714909, 3756443772, 2052902799, 2398762798, 4132756931, 1749251849, 288206272, 3139750337, 1452280478, 4249248955, 3642730401, 4263907933, 300497152, 244678442, 2164026793, 3368015349, 1747765291, 1544449919, 2323540608, 3449351061, 3318249449, 1235299827, 1236869516, 534102031, 2150952037, 1946608821, 3846947004, 156796149, 3664332868, 3622749249, 3548690471, 3447540347, 1216344119, 3240960213, 2431686059, 3147836326, 3689849998, 2615977993, 3933841258, 3086013446, 3162176977, 1050392402, 3259747682, 3402079606, 453260489, 1724385317, 3726922254, 3561158417, 2331657149, 903543699, 1955992812, 3350213799, 3129599431, 1204806503, 2318278125, 970369480, 4280033002, 3398549591, 3740035835, 2506961611, 1228803346, 3011503365, 2319111903, 3601980161, 3899242283, 3187878251, 3873898064, 1226401497, 743252024, 3349838751, 487509351, 1509188604, 739764339, 1841367190, 3813769138, 1199623489, 2072678384, 2782883444, 1204391930, 3109726012, 3208552641, 2459543794, 1646113327, 1056555477, 1104545259, 2111877489, 3918236631, 304690649, 345631574, 1256922725, 3885663848, 2488965730, 1332210908, 361612355, 2046395415, 211693939, 770707719, 1655708829, 2939655947, 2058678862, 666037554, 1335844298, 3186209461, 353436614, 2614104782, 2404501637, 2826248280, 3435756378, 3051786868, 1488398128, 2815568694, 1150186075, 3767727797, 1565209903, 3312443516, 2078515121, 3092356233, 2687736714, 2206272388, 3451095967, 771942997, 3191390545, 1385032477, 1512092073, 568039569, 1040438580, 3553643745, 694161997, 230588118, 240157320, 1255571122, 3911534029, 4004003616, 959441669, 3191915453, 2571924956, 3488499032, 1933228649, 402672349, 3668971997, 2283776492, 1246690548, 169179620, 3866413803, 2573910932, 3840264714, 4009425696, 1868670313, 3409289008, 1386176620, 300022949, 3683089620, 4261074772, 1156161787, 2190620814, 2164480297, 2085716726, 2731644417, 679186770, 3891109066, 1929618794, 887791588, 318184067, 240559328, 618001685, 682463979, 1251454967, 3861828502, 193496807, 618628155, 4175882665, 2556933782, 3036566456, 3882610844, 39546331, 3131935229, 2046057013, 1048592423, 2803950363, 934998461, 2854526632, 1419349482, 2456988835, 1675528853, 2822440635, 1371577468, 2676016229, 4189043309, 1670990688, 1419956678, 3060311380, 143461375, 615505769, 1465427520, 1671751610, 409718796, 825394119, 2462281580, 790192379, 3273680885, 3237370927, 610916032, 3211022031, 89771975, 3281782510, 1107045942, 3000395867, 3892460028, 804166183, 4146579820, 1092175526, 916767101, 3093960408, 3179254196, 1566948118, 2179231536, 1362771296, 4229271054, 1779222407, 865527432, 3274443983, 916936933, 3362323891, 2522511626, 2134515831, 2544623182, 667599532, 939607456, 3006332225, 1216328503, 1409552225, 2615679319, 2587970146, 939756251, 364100779, 3627278440, 2992425029, 3665263057, 837728816, 3769940557, 2666448889, 3238911891, 2408516594, 2666695598, 163478341, 3541305025, 1283441703, 3511082549, 285174722, 2503982968, 3741120720, 3888236707, 558722433, 1207933700, 1339519109, 4148502007, 562448803, 3441908486, 240033264, 1463071700, 1249868764, 3412014390, 1513170806, 399538901, 287156469, 1032861713, 2310599217, 2609554753, 596067147, 1604365668, 722282794, 2181896486, 2504987025, 582798431, 3215159013, 1943389115, 1538775668, 572130555, 3074024543, 4092446552, 2266864712, 1920584866, 857471990, 2941150571, 2521740888, 212362386, 4160701822, 1031162481, 1994394857, 193945768, 2671763257, 680372839, 2109149454, 3011421043, 1738436180, 1159419791, 1939004532, 334072677, 3544069687, 747965543, 1905884981, 497966970, 4108807250, 609584731, 1273230708, 2584171923, 39388224, 821275485, 822026658, 2823686094, 344252759, 3982112233, 3413971639, 1529294133, 3152003976, 1134032003, 680308773, 3679394010, 1101495907, 3104952029, 2476986534, 422404660, 1636043437, 479465007]
> 0
> core.exception.AssertError@runnable/testthread2.d(94): Assertion failure
> ----------------
> ??:? _d_assertp [0x286287ae]
> ??:? ??? [0x80583a8]
> ??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll().__lambda1() [0x28658e92]
> ??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x28658cd5]
> ??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll() [0x28658df0]
> ??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x28658cd5]
> ??:? _d_run_main [0x28658c36]
> ??:? pure nothrow immutable(char)[] std.conv.toImpl!(immutable(char)[], int).toImpl(int, uint, std.ascii.LetterCase).toStringRadixConvert!(24u).toStringRadixConvert(uint) [0x80621af]
> ??:? ??? [0x8058038]
> ??:? ??? [0x8057f17]
> ??:? ??? [0x0]
> ```
> 
> You can reproduce this locally by building DMD and then using `./test/run.d runnable/testthread2.d`

I cannot reproduce this. Maybe it is related to FreeBSD

src/gc/impl/conservative/gc.d
```diff
@@ -43,6 +45,7 @@ import core.gc.gcinterface;
 import rt.util.container.treap;
 
 import cstdlib = core.stdc.stdlib : calloc, free, malloc, realloc;
+import core.stdc.stdio : fflush;
```
rainers: This import seems unnecessary, there is already a local import.

src/gc/impl/conservative/gc.d
```diff
@@ -2557,17 +2585,124 @@ struct Gcx
...
+        // to fulfill the request).
+        // If the collection should block, the parent will wait for the mark phase
+        // to finish before returning control to the mutator,
+        // but other threads are restarted and may run in parallel with the mark phase
+        // (unless they allocate or use the GC themselves, in which case
+        // the global GC lock will stop them).
+        // fork now and sweep later
+        import core.stdc.stdio : fflush;
+        import core.stdc.stdlib : _Exit;
+        fflush(null); // avoid duplicated FILE* output
```
rainers: I wonder if `fflush` is necessary here at all as writing a number of file buffers to disk can be an expensive operation. According to the documentation of `_Exit` no cleanup is done, and as no FILE i/o is happening in the child process there should be no duplications. The only exception are debug `printf`s, so maybe restrict calling `fflush` with the respective conditions or just on `stdout`?

src/gc/impl/conservative/gc.d
```diff
@@ -2557,17 +2585,124 @@ struct Gcx
...
+                return ChildStatus.done; // bogus
+            default: // the parent
+                thread_resumeAll();
+                if (!block)
+                {
+                    markProcPid = pid;
+                    return ChildStatus.running;
+                }
+                ChildStatus r = wait_pid(pid); // block until marking is done
+                assert(r == ChildStatus.done);
```
rainers: With the unsafe semantics of assert as proposed by Walter, this assert could allow the optimizer to remove the error-conditional below. Better just remove it or put it into an else branch of the conditional.

rainers:
> should GC.collect block when called explicitely? Or should we provide a default block=true parameter on GC.collect?
> I would go with the first.

I agree, it should just block. If there is no concurrent thread running, it could even avoid the fork and collect immediately.
That would probably lead to a problem in combination with parallel marking: the additional threads will be created in the parent process, so they are assumed to exist in the child process after forking, too, but only the "main" thread is actually alive. I think that would still work, but without marking in parallel.

rainers:
> I cannot reproduce this. Maybe it is related to FreeBSD

hospital.d seems another candidate that I have seen in a number of test failures. But there are others in druntime unittest and elsewhere, too.

Maybe try using atomic operations for the mark bits as discussed, i.e. `mark.setLocked` and `mark!(precise, true)`.

FraMecca:
> > I cannot reproduce this. Maybe it is related to FreeBSD
> 
> hospital.d seems another candidate that I have seen in a number of test failures. But there are others in druntime unittest and elsewhere, too.
> 
> Maybe try using atomic operations for the mark bits as discussed, i.e. `mark.setLocked` and `mark!(precise, true)`.

parallel is enabled by default so CI test should run using markParallel that uses mark.setLocked.

In any case, I can't reproduce  the  failures on my machine so I don't know how to proceed with debugging CI issues.

rainers:
> parallel is enabled by default so CI test should run using markParallel that uses mark.setLocked.

But the `mark.set` calls in concurrent allocations are not atomic,

rainers:
> In any case, I can't reproduce the failures on my machine so I don't know how to proceed with debugging CI issues.

I've setup a VM with Ubuntu 19 and observe that the std.stdio unittest is running into a ForkError on termination. Maybe this has to do with the unittest also using fork to delete temporary files.

FraMecca:
> > In any case, I can't reproduce the failures on my machine so I don't know how to proceed with debugging CI issues.
> 
> I've setup a VM with Ubuntu 19 and observe that the std.stdio unittest is running into a ForkError on termination. Maybe this has to do with the unittest also using fork to delete temporary files.

This prompted me to open a PR to phobos https://github.com/dlang/phobos/pull/7111
but also to recognize the fact that the user may catch the GC's child termination erroneously so we should ignore ECHILD and acknowledge the end of the concurrent mark phase

rainers:
I can now reproduce failures by executing `./run.d -f runnable/hospital.d` a couple of times in ddm/test.

The failures seem to go away if I disable either forking or parallel marking. Changing the single call to pool.mark.set during allocation to setLocked didn't help.

src/core/gc/config.d
```diff
@@ -39,6 +40,7 @@ struct Config
 
         printf("GC options are specified as white space separated assignments:
     disable:0|1    - start disabled (%d)
+    fork:0|1       - set fork behaviour (%d)
     profile:0|1|2  - enable profiling with summary when terminating program (%d)
     gc:".ptr, disable, profile);
```
rainers: `fork` must be passed as an argument here, too.

src/core/gc/config.d
```diff
@@ -39,8 +40,9 @@ struct Config
 
         printf("GC options are specified as white space separated assignments:
     disable:0|1    - start disabled (%d)
+    fork:0|1       - set fork behaviour (%d)
     profile:0|1|2  - enable profiling with summary when terminating program (%d)
-    gc:".ptr, disable, profile);
+    gc:".ptr, disable, profile, fork);
```
rainers: That's the wrong order, now. It's in the middle in the format string.

src/gc/impl/conservative/gc.d
```diff
@@ -3078,7 +3279,7 @@ struct Pool
             bPageOffsets = null;
         }
 
-        mark.Dtor();
+        mark.Dtor(config.fork);
```
rainers: Nit: this is an unconditional call, while `mark.alloc` uses `version(COLLECT_FORK)`. They should be consistent, probably simpler to leave the evaluation of the  `share` parameter to `GCbits` and remove the version condition for `mark.alloc`..

rainers:
I made a couple more experiments with hospital.d:
- using static linkage instead of shared seems to reduce the probability of a failure (2 out of 100 runs instead of about 10 out of 100).
- using less cores for parallel marking also reduces the chances of failure.
- adding logging during collection (COLLECT_PRINTF) shows differences between runs, but that is to be expected given the concurrency.
- adding logging to the allocation (PRINTF) seems to make the issue go away. I suspect that's because the slowdown causes less concurrent accesses.
- using setLocked instead of set doesn't make a notable difference.

BTW: the failure shows as a small difference in the results, not a crash. Correct run:
```
Patients: 102515
Time:     33730654
Visits:   106371
```
Failed run:
```
Patients: 102515
Time:     33730469
Visits:   106370

rainers:
If this passes and we disable `fork` in the default configuration, it would be good to add some test coverage by running the unittests with the forking GC., See test/gc/Makefile for similar runs.

wilzbach:
Ping @FraMecca. Any chance you can bring this to the finish line?

FraMecca:
@wilzbach I am still working from time to time in understanding the source of the bug.
Right now I can't allocate much time to it, if you need to close it do it, I'll reopen when I have a solution.

If there is something I can do in the meantime let me know.

Ambyjkl:
Happy Birthday to this PR :(. @FraMecca any luck with the bug?

RazvanN7:
I'm going to close this since #2604 was merged.

thewilsonator:
#2604 is this PR.. you mean #3514

RazvanN7:
Yes, I misread the link description.

PR #2607, opened 2019-05-15
Rework of -profile=gc calculations

RazvanN7:
This is a revamp of #1846 . I'm having a hard time understanding the results of the test cases added in the original PR. Any help would be appreciated.

cc @leandro-lucarella-sociomantic

Edit: It looks like there is a missing call to resetThreadStats that causes the tests to fail. Can anyone help?

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -172,7 +172,7 @@ struct GC
         size_t freeSize;
         /// number of bytes allocated for current thread since program start or
         /// call to `GC.resetThreadLocalStats()`
-        size_t allocatedInCurrentThread;
+        ulong allocatedInCurrentThread;
     }
 
     /**
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -140,6 +140,7 @@ private
     extern (C) BlkInfo_ gc_query( void* p ) pure nothrow;
     extern (C) GC.Stats gc_stats ( ) nothrow @nogc;
     extern (C) GC.ProfileStats gc_profileStats ( ) nothrow @nogc @safe;
+    extern (C) void gc_resetThreadLocalStats( ) nothrow @nogc;
 
     extern (C) void gc_addRoot( in void* p ) nothrow @nogc;
     extern (C) void gc_addRange( in void* p, size_t sz, const TypeInfo ti = null ) nothrow @nogc;
@@ -169,6 +170,9 @@ struct GC
         size_t usedSize;
         /// number of free bytes on the GC heap (might only get updated after a collection)
         size_t freeSize;
+        /// number of bytes allocated for current thread since program start or
+        /// call to `GC.resetThreadLocalStats()`
+        size_t allocatedInCurrentThread;
     }
 
     /**
@@ -706,6 +710,15 @@ struct GC
     static Stats stats() nothrow
     {
         return gc_stats();
+    }
+
+    /**
+     * Resets runtime stats for currently active GC implementation of current thread
+     * See `core.memory.GC.Stats` for list of available metrics.
+     */
+    static void resetThreadLocalStats() nothrow
+    {
+        gc_resetThreadLocalStats();
     }
 
     /**
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -79,7 +79,7 @@ __gshared long extendTime;
 __gshared long otherTime;
 __gshared long lockTime;
 
-size_t bytesAllocated;   // thread local counter
+ulong bytesAllocated;   // thread local counter
 
 private
 {
@@ -397,7 +397,7 @@ class ConservativeGC : GC
             alloc_size = size;
         }
         gcx.leakDetector.log_malloc(p, size);
-        .bytesAllocated += size;
+        bytesAllocated += alloc_size;
 
         debug(PRINTF) printf("  => p = %p\n", p);
         return p;
@@ -1116,12 +1116,12 @@ class ConservativeGC : GC
 
         stats.usedSize -= freeListSize;
         stats.freeSize += freeListSize;
-        stats.allocatedInCurrentThread = .bytesAllocated;
+        stats.allocatedInCurrentThread = bytesAllocated;
     }
 
     void resetThreadLocalStats() nothrow @nogc
     {
-        .bytesAllocated = 0;
+        bytesAllocated = 0;
     }
 }
 
@@ -4379,7 +4379,7 @@ version (D_LP64) unittest
     {
         // only run if the system has enough physical memory
         size_t sz = 2L^^32;
-        import core.stdc.stdio;
+        //import core.stdc.stdio;
         //printf("availphys = %lld", os_physical_mem());
         if (os_physical_mem() > sz)
         {
@@ -4396,8 +4396,7 @@ version (D_LP64) unittest
             auto nstats = GC.stats();
             assert(nstats.usedSize == stats.usedSize);
             assert(nstats.freeSize == stats.freeSize);
-            //printf("%d, %d\n", nstats.allocatedInCurrentThread, stats.allocatedInCurrentThread);
-            //assert(nstats.allocatedInCurrentThread == stats.allocatedInCurrentThread);
+            assert(nstats.allocatedInCurrentThread - sz == stats.allocatedInCurrentThread);
         }
     }
 }
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -79,6 +79,8 @@ __gshared long extendTime;
 __gshared long otherTime;
 __gshared long lockTime;
 
+size_t bytesAllocated;   // thread local counter
+
 private
 {
     extern (C)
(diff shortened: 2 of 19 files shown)


src/gc/impl/conservative/gc.d
```diff
@@ -4370,7 +4381,7 @@ version (D_LP64) unittest
     {
         // only run if the system has enough physical memory
         size_t sz = 2L^^32;
-        //import core.stdc.stdio;
+        import core.stdc.stdio;
```
thewilsonator: Left on after debugging?

src/rt/profilegc.d
```diff
@@ -41,24 +42,26 @@ __gshared
 
 extern (C) void profilegc_setlogfilename(string name)
 {
-    logfilename = name;
+    logfilename = name ~ "\0";
```
thewilsonator: Why this?

src/rt/profilegc.d
```diff
@@ -72,14 +75,19 @@ public void accumulate(string file, uint line, string funcname, string type, siz
     buffer[type.length + 1 + funcname.length + 1 + file.length] = ':';
     buffer[type.length + 1 + funcname.length + 1 + file.length + 1 ..
            type.length + 1 + funcname.length + 1 + file.length + 1 + buflen] = buf[0 .. buflen];
+    buffer[length] = 0;
```
thewilsonator: Why null terminate things?

rainers:
As noted in the original PR, I'm not a fan of this command line switch. Here's an example how to implement proper and extended tracing with a proxy GC: https://github.com/rainers/visuald/blob/dmdserver/tools/tracegc.d#L231

src/core/memory.d
```diff
@@ -708,6 +712,15 @@ struct GC
         return gc_stats();
     }
 
+    /**
+     * Resets runtime stats for currently active GC implementation of current thread
+     * See `core.memory.GC.Stats` for list of available metrics.
+     */
+    static void resetThreadLocalStats() nothrow
```
rainers: A reset function is never a good idea with independent clients checking the same value. Just use difference to a previous value.

src/gc/impl/conservative/gc.d
```diff
@@ -395,6 +398,8 @@ class ConservativeGC : GC
             alloc_size = size;
         }
         gcx.leakDetector.log_malloc(p, size);
+        bytesAllocated += alloc_size;
```
rainers: Thread locals can be slow. Versioning this under a compile time condition might be acceptable.

src/rt/profilegc.d
leandro-lucarella-sociomantic: Because they are passed to C functions.

leandro-lucarella-sociomantic:
> I'm having a hard time understanding the results of the test cases added in the original PR. Any help would be appreciated.

The test is just allocating some memory and then verified the profiler got the right places and the amount of memory allocated is reported correctly. If you have more details on what needs clarification, please let me know.

Also ping @mihails-strasuns just in case you have some time.

leandro-lucarella-sociomantic:
> As noted in the original PR, I'm not a fan of this command line switch. Here's an example how to implement proper and extended tracing with a proxy GC: https://github.com/rainers/visuald/blob/dmdserver/tools/tracegc.d#L231

Interesting.

src/core/memory.d
leandro-lucarella-sociomantic: The problem with this is detecting overflows of the counter for long lived processes. But it might be acceptable anyway. I have the feeling there were other reasons for the reset, but I can't remember right now.

src/gc/impl/conservative/gc.d
leandro-lucarella-sociomantic: Compile time is not a good option IMHO, because then you can't use this feature with a normal compiler. Two solutions come in mind. One is doing the proxy trick and providing some sort of flag to pick the GC you want to use, the tracing GC being one of the options. I don't remember if that mechanism is already in place and if currently any other GC is provided by the official distribution. Another option would be to do a simple runtime `if (tracing) bytesAllocated += alloc_size;`, that should be fast enough for all users.

leandro-lucarella-sociomantic:
Thanks @RazvanN7 for picking this up!

RazvanN7:
Thanks for the reply @leandro-lucarella-sociomantic . What I don't understand regarding the tests is the amount that is reported to be allocated. For exmample, the first line for linux 64 in the expected results file says:

```
288	              1	immutable(char)[][int] D main src/profilegc.d:23
```

Looking at line 23, we have:

```d
string[int] aa = [1:"one", 2:"two", 3:"three"];
```
This implementation reports 464 bytes allocated instead 288 for this particular line. I have no idea why this difference occurs; I don't even know why it should report 288; as far as I can tell, it might be possible that the number represents the sum of previous allocations + the current one; it is very likely that a reset should have occurred but it didn't, however I do not know what can/should cause resets. Any information on this is welcome.

mihails-strasuns:
@RazvanN7 naturally this test case is extremely fragile - it has to check for allocation size which is implementation defined and thus test needs to be updated each time implementation changes. It may also differ between platforms even at a single point of time (thus separate expected outputs per platform).

In case of AA - most likely it used to allocate exactly 288 bytes at the point of implementing original PR, all these numbers are taken from actual GC reports and not calculated manually. It is easily possible that something has changed in druntime/compiler since than and now the expected amount is different.

RazvanN7:
So, simply tweaking the test is an acceptable option?

mihails-strasuns:
Sure, it must match what GC actually does.

RazvanN7:
This seems to be green now, so it will probably be merged soon.

@leandro-lucarella-sociomantic what other stats are you interested in? After this is merged, can we hop on implementing other stats or should be try to come up with a generic design that would potentially expose anything that the GC has to offer?

RazvanN7:
Windows failure is unrelated.

src/gc/impl/conservative/gc.d
andralex: The cost is there but should be evaluated relative to the surrounding time. I recall I measured a while ago and an allocation took several thousand cycles even in the best case. The additional overhead would be difficult to measure.

src/rt/profilegc.d
andralex: namely `fopen`

leandro-lucarella-sociomantic:
> @leandro-lucarella-sociomantic what other stats are you interested in? After this is merged, can we hop on implementing other stats or should we try to come up with a generic design that would potentially expose anything that the GC has to offer?

Thanks a lot @RazvanN7 !

I recently notice dmd 2.085 already have some improvements on that front! https://dlang.org/phobos/core_memory.html#.GC.ProfileStats

Probably the most important stats for us are `numCollections` and `totalPauseTime`. I just created an issue in ocean so we can make use of it: https://github.com/sociomantic-tsunami/ocean/issues/711.

As I mentioned in https://github.com/dlang/druntime/pull/1846#issuecomment-386563504, more info could be provided (allocation info) but at this point this is not a blocker for us, just a nice to have.

The other pending issue we have is about `GC.stats` being O(n) instead of O(1) (like in D1): https://github.com/dlang/druntime/pull/2164

The trade off for this is introducing bugs in the calculation if much easier if calculations are kept up to date all the time (but I think that also happens with the profiling info, so I'm not sure now.

src/core/memory.d
rainers: The counter will overflow just the same as the difference between two consecutive calls.

src/gc/impl/conservative/gc.d
rainers: I did a short measurement with the vdparser benchmark, adding the TLS counter costs about 1% for Win64 (which probably has one of the faster implementation of TLS), though that measurement was done on a mobile processor with disabled speedstepping. Thats probably also consistent with the full allocation taking about 1000 cycles, given the additional instructions generated.

rainers:
> Here's an example how to implement proper and extended tracing with a proxy GC: https://github.com/rainers/visuald/blob/dmdserver/tools/tracegc.d#L231

For reference, here is a skeleton that works for hooking the conservative GC with a tracing proxy: https://gist.githubusercontent.com/rainers/6cdf73b48837defb9f88/raw/7e776acd806f44668e5d615a17bdac20047fec64/tracegc.d 
One downside of that approach is that you have to analyze the callstack to find the allocating call. That needs to be done anyway for all calls in phobos/druntime that have not been compiled with -profile=gc.

src/core/memory.d
andralex: One advantage for applications is they can reset the counter and then just look at it later. If they can't reset, they need to save the old value then subtract it from the current value.

rainers: It's still a bad API especially with multiple clients. Do you also reset the system clock if you are doing benchmarks?

leandro-lucarella-sociomantic: @rainers has a point. The overflowing is still passing quite some complexity to the client but it's true there are workarounds to detect it and deal with it.

leandro-lucarella-sociomantic:
I see this was finally merged, thanks again for picking it up and pushing for it.

About the reset, I agree with @rainers that, even when convenient for most use cases, having a reset is a bad API design and we should consider removing it before a first release.

wilzbach:
> we should consider removing it before a first release.

The next release is on 1.7 - with the first beta appearing on 15.6.

leandro-lucarella-sociomantic:
https://github.com/dlang/druntime/pull/2629

PR #2608, opened 2019-05-15
Add __RefCount struct

edi33416:
This PR adds the `__RefCount` struct which is designed to be composed inside any user defined type that desires reference counting.

A simple example of usage

```
struct TestRC {   
        private __RefCount rc; 
        int[] payload;

        this(int sz) {
                rc = __RefCount(1);
                payload = (cast(int*) malloc(sz * int.sizeof))[0 .. sz];
        }

        void opAssign(ref TestRC rhs) {
                if (rc.isUnique) free(payload.ptr);
                
                rc = rhs.rc;
                payload = rhs.payload;
        }

        /* Implement copy ctors */

        ~this() {
                if (rc.isUnique) free(payload.ptr);
        }
}
```

### DIFF
--- a/src/core/experimental/refcount.d
+++ b/src/core/experimental/refcount.d
@@ -1,107 +1,7 @@
 module core.experimental.refcount;
 
-private struct StatsAllocator
+struct _RefCount
 {
-    version(CoreUnittest) size_t bytesUsed;
-
-    @trusted @nogc nothrow pure
-    void* allocate(size_t bytes) shared
-    {
-        import core.memory : pureMalloc;
-        if (!bytes) return null;
-
-        auto p = pureMalloc(bytes);
-        if (p is null) return null;
-        enum alignment = size_t.sizeof;
-        assert(cast(size_t) p % alignment == 0);
-
-        version (CoreUnittest)
-        {
-            static if (is(typeof(this) == shared))
-            {
-                import core.atomic : atomicOp;
-                atomicOp!"+="(bytesUsed, bytes);
-            }
-            else
-            {
-                bytesUsed += bytes;
-            }
-        }
-        return p;
-    }
-
-    @system @nogc nothrow pure
-    bool deallocate(void[] b) shared
-    {
-        import core.memory : pureFree;
-        assert(b !is null);
-
-        version (CoreUnittest)
-        {
-            static if (is(typeof(this) == shared))
-            {
-                import core.atomic : atomicOp;
-                assert(atomicOp!">="(bytesUsed, b.length));
-                atomicOp!"-="(bytesUsed, b.length);
-            }
-            else
-            {
-                assert(bytesUsed >= b.length);
-                bytesUsed -= b.length;
-            }
-        }
-        pureFree(b.ptr);
-        return true;
-    }
-
-    static shared StatsAllocator instance;
-}
-
-version (CoreUnittest)
-{
-    private shared StatsAllocator allocator;
-
-    private @nogc nothrow pure @trusted
-    void* pureAllocate(size_t n)
-    {
-        return (cast(void* function(size_t) @nogc nothrow pure)(&_allocate))(n);
-    }
-
-    private @nogc nothrow @safe
-    void* _allocate(size_t n)
-    {
-        return allocator.allocate(n);
-    }
-
-    private @nogc nothrow pure
-    void* pureDeallocate(T)(T[] b)
-    {
-        return (cast(void* function(T[]) @nogc nothrow pure)(&_deallocate!(T)))(b);
-    }
-
-    private @nogc nothrow
-    void* _deallocate(T)(T[] b)
-    {
-        allocator.deallocate(b);
-        return null;
-    }
-}
-
-struct __RefCount
-{
-    version(CoreUnittest) {} else
-    {
-        import core.memory : pureMalloc, pureFree;
-
-        private alias pureAllocate = pureMalloc;
-
-        @nogc nothrow pure
-        private static void* pureDeallocate(T)(T[] b)
-        {
-            pureFree(b.ptr);
-            return null;
-        }
-    }
     import core.atomic : atomicOp;
 
     alias CounterType = uint;
@@ -221,7 +121,7 @@ struct __RefCount
     // } Get an immutable obj
 
     @nogc nothrow pure @safe scope
-    ref __RefCount opAssign(return scope ref typeof(this) rhs) return
+    ref _RefCount opAssign(return scope ref typeof(this) rhs) return
     {
         if (rhs.isInitialized() && rc == rhs.rc)
         {
@@ -283,7 +183,7 @@ struct __RefCount
     pure nothrow @safe @nogc scope
     bool isUnique() const
     {
-        assert(isInitialized(), "[__RefCount.isUnique] __RefCount is uninitialized");
+        assert(isInitialized(), "[_RefCount.isUnique] _RefCount is uninitialized");
         return !!rcOp!"=="(1);
     }
 
@@ -305,11 +205,11 @@ unittest
 {
     () @safe @nogc pure nothrow
     {
-        __RefCount a = __RefCount(1);
+        _RefCount a = _RefCount(1);
         assert(a.isUnique);
-        const __RefCount ca = const __RefCount(1);
+        const _RefCount ca = const _RefCount(1);
         assert(ca.isUnique);
-        immutable __RefCount ia = immutable __RefCount(1);
+        immutable _RefCount ia = immutable _RefCount(1);
         assert(ia.isUnique);
 
         // A const reference will increase the ref count
@@ -338,14 +238,14 @@ unittest
         assert((() @trusted => *cast(int*)i_cp_c_cp_ia.getUnsafeValue() == 4)());
         assert((() @trusted => i_cp_c_cp_ia.getUnsafeValue() == c_cp_ia.getUnsafeValue())());
 
-        __RefCount t;
+        _RefCount t;
         assert(!t.isInitialized());
-        __RefCount t2 = t;
+        _RefCount t2 = t;
         assert(!t.isInitialized());
         assert(!t2.isInitialized());
     }();
 
-    assert(allocator.bytesUsed == 0, "__RefCount leakes memory");
+    assert(allocator.bytesUsed == 0, "_RefCount leakes memory");
 }
 
 version(CoreUnittest)
@@ -353,17 +253,17 @@ unittest
 {
     () @safe @nogc pure nothrow scope
     {
-        __RefCount a = __RefCount(1);
+        _RefCount a = _RefCount(1);
         assert(a.isUnique);
-        __RefCount a2 = a;
+        _RefCount a2 = a;
         assert((() @trusted => *cast(int*)a.getUnsafeValue() == 2)());
-        __RefCount a3 = __RefCount(1);
+        _RefCount a3 = _RefCount(1);
         a2 = a3;
         assert((() @trusted => *cast(int*)a.getUnsafeValue() == 1)());
         assert(a.isUnique);
     }();
 
-    assert(allocator.bytesUsed == 0, "__RefCount leakes memory");
+    assert(allocator.bytesUsed == 0, "_RefCount leakes memory");
 }
 
 version(CoreUnittest)
@@ -371,7 +271,7 @@ unittest
 {
     struct TestRC
     {
-        private __RefCount rc;
+        private _RefCount rc;
         int[] payload;
 
         @nogc nothrow pure @trusted scope
@@ -379,12 +279,12 @@ unittest
         {
             static if (is(Q == immutable))
             {
-                rc = immutable __RefCount(1);
+                rc = immutable _RefCount(1);
                 payload = (cast(immutable int*) pureAllocate(sz * int.sizeof))[0 .. sz];
             }
             else
             {
-                rc = __RefCount(1);
+                rc = _RefCount(1);
                 payload = (cast(int*) pureAllocate(sz * int.sizeof))[0 .. sz];
             }
         }
@@ -537,5 +437,103 @@ unittest
         t2 = t3;
     }();
 
-    assert(allocator.bytesUsed == 0, "__RefCount leakes memory");
+    assert(allocator.bytesUsed == 0, "_RefCount leakes memory");
+}
+
+version (CoreUnittest)
+{
+    private struct StatsAllocator
+    {
+        version(CoreUnittest) size_t bytesUsed;
+
+        @trusted @nogc nothrow pure
+        void* allocate(size_t bytes) shared
+        {
+            import core.memory : pureMalloc;
+            if (!bytes) return null;
+
+            auto p = pureMalloc(bytes);
+            if (p is null) return null;
+            enum alignment = size_t.sizeof;
+            assert(cast(size_t) p % alignment == 0);
+
+            version (CoreUnittest)
+            {
+                static if (is(typeof(this) == shared))
+                {
+                    import core.atomic : atomicOp;
+                    atomicOp!"+="(bytesUsed, bytes);
+                }
+                else
+                {
+                    bytesUsed += bytes;
+                }
+            }
+            return p;
+        }
+
+        @system @nogc nothrow pure
+        bool deallocate(void[] b) shared
+        {
+            import core.memory : pureFree;
+            assert(b !is null);
+
+            version (CoreUnittest)
+            {
+                static if (is(typeof(this) == shared))
+                {
+                    import core.atomic : atomicOp;
+                    assert(atomicOp!">="(bytesUsed, b.length));
+                    atomicOp!"-="(bytesUsed, b.length);
+                }
+                else
+                {
+                    assert(bytesUsed >= b.length);
+                    bytesUsed -= b.length;
+                }
+            }
+            pureFree(b.ptr);
+            return true;
+        }
+    }
+
+    private shared StatsAllocator allocator;
+
+    private @nogc nothrow pure @trusted
+    void* pureAllocate(size_t n)
+    {
+        return (cast(void* function(size_t) @nogc nothrow pure)(&_allocate))(n);
+    }
+
+    private @nogc nothrow @safe
+    void* _allocate(size_t n)
+    {
+        return allocator.allocate(n);
+    }
+
+    private @nogc nothrow pure
+    void* pureDeallocate(T)(T[] b)
+    {
+        return (cast(void* function(T[]) @nogc nothrow pure)(&_deallocate!(T)))(b);
+    }
+
+    private @nogc nothrow
+    void* _deallocate(T)(T[] b)
+    {
+        allocator.deallocate(b);
+        return null;
+    }
+}
+else
+{
+    import core.memory : pureMalloc, pureFree;
+
+    private alias pureAllocate = pureMalloc;
+
+    @nogc nothrow pure
+    private static void* pureDeallocate(T)(T[] b)
+    {
+        pureFree(b.ptr);
+        return null;
+    }
 }
--- a/src/core/experimental/refcount.d
+++ b/src/core/experimental/refcount.d
@@ -221,12 +221,11 @@ struct __RefCount
     // } Get an immutable obj
 
     @nogc nothrow pure @safe scope
-    ref __RefCount opAssign(return scope ref typeof(this) rhs)
+    ref __RefCount opAssign(return scope ref typeof(this) rhs) return
     {
         if (rhs.isInitialized() && rc == rhs.rc)
         {
-            return rhs;
-            //return this;
+            return this;
         }
         if (rhs.isInitialized())
         {
@@ -237,8 +236,7 @@ struct __RefCount
             delRef();
         }
         () @trusted { rc = rhs.rc; }();
-        return rhs;
-        //return this;
+        return this;
     }
 
     @nogc nothrow pure @safe scope
@@ -461,12 +459,11 @@ unittest
         // } Get an immutable obj
 
         @nogc nothrow pure @safe scope
-        ref TestRC opAssign(return ref typeof(this) rhs)
+        ref TestRC opAssign(return ref typeof(this) rhs) return
         {
             if (payload is rhs.payload)
             {
-                return rhs;
-                //return this;
+                return this;
             }
             if (rc.isInitialized && rc.isUnique)
             {
@@ -474,8 +471,7 @@ unittest
             }
             payload = rhs.payload;
             rc = rhs.rc;
-            return rhs;
-            //return this;
+            return this;
         }
 
         @nogc nothrow pure @trusted scope
--- a/src/core/experimental/refcount.d
+++ b/src/core/experimental/refcount.d
@@ -28,7 +28,7 @@ struct _RefCount
     }
 
     @nogc nothrow pure @trusted scope
-    this(this Q)(int) const
+    this(this Q)(int)
     {
         CounterType* support = cast(CounterType*) pureAllocate(2 * CounterType.sizeof);
         static if (is(Q == immutable))
--- a/src/core/experimental/refcount.d
+++ b/src/core/experimental/refcount.d
@@ -74,47 +74,21 @@ version (CoreUnittest)
     }
 
     private @nogc nothrow pure
-    void pureDeallocate(T)(T[] b)
+    void* pureDeallocate(T)(T[] b)
     {
-        return (cast(void function(T[]) @nogc nothrow pure)(&_deallocate!(T)))(b);
+        return (cast(void* function(T[]) @nogc nothrow pure)(&_deallocate!(T)))(b);
     }
 
     private @nogc nothrow
-    void _deallocate(T)(T[] b)
+    void* _deallocate(T)(T[] b)
     {
         allocator.deallocate(b);
+        return null;
     }
 }
 
 struct __RefCount
 {
-    private static struct __mutable(T)
-    {
-        private union
-        {
-            T _unused;
-            size_t _ref = cast(size_t) null;
-        }
-
-        @nogc nothrow pure @trusted
-        this(T val) const
-        {
-            _ref = cast(size_t) val;
-        }
-
-        @nogc nothrow pure @trusted
-        this(shared T val) const
-        {
-            _ref = cast(size_t) val;
-        }
-
-        @nogc nothrow pure @trusted
-        T unwrap() const
-        {
-            return cast(T) _ref;
-        }
-    }
-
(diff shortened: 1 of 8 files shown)


src/core/experimental/refcount.d
```diff
@@ -0,0 +1,545 @@
...
+
+    private @nogc nothrow
+    void* _deallocate(T)(T[] b)
+    {
+        allocator.deallocate(b);
+        return null;
+    }
+}
+
+struct __RefCount
```
jacob-carlborg: Why the `__` prefix?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    () @safe @nogc pure nothrow scope
+    {
+        enum numElem = 10;
+        auto t = TestRC(numElem);
+        assert(t.rc.isUnique);
+        const TestRC ct = const TestRC(numElem);
+        assert(ct.rc.isUnique);
+        immutable TestRC it = immutable TestRC(numElem);
```
rainers: What happens when a return value of a pure function is implicitely converted to immutable?

n8sh:
Doesn't something like this really belong in Phobos?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,545 @@
+module core.experimental.refcount;
```
JinShil: Please sufficiently document this module.  It very much helps with the review.

JinShil:
Where does this fit in the grander picture?  I see the tree, but what role does it play in the ecosystem of the forest.  Also, this appears to be a runtime lowering.  Is there a compiler PR being planned to so ref-counted objects are first-class citizens in the D language?

src/core/experimental/refcount.d
wilzbach: `__` is very problematic is it's reserved by the compiler and can lead to strange errors. `_` should be used for library symbols.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,545 @@
...
+        }
+        if (rhs.isInitialized())
+        {
+            rhs.addRef();
+        }
+        if (isInitialized())
+        {
+            delRef();
+        }
+        () @trusted { rc = rhs.rc; }();
```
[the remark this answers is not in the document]
edi33416: This is because `rhs` is `scope` and `this` is not. In hindsight, `rhs` probably should not be `scope` but it seems to be required by the L362 in the unittest `a2 = a3;`. Removing `scope` from `opAssign`'s `rhs` declaration yields the following error
```
refcount.d(362): Error: scope variable a3 assigned to non-scope parameter rhs calling core.experimental.refcount.__RefCount.opAssign
```

Maybe I should remove `scope` from the function attributes list?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,545 @@
...
+        return rhs;
+        //return this;
```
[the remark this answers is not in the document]
edi33416: Had the same reaction. It's related to `dip1000`
Returning `this` yields
```
refcount.d(228): Error: returning this escapes a reference to parameter this, perhaps annotate with return
```

This was my mistake as I poorly tried to annotate with return.
Doing
```
@safe return
ref __RefCount opAssign(...)
```
Will throw a bunch of compiler errors. The right way of annotating is
```
@safe
ref __RefCount opAssign(...) return
```
It's exactly how you'd annotate a `const` method

src/core/experimental/refcount.d
edi33416: Can you please give me an example to help me understand the problem?

edi33416:
> Doesn't something like this really belong in Phobos?

The reason why this is placed in druntime is because achieving reference counting with all the attributes that the language provides, is a difficult problem and we want to make it available to all our users.

We also desire to use it in `rcarray`, `rcstring` and the like.

src/core/experimental/refcount.d
edi33416: The prefix is to make the user aware that there is some `magic` going on under the hood in order to make it work.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,545 @@
...
+    import core.atomic : atomicOp;
+
+    alias CounterType = uint;
+    private CounterType* rc = null;
+
+    @nogc nothrow pure @safe scope
+    bool isShared() const
+    {
+        // Faster than ((cast(size_t) rc) % 8) == 0;
+        return !((cast(size_t) rc) & 7);
```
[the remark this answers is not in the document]
edi33416: I looked at dmd's `obj` output for this and it does convert to `& 7`. Did not know about this. Thank you.
Are you suggesting i should leave `% 8`?

One thing that I've noticed here is that with `-O` (all dmd, ldc2 and gdc) result in the same code.
Without it, dmd and gdc convert to `& 7` but it takes more ops. Interestingly, ldc2 doesn't convert (https://godbolt.org/z/QTeary)

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,545 @@
...
+        allocator.deallocate(b);
+        return null;
+    }
+}
+
+struct __RefCount
+{
+    version(CoreUnittest) {} else
+    {
+        import core.memory : pureMalloc, pureFree;
```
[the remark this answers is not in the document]
edi33416: So I don't have any top level imports. I had pay as you go (for the user) in mind, but since `__RefCount` is not a template `core.memory` will be opened. If in the future the compiler will be able to avoid opening the internal imports if the type is not used, then this will behave nicely without changes.

edi33416:
> Where does this fit in the grander picture? I see the tree, but what role does it play in the ecosystem of the forest. Also, this appears to be a runtime lowering. Is there a compiler PR being planned so ref-counted objects are first-class citizens in the D language?

Reference counting is a pressing issue in the D language and we want to have a very basic type that takes care of all the magic required to get reference counting to work with `pure` and `immutable` objects/functions.

There is no planned lowering, for now, that I am aware of.

src/core/experimental/refcount.d
rainers: If there is no specific compiler support, there is no magic. Please note that there are subtle differences in how the compiler treats these symbols, e.g. https://github.com/dlang/dmd/blob/master/src/dmd/traits.d#L1395

src/core/experimental/refcount.d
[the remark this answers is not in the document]
rainers: I think lazy importing will more likely happen for static or selective imports.

src/core/experimental/refcount.d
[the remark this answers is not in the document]
rainers: > Are you suggesting i should leave `% 8`?

`& 7` is fine and the comment is unnecessary IMO.

> ldc2 doesn't convert (https://godbolt.org/z/QTeary)

`% 8` and `& 7` are only the same for unsigned integers.

src/core/experimental/refcount.d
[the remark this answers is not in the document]
rainers: Removing scope doesn't work, but I'm no export on scope/return.

src/core/experimental/refcount.d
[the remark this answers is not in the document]
rainers: `const` before the function does the same, but `return` seems an exception. BTW: having modifiers both before and after the function declarations seems a bit odd.

src/core/experimental/refcount.d
rainers: Something like this
```
struct _RefCount { int* prc; }

struct S
{
	_RefCount rc;
	int payload;
}

S* createS(int x) pure
{
	S* s = new S;
	s.payload = x;
	return s;
}

void main()
{
	immutable(S)* ps = createS(3);
}
```

n8sh:
>The reason why this is placed in druntime is because achieving reference counting with all the attributes that the language provides, is a difficult problem and we want to make it available to all our users.

That's not a great reason in my opinion. Lots of non-trivial stuff is in Phobos, and Phobos is available to all users and has similar things like `std.typecons.RefCounted` which [works in betterC](https://github.com/dlang/phobos/blob/56438a49fc0b37cf740ef2e363c4ecd6e087f367/std/typecons.d#L6453-L6465).

>We also desire to use it in `rcarray`, `rcstring` and the like.

If those are going to be added to the D runtime then that would be a good reason. `rcarray` sounds a lot like the standard library's `std.container.array`, so I would similarly ask why that would be added to the D runtime instead of to the standard library.

src/core/experimental/refcount.d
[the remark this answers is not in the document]
thewilsonator: @edi33416 it does with `-O1`. @rainers correct, but the logical negation that precedes that makes that point moot.

src/core/experimental/refcount.d
andralex: @edi33416 please use double underscore. This is a special/magic type that _is_ meant to be recognized by the compiler as special, even if today it happens to not be. Thanks.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,782 @@
...
+            mixin(copyCtorIncRef);
+        }
+        // } Get a const obj
+
+        // { Get an immutable obj
+        @nogc nothrow pure @trusted scope
+        this(return scope ref typeof(this) rhs) immutable
+        {
+            // Can't have an immutable ref to a mutable. Create a new RC
+            rc = rhs.rc;
```
andralex: this should be `rc = __RefCount(1);`

n8sh:
I have concerns about promoting the use of `__RefCount` as a building-block for other types. By necessity it allocates storage for the counter separately from whatever resource is associated with the `__RefCount`. In a handwritten reference-counted data structure the storage for the payload and the storage for the counter could be allocated together and deallocated together. Using this `__RefCount` would increase the number of allocations and decrease locality of reference, while not doing much to decrease effort of implementation (compared to for example `std.typecons.RefCounted!T` which in addition to incrementing/decrementing a counter also calls a destructor when the counter hits zero and has optional auto-initialization).

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,782 @@
...
+        assert((() @trusted => i_cp_c_cp_ia.getUnsafeValue() == c_cp_ia.getUnsafeValue())());
+
+        __RefCount t;
+        assert(!t.isInitialized());
+        __RefCount t2 = t;
+        assert(!t.isInitialized());
+        assert(!t2.isInitialized());
+    }();
+
+    assert(allocator.bytesUsed == 0, "__RefCount leakes memory");
```
lesderid: leaked (2 more later)

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,782 @@
...
+     * This is unsafe and may lead to dangling pointers to invalid memory.
+     *
+     * Returns:
+     *      A raw pointer to the reference count.
+     *
+     * Complexity:
+     *      $(BIGOH 1).
+     */
+    pure nothrow @nogc @system
+    CounterType* getUnsafeValue() const
```
lesderid: Couldn't this return `const(CounterType*)`?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,782 @@
...
+            return cast(CounterType)(atomicOp!op(*(cast(shared CounterType*) rc), val));
+        }
+        else
+        {
+            mixin("return cast(CounterType)(*(cast(CounterType*) rc)" ~ op ~ "val);");
+        }
+    }
+
+    /**
+     * Creates a new `__RefCount` instance. It's memory is internally managed with
```
atilaneves: s/It's/Its/

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,782 @@
...
+     * malloc/free.
```
atilaneves: So no other allocators then? I know this is a druntime PR so it wouldn't be able to use std.experimental.allocator, but then I wonder why not put this in phobos.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,782 @@
...
+     *
+     * Params:
+     *      _ = an unused int value; required because structs don't have a
+     *          user defined default constructor
+     */
+    @nogc nothrow pure @trusted scope
+    this(this Q)(int)
+    {
+        /* We allocate a `size_t` chunk that will save as our support. We
```
atilaneves: I'm guessing this was supposed to be "will serve as our support"?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,782 @@
...
+    {
+        mixin(copyCtorIncRef);
+    }
+
+    /**
+     * Copy constructs a const `__RefCount` from an immutable reference, `rhs`.
+     * This increases the reference count.
+     */
+    @nogc nothrow pure @safe scope
+    this(return scope immutable ref typeof(this) rhs) const
```
atilaneves: Is there any reason all of these constructors aren't just a template?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,782 @@
...
+            mixin("return cast(CounterType)(*(cast(CounterType*) rc)" ~ op ~ "val);");
+        }
+    }
+
+    /**
+     * Creates a new `__RefCount` instance. It's memory is internally managed with
+     * malloc/free.
+     *
+     * Params:
+     *      _ = an unused int value; required because structs don't have a
```
atilaneves: Wouldn't a factory function be better?

atilaneves:
I would take all of the code out of druntime, compile it with ldc and check with asan that there are no leaks or other memory problems.

src/core/experimental/refcount.d
edi33416: It's placed in druntime as it might require some compiler "help" in the future. For now, all the functions that are called for a dec ref return a `void*`, ex `void* delRef() const`. This is so in the context of an `immutable` object the `pure const delRef` call won't be elided. Currently there is no magic involved, but If, in the future, the compiler will see that the returned value is never used and it decides to elide, we might need a helping hand.

src/core/experimental/refcount.d
edi33416: Can't use templates with the copy constructor.

src/core/experimental/refcount.d
edi33416: Something like `makeRefCount` and `makeSharedRefCount` ?

src/core/experimental/refcount.d
thewilsonator: > call won't be elided. 

Then I'd be making sure to thoroughly with LDC and GDC. Also in the case of dead object elimination (as opposed to just dead call) that is the desired outcome.

edi33416: > Also in the case of dead object elimination (as opposed to just dead call) that is the desired outcome.

Good point. I'd assume that in this case, since the object is "dead" it will elide the calls regardless of any other considerations, as the calls are made through the object

src/core/experimental/refcount.d
wilzbach: Isn't this a major design deficient in copy constructor? CC @RazvanN7

wilzbach:
> I would take all of the code out of druntime, compile it with ldc and check with asan that there are no leaks or other memory problems.

Excellent point, but we should better setup a CI step/task for this as otherwise future additions and improvements won't be checked.

src/core/experimental/refcount.d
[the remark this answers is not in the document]
edi33416: > `const` before the function does the same, but `return` seems an exception.

Having `const` before wouldn't mean I'm returning a `const ref __RefCount`?

>  BTW: having modifiers both before and after the function declarations seems a bit odd.

It looks a bit odd, but I prefer it to having all the attributes after.

src/core/experimental/refcount.d
edi33416: @rainers if you want to use `__RefCount` you shouldn't use pointers as you'll bypass calls to the copy ctors/opAssign.

> Also works with values instead of pointers.

Changing your code to
> ```
> struct _RefCount { int* prc; }
> 
> struct S
> {
> 	_RefCount rc;
> 	int payload;
>      this(int x) { rc = _RefCount(1); payload = x; }
> }
> 
> S createS(int x) pure
> {
> 	S s = S(x);
> 	return s;
> }
> 
> void main()
> {
> 	immutable(S) ps = createS(3);
> }
> ```

I didn't know this will implicitly convert to immutable.
This could be problematic.

If this calls the copy ctor, then it should be fine; otherwise we'll have a problem.

src/core/experimental/refcount.d
atilaneves: Yes. In the unittest it looks weird to pass 1 to a constructor just to get around no-arg ctors.

atilaneves:
> > I would take all of the code out of druntime, compile it with ldc and check with asan that there are no leaks or other memory problems.
> 
> Excellent point, but we should better setup a CI step/task for this as otherwise future additions and improvements won't be checked.

Good point there too. I'm not sure how to set up compiling druntime itself with asan though, and only for ldc.

src/core/experimental/refcount.d
atilaneves: Any such compiler help would require a DIP, no?

src/core/experimental/refcount.d
thewilsonator: > If this calls the copy ctor, then it should be fine; otherwise I think we'll have a problem.

It should be subject to named value return optimisation (which means moving I think).

src/core/experimental/refcount.d
edi33416: > Any such compiler help would require a DIP, no?

I believe so

wilzbach:
> I'm not sure how to set up compiling druntime itself with asan though, and only for ldc.

Well, yet another of those many reasons why we should have LDC/LLVM as official backend. Anyhow, that not being a short-term option, the easiest way would be to not compile full druntime (because of it's compiler magic that might be problematic), but only compile the specific __RefCount test(s) with `ldc` and `-fsanitize=address`?
LDC also has its own runtime build too (see e.g. https://wiki.dlang.org/Building_LDC_runtime_libraries), but I'm not sure how well this will work with master (e.g. name mangling changes in master etc.).

In other words:
- use Buildkite (we could also use other CIs, but we want to unite all CIs to Buildkite) and setup a new target (LDC runtime) - the existing DMD druntime target is [here](https://github.com/dlang/ci/blob/40031edf53a9994790455afe558fa66f8752c720/buildkite/build_project.sh#L263)
- install LDC (we typically use this [script](https://github.com/dlang/dmd/blob/166a623a6f881fc63b8edae6fc0ae53dc12d1791/ci.sh#L137))
- invoke the respective test script with `DMD=ldc2` (and thus let it add sanitization)

9il:
> This PR adds the `__RefCount` struct which is designed to be composed inside any user defined type that desires reference counting.

1. http://mir-algorithm.libmir.org/mir_rc_array.html
2. http://mir-algorithm.libmir.org/mir_rc_ptr.html

are based on the Mir's own RC context, which is more memory efficient and can be used with allocators. Please don't make the language and runtime depend on this type the way users would not be able to use their own type with the same experience. The docs shouldn't define it as a unique RC context for D, just a reference implementation.

src/core/experimental/refcount.d
jacob-carlborg: Would be nice with a usage example.

src/core/experimental/refcount.d
```diff
@@ -5,16 +5,16 @@
...
- * The user is responsible to initialize the `__RefCount` in the constructor of his
- * type. The user will call the `isUnique()` method to decide if this is the last
- * reference to his type so he can safely deallocate his own managed memory.
+ * `__RefCount` was designed to be composed as a field inside a user-defined type.
+ * The user is responsible to initialize the `__RefCount` in that type's constructors.
+ * The user will call the `isUnique()` method to decide if this is the last
+ * reference to the enclosing type so memory can be safely deallocated.
  *
- * `Important`: the `__RefCount` must be initialized through a call to the
+ * $(B Important:) the `__RefCount` member must be initialized through a call to it's
```
lesderid: s/it's/its/

JinShil:
I still don't see how this fits into the larger picture, probably because this PR doesn't explain what the long-term reference counting plan is.  Do users have to add this `__RefCount` field and supporting boiler-plate to their types any time they want to have reference-counted semantics?   Or is there some future compiler integration planned where the boilerplate is added by the compiler?   How is this different from Phobos's `RefCounted` type.  Can a `__RefCount` field also be added to classes?

src/core/experimental/refcount.d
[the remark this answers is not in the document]
andralex: >What happens when structs like TestRC are cast between mutable/shared/immutable?

Good food for thought. E.g. what if someone uses the `assumeUnique` idiom with an object that has a `__RefCount` as a member?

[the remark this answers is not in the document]
andralex: My take: should be transparent to casting, i.e. work as things have worked until now, which means (for this and other reasons) that `__RefCount` must always use interlocked operations :(.

src/core/experimental/refcount.d
andralex: This is a bummer. It means we need to always use interlocked inc/dec conservatively. I was hoping we'd do much better than C++'s `shared_ptr`.

andralex:
>Doesn't something like this really belong in Phobos?

This is a core type virtually part of the language definition.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,545 @@
...
+        {
+            mixin("return cast(CounterType)(*(cast(CounterType*) rc)" ~ op ~ "val);");
+        }
+    }
+
+    @nogc nothrow pure @trusted scope
+    this(this Q)(int) const
+    {
+        CounterType* support = cast(CounterType*) pureAllocate(2 * CounterType.sizeof);
+        static if (is(Q == immutable))
```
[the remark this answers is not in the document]
andralex: We'll eliminate the distinction altogether sadly.

andralex:
> I have concerns about promoting the use of `__RefCount` as a building-block for other types. By necessity it allocates storage for the counter separately from whatever resource is associated with the `__RefCount`. In a handwritten reference-counted data structure the storage for the payload and the storage for the counter could be allocated together and deallocated together. Using this `__RefCount` would increase the number of allocations and decrease locality of reference, while not doing much to decrease effort of implementation (compared to for example `std.typecons.RefCounted!T` which in addition to incrementing/decrementing a counter also calls a destructor when the counter hits zero and has optional auto-initialization).

We don't have a better solution, and we don't even know very well how to get this to work (regardless of the intrusive aspect and orthogonal to it). In the future we may add an intrusive version a la `__RefCountedIntrusive!(T, "member")` to use space lent by the host object. This is from the "I have a dream" cycle.

src/core/experimental/refcount.d
andralex: >So no other allocators then? I know this is a druntime PR so it wouldn't be able to use std.experimental.allocator, but then I wonder why not put this in phobos.

We wanted to get a minimal version going because we have these huge problems with purity and immutable etc. Worrying about custom memory allocation is additional distraction. Too many holes in the dam, too few fingers.

We must put this in druntime because that's where the refcounted slice needs to be (i.e. a slice that is just like `T[]` but manages its own memory).

src/core/experimental/refcount.d
andralex: > I still don't see how this fits into the larger picture, probably because this PR doesn't explain what the long-term reference counting plan is. Do users have to add this `__RefCount` field and supporting boiler-plate to their types any time they want to have reference-counted semantics? Or is there some future compiler integration planned where the boilerplate is added by the compiler? How is this different from Phobos's `RefCounted` type? Can a `__RefCount` field also be added to classes?

This is the basic mechanism for allowing reference counted objects usable with all attributes and qualifiers. Libraries can use it as a building block. There are no current plans for automatic integration in generated code, though that's a future possibility.

9il:
> > I have concerns about promoting the use of `__RefCount` as a building-block for other types. By necessity it allocates storage for the counter separately from whatever resource is associated with the `__RefCount`. In a handwritten reference-counted data structure the storage for the payload and the storage for the counter could be allocated together and deallocated together. Using this `__RefCount` would increase the number of allocations and decrease locality of reference, while not doing much to decrease effort of implementation (compared to for example `std.typecons.RefCounted!T` which in addition to incrementing/decrementing a counter also calls a destructor when the counter hits zero and has optional auto-initialization).
> 
> We don't have a better solution, and we don't even know very well how to get this to work (regardless of the intrusive aspect and orthogonal to it). In the future we may add an intrusive version a la `__RefCountedIntrusive!(T, "member")` to use space lent by the host object. This is from the "I have a dream" cycle.

It is cool that we would have RC support in DRuntime.

mir.rc uses intrusive RCs for pointers and arrays.

Please don't introduce a compiler/druntime changes that would optimize only DRuntime defined RC context. User-defined RC context implementation should have a clear way to get the optimizations as well, hope so.

src/core/experimental/refcount.d
thewilsonator: Pardon my ignorance, but why? It seems to me that moving a copy constructible should be no problem since the total number of instances stays the same and you only care about the refcount in a binary fashion, unless opPostMove messes with the refcount (why??) then all should be good.

edi33416: The issue is the implicit conversion to `immutable`.

The implementation allocates a buffer of two `uint`s: it uses the one aligned at 8 bytes for `immutable` and `shared` types, and the one aligned at 4 bytes for the rest. In order to check if we are dealing with a "possibly" `shared` reference we look at the alignment, and based on this we either do atomics or not.

In the pure factory function, we will create a non-shared `__RefCount`, so the internal counter will use the `uint` allocated at 4 bytes, thus marking the non-shared property. When this is returned and implicitly cast to `immutable`, the contract is broken.

The only solution to this, as far as we can tell, is to always assume `shared` and use atomics. This is what C++ had to do as well.

lesderid:
What's keeping this from getting merged? I realise that an intrusive approach might be preferable and it's a shame that we're currently stuck with using atomic ops for all instances (https://github.com/dlang/druntime/pull/2608#discussion_r288500379).

That said, this PR is targeting `core.experimental` and it's currently blocking progress on `rcarray` et al.

JinShil:
> I still don't see how this fits into the larger picture,

I understand this now after seeing https://github.com/dlang/druntime/pull/2646

n8sh:
>it's currently blocking progress on `rcarray` et al.

Perhaps this is a good candidate for using a development branch.

wilzbach:
> Perhaps this is a good candidate for using a development branch.

Unfortunately, they have never worked in the past. A core.experimental package is probably the only realistic chance we have.

RazvanN7:

PR #2609, opened 2019-05-16
fix Issue 19701 - undefined reference to _D6object__T6hashOf

rainers:
use CoreDdoc instead of D_Ddoc

the test case isn't so easy to reproduce as it relies on phobos.


thewilsonator:
Thanks!

PR #2611, opened 2019-05-17
Better formatting of unittest failure

andralex:
This is a step toward making unittest reporting better as discussed [here](https://forum.dlang.org/post/qbnamt$2h5h$1@digitalmars.com).

### DIFF
--- a/src/core/runtime.d
+++ b/src/core/runtime.d
@@ -630,23 +630,42 @@ extern (C) UnitTestResult runModuleUnitTests()
     UnitTestResult results;
     foreach ( m; ModuleInfo )
     {
-        if ( m )
-        {
-            auto fp = m.unitTest;
+        if ( !m )
+            continue;
+        auto fp = m.unitTest;
+        if ( !fp )
+            continue;
 
-            if ( fp )
+        import core.exception;
+        ++results.executed;
+        try
+        {
+            fp();
+            ++results.passed;
+        }
+        catch ( Throwable e )
+        {
+            import core.stdc.stdio;
+            printf("%.*s(%zu): [unittest] %.*s\n",
+                cast(int) e.file.length, e.file.ptr, e.line,
+                cast(int) e.message.length, e.message.ptr);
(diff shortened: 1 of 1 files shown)


andralex:
With this PR, the program:

```d
unittest
{
    assert(0, "Thou shall not pass!");
}
```
built with:
```
dmd -main -unittest -run test
```
prints:
```
test.d(3): [unittest] Thou shall not pass!
1/1 unittests FAILED
```

thewilsonator:
Please add that as a test.

wilzbach:
> With this PR, the program:

This ain't a test!! Why was this merged??

src/core/runtime.d
```diff
@@ -630,23 +630,42 @@ extern (C) UnitTestResult runModuleUnitTests()
...
     {
-        if ( m )
+        if ( !m )
+            continue;
+        auto fp = m.unitTest;
+        if ( !fp )
+            continue;
+
+        import core.exception;
+        ++results.executed;
```
wilzbach: Should have been a selective import :/

src/core/runtime.d
```diff
@@ -630,23 +630,42 @@ extern (C) UnitTestResult runModuleUnitTests()
...
+                auto moduleName = m.name;
+                if (moduleName.length && e.file.length > moduleName.length
+                    && e.file[0 .. moduleName.length] == moduleName)
                 {
-                    _d_print_throwable(e);
+                    // Exception originates in the same module, don't print
+                    // the stack trace.
+                    // TODO: omit stack trace only if assert was thrown
+                    // directly by the unittest.
+                    continue;
```
wilzbach: How did this get merged???

atilaneves: @wilzbach What do you think is wrong with it?

wilzbach: - Silent change of behavior without any discussion nor tests (people may have been relying on the stack trace getting printed) 
- For all other contributors we require to resolve TODOs as they will never be fixed otherwise

PR #2612, opened 2019-05-20
Add missing enum for FindFirstFileEx.

look-at-me:
https://docs.microsoft.com/en-us/windows/desktop/api/minwinbase/ne-minwinbase-findex_info_levels


CyberShadow:
> Windows Server 2008, Windows Vista, Windows Server 2003 and Windows XP:  This value is not supported until Windows Server 2008 R2 and Windows 7.

Probably why it wasn't there in the first place.

A more correct fix would be to wrap it in `static if (_WIN32_WINNT >= 0x601)`.

PR #2613, opened 2019-05-20
 Improve rt.util.container.common.initialize!T(ref T) if is(T == struct)

n8sh:
Move the zero-init check to compile time. Plus if the struct doesn't have elaborate assignment or an elaborate destructor we can use `t = T.init` instead of memset/memcpy. LDC optimizes memset to the same as `t = T.init` but for memcpy it is an improvement as the result of `typeid(T).initializer()` isn't a compile-time constant.

### DIFF
--- a/src/rt/util/container/common.d
+++ b/src/rt/util/container/common.d
@@ -45,10 +45,12 @@ void destroy(T)(ref T t) if (!is(T == struct))
 void initialize(T)(ref T t) if (is(T == struct))
 {
     import core.stdc.string;
-    if (auto p = typeid(T).initializer().ptr)
-        memcpy(&t, p, T.sizeof);
-    else
+    static if (__traits(isPOD, T)) // implies !hasElaborateAssign!T && !hasElaborateDestructor!T
+        t = T.init;
+    else static if (__traits(isZeroInit, T))
         memset(&t, 0, T.sizeof);
+    else
+        memcpy(&t, typeid(T).initializer().ptr, T.sizeof);
 }
 
 void initialize(T)(ref T t) if (!is(T == struct))


src/rt/util/container/common.d
```diff
@@ -45,10 +45,12 @@ void destroy(T)(ref T t) if (!is(T == struct))
 void initialize(T)(ref T t) if (is(T == struct))
 {
     import core.stdc.string;
-    if (auto p = typeid(T).initializer().ptr)
-        memcpy(&t, p, T.sizeof);
-    else
+    static if (__traits(isPOD, T)) // implies !hasElaborateAssign!T && !hasElaborateDestructor!T
+        t = T.init;
+    else static if (__traits(isZeroInit, T))
```
thewilsonator: shouldn't this come first?

n8sh: If both conditions are applicable then both would have the same result so the only question is efficiency. My idea is that DMD does better with `t = T.init` for small structs and LDC doesn't seem to be harmed by it. Take for instance:
https://run.dlang.io/is/II3tqe
```d
extern (C):
struct S
{
    int a;
}

void fillUsingAssignment(scope ref S s)
{
    s = S.init;
}

void fillUsingMemset(scope ref S s)
{
    import core.stdc.string : memset;
    memset(&s, 0, S.sizeof);
}

void main()
{
}
```

The generated ASM with `dmd -betterC -O`:

```
[...]
fillUsingAssignment:
		push	RBP
		mov	RBP,RSP
		mov	dword ptr [RDI],0
		pop	RBP
		ret
[...]
fillUsingMemset:
		push	RBP
		mov	RBP,RSP
		mov	EDX,4
		xor	ESI,ESI
		call	  memset@PLT32
		pop	RBP
		ret
		add	[RAX],AL
```
The risk is potentially worse generated code with DMD for very large structs but I believe those will be less common. And with LDC -O2 assignment and memset compile the same.

PR #2614, opened 2019-05-23
updated _IO_FILE of the glibc version

burner:
_IO_FILE aka FILE does not mirror the current version of glibc.
This PR fixes that. This came up through https://github.com/ldc-developers/ldc/issues/2782#issuecomment-495221392


burner:
thanks @jacob-carlborg and @thewilsonator

ZombineDev:
@burner FYI, you should have opened the pull request against `stable` if you want to be sure that this change would be part of LDC 1.16.

burner:
@ZombineDev thank you for that information. Next time!

PR #2615, opened 2019-05-26
Make Windows declarations usable on other platforms

CyberShadow:
Many Windows API headers contain declarations which are useful on other platforms. This includes struct and enum declarations which describe file formats, such as BMP (Windows Bitmap, an image format) and PE (Portable Executable, used in .NET / CLR, UEFI, and other applications).

Make these available by:

1. Removing the `version (Windows):` guards from the top of files   corresponding to Windows headers;

2. Adding `version (Windows)` guards to `pragma(lib, ...)` lines, to   pulling in Windows libraries on non-Windows platforms.

### DIFF
--- a/changelog/windows_api.dd
+++ b/changelog/windows_api.dd
@@ -1,0 +1,10 @@
+$(D core.sys.windows) modules are now usable on POSIX
+
+Many Windows API bindings in Druntime contain declarations which are useful on other platforms.
+This includes struct and enum declarations which describe file formats,
+such as BMP (Windows Bitmap, an image format)
+and PE (Portable Executable, used in .NET / CLR, UEFI, and other applications).
+
+Previously, the contents of the Windows API bindings were not usable on other platforms
+due to a `version (Windows):` guard at the top of each file.
+This has now been removed, making the declarations available on all platforms.
--- a/src/core/sys/windows/accctrl.d
+++ b/src/core/sys/windows/accctrl.d
@@ -8,7 +8,6 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_accctrl.d)
  */
 module core.sys.windows.accctrl;
-version (Windows):
 
 version (ANSI) {} else version = Unicode;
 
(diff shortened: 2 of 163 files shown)


CyberShadow:
This idea was discussed previously on a few occasions, and I don't think there were any arguments against it. The only one that comes to mind is that this is a potentially breaking change: previously, importing a Windows header on non-Windows was a no-op. Now, this PR "introduces" a ton of symbols on those platforms, which could hypothetically result in collisions with shims / alternative implementations of these APIs for other platforms.

CyberShadow:
One downside I remembered is that it makes the Phobos library file larger, mostly due to initializers for Windows API struct type declarations. It does increase the size of libphobos2.a from 68 to 88 MB, though doesn't affect the size of the .so.

CyberShadow:
I guess it's doubtful this approach will work.

Will open a new PR that activates only certain modules that we know are useful on non-Windows.

marler8997:
@CyberShadow 

You could split up the modules based on "link" vs "nolink".  Put the enums/types in the "nolink" module and the extern function declarations into the "link" module.

CyberShadow:
The problem is really not the function declarations, but the D functions which are translations of C macros. Because these D functions call Windows APIs, the linker is pulling in Windows APIs. (Looks like, additionally to that, this only happens on FreeBSD - the linker there apparently doesn't care if the D function is never called, it still tries to pull in its dependencies.)

https://wiki.dlang.org/Deimos recommends to use templated D functions to translate macros to avoid this exact problem, however, doing that here now is a potentially breaking change, as e.g. you can't take the address of a template (only an explicit instantiation, even though it has zero arguments). So, adding `version(Windows)` to the D functions is the solution, but, there's just a lot of such functions intermixed with other declarations.

marler8997:
Put the macro template in the `nolink` module, and have it call the explicit function from the `link` module.  Best of both worlds.   You have access to the function address if you need it, but can still use the macro to lazily instantiate it.

PR #2616, opened 2019-05-27
Make some Windows declarations usable on other platforms

CyberShadow:
A different approach to #2615:

- In #2615, I tried to enable all modules for other platforms, and fix problems / disable the modules with big problems.

- Here I am only enabling the modules (and their dependencies) needed for things we know are useful for other platforms. The list of modules covering these known use cases is much smaller.

### DIFF
--- a/changelog/windows_api.dd
+++ b/changelog/windows_api.dd
@@ -1,0 +1,22 @@
+Some $(D core.sys.windows) modules are now usable on POSIX
+
+The Windows API bindings in Druntime contain declarations which are useful on other platforms.
+This includes struct and enum declarations which describe file formats,
+such as BMP (Windows Bitmap, an image format)
+and PE (Portable Executable, used in .NET / CLR, UEFI, and other applications).
+
+Previously, the contents of the Windows API bindings were not usable on other platforms
+due to a `version (Windows):` guard at the top of each file.
+This has now been removed for the following modules, making the declarations available on all platforms:
+
+$(UL
+$(LI $(MREF core,sys,windows,basetsd))
+$(LI $(MREF core,sys,windows,basetyps))
+$(LI $(MREF core,sys,windows,w32api))
+$(LI $(MREF core,sys,windows,winbase))
+$(LI $(MREF core,sys,windows,windef))
+$(LI $(MREF core,sys,windows,wingdi))
+$(LI $(MREF core,sys,windows,winnt))
+$(LI $(MREF core,sys,windows,winuser))
+$(LI $(MREF core,sys,windows,winver))
+)
--- a/src/core/sys/windows/basetsd.d
+++ b/src/core/sys/windows/basetsd.d
@@ -8,7 +8,6 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_basetsd.d)
  */
 module core.sys.windows.basetsd;
-version (Windows):
 
 /*  This template is used in these modules to declare constant pointer types,
  *  in order to support both D 1.x and 2.x.
--- a/src/core/sys/windows/basetyps.d
+++ b/src/core/sys/windows/basetyps.d
@@ -7,7 +7,6 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_basetyps.d)
  */
 module core.sys.windows.basetyps;
-version (Windows):
 
 private import core.sys.windows.windef, core.sys.windows.basetsd;
 
--- a/src/core/sys/windows/w32api.d
+++ b/src/core/sys/windows/w32api.d
@@ -8,7 +8,6 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_w32api.d)
  */
 module core.sys.windows.w32api;
-version (Windows):
 
 version (ANSI) {} else version = Unicode;
 
--- a/src/core/sys/windows/winbase.d
+++ b/src/core/sys/windows/winbase.d
@@ -7,10 +7,9 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_winbase.d)
  */
 module core.sys.windows.winbase;
-version (Windows):
 
 version (ANSI) {} else version = Unicode;
-pragma(lib, "kernel32");
+version (Windows) pragma(lib, "kernel32");
 
 /**
 Translation Notes:
--- a/src/core/sys/windows/winbase.d
+++ b/src/core/sys/windows/winbase.d
@@ -1916,7 +1916,7 @@ WINBASEAPI DWORD WINAPI GetCurrentThreadId(void);
     UINT GlobalGetAtomNameA(ATOM, LPSTR, int);
     UINT GlobalGetAtomNameW(ATOM, LPWSTR, int);
 
-    bool HasOverlappedIoCompleted(LPOVERLAPPED lpOverlapped) {
+    extern(D) bool HasOverlappedIoCompleted(LPOVERLAPPED lpOverlapped) {
         return lpOverlapped.Internal != STATUS_PENDING;
     }
 
--- a/src/core/sys/windows/winbase.d
+++ b/src/core/sys/windows/winbase.d
@@ -2481,7 +2481,7 @@ WINBASEAPI BOOL WINAPI SetEvent(HANDLE);
 }
 
 // For compatibility with old core.sys.windows.windows:
-version (LittleEndian) nothrow @nogc
+version (Windows) version (LittleEndian) nothrow @nogc
 {
     BOOL QueryPerformanceCounter(long* lpPerformanceCount) { return QueryPerformanceCounter(cast(PLARGE_INTEGER)lpPerformanceCount); }
     BOOL QueryPerformanceFrequency(long* lpFrequency) { return QueryPerformanceFrequency(cast(PLARGE_INTEGER)lpFrequency); }
--- a/src/core/sys/windows/windef.d
+++ b/src/core/sys/windows/windef.d
@@ -8,7 +8,6 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_windef.d)
  */
 module core.sys.windows.windef;
-version (Windows):
 
 public import core.sys.windows.winnt;
 private import core.sys.windows.w32api;
--- a/src/core/sys/windows/wingdi.d
+++ b/src/core/sys/windows/wingdi.d
@@ -7,10 +7,9 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_wingdi.d)
  */
 module core.sys.windows.wingdi;
-version (Windows):
 
 version (ANSI) {} else version = Unicode;
-pragma(lib, "gdi32");
+version (Windows) pragma(lib, "gdi32");
 
 // FIXME: clean up Windows version support
 
--- a/src/core/sys/windows/winnt.d
+++ b/src/core/sys/windows/winnt.d
@@ -7,7 +7,6 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_winnt.d)
  */
 module core.sys.windows.winnt;
-version (Windows):
 
 version (ANSI) {} else version = Unicode;
 
--- a/src/core/sys/windows/winuser.d
+++ b/src/core/sys/windows/winuser.d
@@ -3667,7 +3667,7 @@ extern (Windows) nothrow @nogc {
 
 } // extern (Windows)
 
-nothrow @nogc {
+version (Windows) nothrow @nogc {
     HCURSOR CopyCursor(HCURSOR c) {
         return cast(HCURSOR)CopyIcon(cast(HICON)c);
     }
@@ -4423,9 +4423,9 @@ int BroadcastSystemMessageW(DWORD, LPDWORD, UINT, WPARAM, LPARAM);
 UINT SendInput(UINT, LPINPUT, int);
 BOOL EnumDisplayMonitors(HDC, LPCRECT, MONITORENUMPROC, LPARAM);
 BOOL GetMonitorInfoA(HMONITOR, LPMONITORINFO);
-extern(D) BOOL GetMonitorInfoA(HMONITOR m, LPMONITORINFOEXA mi) { return GetMonitorInfoA(m, cast(LPMONITORINFO)mi); }
+version (Windows) extern(D) BOOL GetMonitorInfoA(HMONITOR m, LPMONITORINFOEXA mi) { return GetMonitorInfoA(m, cast(LPMONITORINFO)mi); }
 BOOL GetMonitorInfoW(HMONITOR, LPMONITORINFO);
-extern(D) BOOL GetMonitorInfoW(HMONITOR m, LPMONITORINFOEXW mi) { return GetMonitorInfoW(m, cast(LPMONITORINFO)mi); }
+version (Windows) extern(D) BOOL GetMonitorInfoW(HMONITOR m, LPMONITORINFOEXW mi) { return GetMonitorInfoW(m, cast(LPMONITORINFO)mi); }
 HMONITOR MonitorFromPoint(POINT, DWORD);
 HMONITOR MonitorFromRect(LPCRECT, DWORD);
 HMONITOR MonitorFromWindow(HWND, DWORD);
@@ -4521,7 +4521,33 @@ version (Unicode) {
     alias MONITORINFOEXW MONITORINFOEX;
     alias ICONMETRICSW ICONMETRICS;
     alias NONCLIENTMETRICSW NONCLIENTMETRICS;
+} else { // ANSI
+    alias EDITWORDBREAKPROCA EDITWORDBREAKPROC;
+    alias PROPENUMPROCA PROPENUMPROC;
+    alias PROPENUMPROCEXA PROPENUMPROCEX;
+    alias DESKTOPENUMPROCA DESKTOPENUMPROC;
+    alias WINSTAENUMPROCA WINSTAENUMPROC;
+    alias MAKEINTRESOURCEA MAKEINTRESOURCE;
 
+    alias WNDCLASSA WNDCLASS;
+    alias WNDCLASSEXA WNDCLASSEX;
+    alias MENUITEMINFOA MENUITEMINFO;
+    alias LPCMENUITEMINFOA LPCMENUITEMINFO;
+    alias MSGBOXPARAMSA MSGBOXPARAMS;
+    alias HIGHCONTRASTA HIGHCONTRAST;
+    alias SERIALKEYSA SERIALKEYS;
+    alias SOUNDSENTRYA SOUNDSENTRY;
+    alias CREATESTRUCTA CREATESTRUCT;
+    alias CBT_CREATEWNDA CBT_CREATEWND;
+    alias MDICREATESTRUCTA MDICREATESTRUCT;
+    alias MULTIKEYHELPA MULTIKEYHELP;
+    alias MONITORINFOEXA MONITORINFOEX;
+    alias ICONMETRICSA ICONMETRICS;
+    alias NONCLIENTMETRICSA NONCLIENTMETRICS;
+
+}
+
+version (Windows) { version (Unicode) {
     alias AppendMenuW AppendMenu;
     alias BroadcastSystemMessageW BroadcastSystemMessage;
     static if (_WIN32_WINNT >= 0x501) {
@@ -4670,29 +4696,6 @@ version (Unicode) {
 
 } else { // ANSI
 
-    alias EDITWORDBREAKPROCA EDITWORDBREAKPROC;
-    alias PROPENUMPROCA PROPENUMPROC;
-    alias PROPENUMPROCEXA PROPENUMPROCEX;
-    alias DESKTOPENUMPROCA DESKTOPENUMPROC;
-    alias WINSTAENUMPROCA WINSTAENUMPROC;
-    alias MAKEINTRESOURCEA MAKEINTRESOURCE;
-
-    alias WNDCLASSA WNDCLASS;
-    alias WNDCLASSEXA WNDCLASSEX;
-    alias MENUITEMINFOA MENUITEMINFO;
-    alias LPCMENUITEMINFOA LPCMENUITEMINFO;
-    alias MSGBOXPARAMSA MSGBOXPARAMS;
-    alias HIGHCONTRASTA HIGHCONTRAST;
-    alias SERIALKEYSA SERIALKEYS;
-    alias SOUNDSENTRYA SOUNDSENTRY;
-    alias CREATESTRUCTA CREATESTRUCT;
-    alias CBT_CREATEWNDA CBT_CREATEWND;
-    alias MDICREATESTRUCTA MDICREATESTRUCT;
-    alias MULTIKEYHELPA MULTIKEYHELP;
-    alias MONITORINFOEXA MONITORINFOEX;
-    alias ICONMETRICSA ICONMETRICS;
-    alias NONCLIENTMETRICSA NONCLIENTMETRICS;
-
     alias AppendMenuA AppendMenu;
     alias BroadcastSystemMessageA BroadcastSystemMessage;
     static if (_WIN32_WINNT >= 0x501) {
@@ -4838,7 +4841,7 @@ version (Unicode) {
     alias EnumDisplaySettingsA EnumDisplaySettings;
     alias EnumDisplaySettingsExA EnumDisplaySettingsEx;
     alias EnumDisplayDevicesA EnumDisplayDevices;
-}
+} }
 
 alias WNDCLASS* LPWNDCLASS, PWNDCLASS;
 alias WNDCLASSEX* LPWNDCLASSEX, PWNDCLASSEX;
--- a/src/core/sys/windows/winuser.d
+++ b/src/core/sys/windows/winuser.d
@@ -7,10 +7,9 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_winuser.d)
  */
 module core.sys.windows.winuser;
-version (Windows):
 
 version (ANSI) {} else version = Unicode;
-pragma(lib, "user32");
+version (Windows) pragma(lib, "user32");
 
 // Conversion Notes:
 // The following macros were for win16 only, and are not included in this file:
--- a/src/core/sys/windows/winver.d
+++ b/src/core/sys/windows/winver.d
@@ -8,10 +8,9 @@
  * Source: $(DRUNTIMESRC src/core/sys/windows/_winver.d)
  */
 module core.sys.windows.winver;
-version (Windows):
 
 version (ANSI) {} else version = Unicode;
-pragma(lib, "version");
+version (Windows) pragma(lib, "version");
 
 private import core.sys.windows.windef;
 


JinShil:
This seems odd.  It appears this is only being done because there is currently no portable library containing such things?  Wouldn't creating such a library would be a better way forward?

CyberShadow:
No. Why have the same code in two places?

CyberShadow:
Having the same identifiers copy-pasted in two places also creates problems such as:

- Windows-only library/application A uses Druntime bindings
- Cross-platform library/application B uses stand-alone declarations

Now these two cannot talk to each other (without ugly hacks like casts all over the place), because they use different types, even though the type declarations are identical.

We had these problems back before I added the Windows headers to Druntime. Code using the limited Druntime headers had a hard time talking to code using the `win32` package from the bindings project.

atilaneves:
No tests?

CyberShadow:
I think we generally don't test bindings for APIs others have created, but I guess it's possible to write some cross-platform D code that makes use of these declarations, if that's what you had in mind!

I plan to use these in ae as soon as a new release rolls out with this PR, so, it will be tested by the project tester anyway :)

jmdavis:
In theory, the C bindings in druntime are supposed to mirror the actual C headers so that you import them basically the same way that you'd #include them in C, and I've never heard of anyone #including any windows headers on Linux. So, it seems _really_ wrong to be importing anything from core.sys.windows on any other platform. Maybe I'm just clueless here, but under what circumstances would _anything_ from the win32 API be #included on any non-Windows system? Is there some functionality in there that happens to be part of the Win32 API but is also used by other systems? If so, how is it #included on Linux? Are there other headers on Linux, or are Linux projects actually #including files like winnt.h?

CyberShadow:
> Maybe I'm just clueless here, but under what circumstances would _anything_ from the win32 API be #included on any non-Windows system?

Please read the commit message :) I've provided a few examples.

CyberShadow:
Here is how I'd like to make use of this PR in my own code:

- PE file format:

  - `rclidasm` is a WIP robust assembler/disassembler for Common Language Infrastructure executables:
    https://github.com/CyberShadow/rclidasm

  - `ae.sys.windows.pe` is a package which allows manipulating Windows portable executables, including access to resource directories and version info:
    https://github.com/CyberShadow/ae/tree/master/sys/windows/pe
    We use `ae.sys.windows.pe` as part of our Worms Armageddon build process on POSIX.

- BMP:

  - `ae.utils.graphics.image` has support for reading and writing BMP files:
     https://github.com/cybershadow/ae/blob/master/utils/graphics/image.d#L640

  - `ae` also uses BMP as a graphics interchange format, to read/write image from/to ffmpeg and ImageMagick:
     https://github.com/cybershadow/ae/blob/master/utils/graphics/im_convert.d
     https://github.com/cybershadow/ae/blob/master/utils/graphics/ffmpeg.d

     I demo'd the ffmpeg thing the last time I was at DConf.

> If so, how is it #included on Linux? Are there other headers on Linux, or are Linux projects actually #including files like winnt.h?

I think, in case of the C language, the copyright status and redistribution conditions of Microsoft-provided header files caused independent projects to recreate those headers from the specifications. By the time free headers became available (MinGW etc.), that practice had already taken hold.

jmdavis:
But how are these used on Linux in C? Would someone be #including one of the corresponding Windows C headers to get at bitmap functionality? Or would they be #including something else entirely? My point is that we're supposed to be trying to mirror what's done in C with the C headers and how they would be used in C. We don't follow that 100%, because we're providing bindings for multiple platforms in the same library, whereas with C, they just provide different version of the headers for each platform (e.g. time.h on FreeBSD will just have the FreeBSD version of things, and time.h on Linux will just have the Linux version of things, whereas we have core.sys.posix.time for the standard POSIX stuff and core.sys.freebsd.time and core.sys.linux.time for the FreeBSD and Linux-specific stuff), but unless someone is actually going to #include something like winnt.h on Linux in C, it really doesn't follow how we've set the bindings up to have anything in core.sys.windows not be versioned for Windows.

CyberShadow:
I don't think anything of what you said above contradicts this PR. The C header files are not guarded by some `#ifdef Windows` either, it's just that including the Microsoft-provided ones simply won't work on other platforms. Not sure about the MinGW ones.

jmdavis:
But if it doesn't work to #include the windows header files in C, then that's not what a C program would be doing, and the D usage would not follow the C usage. Any library on Linux would be using headers associated with that library, and what guarantee is there that everything you're now making non-Windows specific here is going to be exactly the same in those Linux libraries? Sure, many of them may be, but if there's any chance that any of those declarations will be different (e.g. a different enum value), then we have a problem.

Overall, the approach with druntime is to purposefully duplicate bindings and use `static assert(false, ...)` for platforms we aren't sure about rather than assuming that some set of bindings will work on multiple platforms. And even if the bindings _do_ work on multiple platforms, we _still_ duplicate them. Walter has been very insistent about that approach, and what you're proposing here does not follow that approach at all. IMHO, if any of these declarations are used on other platforms, they should be declared in a way that matches the way that they're used on those platforrms in C. That way, the D usage matches the C usage, and if any declarations need to be changed for any reason, we don't have to worry about screwing up other platforms because they changed on one platform but not another.

CyberShadow:
The idea that modules containing definitions specific to a platform should not be importable from other platforms is fallacious. Platformisms will inevitably creep across the platform boundary, through file formats and protocols, whether you wish it or not. I've already mentioned the case for PE and BMP files. Another example is ZIP files: their files can contain either POSIX or Windows attribute flags; therefore, cross-platform programs that need to deal with them correctly need to have access to both `FILE_ATTRIBUTE_READONLY` and `S_IWOTH` declarations. UNIX time is another example you're familiar with.

CyberShadow:
> Any library on Linux would be using headers associated with that library, and what guarantee is there that everything you're now making non-Windows specific here is going to be exactly the same in those Linux libraries? Sure, many of them may be, but if there's any chance that any of those declarations will be different (e.g. a different enum value), then we have a problem.

I really can't think how this could come to happen. If there is a mismatch, then there is a bug in either library. Can you provide a counter-example?

> And even if the bindings _do_ work on multiple platforms, we _still_ duplicate them.

This PR is not at all about API bindings. It is about making parts of the modules that are useful for file formats and interoperability usable as such.

I really don't understand where you're going with your argument.

jmdavis:
> I really don't understand where you're going with your argument.

If you wouldn't #include a windows header on Linux, then we shouldn't be making it importable on Linux. We should be matching how this is done in C so that the code matches C and so that any platform differences are properly maintained. We have historically taken the approach that bindings should be platform-specific and match the platform they're for. We have specifically avoided declaring bindings that are multi-platform - even with POSIX where it would arguably make sense and reduce the amount of code duplication. As such, whatever declarations we have for Linux or FreeBSD or Mac OS X or whatever should match what would actually be used on those platforms, not pull in our equivalent of windows.h and friends. If that means code duplication, then that means code duplication. We've been duplicating code in that manner very much on purpose through the druntime bindings for pretty much all of druntime's existence AFAIK.

You're taking huge swaths of declarations for Windows and making them importable on other platforms. Even if we were to assume that bindings from other platforms should be available on other platforms (which I don't agree with at all), you're taking a lot of stray windows-specific declarations and treating them as if they were cross-platform. For instance, winnt.d, you've made stuff like `FCHAR`, `PWCHAR`, and `TCHAR` available on platforms where they're clearly not appropriate.

If there are declarations that are useful here for dealing with file formats and whatnot, then for Linux, they should be put in our equivalent of wherever they would be for Linux in C. Then only the stuff that actually makes sense for Linux will be available and folks maintaining core.sys.windows don't have to worry about accidentally adding bindings that definitely wouldn't work on Linux or any other non-Windows system.

CyberShadow:
BTW, since you mentioned our platform headers, I think the way we deal with them could be improved.

Right now it's a giant mess of `version(X) { ... 1000 lines of code ... } else version (Y) { ... etc.`.

I think it would be better if there were:

- `core.sys.linux.sys.time`
- `core.sys.freebsd.sys.time`
- `core.sys.openbsd.sys.time`
- etc.

and

- `core.sys.posix.sys.time`

which **only** contains a public import based on the current platform version.

This would be better in that:

- Each D file in the platforms' bindings maps directly to one C header file, thus making updates much easier (no more `version`s)

- Each platform's headers are accessible directly if necessary, as discussed above

- If any code is shared, it is now much easier to find than going through the `version{...}` directives

- Adding a new platform becomes much easier as well.

jmdavis:
> which only contains a public import based on the current platform version.

We purposefully do it they way that we do so that if you import any of the core.sys.posix modules, you don't accidentally use functionality that is platform-specific.

CyberShadow:
> We purposefully do it they way that we do so that if you import any of the core.sys.posix modules, you don't accidentally use functionality that is platform-specific.

And this is useful, how? All this does is artificially close off some use cases. If you are putting a specific operating system's name in your import statement, you would know what you're doing, right? AFAICS, it is making D worse, not better.

> For instance, winnt.d, you've made stuff like `FCHAR`, `PWCHAR`, and `TCHAR` available on platforms where they're clearly not appropriate.

This seems like a *really* minor issue compared to the benefit this PR will bring.

Here is what's upsetting me about this conversation:

1. I try to build cool things with D
2. D needs a small change to make cool thing possible / better
3. Submit small change in PR. D is better, I make cool thing, all good yes?
4. No! @jmdavis appears and lists philosophical hypothetical arguments at least three layers removed from reality or actually building cool things that makes someone's life easier.

Maybe we just have different interpretations of what "make D better" means. To me, it means making D more useful to people who want to build cool things with D. But, lately, the trend I've been seeing is that a better D means more "strict", more "correct", and, well, to hell with anyone actually trying to build things with it. I don't understand how dlang/phobos#6867 will help people build cool things with D. I don't understand how dlang/dmd#8941 will help people build cool things with D. And, I don't understand how rejecting this PR will help people build cool things with D.

@jmdavis Please, help me make D better. Can we put away the philosophy and the triple what-ifs and find a practical solution that allows building more better cool things with D? As far as I've seen, I don't think I noticed you suggest an alternative approach that would solve the problems solved by this PR in a way you would find satisfactory.

jmdavis:
> And this is useful, how? All this does is artificially close off some use cases. If you are putting a specific operating system's name in your import statement, you would know what you're doing, right? AFAICS, it is making D worse, not better.

If someone imports `core.sys.linux.whatever` they may accidentally use stuff that is not actually POSIX (especially if they're not familiar with where the line is between POSIX and what their platform is using), and the code won't work on other POSIX platforms, whereas if they import `core.sys.posix.whatever`, they should be able to rely on it working on other POSIX platforms. Each platform (especially Linux) likes to add their own platform-specific stuff on top of POSIX. So, if you want to use platform-specific stuff, you import the platform-specific module, and if you want to restrict what you're doing to POSIX, then you import the core.sys.posix module.

> As far as I've seen, I don't think I noticed you suggest an alternative approach that would solve the problems solved by this PR in a way you would find satisfactory.

Because the whole approach goes against how we've set up the druntime bindings. The approach that we follow throughout all of core.sys is to duplicate bindings for each platforms even if they happen to be the same - even if a standard requires that they be the same. @WalterBright  is _very_ insistent about that, because it prevents bugs related to accidentally creating bindings that work on one platform but not another or otherwise having platform-specific stuff creep into other platforms. Based on that, if you want these on Linux, then the correct solution is to create a module that corresponds to the header where they would be in on Linux. If they're not standard on other systems such that they don't belong in core.sys.linux or core.sys.posix, then they should go in a dub package (possibly in deimos), which is where non-OS level bindings are supposed to go. Sure, it might be more work than simply removing version statements, but the result will be cleaner and avoid certain classes of bugs - it's also in line with the policies that @WalterBright has had for druntime.

This is one of the issues that @WalterBright  has been very firm about. So, if you want this, convince him. Personally, I don't see any problem with the requirement that the bindings be OS-specific. Yes, it creates code duplication, which can sometimes be annoying, but it prevents many of the kinds of bugs that you get with the #ifdef mess in C. And as things stand, anyone working on any of the bindings in core.sys can rely on the fact that anything in core.sys.X is only going to be used on system X and doesn't have to worry about accidentally introducing code breakage on other platforms.

CyberShadow:
> If someone imports `core.sys.linux.whatever` they may accidentally use stuff that is not actually POSIX (especially if they're not familiar with where the line is between POSIX and what their platform is using), and the code won't work on other POSIX platforms

This is a fictional problem!

1. You need to be careful to make your code POSIX-compliant anyway. You generally don't write a non-trivial program against Linux/glibc headers and it's automatically POSIX-compliant (or even useful on other systems).

2. You are assuming you know better than the user. If the user is importing a header with "linux" in the name, it's not your call to tell them they actually mean something else.

I see you're ignoring the second part of my post and again going into completely hypothetical what-ifs...

jmdavis:
> You are assuming you know better than the user. If the user is importing a header with "linux" in the name, it's not your call to tell them they actually mean something else.

We're not telling users that we know better. We're splitting the bindings between standard POSIX and system-specific stuff so that those who want to use stuff that is only standard POSIX can easily do so without having platform-specific stuff creep in. No, that doesn't absolutely guarantee that they don't then make assumptions about what's going on that's specific to their platform or otherwise screw that up, but they can rely on the fact that any bindings they use are standard POSIX and not specific to a particular platform. Anyone who doesn't care can simply use core.sys.X for platform X, and then they not only don't have to worry about it, but the imports make it crystal clear that the code was written for platform X and not POSIX.

> I see you're ignoring the second part of my post and again going into completely hypothetical what-ifs...

Being sloppy with bindings does not make D better. It may be the quick and easy path to make some Windows-specific declarations that relate to stuff used on other platforms usable on those platforms in D, but there is a real risk that it will cause problems with the bindings over time for exactly the reasons that being sloppy with #ifdefs in C causes problems. That's not just hypothetical, and it's something that Walter is very insistent on. There is no reason why the relevant bindings can't be duplicated wherever they belong in core.sys.linux if they make sense there, and if they don't, then there's no reason why they can't be duplicated in a dub package. Either way, the result is that the platform-specific stuff stays segregated, which is the current policy for druntime bindings.

adamdruppe:
But, if a user explicitly `import core.sys.linux.epoll`, for example, even if they are building on Windows, what harm does it do to let them see the declarations?

Similarly, if someone explicitly writes `import core.sys.windows.windows;`, why not let them see those declarations, even if they won't actually work? Maybe they are writing a program on Linux that is trying to reflect over those modules to communicate with a Windows box or something. They might not actually call it, but just want to access the data.

I can't see any downside to allowing it. Again, the user is *explicitly importing the OS-specific module*. They should know what they're getting into. And if they only want it to be visible - which I know I already do - it is simple to put `version(Windows) import core.sys.windows.windows;`

CyberShadow:
> We're not telling users that we know better. We're splitting the bindings between standard POSIX and system-specific stuff so that those who want to use stuff that is only standard POSIX can easily do so without having platform-specific stuff creep in. No, that doesn't absolutely guarantee that they don't then make assumptions about what's going on that's specific to their platform or otherwise screw that up, but they can rely on the fact that any bindings they use are standard POSIX and not specific to a particular platform. Anyone who doesn't care can simply use core.sys.X for platform X, and then they not only don't have to worry about it, but the imports make it crystal clear that the code was written for platform X and not POSIX.

You can still have a `core.sys.posix` package that contains only POSIX declarations, while also having a `core.sys.linux` package that's importable from any platform if one so wishes.

> That's not just hypothetical, and it's something that Walter is very insistent on.

What is the actual argument? Does it actually apply to this case? Can the problem be solved in other ways?

> There is no reason why the relevant bindings can't be duplicated wherever they belong in core.sys.linux if they make sense there, and if they don't, then there's no reason why they can't be duplicated in a dub package.

I described the problem with duplicating the headers [here](https://github.com/dlang/druntime/pull/2616#issuecomment-496053494). There is also the implied problems with copying code:

- One copy may go out of date
- The effort to maintain them is doubled
- The two copies might diverge into incompatible forks

Your parallel also doesn't really work, because in case of Linux / POSIX / FreeBSD / etc., these are all different OSes; similarities in syscall numbers etc. are coincidental. There is no case of one operating system's headers being based off (a fork of) another's. This will not be true if the Windows headers are copied elsewhere.

CyberShadow:
I realize this isn't quite how we've been doing things so far. But, I think the practical applications of moving in this direction outweigh any ideological benefits granted by the status quo.

As such, could we please move forward with this? This change is blocking my current project.

CyberShadow:
> and I think that any cross-platform bindings like you're trying to do here

Again, these are not "cross-platform bindings". These are data types describing file and interchange formats that are incidentally part of the Windows API bindings.

>  should be on code.dlang.org

Again, this creates problems due to the declarations being duplicated.

CyberShadow:
@jmdavis Though I still disagree with you in principle, would you be OK with a patch which makes available *only* the declarations in the examples I mentioned above? I.e. no `TCHAR`. The diff will be much bigger, and will undoubtedly make updating the headers more difficult.

CyberShadow:
> (especially if a large number of declarations are not versioned specifically for Windows, since that greatly increases the chance that a symbol that happens to have the same name as one on another platform gets pulled in).

I'm not sure how closely you've looked at the diff, but in any case, perhaps I should clarify. 

In #2615, I tried making all the headers importable. I abandoned that because it had an unreasonable impact on the size of the Phobos static library file on non-Windows systems.

In this pull request, I make available *only* those modules which contain declarations I know to be useful. You mentioned the socket API - WinSock is not among the modules which have been made available.

jmdavis:
> @jmdavis Though I still disagree with you in principle, would you be OK with a patch which makes available only the declarations in the examples I mentioned above? I.e. no TCHAR. The diff will be much bigger, and will undoubtedly make updating the headers more difficult.

That would be an improvement, but I'm completely against anything in core.sys having any symbols available for any platform other than the one that the module is specifically for. It's a code smell to see an import for core.sys.windows that's not versioned for Windows, just like it's a code smell to see an import for core.sys.linux that's not versioned for Linux. These modules are supposed to correspond to the C headers that share their names for the system that they're for, and those don't get #included on other systems. I would expect most anyone seeing a Windows module imported on Linux to think that something was wrong, and if a Windows module has declarations available for other platforms, then anyone who makes any changes to the module has to worry about how it will affect other platforms.

I think that what really should happen is that a deimos project get put together for whatever declarations make sense on other platforms. But if you really want to have this stuff importable from other platforms without any code duplication (meaning that it has to be in druntime), then I think that one or more modules specific to the functionality you're trying to make available should be created (e.g. one specific for declarations related to bitmaps) and that any such modules should not be in core.sys. Then the core.sys.windows modules that have those declarations can import and alias those symbols from the cross-platform modules rather than having any other platforms importing modules that are supposed to be specific to Windows.

adamdruppe:
On Thu, Jun 13, 2019 at 02:45:35PM -0700, Jonathan M Davis wrote:
> These modules are supposed to correspond to the C headers that share their names for the system that they're for, and those don't get #included on other systems.

But there's nothing stopping you from doing it. The C headers don't all
say #ifdef whatever.

> I would expect most anyone seeing a Windows module imported on Linux to think that something was wrong, and if a Windows module has declarations available for other platforms, then anyone who makes any changes to the module has to worry about how it will affect other platforms.

I think this worry is completely unrealistic.

CyberShadow:
> one or more modules specific to the functionality you're trying to make available should be created

I would say that this is also not quite right, because the *canonical* definitions of these file and interchange formats are dictated by Microsoft, and as such these is nothing wrong with them being in `core.sys.windows`. I think we should just stop thinking of `core.sys` as modules that should exclusively  be imported only on their given platform - that assumption doesn't seem very useful in the first place.

jmdavis:
> Since you feel strongly about this, and you seem to have a good understanding of the issues involved, could you please propose a plan that we could move towards implementing? It feels like we are on the verge of finding a solution that would be satisfactory from all angles, whether it involves restructuring the Windows headers or redefining how we think about platform-specific headers. If you'd rather not proceed without Walter's opinion, I can email him and ask him to add his voice to the discussion.

I am against any solution which involves making our bindings for system-specific headers importable on any system other than the one the header is from. In general, allowing that would be a big problem - especially with the POSIX stuff; the same symbols exist on different systems, but they're declared differently. As such, these files are specifically designed to only be used on the platforms that they're for. It's less of a problem with Windows, because there's less overlap, but overlap still does exist, and in general, it's completely inappropriate to use the Windows symbols on other systems even if they're just aliases or enums, because they're specific to Windows. IMHO, it should be a considered a bug if anything from core.sys.X is imported on any system other than X. The headers that these files correspond to are not intended to be used on any other systems and in general will not work on other systems. And for the most part, the same is true of the D versions.

You do have an odd case here in that there are some symbols declared in a few of the Windows headers which do not require linking and which correspond to file formats that are not Windows-specific, but I do not believe that that justifies making it possible to import core.sys.windows files on other platforms - especially when it involves making a bunch of symbols available that clearly are inappropriate to use on other systems and are not for the file format stuff.

And even if you want to consider the windows headers the "canonical" source of some of these declarations, no Linux program written in C/C++ would ever #include any of these files, so I don't think that it's appropriate for a D program to import their D counterparts, and the values are already duplicated in D simply by the fact that we're declaring bindings rather than using the actual C headers. So, it's not like using these declarations as opposed to a different copy of them which is actually intended to be used on multiple platforms is necessarily going to be any more correct.

Personally, I think that the correct solution is to just create a cross-platform project on code.dlang.org based on whatever the symbols are actually for and put them in there so that they can be used on any platform without mucking with how core.sys is organized. However, if avoiding code duplication with the declarations in core.sys.windows is actually a big enough concern to merit having them be importable from druntime by multiple platforms, then I think that it's far better to create a new module in druntime separate from core.sys for those symbols and have core.sys.windows.* publicly import them rather than making anything in core.sys importable by other systems.

I will not budge on making core.sys.* importable by other systems short of @WalterBright or @atilaneves declaring that it's okay. Walter has made it quite clear in the past that he considers duplicate declarations for bindings far better than sharing them across systems even when they can theoretically can be shared. I agree with that approach, and it's what we've been doing with core.sys for years. And given how firm Walter has been about it, I think that if we're going to change that approach, we need Walter's approval. So, if you want to e-mail Walter to get his input, then go ahead.

CyberShadow:
That's a good summary of your arguments in the discussion so far, thanks. I'll do the same and summarize my points on why I still think we should consider going in this direction despite your points above, and why I think this situation is not quite like the previous ones you refer to:

1. The intention is to expose structs and enums, not functions. These declarations describe file and interchange formats, and not OS APIs.
1. The declarations are canonically defined by Microsoft and included in Microsoft OS SDKs, and as such are not out-of-place being defined in `core.sys.windows`.
1. [The file/interchange formats are widely used outside of Windows, and I have encountered and used them many times while writing D code](https://github.com/dlang/druntime/pull/2616#issuecomment-496307985).
1. Duplicating these declarations here is not the same as duplicating bindings between e.g. Linux and FreeBSD:
   - If the Linux kernel adds a new syscall parameter tomorrow, FreeBSD does not automatically get it too. However, if Microsoft releases a new BMP version or extension (in a new OS/SP/SDK version).
   - If we want to add support for a new POSIX OS, then the way the POSIX headers are structured will greatly affect this effort; however, the Windows headers are unaffected by this, and the OS-independent declarations within doubly so.
1. Duplicating the definitions creates a large number of issues:
    - One copy of the definitions may go out of date
    - The effort to maintain them is doubled
    - The two copies might diverge into incompatible forks
    - [Two libraries using different copies of the definitions can't talk to each other](https://github.com/dlang/druntime/pull/2616#issuecomment-496053494)
    - [People will still use the Windows-only copy when programming on Windows, even when a cross-platform version exists](https://github.com/dlang/druntime/pull/2616#issuecomment-502699004)
1. Most likely none of the symbols exposed in this pull request exist in other platforms (due to the sheer difference between these platforms); as such, there is no danger of accidentally using an OS-specific definition on the wrong platform.
1. [Many declarations that are clearly OS-specific at first sight inevitably leak their way to other platforms through file formats, making it dubious that we actually want to completely forbid allowing the use of any OS-specific symbols on other platforms](https://github.com/dlang/druntime/pull/2616#issuecomment-496310439). Rather, what we really want is making it difficult to do so *unintentionally*.

WalterBright:
> I think it would be better if there were:

That was what I wanted to do. There was a huge argument about it, and everyone but me wanted to use embedded version declarations in order to minimize the number of lines of code. I lost.

jmdavis:
@WalterBright 
> That was what I wanted to do. There was a huge argument about it, and everyone but me wanted to use embedded version declarations in order to minimize the number of lines of code. I lost.

In some ways, it would be better to have all of the declarations in the system-specific files and then just have public imports for the POSIX symbols in the core.sys.posix file, but there's no point in even having the core.sys.posix file if it's publicly importing the entire system-specific module, since then it's more than just POSIX, and anyone importing the core.sys.posix file would trivially end up using system-specific stuff accidentally. At that point, they might as well just import the system-specific module. By having the core.sys.posix modules only contain the POSIX symbols (be it directly or via public imports), it's easy to make sure that you're not accidentally using something system-specific from that module.

Regardless, I think that the real question here is how you feel about making any of the symbols in core.sys.windows not be versioned only for Windows like Vladamir is pushing for with this PR.

jacob-carlborg:
> In some ways, it would be better to have all of the declarations in the system-specific files and then just have public imports for the POSIX symbols in the core.sys.posix file, but there's no point in even having the core.sys.posix file if it's publicly importing the entire system-specific module, since then it's more than just POSIX, and anyone importing the core.sys.posix file would trivially end up using system-specific stuff accidentally

When I added support for macOS to Mecca I managed to hide all platform specifics (at least the differences between Linux and macOS) in a single package. I think that was a great success limiting the the number of version conditions. 

The idea would be to only publicly import the Posix symbols and not the whole modules.

jmdavis:
> The idea would be to only publicly import the Posix symbols and not the whole modules.

Personally, I would have no problem with the posix stuff being rearranged so that all of the actual declarations are in the system-specific files, and the posix files just have public imports or aliases. As far as I'm concerned, the key thing with core.sys.posix is that when you import any file from there, you just get the POSIX declarations from the corresponding header file, whereas you get them all when you import the system-specific one. And of course, the files shouldn't be set up to have the symbols available on platforms other than the target system. Rearranging the contents of the core.sys.posix modules in such a manner shouldn't even break any code, since you'd still have the same symbols available when importing. They'd just be organized differently.

> Currently it’s quiet a pain to add support for a new platform because the platform specific code is spread all over the place instead of limited to a specific package

Honestly, it's a pain to even find declarations, because you have to wade through a bunch of similar declarations for other platforms to find the one you're looking for.

jacob-carlborg:
> As far as I'm concerned, the key thing with core.sys.posix is that when you import any file from there, you just get the POSIX declarations from the corresponding header file, whereas you get them all when you import the system-specific one

Exactly. 

> Honestly, it's a pain to even find declarations, because you have to wade through a bunch of similar declarations for other platforms to find the one you're looking for.

Yes, that’s definitely the case. 

This WWDC Apple announced a new platform, iPadOS. They thought iOS for iPhone and iPad got too different. But in this case I think it’s enough to define the `Darwin` version identifier for this new platform and we’re good to go.

WalterBright:
tl,dr: I agree with @jmdavis 

One possible solution is to maintain a genericfoo.d, and create a windowsfoo.d that imports genericfoo.d and adds Windows specific seasonings to it, and a linuxfoo.d that imports genericfoo.d and adds Linux specific salt, etc.

CyberShadow:
> tl,dr: I agree with @jmdavis

Alright. Thank you, and thank you @jmdavis for taking the time to have this extensive discussion.

> One possible solution is to maintain a genericfoo.d, and create a windowsfoo.d that imports genericfoo.d and adds Windows specific seasonings to it, and a linuxfoo.d that imports genericfoo.d and adds Linux specific salt, etc.

I am not going to purse this matter further in Druntime, since I think that any way we try to split up the Windows headers, the result is going to be unsatisfactory due to how interwoven the platform-agnostic definitions are with the platform-specific ones.

CyberShadow:
Here is the solution I came up with for this problem:
https://github.com/CyberShadow/win32

WalterBright:
I've run into a problem. Specifically, I am working at turning DMD into a cross-compiler, so a DMD on Windows can generate a Linux binary. See https://github.com/dlang/druntime/pull/3279

But we should be judicious about this, and only do this where it is really needed. After all, importing a file for Linux and then compiling it on FreeBSD and oops, declarations are different, now there are mysteriously crashing executables.

I'm reopening this for discussion.

adamdruppe:
I still see no reason not to allow the Windows headers elsewhere - the bitmap structs and such are legitimately useful everywhere.

But the dmd cross compiler thing, are you just using the structs to make some too?

CyberShadow:
Well, here is an idea. At the top of each platform-specific file, add:

```d
version(linux) version = EnableLinuxDefinitions;
version(EnableLinuxDefinitions):
```

Substitute "Linux" with the relevant platform for other platforms' modules.

This way, platform specific declarations are enabled if either 1) the compiler is targeting that platform, or 2) the user has explicitly asking for them to be enabled.

Something would need to be done about `.init` values for structs and other things that linkers would expect to be in `phobos.lib` etc. but won't. I think ensuring all initial values are zero bytes (in which case I think the compiler won't need to emit an `.init` object file entry), and using template functions instead of regular functions, will do that, though it may take some work to get there. Same situation as for Deimos: https://wiki.dlang.org/Deimos#Translation

> I'm reopening this for discussion.

Not sure if you intended but didn't actually reopen or changed your mind and forgot to remove that line but this is still closed, FWIW.

PR #2617, opened 2019-05-27
Align fibers to 16 bytes on Linux

WalterBright:
Blocking https://github.com/dlang/dmd/pull/9895


WalterBright:
@wilzbach does the autotester use fibers to run the test/runnable tests?

PR #2620, opened 2019-05-30
fix Issue 19861 - core.cpuid reports the wrong number of threads

rainers:
do not use i7 detection on AMD processors
use cpuid 0x8000_001E to detect the number of threads per core

It seems AMD uses the term "core" for "logical cores" or threads, too.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -255,7 +255,7 @@ UT_MODULES:=$(patsubst src/%.d,$(ROOT)/unittest/%,$(SRCS))
 HAS_ADDITIONAL_TESTS:=$(shell test -d test && echo 1)
 ifeq ($(HAS_ADDITIONAL_TESTS),1)
 	ADDITIONAL_TESTS:=test/init_fini test/exceptions test/coverage test/profile test/cycles test/allocations test/typeinfo \
-	    test/aa test/gc test/hash \
+	    test/aa test/cpuid test/gc test/hash \
 	    test/thread test/unittest test/imports test/betterc test/stdcpp test/config
 	ADDITIONAL_TESTS+=$(if $(SHARED),test/shared,)
 endif
--- a/src/core/cpuid.d
+++ b/src/core/cpuid.d
@@ -937,13 +937,27 @@ void cpuidX86()
             datacache[0].lineSize = 32;
         }
     }
-    if (max_cpuid >= 0x0B) {
+    if (cf.probablyIntel && max_cpuid >= 0x0B) {
         // For Intel i7 and later, use function 0x0B to determine
         // cores and hyperthreads.
         getCpuInfo0B();
     } else {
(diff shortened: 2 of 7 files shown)


rainers:
@FraMecca This worked for an AMD Ryzen, could you check it on your system, too?

thewilsonator:
```
--
  | grep -nrE "\<(for\|foreach\|foreach_reverse\|if\|while\|switch\|catch\|version)\(" $(find src -name '*.d') ; test $? -eq 1
  | src/core/cpuid.d:949:            version(GNU) asm pure nothrow @nogc {
  | posix.mak:383: recipe for target 'style_lint' failed
```

rainers:
Anyone else with an AMD processor who can try this? Running and verifying this should be enough:
```
import core.cpuid;
import core.stdc.stdio;

void main()
{
    printf("coresPerCPU: %d\n", coresPerCPU);
    printf("threadsPerCPU: %d\n", threadsPerCPU);
}
```
Do we know whether there is an AMD used by any CI?

Parallel marking relies on this information, so would be good to have this in before the next release.

JinShil:
> Anyone else with an AMD processor who can try this?

cc @baziotis

baziotis:
I have an AMD Ryzen 7 1700. It is 8-core CPU, with 2 threads per core.
But I get:
```
coresPerCPU: 16
threadsPerCPU: 1
```
`lscpu` data:
```
CPU(s):                    16
Thread(s) per core:  2
Core(s) per socket:  8
```

rainers:
>  have an AMD Ryzen 7 1700. It is 8-core processor, with 2 threads per core.
> But I get:
> coresPerCPU: 16
> threadsPerCPU: 1

Thanks for testing. I guess this is without this patch? With the patch it should be correct (8/16), but unfortunately your CPU is exactly the same I've been trying it on.

baziotis:
> but unfortunately your CPU is exactly the same I've been trying it on.
What are the odds..

> I guess this is without this patch? With the patch it should be correct (8/16)
Yes and yes.

rainers:
Rebased on stable as this is important for parallel marking to work on AMD systems.
I added a test to print out cpuid values, mostly to see if there is an AMD processor in our test fleet.

rainers:
auto tester redmond reports dubious values for Intel Core2, too:
```
processor: Intel(R) Core(TM)2 Quad CPU    Q9400  @ 2.66GHz
threadsPerCPU:    1
coresPerCPU:      4
```

FraMecca:
Sorry for the delay. It works on my machines.

rainers:
> Sorry for the delay. It works on my machines.

Thanks for testing. I'll remove the verbose output of all the flags from the test.

thewilsonator:
this good to go?

rainers:
> this good to go?

I think so. The bad result for Intel Core2  is another issue, but it is a pretty old CPU (released 2008).

PR #2621, opened 2019-05-31
Issue 19924 - Make core.bitop.bswap(ulong) work in betterC

n8sh:


src/core/bitop.d
```diff
@@ -501,19 +501,42 @@ struct BitRange
...
-/**
- * Swaps bytes in an 8 byte ulong end-to-end, i.e. byte 0 becomes
- * byte 7, byte 1 becomes byte 6, etc.
- */
-ulong bswap(ulong v) pure
-{
-    auto sv = Split64(v);
+version (CoreDdoc) {}
+else version (LDC) { /+ LDC treats bswap(ulong) as an intrinsic so don't change the mangling. +/ }
+else version (GNU) { /+ GDC treats bswap(ulong) as an intrinsic so don't change the mangling. +/ }
```
n8sh: Does it make sense to special-case these since LDC and GDC maintain their own versions of druntime anyway?

kinke: I don't think so. You could also just use `pragma(inline, true)` to make it work for betterC instead of dummy-templatizing it.

n8sh: That works with `dmd -betterC -inline` but not if `-inline` is omitted (I just tested).

kinke: Should be a DMD bug then, https://dlang.org/spec/pragma.html#inline:

> The default inline behavior is typically selectable with a compiler switch such as -inline.

ibuclaw: I don't think that's factually accurate.  GDC maintains its own implementation of `rt`, everything else is ad verbatim.

jacob-carlborg: > Does it make sense to special-case these since LDC and GDC maintain their own versions of druntime anyway?

Even if they maintain their own fork of druntime I think they prefer to have as few differences as possible.

ibuclaw: As far as I'm aware, the regression introduced in #2426 has never been fixed either (GDC treats all of core.checkedint as intrinsics, so mangling was broken there too).

n8sh: https://github.com/dlang/druntime/pull/3162

PR #2622, opened 2019-06-01
Issue 19561 - Enable array ops in betterC

n8sh:
Known limitation: slice assignment to a scalar won't work in betterC for types of size other than 1 because the compiler will try to use _memset16/_memset32/etc. This can only be addressed in DMD rather than druntime.


JinShil:
> Known limitation: slice assignment to a scalar won't work in betterC for types of size other than 1 because the compiler will try to use _memset16/_memset32/etc.

Could this be addressed by making the `_memset` functions templates?

PR #2624, opened 2019-06-03
Add druntime core.deadbeef

WalterBright:
This will be handy for making testing the code generation more reliable. It's also useful for preventing data from leaking via the registers from the internal implementation of a library.

### DIFF
--- a/src/core/deadbeef.d
+++ b/src/core/deadbeef.d
@@ -1,0 +1,147 @@
+/**
+ * This module contains deadbeef().
+ *
+ * Copyright: D Language Foundation 2019
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   Walter Bright
+ * Source:    $(DRUNTIMESRC core/_deadbeef.d)
+ */
+
+module core.deadbeef;
+
+/****************************************
+ * Step on all the register contents that
+ * are not saved by the calling convention
+ * of the function.
+ *
+ * Useful for:
+ *
+ * 1. testing the code generator
+ * 2. testing code that uses inline assembler to help ensure that
+ * it follows the function ABI
+ * 2. avoiding leaking register information that should be hidden
+ * from the caller (clobbering the released part of the stack is
+ * a separate concern)
+ * 3. avoiding leaking register information to a callee (although
+ * clobbering information on the stack is a separate concern)
+ *
+ * Inserting a call to this function at any point in the code
+ * should not produce any failures.
+ */
+
+nothrow
+@safe
+@nogc
+pure
+void clobberRegisters()
+{
+    /* Rely on the inline assembler to know which registers
+     * are to be preserved by the ABI.
+     */
+    version (D_InlineAsm_X86)
+    {
+        asm pure nothrow @safe @nogc
+        {
+            mov EAX,0xDEADBEEF  ;
+            mov EBX,EAX         ;
+            mov ECX,EAX         ;
+            mov EDX,EAX         ;
+            //mov EBP,EAX       ;
+            mov ESI,EAX         ;
+            mov EDI,EAX         ;
+
+            push EAX            ;
+            movd XMM0,[ESP]     ;
+            pshufd XMM0,XMM0,0  ;
+            movdqa XMM1,XMM0    ;
+            movdqa XMM2,XMM0    ;
+            movdqa XMM3,XMM0    ;
+            movdqa XMM4,XMM0    ;
+            movdqa XMM5,XMM0    ;
+            movdqa XMM6,XMM0    ;
+            movdqa XMM7,XMM0    ;
+            pop EAX             ;
+
+            fldlg2              ;
+            fldlg2              ;
+            fldlg2              ;
+            fldlg2              ;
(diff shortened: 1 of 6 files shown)


WalterBright:
> Should go in core/internal/

Since it has utility for users (see comment 2 in the function) it should be user facing, not internal.

WalterBright:
I've been using `printf` as a register stomper, which is clearly silly. This function is better.

thewilsonator:
> it should be user facing

In which case it needs to a) have a much better name, and b) should go though a cycle (or whatever the policy is) of `@__future`.

> I've been using printf as a register stomper, which is clearly silly.

How so? Yes it is perhaps overkill but, so what?

> This function is better.

In what way?

src/core/deadbeef.d
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * Inserting a call to this function at any point in the code
+ * should not produce any failures.
+ */
+
+nothrow
+@safe
+@nogc
+pure
+void deadbeef()
```
JinShil: Seems like a pretty poor name, and I don't see the relationship to uses mentioned here:  https://en.wikipedia.org/wiki/Hexspeak
Maybe something like `clobberRegisters` would be more appropriate.

src/core/deadbeef.d
```diff
@@ -0,0 +1,141 @@
...
+            movdqa XMM6,XMM0    ;
+            movdqa XMM7,XMM0    ;
+            movdqa XMM8,XMM0    ;
+            movdqa XMM9,XMM0    ;
+            movdqa XMM10,XMM0   ;
+            movdqa XMM11,XMM0   ;
+            movdqa XMM12,XMM0   ;
+            movdqa XMM13,XMM0   ;
+            movdqa XMM14,XMM0   ;
+            movdqa XMM15,XMM0   ;
```
JinShil: What about YMM registers for AVX?

WalterBright: I thought for the first version I'd stick with the baseline.

src/core/deadbeef.d
WalterBright: `deadbeef` has a long history of what uninitialized things should be set to, so they stand out in the debugger.

WalterBright:
> In what way?

In the way that it affirmatively sets all the scratch register values, rather than hoping it does.

src/core/deadbeef.d
JinShil: That's fine, but please change the name of the function to something more appropriate.  You might even consider allowing the caller to pass in a value suitable for their needs, perhaps defaulting to `0xdeadbeef`.  Some may prefer to just use `0`.

Geod24:
I don't see why this should be in druntime. It is not required for the language to work. Just put it in Phobos.

src/core/deadbeef.d
WalterBright: Using 0 is a poor idea, since 0 is so often a valid value for a register. The reason for `0xdeadbeef` is because it is highly unlikely to be a correct value, because it shows up in the debugger much like a NaN does. I don't see a point to a user-specified value.

WalterBright:
> I don't see why this should be in druntime.

1. So the compiler can access it for the test suite before relying on a built phobos. Having the test suite rely on successful compilation of Phobos makes for a harder bootstrapping process.
2. It is highly target dependent, and the idea of druntime is the highly target dependent stuff goes there as that's what is expected to be redone for new targets.

> The changelog mentions it for testing the compiler. I suggest moving it there

1. There is no special library for compiler testing.
2. As stated multiple times, it is also useful for users to prevent leakage of information up the program stack. For example, if the function processes plaintext passwords. Come to think of it, it can also be used to prevent leakage downstack.

> needs to add support for other platforms as well

I expect people who support non-x86 platforms to add support for them. It's why it's in druntime.

> It has two commits for no reason

Because I did that by mistake and don't know how to squash them :-( and it didn't seem important.

> Why a new module for this single function?

Because there didn't seem to be an existing module with any semantic connection to its functionality. Besides, Phobos has a bit too much "kitchen sink" aggregations of fairly unrelated functions.

> It has a very poor name

I like the name. But nobody else does, so I'll change it.

wilzbach:
> So the compiler can access it for the test suite before relying on a built phobos. 

The testsuite already relies on a built Phobos. Anyhow, this should be in dmd/test then.

WalterBright:
> The testsuite already relies on a built Phobos.

test/runnable should not. I've tried to remove such dependencies when I found them.

>  Anyhow, this should be in dmd/test then.

Again, **this is useful to users**.

jacob-carlborg:
> Again, this is useful to users.

A lot of things are useful to users which are not in Phobos or druntime. Let's take an example: every package on http://dub.pm.

WalterBright:
> A lot of things are useful to users which are not in Phobos or druntime

This function is not trivial (although it is short) and requires some expertise with function ABIs and target architecture - things a compiler dev team should be expert at, and something not expected of users. Having a portable interface to it across architectures supported by the compiler is of significant value. It *should* be part of the compiler package.

RazvanN7:
@WalterBright what are your thoughts on this PR? Should it be pushed forward?

PR #2625, opened 2019-06-03
Add pthread/math libraries to a couple more rules

marler8997:


marler8997:
A good goal, but we're likely still a ways away from that. Once we can remove the make files in dmd, then I think we can start looking into that for druntime/photos.

In any case, this isn't improving the makefile, it is fixing a rule in the makefile and this fix will also be needed once we move to D.

jacob-carlborg:
I thought bug fixes were improvements 😉

marler8997:
True. Would have been better to say it's not an improvement specific for the makefile.

Hopefully dmd can drop the makefiles sooner rather than later. I've been pushing fixes to build.d to get it ready.

jacob-carlborg:
They way to get someone to move to a new tool/system/whatever is to stop updating the old system and only update the new. Then eventually there will be enough features or bug fixes that exist in the new system and not the old that the user moves to the new system.

In this case this would mean not updating the makefiles and only update the D files. Even if the same bug exists in both.

marler8997:
ready to merge?

PR #2628, opened 2019-06-03
Add unittests for load/storeUnaligned in core.simd

JinShil:
I couldn't find any tests for these in DMD or druntime.  Also, it appears the template constraint was missing the floating point types, but the implementation included them.

### DIFF
--- a/src/core/simd.d
+++ b/src/core/simd.d
@@ -608,7 +608,9 @@ version (D_SIMD)
             is(V == int4) ||
             is(V == uint4) ||
             is(V == long2) ||
-            is(V == ulong2))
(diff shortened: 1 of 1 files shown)


src/core/simd.d
```diff
@@ -619,6 +621,50 @@ version (D_SIMD)
...
     }
 
+    @system
+    unittest
+    {
+        // Memory to load into the vector:
+        // Should have enough data to test all 16-byte alignments, and still
+        // have room for a 16-byte vector
+        ubyte[32] data;
+        for (int i = 0; i < data.length; i++)
```
jacob-carlborg: `foreach`

PR #2632, opened 2019-06-06
Translate _d_arrayappend{T,cTX} to templates

Vild:
dmd PR: https://github.com/dlang/dmd/pull/9982

This PR adds templated version of `_d_arrayappend{T,cTX}`.

## Notes
- There are wrapper around each hook to fake the safety level of the real implementations to not break code.
- I only translated the entrypoints for the hooks and not the called functions. Because this would have taken took long. Future working can be done to rewrite these functions.
- `_d_arrayappend{c,w}d` were not translated because they did not use TypeInfo in their interface. But they were changed to use the new hooks.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -38,6 +38,10 @@ alias dstring = immutable(dchar)[];
 
 version (D_ObjectiveC) public import core.attribute : selector;
 
+/// See $(REF _d_arrayappendTImpl, core,internal,array,appending)
+public import core.internal.array.appending : _d_arrayappendTImpl;
+/// See $(REF _d_arrayappendcTXImpl, core,internal,array,appending)
+public import core.internal.array.appending : _d_arrayappendcTXImpl;
 /// See $(REF __cmp, core,internal,array,comparison)
 public import core.internal.array.comparison : __cmp;
 /// See $(REF __equals, core,internal,array,equality)
--- a/mak/COPY
+++ b/mak/COPY
@@ -33,6 +33,7 @@ COPY=\
 	$(IMPDIR)\core\internal\utf.d \
 	$(IMPDIR)\core\internal\lifetime.d \
 	\
+	$(IMPDIR)\core\internal\array\appending.d \
 	$(IMPDIR)\core\internal\array\comparison.d \
 	$(IMPDIR)\core\internal\array\construction.d \
 	$(IMPDIR)\core\internal\array\equality.d \
--- a/mak/DOCS
+++ b/mak/DOCS
@@ -73,6 +73,7 @@ DOCS=\
 	$(DOCDIR)\core_sys_darwin_mach_thread_act.html \
 	$(DOCDIR)\core_sys_darwin_netinet_in_.html \
 	\
+    $(DOCDIR)\core_internal_array_appending.html \
     $(DOCDIR)\core_internal_array_capacity.html \
     $(DOCDIR)\core_internal_array_casting.html \
     $(DOCDIR)\core_internal_array_comparison.html \
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -33,6 +33,7 @@ SRCS=\
 	src\core\internal\utf.d \
 	src\core\internal\lifetime.d \
 	\
+	src\core\internal\array\appending.d \
 	src\core\internal\array\comparison.d \
 	src\core\internal\array\construction.d \
 	src\core\internal\array\equality.d \
--- a/mak/WINDOWS
+++ b/mak/WINDOWS
@@ -153,6 +153,9 @@ $(IMPDIR)\core\internal\utf.d : src\core\internal\utf.d
 $(IMPDIR)\core\internal\lifetime.d : src\core\internal\lifetime.d
 	copy $** $@
 
+$(IMPDIR)\core\internal\array\appending.d : src\core\internal\array\appending.d
+	copy $** $@
+
 $(IMPDIR)\core\internal\array\comparison.d : src\core\internal\array\comparison.d
 	copy $** $@
 
--- a/src/core/internal/array/appending.d
+++ b/src/core/internal/array/appending.d
@@ -1,0 +1,237 @@
+/**
+ This module contains support for controlling dynamic arrays' appending
+
+  Copyright: Copyright Digital Mars 2000 - 2019.
+  License: Distributed under the
+       $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+     (See accompanying file LICENSE)
+  Source: $(DRUNTIMESRC core/_internal/_array/_appending.d)
+*/
+module core.internal.array.appending;
+
+/// See $(REF _d_arrayappendcTX, rt,lifetime,_d_arrayappendcTX)
+private extern (C) byte[] _d_arrayappendcTX(const TypeInfo ti, ref byte[] px, size_t n) @trusted pure nothrow;
+
+/// Implementation of `_d_arrayappendcTX` and `_d_arrayappendcTXTrace`
+template _d_arrayappendcTXImpl(Tarr : T[], T)
+{
+    import core.internal.array.utils : _d_HookTraceImpl, isPostblitNoThrow;
+
+    private enum errorMessage = "Cannot append to array if compiling without support for runtime type information!";
+
+    /**
+     * Extend an array `px` by `n` elements.
+     * Caller must initialize those elements.
+     * Params:
+     *  px = the array that will be extended, taken as a reference
+     *  n = how many new elements to extend it with
+     * Returns:
+     *  The new value of `px`
+     * Bugs:
+    *   This function template was ported from a much older runtime hook that bypassed safety,
+    *   purity, and throwabilty checks. To prevent breaking existing code, this function template
+    *   is temporarily declared `@trusted pure` until the implementation can be brought up to modern D expectations.
+     */
+    static if (isPostblitNoThrow!T) // `nothrow` deduction doesn't work, so this is needed
+        ref Tarr _d_arrayappendcTX(return scope ref Tarr px, size_t n) @trusted pure nothrow
+        {
+            pragma(inline, false);
+
+            mixin(_d_arrayappendcTXBody);
+        }
+    else
+        ref Tarr _d_arrayappendcTX(return scope ref Tarr px, size_t n) @trusted pure nothrow
+        {
+            pragma(inline, false);
+
+            mixin(_d_arrayappendcTXBody);
+        }
+
+    private enum _d_arrayappendcTXBody = q{
+        version (D_TypeInfo)
+        {
+            auto ti = typeid(Tarr);
+
+            // _d_arrayappendcTX takes the `px` as a ref byte[], but its length
+            // should still be the original length
+            auto pxx = (cast(byte*)px.ptr)[0 .. px.length];
+            ._d_arrayappendcTX(ti, pxx, n);
+            px = (cast(T*)pxx.ptr)[0 .. pxx.length];
+
+            return px;
+        }
+        else
+            assert(0, "Cannot append arrays if compiling without support for runtime type information!");
+    };
+
+    /**
+     * TraceGC wrapper around $(REF _d_arrayappendcTX, rt,array,appending,_d_arrayappendcTXImpl).
+     * Bugs:
+     *  This function template was ported from a much older runtime hook that bypassed safety,
+     *  purity, and throwabilty checks. To prevent breaking existing code, this function template
+     *  is temporarily declared `@trusted pure` until the implementation can be brought up to modern D expectations.
+     */
+    alias _d_arrayappendcTXTrace = _d_HookTraceImpl!(Tarr, _d_arrayappendcTX, errorMessage);
+}
+
+/// Implementation of `_d_arrayappendT` and `_d_arrayappendTTrace`
+template _d_arrayappendTImpl(Tarr : T[], T)
+{
+    import core.internal.array.utils : _d_HookTraceImpl, isPostblitNoThrow;
+
+    private enum errorMessage = "Cannot append to array if compiling without support for runtime type information!";
+
+    /**
+     * Append array `y` to array `x`.
+     * Params:
+     *  x = what array to append to, taken as a reference
+     *  y = what should be appended
+     * Returns:
+     *  The new value of `x`
+     * Bugs:
+    *   This function template was ported from a much older runtime hook that bypassed safety,
+    *   purity, and throwabilty checks. To prevent breaking existing code, this function template
+    *   is temporarily declared `@trusted pure` until the implementation can be brought up to modern D expectations.
+     */
+    static if (isPostblitNoThrow!T)
+        ref Tarr _d_arrayappendT(return scope ref Tarr x, scope Tarr y) @trusted pure nothrow
+        {
+            pragma(inline, false);
+
+            mixin(_d_arrayappendTBody);
+        }
+    else
+        ref Tarr _d_arrayappendT(return scope ref Tarr x, scope Tarr y) @trusted pure
+        {
+            pragma(inline, false);
+
+            mixin(_d_arrayappendTBody);
+        }
+
+    private enum _d_arrayappendTBody = q{
+        import core.stdc.string : memcpy;
+        import core.internal.traits : Unqual;
+
+        auto length = x.length;
+        auto sizeelem = T.sizeof;
+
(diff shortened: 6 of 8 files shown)


thewilsonator:
This uses `scope`, this should also be using `return ref` if I am understating the code correctly. This may alleviate the need for trusted.

src/object.d
```diff
@@ -4956,3 +4956,315 @@ string _d_assert_fail(string comp, A, B)(A a, B b) @nogc @safe nothrow pure
...
+    auto result = _d_arrayappendcTX(px, 1);
+    memcpy(cast(Unqual!T *)&result[$ - 1], cast(Unqual!T *)&y, T.sizeof);
+    return result;
+}
+
+/**
+ * Run postblit on `t` if it is a struct and needs it.
+ * Or if `t` is a array, run it on the children if they have a postblit.
+ */
+void __doPostblit(T)(auto ref T t)
```
JinShil: Can this be `private` in this module?  Also, why the `__` prefix?

src/object.d
```diff
@@ -4956,3 +4956,315 @@ string _d_assert_fail(string comp, A, B)(A a, B b) @nogc @safe nothrow pure
...
+            );
+    }
+    accumulate(file, line, funcname, typename, T.sizeof * y.length);
+    return _d_arrayappendT(x, y);
+}
+
+/**
+ * Append y[] to array x[]
+ */
+T[] _d_arrayappendT_(T)(ref T[] x, scope T[] y) @trusted
```
JinShil: It looks like this is only used in one place.  What about embedding it in the caller as a nested function to make it more obvious that this is an implementation detail of that function?

src/object.d
```diff
@@ -4956,3 +4956,315 @@ string _d_assert_fail(string comp, A, B)(A a, B b) @nogc @safe nothrow pure
...
+    }
+    accumulate(file, line, funcname, typename, T.sizeof * n);
+    return _d_arrayappendcTX(px, n);
+}
+
+/**************************************
+ * Extend an array by n elements.
+ * Caller must initialize those elements.
+ */
+T[] _d_arrayappendcTX_(T)(ref T[] px, size_t n) @trusted
```
JinShil: Same for this.  I'm recommending that this be embedded as a nested function in its only caller to keep the implementations together.

src/object.d
```diff
@@ -4956,3 +4956,315 @@ string _d_assert_fail(string comp, A, B)(A a, B b) @nogc @safe nothrow pure
     buffer[n .. n + valB.length] = valB;
     return (() @trusted => cast(string) buffer)();
 }
+
+// These are needed to fake the safety level to `@safe pure nothrow`.
+// Without this code will break as these hooks where previously exported
+// as having this safety level.
+T[] _d_arrayappendT(T)(ref T[] x, scope T[] y) @trusted
```
JinShil: Do we need to return `T[]`?

JinShil: Is the `@trusted` on this declaration redundant?  It seems that since the call to `_d_arrayappendT_` is wrapped `@trusted.` the  `@trusted` attribute at entry point is not necessary.

thewilsonator: > Do we need to return T[]?

For chaining?

src/object.d
```diff
@@ -4956,3 +4956,315 @@ string _d_assert_fail(string comp, A, B)(A a, B b) @nogc @safe nothrow pure
...
+    enum size_t LARGEPREFIX = 16;
+    enum size_t LARGEPAD = LARGEPREFIX + 1;
+
+    // Can not use `externDFunc` as that will cause a template instantiation loop.
+    template externDFuncMangle(string mangle, string func)
+    {
+        pragma(mangle, mangle) mixin(func);
+    }
+
+    alias __arrayStart = externDFuncMangle!("_D2rt8lifetime12__arrayStartFNaNbS4core6memory8BlkInfo_ZPv", "extern(D) void * externDFuncMangle(BlkInfo info) nothrow pure;");
```
JinShil: Looks like we have some structural problems in druntime, if this is required.  I don't know the solution at this time.

JinShil: Can a lot of this logic be encapsulated in a function in `rt.lifetime` so most of this can be replaced with a single function call, and avoid having to expose so many internals of `rt.lifetime` to  here?

JinShil:
> Should I keep the changes to _d_arrayappend{c,w}d for this PR?
> If those changes gets removed and this PR and dmds PR gets merged. The old code will still be used when assigning wchar and dchars to char[]. But then yet again, those changed do not really belong in this PR.
Not sure should that have been a prerequisite to this PR?

jpf91:
>libphobos2.a(process.o):std/process.d:_D3std7process12__ModuleInfoZ: error: undefined reference to '_D3std8internal6memory12__ModuleInfoZ'
> I have not figured out what causes this to happen. I guess it is something related to the templates using internal imports, but I'm not sure. I would appreciate some help.

The really strange thing here and the reason this fails to link is that `std.internal.memory` is in phobos. 
Can you run `process.o` through objdump and see where exactly the symbol is referenced?

> _D4core3sys7openbsdQm5cdefs12__ModuleInfoZ

Also strange The `core` modules are not compiled into the library, so there's not ModuleInfo, TypeInfo, ... emitted for these. There's nothing in `cdefs.d` which would need ModuleInfo though and I wonder why it even pulls in OpenBSD code?

jacob-carlborg:
> Also strange The core modules are not compiled into the library, so there's not ModuleInfo, TypeInfo, ... emitted for these. 

I’ve run into this before. The C bindings are not being compiled. So the few macros that have been translated to functions are not compiled and causes linker errors.

jacob-carlborg:
> There's nothing in cdefs.d which would need ModuleInfo though and I wonder why it even pulls in OpenBSD code?

Perhaps compiling with `-vgc-ast` will reveal something.

src/object.d
Vild: Yes, I can add private. The `__` is because it called this inside rt.lifetime: https://github.com/dlang/druntime/blob/master/src/rt/lifetime.d#L701

src/object.d
Vild: Good suggestion, will do.

src/object.d
Vild: It returns `T[]` because the original hook returned `void[]`. I also just found a problem with the current implementation in dmd:
```d
int[] arr;
((arr ~= [1,2,3]) ~= 4) ~= 5;
```
This crashes in the backend for dmd (https://run.dlang.io/is/UZAIuX), while working as expected in ldc (`arr == [1, 2, 3, 4, 5]`).
So I should probably change this to return a `ref T[]`, to match ldc's behavior.

I think I still need `@trusted` for the outer functions as I get this error if I remove `@trusted` on `_d_arrayappendcT`.
```
std/uni.d(6165): Error: @safe function std.uni.Stack!(InversionList!(GcPolicy)).Stack.push cannot call @system function object._d_arrayappendcT!(InversionList!(GcPolicy))._d_arrayappendcT
```

src/object.d
Vild: One thing I could do is that I could call `rt.lifetime._d_arrayappendcTX` directly instead of porting it. As I'm not sure I have time to start refactoring that function for this 2 week-block.

Vild:
> > Should I keep the changes to _d_arrayappend{c,w}d for this PR?
> > If those changes gets removed and this PR and dmds PR gets merged. The old code will still be used when assigning wchar and dchars to char[]. But then yet again, those changed do not really belong in this PR.
> 
> I'm not sure. Should that have been a prerequisite to this PR?

I think I will remove that commit, and then apply it later when I do a clean-up PR.

src/object.d
```diff
@@ -4956,3 +4956,315 @@ string _d_assert_fail(string comp, A, B)(A a, B b) @nogc @safe nothrow pure
...
+    memcpy(cast(Unqual!T *)&x[length], cast(Unqual!T *)&y[0], y.length * sizeelem);
+
+    // do postblit
+    __doPostblit(cast(Unqual!T[])x[length .. length + y.length]);
+    return x;
+}
+
+T[] _d_arrayappendcTXTrace(T)(string file, int line, string funcname, ref T[] px, size_t n) @trusted pure
+{
+    import rt.profilegc : accumulate;
```
rainers: Note that the implementation of the tracing functions has changed recently, see https://github.com/dlang/druntime/pull/2607

I wouldn't mind if the trace-functions are left unmodified, it's a half-baked feature anyway and slows down execution in accumulate.

src/object.d
jpf91: Maybe that's best for now. If the compiler instantiates the template (wrapper), we can always port the implementation later on.

src/object.d
Vild: The problem is that the new entrypoints for the hooks are templates, so I need to manually write a trace entrypoint. It also need to be in object.d so the compiler can reference it from the lowering code. https://github.com/dlang/dmd/pull/9982/files#diff-0c1aa1ff99402eb87e11cc3fa1b80da0

Vild:
I've updated the code according to the comments, but the unittest will still fail. This time because when I changed the compiler to rewrite `arr ~= elem` to `_d_arrayappendcT(arr, elem)` it will call the postblit twice on elem, once before the hook is called, and once during the hook. Here I either need to figure out how to never make it call the postblit before calling the function, or revert to the previous behavior where the assignment step is done outside the hook.

Vild:
Update:
All the druntime unittests passes.
In phobos everything related to `CodepointSet` fails, like this one: https://github.com/dlang/phobos/blob/master/std/uni.d#L6916

posix.mak
```diff
@@ -176,6 +176,9 @@ $(DOCDIR)/core_sys_darwin_netinet_%.html : src/core/sys/darwin/netinet/%.d $(DMD
 $(DOCDIR)/rt_%.html : src/rt/%.d $(DMD)
 	$(DMD) $(DDOCFLAGS) -Df$@ project.ddoc $(DOCFMT) $<
 
+$(DOCDIR)/rt_array_array.html : src/rt/array/array.d $(DMD)
```
JinShil: I think you mean `src/rt/array/append.d`.

src/object.d
```diff
@@ -49,6 +49,14 @@ public import rt.array.capacity: reserve;
 /// See $(REF assumeSafeAppend, rt,array,capacity)
 public import rt.array.capacity: assumeSafeAppend;
 
+/// See $(REF capacity, rt,array,capacity)
+public import rt.array.capacity: capacity;
+
+/// See $(REF _d_arrayappendT, rt,array,append) and $(REF _d_arrayappendTTrace, rt,array,append)
```
JinShil: You actually don't need to document these, because they are not called directly by the user.  `capacity.d` has functions that are called directly by the user, so it is an exception.  I'd like to add a `version(Core_Doc)` block to filter out the runtime hooks so they don't appear in the documentation.

posix.mak
Vild: Thanks :)

src/object.d
Vild: Okey, thanks.

It think it is better to do the `version(Core_Doc)` in a separate PR.

src/object.d
```diff
@@ -41,6 +41,7 @@ version (D_ObjectiveC) public import core.attribute : selector;
 public import rt.array.comparison : __cmp;
 public import rt.array.equality : __ArrayEq, __equals;
 public import rt.array.casting: __ArrayCast;
+public import rt.array.append : _d_arrayappendT, _d_arrayappendTTrace,_d_arrayappendcTX, _d_arrayappendcTXTrace;
```
JinShil: Nit: space after comma

src/rt/array/append.d
```diff
@@ -0,0 +1,290 @@
...
+    alias func = ref T[] delegate(string file, int line, string funcname, return scope ref T[] x, scope T[] y) @trusted pure nothrow;
+    auto f = cast(func)&impl;
+    return f(file, line, funcname, x, y);
+}
+
+// This function cannot be imported because then rt.lifetime, and its dependencies needs to be included in the druntime
+// release. It also cannot be core.internal.externDFunc'd because that causes recursive template instantiation.
+
+/// See $(REF _d_arrayappendcTX, rt,lifetime,_d_arrayappendcTX)
+private extern (C) byte[] _d_arrayappendcTX(const TypeInfo ti, ref byte[] px, size_t n);
```
JinShil: I think I commented about this before, but I can't find my comment now.  I think this can be imported instead of being declared as an `extern`

Vild: I don't think it can because it will be instantiated inside the template and rt.lifetime not part of a dmd installation.

JinShil: Ok, it looks like we'll eventually need to change the DMD installation to make all this templating of druntime work, but that is out of scope of the task at hand.

src/rt/array/append.d
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * Run postblit on `t` if it is a struct and needs it.
+ * Or if `t` is a array, run it on the children if they have a postblit.
+ */
+private void __doPostblit(T)(auto ref T t)
+{
+    import core.internal.traits : hasElaborateCopyConstructor;
+
+    static if (is(T == struct)) {
```
JinShil: Curly brace on next line

src/rt/array/append.d
```diff
@@ -0,0 +1,290 @@
...
+ *   The safety level of this function is faked. It shows itself as `@trusted pure nothrow` to not break existing code.
+ */
+ref T[] _d_arrayappendcTX(T)(return scope ref T[] px, size_t n) @trusted
+{
+    ref T[] impl(return scope ref T[] px, size_t n) @trusted
+    {
+        // This template will be instanciated even in betterC due to the lowering code.
+        // So we need to disallow the calling of this function if that happens.
+        version (D_BetterC)
+            assert(0, "_d_arrayappendcTX does not support betterC!");
```
JinShil: What happens if you don't specifically have this check.  Won't the -betterC build just emit a generic error about not supporting `TypeInfo`?

JinShil: Also, if you do choose to keep this message, I don't think it should specifically mention `_d_arrayappendcTX` because that never appears in the user's code.  I suggest something worded more as "betterC does not support appending arrays".  

And I think that it can be a static assert at the beginning of the `_d_arrayappendcTX` template so the user gets a compile-time assertion failure instead of a runtime assertion failure.

Vild: So without this check `test19561.d` will fail with `Error: TypeInfo cannot be used with -betterC`, even if that expression will be intercepted by the CTFE later.

and it cannot be a static assert because that breaks the lowering code.
```
import/rt/array/append.d(121): Error: static assert:  "_d_arrayappendcTX does not support betterC!"
import/rt/array/append.d(35):        instantiated from here: _d_arrayappendcTX!(immutable(char))
import/core/internal/arrayop.d(320):        instantiated from here: _d_arrayappendT!(immutable(char))
import/core/internal/arrayop.d(45):        instantiated from here: initScalarVecs!(int, "+=")
import/object.d(3696):        instantiated from here: arrayOp!(int[], int, "+=")
```

src/rt/array/append.d
```diff
@@ -0,0 +1,292 @@
...
+        version (D_BetterC)
+            assert(0, "The runtime does not support appending arrays in betterC!");
+        else
+        {
+            static if (is(T == shared))
+                auto ti = typeid(shared(T[]));
+            else
+                auto ti = typeid(T[]);
+
+            // _d_arrayappendcTX takes the `px` as a ubyte[], but its length
```
jpf91: nitpick: as a byte[]

jpf91:
So as it seems we can't avoid the `D_BetterC` workaround right now, after changing the version condition to `D_TypeInfo`, is there anything left to be done here or is this ready?

Code LGTM.

Vild:
> [...] is there anything left to be done here or is this ready?
I did the last few fixes and I'm happy with it for now.
In the future I would like to add `nothrow` deducting, but as it still goes though the old hooks it is pretty hard to add right now.

Vild:
This PR needs to implements the same fixes as in https://github.com/dlang/druntime/pull/2667, before it should be merged

RazvanN7:
@Vild Can you rebase this to get rid of the conflicts and see if we can get this in?

Vild:
> @Vild Can you rebase this to get rid of the conflicts and see if we can get this in?

This PR requires https://github.com/dlang/druntime/pull/2673.
So as soon as I fix that one and it gets merge I will rebase this one.

thewilsonator:
> and it gets merge

Done.

Vild:
This PR is ready to be merged

PR #2633, opened 2019-06-06
Remove __cmp from object.d for aggregates

JinShil:
Surprisingly it doesn't appear to be utilized anywhere.  Let's see what the test suite says.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -111,20 +111,6 @@ int __cmp(T)(scope const T[] lhs, scope const T[] rhs) @trusted
     }
 }
 
-// Compare class and interface objects for ordering.
-private int __cmp(Obj)(Obj lhs, Obj rhs)
-if (is(Obj : Object))
-{
-    if (lhs is rhs)
-        return 0;
-    // Regard null references as always being "less than"
-    if (!lhs)
-        return -1;
-    if (!rhs)
-        return 1;
-    return lhs.opCmp(rhs);
-}
-
 // This function is called by the compiler when dealing with array
 // comparisons in the semantic analysis phase of CmpExp. The ordering
 // comparison is lowered to a call to this template.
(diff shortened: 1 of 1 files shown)


JinShil:
Looks like it was introduced here:  https://github.com/dlang/druntime/pull/1787

JinShil:
Please don't merge this PR until @andralex responds.  I'm wondering if we need to add a lowering to DMD instead.

andralex:
The plan here was to have the compiler lower object comparison to this. Also @edi33416 is working on defining `__cmp` for structs as well (lexicographic in field order).

JinShil:
Thank you @andralex Closing pending a DMD PR that will utilize this.

JinShil:
I'm reopening this.

- `RTLSYM_OBJ_CMP` is [no longer used in the compiler](https://github.com/dlang/dmd/blob/1437052671306e289cd21a862c662b341e66ff7f/src/dmd/backend/rtlsym.d#L110)
- The old runtime hook [`_d_obj_cmp`](https://github.com/dlang/druntime/blob/e47a00bff935c3f079bb567a6ec97663ba384487/src/rt/obj.d#L32) no longer exists in druntime.  In fact the entire file `rt/obj.d` no longer exists in druntime.
- It appears that comparing objects in the runtime [has been superceded by `opCmp`](https://run.dlang.io/is/uPUOQD)

I don't know what the plan is for the code that this PR removes.  @andralex, If this is not dead code, could you please elaborate on what needs to be done to utilize it.

edi33416:
I have this dmd [pr](https://github.com/dlang/dmd/pull/9629) and druntime [pr](https://github.com/dlang/druntime/pull/2562) to lower calls to `__cmp`. This was influenced by work at `ProtoObject`. Should I still keep those or two `ProtoObjects` shouldn’t be compared and have the user do a dynamic/static cast at call site?

JinShil:
Thank you @edi33416 It appears the utility of this code has yet to be determined and is pending the adjudication of work on ProtoObject.  I'll close this for now until those design decisions are made.

JinShil:
> Should I still keep those or two ProtoObjects shouldn’t be compared and have the user do a dynamic/static cast at call site?

andralex:
@edi33416 we need that lowering to compare properly null objects. Otherwise code like this will crash:

```d
class Widget { ... defines opCmp ... }
auto w1 = new Widget;
Widget w2 = null;
if (w2 < w1) {}
```

If all we do is lower to `w1.opCmp(w2)` we end up calling a method against a `null` reference. So please proceed and be bold.

PR #2638, opened 2019-06-10
 Implement druntime side of DIP1014

lesderid:
This PR implements the druntime side of [DIP1014](https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1014.md).

The accepted DIP is a little ambiguous about whether `nothrow` should be required on `opPostMove` (i.e. applied to `__move_post_blt`). Quoting: 

> `opPostMove`, if implemented, MUST be `nothrow`.
>
> [...]
> This proposal does require that `nothrow` be defined on `opPostMove`, because throwing would mean a change in the program flow from a place that appears to execute no code. There is a danger that this will be too confusing for the programmer to properly take into account.
>
> [...]
>
> The DIP author suggested the requirement could be eased to "a very hearty recommendation" because D moves "everywhere" and it is "impossible" to prevent it from doing so.

cc @Shachar

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -294,7 +294,7 @@ template Fields(T)
         alias Fields = TypeTuple!T;
 }
 
-// std.traits.hasElaborateMove
+/// See $(REF hasElaborateMove, std,traits)
 template hasElaborateMove(S)
 {
     static if (__traits(isStaticArray, S) && S.length)
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -294,6 +294,25 @@ template Fields(T)
         alias Fields = TypeTuple!T;
 }
 
+// std.traits.hasElaborateMove
+template hasElaborateMove(S)
+{
+    static if (__traits(isStaticArray, S) && S.length)
+    {
+        enum bool hasElaborateMove = hasElaborateMove!(typeof(S.init[0]));
+    }
+    else static if (is(S == struct))
+    {
+        enum hasElaborateMove = (is(typeof(S.init.opPostMove(lvalueOf!S))) &&
+                                    !is(typeof(S.init.opPostMove(rvalueOf!S)))) ||
+                                anySatisfy!(.hasElaborateMove, Fields!S);
+    }
+    else
+    {
+        enum bool hasElaborateMove = false;
+    }
+}
+
 // std.traits.hasElaborateDestructor
 template hasElaborateDestructor(S)
 {
--- a/src/object.d
+++ b/src/object.d
@@ -127,6 +127,23 @@ if (is(Obj : Object))
     assert(a <  "я");
 }
 
+/**
+ * Recursively calls the `opPostMove` callbacks of a struct and its members if
+ * they're defined.
+ *
+ * When moving a struct instance, the compiler emits a call to this function
+ * after blitting the instance and before releasing the original instance's
+ * memory.
+ *
+ * Params:
+ *      newLocation = reference to struct instance being moved into
+ *      oldLocation = reference to the original instance
+ *
+ * Note:
+ *      This function is tentatively defined as `nothrow` to prevent
+ *      `opPostMove` from being defined without `nothrow`, which would allow
+ *      for possibly confusing changes in program flow.
+ */
 void __move_post_blt(S)(ref S newLocation, ref S oldLocation) nothrow
     if (is(S == struct))
 {
--- a/src/object.d
+++ b/src/object.d
@@ -127,6 +127,68 @@ if (is(Obj : Object))
     assert(a <  "я");
 }
 
+void __move_post_blt(S)(ref S newLocation, ref S oldLocation) nothrow
+    if (is(S == struct))
+{
+    static foreach (memberName; __traits(allMembers, S))
+    {
+        static if (is(typeof(__traits(getMember, S, memberName)) == struct))
+        {
+            __move_post_blt(__traits(getMember, newLocation, memberName), __traits(getMember, oldLocation, memberName));
+        }
+    }
+
+    static if (__traits(hasMember, S, "opPostMove"))
+    {
+        import core.internal.traits : lvalueOf, rvalueOf;
+        static assert( is(typeof(S.init.opPostMove(lvalueOf!S))) &&
+                      !is(typeof(S.init.opPostMove(rvalueOf!S))),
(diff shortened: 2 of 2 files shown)


Shachar:
I don't think I'm authorize to make a definite decision on this point, but here's my hearty recommendation. I'll give my rational.

Why it should require `nothrow`:

- The rational given in the DIP still stands, and it is only through people actually using the functionality that we will find out how restrictive it is.
- Changing behavior in the language is easier from code disallowing something to code allowing something, as that direction creates no backwards compatibility problems.

Why it should not require `nothrow`:

- I fear that the implementation might become sacred, with the above rational lost to memory. This is especially since I am personally no longer involved in D.

So my suggestion is to require `nothrow`, but to document that this is an interim decision in both commit message and in the code itself, with reference to this discussion.

src/object.d
```diff
@@ -529,6 +529,68 @@ bool __equals(T1, T2)(T1[] lhs, T2[] rhs)
...
 }
 
+void __move_post_blt(S)(ref S newLocation, ref S oldLocation) nothrow
+    if (is(S == struct))
+{
+    static foreach (memberName; __traits(allMembers, S))
+    {
+        static if (is(typeof(__traits(getMember, S, memberName)) == struct))
+        {
+            mixin("__move_post_blt(newLocation." ~ memberName ~ ", oldLocation." ~ memberName ~ ");");
```
jacob-carlborg: Can `__traits(getMember)` be used instead of the string mixin?

src/object.d
```diff
@@ -529,6 +529,68 @@ bool __equals(T1, T2)(T1[] lhs, T2[] rhs)
     assert(a1 != a2);
 }
 
+void __move_post_blt(S)(ref S newLocation, ref S oldLocation) nothrow
```
jacob-carlborg: Can this be private and have the compiler bypass the protection?

lesderid: It can't be private because `std.algorithm.mutation.move` has to be able to call it:

> The `move` family of functions defined in `std.algorithm` MUST be updated to add a call to `__move_post_blt`.

src/object.d
lesderid: Good call, fixed.

12345swordy:
cc @thewilsonator  @TurkeyMan

TurkeyMan:
Yeah it'd be really good if this got merged!

lesderid:
@thewilsonator Rebased it.

What about the `nothrow` though?

lesderid:
> What about the `nothrow` though?

Actually it's probably fine like this for now.

src/object.d
```diff
@@ -127,6 +127,68 @@ if (is(Obj : Object))
     assert(a <  "я");
 }
 
+void __move_post_blt(S)(ref S newLocation, ref S oldLocation) nothrow
```
JinShil: Please add a complete DDoc header to document this for other developers.

src/core/internal/traits.d
```diff
@@ -294,6 +294,25 @@ template Fields(T)
         alias Fields = TypeTuple!T;
 }
 
+// std.traits.hasElaborateMove
+template hasElaborateMove(S)
```
JinShil: Please add a complete DDoc header to document this for other developers.

JinShil:
> What about the nothrow though?

I suggest making choosing the most restrictive option for now; it can always be relaxed later after there's more experience with it. It can always be relaxed later.

src/core/internal/traits.d
lesderid: The other traits in this module have no ddoc either, but they are all documented on the Phobos side (see https://github.com/dlang/phobos/pull/7075/files#diff-3de50c374a825bac70efe9457fa16e85R3795 for this one), presumably because this is an internal module. AFAICT docs aren't generated for `core.internal`, so using `$(REF ...)` wouldn't really make a difference.

JinShil: Actually, `hasElaborateMove` should go into `core.internal.traits` and `__move_post_blt` should go somewhere in `rt`, but we haven't yet established a strong precedent for that yet.  I don't care if that's done in this PR;  I will probably do it later in a followup if this PR is merged.

The documentation for `rt` is generated for the "Internal API" here:  https://dlang.org/phobos/index.html  The documentation there isn't really for the user; it's for other developers.  Even if the documentation is not published on the website, those developers who utilize a modern editor can see the DDoc comments at each usage of the function, and that makes it quite convenient to browse and understand the code, so that is the motivation for documenting the implementations in this PR.

What was done in the past is not always the best practice to follow.  Some of us are trying to raise the bar for any new code submissions.  Please help the next person that has to maintain this code by documenting the interface.  You can see other contributors are doing this recently with the array runtime hooks, and that would be the preferred precedent to follow.

https://github.com/dlang/druntime/pull/2655/files
https://github.com/dlang/druntime/pull/2648/files
https://github.com/dlang/druntime/pull/2632/files

src/object.d
JinShil: Those of us that are trying to templatize the compiler<-->runtime interface plant to use the naming convention with a `_d_` prefix for runtime hooks.  The `__` prefix is reserved for compiler generated symbols and these aren't compiler-generated.  I realize it's splitting hairs somewhat because the compiler must be intimately aware of these symbols, but that's what we would prefer.  I realize that the `_d_` convention has not been followed well, and I plan to do something about that soon, but the `_d_` convention is what we've discussed and would prefer.

I don't expect you to change this PR to follow the `_d_` convention, though I wouldn't mind it either.  We have others to change, and we can change this with them when the time comes.  It's just an FYI to be aware of and to consider.

cc @wilzbach

lesderid: > I don't expect you to change this PR to follow the `_d_` convention, though I wouldn't mind it either. We have others to change, and we can change this with them when the time comes. It's just an FYI to be aware of and to consider.

I was aware of this when I implemented it, but I decided to stay as close as possible to the interface and implementation as defined by the DIP. I agree it should be changed to `_d_` in a later PR.

lesderid: Ddoc added.

src/core/internal/traits.d
lesderid: > The documentation there isn't really for the user; it's for other developers. Even if the documentation is not published on the website, those developers who utilize a modern editor can see the DDoc comments at each usage of the function, and that makes it quite convenient to browse and understand the code, so that is the motivation for documenting the implementations in this PR.

That's fair. I've added a `$(REF ...)`.

PR #2639, opened 2019-06-12
issue 19949 - C++ Mangling: add support for abi-tags from the Itanium ABI

SSoulaimane:
Special UDA for mangling `extern(C++)` symbols.
Required by https://github.com/dlang/dmd/pull/9995 to fix issue [19949](https://issues.dlang.org/show_bug.cgi?id=19949).


src/core/attribute.d
```diff
@@ -55,3 +55,60 @@ version (D_ObjectiveC) struct selector
...
+ * }
+ *
+ * @gnu_abi_tag("B")
+ * __gshared A var;
+ *
+ * @gnu_abi_tag("E")
+ * enum ABC { a, b, c }
+ * ---
+ */
+version (Posix) struct gnu_abi_tag
```
JinShil: Please prefer naming conventions documented in the [D Style](https://dlang.org/dstyle.html).

src/core/attribute.d
```diff
@@ -55,3 +55,60 @@ version (D_ObjectiveC) struct selector
...
+{
+    string[] tags;
+
+    @disable this();
+
+    this(string[] tags...)
```
jacob-carlborg: It looks like at least one string is required. In that case, I suggest reflecting that in the signature by specifying one string parameter followed by a variadic parameter.

src/core/attribute.d
```diff
@@ -55,3 +55,60 @@ version (D_ObjectiveC) struct selector
...
+ * {
+ *     @gnu_abi_tag("C")
+ *     void method() {}
+ * }
+ *
+ * @gnu_abi_tag("B")
+ * __gshared A var;
+ *
+ * @gnu_abi_tag("E")
+ * enum ABC { a, b, c }
```
jacob-carlborg: Perhaps an example passing multiple strings.

src/core/attribute.d
SSoulaimane: Updated.
I didn't use concatenation because it doesn't work with `-betterC`, even though it's compile time only.

SSoulaimane:
How about something like this:
```d
version (Posix)
    version = UdaGNUAbiTag;

version (D_ObjectiveC)
    version = UdaSelector;

version (CoreDdoc)
{
    version = UdaGNUAbiTag;
    version = UdaSelector;
}
```

jacob-carlborg:
> How about something like this:

Hmm, that might work. I was actually only referring to the documentation you added but if you want to fix the `selector` UDA as well, that would be great.

src/core/attribute.d
```diff
@@ -51,7 +63,67 @@ module core.attribute;
  * }
  * ---
  */
-version (D_ObjectiveC) struct selector
+version (UdaSelector) struct selector
```
Geod24: Please split this change in its own commit

src/core/attribute.d
```diff
@@ -51,7 +63,67 @@ module core.attribute;
...
+ * in order to distinguish from an earlier version with a different ABI.
+ * See: $(LINK2 https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Attributes.html, GCC attributes documentation).
+ *
+ * This is a special compiler recognized attribute, it has a few
+ * requirements, which all will be enforced by the compiler:
+ *
+ * $(UL
+ *  $(LI
+ *      The attribute can only be attached to a declaration of
+ *      an $(D_CODE extern(C++)) variable, function, aggregate,
```
Geod24: "The attribute can only be attached to an `extern(C++)` declaration" ?

src/core/attribute.d
```diff
@@ -51,7 +63,67 @@ module core.attribute;
...
+ *
+ *  $(LI
+ *      The string arguments must only contain valid characters
+ *      for C++ name mangling which currently include alphanumerics
+ *      and the underscore character.
+ *  ),
+ * )
+ *
+ * Note: Unlike other attributes, if applied to a namespace,
+ * this attribute it is not inherited by the members.
```
Geod24: >  Tags on an inline namespace are not represented in the mangled name of the namespace, but they are subject to the above tag propagation

https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.abi-tag

> The attribute can also be applied to an inline namespace, but does not affect the mangled name of the namespace; in this case it is only used for -Wabi-tag warnings and automatic tagging of functions and variables. Tagging inline namespaces is generally preferable to tagging individual declarations, but the latter is sometimes necessary, such as when only certain members of a class need to be tagged.

https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Attributes.html

So it should be inherited by the member. But that's a C++ side thing, we should represent this with `@gnuAbiTag("XXX") { /* decls */ }` not by trying to mimic the C++ code.
I would just remove this sentence.

SSoulaimane: I don't understand what you mean.
That sentence is about the current behavior of DMD of inheriting UDAs for a namespace.
for example:
```d
@("abc")
extern (C++, "namespace")
{
    void fun();
}
```
`fun` has the attribute `"abc"`.

while it is not the same in the following case:
```d
@("abc")
extern (C++)
class C
{
    void fun();
}
```
`fun` doesn't have the attribute `"abc"`.

SSoulaimane: This special UDA is not supposed to be inherited from a namespace.

Geod24: Yes and that makes sense. But what your sentence reads to me is that in this case:
```d
@gnuAbiTag("Test")
extern (C++, "Test") {
    void fun();
}
```
`fun` would not have the `funAbiTag("Test")` attribute.

SSoulaimane: Yes that's what I meant, It is handled specially by the compiler that's why it's documented.

Geod24:
There were several issues raised with this.

thewilsonator:
Superseded by #2990

PR #2643, opened 2019-06-17
Refactor object.d: Move __ArrayCast to rt

JinShil:
Followup to https://github.com/dlang/druntime/pull/2634

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -405,5 +405,6 @@ COPY=\
 	\
 	$(IMPDIR)\rt\array\comparison.d \
 	$(IMPDIR)\rt\array\equality.d \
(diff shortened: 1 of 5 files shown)


n8sh:
Is putting each function in a separate file the direction this is going? One file for `__equals`, one file for ` __ArrayCast`?

JinShil:
> Is putting each function in a separate file the direction this is going?

No.  The idea is to leverage modules and packages as D's unit of encapsulation.  Those functions/features/etc. that don't depend on each other should go into their own module.

When I attempt to move `dup`, `__dopostblit`, `_trustedDup`, etc., those would go together in one module, because there is a dependency between them.

PR #2646, opened 2019-06-18
[WIP] Add rcarray (using __RefCount)

lesderid:
This PR adds an array/slice type (`rcarray!T`) that behaves like built-in arrays (`T[]`) but manages its own memory through reference counting.

The implementation is largely based on https://github.com/dlang/druntime/pull/2348 but uses the upcoming `__RefCount` struct for reference counting. (This PR contains the `__RefCount` commits as they aren't merged yet, but please review them separately: https://github.com/dlang/druntime/pull/2608.)

Left to do:
- [x] Fix arrays of class types
- [x] Add more comments and documentation
- [x] Add factory function to work around no-arg ctors
- [ ] Benchmark and compare vs similar types
- [x] Check for memory errors with asan

### DIFF
--- a/src/core/experimental/array.d
+++ b/src/core/experimental/array.d
@@ -330,11 +330,24 @@ struct rcarray(T)
 
         if (support !is null)
         {
-            () @trusted { pureDeallocate(support); }();
+            if (rc.isUnique)
+            {
+                //if we had a non-empty array and we hold the only reference to it,
+                //then we can free the original (smaller) array
+                () @trusted { pureDeallocate(support); }();
+            }
+            else
+            {
+                //if there are others holding references to the original array,
+                //then we make a new reference count for the new (larger) array
+                auto newrc = __RefCount.make!(typeof(rc));
+                rc = newrc;
+            }
         }
 
         support = tmpSupport;
         payload = (() @trusted => cast(T[])(support[0 .. payload.length]))();
+
         assert(capacity >= n);
     }
 
--- a/src/core/experimental/array.d
+++ b/src/core/experimental/array.d
@@ -604,6 +604,15 @@ struct rcarray(T)
         return this;
     }
 
+    ///
+    static if (is(T == int))
+    @safe unittest
+    {
+        auto stuff = [1, 2, 3];
+        auto a = immutable rcarray!int(stuff);
+        assert(a[] == stuff);
+    }
+
     /**
     Return a slice to the current array that is bounded by `start` and `end`.
     `start` must be less than or equal to `end` and `end` must be less than
@@ -629,7 +638,6 @@ struct rcarray(T)
     {
         auto stuff = [1, 2, 3];
         auto a = immutable rcarray!int(stuff);
-        assert(a[] == stuff);
         assert(a[1 .. $] == stuff[1 .. $]);
     }
 
--- a/src/core/experimental/array.d
+++ b/src/core/experimental/array.d
@@ -158,36 +158,25 @@ struct rcarray(T)
         mixin(immutableInsert!(typeof(rhs.payload), "rhs.payload")());
     };
 
+    //Copy construct mutable from mutable
     this(return scope ref typeof(this) rhs)
     {
         mixin(copyCtorIncRef);
     }
 
-    this(return scope ref typeof(this) rhs) const
+    //Copy construct const from inout
+    this(return scope inout ref typeof(this) rhs) const
     {
         mixin(copyCtorIncRef);
     }
 
-    this(return scope const ref typeof(this) rhs) const
-    {
-        mixin(copyCtorIncRef);
-    }
-
-    this(return scope immutable ref typeof(this) rhs) const
-    {
-        mixin(copyCtorIncRef);
-    }
-
-    this(return scope ref typeof(this) rhs) immutable
+    //Copy construct immutable from mutable/const
+    this(return scope inout ref typeof(this) rhs) immutable
     {
         mixin(copyCtorAlloc);
     }
 
-    this(return scope const ref typeof(this) rhs) immutable
-    {
-        mixin(copyCtorAlloc);
-    }
-
+    //Copy construct immutable from immutable
     this(return scope immutable ref typeof(this) rhs) immutable
     {
         mixin(copyCtorIncRef);
--- a/src/core/experimental/array.d
+++ b/src/core/experimental/array.d
@@ -364,17 +364,17 @@ struct rcarray(T)
 
         foreach (i; 0 .. tmpSupport.length)
         {
-            import core.lifetime : emplace;
+            import core.internal.lifetime : emplaceRef;
 
             if (i < payload.length)
             {
                 // Copy the existing items
-                emplace(&tmpSupport[i], payload[i]);
+                emplaceRef(tmpSupport[i], payload[i]);
             }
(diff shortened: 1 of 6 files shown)


src/core/experimental/array.d
```diff
@@ -0,0 +1,1480 @@
...
+     *
+     * Returns:
+     *      a reference to the element found at `idx`.
+     *
+     * Params:
+     *      idx = a positive integer
+     *
+     * Complexity: $(BIGOH 1).
+     */
+    ref auto opIndex(this _)(size_t idx)
```
jacob-carlborg: What's `this _` for?

src/core/experimental/array.d
```diff
@@ -0,0 +1,1480 @@
...
+    }
+
+    ///
+    bool opEquals(U)(const U rhs) const
+    if (is(U : const typeof(this))
+        || (is(U : const V[], V) && is(typeof(T.init == V.init))))
+    {
+        if (this.length != rhs.length) return false;
+
+        for (size_t i = 0; i < this.length; i++)
```
jacob-carlborg: Is there a reason this is not a `foreach` loop.

jacob-carlborg:
What happens if you slice a `rcarray` and then append to the slice?

src/core/experimental/array.d
lesderid: Nope, leftover from orginal implementation. Fixed

lesderid:
> What happens if you slice a `rcarray` and then append to the slice?

`opOpAssign` uses `opBinary`, which creates a copy of the array, so this behaves like built-in arrays.

src/core/experimental/array.d
lesderid: Attribute inference. I could make it `inout` instead?

ZombineDev: Yes, it `inout` should be used exactly in this case: when the mutability qualifier of the return type needs to be same as the one of `this`.

ZombineDev:
@lesderid thanks for picking up the torch of @edi33416's work!
Since this PR is based on #2348, can you include the last version of #2348 as the first commit in your branch, so that we can track the change that you made on top of it?

lesderid:
> Since this PR is based on #2348, can you include the last version of #2348 as the first commit in your branch, so that we can track the change that you made on top of it?

I'll do that but the diff probably won't be pretty...

src/core/experimental/array.d
jacob-carlborg: Otherwise, to make a dummy template, it’s enough to add an extra set of empty parentheses.

src/core/experimental/array.d
```diff
@@ -0,0 +1,1486 @@
...
+module core.experimental.array;
+
+import core.experimental.refcount;
+
+import core.internal.traits : Unqual;
+
+/**
+Returns the size in bytes of the state that needs to be allocated to hold an
+object of type `T`. `stateSize!T` is zero for `struct`s that are not
+nested and have no nonstatic member variables.
```
n8sh: If that's your intention you need to special-case it below, since `T.sizeof` for a struct with no fields is always at least 1. Check it out:
https://run.dlang.io/is/USB7aW
```d
struct S {}
static assert(S.sizeof == 1);
void main()
{    
}
```

lesderid: Yes, this is currently not implemented. Thanks for reminding me, replacing this with a TODO.

src/core/experimental/array.d
```diff
@@ -23,6 +23,34 @@ private template ElementType(R)
...
 }
 
+
+/**
+ * Create an empty (qualified) `rcarray`.
+ *
+ * Returns:
+ *      an empty `rcarray`
+ */
+auto make(Q : rcarray!T, T)(size_t initialCapacity = 0)
```
lesderid: This factory function is at the module level. The reason for this is that this way one is able to specify the mutability qualifier, without repeating the name of the struct (i.e. `auto ia = make!(immutable rcarray!int);` instead of `auto ia = rcarray!int.make!(immutable rcarray!int);`). This does pollute the module and complicates selective imports.

The alternative is to simply not allow construction of `const`/`immutable` arrays through a factory function. That way, the factory function can be moved into the type (and optionally renamed, giving e.g `auto a = rcarray!int.empty;`). This might be fine, as I can't think of many legitimate use cases of constructing empty `const`/`immutable` arrays anyway.

(Sadly, a factory function is necessary to be able to create empty `rcarray`s without having to explicitly specify the initial capacity. We can't use the default constructor, as that would leave `__RefCount`, which allocates memory in its constructors, uninitialised).

src/core/experimental/array.d
```diff
@@ -0,0 +1,1549 @@
...
+
+    this(return scope const ref typeof(this) rhs) immutable
+    {
+        mixin(copyCtorAlloc);
+    }
+
+    this(return scope immutable ref typeof(this) rhs) immutable
+    {
+        mixin(copyCtorIncRef);
+    }
```
wilzbach: Wouldn't `inout` help here?

src/core/experimental/array.d
```diff
@@ -0,0 +1,1549 @@
...
+
+    private this(RCQual, SuppQual, PaylQual, this Qualified)(RCQual _rc, SuppQual _support, PaylQual _payload)
+        if (is(typeof(support) : typeof(_support)))
```
wilzbach: [DStyle](https://dlang.org/dstyle.html) is to put the contracts on the same indentation level.

src/core/experimental/array.d
```diff
@@ -0,0 +1,1549 @@
...
+    {
+        rc = _rc;
```
wilzbach: FYI: it's much more common to do `this.rc = rc` as then documentation for the function is without the `_` prefix. I know that this is private, but it's still not a very good style to prefix your arguments with an underscore.

src/core/experimental/array.d
```diff
@@ -0,0 +1,1549 @@
...
+        assert(tmpSupport !is null);
+
+        for (size_t i = 0; i < tmpSupport.length; i++)
+        {
+            import core.lifetime : emplace;
+
+            if (i < payload.length)
+            {
+                // Copy the existing items
+                emplace(&tmpSupport[i], payload[i]);
```
wilzbach: Can't we use `emplaceRef` here?

src/core/experimental/array.d
```diff
@@ -0,0 +1,1549 @@
...
+     *
+     * Returns:
+     *      an array that references the current array.
+     *
+     * Complexity: $(BIGOH 1)
+     */
+    auto ref opSlice() inout
+    {
+        return this;
+    }
```
wilzbach: public unittest as an example

src/core/experimental/array.d
lesderid: Partially, see new commit (https://github.com/dlang/druntime/pull/2646/commits/0c762f2e8a82d9b0958f6ef3e51e0c6df91d47d0)

src/core/experimental/array.d
lesderid: We already have https://dlang.org/phobos/std_container_util.html#.make. I don't think it would be particularly confusing. That said, I still don't know if it's the best option.

RazvanN7:

PR #2648, opened 2019-06-19
Add templated _d_arraycatnTX with tests

Vild:
dmd PR: https://github.com/dlang/dmd/pull/10064

This adds a templated entrypoint for _d_arraycatnTX entrypoints.
I don't think there should be any problems with this entrypoint but I do have problem with the dmd implementation.

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -406,5 +406,6 @@ COPY=\
 	$(IMPDIR)\rt\array\equality.d \
 	$(IMPDIR)\rt\array\casting.d \
 	$(IMPDIR)\rt\array\capacity.d \
+	$(IMPDIR)\rt\array\concatenation.d \
 	\
 	$(IMPDIR)\etc\linux\memoryerror.d
--- a/src/object.d
+++ b/src/object.d
@@ -46,6 +46,8 @@ public import rt.array.equality : __equals;
 public import rt.array.equality : __ArrayEq;
 /// See $(REF __ArrayCast, rt,array,casting)
 public import rt.array.casting: __ArrayCast;
+/// See $(REF _d_arraycatnTXImpl, rt,array,concatenation)
+public import rt.array.concatenation : _d_arraycatnTXImpl;
 
 /// See $(REF capacity, rt,array,capacity)
 public import rt.array.capacity: capacity;
--- a/mak/DOCS
+++ b/mak/DOCS
@@ -83,6 +83,7 @@ DOCS=\
     $(DOCDIR)\rt_array_casting.html \
     $(DOCDIR)\rt_array_comparison.html \
     $(DOCDIR)\rt_array_equality.html \
+    $(DOCDIR)\rt_array_concatenation.html \
     \
     $(DOCDIR)\rt_arrayassign.html \
     $(DOCDIR)\rt_arraycat.html \
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -454,6 +454,7 @@ SRCS=\
 	src\rt\array\equality.d \
 	src\rt\array\casting.d \
 	src\rt\array\capacity.d \
+	src\rt\array\concatenation.d \
 	\
 	src\rt\backtrace\dwarf.d \
 	src\rt\backtrace\elf.d \
--- a/mak/WINDOWS
+++ b/mak/WINDOWS
@@ -1195,3 +1195,6 @@ $(IMPDIR)\rt\array\casting.d : src\rt\array\casting.d
 
 $(IMPDIR)\rt\array\capacity.d : src\rt\array\capacity.d
 	copy $** $@
+
+$(IMPDIR)\rt\array\concatenation.d : src\rt\array\concatenation.d
+	copy $** $@
--- a/src/rt/array/concatenation.d
+++ b/src/rt/array/concatenation.d
@@ -1,0 +1,117 @@
+/**
+ This module contains support for controlling dynamic arrays' concatenation
+  Copyright: Copyright Digital Mars 2000 - 2019.
+  License: Distributed under the
+       $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+     (See accompanying file LICENSE)
+  Source: $(DRUNTIMESRC rt/_array/_concatenation.d)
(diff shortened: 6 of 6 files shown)


JinShil:
Since we won't have any code utilizing this until the DMD PR is merged (but the DMD PR needs this PR), please add some unittest that directly instantiate these templates to demonstrate their functionality.

src/rt/array/concat.d
```diff
@@ -0,0 +1,83 @@
+module rt.array.concat;
+
+private extern (C) void[] _d_arraycatnTX(const TypeInfo ti, byte[][] arrs);
+
+T[] _d_arraycatnTX(T)(scope T[][] arrs) @trusted
```
JinShil: Please add DDoc comment including `Params` and `Returns`.  A note about `@trusted` being a temporary hack would also be welcome.

src/rt/array/concat.d
```diff
@@ -0,0 +1,83 @@
+module rt.array.concat;
+
+private extern (C) void[] _d_arraycatnTX(const TypeInfo ti, byte[][] arrs);
```
JinShil: Can this be imported instead of being declared as an `extern`?

JinShil:
For convenience, a link to the existing runtime hook:  https://github.com/dlang/druntime/blob/d50327087bdf7b880cd8b166abb5fcb877ad0736/src/rt/lifetime.d#L2239

JinShil:
You need to add an entry to `mak/WINDOWS` to get the autotester to pass.  See https://github.com/dlang/druntime/pull/2643/files for an example

src/object.d
```diff
@@ -41,6 +41,7 @@ version (D_ObjectiveC) public import core.attribute : selector;
 public import rt.array.comparison : __cmp;
 public import rt.array.equality : __ArrayEq, __equals;
 public import rt.array.casting: __ArrayCast;
+public import rt.array.concat : _d_arraycatnTX, _d_arraycatnTXTrace;
```
jacob-carlborg: I don't like that we have these public implementation specific functions. Can the compiler look directly in `rt.array.concat`? Or can it be a private import and have the compiler bypass the protection?

JinShil: These templates are instantiated by user code, not by the compiler.  The compiler just lowers expressions to them.  So they need to be, source code and all, available to user programs.

jacob-carlborg: Yes, I understand that. But currently all these functions are a weird hybrid of public API and compiler internal API. 

Either these functions should be called by the user and in that case they should be well documented and have good names following the standard naming conventions. Or they should only be called by the compiler. In the latter case we should try to make them only be callable by the compiler and not by users.

JinShil: They can't be *called* by the compiler.  They can *only* be instantiated, and then called by user code.  However, the user should never instantiate and call these implementations explicitly.  The user writes `auto concatenatedArray = array1 ~ array2;`  The compiler lowers that to something like `auto concatenatedArray = _d_arraycatnTX([array1, array2]);` specifically so the user doesn't have to instantiate and call the template explicitly.  Perhaps I'm not understanding.  I intend to add a upcoming PR to do something like this...

jacob-carlborg: For example: `_d_arraycatnTX` can be private and the compiler can lower the code to:

```d
auto concatenatedArray = __traits(getMember, rt.array, "_d_arraycatnTX")([array1, array2]);
```

Because `__traits(getMember)` can bypass the protection. Although I just noticed that it's possible to access a private function using `__traits(getMember)` but it's not possible to call the private function. I'm not sure if that's intentional.

jpf91: Well, the compiler could disable access checks for these specific calls.

However, I'd simply keep things as-is for now and later on change the compiler to directly access the hooks in the original modules, not in object.d: `import rt.array : _d_arraycatnTX = __some_name_which_will_never_conflict`. This change might be somewhat controversial, so it's better to do this later on and not stall further progress.

src/rt/array/concat.d
jpf91: We should probably avoid that, IIRC druntime always uses extern declarations instead of imports for these internal functions. I guess it's to break cyclic module dependencies.

jpf91:
@Vild Can you adjust the windows Makefile as suggested by @JinShil to fix the Windows build? And the `D_TypeInfo` version as mentioned in the other PR?

After these changes, is this ready to go? Code LGTM.

Vild:
There we go, I pushed the fixes.
I'm happy now with the PR.

mak/COPY
```diff
@@ -406,5 +406,6 @@ COPY=\
 	$(IMPDIR)\rt\array\equality.d \
 	$(IMPDIR)\rt\array\casting.d \
 	$(IMPDIR)\rt\array\capacity.d \
+	$(IMPDIR)\rt\array\concat.d \
 	\
```
jpf91: Minor nitpick: I guess to match the other modules, this should maybe be called rt.array.concatenation?

src/rt/array/concat.d
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Concatenating the arrays inside of `arrs`.
+ * `_d_arraycatnTX([a, b, c])` means `a ~ b ~ c`.
+ * Params:
+ *  arrs = Array containing arrays that will be concatenated.
+ * Returns:
+ *  A newly allocated array that contains all the elements from all the arrays in `arrs`.
+ * Bugs:
+ *   The safety level of this function is faked. It shows itself as `@trusted pure nothrow` to not break existing code.
+ */
```
jpf91: Maybe also copy the more detailed explanation regarding faked safety from the other PR here.

mak/COPY
Vild: :+1:

src/rt/array/concat.d
Vild: :+1:

Vild:
This PR needs to implements the same fixes as in https://github.com/dlang/druntime/pull/2667, before it should be merged

thewilsonator:
OK, ping me when this should be merged.

Vild:
@thewilsonator I'm done with all the changes

PR #2649, opened 2019-06-20
Refactor object.d: Move isStaticArray to core.internal.traits

JinShil:
There was an `_isStaticArray` implementation in `object.d` to support `destroy`.  However, [that functionality is already in `std.traits`](https://dlang.org/phobos/std_traits.html#isStaticArray).  So I removed the duplicate implementation in `object.d` and copied it to `core.internal.traits`.  A subsequent PR to Phobos can then be created to forward to the `core.internal.traits` implementation.

I may even follow up with a PR to deprecate [`__traits(isStaticArray)`](https://dlang.org/spec/traits.html#isStaticArray) in favor of the library implementation.

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -164,6 +164,69 @@ template isInnerClass(T) if (is(T == class))
         enum isInnerClass = false;
 }
 
+/// Returns the type of `alias T` if it is not a type, or just returns `T`
+/// if it is a type
+private template Type(alias T)
+{
+    static if (is(typeof(T) == X, X))
+    {
+        alias Type = X;
+    }
+    else
+    {
+        alias Type = T;
+    }
+}
+
+/// Detect whether type `T` is a static array.
+template isStaticArray(alias T)
+{
+    enum isStaticArray = is(Type!T : U[N], U, size_t N) && __traits(getAliasThis, Type!T).length == 0;
+}
+
+///
+@safe unittest
+{
+    static assert( isStaticArray!(int[3]));
+    static assert( isStaticArray!(const(int)[5]));
+    static assert( isStaticArray!(const(int)[][5]));
+
+    static assert(!isStaticArray!(const(int)[]));
+    static assert(!isStaticArray!(immutable(int)[]));
+    static assert(!isStaticArray!(const(int)[4][]));
+    static assert(!isStaticArray!(int[]));
+    static assert(!isStaticArray!(int[char]));
+    static assert(!isStaticArray!(int[1][]));
+    static assert(!isStaticArray!(int[int]));
+    static assert(!isStaticArray!int);
+
+    // This should NOT be considered a static array
+    struct AliasThisStaticArray
+    {
+        int[1] x;
+        alias x this;
+    }
+
+    AliasThisStaticArray atsa;
+    static assert(!isStaticArray!AliasThisStaticArray);
+    static assert(!isStaticArray!atsa);
+
+    // This is an enumeration of static array, so they are indeed static arrays.
+    enum EnumStaticArray : int[2]
+    {
+        a = [1, 2],
+        b = [2, 3]
+    }
+
+    static assert(isStaticArray!EnumStaticArray);
+    static assert(isStaticArray!(EnumStaticArray.a));
+    static assert(isStaticArray!(EnumStaticArray.b));
+
+    int[1] x;
+
+    static assert(isStaticArray!(x));
+}
+
 template dtorIsNothrow(T)
 {
     enum dtorIsNothrow = is(typeof(function{T t=void;}) : void function() nothrow);
@@ -234,7 +297,7 @@ template Fields(T)
 // std.traits.hasElaborateDestructor
 template hasElaborateDestructor(S)
 {
-    static if (__traits(isStaticArray, S) && S.length)
+    static if (isStaticArray!S && S.length)
     {
         enum bool hasElaborateDestructor = hasElaborateDestructor!(typeof(S.init[0]));
     }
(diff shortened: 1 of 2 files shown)


src/core/internal/traits.d
```diff
@@ -164,6 +164,26 @@ template isInnerClass(T) if (is(T == class))
         enum isInnerClass = false;
 }
 
+// Detect whether type `T` is a static array.
+enum bool isStaticArray(T) = __traits(isStaticArray, T);
```
JinShil: Copied from https://github.com/dlang/phobos/blob/991331a951d68de9b36e699ac7572c1d407f7560/std/traits.d#L6646-L6663

ZombineDev:
Replace both declarations with:
```D
enum isStaticArray(T) = is(T : U[N], U, size_t N);
```

(GitHub is acting a bit strange - it didn't include my comment in the review.)

jacob-carlborg:
> I may even follow up with a PR to deprecate __traits(isStaticArray) in favor of the library implementation.

Do we know why `__traits(isStaticArray)` was added?

JinShil:
> Do we know why `__traits(isStaticArray)` was added?

I don't know for sure yet.  My guess at the moment is that it was added before D had all of these fantastic introspection features.  I'll research it before submitting a PR to deprecate it (assuming I even get around to it).

jacob-carlborg:
> Do we know why __traits(isStaticArray) was added?

Seems it was part of the first implementation of `__traits` https://github.com/dlang/dmd/commit/e56081448f and that Brad committed the whole release. Most likely before GitHub.

src/core/internal/traits.d
```diff
@@ -164,6 +164,26 @@ template isInnerClass(T) if (is(T == class))
         enum isInnerClass = false;
 }
 
+/// Detect whether type `T` is a static array.
+enum isStaticArray(T) = is(T : U[N], U, size_t N);
```
ZombineDev: I think the only question that remains is whether we want to allow implicit conversions, i.e. types which have an `alias this` to a static array. Or `enum`s whose base type is a static array:
https://run.dlang.io/gist/run-dlang/e18c94ccff92c2d09feba048b7700860?compiler=dmd

JinShil: > I think the only question that remains is whether we want to allow implicit conversions, i.e. types which have an alias this to a static array

I say *no*.  We want to know if it *is* a static array not whether or not it implicitly converts to a static array or behaves like a static array.

> Or enums whose base type is a static array:
I say *yes*.  That is an enumeration of static arrays.  The enum items are indeed static arrays.

The implementation has been updated and test cases have been added.  Nice work identifying these edge cases!

JinShil:
Another test case that needs adjudication:
```D
int[1] x;

pragma (msg, "isStaticArray!x: ", isStaticArray!x);  // Compiler error
pragma (msg, "__traits(isStaticArray, x): ", __traits(isStaticArray, x)); // true
```
I'd like for the `isStaticArray` template to return `true` for this case, but I'm not sure how to write it to do so.

MoonlightSentinel:
> I'd like for the `isStaticArray` template to return `true` for this case, but I'm not sure how to write it to do so.

You could make `T` an alias parameter and obtain its type by using something like
```
template Type(alias T) {
    static if(is(typeof(T) == X, X)) {
        alias Type = X;
    }
    else {
        alias Type = T;
}
```

But that is probably out of scope for this PR.

JinShil:
> 
> 
> Another test case that needs adjudication:
> 
> ```d
> int[1] x;
> 
> pragma (msg, "isStaticArray!x: ", isStaticArray!x);  // Compiler error
> pragma (msg, "__traits(isStaticArray, x): ", __traits(isStaticArray, x)); // true
> ```
> 
> I'd like for the `isStaticArray` template to return `true` for this case, but I'm not sure how to write it to do so.

Actually, I think the user should probably be required to type `isStaticArray!(typeof(x))`

JinShil:
I used @MoonlightSentinel's technique and added a few more tests to keep the implementation consistent with `__traits(istStaticArray)`.  This PR is ready for additional scrutiny.

src/core/internal/traits.d
```diff
@@ -164,6 +164,75 @@ template isInnerClass(T) if (is(T == class))
...
+        alias Type = X;
+    }
+    else
+    {
+        alias Type = T;
+    }
+}
+
+/// Detect whether type `T` is a static array.
+template isStaticArray(T)
```
MoonlightSentinel: Question:
With dlang/dmd#9769 template alias parameters should accept all types and hence this overload would not be required. But removing it results in the following error message:

```
traits.d(196): Error: template instance `isStaticArray!(int[3])` does not match template declaration isStaticArray(alias T)
traits.d(196):        while evaluating: static assert(isStaticArray!(int[3]))
```

JinShil: I don't know. @FeepingCreature ?

FeepingCreature: Works with run.dlang.org on dmd-nightly... https://run.dlang.io/is/DYSMCo

Are you sure you're using current dmd?

MoonlightSentinel: @FeepingCreature My bad, i was using 2.086.1.

JinShil:
I believe I've addressed all new issues. This PR is ready for additional scrutiny.

jacob-carlborg:
Are we sure it's better to not use `__traits(isStaticArray)`? The library implementation will cause two template instantiations. Is that slower?

JinShil:
> Are we sure it's better to not use __traits(isStaticArray)?

I am of the opinion that anything that can be implemented in the library instead of the language should be.  I'm not sure if that's a very popular opinion, a consensus of D contributors, or a policy of the D Language Foundation.

> Is that slower?

Probably, but I doubt that anyone will notice especially given how infrequently it's used.  Also, in my opinion, any performance issues at compile-time are first and foremost a problem with the compiler implementation, not the language.

PR #2650, opened 2019-06-23
Do not execute commands from any .gdbinit initialization files

Vild:
This unittest can fail if there is a ~/.gdbinit file that changes the output of gdb.
For example this project: https://github.com/cyrus-and/gdb-dashboard


ZombineDev:
(I restarted CircleCI from their UI.)

PR #2651, opened 2019-06-24
Remove the newly-added core.internal.traits.isStaticArray

n8sh:
https://github.com/dlang/druntime/pull/2649  should not have been approved. The private `_isStaticArray!T` in `object` was only used in one place. The correct "refactoring" was to remove it not to promote it to `core.internal.traits` and make it public and start using it everywhere. The `isStaticArray` template is both slower and more complicated than what what it was replacing, with no justification for the replacement other than "it's possible".

Please save the library-building for Phobos.

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -164,69 +164,6 @@ template isInnerClass(T) if (is(T == class))
         enum isInnerClass = false;
 }
 
-/// Returns the type of `alias T` if it is not a type, or just returns `T`
-/// if it is a type
-private template Type(alias T)
-{
-    static if (is(typeof(T) == X, X))
-    {
-        alias Type = X;
-    }
-    else
-    {
-        alias Type = T;
-    }
-}
-
-/// Detect whether type `T` is a static array.
-template isStaticArray(alias T)
-{
-    enum isStaticArray = is(Type!T : U[N], U, size_t N) && __traits(getAliasThis, Type!T).length == 0;
-}
-
-///
-@safe unittest
-{
-    static assert( isStaticArray!(int[3]));
-    static assert( isStaticArray!(const(int)[5]));
-    static assert( isStaticArray!(const(int)[][5]));
-
-    static assert(!isStaticArray!(const(int)[]));
-    static assert(!isStaticArray!(immutable(int)[]));
-    static assert(!isStaticArray!(const(int)[4][]));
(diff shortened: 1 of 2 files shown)


n8sh:
Phobos `std.traitss`:
```d
/**
 * Detect whether type `T` is a static array.
 */
enum bool isStaticArray(T) = __traits(isStaticArray, T);
```

https://github.com/dlang/druntime/pull/2649 `core.internal.traits`:
```d
/// Returns the type of `alias T` if it is not a type, or just returns `T`
/// if it is a type
private template Type(alias T)
{
    static if (is(typeof(T) == X, X))
    {
        alias Type = X;
    }
    else
    {
        alias Type = T;
    }
}

/// Detect whether type `T` is a static array.
template isStaticArray(alias T)
{
    enum isStaticArray = is(Type!T : U[N], U, size_t N) && __traits(getAliasThis, Type!T).length == 0;
}
```

Did no one look at this and say W-T-F?

JinShil:
> Please save the library-building for Phobos.

The separation of Phobos and druntime is an artifact of poor design.  See https://forum.dlang.org/post/q7j4sl$17pe$1@digitalmars.com

JinShil:
> The correct "refactoring" was to remove it not to promote it to core.internal.traits and make it public and start using it everywhere.

`core.internal.traits` is an internal package to druntime.  It should not be used directly by any user; hence the `internal` in the name.  Forwarding Phobos functions to `core.internal.traits` is already an established precedent.  See https://github.com/dlang/phobos/search?q=%22core.internal.traits%22&unscoped_q=%22core.internal.traits%22

JinShil:
> The isStaticArray template is both slower and more complicated than what what it was replacing, with no justification for the replacement other than "it's possible".

No, the justification is not that "it's possible".  The justification is that we should keep the language and compiler as simple as possible and defer and complexity to library implementations.  This was one of the motivating factors for deprecating `delete` and was on full display in the discussions around adding interpolated strings and named parameters.  Some features, like interpolates strings and named parameters, can't be implemented in the library with the same elegance as it can in the compiler, and that is a good argument for adding the feature to the compiler.  The `istStaticArray` implementation, whether implemented in the compiler or the library, is transparent to the user, so the preference is to put it in the library and keep the compiler and language implementation to a set of fundamental composable primitives.

PR #2652, opened 2019-06-25
Remove compiler lowerings from the user documentation

JinShil:
To see the effect of this you need to use a recent compiler that includes https://github.com/dlang/dmd/pull/9068  DAutoTest currently uses 2.079 so you will not see the effect in its output.


JinShil:
This doesn't work the way I thought it did.  It appears I'll have to make a change to https://github.com/dlang/dmd/pull/9068

PR #2647, opened 2019-06-19
Refactor objectd: Move array capacity functions to rt/array/capacity.d

JinShil:
Followup to https://github.com/dlang/druntime/pull/2644, https://github.com/dlang/druntime/pull/2643, and https://github.com/dlang/druntime/pull/2634

PR #2644, opened 2019-06-17
object.d: Move __ArrayEq to rt

JinShil:
Followup to https://github.com/dlang/druntime/pull/2634 and https://github.com/dlang/druntime/pull/2643

PR #2653, opened 2019-06-25
Add documentation for array runtime lowerings for the internal API docs

JinShil:
In my prior PRs (#2647, #2644, #2643, #2634), I did not realize we actually publish documentation on the internal API.  This PR repairs that:

image

cc @Vild.  I'm sorry, but I was wrong about not needing to document these; they need to be documented in the "Internal API" docs.  You'll just need to add an entry to the `mak/DOCS` file for any new files.

I'd like to remove the runtime lowerings from the object.d documentation, but I haven't yet found a solution to that.  I thought versioning it out with `version(CoreDoc)` would work, but it didn't because these templates are necessary for correct language behavior.  I'm investigating other methods.

### DIFF
--- a/mak/DOCS
+++ b/mak/DOCS
@@ -80,6 +80,9 @@ DOCS=\
     $(DOCDIR)\rt_alloca.html \
     \
     $(DOCDIR)\rt_array_capacity.html \
+    $(DOCDIR)\rt_array_casting.html \
+    $(DOCDIR)\rt_array_comparison.html \
+    $(DOCDIR)\rt_array_equality.html \
     \
     $(DOCDIR)\rt_arrayassign.html \
     $(DOCDIR)\rt_arraycat.html \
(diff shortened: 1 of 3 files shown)


wilzbach:
I'm not sure what the problem here is? It's clearly marked as internal API, no?

JinShil:
> I'm not sure what the problem here is? It's clearly marked as internal API, no?

Yes, but it is publicly available at https://dlang.org/phobos/index.html

Currently it looks like this:

image


After this PR it will look like this:

image

So, this PR is trying to maintain the status quo.

JinShil:
Nothing in `rt` should be called directly by the user, but it is, nevertheless, currently documented and published at https://dlang.org/phobos/index.html

wilzbach:
> currently documented and published at 

Yeah, it was an effort I made once to improve the documentation for developers.
Though most symbols don't have Ddoc comments, so while it's always welcome to improve the status quo / keep it, for now it's a failed effort.

JinShil:
>  so while it's always welcome to improve the status quo / keep it, for now it's a failed effort.

We need to know which way to go though.  If you'd like to maintain the status quo, please approve this PR.  Otherwise, let me know what you'd like to do; should I submit an alternate PR to remove the `rt` folder from the docs?

PR #2655, opened 2019-06-27
Translate _d_array{,set}ctor to templates

Vild:
DMD PR: https://github.com/dlang/dmd/pull/10102

- This adds `rt.util.array` to `mak/COPY` as its a dependency for `_d_arrayctor`
- `_d_array{,set}ctor` uses `@trusted` to not break code, `pure` and `nothrow` can be automatically deduced by the compiler
- `_d_arraysetctor` no longer returns a return value as it causes warnings:
```
std/container/array.d(97): Warning: calling rt.array.constructor._d_arraysetctor!(const(Array!int))._d_arraysetctor without side effects discards return value of type const(Array!int)[], prepend a cast(void) if intentional
```

### DIFF
--- a/mak/DOCS
+++ b/mak/DOCS
@@ -82,6 +82,7 @@ DOCS=\
     $(DOCDIR)\rt_array_capacity.html \
     $(DOCDIR)\rt_array_casting.html \
     $(DOCDIR)\rt_array_comparison.html \
+    $(DOCDIR)\rt_array_construction.html \
     $(DOCDIR)\rt_array_equality.html \
     \
     $(DOCDIR)\rt_arrayassign.html \
--- a/src/rt/array/construction.d
+++ b/src/rt/array/construction.d
@@ -1,0 +1,264 @@
+/**
+ This module contains compiler support for constructing dynamic arrays
+
+  Copyright: Copyright Digital Mars 2000 - 2019.
+  License: Distributed under the
+       $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+     (See accompanying file LICENSE)
+  Source: $(DRUNTIMESRC rt/_array/_construction.d)
+*/
+module rt.array.construction;
+
+/**
+ * Does array initialization (not assignment) from another array of the same element type.
+ * Params:
+ *  to = what array to initialize
+ *  from = what data the array should be initialized with
+ * Returns:
+ *  The constructed `to`
+ * Bugs:
+ *  This function template was ported from a much older runtime hook that bypassed safety,
+ *  purity, and throwabilty checks. To prevent breaking existing code, this function template
+ *  is temporarily declared `@trusted` until the implementation can be brought up to modern D expectations.
+ */
+Tarr _d_arrayctor(Tarr : T[], T)(return scope Tarr to, scope Tarr from) @trusted
+{
+    pragma(inline, false);
+    import core.internal.traits : Unqual;
+    import core.stdc.string : memcpy;
+    debug(PRINTF) import core.stdc.stdio;
+
+    // Force `enforceRawArraysConformable` to be `pure`
+    void enforceRawArraysConformable(const char[] action, in size_t elementSize, const void[] a1, const void[] a2, in bool allowOverlap = false) @trusted
+    {
+        import rt.util.array : enforceRawArraysConformable;
+
+        alias Type = void function(const char[] action, in size_t elementSize, const void[] a1, const void[] a2, in bool allowOverlap = false) pure nothrow;
+        (cast(Type)&enforceRawArraysConformable)(action, elementSize, a1, a2, allowOverlap);
+    }
+
+    debug(PRINTF) printf("_d_arrayctor(to = %p,%d, from = %p,%d) size = %d\n", from.ptr, from.length, to.ptr, to.length, T.tsize);
+
+    auto element_size = T.sizeof;
+
+    void[] vFrom = (cast(void*)from.ptr)[0..from.length];
+    void[] vTo = (cast(void*)to.ptr)[0..to.length];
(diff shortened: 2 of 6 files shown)


src/rt/array/constructor.d
```diff
@@ -0,0 +1,165 @@
+/**
+ This module contains compiler support for constructing dynamic arrays
+
+  Copyright: Copyright Digital Mars 2000 - 2019.
+  License: Distributed under the
+       $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+     (See accompanying file LICENSE)
+  Source: $(DRUNTIMESRC rt/_array/_constructor.d)
```
JinShil: To maintain the "tense" (not sure if that's the right word) of the other modules in this folder, I recommend naming it "construction".

Vild: :+1:

mak/DOCS
```diff
@@ -82,6 +82,7 @@ DOCS=\
     $(DOCDIR)\rt_array_capacity.html \
     $(DOCDIR)\rt_array_casting.html \
     $(DOCDIR)\rt_array_comparison.html \
+    $(DOCDIR)\rt_array_constructor.html \
```
jpf91: rt_array_construction?

Vild: Thanks, missed that one.

src/rt/array/construction.d
```diff
@@ -0,0 +1,165 @@
...
+            memcpy(elem, &from[i], element_size);
+
+            static if (__traits(hasMember, T, "__xpostblit") &&
+                // Bugzilla 14746: Check that it's the exact member of S.
+                __traits(isSame, T, __traits(parent, elem.__xpostblit)))
+                elem.__xpostblit();
+        }
+    }
+    catch (Exception o)
+    {
```
jpf91: Just for reference, this means calling destructors will not be done if the postblit threw a non-exception throwable. IIRC the language spec allows this behaviour.

jpf91:
LGMT, but the windows Makefile needs to be fixed. @Vild is this PR ready?

src/rt/array/construction.d
JinShil: See https://github.com/dlang/druntime/blob/ab890e0fb09cb6142806bf3d2cfd0a363936a298/src/rt/arrayassign.d#L180-L199 for the genesis of this code.

src/rt/array/construction.d
```diff
@@ -0,0 +1,165 @@
...
+        // Destroy, in reverse order, what we've constructed so far
+        while (walker > 0)
+        {
+            walker--;
+            auto elem = cast(Unqual!T*)&p[walker];
+            static if (__traits(hasMember, T, "__xdtor") &&
```
JinShil: Should this code be calling `object.destroy` instead?

src/rt/array/construction.d
```diff
@@ -0,0 +1,165 @@
...
+    import core.internal.traits : Unqual;
+    import core.stdc.string : memcpy;
+
+    // Force `enforceRawArraysConformable` to be `pure`
+    void enforceRawArraysConformable(const char[] action, in size_t elementSize, const void[] a1, const void[] a2, in bool allowOverlap = false)
+    {
+        import rt.util.array : enforceRawArraysConformable;
+        mixin("alias Type = " ~ typeof(&enforceRawArraysConformable).stringof ~ " pure;");
+        (cast(Type)&enforceRawArraysConformable)(action, elementSize, a1, a2, allowOverlap);
+    }
```
JinShil: I understand you need to make `enforceRawArraysConformable` but why this technique?  Can you just type it out instead of using a string mixin?

src/rt/array/construction.d
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Does array initialization (not assignment) from another array of the same element type.
+ * Params:
+ *  to = what array to initialize
+ *  from = what data the array should be initialized with
+ * Returns:
+ *  The constructed `to`
+ * Bugs:
+ *  The functions `@safe` safety is faked.
```
JinShil: I think this sentence is too vague.  I suggest "This function template was ported from a much older runtime hook that bypassed safety, purity, and throwabilty checks.  To prevent breaking existing code, this function template is temporarily declared `@trusted` until the implementation can be brought up to modern D expectations."

src/rt/array/construction.d
Vild: I changed it `object.destroy` and `object._postblitRecurse` for the postblit.

src/rt/array/construction.d
Vild: Thanks :)

Vild:
I would call this PR ready.

About the windows build problem, I'm not sure why this is happening.
```
copy src\rt\util\array.d  import\rt\util\array.d
        0 file(s) copied.
```
But the file exist when I check: `ls src/rt/util/array.d`

JinShil:
@rainers We're getting the following build error on Windows

```
..\..\druntime\import\rt\array\construction.d(32): Error: module `array` is in file 'rt\util\array.d' which cannot be read
```

It appears all the right makefile modifications have been made.  Do you happen to know why this error is occurring?

rainers:
> About the windows build problem, I'm not sure why this is happening.
> 
> ```
> copy src\rt\util\array.d  import\rt\util\array.d
> The system cannot find the path specified.
> ```

The problem is that the output directory does not exist. It should be added similar to https://github.com/dlang/druntime/blob/master/mak/WINDOWS#L57

While you are at it, you could also add `core\sys\netbsd\sys` and `core\sys\netbsd`.

thewilsonator:
It seems like the exception paths are not tested, could you add some tests that exercise those? Otherwise looks good.

Vild:
@thewilsonator done

This PR is ready to be merged.

PR #2656, opened 2019-06-28
Translate _d_arraysetlengthT to template

Vild:
dmd PR: https://github.com/dlang/dmd/pull/10106

`rt.array.capacity._d_arraysetlengthT` will call the `_d_arraysetlength{,iT}` depending on if `T` is zero initialized, this logic used to be inside of dmd.

### DIFF
--- a/src/rt/array/capacity.d
+++ b/src/rt/array/capacity.d
@@ -1,5 +1,5 @@
 /**
- This module contains support for controlling dynamic arrays' capacity
+ This module contains support for controlling dynamic arrays' capacity and length
 
   Copyright: Copyright Digital Mars 2000 - 2019.
   License: Distributed under the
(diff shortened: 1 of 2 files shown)


src/rt/array/capacity.d
```diff
@@ -195,3 +195,98 @@ auto ref inout(T[]) assumeSafeAppend(T)(auto ref inout(T[]) arr) nothrow @system
...
+ *  a = the array that will be resized, taken as a reference
+ *  newlength = new length of array
+ * Returns:
+ *  The new value of `arr`
+ * Bugs:
+ *   The safety level of this function is faked. It shows itself as `@trusted pure nothrow` to not break existing code.
+ */
+Tarr _d_arraysetlengthT(Tarr : T[], T)(return scope ref Tarr arr, size_t newlength) @trusted
+{
+    version (D_BetterC)
```
JinShil: I'm trying to figure out why this check is necessary, but I haven't succeeded yet.  It shouldn't be required, but is a successful workaround.  Until then, please use `D_TypeInfo` instead of `D_BetterC`.  I think GDC does not support betterC, but instead has individual flags like `-fno-rtti`, `-fno-exceptions`, etc.  Using the more specific version specifier should make this code more portable.  The error message would also need to be updated e.g. "Cannot resize arrays if compiling without support for runtime type information."

Vild: Thanks, I will do the same changes on the other PRs.

src/rt/array/capacity.d
```diff
@@ -195,3 +195,98 @@ auto ref inout(T[]) assumeSafeAppend(T)(auto ref inout(T[]) arr) nothrow @system
...
+ *  a = the array that will be resized, taken as a reference
+ *  newlength = new length of array
+ * Returns:
+ *  The new value of `arr`
+ * Bugs:
+ *   The safety level of this function is faked. It shows itself as `@trusted pure nothrow` to not break existing code.
+ */
+Tarr _d_arraysetlengthT(Tarr : T[], T)(return scope ref Tarr arr, size_t newlength) @trusted
+{
+    version (D_TypeInfo)
```
jpf91: @Vild seems like the if/else branchs here should be swapped.

jpf91:
LGTM

src/rt/array/capacity.d
Vild: Fixed

Vild:
```
posix.mak:395: recipe for target 'unittest/std/datetime/systime.run' failed
gmake: *** [unittest/std/datetime/systime.run] Error 143
```
I don't think this auto-tester error have anything to do with my changes.

thewilsonator:
This good to go?

Vild:
> This good to go?

Yes! :)

PR #2661, opened 2019-07-04
Fix Issue 20026 - retrying while pthread_cond_signal/pthread_cond_broadcast…

kubo39:
… return EAGAIN


kubo39:
It would be possible, but will take a long time to execute.

thewilsonator:
Fair enough.

PR #2662, opened 2019-07-04
memutils: Replacement of libc string.h functions - currently only Dmemset()

baziotis:
As this is my first PR to D runtime, here are some notes:

1) You can refer to this repo in which the work progresses: https://github.com/baziotis/Dmemutils
2) Generally, I'm seeing tests only in `unittest` but these tests can't really be put in `unittest`, unless I miss something.
3) Progressively, more functions will be put in this file, that have common code. Such code is for example the SIMD intrinsics used in `Dmemset`. I think that a better choice would be to have such code in something like `experimental/simd.d` but I was not sure if that's desired.


src/core/experimental/memutils.d
```diff
@@ -0,0 +1,352 @@
...
+            // but the fact that it's more difficult to optimize it as part of the rest of the code.
+            if (n <= 16)
+            {
+                Dmemset_naive(cast(ubyte*)d, cast(ubyte)val, n);
+                return;
+            }
+            void *temp = d + n - 0x10;                  // Used for the last 32 bytes
+            int4 xmm0;
+            // Broadcast v to all bytes.
+            broadcast_int(xmm0, v);
```
rainers: `auto xmm0 = int4(v);`  works for me.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,352 @@
...
+            ubyte rem = cast(ulong)d & 15;              // Remainder from the previous 16-byte boundary.
```
rainers: No need to extend to ulong for 32-bit platforms, cast to size_t instead.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,352 @@
...
+    version (D_SIMD)
+    {
+        static if (isArray!T)
+        {
+            size_t n = dst.length * typeof(dst[0]).sizeof;
+            Dmemset(dst.ptr, v, n);
+
+            version (unittest)
+            {
+                Dmemset_naive(dst.ptr, v, n);
```
rainers: Why are these calls here?

src/core/experimental/memutils.d
baziotis: Thanks, I didn't know it.

src/core/experimental/memutils.d
baziotis: Yes, actually, that should be `cast(ubyte)` anyway.

src/core/experimental/memutils.d
baziotis: I thought when testing the SIMD version, to test both versions. It was convenient locally.
But, probably not a good idea for the D runtime, since the CI will run in all platforms anyway.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,332 @@
+/**
+ * Pure D replacement of the C Standard Library basic memory building blocks of string.h
+ * Source: $(DRUNTIMESRC core/experimental/memutils.d)
+ */
+module core.experimental.memutils;
+
+unittest
```
wilzbach: At the top level a user `unittest` should be provided which shows off the core features.
Example: https://dlang.org/phobos/std_container_binaryheap.html

tl;dr: All of these tests should be moved below the declaration of the function.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,332 @@
...
+        import ldc.llvmasm;
+        __asm("", "r,~{memory}", p);
+    }
+    version (GNU)
+    {
+        asm { "" : : "g" p : "memory"; }
+    }
+}
+
+void Dmemset_verifyArray(T)(int j, const ref T[] a, const ubyte v)
```
wilzbach: Mind the [DStyle](https://dlang.org/dstyle.html) when naming things.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,332 @@
...
+    Dmemset_testDynamicArray!(ubyte)(5, 32344);
+    Dmemset_testStaticArray!(ubyte, 32344)(5);
+    Dmemset_testDynamicArray!(ubyte)(5, 46830);
+    Dmemset_testStaticArray!(ubyte, 46830)(5);
+    Dmemset_testDynamicArray!(ubyte)(5, 64349);
+    Dmemset_testStaticArray!(ubyte, 64349)(5);
+}
+
+// From a very good Chandler Carruth video on benchmarking: https://www.youtube.com/watch?v=nXaxk27zwlk
+void escape(void* p)
```
wilzbach: We only need this for benchmarking, so this doesn't need to be part of this module. We have the https://github.com/dlang/druntime/tree/master/benchmark folder here (though these test run on linux only), but if you have any benchmark scripts they should be added to this folder. They don't have to be integrated with the suite, but they shouldn't be part of the released source code which they would if added here.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,332 @@
...
+        __asm("", "r,~{memory}", p);
+    }
+    version (GNU)
+    {
+        asm { "" : : "g" p : "memory"; }
+    }
+}
+
+void Dmemset_verifyArray(T)(int j, const ref T[] a, const ubyte v)
+{
```
wilzbach: Also, if as this is for testing purposes only it should be `private` and probably even have the `CoreUnittest` version around it to make sure it doesn't end up in the released sources. If you don't like the version tags, you could create a folder in `tests` and move all the test functions over there.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,332 @@
...
+    escape(&t);
+    Dmemset(t, v);
+    Dmemset_verifyStaticType(t, v);
+}
+
+version (GNU)
+{
+    void Dmemset(void *d, const uint val, size_t n)
+    {
+        Dmemset_naive(d, cast(const(ubyte))val, n);
```
wilzbach: DStyle: 

```suggestion
        Dmemset_naive(d, cast(const(ubyte)) val, n);
```

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,332 @@
...
+{
+    version (D_SIMD)
+    {
+        // NOTE(stefanos): I could not GDC respective intrinsics.
+        void Dmemset(void *d, const uint val, size_t n)
+        {
+            import core.simd: int4;
+            version (LDC)
+            {
+                import ldc.simd: loadUnaligned, storeUnaligned;
```
wilzbach: DStyle:
```suggestion
                import ldc.simd : loadUnaligned, storeUnaligned;
```

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,332 @@
...
+            }
+            else
+            version (DigitalMars)
```
wilzbach: It's okay to have `else version` in one line.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,332 @@
...
+version (GNU)
+{
+    void Dmemset(void *d, const uint val, size_t n)
+    {
+        Dmemset_naive(d, cast(const(ubyte))val, n);
+    }
+}
+else
+{
+    version (D_SIMD)
```
wilzbach: We should also provide a version if SIMD isn't supported

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,332 @@
...
+    static if (is(T U ==          immutable U)) alias ModifyTypePreservingTQ =          immutable Modifier!U;
+    else static if (is(T U == shared inout const U)) alias ModifyTypePreservingTQ = shared inout const Modifier!U;
+    else static if (is(T U == shared inout       U)) alias ModifyTypePreservingTQ = shared inout       Modifier!U;
+    else static if (is(T U == shared       const U)) alias ModifyTypePreservingTQ = shared       const Modifier!U;
+    else static if (is(T U == shared             U)) alias ModifyTypePreservingTQ = shared             Modifier!U;
+    else static if (is(T U ==        inout const U)) alias ModifyTypePreservingTQ =        inout const Modifier!U;
+    else static if (is(T U ==        inout       U)) alias ModifyTypePreservingTQ =              inout Modifier!U;
+    else static if (is(T U ==              const U)) alias ModifyTypePreservingTQ =              const Modifier!U;
+    else                                             alias ModifyTypePreservingTQ =                    Modifier!T;
+}
```
wilzbach: If this is copied from std.traits, please move them to `core.internal.traits`

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,332 @@
...
+    }
+    else
+        static assert(0, T.stringof~" is not a dynamic array");
+}
+
+enum bool isDynamicArray(T) = is(DynamicArrayTypeOf!T) && !isAggregateType!T;
+enum bool isStaticArray(T) = __traits(isStaticArray, T);
+enum bool isArray(T) = isStaticArray!T || isDynamicArray!T;
+
+void Dmemset(T)(ref T dst, const ubyte val)
```
wilzbach: Needs proper documentation. Also why do we call we need the `D` prefix?
Also, this should be  before all of its tests and specialization, s.t. a reader has a consistent flow.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,332 @@
...
+    for (size_t i = 0; i < a.length * T.sizeof; i++)
+    {
+        assert(p[i] == v);
+    }
+}
+
+void Dmemset_verifyStaticType(T)(const ref T t, const ubyte v)
+{
+    const ubyte *p = cast(const ubyte *) &t;
+    for (size_t i = 0; i < T.sizeof; i++)
```
wilzbach:
```suggestion
    foreach (i; 0 .. T.sizeof)
```

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,332 @@
...
+}
+
+void Dmemset_testStaticArray(T, size_t n)(const ubyte v)
+{
+    T[n + 32] buf;
+
+    enum alignments = 32;
+    size_t len = n;
+
+    foreach (i; 0..alignments)
```
wilzbach: DStyle: 

```suggestion
    foreach (i; 0 .. alignments)
```

src/core/experimental/memutils.d
baziotis: I guess it was not clear, but I don't know a better way to do it.
If SIMD isn't supported, only `Dmemset_naive` will be compiled _and_ called.
Check the `void Dmemset(T)(ref T dst, const ubyte val)` definition (currently at the bottom of the file).

src/core/experimental/memutils.d
baziotis: Yes, I'll move them along with the other layout changes.
The `D` prefix is from when the project started. There's no specific reason for it, I'm just afraid that it will collide with the libc versions.

src/core/experimental/memutils.d
baziotis: I think that `std.traits` includes `core.internal.traits`. If I move them there, wouldn't that create problem?

wilzbach: No. On the contrary. You can remove the implementation from `std.traits` after moving them there to avoid a duplicate implementation.
Example: https://github.com/dlang/phobos/pull/6847

src/core/experimental/memutils.d
wilzbach: > I'm just afraid that it will collide with the libc versions.

I don't think it will as the C `memset` accepts `void*` whereas this here accepts actual D types. Also, people generally don't import things from `core.stdc`, so in most cases both symbols won't be in the same scope.

src/core/experimental/memutils.d
wilzbach: OK, but if that's the SIMD version I would do keep SIMD versioning in the `memset` function. No reason to duplicate it.

baziotis: In which of 2? I see 2 possibilities:
a) Remove it altogether from there and just have it in `Dmemset(T)`
b) Put it inside the function instead of around it.

With a), the problem is that `Dmemset(void *...` will be compiled. And if SIMD is not available, it won't compile.
With b), without SIMD an empty function will be compiled. If that's good, ok.

src/core/experimental/memutils.d
baziotis: Done, but not sure if done correctly.

src/core/experimental/memutils.d
baziotis: Yes, but to do that, wouldn't those 2 have to be done at the same time?
Actually, first merge the changes in `core.internal.traits`, then a PR to phobos.
Otherwise, I guess phobos will break.
If that's the case, I guess this means that I have to do a different PR for the move in `core.internal.traits`.

wilzbach: > Actually, first merge the changes in core.internal.traits, then a PR to phobos.

Yes, but things won't break if you add them in `core.internal.traits`. 

> If that's the case, I guess this means that I have to do a different PR for the move in core.internal.traits.

It can be done directly here as this way we know that these traits are actually used. Though that's your choice.

src/core/experimental/memutils.d
baziotis: Good. I made the change but left the `D` prefix in a couple of places for either conceptual reasons or practical reasons (for example, there is a `Dmemset` that gets `void *`).
I'm open to further suggestions though.

src/core/experimental/memutils.d
baziotis: Ok, good.
> It can be done directly here as this way we know that these traits are actually used. Though that's your choice.
I agree, I proposed a different PR just because this PR might never be accepted, but ok, I'll do it here.

src/core/experimental/memutils.d
baziotis: I have not forgotten this one, I just picked other things first. The thing with escaping is that the reason we want it for benchmarking purposes is so that the compiler doesn't optimize it away.
But optimizing it away is a problem when testing as well. I don't remember the last time I checked _without_ escaping, when I have time I'll do and I'll post an update.

src/core/experimental/memutils.d
wilzbach: One variant:

```d
version(D_SIMD) memsetSIMD() {

...
}

auto memset(...) {

   version (D_SIMD)  {
     version (GNU) { memsetNaive(); }  else {
         memsetSIMD();
      }
   }
}
```

Another one:

```d
version (D_SIMD)  {
    version (GNU) {
       enum memsetWithSIMD = false;
    } else {
       memsetWithSIMD = true;
   }
} else {
   enum memsetWithSIMD = false;
} 

version(memsetWithSIMD) memsetSIMD() {
...
}

auto memset(...) {
     version (memsetWithSIMD) { 
       memsetSIMD();
      }  else {
        memsetNaive();
      }
   }
}
```

baziotis: Thanks, I did something like what you proposed. There is a problem though with GDC. Eventually, GDC should probably use intrinsics as well. That's not necessary, as @jpf91 said seems to be more interested in completely pure D implementation (Johannes you can give feedback here).

But if this happens, GDC's intrinsics are x86 only. Meaning, then the code will look like:
```
version (GNU)
{
    useSIMD = false;
    version (x86_64)
    {
         version (D_SIMD) { useSIMD = true; }
    }
    version (useSIMD)
    {
         // A same implementation as the current SIMD one, but with GDC intrinsics.
    }
    else
    {
        // What is done now - i.e. forward to naive.
    }
}
```
Which pretty much means that the code will get ugly.

baziotis: I made an update in the forums. GDC (and LDC if the SIMD version is not used) reaches complete parity with libc memset() in my benchmarks (AMD Ryzen). If that holds true for different systems (unfortunately, I don't have Intel to test on) then we could even use the SIMD version only for DMD.

src/core/experimental/memutils.d
baziotis: Ok, escaping was removed. It's not needed and my best _guess_ is that this is because of the verification (if the code is optimized away, an assert will fire, which does not satisfy correctness).
Do you want to add benchmarks in `/benchmark` ?
I think that it's easier for people to do benchmarks using the main repo of the project. So maybe somehow redirect them there.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,179 @@
...
+        size_t n = dst.length * typeof(dst[0]).sizeof;
+        Dmemset(dst.ptr, v, n);
+    }
+    else
+    {
+        Dmemset(&dst, v, T.sizeof);
+    }
+}
+
+version (GNU)
```
jpf91: AFAIK GDC does not set the `D_SIMD` version, so you probably don't need the explicit GNU branch here.

jpf91:
BTW, with these kind of things I highly recommend checking the ASM in GDC, as the compiler might silently turn your code back into a call to the C memset function ;-)

Seems to work fine in this case though: https://explore.dgnu.org/z/-uJfsJ

I'll need to have a look at the GCC sources to see if we can somehow disable this pattern matching / rewriting for individual functions, just to be safe.

src/core/experimental/memutils.d
ibuclaw: Indeed it doesn't.  `D_SIMD` is to be defined when you have `__simd` as intrinsics.

baziotis: Your comment revealed an important let's say "bug". LDC doesn't set `D_SIMD` as well, which means that LDC uses the naive version as well. Although that reaches parity with libc, using the actual SIMD version surpasses libc quite a lot, especially for small sizes. And I don't know a way to check if SIMD is available in LDC. Is there such a thing in GDC?

baziotis:
Add an -O3 there and you'll see it jumps to memset. :P 
We have to see that yes.

ibuclaw:
Just a general high-level question, how is this meant to scale to the plethora of already supported targets? - ARM, MIPS, RISCV, PPC, SystemZ, HPPA, SPARC, and their 64-bit variants to name few.

Have considerations been made for strict alignment targets?  BigEndian targets?

My assumption would be that the performance would be a net-loss for everyone, leaving some improvements left wanting for the naive implementation.  Which could be done by handling small arrays with minimal branching, then setting values 32 or 64 bytes at a time depending on how big the array is.

jpf91:
@ibuclaw BTW, `-fno-builtin` seems to be broken:
https://explore.dgnu.org/z/I2RZju
https://godbolt.org/z/ZYp1AY

src/core/experimental/memutils.d
ibuclaw: GDC pretty much supports any kind of vector size. `__traits(compiles, the vector you want to use)` is the best way to detect if it is rejected for unsupported reasons.

baziotis:
@ibuclaw The project was initially focused on DMD and x86_64. For all the rest of the targets the goal was to just have a naive working implementation. That changed the last week, which meant throwing most of the (ASM) code away and I started again with intrinsics, optimizing for LDC etc. But now time is constrained. Seb and I try to make the best of the time left, but the community can influence us. So, I'm open to suggestions.

I didn't really understand what you meant about the naive implementation. Settings 32 or 64 bytes at a time requires SIMD, which is done in the SIMD implementation, along with other optimizations (reaching alignment and there was a switch fall-through for small sizes, but that had problems, so it forwards to naive. There are some other tweaks but those require ASM-level control).

The without SIMD optimizations are not a lot. They're things like the GNU algorithm, which pretty much reaches 4/8-byte alignment. But those things are trivial for a compiler to do it itself.

ibuclaw:
> Settings 32 or 64 bytes at a time requires SIMD, which is done in the SIMD implementation, along with other optimizations (reaching alignment and there was a switch fall-through for small sizes, but that had problems, so it forwards to naive. There are some other tweaks but those require ASM-level control).

```
// val = 0b00100000
ubyte val = 32;

// val32 = 0b00100000_00100000_00100000_00100000
uint val32 = -1u / 255 * c;

// val64 = 0b00100000_00100000_00100000_00100000_00100000_00100000_00100000_00100000
ulong val64 = val32 | (cast(ulong)val32 << 32);
```

Setting then becomes (make sure pointer it aligned!)
```
*cast(uint*)(dest + i) = val32;
```
or
```
*cast(ulong*)(dest + i) = val64;
```

ibuclaw:
> The without SIMD optimizations are not a lot. They're things like the GNU algorithm, which pretty much reaches 4/8-byte alignment. But those things are trivial for a compiler to do it itself.

I'm sure the compiler could go a long way by itself, but sometimes you need to give it a gentle nudge to make it go one better.

Anyway, I'll wait to see how this ends up, as things are still in flux.

baziotis:
So, you mean 32/64 _bits_ at a time, not bytes. Then yes, as I said, we're going back to the modified GNU algorithm: https://www.embedded.com/print/4024961 (that's for memcpy() but the "GNU algorithm" is thrown around when talking about the idea of "align to 4/8-byte boundary and do appropriate moves").
It's not a big thing to add, but you know better than be that the compiler will do way more crazy stuff than this.
For example, in the link that @jpf91 posted above: https://godbolt.org/z/ZYp1AY
The compiler reaches 16-byte alignment, uses SIMD and does a switch fall-through in the end. :P

wilzbach:
For reference, the goal of this project is to provide common generic building blocks like this one in pure D, so that they can be used for no matter what target architecture - from wasm to baremetal. Mike Franklin (and now this project) have shown that it's possible to beat libc if you know more than `void*` at compile-time, and while this is a very interesting aspect of this project, I think the full platform compatibility is the biggest goal. AFAICT the idea is roughly to see how well we fare against libc and maybe optimize for x86_64 to be libc-free for this architecture, but fallback to libc on other platforms if available (or the generic naive D if there's no libc).

baziotis:
Yes, a big part initially was compile time information, which can be seen in `Dmemcpy` / `Dmemset`. As a comment, my guess is that it doesn't matter for LDC / GDC (i.e. a simple *dst = *src is enough to let the compiler do better work that all the Dmemcpy).
At some point it was mentioned that for other targets we would have libc or use a simple implementation or incrementally let other people, who are experts in those targets, create optimized versions. For x86 I think that Dmemset, Dmemcpy and Dmemmove are good and could replace libc. That's true for dynamic sizes as well. For a multi-target implementation, a lot more work is needed that cannot be done in a summer and possibly by a single person.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,277 @@
...
+    }
+}
+
+// NOTE(stefanos): We're using naive for the < 32 case in the SIMD version.
+// To be more performant, for that case, we would have a big fall-through switch
+// for all < 32 sizes.
+private void memsetNaive(void *dst, const uint val, size_t n)
+{
+    const ulong v = cast(ulong) val * 0x0101010101010101;  // Broadcast val to all 8 bytes
+    enum handleLT16Sizes = "
```
thewilsonator: You should see if using a `pragma(inline, true)` function does what you want it to there rather than use an enum + mixin. If you have to the enum + mixin please use [q{} strings](https://dlang.org/spec/lex.html#delimited_strings) instead of quoted strings.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,277 @@
...
+    // stuff below. So we start by default with useSIMD and decide later.
+    version = useSIMD;
+}
+
+version (useSIMD)
+{
+    /* SIMD implementation
+     */
+    //pragma(msg, "SIMD used");
+    extern(C) private void Dmemset(void *d, const uint val, size_t n)
```
thewilsonator: Is there any reason for this to be `extern(C)`?

baziotis: No.

src/core/experimental/memutils.d
baziotis: I don't think dmd will be able to inline but yes, I agree.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,285 @@
+/**
+ * Pure D replacement of the C Standard Library basic memory building blocks of string.h
+ * Source: $(DRUNTIMESRC core/experimental/memutils.d)
+ */
+module core.experimental.memutils;
+
+/** memset() implementation */
```
thewilsonator: Remove this comment

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,285 @@
...
+
+/**
+ * NOTE(stefanos):
```
thewilsonator: Make this an N.B. of the doc comment below or remove

baziotis:
The TODOs were removed and some NOTEs. The NOTEs that have remained I think are important. Are they ok as they are?

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,251 @@
...
+ */
+module core.experimental.memutils;
+
+/*
+  If T is an array, set all `dst`'s bytes
+  (whose count is the length of the array times
+  the size of the array element) to `val`.
+  Otherwise, set T.sizeof bytes to `val` starting from the address of `dst`.
+*/
+void memset(T)(ref T dst, const ubyte val)
```
lesderid: Is this undocumented (ddoc) on purpose?

baziotis: You mean it should be:
`
/**
 * ...
 */`
?

lesderid: Yes, and a `Params` section (https://dlang.org/spec/ddoc.html#standard_sections) is probably best too.

baziotis: Hmm, thanks. I'm not very accustomed yet to the logistics of contributing to D.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,256 @@
...
+ * (whose count is the length of the array times
+ * the size of the array element) to `val`.
+ * Otherwise, set T.sizeof bytes to `val` starting from the address of `dst`.
+ * 
+ * Params
+ *  val = The byte with which we want to fill memory with.
+ *  dst = Memory Destination whose bytes are to be set to `val`.
+ *
+ * Returns:
+ *  Nothing.
```
lesderid: You can just leave the `Returns` section out if it's `void`. :wink:

baziotis: Haha, I didn't like it too. But I wanted a way to make a note that this `memset` is different from the libc in that it returns nothing. I'll add it as a `N.B.`.

baziotis:
I think something went wrong with buildkite that is not a problem of the PR.

wilzbach:
see e.g. https://github.com/dlang/dmd/pull/10244#pullrequestreview-268622608

baziotis:
Sorry but I don't understand what I should do.

thewilsonator:
Nothing, it is unrelated to this PR.

baziotis:
Ah, ok, that's what I figured, just pointing to the problem. Thanks

JohanEngelen:
One thing to take into account is that the LLVM optimizer recognizes `memset` and `memcpy` and knows the semantics of those functions. It uses that knowledge for optimization (e.g. if you memcpy something, the optimizer knows that the dst memory now contains the values of src), which can be a massive performance boost. 
When you provide your own memset function, and the optimizer doesn't know the semantics of it, you may loose performance because of it. Cross-module inlining is tricky to work here:
- your new memset needs to be compiled without the compiler turning the algorithm into a `memset`/`memcpy` call
- application code needs to be compiled _with_ the compiler turning any copying algorithm into `memcpy` or `memset`.
--> this cannot be done in one compiler optimization unit..............

Before using/enabling this on the optimizing compilers, please do _extensive_ testing and assembly output checking.


The current set of tests is way too small. After just a few seconds of thinking about it I already see things untested: misaligned pointer, `null`, different memset sizes to trigger all corner cases of simd implementations (4-ply, 8-ply, ...), ...

src/core/experimental/memutils.d
JohanEngelen: Probably you can add `scope` to the reference? Explicit `nogc` `nothrow` may be good too.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,254 @@
...
+ * Source: $(DRUNTIMESRC core/experimental/memutils.d)
+ */
+module core.experimental.memutils;
+
+/**
+ * If T is an array, set all `dst`'s bytes
+ * (whose count is the length of the array times
+ * the size of the array element) to `val`.
+ * Otherwise, set T.sizeof bytes to `val` starting from the address of `dst`.
+ * N.B.: Contrary to the C Standard Library memset(), this functions returns nothing.
```
JohanEngelen: why deviate from c stdlib memset on this point?

baziotis:
Oh crap, I did not mean to close the PR.

baziotis:
> One thing to take into account is that the LLVM optimizer recognizes `memset` and `memcpy` and knows the semantics of those functions. It uses that knowledge for optimization (e.g. if you memcpy something, the optimizer knows that the dst memory now contains the values of src), which can be a massive performance boost.
> When you provide your own memset function, and the optimizer doesn't know the semantics of it, you may loose performance because of it. Cross-module inlining is tricky to work here:
> 
> * your new memset needs to be compiled without the compiler turning the algorithm into a `memset`/`memcpy` call
> * application code needs to be compiled _with_ the compiler turning any copying algorithm into `memcpy` or `memset`.
>   --> this cannot be done in one compiler optimization unit..............
> 
> Before using/enabling this on the optimizing compilers, please do _extensive_ testing and assembly output checking.
> 

That is true. It's supposed to be done on GDC as well. It's supposed that on GDC, you can set the `-fno-builtin` but that seems broken: https://github.com/dlang/druntime/pull/2662#issuecomment-508988793

As far as I have tested, the functions are not turned into LLVM / GDC counterparts, except for the naive versions, which is ok. Other than that, I'm not sure what exactly you seem to want tested.

> The current set of tests is way too small. After just a few seconds of thinking about it I already see things untested: misaligned pointer, `null`, different memset sizes to trigger all corner cases of simd implementations (4-ply, 8-ply, ...), ...

The things you mention are tested except for `null`, which seems like a redundant test for such a performance critical function (as far as I'm aware, it's not done in any other public `mem`* functions).
The tests are in this file: https://github.com/dlang/druntime/pull/2662/files#diff-a414cc2209b63799adab22dd7971e59c

Meaning, except for the basic unittests.

src/core/experimental/memutils.d
baziotis: Because these utilities were not created with the idea to have the exact C interface. We decided to drop legacy C stuff that seem useless. Like the return value and the fact that `memset` gets an `int` instead of a byte.

baziotis: I'm not saying this is necessarily the best idea ever, because no matter how irrelevant legacy stuff is, the thing is, people have been used to that for years.

JohanEngelen:
> > One thing to take into account is that the LLVM optimizer recognizes `memset` and `memcpy` and knows the semantics of those functions. It uses that knowledge for optimization (e.g. if you memcpy something, the optimizer knows that the dst memory now contains the values of src), which can be a massive performance boost.
> > When you provide your own memset function, and the optimizer doesn't know the semantics of it, you may loose performance because of it. Cross-module inlining is tricky to work here:
> > 
> > * your new memset needs to be compiled without the compiler turning the algorithm into a `memset`/`memcpy` call
> > * application code needs to be compiled _with_ the compiler turning any copying algorithm into `memcpy` or `memset`.
> >   --> this cannot be done in one compiler optimization unit..............
> > 
> > Before using/enabling this on the optimizing compilers, please do _extensive_ testing and assembly output checking.
> 
> That is true. It's supposed to be done on GDC as well. It's supposed that on GDC, you can set the `-fno-builtin` but that seems broken: [#2662 (comment)](https://github.com/dlang/druntime/pull/2662#issuecomment-508988793)

Using `-fno-builtin` may break if cross-module inlining is needed to get performance. (because in user code we don't want `-fno-builtin` and then memset may get local codegen without being inlined and inside it the algorithm going to be turned into `memset`. This heavily depends on how cross-module inlining is implemented though. But imagine your `memset` being instantiated in user code for a specific type T; then `Dmemset` gets inlined into that and the code is converted to a call to `memset`. Is that bad? Is the plan to supply a C stdlib compatible `memset` without the need to link to C stdlib? Sorry if unclear, I'm trying to sketch potential issues in more complex use cases).

If you haven't seen it yet, this thread is interesting: https://lists.llvm.org/pipermail/llvm-dev/2019-April/131973.html

> As far as I have tested, the functions are not turned into LLVM / GDC counterparts, except for the naive versions, which is ok. Other than that, I'm not sure what exactly you seem to want tested.

I meant that the performance of this may heavily depend on how it is used in user code + separate compilation. Will the function be opaque or not to the optimizer? Think of user code like this:
```
ubyte[100] buf;
memset(&buf[0], 5, 5);
auto a = buf[3]; // will optimizer know that this is 5? 
```

> > The current set of tests is way too small. After just a few seconds of thinking about it I already see things untested: misaligned pointer, `null`, different memset sizes to trigger all corner cases of simd implementations (4-ply, 8-ply, ...), ...
> 
> The things you mention are tested except for `null`, which seems like a redundant test for such a performance critical function (as far as I'm aware, it's not done in any other public `mem`* functions).
> The tests are in this file: https://github.com/dlang/druntime/pull/2662/files#diff-a414cc2209b63799adab22dd7971e59c

Ah yes I see now how the tests work.
What is still missing from the tests is whether anything is overwritten that shouldn't be overwritten. (If you start implementing `memcpy` you should also check for buffer overreads.)
About `null`: whether the function is performance critical or not is not relevant for testing a corner case input: just test that `memset([], 1)` doesn't do anything. Similar to testing whether it works for structs, just sanity checking and making sure in the future people are not going to break things accidentally.

By the way, perhaps more clear to have all functions take a `ubyte` instead of an `int` (because they are going to write only the ubyte)?

baziotis:
> 
> Using `-fno-builtin` may break if cross-module inlining is needed to get performance. (because in user code we don't want `-fno-builtin` and then memset may get local codegen without being inlined and inside it the algorithm going to be turned into `memset`. This heavily depends on how cross-module inlining is implemented though. But imagine your `memset` being instantiated in user code for a specific type T; then `Dmemset` gets inlined into that and the code is converted to a call to `memset`. Is that bad? Is the plan to supply a C stdlib compatible `memset` without the need to link to C stdlib? Sorry if unclear, I'm trying to sketch potential issues in more complex use cases).
>

Ah, I see ok. I mentioned `-fno-builtin` for.. how should I say.. microbenchmarks (?) I don't know, but basically, when you want to benchmark some `memset` you've made, you have some benchmark code and you try to verify that LLVM uses _your_ code and not his own patch, so `-fno-builtin` is useful (although again, it currently seems broken). Anyway, I get your point. Ok, so I don't know what `memset` LLVM and GCC uses. Is it better than mine or is it a more conservative? And on what platforms is it replaced ? And on what platforms is it faster and vice versa ?

The way I see it, my `memset()` is targetted to x86_64. As far as I have tested, neither LDC or GDC replace it (the non-naive version) with their own. And mine is faster last time I checked.
Now, in other platforms, the compiler always knows better what to do, since my implementation is the most trivial one (so it is trivial for the compiler to know that "hey, this is a e.g. memcpy(), replace it with what you think is better"). If versions for other platforms get added, those are supposed to be better than the LLVM / GCC patch, so I think we're ok there as well.

Although ideally, we should have some control besides flags for saying "we don't want this to be replaced" or smth.

Ok, this was somewhat long, I hope it was clear.

> If you haven't seen it yet, this thread is interesting: https://lists.llvm.org/pipermail/llvm-dev/2019-April/131973.html
>

No, thanks! I'll check it.

> I meant that the performance of this may heavily depend on how it is used in user code + separate compilation. Will the function be opaque or not to the optimizer? Think of user code like this:
> 
> ```
> ubyte[100] buf;
> memset(&buf[0], 5, 5);
> auto a = buf[3]; // will optimizer know that this is 5? 
> ```
> 

Yes, this is an important question. It adds to the questions I wrote above and it's on the point. Well, I don't know how one can tell the compiler that. As I said, I don't even know how to have something like assume_aligned, which is quite important.

> Ah yes I see now how the tests work.
> What is still missing from the tests is whether anything is overwritten that shouldn't be overwritten. (If you start implementing `memcpy` you should also check for buffer overreads.)

Good point. I have tested by repeatedly inspecting on the debugger, but a more formal test should be added.

Here's the `memcpy` PR: https://github.com/dlang/druntime/pull/2687
`memmove` has also been implemented, but it's not PR'd yet (since these will take quite some time to be merged anyway).

> About `null`: whether the function is performance critical or not is not relevant for testing a corner case input: just test that `memset([], 1)` doesn't do anything. Similar to testing whether it works for structs, just sanity checking and making sure in the future people are not going to break things accidentally.
>

Well, it's just another `if`. It's not the end of the world, but it's generally not added.
For example: https://github.com/tpn/agner/blob/master/asmlib/asmlibSrc/memset64.asm#L189
I don't think there's a right answer, it's just a decision. We can change it.

> By the way, perhaps more clear to have all functions take a `ubyte` instead of an `int` (because they are going to write only the ubyte)?

Hmm, you mean the `Dmemset` right? Because the `memset` (I mean, my `memset` not the libc, the one that takes arrays, `ref`s etc.) do take `ubyte`.

src/core/experimental/memutils.d
baziotis: Yes, I'm not very familiar with `scope` but `nogc`, `nothrow` is good.

JohanEngelen:
> > About `null`: whether the function is performance critical or not is not relevant for testing a corner case input: just test that `memset([], 1)` doesn't do anything. Similar to testing whether it works for structs, just sanity checking and making sure in the future people are not going to break things accidentally.
> 
> Well, it's just another `if`. It's not the end of the world, but it's generally not added.
> For example: https://github.com/tpn/agner/blob/master/asmlib/asmlibSrc/memset64.asm#L189
> I don't think there's a right answer, it's just a decision. We can change it.

Ah I didn't mean to add an `if` to the memset functions.
Currently, I think it already works for an empty array (length == 0). Just add a testcase that tests that indeed it doesn't crash or something. Same as adding a testcase like this would be nice:
```
struct A { int i; ubyte b; };
memset(A[] or A...);
```
 
> > By the way, perhaps more clear to have all functions take a `ubyte` instead of an `int` (because they are going to write only the ubyte)?
> 
> Hmm, you mean the `Dmemset` right? Because the `memset` (I mean, my `memset` not the libc, the one that takes arrays, `ref`s etc.) do take `ubyte`.

Yes indeed.

baziotis:
> Ah I didn't mean to add an `if` to the memset functions.
> Same as adding a testcase like this would be nice:
> 
> ```
> struct A { int i; ubyte b; };
> memset(A[] or A...);
> ```

Sorry, I did not get that one.

> > Hmm, you mean the `Dmemset` right? Because the `memset` (I mean, my `memset` not the libc, the one that takes arrays, `ref`s etc.) do take `ubyte`.
> 
> Yes indeed.

Yes, `ubyte` makes more sense for `Dmemset` as well, I just left it because `Dmemset` has a  more C-like interface. However, since it already has deviated from it, it's better to change to `ubyte`.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,258 @@
...
+    {
+        enum useSIMD = false;
+    }
+}
+
+version (useSIMD)
+{
+    /* SIMD implementation
+     */
+    private void Dmemset(void *d, const uint val, size_t n) nothrow @nogc
```
TurkeyMan: I can't see a path for `val == 0`... It is usually possible to make zeromem substantially faster.
But again, none of this code should apply to LDC/GDC, those compilers have intrinsics for this stuff; this should be considered strictly for DMD.
For instance, on PowerPC, there's an instruction `dcbz` (data cache-block zero) which resets an entire 256byte cache line to zero in one instantaneous operation... I can't see your PPC branch that takes advantage of 256byte alignment and `bcbz`... but the backends all do.
Assuming this is targetting DMD, then your assumptions that the compiler won't do anything interesting, and that the target is x64 are both reasonable assumptions, otherwise I can't get behind any of this code.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,258 @@
...
+        // than the previous classic switch. BUT. Using the switch had a significant
+        // drop in the rest of the sizes. It's not the branch that is responsible for the drop,
+        // but the fact that it's more difficult to optimize it as part of the rest of the code.
+        if (n < 32)
+        {
+            memsetNaive(d, val, n);
+            return;
+        }
+        void *temp = d + n - 0x10;                  // Used for the last 32 bytes
+        const uint v = val * 0x01010101;            // Broadcast c to all 4 bytes
```
TurkeyMan: mul is almost never a good idea. I suspect the mul may be a bottleneck in the ~32-128 byte range. If you're gonna do it, why not extend to 8 bytes and get more value from it?

Do you have a reference for this code? It doesn't look very optimal to me. If you're gonna use SSE, there are broadcast and permute functions, which not introduce hazards as bad as mul...

src/core/experimental/memutils.d
baziotis: > I can't see a path for val == 0... It is usually possible to make zeromem substantially faster.

No `memset` that I know of does anything different for `val == 0` and targets x64.

> But again, none of this code should apply to LDC/GDC, those compilers have intrinsics for this stuff; this should be considered strictly for DMD.

Check this: https://github.com/dlang/druntime/pull/2687#discussion_r310724318

PowerPC was not a target of this project.

> Assuming this is targetting DMD, then your assumptions that the compiler won't do anything interesting, and that the target is x64 are both reasonable assumptions, otherwise I can't get behind any of this code.

The focus was initially DMD, then LDC and GDC by not using their intrinsics. It's again the same question - What happens when libc is not available and do we care?

src/core/experimental/memutils.d
baziotis: Yes, it's inspired by Agner Fog. I didn't copy code from him but I have read his optimization manual. This mul trick is his. And no, a mul in x86 is not a problem nowadays.

Our code turned out to be similar e.g. his AVX version: https://github.com/tpn/agner/blob/master/asmlib/asmlibSrc/memset64.asm#L188

It's also inspired by GCC.

TurkeyMan: I'm hyper-aware of this mul trick, I've been doing this for decades, but it absolutely IS a problem 'nowdays'... I've written a lot of code using mul-tricks like this, and been surprised when they become an unintuitive bottleneck.
It might not be a problem here, because I think it's usually the case that memset is not called in a hot-loop, but multiplies usually have a lot of limitations on the pipeline.
There's usually only one imul pipeline, and while the latency might not be so bad these days, the circuit is kinda recursive, so it can't accept a new instruction per cycle like add/sub circuits, so if there are multiple mul's contending for the limited imul resource, they create stalling hazards on each other.
imul is often used in address calculation, so there's a reasonable possibility of contention, but there's an if above this, so the pipeline might already have flushed...

I do a lot of work with colour precision scaling using this same trick; the imul is almost always the limiting factor in my loops.

Anyway, I only say this because you're shoving the value straight into SSE in the following line, where you can use much faster permute operations to broadcast the value very easily (like pshufb, or complements on other architectures).

src/core/experimental/memutils.d
TurkeyMan: > PowerPC was not a target of this project.

Right, that's my point, and why I think it's not a good idea to re-implement these functions. There's too many versions of them to worry about, and maintaining great perf is a moving target as arch like x86 and arm evolve.

> What happens when libc is not available and do we care?

libc is always available... what is a case when it's not?
I'd also like to know if the intrinsics definitely call out to libc, or if the intrinsics call into the compilers runtime lib? What is the symbol that the intrinsics call?

src/core/experimental/memutils.d
baziotis: I didn't know all that stuff about muls limiting the pipeline. I just knew that the cycles they consume are about the same as say add / sub. Is there somewhere I can read more about those?

But the probabilities of Agner not having thought this is close to 0. :P

src/core/experimental/memutils.d
baziotis: I'll have to look at this. Let me be clear again, I don't know for a fact that they call libc, it just seemed that way as they execute the same code. I'll definitely have to look at that if we move forward with this PR.

baziotis:
Closed due to not being useful - more info [here](http://users.uoa.gr/~sdi1600105/google-summer-of-code.html).

PR #2665, opened 2019-07-06
Add all missing symbols to `loader.d`

jacob-carlborg:
This also updates the file to match the macOS 10.15 SDK.

It's probably easiest to not look at the diff but look at the actual file.

### DIFF
--- a/src/core/sys/darwin/mach/loader.d
+++ b/src/core/sys/darwin/mach/loader.d
@@ -1,27 +1,36 @@
 /**
- * Copyright: Copyright Digital Mars 2010-2018.
- * License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
- * Authors:   Jacob Carlborg
+ * This file describes the format of Mach-O object files.
+ *
+ * D header file for `mach-o/loader.h` from the macOS 10.15 SDK.
+ *
+ * Copyright: Copyright Digital Mars 2010-2019.
+ * License: $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Authors: Jacob Carlborg
  * Version: Initial created: Feb 20, 2010-2018
- * Source: $(DRUNTIMESRC core/sys/darwin/mach/_loade.d)
+ * Source: $(DRUNTIMESRC core/sys/darwin/mach/_loader.d)
  */
 module core.sys.darwin.mach.loader;
 
+import core.stdc.config;
+
 version (CoreDdoc)
 {
-    /// Represents the header of a Mach-O file for 32-bit architecture.
+    /**
+     * The 32-bit mach header appears at the very beginning of the object file
+     * for 32-bit architectures.
+     */
     struct mach_header
     {
-        /// The mach magic number identifier.
+        /// Mach magic number identifier.
         uint magic;
 
-        /// CPU specifier.
+        /// Cpu specifier.
         int cputype;
 
         /// Machine specifier.
         int cpusubtype;
 
-        /// The type of the file.
+        /// Type of file.
         uint filetype;
 
         /// Number of load commands.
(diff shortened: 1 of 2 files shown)


src/core/sys/darwin/mach/loader.d
```diff
@@ -62,89 +84,489 @@ version (CoreDdoc)
...
         uint cmd;
 
-        /// The size of this segment, includes size of section structs.
+        /// Includes sizeof section structs.
         uint cmdsize;
 
-        /// The name of this segment.
-        char[16] segname = 0;
+        /// Segment name.
+        char[16] segname;
```
thewilsonator: Keep the `= 0`

src/core/sys/darwin/mach/loader.d
```diff
@@ -159,252 +581,2853 @@ version (CoreDdoc)
...
         uint cmd;
 
-        /// The size of this segment, includes size of section structs.
+        /// Includes sizeof section_64 structs.
         uint cmdsize;
 
-        /// The name of this segment.
-        char[16] segname = 0;
+        /// Segment name.
+        char[16] segname;
```
thewilsonator: Ditto throughout

src/core/sys/darwin/mach/loader.d
```diff
@@ -159,252 +581,2853 @@ version (CoreDdoc)
...
+    /**
+     * The entries in the reference symbol table are used when loading the
+     * module (both by the static and dynamic link editors) and if the module is
+     * unloaded or replaced. Therefore all external symbols
+     * (defined and undefined) are listed in the module's reference table. The
+     * flags describe the type of reference that is being made. The constants
+     * for the flags are defined in <mach-o/nlist.h> as they are also used for
+     * symbol table entries.
+     */
+    struct dylib_reference
```
thewilsonator: is this supposed to have members?

src/core/sys/darwin/mach/loader.d
```diff
@@ -159,252 +581,2853 @@ version (CoreDdoc)
...
+     * link editor. If isub-image is 0 the the symbol is expected to be defined
+     * in library and not in the sub-images. If isub-image is non-zero it is an
+     * index into the array of sub-images for the umbrella with the first index
+     * in the sub-images being 1. The array of sub-images is the ordered list of
+     * sub-images of the umbrella that would be searched for a symbol that has
+     * the umbrella recorded as its primary library. The table of contents index
+     * is an index into the library's table of contents. This is used as the
+     * starting point of the binary search or a directed linear search.
+     */
+    struct twolevel_hint
```
thewilsonator: ditto

src/core/sys/darwin/mach/loader.d
jacob-carlborg: Is it only necessary for static arrays of type `char`?

src/core/sys/darwin/mach/loader.d
jacob-carlborg: No. The C struct contains bitfields. So what I did was to use `std.bitmanip.bitfields` and added the code that `bitfields` expands to. What you have commented here on is a `CoreDdoc` version block. This is at the top to be able to generate documentation on all platforms. Otherwise the file is guarded with `version (Darwin):` and the documentation is not generated on macOS. There is one field in the actual definition of the struct: https://github.com/dlang/druntime/pull/2665/files#diff-3ca62e3adaf537d37782e971d010c0e5R3112.

src/core/sys/darwin/mach/loader.d
jacob-carlborg: See above comment: https://github.com/dlang/druntime/pull/2665#discussion_r300856637.

src/core/sys/darwin/mach/loader.d
jacob-carlborg: Fixed.

src/core/sys/darwin/mach/loader.d
jacob-carlborg: Fixed.

thewilsonator:
DAutoTest: `src/core/sys/darwin/mach/loader.d(955): Error: integral constant must be scalar type, not `string``

thewilsonator:
Please taget stable.

jacob-carlborg:
> Please taget stable.

Done.

jacob-carlborg:
What’s up with buildkite?

Geod24:
> What’s up with buildkite?

It does not support stable

jacob-carlborg:
> It does not support stable

So manual merge is required?

PR #2666, opened 2019-07-07
 Fix issue 20019: Symbol not found: _dyld_enumerate_tlv_storage on macOS 10.15

jacob-carlborg:
The symbol `dyld_enumerate_tlv_storage` has been used to iterate all the thread local storage addresses and registered them as roots with the garbage collector.

The symbol has always been a private symbol, not part of the public API. In the latest version of macOS, 10.15 it has been removed.

This replaces the usage of `dyld_enumerate_tlv_storage` by manually iterating the TLV sections of all images and get key that the dynamic loader has setup for each TLS symbol. The key is then used to get the base address of the TLV storage.

The advantage of this change is that it doesn't rely on any private symbols anymore. The disadvantage is that it relies on some implementation details of the dynamic loader. So this might be even worse than the usage of the private symbol `dyld_enumerate_tlv_storage`.

This fix is backwards compatible and works with all version of macOS that D supports.

This PR depends on https://github.com/dlang/druntime/pull/2665.


jacob-carlborg:
Should I target stable? This depends on https://github.com/dlang/druntime/pull/2665.

Geod24:
If I understand the implications correctly, it means that not a single D program will work on OSX 10.15 until they are compiled with a release which includes this PR ?

jacob-carlborg:
> If I understand the implications correctly, it means that not a single D program will work on OSX 10.15 until they are compiled with a release which includes this PR ?

That's correct.

thewilsonator:
> Should I target stable?

Yes.

jacob-carlborg:
Targeting stable now.

p0nce:
Will programs compiled with LDC also need this fix?

thewilsonator:
Yes..

src/rt/sections_osx_x86_64.d
```diff
@@ -183,3 +178,177 @@ ubyte[] getSection(in mach_header* header, intptr_t slide,
...
+extern (C) size_t malloc_size(const void* ptr) nothrow @nogc;
+
+/// Represents a TLS range.
+struct TLSRange
+{
+    /// The start of the range.
+    void* start;
+
+    /// The size of the range.
+    size_t size;
```
WalterBright: The fields should be reversed, so they match D slices.

src/rt/sections_osx_x86_64.d
```diff
@@ -183,3 +178,177 @@ ubyte[] getSection(in mach_header* header, intptr_t slide,
...
+    {
+        return start !is null && size > 0;
+    }
+
+    /// Returns the range as an array.
+    void[] toArray() pure nothrow @nogc
+    {
+        return start[0 .. size];
+    }
+}
```
WalterBright: I can't see the difference between this struct and a simple D dynamic array.

jacob-carlborg:
FYI, I have now finally got a reply from Apple regarding this issue. I asked if there's an alternative to the `_dyld_enumerate_tlv_storage` symbol. The answer was basically what I have already implemented here.

PR #2668, opened 2019-07-07
Link to 32-bit inode symbols on FreeBSD

myfreeweb:
This fixes building on FreeBSD >= 12 without introducing OS version conditions.

---

Since https://github.com/dlang/dmd/pull/8567 is still in endless bikeshedding mode, let's just do this already.

### DIFF
--- a/src/core/sys/freebsd/sys/event.d
+++ b/src/core/sys/freebsd/sys/event.d
@@ -144,6 +144,7 @@ enum
 }
 
 int kqueue();
+pragma(mangle, "kevent@FBSD_1.0")
 int kevent(int kq, const kevent_t *changelist, int nchanges,
            kevent_t *eventlist, int nevents,
            const timespec *timeout);
--- a/src/core/sys/freebsd/sys/mount.d
+++ b/src/core/sys/freebsd/sys/mount.d
@@ -288,17 +288,17 @@ enum uint VQ_FLAG2000 = 0x2000;
 enum uint VQ_FLAG4000 = 0x4000;
 enum uint VQ_FLAG8000 = 0x8000;
 
-int fhopen(const fhandle_t*, int);
-int fhstat(const fhandle_t*, stat_t*);
-int fhstatfs(const fhandle_t*, statfs_t*);
-int fstatfs(int, statfs_t*);
-int getfh(const char*, fhandle_t*);
-int getfsstat(statfs_t*, c_long, int);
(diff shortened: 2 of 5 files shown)


jmdavis:
Also, this approach introduces the problem that any attempt to use the druntime structs with any other C libraries on FreeBSD 12 won't work, because the structs will be wrong.

myfreeweb:
> how would we move to using the correct version of these functions on FreeBSD 12 down the line?

Exactly like how you would without this patch — the existing declarations would be versioned to <=11, and the >=12 case would be either unversioned symbols or symbols with the new versions (FBSD_1.5 in most cases).

Note that linking to versioned symbols would allow targeting <=11 when working on >=12.

[This is what Rust does](https://github.com/rust-lang/libc/commit/7437d0a6f18625b90a9d2fc40f3ecbc98a2e95fc#diff-12f5e3a0b920a79f7608cd3ff2982c65R822) (though they only added the 12 ABI for testing right now, it's not shipped at all).

> any other C libraries

Do they even exist? Has anyone ever actually made a third party libc that works on FreeBSD?

myfreeweb:
(Actually, the libc *is the public API* of FreeBSD, which is why alternative libcs were never encouraged or supported.)

jmdavis:
> Exactly like how you would without this patch — the existing declarations would be versioned to <=11, and the >=12 case would be either unversioned symbols or symbols with the new versions (FBSD_1.5 in most cases).

Okay. So, basically, long term, this doesn't fix anything, but it shouldn't get in the way of anything either. It just lets us build what are essentially FreeBSD 11 binaries on FreeBSD 12 for the time being.

> Do they even exist? Has anyone ever actually made a third party libc that works on FreeBSD?

I wasn't talking about alternate libcs. I was talking about other C libraries that might use any of the definitions from libc in their own APIs. I don't know how big an issue that is in practice, though right now, nothing works on FreeBSD 12 anyway. So, while this could have problems, it's better than where we are at present.

myfreeweb:
> other C libraries that might use any of the definitions from libc in their own APIs

Ah, that's a non-issue. There's absolutely nothing wrong with using the 1.0 (<=11, 32-bit inode) symbols and 1.5 (>=12, 64-bit inode) symbols in the same program!

The only way to screw something up is *passing statfs etc. structs* between druntime and random external C libs. I don't see that happening..

PR #2671, opened 2019-07-09
memutils: memmove() / memcpy() implementation

baziotis:
Notes:
1) I could not find a better way to handle both static and dynamic arrays (that includes `isArray`).
2) More tests to be added.
3) To check benchmarks, I recommend the main repo: https://github.com/baziotis/Dmemutils

### DIFF
--- a/src/core/experimental/memutils.d
+++ b/src/core/experimental/memutils.d
@@ -1,0 +1,685 @@
+/**
+ * Pure D replacement of the C Standard Library basic memory building blocks of string.h
+ * Source: $(DRUNTIMESRC core/experimental/memutils.d)
+ */
+module core.experimental.memutils;
+
+void memmove(T)(T *dst, const T *src)
+{
+    void *d = cast(void*) dst;
+    const(void) *s = cast(void*) src;
+    if ((cast(ulong)d - cast(ulong)s) < T.sizeof)
+    {  // There is overlap with dest being ahead. Use backwards move.
+        Dmemmove_back(d, s, T.sizeof);
+    }
+    else if ((cast(ulong)s - cast(ulong)d) < T.sizeof)
+    {  // There is overlap with src being ahead. Use backwards move.
+        Dmemmove_forw(d, s, T.sizeof);
+    }
+    else
+    {  // There is no overlap, use memcpy.
+        Dmemcpy(dst, src);
+    }
+}
+
+unittest
+{
+    real a = 1.2;
+    real b;
+    memmove(&b, &a);
+    assert(b == 1.2);
+ 
+    // Overwrite the type system and create overlap with dst forward.   
+    ubyte[8] buf;
+    int *p = cast(int*) buf.ptr;
+    int *q = cast(int*) (buf.ptr + 2);
+    *p = 203847;
+    memmove(q, p);
+    assert(*q == 203847);
+
+    // Create overlap with src forward.
+    *q = 92239;
+    memmove(p, q);
+    assert(*p == 92239);
+}
+
+/* Dynamic Arrays
+ */
+void memmove(T)(T[] dst, const T[] src)
+{
+    mixin(arrayCode);
+}
+
+/* Static Arrays
+ */
+void memmove(T, size_t len)(ref T[len] dst, ref const T[len] src)
+{
+    mixin(arrayCode);
+}
+
+enum arrayCode = "
+    assert(dst.length == src.length);
+    void *d = cast(void*) dst.ptr;
+    const void *s = cast(const(void)*) src.ptr;
+    size_t n = dst.length * typeof(dst[0]).sizeof;
+    if ((cast(ulong)d - cast(ulong)s) < n)
+    {  // There is overlap with dest being ahead. Use backwards move.
+        Dmemmove_back(d, s, n);
+    }
+    else if ((cast(ulong)s - cast(ulong)d) < n)
+    {  // There is overlap with src being ahead. Use backwards move.
+        Dmemmove_forw(d, s, n);
+    }
+    else
+    {  // There is no overlap, use memcpy.
+        pragma(inline, true);
+        Dmemcpy(d, s, n);
+    }";
+
+
+
+unittest
+{
+    const float[3] a = [1.2, 3.4, 5.8];
+    float[3] b;
+    memmove(b, a);
+    assert(b[0] == 1.2f);
+    assert(b[1] == 3.4f);
+    assert(b[2] == 5.8f);
+}
+
+/* Can we use SIMD?
+ */
+import core.simd: float4;
+version (D_SIMD)
+{
+    enum useSIMD = true;
+}
+else version (LDC)
+{
+    // LDC always supports SIMD (but doesn't ever set D_SIMD) and
+    // the back-end uses the most appropriate size for every target.
+    enum useSIMD = true;
+}
+else version (GNU)
+{
+    // GNU does not support SIMD by default.
+    version (X86_64)
+    {
+        enum isX86 = true;
+    }
+    else version (X86)
+    {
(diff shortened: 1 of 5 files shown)


src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
...
+    }
+    else if ((cast(ulong)s - cast(ulong)d) < n)
+    {  // There is overlap with src being ahead. Use backwards move.
+        Dmemmove_forw(d, s, n);
+    }
+    else
+    {  // There is no overlap, use memcpy.
+        pragma(inline, true);
+        Dmemcpy(d, s, n);
+    }";
```
wilzbach: Private. Also, what's the motivation for the split-up? Why can't the static array implementation call the other one?

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
...
+    {  // There is overlap with src being ahead. Use backwards move.
+        Dmemmove_forw(d, s, T.sizeof);
+    }
+    else
+    {  // There is no overlap, use memcpy.
+        Dmemcpy(dst, src);
+    }
+}
+
+unittest
```
wilzbach: This needs `///` in order to be part of the documentation.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
...
+    const float[3] a = [1.2, 3.4, 5.8];
+    float[3] b;
+    memmove(b, a);
+    assert(b[0] == 1.2f);
+    assert(b[1] == 3.4f);
+    assert(b[2] == 5.8f);
+}
+
+/* Can we use SIMD?
+ */
```
wilzbach: This comment isn't really helpful.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
...
+        Dmemmove_forw(d, s, n);
+    }
+    else
+    {  // There is no overlap, use memcpy.
+        pragma(inline, true);
+        Dmemcpy(d, s, n);
+    }";
+
+
+
```
wilzbach: Only one line of whitespace between functions.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
...
+    assert(b[0] == 1.2f);
+    assert(b[1] == 3.4f);
+    assert(b[2] == 5.8f);
+}
+
+/* Can we use SIMD?
+ */
+version (D_SIMD)
+{
+    import core.simd: float4;
```
wilzbach: DStyle: ` : `

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
+/**
+ * Pure D replacement of the C Standard Library basic memory building blocks of string.h
+ * Source: $(DRUNTIMESRC core/experimental/memutils.d)
+ */
+module core.experimental.memutils;
+
+void memmove(T)(T *dst, const T *src)
```
wilzbach: Needs full Ddoc comment.
Also, what's the reason for not reason `ref`?

wilzbach: Also, thinking more about this interface `src, dst` would be more D-like, but I guess we can't really change that.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
...
+    real b;
+    memmove(&b, &a);
+    assert(b == 1.2);
+    // Overwrite the type system and create overlap with dst forward.
+    ubyte[8] buf;
+    int *p = cast(int*) buf.ptr;
+    int *q = cast(int*) (buf.ptr + 2);
+    *p = 203847;
+    memmove(q, p);
+    assert(*q == 203847);
```
wilzbach: Split up into multiple tests, no?

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
...
+    assert(b[1] == 3.4f);
+    assert(b[2] == 5.8f);
+}
+
+/* Can we use SIMD?
+ */
+version (D_SIMD)
+{
+    import core.simd: float4;
+    enum useSIMD = true;
```
wilzbach: Needs to be private.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
...
+    else version (DigitalMars)
+    {
+        import core.simd: void16, loadUnaligned, storeUnaligned;
+    }
+    else version (GNU)
+    {
+        import gcc.builtins : __builtin_ia32_storeups, __builtin_ia32_loadups;
+    }
+
+    void store16f_sse(void *dest, float4 reg)
```
wilzbach: Private

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
...
+        enum useSIMD = false;
+    }
+}
+else
+{
+    enum useSIMD = false;
+}
+
+/* Little SIMD library
+ */
```
wilzbach: Maybe move into a separate module?

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
...
+        // Move pointers to their start.
+        s -= n;
+        d -= n;
+        goto START;
+    }
+}
+
+/* Handle dynamic sizes < 64 with forwards move. Overlap is possible.
+ */
+void Dmemmove_forw_lt64(void *d, const(void) *s, size_t n)
```
wilzbach: Private

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
...
+        else version (DigitalMars)
+        {
+            storeUnaligned(cast(void16*)dest, reg);
+        }
+        else version (GNU)
+        {
+            __builtin_ia32_storeups(cast(float*) dest, reg);
+        }
+    }
+    float4 load16f_sse(const(void) *src)
```
wilzbach: No underscore - see the DStyle.

src/core/experimental/memutils.d
baziotis: The most important reason is that I couldn't make `(ref T[] dst, ref const T[] src)` work when the arguments are `!()(float[3], const(float[3]))`. I don't know why that is.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
+/**
+ * Pure D replacement of the C Standard Library basic memory building blocks of string.h
+ * Source: $(DRUNTIMESRC core/experimental/memutils.d)
+ */
+module core.experimental.memutils;
+
+void memmove(T)(T *dst, const T *src)
+{
+    void *d = cast(void*) dst;
```
thewilsonator: please write `void* d` not `void *d`, this is D not C. ditto throughout.

src/core/experimental/memutils.d
```diff
@@ -0,0 +1,685 @@
...
+}
+
+/* Static Arrays
+ */
+void memmove(T, size_t len)(ref T[len] dst, ref const T[len] src)
+{
+    mixin(arrayCode);
+}
+
+enum arrayCode = "
```
thewilsonator: if you're going to use enum+mixin, please use `q{}` strings.

src/core/experimental/memutils.d
thewilsonator: one possibility is to make `core.experimental.mem` a package with modules `cpy`, `move`, and `simd`.

wilzbach: Better name it `memory` as there's `core.memory`.

baziotis: Ok, I'm making a new PR. I have seen all the comments, I left them to be resolved in the new PR.

PR #2673, opened 2019-07-13
Cleanup and fix implementation of _d_array{setlengthT,catnTX}Trace

Vild:
This should be the final fix needed to be able to get https://github.com/dlang/dmd/pull/10106 to pass all the unittest.

This code uses a pure bypass technique I found inside of dmd's source code so no more ugly hacky casts are needed for calling `GC.stats()` and `accumulate`.
I've also refactored the two previous *Trace functions into `HookTraceImpl` to make sure that all the Trace functions are implemented correctly and to make it easier to change in the future.
The `HookTraceImpl` also contains a `if (...) assert(0);` check to make sure that the compiler does not remove the call to `accumulate`, which is what is currently happening in https://github.com/dlang/dmd/pull/10106.

### DIFF
--- a/src/rt/array/utils.d
+++ b/src/rt/array/utils.d
@@ -1,0 +1,94 @@
+/**
+ This module contains utility functions to help the implementation of the runtime hook
+
+  Copyright: Copyright Digital Mars 2000 - 2019.
+  License: Distributed under the
+       $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+     (See accompanying file LICENSE)
+  Source: $(DRUNTIMESRC rt/_array/_utils.d)
+*/
+module rt.array.utils;
+
+import core.internal.traits : Parameters;
+
+private auto gcStatsPure() nothrow pure
+{
+    import core.memory : GC;
+
+    auto impureBypass = cast(GC.Stats function() pure nothrow)&GC.stats;
+    return impureBypass();
+}
+
+private ulong accumulatePure(string file, int line, string funcname, string name, ulong size) nothrow pure
+{
+    static ulong impureBypass(string file, int line, string funcname, string name, ulong size) @nogc nothrow
+    {
+        import core.internal.traits : externDFunc;
+
+        alias accumulate = externDFunc!("rt.profilegc.accumulate", void function(string file, uint line, string funcname, string type, ulong sz) @nogc nothrow);
+        accumulate(file, line, funcname, name, size);
+        return size;
+    }
+
+    auto func = cast(ulong function(string file, int line, string funcname, string name, ulong size) @nogc nothrow pure)&impureBypass;
+    return func(file, line, funcname, name, size);
+}
+
+/**
+ * TraceGC wrapper around runtime hook `Hook`.
+ * Params:
+ *  T = Type of hook to report to accumulate
+ *  Hook = The hook to wrap
+ *  errorMessage = The error message incase `version != D_TypeInfo`
+ *  file = File that called `HookTraceImpl`
+ *  line = Line inside of `file` that called `HookTraceImpl`
+ *  funcname = Function that called `HookTraceImpl`
+ *  parameters = Parameters that will be used to call `Hook`
+ * Bugs:
+ *  This function template was ported from a much older runtime hook that bypassed safety,
+ *  purity, and throwabilty checks. To prevent breaking existing code, this function template
+ *  is temporarily declared `@trusted pure nothrow` until the implementation can be brought up to modern D expectations.
+*/
+auto HookTraceImpl(T, alias Hook, string errorMessage)(string file, int line, string funcname, Parameters!Hook parameters) @trusted pure nothrow
+{
+    version (D_TypeInfo)
+    {
+        pragma(inline, false);
+        string name = T.stringof;
+
+        // FIXME: use rt.tracegc.accumulator when it is accessable in the future.
+        version (tracegc)
+        {
+            import core.stdc.stdio;
+
+            printf("%sTrace file = '%.*s' line = %d function = '%.*s' type = %.*s\n",
+                Hook.stringof.ptr,
+                file.length, file.ptr,
+                line,
+                funcname.length, funcname.ptr,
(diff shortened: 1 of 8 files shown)


aG0aep6G:
That "pure bypass technique" looks like a compiler bug.

Vild:
> That "pure bypass technique" looks like a compiler bug.

I don't fully understand why this technique exist. It (or at least the comment) was added 5 years ago:
https://github.com/dlang/dmd/commit/0acb94c7b8198b28386ed3c298e8c8a6829340be#diff-31ec3d6838ade0264adec0944948507dR2399

aG0aep6G:
> I don't fully understand why this "technique" exist. It (or at least the comment) was added 5 years ago:
> [dlang/dmd@0acb94c#diff-31ec3d6838ade0264adec0944948507dR2399](https://github.com/dlang/dmd/commit/0acb94c7b8198b28386ed3c298e8c8a6829340be#diff-31ec3d6838ade0264adec0944948507dR2399)

I can't say that I understand what that comment is trying to say. I suspect it might just be wrong.

Anyway, I'm sure that we're looking at a compiler bug. Calling an impure function from a pure one can't be allowed (without a cast).

Filed: https://issues.dlang.org/show_bug.cgi?id=20047

JinShil:
> I don't fully understand why this "technique" exist.

This behavior was introduced in https://github.com/dlang/dmd/pull/4344 as a fix to https://issues.dlang.org/show_bug.cgi?id=14039

JinShil:
@WalterBright, you filed [Issue 14039](https://issues.dlang.org/show_bug.cgi?id=14039) and approved and merged the fix at https://github.com/dlang/dmd/pull/4344.  Can you please help reconcile this?  Is the technique employed in this PR a bug or a legitimate feature?

aG0aep6G:
> Anyway, I'm sure that we're looking at a compiler bug. Calling an impure function from a pure one can't be allowed (without a cast).
> 
> Filed: https://issues.dlang.org/show_bug.cgi?id=20047

And a PR to fix it: https://github.com/dlang/dmd/pull/10172

jpf91:
@Vild if you use the function pointer / literal / cast bypass technique instead, does that trigger the optimizer problem? AFAIK this is the commonly used technique to bypass pure.

RazvanN7:
Is there a possibility that if dlang/dmd#10172 is merged, then this PR might not be needed?

JinShil:
> Is there a possibility that if dlang/dmd#10172 is merged, then this PR might not be needed?

I believe that if dlang/dmd#10172 is merged, this PR will no longer work.

Vild:
> @Vild if you use the function pointer / literal / cast bypass technique instead, does that trigger the optimizer problem? AFAIK this is the commonly used technique to bypass pure.

I could move back to the pointer cast method, I just felt that it looked too hacky compared to this bypass method when I found it inside of dmd's code. 

> Is there a possibility that if dlang/dmd#10172 is merged, then this PR might not be needed?

I still need to apply the `if (!accumulate(...)) assert(0);` fix to not allow the optimizer to remove that call. So I just need to use `cast`s instead of this bypass method.

Vild:
I've changed it so it uses cast instead of that other method to fake purity.

src/rt/array/utils.d
```diff
@@ -0,0 +1,94 @@
...
+            ulong size = gcStatsPure().allocatedInCurrentThread - currentlyAllocated;
+            if (size > 0)
+                if (!accumulatePure(file, line, funcname, name, size)) {
+                    // This 'if' and 'assert' is needed to force the compiler to not remove the call to
+                    // `accumulatePure`. It really want to do that while optimizing as the function is
+                    // `pure` and it does not influence the result of this hook.
+
+                    // `accumulatePure` returns the value of `size`, which can never be zero due to the
+                    // previous 'if'. So this assert will never be triggered.
+                    assert(0);
```
jpf91: BTW, reading this I think there's may be a deeper issue whenever we have 'fake' pure code which internally calls impure code.

I guess the intent of such code is to not influence purity of the caller (i.e. completely optimizing away the caller would be fine), but the callee is not really pure, so it should not be removed on it's own. I wonder whether this can be expressed in any theoretically sound way, @tgehr any insights here?

It seems to me that the problem is that we lie to the compiler and simply tell it `this function is pure`. Maybe we should rather tell it `Ignore purity checks for this call, but the function is not pure`. Intuitively I'd say a
```d 
pure void foo()
{
    __traits(forcepure, bar());
}
```
Should work: if bar is not inlined, this would tell the compiler to keep the call, but ignore purity for purity calculation of the callee. With inlining, the compiler should be able to see the side effects anyway and should not optimize these away.

jpf91: I can see how your `assert` code has the same effect though: It marks the return value as used, as long as the call is not inlined. When it's inlined, the compiler may remove the assert, but it will see the impure parts of accumulatePure and keep these. This can probably also somehow solved like this: https://stackoverflow.com/questions/40122141/preventing-compiler-optimizations-while-benchmarking, but I'd prefer a compiler hook which should also work for void functions.

Vild: I've been thinking about introducing `HookCallExp` or a bool in `CallExp` that ignore the purity and somehow allowed the code to be inline, which is disallowed right now `pragma(inline, false)`, whilst being able to be detected as the hook (`HookCompoundStatement`?) so the CTFE intercept code works.

thewilsonator: That sounds like a great idea.

tgehr: > BTW, reading this I think there's may be a deeper issue whenever we have 'fake' pure code which internally calls impure code.
> 
> I guess the intent of such code is to not influence purity of the caller (i.e. completely optimizing away the caller would be fine), but the callee is not really pure, so it should not be removed on it's own. I wonder whether this can be expressed in any theoretically sound way, @tgehr any insights here?
> ...

@jpf91 This was what `__mutable` functions were for in the original `__mutable` proposal. Inlining for such functions is not so easy if you want to preserve constraints on evaluation order precisely.

PR #2674, opened 2019-07-14
Fix issue 20049 - object.destroy doesn't propagate attributes 

ShigekiKarita:
A declaration of `rt_finalize` in object.d have not match to its implementation https://github.com/dlang/druntime/blob/3ba98e659dc6857d7f5ad6383adc2b02510ce597/src/rt/lifetime.d#L1423
Because of this mismatch, we could not destroy class objects in nothrow. 

PS. This issue is found in mir-algorithm https://github.com/libmir/mir-algorithm/issues/208


9il:
Issue https://issues.dlang.org/show_bug.cgi?id=20049

thewilsonator:
If this fixes the linked issue please change the commit title to "Fix issue 20049 - [rest of title]". 

Also please target stable.

JinShil:
Also, please add a test case to help prevent a regression.

ShigekiKarita:
Thanks. Just a moment. I'll fix that.

thewilsonator:
You've pulled all of masters commits along with it. You need to rebase your commit on top of stable in addition to retargeting the branch in GitHub.

ShigekiKarita:
@JinShil I added the regression test. Can you check if it is correctly placed (test/exceptions/src/nothrow_dtor.d)?

JinShil:
@ShigekiKarita  Can you just add a `unittest` in object.d just below the `destroy` implementation?  I think something like this should work:

```D
// Test to ensure proper behavior of `nothrow` destructors
unittest
{
    class C
    {
        static int dtorCount = 0;
        this() nothrow {}
        ~this() nothrow { dtorCount++; }
    }

    auto c = new C;
    destroy(c);
    assert(C.dtorCount == 1);
}
```

ShigekiKarita:
Thanks. I think `nothrow unittest` is also fine.

src/object.d
```diff
@@ -882,6 +882,21 @@ nothrow @safe @nogc unittest
     }
 }
 
+// Test to ensure proper behavior of `nothrow` destructors
```
thewilsonator: Please reference the buzilla issue number, otherwise looks good to go.

ShigekiKarita: Thanks.

PR #2675, opened 2019-07-14
Fix issue 20048: Unittest printing fails on Windows when using msvcr100.dll

BorisCarvajal:
The format specifier "%zu" doesn't exist in the MS C runtime before VS2015.
The segfault happens using DMD with Mingw-w64 libraries.


rainers:
Please target stable.

PR #2676, opened 2019-07-15
Replace usages of `in` with `const scope`

JinShil:
In support of https://github.com/dlang/dmd/pull/10179


JinShil:
Mecca failure seems to be unrelated.  There may be a runtime mismatch.  See https://github.com/weka-io/mecca/pull/23

Geod24:
Uh, Mecca should be fully fixed :/

PR #2679, opened 2019-07-16
Add core.experimental.refcount (continuation of #2608)

lesderid:
Continuation of https://github.com/dlang/druntime/pull/2608

This PR adds a struct `__RefCount` that can be used as a common building block for reference counted types through composition.

A short summary of the known issues that were brought up in the previous thread and earlier discussions:
 1. The type needs to be prefixed with `__` because the compiler might need to be made aware of its existence in the future (e.g. to inhibit elision of calls to `const pure` functions like `addRef`/`delRef`)
 2. It's a non-intrusive approach, i.e. it doesn't use space lent by the object it's tracking, which means worse locality of reference and more (de)allocations
 3. Because of implicit conversion of return values of `pure` functions to `immutable` (and `immutable` implying `shared`), we have to conservatively use atomic increments/decrements for the counter variable

None of these issues are bad enough to keep blocking progress. Early inclusion in `core.experimental` allows us to experiment with reference counted containers using this reference counting scheme (e.g. `rcarray!T`, https://github.com/dlang/druntime/pull/2646) while not yet fully committing to a specific implementation or API.


src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+             user defined default constructor
+    */
+    @nogc nothrow pure @trusted scope
+    this(this Q)(int _)
+    {
+        // We are required to always use a shared support as the result of a `pure`
+        // function is implicitly convertible to `immutable`.
+
+        shared CounterType* support = cast(shared CounterType*) pureAllocate(CounterType.sizeof);
+        *support = 1; // Start with 1 to avoid calling an `addRef` from 0 to 1
```
thewilsonator: This will fail to compile with the new (not yet implemented) restricted shared, please do `*cast()support = 1;.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+reference counting.
+
+`__RefCount` was designed to be composed as a field inside a user-defined type.
+The user is responsible to initialize the `__RefCount` in that type's constructors.
+The user will call the `isUnique()` method to decide if this is the last
+reference to the enclosing type so memory can be safely deallocated.
+
+Implementation: The internal implementation of `__RefCount` uses `malloc`/`free`.
+
+Important: The `__RefCount` member must be initialized through a call to its
```
wilzbach: Can't we just disable the default initialization to enforce this?

```d
@disable this();
```

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+    /*
+    Perform `rc op val` operation. Always use atomics as the counter is shared.
+
+    Returns:
+         The result of the operation.
+    */
+    @nogc nothrow pure @trusted scope
+    private CounterType rcOp(this Q, string op)(CounterType val) const
+    {
+        return cast(CounterType)(atomicOp!op(*(cast(shared CounterType*) rc), val));
```
wilzbach:
```suggestion
        return cast(CounterType) (atomicOp!op(*(cast(shared CounterType*) rc), val));
```

Dstyle spacing

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+    }
+
+    /**
+    Creates a new `__RefCount` instance. Its memory is internally managed with
+    malloc/free.
+
+    Params:
+         _ = an unused int value; required because structs don't have a
+             user defined default constructor
```
wilzbach: Hmm shouldn't we make this private then and only provide a `static` create/make function around it?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+    {
+        mixin(copyCtorIncRef);
+    }
+
+    /**
+    Copy constructs a $(B const __RefCount) from a mutable reference, `rhs`.
+    This increases the reference count.
+    */
+    @nogc nothrow pure @safe scope
+    this(return scope ref typeof(this) rhs) const
```
wilzbach: Could this be `inout` to avoid some duplications?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+         $(BIGOH 1).
+    */
+    @nogc nothrow pure @safe scope
+    ref __RefCount opAssign(return scope ref typeof(this) rhs) return
+    {
+        if (rhs.isInitialized() && rc == rhs.rc)
+        {
+            return this;
+        }
+        if (rhs.isInitialized())
```
wilzbach: why not move this into the first clause? this would save checking this twice

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+        {
+            rhs.addRef();
+        }
+        if (isInitialized())
+        {
+            delRef();
+        }
+        () @trusted { rc = rhs.rc; }();
+        return this;
```
wilzbach: Can we use a mixin here to avoid the duplication?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+    Increase the reference count. This asserts that `__RefCount` is initialized.
+
+    Returns:
+         This returns a `void*` so the compiler won't optimize away the call
+         to this `const pure` function.
+    */
+    @nogc nothrow pure @safe scope
+    private void* addRef(this Q)() const
+    {
+        assert(isInitialized(), "[__RefCount.addRef] __RefCount is uninitialized");
```
wilzbach: Maybe use `.stringof` to get the full name of the RefCount member, s.t. the user knows which one it was?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+        () @trusted { rc = rhs.rc; }();
+        return this;
+    }
+
+    /*
+    Increase the reference count. This asserts that `__RefCount` is initialized.
+
+    Returns:
+         This returns a `void*` so the compiler won't optimize away the call
+         to this `const pure` function.
```
wilzbach: IIRC DMD never does this and LDC has a special attribute for this, though I could only find `@weak` :/

https://wiki.dlang.org/LDC-specific_language_changes

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+    }
+
+    version (CoreUnittest)
+    {
+        pure nothrow @nogc @trusted scope
+        private bool isValueEq(this Q)(uint val) const
+        {
+            return *getUnsafeValue == val;
+        }
+    }
```
wilzbach: Why do we need this? Using `version(unittest)` directives has bitten us a lot in the past as all tests passed, but the end-users got a different behavior -> I recommend moving this into the actual `unittest` block.
One can access private members in the same file.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+        assert(!a.rc.isUnique);
+        // a2 and a3 go out of scope here
+        // a2 is the last ref to rcarray(4242) -> gets freed
+    }
+    assert(a.rc.isUnique);
+}
+
+version (CoreUnittest)
+{
+    private struct StatsAllocator
```
wilzbach: This isn't used anymore?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+        auto a3 = rcarray(4242);
+        a2 = a3; // Assign a3 into a2; a's ref count drops
+        assert(a.rc.isUnique);
+        a3 = a; // Assign a into a3; a's ref count increases
+        assert(!a.rc.isUnique);
+        // a2 and a3 go out of scope here
+        // a2 is the last ref to rcarray(4242) -> gets freed
+    }
+    assert(a.rc.isUnique);
+}
```
wilzbach: Can we add a few more test to increase the coverage of this module and ensure it works as expected?
Maybe even one with multiple threads to ensure the atomic ops work too.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
+// Written in the D programming language.
+/**
+This module provides a composable reference count implementation in the form
+of `__RefCount`.
+
+License: $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+
+Authors: Eduard Staniloiu
```
wilzbach: @lesderid please feel free to add your name here.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,548 @@
...
+*/
+struct __RefCount
+{
+    import core.atomic : atomicOp;
+
+    alias CounterType = uint;
+    private shared CounterType* rc = null;
+
+    /*
+    Perform `rc op val` operation. Always use atomics as the counter is shared.
```
wilzbach: Hmm, I guess this is the most controversial part of this PR as this will make it very slow for single-threaded programs.
Can't we separate the overloads, s.t. this is only done when `__RefCount` is actually shared (and thus its `shared` functions are called?).
Alternatively, maybe we need to add a `__TLSRefCount`?

src/core/experimental/refcount.d
thewilsonator: why is this and rcOp marked pure in the first place? They very clearly aren't.

src/core/experimental/refcount.d
lesderid: Sadly, we can't. With implicit conversion of `immutable(__RefCount)` to `const(__RefCount)` we can't rely on `this` being shared.

The solution @edi33416 was trying out for this used alignment as a marker for `__RefCount`s that were constructed as `shared` (e.g. `immutable`), so after implicit conversion to `const` we can still know it was initially `immutable`. 

However this doesn't work either, as the result of `pure` functions can be implicitly converted to `immutable`. This means that when creating an returning a `__RefCount` from a `pure` function, the non-`shared` constructor ends up being called for a `shared` object, meaning it gets constructed with non-`shared` alignment, and we're right back at square one.

src/core/experimental/refcount.d
lesderid: That just prints `module refcount` (because `__RefCount`isn't a template?). Anyway, this is `private` and the stacktrace should be enough, so I'm just removing the bracketed stuff.

src/core/experimental/refcount.d
lesderid: Yes, made the `const` ones `inout`.

src/core/experimental/refcount.d
lesderid: This breaks `rcarray` for some reason: ```src/core/experimental/array.d(676): Error: variable `core.experimental.array.rcarray!int.rcarray.opSlice!(immutable(rcarray!int)).opSlice.__copytmp2437` default construction is disabled for type immutable(__RefCount)```

src/core/experimental/refcount.d
lesderid: I changed `shared CounterType*` to `shared(CounterType)*`, as it's the counter that should be `shared`, not the pointer to it (as @thewilsonator pointed out in private). This makes this line compile with `-preview=restrictiveshared` (https://github.com/dlang/dmd/pull/10142), but that seems like a bug in the restrictive shared implementation to me?

src/core/experimental/refcount.d
lesderid: Can we just get away with returning void then?

@wilzbach Are you sure LDC has an UDA for this? Can't find it here: https://github.com/ldc-developers/ldc/blob/eaa65fca18562dd71b4274a09763124cc2643934/gen/uda.cpp#L409

lesderid: @thewilsonator I'm not entirely sure, but I believe it should be weakly `pure` because it only modifies `this` and only calls other `pure` functions?

aG0aep6G: > I believe it should be weakly `pure` because it only modifies `this` and only calls other `pure` functions?

If it modifies `this`, it's clearly not `const`.

lesderid: > If it modifies `this`, it's clearly not `const`.

You're right, it only uses `atomicOp` to increment/decrement `*rc` so it doesn't even modify `this`.

aG0aep6G: > it only uses `atomicOp` to increment/decrement `*rc` so it doesn't even modify `this`

If you're reaching `*rc` via a const `this`, that's still invalid. Casting away const and mutating the data is not allowed. That includes const pointers to otherwise mutable data.

lesderid: I don't think there's really any way around that right now, until we get `__mutable`/`__metadata` or give the type special treatment (the `__` prefix is there in case things like this become an issue). Is this currently a problem on any compilers?

src/core/experimental/refcount.d
thewilsonator: I think so, probably best to change it anyway.

src/core/experimental/refcount.d
thewilsonator: Sorry, where is `src/core/experimental/array.d`? somewhere in your fork of druntime?

src/core/experimental/refcount.d
TurkeyMan: > This makes this line compile with -preview=restrictiveshared 

No it doesn't... shared(int) can't be incremented.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,390 @@
...
+
+Source: $(DRUNTIMESRC core/experimental/refcount.d)
+*/
+module core.experimental.refcount;
+
+///
+unittest
+{
+    auto rc = __RefCount.make();
+    assert(rc.isUnique);
```
TurkeyMan: You talk about atomic rc, but then there's `isUnique`? They're incompatible.

src/core/experimental/refcount.d
TurkeyMan: I have a whole lot of problems with this conversation. It seems like there are mistakes everywhere.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,390 @@
...
+
+Important: The `__RefCount` member must be initialized through a call to its
+constructor before being used.
+*/
+struct __RefCount
+{
+    import core.atomic : atomicOp;
+
+    alias CounterType = uint;
+    private shared(CounterType)* rc = null;
```
TurkeyMan: No offense personally, but this is the most disappointing implementation of reference counting that is possible to write. We can do better than this, and if we can't, then we have serious language issues.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,390 @@
...
+    /*
+    Perform `rc op val` operation. Always use atomics as the counter is shared.
+
+    Returns:
+         The result of the operation.
+    */
+    @nogc nothrow pure @trusted scope
+    private shared(CounterType) rcOp(this Q, string op)(CounterType val) const
+    {
+        return cast(shared(CounterType)) (atomicOp!op(*(cast(shared(CounterType)*) rc), val));
```
TurkeyMan: How does this work? `rc` is `const`, yet you mutate it... D's `const` is transitive, and D doesn't have `mutable` or anything like it, and the optimiser is free to do whatever it likes with the knowledge a value is `const`.
I don't think this line can go in the core runtime...?

src/core/experimental/refcount.d
TurkeyMan: Perhaps `*cast(CounterType*)support = 1`?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,390 @@
...
+            addRef();
+        }
+    };
+
+    /**
+    Copy constructs a mutable `__RefCount` from a mutable reference, `rhs`.
+    This increases the reference count.
+    */
+    @nogc nothrow pure @safe scope
+    this(return scope ref typeof(this) rhs)
```
TurkeyMan: Why `typeof(this)` and not `__RefCount`?

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,390 @@
...
+        // We are required to always use a shared support as the result of a `pure`
+        // function is implicitly convertible to `immutable`.
+
+        import core.memory : pureMalloc;
+        shared(CounterType)* support = cast(shared(CounterType)*) pureMalloc(CounterType.sizeof);
+        *support = 1; // Start with 1 to avoid calling an `addRef` from 0 to 1
+        rc = cast(typeof(rc)) support;
+    }
+
+    private enum copyCtorIncRef = q{
```
TurkeyMan: Please don't do this, it's not debug-able.

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,390 @@
...
+        rc = rhs.rc;
+        assert(rc == rhs.rc);
```
TurkeyMan: What kind of failure is this assert trying to catch?

TurkeyMan:
An interesting observation to me, is that this seems like a whole lot more code (maybe 3-4 times as much code) as the same functionality would be in C++, and that's even considering that we're using those ugly text mixins everywhere to reduce the code could... I think it's possible that D is still lacking necessary foundation for this :/

src/core/experimental/refcount.d
```diff
@@ -0,0 +1,390 @@
...
+
+    Returns:
+         `true` if this reference count is unique; `false` if this `__RefCount`
+         object is uninitialized or there are multiple references to it.
+
+    Complexity:
+         $(BIGOH 1).
+    */
+    pure nothrow @safe @nogc scope
+    bool isUnique(this Q)() const
```
TurkeyMan: This function must be removed. I can't imagine anything you could do with this that's not a data race.

src/core/experimental/refcount.d
TurkeyMan: Also, where's the memory order argument? It looks like it would execute this seq-cst (assuming that's the default), which is wrong.

src/core/experimental/refcount.d
TurkeyMan: Do we have reason to believe the language will definitely be changed to allow `__mutable`? Can we commit to something based on a future-truth that never happens? :/

TurkeyMan: Like, i'm skeptical `__mutable` will ever appear; if `__mutable` is introduced, then any optimisation opportunity that `const` offers must immediately be explicitly denied by the spec... there's been tension on that issue for decades.

src/core/experimental/refcount.d
wilzbach: It's idiomatic D to not repeat the struct name and reference it via typeof.

src/core/experimental/refcount.d
wilzbach: What would be the alternative?

src/core/experimental/refcount.d
TurkeyMan: Where's that written? I've seen Walter recommending against `typeof(this)` used in this way on numerous occasions.

wilzbach: In Phobos source code.

src/core/experimental/refcount.d
TurkeyMan: Does a mixin template work here? If not, just repeat the code; its only 3 lines.

src/core/experimental/refcount.d
TurkeyMan: Not a good place to find great D code ;)

src/core/experimental/refcount.d
lesderid: It does use `inout`?

UplinkCoder:
I can't review this properly right now. Please hold off on merging until I had a proper look. Which I'll signal by approving this pr.

lesderid:
Yes, this shouldn't be merged until we've been able to address the issues Manu pointed out.

(Having a feature to turn PRs into draft PRs would be very useful for cases like this, so it could be helpful to give that suggestion some support here: https://github.community/t5/H/F/m-p/19107)

src/core/experimental/refcount.d
TurkeyMan: Oh yeah, sorry I didn't see it in the sea of attributes. I was expecting it attributed to the context.
Can a constructor be `inout`? Wouldn't that cut the number of ctors down to 2; one `inout` and one `inout shared`?

src/core/experimental/refcount.d
TurkeyMan: We should really have `cast(ref CounterType)lvalue` to perform lvalue casting...

src/core/experimental/refcount.d
lesderid: @TurkeyMan Any insights you might have about this particular issue would be very much appreciated.

TurkeyMan: I haven't had time to think on it sorry, as funny as it might seem to some, RC is not actually an urgent issue for me as it has been for the last 10 years >_<  
This is very important, but I need to devote 100% attention to `shared` right now.

TurkeyMan: My instinct is that RC is so important, that `immutable`->`shared` conversion causing us an issue here is satisfactory grounds to re-consider `immutable`->`shared` conversion rules... the problem might not be here, maybe it's there. (I don't know, I haven't thought on it at all, that's just my feeling)

lesderid: I've had the same thought. I think it's definitely worth looking into.

lesderid:
We're trying out a possible alternative, a shared pointer type: https://github.com/dlang/druntime/pull/2690.

src/core/experimental/refcount.d
edi33416: Removing the call to `isInitialized` will remove the data race at the cost of generating a `segfault` for an invalid (uninitialized) `__RefCount`ed instance.

This function provides a simple interface towards the user to help him decide if he needs to deallocate the internal state of his data structure.
An "alternative" to this would be to expose the internal `delRef` to the user, which could break / invalidate the automatic reference counting. The user should never be able to explicitly modify the RC, but only "view" the value.

I propose that we bite the bullet and take the segfault.

src/core/experimental/refcount.d
edi33416: > Can we add a few more test to increase the coverage of this module and ensure it works as expected?

A bunch of tests were deleted with [this commit](https://github.com/dlang/druntime/pull/2679/commits/38d8ecf5248f4863fc2e94f8dea286d57830b400) 

> Maybe even one with multiple threads to ensure the atomic ops work too.

There was one which got deleted as well.

src/core/experimental/refcount.d
TurkeyMan: If the RC is atomic, then none of those things you say can exist. Everything you just said is a data race.
If RC is thread-local (as it should be in D), then we have options.

anon17:
You can have local counter, if you share immutable data in a shared container that would provide local rc wrappers for threads. When counter reaches zero you can see if it came from a shared container. [See also this](https://wiki.dlang.org/FatPointer).

PR #2680, opened 2019-07-16
Replace `in` with `const scope` in `core.stdc`

JinShil:
## About This PR
Followup to https://github.com/dlang/druntime/pull/2676 and https://github.com/dlang/phobos/pull/7110

This is the first of several PRs I intend to submit, breaking up https://github.com/dlang/druntime/pull/2677 into multiple PRs.

This PR only addresses C language bindings in `core.stdc`.  Since C does not have any concept of scoped pointers, I think it is arguably a mistake for these to be `in`.

## Background

### DIFF
--- a/src/core/stdc/fenv.d
+++ b/src/core/stdc/fenv.d
@@ -875,7 +875,7 @@ int feholdexcept(fenv_t* envp);
 ///
 int fegetexceptflag(fexcept_t* flagp, int excepts);
 ///
-int fesetexceptflag(in fexcept_t* flagp, int excepts);
+int fesetexceptflag(const scope fexcept_t* flagp, int excepts);
 
 ///
 int fegetround();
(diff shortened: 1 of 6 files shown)


thewilsonator:
> I think it is arguably a mistake for these to be in.

But none of those function leak the pointers given, which is the whole point of scope.

JinShil:
> But none of those function leak the pointers given, which is the whole point of scope.

Hopefully they don't, but C has no way to express such a thing and no way to statically enforce it, so why would we assume that that's the case unless we were to manually audit the C code?

thewilsonator:
> but C has no way to express such a thing

The C standard.

(From memory the only function that is obviously not `scope` is `getopt`, as it sets `optarg`.)

> no way to statically enforce it.  so why would we assume that that's the case unless we were to manually audit the C code?

Because this isn't some random C library downloaded from some dark and dodgy corner of the internet, this is the C standard library we're talking about. Its only dependance in on the hardware and the OS system calls. If its buggy, malicious, swapped out with LD_PRELOAD, then a) there is nothing we can do to protect against that, and b) you have much more pressing concerns, namely that your C standard lib is dud. This is a pragmatism vs purism issue, and I'm firmly for the pragmatism. The threat model of having a malfunctioning C standard library is not one we should be concerning ourselves with.

JinShil:
Updated to use `const scope` now.

PR #2683, opened 2019-07-17
Add PowerPC64 support for FreeBSD

lenoil98:
This initial set of changes are to add powerpc64 support for FreeBSD.  Additional work needs to be done to fix test build and failures.


thewilsonator:
You have tabs everywhere, that cause build kite to fail.

thewilsonator:
cc @redstar

lenoil98:
I'll remove tabs and make adjustments.

thewilsonator:
Thanks.

thewilsonator:
If there isn't already (I'm pretty sure there is, there definitely is for X86) please add

```
version (PPC)
    version = PPC_Any;
else version (PPC64)
    version = PPC_Any;
```

and then use `version (PPC_Any)` to deduplicate the definitions. If you could also please clean up the commit history that would be great too.

src/core/sys/posix/ucontext.d
```diff
@@ -1032,6 +1032,46 @@ else version (FreeBSD)
             ulong[8]        mc_spare;
         }
     }
+    else version (PPC)
+    {
+        alias uint   __register_t;
```
thewilsonator: This is the only thing that is different between PPC and PPC64, please alias it to `size_t` and merge the two into a `version (PPC_Any)`.

PR #2684, opened 2019-07-17
Replace `in` with `const scope` in `core.sys.darwin`

JinShil:
## About This PR
Followup to 

This PR only addresses C language bindings in `core.sys.darwin`.

## Background

### DIFF
--- a/src/core/sys/darwin/dlfcn.d
+++ b/src/core/sys/darwin/dlfcn.d
@@ -33,7 +33,7 @@ struct Dl_info
     void*        dli_saddr;
 }
 
-int dladdr(in void* addr, Dl_info* info);
+int dladdr(const scope void* addr, Dl_info* info);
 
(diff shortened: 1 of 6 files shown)


thewilsonator:
```
src/rt/sections_osx_x86_64.d(83): Error: function `core.sys.darwin.mach.dyld._dyld_register_func_for_add_image(extern (C) void function(scope const(mach_header*) mh, long vmaddr_slide))` is not callable using argument types `(extern (C) void function(const(mach_header*) h, long slide))`
src/rt/sections_osx_x86_64.d(83):        cannot pass argument `& sections_osx_onAddImage` of type `extern (C) void function(const(mach_header*) h, long slide)` to parameter `extern (C) void function(scope const(mach_header*) mh, long vmaddr_slide)`

```

JinShil:
All Darwin platforms are passing now.  This should be good to go.

PR #2685, opened 2019-07-17
Replace `in` with `const scope` for object.d rt/dmain2.d and rt/typeinfo

JinShil:
## About This PR
Followup to 

This PR only addresses object.d, rt/dmain2.d, and files in the rt/typeinfo folder.

## Background

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -969,10 +969,10 @@ class TypeInfo
     }
 
     /// Compares two instances for equality.
-    bool equals(in void* p1, in void* p2) const { return p1 == p2; }
+    bool equals(const scope void* p1, const scope void* p2) const { return p1 == p2; }
 
     /// Compares two instances for &lt;, ==, or &gt;.
-    int compare(in void* p1, in void* p2) const { return _xopCmp(p1, p2); }
+    int compare(const scope void* p1, const scope void* p2) const { return _xopCmp(p1, p2); }
 
     /// Returns size of the type.
     @property size_t tsize() nothrow pure const @safe @nogc { return 0; }
@@ -1044,8 +1044,8 @@ class TypeInfo_Enum : TypeInfo
     }
 
     override size_t getHash(scope const void* p) const { return base.getHash(p); }
-    override bool equals(in void* p1, in void* p2) const { return base.equals(p1, p2); }
-    override int compare(in void* p1, in void* p2) const { return base.compare(p1, p2); }
+    override bool equals(const scope void* p1, const scope void* p2) const { return base.equals(p1, p2); }
+    override int compare(const scope void* p1, const scope void* p2) const { return base.compare(p1, p2); }
     override @property size_t tsize() nothrow pure const { return base.tsize; }
     override void swap(void* p1, void* p2) const { return base.swap(p1, p2); }
 
@@ -1097,12 +1097,12 @@ class TypeInfo_Pointer : TypeInfo
         return addr ^ (addr >> 4);
     }
 
-    override bool equals(in void* p1, in void* p2) const
+    override bool equals(const scope void* p1, const scope void* p2) const
     {
         return *cast(void**)p1 == *cast(void**)p2;
     }
 
-    override int compare(in void* p1, in void* p2) const
+    override int compare(const scope void* p1, const scope void* p2) const
     {
         if (*cast(void**)p1 < *cast(void**)p2)
             return -1;
@@ -1153,7 +1153,7 @@ class TypeInfo_Array : TypeInfo
         return getArrayHash(value, a.ptr, a.length);
     }
 
-    override bool equals(in void* p1, in void* p2) const
+    override bool equals(const scope void* p1, const scope void* p2) const
     {
         void[] a1 = *cast(void[]*)p1;
         void[] a2 = *cast(void[]*)p2;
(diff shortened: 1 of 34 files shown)


src/object.d
```diff
@@ -2738,13 +2738,13 @@ extern (C)
...
-    private void* _aaGetX(AA* paa, const TypeInfo_AssociativeArray ti, in size_t valsz, in void* pkey, out bool found) pure nothrow;
-    // inout(void)* _aaGetRvalueX(inout AA aa, in TypeInfo keyti, in size_t valsz, in void* pkey);
-    inout(void[]) _aaValues(inout AA aa, in size_t keysz, in size_t valsz, const TypeInfo tiValueArray) pure nothrow;
-    inout(void[]) _aaKeys(inout AA aa, in size_t keysz, const TypeInfo tiKeyArray) pure nothrow;
-    void* _aaRehash(AA* paa, in TypeInfo keyti) pure nothrow;
+    // size_t _aaLen(const scope AA aa) pure nothrow @nogc;
+    private void* _aaGetY(AA* paa, const TypeInfo_AssociativeArray ti, const size_t valsz, const scope void* pkey) pure nothrow;
+    private void* _aaGetX(AA* paa, const TypeInfo_AssociativeArray ti, const size_t valsz, const scope void* pkey, out bool found) pure nothrow;
+    // inout(void)* _aaGetRvalueX(inout AA aa, const scope TypeInfo keyti, const size_t valsz, const scope void* pkey);
+    inout(void[]) _aaValues(inout AA aa, const size_t keysz, const size_t valsz, const TypeInfo tiValueArray) pure nothrow;
```
JinShil: I only used `const` for `size_t` as I believe `scope` is not applicable to value types.

thewilsonator: Yes, it is ignored to make generic code agnostic to valueness.

thewilsonator:
```
src/object.d(2515): Error: none of the overloads of `toString` are callable using argument types `(void)`, candidates are:
src/object.d(2512):        `object.Throwable.toString()`
src/object.d(2525):        `object.Throwable.toString(scope void delegate(scope const(char[])) sink)`
```

src/object.d
```diff
@@ -2065,8 +2065,8 @@ class TypeInfo_Const : TypeInfo
     }
 
     override size_t getHash(scope const void *p) const { return base.getHash(p); }
-    override bool equals(in void *p1, in void *p2) const { return base.equals(p1, p2); }
-    override int compare(in void *p1, in void *p2) const { return base.compare(p1, p2); }
+    override bool equals(const scope void *p1, const scope void *p2) const { return base.equals(p1, p2); }
+    override int compare(const scope void *p1, const scope void *p2) const { return base.compare(p1, p2); }
```
ZombineDev: `base` can be the `TypeInfo` for a `class` or `struct` which implements `opEquals`|`opCompare` without using the `scope` attribute.

src/object.d
```diff
@@ -1908,10 +1908,10 @@ class TypeInfo_Struct : TypeInfo
...
     @safe pure nothrow
     {
-        size_t   function(in void*)           xtoHash;
-        bool     function(in void*, in void*) xopEquals;
-        int      function(in void*, in void*) xopCmp;
-        string   function(in void*)           xtoString;
+        size_t   function(const scope void*)           xtoHash;
+        bool     function(const scope void*, const scope void*) xopEquals;
+        int      function(const scope void*, const scope void*) xopCmp;
+        string   function(const scope void*)           xtoString;
```
ZombineDev: Pedantically speaking, I don't think those function parameters can be neither `scope` nor `const` :(

Only when the TypeInfo generation is fully template-driven we can have true guarantees about `const`/`scope`/insert your favorite function attribute here correctness.

src/object.d
JinShil: What do you recommend?

src/object.d
JinShil: What do you recommend?

ZombineDev: There are two options:
1. Use honest function signature and change those parameters to `void*` as in reality even the `const`-ness can't be guaranteed. This means that `TypeInfo_Struct.{equals,compare}` must also take `void*` parameters which is likely a breaking change.

2. Continue with the deceit and deal with problems somehow later down the line, when we can be technically honest.

@jmdavis @schveiguy @atilaneves What's your point of view?

atilaneves: Not sure. I agree we can't make them `scope` or `const` though.

JinShil: @ZombineDev Why can't these be `const`?  They shouldn't be modifying `p`.

JinShil:
A decision about `in` has finally been made:  https://github.com/dlang/dmd/pull/10382#issuecomment-527081559

src/object.d
ZombineDev: @JinShil  Because they are function pointers to arbitrary user code. The pointer itself may not be modified (as it will be copied), but the object pointed by `p` may be. And it can also escape `p` without any issues.

Here's the sort of techniques that we need to use in order to be type safe:
https://gist.github.com/ZombineDev/ac19e93dff4f7b251be0c7af8e8d150d#file-proper_type_info-d-L62-L73

Live demo: https://run.dlang.io/gist/ZombineDev/ac19e93dff4f7b251be0c7af8e8d150d?compiler=dmd&args=-dip1000

PR #2687, opened 2019-07-19
core.experimental.memory module with D replacement for memcpy added

baziotis:
This is a D replacement for the C Standard Library `memcpy`. It provides 2 interfaces:
- The "D" one, which takes, one of `ref`, dynamic arrays or static arrays.
- The "C" one, via the name `Dmemcpy` (for disambiguation with the libc), which takes the classic
`(void*, void*, size_t)`.

Since this function uses SIMD, I found convenient to PR a simple SIMD library, which I hope will be extended by the community. Its main purpose is to provide cross-compilation intrinsics for D.

Ideally, the `core.experimental.memory.simd` would be a different PR, but without it basically a PR for `memcpy` can't happen.

The change in `core.internal.traits` is the same done here: https://github.com/dlang/druntime/pull/2662
I have copied everything necessary from `std.traits` to have `isArray` available.

There are a couple things I'm not sure about but are minor, so better wait for feedback.

### DIFF
--- a/src/core/experimental/memory/memcpy.d
+++ b/src/core/experimental/memory/memcpy.d
@@ -15,6 +15,13 @@ import core.internal.traits : isArray;
  * Handle Static Types
  * N.B.: No need for more sophisticated code for static types. The compiler
  * knows better how to handle them in every target case.
+ *
+ * Params:
+ *  dst = Reference to memory destination to copy bytes to.
+ *  src = Reference to memory source to copy bytes from.
+ *
+ * Returns:
+ *  Nothing.
  */
 pragma(inline, true)
 void memcpy(T)(ref T dst, ref const T src)
@@ -24,7 +31,14 @@ if (!isArray!T)
 }
 
 /**
- * Dynamic Arrays
+ * Handle Dynamic Types
+ *
+ * Params:
+ *  dst = Reference to destination dynamic array to copy bytes to.
+ *  src = Reference to source dynamic array to copy bytes from.
+ *
+ * Returns:
+ *  Nothing.
  */
 void memcpy(T)(ref T[] dst, ref const T[] src)
 {
@@ -38,7 +52,15 @@ void memcpy(T)(ref T[] dst, ref const T[] src)
 }
 
 /**
- * Static Arrays
+ * Handle Dynamic Types
+ *
+ * Params:
+ *  len = Length of the static arrays.
+ *  dst = Reference to destination static array to copy bytes to.
+ *  src = Reference to source static array to copy bytes from.
+ *
+ * Returns:
+ *  Nothing.
  */
 void memcpy(T, size_t len)(ref T[len] dst, ref const T[len] src)
 {
@@ -231,8 +253,16 @@ static if (useSIMD)
 import core.simd : float4;
 
 /**
- * Handle dynamic sizes. `d` and `s` must not overlap.
+ * Handle Dynamic Types
  * N.B.: While Dmemcpy's interface is C-like, it returns _nothing_.
+ *
+ * Params:
+ *  len = Length of the static arrays.
+ *  dst = Pointer to memory destination to copy bytes to.
+ *  src = Pointer to memory source to copy bytes from.
+ *
+ * Returns:
+ *  Nothing.
  */
 void Dmemcpy(void* d, const(void)* s, size_t n)
 {
--- a/src/core/experimental/memory/memcpy.d
+++ b/src/core/experimental/memory/memcpy.d
@@ -250,9 +250,9 @@ import core.simd : float4;
  * N.B.: While Dmemcpy's interface is C-like, it returns _nothing_.
  *
  * Params:
- *  len = Length of the static arrays.
- *  dst = Pointer to memory destination to copy bytes to.
- *  src = Pointer to memory source to copy bytes from.
+ *  d = Pointer to memory destination to copy bytes to.
+ *  s = Pointer to memory source to copy bytes from.
+ *  n = Number of bytes to copy.
  */
 void Dmemcpy(void* d, const(void)* s, size_t n)
 {
--- a/src/core/experimental/memory/memcpy.d
+++ b/src/core/experimental/memory/memcpy.d
@@ -23,7 +23,7 @@ import core.internal.traits : isArray;
  *  src = Reference to memory source to copy bytes from.
  */
 pragma(inline, true)
-void memcpy(T)(ref T dst, ref const T src)
+void memcpy(T)(ref T dst, ref const T src) nothrow @nogc
 if (!isArray!T)
 {
     dst = src;
@@ -36,7 +36,7 @@ if (!isArray!T)
  *  dst = Reference to destination dynamic array to copy bytes to.
  *  src = Reference to source dynamic array to copy bytes from.
  */
-void memcpy(T)(ref T[] dst, ref const T[] src)
+void memcpy(T)(ref T[] dst, ref const T[] src) nothrow @nogc
 {
     assert(dst.length == src.length);
     void* d = cast(void*) dst.ptr;
@@ -47,15 +47,8 @@ void memcpy(T)(ref T[] dst, ref const T[] src)
     Dmemcpy(d, s, n);
 }
 
-/**
- * Handle Dynamic Types
- *
- * Params:
- *  len = Length of the static arrays.
- *  dst = Reference to destination static array to copy bytes to.
- *  src = Reference to source static array to copy bytes from.
- */
-void memcpy(T, size_t len)(ref T[len] dst, ref const T[len] src)
+/// Ditto
+void memcpy(T, size_t len)(ref T[len] dst, ref const T[len] src) nothrow @nogc
 {
     T[] d = dst[0 .. $];
     const T[] s = src[0 .. $];
@@ -67,7 +60,7 @@ void memcpy(T, size_t len)(ref T[len] dst, ref const T[len] src)
 
 /* Basic features tests
  */
-unittest
+nothrow @nogc unittest
 {
     real a = 1.2;
     real b;
@@ -75,7 +68,7 @@ unittest
     assert(b == 1.2);
 }
 ///
-unittest
+nothrow @nogc unittest
 {
     const float[3] a = [1.2, 3.4, 5.8];
     float[3] b;
@@ -87,7 +80,7 @@ unittest
 
 /* More sophisticated test suite
  */
-version (unittest)
+nothrow @nogc unittest
 {
     /* Handy struct
      */
@@ -95,7 +88,98 @@ version (unittest)
     {
         ubyte[Size] x;
     }
-    void tests()
+
+    pragma(inline, false)
+    void initStatic(T)(T *v) nothrow @nogc
+    {
+        auto m = (cast(ubyte*) v)[0 .. T.sizeof];
+        foreach (i; 0..m.length)
+        {
+            m[i] = cast(ubyte) i;
+        }
+    }
+
+    pragma(inline, false)
+    void verifyStaticType(T)(const T *a, const T *b) nothrow @nogc
+    {
+        const ubyte* aa = cast(const ubyte*) a;
+        const ubyte* bb = cast(const ubyte*) b;
+        // Note: `real` is an exceptional case,
+        // in that it behaves differently across compilers
+        // because it's not a power of 2 (its size is 10 for x86)
+        // and thus padding is added (to reach 16). But, the padding bytes
+        // are not considered (by the compiler) in a move (for instance).
+        // So, Dmemcpy, for static types, is *dst = *src. And the compiler
+        // might output `fld` followed by `fstp` instruction. Those intructions
+        // operate on extended floating point values (whose size is 10). And so,
+        // the padding bytes are not copied to dest.
+        static if (is(T == real))
+        {
+            enum n = 10;
+        }
+        else
+        {
+            enum n = T.sizeof;
+        }
+        foreach (i; 0..n)
+        {
+            assert(aa[i] == bb[i]);
+        }
+    }
+
+    pragma(inline, false)
+    void testStaticType(T)() nothrow @nogc
+    {
+        T d, s;
+        initStatic!(T)(&d);
+        initStatic!(T)(&s);
+        memcpy(d, s);
+        verifyStaticType(&d, &s);
+    }
+
+    pragma(inline, false)
+    void init(T)(ref T[] v) nothrow @nogc
+    {
+        foreach (i; 0..v.length)
+        {
+            v[i] = cast(ubyte) i;
+        }
+    }
+
+    pragma(inline, false)
+    void verifyArray(size_t j, const ref ubyte[] a, const ref ubyte[80000] b) nothrow @nogc
+    {
+        foreach (i; 0..a.length)
+        {
+            assert(a[i] == b[i]);
+        }
+    }
+
+    pragma(inline, false)
+    void testDynamicArray(size_t n)() nothrow @nogc
+    {
+        ubyte[80000] buf1;
+        ubyte[80000] buf2;
+        enum alignments = 32;
+        foreach (i; 0..alignments)
+        {
+            ubyte[] p = buf1[i..i+n];
+            ubyte[] q;
+            // Relatively aligned
+            q = buf2[0..n];
+            // Use a copy for the cases of overlap.
+            ubyte[80000] copy;
+            init(q);
+            init(p);
+            foreach (k; 0..p.length)
+            {
+                copy[k] = p[k];
+            }
+            memcpy(q, p);
+            verifyArray(i, q, copy);
+        }
+    }
+    void tests() nothrow @nogc
     {
         testStaticType!(byte);
         testStaticType!(ubyte);
@@ -134,107 +218,12 @@ version (unittest)
         testStaticType!(S!32768);
         testStaticType!(S!65536);
     }
-    pragma(inline, false)
-    void initStatic(T)(T *v)
-    {
-        auto m = (cast(ubyte*) v)[0 .. T.sizeof];
-        for (int i = 0; i < m.length; i++)
-        {
-            m[i] = cast(ubyte) i;
-        }
-    }
-    pragma(inline, false)
-    void verifyStaticType(T)(const T *a, const T *b)
-    {
-        const ubyte* aa = (cast(const ubyte*) a);
-        const ubyte* bb = (cast(const ubyte*) b);
-        // NOTE(stefanos): `real` is an exceptional case,
-        // in that it behaves differently across compilers
-        // because it's not a power of 2 (its size is 10 for x86)
-        // and thus padding is added (to reach 16). But, the padding bytes
-        // are not considered (by the compiler) in a move (for instance).
-        // So, Dmemcpy, for static types, is *dst = *src. And the compiler
-        // might output `fld` followed by `fstp` instruction. Those intructions
-        // operate on extended floating point values (whose size is 10). And so,
-        // the padding bytes are not copied to dest.
-        static if (is(T == real))
-        {
-            enum n = 10;
-        }
-        else
-        {
-            enum n = T.sizeof;
-        }
-        for (size_t i = 0; i < n; i++)
-        {
-            assert(aa[i] == bb[i]);
-        }
-    }
-    pragma(inline, false)
-    void testStaticType(T)()
-    {
-        T d, s;
-        initStatic!(T)(&d);
-        initStatic!(T)(&s);
-        memcpy(d, s);
-        verifyStaticType(&d, &s);
-    }
-    pragma(inline, false)
-    void init(T)(ref T[] v)
-    {
-        for (int i = 0; i < v.length; i++)
-        {
-            v[i] = cast(ubyte) i;
-        }
-    }
-    pragma(inline, false)
-    void verifyArray(size_t j, const ref ubyte[] a, const ref ubyte[80000] b)
-    {
-        //assert(a.length == b.length);
-        for (int i = 0; i < a.length; i++)
-        {
-            assert(a[i] == b[i]);
-        }
-    }
-    pragma(inline, false)
-    void testDynamicArray(size_t n)()
-    {
-        ubyte[80000] buf1;
-        ubyte[80000] buf2;
-        enum alignments = 32;
-        foreach (i; 0..alignments)
-        {
-            ubyte[] p = buf1[i..i+n];
-            ubyte[] q;
-            // Relatively aligned
-            q = buf2[0..n];
-            // Use a copy for the cases of overlap.
-            ubyte[80000] copy;
-            pragma(inline, false);
-            init(q);
-            pragma(inline, false);
-            init(p);
-            for (size_t k = 0; k != p.length; ++k)
-            {
-                copy[k] = p[k];
-            }
-            pragma(inline, false);
-            memcpy(q, p);
-            pragma(inline, false);
-            verifyArray(i, q, copy);
-        }
-    }
-}
-///
-unittest
-{
+
     tests();
 }
 
-/* Implementation
- */
-
-import core.experimental.memory.simd;
+import core.experimental.memory.simd : useSIMD, load16fSSE, store16fSSE, lstore128fpSSE,
+                                       lstore128fSSE, lstore64fSSE, lstore32fSSE;
 
 /*
  * Dynamic implementation
@@ -247,14 +236,14 @@ import core.simd : float4;
 
 /**
  * Handle Dynamic Types
- * N.B.: While Dmemcpy's interface is C-like, it returns _nothing_.
+ * N.B.: While Dmemcpy's interface is C-like, it returns void.
  *
  * Params:
  *  d = Pointer to memory destination to copy bytes to.
  *  s = Pointer to memory source to copy bytes from.
  *  n = Number of bytes to copy.
  */
-void Dmemcpy(void* d, const(void)* s, size_t n)
+void Dmemcpy(void* d, const(void)* s, size_t n) nothrow @nogc
 {
     if (n <= 128)
     {
@@ -268,7 +257,7 @@ void Dmemcpy(void* d, const(void)* s, size_t n)
 
 /* Handle dynamic sizes <= 128. `d` and `s` must not overlap.
  */
-private void Dmemcpy_small(void* d, const(void)* s, size_t n)
+private void Dmemcpy_small(void* d, const(void)* s, size_t n) nothrow @nogc
 {
     if (n < 16) {
         if (n & 0x01)
@@ -303,9 +292,6 @@ private void Dmemcpy_small(void* d, const(void)* s, size_t n)
         store16fSSE(d-16+n, xmm1);
         return;
     }
-    // NOTE(stefanos): I'm writing using load/storeUnaligned() but you possibly can
-    // achieve greater performance using naked ASM. Be careful that you should either use
-    // only D or only naked ASM.
     if (n <= 64)
     {
         float4 xmm0 = load16fSSE(s);
@@ -329,14 +315,14 @@ private void Dmemcpy_small(void* d, const(void)* s, size_t n)
 
 /* Handle dynamic sizes > 128. `d` and `s` must not overlap.
  */
-private void Dmemcpy_large(void* d, const(void)* s, size_t n)
+private void Dmemcpy_large(void* d, const(void)* s, size_t n) nothrow @nogc
 {
-    // NOTE(stefanos): Alternative - Reach 64-byte
+    // NOTE: Alternative - Reach 64-byte
     // (cache-line) alignment and use rep movsb
     // Good for bigger sizes and only for Intel.
 
     // Align destination (write) to 32-byte boundary
-    // NOTE(stefanos): We're using SSE, which needs 16-byte alignment.
+    // Note: We're using SSE, which needs 16-byte alignment.
     // But actually, 32-byte alignment was quite faster (probably because
     // the loads / stores are faster and there's the bottleneck).
     uint rem = cast(ulong) d & 15;
@@ -352,14 +338,14 @@ private void Dmemcpy_large(void* d, const(void)* s, size_t n)
     {
         return
         "
-        while (n >= 128)
-        {
-            // Aligned stores / writes
-            " ~ prefetchChoice ~ "(d, s);
-            d += 128;
-            s += 128;
-            n -= 128;
-        }
+            while (n >= 128)
+            {
+                // Aligned stores / writes
+                " ~ prefetchChoice ~ "(d, s);
+                d += 128;
+                s += 128;
+                n -= 128;
+            }
         ";
     }
 
@@ -386,11 +372,11 @@ else
 
 /* Non-SIMD version
  */
-void Dmemcpy(void* d, const(void)* s, size_t n)
+void Dmemcpy(void* d, const(void)* s, size_t n) nothrow @nogc
 {
     ubyte* dst = cast(ubyte*) d;
     const(ubyte)* src = cast(const(ubyte)*) s;
-    for (size_t i = 0; i != n; ++i)
+    foreach (i; 0..n)
     {
         *dst = *src;
         dst++;
--- a/src/core/experimental/memory/memcpy.d
+++ b/src/core/experimental/memory/memcpy.d
@@ -4,7 +4,9 @@
  * One taking static types, one dynamic arrays and one static arrays.
  * Also, there is available a C-like interface, the Dmemcpy() (which is named Dmemcpy
  * for disambiguation with the C memcpy() a _similar_ interface) that
- * is the classic (void*, void*, size_t) interface. _But_ Dmemcpy returns nothing.
+ * is the classic (void*, void*, size_t) interface.
+ * N.B.: Both the memcpy() here and Dmemcpy() return nothing, contrary to the C Standard
+ * Library version.
  * Source: $(DRUNTIMESRC core/experimental/memory/memcpy.d)
  */
 module core.experimental.memory.memcpy;
@@ -19,9 +21,6 @@ import core.internal.traits : isArray;
  * Params:
  *  dst = Reference to memory destination to copy bytes to.
  *  src = Reference to memory source to copy bytes from.
- *
- * Returns:
- *  Nothing.
  */
 pragma(inline, true)
 void memcpy(T)(ref T dst, ref const T src)
@@ -36,9 +35,6 @@ if (!isArray!T)
  * Params:
  *  dst = Reference to destination dynamic array to copy bytes to.
  *  src = Reference to source dynamic array to copy bytes from.
- *
- * Returns:
- *  Nothing.
  */
 void memcpy(T)(ref T[] dst, ref const T[] src)
 {
@@ -58,9 +54,6 @@ void memcpy(T)(ref T[] dst, ref const T[] src)
  *  len = Length of the static arrays.
  *  dst = Reference to destination static array to copy bytes to.
  *  src = Reference to source static array to copy bytes from.
- *
- * Returns:
- *  Nothing.
  */
 void memcpy(T, size_t len)(ref T[len] dst, ref const T[len] src)
 {
@@ -260,9 +253,6 @@ import core.simd : float4;
  *  len = Length of the static arrays.
  *  dst = Pointer to memory destination to copy bytes to.
  *  src = Pointer to memory source to copy bytes from.
- *
- * Returns:
- *  Nothing.
  */
 void Dmemcpy(void* d, const(void)* s, size_t n)
 {
--- a/src/core/experimental/memory/memcpy.d
+++ b/src/core/experimental/memory/memcpy.d
@@ -258,7 +258,7 @@ unittest
 /* Implementation
  */
 
-import simd;
+import core.experimental.memory.simd;
 
 /*
  * Dynamic implementation
--- a/src/core/experimental/memory/memcpy.d
+++ b/src/core/experimental/memory/memcpy.d
@@ -1,0 +1,447 @@
+/**
+ * Pure D replacement of the C Standard Library memcpy().
+ * There is an idiomatic-D interface, memcpy(), which is split into 3 overloads.
+ * One taking static types, one dynamic arrays and one static arrays.
+ * Also, there is available a C-like interface, the Dmemcpy() (which is named Dmemcpy
+ * for disambiguation with the C memcpy() which has the exact same interface) that
+ * is the classic (void*, void*, size_t) interface.
+ * Source: $(DRUNTIMESRC core/experimental/memory/memcpy.d)
+ */
+module core.experimental.memory.memcpy;
+
+import core.internal.traits : isArray;
+
+/* Static Types
+ */
+// NOTE(stefanos): Previously, there was more sophisticated code
+// for static types. But the rationale of removing it is that
+// the compiler knows better how to optimize static types.
+pragma(inline, true)
+void memcpy(T)(ref T dst, ref const T src)
+if (!isArray!T)
+{
+    dst = src;
+}
+
+/* Dynamic Arrays
+ */
+void memcpy(T)(ref T[] dst, ref const T[] src)
+{
+    assert(dst.length == src.length);
+    void* d = cast(void*) dst.ptr;
+    const(void)* s = cast(const(void)*) src.ptr;
+    size_t n = dst.length * typeof(dst[0]).sizeof;
+    // Assume that there is no overlap.
+    pragma(inline, true);
+    Dmemcpy(d, s, n);
+}
+
+/* Static Arrays
+ */
+void memcpy(T, size_t len)(ref T[len] dst, ref const T[len] src)
+{
+    T[] d = dst[0 .. $];
+    const T[] s = src[0 .. $];
+    memcpy(d, s);
+}
+
+/** Tests
+  */
+
+/* Basic features tests
+ */
+unittest
+{
+    real a = 1.2;
+    real b;
+    memcpy(b, a);
+    assert(b == 1.2);
+}
+///
+unittest
+{
+    const float[3] a = [1.2, 3.4, 5.8];
+    float[3] b;
+    memcpy(b, a);
+    assert(b[0] == 1.2f);
+    assert(b[1] == 3.4f);
+    assert(b[2] == 5.8f);
+}
+
+/* More sophisticated test suite
+ */
+version (unittest)
+{
+
+    import std.random;
+    
+    /* Handy struct
+     */
+    struct S(size_t Size)
+    {
+        ubyte[Size] x;
+    }
+    
+    static string genTests()
+    {
+        import std.conv : text;
+        string res;
+        foreach (i; 1..100)
+        {
+            res ~=
+            "
+            testStaticType!(S!"~text(i)~");
+            testDynamicArray!("~text(i)~")();
+            ";
+        }
+        return res;
+    }
+    
+    void tests()
+    {
+        testStaticType!(byte);
+        testStaticType!(ubyte);
+        testStaticType!(short);
+        testStaticType!(ushort);
+        testStaticType!(int);
+        testStaticType!(uint);
+        testStaticType!(long);
+        testStaticType!(ulong);
+        testStaticType!(float);
+        testStaticType!(double);
+        testStaticType!(real);
+        mixin(genTests());
+        testStaticType!(S!3452);
+        testDynamicArray!(3452)();
+        testStaticType!(S!6598);
+        testDynamicArray!(6598);
+        testStaticType!(S!14928);
+        testDynamicArray!(14928);
+        testStaticType!(S!27891);
+        testDynamicArray!(27891);
+        testStaticType!(S!44032);
+        testStaticType!(S!55897);
+        testStaticType!(S!79394);
+    
+        testStaticType!(S!256);
+        testStaticType!(S!512);
+        testStaticType!(S!1024);
+        testStaticType!(S!2048);
+        testStaticType!(S!4096);
+        testStaticType!(S!8192);
+        testStaticType!(S!16384);
+        testStaticType!(S!32768);
+        testStaticType!(S!65536);
+    }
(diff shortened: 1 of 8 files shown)


src/core/experimental/memory/simd.d
```diff
@@ -0,0 +1,170 @@
+/**
```
lesderid: Should this be in `core.experimental.memory`? Wouldn't `core.experimental` make more sense?

lesderid:
> Ideally, the `core.experimental.memory.simd` would be a different PR, but without it basically a PR for `memcpy` can't happen.

Perhaps make two PRs but include the SIMD commits in the `memcpy` one too?

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,431 @@
...
+    float[3] b;
+    memcpy(b, a);
+    assert(b[0] == 1.2f);
+    assert(b[1] == 3.4f);
+    assert(b[2] == 5.8f);
+}
+
+/* More sophisticated test suite
+ */
+version (unittest)
```
lesderid: No Phobos in druntime.

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,431 @@
...
+{
+    import std.random;
```
lesderid: No Phobos in druntime.

baziotis: Hey, we saw it the same time. Just updated with non-random values since it doesn't really matter.

src/core/experimental/memory/simd.d
baziotis: It depends on how you look at it. It was proposed here: https://github.com/dlang/druntime/pull/2671#discussion_r305198634
Which makes sense if the SIMD operations are only memory operations (which they are currently). Eventually, I hope that this will become a bigger SIMD library and be moved out to `experimental` (or wherever) because D needs one that is able to cross-compile. At least I don't know of one and so I made this.

baziotis:
> Perhaps make two PRs but include the SIMD commits in the memcpy one too?

Yes, that's an alternative. But then if changes happen in this or the other PR, then the changes have to be made on both PRs. Ideally, first the PR on SIMD would be made, then this, but it was done to save time. It makes _some_ sense to be here, as the `internal.traits`, since besides the code, we see a use case. I'll leave it an open discussion for a little time, thanks for the feedback. :)

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,391 @@
...
+ *
+ */
+static if (useSIMD)
+{
+
+import core.simd : float4;
+
+/* Handle dynamic sizes. `d` and `s` must not overlap.
+ */
+void Dmemcpy(void* d, const(void)* s, size_t n)
```
thewilsonator: `private`

baziotis: That is public because of the description of this PR:

> This is a D replacement for the C Standard Library memcpy. It provides 2 interfaces:

> The "D" one, which takes, one of ref, dynamic arrays or static arrays.
> The "C" one, via the name Dmemcpy (for disambiguation with the libc), which takes the classic: `(void*, void*, size_t)`.

So, because the idea was to provide the C interface as well. Is this desired?
Otherwise, yes it should be `private`.

thewilsonator: Isn't that the C version then? I think all bases should be covered with static arrays and dynamic arrays, you can always slice a pointer to create a dynamic array from it and a length..

baziotis: Yes, it is the C version. It is named _D_`memcpy` for disambiguation with the libc `memcpy`.
We changed the `Dmemcpy` name to `memcpy` for the D interface, which provides 3 overloads, the ones you can see on top of the file here, named `memcpy`. I guess this was done with the purpose that they are overloads for the libc `memcpy`. They provide handling for `ref`s, static arrays and dynamic arrays.
But I thought it would be good to also provide a C interface but for the D version (aka the druntime version) named `Dmemcpy`, that is different from the libc.

thewilsonator: OK.

thewilsonator:
`make[1]: *** No rule to make target '/dev/shm/dtest/work/repo/dlang.org/web/phobos-prerelease/core_experimental_memory_memcpy.html', needed by 'doc'.  Stop.`

I think you need to add this to the doc makefile.

baziotis:
Isn't the doc makefile the `mak/DOCS`?

lesderid:
> Isn't the doc makefile the `mak/DOCS`?

Yes, but new (sub)directories have to be added in `posix.mak` too: https://github.com/dlang/druntime/blob/master/posix.mak#L157.

baziotis:
Thanks Les!

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+ *  src = Reference to source dynamic array to copy bytes from.
+ */
+void memcpy(T)(ref T[] dst, ref const T[] src)
+{
+    assert(dst.length == src.length);
+    void* d = cast(void*) dst.ptr;
+    const(void)* s = cast(const(void)*) src.ptr;
+    size_t n = dst.length * typeof(dst[0]).sizeof;
+    // Assume that there is no overlap.
+    pragma(inline, true);
```
TurkeyMan: What does this mean?

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+ *
+ * Params:
+ *  len = Length of the static arrays.
+ *  dst = Reference to destination static array to copy bytes to.
+ *  src = Reference to source static array to copy bytes from.
+ */
+void memcpy(T, size_t len)(ref T[len] dst, ref const T[len] src)
+{
+    T[] d = dst[0 .. $];
+    const T[] s = src[0 .. $];
```
TurkeyMan: Why write these on separate lines?

src/core/experimental/memory/memcpy.d
TurkeyMan: So, LDC/GDC would never use this function right... they're re-route to intrinsics?

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+else
+{
+
+/* Non-SIMD version
+ */
+void Dmemcpy(void* d, const(void)* s, size_t n)
+{
+    ubyte* dst = cast(ubyte*) d;
+    const(ubyte)* src = cast(const(ubyte)*) s;
+    for (size_t i = 0; i != n; ++i)
```
TurkeyMan: Why isn't this at least moving `ulong`'s?

src/core/experimental/memory/memcpy.d
baziotis: They will use the SIMD version. They will probably use their intrinsics for when SIMD is not available.

src/core/experimental/memory/memcpy.d
baziotis: What? It shows me 4 lines. Please point me to the line.

src/core/experimental/memory/memcpy.d
TurkeyMan: They should use their intrinsics unconditionally. There's no way we should take that knowledge away from the optimiser.

src/core/experimental/memory/memcpy.d
TurkeyMan: The one I tagged... `pragma(inline, true);`
That's a weird statement, what does it mean? What does it apply to?

src/core/experimental/memory/memcpy.d
baziotis: Your comment made me think, thanks.

I could use the modified GNU algorithm: https://www.embedded.com/print/4024961
but to me this only complicates the code without any real gain. The reason is that it only _tests_ if they're aligned and doesn't _reach_ alignment. To do a more serious 4-byte or 8-byte move, one has reach alignment etc. like this: https://github.com/dlang/druntime/pull/2662/files#diff-b2f13b9afdcf6d39739b2bf55035f8e2R166

The reason I did not do either of those is that the compiler can do this and other things itself. It should be able to understand that moving either 4 or 8 bytes at a time is faster, reach alignment, even add SIMD. For example
LDC: https://godbolt.org/z/2V1Wl8
GDC: https://godbolt.org/z/XXzLUQ (GDC needs -O3 to do the crazy stuff).

Here, the compiler does all of the above. It even added something like a switch.

But, here's the part I hadn't tested. If on LDC we pass the `-disable-loop-vectorization`
LDC: https://godbolt.org/z/WVVJlz
or we specify an architecture that does not support SIMD, like let's say `pentium2`.
LDC: https://godbolt.org/z/Q1vqOS
then it doesn't do 4-byte moves (It should be 4 bytes there since it is 32-bit compilation).

The same is true for GDC: https://godbolt.org/z/PK8YDQ

So, I don't know if it pays to _reach_ alignment but definetely I will use the GNU algorithm. It adds value without complicating the code much.

src/core/experimental/memory/memcpy.d
baziotis: `d` and `s`? How else could they be?

src/core/experimental/memory/memcpy.d
baziotis: As far as I know, `pragma(inline, true)` instructs the compiler to inline the function call that follows.

src/core/experimental/memory/memcpy.d
baziotis: That is something important that I have thought and that Johan Engelen pointed out as well: https://github.com/dlang/druntime/pull/2662#issuecomment-517742489

For one, to the best of my knowledge (meaning, by inspecting the assembly), ldc and gdc, when they reroute, they re-route to the c standard library. They just, as you said, know that it is a memory copy, with all the implications. But this is what we try to avoid here (meaning the libc).
I don't know what ldc and gdc do if libc is not available. Do they just use our function or they have some other intrinsic? If they have their intrinsic, what code is it? Should we use that or ours but somehow give the compiler info that this is a memory copy ?
And I don't know of a way to give ldc and gdc such info.

src/core/experimental/memory/memcpy.d
TurkeyMan: You shouldn't be depending on optimisations that DMD can't do.
And LDC/GDC should be using intrinsics exclusively, definitely not calling any of these functions. Taking that knowledge away from the optimiser may create some very lame copies in optimised code.

src/core/experimental/memory/memcpy.d
TurkeyMan: `memcpy(dst[0 .. $], src[0 .. $]);`

I only say because I stopped and re-read these lines carefully looking for the gotcha I assumed must exist to justify this long-written version. To me, writing it out this way makes it look like there's unexpected detail that should be distinctly stated... maybe that's just me though. I was looking for the substance in the noise, and there wasn't any.

src/core/experimental/memory/memcpy.d
TurkeyMan: as a stand-alone statement on its own hanging there? What does it apply to?

Can you show the doco where you read that? If that exists I'll be very surprised (I've been asking for it for many years).

src/core/experimental/memory/memcpy.d
TurkeyMan: Exactly; all the things he said. For my money, re-implementing memcpy and memset for LDC/GDC is a very bad idea. Just don't do it. Even if you test some cases and we don't see a disadvantage today, it might appear tomorrow, next thursday, in may 2022...
You should assume this whole thing only applies only to DMD, and then you need to carefully consider that compiler's optimisation capabilities.

src/core/experimental/memory/memcpy.d
baziotis: DMD was the initial focus. Then, it changed LDC and GDC so pretty much, what DMD does became irrelevant. Otherwise, the code would be way different. This is not a hypothesis. I spent one month writing DMD versions of these functions that had to reach the libc ones. They were completely different from the current ones due to the optimizations that DMD could not do.

Meaning, if LDC and GDC use their intrinsics, this project becomes irrelevant. But to use their intrinsics, the above question should be answered, which is what LDC and GDC do when libc is not available.

For me, the logical move is to inform the optimizer that e.g. our `memset` is supposed to do what `memset` does. Which enables us to write our own versions of these functions (aka libc independency) and not lose the optimization info.

src/core/experimental/memory/memcpy.d
baziotis: It has effect, but not the effect I said. My mistake: https://dlang.org/spec/pragma.html#inline
It inlines the function it is _enclosed_ by.

src/core/experimental/memory/memcpy.d
baziotis: I agree. And yes, there's no gotcha.

src/core/experimental/memory/memcpy.d
baziotis: As I said here: https://github.com/dlang/druntime/pull/2687#discussion_r310724318
The focus was initially DMD. Then that was considered a bad idea so we switched to LDC / GDC.
For me, depending on the goal, targetting DMD is not a very good idea. Unless we either want to write ASM or not reach libc.

Now, that seems a good idea if there's a problem with intrinsics if libc is not available. It's also a good idea if one plans to completely get rid of libc (this is not planned). But it also seems good if one wants to have their own version but somehow can instruct the compiler about what the function does.

src/core/experimental/memory/memcpy.d
TurkeyMan: > which is what LDC and GDC do when libc is not available.

There's a difference between a call to a function called `mempy` (which is hooked by an intrinsic), and a symbol called `memcpy` in a lib somewhere. You're writing the latter, but the former is what appears in the code. LLVM/GCC intrinsics don't always emit a call to a memcpy function, they can do anything they like. They *might* emit a call to a memcpy function, and if they do, it would be an advantage to understand in what context that can occur.

I don't know the answer here; I haven't looked into this, it might be that they just call the libc memcpy when they decide not to do anything special; or they might call into their own small runtime libraries? Perhaps their runtime libraries have a small suite of such functions optimised for different cases? I don't know, but anything's possible here.
All the compilers have a runtime library which is a lower-level library than libc. I'm not sure if those libraries contain memcpy/memset's for when the intrinsics bail out... or if they all just leave it to the libc implementation? If the intrinsic's fall-back functions are in the runtime library and not taken from libc, then we already have libc independence in those cases.

Anyway, I can see a world where it may be useful to have these functions for when the intrinsic does bail-out, but I think it's important to understand the implementation details of each compilers intrinsics to understand what they expect from these functions and the conditions when they're called.

You'll never be able to write optimised implementations for all platforms (or even x86, because it's a moving target); so I feel like the most interesting victory here is for DMD.

I'm sorry, I don't really know anything about this project you're working on, or the parameters for completion... but I think tackling memcpy/memset are probably lower priority when considering libc independence, and possibly the 2 most highly contentious items on the list.

src/core/experimental/memory/memcpy.d
TurkeyMan: Oh wow. I had no idea you could do that. That's very weird.

src/core/experimental/memory/memcpy.d
TurkeyMan: I don't understand of follow any of those propositions, but I'll assume that the discussion has been had elsewhere and makes sense.
I don't know what your goals are, I was just browsing the PR. At face value though, the motive feels weird... is there a reference to the goals and initial discussion? You mentioned that the goals changed, I'd kinda like to see how and why?

src/core/experimental/memory/memcpy.d
TurkeyMan: > Meaning, if LDC and GDC use their intrinsics, this project becomes irrelevant.

Just to be clear; we should assume they are, and do absolutely everything we must to make sure we don't interfere with that. The case for us to consider here is when the intrinsics decide not to do anything special and fall-back to a lib-call... what functions do they call when they bail out? Is it libc memcpy/set, or something else customised for their runtimes?

src/core/experimental/memory/memcpy.d
baziotis: It seems weird to me as well..

src/core/experimental/memory/memcpy.d
ibuclaw: Musl may be a good reference too.

https://github.com/bminor/musl/blob/master/src/string/memcpy.c
https://github.com/bminor/musl/blob/master/src/string/memset.c

src/core/experimental/memory/memcpy.d
baziotis: > I don't understand of follow any of those propositions, but I'll assume that the discussion has been had elsewhere and makes sense.

Could you be more specific? You seem to like DMD as a target, in which case I think at least one thing is true. Targetting DMD is a good idea if we either want to write ASM or we don't care for parity with libc.

Other than that, this specific discussion has not been done elsewhere and I too try to understand more and more things as this project goes forward and the oppositions are more and more. Like yours, Johan's, Iain's, Andrei's initially etc. Thank you for your time, I mean it.
But just have in your mind that:
I don't have a mentor or anyone interested in this project for that matter. So, I face the oppositions myself for a lot of things that I don't know and / or have experience with. Like the compiler intrinsics. And most of the oppositions are not about something I _wrote_, like "hey, why is this mul there?". Yes, we can discuss this, I'm supposed to have knowledge about the code I've written. But for the project _goals_.. I don't know, I don't think it's the best idea to leave to a student to justify why the project is needed. And btw, I'm not too very sure about this project.

Anyway, the initial goals were these: https://forum.dlang.org/thread/izaufklyvmktnwsrmhci@forum.dlang.org

Which basically is DMD and close performance to libc. To have that, one has to write ASM, so most of the code was inline ASM.
You can see some very early oppositions I had about the project goals very early on: https://forum.dlang.org/post/juipzgghcwcengyackgt@forum.dlang.org
And that was with much less knowledge about what can go wrong.

Then, after the project was left without a mentor, the goals changed to:
https://forum.dlang.org/post/qbicnduehcmxeaqwervf@forum.dlang.org

And _all_ the code was re-written to D.

src/core/experimental/memory/memcpy.d
baziotis: @TurkeyMan I agree with everything you wrote. You see, when this project started (or when it moved to LDC / GDC), I didn't know any of these proeblems. It was Johan's comment (about a week ago) that made me realize "Oh, not using the intrinsics is like a really bad thing". I'm not sure how to move forward..

baziotis: @ibuclaw Yes, they're good reference, thanks. It was one of the early codes I was referred to. Generally though, Agner's implementations are supposed to be the fastest around (although that's not exactly true compared to the libc. This could be a very long discussion). He's considered a master in this field.

Other than that, unfortunately, I don't think that the main problems are with the code but with the goals.

src/core/experimental/memory/memcpy.d
TurkeyMan: Actaully... I don't really care much at all about DMD; it's mostly a toy. But what I'm saying is that DMD could absolutely *benefit* from good implementations of these functions, whereas LDC/GDC will likely rather suffer.

> I don't have a mentor or anyone interested in this project for that matter. So, I face the oppositions myself for a lot of things that I don't know and / or have experience with.

You're diving into some really complex and heavy shit, so I give you credit for working on this challenging stuff!

> Anyway, the initial goals were these: [...]

Thanks, I didn't actually know the details of your project ;)

Okay, I think I can see something about your process here. Sorry for jumping in and messing the waters so late. I was just reviewing some PR's and saw what looked like very interesting PR's... but as you have determined, also very complex ;)

Who was the mentor initially? Have you had any offers for a replacement?

To be clear; I'm not against your current work. The goals of your project are interesting and useful (although libc independence is not interesting to me personally, there's a libc everywhere I've ever worked), but if you want to do it, then it's important to understand the context it must be plugged into.
The goals seem to have changed, which is never fun, but at very least you're learning a whole lot about a very complicated ecosystem. Progress here was never going to be easy ;)

The intrinsics do call out to library functions when the compilers determine that's what they want to do... so supplying a source for those symbols other than libc may be good for D to have, but producing optimal code is unlikely, and frankly, not worth the time invested. You should probably define initial success as a 'compatibility grade' implementation, which would be a dignified outcome for users that specifically want a stand-alone environment. No need to waste a heap of time benchmarking and micro-tuning on those terms.

If you take it from that perspective, then as-portable-as-possible implementations with no asm or any nonsense makes sense, and what you could look into, is defining these functions in druntime but specified with weak-linkage, then the libc functions will override them if the user links a libc.

But in addition to that, I think it's very important to review the GCC and LDC intrinsics, understand how they are used by the optimiser, and how the compilers will emit library calls when they choose to. Do they call the standard memcpy/set functions in the libc, or are there other implementations integrated into the toolchain's runtime libraries designed to work together with the intrinsics?
It's probably the case that they just call out to libc, but I think it's important to understand how and when that works when considering your implementations here.

I think @ibuclaw and @JohanEngelen will have much better advice than me on this whole matter; they are experts on this.

baziotis: > Actaully... I don't really care much at all about DMD; it's mostly a toy. But what I'm saying is that DMD could absolutely benefit from good implementations of these functions, whereas LDC/GDC will likely rather suffer.

Oh, I agree, me neither. From a personal point of view, it was really fun optimizing for DMD exactly because it doesn't have the GCC / LLVM optimizer. But for D, I don't think it's worth. So yes, I see that DMD could definitely benefit from this, but I don't think we should invest time.

> You're diving into some really complex and heavy shit, so I give you credit for working on this challenging stuff!

Thank you! You know, unfortunately the challenging stuff turned out not to be what I expected / told it would be (aka reaching libc in performance) rather turned out to be things about I don't have experience or knowledge (aka justifying the goals, specifics about compiler intrinsics etc.).

> Thanks, I didn't actually know the details of your project ;)
>
> Okay, I think I can see something about your process here. Sorry for jumping in and messing the waters so late. I was just reviewing some PR's and saw what looked like very interesting PR's... but as you have determined, also very complex ;)

No thank you for that. And all of you. You know these PRs might never be merged, but at least I'll know why and I will have learned something.

> Who was the mentor initially?

Mostly Mike Franklin.

> Have you had any offers for a replacement?

No. Although I thank @thewilsonator for jumping in and helping me move these PRs forward (and for other things, but anyway).

> To be clear; I'm not against your current work. The goals of your project are interesting and useful (although libc independence is not interesting to me personally, there's a libc everywhere I've ever worked), but if you want to do it, then it's important to understand the context it must be plugged into.
The goals seem to have changed, which is never fun, but at very least you're learning a whole lot about a very complicated ecosystem. Progress here was never going to be easy ;)

Yes, indeed! Well, again, the tough part is that the problems are things I don't know. And things that require experience. I was unsure for the goals of this project in the first 5 days or so. Then, even when these goals changed, there were unexpected problems that I just didn't know of. Problems with the very _goals_ not the implementation.

> The intrinsics do call out to library functions when the compilers determine that's what they want to do...

Yes that seems reasonable.

> so supplying a source for those symbols other than libc may be good for D to have, but producing optimal code is unlikely, and frankly, not worth the time invested.
You should probably define initial success as a 'compatibility grade' implementation, which would be a dignified outcome for users that specifically want a stand-alone environment. No need to waste a heap of time benchmarking and micro-tuning on those terms.

I might have missed this. You mean, an implementation but not e.g. with a constraint to be close to libc. Just an implementation that is D, it is general and it works.

> If you take it from that perspective, then as-portable-as-possible implementations with no asm or any nonsense makes sense, and what you could look into, is defining these functions in druntime but specified with weak-linkage, then the libc functions will override them if the user links a libc.

Oh, yeah, I see you mention a similar thing. Weak-linkage is a term? I understand the meaning from the context, but I've never heard of it.

> But in addition to that, I think it's very important to review the GCC and LDC intrinsics, understand how they are used by the optimiser, and how the compilers will emit library calls when they choose to. Do they call the standard memcpy/set functions in the libc, or are there other implementations integrated into the toolchain's runtime libraries designed to work together with the intrinsics? Is that decision back-end specific, could it change for an odd platform?
It's probably the case that they always just call out to libc, but I think it's important to understand how and when that works when considering your implementations here.

Intrinsics had been mentioned initially. But the focus was DMD then, so nobody really cared.
Even when it moved to LDC / GDC, I didn't know that the optimizers _know_ about the behavior of e.g. memcpy (i.e. after the call, this will have a copy of this). Then Johan happened :p. Thanks Johan.
Obviously, with that information, and considering that anyone who wants performance in x86_64 (this project's performance is focused on x86_64) will probably have libc, then this project becomes kind of irrelevant.

> I think @ibuclaw and @JohanEngelen will have much better advice than me on this whole matter; they are experts on this.

Their advice is welcome!

wilzbach: > this project becomes kind of irrelevant.

Don't forget web assembly and similar targets.

baziotis: The thing is that the whole code and time invested is optimizing for x86 (and x86_64). For WASM and other platforms, the compiler will branch to the naive versions, which are like _really_ naive. Meaning, a musl-like implementation (which doesn't have SIMD) would be more appropriate if such platforms were the main focus.

lesderid: > Weak-linkage is a term? I understand the meaning from the context, but I've never heard of it.

Yes, see e.g. [ldc.attributes.weak](https://wiki.dlang.org/LDC-specific_language_changes#.40.28ldc.attributes.weak.29), but note that D sadly doesn't have a standard attribute for it and that it's not supported on all platforms.

baziotis: Ah, good to know!

TurkeyMan: > it's not supported on all platforms.

Where is it not supported?

D just doesn't have a standard for it... but that's very possible to fix.

lesderid: > Where is it not supported?

AFAIK MSVC (or whatever object format that Windows uses, COFF/PE?) doesn't support weak linkage, but I might be wrong.

> D just doesn't have a standard for it... but that's very possible to fix.

I agree, would be great to have this.

TurkeyMan: > but I might be wrong.

Yeah. I'm pretty sure weak linkage is available everywhere, and it might be particularly useful in this project.

I wonder if there's any reason that we don't have a weak attribute already? I think I've asked for it in the past and it was denied because reasons... I wonder what they were :/

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+        testStaticType!(S!8192);
+        testStaticType!(S!16384);
+        testStaticType!(S!32768);
+        testStaticType!(S!65536);
+    }
+    pragma(inline, false)
+    void initStatic(T)(T *v)
+    {
+        auto m = (cast(ubyte*) v)[0 .. T.sizeof];
+        for (int i = 0; i < m.length; i++)
```
wilzbach: Foreach

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+        {
+            m[i] = cast(ubyte) i;
+        }
+    }
+    pragma(inline, false)
+    void verifyStaticType(T)(const T *a, const T *b)
+    {
+        const ubyte* aa = (cast(const ubyte*) a);
+        const ubyte* bb = (cast(const ubyte*) b);
```
wilzbach: Useless parentheses

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+        // the padding bytes are not copied to dest.
+        static if (is(T == real))
+        {
+            enum n = 10;
+        }
+        else
+        {
+            enum n = T.sizeof;
+        }
+        for (size_t i = 0; i < n; i++)
```
wilzbach: F

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+        T d, s;
+        initStatic!(T)(&d);
+        initStatic!(T)(&s);
+        memcpy(d, s);
+        verifyStaticType(&d, &s);
+    }
+    pragma(inline, false)
+    void init(T)(ref T[] v)
+    {
+        for (int i = 0; i < v.length; i++)
```
wilzbach: Foreach

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+ */
+static if (useSIMD)
+{
+
+import core.simd : float4;
+
+/**
+ * Handle Dynamic Types
+ * N.B.: While Dmemcpy's interface is C-like, it returns _nothing_.
+ *
```
wilzbach: Mention void.

src/core/experimental/memory/memcpy.d
wilzbach: Function should still be private.

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+        store16fSSE(d, load16fSSE(s));
+        s += 16 - rem;
+        d += 16 - rem;
+        n -= 16 - rem;
+    }
+
+    static string loop(string prefetchChoice)()
+    {
+        return
+        "
```
wilzbach: Use token strings

src/core/experimental/memory/simd.d
```diff
@@ -0,0 +1,170 @@
...
+ * An currently small experimental SIMD library for D, with the main purpose
+ * to be easily used by DMD, GDC and LDC users.
+ * There is an idiomatic-D interface, memcpy(), which is split into 3 overloads.
+ * One taking static types, one dynamic arrays and one static arrays.
+ * Also, there is available a C-like interface, the Dmemcpy() (which is named Dmemcpy
```
wilzbach: This isn't part of this package

src/core/experimental/memory/simd.d
```diff
@@ -0,0 +1,170 @@
+/**
+ * An currently small experimental SIMD library for D, with the main purpose
+ * to be easily used by DMD, GDC and LDC users.
```
wilzbach: Cross-compiler + platform

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+///
+unittest
+{
+    tests();
+}
+
+/* Implementation
+ */
+
+import core.experimental.memory.simd;
```
wilzbach: Selective imports

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+ * Pure D replacement of the C Standard Library memcpy().
+ * There is an idiomatic-D interface, memcpy(), which is split into 3 overloads.
+ * One taking static types, one dynamic arrays and one static arrays.
+ * Also, there is available a C-like interface, the Dmemcpy() (which is named Dmemcpy
+ * for disambiguation with the C memcpy() a _similar_ interface) that
+ * is the classic (void*, void*, size_t) interface.
+ * N.B.: Both the memcpy() here and Dmemcpy() return nothing, contrary to the C Standard
+ * Library version.
+ * Source: $(DRUNTIMESRC core/experimental/memory/memcpy.d)
+ */
```
wilzbach: DStyle is without leading stars.

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+module core.experimental.memory.memcpy;
+
+import core.internal.traits : isArray;
+
+/**
+ * Handle Static Types
+ * N.B.: No need for more sophisticated code for static types. The compiler
```
wilzbach: Use full sentences and no abbreviation. Have a look at std.algorithm for good examples on how to write good doc headers.

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+    dst = src;
+}
+
+/**
+ * Handle Dynamic Types
+ *
+ * Params:
+ *  dst = Reference to destination dynamic array to copy bytes to.
+ *  src = Reference to source dynamic array to copy bytes from.
+ */
```
wilzbach: Why not join them with ditto?

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+ * N.B.: Both the memcpy() here and Dmemcpy() return nothing, contrary to the C Standard
+ * Library version.
+ * Source: $(DRUNTIMESRC core/experimental/memory/memcpy.d)
+ */
+module core.experimental.memory.memcpy;
+
+import core.internal.traits : isArray;
+
+/**
+ * Handle Static Types
```
wilzbach: This distinction isn't important to the user.

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+ * N.B.: No need for more sophisticated code for static types. The compiler
+ * knows better how to handle them in every target case.
+ *
+ * Params:
+ *  dst = Reference to memory destination to copy bytes to.
+ *  src = Reference to memory source to copy bytes from.
+ */
+pragma(inline, true)
+void memcpy(T)(ref T dst, ref const T src)
+if (!isArray!T)
```
wilzbach: Why not use this as the user facing API and delegate with static if? Reduces the exposed interfaces -> less confusion.

src/core/experimental/memory/memcpy.d
wilzbach: I agree that writing it in one line expresses the intent clearer.

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+{
+    dst = src;
+}
+
```
wilzbach: Needs public example

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+ */
+unittest
+{
+    real a = 1.2;
+    real b;
+    memcpy(b, a);
+    assert(b == 1.2);
+}
+///
+unittest
```
wilzbach: Attributes + what's the motivation behind making this public, but above not?

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+        testStaticType!(ubyte);
+        testStaticType!(short);
+        testStaticType!(ushort);
+        testStaticType!(int);
+        testStaticType!(uint);
+        testStaticType!(long);
+        testStaticType!(ulong);
+        testStaticType!(float);
+        testStaticType!(double);
+        testStaticType!(real);
```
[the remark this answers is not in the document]
baziotis: You mean with `AliasSeq` probably. The static foreach is a problem for GDC though. Do you want 2 versions?

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+        testStaticType!(S!79394);
+        testStaticType!(S!256);
+        testStaticType!(S!512);
+        testStaticType!(S!1024);
+        testStaticType!(S!2048);
+        testStaticType!(S!4096);
+        testStaticType!(S!8192);
+        testStaticType!(S!16384);
+        testStaticType!(S!32768);
+        testStaticType!(S!65536);
```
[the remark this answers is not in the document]
baziotis: Again, the GDC thing.

src/core/experimental/memory/memcpy.d
baziotis: I don't really get where to use the one and where the other.
For example, with leading stars: https://github.com/dlang/phobos/blob/master/std/algorithm/searching.d#L277
Without leading stars:
https://github.com/dlang/phobos/blob/master/std/algorithm/searching.d#L213

src/core/experimental/memory/memcpy.d
baziotis: I think it was a note and I was referred to a N.B. So, `Note`?

src/core/experimental/memory/memcpy.d
baziotis: You mean it should be just a `//` that is not part of the doc?

src/core/experimental/memory/memcpy.d
baziotis: Because for me it's more confusing with one interface. I'll change it.

src/core/experimental/memory/memcpy.d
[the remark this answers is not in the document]
ibuclaw: For testing, that makes sense.  However, this module won't be backported to dmd-cxx, so you don't need to worry about backwards compatibility for older versions of D.

src/core/experimental/memory/memcpy.d
baziotis: I can't find an easy way to interpolate the `prefetchChoice` with token strings.

baziotis: I can't somehow answer on the comment above, so:

```
Function should still be private or have a clear confusion free name
```
If it is private, then the whole discussion about providing a C-like has no meaning.
So, if you don't want a C-like interface then I make it private otherwise, what would be a confusion-free name?

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+    }
+    pragma(inline, false)
+    void initStatic(T)(T *v)
+    {
+        auto m = (cast(ubyte*) v)[0 .. T.sizeof];
```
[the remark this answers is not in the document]
baziotis: I think this is not useless. We want first to cast it to a `ubyte*`. Even if precedence handles that, I think it's more clear to parenthesize it.

src/core/experimental/memory/memcpy.d
```diff
@@ -0,0 +1,401 @@
...
+        testStaticType!(S!512);
+        testStaticType!(S!1024);
+        testStaticType!(S!2048);
+        testStaticType!(S!4096);
+        testStaticType!(S!8192);
+        testStaticType!(S!16384);
+        testStaticType!(S!32768);
+        testStaticType!(S!65536);
+    }
+    pragma(inline, false)
```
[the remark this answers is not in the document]
baziotis: Why?

src/core/experimental/memory/memcpy.d
baziotis: Ah, easy. I'm still not familiar with the Dstyle.

src/core/experimental/memory/memcpy.d
baziotis: I have a problem that I think I had when I had tried to join in one interface all of them. Basically, somehow the arrays get confused. e.g. if I have this code:

```
void memcpy(T)(ref T dst, ref const T src) nothrow @nogc
{
    static if (!isArray!T)
    {
        dst = src;
    }
    else
    {
        void memcpyArray(T)(ref T[] dst, ref const T[] src) nothrow @nogc
        {
            assert(dst.length == src.length);
            void* d = cast(void*) dst.ptr;
            const(void)* s = cast(const(void)*) src.ptr;
            size_t n = dst.length * typeof(dst[0]).sizeof;
            // Assume that there is no overlap.
            Dmemcpy(d, s, n);
        }

        static if (isDynamicArray!T)
        {
            memcpyArray(dst, src);
        }
        else
        {
            memcpyArray(dst[0 .. $], src[0 .. $]);
        }
    }
}
```

baziotis:
I have the one problem I mentioned in the comment about trying to merge all of the `memcpy` in one interface.

Also, this:
```
    memcpy(dst[0 .. $], src[0 .. $]);
```
produces a segfault.

Other than those:
> move traits in a separate PR into druntime, but with all tests.

src/core/experimental/memory/memcpy.d
lesderid: If you're not sure, you can check the D Style document[1], which should always be followed for new code: "Documentation comments should not have leading stars on each line."

A lot of the code in druntime (and sadly also some in Phobos) is old and doesn't follow the style very closely or at all.

[1] https://dlang.org/dstyle.html

baziotis: Thanks, I'm trying to follow but things escape here and there.

src/core/experimental/memory/memcpy.d
lesderid: > Yeah. I'm pretty sure weak linkage is available everywhere

The accepted answer and the most up-voted answer on StackOverflow[1] suggest that at least the semantics are different with MSVC. Other compilers/linkers might implement workarounds though, I don't have access to a Windows machine to test on.

[1] https://stackoverflow.com/questions/2290587/gcc-style-weak-linking-in-visual-studio?answertab=votes#tab-top

src/core/experimental/memory/memcpy.d
wilzbach: > can't find an easy way to interpolate the prefetchChoice with token strings.

```
q{ ...} ~ ...
```

src/core/experimental/memory/memcpy.d
wilzbach: Yes there's lots of old code here and in Phobos that was never changed, but at least new code should follow the guidelines.

src/core/experimental/memory/memcpy.d
wilzbach: No I meant that as you have both overloads defined, this is an implementation detail. Though as said on another point there's a case to simplify the interface even further, but it definitely shouldn't expose such details to the user.

src/core/experimental/memory/memcpy.d
baziotis: Yes ok

src/core/experimental/memory/memcpy.d
baziotis: > this is an implementation detail
Yes, so the `//` wouldn't be the right choice i.e. not be part of the doc?
On the merging of the interfaces there were problems that I could not solve: https://github.com/dlang/druntime/pull/2687#discussion_r311488051

JohanEngelen:
A note about compiler intrinsics: LLVM has intrinsics for memset and memcpy, but one doesn't need to use them to get optimizations going. E.g. in code like this, no intrinsics are used:
```
extern(C) void *memset(void *ptr, int x, ulong n);
void foo(int* ptr) {
    memset(ptr, 0, 2);
}
```
yet the optimizer recognizes `memset` (by C name and signature) and is assuming it correctly knows all about its semantics. This is an aggressive optimization (recognizing by name) that some people may frown upon.  See: https://d.godbolt.org/z/H-yeMW . It is similar to the optimizer recognizing a code pattern and turning that into a `memset` call, again assuming it correctly knows the semantics of the memset provided at link time. As far as I know, there is currently no way to tell the optimizer that a function has the same semantics as `memset`, other than naming that function exactly like that.
There are a host of functions that LLVM recognizes, with elaborate logic related to it, see e.g. https://github.com/llvm/llvm-project/blob/7f7ef0208b57a4d253cd8b07053460f40ad7cbc8/llvm/lib/Analysis/TargetLibraryInfo.cpp#L64-L67

(For the interested, the same works for allocation functions: see here how Rust adds to its LLVM fork the names of its own stdlib functions to the list of recognized memory functions: https://github.com/rust-lang/llvm-project/commit/1f4c17d166789784297c3f408e277f3728f11228 . Although for that, we can use the `allocSize` attribute https://github.com/ldc-developers/druntime/blob/ldc/src/ldc/attributes.d#L16-L56)

baziotis:
Thanks for the info Johan!

As I said in a previous comment, when I started the implementation for LDC, GDC, I didn't know that:
a) What `memset`, `memcpy` etc. do is info known to the compiler that is used during optimization.
b) That info is very important to have.

As I said in the comments above, sadly (very sadly, I have written this code one for DMD and one for LDC, GDC) that info pretty much makes this project irrelevant.
To be more specific:
* Not using the intrinsics is quite bad. And since, as you said, there's no way to specifically inform the compiler that some code has the same semantics as e.g. `memset`, it's not worth it. And writing code on the assumption that the compiler will assume the same semantics on either the code or naming, seems bad to me.
* The one question that I had was what these intrinsics forward to when there is no libc. But the thing is, this code is x86_64 specific. And probably no-one who cares about performance won't have the libc (AFAIK).

So, this project is worth it if:
* We're targetting DMD. Although I don't think that the initial goal of reaching libc (and so, my then approach) was good because a) it requires inline ASM b) keeping up to date with x86_64 performance is difficult. But DMD could
benefit from performant, but generic D implementations.
* Even if we're not targetting DMD, a generic implementation would be candidate. An implementation that doesn't try to optimize for x86_64 because it probably won't ever be used there (as the libc will probably be available). That could be something like the musl implementation which is generic C.

So, well I don't know.. I'm open to your thoughts.

Btw
> A note about compiler intrinsics: LLVM has intrinsics for memset and memcpy, but one doesn't need to use them to get optimizations going. E.g. in code like this, no intrinsics are used:
> 
> yet the optimizer recognizes `memset` (by C name and signature) and is assuming it correctly knows all about its semantics. This is an aggressive optimization (recognizing by name) that some people may frown upon. See: https://d.godbolt.org/z/H-yeMW . It is similar to the optimizer recognizing a code pattern and turning that into a `memset` call, again assuming it correctly knows the semantics of the memset provided at link time.
> There are a host of functions that LLVM recognizes, with elaborate logic related to it, see e.g. https://github.com/llvm/llvm-project/blob/7f7ef0208b57a4d253cd8b07053460f40ad7cbc8/llvm/lib/Analysis/TargetLibraryInfo.cpp#L64-L67

Wow!

> (For the interested, the same works for allocation functions: see here how Rust adds to its LLVM fork the names of its own stdlib functions to the list of recognized memory functions: [rust-lang/llvm-project@1f4c17d](https://github.com/rust-lang/llvm-project/commit/1f4c17d166789784297c3f408e277f3728f11228) . Although for that, we can use the `allocSize` attribute https://github.com/ldc-developers/druntime/blob/ldc/src/ldc/attributes.d#L16-L56)

Interersting

baziotis:
Closed due to not being useful - more info [here](http://users.uoa.gr/~sdi1600105/google-summer-of-code.html).

PR #2688, opened 2019-07-19
Use MONOTONIC clock with condition if available

kubo39:


src/core/sync/condition.d
```diff
@@ -74,7 +74,7 @@ class Condition
      * Throws:
      *  SyncError on error.
      */
-    this( Mutex m ) nothrow @safe
+    this( Mutex m ) nothrow @trusted
```
Geod24: The whole function shouldn't be `@trusted`. Instead, the part that needs to escape `@safety` should use `@trusted` delegates.

kubo39:
ping?

PR #2689, opened 2019-07-21
core.thread refactor: osthread module

baziotis:
This is the start of an effort to refactor `core.thread`. With Nicholas Wilson's help, this takes the form of a separate folder that will split `thread` to modules. Currently, the plan is to split it to:
- `osthread` : Low-level, OS-dependent code for threads.
- `fiber` : The interface for fibers.
- `context` : The stack context needed for threads and fibers.

`osthread` seems to be the easiest to start from so that the review is easier.


thewilsonator:
You should start by moving `core/thread.d` into `core/thread/package.d` and then move things out of there into other files, making `private` symbols `package(core.thread)` where needed. 

It will be much easier to keep track of things that way.

baziotis:
Ok, I just didn't get the last phrase:
> making private symbols package(core.thread) where needed.

thewilsonator:
some symbols in core.thread are private. 

When you move symbol around that require access to those symbols, as they are now in different modules they will give an error if used. 

making those symbols `package(core.thread)` will resolve that problem and still have the symbols not be visible outside of `core.thread`

baziotis:
Ok, thanks. I didn't know that there's the `package` specifier. I'll learn about it and update the PR.

ZombineDev:
@baziotis thanks for working on this! I will try to review your later today or tomorrow.

baziotis:
@ZombineDev Don't mention it. It might seem like a big PR but it's actually not. I just moved the OS-dependent stuff from `core/thread.d` to `/thread/osthread.d`. I'll continue with Nicholas' proposal.

Geod24:
> I'll continue with Nicholas' proposal.

@thewilsonator 's proposal was to *start* with something else. And IMO it makes a lot of sense.
This PR is currently only adding code, nothing is removed yet, so I assume this is in essence unused code ? Which also means untested. It's harder to review when the changes are actually disconnected, like adding in a PR and removing in another (unless there's a good reason to do so, e.g. when dealing with cross-repository dependencies).

baziotis:
@Geod24 To be honest, I'm not sure what is your point. This PR is a draft because it is not supposed to be merged as it is. It basically asks "have I moved the correct amount of things (i.e. the OS stuff) from `thread.d` to `osthread`". Practically, it adds no code.

thewilsonator:
> have I moved the correct amount of things

In this case, no, because there should be a (roughly) equal amount of code removed.

src/core/thread/package.d
```diff
@@ -0,0 +1,16 @@
...
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors:   Sean Kelly, Walter Bright, Alex Rønne Petersen, Martin Nowak
+ * Source:    $(DRUNTIMESRC core/thread/package.d)
+ */
+
+module core.thread;
+
+public import core.time;  // Should be removed
```
thewilsonator: can these be `package(core.thread) import`s?

baziotis: Let me post the PM:

The initial `thread.d` had `public import core.time`. So, I removed `public` but then `spinlock` (iirc) used something of `core.time` which was coming from `thread`.
It was kind of a mess so I removed the `public` from `core.thread` and imported on `spinlock`.
As you can see, I get an `undefined reference` error which I don't understand. Thinking that it might have been a problem of with removing the `public import` from `core.thread`, I put it again on `package.d`.

Tl;dr, the `public import core.time` should not be needed. The rest yes, they can.

baziotis:
Can I have some help on why the tests are failing? I can't seem to find a way to fix them.

aG0aep6G:
> Can I have some help on why the tests are failing? I can't seem to find a way to fix them.

I think you forgot to add thread/package.d to the make files.

baziotis:
> I think you forgot to add thread/package.d to the make files.
I thought it wasn't needed, thanks for the response. I added in all but DOCS. Unfortunately, it doesn't seem to fix the error.

aG0aep6G:
> Unfortunately, it doesn't seem to fix the error.

The error is different now. [You've got a merge conflict in mak/DOCS.](https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3712300&isPull=true) You need to rebase.

baziotis:
> The error is different now. You've got a merge conflict in mak/DOCS. You need to rebase.

Yes, I had this error in the previous commit as well. But, besides this, there's also the previous error for example in circleci.

thewilsonator:
[This test](https://github.com/dlang/dmd/blob/master/test/runnable/test15779.d) seems to fail on windows consistently.

src/core/thread/osthread.d
```diff
@@ -1579,8 +1648,27 @@ private:
         sm_this = t;
     }
 
+package(core.thread):
```
thewilsonator: One of the reasons for wanting to do this refactoring in the first place was to have this be pubic so that Weka can use it in their own fiber code. Please remove this line.

thewilsonator: and move it outside of `class Thread`

baziotis:
> [This test](https://github.com/dlang/dmd/blob/master/test/runnable/test15779.d) seems to fail on windows consistently.

Thanks, I currently can't build either dmd or druntime on Windows. With either Visual Studio or the makefiles. Building those from source can have a lot of complications.

baziotis:
I still don't understand the error much. It seems it has something to do with [the constructor of Fiber](https://github.com/dlang/druntime/blob/e36084b951eb94b382bd0a6b253372e51f5c4c4d/src/core/thread/fiber.d#L546) or [call()](https://github.com/dlang/druntime/blob/e36084b951eb94b382bd0a6b253372e51f5c4c4d/src/core/thread/fiber.d#L637).

The second case AFAIK is harder since it goes quite deep in the calls. However for the first, could you please verify that the module construction works across the package ? Meaning [this](https://github.com/dlang/druntime/blob/e36084b951eb94b382bd0a6b253372e51f5c4c4d/src/core/thread/osthread.d#L76).

baziotis:
Finally that works. :sweat_smile: 
@thewilsonator Regarding `Context`. I'm not sure whether moving `Context` outside is better done in this PR. It might not be enough to just move the struct and also I liked the version in your diff where there was a separate file for all the context stuff.

thewilsonator:
I take it this is no linger a draft PR?

baziotis:
> I take it this is no longer a draft PR?

Yes, let me make it an actual PR.

baziotis:
Oh crap, I thought `Ready for review` makes it an actual PR, not requesting actual reviews.

lesderid:
> Edit: Should this somehow be undone or is it supposed to be that way?

Supposed to be that way.

thewilsonator:
There are a couple of things I'd like to see changed, but theres no point keeping thins waiting any further.

baziotis:
> There are a couple of things I'd like to see changed, but theres no point keeping thins waiting any further.

You can tell me. :)
As this is a PR in a series, trying to break the changes in reasonable PRs.
To the best of my knowledge, I have done the things that were discussed in this one.

thewilsonator:
https://github.com/dlang/druntime/pull/2689#discussion_r312720591
Fair enough.
Also fair enough.

For the next one split out the context stuff into `thread/context.d` like in the diff I gave you. (it has some typos and naming inconsistencies but I'm sure you'll discover that when you try to compile it ;) )

baziotis:
> [#2689 (comment)](https://github.com/dlang/druntime/pull/2689#discussion_r312720591)
> Fair enough.
> Also fair enough.
> 

Yes as said here: https://github.com/dlang/druntime/pull/2689#issuecomment-520237162
I think it makes this easier to do in a separate PR.

> For the next one split out the context stuff into `thread/context.d` like in the diff I gave you. (it has some typos and naming inconsistencies but I'm sure you'll discover that when you try to compile it ;) )

Haha yes I don't compile it. As you can see from this PR there are a lot of subtle things
in these changes so I take your idea and try to figure out how to make the implementation work.

thewilsonator:
Indeed, I look forward to it.

rainers:
As one of the tests failing in https://dev.azure.com/dlanguage/dmd/_build/results?buildId=4590 and other PRs is runnable/test15779.d which checks exceptions in fibers, I suspect it has been introduced by this PR. Any idea what might be causing this?

Apart from the auto tester, druntime just runs a partial win64 test on Appveyor. @wilzbach Can we easily run the azure tests from dmd for druntime, too?

thewilsonator:
No, but I note that there are a [couple](https://github.com/dlang/druntime/blob/master/src/core/thread/fiber.d#L1721) of [unittests](https://github.com/dlang/druntime/blob/master/src/core/thread/fiber.d#L1761) related to exceptions and fibers that are explicitly disabled for Win32. Maybe a subtle change has caused it to hit that bug?

rainers:
Indeed, the disabled unittests look very similar. I have failed to reproduce the issues reported by the azure CI locally, will try the unittests tonight.

thewilsonator:
Of note, that test enables stack stomping.

baziotis:
Follow-up: https://github.com/dlang/druntime/pull/2821

PR #2690, opened 2019-07-22
Add __rcptr!T (possible alternative to #2679)

lesderid:
This PR adds a shared pointer type `__rcptr!T` that does reference counting.

It's an alternative to `__RefCount` (https://github.com/dlang/druntime/pull/2679) in the sense that `__RefCount` would almost always be used to manage a `malloc`/`free`-managed block of memory.

`__rcptr!T` improves upon `__RefCount` by not exposing an thread-unsafe `isUnique` and by being simpler to use (e.g. default construction is allowed).

For simplicity, in this PR the counter is external to the memory managed by `__rcptr`, but it should be possible to modify `__rcptr` to use intrusive reference counting.

Most of the other issues from `__RefCount` do still persist. We're casting `const`/`immutable` away, but we hope to solve this through `__mutable`/`__metadata`. We also still have to use atomic increments/decrements due to implicit conversions to/from `immutable`.

### DIFF
--- a/src/core/experimental/rcptr.d
+++ b/src/core/experimental/rcptr.d
@@ -1,0 +1,208 @@
+// Written in the D programming language.
+/**
+This module provides a shared pointer implementation with memory management
+through reference counting.
+
+License: $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+
+Authors: Les De Ridder
+
+Source: $(DRUNTIMESRC core/experimental/rcptr.d)
+*/
+module core.experimental.rcptr;
+
+///
+unittest
+{
+    import core.stdc.stdlib : calloc;
+    auto allocInts = (size_t count) => cast(int*) calloc(count, int.sizeof);
+
+    __rcptr!int a; //default ctor
+    assert(a.get is null);
+
+    auto b = __rcptr!int(allocInts(10)); //ptr ctor
+
+    {
+        auto c = a; //copy ctor
+        assert(c.get == a.get);
+        assert(c.get != b.get);
+
+        c = b; //opAssign
+        assert(c.get != a.get);
+        assert(c.get == b.get);
+    }
+
+    assert(a.count is null);
+    assert(*b.count == 0);
+}
+
+/**
+A reference counted shared pointer that can be used to implement reference
+counted data structures.
+
+The reference count is automatically incremented/decremented on assignment,
+copy (construction), and destruction. When there are no more references to the
+pointer, the reference count is automatically deallocated and the referenced
+pointer is `free`d.
+
+Implementation: The internal implementation of `__rcptr` uses `malloc`/`free`.
+*/
+struct __rcptr(T)
+{
+    alias CounterType = uint;
+
+    private T* ptr = null;
+    private shared(CounterType)* count = null;
+
+    /**
+    Creates a new `__rcptr` instance, tracking the provided pointer.
+
+    This implies that the ownership of the pointer is transferred to
+    `__rcptr`.
+
+    Params:
+         ptr = pointer to memory to be managed by `__rcptr`
+    */
+    this(inout(T*) ptr) inout
+    {
+        this.ptr = ptr;
+
+        if (ptr !is null)
+        {
+            import core.memory : pureCalloc;
+
+            // We use `calloc` so we don't have to manually initialise count/addRef
+            count = (() @trusted inout => cast(typeof(count)) pureCalloc(1, CounterType.sizeof))();
+        }
+    }
+
+    /**
+    Get the owned pointer.
+
+    Returns:
+        The pointer owned by this instance of `__rcptr`.
+    */
+    @system
+    inout(T*) get() inout
+    {
+        return ptr;
+    }
+
+    ~this()
+    {
+        delRef();
+    }
+
+    ///
+    void opAssign(ref __rcptr!T rhs)
+    {
+        if (rhs.count == count)
+        {
+            return;
+        }
+
+        delRef();
+
+        ptr = rhs.ptr;
+        count = rhs.count;
+
+        addRef();
+    }
+
+    ///
+    this(scope ref inout __rcptr!T rhs) inout
+    {
+        ptr = rhs.ptr;
+        count = rhs.count;
+
+        addRef();
+    }
+
+    private void addRef() inout
+    {
+        import core.atomic : atomicOp;
+
+        if (ptr is null)
+        {
+            return;
+        }
+
+        () @trusted { atomicOp!"+="(*(cast(shared(CounterType)*) count), 1); } ();
+    }
+
+    private void delRef()
+    {
+        import core.atomic : atomicOp;
+
+        if (ptr is null)
+        {
+            return;
+        }
+
+        // The counter is left at -1 when this was the last reference
+        // (i.e. the counter is 0-based, because we use calloc)
+        if (atomicOp!"-="(*count, 1) == -1)
+        {
+            deallocate();
+        }
+    }
+
+    @trusted
+    private void deallocate()
+    {
+        import core.memory : pureFree;
+
+        pureFree(ptr);
+        pureFree(cast(CounterType*) count);
+    }
+}
+
+unittest
+{
+    struct rcarray
+    {
+        @safe @nogc nothrow:
+
+        private __rcptr!int ptr;
+        private size_t size;
+
+        this(size_t size)
+        {
+            import core.stdc.stdlib : calloc;
+
+            this.ptr = __rcptr!int(() @trusted { return cast(int*) calloc(size, int.sizeof); }());
+            this.size = size;
+        }
+
+        void opAssign(ref rcarray rhs)
+        {
+            // This will update the reference count
+            ptr = rhs.ptr;
+            // Update the size
+            size = rhs.size;
+        }
+
+        this(ref rcarray rhs)
+        {
+            // This will update the reference count
+            ptr = rhs.ptr;
+            // Update the size
+            size = rhs.size;
+        }
+    }
+
+    auto a = rcarray(42);
+    assert(*a.ptr.count == 0);
+    {
+        auto a2 = a; // Construct a2 by copy construction
+        assert(*a.ptr.count == 1);
+        auto a3 = rcarray(4242);
+        a2 = a3; // Assign a3 into a2; a's ref count drops
+        assert(*a.ptr.count == 0);
+        a3 = a; // Assign a into a3; a's ref count increases
+        assert(*a.ptr.count == 1);
+        // a2 and a3 go out of scope here
+        // a2 is the last ref to rcarray(4242) -> gets freed
+    }
+    assert(*a.ptr.count == 0);
+}
--- a/mak/COPY
+++ b/mak/COPY
@@ -21,6 +21,8 @@ COPY=\
 	$(IMPDIR)\core\time.d \
 	$(IMPDIR)\core\vararg.d \
 	\
+	src\core\experimental\rcptr.d \
+	\
 	$(IMPDIR)\core\internal\abort.d \
 	$(IMPDIR)\core\internal\arrayop.d \
 	$(IMPDIR)\core\internal\convert.d \
--- a/mak/DOCS
+++ b/mak/DOCS
@@ -16,6 +16,8 @@ DOCS=\
 	$(DOCDIR)\core_time.html \
 	$(DOCDIR)\core_vararg.html \
 	\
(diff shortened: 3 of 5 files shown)


src/core/experimental/rcptr.d
```diff
@@ -0,0 +1,208 @@
...
+            deallocate();
+        }
+    }
+
+    @trusted
+    private void deallocate()
+    {
+        import core.memory : pureFree;
+
+        pureFree(ptr);
```
wilzbach: So this is a big problem here as we assume that `ptr` as allocated via `malloc`. Maybe we can templatize `__rcptr` and allow the user to provide his own deallocate function? Alternatively, we would need to store a pointer to such a deallocation function in the `count` payload.

src/core/experimental/rcptr.d
```diff
@@ -0,0 +1,208 @@
...
+        import core.atomic : atomicOp;
+
+        if (ptr is null)
+        {
+            return;
+        }
+
+        // The counter is left at -1 when this was the last reference
+        // (i.e. the counter is 0-based, because we use calloc)
+        if (atomicOp!"-="(*count, 1) == -1)
```
thewilsonator: `if (atomicOp!"-="(*count, 1) < 0)`

also make counter type `int`.

lesderid: Why `< 0` instead of `== -1`? Readability?

wilzbach: Because `delRef` could be wrongfully called twice by threads. With == you ensure that at least you don't deallocate twice.

thewilsonator: I was going to say speed, but I'll defer to @andralex since he actually knows what he's talking about here.

andralex: Yah, comparisons against 0 are smaller (and therefore potentially faster).

thewilsonator: and potential races / double frees?

src/core/experimental/rcptr.d
lesderid: If we modify `__rcptr` to use intrusive ref counting we could properly support custom allocators (once we have them in druntime).

For now, should I just add a template parameter?

src/core/experimental/rcptr.d
andralex: That's a different consideration - an `assert` seems appropriate for those cases.

src/core/experimental/rcptr.d
edi33416: Sadly, I'm not sure how we'll be able to use custom allocators as we'll be hit by the same problem as in the past: calling allocate / deallocate on an immutable allocator instance.

The very first implementation that I tried (long ago) was with intrusive reference counting and custom allocators and that hit a lot of bumps down the road. This is why we've over-simplified the design, going from an array with intrusive RC and custom allocators support to the implementation of `rcarray`

src/core/experimental/rcptr.d
```diff
@@ -0,0 +1,208 @@
...
+pointer is `free`d.
+
+Implementation: The internal implementation of `__rcptr` uses `malloc`/`free`.
+*/
+struct __rcptr(T)
+{
+    alias CounterType = int;
+
+    private T* ptr = null;
+    private shared(CounterType)* count = null;
```
TurkeyMan: Still `shared` :(

TurkeyMan: Also not invasive; I will not use this. As the only person I'm aware of who has been harping on an on about RC since 2009, it would be a shame to see someone like me not use this.

src/core/experimental/rcptr.d
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    /**
+    Get the owned pointer.
+
+    Returns:
+        The pointer owned by this instance of `__rcptr`.
+    */
+    @system
+    inout(T*) get() inout
```
TurkeyMan: Pointer to payload must be `scope`.

src/core/experimental/rcptr.d
```diff
@@ -0,0 +1,208 @@
...
+
+    ///
+    void opAssign(ref __rcptr!T rhs)
+    {
+        if (rhs.count == count)
+        {
+            return;
+        }
+
+        delRef();
```
TurkeyMan: You need to `addRef` before `delRef`. It's possible that the `delRef` may destroy `this` before `addRef` down below retains it.

src/core/experimental/rcptr.d
```diff
@@ -0,0 +1,208 @@
...
+        import core.atomic : atomicOp;
+
+        if (ptr is null)
+        {
+            return;
+        }
+
+        // The counter is left at -1 when this was the last reference
+        // (i.e. the counter is 0-based, because we use calloc)
+        if (atomicOp!"-="(*count, 1) < 0)
```
TurkeyMan: This isn't how to do atomic ref releasing. You should compare == 0 to avoid decrementing when possible.

src/core/experimental/rcptr.d
lesderid: We're in `opAssign`, the `addRef` is after changing the counter over.

src/core/experimental/rcptr.d
TurkeyMan: > This is why we've over-simplified the design, going from an array with intrusive RC and custom allocators support to the implementation of rcarray

I understand the troubles you face, but this solution kinda buys us nothing. At least, it buys me nothing personally.
I won't use not-invasive ref counting.

What are we trying to do with this solution? Is it a stop-gap, or is it a thing?

src/core/experimental/rcptr.d
TurkeyMan: `rhs` might be dead.
You need to:
```d
rhs.addRef();
swap(this, rhs);
rhs.decRef();
```

TurkeyMan: `rhs` is a `ref`, which means the calling stack frame might not have acquired a ref for it.

TurkeyMan: An alternative may be to accept the argument by value...

src/core/experimental/rcptr.d
TurkeyMan: The proper way to do this is
```d
if (atomicLoad(rc) == 1 || atomicDecrement(rc) == 0)

File: src/core/experimental/rcptr.d [hunk 7]
```diff
@@ -0,0 +1,208 @@
...
+        import core.atomic : atomicOp;
+
+        if (ptr is null)
+        {
+            return;
+        }
+
+        // The counter is left at -1 when this was the last reference
+        // (i.e. the counter is 0-based, because we use calloc)
+        if (atomicOp!"=="(*count, 0) || atomicOp!"-="(*count, 1) == -1)
```
TurkeyMan: I don't know what `atomicOp!"=="` actually is, but I think this is not right either.
It should be a non-atomic comparison:
```d
if (atomicLoad!(MemoryOrder.raw)(*count) == 0 || atomicOp!"-="(*count, 1) == -1)
```
The decrement should be relaxed too.

src/core/experimental/rcptr.d
```diff
@@ -0,0 +1,208 @@
...
+        return ptr;
+    }
+
+    ~this()
+    {
+        delRef();
+    }
+
+    ///
+    void opAssign(__rcptr!T rhs)
```
TurkeyMan: Now I'm really confused... you changed it to accept the arg by value, but then still ref-fiddling in the function?
All you need to do is swap the pointers. Taking the arg by value will increment the ref count, and then destruction of the local on return will decrement; you need to swap them before returning so it decrements the right object.

Sadly, this approach accepting the argument by-val does not optimise for the `rhs.count == count` case; it will do the ref fiddling in that case when it doesn't need to.

lesderid: Ah, my bad, I see what you're saying now. It does get a little messy because there is no swap function in druntime and we have to swap both `ptr` and `count`.

It indeed doesn't optimise for the `rhs.count == count` case though, but I'm not sure how much of an issue that is?

TurkeyMan: Maybe we should have a `swap` function ;)

src/core/experimental/rcptr.d
lesderid: Looks like it should be right like this? https://github.com/dlang/druntime/blob/3c9b39f3d9f6575d8df5289747f7e6d0f0680b6b/src/core/atomic.d#L819

TurkeyMan: I'm not sure what or why you just linked to that? Or what you mean by 'like this'?

TurkeyMan: Oh , maybe I see what you're saying...
Considering this code:
```
         TailShared!T get = atomicLoad!(MemoryOrder.raw)( val ); 
         mixin( "return get " ~ op ~ " mod;" ); 
```
It's clear that the comparison is not atomic; despite the name of the function...

So yeah you're right, but while that may be true, I would recommend NOT calling a function called `atomicOp` that performs a non-atomic op... it's misleading at best, and for my money, ugly and harder to understand.

lesderid: > So yeah you're right, but while that may be true, I would recommend NOT calling a function called `atomicOp` that performs a non-atomic op... it's misleading at best, and for my money, ugly and harder to understand.

That's fair, especially considering `atomicOp`'s documentation doesn't specify it.

TurkeyMan: It's really not a great library.

src/core/experimental/rcptr.d
wilzbach: > I won't use not-invasive ref counting.

Why not?

> What are we trying to do with this solution?

Library ref-counting for RefCounting for rcarray & co . with some (future) compiler magic.

TurkeyMan: It would be incompatible with any of my existing code. It's also inefficient.

What compiler magic are we envisioning?

wilzbach: > It would be incompatible with any of my existing code. 

This isn't intended to be for users. Hence, __

> What compiler magic are we envisioning?

I think it was planned to detect __RefCounted for better @safe safety, but it currently looks like Water is trying to generalize this with DIP 1021.

TurkeyMan: > > It would be incompatible with any of my existing code.
> 
> This isn't intended to be for users. Hence, __

I'm not sure what you mean... how do I make use of it then? What does it offer if not a platform the specifying how ref-counting should function?

I want to know why `opInc`/`opDec` failed? After 10 years of waiting for this, this direction is embarrassingly feeble, and even slightly worse than C++, which we determined was so disappointing that we spent **TEN YEARS** trying to do something better. If we can't do something *at least* as good as C++, then common-sense suggests we just copy C++, at least then we'll be fully compatible with C++.

wilzbach: > how do I make use of it then?

You don't.


>  What does it offer if not a platform specifying how ref-counting should function?

It's internal for now and the forseeable future as no one is fully sure on the direction.

> I want to know why opInc/opDec failed? 

I don't know.

> this direction is embarrassingly feeble

We need rcarray __now__, but we don't want to expose the exact details of it yet. What's so wrong with that?

BTW I'm the wrong person to ask these questions to.

TurkeyMan: > > how do I make use of it then?
> 
> You don't.

Huh? Why are we introducing code that has no use?

> > What does it offer if not a platform specifying how ref-counting should function?
> 
> It's internal for now and the forseeable future as no one is fully sure on the direction.

Evidently.

> > I want to know why opInc/opDec failed?
> 
> I don't know.

I feel like we need to understand this very clearly.

> > this direction is embarrassingly feeble
> 
> We need rcarray **now**, but we don't want to expose the exact details of it yet. What's so wrong with that?

What do we need it for? We've done without it for a very long time...

> BTW I'm the wrong person to ask these questions to.

Okay, but you're the only one posting on the thread that appears to have a sense of authority :)

wilzbach: > Huh? Why are we introducing code that has no use?

It has a used. It will be immediately used for rcarray, but allowing us to not commit to the interface/design just yet.

> What do we need it for? We've done without it for a very long time... 

Sure, and I survived without a million dollars for a very long time too. However, we need to start somewhere if the vision of a fully @nogc D is ever to become a reality.

> you're the only one posting on the thread that appears to have a sense of authority

Yeah that's pretty sad. Maybe this conversation is better moved to the NG?

RazvanN7:

PR #2691, opened 2019-07-22
Test fix for 19831

ibuclaw:
currentLsd starts off with being null, so the function always breaks on the first loop without doing anything.

@FeepingCreature


FeepingCreature:
Not much point @ing me, I didn't understand the code the first time.

PR #2692, opened 2019-07-24
std::optional

TurkeyMan:
I think this is done; just gotta get the tests in order...

### DIFF
--- a/src/core/stdcpp/optional.d
+++ b/src/core/stdcpp/optional.d
@@ -1,0 +1,201 @@
+/**
+ * D header file for interaction with C++ std::optional.
+ *
+ * Copyright: Copyright (c) 2018 D Language Foundation
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors:   Manu Evans
+ * Source:    $(DRUNTIMESRC core/stdcpp/optional.d)
+ */
+
+module core.stdcpp.optional;
+
+import core.stdcpp.exception : exception;
+
+version (CppRuntime_DigitalMars)
+{
+    pragma(msg, "std::optional not supported by DMC");
+}
+version (CppRuntime_Clang)
+{
+    private alias AliasSeq(Args...) = Args;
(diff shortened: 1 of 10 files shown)


thewilsonator:
LMK when its good.

TurkeyMan:
It's not good. Unit tests never work, and there's no way to build and run tests locally, and I just CBF.

12345swordy:
You going to finished up std::vector hopefully?

TurkeyMan:
`std::vector` works on windows.

TurkeyMan:
Help make the unit tests work and I'll do the work to support the other platforms. The unit test infrastructure is SOOO HARD and it's the thing that's holding this whole process up.

TurkeyMan:
What are we supposed to do about all these ancient test boxes?

TurkeyMan:
The unit test fails because `hasElaborateCopyConstructor` is broken:
https://issues.dlang.org/show_bug.cgi?id=19902

TurkeyMan:
@12345swordy `std::vector` is in, with a few known issues...

src/core/stdcpp/optional.d
```diff
@@ -0,0 +1,201 @@
...
+ * Source:    $(DRUNTIMESRC core/stdcpp/optional.d)
+ */
+
+module core.stdcpp.optional;
+
+import core.stdcpp.exception : exception;
+
+version (CppRuntime_DigitalMars)
+{
+    pragma(msg, "std::optional not supported by DMC");
```
thewilsonator: I think this should be a generic "DMC does not support C++11 or newer" generated by the compiler. i.e. if one passes `-extern-std=c++1[1|4|7] -m32`.

TurkeyMan: DMC does have some sparse C++11 features.

thewilsonator: Urgh, carry on then.

TurkeyMan:
Buildkite fails... that other PR really needs attention ;)

Geod24:
No it did not. And now master is broken. Once again.

AndrejMitrovic:
Isn't it possible to make a github rule preventing merges until the PR is green? I also don't understand the lax attitude of merging every PR under the sun.

PR #2696, opened 2019-07-26
Replace `in` with `const` and `const scope` for src/core/sys/windows

JinShil:
## About This PR
Followup to 

This PR only addresses code in src/core/sys/windows

## Background

### DIFF
--- a/src/core/sys/windows/winreg.d
+++ b/src/core/sys/windows/winreg.d
@@ -97,58 +97,58 @@ static if (_WIN32_WINNT >= 0x600) {
 }
 
 extern (Windows) nothrow @nogc {
-    LONG RegCloseKey(in HKEY);
+    LONG RegCloseKey(const scope HKEY);
     LONG RegConnectRegistryA(LPCSTR, HKEY, PHKEY);
     LONG RegConnectRegistryW(LPCWSTR, HKEY, PHKEY);
-    LONG RegCreateKeyExA(in HKEY, LPCSTR, DWORD, LPSTR, DWORD, REGSAM,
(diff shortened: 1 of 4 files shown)


src/core/sys/windows/winreg.d
```diff
@@ -97,58 +97,58 @@ static if (_WIN32_WINNT >= 0x600) {
 }
 
 extern (Windows) nothrow @nogc {
-    LONG RegCloseKey(in HKEY);
+    LONG RegCloseKey(const HKEY);
```
n8sh: Why `const` here instead of `scope const`?

JinShil: Because `HKEY` is a value type.

JinShil: Actually, I think `HKEY` may be `const(void*)`.  It's defined here here:  https://github.com/dlang/druntime/blob/8fd52019826259dc92ab712f4b37a3f0ea4d8265/src/core/sys/windows/windef.d#L94

And `DECLARE_HANDLE` is defined here:  https://github.com/dlang/druntime/blob/8fd52019826259dc92ab712f4b37a3f0ea4d8265/src/core/sys/windows/basetsd.d#L41-L50

I'll update it to `const scope`.

thewilsonator: I think there was a whole pr that you wrote with HKEY as const instead of const scope.

JinShil: That was this PR.

thewilsonator: It seems it was, for some reason I thought I had merged it already.

PR #2704, opened 2019-07-29
Replace `in` with `const` and `const scope` in core

JinShil:
## About This PR
Followup to 

This PR predominately addresses code in src/core/

## Background

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -267,7 +267,7 @@ unittest
  * (No longer an intrisic - the compiler recognizes the patterns
  * in the body.)
  */
-int bt(in size_t* p, size_t bitnum) pure @system
+int bt(const scope size_t* p, size_t bitnum) pure @system
 {
     static if (size_t.sizeof == 8)
         return ((p[bitnum >> 6] & (1L << (bitnum & 63)))) != 0;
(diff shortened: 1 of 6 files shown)


src/core/bitop.d
```diff
@@ -954,28 +954,28 @@ version (D_InlineAsm_X86_64)
  *  Bitwise rotate `value` left (`rol`) or right (`ror`) by
  *  `count` bit positions.
  */
-pure T rol(T)(in T value, in uint count)
+pure T rol(T)(const T value, const uint count)
     if (__traits(isIntegral, T) && __traits(isUnsigned, T))
```
JinShil: Due to the `isIntegral` constraint I made this `const`.

src/core/memory.d
```diff
@@ -141,12 +141,12 @@ private
     extern (C) GC.Stats gc_stats ( ) nothrow @nogc;
     extern (C) GC.ProfileStats gc_profileStats ( ) nothrow @nogc @safe;
```
ZombineDev: Everything except the following should not accept `scope` arguments. Especially the `add(Root|Range)` as their sole purpose is save (i.e. escape) the pointer argument.

Can accept `scope`:
* `(get|set|clr)Attr`
* `sizeOf`
* `query`

JinShil: Addressed with https://github.com/dlang/druntime/pull/2706

It appears the original code was wrong.  #2706 should set things straight.

PR #2705, opened 2019-07-29
Make TempStringNoAlloc a template

skoppe:
Fixes https://issues.dlang.org/show_bug.cgi?id=20088

Since the new `__ArrayCast` template hook, casting arrays is broken in betterC. The reason is that `__ArrayCast` indirectly depends on `TempStringNoAlloc.get`, which is not a template function and doesn't get linked in betterC.

### DIFF
--- a/src/core/internal/string.d
+++ b/src/core/internal/string.d
@@ -53,7 +53,7 @@ char[] unsignedToTempString()(ulong value, return scope char[] buf, uint radix =
     return buf[i .. $];
 }
 
-private struct TempStringNoAlloc
+private struct TempStringNoAlloc()
 {
     // need to handle 65 bytes for radix of 2 with negative sign.
     private char[65] _buf = void;
@@ -79,7 +79,7 @@ Returns:
 */
 auto unsignedToTempString()(ulong value, uint radix = 10) @safe
 {
-    TempStringNoAlloc result = void;
+    TempStringNoAlloc!() result = void;
     result._len = unsignedToTempString(value, result._buf, radix).length & 0xff;
     return result;
 }
--- a/test/betterc/Makefile
+++ b/test/betterc/Makefile
@@ -1,6 +1,6 @@
 include ../common.mak
 
-TESTS:=test18828 test19416 test19421 test19561
+TESTS:=test18828 test19416 test19421 test19561 test20088
 
 .PHONY: all clean
 all: $(addprefix $(ROOT)/,$(addsuffix ,$(TESTS)))
--- a/test/betterc/src/test20088.d
+++ b/test/betterc/src/test20088.d
@@ -1,0 +1,14 @@
+/*******************************************/
+// https://issues.dlang.org/show_bug.cgi?id=20088
+
+struct S {
+    int i;
+}
+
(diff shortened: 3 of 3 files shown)


thewilsonator:
Please add the test case from the bug report.

ZombineDev:
In Phobos we have the option to run `@betterC` tests, and I think we should set up something like that for druntime as well. In the mean time, you can add a test to Phobos as well. See:
https://github.com/dlang/phobos/pull/5952

skoppe:
> In Phobos we have the option to run `@betterC` tests, and I think we should set up something like that for druntime as well. In the mean time, you can add a test to Phobos as well

Sure, but what would be the benefit? Plus, where do you want me to put it?

wilzbach:
Test fails with:

```
generated/linux/release/64/test20088.o: In function `_D6object__T16onArrayCastErrorZQtFNaNbNiNeAyamQemZv':
src/test20088.d:(.text._D6object__T16onArrayCastErrorZQtFNaNbNiNeAyamQemZv[_D6object__T16onArrayCastErrorZQtFNaNbNiNeAyamQemZv]+0x5e): undefined reference to `_D4core8internal6string17TempStringNoAlloc__T3getZQfMFNaNbNiNjNfZAa'
src/test20088.d:(.text._D6object__T16onArrayCastErrorZQtFNaNbNiNeAyamQemZv[_D6object__T16onArrayCastErrorZQtFNaNbNiNeAyamQemZv]+0x9e): undefined reference to `_D4core8internal6string17TempStringNoAlloc__T3getZQfMFNaNbNiNjNfZAa'
```

> In Phobos we have the option to run @betterC unit tests, and I think we should set up something like that for druntime as well - see:

`@betterC` should be a thing soon here as well (e.g. https://github.com/dlang/druntime/pull/2715).

skoppe:
> Test fails with:
> 
> ```
> generated/linux/release/64/test20088.o: In function `_D6object__T16onArrayCastErrorZQtFNaNbNiNeAyamQemZv':
> src/test20088.d:(.text._D6object__T16onArrayCastErrorZQtFNaNbNiNeAyamQemZv[_D6object__T16onArrayCastErrorZQtFNaNbNiNeAyamQemZv]+0x5e): undefined reference to `_D4core8internal6string17TempStringNoAlloc__T3getZQfMFNaNbNiNjNfZAa'
> src/test20088.d:(.text._D6object__T16onArrayCastErrorZQtFNaNbNiNeAyamQemZv[_D6object__T16onArrayCastErrorZQtFNaNbNiNeAyamQemZv]+0x9e): undefined reference to `_D4core8internal6string17TempStringNoAlloc__T3getZQfMFNaNbNiNjNfZAa'
> collect2: error: ld returned 1 exit status
> ```

Hmm, I see. Will look into it this weekend.

skoppe:
@wilzbach I found the mistake. Some CI's are still red though, but it doesn't seem to be caused by my changes. Would it help if I rebase on stable again?

wilzbach:
> Would it help if I rebase on stable again?

Probably. Just tried this for you.

thewilsonator:
Appveyor is expected to fail on stable.

PR #2706, opened 2019-07-29
Replace `in` with `const scope` for gc implementations

JinShil:
This PR also addresses https://github.com/dlang/druntime/pull/2704#pullrequestreview-268044024

cc @ZombineDev 

## About This PR
Followup to 
https://github.com/dlang/druntime/pull/2704

This PR predominately addresses code in the gc implementations

## Background

### DIFF
--- a/src/core/gc/gcinterface.d
+++ b/src/core/gc/gcinterface.d
@@ -182,7 +182,7 @@ interface GC
     /**
      * run finalizers
      */
-    void runFinalizers(in void[] segment) nothrow;
+    void runFinalizers(const scope void[] segment) nothrow;
 
     /*
      *
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -141,11 +141,11 @@ private
     extern (C) GC.Stats gc_stats ( ) nothrow @nogc;
     extern (C) GC.ProfileStats gc_profileStats ( ) nothrow @nogc @safe;
 
-    extern (C) void gc_addRoot( const scope void* p ) nothrow @nogc;
-    extern (C) void gc_addRange( const scope void* p, size_t sz, const TypeInfo ti = null ) nothrow @nogc;
+    extern (C) void gc_addRoot(const void* p ) nothrow @nogc;
+    extern (C) void gc_addRange(const void* p, size_t sz, const TypeInfo ti = null ) nothrow @nogc;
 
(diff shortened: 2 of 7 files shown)


rainers:
`const` should be ok for add/remove root/range, the GC isn't going to change the memory. `scope` should be removed, though. I think that is what @ZombineDev meant.

JinShil:
> `const` should be ok for add/remove root/range

But that's not how the GC API is written.  See...

https://github.com/dlang/druntime/blob/a370406fae72602ab2e0fe632d7a70a71da79267/src/gc/proxy.d#L228-L231

https://github.com/dlang/druntime/blob/a370406fae72602ab2e0fe632d7a70a71da79267/src/gc/proxy.d#L238-L241

https://github.com/dlang/druntime/blob/a370406fae72602ab2e0fe632d7a70a71da79267/src/gc/impl/proto/gc.d#L27-L28

https://github.com/dlang/druntime/blob/a370406fae72602ab2e0fe632d7a70a71da79267/src/gc/impl/proto/gc.d#L202

https://github.com/dlang/druntime/blob/a370406fae72602ab2e0fe632d7a70a71da79267/src/gc/impl/proto/gc.d#L207

Should I modify `proto/gc.d` and `proxy.d` to use `const`?

rainers:
> But that's not how the GC API is written. See...

But it used to have `in` in core.memory. I'd rather adjust the proxy functions and the GC implementations.

rainers:
> Should I modify `proto/gc.d` and `proxy.d` to use `const`?

Would be nice to be more const-correct, but it might be rather viral as you'll have to change all GC implementations. Should be ok to leave the proxy function as is for now.

JinShil:
I don't mind making all the changes.  I'll try to change them to `const void*` and see how it goes.

JinShil:
I tried to make everything const-correct, but as @rainers predicted it snowballed into too many changes outside the scope of this PR.  So, I did only what was necessary to keep the status quo.

It's clear there are some const-ness problems in the GC interface and implementations, but fixing that will have to be a different PR.

rainers:
Looks like core.gc.gcinterface.GC.runFinalizers() should be updated, too.

JinShil:
> Looks like core.gc.gcinterface.GC.runFinalizers() should be updated, too.

Done.  Thank you!

PR #2707, opened 2019-07-30
Enable C++17 tests

TurkeyMan:
Enable C++17 unit tests (including `string_view` which hasn't had unit tests run before) for windows where the compiler supports it.
Fixed a bug in `new` when C++17 is enabled.

### DIFF
--- a/setmscver.bat
+++ b/setmscver.bat
@@ -4,4 +4,10 @@ cl /nologo /EP ver.c > ver_raw.txt
 findstr /v /r /c:"^$" "ver_raw.txt" > "ver.txt"
 set /P _MSC_VER=< ver.txt
 echo set _MSC_VER=%_MSC_VER%
+if exist cflags.txt del /q cflags.txt
+if exist dflags.txt del /q dflags.txt
+if exist add_tests.txt del /q add_tests.txt
+if %_MSC_VER% GTR 1900 echo /std:c++17 > cflags.txt
+if %_MSC_VER% GTR 1900 echo -extern-std=c++17 > dflags.txt
+if %_MSC_VER% GTR 1900 echo string_view > add_tests.txt
 del ver.c ver_raw.txt
--- a/test/stdcpp/win64.mak
+++ b/test/stdcpp/win64.mak
@@ -5,29 +5,34 @@ MODEL=64
 DRUNTIMELIB=druntime64.lib
 CC=cl
 
-_MSC_VER = $(file < ..\..\ver.txt)
+TESTS=array allocator new string vector
 
-TESTS= array allocator new string vector
+_MSC_VER=$(file < ..\..\ver.txt)
+ADD_CFLAGS=$(file < ..\..\cflags.txt)
+ADD_DFLAGS=$(file < ..\..\dflags.txt)
+ADD_TESTS=$(file < ..\..\add_tests.txt)
+
+TESTS=$(TESTS) $(ADD_TESTS)
 
 test: $(TESTS)
 
 $(TESTS):
-	"$(CC)" -c /Fo$@_cpp.obj test\stdcpp\src\$@.cpp /EHsc /MT
-	"$(DMD)" -of=$@.exe -m$(MODEL) -conf= -Isrc -defaultlib=$(DRUNTIMELIB) -main -unittest -version=_MSC_VER_$(_MSC_VER) -mscrtlib=libcmt test\stdcpp\src\$@_test.d $@_cpp.obj
+	"$(CC)" -c /Fo$@_cpp.obj test\stdcpp\src\$@.cpp /EHsc /MT $(ADD_CFLAGS)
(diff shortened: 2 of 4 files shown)


TurkeyMan:
I shunted the if's out of the makefile... what a hack!

thewilsonator:
Lol, nice!

test/stdcpp/win64.mak
```diff
@@ -5,29 +5,34 @@ MODEL=64
 DRUNTIMELIB=druntime64.lib
 CC=cl
 
-_MSC_VER = $(file < ..\..\ver.txt)
+TESTS=array allocator new string
 
-TESTS= array allocator new string
+_MSC_VER=$(file < ..\..\ver.txt)
```
rainers: I would be very surprised if this works with Digital Mars make.
AFAICT including a file only works with a line like this (from druntime/win64.mak):
```
$(mak\COPY)
```
This reads a single assignment from the file.

TurkeyMan: What do you expect doesn't work?
This does work... the tests all pass.

TurkeyMan: Look for `-version=_MSC_VER_xxxx` in the CI logs. That value is read from this line at your comment.

TurkeyMan:
Why did that fail?!

test/stdcpp/win64.mak
```diff
@@ -5,29 +5,34 @@ MODEL=64
...
+ADD_TESTS=$(file < ..\..\add_tests.txt)
+
+TESTS=$(TESTS) $(ADD_TESTS)
 
 test: $(TESTS)
 
 $(TESTS):
-	"$(CC)" -c /Fo$@_cpp.obj test\stdcpp\src\$@.cpp /EHsc /MT
-	"$(DMD)" -of=$@.exe -m$(MODEL) -conf= -Isrc -defaultlib=$(DRUNTIMELIB) -main -unittest -version=_MSC_VER_$(_MSC_VER) -mscrtlib=libcmt test\stdcpp\src\$@_test.d $@_cpp.obj
+	"$(CC)" -c /Fo$@_cpp.obj test\stdcpp\src\$@.cpp /EHsc /MT $(ADD_CFLAGS)
```
rainers: Maybe this also works directly if there is a file with the name ADD_CFLAGS that has the settings, see https://github.com/DigitalMars/Compiler/blob/1dbd3e3381c8f7f7ab2e35214dcd63455ae38c29/dm/src/make/dmake.d#L1130.

TurkeyMan: I didn't see any documentation to that effect in the scant docs.
That would be highly unusual though I think... I kinda like that way I've written it makes it terribly clear what mischief is at play.

test/stdcpp/win64.mak
rainers: Surprising. `$(file < MAK\COPY)` is empty for me. `_MSC_VER=$(..\..\ver.txt)` would work.

I suspect the auto-tester is not using Digital Mars make here?

TurkeyMan: I ran the same script locally... It works, and the tests invoke with all proper args... How are you running it?
auto-tester and appveyor both seem to work... My local machine too. I'm surprised...

TurkeyMan: Like, check the CI logs. The msc ver and the cflags and dflags are supplied correctly. It also did build `string_view`, which is pulled the same way...

rainers: Your PR passes for me locally, too. I'm confused ;-)

rainers:
> Why did that fail?!

Looks like one of the spurious failures (systime timed out).

setmscver.bat
```diff
@@ -4,4 +4,7 @@ cl /nologo /EP ver.c > ver_raw.txt
 findstr /v /r /c:"^$" "ver_raw.txt" > "ver.txt"
 set /P _MSC_VER=< ver.txt
 echo set _MSC_VER=%_MSC_VER%
+if %_MSC_VER% GTR 1900 echo /std:c++17 > cflags.txt
+if %_MSC_VER% GTR 1900 echo -extern-std=c++17 > dflags.txt
+if %_MSC_VER% GTR 1900 echo string_view > add_tests.txt
```
rainers: Would be good to remove these files for lower versions, so you don't get a stale copy when trying different versions of VS one after the other.

TurkeyMan: I though that too, but I couldn't work out how to delete them. `del /f` errors if they don't exist, and they don't exist on the auto-tester build.

rainers: Just put `if exist cflags.txt del /q cflags.txt` etc. before the if statements.

TurkeyMan: DOS is so shit ;)

TurkeyMan:
What's going on with buildkite?

wilzbach:
see e.g. https://github.com/dlang/dmd/pull/10244#pullrequestreview-268622608

TurkeyMan:
Okay, so this PR is good then?

TurkeyMan:
Wait.. it just switched from green to red... what's happening here? O_o

thewilsonator:
the concurrency tests have always been flaky

TurkeyMan:
So... am I supposed to do something now?

thewilsonator:
Ive restarted the tests, they might take a bit to come online.

PR #2713, opened 2019-08-01
Fix shared libraries for LDC on macOS

jacob-carlborg:
This also avoids relying on the assumption that the current executable is always the first image.

### DIFF
--- a/src/rt/util/utility.d
+++ b/src/rt/util/utility.d
@@ -1,0 +1,27 @@
+/**
+ * Contains various utility functions used by the runtime implementation.
+ *
+ * Copyright: Copyright Digital Mars 2016.
+ * License: Distributed under the
+ *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
+ *    (See accompanying file LICENSE)
+ * Authors: Jacob Carlborg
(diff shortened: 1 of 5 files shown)


jacob-carlborg:
@kinke upstream fix for shared libraries on macOS.

jacob-carlborg:
I don't think this deserves a changelog entry.

src/rt/util/utility.d
```diff
@@ -0,0 +1,27 @@
...
+ * throw an AssertError and thus require an initialized GC, which isn't the case
+ * (yet or anymore) for the startup/shutdown code in this module
+ * (called by CRT ctors/dtors etc.).
+ */
+package(rt) void safeAssert(
+    bool condition, scope string msg, size_t line = __LINE__
+) nothrow @nogc @safe
+{
+    import core.internal.abort;
+    condition || abort(msg, __FILE__, line);
```
kinke: The `__FILE__` here (`rt.util.utility`) defeats the original purpose.

jacob-carlborg: Right, good point. I'll fix that.

kinke: => #2714

PR #2716, opened 2019-08-02
Clang std::string

TurkeyMan:


TurkeyMan:
There's a mangling issue that I can't isolate... :/

thewilsonator:
For reference the issue is:
```
/tmp/string-abbbdd.o: In function `sumOfElements_val(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >)':
src/string.cpp:(.text+0x197): undefined reference to `fromC_ref(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)'
src/string.cpp:(.text+0x1c6): undefined reference to `fromC_val(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >)'
```

TurkeyMan:
I can read the logs ;) ... but the issue is this:
`std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >`
See the `std::__1::basic_string`, `std::__1::char_traits`, and `std::__1::allocator`?
It should be `std::__1::basic_string`, `std::char_traits`, and `std::allocator`... and it's definitely written that way in the code.

I suspect perhaps an issue with the mangling exists where the outer template's namespace it being transitively applied to the template arguments or something, rather than those arguments being mangled with their own namespaces...
I can't reproduce that though, I have no system that can reproduce those build results... so I don't know how to test this.

Geod24:
I can reproduce on my system. I'll take a look at it.

TurkeyMan:
I think I spotted it; my libc++ reference was inconsistent with other material I've found on the net... I think those things SHOULD be mangled with `__1`. I've pushed that change, and we'll see if it links...

Geod24:
If it's any help, the following runs on Mac: https://github.com/bpfkorea/agora/blob/v0.x.x/source/scpd/Cpp.d

And yeah `std::allocator` is in `__1`

thewilsonator:
> I can read the logs ;)

It took me a while to find it ;) thats why I posted that.

TurkeyMan:
@Geod24 If CI fails, can you grab this patch and see if you can see where I've fucked up the mangling?

Geod24:
Looks like you won't be needing it :)

TurkeyMan:
Oh my! Huzzah!

PR #2717, opened 2019-08-03
std::string GCC tests

TurkeyMan:
Add GCC old ABI tests. Modern ABI has an interior pointer and is blocked on move constructors.

### DIFF
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -14,7 +14,8 @@ ifeq (osx,$(OS))
 endif
 ifeq (linux,$(OS))
 #	TESTS+=string vector
-#	OLDABITESTS+=string vector
+	OLDABITESTS+=string
+#	OLDABITESTS+=vector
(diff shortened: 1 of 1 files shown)


TurkeyMan:
I think this should fail... I'm not confident the CI is actually running the tests :/
How can I see if they're running? They appear to be building, but what's weird is there's lines in the log like `stdcpp/xutility.d` which doesn't have tests. Why is that there, as if it's being tested? It's certainly not in the test makefile...?

TurkeyMan:
So, are any of my posix unit tests being run?
How can I tell? It seemed like they are... but the logs are weird and not very clear.

TurkeyMan:
I mean, it might actually work... I just find it very unlikely that it worked first try.

rainers:
> So, are any of my posix unit tests being run?
> How can I tell? It seemed like they are... but the logs are weird and not very clear.

You can grep for "Testing old_string" to see it is actually run.

PR #2719, opened 2019-08-03
Fix string tests not running

TurkeyMan:
I knew something was wrong!!
It was all too easy...

### DIFF
--- a/src/core/stdcpp/string.d
+++ b/src/core/stdcpp/string.d
@@ -13,37 +13,42 @@
 module core.stdcpp.string;
 
 import core.stdcpp.allocator;
+import core.stdcpp.xutility : StdNamespace;
 import core.stdc.stddef : wchar_t;
 
+version (OSX)
+{
+    // Apple decided to rock a different ABI... good for them!
+    version = _LIBCPP_ABI_ALTERNATE_STRING_LAYOUT;
+}
 version (CppRuntime_Gcc)
 {
     version (_GLIBCXX_USE_CXX98_ABI)
     {
-        private enum StdNamespace = "std";
+        private enum StringNamespace = "std";
 //        version = __GTHREADS; // TODO: we need to make ref-count interactions atomic
     }
     else
     {
         import core.internal.traits : AliasSeq;
-        private enum StdNamespace = AliasSeq!("std", "__cxx11");
+        private enum StringNamespace = AliasSeq!("std", "__cxx11");
     }
 }
 else
-    import core.stdcpp.xutility : StdNamespace;
+    alias StringNamespace = StdNamespace;
 
 enum DefaultConstruct { value }
 
 /// Constructor argument for default construction
 enum Default = DefaultConstruct();
 
-extern(C++, (StdNamespace)):
 @nogc:
 
 /**
  * Character traits classes specify character properties and provide specific
  * semantics for certain operations on characters and sequences of characters.
  */
-struct char_traits(CharT) {}
+extern(C++, (StdNamespace)) struct char_traits(CharT) {}
 
 // I don't think we can have these here, otherwise symbols are emit to druntime, and we don't want that...
 //alias std_string = basic_string!char;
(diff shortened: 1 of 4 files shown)


test/stdcpp/Makefile
```diff
@@ -6,23 +6,22 @@ TESTS:=allocator new
 TESTS11:=array
 TESTS17:=string_view
 OLDABITESTS:=
-LIBCPPTESTS:=
 
 ifeq (osx,$(OS))
-	TESTS+=string
+#	TESTS+=string
```
TurkeyMan: @Geod24 @jacob-carlborg
Can one of you guys uncomment this line and test on your OSX environments?
There's an error running the unit tests which I have no way to resolve. I'm fairly confident I have the struct layout correct for Apple, so maybe there's a calling convention thing?

TurkeyMan:
This should be merged to actually enable the tests. The OSX fix can appear in another patch (I don't have the means to fix it).

test/stdcpp/Makefile
TurkeyMan: Fucking typical that Apple can't just use libc++ unmolested! They just have to change some shit don't they!!

thewilsonator:
I just noticed on line 403 `static_assert(false, "DO WE NEED THIS?");`

TurkeyMan:
Yeah, and I still don't know :)

TurkeyMan:
Why is this sitting forever and the tests not completing?

thewilsonator:
Dunno.

src/core/stdcpp/string.d
```diff
@@ -727,6 +733,8 @@ extern(D):
         }
         else
         {
+            pragma(msg, "libstdc++ std::__cxx11::basic_string is not yet supported; the struct contains an interior pointer which breaks D move semantics!");
```
Geod24: Instead, can't we just disable `opAssign` and postblit to make it un-copyable ?

TurkeyMan: Copying is not the problem, moving is. D has no control over move semantics.

test/stdcpp/Makefile
Geod24: Okay, looking at it now. The `size` of "Hello" is 36 on the C++ size, and the iteration starts on `e`, not `H`. Will see if I can spot what's wrong.

TurkeyMan: Check that `_LIBCPP_ABI_ALTERNATE_STRING_LAYOUT` is appropriate for OSX?
I'm pretty sure it is... Evidence I could find suggested osx rocks the alternate ABI...

TurkeyMan: Check 1, does the by-ref pass correctly? If not, the structure is wrong. If it does, and by-val doesn't work, it's probably calling convention.

Geod24: No, `_LIBCPP_ABI_ALTERNATE_STRING_LAYOUT` does not seem to be enabled by default.
Disabling it make the first 2 lines pass, but then after assignment you'll get:
```
string(39567,0x109c275c0) malloc: *** error for object 0x7fdc29c02d10: pointer being freed was not allocated
string(39567,0x109c275c0) malloc: *** set a breakpoint in malloc_error_break to debug
```

TurkeyMan: If the struct looks right, and the values carry correctly from D to C++, maybe check that allocator.d is doing the right thing?
Maybe the allocation is being aligned or something? Windows does that, but I didn't see that business in libc++. Maybe Apple do something special in the allocator?
(I would expect the allocator tests to fail if they do)

jacob-carlborg: @TurkeyMan if you need access to a Mac you can use Circle CI. You can setup a job that fails immediately. For failed jobs you get the option, in the UI, to rerun the job with SSH enabled. Then you will have a session available for one hour.

TurkeyMan: I mean, that's very helpful information, but my personal policy is not to engage with the devil ;)

jacob-carlborg: Well, your loss 😃.

TurkeyMan: Not really. I'll never be upset that the Mac doesn't work :P

PR #2720, opened 2019-08-03
-checkaction=context: Improve struct output + add tests for issue 20100

wilzbach:
Adds the tests for https://github.com/dlang/dmd/pull/10270

Improved the struct output while I was at it because otherwise it would print `S(void*)` for an empty struct to to it printing the `this` pointer.

Requires https://github.com/dlang/dmd/pull/10270 to pass.


thewilsonator:
This fails on OSX and FreeBSD

RazvanN7:
Closing, since alternative #3462 was merged

PR #2721, opened 2019-08-03
Fix Issue 20066 - Assertion on void[] does not compile with -checkaction=context

wilzbach:


test/exceptions/src/assert_fail.d
```diff
@@ -119,16 +119,28 @@ void testAttributes() @safe pure @nogc nothrow
...
-    testAttributes();
+    //testIntegers();
+    //testIntegerComparisons();
+    //testFloatingPoint();
+    //testStrings();
+    //testToString();
+    //testArray();
+    //testStruct();
+    //testAA();
+    //testAttributes();
```
thewilsonator: Looks good mod this.

wilzbach: Sorry about that. Fixed.

PR #2722, opened 2019-08-04
core.stdcpp.string: Fix definition, enable tests for OSX

Geod24:
CC @TurkeyMan : The error I had suddenly disappeared, so might just be bad dependency tracking.

### DIFF
--- a/src/core/stdcpp/string.d
+++ b/src/core/stdcpp/string.d
@@ -18,8 +18,10 @@ import core.stdc.stddef : wchar_t;
 
 version (OSX)
 {
-    // Apple decided to rock a different ABI... good for them!
-    version = _LIBCPP_ABI_ALTERNATE_STRING_LAYOUT;
+    // There is an alternate ABI which is supposed to make the performance
+    // marginally better to to better alignment,
+    // however it is not enabled by default
+    // version = _LIBCPP_ABI_ALTERNATE_STRING_LAYOUT;
 }
 version (CppRuntime_Gcc)
 {
--- a/test/stdcpp/Makefile
+++ b/test/stdcpp/Makefile
@@ -8,7 +8,7 @@ TESTS17:=string_view
 OLDABITESTS:=
 
 ifeq (osx,$(OS))
-#	TESTS+=string
+	TESTS+=string
 #	TESTS+=vector
 endif
 ifeq (linux,$(OS))


TurkeyMan:
Ummm... I'm not so sure random crashes just 'go away' :/
If you have been able to step the code and watch the argument passing correct by ref AND by val, then I strongly suspect allocator alignment or pointer adjustment or something like that.
Did you check `allocator.d` in comparison to `<memory>` on OSX?

Actually, there is an option for libc++ iterator debugging which is guarded with `_LIBCPP_DEBUG_LEVEL`; as far as I can tell, it's not on by default; can you check if OSX keeps it disabled by default?

src/core/stdcpp/string.d
```diff
@@ -18,8 +18,10 @@ import core.stdc.stddef : wchar_t;
 
 version (OSX)
 {
-    // Apple decided to rock a different ABI... good for them!
-    version = _LIBCPP_ABI_ALTERNATE_STRING_LAYOUT;
+    // There is an alternate ABI which is supposed to make the performance
+    // marginally better to to better alignment,
+    // however it is not enabled by default
+    // version = _LIBCPP_ABI_ALTERNATE_STRING_LAYOUT;
 }
```
TurkeyMan: Remove this entire version block if it's not right.

Geod24:
It seems that `libc++` uses a clang builtin function for memory allocation so that it can remove some of them:

```C++
// /Library/Developer/CommandLineTools/usr/include/c++/v1/memory
    _LIBCPP_NODISCARD_AFTER_CXX17 _LIBCPP_INLINE_VISIBILITY
    pointer allocate(size_type __n, allocator<void>::const_pointer = 0)
        {
        if (__n > max_size())
            __throw_length_error("allocator<T>::allocate(size_t n)"
                                 " 'n' exceeds maximum supported size");
        return static_cast<pointer>(_VSTD::__libcpp_allocate(__n * sizeof(_Tp), _LIBCPP_ALIGNOF(_Tp)));
        }
    _LIBCPP_INLINE_VISIBILITY void deallocate(pointer __p, size_type __n) _NOEXCEPT
        {_VSTD::__libcpp_deallocate((void*)__p, __n * sizeof(_Tp), _LIBCPP_ALIGNOF(_Tp));}
    _LIBCPP_INLINE_VISIBILITY size_type max_size() const _NOEXCEPT
```

```c++
inline _LIBCPP_INLINE_VISIBILITY void *__libcpp_allocate(size_t __size, size_t __align) {
#ifndef _LIBCPP_HAS_NO_ALIGNED_ALLOCATION
  if (__is_overaligned_for_new(__align)) {
    const align_val_t __align_val = static_cast<align_val_t>(__align);
# ifdef _LIBCPP_HAS_NO_BUILTIN_OVERLOADED_OPERATOR_NEW_DELETE
    return ::operator new(__size, __align_val);
# else
    return __builtin_operator_new(__size, __align_val);
# endif
  }
#else
  ((void)__align);
#endif
#ifdef _LIBCPP_HAS_NO_BUILTIN_OPERATOR_NEW_DELETE
  return ::operator new(__size);
#else
  return __builtin_operator_new(__size);
#endif
}
```

```C++
#if !__has_builtin(__builtin_operator_new) || \
   __has_builtin(__builtin_operator_new) < 201802L
#define _LIBCPP_HAS_NO_BUILTIN_OVERLOADED_OPERATOR_NEW_DELETE
#endif
```

But in the end, it is calling [operator new](https://github.com/llvm-mirror/clang/blob/a25d062264efe6f7af3c8ad18e73989ae20909c5/lib/Sema/SemaExprCXX.cpp#L3522) when an allocation happens.

> If you have been able to step the code and watch the argument passing correct by ref AND by val, then I strongly suspect allocator alignment or pointer adjustment or something like that.

Not really, because debugging D on OSX is just terrible. But the test passes if you comment either [the cleanup](https://github.com/dlang/druntime/blob/aace1c5395862e1d6c7b07fb7cdeb59b55c2f3d8/src/core/stdcpp/string.d#L1019-L1020) or [the pass by value](https://github.com/dlang/druntime/blob/aace1c5395862e1d6c7b07fb7cdeb59b55c2f3d8/test/stdcpp/src/string_test.d#L20).

TurkeyMan:
Wow... so... what does `__builtin_operator_new` do? How do we emulate it's different behaviour? O_o
Is this actually impossible to support on OSX?

TurkeyMan:
Anyway, whatever it does, we need to do the same thing... can we find the source code?

Geod24:
> Anyway, whatever it does, we need to do the same thing... can we find the source code?

Already linked:
> But in the end, it is calling [operator new](https://github.com/llvm-mirror/clang/blob/a25d062264efe6f7af3c8ad18e73989ae20909c5/lib/Sema/SemaExprCXX.cpp#L3522) when an allocation happens.

But I don't think that's the problem here.
If you want me to look at this I'm going to need a bit more time tho.

TurkeyMan:
No rush at all. Only affects Apple users ;)

ZombineDev:
Is this `__builtin_operator_new` because they added support (though in C++20, IIRC) for using STL types like vector and string inside `constexpr` function?

SSoulaimane:
The issue here is the argument destruction convention in the frontend which deosn't match the C++ convention on POSIX (see issue [20235](https://issues.dlang.org/show_bug.cgi?id=20235)) which should be fixed with this https://github.com/dlang/dmd/pull/10593.

Geod24:
Great, thanks for tracking that down! Feel free to ping me if this doesn't get rebased once your PR is in.

Geod24:
Superseded by #2878

PR #2723, opened 2019-08-05
Add std::unique_ptr

TurkeyMan:


src/core/stdcpp/new_.d
```diff
@@ -40,6 +40,39 @@ extern (C++, "std")
 }
 
 
+///
+T* cpp_new(T, Args...)(auto ref Args) if (!is(T == class))
```
Geod24: What about dynamic and associative arrays ?

TurkeyMan: Yeah I was pondering that.
You can do `cpp_new!(T[N])`, which will allocate a static array of the desired length... maybe that's fine?
Alternatively we have some other form that receives a length... but I wasn't sure how to overload the function. Maybe it should be a separate function like `cpp_new_array()`?

Geod24: Yeah `cpp_new!(T[N])` is fine but won't `cpp_new!(T[])` allocate a `T[]*` technically ?

TurkeyMan: Yes, `T[]` will allocate a `T[]*`, which seems... fine?
If we want it to allocate a dynamic array and return a slice, then we need to decide how to supply the length.
Overload (difficult because `Args...`)? Separate function with a different name?

TurkeyMan:
What's going on with auto-tester?

TurkeyMan:
I don't know what's going on here... I'm kinda worried it's the same problem as the string on OSX... :/

test/stdcpp/src/memory_test.d
```diff
@@ -0,0 +1,39 @@
...
+    assert(!!x && !up2);
+
+    up = x.move;
+    assert(!!up && !x);
+
+    int* p = up.get;
+    up = passThrough(up.move);
+    assert(up.get == p);
+    up = changeIt(up.move);
+    assert(up.get != p);
```
SSoulaimane: It seems malloc returns the same pointer that was just freed, this the reason the ci fails on osx. mybe check the value instead `assert(*up.get == 20);`?

rainers: Good catch. I've added a commit to ensure the pointers in changeIt are different.

SSoulaimane: Perfect. It's working now.

rainers:
Merging manually as the buildkite failure is unrelated (AFAICT it doesn't merge the PR to master, but builds the PR branch as is).

rainers:
> Merging manually

Argh, suddenly blocked by a spurious doc-tester-failure. Rebased now...

thewilsonator:
and now Circle has failed, which is required.

PR #2724, opened 2019-08-05
Use `backtrace` on POSIX when available

Geod24:
Couple of low hanging fruit I caught while trying to debug the debug infos on OSX.

### DIFF
--- a/src/core/runtime.d
+++ b/src/core/runtime.d
@@ -739,8 +739,17 @@ static if (hasExecinfo) private class DefaultTraceInfo : Throwable.TraceInfo
 
     this()
     {
-        numframes = 0; //backtrace( callstack, MAXFRAMES );
-        if (numframes < 2) // backtrace() failed, do it ourselves
+        // it may not be 1 but it is good enough to get
+        // in CALL instruction address range for backtrace
+        enum CALL_INSTRUCTION_SIZE = 1;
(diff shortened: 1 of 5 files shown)


Geod24:
Interesting. This changes the line information on one of the test, to be the line after the `throw`.
But it works correctly on Mac OSX... So I'll just version it for now.

thewilsonator:
Whats the status of this?

Geod24:
I was puzzled for a very long time as to why this changed the debug info. That might be due to another bug I found, and on which I worked last month (https://github.com/dlang/dmd/pull/10722#issue-362055337) but I'm not 100% positive.
No ETA as to when I'll come back to this yet. I will maybe separate some of the refactoring.

thewilsonator:
OK, just clearing out some old GH notifications.

yazd:
The reason for the difference is most likely due to the fact that `backtrace` returns the *return* addresses, however, for the stack trace, we'll need the *caller* addresses.

The code here https://github.com/dlang/druntime/blob/ef788b3be2b1074787668dd45f29ae20e38a335b/src/core/runtime.d#L768-L771 subtracts "CALL_INSTRUCTION_SIZE" to point to the *caller* address instead.

You can try doing the same with the output from `backtrace`.

PR #2727, opened 2019-08-06
Serialize rdtsc instruction on x86_64

kubo39:
Unfortunately rdtsc can produce skewed results without mfence sometimes.
This is what GNU/Linux does in rdtsc_ordered function.

### DIFF
--- a/src/rt/trace.d
+++ b/src/rt/trace.d
@@ -862,6 +862,10 @@ else
             asm
             {
                 naked                   ;
+                // rdtsc can produce skewed results without preceding lfence/mfence.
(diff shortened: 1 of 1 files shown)


src/rt/trace.d
```diff
@@ -862,6 +862,9 @@ else
             asm
             {
                 naked                   ;
+                // rdtsc can produce skewed results without preceding lfence/mfence.
+                // this is what GNU/Linux does (see rdtsc_ordered), but only use mfence here.
```
thewilsonator: could you provide a link in the comment please.

PR #2734, opened 2019-08-15
Fix cas register usage

TurkeyMan:
I made a mistake in the last PR choosing the wrong registers.
My memory was faulty! (I thought EDI was a caller-save reg; that's how long it's been!)

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -628,6 +628,7 @@ else version (AsmX86_32)
 
             asm pure nothrow @nogc @trusted
             {
+                push EDI;
                 mov DL, writeThis;
                 mov EDI, ifThis;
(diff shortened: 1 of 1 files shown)


TurkeyMan:
A test that I respected the ABI properly?
The existing tests didn't run into any cases where calling code expected that register to be saves across the call, so it didn't break.

I have no idea how to test that... but I don't think it's necessary; the ABI won't ever change. This is correct, I just forgot which registers were in the volatile list, because it's been so many years since I wrote x86 asm.

TurkeyMan:
That said, while pondering on the best register assignment, I realised the existing code is terribly inefficient... I might do another pass later to select the best registers and minimise the shuffling.

PR #2737, opened 2019-08-16
UnicodeException.this() should be @nogc

WalterBright:


WalterBright:
Blocking https://github.com/dlang/phobos/pull/7145

PR #2740, opened 2019-08-18
Remove TypeInfo from GC memory block

JinShil:


rainers:
Maybe you are referring to https://github.com/dlang/druntime/pull/2513, but I don't think calling struct dtors from GC is expected to be deprecated, but the option to turn that off is.

JinShil:
The reason for this PR is this:

Some of us are trying to remove the dependency on runtime type information (a.k.a `TypeInfo`) in favor of compile-time type information by templating some of the runtime hooks.  As we progress with that endeavor it has become apparent that `TypeInfo` is being stored in the GC memory blocks.  While I was able to see [where the `TypeInfo` was being stored to the memory block](https://github.com/dlang/druntime/blob/e9cdc29346b232b16dcedb18b282eaf09a91e10e/src/rt/lifetime.d#L308), I was not able to figure out when and where that `TypeInfo` was being read from the memory block.  @rainers, if you know, please point it out for me.

I submitted this PR to see what would break if I removed the `TypeInfo` from the memory block in order to learn how things currently work.  @rainers if you have any advice for removing the dependency on `TypeInfo` from the GC, please share.

rainers:
> I was not able to figure out when and where that `TypeInfo` was being read from the memory block. @rainers, if you know, please point it out for me.

Structs with dtor and arrays of structs with dtor store the typeinfo at the end of the allocation (large array allocations use an unused slot of the 16 byte header). In addtion BlkAttr.STRUCTFINALIZE is passed with the memoryallocation.

The GC than calls 
https://github.com/dlang/druntime/blob/e9cdc29346b232b16dcedb18b282eaf09a91e10e/src/rt/lifetime.d#L1315
and
https://github.com/dlang/druntime/blob/e9cdc29346b232b16dcedb18b282eaf09a91e10e/src/rt/lifetime.d#L1362
via
https://github.com/dlang/druntime/blob/e9cdc29346b232b16dcedb18b282eaf09a91e10e/src/rt/lifetime.d#L1428
during a collection.

kinke:
The appended TypeInfo is used e.g. here: https://github.com/dlang/druntime/blob/90766698f65219f6691fffcfe1e306e82d3ad499/src/rt/lifetime.d#L1361-L1376

If it's really only used to destruct an instance, a pointer to the dtor function should work too. The actual dtor might have to be wrapped though, to account for ABI differences wrt. a method and a function pointer and `extern(D)` vs. `extern(C++)` dtors as well.

JinShil:
Instead of `ti.destroy(p)`, can't we just call `object.destroy(p)`?

kinke:
That needs a type. ;)

rainers:
> if you have any advice for removing the dependency on `TypeInfo` from the GC, please share.

For calling dtors on structs, we don't have to actually store the TypeInfo, an appropriate dtor function pointer could also work.

The TypeInfo passed to allocations is also used by the precise GC (reading the rtInfo). If you pass null instead, it will fall back to conservative scanning.

JinShil:
> For calling dtors on structs, we don't have to actually store the TypeInfo, an appropriate dtor function pointer could also work.

Could you perhaps share how I can obtain a pointer to the destructor without `TypeInfo`?

rainers:
You can save and pass around this lambda generated by the templated code:
```
    auto dg = (ref S s) { destroy(s); };
```
This would also defer proper ABI adjustments to the destroy function.

JinShil:
Thank you @kinke and @rainers.  I believe I have the information I was looking for.  Closing.

PR #2741, opened 2019-08-18
[dmd-cxx] Synchronize extern(C) bindings

ibuclaw:
Only parts of note are Musl and Darwin mach.loader.

### DIFF
--- a/src/core/sys/posix/pthread.d
+++ b/src/core/sys/posix/pthread.d
@@ -854,7 +854,15 @@ else version (CRuntime_Bionic)
 }
 else version (CRuntime_Musl)
 {
+    enum PTHREAD_BARRIER_SERIAL_THREAD = -1;
 
+    int pthread_barrier_destroy(pthread_barrier_t*);
+    int pthread_barrier_init(pthread_barrier_t*, in pthread_barrierattr_t*, uint);
+    int pthread_barrier_wait(pthread_barrier_t*);
+    int pthread_barrierattr_destroy(pthread_barrierattr_t*);
+    int pthread_barrierattr_getpshared(in pthread_barrierattr_t*, int*);
+    int pthread_barrierattr_init(pthread_barrierattr_t*);
+    int pthread_barrierattr_setpshared(pthread_barrierattr_t*, int);
 }
 else version (CRuntime_UClibc)
 {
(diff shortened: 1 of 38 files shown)


src/core/sys/posix/pthread.d
```diff
@@ -880,6 +880,53 @@ else
...
+else version (Solaris)
+{
+    int pthread_condattr_getclock(in pthread_condattr_t*, clockid_t*);
+    int pthread_condattr_setclock(pthread_condattr_t*, clockid_t);
+}
+else version (CRuntime_Bionic)
+{
+}
+else version (CRuntime_Musl)
+{
```
Geod24: Why is there no Musl definition ?
https://github.com/ifduyue/musl/blob/79f653c6bc2881dd6855299c908a442f56cb7c2b/include/pthread.h#L181-L183

ibuclaw: Fix it in master and I'll pull it in?

ibuclaw:
@Geod24 - rebased.

ibuclaw:
@Geod24 - Oh well... :-)

Geod24:
@ibuclaw : So, `backtrace` & co are not part of Musl, and I doubt they'll get integrated.
On Alpine, `libexecinfo` provides those symbols. Is there a good way to express this ? In my packages I just link `libexecinfo`, but open to suggestions.

ibuclaw:
@Geod24 - In my build, I can see that gcc's in-tree libbacktrace is built, and I can't see it disabled from the system installed gcc either, so I don't think that's a problem.

```
Configured with: /home/buildozer/aports/main/gcc/src/gcc-8.3.0/configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --build=x86_64-alpine-linux-musl --host=x86_64-alpine-linux-musl --target=x86_64-alpine-linux-musl --with-pkgversion='Alpine 8.3.0' --enable-checking=release --disable-fixed-point --disable-libstdcxx-pch --disable-multilib --disable-nls --disable-werror --disable-symvers --enable-__cxa_atexit --enable-default-pie --enable-default-ssp --enable-cloog-backend --enable-languages=c,c++,objc,fortran,ada --disable-libssp --disable-libmpx --disable-libmudflap --disable-libsanitizer --enable-shared --enable-threads --enable-tls --with-system-zlib --with-linker-hash-style=gnu
```

PR #2742, opened 2019-08-18
Move some traits to druntime

TurkeyMan:
I need these for core.atomic...

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -66,7 +66,7 @@ template Unqual(T)
 }
 
 // [For internal use]
-package template ModifyTypePreservingTQ(alias Modifier, T)
+template ModifyTypePreservingTQ(alias Modifier, T)
 {
          static if (is(T U ==          immutable U)) alias ModifyTypePreservingTQ =          immutable Modifier!U;
     else static if (is(T U == shared inout const U)) alias ModifyTypePreservingTQ = shared inout const Modifier!U;
@@ -77,6 +77,19 @@ package template ModifyTypePreservingTQ(alias Modifier, T)
     else static if (is(T U ==        inout       U)) alias ModifyTypePreservingTQ =              inout Modifier!U;
     else static if (is(T U ==              const U)) alias ModifyTypePreservingTQ =              const Modifier!U;
     else                                             alias ModifyTypePreservingTQ =                    Modifier!T;
+}
+@safe unittest
+{
+    alias Intify(T) = int;
+    static assert(is(ModifyTypePreservingTQ!(Intify,                    real) ==                    int));
+    static assert(is(ModifyTypePreservingTQ!(Intify,              const real) ==              const int));
+    static assert(is(ModifyTypePreservingTQ!(Intify,        inout       real) ==        inout       int));
+    static assert(is(ModifyTypePreservingTQ!(Intify,        inout const real) ==        inout const int));
+    static assert(is(ModifyTypePreservingTQ!(Intify, shared             real) == shared             int));
+    static assert(is(ModifyTypePreservingTQ!(Intify, shared       const real) == shared       const int));
+    static assert(is(ModifyTypePreservingTQ!(Intify, shared inout       real) == shared inout       int));
+    static assert(is(ModifyTypePreservingTQ!(Intify, shared inout const real) == shared inout const int));
+    static assert(is(ModifyTypePreservingTQ!(Intify,          immutable real) ==          immutable int));
 }
 
 // Substitute all `inout` qualifiers that appears in T to `const`


src/core/internal/traits.d
```diff
@@ -303,6 +305,110 @@ template hasElaborateAssign(S)
...
+        alias X = OriginalType!T;
+
+    static if (is(Unqual!X : E[], E) && !is(typeof({ enum n = X.length; })))
+        alias DynamicArrayTypeOf = X;
+    else
+        static assert(0, T.stringof ~ " is not a dynamic array");
+}
+
+private template AliasThisTypeOf(T)
+    if (isAggregateType!T)
```
lesderid: Shouldn't be indented.

src/core/internal/traits.d
```diff
@@ -303,6 +305,110 @@ template hasElaborateAssign(S)
...
+    alias members = __traits(getAliasThis, T);
+
+    static if (members.length == 1)
+        alias AliasThisTypeOf = typeof(__traits(getMember, T.init, members[0]));
+    else
+        static assert(0, T.stringof~" does not have alias this type");
+}
+
+template isFunctionPointer(T...)
+    if (T.length == 1)
```
lesderid: ditto

src/core/internal/traits.d
```diff
@@ -303,6 +305,110 @@ template hasElaborateAssign(S)
...
+            enum bool isFunctionPointer = true;
+        else
+            enum bool isFunctionPointer = false;
+    }
+    else
+        enum bool isFunctionPointer = false;
+}
+
+template isDelegate(T...)
+    if (T.length == 1)
```
lesderid: ditto

src/core/internal/traits.d
```diff
@@ -8,23 +8,22 @@
...
 
-/// taken from std.typetuple.TypeTuple
-template TypeTuple(TList...)
-{
-    alias TypeTuple = TList;
-}
-alias AliasSeq = TypeTuple;
 
-template FieldTypeTuple(T)
+// TODO: deprecate these old names...?
```
lesderid: Can't we just remove them? It's `core.internal` so should be fine.

JinShil:
If these came from Phobos, I think the unittests should be moved as well.  A followup PR to Phobos to import them from druntime (and remove them from Phobos) would also be prudent IMO.

lesderid:
> If these came from Phobos, I think the unittests should be moved as well. A followup PR to Phobos to import them from druntime (and remove them from Phobos) would also be prudent IMO.

Other traits that were moved to druntime still have their unittests in Phobos because then they can still be used as examples (documentation isn't generated for core.internal).

src/core/internal/traits.d
TurkeyMan: Really? Every constraint I've seen has been intended... that's how you know it's a constraint.

lesderid: [Yes](https://dlang.org/dstyle.html#phobos_declarations), "Constraints on declarations should have the same indentation level as their declaration".

TurkeyMan:
>  A followup PR to Phobos to import them from druntime (and remove them from Phobos) would also be prudent IMO.

I have a follow-up, but it won't build until this is merged... so I haven't pushed it.

src/core/internal/traits.d
TurkeyMan: That's a large project.

src/core/internal/traits.d
TurkeyMan: 1. that's horrible.
2. it's already a 50/50 split in druntime. i think it actually leans in favour of this way, which is why i went with this way.

src/core/internal/traits.d
lesderid: `FieldTypeTuple` is only used in this file. `TypeTuple` is used elsewhere in druntime too, but I can make a follow-up PR.

src/core/internal/traits.d
lesderid: AIUI, it's only split because the style guide was written/followed after a lot of the code had already been already written. I'm not sure I agree with this rule either, but it's required by the style guide...

src/core/internal/traits.d
TurkeyMan: Go for it!

src/core/internal/traits.d
TurkeyMan: Well someone can produce a PR that fixes all the code in one hit.

TurkeyMan:
Why has buildkite gone on holiday? All PR's are not building...

lesderid:
> Why has buildkite gone on holiday? All PR's are not building...

Looks like the CI agent went offline, I can't ping the IP and Buildkite says 'lost connection'.

wilzbach:
The CI agents are continuously spawned based on demand. Looks like there's a problem with that logic currently, but it should very likely recover over the next the hours. Not blocking the approval of this PR.

TurkeyMan:
> Their unittests should be moved too, because then Phobos can be a mere alias (and public example), but it's a bit tricky if we separate implementation from their tests.

Documented tests stay in phobos, and the tests are documented tests.
We do this for all the other tests.

We would need to do a pass to move all the docs over to druntime, which would mean moving `core.internal.traits` to `core.traits`... that's way outside the scope of this PR, I'm not changing any organisation here.

JinShil:
I think only those unittests with a `///` DDoc header are used for documentation (@wilzbach, please correct me if I'm wrong).  Can we at least move the unittests which are not used for documentation?  For example:  https://github.com/dlang/phobos/blob/81572ffce650cb17319c63927cffa6f1e345a094/std/traits.d#L7543-L7555

TurkeyMan:
@JinShil That is duplicated in druntime and phobos (because not public)... are you saying the unit test should exist in both places?

JinShil:
> are you saying the unit test should exist in both places?

With regard to `ModifyTypePreservingTQ`, it is `package`. and since these traits are being moved to `core.internal`, I would expect the entire thing (unittests and implementation) to be moved to druntime and made `public`.  Then it can be package-imported in Phobos.  e.g.

```D
package import core.internal : ModifyTypePreservingTQ;
```
I don't know if `package import` is a thing though; I'll have to try it out.

JinShil:
> I don't know if package import is a thing though; I'll have to try it out.

Even if `package import` is not legal, you can still make `ModifyTypePreservingTQ` public in druntime and then make a `package` wrapper like you did with some of the other imports in https://github.com/dlang/phobos/pull/7148

TurkeyMan:
Oh FFS!! See what happens! I move the tests over and they depend on 90% of phobos...
This happens every time! That's why I don't do it.

This is blocking me, one of you can fix this. Please be fast...

TurkeyMan:
I want to start getting paid for the time wasted by bullshit like this. I have finite lifetime, and it's worth $100/hr.

TurkeyMan:
Should merge this; everyone got bent-out-of-shape about the tests... and from here I can push the phobos PR...

TurkeyMan:
Surely this is trivial?

PR #2745, opened 2019-08-20
Renovate core.atomic, improve and simplify the constraints on functions

TurkeyMan:
Add missing features to the API.
Most functions are slightly more efficient.
Concentrate more specific work into less surface area.

Fixes Issue 20107 - core.atomic : Memory order is missing keys
Fixes Issue 20106 - core.atomic : atomicFence doesn't accept MemoryOrder
Fixes Issue 20105 - core.atomic 'cas' function is incomplete
Fixes Issue 18643 - Compiling error when combining CAS and numeric literal.
Fixes Issue 15007 - core.atomic match C++11
Fixes Issue 8831 - core.atomic: add compare-and-swap function with other result type

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -276,7 +276,7 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
  * Returns:
  *  true if the store occurred, false if not.
  */
-bool cas(T,V1,V2)(T* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
+bool cas(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V1,V2)(T* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
     if (!is(T == shared S, S) && is(T : V1))
 in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 {
@@ -287,14 +287,14 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
     static if (__traits(isFloating, T))
     {
         alias IntTy = IntForFloat!T;
-        return atomicCompareExchangeStrongNoResult(cast(IntTy*)here, *cast(IntTy*)&arg1, *cast(IntTy*)&arg2);
+        return atomicCompareExchangeStrongNoResult!(succ, fail)(cast(IntTy*)here, *cast(IntTy*)&arg1, *cast(IntTy*)&arg2);
     }
     else
-        return atomicCompareExchangeStrongNoResult(here, arg1, arg2);
+        return atomicCompareExchangeStrongNoResult!(succ, fail)(here, arg1, arg2);
 }
 
 /// Ditto
-bool cas(T,V1,V2)(shared(T)* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
+bool cas(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V1,V2)(shared(T)* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
     if (!is(T == class) && (is(T : V1) || is(shared T : V1)))
 in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 {
@@ -310,15 +310,15 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
         static assert(!hasUnsharedIndirections!V2, "Copying `" ~ V2.stringof ~ "* writeThis` to `" ~ shared(T).stringof ~ "* here` would violate shared.");
         alias Thunk2 = V2;
     }
-    return cas(cast(T*)here, *cast(Thunk1*)&ifThis, *cast(Thunk2*)&writeThis);
+    return cas!(succ, fail)(cast(T*)here, *cast(Thunk1*)&ifThis, *cast(Thunk2*)&writeThis);
 }
 
 /// Ditto
-bool cas(T,V1,V2)(shared(T)* here, shared(V1) ifThis, shared(V2) writeThis) pure nothrow @nogc @trusted
+bool cas(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V1,V2)(shared(T)* here, shared(V1) ifThis, shared(V2) writeThis) pure nothrow @nogc @trusted
     if (is(T == class))
 in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 {
-    return atomicCompareExchangeStrongNoResult(cast(T*)here, cast(V1)ifThis, cast(V2)writeThis);
+    return atomicCompareExchangeStrongNoResult!(succ, fail)(cast(T*)here, cast(V1)ifThis, cast(V2)writeThis);
 }
 
 /**
@@ -335,7 +335,7 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
  * Returns:
  *  true if the store occurred, false if not.
  */
-bool cas(T,V)(T* here, T* ifThis, V writeThis) pure nothrow @nogc @trusted
+bool cas(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V)(T* here, T* ifThis, V writeThis) pure nothrow @nogc @trusted
     if (!is(T == shared S, S) && !is(V == shared U, U))
 in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 {
@@ -345,14 +345,14 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
     static if (__traits(isFloating, T))
     {
         alias IntTy = IntForFloat!T;
-        return atomicCompareExchangeStrong(cast(IntTy*)here, cast(IntTy*)ifThis, *cast(IntTy*)&writeThis);
+        return atomicCompareExchangeStrong!(succ, fail)(cast(IntTy*)here, cast(IntTy*)ifThis, *cast(IntTy*)&writeThis);
     }
     else
-        return atomicCompareExchangeStrong(here, ifThis, writeThis);
+        return atomicCompareExchangeStrong!(succ, fail)(here, ifThis, writeThis);
 }
 
 /// Ditto
-bool cas(T,V1,V2)(shared(T)* here, V1* ifThis, V2 writeThis) pure nothrow @nogc @trusted
+bool cas(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V1,V2)(shared(T)* here, V1* ifThis, V2 writeThis) pure nothrow @nogc @trusted
     if (!is(T == class) && (is(T : V1) || is(shared T : V1)))
 in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 {
@@ -373,15 +373,15 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
         alias Thunk2 = V2;
     }
     static assert (is(T : Thunk1), "Mismatching types for `here` and `ifThis`: `" ~ shared(T).stringof ~ "` and `" ~ V1.stringof ~ "`.");
-    return cas(cast(T*)here, cast(Thunk1*)ifThis, *cast(Thunk2*)&writeThis);
+    return cas!(succ, fail)(cast(T*)here, cast(Thunk1*)ifThis, *cast(Thunk2*)&writeThis);
 }
 
 /// Ditto
-bool cas(T,V)(shared(T)* here, shared(T)* ifThis, shared(V) writeThis) pure nothrow @nogc @trusted
+bool cas(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V)(shared(T)* here, shared(T)* ifThis, shared(V) writeThis) pure nothrow @nogc @trusted
     if (is(T == class))
 in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 {
-    return atomicCompareExchangeStrong(cast(T*)here, cast(T*)ifThis, cast(V)writeThis);
+    return atomicCompareExchangeStrong!(succ, fail)(cast(T*)here, cast(T*)ifThis, cast(V)writeThis);
 }
 
 /**
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -385,6 +385,133 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 }
 
 /**
+* Stores 'writeThis' to the memory referenced by 'here' if the value
+* referenced by 'here' is equal to 'ifThis'.
+* The 'weak' version of cas may spuriously fail. It is recommended to
+* use `casWeak` only when `cas` would be used in a loop.
+* This operation is both
+* lock-free and atomic.
+*
+* Params:
+*  here      = The address of the destination variable.
+*  writeThis = The value to store.
+*  ifThis    = The comparison value.
+*
+* Returns:
+*  true if the store occurred, false if not.
+*/
+bool casWeak(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V1,V2)(T* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
+    if (!is(T == shared S, S) && is(T : V1))
+in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
+{
+    // resolve implicit conversions
+    T arg1 = ifThis;
+    T arg2 = writeThis;
+
+    static if (__traits(isFloating, T))
+    {
+        alias IntTy = IntForFloat!T;
+        return atomicCompareExchangeWeakNoResult!(succ, fail)(cast(IntTy*)here, *cast(IntTy*)&arg1, *cast(IntTy*)&arg2);
+    }
+    else
+        return atomicCompareExchangeWeakNoResult!(succ, fail)(here, arg1, arg2);
+}
+
+/// Ditto
+bool casWeak(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V1,V2)(shared(T)* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
+    if (!is(T == class) && (is(T : V1) || is(shared T : V1)))
+in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
+{
+    static if (is (V1 == shared U1, U1))
+        alias Thunk1 = U1;
+    else
+        alias Thunk1 = V1;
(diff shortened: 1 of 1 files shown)


TurkeyMan:
That buildkite error is problematic... it's in master now.

TurkeyMan:
See, the old `atomicLoad` accepted a `shared const T`, and returned a `T`, which means the const was lost through the `atomicLoad` call... which is REALLY bad.
I fixed that... and now this issue where vibe.d is trying to assign to an atomic thing shows up.

I think this is acceptable breakage, because this fixes a heinous const violation bug.

wilzbach:
> I don't know how that ever worked; but it looks like I've broken it.

Yes -> https://github.com/dlang/druntime/pull/2739#issuecomment-522862487

> I think this is acceptable breakage, because this fixes a heinous const violation bug.

You just broke ten of the top 50 dub packages. It would have been okay if Vibe.d was fixed __before__.

TurkeyMan:
I can hack the bug back in there... but it's a terrible bug, no matter how you slice it.

TurkeyMan:
Also, I didn't break it, I don't have merge rights!

Geod24:
> I can hack the bug back in there... but it's a terrible bug, no matter how you slice it.

Surely is a terrible code, but breaking the ability to compile older Vibe.d with newer release is terrible for the ecosystem.

TurkeyMan:
https://github.com/vibe-d/vibe.d/issues/2348

I'll give them a day or 2 to respond, or I'll commit the bug back in there.
We missed 2.088, so we're months away from this code being released.

TurkeyMan:
This PR fixes 5 issues, why does dlang-bot only mention one?

TurkeyMan:
> compile older Vibe.d with newer release

What do you suggest? I don't think tremendously bad bugs in druntime are acceptable... :/

wilzbach:
Dlang-bot only looks at commit messages.

TurkeyMan:
All 5 are in the commit message.

wilzbach:
> we're months away from this code being released.

The CI should never be red though :/

TurkeyMan:
> The CI should never be red though :/

I can hack the bug back in there... although I reckon they'll fix it in vibe.d by tomorrow.

TurkeyMan:
Why doesn't dlang-bot see the issues that are fixed? They won't appear in the changelog unless they're referenced here...

TurkeyMan:
Oh well, I guess the changelog is just missing some great info!

Why would dlang-bot only look at the commit messages? He should look at the PR description... I always forget to add text to the commit message, and then I need to cause a full rebuild of the CI to amend the message... that's terribly wasteful.

wilzbach:
> Why would dlang-bot only look at the commit messages? He should look at the PR description

Because _only_ the commit messages are available when you look at the git log between two releases.

> although I reckon they'll fix it in vibe.d by tomorrow.

No, I doubt it will get fixed within the two months. Vibe.d isn't really maintained + you need to release the change in Vibe.d and then upgrade all libraries/apps that use it + release them.

wilzbach:
> Why doesn't dlang-bot see the issues that are fixed? 

Looks like it only looks at the first "fixes" string. However, there's this:

https://github.com/dlang/dlang-bot#referencing-multiple-issues

TurkeyMan:
> No, I doubt it will get fixed within the two months.

So, should I put the bug back? Should we give it a short while and put it back later?
I think anyone of you who can build vibe.d could make the PR in about 3 minutes...

wilzbach:
> So, should I put the bug back?

Unfortunately, you have to :/ 

> I think anyone of you who can build vibe.d could make the PR in about 3 minutes...

Yes, but you also need to release a new version which only happens every 2-3 months and afterwards update all the other projects that depend on Vibe.d.
So I guess the only thing we can do is revert it for now, and open a revert of that revert, s.t. it's not forgotten and can be merged once Vibe.d is updated.

TurkeyMan:
It is done. Someone should make a PR for vibe.d...

aG0aep6G:
> See, the old `atomicLoad` accepted a `shared const T`, and returned a `T`, which means the const was lost through the `atomicLoad` call... which is REALLY bad.

It's not bad at all, as long as:

* `atomicLoad` makes a copy. It does make a copy, right?
* `T` isn't a class or struct. `TailShared` was checking for this case.

TurkeyMan:
Right, if `T` has any indirection.
- is a pointer
- is a slice
- is a delegate
- is a struct with any of those things

How could TailShared check for that? `TailShared`T` can't know the source was const... so it would have no reason to believe a contained indirection should remain const?

TurkeyMan:
https://github.com/vibe-d/vibe.d/pull/2349

aG0aep6G:
> Right, if `T` has any indirection.
> 
>     * is a pointer

Not a problem. `T` becomes `const(Foo)*`.

>     * is a slice

Not a problem. `T` becomes `const(Foo)[]`.

>     * is a delegate

I'm not sure about this one. Does `const` on a delegate mean anything beyond "can't reassign it"?

>     * is a struct with any of those things
> 
> 
> How could TailShared check for that? `TailShared!T` can't know the source was const... so it would have no reason to believe a contained indirection should remain const?

You're right about structs. I was thinking of `shared`, not `const`. `TailShared` does nothing here, and it is a `const` violation.

TurkeyMan:
> Not a problem. `T` becomes `const(Foo)*`.

Huh?

```d
        const(int*) fun();
        auto y = fun();
        pragma(msg, typeof(y));

 > const(int*)
```

Same for the others... are you sure?

WalterBright:
The buildkite/dmd error:
```
```
The code https://github.com/vibe-d/vibe.d/blob/master/crypto/vibe/crypto/cryptorand.d
```
auto p = atomicLoad(*cast(const shared GET_RANDOM*) &hasGetRandom);
if (p == GET_RANDOM.UNINITIALIZED)
{
	p = initHasGetRandom() ? GET_RANDOM.AVAILABLE   <= error on this line
					: GET_RANDOM.NOT_AVAILABLE;
```

TurkeyMan:
Thanks @WalterBright ;)
Bit late... we've been discussing this for hours, and I fixed it already.

aG0aep6G:
> > Not a problem. `T` becomes `const(Foo)*`.
> 
> Huh?

I mean this:

```d
T f(T)(const T thing) { return thing; }
const int* p;
auto x = f(p);
pragma(msg, typeof(x)); /* const(int)* */
```

TurkeyMan:
How does that work O_o ... `T` should be inferred to `int*` no? How does `T` keep the `const`?

aG0aep6G:
> How does that work O_o ... `T` should be inferred to `int*` no? How does `T` keep the `const`?

`const(int*)` is the same as `const(const(int)*)`. Removing `const` from that type leaves `const(int)*`. Working as intended.

TurkeyMan:
Oh yeah!!
So obvious! :P
Sorry, that's a massive derp moment.

TurkeyMan:
@thewilsonator Should probably prioritise this patch to un-break the CI ;)

UplinkCoder:
All the formatting changes make it hard to read the diff;
any chance you can commit a version without the white-space changes?

TurkeyMan:
That's a lot of work >_<

lesderid:
If you're using the GitHub diff view, you can hide lines that only have whitespace changes: https://github.com/dlang/druntime/pull/2745/files?w=1.

CyberShadow:
The longer master remains broken, the more easier it is for more breaking changes to slip in, and the longer the resulting mess will take to figure out. We should not be fixing master breakage due to a non-trivial PR with another non-trivial PR. Can we revert #2739? Then it can be folded into this PR and we can ensure that the full result is not breaking.

Also, the first commit in this PR is currently a mess and should have been several commits.

TurkeyMan:
Sure, do what you like. I didn't merge the last one; I don't have merge rights.

But I've spend weeks on this and if I were billing D, something like $3000-4000... what's broken is broken, this is the final destination, so it needs to be reviewed regardless.

You're right, I shouldn't have made the whitespace change; I was really struggling to maintain the original file's unusual extra-spacing style, so I made it consistent with the rest of druntime.
Personally, I can disable whitespace diff's in my diff view (including github) and it just doesn't bother me.

CyberShadow:
> Sure, do what you like. I didn't merge the last one; I don't have merge rights.

As you are familiar with the code, you have the opportunity to take initiative and help fix the situation (in a way that also doesn't obstruct your work too much), but I guess you have no interest in such things?

> But I've spend weeks on this and if I were billing D, something like $3000-4000...

Then, you must be a very good salesman.

Please get over yourself. Almost all contributors to D do it voluntarily in their free time and work together to create a better project. Money has nothing to do with any of it; putting a monetary value on your contribution is frankly insulting to those who do it in their free time.

> this is the final destination, so it needs to be reviewed regardless.

I'm not sure how you meant it, but one interpretation of that is disrespectful to reviewers and future contributors who will need to read and update code affected by your changes.

> You're right, I shouldn't have made the whitespace change; I was really struggling to maintain the original file's unusual extra-spacing style, so I made it consistent with the rest of druntime.

No problem with that; but, it is much easier to review changes when they are split into minimal commits. In case of whitespace, you could either fix the style first in one commit then use the correct style in new code, or add code using whatever style then fix the style in a separate commit. Ideally, each commit should be a minimal change (before and after which the project still works) constrained to a single matter.

You specified the first commit to fix 6 bugs. This is abnormal. Ideally, there ought to be either 6 commits each fixing one bug, or all but one of those bugs should be closed as duplicates.

TurkeyMan:
> As you are familiar with the code, you have the opportunity to take initiative and help fix the situation (in a way that also doesn't obstruct your work too much), but I guess you have no interest in such things?

I'm not sure what you mean... This PR fixes it, and it's all green. It's been all green for days. I spent a lot of time working on this, so to say I have no interest in fixing it is very unfair. I worked to fix it immediately, and it's sitting here.

> putting a monetary value on your contribution is frankly insulting to those who do it in their free time.

I removed that comment immediately, and I apologies for it to the reasonable extent that it was insulting; but what's frustrating to me is trying to justify my commitment to the community to myself repeatedly for so long. My energy here DOES have a monetary value (as is true for every contributor) whether it's politically correct to recognise that or not.
What annoyed me is the idea that now I need to spend another evening wasting time doing busy work when you can hide whitespace changes in the diff-view. This PR (which fixes the issue in question) was a lot of work, and has been sitting here for a couple of days since I finished it while people are upset that the build is broken.
It can be reviewed and merged: build is not broken. I can spend my night teasing it apart into bits... but I need to find time for that.

We just waste so much time here, and it makes me unreasonably distressed.

> > this is the final destination, so it needs to be reviewed regardless.
>
> I'm not sure how you meant it, but one interpretation of that is disrespectful to reviewers and future contributors who will need to read and update code affected by your changes.

I don't know what you mean, I'm just saying I'm done with this PR. I don't have any more changes planned, I think this is fully baked.

> You specified the first commit to fix 6 bugs. This is abnormal. Ideally, there ought to be either 6 commits each fixing one bug, or all but one of those bugs should be closed as duplicates.

They're definitely not duplicates, just that a lot of issues are resolved by tidying and fixing the API in general.
I agree it would be better to fix each function one at a time, but that's not what I did. Is it necessary that I burn a few more hours trying to tease it apart?

CyberShadow:
> I removed that comment immediately

OK, I removed the replies too. Sorry if I misunderstood anything.

> I worked to fix it immediately, and it's sitting here.

I understand how you feel. Some of my PRs stayed open for over a year, and many never land at all. This is the nature of open-source development. Reviewing others' contributions is a thankless chore, so you often have to make an extra effort to make changes as easy as possible to review.

> Is it necessary that I burn a few more hours trying to tease it apart?

I think it would certainly be appreciated! Consider the possibility that this PR introduces a regression. Finding exactly which part of a change in a big PR introduced it is much easier when the changes are split into many commits.

TurkeyMan:
> so you often have to make an extra effort to make changes as easy as possible to review.

I knew I fucked that up the moment I got there, but the damage was done, and I haven't had time (or will) to spend a few more hours splitting it apart.
It's also counter-productive at a number of levels... having N small PR's merged takes N review cycles, and sitting and waiting for that which might block for a day becomes a blocker for a week.

I don't know how to be more efficient here; by I can and do game it to increase efficiency in whatever ways I can affect that. Is has been my experience that merging one PR is easier than merging 5, and to that extend, there's an obvious pressure to keep doing that. I understand the ways that that's bad, but the path of least resistance will virtually always win.

We should just all meet up somewhere more frequently and get shit done in the same room.

> Consider the possibility that this PR introduces a regression.

It probably will; I changed some subtle broken things. It exposed a bug in vibe.d for instance... which is a really bad bug, and it pains me to revert the fix.

CyberShadow:
> It's also counter-productive at a number of levels... having N small PR's merged takes N review cycles, and sitting and waiting for that which might block for a day becomes a blocker for a week.

I agree with you here, which is why I suggested splitting it into commits, not PRs. However, Walter historically prefers them spread even over separate PRs. One advantage of using PRs is that if you agree with all changes up to some point, you can merge just those. Per-commit reviews is also how software forges used on larger scales seem to work (Phabricator and Gerrit).

TurkeyMan:
Another problem with many PR's, is that each one merged re-runs CI on the others which takes hours each time.

Geod24:
> What do you suggest? I don't think tremendously bad bugs in druntime are acceptable... :/

According to https://github.com/dlang/druntime/pull/2745#issuecomment-522898187 the "tremendous" part was quite exaggerated.

As mentioned in the Vibe.d PR, the bug you fixed didn't have any impact on the user. But fixing it means that:
- Other projects not depending on Vibe.d could use this pattern and be broken by the change
- People updating to newer compilers directly loose the ability to bisect on Vibe.d
- We force anyone updating to update to the newest compiler to update their Vibe.d dependency
- We potentially force any library depending on Vibe.d to release a new version

Don't get me wrong, this needs to be fixed eventually, but fixing it only makes life worse for users of druntime at the moment, not better. So having everything fixed beforehand and giving it at least a release to propagate is, I think, not too much to ask for.

TurkeyMan:
> According to #2745 (comment) the "tremendous" part was quite exaggerated.

Well, it's just applicable to `class` and `struct`, but not to `int*` or `int[]` as I initially thought... so the bad-ness is identical, it just affects less things that I initially thought.

I'm not good with a function that silently casts const away from a struct.

> So having everything fixed beforehand and giving it at least a release to propagate is, I think, not too much to ask for.

I buy your arguments, which is why I reintroduced it to this PR. It is also in isolation here: https://github.com/dlang/druntime/pull/2753

TurkeyMan:
I separated the whitespace change and casWeak into separate patches.

TurkeyMan:
And split it some more for good measure.

n8sh:
> I think this is acceptable breakage, because this fixes a heinous const violation bug.

The old behavior was not a bug because it only could occur for types without indirections. It is legal to convert `const int` to `int`. You almost never want to return a `const` scalar because it causes problems with `auto`, like happened to the code that broke in vibe.d.

aG0aep6G:
> The old behavior was not a bug

There is a bug. To clear up the confusion (to which I contributed earlier), here is an example:

```d
import core.atomic;
struct S { int* p; }
void main()
{
    immutable int* p = new immutable int(42);
    auto s = immutable S(p);
    auto x = atomicLoad(s);
    pragma(msg, typeof(x.p)); /* shared(int)* - uh oh */
    *x.p = 13; /* mutating an immutable int */
}
```

The problem is that `atomicLoad` strips away const from structs and classes. There is no issue with scalars, pointers, arrays. It's just structs and classes.

TurkeyMan:
Is there something wrong with this?

src/core/atomic.d
```diff
@@ -113,28 +113,30 @@ TailShared!T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)(ref const shared T
  *  newval = The value to store.
  */
 void atomicStore(MemoryOrder ms = MemoryOrder.seq, T, V)(ref T val, V newval) pure nothrow @nogc @trusted
-    if (__traits(compiles, { val = newval; }) && !is(T == shared S, S) && !is(V == shared U, U))
+    if (!is(T == shared S, S) && !is(V == shared U, U))
```
lesderid: Couldn't this just be `if (!is(T == shared) && !is(V == shared))`?

src/core/atomic.d
```diff
@@ -539,9 +666,9 @@ private
     }
 
     // TODO: it'd be nice if we had @trusted scopes; we could remove this...
```
lesderid: Could be replaced with a `@trusted` lambda (something like `(() @trusted => casWeak(&val, &get, set))()`)? Not sure if that's better here though.

src/core/atomic.d
TurkeyMan: Oh yeah... I guess so. I didn't think of that ;)

src/core/atomic.d
TurkeyMan: Yeah, possibly the lamest pattern D ever manifested. I don't want to put another another another function call around an asm opcode. There's already 3-4 more than enough.

src/core/atomic.d
```diff
@@ -112,39 +112,41 @@ TailShared!T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)( ref const shared T
...
  *  newval = The value to store.
  */
-void atomicStore(MemoryOrder ms = MemoryOrder.seq, T, V)( ref T val, V newval ) pure nothrow @nogc @trusted
-    if ( __traits( compiles, { val = newval; } ) && !is(T == shared S, S) && !is(V == shared U, U) )
+void atomicStore(MemoryOrder ms = MemoryOrder.seq, T, V)(ref T val, V newval) pure nothrow @nogc @trusted
+    if (!is(T == shared S, S) && !is(V == shared U, U))
 {
-    static if ( __traits(isFloating, T) )
+    // resolve implicit conversions
+    T arg = newval;
```
dnadlinger: Hmm, doesn't this end up `@trust`ing potentially arbitrary user code?

TurkeyMan: That's an interesting point.
It's invalid for any type that interacts with atomic to have non-trivial copy anyway... so I can assert that in the constraint and then we can have confidence this is safe.
I actually thought about that while doing this (not for this exact reason), but it's stuck on https://issues.dlang.org/show_bug.cgi?id=19902, which is a really bad regression, and it's been sitting there for months!
I figured I'd add that restriction whenever this is solved.

TurkeyMan: I put the assert.
I'm not sure if the same assert should be everywhere though; `atomicStore` is definitely a copying assignment, whereas exchange and cas are both actually swap operations, which are move operations, so they don't actually interact with copy semantics.

TurkeyMan:
2 weeks feels like a long time to twiddle my thumbs... I could be programming, but I'm playing video games instead ;)

src/core/atomic.d
kinke: @TurkeyMan: This line causes a regression for:
```
struct S
{
    int x;
    alias x this;
}
shared S s;
s = 0x1234_5678;
atomicStore(s, 0); // => Error: cannot implicitly convert expression `newval` of type `int` to `S`
assert(s.x == 0);
```

Note that the removed `__traits( compiles, { val = newval; } )` constraint doesn't play a role due to `alias this` magic, i.e., assignment is fine, but construction via implicit conversion fails. An explicit cast would work.

TurkeyMan: interesting.
so you're saying the semantic difference is that i changed a copy to a construction?

is this real code? does this occur somewhere?

kinke: It's only an additional LDC test (https://github.com/ldc-developers/druntime/blob/77066d7d399b0da1e94699da37436066589898ef/src/core/atomic.d#L1073-L1081), so I hope code like this isn't in the wild.

I'm not sure how to handle it best, i.e., whether such code should be disallowed or whether an explicit cast in `atomicStore` is justified.

PR #2750, opened 2019-08-21
[Musl] Remove duplicate definition of SOCK_RAW

ibuclaw:
As it was added in the proper location in an earlier PR.


Geod24:
Why are the raw socket definition separate ?

ibuclaw:
According to IEEE Std 1003.1-2017, SOCK_RAW is an optional, and anextension to the ISO C standard.

ibuclaw:
I'll merge as the dmd-cxx branch already has this.

PR #2752, opened 2019-08-21
Revert "Extract atomic platform specific implementation into an implementation file"

CyberShadow:
Reverts dlang/druntime#2739

#2739 contains a bugfix that happened to be a breaking change. The breaking change affects vibe.d and many projects using vibe.d, thus breaking the Buildkite autotester.

A follow-up PR, #2745, contains a change which reverts the bugfix, thus unbreaking VIbe.d. However, it looks like that will take a while to fully review. 

For now, we can revert #2739 to unbreak master. The submitter [seems to agree](https://github.com/dlang/druntime/pull/2745#issuecomment-523572533) with this approach.

CC @TurkeyMan


dnadlinger:
Yep; first order of business should always be to unbreak the CI, since that brings all progress globally.

TurkeyMan:
Why is buildkite still red?

Alternatively, I could just reintroduce the bug in isolation... that's a 3-line patch.

TurkeyMan:
Alternative approach: https://github.com/dlang/druntime/pull/2753

I don't care which one is chosen.

CyberShadow:
> Why is buildkite still red?

Uh-oh.

It looks like a different error.

master fails with:

master + this PR fails with:

    ../../../.dub/packages/eventcore-0.8.44/eventcore/source/eventcore/drivers/posix/processes.d(305,25): Error: undefined identifier `intptr_t`

> Alternatively, I could just reintroduce the bug in isolation... that's a 3-line patch.

If it's not too much trouble.

TurkeyMan:
Link in comment above.

Geod24:
> Why is buildkite still red?

A new release of eventcore was tagged yesterday and that release is broken. It's already fixed: https://github.com/vibe-d/eventcore/pull/120
Now @s-ludwig just needs to tag a new release of eventcore.

CyberShadow:
> A new release of eventcore was tagged yesterday and that release is broken.

How did that happen? Is it broken in all D versions? Or is it broken only with D master (meaning, the new release exposed a regression in master)?

TurkeyMan:
I also pushed a fix to vibe.d which corrects against the bug in DMD which I fixed, and that's just waiting for their CI to flush... so we could not do anything, and it will resolve when vibe.d merges the fix... too many choices.

Geod24:
> it will resolve when vibe.d merges the fix

No, it also needs a new release for Buildkite to pick it up.
And as mentioned in [my other comment](https://github.com/dlang/druntime/pull/2745#issuecomment-523629648), it transitively breaks a lot of projects. Libraries could get the update for free, but apps that commit `dub.selections.json` for example will need a manual update.

> How did that happen? Is it broken in all D versions?

I don't know, and asked the same question. All D versions AFAICS.

@wilzbach (and `@others`): Would now be the time to make buildkite `required` ?
Since it's not the first time it is flat out ignored when relevant.

wilzbach:
Yeah I will set it as required once it's green again.

CyberShadow:
Merged #2753.

I understand that the remaining failures should be cleared up once the eventcore bump gets propagated to Vibe.d.

PR #2753, opened 2019-08-21
Reintroduce const violation bug to fix vibe.d build failure

TurkeyMan:
This should fix the vibe.d build failure.

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -66,7 +66,7 @@ enum MemoryOrder
  * Returns:
  *  The value of 'val'.
  */
-T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)( ref T val ) pure nothrow @nogc @trusted
+T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)( ref const T val ) pure nothrow @nogc @trusted
     if ( !is( T == shared U, U ) && !is( T == shared inout U, U ) && !is( T == shared const U, U ) )
 {
     static if ( __traits(isFloating, T) )
@@ -76,11 +76,11 @@ T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)( ref T val ) pure nothrow @nog
         return *cast(T*)&r;
     }
     else
-        return core.internal.atomic.atomicLoad!ms(&val);
+        return core.internal.atomic.atomicLoad!ms(cast(T*)&val);
 }
 
 /// Ditto
-T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)( ref shared T val ) pure nothrow @nogc @trusted
+T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)( ref const shared T val ) pure nothrow @nogc @trusted
     if ( !hasUnsharedIndirections!T )
 {
     import core.internal.traits : hasUnsharedIndirections;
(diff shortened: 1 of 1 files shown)


CyberShadow:
This has fewer failures on BuildKite than master, so, I'm going to merge this.

TurkeyMan:
The vibe.d issue that this was intended to resolve is resolved. That other one is unrelated to anything I've done.

n8sh:
Noting that the title of this PR is incorrect. It _is not a bug_ to return `int` rather than `const int` and is to be preferred https://github.com/dlang/phobos/pull/6682#discussion_r213024113:
>That ensures no surprise with functions that initialize an auto with the result of the call.

CyberShadow:
Thanks for pointing that out! So, this is actually a regression fix affecting value types.

TurkeyMan:
@n8sh Supplying a const ref to a function, and returning it with const stripped away is a bug, and it's a bad one.

CyberShadow:
> Supplying a const ref to a function, and returning it with const stripped away is a bug

If it returns by non-ref, then it should be perfectly fine to strip head const.

> `T` may have indirections.

So, don't return `T`, return `T` with head const stripped. That should both fix the bug and prevent a regression for pure value types.

TurkeyMan:
> If it returns by non-ref, then it should be perfectly fine to strip head const.

Stripping head-const is not valid for struct's or classes.
This code strips const completely.

CyberShadow:
> Stripping head-const is not valid for struct's or classes.

Yes, it is impossible to have tail-const in these cases. So, for structs or classes, it should return `T` as-is.

TurkeyMan:
That's possible... but my general feeling is that so many 'cases' are bad. It makes meta way more complicated when it decides to do a special thing in one case and not some other...
Someone will write their own meta that assumes `atomicLoad` strips const from the return, and then we just cascade the issue into their meta when they encounter `T == struct`.

CyberShadow:
> That's possible... but my general feeling is that so many 'cases' are bad.

That really isn't anything new, it's a consequence of D's type system being generally very complicated. Have a look at the implementations of some templates in std.traits and std.typecons. This is why we encapsulate the complexity into a template which does a single thing and is documented to do this certain thing, then use it throughout. (Actually a bit surprised there isn't already something like that in Phobos, but maybe I missed it.)

> Someone will write their own meta that assumes `atomicLoad` strips const from the return

I don't really see why this is an issue. It will result in a compilation error, and they will have to fix the code if they want it to support more kinds of types.

The bottom line is, the code in vibe.d was correct, so we need to avoid breaking it and other such code, even if that means adding a little bit of complexity here.

CyberShadow:
BTW, perhaps leveraging the implementation in the compiler will make this simpler. Does `Unqual!T v = T.init;` compile? Then, `Unqual!T` is a safely head-const-stripped version of `T`. Then there is no need to check for every kind of D type.

TurkeyMan:
I can make a rule like that. It still feels pretty complex though.
Here's a thought; the argument is that it interacts poorly with `auto`, but I wonder why `auto` works the way it does?
```D
const(int) foo();
auto x = foo();
pragma(msg, typeof(x));

> const(int)
```
This seems unnecessary... why doesn't auto strip away pointless qualifiers naturally?
I feel like that would have a simplifying effect on a lot of code. It may also reduce instantiation bloat; since functions that receive `auto` args would coalesce various qualified permutations into fewer instantiations.

PR #2754, opened 2019-08-22
Annotate escaping references with `return`

JinShil:
I think these are escaping bugs in druntime that have been exposed with the fix in https://github.com/dlang/dmd/pull/10333 (Assuming my understanding is correct)

### DIFF
--- a/src/core/internal/convert.d
+++ b/src/core/internal/convert.d
@@ -34,7 +34,7 @@ private ubyte[] ctfe_alloc()(size_t n)
 }
 
 @trusted pure nothrow @nogc
-const(ubyte)[] toUbyte(T)(const ref T val) if (is(Unqual!T == float) || is(Unqual!T == double) || is(Unqual!T == real) ||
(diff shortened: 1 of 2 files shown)


src/core/internal/convert.d
```diff
@@ -34,8 +34,9 @@ private ubyte[] ctfe_alloc()(size_t n)
 }
 
 @trusted pure nothrow @nogc
-const(ubyte)[] toUbyte(T)(const ref T val) if (is(Unqual!T == float) || is(Unqual!T == double) || is(Unqual!T == real) ||
-                                        is(Unqual!T == ifloat) || is(Unqual!T == idouble) || is(Unqual!T == ireal))
+const(ubyte)[] toUbyte(T)(return const ref T val)
```
JinShil: The only actual change here is the added return to the val parameter. The code was just so horribly formatted, I had to do something about it.

JinShil:
I think `return` is supposed to be inferred for these instances, so closing for now until I figure this out.

PR #2755, opened 2019-08-23
Issue 20155 - Allocating a struct with dtor on the GC heap can produce false pointers

rainers:
clear padding area not completely zeroed for struct allocations with dtor

### DIFF
--- a/src/rt/lifetime.d
+++ b/src/rt/lifetime.d
@@ -1096,7 +1096,8 @@ extern (C) void* _d_newitemU(in TypeInfo _ti)
     auto ti = unqualify(_ti);
     auto flags = !(ti.flags & 1) ? BlkAttr.NO_SCAN : 0;
     immutable tiSize = structTypeInfoSize(ti);
-    immutable size = ti.tsize + tiSize;
+    immutable itemSize = ti.tsize;
+    immutable size = itemSize + tiSize;
     if (tiSize)
         flags |= BlkAttr.STRUCTFINAL | BlkAttr.FINALIZE;
 
@@ -1104,7 +1105,10 @@ extern (C) void* _d_newitemU(in TypeInfo _ti)
     auto p = blkInf.base;
 
     if (tiSize)
+    {
+        *cast(TypeInfo*)(p + itemSize) = null; // the GC might not have cleared this area
         *cast(TypeInfo*)(p + blkInf.size - tiSize) = cast() ti;
+    }
 
     return p;
 }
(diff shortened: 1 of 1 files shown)


rainers:
Added fix for arrays of structs with dtor and any large arrays.

src/rt/lifetime.d
```diff
@@ -1096,15 +1118,19 @@ extern (C) void* _d_newitemU(in TypeInfo _ti)
...
+    immutable size = itemSize + tiSize;
     if (tiSize)
         flags |= BlkAttr.STRUCTFINAL | BlkAttr.FINALIZE;
 
     auto blkInf = GC.qalloc(size, flags, ti);
     auto p = blkInf.base;
 
     if (tiSize)
+    {
+        *cast(TypeInfo*)(p + itemSize) = null; // the GC might not have cleared this area
```
kinke: Is there some guarantee that there isn't anymore padding between this and the `ti` pointer at the end of the block?

rainers: Everything else is already cleared by the GC:
https://github.com/dlang/druntime/blob/396a0ec7ef2ba38bb6314f7992d5d99be3645f02/src/gc/impl/conservative/gc.d#L417-L422

rainers:
Added a tiny optimization that memset isn't called if the NO_SCAN attribute is set.

src/rt/lifetime.d
```diff
@@ -436,14 +455,17 @@ BlkInfo __arrayAlloc(size_t arrsize, ref BlkInfo info, const TypeInfo ti, const
     if (!info.base)
         return __arrayAlloc(arrsize, ti, tinext);
 
+    size_t padsize = __arrayPad(arrsize, tinext);
```
Geod24: `immutable` ?

src/rt/lifetime.d
```diff
@@ -406,6 +406,22 @@ size_t __arrayPad(size_t size, const TypeInfo tinext) nothrow pure @trusted
     return size > MAXMEDSIZE ? LARGEPAD : ((size > MAXSMALLSIZE ? MEDPAD : SMALLPAD) + structTypeInfoSize(tinext));
 }
 
+/**
+  clear padding that might not be zeroed by the GC (it assumes it is within the
+  requested size from the start, but it is actually at the end of the allocated block)
+  */
+void __arrayClearPad(ref BlkInfo info, size_t arrsize, size_t padsize) nothrow pure
```
Geod24: `private` ?

src/rt/lifetime.d
rainers: done.

src/rt/lifetime.d
rainers: done.

Geod24: On the wrong function?

rainers: Oops, fixed. Just shows how arbitrary it looks to mark only this function private.

rainers:
Ping.

ibuclaw:
This PR caused a regression https://issues.dlang.org/show_bug.cgi?id=22844

src/rt/lifetime.d
ibuclaw: This set is unaligned, breaking strict alignment targets.

PR #2757, opened 2019-08-24
Failure test was at the wrong indentation level.

TurkeyMan:
@ibuclaw Thanks for spotting it!

### DIFF
--- a/src/core/internal/atomic.d
+++ b/src/core/internal/atomic.d
@@ -640,9 +640,9 @@ void atomicFence(MemoryOrder order = MemoryOrder.seq)() nothrow @nogc @safe
                 ret;
             }
         }
+        else
+            static assert (false, "Unsupported architecture.");
     }
-    else
-        static assert (false, "Unsupported architecture.");
 }
 
 


src/core/internal/atomic.d
```diff
@@ -640,9 +640,9 @@ void atomicFence(MemoryOrder order = MemoryOrder.seq)() nothrow @nogc @safe
                 ret;
             }
         }
+        else
+            static assert (false, "Unsupported architecture.");
     }
-    else
-        static assert (false, "Unsupported architecture.");
```
Geod24: Shouldn't this be `Unsupported memory order` ?

TurkeyMan:
It's fine, it just doesn't do anything.

Geod24:
> It's fine, it just doesn't do anything.

I was just going by @ibuclaw 's comment on Slack:
> Actually, it should be a `static assert (order != MemoryOrder.raw, "Invalid MemoryOrder for atomicFence");`, same as the atomicLoad/Store.

TurkeyMan:
I saw, but I don't think that's correct.

src/core/internal/atomic.d
ibuclaw: Unless it is intentional that `MemoryOrder.raw` be a no-op for atomicFence.

TurkeyMan: It is intentional. That's how other atomic API's roll. I couldn't find why, but I'd rather minimise surprise for users.

Geod24: From what I can tell, there's two approaches:
- `atomicLoad` and `atomicStore` `static assert` that the memory order is what they expect
- Other API that do not use `MemoryOrder` just ignore the parameter's value

`atomicFence` is in a third category, where it expects a specific `MemoryOrder`, but does not do anything if that order is not provided.

ibuclaw: Actually, having thought about it, the MemoryOrder should be enforced in the public API, not the internal.

@Geod24 - This is what the documentation says on C++11 (renamed relevant names to core.atomic equivalents):
- `atomicLoad`: The valid memory order variants are raw, seq, and acq (also valid is consume, but doesn't exist in D API).
- `atomicStore`: The valid memory order variants are raw, seq, and rel.
- `atomicFetch(Add|Sub)`: All memory orders are valid. 
- `atomicExchange`: The valid memory order variants are raw, seq, acq, rel, and acq_rel (all that are present in D API)
- `cas`: There are no restrictions on what memory order can be used here. 
- `atomicFence`: All memory orders are valid.

ibuclaw: Actually, I've just noticed the comment at the top of this function, doing no-op is fine for `MemoryOrder.raw`.  For X86, it should be a no-op for all memory orders except `seq`.

ibuclaw:
@TurkeyMan - For x86 without sse2: https://explore.dgnu.org/z/Kpt1cm

src/core/internal/atomic.d
TurkeyMan: > Actually, having thought about it, the MemoryOrder should be enforced in the public API, not the internal.

Yes, this too. I tried to leave all the noise outside this `internal` file, which was just intended to map very closely to the intrinsics, so it was easy to swap out, and keep track of many implementations for a lot of platforms, which don't each have to be bogged down by a lot of noise.

My initial implementation even restricted `T` to a sized integers in `internal`, and the calling code wrangled the source arg to the appropriate sized int and did a thunk; but that was bad because I noticed the calling convention was different for int and some int-sized structs, which resulted in unnecessary conversions of calling convention between the client code and the internal code, which I couldn't optimise away (because inline asm).

I think this is the most tasteful arrangement I could find that was also efficient.

Also, this was a LOT of work.

> `atomicLoad`: The valid memory order variants are raw, seq, and acq (also valid is consume, but doesn't exist in D API).

I thought on adding `consume` to the D enum, but I wasn't sure if we could implement it. DMD probably can't... @ibuclaw is it simple to route through to the GDC/LDC backends? I presume so.
Consume is a bit niche, I've never needed it personally, and I work with atomics a lot. Maybe I've missed opportunities to use it though...

TurkeyMan:
Yeah, that implementation looks reasonable ;)

src/core/internal/atomic.d
ibuclaw: @TurkeyMan - The __ATOMIC_XXX defines and builtins all come from the compiler middle-end, so all you need to do is add the enum to MemoryOrder with the right value to use it with gcc atomics.

As for whether it makes sense to add consume, I'll let the following speak for itself:

From [gcc internals](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html):
> __ATOMIC_CONSUME: This is currently implemented using the stronger __ATOMIC_ACQUIRE memory order because of a deficiency in C++11’s semantics for memory_order_consume. 

From well-informed [blog post](https://preshing.com/20140709/the-purpose-of-memory_order_consume-in-cpp11/):
> memory_order_consume is probably the least well-understood. It’s the most complicated ordering constraint, and it offers the least reward for using it correctly.

It also goes on to link reasons why both gcc and llvm treat consume in the same way as acquire.

PR #2760, opened 2019-08-26
Add core.experimental.rc

lesderid:
This PR adds four `@safe @nogc @nothrow` reference counted data structures to `core.experimental`:
* a bare-bones slice type with reference semantics,
* a proper dynamic array type using the slice type,
* a singly linked list type, and
* a hash map type.

The slice in this PR differs from `__rcptr!T` (https://github.com/dlang/druntime/pull/2690) by using intrusive reference counting instead of the refcount being external to the object, increasing performance. This also means we can safely call `free` on the pointer because we always allocate memory ourselves with `malloc`. 

Sadly, we're still stuck with using atomics for the reference count (see discussion here: https://github.com/dlang/druntime/pull/2679#discussion_r303841362). I believe this could only reasonably be solved by changing the `immutable` => `shared` conversion rules. This would probably break a lot of code (and obviously requires a DIP) and I'm not sure what the deprecation process would look like. I'll start a thread on the NG soon to discuss this issue further.

(The hash map does not yet support (non-empty) creating immutable instances. For now, one should be able to use the same idiom that can be used for the built-in ones: https://dlang.org/spec/hash-map.html#runtime_initialization.)

### DIFF
--- a/src/core/experimental/rc/slice.d
+++ b/src/core/experimental/rc/slice.d
@@ -1,0 +1,228 @@
+// Written in the D programming language.
+/**
+This module provides a barebones reference counted slice.
+
+License: $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+
+Authors: Les De Ridder
+
+Source: $(DRUNTIMESRC core/experimental/rcslice.d)
+*/
+module core.experimental.rc.slice;
+
+///
+unittest
+{
+    __rcslice!int a; //default ctor
+    assert(a.ptr is null);
+
+    auto b = __rcslice!int(10); //allocating ctor
+
+    {
+        auto c = a; //copy ctor
+        assert(c.ptr == a.ptr);
+        assert(c.ptr != b.ptr);
+
+        c = b; //opAssign
+        assert(c.ptr != a.ptr);
+        assert(c.ptr == b.ptr);
+    }
+
+    assert(a.count is null);
+    assert(*b.count == 0);
+}
+
+/**
+A reference counted slice can be used to implement reference counted data
+structures.
+
+The reference count is automatically incremented/decremented on assignment,
+copy (construction), and destruction. When there are no more references to the
+slice, the slice and the reference count are automatically deallocated.
+
+Implementation: The internal implementation of `__rcslice` uses `malloc`/`free`.
+*/
+struct __rcslice(T, alias onDeallocate = null, bool defaultInitialize = true)
+{
+    alias CounterType = int;
+
+    private T[] slice;
+    private shared(CounterType)* count = null;
+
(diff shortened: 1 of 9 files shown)


src/core/experimental/rc/slice.d
```diff
@@ -0,0 +1,217 @@
...
+        // (i.e. the counter is 0-based, because we use calloc)
+        //TODO: Use `atomicFetchSub` with `MemoryOrder.raw` once LDC has it
+        if (atomicLoad!(MemoryOrder.raw)(*count) == 0 || atomicOp!("-=")(*count, 1) == -1)
+        {
+            deallocate();
+        }
+    }
+
+    @trusted
+    private void deallocate()
```
aG0aep6G: Strictly speaking, this can't be `@trusted`. It's essentially an alias of `free`, which can't possibly be `@trusted`.

One might argue that it's ok here, because it's private and the rest of the code takes care to call it correctly. But that's prone to the kind of mistakes that `@safe` is supposed to prevent.

And there already is such a mistake in the code: `opAssign(null)` is public and calls `delRef` which calls `deallocate`. We effectively have a public `@trusted` alias of `free`, and we can use it to create dangling references in `@safe` code:

```d
auto rcs = __rcslice!int(1);
int[] s = rcs[];
rcs = null; /* leaves s dangling */
```

lesderid: Thanks for pointing this out!

lesderid:
With the last commit (https://github.com/dlang/druntime/pull/2760/commits/57b38e1219a61d5181c15a727a637a9e2fcc9d45) and using non-atomic ref counting, `rcarray` is faster than `std.container.array` for most mixed benchmarks and nearly as fast as `std::vector` (with a wrapper):

<details>
<summary>Benchmark plots</summary>

Mixed 1
Mixed 2
Mixed 3
</details>

cc @wilzbach

aG0aep6G:
As far as I can tell, these containers are not `@safe`.

They allow obtaining references to their elements, and they don't manage to do it `@safe`ly. I'm not sure where we stand with DIP 1000, etc. Maybe it just can't be done, yet.

Examples where references to elements outlive their containers (i.e. they become dangling):

```d
ref int slice() @safe
{
    auto s = __rcslice!int(1);
    s[0] = 42;
    return s[0];
}

ref int array() @safe
{
    rcarray!int a;
    a ~= 42;
    return a[0];
}

ref int slist() @safe
{
    rcslist!int s;
    s.insertFront(42);
    return s.front;
}

ref int map() @safe
{
    rcmap!(int, int) m;
    m[0] = 42;
    return *(0 in m);
}
```

RazvanN7:

PR #2761, opened 2019-08-27
Move array dup properties from object.d to core.internal

JinShil:
Followup to https://github.com/dlang/druntime/pull/2647, https://github.com/dlang/druntime/pull/2644, https://github.com/dlang/druntime/pull/2643, and https://github.com/dlang/druntime/pull/2634

This is a continuation of work to clean up object.d.  It's just a cut-and-paste with the exception of `getPostblit` and `doPostblit` which were made public (though still private to druntime because it is in `core.internal`) so it could be used by some associative array code that still resides in object.d

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -43,6 +43,7 @@ COPY=\
 	$(IMPDIR)\core\internal\array\casting.d \
 	$(IMPDIR)\core\internal\array\capacity.d \
 	$(IMPDIR)\core\internal\array\concatenation.d \
+	$(IMPDIR)\core\internal\array\duplication.d \
 	$(IMPDIR)\core\internal\array\utils.d \
 	\
 	$(IMPDIR)\core\internal\util\array.d \
--- a/mak/DOCS
+++ b/mak/DOCS
@@ -82,6 +82,7 @@ DOCS=\
     $(DOCDIR)\core_internal_array_construction.html \
     $(DOCDIR)\core_internal_array_equality.html \
     $(DOCDIR)\core_internal_array_concatenation.html \
+    $(DOCDIR)\core_internal_array_duplication.html \
     $(DOCDIR)\core_internal_array_utils.html \
     $(DOCDIR)\core_internal_util_array.html \
     \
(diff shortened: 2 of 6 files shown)


JinShil:
Weird!  All packages in buildkite pass, but the unittests in druntime fail.

thewilsonator:
```
src/core/internal/array/duplication.d(243): Error: cannot implicitly convert expression `dup(s)` of type `char[]` to `immutable(string)`
--
  | src/core/internal/array/duplication.d(268): Error: cannot implicitly convert expression `dup(m)` of type `void[]` to `immutable(void)[]`
  | src/core/internal/array/duplication.d(272): Error: cannot implicitly convert expression `dup(s)` of type `shared(void)[]` to `immutable(void)[]`
  | src/core/internal/array/duplication.d(243): Error: cannot implicitly convert expression `dup(s)` of type `char[]` to `immutable(string)`
  | src/core/internal/array/duplication.d(268): Error: cannot implicitly convert expression `dup(m)` of type `void[]` to `immutable(void)[]`
  | src/core/internal/array/duplication.d(272): Error: cannot implicitly convert expression `dup(s)` of type `shared(void)[]` to `immutable(void)[]`
```

JinShil:
@ thewilsonator.  Thanks.  I see the error message, but any idea why it is occurring? All I did was move the code.

thewilsonator:
Dunno, but it seems that all of those should be `idup`s.

JinShil:
> it seems that all of those should be `idup`s.

Line 38 explicitly says `const overload to support implicit conversion to immutable (unique result, see DIP29)`, so there must be something else at play.  Something's not right though.  Just moving code shouldn't trigger this.

thewilsonator:
IDK, try asking on slack or the NG,

JinShil:
I don't think these symbols should be in `core.internal`.  I think there are a couple of better ways to do this.

ZombineDev:
@JinShil as mentioned in #2784, I think they're perfect candidates for `core.array`.

PR #2763, opened 2019-08-27
Move core.internal.arrayop.d to core.internal.array/operations.d to join other array-related files

JinShil:
Followup to https://github.com/dlang/druntime/pull/2647, https://github.com/dlang/druntime/pull/2644, https://github.com/dlang/druntime/pull/2643, and https://github.com/dlang/druntime/pull/2634

This is a continuation of work to clean up object.d and bring some older conventions in line with the current trends.  There's no change in functionality; this PR has only moved code to a new file ~and renamed `arrayOp` to `_arrayOp` to avoid needing a forwarding template~.

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -22,7 +22,6 @@ COPY=\
 	$(IMPDIR)\core\vararg.d \
 	\
 	$(IMPDIR)\core\internal\abort.d \
-	$(IMPDIR)\core\internal\arrayop.d \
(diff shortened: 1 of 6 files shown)


JinShil:
This should be ready to go now.  For some reason renaming from `_arrayOp` to `arrayOp` cause 32-bit platforms to fail in `core.internal.simd`.  So I removed it for now.  I'll address it in a followup PR where I can deal with it in isolation.

PR #2766, opened 2019-08-27
Move _d_assert_fail from object.d to core.internal.dassert.d

JinShil:
Followup to #2647, #2644, #2643, #2634, #2763, and #2765

This is a continuation of work to clean up object.d

### DIFF
--- a/mak/DOCS
+++ b/mak/DOCS
@@ -75,6 +75,7 @@ DOCS=\
 	$(DOCDIR)\core_sys_darwin_mach_thread_act.html \
 	$(DOCDIR)\core_sys_darwin_netinet_in_.html \
     \
+    $(DOCDIR)\core_internal_dassert.html \
     $(DOCDIR)\core_internal_switch_.html \
 	\
     $(DOCDIR)\core_internal_array_appending.html \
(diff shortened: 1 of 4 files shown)


CyberShadow:
All these object.d refactorings do seem to have a measurable effect on compilation times:
https://perf.dlang.io/#program-empty-compile-realtime-10;program-empty-compile-callgrind-insns;1560217958.0705755;1567251986

However, the cost is very small - merely two milliseconds.

JinShil:
For instruction count, I see the bump, but for real time, there seems to be a lot of noise, and I don't think it's possible to say one way or another.  For example, there was a bump in real time for [this PR](https://github.com/dlang/druntime/pull/2774), but all I did was delete code.

CyberShadow:
Measuring CPU time is always going to be imprecise, however we can still see that the time is increasing (compare visually the left side and right side of the chart at that link). To find the cause why CPU time has increased, we can use a synthetic but exact metric which is generally directly correlated with it (in this case, # of instructions executed as reported by Callgrind). Using this method we can see that the execution time was increased by this & similar recent PRs.

PR #2768, opened 2019-08-28
Demphasize runtime hooks in object.d's documentation

JinShil:
Followup to #2647, #2644, #2643, #2634, #2763, #2765, and #2766

This is a continuation of work to clean up object.d.

The runtime hooks are only used by the compiler and should not be invoked directly by the user.  This PR moves them to the bottom of object.d's file so they appear last in the documentation.  `_cmp` had to be moved too due to a bug in the compiler causing a conflict with `core.internal.array.comparison.__cmp`.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -38,117 +38,12 @@ alias dstring = immutable(dchar)[];
 
 version (D_ObjectiveC) public import core.attribute : selector;
 
-/// See $(REF _d_cmain, core,internal,entrypoint)
-public import core.internal.entrypoint : _d_cmain;
-
-/// See $(REF _d_arrayappendTImpl, core,internal,array,appending)
-public import core.internal.array.appending : _d_arrayappendTImpl;
-/// See $(REF _d_arrayappendcTXImpl, core,internal,array,appending)
-public import core.internal.array.appending : _d_arrayappendcTXImpl;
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -4178,3 +4073,108 @@ void __ArrayDtor(T)(T[] a)
...
+/// See $(REF _d_assert_fail, core,internal,dassert)
+public import core.internal.dassert: _d_assert_fail;
+
+/// See $(REF __switch, core,internal,switch_)
+public import core.internal.switch_: __switch;
+/// See $(REF __switch_error, core,internal,switch_)
+public import core.internal.switch_: __switch_error;
+
+// Compare class and interface objects for ordering.
+private int __cmp(Obj)(Obj lhs, Obj rhs)
```
ZombineDev: This looks like another candidate for moving to `core.internal`.

src/object.d
```diff
@@ -4178,3 +4073,108 @@ void __ArrayDtor(T)(T[] a)
     foreach_reverse (ref T e; a)
         e.__xdtor();
 }
+
```
ZombineDev: I wonder if it makes sense to add a big comment along the lines of:
```D
//******************
//* Runtime Hooks: *
//******************
```

To better designate the file structure.

PR #2769, opened 2019-08-28
Move __move_post_blt from object.d to core.internal.moving.d

JinShil:
Followup to #2647, #2644, #2643, #2634, #2763, #2765, and #2766

This is a continuation of work to clean up object.d.

~This code was added in https://github.com/dlang/druntime/pull/2638, but it doesn't currently appear to be used anywhere.~

~During the GSoC 2019 project for converting some runtime hooks to templates, we decided that we should be using the naming convention `_d_whatever` for any runtime templates such as this.  That naming convention is already the dominant convention in druntime, so it makes sense to keep it.  It also distinguishes runtime hooks from compiler-generated symbols that are prefixed with `__`.  A few of the templates don't follow the `_d_` convention, but I intend to remedy that with a later PR so all are consistent.~

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -28,6 +28,7 @@ COPY=\
 	$(IMPDIR)\core\internal\dassert.d \
 	$(IMPDIR)\core\internal\entrypoint.d \
 	$(IMPDIR)\core\internal\hash.d \
+	$(IMPDIR)\core\internal\moving.d \
 	$(IMPDIR)\core\internal\parseoptions.d \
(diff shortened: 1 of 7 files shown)


JinShil:
It appears [code in Phobos calls this directly](https://github.com/dlang/phobos/blob/f7fc24c3484eaa37cae1b3507f5f7fd616103a06/std/algorithm/mutation.d#L1420), so I reverted the rename.

PR #2770, opened 2019-08-29
Give TypeInfo_Class/TypeInfo_Interface.isBaseOf like C#/Java isAssignableFrom

n8sh:
This is equivalent to C#/Java `isAssignableFrom` ~~with the argument swapped with the receiver~~. Naming the method "isAssignableFrom" would be more familiar to people coming from C#/Java (and I would find it convenient since it's one fewer difference to remember between languages) but is potentially misleading: "alias this" and overloadable opAssign mean that this would not actually indicate whether values of one type could be assigned to another.

~~I used the name "isDerivedFrom" since the phrase "derives from" appears in the D online documentation. Swapping the argument order is not ideal, so if someone can think of a sensible name that would work with the same argument order as "isAssignableFrom" please suggest one.~~

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -960,6 +960,8 @@ class TypeInfo_Delegate : TypeInfo
 }
 
 private extern (C) Object _d_newclass(const TypeInfo_Class ci);
+private extern (C) int _d_isbaseof(scope TypeInfo_Class child,
+    scope const TypeInfo_Class parent) @nogc nothrow pure @safe; // rt.cast_
 
 /**
  * Runtime type information about a class.
@@ -1103,6 +1105,36 @@ class TypeInfo_Class : TypeInfo
         }
         return o;
     }
+
+   /**
+    * Returns true if the class described by `child` derives from or is
+    * the class described by this `TypeInfo_Class`. Always returns false
+    * if the argument is null.
+    *
+    * Params:
+    *  child = TypeInfo for some class
+    * Returns:
+    *  true if the class described by `child` derives from or is the
+    *  class described by this `TypeInfo_Class`.
+    */
+    final bool isBaseOf(scope const TypeInfo_Class child) const @nogc nothrow pure @trusted
+    {
+        if (m_init.length)
+        {
+            // If this TypeInfo_Class represents an actual class we only need
+            // to check the child and its direct ancestors.
+            for (auto ti = cast() child; ti !is null; ti = ti.base)
+                if (ti is this)
+                    return true;
+            return false;
+        }
+        else
+        {
+            // If this TypeInfo_Class is the .info field of a TypeInfo_Interface
+            // we also need to recursively check the child's interfaces.
+            return child !is null && _d_isbaseof(cast() child, this);
+        }
+    }
 }
 
 alias ClassInfo = TypeInfo_Class;
@@ -1192,6 +1224,38 @@ class TypeInfo_Interface : TypeInfo
     override @property uint flags() nothrow pure const { return 1; }
 
     TypeInfo_Class info;
+
+   /**
+    * Returns true if the class described by `child` derives from the
+    * interface described by this `TypeInfo_Interface`. Always returns
+    * false if the argument is null.
+    *
+    * Params:
+    *  child = TypeInfo for some class
+    * Returns:
+    *  true if the class described by `child` derives from the
+    *  interface described by this `TypeInfo_Interface`.
+    */
+    final bool isBaseOf(scope const TypeInfo_Class child) const @nogc nothrow pure @trusted
+    {
+        return child !is null && _d_isbaseof(cast() child, this.info);
+    }
+
+   /**
+    * Returns true if the interface described by `child` derives from
+    * or is the interface described by this `TypeInfo_Interface`.
+    * Always returns false if the argument is null.
+    *
+    * Params:
+    *  child = TypeInfo for some interface
+    * Returns:
+    *  true if the interface described by `child` derives from or is
+    *  the interface described by this `TypeInfo_Interface`.
+    */
+    final bool isBaseOf(scope const TypeInfo_Interface child) const @nogc nothrow pure @trusted
+    {
+        return child !is null && _d_isbaseof(cast() child.info, this.info);
+    }
 }
 
 class TypeInfo_Struct : TypeInfo
--- a/src/rt/cast_.d
+++ b/src/rt/cast_.d
@@ -14,6 +14,9 @@
 module rt.cast_;
 
 extern (C):
+@nogc:
+nothrow:
+pure:
 
 /******************************************
  * Given a pointer:
@@ -74,7 +77,7 @@ void* _d_dynamic_cast(Object o, ClassInfo c)
     return res;
 }
 
-int _d_isbaseof2(ClassInfo oc, ClassInfo c, ref size_t offset)
+int _d_isbaseof2(scope ClassInfo oc, scope const ClassInfo c, scope ref size_t offset) @safe
 {
     if (oc is c)
         return true;
@@ -101,7 +104,7 @@ int _d_isbaseof2(ClassInfo oc, ClassInfo c, ref size_t offset)
     return false;
 }
 
-int _d_isbaseof(ClassInfo oc, ClassInfo c)
+int _d_isbaseof(scope ClassInfo oc, scope const ClassInfo c) @safe
 {
     if (oc is c)
         return true;
--- a/test/typeinfo/src/isbaseof.d
+++ b/test/typeinfo/src/isbaseof.d
@@ -1,0 +1,46 @@
+// https://issues.dlang.org/show_bug.cgi?id=20178
+
+interface I {}
+interface J : I {}
+interface K(T) {}
+class C1 : I {}
+class C2 : C1 {}
+class C3 : J {}
+class C4(T) : C3, K!T {}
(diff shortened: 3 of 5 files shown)


src/object.d
```diff
@@ -1736,6 +1736,42 @@ class TypeInfo_Class : TypeInfo
         }
         return o;
     }
+
+   /**
+    * Returns true if the class described by this TypeInfo_Class derives from
+    * the interface described by `parent`.
+    */
+    bool isDerivedFrom(scope const TypeInfo_Interface parent) const @nogc nothrow pure @safe
```
JinShil: DDoc should also include `Params` and `Returns` when applicable.

src/object.d
```diff
@@ -1736,6 +1736,42 @@ class TypeInfo_Class : TypeInfo
...
+    bool isDerivedFrom(scope const TypeInfo_Class parent) const @nogc nothrow pure @safe
+    {
+        if (parent is null)
+            return false;
+        if (parent is this || parent.name == this.name)
+            return true;
+        return this.base !is null && this.base.isDerivedFrom(parent);
+    }
+
+    private static bool interfaceConvertibleFrom(scope const TypeInfo_Class interfaceInfo,
```
JinShil: I also recommend DDoc headers on private functions because it makes it easier for those browsing the code with a modern editor to see the documentation at usage sights.

JinShil:
Seems like a good idea to me.  I'm not sure about the implementation though.  Is `name` reliable for comparisons.  I think the tests could be filled out with more corner cases (e.g. packages, nested types, multiple interface inheritance, etc.)

Geod24:
What's the use case for this ? Java / C# reflection is primarily runtime, while D's is primarily compile time.

src/object.d
```diff
@@ -1736,6 +1736,42 @@ class TypeInfo_Class : TypeInfo
...
+
+   /**
+    * Returns true if the class described by this TypeInfo_Class describes a class that
+    * derives from the interface described by this TypeInfo_Interface.
+    */
+    bool isDerivedFrom(scope const TypeInfo_Class parent) const @nogc nothrow pure @safe
+    {
+        if (parent is null)
+            return false;
+        if (parent is this || parent.name == this.name)
```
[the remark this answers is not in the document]
rainers: Dynamic casting just compares TypeInfo pointers, so if you don't expect copies of TypeInfo being made at runtime, you don't have to compare names. You might also just call `rt.cast_._d_isbaseof`.

n8sh:
My personal use case was implementing multimethods when the classes involved aren't known at compile time.

ZombineDev:
@n8sh Perhaps rebasing on top of master should fix the error on BuildKite.

src/object.d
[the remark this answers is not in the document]
n8sh: If pointer comparison is good enough for dynamic casting then it is good enough for this purpose too. Thank you for the tip regarding `_d_isbaseof`.

n8sh:
Changing to use `_d_isbaseof` and renaming to `isBaseOf` with the same argument order as `isAssignableFrom`.

src/object.d
```diff
@@ -1102,6 +1104,24 @@ class TypeInfo_Class : TypeInfo
...
+    * Returns:
+    *  true if the class described by `child` derives from the class described
+    *  by `child`, false otherwise.
+    */
+    bool isBaseOf(scope const TypeInfo_Class child) const @nogc nothrow pure @trusted
+    {
+        for (auto ti = cast() child; ti !is null; ti = ti.base)
+            if (ti is this)
+                return true;
+        return false;
```
n8sh: If there's some case for classes where this gives the wrong answer but `_d_isbaseof` gives the right answer I haven't found it. Unless such a case can be found I'll leave it here since `_d_isbaseof` potentially does a lot more work and `extern(C)` function calls are opaque to the optimizer.

src/rt/cast_.d
```diff
@@ -14,6 +14,9 @@
 module rt.cast_;
 
 extern (C):
+@nogc:
+nothrow:
+pure:
```
n8sh: Adding some attributes here so the compiler can verify the attributes I claim in the `private extern(C)` declaration in `object.d`.

thewilsonator:
@n8sh whats the status of this?

n8sh:
@thewilsonator The code is ready but discussion is open regarding whether this should be added.

src/object.d
```diff
@@ -1191,6 +1211,38 @@ class TypeInfo_Interface : TypeInfo
...
+    * Returns true if the class described by `child` derives from the
+    * interface described by this `TypeInfo_Interface`.
+    *
+    * Params:
+    *  child = TypeInfo for some class
+    * Returns:
+    *  true if the class described by `child` derives from the
+    *  interface described by this `TypeInfo_Interface`.
+    */
+    bool isBaseOf(scope const TypeInfo_Class child) const @nogc nothrow pure @trusted
```
n8sh: Maybe this and the other functions should be `final`. The limitation could be lifted later if necessary.

thewilsonator:
It could also do with a changelog entry.

src/object.d
thewilsonator: Sounds good to me.

src/object.d
n8sh: Looking back at this, checking straight up the inheritance chain rather than using _d_isbaseof means that the receiver TypeInfo_Class and argument TypeInfo_Class must represent actual classes rather than being the .info field of some TypeInfo_Interface.

thewilsonator:
Still needs a changelog entry.

n8sh:
Any feedback regarding whether the current behavior with TypeInfo_Interface's ClassInfo is appropriate?

thewilsonator:
None, but I don't see any reason to not have it.

n8sh:
I'd like to hold off to see if any comments on that are forthcoming, unless you're in a hurry to move this along.

12345swordy:
I don't see any reason to reject this giving that there are attempts to call c# code from d.

src/object.d
```diff
@@ -1102,6 +1104,24 @@ class TypeInfo_Class : TypeInfo
         }
         return o;
     }
+
+   /**
+    * Returns true if the class described by `child` derives from the
```
rainers: I think the description should explicitly mention that the class is considered a base of itself (making the name a bit misleading).

src/object.d
```diff
@@ -1102,6 +1104,24 @@ class TypeInfo_Class : TypeInfo
...
+    * class described by this `TypeInfo_Class`.
+    *
+    * Params:
+    *  child = TypeInfo for some class
```
rainers: Mention that `null` is also a valid argument?

12345swordy:
@n8sh ping. Address rainers comments.

src/object.d
n8sh: Solution: detect an interface's ClassInfo (TypeInfo_Class that does not represent a class) by checking if the initializer has length zero.

12345swordy:
@thewilsonator Any objections so far.

thewilsonator:
None from me, but it doesn't seem to pass.

12345swordy:
@n8sh the test are not passing, could you rebase?

12345swordy:
@thewilsonator the tests have pass, need to add auto merge label to this

Geod24:
Does this work with `shared` druntime ?

n8sh:
@Geod24 I don't know the answer to that but it works the same as `rt.cast_._d_isbaseof`.

test/typeinfo/src/isbaseof.d
```diff
@@ -0,0 +1,35 @@
+// https://issues.dlang.org/show_bug.cgi?id=20178
```
Geod24: Can you add:

- A test with a templated interface / templated class (both `C4(T) : I!T` and `C4(T) : I!int` let's say).
- A test that it can't be called with any other typeinfo (`static assert(!__traits(compiles, ...));`)

Geod24:
Ah I found it: https://github.com/dlang/druntime/pull/2828

src/object.d
```diff
@@ -1192,6 +1224,38 @@ class TypeInfo_Interface : TypeInfo
...
+    *
+    * Params:
+    *  child = TypeInfo for some interface
+    * Returns:
+    *  true if the interface described by `child` derives from or is
+    *  the interface described by this `TypeInfo_Interface`.
+    */
+    final bool isBaseOf(scope const TypeInfo_Interface child) const @nogc nothrow pure @trusted
+    {
+        return child !is null && cast(bool) _d_isbaseof(cast() child.info, this.info);
```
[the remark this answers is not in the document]
n8sh: Even better I can omit the `cast`/`!!` altogether since bool && int works.

src/rt/cast_.d
```diff
@@ -101,7 +104,7 @@ int _d_isbaseof2(ClassInfo oc, ClassInfo c, ref size_t offset)
     return false;
 }
 
-int _d_isbaseof(ClassInfo oc, ClassInfo c)
+int _d_isbaseof(scope ClassInfo oc, scope const ClassInfo c) @safe
```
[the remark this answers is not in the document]
n8sh: I'd rather not mess with the implementation of `_d_isbaseof` in this PR.

12345swordy:
What is the hold up here?

12345swordy:
The objection time limit is up
cc @thewilsonator

PR #2771, opened 2019-08-29
Undocument public templates in object.d that are only intended for use by the compiler

JinShil:
Now that https://github.com/dlang/dmd/pull/10365 has been merged, we don't need to document those templates in object.d that are only intended for use by the compiler.  Unfortunately, these may appear in the documentation for one release until https://github.com/dlang/dmd/pull/10365 is folded into the release and dlang.org is updated to use the new compiler.

Also includes @ZombineDev's [suggestion](https://github.com/dlang/druntime/pull/2768#discussion_r318448688).

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3229,43 +3229,37 @@ private void _doPostblit(T)(T[] arr)
     auto a = arr.dup; // dup does escape
 }
 
-/// See $(REF _d_cmain, core,internal,entrypoint)
+/* ************************************************************************
+                           COMPILER SUPPORT
+The compiler lowers certain expressions to instantiations of the following
+templates.  They must be implicitly imported, which is why they are here
+in this file. They must also be `public` as they must be visible from the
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -3995,39 +3995,33 @@ void __ArrayDtor(T)(T[] a)
         e.__xdtor();
 }
 
-/// See $(REF _d_cmain, core,internal,entrypoint)
+/**************************************************************************
```
Geod24:
```suggestion
/* ************************************************************************
```

Otherwise it's a documentation comment

JinShil: I thought the newline between the comment and the declaration would handle that.

JinShil:
ping. It's an easy review and causes merge conflicts with some other work I'm trying to do.

thewilsonator:
auto merge added.

JinShil:
Thank you.

PR #2773, opened 2019-08-29
Fix Issue 13826 - Move volatileLoad/Store to core.volatile when the volatile keyword is removed

ErnyTech:
This PR move the intrinsic volatileLoad and volatileStore into core.volatile,
as promised by the discussion in https://github.com/dlang/druntime/pull/892
and https://github.com/dlang/dmd/pull/4155.

Currently an alias is maintained in core.bitop to avoid broken code,
the alias is marked as deprecated to warn users of the displacement
occurred in the core.volatile

You should probably schedule a removal of the alias from core.bitop and
publish it in https://dlang.org/deprecate.html

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -20,6 +20,7 @@ COPY=\
 	$(IMPDIR)\core\thread.d \
 	$(IMPDIR)\core\time.d \
 	$(IMPDIR)\core\vararg.d \
+	$(IMPDIR)\core\volatile.d \
 	\
 	$(IMPDIR)\core\internal\abort.d \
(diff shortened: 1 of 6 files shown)


ZombineDev:
Don't forget to open a PR to delete https://github.com/dlang/dmd/blob/93363cb191ed76d6f50b665e123dfe3fbc2b2bb7/src/dmd/toir.d#L497-L507 ;)

ErnyTech:
@ZombineDev the PR for DMD is ready https://github.com/dlang/dmd/pull/10370

ZombineDev:
Super, tnx!

PR #2777, opened 2019-08-30
Deprecate `sizediff_t`, `hash_t`, and `equals_t` in object.d

JinShil:


joseph-wakeling-frequenz:
Um.

I'm sorry to be late to the party, and grumpy with it, but ... this feels like a pretty major deprecation step to take, and it feels a bit shocking that it would be done without any description or discussion (in the PR or commit message) of _why_.

If the discussion/motivation/decision happened elsewhere, then it would be nice to have a link to that in the PR and commit.

What gives, folks? :-\

joseph-wakeling-frequenz:
Ah wait.  I see it has been closed without being merged.  Still, I'm seeing a bunch of changes in other projects influenced by this PR.  So I'd still like to understand what the motivation is and why other projects are carrying those (invalid?) patches.

I'm also (grumpy hat on again) kind of shocked that a change like this would be approved by reviewers without any description, discussion, or cross-referencing of the motivation, or any update to the changelog (!!).

Don't you CARE about communicating changes effectively to other developers? :-\

PR #2778, opened 2019-08-30
Fix Issue 16380 - no bindings for err.h

ErnyTech:
err.h although not present in the posix standard it is however present
in most distributions based on Darwin, BSD and Linux

### DIFF
--- a/src/core/sys/darwin/err.d
+++ b/src/core/sys/darwin/err.d
@@ -1,0 +1,41 @@
+/**
+  * D header file for Darwin err.h.
+  *
+  * Copyright: Copyright © 2019, The D Language Foundation
+  * License: <a href="http://www.boost.org/LICENSE_1_0.txt">Boost License 1.0</a>.
+  * Authors: Ernesto Castellotti
+  */
+module core.sys.darwin.err;
+import core.stdc.stdarg : va_list;
+
+version (OSX)
+    version = Darwin;
+else version (iOS)
+    version = Darwin;
+else version (TVOS)
+    version = Darwin;
+else version (WatchOS)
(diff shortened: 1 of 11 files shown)


thewilsonator:
Seems fine to me, I'll wait for others to chime in though.

src/core/sys/darwin/err.d
```diff
@@ -0,0 +1,44 @@
...
+void verrc(int eval, int code, scope const char* fmt, va_list args);
+void verrx(int eval, scope const char* fmt, va_list args);
+void vwarn(scope const char* fmt, va_list args);
+void vwarnc(int code, scope const char* fmt, va_list args);
+void vwarnx(scope const char* fmt, va_list args);
+void err_set_file(void* vfp);
+void err_set_exit(ExitFunction exitf);
+
+// ^ __OSX_AVAILABLE_STARTING(__MAC_10_6, __IPHONE_3_2);
+err_set_exit_b(ExitFunction exitf);
```
jacob-carlborg: `err_set_exit_b` takes a block not a function pointer. A block is an Apple/LLVM extension to C, which is not representable in D.

ErnyTech:
@jacob-carlborg so i need remove err_set_exit_b?

jacob-carlborg:
Yes.

jacob-carlborg:
It’s technically possible to create a struct in D with the same layout as a block, but that was not accepted in the past.

ErnyTech:
Also on opensolaris there is the same thing
https://repo.or.cz/opensolaris.git/blob_plain/HEAD:/usr/src/head/err.h

PR #2779, opened 2019-08-31
[stable] core.sys.posix.unistd: Add constants for Bionic/Android

kinke:
Fixes the druntime build on Android; e.g., gc.os needs _SC_PHYS_PAGES. Based on Android NDK r20 (`bits/syconf.h`).


kinke:
Wrt. stable - I'm upgrading the LDC Android/Termux package, and v1.17 needs this extra patch (for druntime to compile); so I'd like to have this fixed in 2.088, so that LDC v1.18 won't need it anymore.

PR #2782, opened 2019-09-01
Replace `in` with `scope const` for `Throwable.toString`

JinShil:
Followup to 
https://github.com/dlang/druntime/pull/2704
https://github.com/dlang/druntime/pull/2781

## Background
This PR is in support of https://github.com/dlang/dmd/pull/10382

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -1897,7 +1897,7 @@ class Throwable : Object
     override string toString()
     {
         string s;
-        toString((buf) { s ~= buf; });
+        toString((scope const buf) { s ~= buf; });
         return s;
     }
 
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -1897,7 +1897,7 @@ class Throwable : Object
     override string toString()
     {
         string s;
-        toString((buf) { s ~= buf; });
+        toString((scope const buf) { s ~= buf; });
```
JinShil: Shouldn't the compiler be able to infer the `scope const` here?

JinShil:
A decision about `in` has finally been made:  https://github.com/dlang/dmd/pull/10382

PR #2783, opened 2019-09-02
Remove core.internal documentation that only contain symbols for the compiler

JinShil:
These files only contain symbols for the compiler; not for the user.  Since they are in `core.internal` they probably should not be in the documentation.

This is now possible due to the fix in https://github.com/dlang/dmd/pull/10365 and https://github.com/dlang/druntime/pull/2771


JinShil:
DAutoTest had trouble downloading packages so I had to restart it which removed the auto-merge label.  It should be good now.  cc @thewilsonator

PR #2786, opened 2019-09-05
Add std::pair<>

TurkeyMan:


TurkeyMan:
What's the fail? I don't get it :/

thewilsonator:
```
/tmp/ccKx6Emz.o: In function `callC_val(std::pair<int, float>, std::pair<int, float>&, std::pair<int, Elaborate>, std::pair<int, Elaborate>&)':
utility.cpp:(.text+0x40): undefined reference to `fromC_val(std::pair<int, float>, std::pair<int, float>&, std::pair<int, Elaborate>, std::pair<int, Elaborate>&)'
./generated/linux/debug/64/utility_d.o: In function `_D12utility_test16__unittest_L6_C1FZv':
/media/ephemeral0/sandbox/at-client/pull-3760051-Linux_32_64/druntime/test/stdcpp/src/utility_test.d:21: undefined reference to `callC_val(std::pair<int, float>, std::pair<int, float>&, std::pair<int, Elaborate>, std::pair<int, Elaborate>&)'
```

TurkeyMan:
Yeah... but I don't get it... the functions are there, the mangling looks right :/

TurkeyMan:
Wait... the C++ code tried to call a function with `@PLT` stuck on the end of the name O_o

TurkeyMan:
Everything's hard! *sigh*
Time for bed.

TurkeyMan:
Here I was thinking this one would be the easiest one ever!

SSoulaimane:
https://github.com/dlang/dmd/pull/10601 should fix the mangling issue here.

thewilsonator:
Hmm, looks like this (still?) fails on FreeBSD.

wilzbach:
Can't we just disable the tests for FreeBSD? It's only partially supported anyhow and with FreeBSD 12 it won't be supported anymore.

TurkeyMan:
That freebsd error looks unrelated? O_o

src/core/stdcpp/utility.d
```diff
@@ -0,0 +1,35 @@
...
+    ///
+    alias first_type = T1;
+    ///
+    alias second_type = T2;
+
+    ///
+    T1 first;
+    ///
+    T2 second;
+}
```
SSoulaimane: The freebsd failure is real unfortunately. `pair` is define here as a POD while in C++ it isn't. A copy constructor definition here should fix it.

SSoulaimane: It seems `std::pair` is a POD by default on every platform except on freebsd.
https://github.com/llvm-mirror/libcxx/blob/41e7283b76a113234f40b72ab3b92644c0f04613/include/__config#L116-L118
https://github.com/llvm-mirror/libcxx/blob/8ef0e6a4fe63c27be4bddf80b8932133e102ba17/include/utility#L286-L301

n8sh: Adding an explicit constructor would mean that brace-style struct initializers for `pair` would work on all platforms other than FreeBSD. As an alternative, how about adding a trivial destructor on FreeBSD? I'll try pushing that and see if the tests then pass.

12345swordy:
@TurkeyMan @thewilsonator has freebsd been drop for the testing phase? It mostly green so far...

12345swordy:
@TurkeyMan Can we merge this?

src/core/stdcpp/utility.d
```diff
@@ -0,0 +1,43 @@
...
+    alias second_type = T2;
+
+    ///
+    T1 first;
+    ///
+    T2 second;
+
+    // FreeBSD has pair as non-POD so add a contructor
+    version (FreeBSD)
+    this(ref return scope inout(S) src) inout
```
MoonlightSentinel, line 38:
```suggestion
    this(ref return scope inout pair!(T1, T2) src) inout
```

TurkeyMan: Does this change have any affect, or do you just prefer it?

TurkeyMan:
@12345swordy I mean, I've love to merge it! It's pretty important!
...but the FreeBSD test still fails :/

src/core/stdcpp/utility.d
MoonlightSentinel: `S` isn't defined in the current scope:
```
/usr/home/braddr/sandbox/at-client/pull-4040470-FreeBSD_64_64/dmd/generated/freebsd/release/64/dmd -m64 -fPIC -w -I../../src -I../../import -Isrc -defaultlib= -debuglib= -dip1000 -L-lpthread -L-lm -L/usr/home/braddr/sandbox/at-client/pull-4040470-FreeBSD_64_64/druntime/generated/freebsd/release/64/libdruntime.a -O -release -extern-std=c++98 -main -unittest -version=CoreUnittest -c -of=./generated/freebsd/release/64/utility_d.o src/utility_test.d
../../src/core/stdcpp/utility.d(38): Error: undefined identifier `S`
```

TurkeyMan: Yeah I just noticed that from the test failure ;)

thewilsonator: Welp, thats my fault for just blindly copying the language docs without thinking.

thewilsonator:
I̵͎̖̼̜̥̙̓̂̎̓̉̆͐͆̚̕t̶̫̜̜̝̰͚̳̟̂̒̈͂͆͟ p̷̨̼̞͎̭̤̤̣̑͋̈́̒͐̀̈́͡͝a̮̯̬͎̟͈̳̠͗͐̅͑̐̋̓͛͜͠s̵̢̜̤̹͔̰̣̅̉̎͂̊͌̉͞s̢̨͙̮̯̔̑̀̓̐̍͘ę̜̣̪̙̼̣̰̐̿̀̊͗͂̌̚̚s̸̬̯̣̩͓̀́́̓͛̈́̂

TurkeyMan:
Huzzah! Why the other test fails?

thewilsonator:
The clang tests fail because of an issue with caching the LLVM packages, this is unrelated. Buildkite/phobos fails because it appears to be using an out of date druntime, this has occurred on other PRs in the past and is also unrelated.

PR #2788, opened 2019-09-08
[WIP] core.stdcpp.{exception,typeinfo}: Fix D/C++ symbol conflicts and undefined D vtable entries

kinke:
Fixes linker errors for the LDC druntime test runner.

Used source refs: [libstdc++](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/libsupc%2B%2B/exception), [libc++](https://android.googlesource.com/platform/ndk/+/5b3a49bdbd08775d0e6f9727221fe98946f6db44/sources/cxx-stl/llvm-libc++/libcxx/include/exception), Visual Studio 2017 headers

### DIFF
--- a/src/core/stdcpp/exception.d
+++ b/src/core/stdcpp/exception.d
@@ -67,13 +67,14 @@ version (GenericBaseException)
     class exception
     {
     @nogc:
+    extern(D):
         ///
         this() nothrow {}
         ///
-        ~this() nothrow {} // HACK: this should extern, but then we have link errors!
+        ~this() nothrow {}
 
         ///
-        const(char)* what() const nothrow { return "unknown"; } // HACK: this should extern, but then we have link errors!
+        const(char)* what() const nothrow { return "unknown"; }
 
     protected:
         this(const(char)*, int = 1) nothrow { this(); } // compat with MS derived classes
@@ -85,19 +86,20 @@ else version (CppRuntime_Microsoft)
     class exception
     {
     @nogc:
+    extern (D):
         ///
         this(const(char)* message = "unknown", int = 1) nothrow { msg = message; }
         ///
-        extern(D) ~this() nothrow {}
+        ~this() nothrow {}
 
         ///
-        extern(D) const(char)* what() const nothrow { return msg != null ? msg : "unknown exception"; }
+        const(char)* what() const nothrow { return msg != null ? msg : "unknown exception"; }
 
         // TODO: do we want this? exceptions are classes... ref types.
 //        final ref exception opAssign(ref const(exception) e) nothrow { msg = e.msg; return this; }
 
     protected:
-        void _Doraise() const {}
+        void _Doraise() const { assert(0); }
 
     protected:
         const(char)* msg;
@@ -111,6 +113,15 @@ else
 class bad_exception : exception
 {
 @nogc:
+extern(D):
+    private static immutable msg = "bad exception";
+
     ///
-    this(const(char)* message = "bad exception") { super(message); }
+    this(const(char)* message = msg.ptr) nothrow { super(message); }
+
+    version (GenericBaseException)
+    {
+        ///
+        override const(char)* what() const nothrow { return msg.ptr; }
+    }
 }
--- a/src/core/stdcpp/typeinfo.d
+++ b/src/core/stdcpp/typeinfo.d
@@ -70,12 +70,12 @@ else version (CppRuntime_Microsoft)
 
     class type_info
     {
-        //virtual ~this();
-        void dtor() { }     // reserve slot in vtbl[]
+    @nogc:
+        extern(D) ~this() nothrow {}
         //bool operator==(const type_info rhs) const;
         //bool operator!=(const type_info rhs) const;
-        final bool before(const type_info rhs) const;
-        final const(char)* name(__type_info_node* p = &__type_info_root_node) const;
+        final bool before(const type_info rhs) const nothrow;
+        final const(char)* name(__type_info_node* p = &__type_info_root_node) const nothrow;
 
     private:
         void* pdata;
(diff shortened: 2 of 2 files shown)


src/core/stdcpp/exception.d
```diff
@@ -112,5 +112,13 @@ class bad_exception : exception
 {
 @nogc:
     ///
-    this(const(char)* message = "bad exception") { super(message); }
+    this(const(char)* message = "bad exception") nothrow { super(message); }
+
+    version (GenericBaseException)
+    {
+        ///
+        ~this() nothrow;
```
kinke: Not declaring the dtor means it gets an implicit one (base class has one), and that one collides with the C++ one. [For MSVC, everything is defined inline in the C++ headers, so the implicit one does the job.]

src/core/stdcpp/exception.d
```diff
@@ -112,5 +112,13 @@ class bad_exception : exception
...
+        ///
+        override const(char)* what() const nothrow;
```
kinke: Required to have the vtable generated on the D side feature the right override in that vtable slot.

src/core/stdcpp/typeinfo.d
```diff
@@ -71,11 +71,11 @@ else version (CppRuntime_Microsoft)
...
     {
         //virtual ~this();
-        void dtor() { }     // reserve slot in vtbl[]
+        void dtor() nothrow { }     // reserve slot in vtbl[]
         //bool operator==(const type_info rhs) const;
         //bool operator!=(const type_info rhs) const;
-        final bool before(const type_info rhs) const;
-        final const(char)* name(__type_info_node* p = &__type_info_root_node) const;
+        final bool before(const type_info rhs) const nothrow;
+        final const(char)* name(__type_info_node* p = &__type_info_root_node) const nothrow;
```
kinke: That's actually the wrong signature for VS 2017 (which takes no params), but I guess DMD has to maintain VS 2010 compatibility or whatever.

src/core/stdcpp/typeinfo.d
```diff
@@ -85,13 +85,13 @@ else version (CppRuntime_Microsoft)
 
     class bad_cast : exception
     {
-        this(const(char)* msg = "bad cast");
+        this(const(char)* msg = "bad cast") nothrow { super(msg); }
```
kinke: Defined inline in C++ header.

src/core/stdcpp/typeinfo.d
```diff
@@ -108,8 +108,8 @@ else version (CppRuntime_Gcc)
 
     class type_info
     {
-        void dtor1();                           // consume destructor slot in vtbl[]
-        void dtor2();                           // consume destructor slot in vtbl[]
+        void dtor1() {}                         // consume destructor slot in vtbl[]
+        void dtor2() {}                         // consume destructor slot in vtbl[]
```
kinke: The vtable on the D side needs dummy definitions to put into the first 2 vtable slots.

src/core/stdcpp/typeinfo.d
```diff
@@ -131,16 +131,16 @@ else version (CppRuntime_Gcc)
 
     class bad_cast : exception
     {
-        this();
-        //~this();
-        override const(char)* what() const;
+        this() nothrow {}
+        ~this() nothrow;
```
kinke: Again, implicit dtor colliding with C++ one when linking in `libstdc++`.

kinke:
Due to Phobos including druntime, Phobos would need to be adapted, i.e., linking shared Phobos against libstdc++ too.

Geod24:
> Due to Phobos including druntime, Phobos would need to be adapted, i.e., linking shared Phobos against lib[std]c++ too.

That's a huge issue IMO. We can't link druntime/phobos with the C++ standard library for the sake of those bindings.

kinke:
It only affects the shared libraries and should thus just be an additional dependency to lib[std]c++.{so,dylib}. When using the static libraries, you only need the C++ runtime when using `core.stdcpp.{exception,typeinfo}`. If that's not acceptable (I don't have a strong opinion about this), then core.stdcpp.* should probably be extracted out of druntime.

JohanEngelen:
I also think it is bad if we need to link with the C++ standard library, even if it is only for the shared libraries. In my opinion that is too high a price to pay for the convenience of having C++ bindings in the D std library. Simply moving the bindings into their own separate package seems much more appropriate.

kinke:
Pinging @TurkeyMan.

src/core/stdcpp/typeinfo.d
TurkeyMan: O_o ... this is like 12 months out of date.

src/core/stdcpp/typeinfo.d
TurkeyMan: We have `-extern-std=`, so we can version the C++ code according to the C++ version it binds to.

TurkeyMan:
These are the exact 2 modules that I didn't write... I think the need a complete overhaul, but i'll give this patch a better look when I'm home.
Generally; local methods should be `extern(D)` so they don't collide with C++, and declare proper ctor/dtor's because they mangle and consume the vtable properly now.

TurkeyMan:
Where's this at now? I can't really see from the CI output what's going on here.
Are there multiple-definition errors? It looks like that's what I'd expect...

kinke:
Had some time to re-look into this and tried the `extern(D)` way as you suggested. Now the druntime testrunner compiles and links fine for LDC (https://github.com/ldc-developers/ldc/pull/3158), i.e., the D vtables can be populated; no idea what's going on with the failing subset of Mac stdcpp CI jobs here.

See the 2nd paragraph in https://github.com/ldc-developers/ldc/pull/3158#issuecomment-624885024 as to why `extern(D)` isn't the perfect solution when it comes to virtual functions and trying to avoid a C++ runtime dependency while not conflicting with its symbols. For LDC, `@weak` might be the solution (https://github.com/ldc-developers/ldc/pull/3424).

TurkeyMan:
Yeah, I was gonna say, `weak` linkage is the obvious answer, but Walter has rejected it for years.
Perhaps we can push for it again? We need to standardise a weak attribute, it's a PITA being compiler specific!

kinke:
> but Walter has rejected it for years

I guess that's solvable by providing him with a finished implementation. ;) - None of my business though.

ibuclaw:
Revived as #3552

PR #2789, opened 2019-09-08
POSIX: Add core.sys.posix.{string,strings,locale}

Geod24:
```
It looks like most of POSIX 2008 is missing from Druntime.
This particular subset is used by Sociomantic's Ocean,
and were ported in an effort to remove duplication.
```

In the process, I've also altered `core.stdc.string` in order to be more future proof (attributes on the right), and alphabetically ordered (as the online documentation is).

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -205,6 +205,7 @@ COPY=\
 	$(IMPDIR)\core\sys\posix\iconv.d \
 	$(IMPDIR)\core\sys\posix\inttypes.d \
 	$(IMPDIR)\core\sys\posix\libgen.d \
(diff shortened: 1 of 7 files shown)


thewilsonator:
`Failed to download package memutils from https://code.dlang.org/packages/memutils/0.4.11.zip`

Geod24:
Yep the doc tester will retrigger... Eventually

thewilsonator:
I thought it only retriggered on commit.

Geod24:
I asked a while back and it would retrigger from time to time (on merge I believe), but it does take a while, so I amended the commit.

PR #2790, opened 2019-09-11
demangle.d: remove use of lazy

WalterBright:
Since `lazy` is easily replaced with a delegate, I wonder if it is worth it at all. Should just use delegates.

### DIFF
--- a/src/core/demangle.d
+++ b/src/core/demangle.d
@@ -300,7 +300,7 @@ pure @safe:
     }
 
 
-    void silent( lazy void dg )
+    void silent( void delegate() pure @safe dg )
     {
         debug(trace) printf( "silent+\n" );
         debug(trace) scope(success) printf( "silent-\n" );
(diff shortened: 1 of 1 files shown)


src/core/demangle.d
```diff
@@ -1612,7 +1612,7 @@ pure @safe:
                 char t = front; // peek at type for parseValue
                 if ( t == 'Q' )
                     t = peekBackref();
-                char[] name; silent( name = parseType() );
+                char[] name; silent( delegate void() { name = parseType(); } );
```
Geod24:
```suggestion
                char[] name; silent( () { name = parseType(); } );
```

WalterBright: `return` expression expected

Geod24: Right, corrected

rainers:
> Since `lazy` is easily replaced with a delegate, I wonder if it is worth it at all.

So the arguments in https://dlang.org/articles/lazy-evaluation.html don't hold anymore?

WalterBright:
> don't hold anymore?

In looking through existing code, the only good use of lazy seems to be for logging messages. I had originally expected much wider use of it. Other languages don't seem to have such a feature. I suspect that making a more convenient delegate syntax at the call site would be better than adding the cognitive complexity of `lazy`.

PR #2794, opened 2019-09-13
core/internal/array.d: replace Error with assert()

WalterBright:
assert() is better


JinShil:
> assert() is better

I'm in favor of this for my own reasons, but I'd like to hear your justification.

WalterBright:
> your justification

Error and assert() do the same thing, we should just use one method. I'd like to eventually deprecate Error.

Geod24:
> Error and assert() do the same thing, we should just use one method. I'd like to eventually deprecate Error.

No they don't. First, we have `checkaction` to change the behavior. Second, in release mode (and we distribute druntime / phobos in release mode), the `assert` you are putting lead to a HLT, without error message, while `throw new Error` will lead to sensible error messages.

PR #2796, opened 2019-09-13
[stable] Fix regression for POSIX systems without backtrace[_symbols]

kinke:
Such as Android, OpenBSD, musl, ...

### DIFF
--- a/src/core/runtime.d
+++ b/src/core/runtime.d
@@ -745,8 +745,16 @@ unittest
     }
 }
 
-/// Default implementation for POSIX systems
-version (Posix) private class DefaultTraceInfo : Throwable.TraceInfo
+version (CRuntime_Glibc)       version = HasBacktrace;
+else version (Darwin)          version = HasBacktrace;
+else version (FreeBSD)         version = HasBacktrace;
+else version (NetBSD)          version = HasBacktrace;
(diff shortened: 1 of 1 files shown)


Geod24:
Hum sorry about this. Regarding `backtrace`, could we somehow combine the check with those in `runModuleUnitTests`, as they're doing the same thing.

Additionally, while `Musl` does not have `backtrace`, some systems using Musl (Alpine Linux) have implementations for it, such as `libexecinfo`.

kinke:
> Regarding backtrace, could we somehow combine the check with those in runModuleUnitTests, as they're doing the same thing.

That one includes Windows too, but most importantly, there we can simply use a `__traits(compiles, backtrace)`, as the local imports are before the check.
With this new `DefaultTraceInfo` class, we need to check the availability before the local imports.
Not sure if the usage of local `execinfo` imports (the two almost identical blocks) provides a benefit over just importing them once in the header.

Another thing that isn't really correct is `version (linux) enum enableDwarf = true;` in opApply - `rt.backtrace.dwarf` is actually only available for glibc/uclibc too. That module also doesn't support NetBSD and Solaris (for its private `has_backtrace` version), not sure if that's just an oversight or if it really cannot be used on those platforms.

PR #2797, opened 2019-09-14
Decouple Stack Context

baziotis:
Continuing from https://github.com/dlang/druntime/pull/2689

This PR decouples `Context` (now `StackContext`) from `Thread / Fiber` in the form of a super class for both.
Also, it adds a `GlobalStackContext` for commonly used global data.

@thewilsonator I did it kind of quickly, tell me what you think. I did not follow your diff precisely as some things seemed not necessary. One example is `push/popContext` etc. which seem to be relevant only to `Thread`.
Also, you seem to want a `LockGuardedBlock` for which I found no use and so it's not added yet.
Finally, it may be good to alias `GlobalStackContext` as it is used very much and it is a little verbose.


thewilsonator:
> One example is push/popContext etc. which seem to be relevant only to Thread.

Yes, but the operations only require the data members of `StackContextExecutor`. Also `osthread.d` is (still) a massive file anything that can logically be removed from it is a good thing.

> Also, you seem to want a LockGuardedBlock for which I found no use and so it's not added yet.

I think I was trying to factorise the horrible mess of globals somewhat. See the low level thread stuff at the bottom of `osthread.d`, although tbh it might be better to split that off completely in a separate PR and refactoring it there (making a `struct LowLevelThread { ThreadID tid; }` with the free functions as methods and the locks as static members. Be sure to ping Rainer for that one.

> Finally, it may be good to alias GlobalStackContext as it is used very much and it is a little verbose.

I don't think so, as its kind of the point to make it obvious that the variable being accessed is global (and therefore synchronisation may be required, and should be suspicious if absent).

baziotis:
> > One example is push/popContext etc. which seem to be relevant only to Thread.
> 
> Yes, but the operations only require the data members of `StackContextExecutor`.

Yes, but if we put them in `StackContextExecutor`, then it seems like `Fiber` needs them, while
it doesn't seem to. If that's ok, I'll put them there.

> Also `osthread.d` is (still) a massive file anything that can logically be removed from it is a good thing.

Yes, I'll see if I have something when the tests pass. Which btw, fail currently. If you happen to have any suggestion, in [this unittest](https://github.com/dlang/druntime/blob/757bc24523c49d1a692fa64c2fdd5ffbee964f25/src/core/thread/fiber.d#L1525) calling the `derived.call` or `composed.call` (and even having everything else commented out), yields a seg fault. GDB shows that it tries to call `m_dg` or `m_fn` and they're `null` - [here](https://github.com/dlang/druntime/pull/2797/files#diff-3234fff9533bd176714696695fe0df32R65).
Which is weird if you consider that to reach this point (because of the `switch`) , `m_call` has
to be set to `Call.DG`, which is set in `reset()`, which should set `m_dg` as well.

So.. the hard part on these refactorings is that something can break with 0 chances of finding why. :p

> > Also, you seem to want a LockGuardedBlock for which I found no use and so it's not added yet.
> 
> I think I was trying to factorise the horrible mess of globals somewhat. See the low level thread stuff at the bottom of `osthread.d`, although tbh it might be better to split that off completely in a separate PR and refactoring it there (making a `struct LowLevelThread { ThreadID tid; }` with the free functions as methods and the locks as static members. Be sure to ping Rainer for that one.
>

Ok great, interesting.

> > Finally, it may be good to alias GlobalStackContext as it is used very much and it is a little verbose.
> 
> I don't think so, as its kind of the point to make it obvious that the variable being accessed is global (and therefore synchronisation may be required, and should be suspicious if absent).

Ok, clear.

thewilsonator:
> Yes, but if we put them in StackContextExecutor, then it seems like Fiber needs them, while
it doesn't seem to.

Even if Fiber doesn't need it, having the stuff related solely to StackContext management all in one place is good practice.

> If that's ok, I'll put them there.

Please do.

src/core/thread/context.d
```diff
@@ -0,0 +1,312 @@
...
+    void* ehContext;
+
+    StackContext* within, next, prev;
+}
+
+/**
+A class that represents a thread of execution that manages a stack.
+This serves primarily as a superclass for Thread and Fiber.
+*/
+class StackContextExecutor
```
thewilsonator: part of the purpose of this class was to deduplicate as much as possible from Fiber and Thread. 

In particular:
* Thread's `m_curr` and Fiber's `m_ctxt` (this being a `StackContextExecutor` and all :) ) 
* Thread's `m_sz` and Fiber's `m_size`

are effectively the same

`m_unhandled` is identical.

There may be other ones as well.

src/core/thread/context.d
```diff
@@ -0,0 +1,312 @@
...
+        FN,
+        DG
+    }
+
+    // Common standard data for Thread / Fiber.
+    Call                m_call = Call.NO;
+    union
+    {
+        void function() m_fn;
+        void delegate() m_dg;
```
thewilsonator: It would be nice to package this up in a way that deduplicates the amount of work in the constructors of Thread and Fiber. something like 

```d
static struct Entrypoint
{
    enum Call {  NO, FN, DG }
    Call                m_call = Call.NO;
    union
     {
         void function() m_fn;
         void delegate() m_dg;
    }
    void opAssign(void function() fn) @trusted
    {
         m_fn = fn;
         m_call = Call.FN; 
    }

    // ditto for delegate

    void opCall()
    {
        run();
    }
}
```

src/core/thread/context.d
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    // Thread / Fiber entry point.  Invokes the function or delegate passed on
+    // construction (if any).
+    final void run()
```
thewilsonator: note that somehow this has triggered 
```
../.dub/packages/gtk-d-3.3.1/gtk-d/src/glib/Spawn.d(204,15): Error: function `glib.Spawn.Spawn.ReadFile.run` cannot override `final` function `core.thread.context.StackContextExecutor.run`
```
in build kite, why I'm not sure.

src/core/thread/context.d
baziotis: Ok, good. Yes, the first pass is to make the tests work and then one small change at a time. :)

src/core/thread/context.d
baziotis: Ok but it seems kind of risky at this point. I'll leave for the end.

thewilsonator: Oh, thats definitely for another PR, I just wanted to make a note of it while I remembered.

src/core/thread/context.d
baziotis: I can't find the source of either of these files _but_ the error most probably is something like [here](https://github.com/dlang/druntime/pull/2797/files#diff-897dcc1ea84bed3f92a8ba5e1bc6e619R1576). Note that the `private` was not there before. I put it to fix an identical error (following the example of other unittests).
So, the point is that this is _user_ code and it's not a problem of `Fiber`'s code.
But this may lead to breaking change in existing code.

thewilsonator: Yeah its probably a visibility/protection issue, you can ask on the forums for an explanation if you can't find a solution. (Alternately you could call it something else that won't cause a collision)

baziotis: Well, the problem is that a `final` function cannot be overriden, unless it is also `private`.
_Previously_: `run` was member of `Fiber` but it was `private` so overriding in e.g. `DerivedFiber` was ok.
_Now_: `run` is a member of `StackContextExecutor` and it can't be `private` as it has to be used by `Fiber`. Because of that, overriding it causes a problem.

I think of a couple solutions to that, of which the simplest is the one you said - chaning the name.

src/core/thread/context.d
```diff
@@ -28,6 +28,72 @@ struct StackContext
     StackContext* within, next, prev;
 }
 
+package(core.thread)
```
thewilsonator: this can be private, also please move this to the bottom of the file.

src/core/thread/context.d
thewilsonator: Hmm, perhaps doing https://github.com/dlang/druntime/pull/2797#discussion_r324444355 and having it as the opCall in this PR would be worth doing then. Saves having to come up with a name that won't collide. Or just move run back to Fiber for the time being and do that when you address https://github.com/dlang/druntime/pull/2797#discussion_r324444355 in a future PR. Up to you.

src/core/thread/context.d
```diff
@@ -72,6 +138,41 @@ class StackContextExecutor
             break;
         }
     }
+
```
thewilsonator:
```
src/core/thread/context.d:141:    
posix.mak:361: recipe for target 'checkwhitespace' failed
```
???

src/core/thread/context.d
baziotis: Ok, I'll see if there are any simple things left and do that.

src/core/thread/context.d
baziotis: No worries, I usually forget to check style. :)

thewilsonator: The `???` was not aimed at you, sorry. Usually it complains about spaces before the end of a line, bt as far as i ca tell there is no whitespace there at all! Oh well, I'm sure you'll figure out how to appease that useless checker somehow. 'Night.

baziotis: Hahaha. I probably have fought the style checker as few, but here, there is space. :P 
You know, it's: ---\n (where - is a whitespace) instead of: \n
Style checker - Stefanos: 1-0
Good night.

src/core/thread/context.d
```diff
@@ -28,6 +28,10 @@ struct StackContext
     StackContext* within, next, prev;
 }
 
+
+/// Flag to control rethrow behavior of call, join and friends
+enum Rethrow : bool { no, yes }
```
baziotis: I don't know how good of an idea is to put it here. But basically the reason that it can't go inside `StackContextExecutor` is that it's needed by `ThreadGroup`.

thewilsonator: Thats fine, this is also user facing code so I think it needs to anyway.

src/core/thread/fiber.d
```diff
@@ -518,7 +518,10 @@ private
  *
  * Authors: Based on a design by Mikola Lysenko.
  */
-class Fiber
+
+import core.thread.context : StackContext, StackContextExecutor;
```
thewilsonator: You have multiple import of the same module just collapse them into one, and the public import for `Rethrow` just do `alias Rethrow = .Rethow;` (or `alias Rethrow = core.thread.context.Rethrow;` if that doesn't work)

baziotis: Yes, the imports are a little bad at the moment, along with the versioning. The last one is good thought, thanks.

baziotis:
This was supposed to be a follow-up to https://github.com/dlang/druntime/pull/2821,
continuing the core.thread refactoring by decoupling the stack context.
I don't have time at the moment to continue it, but I'm leaving this for reference.

baziotis:
This was supposed to be a follow-up to https://github.com/dlang/druntime/pull/2821,
continuing the core.thread refactoring by decoupling the stack context.
I don't have time at the moment to continue it, but I'm leaving this for reference.

PR #2800, opened 2019-09-17
copy dmd's azure pipelines

thewilsonator:
so that breakages like #2689 won't happen again.


thewilsonator:
This is currently just a direct copy of DMD's it will likely need be adapted. (e.g. It doesn't appear to be triggering on this PR).

cc @Geod24

Geod24:
Yeah you probably need to configure something on Azure's side, for which you'll need @wilzbach .
And we should be dropping AppVeyor I suppose.

PR #2801, opened 2019-09-17
fix issue 13821 - fiber + exception + win server 2012 failures

rainers:
the Windows server exception chain validation fails because the initial exception entry is wrong in case the stack was aligned manually. 
See https://github.com/dlang/druntime/compare/master...rainers:issue13821?expand=1#diff-8bb12ed976acf0a5132e877ec5a01ea8L4717

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -4786,13 +4786,14 @@ private:
             pstack -= reserve;
             *(cast(EXCEPTION_REGISTRATION*)pstack) =
                 EXCEPTION_REGISTRATION( sehChainEnd, finalHandler );
+            auto pChainEnd = pstack;
 
             push( cast(size_t) &fiber_entryPoint );                 // EIP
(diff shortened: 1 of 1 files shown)


src/core/thread.d
```diff
@@ -4786,13 +4786,14 @@ private:
             pstack -= reserve;
             *(cast(EXCEPTION_REGISTRATION*)pstack) =
                 EXCEPTION_REGISTRATION( sehChainEnd, finalHandler );
+            auto pChainEnd = pstack;
```
Geod24: `const` ?

rainers: Not really the style of this file.

thewilsonator: Normally I'd disagree that that is a good reason not to do it, but the whole file is about to get a re-architect anyway, so it doesn't really matter.

Geod24: Fair enough

PR #2802, opened 2019-09-17
Issue 20219 - Idle D programs keep consuming CPU in Gcx.scanBackground

rainers:
background scan threads now wait indefinitely, with termination continuously triggering the condition until all threads have woken up

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -2716,6 +2716,7 @@ struct Gcx
     Event evDone;
 
     shared uint busyThreads;
+    shared uint stoppedThreads;
     bool stopGC;
 
     void markParallel(bool nostack) nothrow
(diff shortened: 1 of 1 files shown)


thewilsonator:
1) Does this fix this issue? If so the commit title needs to reflect that.
2) should this target stable?
3) I assume it is imposible to test this?

rainers:
> * Does this fix this issue? If so the commit title needs to reflect that.

Oops, missed to add "fix" when copying the title from the issue. Amended.

> * should this target stable?

I guess the next version (2.089) will start from master anyway. Or are we already past that point?

> * I assume it is imposible to test this?

All the existing tests run with the parallel marking, so the test for correct termination should be that the test suite doesn't run longer than before. Actually testing whether there are too many spurious wake ups of the background threads might be a bit invasive.

thewilsonator:
>  Amended.

Thanks.

> Or are we already past that point?

not sure, but targeting stable will make it into the betas if nothing else. I only asked because this is marked regression.

> Actually testing whether there are too many spurious wake ups of the background threads might be a bit invasive

Thats what I though was going to be the case.

rainers:
> Is the Buildkite Ocean failure related?

Maybe. Both failing tests are using fork()...

CyberShadow:
Speaking of `fork`, how is it handled by the new GC? It would need to detect it (using `atfork`) and manually start its own copy of the threads, right? It doesn't seem like it does that currently...

CyberShadow:
> Speaking of `fork`, how is it handled by the new GC?

I tried it and here's what I found:
https://issues.dlang.org/show_bug.cgi?id=20227

rainers:
Rebased to stable to pick up https://github.com/dlang/druntime/pull/2805

PR #2803, opened 2019-09-17
Add cppdemangle in druntime

ErnyTech:
Having a C++ symbol demangler is very useful, with this it will
be possible to implement the C++ symbol demangle in the stacktrace.

To avoid errors in the demangle implementation I used __cxa_demangle,
this function is standard on Itanium C++ ABI
(https://itanium-cxx-abi.github.io/cxx-abi/abi.html#demangler).

Itanium C++ ABI is used practically in all modern C++ compilers on Posix,
however old compilers (like GCC < 3.0) used a different name mangling,
this is not a problem because DMD in fact only supports Itanium C++ ABI on Posix.

This implementation obviously only supports Posix, on Windows it is not possible to use it.
You could probably do the same thing on Windows using
https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-undecoratesymbolname

### DIFF
--- a/src/core/internal/cppdemangle.d
+++ b/src/core/internal/cppdemangle.d
@@ -1,0 +1,199 @@
+/**
+* This module provides routines to demangle C++ symbols
+*
+* Demangling is done through the standard C++ `__cxa_demangle` function.
+* The C++ standard library is lazily loaded via `dlsym`. If not found, those routines will return their arguments.
+* This module currently only supports POSIX.
+*
+* Copyright: Copyright © 2019, The D Language Foundation
+* License: $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
+* Authors: Ernesto Castellotti
+* Source: $(DRUNTIMESRC core/internal/_cppdemangle.d)
+*/
+module core.internal.cppdemangle;
+
+/**
+* Demangles C++ mangled names.
+*
+* If it is not a C++ mangled name, it returns its argument name.
+*
+ * Params:
+ *  buf = The string to demangle.
+ *
+ * Returns:
+ *  The demangled name or the original string if the name is not a mangled C++
+ *  name.
+ * The returns is always allocated with malloc, use free to deallocate the memory.
+ */
+version (Posix)
+{
+    char[] cppdemangle(const(char)[] buf) @nogc @safe nothrow
+    {
+        char[] fakeDst = null;
+        return cppdemangle(buf, fakeDst);
+    }
+}
+
+extern(C++)  private class Test
+{
+    int var;
+}
+
+unittest
+{
+    version (Posix)  // Run tests only in Posix since is the only supported platform
+    {
+        import core.stdc.stdlib : free;
+        auto demangledName = cppdemangle("_Znwm");
+        assert(demangledName == "operator new(unsigned long)");
+        free(demangledName.ptr);
+        demangledName = cppdemangle(Test.var.mangleof);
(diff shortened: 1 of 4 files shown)


src/core/internal/cppdemangle.d
```diff
@@ -0,0 +1,196 @@
+/**
+* This module provide the demangle for converts mangled C++ symbols
```
Geod24:
```suggestion
* This module provides routines to demangle C++ symbols
*
* Demangling is done through the standard C++ `__cxa_demangle` function.
* The C++ standard library is lazily loaded via `dlsym`. If not found, those routines will return their arguments.
* This module currently only supports POSIX, Windows symbols are never demangled.
```

src/core/internal/cppdemangle.d
```diff
@@ -0,0 +1,196 @@
...
+*
+* Copyright: Copyright © 2019, The D Language Foundation
+* License: <a href="http://www.boost.org/LICENSE_1_0.txt">Boost License 1.0</a>.
+* Authors: Ernesto Castellotti
+* Source: $(DRUNTIMESRC core/internal/_cppdemangle.d)
+*/
+
+/**
+* Demangles C++ mangled names. If it is not a C++ mangled name, it returns its argument name.
```
Geod24: Take a look at how the doc looks like. My guess is that the second sentence should be part of the description (so you need a double newline, like in the `suggestion` above).

src/core/internal/cppdemangle.d
```diff
@@ -0,0 +1,196 @@
+/**
+* This module provide the demangle for converts mangled C++ symbols
+*
+* Copyright: Copyright © 2019, The D Language Foundation
+* License: <a href="http://www.boost.org/LICENSE_1_0.txt">Boost License 1.0</a>.
```
Geod24: That shouldn't be HTML

src/core/internal/cppdemangle.d
```diff
@@ -0,0 +1,196 @@
...
+*
+ * Params:
+ *  buf = The string to demangle.
+ *
+ * Returns:
+ *  The demangled name or the original string if the name is not a mangled C++
+ *  name.
+ * The returns is always allocated with malloc, use free to deallocate the memory.
+ */
+@nogc @trusted nothrow char[] cppdemangle(const(char)[] buf)
```
Geod24:
```suggestion
char[] cppdemangle(const(char)[] buf) @nogc @safe nothrow
```

src/core/internal/cppdemangle.d
```diff
@@ -0,0 +1,196 @@
...
+        free(demangledName.ptr);
+
+        extern(C++)  class Test
+        {
+            int var;
+        }
+
+        demangledName = cppdemangle(Test.var.mangleof);
+        writeln(demangledName);
+        assert(demangledName == "__unittest_L27_C1::Test::var");
```
Geod24: The `unittest` "function" is a D symbol. It would be much better to have this class defined outside of it.
But anyhow, this unittest depends on the C++ standard library to run, and I don't know if that's acceptable.

Geod24:
@ErnyTech : Do you have interest in pursuing this further ? This would be very, very valuable.

ErnyTech:
yes I am going to continue
The problem is that I don't have much time,

Geod24:
Great! Feel free to reach out on Slack if you want faster turnaround.

ErnyTech:
Superseded by #3002

PR #2804, opened 2019-09-18
Fix Issue 20222 -- druntime fails to build for Android

puneet:
Commit https://github.com/dlang/druntime/commit/d931921a adds a
dependency on _SC_PHYS_PAGES which is not defined in the module
core.sys.posix.unistd for Android. This breaks compilation of druntime
on/for Android platform. Porting this constant from Android bionic
source file
https://android.googlesource.com/platform/bionic/+/master/libc/include/bits/sysconf.h
fixes the issue.


thewilsonator:
Please target stable.

PR #2805, opened 2019-09-20
fix Issue 20227 - "Error: phread_mutex_destroy failed." after fork()

rainers:
clear thread handles and events for parallel marking in child process after fork

might also fix failing tests in https://github.com/dlang/druntime/pull/2802

### DIFF
--- a/src/core/thread.d
+++ b/src/core/thread.d
@@ -2045,6 +2045,16 @@ extern (C) void thread_init() @nogc
 
     version (Darwin)
     {
+        // thread id different in forked child process
+        static extern(C) void initChildAfterFork()
+        {
+            auto thisThread = Thread.getThis();
+            thisThread.m_addr = pthread_self();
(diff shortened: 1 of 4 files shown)


rainers:
Should be better rebased to stable?

thewilsonator:
Yes.

rainers:
Done.

CyberShadow:
On Darwin_64_64:
```
Testing forkgc
timelimit -t 10 ./generated/osx/debug/64/forkgc 
core.thread.ThreadError@src/core/thread.d(3093): Unable to load thread state
----------------
make[2]: *** [generated/osx/release/64/forkgc.done] Error 143
```

rainers:
Turns out forking under OSX didn't really work at all (when thread handles were needed in the child process). Any more involved program will probably also fail on Linux, e.g. because the chain of threads is invalid.

PR #2806, opened 2019-09-20
try using dmd's azure-pipeline

rainers:
Will need some adaptions in dmd's files...

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -61,25 +61,4 @@ jobs:
     steps:
       - script: git clone --branch $(DMD_BRANCH) https://github.com/dlang/dmd.git ../dmd
       - script: cd ../dmd && sh --login .azure-pipelines/windows-visual-studio.sh
-      - script: cd ../dmd && call .azure-pipelines/windows-msbuild.bat
-
-  - job: Windows_VisualD_Debug
-    timeoutInMinutes: 120
-    pool:
-      vmImage: 'vs2017-win2016'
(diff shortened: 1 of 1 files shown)


rainers:
Using YAML templates from another repo doesn't seem to work, as they are read before any command is executed. Calling bash/cmd instead needs a bit more duplication, but with https://github.com/dlang/dmd/pull/10421 it should be rather small.

This PR could work with the dmd PR, but is currently only using dmd master (the first two jobs passed on a personal pipeline).

@wilzbach Can you enable azure for druntime?

rainers:
BTW: here is the azure run on my account: https://rainers8.visualstudio.com/druntime/_build/results?buildId=20

thewilsonator:
@wilzbach ping

wilzbach:
Sorry, added it to Azure. thanks a lot @rainers! I also sent out an invite to the dlanguage Azure group to you in case access is required again in the future!

wilzbach:
FYI: I also configured Azure Pipelines for Phobos and tools in case it comes ever up as I presume we can do a similar PR to Phobos once this is passing.

rainers:
> I also sent out an invite to the dlanguage Azure group to you in case access is required again in the future!

Thanks, but it seems that went to the wrong account. My MS account is rainers8 at hotmail dot com.

rainers:
> Will need some adaptions in dmd's files...

It needs https://github.com/dlang/dmd/pull/10421

PR #2811, opened 2019-10-02
fix issue 20256

ikod:
This code fails with dmd 2.087 and 2.088 if parallel GC scanning enabled.
```
import std.stdio;
import core.sys.posix.signal;
import core.thread;
import core.sys.posix.unistd;
import core.sys.posix.sys.wait;
import core.memory;
import std.algorithm;
import std.range;

void block(int signum) {
    sigset_t m;
    sigemptyset(&m);
    sigaddset(&m, signum);
    sigprocmask(SIG_BLOCK, &m, null);
}

void main()
{
    auto x = new int[](10000);
    iota(10000).each!(i => x ~= i); 
    GC.collect();  // startScanThreads create threads here
    block(SIGHUP); // block works only for current thread, not for thread created by GC.collect, so scanthread receive signal and call default handler (kill program)
    auto parent_pid = getpid();
    auto child_pid = fork();
    if ( child_pid == 0 ) { 
        //child
        kill(parent_pid, SIGHUP); // send signal to parent
        _exit(0);
    }   
    if ( child_pid == -1 ) { 
        _exit(0);
    }   
    Thread.sleep(1.seconds);
    writeln("This should be printed");
}
```
Expected output "This should be printed". Instead program killed by signal SIGHUP. This code works as expected in dmd 2.083 or when program started with --DRT-gcopt=parallel:0 with dmd 2.087, 2088.

This is because SIGHUP is not masked in GC thread (as if it were cloned after calling sigprocmask from the main thread).

This PR solve this problem by blocking all signals in scan threads under Linux.


src/gc/impl/conservative/gc.d
```diff
@@ -2854,6 +2854,16 @@ struct Gcx
 
     void scanBackground() nothrow
     {
+        version(linux) {
+            import core.sys.posix.signal;
+
+            // block all signals, so that scanThread will not receive a signal not intended for him.
+			// see https://issues.dlang.org/show_bug.cgi?id=20256
+            sigset_t new_mask;
+            sigfillset(&new_mask);
```
thewilsonator: Please make the indenting consistent with the code below.

thewilsonator:
cc @rainers

src/gc/impl/conservative/gc.d
ikod: Sorry, should be ok now.

thewilsonator: This is still indented by two spaces. Also the please place the brace in `version(linux) {` on a new line.

ikod: sorry for this stupid noise. Is it ok now?

src/gc/impl/conservative/gc.d
```diff
@@ -2854,15 +2854,16 @@ struct Gcx
...
             // block all signals, so that scanThread will not receive a signal not intended for him.
             // see https://issues.dlang.org/show_bug.cgi?id=20256
             sigset_t new_mask;
             sigfillset(&new_mask);
-			int sigmask_rc = pthread_sigmask(SIG_BLOCK, &new_mask, null);
-			assert(sigmask_rc == 0, "failed to set up GC scan thread sigmask");
+            int sigmask_rc = pthread_sigmask(SIG_BLOCK, &new_mask, null);
+            assert(sigmask_rc == 0, "failed to set up GC scan thread sigmask");
 		}
         while (!stopGC)
```
thewilsonator: These two ought to line up. I think you need to indent the entire `version(linux)` block by two spaces.

thewilsonator:
Thanks.

src/gc/impl/conservative/gc.d
ikod: sorry, it was new unconfigured vim instance...

src/gc/impl/conservative/gc.d
```diff
@@ -2854,6 +2854,17 @@ struct Gcx
 
     void scanBackground() nothrow
     {
+        version(linux)
```
thewilsonator:
```
--
  | grep -nrE "\<(for\|foreach\|foreach_reverse\|if\|while\|switch\|catch\|version)\(" $(find src -name '*.d') ; test $? -eq 1
  | src/gc/impl/conservative/gc.d:2857:        version(linux)
```
you need to add a space, like : `version (linux)`

ikod:
> Include the test contained in the issue(?)

added unittest to gc.d.

You may find it is better to run this test as separate file, as it change (but carefully restore) some global state - sigmask. Just let me know.

src/gc/impl/conservative/gc.d
```diff
@@ -2854,6 +2854,17 @@ struct Gcx
 
     void scanBackground() nothrow
     {
+        version (linux)
```
rainers: Is this really specific to linux, or would Posix be the better version?

rainers:
> You may find it is better to run this test as separate file, as it change (but carefully restore) some global state - sigmask. Just let me know.

Forking might have a couple of more side effects that interfere with other tests, e.g. existing threads will not be handled correctly. I'd recommend making this a separate test similar to https://github.com/dlang/druntime/blob/stable/test/gc/forkgc.d (this PR might better target stable, too).

src/gc/impl/conservative/gc.d
```diff
@@ -4469,3 +4480,41 @@ unittest
...
+    import core.thread;
+    import core.memory;
+
+    int sig;
+    sigset_t m, o;
+    sigemptyset(&m);
+    sigaddset(&m, SIGHUP);
+
+    auto x = new int[](10000);
+    for(int i=0;i<10000;i++)
```
rainers: The style checker expects a space after `for` and a few more for the expressions might be good, too.

src/gc/impl/conservative/gc.d
ikod: I think it should be valid for any Posix system, although I have access only to linux and OSX.

I don't have such problem with dmd 2.087 under OSX, so I targeted this patch for linux only. I can spend some time to investigate why OSX is not affected.

thewilsonator:
You need to rebase this on stable, not just change the branch (it pull all of master along with it).

ikod:
> You need to rebase this on stable, not just change the branch (it pull all of master along with it).

Sorry, I make this first time.

Locally I made
```
git rebase --onto origin/stable origin/master
```
my git/config
```
[remote "origin"]
	url = https://github.com/ikod/druntime.git
	fetch = +refs/heads/*:refs/remotes/origin/*
[branch "master"]
	remote = origin
	merge = refs/heads/master
[branch "stable"]
	remote = origin
	merge = refs/heads/stable
[branch "fixIssue20256"]
	remote = origin
	merge = refs/heads/fixIssue20256
```
Then I changed base for the PR.
Is there something wrong here?

rainers:
> git rebase --onto origin/stable origin/master

Maybe your current branch was not the PR? I'm not a git command line expert, but this is supposed to work:
```
git rebase --onto stable master fixIssue20256
```

ikod:
@rainers 
> > git rebase --onto origin/stable origin/master
> 
> Maybe your current branch was not the PR? I'm not a git command line expert, but this is supposed to work:
> 
> ```
> git rebase --onto stable master fixIssue20256
> ```

I'll make new PR, based on stable, as I do not want to make experiments with git on this repo.

PR #2812, opened 2019-10-03
Set up CI with Azure Pipelines

wilzbach:


wilzbach:
Sorry for the spam. Unfortunately, this is required to create an Azure Pipeline.

PR #2813, opened 2019-10-05
problem with signal handling and parallel GC on linux

ikod:
Issue https://issues.dlang.org/show_bug.cgi?id=20256

Conservative GC scan threads do not block signals when created, which can lead to unexpected behaviour if developer have to block signals for whole process.

This PR fix issue  by blocking all signals in conservative GC scanThreads.

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -2822,8 +2822,25 @@ struct Gcx
         evStart.initialize(false, false);
         evDone.initialize(false, false);
 
+        version (Posix)
+        {
+            import core.sys.posix.signal;
+            // block all signals, scanBackground inherits this mask.
+            // see https://issues.dlang.org/show_bug.cgi?id=20256
+            sigset_t new_mask, old_mask;
+            sigfillset(&new_mask);
+            auto sigmask_rc = pthread_sigmask(SIG_BLOCK, &new_mask, &old_mask);
+            assert(sigmask_rc == 0, "failed to set up GC scan thread sigmask");
+        }
+
         for (int idx = 0; idx < numScanThreads; idx++)
             scanThreadData[idx].tid = createLowLevelThread(&scanBackground, 0x4000, &stopScanThreads);
+
+        version (Posix)
+        {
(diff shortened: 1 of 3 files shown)


src/gc/impl/conservative/gc.d
```diff
@@ -2874,6 +2874,18 @@ struct Gcx
 
     void scanBackground() nothrow
     {
+        version (posix)
```
rainers: It's `Posix`, not `posix`.

src/gc/impl/conservative/gc.d
```diff
@@ -2822,9 +2822,26 @@ struct Gcx
         evStart.initialize(false, false);
         evDone.initialize(false, false);
 
+        version (Posix)
+        {
+            import core.sys.posix.signal;
+            // block all signals, scanThread inherits this mask.
```
rainers: I guess you mean `scanBackground` here.

test/gc/sigmaskgc.d
```diff
@@ -0,0 +1,42 @@
...
+        import core.sys.posix.unistd;
+        import core.thread;
+        import core.memory;
+
+        sigset_t m;
+        sigemptyset(&m);
+        sigaddset(&m, SIGHUP);
+
+        auto x = new int[](10000);
+        for (int i=0; i<10000; i++)
```
rainers: code style nit: `for (int i = 0; i < 10000; i++)`

test/gc/sigmaskgc.d
```diff
@@ -0,0 +1,42 @@
...
+            x ~= i;
+        }
+        GC.collect();  // GC create thread
+
+        sigprocmask(SIG_BLOCK, &m, null); // block SIGHUP from delivery to main thread
+
+        auto parent_pid = getpid();
+        auto child_pid = fork();
+        assert(child_pid >= 0);
+        if ( child_pid == 0 )
```
rainers: nit: `if (child_pid == 0)`

test/gc/sigmaskgc.d
wilzbach: Bonus for `foreach (i; 0 .. 10000)`

ikod:
Thanks!

PR #2815, opened 2019-10-06
syscall: add syscall for missing platform specific implementations like `gettid`

carun:


wilzbach:
See https://github.com/dlang/druntime/pull/1984

carun:
@wilzbach Nice. I'm abandoning this in favour of yours.

PR #2816, opened 2019-10-06
fix Issue 20270 - [REG2.087] Deadlock in garbage collection when runn…

rainers:
…ing processes in parallel

start scan threads while the world isn't suspended

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -2870,8 +2870,12 @@ struct Gcx
             if (scanThreadData[idx].tid != scanThreadData[idx].tid.init)
                 startedThreads++;
 
+        version (Windows)
+            alias allThreadsDead = thread_DLLProcessDetaching;
+        else
+            enum allThreadsDead = false;
         stopGC = true;
-        while (atomicLoad(stoppedThreads) < startedThreads)
+        while (atomicLoad(stoppedThreads) < startedThreads && !allThreadsDead)
         {
             evStart.set();
             evDone.wait(dur!"msecs"(1));
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -2597,6 +2597,15 @@ struct Gcx
         //printf("\tpool address range = %p .. %p\n", minAddr, maxAddr);
 
         {
+            version (COLLECT_PARALLEL)
+            {
+                bool doParallel = config.parallel > 0;
+                if (doParallel && !scanThreadData)
+                    startScanThreads();
+            }
+            else
+                enum doParallel = false;
+
             // lock roots and ranges around suspending threads b/c they're not reentrant safe
             rangesLock.lock();
             rootsLock.lock();
@@ -2614,11 +2623,6 @@ struct Gcx
             stop = currTime;
             prepTime += (stop - start);
             start = stop;
-
-            version (COLLECT_PARALLEL)
-                bool doParallel = config.parallel > 0;
-            else
-                enum doParallel = false;
 
             if (doParallel)
             {
@@ -2743,9 +2747,6 @@ struct Gcx
 
         void** pbot = toscanRoots._p;
         void** ptop = toscanRoots._p + toscanRoots._length;
-
-        if (!scanThreadData)
-            startScanThreads();
 
         debug(PARALLEL_PRINTF) printf("markParallel\n");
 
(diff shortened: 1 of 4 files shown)


thewilsonator:
Is it possible to test?

CyberShadow:
Thank you! Confirmed the fix on my machine.

> Is it possible to test?

There is a small test program in the bug report. It should be suitable to be added to the test suite.

rainers:
> There is a small test program in the bug report. It should be suitable to be added to the test suite.

It depends on phobos which is not allowed in druntime. The root cause is much more basic, but I'm not sure it is easy to reproduce.

CyberShadow:
I'll try to create a smaller test case.

CyberShadow:
Reduced test case:

```d
import core.stdc.stdlib : exit;
import core.sys.posix.unistd : fork;
import core.thread : Thread;

void main()
{
	foreach (t; 0 .. 10)
		new Thread({
			foreach (n; 0 .. 100)
			{
				foreach (x; 0 .. 100)
					new ubyte[x];
				auto f = fork();
				assert(f >= 0);
				if (f == 0)
					exit(0);
			}
		}).start();
}
```

CyberShadow:
I think I found a new bug while trying to create a test case:

```d
import core.stdc.stdlib : exit;
import core.sys.posix.sys.wait;
import core.sys.posix.unistd : fork;
import core.thread : Thread;

void main()
{
	foreach (t; 0 .. 10)
		new Thread({
			foreach (n; 0 .. 100)
			{
				foreach (x; 0 .. 100)
					new ubyte[x];
				auto f = fork();
				assert(f >= 0);
				if (f == 0)
					exit(0);
				else
					waitpid(f, null, 0);
			}
		}).start();
}
```

This program has a high chance of never exiting, because it is waiting on forked processes which are deadlocked. The forked processes have a single thread in `core.internal.spinlock.SpinLock.yield` (called by `Gcx.removeRange`).

It looks like the GC lock was not cleared after a fork?

CyberShadow:
> Reduced test case:

For this bug / PR:

```d
import core.sys.posix.sys.wait : waitpid;
import core.sys.posix.unistd : fork, _exit;
import core.thread : Thread;

void main()
{
	foreach (t; 0 .. 10)
		new Thread({
			foreach (n; 0 .. 100)
			{
				foreach (x; 0 .. 100)
					new ubyte[x];
				auto f = fork();
				assert(f >= 0);
				if (f == 0)
					_exit(0);
				else
					waitpid(f, null, 0);
			}
		}).start();
}
```

CyberShadow:
> I think I found a new bug while trying to create a test case:

I meant to call `_exit` instead of `exit`. The latter calls finalizers incl. GC destruction, but the fork might have happened while the GC lock was being held. This is a very separate issue. Probably, it would be good to keep the GC working after a fork, so it would need to clear its lock in its `atfork` handler.

rainers:
Thanks, I added the test case. 

> The latter calls finalizers incl. GC destruction, but the fork might have happened while the GC lock was being held. This is a very separate issue. Probably, it would be good to keep the GC working after a fork, so it would need to clear its lock in its `atfork` handler.

I think there are more issues in druntime like that. For example the chain of Threads all point to dead threads in the fork-child, so joining these will fail, too.

CyberShadow:
> I think there are more issues in druntime like that. For example the chain of Threads all point to dead threads in the fork-child, so joining these will fail, too.

Hmm, probably we should try to fix them one by one.

What do you think about https://github.com/dlang/druntime/pull/2817 ?

rainers:
@CyberShadow This is working now, unfortunately a day too late for the release.

wilzbach:
@rainers FYI: the master-stable branchoff hasn't happened yet.

rainers:
@wilzbach I meant the 2.088.1 release, this PR is targeting stable.

src/gc/impl/conservative/gc.d
```diff
@@ -2871,8 +2871,12 @@ struct Gcx
             if (scanThreadData[idx].tid != scanThreadData[idx].tid.init)
                 startedThreads++;
 
+        version (Windows)
+            alias allThreadsDead = thread_DLLProcessDetaching;
+        else
+            enum allThreadsDead = false;
```
CyberShadow: So what happened here? A DLL with running threads was unloaded, terminating them (causing them to segfault)?

rainers: No, the process terminates while threads are still running. The threads are just terminated as well without `DllMain(DLL_THREAD_DETACH)` being called. `DllMain(DLL_PROCESS_DETACH) is still ìnvoked, though.

rainers:
Rebased. Would be good to have this in the coming release as the Visual D installer is hit by the non-termination issue, too.

PR #2817, opened 2019-10-06
Add fork handling to GC

CyberShadow:
Make sure a fork does not happen while GC code is running, thus leaving it in an inconsistent state.

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -1251,6 +1251,7 @@ struct Gcx
         alias leakDetector = LeakDetector;
 
     SmallObjectPool*[B_NUMSMALL] recoverPool;
+    version (Posix) __gshared Gcx* instance;
 
     void initialize()
     {
@@ -1262,9 +1263,22 @@ struct Gcx
         usedSmallPages = usedLargePages = 0;
         mappedPages = 0;
         //printf("gcx = %p, self = %x\n", &this, self);
+        version (Posix)
+        {
+            import core.sys.posix.pthread : pthread_atfork;
+            instance = &this;
+            __gshared atforkHandlersInstalled = false;
+            if (!atforkHandlersInstalled)
+            {
+                pthread_atfork(
+                    &_d_gcx_atfork_prepare,
+                    &_d_gcx_atfork_parent,
+                    &_d_gcx_atfork_child);
+                atforkHandlersInstalled = true;
+            }
+        }
         debug(INVARIANT) initialized = true;
     }
-
 
     void Dtor()
     {
@@ -1310,6 +1324,8 @@ struct Gcx
                    pauseTime, maxPause, apitxt.ptr);
         }
 
+        version (Posix)
+            instance = null;
         version (COLLECT_PARALLEL)
             stopScanThreads();
 
@@ -2841,13 +2857,6 @@ struct Gcx
             sigmask_rc = pthread_sigmask(SIG_SETMASK, &old_mask, null);
             assert(sigmask_rc == 0, "failed to set up GC scan thread sigmask");
         }
-
-        version (Posix)
-        {
-            import core.sys.posix.pthread;
-            forkedGcx = &this;
-            pthread_atfork(null, null, &initChildAfterFork);
-        }
     }
 
     void stopScanThreads() nothrow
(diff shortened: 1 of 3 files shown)


src/gc/impl/conservative/gc.d
```diff
@@ -1118,6 +1128,41 @@ class ConservativeGC : GC
...
+        }
+
+        extern(C) static void _d_gcx_atfork_parent()
+        {
+            gcLock.unlock();
+        }
+
+        extern(C) static void _d_gcx_atfork_child()
+        {
+            gcLock.unlock();
```
rainers: Shouldn't this reinitialize the lock?

CyberShadow: Sorry. Reinitialize how? I understand that the lock here is a spinlock with yield/sleep backoff, so since it's just a value in memory there's nothing to reinitialize.

src/gc/impl/conservative/gc.d
```diff
@@ -2883,8 +2921,6 @@ struct Gcx
         cstdlib.free(scanThreadData);
         // scanThreadData = null; // keep non-null to not start again after shutdown
         numScanThreads = 0;
-        version (Posix)
-            forkedGcx = null;
```
rainers: I think this cleanup and the check in the callback should be kept in case the runtime is started and stopped manually and a fork call happens later.

CyberShadow: You're right, forgot about that. Fixed.

src/gc/impl/conservative/gc.d
rainers: You are right, SpinLock doesn't have an initialization, but unlock happens to have the same effect. I was rather thinking of a pthread mutex where a handle might be involved.

CyberShadow: It might be worth looking into switching to a mutex, as I've seen situations where CPU-bound programs are idle in all threads because of the `Thread.sleep(1.msecs)` backoff. But that's something else.

src/gc/impl/conservative/gc.d
```diff
@@ -193,6 +208,8 @@ class ConservativeGC : GC
             cstdlib.free(gcx);
             gcx = null;
         }
+        version (Posix)
+            Gcx.instance = gcx;
```
rainers: I think this pointer should be reset before gcx is destroyed. BTW: as only members of gcx are affected, how about moving this and the initialization into Gcx.Dtor and initialize, respectively?

CyberShadow: The lock is in `ConservativeGC`, and we need to refer to it from the callbacks. Accessing `ConservativeGC` from `Gcx` which seemed to me like a layering violation, but I see now there is precedent.

CyberShadow:
I found a new bug while playing around to create a test case for this PR.

```d
import core.stdc.stdio : printf;
import core.stdc.stdlib : exit;
import core.sys.posix.sys.types : pid_t;
import core.sys.posix.sys.wait : waitpid;
import core.sys.posix.unistd : fork;
import core.thread : Thread, getpid;

extern(C) int syscall(int, ...);
pid_t gettid() { return syscall(186/*SYS_gettid*/); }

void main()
{
	foreach (t; 0 .. 10)
		new Thread({
			auto parent = gettid();
			foreach (n; 0 .. 100)
			{
				foreach (x; 0 .. 100)
					new ubyte[x];
				auto child = fork();
				assert(child >= 0);
				if (child == 0)
				{
					child = gettid();
					printf("%d has forked.\n", child);
					foreach (x; 0 .. 100)
						new ubyte[x];
					printf("%d is exiting.\n", child);
					exit(0);
				}
				else
				{
					printf("%d waiting on %d...\n", parent, child);
					waitpid(child, null, 0);
					printf("%d done waiting on %d.\n", parent, child);
				}
			}
		}).start();
}
```

This program doesn't exit for me, because some threads are locked in a `waitpid`. The strange thing, however, is that the process they're waiting on actually has exited and is a zombie. I don't yet understand what is happening.

src/gc/impl/conservative/gc.d
```diff
@@ -2942,20 +2949,40 @@ struct Gcx
...
+        {
+            if (instance)
+                ConservativeGC.lockNR();
+        }
+
+        extern(C) static void _d_gcx_atfork_parent()
+        {
+            if (instance)
+                ConservativeGC.gcLock.unlock();
+        }
```
CyberShadow: I will probably move these out to a general Druntime fork handler which also takes care of the thread stuff.

rainers: That probably won't work as the GC can be swapped out at link time.

CyberShadow: Not sure what you mean. The atfork notifications would then be part of the abstract GC interface.

PR #2819, opened 2019-10-06
Added `pause` to core.atomic

TurkeyMan:
So that other architectures can add their special sauce.

### DIFF
--- a/changelog/atomic_cas.dd
+++ b/changelog/atomic_cas.dd
@@ -1,4 +1,4 @@
-Added overloads for `core.atomic.cas` which don't discard their result.
+Added overloads for `core.atomic : cas` which don't discard their result
 
-Existing `core.atomic.cas` functions discard the result, which make it impossible to implement certain operations.
+Existing `core.atomic : cas` functions discard the result, which make it impossible to implement certain operations.
 A new set of overloads was added which take `ifThis` by pointer and write the result back to the argument.
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -466,7 +466,7 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 *  true if the store occurred, false if not.
 */
 bool casWeak(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V)(T* here, T* ifThis, V writeThis) pure nothrow @nogc @trusted
-if (!is(T == shared S, S) && !is(V == shared U, U))
+    if (!is(T == shared S, S) && !is(V == shared U, U))
 in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 {
     // resolve implicit conversions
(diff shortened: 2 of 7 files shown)


src/core/atomic.d
```diff
@@ -466,7 +466,7 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 *  true if the store occurred, false if not.
 */
 bool casWeak(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V)(T* here, T* ifThis, V writeThis) pure nothrow @nogc @trusted
-if (!is(T == shared S, S) && !is(V == shared U, U))
+    if (!is(T == shared S, S) && !is(V == shared U, U))
```
wilzbach: Before your change this was the official DStyle (dlang.org/dstyle).

TurkeyMan: It was inconsistent with the rest of the file... I would always favour internal consistency :/

CyberShadow:
Maybe refactor the one in `core.internal.spinlock` to use your implementation?

TurkeyMan:
Those spinlock backoff counters look sus; they're *very* small numbers.
With backoff set to 4/16 cycles, it's not really a spinlock at all! It's just a sleep-lock.
NT kernel uses ~4000 cycles before backoff... and it probably shouldn't be called a "spin lock" if it backs off.

CyberShadow:
> Those spinlock backoff counters look sus; they're _very_ small numbers.

Yes, [I ran into issues with that too](https://github.com/dlang/druntime/pull/2817#discussion_r331786943).

TurkeyMan:
Yeah, it's not a spin-lock... and it's also really bad >_<

TurkeyMan:
Is this controversial?

changelog/atomic_cas.dd
```diff
@@ -1,4 +1,4 @@
-Added overloads for `core.atomic.cas` which don't discard their result.
+Added overloads for `core.atomic : cas` which don't discard their result
```
MoonlightSentinel: Unrelated style change
Ditto for all other changelogs and template constraints

src/core/atomic.d
```diff
@@ -519,11 +519,20 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
  * that all loads and stores before a call to this function are executed before any
  * loads and stores after the call.
  */
-void atomicFence(MemoryOrder order = MemoryOrder.seq)() nothrow @nogc @safe
+void atomicFence(MemoryOrder order = MemoryOrder.seq)() pure nothrow @nogc @safe
```
MoonlightSentinel: Unrelated to pause(), please move this into another PR
Ditto for all other `pure` additions

src/core/internal/spinlock.d
```diff
@@ -52,7 +52,7 @@ shared struct SpinLock
     {
         import core.time;
         if (k < pauseThresh)
-            return pause();
+            return core.atomic.pause();
```
MoonlightSentinel: Correct if im wrong but this is more than a simple refactoring as the new inline assembly differs:
Old:
```
rep
nop
```
New:
```
naked
rep
nop
ret
```

src/core/internal/atomic.d
```diff
@@ -645,6 +645,33 @@ void atomicFence(MemoryOrder order = MemoryOrder.seq)() nothrow @nogc @safe
...
+    version (D_InlineAsm_X86)
+    {
+        asm pure nothrow @nogc @trusted
+        {
+            naked;
+            rep; nop;
+            ret;
+        }
+    }
+    else version (D_InlineAsm_X86_64)
```
MoonlightSentinel: Keep the old enum/version declaration instead of duplicating the assembly? Or should there be different implementations according to your comment regarding the `pause` opcode?

MoonlightSentinel:
Damn ...

src/core/atomic.d
wilzbach: We have a style guide for a reason!!

PR #2820, opened 2019-10-06
linux: add system call bindings

carun:
Reviving #1984


ZombineDev:
Remaining style issues:
```
make -f posix.mak style
grep -n -U -P "([ \t]$|\r)" $FILES ; test $? -eq 1
src/core/sys/linux/unistd.d:3:version(linux):
src/core/sys/linux/syscalls.d:3:version(linux):
src/core/sys/linux/syscalls.d:10:version(CoreDdoc)
src/core/sys/linux/syscalls.d:15:version(X86_64)
src/core/sys/linux/syscalls.d:356:else version(X86)
posix.mak:423: recipe for target 'style_lint' failed
make: *** [style_lint] Error 1
```

carun:
@ZombineDev not sure why some checks are failing. Do you mind explaining? Thanks.

thewilsonator:
```
../druntime/src/core/sys/linux/syscalls.d(19): Error: enum `core.sys.linux.syscalls.SYS` conflicts with enum `core.sys.linux.syscalls.SYS` at ../druntime/src/core/sys/linux/syscalls.d(13)
```

Try 

```
version (CoreDdoc)
 {
     /// Linux system call number from Linux's asm/unistd.h
     enum SYS : c_long;
 }
 _else_ version (X86_64)
 {
    //...
}
```

carun:
@thewilsonator I'm not sure why it still fails. The errors seem to indicate unittest failure. Apparently, I don't see the errors explaining where/why it failed.

rainers:
The Windows failures look unrelated, maybe the additional files trigger some compiler bug. The win32-ldc failure looks like the compiler silently exits while compiling the unittest executable. I haven't been able to reproduce locally, though.

carun:
@ZombineDev @thewilsonator help pls!

rainers:
The failing test hints at a problem: PR https://github.com/dlang/druntime/pull/2821 has undone the changes in https://github.com/dlang/druntime/pull/2801. Not sure why it has passed the CI, though.

wilzbach:
As I pointed out in my PR, the [mir-kernel](https://github.com/libmir/mir-linux-kernel) package is far superior. It's all Boost, so the least we could do here would be to do a proper copy/paste as otherwise soon users on ARM etc. will have to add their bindings as well.

carun:
@wilzbach I can redo this PR with a copy paste from https://github.com/libmir/mir-linux-kernel. Honestly, the copy-paste makes me cringe. But we can't have Phobos to depend on libmir. I see no other solution at the moment!

ZombineDev:
@carun libmir/mir-linux-kernel exists for 2 reasons:
1. Druntime didn't offer the functionality
2. It's a dub package which makes it easy to use, independently of the compiler used.

We can fix 1 and then after all compilers catch up in a couple of releases `mir-linux-kernel` will stop needing to exist.

I suggest we move forward by adding all the functionality to druntime so that everyone (not only those using `mir-linux-kernel`) can benefit.

In long term, I believe that we should split druntime into packages that implement compiler functionality (directly tied to the version of the compiler) and bindings for the system libraries, which can be used by any compiler version. The latter should be put on code.dlang.org.

RazvanN7:
@PetarKirov @Geod24 @maxhaton I rebased this. Should we get this on or should we make sure we are properly copy pasting from libmir?

PetarKirov:
@RazvanN7 Thank you for the reminder, this PR completely fell off my radar. Should be good to go modulo the above comments.

Perhaps if one want to go the extra mile, we could add a unittest that actually verifies that the `syscall` function works (e.g. that we haven't made an ABI error in the declarations.

RazvanN7:
The autotester failure is a spurious one and I am tempted to think that the stdcpp test is also a spurious one as it does not have any information about why it is failing.

ibuclaw:
WHY WAS THIS MERGED!!!!!!!

PR #2822, opened 2019-10-08
Revert "core.thread.refactor Part 1"

wilzbach:
Reverts dlang/druntime#2821


rainers:
I suspect the merge on Azure fails because the PR is from the dlang repo. Should we still force-merge this PR?

rainers:
Removed auto-merge, doesn't seem to be the only problem. https://github.com/dlang/druntime/pull/2825 runs the Azure tests with the revert, but still fails.

wilzbach:
@rainers the error happens when trying to clone DMD:

```
Cloning into '../dmd'...
fatal: Remote branch revert-2821-osthread not found in upstream origin
##[error]Cmd.exe exited with code '128'.
```

The problem is here: https://github.com/dlang/druntime/blob/master/azure-pipelines.yml#L43 :

```
  DMD_BRANCH:
    Parsing expression: <coalesce(variables['System.PullRequest.TargetBranch'], variables['Build.SourceBranchName'], 'master')>
    Evaluating: coalesce(variables['System.PullRequest.TargetBranch'], variables['Build.SourceBranchName'], 'master')
    Expanded: coalesce(Null, 'revert-2821-osthread', 'master')
    Result: 'revert-2821-osthread'
```


Maybe the environment variable `SYSTEM_PULLREQUEST_TARGETBRANCH` (see e.g. https://github.com/YakDriver/azure-pipelines-environment-variables#system_pullrequest_targetbranch) is defined?

wilzbach:
It's weird that `variables['System.PullRequest.TargetBranch']` is `Null` though:

> The branch that is the target of a pull request. For example: refs/heads/master. This variable is initialized only if the build ran because of a Git PR affected by a branch policy.

https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#system-variables

Maybe it's null, because it was re-triggered manually?

rainers:
I suspect it makes a difference whether the same repo is used for the PR and the target and then some variables behave differently, We should add printing the environment variables to the yml as in the dmd version.

The revert alone didn't help, though, see #2825 . So resolving this merge issue could be in another PR.

rainers:
> So resolving this merge issue could be in another PR.

It seems the branch isn't built as a PR, but a separate branch. I guess we should configure only master and stable as "triggers": https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#triggers

PR #2823, opened 2019-10-08
fix issue 13821 again

rainers:
This should fix the CI breakage, redoing #2801.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,6 +19,15 @@ jobs:
           ARCH: x64
           D_COMPILER: dmd
     steps:
+      - script: set
+        displayName: environment
+      - script: systeminfo
+        displayName: systeminfo
+      - task: CopyFiles@2
+        inputs:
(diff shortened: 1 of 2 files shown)


rainers:
I have currently no idea why this still fails (it works in a Windows Server VM locally). So maybe reverting with https://github.com/dlang/druntime/pull/2822 could fix the builds sooner.

rainers:
I finally figured the problem:
- druntime is checked out into a folder "s", not "druntime" as for example AppVeyor does.
- the build script clones dmd into "../dmd"
- dmd's installation/build script windows-visual-studio.sh/windows.sh is invoked
- as no folder "../druntime" exists druntime/master is checked out
- dmd's build and tests are run

So, the azure tests never actually tested the PR, but always master (or stable).

This should be fixed now by copying the source folder to "../druntime". There is still a wrong message `[GIT_CLONE] Switched druntime to branch master` in the log, but that has to be fixed in dmd's repo.

thewilsonator:
Auto-merge toggled on

PR #2825, opened 2019-10-09
Revert "core.thread.refactor Part 1 (#2821)"

rainers:
This reverts commit 1d39caad8b00b123709896f1df8494903ee478d2.

trying to see whether it helps and the azure CI works better with a PR not using the dlang repo as in https://github.com/dlang/druntime/pull/2822


rainers:
druntime back to green via #2823

PR #2826, opened 2019-10-11
Fix type of iovec.iov_len

tom-tan:
This request fixes the type of `iovec.iov_len` declared in `core/sys/posix/sys/uio.d` from `uint` to `size_t` for `Runtime_Musl` environment.

The header of `core/sys/posix/sys/uio.d` says:
```d
//
// Required
//
/*
struct iovec
{
    void*  iov_base;
    size_t iov_len;
}
```
and `iov_len` is declared as `size_t` in other environments.

### DIFF
--- a/src/core/sys/posix/sys/uio.d
+++ b/src/core/sys/posix/sys/uio.d
@@ -138,8 +138,8 @@ else version (CRuntime_Musl)
 {
     struct iovec
     {
-        void* iov_base;
-        uint  iov_len;
+        void*  iov_base;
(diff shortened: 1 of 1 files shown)


thewilsonator:
Auto-merge toggled on

tom-tan:
Thanks!

ZombineDev:
For future reference: https://github.com/ifduyue/musl/blob/b4b1e10364c8737a632be61582e05a8d3acf5690/include/alltypes.h.in#L68

PR #2827, opened 2019-10-11
-checkaction=context: Fix recursive struct formatting

dkorpel:
If you compile the following with `-checkaction=context`:
```D
struct S {
    S[] arr;
}

void main() {
    assert(S.init == S.init);
}
```
You get:
```
(...)/src/druntime/import/core/internal/dassert.d(115): Error: forward reference to inferred return type of function call miniFormat(v.arr)
```

Changing the `auto` return type of miniFormat to `string` should fix it. 
I can add the test-case for this, but should that be in dmd or druntime?

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -64,7 +64,7 @@ private template getPrintfFormat(T)
 Minimalistic formatting for use in _d_assert_fail to keep the compilation
 overhead small and avoid the use of Phobos.
 */
-private auto miniFormat(V)(V v)
+private string miniFormat(V)(V v)
 {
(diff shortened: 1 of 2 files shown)


wilzbach:
Thanks! You can add a test here: https://github.com/dlang/druntime/blob/master/test/exceptions/src/assert_fail.d

wilzbach:
Thanks!

Note that we don't need to do this in all cases. I added a commit to your PR to account for this.
See https://github.com/dlang/druntime/pull/2827/commits/2be02c3d2b0ebe0ac43a235aca447408481b34ec

PR #2828, opened 2019-10-11
Fix exceptions and dynamic casts across dll boundaries

John-Colvin:
By making the checks for whether something is a base class use an implementation like `TypeInfo_Class.opEquals` (Can't actually use `==` because that does a dynamic cast, which calls `_d_isbaseof2`, which means infinite recursion).

Relying on name comparison only seems weak, but ultimately the name *is* the id, just like names of symbols (on linux then the loader will implement ODR by eliminating duplicates based on names anyway, right?).

Effectively this is a reboot of https://github.com/dlang/druntime/pull/92

https://issues.dlang.org/show_bug.cgi?id=7020

### DIFF
--- a/src/rt/cast_.d
+++ b/src/rt/cast_.d
@@ -112,17 +112,17 @@ int _d_isbaseof2(ClassInfo oc, ClassInfo c, ref size_t offset)
 
 int _d_isbaseof(ClassInfo oc, ClassInfo c)
 {
-    if (oc is c)
+    if (oc.compareClassInfo(c))
         return true;
 
     do
     {
-        if (oc.base is c)
+        if (oc.base.compareClassInfo(c))
             return true;
 
         foreach (iface; oc.interfaces)
         {
-            if (iface.classinfo is c || _d_isbaseof(iface.classinfo, c))
+            if (iface.classinfo.compareClassInfo(c) || _d_isbaseof(iface.classinfo, c))
                 return true;
         }
 
--- a/src/rt/cast_.d
+++ b/src/rt/cast_.d
@@ -13,6 +13,15 @@
  */
 module rt.cast_;
 
+// because using == does a dynamic cast, but we
+// are trying to implement dynamic cast.
+bool compareClassInfo(ClassInfo a, ClassInfo b)
+{
+    if (a is b)
+        return true;
+    return a.info.name == b.info.name;
+}
+
 extern (C):
 
 /******************************************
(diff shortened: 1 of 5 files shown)


adamdruppe:
any idea why it is failing on the win with lld-link? t seems kinda arbitrary

wilzbach:
This could be related/help: https://github.com/dlang/dmd/pull/10525

adamdruppe:
@John-Colvin so the failure on win64 doesn't seem related to the other thing but have you confirmed this works on win32? or is that talk about the throw function and catch tables needing to be merged mean there's still a lot to do?

rmanthorpe:
Using `ClassInfo.name` isn't good enough for the same reason as https://issues.dlang.org/show_bug.cgi?id=20579

```d
class A(T) {}
template B(T) {
    struct C {}
}

void main() {
    assert(A!(B!int.C).classinfo.name != A!(B!string.C).classinfo.name); // assertion failure!
}
```

I think the only way to fix this properly is to put the mangle of a type into `ClassInfo` and compare that. If you're going to do that, I suggest using that to fix `TypeInfo.opEquals` while we're at it.

adamdruppe:
alas mangleof isn't available for subclasses... that means we need a dmd change to make the information available. and mangles can be absurdly large. idk.

rmanthorpe:
Or some other unique name, but it must be unique that the problem. Something like `std.traits.fullyQualifiedName` looks like it does the trick. There's a reason they get long - they have to be unique.

You can mitigate the cost of doing this by versioning it out on non-Windows platforms and with a little effort you can work out if you're in a different DLL and only do the string comparison in those cases.

adamdruppe:
fullyQualfiedName i think also doesn't work without virtuals

but i did just remember we have RTInfo so it doesn't need a dmd change per se; it can be done at CT with rtinfo in object.d. not a bad idea, might work at least as a hack.

rmanthorpe:
How does RTInfo help?

adamdruppe:
On Thu, Feb 13, 2020 at 07:17:23AM -0800, Richard Manthorpe wrote:
> How does RTInfo help?

it lets druntime extend typeinfo by passing every user-defined class to
a template for CTFE processing.

It can add a instance for the mangleof string.

rmanthorpe:
Ah that's the badger!

kinke:
TypeInfo names of classes, structs and interfaces are unique now after #3527.

src/rt/cast_.d
```diff
@@ -13,6 +13,15 @@
  */
 module rt.cast_;
 
+// because using == does a dynamic cast, but we
+// are trying to implement dynamic cast.
+bool compareClassInfo(ClassInfo a, ClassInfo b)
+{
+    if (a is b)
+        return true;
+    return (a && b) && a.info.name == b.info.name;
```
kinke, line 22: It might be sensible to restrict the name check to `version (Shared)`, i.e., a shared druntime (`.{so,dll}`). I think it's reasonable to expect a process with multiple D binaries to share a single druntime; that's not an option with DMD on Windows, I know.

kinke:
This has landed in #3543, thx John.

John-Colvin:
> This has landed in #3543, thx John.

Thanks for sorting this out, sorry for abandoning it for so long

PR #2829, opened 2019-10-13
core/atomic.d: Fix-up both public and internal API

ibuclaw:
This is so it becomes trivial to use this with GCC/LLVM intrinsics based on C++11 and C11.


TurkeyMan:
merge it now!

PR #2831, opened 2019-10-14
core/atomic.d: Add GDC atomic support code

ibuclaw:
Following on from #2596, rewrite to work with the new atomic API.

This has been peer reviewed by @TurkeyMan 

Best viewed with whitespace ignored, as the removals are only highlighted by indenting the DMD specific code into a version(DigitalMars) block.

### DIFF
--- a/src/core/internal/atomic.d
+++ b/src/core/internal/atomic.d
@@ -12,664 +12,1034 @@ module core.internal.atomic;
 
 import core.atomic : MemoryOrder;
 
-private
+version (DigitalMars)
 {
-    enum : int
+    private
     {
-        AX, BX, CX, DX, DI, SI, R8, R9
-    }
-
-    immutable string[4][8] registerNames = [
-        [ "AL", "AX", "EAX", "RAX" ],
-        [ "BL", "BX", "EBX", "RBX" ],
-        [ "CL", "CX", "ECX", "RCX" ],
-        [ "DL", "DX", "EDX", "RDX" ],
-        [ "DIL", "DI", "EDI", "RDI" ],
-        [ "SIL", "SI", "ESI", "RSI" ],
-        [ "R8B", "R8W", "R8D", "R8" ],
-        [ "R9B", "R9W", "R9D", "R9" ],
-    ];
-
-    template RegIndex(T)
-    {
-        static if (T.sizeof == 1)
-            enum RegIndex = 0;
-        else static if (T.sizeof == 2)
-            enum RegIndex = 1;
-        else static if (T.sizeof == 4)
-            enum RegIndex = 2;
-        else static if (T.sizeof == 8)
-            enum RegIndex = 3;
-        else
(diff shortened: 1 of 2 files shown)


src/core/internal/atomic.d
```diff
@@ -12,664 +12,1075 @@ module core.internal.atomic;
...
+    }
 
-            Lend:
+    private bool atomicCompareExchangeImpl(MemoryOrder succ = MemoryOrder.seq, MemoryOrder fail = MemoryOrder.seq, bool weak, T)(T* dest, T* compare, T value) pure nothrow @nogc @trusted
+        if (CanCAS!T)
+    {
+        bool res = void;
 
-                ret;
+        static if (GNU_Have_Atomics || GNU_Have_LibAtomic)
```
TurkeyMan: 3 out if 5 lines in this section is wasted vertical space :/
Most of the lines here are `{`, `}` and the line split, which is split well before the horiz white space even github shows here in the review render ;)

The 3 if/else blocks up a little waste 50% if lines to `{`, `}`, but this blocks triggers me ;)
Where I'm from, monitors have been 16/9 for almost at least 15 years... so I feel like vertical space is about 16/9th's more valuable ;)

TurkeyMan:
What's sad is that no unit tests will cover this GDC code... is that fine?

src/core/internal/atomic.d
wilzbach: https://dlang.org/dstyle

ibuclaw: I probably have a subconscious habit of keeping to 80 chars as well.

ibuclaw:
> What's sad is that no unit tests will cover this GDC code... is that fine?

Erm, all the unittests in core.atomic.

src/core/internal/atomic.d
ibuclaw: Any though it's just an opinion, I don't personally find compressing the format very readable.

image

ibuclaw: Removed some braces just to make @TurkeyMan happy.

TurkeyMan:
Yeah but the DMD CI won't prove this code path, it could regress against GDC?

src/core/internal/atomic.d
TurkeyMan: left the awkward line breaks though :P

ibuclaw: It's there for readability. :-P

ibuclaw:
> Yeah but the DMD CI won't prove this code path, it could regress against GDC?

If it isn't syntactically valid, an error would be thrown.  Any API changes need to be synchronized, however I would expect the person making the change would do the required due dilligence, otherwise there's no basis for accepting them.

kinke:
Iain, what's your ultimate goal wrt. these upstreamings? Do you aim for not having to maintain druntime/Phobos forks or do you plan to restrict the upstreamings to manageably-sized diffs?

PR #2834, opened 2019-10-17
Refactoring and resolving inconsistency in platfrom dependent execinfo import

0l-l0:
The initial goal was to resolve link error on systems with musl-libc. This issue was partially solved by #2796 but it didn't support the use of an external execinfo implementation.

This very PR is meant to support all platforms (libc versions) with execinfo (backtrace) functionality included and those with an external execinfo library. In the external lib case, we need to set the corresponding version ID depending on which `backtrace_symbols` output format is used.

Modifications:
* many boilerplate code removed
* clarifying the usage of `CRuntime_Glibc`, `linux` and `Posix` version IDs in execinfo related code
* thorough execinfo availability introspection
* compile time helper code placed into a separated module (`core.internal.execinfo`)

### DIFF
--- a/changelog/platform_dependent_execinfo.dd
+++ b/changelog/platform_dependent_execinfo.dd
@@ -1,0 +1,36 @@
+Platform dependent execinfo introspection added
+
+A new module (core.internal.execinfo) has been added for platform dependent
+execinfo detection. On every POSIX system which provides an execinfo
+implementation as part of its C runtime the appropriate implementation-
+dependent execinfo module will be imported automatically.
+
+Besides that, there is an opportunity for using external execinfo
+implementations built as separated libraries and linking DRuntime with them.
+
+$(B IMPORTANT:) On platforms with C runtime not providing execinfo
+functionality one should decide whether an external lib (e.g. libexecinfo)
+will be used, or not. If not, DRuntime should be built normally without any
+additional version ID, but with external lib exactly one of the following
+version IDs should be chosen at compile time. That means, the selected
+external format cannot be changed later without rebuilding DRuntime.
+
+It could be really important to keep this in mind when someone is packaging
+DRuntime for such OSs (e.g. ones using musl libc) and provide packages with
+and without execinfo support (with the specific external library as
+dependency) or in the case of source based packages, make this build option
+selectable (e.g. portage).
+
+$(TABLE
+$(THEAD Version ID, Backtrace format)
+$(TROW $(B ExtExecinfo_BSDFmt), 0x00000000 <_D6module4funcAFZv+0x78> at module)
+$(TROW $(B ExtExecinfo_DarwinFmt),
+1  module    0x00000000 D6module4funcAFZv + 0)
+$(TROW $(B ExtExecinfo_GNUFmt), module(_D6module4funcAFZv) [0x00000000] $(B or)
+module(_D6module4funcAFZv+0x78) [0x00000000]
+$(B or) module(_D6module4funcAFZv-0x78) [0x00000000])
+$(TROW $(B ExtExecinfo_SolarisFmt), object'symbol+offset [pc])
+)
+
+These formats above cover most of the "classic" backtrace outputs but as a new
+important format emerges, it can be easily added.
--- a/src/core/internal/execinfo.d
+++ b/src/core/internal/execinfo.d
@@ -1,0 +1,301 @@
+/**
+ * This module helps to decide whether an appropriate execinfo implementation
+ * is available in the underling C runtime or in an external library. In the
+ * latter case exactly one of the following version identifiers should be
+ * set at the time of building druntime.
+ *
+ * Possible external execinfo version IDs based on possible backtrace output
+ * formats:
+ * $(TABLE
+ * $(THEAD Version ID, Backtrace format)
+ * $(TROW $(B ExtExecinfo_BSDFmt), 0x00000000 <_D6module4funcAFZv+0x78> at module)
+ * $(TROW $(B ExtExecinfo_DarwinFmt), 1  module    0x00000000 D6module4funcAFZv + 0)
+ * $(TROW $(B ExtExecinfo_GNUFmt), module(_D6module4funcAFZv) [0x00000000] $(B or)
+ * module(_D6module4funcAFZv+0x78) [0x00000000] $(B or) module(_D6module4funcAFZv-0x78) [0x00000000])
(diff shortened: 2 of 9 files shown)


kinke:
New files need to be added to the files in the `mak` subdir - archaic, I know.

src/rt/backtrace/dwarf.d
```diff
@@ -12,6 +12,11 @@
 
 module rt.backtrace.dwarf;
 
+private import rt.backtrace.execinfo;
+
+static if (hasExecinfo)
+{
```
kinke: Could be `static if (hasExecinfo):`, then you don't need the closing parenthesis at the end of the file (and make it clear that the module is otherwise empty).

kinke:
`mak/COPY` is still missing, and lines 108/115/124 feature trailing spaces.

0l-l0:
As far as I can see, `mak/COPY` corresponds to the public interface, so none of the private modules (in `rt` and `gc` subdirs) are listed in that file, that's why I left it out. Should it be included anyway?

kinke:
Oh right, my bad; you can't import the internal module in public `core.runtime` then though.

wilzbach:
One can with `externDFunc`.

0l-l0:
> One can with `externDFunc`.

Is there any way to solve this issue for template mixins and compile time functions/values (I mean for `ImportExecinfoPOSIX` and `hasExecinfo`)? Or should I make it work in an another way?

kinke:
Nope, linker tricks are limited.

0l-l0:
The last commit covers a possible (not so elegant) solution.

kinke:
I'd probably just move your new `rt.backtrace.execinfo` to `core.internal.execinfo`. Public `core.runtime` needs parts of it, so I don't see a problem with a new public but internal module.

0l-l0:
I moved it and now it looks much better.

src/core/internal/execinfo.d
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * Imports the appropriate execinfo module depending on the actual
+ * OS and libc version. If the libc implementation does not include
+ * any execinfo methods, it is possible to link with an external
+ * execinfo library. In the latter case one of the $(B ExtExecinfo_)
+ * version identifiers should be set and the corresponding external
+ * library should be linked with druntime.
+ */
+mixin template ImportExecinfoPOSIX()
```
Geod24: Why do you need a `mixin template` ? The three aliases should be enough.

src/core/internal/execinfo.d
```diff
@@ -0,0 +1,312 @@
+/**
+ * This module helps to decide whether an appropriate execinfo implementation
+ * is available in the underling C runtime or in an external library. In the
+ * latter case exactly one of the following version identifiers should be
+ * set at the time of building druntime.
+ *
+ * Possible external execinfo version IDs based on possible backtrace output
```
Geod24: This should be in a release note, it's relevant to packagers (e.g. for Musl)

src/core/internal/execinfo.d
0l-l0: You're right. It's unnecessary.

src/core/internal/execinfo.d
0l-l0: Done.

Geod24:
@thewilsonator : Can you remove auto-merge please ? Just want to discuss a couple details before we merge.

changelog/platform_dependent_execinfo.dd
```diff
@@ -0,0 +1,25 @@
+Platform dependent execinfo introspection added
+
+A new module (core.internal.execinfo) has been added for platform dependent
+execinfo detection. On every POSIX system which provides an execinfo
+implementation as part of its C runtime the appropriate implementation-
+dependent execinfo module will be imported automatically.
+
+Besides that, there is an opportunity for using external execinfo
+implementations built as separated libraries and linking druntime with them.
+In that case exactly one of the following version ID must be set for druntime
```
Geod24: Could you clarify the wording to make it obvious that this version ID needs to be set when Druntime is compiled ?
That means, if one package it for Musl, one has to choose whether it depends on libexecinfo or not, and the end user cannot change this.

src/core/internal/execinfo.d
```diff
@@ -0,0 +1,303 @@
...
+}
+
+private bool atMostOneBTFmt()
+{
+    size_t trueCnt = 0;
+
+    foreach (fmt; __traits(allMembers, BacktraceFmt))
+        if (__traits(getMember, BacktraceFmt, fmt)) ++trueCnt;
+
+    return 2 > trueCnt;
```
Geod24:
```suggestion
    return trueCnt <= 1;
```

src/core/internal/execinfo.d
```diff
@@ -0,0 +1,303 @@
...
+else version (FreeBSD)
+    import _execinfo = core.sys.freebsd.execinfo;
+else version (NetBSD)
+    import _execinfo = core.sys.netbsd.execinfo;
+else version (DragonFlyBSD)
+    import _execinfo = core.sys.dragonflybsd.execinfo;
+else version (Solaris)
+    import _execinfo = core.sys.solaris.execinfo;
+
+static if (__traits(compiles, _execinfo) && is(_execinfo == module))
```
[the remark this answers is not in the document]
Geod24: Nope, since druntime has hooks that can be added / removed from a PR to another, X-version compilation is not a concern.

src/core/internal/execinfo.d
```diff
@@ -0,0 +1,303 @@
...
+    /// supported output format can be obtained by testing
+    /// $(LREF BacktraceFmt) enum values.
+    alias backtrace_symbols = _execinfo.backtrace_symbols;
+
+    /// Always points to the platform's backtrace_symbols_fd function.
+    alias backtrace_symbols_fd = _execinfo.backtrace_symbols_fd;
+}
+
+/// Indicates the availability of backtrace functions
+enum bool hasExecinfo = __traits(compiles, backtrace)
```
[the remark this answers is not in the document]
0l-l0: Yes, it looks nicer. Thanks!

src/core/internal/execinfo.d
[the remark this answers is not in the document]
0l-l0: I don't see any reason to require a minimum version so I removed the redundancy.

PR #2835, opened 2019-10-18
[stable] core.atomic: Make type repainting in atomicLoad cheaper

kinke:
Pinging @TurkeyMan.

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -96,10 +96,7 @@ TailShared!T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)(ref shared const T 
     // HACK: DEPRECATE THIS FUNCTION, IT IS INVALID TO DO ATOMIC LOAD OF SHARED CLASS
     // this is here because code exists in the wild that does this...
 
-    import core.lifetime : move;
-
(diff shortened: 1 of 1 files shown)


TurkeyMan:
How does this make it cheaper? Why isn't this identical?
I'm not on board with this change... I specifically designed against instantiations of the implementation functions with types like that.

TurkeyMan:
On I get it, because `move` is rubbish...
We need to fix `move`... it's my #1 slot for worst problem in D right now.

kinke:
> On I get it, because move is rubbish...

Exactly, and I don't see how `T` is better suited for the internal implementation than the `TailShared!T`; it's not like you'd use an integer type of matching size here or something like that.

The actual reason is that LDC's druntime testrunner wouldn't link due to undefined `move` template instantiations without this; this rewrite hides that problem at least. ;)

TurkeyMan:
Well, `TailShared` should be eradicated somehow, and I'm trying to isolate that as much as possible for the time being.
I guess I can tolerate this patch though, since it's contained in a function that shouldn't exist anyway (see the comment immediately above). It still upsets me :/

PetarKirov:
@wilzbach is @dlang-bot taking a day off? :D

image

PR #2838, opened 2019-10-24
Fix issue 11725: AA.dup should return mutable AA where possible.

quickfur:
Since the original AA has been duplicated, it should be safe to mutate the copy. No need to needlessly restrict the copy to be const too.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2223,7 +2223,7 @@ T rehash(T : shared Value[Key], Value, Key)(T* aa)
  * Params:
  *      aa =     The associative array.
  */
-V[K] dup(T : V[K], K, V)(T aa)
+auto dup(T : V[K], K, V)(T aa)
 {
     //pragma(msg, "K = ", K, ", V = ", V);
 
@@ -2231,7 +2231,14 @@ V[K] dup(T : V[K], K, V)(T aa)
     static assert(is(typeof({ V v = aa[K.init]; })),
         "cannot call " ~ T.stringof ~ ".dup because " ~ V.stringof ~ " is not copyable");
 
-    V[K] result;
+    // bug 11725: return mutable AA if possible
+    import core.internal.traits : Unconst;
+    static if (is(V : Unconst!V) && is(K : Unconst!K))
+        alias Result = Unconst!V[Unconst!K];
+    else // mutable dup not possible, return const object instead.
+        alias Result = V[K];
+
+    Result result;
 
     //foreach (k, ref v; aa)
     //    result[k] = v;  // Bug13701 - won't work if V is not mutable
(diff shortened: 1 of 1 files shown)


quickfur:
Hmm. Running into trouble with `immutable int[int]` here: returning a mutable copy makes it not implicitly convertible back to the original type.  I looked over the implementation of array.dup and it seems there's some kind of hack that triggers DIP25, but I'm not sure I understand exactly what's being done and how/whether to port it to the AA case?

src/object.d
```diff
@@ -2223,15 +2223,22 @@ T rehash(T : shared Value[Key], Value, Key)(T* aa)
...
 
     // Bug10720 - check whether V is copyable
     static assert(is(typeof({ V v = aa[K.init]; })),
         "cannot call " ~ T.stringof ~ ".dup because " ~ V.stringof ~ " is not copyable");
 
-    V[K] result;
+    // bug 11725: return mutable AA if possible
+    import core.internal.traits : Unconst;
+    static if (is(V : Unconst!V) && is(K : Unconst!K))
+        alias Result = Unconst!V[Unconst!K];
```
MoonlightSentinel: This should probably be done for `K` and `V` independently.

Furthermore this should handle indirections by only stripping head const, e.g. `const int*` should be replaced with `const(int)*`. This also raises the question on how to handle structs with indirections:

```
struct S
{
    int* ptr;
}

void foo(const S[int] aa)
{
    auto copy = aa.dup();
    *(copy[0].ptr) = 2; // Can write const data
}

Comment (schveiguy) [reply src/object.d]: Technically, keys should always be const. In fact, D makes sure of that for indirections:

```D
int[char[]] aa;
pragma(msg, typeof(aa)); // int[const(char)[]]
```

And the `is(V : Unconst!V)` should take care of the structs with indirection cases.

schveiguy:
> there's some kind of hack that triggers DIP25

Isn't it the purity? If you accept an immutable item, and return a mutable item in a pure function, it's assumed unique.

Note that you have to accept a const parameter for this to work. So in the case of an AA with mutable values, you would have to accept it as a const parameter explicitly. This is why there are 2 overloads of T[].dup.

src/object.d
MoonlightSentinel: You're right, Unconst was more elegant than I remembered.

schveiguy: Well, `Unconst` could be dumb or poorly implemented :) It's the `is` expression that makes the difference.

quickfur:
Does anyone want to take this PR up?  I've been busy and haven't been able to work on this.  Would be nice if somebody took over who better understand how this all works. :-D

schveiguy:
Sorry, also busy. Should be pretty quick though, I may do it next week if I find some time.

Geod24:

PR #2842, opened 2019-10-29
Fix __WORDSIZE definition for 32 bits Musl

Geod24:
See also: #2843


thewilsonator:
Auto-merge toggled on

PR #2846, opened 2019-11-01
Fix Issue 20346 - std.uuid does not compile with checkaction=context

MoonlightSentinel:
Marking all parameters as `const` additionally reduces the amount of instantiations of `miniFormat`

### DIFF
--- a/test/exceptions/src/assert_fail.d
+++ b/test/exceptions/src/assert_fail.d
@@ -93,6 +93,22 @@ void testToString()()
         }
     }
     test(new Foo("a"), new Foo("b"), "Foo(a) != Foo(b)");
+
+    // Verifiy that the const toString is selected if present
+    static struct Overloaded
+    {
+        string toString()
+        {
+            return "Mutable";
+        }
+
+        string toString() const
+        {
+            return "Const";
+        }
+    }
+
+    test!"!="(Overloaded(), Overloaded(), "Const is Const");
 }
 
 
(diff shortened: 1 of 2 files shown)


test/exceptions/src/assert_fail.d
```diff
@@ -158,6 +158,20 @@ void testTemporary()
...
+    struct UUID {
+        union
+        {
+            ubyte[] data;
+        }
+    }
+
+    ubyte[] data;
+    enum ctfe = UUID();
+    assert(ctfe.data == data);
```
wilzbach: You want to test the output string. I noticed that your earlier addition doesn't check the output string either.  This can be done with the `test` function. See the other tests.

MoonlightSentinel: I didn't test the output string because this is primarily a compilable test like https://github.com/dlang/druntime/blob/e7df7706c224ff6c3d684a52ce5528a500449c7a/test/exceptions/src/assert_fail.d#L131-L135

But i will add the output verfication

MoonlightSentinel: Using `test` would not work as the enum member would be copied into the parameter of `test`

PetarKirov: We would need to add another function that unlike `test` accepts a whole block of code.

PetarKirov: Or just replicate the code inside the `test` function in this one case.

MoonlightSentinel: Added

test/exceptions/src/assert_fail.d
```diff
@@ -158,6 +163,20 @@ void testTemporary()
     assert(Bad() == Bad());
```
wilzbach: Thanks! Would you mind adding a similar output test for `Bad` as well.

MoonlightSentinel: Can do, but its another compilable test (as `struct`formatting is already covered above)

MoonlightSentinel:
@wilzbach Dont merge yet, i will add another test concerning the correct `toString` selection

MoonlightSentinel:
@wilzbach Done

PR #2849, opened 2019-11-06
Correctly use `in` in FreeBSD static asserts instead of `const`

atilaneves:
While currently `in` and `const` mean the same thing, they might not
in the future.

This is blocking [this pull request](https://github.com/dlang/dmd/pull/10506).


thewilsonator:
Auto-merge toggled on

PR #2853, opened 2019-11-29
Fix issue 20354: interface is not supported by cas

lempiji:
I want to fix this issue.
Some packages are breaking now. :cry:

https://issues.dlang.org/show_bug.cgi?id=20354

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -239,7 +239,7 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 
 /// Ditto
 TailShared!T atomicExchange(MemoryOrder ms = MemoryOrder.seq,T,V)(shared(T)* here, V exchangeWith) pure nothrow @nogc @trusted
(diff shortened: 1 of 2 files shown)


thewilsonator:
```
src\core\atomic.d(246): Error: undefined identifier `U`
src\core\atomic.d(830): Error: template instance `core.atomic.atomicExchange!(cast(MemoryOrder)5, Inter, shared(Inter))` error instantiating
src\core\atomic.d(910):        instantiated from here: `testXCHG!(shared(Inter))`
```

I think https://github.com/dlang/druntime/blob/master/src/core/atomic.d#L245-L246 is a typo.

CC @TurkeyMan git blames you for this

Should that be `alias Thunk = Unshared!V;`?

thewilsonator:
Also as this is a regression, please target stable.

PetarKirov:
I rebased the PR to stable and fixed the remaining errors.
@TurkeyMan can you take another look?

thewilsonator:
LGTM

PR #2858, opened 2019-12-07
fix issue 20438 - [Reg 2.086] GC: memory not reusable when calling GC…

rainers:
….collect after GC.free

fix detection of page with unused allocation slots

What this bug means: if no object was collected in a page, and a quarter of the page had no free allocation slots, the whole page was considered to be used, so the free slots were not restored to the free lists.


rainers:
buildkite failure is an unrelated connection error:
```
object.Exception@source/vibe/core/net.d(229): Failed to connect to [2606:2800:220:1:248:1893:25c8:1946]:443: refused
```

PR #2859, opened 2019-12-07
rt.backtrace.elf: Support D_X32 targets

ibuclaw:


src/rt/backtrace/elf.d
```diff
@@ -309,9 +309,18 @@ version (X86)
...
+        alias Elf_Ehdr = Elf32_Ehdr;
+        alias Elf_Shdr = Elf32_Shdr;
+        enum ELFCLASS = ELFCLASS32;
+    }
+    else
+    {
+        alias Elf_Ehdr = Elf64_Ehdr;
+        alias Elf_Shdr = Elf64_Shdr;
+        enum ELFCLASS = ELFCLASS64;
+    }
```
Geod24:
```D
version (X86) { ... }
else version (D_X32) { ... }
else version (X86_64) { ... }
```

Would be more in line with the style used here.

ibuclaw: X32 is a subset rather than a separate architecture in its own right.

PetarKirov: It seems like https://dlang.org/spec/version agrees:

> **D_X32**	Pointers are 32 bits, but words are still 64 bits (x32 ABI) (This can be defined in parallel to X86_64)

PR #2860, opened 2019-12-08
fix Issue 20364 - [REG2.069] changing length for typeof(null)[] array…

WalterBright:
… seg faults in _d_arraysetlengthiT()

A null pointer was being returned from `Typeinfo_n.initializer()`


rainers:
The initializer is explicitly allowed to return a size without an actual pointer to avoid large zero-arrays: https://github.com/dlang/druntime/blob/7d89f3999f2ed4eff80065d37b6b77c87342ba9b/src/object.d#L366-L372
The actual bug is either in `_d_arraysetlengthiT` or in the compiler by not calling `_d_arraysetlengthT`.

PR #2862, opened 2019-12-08
Fix issue 19489 - Null function call results in no stack trace

ahmetsait:


ahmetsait:
```D
if ( stackframe.AddrPC.Offset == stackframe.AddrReturn.Offset )
{
    debug(PRINTF) printf("Endless callstack\n");
    break;
}
```
I could not figure out what the actual reason for this code in traceNoSync() but removing it does not break anything in my tests (and apparently in CI as well) so someone needs to enlighten me.

The issue here is that the statement checks against `stackframe.AddrReturn.Offset` **before** even initializing it so when a null function call happens it will immediatelly break because `stackframe.AddrReturn.Offset` is also `0` in the first iteration of the do..while loop.

Not only that, but this code also fails to generate a stack trace for recursive functions as well:
```D
import std.stdio : writeln;
void main(string[] args)
{
    try r(0);
    catch(Throwable ex) writeln('\n', ex, '\n');
}

void r(int i)
{
    if (i < 5)
        r(i + 1);
    else
    {
        int* p = cast(int*)0;
        try *p = 5;
    }
}
```
Note that these require an access violation to be reproduced, normal exceptions will work as usual.

FYI, 64bit programs will continue to just crash silently, which has to do with druntime (dmd?) being unable to handle it currently.

rainers:
> I could not figure out what the actual reason for this code in traceNoSync()

Indeed, it looks questionable. Maybe a check for a decreasing or identical stack address could avoid an infinite loop (if StackWalk64 doesn't already check this).

Can you add a testcase? See druntime\test\exceptions for posix versions of similar tests.

> FYI, 64bit programs will continue to just crash silently, which has to do with druntime (dmd?) being unable to handle it currently.

Unfortunately, dmd implements its own exception handling derived from the posix implementation, so it is unable to show a callstack for access violations.

ahmetsait:
It's a bit disappointing that only commit messages are picked up by the dlang-bot (and also dlang/dlang-bot#234).

CyberShadow:
> It's a bit disappointing that only commit messages are picked up by the dlang-bot

No, commit messages should be at least as descriptive as the pull request description.

thewilsonator:
> I could not figure out what the actual reason for this code in traceNoSync()

Stack overflow?

ahmetsait:
> Stack overflow?

Well the current situation is kind of pathetic, runtime crashes with another stack overflow while trying to create the stack trace for it.

@rainers can you give it hand with test cases? Make files seem a bit cryptic for my taste.

rainers:
> @rainers can you give it hand with test cases? Make files seem a bit cryptic for my taste.

I have added a test case for Win32 stack traces.

ahmetsait:
Thanks for the test, it's failing because _Dmain is not demangled to "D main" for some reason, should I change the test or check why is it not demangling..?

rainers:
> it's failing because _Dmain is not demangled to "D main" for some reason, should I change the test or check why is it not demangling..?

Hmm, the OMF version is failing, not sure why. I have just added support for the mangled symbol, too.

ahmetsait:
@CyberShadow I consider this good to merge once the autotester passes.

PR #2864, opened 2019-12-10
core.math: Add templated toPrec.

ibuclaw:
Some floating point algorithms, such as Kahan-Babuska-Neumaier Summation,
require rounding to specific precisions. Rounding to precision after
every operation, however, loses overall precision in the general case
and is a runtime performance problem.

Adding these functions guarantee the rounding at required points in the
code, and document where in the algorithm the requirement exists.

Rebase of #1941.

### DIFF
--- a/src/core/math.d
+++ b/src/core/math.d
@@ -164,3 +164,74 @@ unittest
     }
 }
 
+/*************************************
+ * Round argument to a specific precision.
+ *
+ * D language types specify a minimum precision, not a maximum. The
+ * `toPrec()` function forces rounding of the argument `f` to the
+ * precision of the specified floating point type `T`.
+ *
+ * Params:
+ *      T = precision type to round to
+ *      f = value to convert
+ * Returns:
+ *      f in precision of type `T`
+ */
+@safe pure nothrow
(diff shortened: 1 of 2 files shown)


src/core/math.d
```diff
@@ -164,3 +164,76 @@ unittest
...
+ * precision of the specified floating point type `T`.
+ *
+ * Params:
+ *      T = precision type to round to
+ *      f = value to convert
+ * Returns:
+ *      f in precision of type `T`
+ */
+@safe pure nothrow
+T toPrec(T:float)(float f) { pragma(inline, false); return f; }
```
thewilsonator: Why ` pragma(inline, false);`?

ibuclaw: I would assume just in case the inliner removes the call before builtins handler is given a chance to expand it.

ibuclaw:
Disabling problematic tests until compiler support has been added.

src/core/math.d
thewilsonator: I'm worried about performance, but I suppose if this is a CTFE only intrinsic then there is no problem.

PR #2866, opened 2019-12-11
core.stdcpp.vector GCC and Clang versions

SSoulaimane:
This is the GCC an Clang ports of STL's `std::vector`.


SSoulaimane:
Tests need https://github.com/dlang/dmd/pull/10593 otherwise it's good to go.

PR #2867, opened 2019-12-11
Fix Issue 20303 - Memory leak in core.thread

jpiles:
In `thread_entryPoint` we are allocating space using `rt.tlsgc.init`, but it is never given back with `rt.tlsgc.destroy`.

It *could* happen in the destructor for the class, but it checks first that the thread is registered. In this case, by the time the destructor is (eventually) reached, the `scope (exit)` block has already deregistered the thread, so it is basically a NOOP.

I'm adding the relevant cleanup code from the destructor into that `scope (exit)` block, but I'm not 100% sure it's the best place for it. Anyway, there should be no risk of double free, since as I said, the thread is already deregistered at that point.

### DIFF
--- a/src/core/thread/osthread.d
+++ b/src/core/thread/osthread.d
@@ -687,6 +687,12 @@ class Thread
      */
     ~this() nothrow @nogc
     {
+        if (m_tlsgcdata !is null)
+        {
+            rt_tlsgc_destroy( m_tlsgcdata );
+            m_tlsgcdata = null;
+        }
+
         bool no_context = m_addr == m_addr.init;
         bool not_registered = !next && !prev && (sm_tbeg !is this);
 
@@ -710,8 +716,6 @@ class Thread
         {
             m_tmach = m_tmach.init;
         }
-        rt_tlsgc_destroy( m_tlsgcdata );
-        m_tlsgcdata = null;
     }
 
 
--- a/src/core/thread/osthread.d
+++ b/src/core/thread/osthread.d
@@ -413,6 +413,8 @@ else version (Posix)
             {
                 Thread.remove(obj);
                 atomicStore!(MemoryOrder.raw)(obj.m_isRunning, false);
+                rt_tlsgc_destroy( obj.m_tlsgcdata );
+                obj.m_tlsgcdata = null;
             }
             Thread.add(&obj.m_main);
 
--- a/src/core/thread/osthread.d
+++ b/src/core/thread/osthread.d
@@ -302,6 +302,8 @@ version (Windows)
             scope (exit)
             {
                 Thread.remove(obj);
+                rt_tlsgc_destroy( obj.m_tlsgcdata );
+                obj.m_tlsgcdata = null;
             }
             Thread.add(&obj.m_main);
 


src/core/thread/osthread.d
```diff
@@ -413,6 +413,8 @@ else version (Posix)
             {
                 Thread.remove(obj);
                 atomicStore!(MemoryOrder.raw)(obj.m_isRunning, false);
+                rt_tlsgc_destroy( obj.m_tlsgcdata );
+                obj.m_tlsgcdata = null;
```
rainers: Isn't this an issue with the Windows version, too?

What about the calls to pthread_detach/CloseHandle in the dtor? These seem to not be called aswell if the thread is not explicitly joined. Maybe the check in the dtor is too restrictive and it should just free resources according to the state of m_addr and m_tlsgcdata.

jpiles: Sorry about the windows version, I didn't think about it... I don't even have a Windows machine nowadays, so it went totally past me. Thanks for reminding me! I'll just add it to the windows version of the function.

However, now that I look deeper into it, it's a bit of a rabbit hole, and I think the whole design might need to be reconsidered:

* When you "attach" (in the D sense, which also makes it confusing, because it's unrelated to libpthread) a thread, some resources are reserved (see `attachThread`):

```d
    thisThread.m_isDaemon = true;
    thisThread.m_tlsgcdata = rt_tlsgc_init();
    Thread.setThis( thisThread );
```
* If you later "detach" the thread, the resources are **not** freed, but the thread is "unregistered" (see `thread_detachThis`, `thread_detachByAddr`, `thread_detachInstance`). This means, the resources will leak because the destructor will be a noop.

So for me the question would be who is responsible for cleaning up. One option would be to do what you suggested and just let the destructor do it, but it concerns me that there is no warranty about when, or even if, class destructors run, and we might be using up resources for much longer than needed. I wouldn't mind keeping that as a "safety net" just in case, but I guess I'd prefer a more managed approach.

My personal opinion is that whoever grabs the resources should free them, so I'd keep the changes and then add the cleanup code to the detaching functions, mirroring the attaching. What would you think of this?

Alternatively, I'd suggest moving this resource management to `Thread.add( Thread t, bool rmAboutToStart = true )` / `Thread.remove( Thread t )`. As far as I've seen, they are linked (i.e. they are called at the same places), so it should work.

The problem with this is that I don't know if there's any code path / use case for these functions to be called directly, without needing or wanting the resource management. So far I haven't seen it, and they are not public either, but there might be a reason for keeping them apart that I don't know of.

rainers: I think the resources can be released as eagerly as *safely* possible, i.e. `m_tlsgcdata` can be freed after `remove` before the thread finishes as proposed in this PR. It should keep track of released resources by setting the appropriate member (m_addr or m_tlsgcdata) to null, so that the dtor can unconditionally check whether these resources still need to be released (i.e. I'd remove the `no_context || not_registered` check).

Detaching a thread currently relies on the garbage collector to eventually call the dtor (as it does for regular D threads that are not joined). This seems to be ok when `thread_attachThis/detachThis` is called only once per thread, but can waste a lot of resources if called repeatedly, e.g. when "securing" a callback from some other language with a pair of attach and detach. That doesn't seem to work anyway right now, though, as the `sm_this` value is not reset so the second `attach` won't register the thread again. So I'd leave this for another PR.

jpiles:
I just updated the patch as per the previous comments, in the cleanup I only checked the tls data because for the other resources the check seems appropriate.

rainers:
> I just updated the patch as per the previous comments, in the cleanup I only checked the tls data because for the other resources the check seems appropriate.

Ok, let's leave that for another PR. Thanks.

PR #2868, opened 2019-12-11
core.internal.convert: Use toPrec to force double precision when parsing IBM 128bit floats

ibuclaw:


thewilsonator:
Does this fix https://issues.dlang.org/show_bug.cgi?id=9937 ?

ibuclaw:
> Does this fix https://issues.dlang.org/show_bug.cgi?id=9937 ?

It does not. It only finishes off the implementation of #2863 so that it genuinely works on PPC.

ibuclaw:
@thewilsonator - OK for merge?

PR #2873, opened 2019-12-16
Fix issue 17563 - Expose gc_inFinalizer

PetarKirov:
See the changelog entry and source code comments for details.

### DIFF
--- a/changelog/gc_in_finalizer.dd
+++ b/changelog/gc_in_finalizer.dd
@@ -1,0 +1,18 @@
+$(REF inFinalizer, core, GC) was added
+
+A new function exposing information about the state of the GC was added.
+It returns `true` if the current thread is executing destructors (finalizers)
+of objects allocated on the GC heap that are no longer being referenced.
+
+Destructors of objects allocated on the GC heap presently have several
+limitations or oddities:
+* As destructors of all unreachable objects are ran in indeterministic order,
+  there's no guarantee that object members are alive during the object's
+  finalization
+* Allocation during finalization is disallowed
+* In contrast to stack allocated objects, partially constructed objects (object
+  whose constructor threw an exception) may be finalized
+
+$(REF inFinalizer, core, GC) can be used to efficiently guard against
+programming errors such as the above, or to detect whether certain
+objects were left to be destroyed by the GC.
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -148,7 +148,7 @@ private
     extern (C) void gc_removeRange(const void* p ) nothrow @nogc;
     extern (C) void gc_runFinalizers( const scope void[] segment );
 
-    package extern (C) bool gc_inFinalizer();
+    package extern (C) bool gc_inFinalizer() nothrow @nogc @safe;
 }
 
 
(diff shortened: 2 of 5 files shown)


src/core/memory.d
```diff
@@ -847,6 +847,104 @@ struct GC
...
+            auto r = new Resource;
+            r = null;
+        }
+
+        assert(Resource.outcome == Outcome.notExecuted);
+        createGarbage();
+        GC.collect;
+        assert(
+            Resource.outcome == Outcome.notExecuted ||
+            Resource.outcome == Outcome.executedByTheGC);
```
schveiguy: This sucks, but I don't know how to guarantee something becomes garbage.

changelog/gc_in_finalizer.dd
```diff
@@ -0,0 +1,18 @@
...
+* As destructors of all unreachable objects are ran in indeterministic order,
+  there's no guarantee that object members are alive during the object's
+  finalization
+* Allocation during finalization is disallowed
+* In contrast to stack allocated objects, partially constructed objects (object
+  whose constructor threw an exception) may be finalized
+
+$(REF inFinalizer, core, GC) can be used to efficiently guard against
+programming errors such as the above, or to empirically verify that certain
+objects are not left to be destroyed by the GC.
```
schveiguy: On that last point, it's a bit confusing. You can't verify that they aren't *left* to be destroyed by the GC (which happens when you stop referencing an object, and isn't detectable), you can only verify that they aren't destroyed by the GC.

Perhaps "empirically verify if an object is being destroyed by the GC when it shouldn't be."

src/core/memory.d
```diff
@@ -148,7 +148,7 @@ private
     extern (C) void gc_removeRange(const void* p ) nothrow @nogc;
     extern (C) void gc_runFinalizers( const scope void[] segment );
 
-    package extern (C) bool gc_inFinalizer();
+    package extern (C) bool gc_inFinalizer() nothrow @nogc @safe;
```
schveiguy: `core.gc.gcinterface.GC.inFinalizer` is only marked `nothrow`, and so is the extern(C) function. I could say we can probably set it to `@nogc`. `@safe` is probably fine too, but has a slight potential to break code (if there are any other private GC implementations out there).

But at the very least, the actual functions should be updated to reflect these attributes.

src/core/memory.d
```diff
@@ -847,6 +847,104 @@ struct GC
...
+        return gc_inFinalizer();
+    }
+
+    ///
+    @safe nothrow @nogc unittest
+    {
+        // Although unit tests can trigger garbage collection,
+        // and subsequent object finalization, it's highly unlikely
+        // that top-level `unittest` statements are executed from
+        // a GC finalization.
```
rainers: If you don't call this function from a dtor there is no uncertainty here.

src/core/memory.d
```diff
@@ -847,6 +847,104 @@ struct GC
     {
         gc_runFinalizers( segment );
     }
+
+    /**
+     * Queries the GC whether the current thread is running object finalization
+     * as part of a GC collection.
```
rainers: ... or an explicit call to runFinalizers.

src/core/memory.d
rainers: Not completely the same, but you could call runFinalizers explicitly with a range that contains the address of the dtor. Also used here:
https://github.com/dlang/druntime/blob/2429fb822ff11ec5692deea6805abb32f75cbc7f/src/rt/lifetime.d#L2634-L2636

src/core/memory.d
PetarKirov: I know, this what I meant, but perhaps I should have been more clear. Any suggestions on how to reword the comment?

src/core/memory.d
PetarKirov: Actually, initially, before reading the docs on `runFinalizers` I was planning on using it, as I thought that it works on data, but after reading that it works on code segments, I wasn't sure anymore that it fits the bill. 

Given that that I need to update the docs to mention `runFinalizers`, I'll also add it to the unit test.

src/core/memory.d
PetarKirov: Before I started working on this I thought that the only way to check if we're in a finalizer is to read the _`inFinalizer` static variable of the conservative GC implementation - I didn't expect my job to be so easy, as it was already exposed through the GC proxy, and so I forgot to update the proxy. Thanks for reminder, will fix.

On the actual attributes, I was not worried that they may put restrictions private GC implementation details, as after all, I think that in all implements this will amount to reading some variable, which should be safe in all circumstances (at worst the GC will need to add atomicLoads and stores, as it wouldn't be able to assume anything about the callers, but still, I don't see a big deal in that).

changelog/gc_in_finalizer.dd
PetarKirov: Yeah, probably I didn't use the best words here, but I still think this covers my intention.

Of course, you can't know with 100% certainty if an object will be finalized by the GC as that's indeterministic (!= nondeterministic) and that's why I qualified it that by saying "empirically".

But I think you can verify if an object was left to be destroyed by the GC - unless you manually called `runFinalizers` no one else will call the object destructor while having `inFinalizer` also be true, unless the object was no longer referenced and was being collected by the GC.

src/core/memory.d
rainers: If a comment is necessary at all, maybe: "code not called from a dtor is never executed during finalization".

PetarKirov:
@rainers @schveiguy Thanks for the review! I think I address all of your comments:
https://github.com/dlang/druntime/compare/da1a9b17498ccd66a33a476f6997f4e4d289aeac..ec3845aa0d61cba8c05aabcbda80f5286d569d71

src/gc/proxy.d
```diff
@@ -245,7 +245,7 @@ extern (C)
         return instance.runFinalizers( segment );
     }
 
-    bool gc_inFinalizer() nothrow
+    bool gc_inFinalizer() nothrow @nogc
```
PetarKirov: Note: not adding `@safe` here, because we can't access the `__gshared` variable `instance` in `@safe` code.

PetarKirov:
@CyberShadow is there any way to restart DAutoTest without restarting the other CIs? Like a restart button that only people in the GitHub org can click?

CyberShadow:
Unfortunately such a feature does not exist at the moment.

src/gc/proxy.d
schveiguy: Well, then should `@safe` be on the prototype? In essence, if we mark it safe there, then it's de-facto trusted. I'd rather see that be explicit. Or, we can omit safe for the time being (this is, after all, a new feature, so omitting safe shouldn't break any code).

But I find it troubling that this couldn't be made safe. How is memory allocation safe, since it is going to access the instance as well?

schveiguy: Do this I think:

```D
return (() @trusted => instance)().inFinalizer();
```

schveiguy: Rationale: Technically the access to `__gshared` instance is safe because instance is set only when accessing methods that need a real GC (this is how the proxy system works). And it should be safe to use the old instance in that case, because we aren't going to be in a finalizer for that thread anyway.

Note that I think this depends on the load of the instance pointer being atomic. But I think every other method in here assumes that too.

PetarKirov: If you want to go this road, I guess it would be better to rename `instance` to `_instance` and a `@safe` `instance` property, so that all proxy function can be made `@safe`

schveiguy: Yeah, technically, the instance should be loaded atomically. But if that is a problem, it's already a problem, so we can ignore that for now. I'd say a trusted accessor is probably the best route.

Not sure all proxy functions can be safe, but at least the instance fetching can be made safe.

PetarKirov: @schveiguy done, can you confirm if this is what you were looking for?

schveiguy: Ugh, I see you had to mark `profileStats` trusted. But for the same reason. I think it's reasonable.

Thanks.

PR #2874, opened 2019-12-17
Support merging of EH tables across Win64 DLLs

adamdruppe:
I'm not 100% happy with this yet but wanna get your input at the early stages.

Like the GC proxy, I don't think this is a good solution in the end, but it can be a temporary helper to people trying to use dlls today.

Currently, if you throw an exception in a dll, it cannot cross the boundary correctly for two reasons: the dynamic cast not working (John Colvin already has an open PR to address that... hackily again, but eh) and the EH tables being separate.

This PR addresses the latter problem by merging the tables through a little indirection. Combined with Colvin's PR, we can throw and catch exceptions across dll boundaries.

Longer term, I'd like to see druntime as a dll as a viable option, but that's pretty far off right now and we need something short term too.

My worry with this implementation is it might need some synchronization. I also kinda think all the initLibrary/unloadLibrary stuff would better belong in DllMain. But I don't want to get too invasive at this point.

### DIFF
--- a/src/rt/cast_.d
+++ b/src/rt/cast_.d
@@ -112,17 +112,17 @@ int _d_isbaseof2(ClassInfo oc, ClassInfo c, ref size_t offset)
 
 int _d_isbaseof(ClassInfo oc, ClassInfo c)
 {
-    if (oc is c)
+    if (oc.compareClassInfo(c))
         return true;
 
     do
     {
-        if (oc.base is c)
+        if (oc.base.compareClassInfo(c))
             return true;
 
         foreach (iface; oc.interfaces)
         {
-            if (iface.classinfo is c || _d_isbaseof(iface.classinfo, c))
+            if (iface.classinfo.compareClassInfo(c) || _d_isbaseof(iface.classinfo, c))
                 return true;
         }
 
--- a/src/rt/cast_.d
+++ b/src/rt/cast_.d
@@ -13,6 +13,15 @@
  */
 module rt.cast_;
 
+// because using == does a dynamic cast, but we
+// are trying to implement dynamic cast.
+bool compareClassInfo(ClassInfo a, ClassInfo b)
+{
+    if (a is b)
+        return true;
+    return a.info.name == b.info.name;
+}
+
 extern (C):
 
 /******************************************
@@ -76,19 +85,19 @@ void* _d_dynamic_cast(Object o, ClassInfo c)
 
 int _d_isbaseof2(ClassInfo oc, ClassInfo c, ref size_t offset)
 {
-    if (oc is c)
+    if (oc.compareClassInfo(c))
         return true;
 
     do
     {
-        if (oc.base is c)
+        if (oc.base.compareClassInfo(c))
             return true;
 
         // Bugzilla 2013: Use depth-first search to calculate offset
         // from the derived (oc) to the base (c).
         foreach (iface; oc.interfaces)
         {
-            if (iface.classinfo is c || _d_isbaseof2(iface.classinfo, c, offset))
+            if (iface.classinfo.compareClassInfo(c) || _d_isbaseof2(iface.classinfo, c, offset))
             {
                 offset += iface.offset;
                 return true;
--- a/src/rt/cast_.d
+++ b/src/rt/cast_.d
@@ -19,7 +19,7 @@ bool compareClassInfo(ClassInfo a, ClassInfo b)
 {
     if (a is b)
         return true;
-    return a.info.name == b.info.name;
+    return (a && b) && a.info.name == b.info.name;
 }
 
 extern (C):
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -225,7 +225,7 @@ version (Windows)
                             import core.stdc.string;
                             // assumes continuous append! If you ever sort the loading code or something
                             // be sure to fix this too.
-                            memmove(&ehTablesGlobal[idx], &ehTablesGlobal[idx + dllTables.length], typeof(ge).sizeof * (ehTablesGlobal.length - (idx + dllTables.length)));
+                            memmove(cast(void*) &ehTablesGlobal[idx], cast(void*) &ehTablesGlobal[idx + dllTables.length], typeof(ge).sizeof * (ehTablesGlobal.length - (idx + dllTables.length)));
                             ehTablesGlobal = ehTablesGlobal[idx + dllTables.length .. $];
                             // could realloc it down to the new size, but I suspect it will be useful
                             // again in the future anyway so might as well let the realloc on load reuse
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -89,6 +89,9 @@ version (CRuntime_Microsoft)
     extern(C) void init_msvc();
 }
 
+version (Win64)
+import rt.deh_win64_posix;
+
 /***********************************
  * These are a temporary means of providing a GC hook for DLL use.  They may be
  * replaced with some other similar functionality later.
@@ -102,6 +105,13 @@ extern (C)
     alias void* function()      gcGetFn;
     alias void  function(void*) gcSetFn;
     alias void  function()      gcClrFn;
+
+    // for exception handling, at least until druntime.dll actually works.
+    version (Win64)
+    {
+        alias void  function(const(FuncTable)[]*) ehSetFn;
+        alias const(FuncTable)[] function() ehGetFn;
+    }
 }
 
 version (Windows)
@@ -112,12 +122,12 @@ version (Windows)
      *      opaque handle to the DLL if successfully loaded
      *      null if failure
      */
-    extern (C) void* rt_loadLibrary(const char* name)
+    export extern (C) void* rt_loadLibrary(const char* name)
     {
         return initLibrary(.LoadLibraryA(name));
     }
 
-    extern (C) void* rt_loadLibraryW(const WCHAR* name)
+    export extern (C) void* rt_loadLibraryW(const WCHAR* name)
     {
         return initLibrary(.LoadLibraryW(name));
     }
@@ -126,6 +136,7 @@ version (Windows)
     {
         // BUG: LoadLibrary() call calls rt_init(), which fails if proxy is not set!
         // (What? LoadLibrary() is a Windows API call, it shouldn't call rt_init().)
+        // FYI LoadLibrary calls DllMain. DllMain calls rt_init.
         if (mod is null)
             return mod;
         gcSetFn gcSet = cast(gcSetFn) GetProcAddress(mod, "gc_setProxy");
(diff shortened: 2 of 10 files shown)


src/rt/dmain2.d
```diff
@@ -126,13 +133,58 @@ version (Windows)
...
             gcSet(gc_getProxy());
         }
+
+        import rt.sections_win64;
+
+        ehGetFn ehGet = cast(ehGetFn) GetProcAddress(mod, "_d_innerEhTable");
+        ehSetFn ehSet = cast(ehSetFn) GetProcAddress(mod, "_d_setEhTablePointer");
+
+        typeof(ehGet()) libraryEh;
+        if (ehGet)
```
thewilsonator: Please write this as 
```d
if (!ehGet)
    return mod;

libraryEh = ehGet();
...
```

src/rt/dmain2.d
```diff
@@ -144,8 +196,39 @@ version (Windows)
...
      */
-    extern (C) int rt_unloadLibrary(void* ptr)
+    export extern (C) int rt_unloadLibrary(void* ptr)
     {
+        // should this logic just be done in the dll's DllMain instead?
+
+        // need to clear the DLL's table out of the global list, so first that
+        // means fetching the dll's table...
+        ehGetFn ehGet = cast(ehGetFn) GetProcAddress(ptr, "_d_innerEhTable");
+        if (ehGetFn)
```
thewilsonator: ditto

src/rt/dmain2.d
```diff
@@ -144,8 +196,39 @@ version (Windows)
...
+        {
+            auto dllTables = ehGetFn();
+            import rt.sections_win64;
+//__gshared package(rt) const(FuncTable)[] ehTablesGlobal;
+
+            loop: foreach (entry; dllTables)
+            {
+                foreach (idx, ge; ehTablesGlobal)
+                {
+                    if (ge == entry)
```
thewilsonator:
```d
if (ge != entry)
    continue;
...
```

src/rt/dmain2.d
```diff
@@ -191,6 +274,12 @@ extern (C) int rt_init()
     version (CRuntime_Microsoft)
         init_msvc();
 
+    version(Win64)
+    {{
```
thewilsonator: why the double braces?

src/rt/sections_win64.d
```diff
@@ -218,6 +216,31 @@ do
...
+    ehTablesFull = ptr;
+}
+
+extern(C) export const(FuncTable)[] _d_innerEhTable() {
+    auto pbeg = cast(const(FuncTable)*)&_deh_beg;
+    auto pend = cast(const(FuncTable)*)&_deh_end;
+    return pbeg[0 .. pend - pbeg];
+}
+
+// finished
```
thewilsonator: what is this comment referring to?

src/rt/dmain2.d
adamdruppe: introduce a new scope for the local import.

src/rt/sections_win64.d
adamdruppe: it opens with

// Used for EH table merging across DLL boundaries


and then

// finished

ends that block.

adamdruppe:
Another high-level question for y'all: on Win32, this just works because it uses the built-in operating system functions. Why doesn't it just do that on Win64 as well? I'm pretty sure RaiseException works just as well there.

rainers:
> Why doesn't it just do that on Win64 as well?

Yeah, that's pretty unfortunate. Walter didn't want to go the last 5 percent (rough guess, most of the work for unwinding tables seems done). Not using standard exceptions is also a pain in a debugger, because you cannot break on an exception being thrown.

LDC uses C++ exception handling on Windows with a slightly modified termination handler.

src/rt/dmain2.d
```diff
@@ -321,7 +410,7 @@ private alias extern(C) int function(char[][] args) MainFunc;
  * runs embedded unittests and then runs the given D main() function,
  * optionally catching and printing any unhandled exceptions.
  */
-extern (C) int _d_run_main(int argc, char** argv, MainFunc mainFunc)
+export extern (C) int _d_run_main(int argc, char** argv, MainFunc mainFunc)
```
rainers: AFAICT `export` produces exported symbols on every executable, too. I would recommend just using a def file in case you actually need any of these symbols being exported.

adamdruppe: oh yeah i meant to remove that before pushing it here but missed a spot.

I was also experimenting with actually making druntime.dll and sprinkled `export` in parts around the runtime to get it to load. I actually got it sort of working but virtual functions always jumped to 0xffffffffff - i figure that data must not have been indirected correctly by dmd then scaled back my ambitions to this PR.

(I think druntime.dll is the correct solution for my plugin problem but I ran out of time trying to get that path to work.)

rainers: For proper shared runtime/library support, see https://dconf.org/2016/talks/thaut.html

rainers: > oh yeah i meant to remove that before pushing it here but missed a spot.

I meant all other functions with changed export, too. For DLL support, I implemented an `-exportall` switch a couple of years ago. It does what's also done on linux, i.e. export all symbols.

adamdruppe: Yeah, I looked at Thaut's code too, it looks like it did most the work reasonably well so I might try to revive that fork and get it merged when my urgent todo list is a bit shorter.

When/if that gets in, I would then delete all this new code and the GC proxy stuff too I'm pretty sure is useless witht hat.

adamdruppe:
On Tue, Dec 17, 2019 at 11:56:04AM -0800, Rainer Schuetze wrote:
> I meant all other functions with changed export, too. For DLL support, I implemented an `-exportall` switch a couple of years ago. It does what's also done on linux, i.e. export all symbols.

On exported symbols from exe, that's basically useless right?
I'm pretty sure GetProcAddress only works on dlls..

rainers:
> On exported symbols from exe, that's basically useless right?

It was meant to export all symbols of a DLL, an executable would then be linked against its import library and doesn't need to export symbols itself.

> I'm pretty sure GetProcAddress only works on dlls..

It works on any module regardless of extension. That includes the executable.

src/rt/dmain2.d
rainers: > When/if that gets in, I would then delete all this new code and the GC proxy stuff too I'm pretty sure is useless witht hat.

I don't remember whether Benjamins' code handled the non-standard exception handling in Win64. I'm not familiar with the implementation, but it doesn't look like it would work out of the box by just patching up import relocations.

adamdruppe:
sooooo the test

rainers:
> sooooo the test appears to be failing because the output says

Yeah, the output of the test is verified. You can get rid of the error by not automatically exporting the symbols, but rely on a def file for the DLL.

Speaking of tests, can you add one in druntime/test/shared or enable some of the posix tests on Windows, too?

John-Colvin:
The tests from https://github.com/dlang/druntime/pull/2828/files might help here.

adamdruppe:
Yeah, I think my next step will be to merge those PRs. I'm not in love with the name comparison but it works as a short-term thing. And then the def file is OK too.

I just need to fix my Windows build environment, I totally messed it up again this week :(

adamdruppe:
just pulled john colvin's branch into here too, let's see how the tests go with that.....

BTW i updated visual studio earlier and everything broke for the druntime dev. dmd master make -f win64 broke. at least build.d still worked. then druntime make utterly failed - apparently the update killed all the paths it depended on. what a hassle.

adamdruppe:
BTW I also can't run the tests locally, the MS C compiler complains no include path. Again surely related to the path breakage from that VS update. so fickle.

so i'll just push here to run the unittests. at least i can compile my own things again though.

rainers:
> just pulled john colvin's branch into here too, let's see how the tests go with that.....

There are quite some more places that expect pointer identity of TypeInfo instances, e.g.
https://github.com/dlang/druntime/blob/e6b89f6d4d796db66c44d1629e9e4b4e90b80bd3/src/rt/lifetime.d#L205
https://github.com/dlang/druntime/blob/e6b89f6d4d796db66c44d1629e9e4b4e90b80bd3/src/rt/lifetime.d#L779
https://github.com/dlang/druntime/blob/e6b89f6d4d796db66c44d1629e9e4b4e90b80bd3/src/rt/aaA.d#L238
https://github.com/dlang/druntime/blob/e6b89f6d4d796db66c44d1629e9e4b4e90b80bd3/src/gc/impl/conservative/gc.d#L3468

grep for "typeid.*is" for a couple more places. I'm not sure we want to pay the performance penalty of a string compare in all these places. Maybe some of these can be left as is because they are compiler callbacks where both TypeInfo instances are guaranteed to be from the same binary, but that's not the case for its usage in the (precise) GC.

adamdruppe:
yeah, I'd like to fix everything but that's not the short term goal - i just want throw and catch to work here. the rest can stay broken until druntime.dll actually works.

but idk wtf i did here. it worked on my computer last week and now i need to specify the .def file in my own library to export this. Did that visual studio update affect even that? i can't see how but i'm so lost now.

most recent commit removes the `export` qualifier but like those symbols DO have to be exported and putting the .def file on every individual d dll is not great. so weird, with `export`, I see it in the exe but not the dll!

rainers:
optlink doesn't like the empty def file, it must at least contain "EXETYPE  NT"

MoonlightSentinel:
Any updates on this? Getting this to the finish line would be great.

adamdruppe:
oh over christmas i forgot about this and never got back to it after.

I think it basically works though, just a few minor fixes and prolly update to master....

src/rt/cast_.d
```diff
@@ -13,6 +13,15 @@
  */
 module rt.cast_;
 
+// because using == does a dynamic cast, but we
+// are trying to implement dynamic cast.
```
WalterBright, line 17: Since this is comparing for equality, the name for it should be something like "equals".

src/rt/cast_.d
```diff
@@ -13,6 +13,15 @@
...
+bool compareClassInfo(ClassInfo a, ClassInfo b)
+{
+    if (a is b)
+        return true;
+    return (a && b) && a.info.name == b.info.name;
```
WalterBright, line 22: When would the ClassInfo ever be null?

adamdruppe: I vaguely recall seeing it happen in December when working on this originally but I don't remember where. Possible I was just being defensive too.

rainers:
For reference, https://github.com/dlang/druntime/pull/3138 goes the opposite direction, because template names can be the same for different parameter types. I think the name generation should be fixed before it can be used for type comparison.

rmanthorpe:
I think it's well worth pursuing this and getting it ready to merge. If someone can point me at wherever the names get generated (I found it once, but I can't find it now) I'll see if I can do something there.

adamdruppe:
OK, I'm coming back to this now. ... idk where that name generation is, I kinda wish it was in druntime, RTInfo could just plop a hash(T.mangleof) and ctfe that right up.

...in fact i could just add that here...

adamdruppe:
ah the generation is in `toobj.d` in dmd's source, search for `cd.toPrettyChars`. That leads to `dclass.d`'s `class ClassDeclaration` and yada yada yada `dsymbol.d` has `toPrettyChars`' definition... which calls `toPrettyCharsHelper` which has comment about fully qualified template arguments but just calls more toChars in most cases, but `dtemplate.d` overrides it to call `toCBufferInstance` which calls `hdrgen.d`'s `toCBufferInstance` which makes a `DsymbolPrettyPrintVisitor`...

Which is defined in that same file and seems to actually do the work. @rmanthorpe

kinke:
John's PR included here has landed in #3543, so this would definitely need a rebase now, reducing it to the DMD-specific WinEH fix only. The testcase only needs the removal of https://github.com/dlang/druntime/blob/a8df8912d4b47c27d2f8bed6fb75bc6d7b649fe7/test/shared/src/dynamiccast.d#L78-L82

RazvanN7:
@adamdruppe are you willing to pursue this?

adamdruppe:
This was finished almost two years ago. I don't even know what the new merge conflicts are anymore... I think half of it can simply be deleted as other MRs got it, just the table merge itself, and even that only in dmd, since ldc and gdc im pretty sure solved the problem on the compiler side.

Geod24:

PR #2875, opened 2019-12-17
[Musl] add to syslog.d

Cogitri:


Cogitri:
See also: #2876

Geod24:
@Cogitri : Not sure this should be backported. The previous PR corrected system definition which were wrong, but pure additions are probably not worth it.

Cogitri:
Hm, alrighty. Figured it was a simple change so I thought it'd make sense to backport it but I don't mind closing this either.

-------- Original Message --------

PR #2876, opened 2019-12-17
[Musl] add to syslog.d

Cogitri:


Cogitri:
Hm, do I need to do something more for this to land in an actual release? It appears this isn't in the lastest DMD yet so I need to mirror this change in my projects.

thewilsonator:
It should be. 2.091 was released 8th march

Cogitri:
A derp, I built DMD 2.091 but forgot to actually install it. Thanks for the message :)

thewilsonator:
Haha. No problems.

PR #2877, opened 2019-12-18
Switch default behavior of tests to only run tests when they are present

schveiguy:
This should have been done a long time ago, but I forgot about it. The note said it would be switched by 2.080, but we are pushing 2.090 next...

The new default behavior makes it so any unittests present will cancel the running of main, even on a successful test run. The old behavior can be had by passing in a DRT switch (see changelog notes).


Geod24:
This caused a regression on dlang-tour: https://issues.dlang.org/show_bug.cgi?id=20500

PR #2878, opened 2019-12-18
The rest of std::string's API

SSoulaimane:
Implemented:
* atomic ref counting in the GCC version.
* growing with `resize()`.
* `reserve()`, `shrink_to_fit()`, `insert()`, `replace()`, and `swap()`.


SSoulaimane:
Tested manually on linux, windows and osx (with https://github.com/dlang/dmd/pull/10593), and the tests are passing. On linux, only the old ABI is working which can be turned on with `version = _GLIBCXX_USE_CXX98_ABI`, the c++11 ABI is practically unusable until there is a move constructor in D, and issue [20321](https://issues.dlang.org/show_bug.cgi?id=20321) is closed.

src/core/internal/lifetime.d
```diff
@@ -107,3 +107,15 @@ void emplaceInitializer(T)(scope ref T chunk)
...
+/*
+Simple swap function.
+*/
+void swap(T)(ref T lhs, ref T rhs)
+{
+    import core.lifetime : move;
+
+    T tmp = move(lhs);
+    lhs = move(rhs);
+    rhs = move(tmp);
```
kinke: This should be slightly more efficient:
```D
T tmp = move(lhs);
moveEmplace(rhs, lhs);
moveEmplace(tmp, rhs);
```

SSoulaimane: Definitely.

thewilsonator:
This is green. Is it good to go?

SSoulaimane:
It should be good to go unless a mistake is discovered. Note that OSX tests are still disabled and that dlang/dmd#10593 is still a prerequisite.

PR #2879, opened 2019-12-21
Port std::tuple

SSoulaimane:
C++ `std::tuple`.


SSoulaimane:
It seems variadic templates are not get the correct C++ mangling.

PR #2880, opened 2019-12-22
Merge remote-tracking branch 'upstream/stable' into merge_stable

MartinNowak:


Geod24:
Is it necessary to have empty merge commits like this ?

PR #2881, opened 2019-12-23
Fix issue 20459 - Runtime arg parsing should stop at POSIX delimiter

Geod24:
```
This makes druntime argument parsing consistent with the behavior of
options on POSIX systems.
It also ensures that druntime argument can be passed down to other programs,
e.g. for 'dub' and 'rdmd'.
```

### DIFF
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -510,15 +510,18 @@ private extern (C) int _d_run_main2(char[][] args, size_t totalArgsLength, MainF
         char[][] argsCopy = buff[0 .. args.length];
         auto argBuff = cast(char*) (buff + args.length);
         size_t j = 0;
+        import rt.config : rt_cmdline_enabled;
+        bool parseOpts = rt_cmdline_enabled!();
         foreach (arg; args)
         {
-            import rt.config : rt_cmdline_enabled;
-
-            if (!rt_cmdline_enabled!() || arg.length < 6 || arg[0..6] != "--DRT-") // skip D runtime options
-            {
-                argsCopy[j++] = (argBuff[0 .. arg.length] = arg[]);
-                argBuff += arg.length;
-            }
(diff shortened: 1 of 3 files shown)


src/rt/dmain2.d
```diff
@@ -510,15 +510,18 @@ private extern (C) int _d_run_main2(char[][] args, size_t totalArgsLength, MainF
...
-            import rt.config : rt_cmdline_enabled;
-
-            if (!rt_cmdline_enabled!() || arg.length < 6 || arg[0..6] != "--DRT-") // skip D runtime options
-            {
-                argsCopy[j++] = (argBuff[0 .. arg.length] = arg[]);
-                argBuff += arg.length;
-            }
+            // Do not pass Druntime options to the program
+            if (!skipOpts && arg.length >= 6 && arg[0 .. 6] == "--DRT-")
+                continue;
```
Geod24: Also did a small refactor here to avoid the double negative boolean logic.

src/rt/dmain2.d
```diff
@@ -510,15 +510,18 @@ private extern (C) int _d_run_main2(char[][] args, size_t totalArgsLength, MainF
...
+            // https://issues.dlang.org/show_bug.cgi?id=20459
+            if (arg.length == 2 && arg == "--")
```
CyberShadow: `arg.length == 2` is redundant, no?
```suggestion
            if (arg == "--")
```

src/rt/dmain2.d
CyberShadow: I suggest inverting `skipOpts` to avoid that double negative too.

Geod24:
Fixed according to @CyberShadow comments

MoonlightSentinel:
This patch was incomplete. Options encountered after `--` still affect the current process!

https://issues.dlang.org/show_bug.cgi?id=22523

PR #2887, opened 2019-12-26
[changelog]: Improve GC.inFinalizer changelog entry

PetarKirov:
* Correct documentation link
* Add a note about `GC.runFinalizers`


PetarKirov:
Thanks!

PR #2888, opened 2019-12-30
Removed version(unittest) blocks because they introduce big overhead

alexandrumc:
user code:

```d
module foo;

import core.sync.barrier;

unittest 
{
     assert(1 == 1);
}
```

library:

```d
module core.sync.barrier;

version(unittest) 
{
      import core.sync.condition;
      import core.sync.semaphore;
      // other imports
      // classes
      // template instantiations
}
```

```bash
dmd -unittest foo.d

# will trigger semantic analysis on every top-level version(unittest) block from the library; 
# in the example above, this means semantic analysis also on condition, semaphore etc.
# -> makes the compilation take up to 5x longer on some examples
```

This is related to: https://github.com/dlang/phobos/pull/6450

### DIFF
--- a/src/core/sync/barrier.d
+++ b/src/core/sync/barrier.d
@@ -112,40 +112,35 @@ private:
 // Unit Tests
 ////////////////////////////////////////////////////////////////////////////////
 
-
-version (unittest)
+unittest
 {
-    private import core.thread;
+    import core.thread;
 
+    int  numThreads = 10;
+    auto barrier    = new Barrier( numThreads );
+    auto synInfo    = new Object;
+    int  numReady   = 0;
+    int  numPassed  = 0;
 
-    unittest
+    void threadFn()
     {
-        int  numThreads = 10;
-        auto barrier    = new Barrier( numThreads );
-        auto synInfo    = new Object;
-        int  numReady   = 0;
-        int  numPassed  = 0;
-
-        void threadFn()
+        synchronized( synInfo )
         {
-            synchronized( synInfo )
-            {
-                ++numReady;
-            }
-            barrier.wait();
-            synchronized( synInfo )
-            {
-                ++numPassed;
-            }
+            ++numReady;
         }
-
-        auto group = new ThreadGroup;
-
-        for ( int i = 0; i < numThreads; ++i )
+        barrier.wait();
+        synchronized( synInfo )
         {
-            group.create( &threadFn );
+            ++numPassed;
         }
-        group.joinAll();
-        assert( numReady == numThreads && numPassed == numThreads );
     }
+
+    auto group = new ThreadGroup;
+
+    for ( int i = 0; i < numThreads; ++i )
+    {
+        group.create( &threadFn );
+    }
+    group.joinAll();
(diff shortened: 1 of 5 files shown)


Geod24:
Can you define what big overhead you are witnessing ?
The change in itself doesn't seem bad, but you probably want to make those functions static.

alexandrumc:
@atilaneves

alexandrumc:
> Can you define what big overhead you are witnessing ?
> The change in itself doesn't seem bad, but you probably want to make those functions static.

Please see the description I added. Thanks.

wilzbach:
This doesn't fix the actual problem. The bug here is in DMD and this is just white-washing a tiny bit of the problem! There are likely many other occurrences of this in phobos or other libraries.

atilaneves:
> The bug here is in DMD and this is just white-washing a tiny bit of the problem! 

Is `version(unittest)` iffy and does it have to be fixed? Yes. In the main problem in dmd? Yes. But:

* This PR makes things better wrt compilation times
* I think the code after the diff should have been what was there to begin with

> There are likely many other occurrences of this in phobos or other libraries.

Unfortunately, yes. Also unfortunately, the fix in dmd is unlikely to be trivial (@alexandrumc  has been looking into the root cause).

Even if `version(unittest)` didn't add overhead, however, I think this diff is still for the better.

alexandrumc:
> This doesn't fix the actual problem. The bug here is in DMD and this is just white-washing a tiny bit of the problem! There are likely many other occurrences of this in phobos or other libraries.

I know what the problem is. This is only a small part of the solution. The first step is to remove all the version(unittest) blocks, but I have to remove them in small incremental steps to be easier to review.

Geod24:
@atilaneves : What's the plan with `version (unittest)` ? Asking because I hope it won't get the same treatment as `unittest` blocks, as they are two different beasts (namely: one can use `version (unittest)` to provide `unittest`-only behavior / utilities).

@alexandrumc : Thanks for adding the explanation.

atilaneves:
> @atilaneves : What's the plan with `version (unittest)` ? Asking because I hope it won't get the same treatment as `unittest` blocks, as they are two different beasts (namely: one can use `version (unittest)` to provide `unittest`-only behavior / utilities).

I'm not sure yet, I have to discuss it with Walter.

schveiguy:
The issue with version(unittest) specifically is that they generally are for testing the project or sometimes even only the module at hand. They are not usually for importers.

version(unittest) blocks get semantic analysis even if they are irrelevant to the importer. unittest functions do not. There has to be a very very good reason to have a version(unittest), whereas just including unittests is fine.

Thanks for the change, and hope to see more about the problem/solution described somewhere.

CyberShadow:
> Asking because I hope it won't get the same treatment as `unittest` blocks, as they are two different beasts (namely: one can use `version (unittest)` to provide `unittest`-only behavior / utilities).

How would that work? Wouldn't you need to include the module in the compilation to make use of said behavior/utilities, which would then enable `version(unittest)` (and `unittest`) blocks regardless?

PR #2890, opened 2020-01-01
Fix issue 20468 - `emplace` should forward argument lvalueness correctly

SSoulaimane:
`emplace` passes arguments by ref instead of using `forward`, this fixes it.

### DIFF
--- a/src/core/lifetime.d
+++ b/src/core/lifetime.d
@@ -58,7 +58,7 @@ T* emplace(T, Args...)(T* chunk, auto ref Args args)
 {
     import core.internal.lifetime : emplaceRef;
 
-    emplaceRef!T(*chunk, args);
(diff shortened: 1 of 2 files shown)


src/core/lifetime.d
```diff
@@ -1197,6 +1197,28 @@ pure nothrow @safe /* @nogc */ unittest
     // need ctor args
     static assert(!is(typeof(emplace!A(buf))));
 }
+
+//constructor arguments forwarding
+unittest @safe
```
thewilsonator: I think this needs to be `@safe unittest` not `unittest @safe`

SSoulaimane:
It's ready now. I put `forward` everywhere it's needed and updated a couple of tests.

PR #2891, opened 2020-01-02
Fix DDoc REF macro link for GC.inFinalizer

PetarKirov:


PetarKirov:
One does not simply write DDoc without checking the DAutoTest diff :D

PR #2893, opened 2020-01-03
More version(unittest) blocks removed

alexandrumc:
Please see #2888


src/core/atomic.d
```diff
@@ -818,8 +818,7 @@ private
 // Unit Tests
 ////////////////////////////////////////////////////////////////////////////////
 
-
-version (unittest)
+@betterC @safe pure nothrow unittest
 {
     void testXCHG(T)(T val) pure nothrow @nogc @trusted
```
MoonlightSentinel: Keep the old behaviour (ditto for other functions, structs, classes)

```suggestion
    static void testXCHG(T)(T val) pure nothrow @nogc @trusted
```

src/core/atomic.d
```diff
@@ -893,292 +892,289 @@ version (unittest)
         testLoadStore!(MemoryOrder.raw, T)(val);
     }
 
-    @betterC @safe pure nothrow unittest
-    {
```
MoonlightSentinel: Why not keep these blocks to signify that they different tests?
(Bonus: That would imply a much smaller diff and does not screw with the line history)

PR #2897, opened 2020-01-06
[stable] Fix regression for GNU backtraces

kinke:
In case the 'module' part is a path containing dashes, which crept in with #2834.


Geod24:
Dub failure unrelated (I'll fix it soon)

PR #2900, opened 2020-01-08
Remove redundant selective import of MAP_ANON

Geod24:
MAP_ANON is provided by core.sys.posix.sys.mman,
so the selective imports don't make sense.


ibuclaw:
I thought I was having a weird dejavu moment. #2593

PR #2901, opened 2020-01-08
Fix Issue 20490 - malloc and calloc should be `@safe` / `@trusted`

n8sh:
There is no good reason not to annotate them correctly.

core.stdc.stdlib.d has this comment explaining why they aren't `@trusted`:

> We don't mark these `@trusted`. Given that they return a void*, one has to do a pointer cast to do anything sensible with the result. Thus, functions using these already have to be `@trusted`, allowing them to call `@system` stuff anyway.

That comment is longer than just writing `@trusted` twice. Moreover since that comment was written Phobos has adopted the style of wrapping individual statements in `@trusted` lambdas instead of marking entire functions as `@trusted`.

### DIFF
--- a/src/core/stdc/stdlib.d
+++ b/src/core/stdc/stdlib.d
@@ -150,14 +150,10 @@ else
     void    srand(uint seed);
 }
 
-// We don't mark these @trusted. Given that they return a void*, one has
-// to do a pointer cast to do anything sensible with the result. Thus,
-// functions using these already have to be @trusted, allowing them to
-// call @system stuff anyway.
 ///
-void*   malloc(size_t size);
+void*   malloc(size_t size) @trusted;
 ///
-void*   calloc(size_t nmemb, size_t size);
+void*   calloc(size_t nmemb, size_t size) @trusted;
 ///
 void*   realloc(void* ptr, size_t size);
 ///
@@ -209,7 +205,6 @@ size_t  wcstombs(scope char* s, scope const wchar_t* pwcs, size_t n);
 ///
 version (DigitalMars)
 {
-    // See malloc comment about @trusted.
     void* alloca(size_t size) pure; // non-standard
 }
 else version (GNU)


Geod24:
I think you meant `calloc` instead of `free` in the issue and here?

atilaneves:
I tend to agree with this. Does anyone have an objection? While it's true that the returned pointer has to be cast or sliced to be made useful, allocating itself isn't @system.

schveiguy:
I agree with the original comment. Consider using `malloc` in a `@safe` function:

```D
auto data = (() @trusted => cast(int *)malloc(int.sizeof))();
```
If you make malloc `@safe` or `@trusted` (it should never be safe, but trusted, as the code itself is not checked by the compiler), you *still* need the trusted escape for the cast! Why is it important to make malloc trusted then?

The point of marking clearly unchecked and source-unavailable functions as `@trusted` is to make it more convenient to call from safe code. I see zero benefit to marking this trusted.

> That comment is longer than just writing `@trusted` twice.

The point of the comment is not to explain that it's really trusted, it's to prevent people from making PRs to make the functions `@trusted`. If you don't put a comment where the functions are, then people might come along, and say "hey, these really are safe, so let's mark 'em!".

schveiguy:
Is `malloc(0)` guaranteed to return null? Because if not it is definitely not safe

dkorpel:
> Why is it important to make malloc trusted then?

This sounds backwards to me. A function is `@system` a call to it from `@safe` code can result in undefined behavior. If that can't happen, then it is `@safe` or `@trusted`, whether the annotation is useful shouldn't even matter.

I actually would consider it useful though, since I would write your piece of code like this:
```D
int* newIntPtr() @safe {
    auto ptr = malloc(int.sizeof);
    return (() @trusted => cast(int*) ptr)();
}
```
The `@trusted` part should be as small as possible, so I don't want to get an error that 'malloc is not callable from `@safe`' and find out there's no actual risk there, it's just that someone labelled it incorrectly because they wouldn't think the correct label would be useful.

> Is malloc(0) guaranteed to return null? Because if not it is definitely not safe

It's implementation defined, it can return `null` or a pointer. That doesn't make it not safe though, unless you can demonstrate how that can result in memory corruption in `@safe` code.

schveiguy:
> This sounds backwards to me. A function is `@system` if a call to it from `@safe` code can result in undefined behavior. If that can't happen, then it is `@safe` or `@trusted`, whether the annotation is useful shouldn't even matter.

If you don't use malloc for what it's intended, then of course, you cannot incur undefined behavior.

But if you are going to use malloc to allocate memory for any use, you're going to have to use trusted escapes *anyway*. There is no point to having this marked trusted, as it doesn't give you any extra safety -- you have to break the safety guarantees anyway for the cast.

> I actually would consider it useful though, since I would write your piece of code like this:

Your entire function is trusted without it being marked trusted, why would you do that? In other words, your function is exactly the same as the line I wrote in terms of using trusted escapes.

There is no benefit to calling the malloc apart from the cast. In fact, I'd say this is promoting risky behavior -- you want the unsafe cast and the allocation to be as close as possible to inform the reader where the casted memory comes from. I've never seen a malloc with a cast be separated for any good reason.

> I don't want to get an error that 'malloc is not callable from `@safe`' and find out there's no actual risk there

Or, you put it in one line (as everyone who's ever used malloc ALWAYS does), and you don't get this error.

> That doesn't make it not safe though, unless you can demonstrate how that can result in memory corruption in `@safe` code.

You would now have a pointer to memory that is unallocated. It is undefined behavior to access such memory. Granted, you can't access it without a cast in most cases, but if you are not going to cast it, why are you calling `malloc` in the first place? You can't even free it, because that is not `@safe`.

The reason I consider this a problem is because `@safe` code assumes it can always dereference any pointer safely. Either a segfault occurs, or the data is allocated and owned by the caller.

Compare this to std.experimental.allocator -- it returns a `void[]`. This is allowed to point at something not allocated with zero length WITHOUT being unsafe, because the memory region "owned" by the array is 0 size. The distinction is important.

This whole PR seems completely pointless to me. It seems like it is intended to get rid of a comment, without actually making any meaningful change. The comment isn't even public documentation. There's no reason to mark a function as trusted that is not manually checked, and will ultimately be called from a context that is going to have some system or trusted code. Leave it as system, it's not going to hurt anything.

I've said my piece, and if this gets marked as trusted anyway, so be it. I think we should err on the side of leaving non-checkable code that can't possibly be used in a safe context as trusted, but it's not up to me to ultimately make that decision. It's not going to break D for this PR to be pulled, it will just look weird.

aG0aep6G:
> I actually would consider it useful though, since I would write your piece of code like this:
> 
> ```d
> int* newIntPtr() @safe {
>     auto ptr = malloc(int.sizeof);
>     return (() @trusted => cast(int*) ptr)();
> }
> ```
> 
> The `@trusted` part should be as small as possible, so I don't want to get an error that 'malloc is not callable from `@safe`' and find out there's no actual risk there, it's just that someone labelled it incorrectly because they wouldn't think the correct label would be useful.

You made the `@trusted` part smaller than possible there. You can't guarantee in the `@trusted` portion that casting `ptr` to `int*` is safe.

"As small as possible" doesn't mean "as small as the compiler lets you". You still have to make sure that `@trusted` functions (including tiny lambdas) are actually safe.

dkorpel:
> There is no point to having this marked trusted, as it doesn't give you any extra safety

It's not like marking it `@system` gives any extra safety, just confusion as to why `malloc` isn't marked how it is.

> Granted, you can't access it without a cast in most cases

With a `void*`, you can't access it in any case in `@safe` code. That's why malloc is `@safe`.

> but if you are not going to cast it, why are you calling malloc in the first place?

Again, I don't get the notion that an attribute needs to be useful before a function can earn it. Would you say that `BigInt.opUnary!"-"` should not have the `@nogc` attribute, because when are you ever going to use `std.bigint` in a `@nogc` context considering most other operations require the GC? Or would you say: `opUnary!"-"` does not allocate with the garbage collector, so it is `@nogc`, simple as that.

> The reason I consider this a problem is because `@safe` code assumes it can always dereference any pointer safely.

Except a `void*`.

> Or, you put it in one line (as everyone who's ever used malloc ALWAYS does),

[Not true](https://github.com/apache/thrift/blob/30ac2598e84928d9af7066b5d3248b7aea4376b2/lib/d/src/thrift/async/libevent.d#L199).

schveiguy:
> [Not true](https://github.com/apache/thrift/blob/30ac2598e84928d9af7066b5d3248b7aea4376b2/lib/d/src/thrift/async/libevent.d#L199).

There are numerous problems with this. First, there's no reason to use a dynamic array to assign (actually, it's a slice assign, which calls a druntime function or C function). Just dereference. Second, there's no reason not to cast immediately (a `T*` is just as good as a `void *` in all circumstances here). Third, it adds a range even if the T is not a reference-holding type (maybe true in all usages, but I'd rather the function use static if or static assert to verify, no reason not to). Fourth, the responsibility of removing the range and freeing the data is not properly matched with the allocator. It's on the supplied handler to do these things, which could be easily forgotten. A properly written system would match the teardown of the data to the layer which built it (most likely the best thing to do here is use a smart pointer type that does all of it for you, and is `@trusted`, like `RefCounted`). This code is not intended to be `@safe` by any measure.

dkorpel:
> There are numerous problems with this.

That was just a simple counter-example of your claim that 'everyone who's ever used malloc ALWAYS puts the cast on the same line', I'm not saying it demonstrates why `malloc` should be `@safe`. 

I'm more interested in whether you think that `BigInt.opUnary!"-"` shouldn't be `@nogc` by your own logic.

schveiguy:
> I'm more interested in whether you think that `BigInt.opUnary!"-"` shouldn't be `@nogc` by your own logic.

Of course it should be `@nogc`. It's not an apt comparison to extern(C) code that we don't control and cannot manually check being marked as such. It's not genuine to mark it as `@trusted`, so there better be a good reason. There isn't one. The reason here is "so we can delete a comment".

dkorpel:
> It's not an apt comparison to extern(C) code that we don't control and cannot manually check being marked as such

Okay, but by that standard no `extern` function can ever be `@safe` / `@trusted`.

> It's not genuine to mark it as @trusted

It is being trusted in druntime though. If I were to link in a C runtime library with a `malloc` implementation that randomly corrupts memory on certain allocation sizes, then all bets are off right from `Dmain` since [initialization of the garbage collector](https://github.com/dlang/druntime/blob/aff248cb8917767b5fe1c99464b0aa02a174c8d1/src/gc/impl/conservative/gc.d#L126) calls malloc.
(this coincidentally is another example of `malloc` without a pointer cast on the same line ;-) )

If `malloc` is `@system` because we don't trust the underlying C code, then `new` is `@system` too since it relies on `malloc`.

schveiguy:
You misunderstand trusted. The whole point of trusted is to have a `@safe`-style interface to a `@system` function. But malloc itself sees no benefit of being trusted. Things that use malloc absolutely can be trusted. An example (and look, there are uses for delaying the cast!):

```D
T[] allocate(T)(size_t count) @trusted
{
   if(count == 0) return null;
   auto result = malloc(count * T.sizeof);
   if(result is null) return null;
   foreach(i; 0 .. count)
       result[i * T.sizeof .. (i + 1) * T.sizeof] = typeid(T).initializer[];
   return cast(T[])result[0 .. T.sizeof * count];
}
```

I think there are 2 lines that could be safe, it's not critical to have those checked by the compiler though. In other words, marking `malloc` safe solves nothing for this function.

dkorpel:
> The whole point of trusted is to have a `@safe`-style interface to a `@system` function.

`malloc`'s interface is already compliant with `@safe`. If "`@safe`-style" is a requirement, then it should be properly defined in the spec.

> But malloc itself sees no benefit of being trusted

And `BigInt.opBinary!"-"` sees no benefit of being `@nogc`.
I thought you didn't think that was comparable because `malloc` is "extern(C) code that we don't control", but now you say "things that use malloc absolutely can be trusted". 

To me, whether to apply an attribute should be a simple unopinionated matter of "does the function conform to the specification of the attribute". It doesn't matter to me:
- whether the attribute is useful, since 'useful' is subjective
- whether the function has a body or not
- whether the attribute catches bugs or code smells unrelated to its definition

I should be able to write this:
```D
import std;
import core.stdc.stdlib;
void main() @safe {
    void* a = null;
    int i = 0;
    do {
        a = malloc(1024*1024*1024);
        i++;
    } while (a);
    writeln("memory: ", i);
}
```
You probably do not like that code. That code might be ugly, but the `@safe` attribute has no business judging that, and the compiler raising an error suggesting this may not be memory-safe is just a false positive.

But that's just my thought, I am trying to figure out what your rules are for applying an attribute. From your replies I gather that you do find those 3 points I listed important to an extent, but you are not explicit about it. I'd appreciate it if you detail your guidelines to applying attributes.

schveiguy:
I should be more definitive and clearer: I think calling malloc without casting is probably safe, so technically, since casting is unsafe, malloc could be marked `@trusted` (please, not `@safe`, it's not mechanically checked, so that is untrue).

But every time we mark something `@trusted`, it's a certification that someone has checked it to make sure it's safe to call in all cases. We've done this for quite a few functions in libc, justified mostly because they have to follow the standard. Let's not forget though, that memory issues are not design issues, but implementation issues. So we always take a risk to mark things that we don't check.

In some cases, it's worth it! We can't live without certain items that come from libc, and so we have to weigh the benefits against the risks of (in general) marking a function trusted without actually verifying it's safe for all inputs. In the case of malloc, we are 100% dependent on it, so if it's not implemented properly, then we are at risk anyway. But in my opinion, the benefits are immeasurably small. You get no benefits from marking it trusted, as nobody is going to write the function you just wrote. Except in a debate about this PR. If you get no benefits, why take any risk at all? I'd much rather have a wrapper that is actually `@safe` (i.e. returns a typed pointer, or a `void[]` or a typed array) than mark this trusted.

I still am not convinced that `malloc(0)` is perfectly safe. The answer can't be, "well, you shouldn't be casting the result of malloc, that's unsafe". I hope you realize how absurd that sounds to someone who wants to use `malloc`. But in any case, feel free to proceed (please change to `@trusted`, not `@safe`), if you really really think this is important. To me, it's nonsensical, and taking risk without benefit.

dkorpel:
> as nobody is going to write the function you just wrote

Oh, people write that. You don't want to know :-)

> If you get no benefits, why take any risk at all?

You do get benefits. People who are in the process of writing a `malloc` wrapper might stumble on a compile error and think "huh, just calling `malloc` is `@system`? I haven't even done a cast yet. Have i overlooked something?". Then they might find the declaration, open a new PR like this, and potentially have another discussion about it. The existence of the comment above the declaration and this PR demonstrate how the `@system` annotation is counter-intuitive and creates confusion.

Also there is no risk to take. Either `malloc` is indeed memory-safe and the attribute is correct, or it is not and Druntime as a whole is compromised anyway so a misapplied `@trusted` on `malloc` is the least of your problems. 

What's the benefit of `malloc` being `@system`?

> if you really really think this is important

This specific case is not important to me, I just wish that deciding whether something is `@safe` were a simple uncontroversial question about the presence or absence of memory corruption, and nothing else. But somehow I always end up in discussions with people suggesting something should be `@system` without demonstrating presence of memory corruption.

- maybe [union member acces should be `@system`](https://issues.dlang.org/show_bug.cgi?id=19916) as well
- maybe [casting to `void*` should be `@system](https://forum.dlang.org/post/dkzfysvfjofyueaprokz@forum.dlang.org) as well

And now we have:
- maybe functions that have no benefit of being `@safe` (or `@trusted` in this case) should be `@system` as well.

schveiguy:
> What's the benefit of `malloc` being `@system`?

In fact, it is `@system`. It's allowed to do whatever it wants in its implementation, and not worry about safety checks from the compiler. `@system` means "not checked for safety".

> maybe functions that have no benefit of being `@safe` (or `@trusted` in this case) should be `@system` as well.

You misunderstand. I mean functions we *do not control* and are not checkable (and also have no benefit of being `@trusted`). If you're going to lie, it should be for a good reason ;)

n8sh:
> please change to `@trusted`, not `@safe`

Okay.

> To me, it's nonsensical, and taking risk without benefit.

To me there is no risk, because there is no unsafe behavior that an incorrect annotation saves you from. The benefit is _correct and useful error messages_. It puts a light on exactly where things can go "boom" and it doesn't train the programmer to ignore error messages about safety. "Oh yeah, someone probably just forgot to annotate this C function properly, so I'll wrap this in `@trusted`."

dkorpel:
> In fact, it is `@system`.

That's not a benefit.

> It's allowed to do whatever it wants in its implementation

That contradicts "Things that use malloc absolutely can be trusted.".
I am okay with the position that `malloc` is not checked so it is assumed `@system`, but then your wrappers must be `@system` as well. You can't say "I can't promise a simple call to `malloc` won't result in memory corruption, but this wrapper that makes arbitrary calls to `malloc` and then some, _that_ absolutely can be trusted!". `malloc` being `@trusted` is necessary for your wrapper to be `@trusted`. They are either both `@system` or `malloc` is `@trusted`. And since most of Druntime already made the assumption that `malloc` can be `@trusted`, marking it `@system` is just inconsistent.

Attributes on functions without bodies are never mechanically checked, they are always based on an assumption. Is `core.stdc.math` actually `nothrow` `@nogc` `@trusted`? I can link in an implementation of `atan2f` that violates them all if I want, but since the code is opaque, we just have to assume the attributes based on the interface and assumed implementation.

RazvanN7 (approved):

dkorpel:
@ibuclaw Why did you remove the auto-merge label?

RazvanN7:
As previously stated, marking malloc/calloc as trusted/safe brings 0 benefit since you need a cast of the void* to be able to use the data anyway. Therefore, I intend to close this, however, will leave it open for another 72h. If anyone can come up with a compelling case where having malloc trusted brings any benefits then I will reconsider.

atilaneves:
I don't know about this. Especially since [Mallocator.allocate](https://github.com/dlang/phobos/blob/668d43c5df5f8821c19f05f6bc23a98f7f75c5b9/std/experimental/allocator/mallocator.d#L31) is `@trusted`, and for the life of me I don't know why anyone would rather call `malloc` directly instead.

n8sh:
> As previously stated, marking malloc/calloc as trusted/safe brings 0 benefit since you need a cast of the void* to be able to use the data anyway.

Sometimes you also want to do things with the pointer. For instance, checking if the result is null before slicing it instead of after. The common idiom of combining allocation, casting, and slicing in a single `@trusted` lambda actually violates the specification for `@trusted`:

```d
ubyte[] allocate(size_t s) @safe @nogc nothrow
{
    ubyte[] result = (() @trusted => (cast(ubyte*) malloc(s))[0 .. s])();
    if (result is null) // checks if the array's pointer is null
        return null; // returns an array with null pointer and length 0
    return result;
}
```

But see <https://dlang.org/spec/function.html#safe-interfaces>:

>When it is only called with safe values and safe aliasing, a function has a safe interface when:
>* it cannot exhibit undefined behavior, and
>* **it cannot create unsafe values that are accessible from other parts of the program (e.g., via return values, global variables, or ref parameters)**, and
>* it cannot introduce unsafe aliasing that is accessible from other parts of the program.

>Functions that meet these requirements may be `@safe` or `@trusted`. Function that do not meet these requirements can only be `@system`.

When malloc returns null the return value of the `@trusted` lambda is an unsafe value -- a ubyte[] with a null pointer and non-zero length -- and the enclosing function is responsible for sanitizing it. It is the entire enclosing function that should be marked `@trusted`.  However, the style checker for Phobos mandates that `@trusted` lambdas be used instead of entire `@trusted` functions. The correct way to write the above with lambdas is:

```d
ubyte[] allocate(size_t s) @safe @nogc nothrow
{
    void* result = (() @trusted => malloc(s))();
    if (result is null)
        return null;
    return (() @trusted => (cast(ubyte*) result)[0 .. s])();
}
```

but these lambdas are visual noise so it isn't surprising when programmers try to use as few as possible. With this PR the correct version is no uglier than the incorrect version:

```d
ubyte[] allocate(size_t s) @safe @nogc nothrow
{
    void* result = malloc(s);
    if (result is null)
        return null;
    return (() @trusted => (cast(ubyte*) result)[0 .. s])();
}
```

ibuclaw:
AFAIU, the vision is to make these trusted lambda's redundant in the language, and offer something better wrt `@safe`-ty.  @bcarneal was going to draft something up along with @WebDrake I believe, but both are understandably busy. :-)

dkorpel:
> and for the life of me I don't know why anyone would rather call `malloc` directly instead.

`malloc` is a decades old stable and familiar function, while `Mallocator` adds an extra layer of indirection to it. Does it do out of memory checks? Does it initialize the memory in debug mode? Will its behavior change in the future? Especially considering it's located in `std.experimental`, it's iffy to rely on it. Also, `import std.experimental.allocator.mallocator` imports a good chunk of Phobos and increases compile time by 80 ms compared to `import core.stdc.stdlib`.

ljmf00 (approved):

ibuclaw:
What part of malloc's interface is `@safe`?

ljmf00:
> What part of malloc's interface is `@safe`?

According to D `@safe`,... everything. We should think about the inverted question if we want to prove it is not `@safe`.

According to POSIX:
> The functionality described on this reference page is aligned with the ISO C standard. Any conflict between the requirements described here and the ISO C standard is unintentional. This volume of POSIX.1-2017 defers to the ISO C standard.
>
> Upon successful completion with size not equal to 0, malloc() shall return a pointer to the allocated space. If size is 0, either:
> - A null pointer shall be returned and errno may be set to an implementation-defined value, or
> - A pointer to the allocated space shall be returned. The application shall ensure that the pointer is not used to access an object.
>
> Otherwise, it shall return a null pointer and set errno to indicate the error.

Note that:
- the returned `void*` can be either `NULL`, or have a valid pointer with the size described
- `void*` doesn't describe the size of the pointed memory
- `void*` can't be dereferenced in any way, in safe code
- casting `void*` to a type with a size equivalence of the target pointer is valid in safe code (e.g. `ptrdiff_t`, `size_t`)
- storing a `void*` is safe

If you see anything unsafe, from the D perspective, specified in POSIX or ISO C standard about `malloc` interface, please tell.

schveiguy:
I know I said I wouldn't comment more... well I did anyway, but this continual debate just irks me.

Remember that every time we mark something as `@trusted` that is not D code, we accumulate baggage -- baggage of assuming that something is safe that we don't control, and that the compiler that cannot verify.

When we do this, it might be OK. It might be necessary even. But it better be worth it. Because every time we do that we add to the list of possible holes that we allow in the D safety story.

What is the value here? That we can print a pointer value that was just allocated? Is that worth it? Does it give us more capability or functionality? How much more code can be marked as `@safe` without any trusted escapes?

Yes, I know in the case of an odd usage of malloc, it means one extra line would not have to be marked as `@trusted`. Meh.

I'm going to unsubscribe to this thread, just because I don't need more garbage in my email. But I'll leave you with 2 points here:

1. Yes, you are right, this is OK in terms of D safety. Because you can allocate data, and never access it, that use case is now covered.
2. This will move the needle exactly zero in terms of making D safer, simply because nobody will use this without trusted escapes. It's a trivial difference if the call to malloc is split from the cast *because there is still a trusted escape*. But there is a huge difference between fully safe code and code with one or more trusted escapes. That is where the focus should be.

ibuclaw:
> > What part of malloc's interface is `@safe`?
> 
> According to D `@safe`,... everything. We should think about the inverted question if we want to prove it is not `@safe`.
> 
> According to POSIX:
> > The functionality described on this reference page is aligned with the ISO C standard. Any conflict between the requirements described here and the ISO C standard is unintentional. This volume of POSIX.1-2017 defers to the ISO C standard.
> >
> > Upon successful completion with size not equal to 0, malloc() shall return a pointer to the allocated space. If size is 0, either:
> > - A null pointer shall be returned and errno may be set to an implementation-defined value, or
> > - A pointer to the allocated space shall be returned. The application shall ensure that the pointer is not used to access an object.
> >
> > Otherwise, it shall return a null pointer and set errno to indicate the error.
> 
> Note that:
> - the returned `void*` can be either `NULL`, or have a valid pointer with the size described

Or an invalid pointer. 

> - `void*` doesn't describe the size of the pointed memory
> - `void*` can't be dereferenced in any way, in safe code

Exactly, malloc returns untyped, uninitialized garbage.  The only way to make it usable is to do any of the unsafe operations (slicing, dereferencing, etc).

ibuclaw:
NB: At some point in the future, `@trusted` will likely have the same  enforcement checks that `@safe` currently has. Killing off this trusted lambda anti-pattern that undermines `@safe`.

pbackus:
I really don't understand why this is controversial. `malloc`, as specified by the ISO C standard, has a `@safe` interface. The only justification to mark it as `@system` is if we are not willing to trust the ISO C standard—and if that's the case, we have a much larger problem on our hands.

dkorpel:
> Because every time we do that we add to the list of possible holes that we allow in the D safety story.

Only if you consider "number of `@trusted` annotations" as the defining metric for that. Arbitrary calls to `malloc` can be made from `@safe` code, so it is de-facto `@trusted` and a possible risk anyway. The idea that removing a `@trusted` annotation here decreases the surface of possible holes or amount of code that should be checked is only an illusion.

ljmf00:
> 1. Yes, you are right, this is OK in terms of D safety. Because you can allocate data, and never access it, that use case is now covered.
> 2. This will move the needle exactly zero in terms of making D safer, simply because nobody will use this without trusted escapes. It's a trivial difference if the call to malloc is split from the cast _because there is still a trusted escape_. But there is a huge difference between fully safe code and code with one or more trusted escapes. That is where the focus should be.

You can't say nobody will propagate a `void*` to cast it afterwards or simply print it for instrumentation/debugging purposes. That is not exactly zero. Sure, it has a limited use case, but the fact that it can bring as little as the non-zero value doesn't mean it is exactly zero. Furthermore, I can promise, according to ISO C standard, the behaviour of malloc is safe and there is exactly zero possibility of having unsafe code with the returned value, without casting it to a type with indirections or larger than the size of the target machine addresses. And we already agreed on the first point.

> > Note that:
> > 
> > * the returned `void*` can be either `NULL`, or have a valid pointer with the size described
> 
> Or an invalid pointer.

No. All pointers returned by the `malloc` are either `NULL` or unique and valid, there is no way to return an invalid pointer. This is described by ISO C standard. If you find an implementation that doesn't follow that it is considered a bug. The pointer returned by `malloc(0)` is also valid, given the size of the requested memory. You literally request 0 bytes, and the returned pointer points to 0-sized memory. Btw, the pointer uniqueness part is achieved because malloc allocates a small payload of metadata that describes the allocated memory block for functions like `free` and `realloc`.

> > * `void*` doesn't describe the size of the pointed memory
> > * `void*` can't be dereferenced in any way, in safe code
> 
> Exactly, malloc returns untyped, uninitialized garbage. The only way to make it usable is to do any of the unsafe operations (slicing, dereferencing, etc).

1. Uninitialized memory is safe in D, as long as it is memory without indirections. This can be arguable if it should be that way, but that is how the specification treats void initialization. This behaviour is not described as undefined behaviour but instead, implementation-defined. Even if uninitialized memory was unsafe, you are just pointing to valid memory, given the requested size. It is your responsibility, when casting, to ensure memory safety. Other than that, you can do whatever safe operation to that pointer.
2. That is not the only way. As I described you can print the pointer, cast it to memory without indirections or convert it to memory with less or the same size as the target machine addresses. This can be used by the user to do whatever they want to do: for debugging purposes, logging, instrumentation, ...

> I really don't understand why this is controversial. `malloc`, as specified by the ISO C standard, has a `@safe` interface. The only justification to mark it as `@system` is if we are not willing to trust the ISO C standard—and if that's the case, we have a much larger problem on our hands.

Exactly. Also, the file header comment describes that it follows ISO C standard, so any deviation from the ISO C rules can lead to undefined behaviour.

pbackus:
It seems like everyone here agrees that `malloc`, as specified by the C standard, satisfies the requirements of `@trusted`, and therefore could be marked as `@trusted` without violating memory safety. The disagreement, then, is about whether there exist circumstances in which a function that satisfies the requirements of `@trusted` should nonetheless be marked as `@system`.

Personally I think the answer is "no", for reasons which @dkorpel and others have already articulated, but if the consensus is "yes", then we ought to nail down in writing exactly what those circumstances are. Otherwise, we are going to continue getting bogged down in discussions like this forever.

pbackus:
There's also some disagreement w.r.t. how we ought to treat "platform" dependencies like C library functions when it comes to memory safety.

One position is, "the platform is a source of risk, and we should minimize our exposure to that risk by trusting it as little as possible." From this position, it makes sense to decide whether to trust each individual OS or system library function on a case-by-case basis by weighing its potential usefulness against the added risk of memory corruption it exposes us to.

Another position is, "if the platform is compromised, it's already game over and there's nothing we can do about it at the language level, so we might as well just trust it unconditionally." From this position, it makes sense to just assume that the OS and system libraries behave the way they say they will (e.g., that they conform to any relevant standards), and not worry about individual functions.

ibuclaw:
> No. All pointers returned by the malloc are either NULL or unique and valid, there is no way to return an invalid pointer. 

Yes there is, I've encountered it a few times.

RazvanN7:
I, honestly, think that this PR is a 0 sum game. Whatever the outcome, there will be no improvement. I suggest that we stop wasting time debating this and just go one way or the other. 

cc @WalterBright for a final decision.

pbackus:
@ljmf00 The pointer returned from `malloc(0)` is not required to point to allocated memory; it just has to be a valid argument to `free`.

> If the size of the space requested is zero, the behavior is implementation-defined: either a null pointer is returned, or the behavior is as if the size were some nonzero value, except that **the returned pointer shall not be used to access an object.**

Source: http://port70.net/~nsz/c/c11/n1570.html#7.22.3

pbackus:
@ljmf00 I think you and @ibuclaw are using different definitions of "valid", and I was attempting to clarify by pointing out that the result of `malloc(0)` is valid *for some uses* (the ones included in your definition) but not for others (the ones included in Iain's definition). It is probably better for discussion if we all avoid ambiguous terms like "valid" and say exactly what we mean.

I agree that the pointer returned from `malloc` is always a [safe value][1].

[1]: https://dlang.org/spec/function.html#safe-values

ibuclaw:
> I agree that the pointer returned from `malloc` is always a [safe value](https://dlang.org/spec/function.html#safe-values).

One problem with that example is that what malloc returns is untyped, and casting from untyped to typed is unsafe, even if the result of the cast is a safe pointer that can be passed around and dereferenced in `@safe` code.

ljmf00:
> One problem with that example is that what malloc returns is untyped, and casting from untyped to typed is unsafe, even if the result of the cast is a safe pointer that can be passed around and dereferenced in `@safe` code.

Right, but the problem of being unsafe is the cast, not the malloc itself. Everything you manipulate with the returned value of `malloc` is safe, unless you do unsafe operations.

Geod24:
How frequently does one need to call `malloc` from a `@safe` context ?
I'd wager the answer is binary: zero or once. Because writing a wrapper to it will be the go-to for most people, as @n8sh [showed](https://github.com/dlang/druntime/pull/2901#issuecomment-1062075072).

So, given this change is 1) Only providing value for an artificial edge case; 2) Highly controversial; I am of the opinion that we should just close this and move on to do more productive things. As mentioned, `Mallocator` covers this case (and the 80ms compile time argument is IMO invalid, as the added dependencies are likely to be added from somewhere else anyway).

dkorpel:
> 1) Only providing value for an artificial edge case; 

People keep arguing that it shouldn't be `@trusted` because the utility is "near 0", neglecting the fact that marking it `@system` has utility of exactly 0. Just because there is one less `@trusted` annotation doesn't mean there's less actual `@trusted` code.

> we should just close this and move on to do more productive things.

It's not about `malloc`, it's about these kind of discussions that keep popping up. See Paul's comment: https://github.com/dlang/druntime/pull/2901#issuecomment-1063041818

atilaneves:
FWIW, I think malloc should be safe or trusted. It definitely shouldn't be any different from Mallocator. Also, people should use Mallocator instead.

WalterBright:
Lots of good arguments from both sides. @dkorpel brought up the "see how much memory can be allocated" example, and had the best pedantic argument. @schveiguy had the best pragmatic argument.

Some more pragmatic points:

1. Part of the point of having @system code is to flag code for review by the more senior members of the team. I know I always zero in on any use of malloc/calloc/realloc/free as being rich sources of bugs. Being able to mechanically flag all @system code is valuable.
2. The only safe thing one can do with the return value is test it for null or not-null. This is of extremely low utility, and I'd be suspicious of why any code needed to do only that.
3. I like the symmetry of malloc being paired with free. Having one be trusted and the other system just doesn't sit well.
4. If some code took the return from malloc and just passed it off to some other opaque @safe function, I wouldn't want that to pass the compiler.

Everybody has had the chance to clearly advance their views on this, there's nowhere else to go but make a decision.

So "no" it is.

atilaneves:
> I like the symmetry of malloc being paired with free. Having one be trusted and the other system just doesn't sit well.

They're not symmetrical. One can be called willy-nilly, the other is the source of an entire class of bugs.

> If some code took the return from malloc and just passed it off to some other opaque @safe function, I wouldn't want that to pass the compiler.

The only way that this situation would break memory safety is if that paremeter isn't scope.

pbackus:
>>    If some code took the return from malloc and just passed it off to some other opaque `@safe` function, I wouldn't want that to pass the compiler.
>
> The only way that this situation would break memory safety is if that paremeter isn't scope.

Even then, it would not break memory safety unless the `@safe` function called an incorrect `@trusted` function. There is nothing you can do with the return value of `malloc` in `@safe` code that can corrupt memory. `scope` makes no difference either way.

dkorpel:
> Some more pragmatic points:

From what I've read on the newsgroup, some people consider `@safe` formal verification of the absence of memory corruption (modulo bugs), while others consider it to just mean "turn dmd's memory safety checker on". "pragmatic vs pedantic" if you will. 

I've been on the latter camp because I thought that aligned with D's vision, but considering your verdict on this issue, as well as [your recent quote](https://forum.dlang.org/post/sud4gs$aeb$1@digitalmars.com) "`@safe` does not supply complete memory safety. The addition of `@live` fills in much of the rest.", I'm curious what the new vision for `@safe` is.

WalterBright:
Leaving malloc as @system does not compromise the integrity of @safe in any way.

The goal of @safe (along with @live) is to guarantee memory safety.

> Even then, it would not break memory safety unless the @safe function called an incorrect @trusted function. There is nothing you can do with the return value of malloc in @safe code that can corrupt memory. scope makes no difference either way.

That is 100% correct. But I, as a code reviewer, would find it *highly suspicious* to see the result of @trusted malloc being sent to an opaque @safe function. As a code reviewer, I wouldn't want that stench to pass unnoticed.

WalterBright:
@pbackus I know it when I see it. (The real world is rarely cut and dried.)

PR #2896, opened 2020-01-05
More version(unittest) blocks removed

alexandrumc:
Please see #2888

PR #2902, opened 2020-01-08
Replaced version (unittest) with version (DRuntimeUnittest) to avoid unnecessary overhead when compiling with -unittest

alexandrumc:
Related to #2888 and #2896


Geod24:
I appreciate the focus on this, however maybe we should pause and consider how widespread this problem is likely to be. Any library is likely to suffer from this. I'm thinking Vibe.d, for example.

Should the compiler blindly run `unittest`s in all dependencies ? Is Phobos that special that we want this rule to apply to all dependencies but druntime/Phobos ? I don't think so. Running your dependencies' unittests is not scalable and rarely what you want, especially during development.

Can we start to acknowledge this and move to a solution on the language level, rather than pushing the problem down to all libraries ?

alexandrumc:
> Should the compiler blindly run `unittest`s in all dependencies ? Is Phobos that special that we want this rule to apply to all dependencies but druntime/Phobos ? I don't think so. Running your dependencies' unittests is not scalable and rarely what you want, especially during development.

The compiler doesn't run `unittest`s in all dependencies. For example, in the following code

```d
module foo;

import std.smth;

unittest 
{
     import std.stdio : writeln;
     writeln("Target unittest: ", a);
}

version (unittest)
{
     int a  = 10;
}

void main() {}
```

```d
module std.smth;

unittest
{
    import std.stdio : writeln;
    writeln("Dependency unittest: ", b);
}

version (unittest)
{
    int b = 5;
}
```

only `Target unittest: 10` will be printed. However, `int b = 5` will be visible and semantic analyzed without any usefulness. This is why we want to remove / replace all the `version (unittest)` s from Phobos and druntime.

Geod24:
@alexandrumc : Right, it doesn't do it for non-root modules.
But that depends on your build system. If you're using dub with `library` you're fine, if you use `rdmd` (or `dub` with `sourceLibrary`, I'd wager) those are going to be root modules, and your unittests are going to be run.
Druntime is a bit of a special case here, as it's always linked, and its modules are never root modules.

alexandrumc:
> @alexandrumc : Right, it doesn't do it for non-root modules.
> But that depends on your build system. If you're using dub with `library` you're fine, if you use `rdmd` (or `dub` with `sourceLibrary`, I'd wager) those are going to be root modules, and your unittests are going to be run.
> Druntime is a bit of a special case here, as it's always linked, and its modules are never root modules.

Sure, that's true. But by changing the `version (unittest)`s I am not trying to address the issue you described. `version {}` and `unittest{}` are two different mechanisms and by chance two separate aspects of the same problem: the behaviour triggered by the `-unittest` switch. Here I am just trying to solve the problem that I described in my previous comment.

atilaneves:
> I appreciate the focus on this, however maybe we should pause and consider how widespread this problem is likely to be. Any library is likely to suffer from this. I'm thinking Vibe.d, for example.

Correct. This is why none of my own libraries have unittest blocks anywhere that client code could see.

> Should the compiler blindly run `unittest`s in all dependencies ?

No.

> Is Phobos that special that we want this rule to apply to all dependencies but druntime/Phobos ?

Yes. It's the standard library, it's _the_ dependency.

> I don't think so. Running your dependencies' unittests is not scalable and rarely what you want, 
> especially during development.

The issue is semantic analysis, not actually running tests.

> Can we start to acknowledge this and move to a solution on the language level, rather than pushing the problem down to all libraries ?

Yes. But for now this is easier. I need to discuss this with @WalterBright

src/core/time.d
```diff
@@ -117,7 +117,7 @@ ulong mach_absolute_time();
 }
 
 //To verify that an lvalue isn't required.
-version (unittest) private T copy(T)(T t)
+version (DRuntimeUnittest) private T copy(T)(T t)
```
Geod24: I believe you need to keep this because [it is called from a templated unittest](https://github.com/dlang/druntime/blob/26fa565714d76df892eb327e98235aa854468766/src/core/time.d#L2199-L2213).

At least that's what the CI suggests:
```
/var/lib/buildkite-agent/builds/ci-agent-89c4044a-12c8-4cab-b953-c7a634bd4891-5/dlang/druntime/build/distribution/bin/../imports/core/time.d(2202,21): Error: undefined identifier `copy`, did you mean import `core`?
--
  | /var/lib/buildkite-agent/builds/ci-agent-89c4044a-12c8-4cab-b953-c7a634bd4891-5/dlang/druntime/build/distribution/bin/../imports/core/time.d(2210,16): Error: undefined identifier `copy`, did you mean import `core`?
  | /var/lib/buildkite-agent/builds/ci-agent-89c4044a-12c8-4cab-b953-c7a634bd4891-5/dlang/druntime/build/distribution/bin/../imports/core/time.d(2211,16): Error: undefined identifier `copy`, did you mean import `core`?
  | /var/lib/buildkite-agent/builds/ci-agent-89c4044a-12c8-4cab-b953-c7a634bd4891-5/dlang/druntime/build/distribution/bin/../imports/core/time.d(2212,25): Error: undefined identifier `copy`, did you mean import `core`?
  | /var/lib/buildkite-agent/builds/ci-agent-89c4044a-12c8-4cab-b953-c7a634bd4891-5/dlang/druntime/build/distribution/bin/../imports/core/time.d(2272,20): Error: undefined identifier `copy`, did you mean import `core`?
  | /var/lib/buildkite-agent/builds/ci-agent-89c4044a-12c8-4cab-b953-c7a634bd4891-5/dlang/druntime/build/distribution/bin/../imports/core/time.d(2283,13): Error: undefined identifier `assertApprox`
  | /var/lib/buildkite-agent/builds/ci-agent-89c4044a-12c8-4cab-b953-c7a634bd4891-5/dlang/druntime/build/distribution/bin/../imports/core/time.d(2004,18): Error: template instance `core.time.MonoTimeImpl!cast(ClockType)0` error instantiating


```

Geod24:
> Yes. But for now this is easier. I need to discuss this with @WalterBright

@atilaneves : Great, thanks. And happy to know you guys are working on improving compilation speed!

src/core/time.d
alexandrumc: I don't think that's actually a problem. I think that the `-version=DRuntimeUnittest` is somehow not passed to the compiler when unittests are compiled.

alexandrumc: Are the `Makefile`s used for testing exactly the ones that are located in `phobos/`? Are they different?

PetarKirov:
I addressed the issues with `core.time.copy` (see https://github.com/dlang/druntime/commit/372b0f56ddabab9a174323ba860c5fc0dda324d4), rebased and squashed.

@alexandrumc let's wait and see if the CI will be green, hopefully no further work on this PR will be necessary from you.

PetarKirov:
@alexandrumc @atilaneves  @Geod24 @thewilsonator @wilzbach Please take another look.

PetarKirov:
BTW, I just rembered about the discussion in this PR: https://github.com/dlang/phobos/pull/6159. @atilaneves what's your take on it?

src/core/time.d
PetarKirov: For the record, I both of you are correct.
The root cause was that even though the function was used only from druntime's unit tests, they were nested inside the `MonoTime` template, which was instantiated from phobos, without passing `-version=CoreUnittest`. I solved this by guarding all nested (and undocumented) unit tests with `version (CoreUnittest)`.

PR #2904, opened 2020-01-10
fix Issue 20497 - thread with limited stackspace crashes depending on…

rainers:
… size of TLS

on Posix systems, add the TLS size to the requested stack size for new threads

related issues:
- https://github.com/ldc-developers/ldc/issues/3127
- https://github.com/dlang/dub/issues/1812

### DIFF
--- a/src/core/thread/osthread.d
+++ b/src/core/thread/osthread.d
@@ -752,11 +752,13 @@ class Thread
         version (Windows) {} else
         version (Posix)
         {
+            size_t stksz = adjustStackSize( m_sz );
+
             pthread_attr_t  attr;
 
             if ( pthread_attr_init( &attr ) )
                 onThreadError( "Error initializing thread attributes" );
-            if ( m_sz && pthread_attr_setstacksize( &attr, m_sz ) )
+            if ( stksz && pthread_attr_setstacksize( &attr, stksz ) )
                 onThreadError( "Error initializing thread stack size" );
         }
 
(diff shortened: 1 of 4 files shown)


rainers:
The output of the test will hopefully show what platforms are actually using the stack for TLS.

src/core/thread/osthread.d
```diff
@@ -3189,6 +3179,23 @@ private void onThreadError(string msg) nothrow @nogc
...
 
+version (Posix)
+private size_t adjustStackSize(size_t sz) nothrow @nogc
+{
+    if (sz == 0)
+        return 0;
+
+    // TLS uses the top of the stack, so add its size to the requested size
+    sz += externDFunc!("rt.sections_elf_shared.sizeOfTLS",
+                       size_t function() @nogc nothrow)();
```
kubo39: Isn't it better to add `version(CRuntime_Glibc)` here?

kubo39:
Nit: In glibc, we can use `__pthread_get_minstack` which is added in 2.15.

src/core/thread/osthread.d
rainers: Probably, as FreeBsd doesn't seem to have this issue. I've added the version check.

rainers:
> Nit: In glibc, we can use `__pthread_get_minstack` which is added in 2.15.

Indeed, looks like a simpler solution. When searching for it I noticed a couple of bug reports due to undefined references to this symbol. Would it raise the libc version requirements for druntime?

PR #2907, opened 2020-01-14
core.demangle: Add fast path for obvious non-D names

kinke:
Trying to avoid exceptions for most C(++) mangled names.


kinke:
[This is primarily intended to simplify EH debugging - an exception's stack trace generation involving newly thrown & caught exceptions (at least statically allocated) isn't very nice.]

PR #2908, opened 2020-01-16
Fix issue 19909: core.stdc.errno missing POSIX error code on Windows

Geod24:
Taken from https://www.winehq.org/pipermail/wine-cvs/2018-July/128910.html


Geod24:
> Your PR doesn't reference any Bugzilla issue.

@wilzbach : It would be great if it could recognize "Fix ######". "issue" is just redundant and commit title should generally be kept < 50 chars.

PR #2910, opened 2020-01-19
Add support for iOS, tvOS and watchOS

jacob-carlborg:
I've only tested this on a 64 bit iPhone running iOS 12. 32 bit is obsolete and I don't have access to the other platforms. All unit tests pass.

Most of the fixes are just missing declarations.

I recommend looking at the individual commits.

Phobos PR: https://github.com/dlang/phobos/pull/7355.
LDC PR: ldc-developers/ldc#3288.


jacob-carlborg:
@kinke

src/rt/sections_darwin_64.d
```diff
@@ -0,0 +1,178 @@
...
+    version = Darwin;
+else version (iOS)
+    version = Darwin;
+else version (TVOS)
+    version = Darwin;
+else version (WatchOS)
+    version = Darwin;
+
+version (Darwin):
+version (D_LP64):
```
Geod24: The commit message says "both x86-64 and ARM", but I assume you mean `s/ARM/AArch64/` ? `version (ARM)` is 32 bits only.

adamdruppe:
very cool

Geod24:
> 32 bit is obsolete and I don't have access to the other platforms.

Is there a point in having the `version (ARM)` then ?

src/rt/sections_darwin_64.d
jacob-carlborg: Yes, Apple calls AArch64 ARM64 and `arm64-apple-ios12.0` is a valid target triple in Clang/LDC.

jacob-carlborg:
> Is there a point in having the version (ARM) then ?

I added it in a few places where it didn't require any extra effort. The thread related code is just a copy-paste of what Dan Olson already had.

moon-chilled:
Isn't WatchOS still 32-bit?

jacob-carlborg:
> Isn't WatchOS still 32-bit?

Apple Watch series 4 and 5 are 64 bit.

src/core/stdc/assert_.d
```diff
@@ -17,6 +17,15 @@
...
 module core.stdc.assert_;
 
+version (OSX)
+    version = Darwin;
+else version (iOS)
+    version = Darwin;
+else version (TVOS)
+    version = Darwin;
+else version (WatchOS)
+    version = Darwin;
```
kinke: Wrt. this: there's a predefined `darwin` version, but it's deprecated (`use OSX instead`) - time to undo that and redefine its semantics maybe? Or add a new `Darwin` one with these semantics?

jacob-carlborg: I would prefer to do that. Walter was against that https://github.com/dlang/dmd/pull/5231#issuecomment-151361493. IIRC he regrets that we have a version identifier for Posix as well.

PetarKirov:
@jacob-carlborg I reviewed the changes and they look good on their own, but I'd prefer to cross-check with their actual libc headers. Are they available somewhere online in browsable form, or you have to download a zip file (IIRC) from Apple's website?

jacob-carlborg:
> but I'd prefer to cross-check with their actual libc headers. Are they available somewhere online in browsable form, or you have to download a zip file (IIRC) from Apple's website?

The actual SDK is not available. But Apple has their open source part of macOS available here [1]. For iOS there’s nothing of interest, but the header files seem to be the same. As you can see, many of the declarations have some form of an availability macro attached to them. I think most of the interesting part would be in the libc project. 

Some one has uploaded the macOS SDK to Dropbox here [2]. 

[1] https://opensource.apple.com/release/macos-10145.html

[2] https://github.com/multiarch/crossbuild/blob/529881986478b18c9973b8979fa6287d44a77c8f/Dockerfile#L67

src/core/threadasm.S
```diff
@@ -406,21 +406,32 @@ fiber_switchContext:
 
 .text
 .align  2
-.global fiber_switchContext
+#ifdef __APPLE__
+.global _fiber_switchContext
+_fiber_switchContext:
+#else
```
kinke: I guess this name decoration business should look like this:
```asm
.global CSYM(fiber_switchContext)
.type   CSYM(fiber_switchContext), %function
CSYM(fiber_switchContext):
```
It's used like this for the other functions below.

src/core/threadasm.S
```diff
@@ -406,21 +406,32 @@ fiber_switchContext:
...
 #if defined(__ARM_PCS_VFP) || (defined(__ARM_PCS) && !defined(__SOFTFP__)) // ARM_HardFloat  || ARM_SoftFP
```
kinke: This should probably be exactly the same condition as in line 434 according to the comment. So extracting it as its own `#define` should pay off.

jacob-carlborg: Do you have a suggestion for the name of this `#define`?

kinke: I *think* the condition is supposed to reflect whether FP registers need to be saved & restored when switching fibers. I *think* that's the case when hardware units are enabled (`version (D_HardFloat)`), even if FP registers aren't used to pass/return values when calling a function ('softfp' - use hardware units like hard-float, but be ABI-compatible to soft-float).

So if that's correct, something like `ARM_NonSoftFloat` might be workable.

jacob-carlborg: Fixed.

src/core/threadasm.S
jacob-carlborg: Fixed.

jacob-carlborg:
Anyone wants to have a look at this PR?

src/core/sys/darwin/mach/thread_act.d
```diff
@@ -135,3 +140,103 @@ version (i386)
...
+        ARM_CPMU_STATE64 = 18
+    }
+
+    enum
+    {
+        ARM_AMX_STATE = 24,
+        ARM_AMX_STATE_V1 = 25
+    }
+
+    struct arm_thread_state_t
```
MoonlightSentinel: This is identical to `arm_thread_state32_t`, why not provide a simple alias?
```suggestion
    alias arm_thread_state_t = arm_thread_state32_t;
```

src/core/sys/darwin/mach/thread_act.d
```diff
@@ -135,3 +140,103 @@ version (i386)
...
+    {
+        uint[13] r; /// General purpose register r0-r12
+        uint sp;    /// Stack pointer r13
+        uint lr;    /// Link register r14
+        uint pc;    /// Program counter r15
+        uint cpsr;  /// Current program status register
+    };
```
MoonlightSentinel:
```suggestion
    }
```

src/core/sys/darwin/mach/thread_act.d
```diff
@@ -135,3 +140,103 @@ version (i386)
     kern_return_t thread_resume(thread_act_t);
     kern_return_t thread_get_state(thread_act_t, thread_state_flavor_t, thread_state_t*, mach_msg_type_number_t*);
 }
+else version (AppleARM)
```
MoonlightSentinel: Could you include a comment linking to the original definitions?

src/core/sys/posix/locale.d
```diff
@@ -68,7 +76,7 @@ char*    setlocale(int category, const char* locale);
 /// Set the per-thread locale
 locale_t uselocale (locale_t locale);
 
-version (OSXBSDLocale)
+version (BSDLocale)
```
MoonlightSentinel:
```suggestion
version (DarwinBSDLocale)
```

To emphasize that this version still includes OSX & friends.

src/core/sys/posix/sys/types.d
```diff
@@ -852,7 +852,7 @@ else version (Darwin)
 {
     version (D_LP64)
     {
-        enum __PTHREAD_SIZE__               = 1168;
```
MoonlightSentinel: This change applies to existing prorgams. Correct me if I'm wrong but these new values appear to be the defaults for linux, ... which differ from OSX IIRC.

src/core/sys/posix/locale.d
jacob-carlborg: The Posix layer in Darwin is from BSD. But ok.

src/core/sys/posix/sys/types.d
jacob-carlborg: I'm not sure I understand. The new values are what's written in the header files in the SDK for macOS, iOS, tvOS and watchOS. I don't know if this has been updated or where the original values come from.

src/core/sys/darwin/mach/thread_act.d
jacob-carlborg: Fixed.

src/core/sys/darwin/mach/thread_act.d
jacob-carlborg: Fixed.

src/core/sys/darwin/mach/thread_act.d
jacob-carlborg: Done.

src/core/sys/posix/locale.d
jacob-carlborg: Done.

src/core/sys/posix/sys/types.d
MoonlightSentinel: You're right, seems like the current values are outdated (they originate from the first official commit to druntime).

E.g. [this outdated FAQ](https://developer.apple.com/library/archive/qa/qa1419/_index.html) mentions 512KB as the default stack size for secondary threads in pthread.

jacob-carlborg:
Thanks 😊.

PR #2911, opened 2020-01-19
Fix issue 19902 - hasElaborateCopyConstructor doesn't know about copy…

pbackus:
… constructors

Uses the new `__traits` from dlang/dmd#10528.

Previously: #2709

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -292,12 +292,41 @@ template hasElaborateCopyConstructor(S)
     }
     else static if (is(S == struct))
     {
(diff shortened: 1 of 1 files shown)


TurkeyMan:
There's no negative test in this PR. Regressions that produce false-positives may be missed.

MoonlightSentinel:
FYI: Phobos forwards to this implementation and has some additional tests, see https://github.com/dlang/phobos/blob/79cc44b8f5c25ea8be6a254171b17f3606465c1a/std/traits.d#L3665-L3686

TurkeyMan:
👍

PR #2912, opened 2020-01-20
Add missing files to makefiles in 'mak/'

PetarKirov:
As reported by the following script:

```bash
#!/usr/bin/env bash

find_missing_headers() {
    for header_file in $(find src/core/ -type f -name "*.d"); do
        path="$(echo $header_file | sed -e 's,src/,,; s,/,\\\\,g')"
        for make_file in ./mak/{COPY,SRCS,WINDOWS}; do
            if ! grep -q "$path" "$make_file"
            then
                echo "$header_file";
                break;
            fi
        done
    done
}

find_missing_headers | sort
```

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -78,18 +78,21 @@ COPY=\
 	$(IMPDIR)\core\stdcpp\allocator.d \
 	$(IMPDIR)\core\stdcpp\array.d \
 	$(IMPDIR)\core\stdcpp\exception.d \
+	$(IMPDIR)\core\stdcpp\memory.d \
 	$(IMPDIR)\core\stdcpp\new_.d \
 	$(IMPDIR)\core\stdcpp\string.d \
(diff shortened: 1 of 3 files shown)


mak/COPY
```diff
@@ -438,4 +468,4 @@ COPY=\
 	$(IMPDIR)\core\thread\osthread.d \
 	$(IMPDIR)\core\thread\package.d \
 	\
-	$(IMPDIR)\etc\linux\memoryerror.d
+	$(IMPDIR)\etc\linux\memoryerror.d \
```
Geod24: This doesn't look correct?

PetarKirov: Thanks for noticing, fixed. (It had a missing newline escape and I apparently added where it wasn't necessary while trying to quickly amend my commit.)

PetarKirov:
Merging manually, as there are no available BuildKite agents currently.

PR #2915, opened 2020-01-28
Be sure to call pthread_attr_destroy

kubo39:


src/core/thread/osthread.d
```diff
@@ -820,6 +820,7 @@ class Thread
                     if ( pthread_create( &m_addr, &attr, &thread_entryPoint, cast(void*) this ) != 0 )
                         onThreadError( "Error creating thread" );
                 }
+                assert( pthread_attr_destroy( &attr ) == 0 );
```
thewilsonator: This will get removed when asserts are disabled. Please write it as `pthread_attr_destroy( &attr ) && assert(false);` or code to that effect.

kubo39: Ah, missed! Thanks.

src/core/thread/osthread.d
```diff
@@ -820,6 +820,7 @@ class Thread
                     if ( pthread_create( &m_addr, &attr, &thread_entryPoint, cast(void*) this ) != 0 )
                         onThreadError( "Error creating thread" );
                 }
+                pthread_attr_destroy( &attr ) && assert(false);
```
MoonlightSentinel: The remaining error handling uses `onThreadError`. Why use a plain `assert` (without a message) here?

kubo39: There's no particular reason and it should be. Thanks for pointing out!

PR #2919, opened 2020-01-31
Fix Issue 20550 - Use fixed seeds for treaps in GC

n8sh:
Discussion started from https://github.com/ldc-developers/druntime/pull/168.

### DIFF
--- a/src/rt/util/container/treap.d
+++ b/src/rt/util/container/treap.d
@@ -27,9 +27,10 @@ nothrow:
         removeAll();
     }
 
-    void initialize()
+    void initialize(ulong randSeed)
     {
-        rand48.defaultSeed();
+        Rand48 _rand48 = { randSeed };
+        rand48 = _rand48;
     }
 
     void insert(E element) @nogc
@@ -262,8 +263,8 @@ unittest
     OP[] ops;
     uint[] opdata;
 
-    treap.initialize();
     srand(cast(uint)time(null));
+    treap.initialize(rand());
 
     uint[] data;
 initialLoop:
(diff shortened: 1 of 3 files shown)


src/rt/util/container/treap.d
```diff
@@ -27,9 +27,10 @@ nothrow:
         removeAll();
     }
 
-    void initialize()
+    void initialize(ulong randSeed)
     {
-        rand48.defaultSeed();
+        Rand48 _rand48 = { randSeed };
+        rand48 = _rand48;
```
rainers: Doesn't `rand48 = Rand48(randSeed);` work here?

src/rt/util/container/treap.d
```diff
@@ -27,9 +27,10 @@ nothrow:
         removeAll();
     }
 
-    void initialize()
+    void initialize(ulong randSeed)
```
rainers: Would it be a problem to keep the existing `initialize()` method, but also allow passing the seed explicitly in an overload? I'm not expecting Treap to be used often, but having to come up with a magic number can be annoying.

MoonlightSentinel: Maybe define `rand()` as the default parameter value?

src/rt/util/container/treap.d
n8sh: no it doesn't because of the `opCall` override

rainers: ok (though it seems a hole in `@safe` if you are allowed to initialize any private members of a struct like this.) I'm not sure why the compiler tries to invoke a non-static opCall. Adding `this(ulong)` to `Rand48` allows to use the constructor syntax.

PR #2920, opened 2020-02-01
Add missing pthread_attr_init

kubo39:


thewilsonator:
This is also absent on `CRuntime_Bionic`, `CRuntime_Musl` and `CRuntime_UClibc`.

cc @Geod24 These in turn are identical modulo this missing `pthread_attr_init` to `FreeBSD`, `NetBSD`, `DragonFlyBSD`. Should we collapse them into a single version?

Geod24:
@kubo39 : Do you have a test-case for this ? Or an issue number ? How did you realize the call was missing ?

> Should we collapse them into a single version?

The point of having version specs is to just rule out system differences. If those blocks only use POSIX functions, we should be able to merge them without issue yes (save for implementation bugs).

> `pthread_getattr_np`

Erg, those are not POSIX functions (np stands for "not portable"). I wonder if we could refactor things to just keep the attribute around for the lifetime of the thread.

thewilsonator:
W.t.t issue number @kubo39 has added a changelog entry to #2922 

>  If those blocks only use POSIX functions, we should be able to merge them without issue yes.
> Erg, those are not POSIX functions (np stands for "not portable"). I wonder if we could refactor things to just keep the attribute around for the lifetime of the thread.

Good to know. Maybe? What else uses the thread's attributes for?

kubo39:
>  Do you have a test-case for this ? Or an issue number ? How did you realize the call was missing ?

No test-case or issue number, I noticed whie I read the code.
Perhaps we can write a test like

```d
// For glibc
//

// define pthread_attr, same as in glibc.
struct pthread_attr = ...;

pthread_attr iattr = cast(pthread_attr*) &attr;
assert(iattr.guardsize == PAGE_SIZE);
```

I'm not sure it's worthwhile.

PR #2922, opened 2020-02-02
Added missing pthread_attr_destroy one more

kubo39:


kubo39:
Ah, it should be because this would fix memory leak in some platform such as freebsd or solaris 10, though I haven't tested.

kubo39:
good to go, or needs rebase?

PR #2923, opened 2020-02-03
etc.linux.memoryerror: detect stack overflow if possible

kubo39:


kubo39:
I've forget about signal safety...

PR #2930, opened 2020-02-04
Update and fix SDKDDKVERs

LimiQS:
Correct and update SDKDDKVERs based on Windows Kits values.


LimiQS:
Values are from Windows Kits 10.0.18362.0 sdkddkvers.h.

PR #2931, opened 2020-02-04
Fix Issue 19502: Implement correct split behavior

LimiQS:
Implement own parser to fix incorrect command line split behavior.

### DIFF
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -25,17 +25,15 @@ private
 
 version (Windows)
 {
+    private import rt.winargv;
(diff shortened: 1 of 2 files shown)


LimiQS:
This PR directly introduced codes from [.Net Runtime](https://github.com/dotnet/runtime/blob/fcd862e06413a000f9cafa9d2f359226c60b9b42/src/coreclr/src/utilcode/util.cpp#L2799) which have correct command line split behavior to prevent inconsistent behavior and potential security flaw.

PR #2933, opened 2020-02-08
fix Issue 20567 - GC should not start threads for parallel marking in simple programs

rainers:
Avoid collection when there is no small/large pool to reuse anyway.
Don't start parallel threads during shutdown

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -1667,7 +1667,7 @@ struct Gcx
                     recoverNextPage(bin);
                 }
             }
-            else
+            else if (usedSmallPages > 0)
             {
                 fullcollect();
                 if (lowMem)
@@ -1752,7 +1752,7 @@ struct Gcx
                     minimize();
                 }
             }
-            else
+            else if (usedLargePages > 0)
             {
                 fullcollect();
                 minimize();
@@ -2601,7 +2601,12 @@ struct Gcx
             {
                 bool doParallel = config.parallel > 0;
                 if (doParallel && !scanThreadData)
-                    startScanThreads();
+                {
+                    if (nostack) // only used during shutdown, avoid starting threads for parallel marking
+                        doParallel = false;
+                    else
+                        startScanThreads();
+                }
             }
             else
                 enum doParallel = false;
--- a/test/gc/Makefile
+++ b/test/gc/Makefile
@@ -1,6 +1,7 @@
 include ../common.mak
 
-TESTS := sentinel printf memstomp invariant logging precise precisegc forkgc forkgc2 sigmaskgc startbackgc recoverfree
+TESTS := sentinel printf memstomp invariant logging precise precisegc forkgc forkgc2 sigmaskgc startbackgc \
+         recoverfree nocollect
 
 SRC_GC = ../../src/gc/impl/conservative/gc.d
 SRC = $(SRC_GC) ../../src/rt/lifetime.d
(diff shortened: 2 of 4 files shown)


rainers:
> Is there a way to add a test for this?

The first part "Avoid collection when there is no small/large pool" should be possible with a separate test executable. The shutdown behavior is not so easy to verify because, well, everything else is already shut down.

rainers:
Amended a test

src/gc/impl/conservative/gc.d
```diff
@@ -1667,7 +1673,7 @@ struct Gcx
                     recoverNextPage(bin);
                 }
             }
-            else
+            else if (usedSmallPages > 0)
```
schveiguy: I think usedSmallPages is not ever going to be 0 here.

Looking at the if statement above:

```D
if (!lowMem && (disabled || usedSmallPages < smallCollectThreshold))
```

Arguably, `lowMem` is going to be false (always is on 64-bit). If it's false, then the condition becomes `disabled || usedSmallPages < smallCollectThreshold`. If not disabled, and `usedSmallPages < smallCollectThreshold`, then 0 would be included in that range.

Really the only way we get here is if the entire memory space is allocated with large pages. Which should NEVER happen, at the very minimum you are going to have some thread-related allocations.

src/gc/impl/conservative/gc.d
```diff
@@ -1752,7 +1758,7 @@ struct Gcx
                     minimize();
                 }
             }
-            else
+            else if (usedLargePages > 0)
```
schveiguy: Same thing here, but it might be technically more likely to have this happen if you use up all your memory with small allocations. And only on 32-bit systems.

Really, I feel this extra check is not worth it. As this only is true potentially once.

src/gc/impl/conservative/gc.d
```diff
@@ -1346,6 +1346,12 @@ struct Gcx
         toscanPrecise.reset();
     }
 
+    __gshared bool inShutdown = false;
+    shared static ~this()
+    {
+        // detect shutdown to avoid starting threads for parallel marking
+        inShutdown = true;
+    }
```
schveiguy: This works, but why not just tell the GC it's being shut down here?

https://github.com/dlang/druntime/blob/0f9cd780759765241e57fdb0e594ae61798d7e46/src/gc/proxy.d#L120

src/gc/impl/conservative/gc.d
rainers: This is just meant as a detection whether any pool for small objects is allocated, i.e. it avoids running a collection (and starting the threads for parallel marking) for the very first allocation.

src/gc/impl/conservative/gc.d
rainers: Same for the very first pool for large objects.

src/gc/impl/conservative/gc.d
rainers: Indeed, that is a simpler check I didn't think of. I was also slightly worried about introducing module dependencies due to the static dtor. I'll change it.

src/gc/impl/conservative/gc.d
schveiguy: Right, but it won't get here. The very first allocation should go into the if branch.

In the first allocation, `lowMem` is false, disabled is false, and usedSmallPages is 0. `if(!false && (false || 0 < smallCollectThreshold))` should be true, so it will never get to this else if.

Or am I reading it wrong?

rainers: `smallCollectThreshold` is 0, too, see https://github.com/dlang/druntime/blob/3d44d2affda45e638ed52b0c9658435b3418d7f8/src/gc/impl/conservative/gc.d#L1261

schveiguy: My goodness. what the hell is it for then?

schveiguy: OK, I see it's based on some mechanism to tune the collection frequency based on used/allocated memory. I had thought it was a constant to prevent exactly what you were trying to prevent. So much going on in here!

PR #2934, opened 2020-02-08
Adjust minimal stack size for thread-local storage in Glibc

kubo39:
glibc's PTHREAD_MIN_STACK is incorrect, so I think adjustStackSize has still problem after https://github.com/dlang/druntime/pull/2904 merged.

### DIFF
--- a/src/core/thread/osthread.d
+++ b/src/core/thread/osthread.d
@@ -3187,16 +3187,18 @@ private size_t adjustStackSize(size_t sz) nothrow @nogc
     if (sz == 0)
         return 0;
 
+    if (PTHREAD_STACK_MIN > sz)
+        sz = PTHREAD_STACK_MIN;
+
     version (CRuntime_Glibc)
     {
         // TLS uses the top of the stack, so add its size to the requested size
         sz += externDFunc!("rt.sections_elf_shared.sizeOfTLS",
                            size_t function() @nogc nothrow)();
     }
+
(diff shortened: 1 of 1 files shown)


Geod24:
What do you mean by "incorrect" ?

kubo39:
> What do you mean by "incorrect" ?

Ah, sorry for confusing. The word `incorrect` is mistake.
 I mean, PTHREAD_STACK_MIN is not enough when a small stack (in case, < `sz`) is created an application with large thread-local storage.

src/core/thread/osthread.d
```diff
@@ -3182,23 +3182,42 @@ private void onThreadError(string msg) nothrow @nogc
...
+        version (CRuntime_Glibc)
+        {
+            // TLS uses the top of the stack, so add its size to the requested size
+            sz += externDFunc!("rt.sections_elf_shared.sizeOfTLS",
+                               size_t function() @nogc nothrow)();
+        }
+        // stack size must be a multiple of PAGESIZE and at least PTHREAD_STACK_MIN
+        sz = ((sz + PAGESIZE - 1) & ~(PAGESIZE - 1));
+
+        size_t minStackSize = minimalStackSize();
```
rainers: Instead of evaluating sizeOfTLS size twice, it should be simpler and more efficient to save it to a variable and adjust `sz` and `minStackSize` accordingly inside this function.

kubo39: Thank you for notice that. How stupid I was...

PR #2937, opened 2020-02-11
Update winnt.d

etcimon:
Add more symbols


src/core/sys/windows/winnt.d
```diff
@@ -2397,6 +2397,7 @@ struct TOKEN_OWNER {
     PSID Owner;
 }
 alias TOKEN_OWNER* PTOKEN_OWNER;
+const int SECURITY_MAX_SID_SIZE = 68;
```
MoonlightSentinel: Constants should be `enum`s.

MoonlightSentinel: Also this should use the appropriate type alias instead of `int`
(ditto for the other changes)

src/core/sys/windows/winnt.d
```diff
@@ -2451,27 +2456,65 @@ struct SECURITY_DESCRIPTOR {
...
-    TokenStatistics,
-    TokenRestrictedSids,
-    TokenSessionId,
-    TokenGroupsAndPrivileges,
-    TokenSessionReference,
-    TokenSandBoxInert,
-    TokenAuditPolicy,
-    TokenOrigin
+alias TOKEN_ELEVATION_TYPE = int;
+enum : TOKEN_ELEVATION_TYPE {
```
MoonlightSentinel: Making this an anonymous enum breaks existing code.... Why?

src/core/sys/windows/winnt.d
```diff
@@ -3734,6 +3875,7 @@ enum DWORD
     ES_SYSTEM_REQUIRED  = 0x00000001,
     ES_DISPLAY_REQUIRED = 0x00000002,
     ES_USER_PRESENT     = 0x00000004,
+	ES_AWAYMODE_REQUIRED = 0x00000040,
```
MoonlightSentinel: Fix intendation

src/core/sys/windows/winnt.d
etcimon: This is a constant: https://docs.microsoft.com/en-us/windows/win32/secbiomet/general-constants

src/core/sys/windows/winnt.d
etcimon: I was migrating from my own windows headers library and didn't know about this difference between them, thanks!

src/core/sys/windows/winnt.d
MoonlightSentinel: `enum` can be used to declare compile time constants
```suggestion
enum int SECURITY_MAX_SID_SIZE = 68;
```

PR #2943, opened 2020-02-16
Fix Issue 18011 - core.sys.condition cannot be used as shared

tom-tan:
It is to solve [Issue 18011](https://issues.dlang.org/show_bug.cgi?id=18011) by adding shared methods to `core.sync.condition.Condition`.
I also added a shared version of `Mutex#handleAddr` because it is used in `Condition#wait()` method.

### DIFF
--- a/src/core/sync/condition.d
+++ b/src/core/sync/condition.d
@@ -119,7 +119,7 @@ class Condition
                     rc = pthread_condattr_setclock( &attr, CLOCK_MONOTONIC );
                     if ( rc )
                         throw new SyncError( "Unable to initialize condition" );
-                    rc = pthread_cond_init( cast(pthread_cond_t*)&m_hndl, &attr );
+                    rc = pthread_cond_init( cast(pthread_cond_t*) &m_hndl, &attr );
                     if ( rc )
                         throw new SyncError( "Unable to initialize condition" );
                     rc = pthread_condattr_destroy( &attr );
--- a/src/core/sync/condition.d
+++ b/src/core/sync/condition.d
@@ -411,10 +411,10 @@ private:
             }
             else
             {
-                auto op(string op, T, V1)(ref shared T val, V1 mod)
-                {            
+                auto op(string o, T, V1)(ref shared T val, V1 mod)
+                {
                     import core.atomic: atomicOp;
-                    return atomicOp!op(val, mod);
+                    return atomicOp!o(val, mod);
                 }
             }
 
--- a/src/core/sync/condition.d
+++ b/src/core/sync/condition.d
@@ -404,9 +404,9 @@ private:
         {
             static if (is(Q == Condition))
             {
-                auto op(string op, T, V1)(ref T val, V1 mod)
+                auto op(string o, T, V1)(ref T val, V1 mod)
                 {
-                    return mixin("val " ~ op ~ "mod");
+                    return mixin("val " ~ o ~ "mod");
                 }
             }
             else
--- a/src/core/sync/condition.d
+++ b/src/core/sync/condition.d
@@ -92,12 +92,20 @@ class Condition
     {
         version (Windows)
         {
-            m_blockLock = cast() CreateSemaphoreA( null, 1, 1, null );
+            static if (is(Q == Condition))
+            {
+                alias HANDLE_BASE = void*;
+            }
+            else
+            {
+                alias HANDLE_BASE = shared(void*);
+            }
+            m_blockLock = cast(HANDLE_BASE) CreateSemaphoreA( null, 1, 1, null );
             if ( m_blockLock == m_blockLock.init )
                 throw new SyncError( "Unable to initialize condition" );
             scope(failure) CloseHandle( cast(void*) m_blockLock );
 
-            m_blockQueue = cast() CreateSemaphoreA( null, 0, int.max, null );
+            m_blockQueue = cast(HANDLE_BASE) CreateSemaphoreA( null, 0, int.max, null );
             if ( m_blockQueue == m_blockQueue.init )
                 throw new SyncError( "Unable to initialize condition" );
             scope(failure) CloseHandle( cast(void*) m_blockQueue );
--- a/src/core/sync/condition.d
+++ b/src/core/sync/condition.d
@@ -701,3 +701,169 @@ unittest
     testNotifyAll();
     testWaitTimeout();
 }
+
+unittest
+{
+    import core.thread;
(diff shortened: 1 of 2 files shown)


thewilsonator:
Please make one of the commit messages say "Fix issue 18011: ..." so that the bot picks it up.

kubo39:
@tom-tan 
These links could be helpful.
https://wiki.dlang.org/Starting_as_a_Contributor#How_to_choose_a_title_for_your_Pull_Request
https://wiki.dlang.org/Starting_as_a_Contributor#Squashing

CI failed due to style check, please make sure `make -f posix.mak style` success.

kubo39:
This seems reasonable for me, but I'm not sure why these aren't marked as shared.
cc: @PetarKirov, because you are assigned at https://issues.dlang.org/show_bug.cgi?id=18011

src/core/sync/condition.d
```diff
@@ -176,14 +201,27 @@ class Condition
      *  SyncError on error.
      */
     void wait()
+    {
+        wait_();
+    }
+
+    /// ditto
+    void wait() shared
```
MoonlightSentinel: This overload could simply cast away shared and Call the existing wait method.
There is no need for a template.

Ditto for the other methods

tom-tan: I updated to use casts instead of overloads except constructors.

tom-tan:
Thank you for valuable comments!

I update the title of this PR and the commit message.
I will add a test for shared later.

kubo39:
L132: also needs to add `cast(pthread_cond_t*)`.

src/core/sync/condition.d
```diff
@@ -75,6 +75,20 @@ class Condition
...
+    }
+
+    /// ditto
+    this( shared Mutex m ) shared nothrow @safe
+    {
+        this(m, true);
+    }
+
+    //
+    private this(this Q, M)( M m, bool _unused_ ) nothrow @safe
```
MoonlightSentinel: This could also be a plain method instead as it does not initialize const/immutable data AFAICT.

src/core/sync/condition.d
```diff
@@ -105,7 +119,7 @@ class Condition
                     rc = pthread_condattr_setclock( &attr, CLOCK_MONOTONIC );
                     if ( rc )
                         throw new SyncError( "Unable to initialize condition" );
-                    rc = pthread_cond_init( &m_hndl, &attr );
+                    rc = pthread_cond_init( cast(pthread_cond_t*)&m_hndl, &attr );
```
MoonlightSentinel: DStyle requires a space after a cast:

```suggestion
                    rc = pthread_cond_init( cast(pthread_cond_t*) &m_hndl, &attr );
```

Ditto below

src/core/sync/mutex.d
```diff
@@ -296,6 +296,11 @@ package:
         {
             return &m_hndl;
         }
+
+        shared(pthread_mutex_t*) handleAddr() shared
```
MoonlightSentinel: This seems unecessary?

src/core/sync/condition.d
tom-tan: It may cause a breaking change in ABI level.
In the case of using a plain method, we have to use an unsafe cast in `this(shared Mutex)` and have to change the attribute of `this(shared Mutex)` from `@safe` to `@trusted` but it changes ABI of the constructor (See [Type Mangling](https://dlang.org/spec/abi.html#type_mangling)).

But I am not sure it really happens because I am not familiar with low level programming.

src/core/sync/mutex.d
PetarKirov: Actually, does the non-`shared` `handleAddr` need to exist? It's marked as package, so I see no problem with also making it as `final nothrow @nogc`.

src/core/sync/condition.d
PetarKirov: @MoonlightSentinel Templating a member function on the type qualifier of `this` makes it possible to reuse code between `shared` and non-`shared` code without casts. This pattern is also used in `core.sync.mutex`.

MoonlightSentinel: > It may cause a breaking change in ABI level.

Not really, it's a new symbol.

> In the case of using a plain method, we have to use an unsafe cast in `this(shared Mutex)` and have to change the attribute of `this(shared Mutex)` from `@safe` to `@trusted` [...]

You can always use a `@trusted` anonymous lamdba

MoonlightSentinel: > @MoonlightSentinel Templating a member function on the type qualifier of `this` makes it possible to reuse code between `shared` and non-`shared` code without casts. This pattern is also used in `core.sync.mutex`.

Agreed, I was just thinking whether we need to have this kinda akward function header and  generate this constructor twice. 

But anyway, I'm fine if  this remains unchanged

src/core/sync/mutex.d
tom-tan: If we fix this by using overloading such as `Mutex`, both versions are needed because they are called from `shared` and non-`shared` `wait()`, respectively.
If we fix it by using casting as @MoonlightSentinel suggested, only non-`shared` version is needed.

Which solution is more preferable?

tom-tan:
> L132: also needs to add cast(pthread_cond_t*).

Fixed!

I also added a unittest. It is almost the same as non-`shared` version but uses `shared Mutex` and `shared Condition` (I need comments for it because I am not good at low layer concurrent programming).

src/core/sync/mutex.d
tom-tan: I will remove it because all the use of `handleAddr` is to be passed to `pthread_cond_*` that requires `pthread_mutex_t*`.
Casting `Condition#m_assocMutex` to `Mutex` seems better rather than providing an overload of `handleAddr` for `shared Mutex` because it is finally casted to non-`shared`.

tom-tan:
One CI `buildkite/druntime` fails but this failure is not related to this request.

kubo39:
> One CI tester buildkite/druntime fails but this failure is not related to this request.

Yes, this was already fixed in master at https://github.com/dlang/druntime/commit/5cd15b2dc4d3c2678d27aae8ab0f3900f9b93c1, but not merged into stable yet.

tom-tan:
Rebased!

tom-tan:
Now it passes the all CI tests!

tom-tan:
Thanks!

PetarKirov:
Thanks @tom-tan, nice work!

PR #2944, opened 2020-02-16
runbench: add -mcpu=native by default

kubo39:
This would be useful for benchmarking simd optimization.


kubo39:
CI failure seems unrelated.

PR #2946, opened 2020-02-21
allow linking with MinGW libs if phobos is compiled with VS 2017+

rainers:
implement weak symbol __stdio_common_vsprintf emitted by VC when compiling "snprintf" in phobos/etc/c/zlib/gzlib.c
detect msvcrt110 and msvcrt120
add test target to skip C++ tests and uuid.lib-test.

needed by https://github.com/dlang/dmd/pull/10814

### DIFF
--- a/src/rt/msvc.c
+++ b/src/rt/msvc.c
@@ -33,6 +33,8 @@ FILE* __iob_func();                 // VS2013-
 
 int _set_output_format(int format); // VS2013-
 
+int _vsnprintf(char* buffer, int buffer_length, const char* format, void* va_list); // VS2015-
+
 //extern const char* __acrt_iob_func;
 extern const char* _nullfunc = 0;
 
@@ -40,8 +42,10 @@ unsigned char msvcUsesUCRT;
 
 #if defined _M_IX86
     #define C_PREFIX "_"
+    typedef long size_t;
 #elif defined _M_X64 || defined _M_ARM || defined _M_ARM64
     #define C_PREFIX ""
+    typedef long long size_t;
 #else
     #error Unsupported architecture
 #endif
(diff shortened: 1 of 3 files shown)


src/rt/msvc.c
```diff
@@ -76,6 +80,22 @@ void init_msvc()
...
+int _msvc_stdio_common_vsprintf(
+    unsigned __int64 const options,
+    char* const buffer,
+    size_t const buffer_count,
+    char const* const format,
+    void* const locale,
+    void* const arglist
+)
+{
+    return _vsnprintf(buffer, buffer_count, format, arglist);
```
kinke: Nice. `_vsnprintf` probably isn't 100% compatible though, I guess the expected behavior is standard-compliant `vsnprintf` (always null-terminates, and there's a difference wrt. return value too IIRC).

kinke: This should mimic vsnprintf semantics:
```D
const r = _vsnprintf(buffer, buffer_count, format, arglist);
if (buffer_count != 0)
{
    if (r == buffer_count)
    {
        buffer[buffer_count - 1] = 0;
        return buffer_count;
    }
    if (r == -1)
    {
        buffer[buffer_count - 1] = 0;
        return _vsnprintf(null, 0, format, arglist);
    }
}
return r;
```

rainers: The expected behavior is encoded into the Options parameter, not sure we want to support them all. Does `_vsnprintf(null, 0,...)` work with any version of the runtime?

kinke: If it's just these 2 behaviors - legacy _vsnprintf and standard vsnprintf - then I think we should support them both. - I'm pretty sure `_vsnprintf(null, 0, ...)` has always worked, but the MSDN docs for VS < 2015 are dead.

kinke: Slightly extending the condition above should be enough to cover both behaviors, according to my WinSDK headers:
```D
enum _CRT_INTERNAL_PRINTF_STANDARD_SNPRINTF_BEHAVIOR = 2;
if ((options & _CRT_INTERNAL_PRINTF_STANDARD_SNPRINTF_BEHAVIOR) && buffer_count && buffer) ...
```

[My only worry is wrt. user C(++) libs referencing `__stdio_common_vsprintf` too, i.e., Phobos' zlib potentially not being the only user of this default implementation.]

rainers: As mentioned earlier, I think you should not use the MinGW libraries when using other libraries built with a C/C++ compiler. You will very likely run into other incompatibilities anyway.

I'll incorporate your suggested changes, but the respective tests in https://github.com/dlang/dmd/pull/10814 are suddenly broken for me locally (LDC not finding vcruntime.lib).

kinke: > I think you should not use the MinGW libraries when using other libraries built with a C/C++ compiler

Yes, and both of us know that, but most other people probably don't. Just in case all they'd need is `__stdio_common_vsprintf`, this extra logic should decrease the risk of strange runtime failures.

Wrt. the dlang PR, LDC doesn't officially support VC12 anymore, and adds `vcruntime.lib` implicitly. I'd recommend keeping a statically linked dmd.exe, not depending on bundled msvcr120.dll.

rainers: It's supposed to link with the regular VS2017 installation, but I now noticed that changing the LIB variable breaks it (but has to be done to run the tests themselves with lld/mingw-libs). I can reorder compilation, though.

PR #2947, opened 2020-02-22
fix printf format

WalterBright:
Blocking https://github.com/dlang/dmd/pull/10812

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -4087,7 +4087,7 @@ debug (LOGGING)
             printf("    p = %p, size = %lld, parent = %p ", p, cast(ulong)size, parent);
             if (file)
             {
-                printf("%s(%u)", file, line);
+                printf("%s(%u)", file, cast(uint)line);
             }
             printf("\n");
         }
--- a/test/exceptions/src/line_trace.d
+++ b/test/exceptions/src/line_trace.d
@@ -8,7 +8,7 @@ void main()
     {
         import core.stdc.stdio;
         auto str = e.toString();
-        printf("%.*s\n", str.length, str.ptr);
+        printf("%.*s\n", cast(int)str.length, str.ptr);
     }
 }
 


src/gc/impl/conservative/gc.d
```diff
@@ -4087,7 +4087,7 @@ debug (LOGGING)
             printf("    p = %p, size = %lld, parent = %p ", p, cast(ulong)size, parent);
             if (file)
             {
-                printf("%s(%u)", file, line);
+                printf("%s(%zu)", file, line);
```
rainers: %z doesn't work with the older VS runtimes (before VS2015 IIRC).

WalterBright: We're kinda hosed, then. Why are we supporting 5 year old VS?

WalterBright: The `z` format was standardized in 1999, 21 years ago.

rainers: %lld and a cast to long is usually used. 

> Why are we supporting 5 year old VS?

mostly because the autotester uses VS2013 and the release process uses VS2010 (hopefully updated soon, but it seems Martin doesn't want to go further than VS2015).

WalterBright: @MartinNowak what do you think?

WalterBright: (Note that I added hundreds of `zd` and `td` formats to the test suite to resolve those errors.)

rainers: Maybe the compiler should not be too pedantic: it could only warn about uint and size_t mismatches if it actually makes a difference in the ABI: for example with x64, ints are passed as 64-bit values in registers or on the stack, too. Not sure if the higher bits are allowed to contain garbage, though.

WalterBright: It makes a difference with 32 <=> 64 bit compiles.

WalterBright:
I changed this one to use the casting method instead of `z`, just to move things forward.

src/gc/impl/conservative/gc.d
rainers: No idea what you mean. In 32-bit processes there is no size-difference between int and size_t, and in 64-bit processes, single integer arguments are always passed by occupying 64-bit.

WalterBright:
The DAutoTest and buildkite failures are not relevant to this PR.

PR #2948, opened 2020-02-22
Disable FPU fast mode for Apple ARM in test runner

jacob-carlborg:
This is required for the some of the Phobos math related tests to pass.

I'm not sure if this applies to other ARM/AArch64 platforms as well.

### DIFF
--- a/src/test_runner.d
+++ b/src/test_runner.d
@@ -10,6 +10,10 @@ ModuleInfo* getModuleInfo(string name)
 
 UnitTestResult tester()
 {
+    disableFPUFastMode();
+    scope (exit)
+        restoreFPUMode();
+
     return Runtime.args.length > 1 ? testModules() : testAll();
 }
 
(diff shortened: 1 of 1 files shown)


jacob-carlborg:
@kinke

src/test_runner.d
```diff
@@ -96,3 +100,72 @@ shared static this()
...
+[25] DN Default NaN mode enable bit:
+0 = default NaN mode disabled 1 = default NaN mode enabled.
+[24] FZ Flush-to-zero mode enable bit:
+0 = flush-to-zero mode disabled 1 = flush-to-zero mode enabled.
+*/
+void disableFPUFastMode()
+{
+    version (D_HardFloat)
+    {
+        import ldc.llvmasm;
```
thewilsonator: `version(LDC)`?

src/test_runner.d
```diff
@@ -96,3 +100,72 @@ shared static this()
...
+
+        cast(void)__asm!uint(code, "=r");
+    }
+}
+
+void restoreFPUMode()
+{
+    version (D_HardFloat)
+    {
+        import ldc.llvmasm;
```
thewilsonator: Ditto.

src/test_runner.d
jacob-carlborg: Yes, good point.

jacob-carlborg: Done.

src/test_runner.d
jacob-carlborg: Done.

PR #2949, opened 2020-02-22
Adds c_long and c_ulong for bare-metal ARM 32-bit platform

denizzzka:
Based on information from http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0041c/ch03s02s02.html

### DIFF
--- a/src/core/stdc/config.d
+++ b/src/core/stdc/config.d
@@ -144,6 +144,11 @@ else version (Posix)
     alias ulong cpp_ulonglong;
   }
 }
+else version (ARM)
+{
+    alias c_long = int;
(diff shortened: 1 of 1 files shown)


denizzzka:
Is such a PR appropriate or is our goal is to create a separated runtime for bare metal?

denizzzka:
Okay, too global a task if think more deeply.

ibuclaw:
There is version FreeStanding that could be used.

denizzzka:
> There is version FreeStanding that could be used.

Thanks! This is what I was looking for!

PR #2951, opened 2020-02-23
more printf format fixes

WalterBright:


WalterBright:
Blocking https://github.com/dlang/dmd/pull/10812

PR #2953, opened 2020-02-24
Fix Issue 20368 - dmd 2.089.0 Error: expression `main` is `void` and has no value

ErnyTech:
Rebase of #2861 with addition of unittest


RazvanN7:
closing in favor of #3786 since that one is also linked to the bugzilla issue.

PR #2954, opened 2020-02-24
Fix typo in atomic.ValidateStruct

don-clugston-sociomantic:
This template couldn't possibly compile, so it has clearly never been instantiated.
Fix the typo and document the trick which is being used.

The typo is `size_t *2` when clearly `size_t.sizeof*2` was meant.

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -687,7 +687,10 @@ private
     {
         import core.internal.traits : hasElaborateAssign;
 
-        static assert (S.sizeof <= size_t*2 && (S.sizeof & (S.sizeof - 1)) == 0, S.stringof ~ " has invalid size for atomic operations.");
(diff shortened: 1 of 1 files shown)


don-clugston-sociomantic:
> Please add a unittest which instantiates that template

Sorry, that's out of scope. Not going to do that. End of story.

thewilsonator:
cc @TurkeyMan this isn't used. Also there are a typos in `IntForStruct` (which is the only user of this. It is unused) all those `IntForFloat` should be `IntForStruct`.

TurkeyMan:
Haha, yeah there's a whole lot wrong with that code. What's funny is, the assert attempts to prove a thing that's already established in the instantiating context; the case where it attempts instantiation (not power-of-2 and `<= size_t.sizeof*2`) already proves the static assert false ;)

PR #2955, opened 2020-02-25
In preparation for #10828 don't define size_t if it is a type already

UplinkCoder:


UplinkCoder:
@Geod24 why did you remove the auto-merge? 
This is as non intrusive as it gets.

Geod24:
@thewilsonator : Something something don't put auto-merge so quickly for the 10,000,000 time.

dlang/dmd#10828 raised some concerns and it would be good to have a resolution for those before we go ahead with merging code elsewhere.
Otherwise we might end up in a situation where we eagerly merge code in a repository in support for a PR which is eventually abandoned.

wilzbach:
Thanks a lot for catching this one in time @Geod24!

UplinkCoder:
with the dmd PR in question and this specific change chances of failing
would have been rare.

Geod24:
Guess we don't need this one either @UplinkCoder ?

UplinkCoder:
The PR which spawned this one is one ice for now.

PR #2957, opened 2020-02-26
Add sched_getcpu

kubo39:


src/core/sys/linux/sched.d
```diff
@@ -110,6 +110,12 @@ int sched_getaffinity(pid_t pid, size_t cpusetsize, cpu_set_t *mask);
 int clone(int function(void*), void* child_stack, int flags, void* arg, ...);
 int unshare(int flags) @trusted;
 
+version (CRuntime_Glibc)
```
n8sh: Why does `sched_getcpu` need this version guard but not `sched_getaffinity` or `sched_setaffinity`?

kubo39: Bacause these  functions are completely different.
[sched_getcpu](http://man7.org/linux/man-pages/man3/sched_getcpu.3.html) is a glibc-specific library function, and [sched_getaffinity and sched_setaffinity](http://man7.org/linux/man-pages/man2/sched_setaffinity.2.html) are Linux-specific system calls.

PR #2958, opened 2020-02-26
druntime: split stdc and stdcpp from main source

denizzzka:
Splits `core.stdc` and `core.stdcpp` modules from main source while compiling.
Adds `DRUNTIME_NOSTDC` and `DRUNTIME_NOSTDC` flags to druntime Makefiles to convient work with `libc` dependency extraction and removal.

Rationale: druntime should support Posix, but it can't be forced to use Posix for all platforms.

Discussion: https://forum.dlang.org/post/jvjphuovogxcgglvwqak@forum.dlang.org

**Actually doesn't removes `libc` dependency!**
If you try to build with `DRUNTIME_NOSTDC=1` currently you got tons of erros like:
```
../druntime/import/core/time.d(69): Error: module time is in file 'core/stdc/time.d' which cannot be read
import path[0] = ../druntime/import
../druntime/import/core/time.d(70): Error: module stdio is in file 'core/stdc/stdio.d' which cannot be read
import path[0] = ../druntime/import
```


posix.mak
```diff
@@ -127,7 +127,27 @@ SRCS:=$(subst \,/,$(SRCS))
 # NOTE: a pre-compiled minit.obj has been provided in dmd for Win32	 and
 #       minit.asm is not used by dmd for Linux
 
-OBJS= $(ROOT)/errno_c.o $(ROOT)/threadasm.o
+OBJS= $(ROOT)/threadasm.o
+
+# checks if linking against stdc not disabled
+ifndef DRUNTIME_NOSTDC
+include mak/STDC
```
Geod24: Since it affects all files in two packages, wouldn't it be better to use [filter-out](https://www.gnu.org/software/make/manual/html_node/Text-Functions.html) ?

denizzzka: I don't think that this is very important. My option will help to easily move these packages to another location at future .

denizzzka: And I saw that filter-out is already used. I'll redo it now.

denizzzka:
weak idea, is need another approach

PR #2959, opened 2020-02-26
Unused stdc imports removed

denizzzka:


denizzzka:
huh

schveiguy:
You just need to move the imports into the unittests that need them. (at least for printf).  Not sure that this module doesn't actually need `core.stdc.time`.

denizzzka:
> Not sure that this module doesn't actually need core.stdc.time.

I will try to at least wrap into version() branches.
(But for now I'm fighting for selective build of files from druntime tree.)

PR #2962, opened 2020-02-27
Fix 20613 - String switch in -betterC fails for 7+ labels

schveiguy:
Even though it's not actually used at runtime, the `idup` template in object.d uses druntime features not available in betterC. This makes the compiler throw an error, even though that code is never used. That is a separate issue ([19268](https://issues.dlang.org/show_bug.cgi?id=19268)).

This works around the issue and also simplifies the translation between const items and immutable items with a cast instead of an idup. It should be much less CTFE churn as well.

Also added a betterC test for the issue.

### DIFF
--- a/src/core/internal/switch_.d
+++ b/src/core/internal/switch_.d
@@ -61,12 +61,13 @@ int __switch(T, caseLabels...)(/*in*/ const scope T[] condition) pure nothrow @s
     {
         // Need immutable array to be accessible in pure code, but case labels are
         // currently coerced to the switch condition type (e.g. const(char)[]).
-        static immutable T[][caseLabels.length] cases = {
-            auto res = new immutable(T)[][](caseLabels.length);
-            foreach (i, s; caseLabels)
-                res[i] = s.idup;
-            return res;
-        }();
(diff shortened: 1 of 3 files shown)


wilzbach:
FYI for future purposes you could have added a `@betterC` unittest:

```
@betterC unittest {

}
```

Might make your life easier.

(it's currently broken for Phobos, but it's working here for druntime).

schveiguy:
> FYI for future purposes you could have added a `@betterC` unittest:

How does that work? I didn't think conditional compilation could be done via attributes.

schveiguy:
buildkite seemed to have a spurious failure, so I just merged.

wilzbach:
Yes, but we build our own support:

https://github.com/dlang/druntime/blob/7ffd1dc14cf01c3586c83f442141579b9091bb71/posix.mak#L415

> And technically, it wasn't that hard to add another file to the betterC test directory.

Well, it's nice to keep implementation and tests together (or even expose the test publicly).

schveiguy:
OK, I will try to remember next time.

PR #2963, opened 2020-02-27
core.memory: Fix mangle name of core.memory.initialize

ibuclaw:
It doesn't start with a valid character for Assembler/C symbols.

```
/tmp/cc3PpJ3O.s: Assembler messages:
/tmp/cc3PpJ3O.s:6: Error: expected symbol name
/tmp/cc3PpJ3O.s:7: Error: Missing symbol name in directive
/tmp/cc3PpJ3O.s:7: Error: unrecognised symbol type "4"
/tmp/cc3PpJ3O.s:7: Error: junk at end of line, first unrecognised character is `c'
/tmp/cc3PpJ3O.s:8: Error: junk at end of line, first unrecognised character is `4'
/tmp/cc3PpJ3O.s:27: Error: expected comma after name `' in .size directive
/tmp/cc3PpJ3O.s:30: Error: junk at end of line, first unrecognised character is `c'
```

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -227,7 +227,7 @@ unittest
 // make it more difficult to call the function again, manually.
 private void initialize();
 pragma(crt_constructor)
-pragma(mangle, initialize.mangleof)
+pragma(mangle, `_D` ~ initialize.mangleof)
 private extern (C) void initialize() @system
 {
     version (Posix)


thewilsonator:
Auto-merge toggled on

Geod24:
Why wasn't this caught by the tests ?

thewilsonator:
I assume because we (i.e. not GDC) don't use an assember as an intermediate step.

jacob-carlborg:
There’s a another bug here that’s causing this issue. If there are two function declarations with the same mangling, `mangleof` breaks.

Two function declarations:

```d
void foo();
void foo();
static assert(foo.mangleof == “4main3foo”);
```

One function declaration:

```d
void foo();
static assert(foo.mangleof == “_D4main3fooFZv”);
```

Geod24:
Funnily it's a bug I'm working on at the moment. It is at least "documented":
https://github.com/dlang/dmd/blob/ea2280fd9d7048f595d718e080fd0b8a5ffee55c/src/dmd/dmangle.d#L619-L641

PR #2964, opened 2020-02-28
update emplaceRef API

9il:


src/core/internal/lifetime.d
```diff
@@ -10,69 +10,74 @@ build.
 Furthermore, emplaceRef optionally takes a type parameter, which specifies
 the type we want to build. This helps to build qualified objects on mutable
 buffer, without breaking the type system with unsafe casts.
+
+Note: The function is used in mir-core.
```
Geod24: How come ? It is in `internal`, and *explicitly* defined as "for druntime internal use"

9il: This is why the note is here. It is a really useful API function for ref-counted libraries. It is used for a long time, so I just removed mir-core copy of the DRuntime code and start use these API directly.

9il: This MR is the next step required to remove the mir-core definitions.

atilaneves: If it's a useful API for ref-counted libraries, shouldn't it be in Phobos instead? Or, better yet, code.dlang.org?

In any case, why would druntime have a comment about one of its functions being used in a 3rd party library?

9il: nothing has been changed in the community

atilaneves: > nothing has been changed in the community

I don't know what you mean by that.

PR #2966, opened 2020-03-01
Revive 1989: Thread detach stability

MoonlightSentinel:
Rebases #1989. The original PR was already approved but got stalled because of a TLS issue AFAICT,

Let's see whether this issue was resolved since then.


MoonlightSentinel:
Well, seems like there are some remaining issues

PR #2968, opened 2020-03-02
Fix fprintf arg types

Luhrel:
Blocking https://github.com/dlang/dmd/pull/10852

### DIFF
--- a/src/rt/profilegc.d
+++ b/src/rt/profilegc.d
@@ -159,12 +159,12 @@ shared static ~this()
             {
                 fprintf(fp, "%15llu\t%15llu\t%8.*s\n",
                     cast(ulong)c.entry.size, cast(ulong)c.entry.count,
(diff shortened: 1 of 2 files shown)


src/rt/profilegc.d
```diff
@@ -159,12 +159,12 @@ shared static ~this()
             {
                 fprintf(fp, "%15llu\t%15llu\t%8.*s\n",
                     cast(ulong)c.entry.size, cast(ulong)c.entry.count,
-                    c.name.length, c.name.ptr);
+                    cast(int)c.name.length, c.name.ptr);
```
MoonlightSentinel: DStyle requires a space after a cast:

```suggestion
                    cast(int) c.name.length, c.name.ptr);
```

Luhrel: Done.

PR #2969, opened 2020-03-03
rt/trace: Restore `extern(D)` default, apply `extern(C)` selectively

Geod24:
The whole module was `extern(C)` and the code seems very old.
It would probably benefit from some love, but at the moment I'm just cleaning up for the fact that the module ctors/dtors shouldn't be mangled as `extern(C)`.

See https://issues.dlang.org/show_bug.cgi?id=20627

### DIFF
--- a/src/rt/trace.d
+++ b/src/rt/trace.d
@@ -29,8 +29,6 @@ shared static this ()
     trace_deffilename = strdup(DefaultDef.ptr)[0 .. DefaultDef.length + 1];
 }
 
-extern (C):
-
 /**
  * Set the file path for profile reports (`-profile`)
  *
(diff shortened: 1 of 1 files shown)


Geod24:
> /d/a/1/dmd/.azure-pipelines/lib.sh: line 37: HOST_DMD_VERSION: unbound variable

Same issue as in some other PR. Is something up with `Windows_VisualD win32-ldc`. @rainers ?

rainers:
> Same issue as in some other PR. Is something up with `Windows_VisualD win32-ldc`.

Oops, sorry. Some change since the latest merge from stable. This should fix it: https://github.com/dlang/druntime/pull/2970

Geod24:
Thanks for the quick fix @rainers !
Rebased on master.

RazvanN7:
It seems that this change is making windows unhappy :-?

Geod24:
Yep, there seems to be some magic happening in the backend. Will get back to it soon.

Geod24:
Rebased on `master`, much smaller diff, and this works now.

PR #2970, opened 2020-03-03
fix azure pipeline

rainers:
building with LDC now needs DMD as a host compiler for tools, too


Geod24:
I guess Phobos has the same issue ?
https://github.com/dlang/phobos/blob/d19ff00ff314d9997b2b1a87f9b88e395e8dcf29/azure-pipelines.yml#L68-L69

rainers:
> I guess Phobos has the same issue ?

https://github.com/dlang/phobos/pull/7412

thewilsonator:
Could you cherrypick this on to stable?

PR #2972, opened 2020-03-04
Fix printf arg. types 2

Luhrel:
Blocking https://github.com/dlang/dmd/pull/10872

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -88,7 +88,15 @@ private string miniFormat(V)(const ref V v)
     else static if (__traits(isFloating, V))
     {
         char[60] val;
-        const len = sprintf(&val[0], "%g", v);
+        int len;
+        static if (is(V == cfloat) || is(V == cdouble))
+            len = sprintf(&val[0], "%g + %gi", v.re, v.im);
+        else static if (is(V == creal))
+            len = sprintf(&val[0], "%Lg + %Lgi", v.re, v.im);
+        else static if (is(V == real) || is(V == ireal))
+            len = sprintf(&val[0], "%Lg", v);
+        else
+            len = sprintf(&val[0], "%g", v);
         return val.idup[0 .. len];
     }
     // special-handling for void-arrays
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -4023,7 +4023,7 @@ debug(PRINTF_TO_FILE)
                 gcStartTick = MonoTime.currTime;
             immutable timeElapsed = MonoTime.currTime - gcStartTick;
             immutable secondsAsDouble = timeElapsed.total!"hnsecs" / cast(double)convert!("seconds", "hnsecs")(1);
-            len = fprintf(gcx_fh, "%10.6lf: ", secondsAsDouble);
+            len = fprintf(gcx_fh, "%10.6f: ", secondsAsDouble);
         }
         len += fprintf(gcx_fh, fmt, args);
         fflush(gcx_fh);


atilaneves:
I don't see any changes in any tests, how can that be?

thewilsonator:
Because this is blocking a dmd PR and NFC. The tests are in that. 

Note that the first diff is used in assert messages so any change will show up. If that isn't covered by tests I'd be surprised.

The second one is for debug only.

src/core/internal/dassert.d
```diff
@@ -88,7 +88,7 @@ private string miniFormat(V)(const ref V v)
     else static if (__traits(isFloating, V))
     {
         char[60] val;
-        const len = sprintf(&val[0], "%g", v);
+        const len = sprintf(&val[0], "%Lg", cast(real) v);
```
Geod24:
```suggestion
        const len = sprintf(&val[0], "%Lg", real(v));
```

Also not sure about how I feel about everything being upcast to `real`.

Geod24:
> Note that the first diff is used in assert messages so any change will show up. If that isn't covered by tests I'd be surprised.

There is one test, but it does not include `real`: https://github.com/dlang/druntime/blob/b57ca7c86005e1e33cac1623c10b1a2fb6e2adc7/test/exceptions/src/assert_fail.d#L62-L67

src/core/internal/dassert.d
Luhrel: Fixed.

Luhrel: With your suggestion :

> Error: cannot implicitly convert expression v of type const(ifloat) to real

Will this cast change the result of `len`, depending of the base type ?

thewilsonator: you could just enumerate the different floating point types.

Geod24: @Luhrel : Can you add a real for `real` and the various imaginary types [here](https://github.com/dlang/druntime/blob/b57ca7c86005e1e33cac1623c10b1a2fb6e2adc7/test/exceptions/src/assert_fail.d#L62-L67) ?

Luhrel: > you could just enumerate the different floating point types.

It won't work with `float`, because `printf` doesn't accept it.
I simply escape the check.



> @Luhrel : Can you add a real for `real` and the various imaginary types [here](https://github.com/dlang/druntime/blob/b57ca7c86005e1e33cac1623c10b1a2fb6e2adc7/test/exceptions/src/assert_fail.d#L62-L67) ?

Done.

test/exceptions/src/assert_fail.d
```diff
@@ -64,6 +64,7 @@ void testFloatingPoint()()
     test(1.5, 2.5, "1.5 != 2.5");
     test(float.max, -float.max, "3.40282e+38 != -3.40282e+38");
     test(double.max, -double.max, "1.79769e+308 != -1.79769e+308");
+    test(real.max, -real.max, "1.18973e+4932 != -1.18973e+4932");
```
kinke: Please use another constant, `real` isn't 80-bit for most targets.

Luhrel: Maybe this way ?

```
import std.conv : to;
//test(real.max, -real.max, "1.18973e+4932 != -1.18973e+4932");
test(real.max, -real.max, real.max.to!string ~ " != -" ~ real.max.to!string);
```

How shall I test this with a number which change depending of the platform ?

kinke: Well, `real.max` seems totally arbitrary, so any nicely representable values should do (`5.25L` or whatever).

Luhrel: Changed to:
```
test(5.25L, -5.25L "5.25 != -5.25");
```

Luhrel: > Changed to:
> 
> ```
> test(5.25L, -5.25L, "5.25 != -5.25");
> ```

Somehow it returns this:
```
'-nan != -nan' != '5.25 != -5.25'
```
Any idea @kinke @Geod24 ?

src/core/internal/dassert.d
```diff
@@ -88,7 +88,8 @@ private string miniFormat(V)(const ref V v)
     else static if (__traits(isFloating, V))
     {
         char[60] val;
-        const len = sprintf(&val[0], "%g", v);
+        const format = "%g"; // escape printf format validation
+        const len = sprintf(&val[0], format.ptr, v);
```
kinke: That's not a fix; for C varargs, `float` is promoted to double, but D `real` as C `long double` needs a `%Lg` format...

Luhrel: Fixed. However, why `cfloat` aren't promoted to `creal` in this case ?

kinke: AFAIK, there's no complex type support for printf, and no promotions are done (it's seen as a struct with 2 fp values).

Luhrel: I added support for complex numbers in checkPrintFormat (see https://github.com/dlang/dmd/pull/10872/commits/090c02012f063b98bcd06f1d64759f8efe32e893).
Should we stay with that ?

kinke: According to https://stackoverflow.com/questions/4099433/c-complex-number-and-printf, printf doesn't support it.

Luhrel: Then, what should I do with `V` in `miniFormat`, which sometimes is a `creal`, a `cdouble` or a `cfloat` ?

kinke: Be creative - e.g., `"%[L]g + %[L]gi", v.re, v.im`.

Luhrel: I see.

src/core/internal/dassert.d
```diff
@@ -88,7 +88,15 @@ private string miniFormat(V)(const ref V v)
     else static if (__traits(isFloating, V))
     {
         char[60] val;
-        const len = sprintf(&val[0], "%g", v);
+        int len;
+        static if (is(V == cfloat) || is(V == cdouble))
+            len = sprintf(&val[0], "%gi", v.re, v.im);
```
kinke: I literally meant `"%g + %gi"` - you are simply printing 2 separately passed doubles => `1.5 + 0.5i`.

PR #2973, opened 2020-03-04
fix azure pipeline: building with LDC now needs DMD as a host compile…

rainers:
…r for tools, too

Cherry picked from master.


Geod24:
Phobos ?

rainers:
> Phobos ?

https://github.com/dlang/phobos/pull/7414

PR #2974, opened 2020-03-04
rt/trace: Some refactorings & code cleanup

Geod24:
Looking at this code, it is very obvious it hasn't been touched for years, if not over a decade.
It also needlessly exposes a bunch of C symbols which could easily conflict.
Here I am trying to refactor it gradually without triggering the issue I saw in #2969 .

### DIFF
--- a/src/rt/trace.d
+++ b/src/rt/trace.d
@@ -11,17 +11,14 @@
 
 module rt.trace;
 
-private
-{
-    import core.demangle;
-    import core.stdc.ctype;
-    import core.stdc.stdio;
-    import core.stdc.stdlib;
-    import core.stdc.string;
+import core.demangle;
+import core.stdc.ctype;
+import core.stdc.stdio;
+import core.stdc.stdlib;
+import core.stdc.string;
 
-    version (CRuntime_Microsoft)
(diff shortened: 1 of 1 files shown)


Geod24:
Also note that in the first commit, `trace_logfilename` was using `.ptr` without checking for `\0`, so that was a potential bug (which prompted the commit which always allocates the C strings).

src/rt/trace.d
```diff
@@ -77,32 +140,9 @@ __gshared
 
     timer_t trace_ohd;
 
-    string trace_logfilename = "trace.log";
-    string trace_deffilename = "trace.def";
```
thewilsonator: why not just embed a `\0` in them instead?

Geod24: Technically it's already embedded because it's a string literal :)
But the point was that I could not free them, so I'd have to special-case things for the initial case.

thewilsonator: Ah, fair enough.

Geod24: In particular this wouldn't work if we were pointing to a string literal:
https://github.com/dlang/druntime/pull/2974/files#diff-403c31034f603ad2b3ffcd39aa753a2cR80

PR #2976, opened 2020-03-06
[Do not merge!] Posix -> libc proposal

denizzzka:
_Do not merge! Not completed PR!_

Greetings! It is me again :-)

This is another idea how to get rid of libc. This idea is (surprisingly) that on most platforms we will never be able to get rid of libc because only it provides very important things (such as threads) on most platforms.

In druntime code word Posix constantly appears. But fact what we need an Posix interfacing in it, only references to `libc` and some functions from `core.*` and `rt.*` are enough. (Because libc is a part of Posix)

This patch introduces versions `DruntimeAbstractLibc` and `DruntimeAbstractRt`:

* `DruntimeAbstractRt` covers calls to system-depend  `core.*` and `rt.*` and provides way to implement any platform without touching main `druntime` source. It redirects some calls to  `core.*` and `rt.*` to `external.*` what should be provided at compiling stage if this version is used.
* `DruntimeAbstractLibc` is way to use any type of `libc` library by including new external `libc` package.  (External `libc` package also should be provided at compiling stage if this version is used.)

(Maybe `DruntimeAbstractLibc` can be renamed to something like `CRuntime_Abstract`)

In this way we can start gradually get rid of `libc` calls in `druntime`. In the worst case by transferring these calls into external.*, and at best by creating new independent (from `libc`) calls inside of `external.*`. Or maybe in some cases `dpp` generator can be used to provide interfacing with `libc` in external module. Also, by this way, some functions that are unnecessary for some platforms can easily be closed by stubs.

Existing code will not be affected.

Code what implements (actually just stubs) external calls for current state of this PR:
https://github.com/denizzzka/d_c_arm_test/tree/druntime_2att/stm32_druntime_backend

Waiting for your feedback!


denizzzka:
Code what implements external calls for current state of this PR:
https://github.com/denizzzka/d_c_arm_test/tree/druntime_2att/stm32_druntime_backend

PR #2977, opened 2020-03-06
Fix Issue 20640 - LDC: AddressSanitizer failing to catch SIGSEGV in unittest

kubo39:
Stop overriding SIGSEGV/SIGBUS handlers if already exists.


kubo39:
ah, sorry for late response and ok to close this.

PR #2978, opened 2020-03-07
[stable] Fix test_runner regression for LDC

kinke:
The previous code failed to compile with LDC for non-ARM targets. Fix that via proper versioning and switch from LDC-specific `__asm` to GDC-style `asm`, incl. tampering with the FPU mode with GDC too.


kinke:
Pinging @ibuclaw.

ibuclaw:
(gdc testsuite doesn't use this file to run druntime unittests).

PR #2982, opened 2020-03-10
remove wrong @safe attributes

aG0aep6G:
When issue 20655 ("attribute inference accepts unsafe union access as
@safe") is fixed, DMD correctly complains about these.

Needed for <https://github.com/dlang/dmd/pull/10884>.

### DIFF
--- a/src/core/stdcpp/string.d
+++ b/src/core/stdcpp/string.d
@@ -1900,9 +1900,9 @@ extern(D):
         }
 
         ///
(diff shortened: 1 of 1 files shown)


aG0aep6G:
@TurkeyMan, you wrote these, so you might want to have a look.

TurkeyMan:
Erm... these functions don't do anything unsafe... they all only call through to other functions, and all those functions only return counts and stuff.
I kinda feel like the functions these call through to should be marked trusted, and these should remain `@safe`?

PR #2988, opened 2020-03-15
Fix printf format in GC config

Luhrel:
Blocking https://github.com/dlang/dmd/pull/10920

### DIFF
--- a/src/core/gc/config.d
+++ b/src/core/gc/config.d
@@ -38,8 +38,8 @@ struct Config
         import core.gc.registry : registeredGCFactories;
 
         printf("GC options are specified as white space separated assignments:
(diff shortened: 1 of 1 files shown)


rainers:
I don't think this change should be necessary. The check should take "default parameter promotion" into consideration, see 6.5.2.2.7: "The ellipsis notation in a function prototype declarator causes
argument type conversion to stop after the last declared parameter. The default argument
promotions are performed on trailing arguments."

The defintion of the term is in 6.5.2.2.6: "If the expression that denotes the called function has a type that does not include a prototype, the integer promotions are performed on each argument, and arguments that have type float are promoted to double. These are called the default argument promotions."

PR #2991, opened 2020-03-16
Revamp rt.sections_android, switching to parsing the ELF headers

kinke:
Instead of using magic compiler and linker symbols. This solves a bunch of related issues for LDC, see https://github.com/ldc-developers/ldc/issues/3350 - wrong `_end` for x86 architectures, bfd-linker-only support and a requirement wrt. having a D `main` function (in .so libs too) *and* feeding the containing object file as first object to the linker.

In case DMD ever supports Android, its codegen might have to be tweaked (emitting TLS data into sections `.tdata` and `.tbss`; no need for magic `_tls{start,end}` symbols anymore).

### DIFF
--- a/src/rt/sections_android.d
+++ b/src/rt/sections_android.d
@@ -95,14 +95,6 @@ void initSections() nothrow @nogc
             void* end = start + phdr.p_memsz;
             debug(PRINTF) printf("data segment: %p - %p\n", start, end);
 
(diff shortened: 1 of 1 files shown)


thewilsonator:
Is this good to go?

Geod24:
Yep, DMD doesn't use this file anyway.

PR #2993, opened 2020-03-16
pure functions infer scope, but not return 3

WalterBright:
More demangle.d cases to fix.


WalterBright:
The buildkite error is not relevant to this PR.

PR #2994, opened 2020-03-16
Removes platform version() from time.d

denizzzka:
Many modules are scattered by version() keywords that implement common calls necessary when using any platform.

We can improve Makefiles to include only needed platform-dependend subdirs what will contain these modules but without "versions", just code for each supported platform.

After such replacement it will be very easy to clear out remaining libc calls (if it will be need - all supported platforms actually have libc interfaces), and also support for new arbitrary platforms will be ready - for supporting any new platform will be enough to replace platform-specific section with your own at build stage by build system.

https://forum.dlang.org/post/cehxbtoccbyobtjfnawn@forum.dlang.org

This PR can be called initial one or proof of concept. It touches only one D file and removes all platform-related version() directives from time.d file.


denizzzka:
Problem:

It is need to set default import directory depending on the platform. (Like currently compiler does default import for Phobos.) Otherwise Phobos and all other software isn't builds because platform specific code isn't imported.

PR #2995, opened 2020-03-17
pure functions infer scope, but not return 4

WalterBright:


Geod24:
How many more of those are there ? It's pretty hard to understand what's the end goal from this PR.
This code obviously compile at the moment, but won't with a future change, I suppose. Could you link the change when making a PR ? And/or perhaps prior changes.

When searching the `git` history and trying to understand why a change was made, finding those PRs without explanation, comment, or link is really annoying because of how much time it makes you loose. Just having "fix things" or an issue number is pretty useless when using history or blame.

WalterBright:
I have no idea how many more there are. In any case, the annotations are correct so I don't see it as really a problem.

WalterBright:
The buildkite errors are irrelevant to this PR.

aG0aep6G:
> This code obviously compile at the moment, but won't with a future change, I suppose. Could you link the change when making a PR ? And/or perhaps prior changes.

Since Walter apparently skimmed over that question: I believe these changes are in preparation for <https://github.com/dlang/dmd/pull/10924>.

Geod24:
Thanks, that makes sense. I'll make this link more obvious in the other PRs.

PR #2997, opened 2020-03-17
Split core.stdc.time to Posix and Windows versions

denizzzka:
<s>This PR adds proof of concept of splitting platform-dependent code</s>

https://forum.dlang.org/post/cehxbtoccbyobtjfnawn@forum.dlang.org

### DIFF
--- a/src/core/stdc/time.d
+++ b/src/core/stdc/time.d
@@ -15,137 +15,19 @@
 
 module core.stdc.time;
 
-private import core.stdc.config;
+version (Posix)
+    public import core.sys.posix.stdc.time;
+else version (Windows)
+    public import core.sys.windows.stdc.time;
+else
+    static assert(0, "unsupported system");
 
-version (OSX)
-    version = Darwin;
-else version (iOS)
-    version = Darwin;
-else version (TVOS)
-    version = Darwin;
-else version (WatchOS)
-    version = Darwin;
+private import core.stdc.config;
 
 extern (C):
 @trusted: // There are only a few functions here that use unsafe C strings.
 nothrow:
 @nogc:
-
-version (Windows)
-{
-    ///
-    struct tm
-    {
-        int     tm_sec;     /// seconds after the minute - [0, 60]
-        int     tm_min;     /// minutes after the hour - [0, 59]
-        int     tm_hour;    /// hours since midnight - [0, 23]
-        int     tm_mday;    /// day of the month - [1, 31]
-        int     tm_mon;     /// months since January - [0, 11]
-        int     tm_year;    /// years since 1900
-        int     tm_wday;    /// days since Sunday - [0, 6]
-        int     tm_yday;    /// days since January 1 - [0, 365]
-        int     tm_isdst;   /// Daylight Saving Time flag
-    }
-}
-else version (Posix)
-{
-    ///
-    struct tm
-    {
-        int     tm_sec;     /// seconds after the minute [0-60]
-        int     tm_min;     /// minutes after the hour [0-59]
-        int     tm_hour;    /// hours since midnight [0-23]
-        int     tm_mday;    /// day of the month [1-31]
-        int     tm_mon;     /// months since January [0-11]
-        int     tm_year;    /// years since 1900
-        int     tm_wday;    /// days since Sunday [0-6]
-        int     tm_yday;    /// days since January 1 [0-365]
-        int     tm_isdst;   /// Daylight Savings Time flag
-        c_long  tm_gmtoff;  /// offset from CUT in seconds
-        char*   tm_zone;    /// timezone abbreviation
-    }
-}
-
-version (Posix)
-{
-    public import core.sys.posix.sys.types : time_t, clock_t;
-}
-else version (Windows)
-{
-    ///
-    alias c_long time_t;
-    ///
-    alias c_long clock_t;
-}
-
-///
-version (Windows)
-{
-    enum clock_t CLOCKS_PER_SEC = 1000;
-    clock_t clock();
-}
-else version (OSX)
-{
-    enum clock_t CLOCKS_PER_SEC = 1_000_000; // was 100 until OSX 10.4/10.5
-    version (X86)
-        extern (C) pragma(mangle, "clock$UNIX2003") clock_t clock();
-    else
-        clock_t clock();
-}
-else version (Darwin) // other Darwins (iOS, TVOS, WatchOS)
-{
-    enum clock_t CLOCKS_PER_SEC = 1_000_000;
-    clock_t clock();
-}
-else version (FreeBSD)
-{
-    enum clock_t CLOCKS_PER_SEC = 128;
-    clock_t clock();
-}
-else version (NetBSD)
-{
-    enum clock_t CLOCKS_PER_SEC = 100;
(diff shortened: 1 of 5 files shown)


denizzzka:
Everything seems to be done well, but why Windows test are broken?

Geod24:
`mak/COPY` and `mak/SRCS` are missing the windows header

denizzzka:
Related discussion: https://forum.dlang.org/post/oyuxvperqmddgndvmlyx@forum.dlang.org

src/core/stdc/time.d
```diff
@@ -15,138 +15,22 @@
 
 module core.stdc.time;
 
-private import core.stdc.config;
+version (CRuntime_Abstract)
+    public import platform_dependent.stdc.time;
```
ibuclaw: Need something else here perhaps.

denizzzka: if you about
```D
private import core.stdc.config;
```
it moved to 27 line

denizzzka: Please explain more

ibuclaw: Neither the version exists in the spec nor the module exists in druntime.

denizzzka: Yes, this is made intentionally. This version must be used if you want to connect an arbitrary external platform implementation.

Currently it is not used in *.mak scripts. It makes no sense to add it because some CI test implementation code for it does not exist yet.

ibuclaw: I'd just leave such cases to hit the static assert. Don't bother trying to cater for it.

denizzzka: @ibuclaw i.e., should try to compile `import platform_dependent.*` and if it is not compiles then throw static assertion?

ibuclaw: > @ibuclaw i.e., should try to compile `import platform_dependent.*` and if it is not compiles then throw static assertion?

Hi, sorry, I dropped the ball here.

I think for now we need not bother with handling unknown platforms in druntime.

If we ever need such machinery, better to do it all in one go. So please just use `static assert(false, "unsupported platform");` to be consistent with the rest of druntime.

denizzzka: > I think for now we need not bother with handling unknown platforms in druntime.

Not "unknown" but "abstract".

> If we ever need such machinery, better to do it all in one go.

It is already needed! Al these MCUs, thousands of configarations! (like: Bare metal ARM, FreeRTOS ARM, bare metal RISC-V, FreeRTOS RISC-V, FreeRTOS RISC-V with another Memory Manager, etc.)
Also as a side effect we get rid of libc.

denizzzka: In other words, my proposal is to make possible to replace of the low-level calls used by by druntime such as memory allocation or thread creation without changing druntime code.

ibuclaw: > In other words, my proposal is to make possible to replace of the low-level calls used by druntime such as memory allocation or thread creation without changing druntime code.

Then we can do that as a phase-2 step, and add all ~30 places in druntime in one PR. :-) 

First, let's just split up core.time, to keep refactoring and actual changes separated.

denizzzka: > Then we can do that as a phase-2 step, and add all ~30 places in druntime in one PR. :-)

Ok, but first we need to reach an agreement that this is exactly the way to add an "abstract" platform that will be merged. Otherwise this huge amount of work will be wasted.

> First, let's just split up core.time, to keep refactoring and actual changes separated.

Now I think this is not the best candidate for the first change. It is too huge. Maybe I find something smaller.

ibuclaw: > > First, let's just split up core.time, to keep refactoring and actual changes separated.
> 
> Why you talk about  core.time?
> If it isn't typo now I think this is not the best candidate for the first change. It is too huge. Maybe I find something smaller.

I did indeed mean core.stdc.time.

Furthermore CRuntime_Abstract is an  undocumented, unimplented, unapproved version condition; and it ignores an already existing and documented version condition made precisely for bare metal targets.

ibuclaw: So I reiterate, let's remove these two lines for now, and think more broadly about how druntime could support free-standing environments in the future.

denizzzka: Ok, I removed these two lines

denizzzka:
> and think more broadly about how druntime could support free-standing environments in the future.

> Furthermore CRuntime_Abstract is an undocumented, unimplented, unapproved version condition; and it ignores an **already existing and documented version condition made precisely for bare metal targets**.

What do you mean?

An important nuance here: we do not have support for just any abstract bare metal targets. For example, we have support for `CRuntime_UClibc`, but not for `CRuntime_picolibc` or for something like `CRuntime_my_own_best_baremetal_libc`

denizzzka:
> Furthermore CRuntime_Abstract is an undocumented, unimplented, unapproved version condition

About implementation: I can implement some dumb or already-implemented platform as implementation of these abstract platform for CI purposes. This is appropriable?

n8sh:
> @ibuclaw
> 
> > an **already existing and documented version condition made precisely for bare metal targets**.
> 
> What do you mean?

I believe he means [`version (FreeStanding)`](https://dlang.org/spec/version.html#predefined-versions).

ibuclaw (approved):

ibuclaw:
> (As I understand) we do not have support for just any abstract bare metal targets. We have support for `CRuntime_UClibc`, but not for `CRuntime_picolibc` or for something like `CRuntime_my_own_best_baremetal_libc`

- As far as I can tell picolibc is just newlib with some under-the-hood changes. So one can get away with adding and implementing CRuntime_Newlib to cover both, but you'll be expected to link to the right library yourself.

Any further down you go though, it does get harder to justify why you would be using druntime in the first place, as opposed to, say, [minilibd](https://github.com/embedded-boston/awesome-embedded-systems/blob/master/README.md#d).

denizzzka:
> I believe he means version (FreeStanding).

Ah, got it, thanks!

Then I chose the unfortunate term "baremetal target" here. I meant "support for any OS and hardware including my own what will not become mainstream"

denizzzka:
> As far as I can tell picolibc is just newlib with some under-the-hood changes. So one can get away with adding and implementing CRuntime_Newlib to cover both

It is not possible by this way to add rare or only developing systems. D will always be catching up in this case.

> Any further down you go though, it does get harder to justify why you would be using druntime in the first place, as opposed to, say, minilibd.

minilibd (and others) does not provide all features that provides "official" druntime. Its author decided to be content with the minimum necessary because writing and maintaining of druntime is not easy.

In turn, I see no reason for separation of "main" druntime and third party "mini" druntimes. Most likely, the authors of these libraries did not want to go through this bureaucratic way to improve official druntime. 

Another important consequence is "marking" (or disposal) of some libc calls.

Panke:
Would you ideally want to have the posix file split into seperate platforms as well?

denizzzka:
No. In the future I want to separate druntime code into Windows, Posix and External, where External will be implemented as ability to pass "low-level backend" to druntime code in compile time.

This is how it looks (this is fork of LDC druntime) with such additions: https://github.com/denizzzka/druntime/tree/non_posix

PR #2998, opened 2020-03-19
Reduces build scripts code duplication

denizzzka:


Geod24:
A good example of [Chesterton's fence](https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence).
This duplication is in place because DM make does not support wildcard rules.

denizzzka:
Can someone tell me an easy way to test Windows builds on Linux?

> This duplication is in place because DM make does not support wildcard rules.

According to the latest fashion trends (see Meson build system) this is not bad, but good. In any case, this will not be a problem.

denizzzka:
> This duplication is in place because DM make does not support wildcard rules.

In fact, I am very indignant: duplication here is because someone was too lazy to write a more smart script, not because wildcards isn't supported.

Also, in order to develop such a core thing of the language as a runtime library developer actually forced to use Windows. Right now I, probably, violate MS license by downloading Virtualbox image with Windows inside from torrents.

Remembering my previous attempts to add changes to Phobos or druntime: about ~90% of errors were on Windows CI checks. If I could I would drop Windows support into an external library to let fans of proprietary OSes support his stuff separately.

ibuclaw:
> Also, in order to develop such a core thing of the language as a runtime library developer actually forced to use Windows. Right now I, probably, violate MS license by downloading Virtualbox image with Windows inside from torrents.

Microsoft provides development images for free. They are only valid for a limited time, but they seem to regularly update the image when the previous one has expired.

https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/

ibuclaw:
> Can someone tell me an easy way to test Windows builds on Linux?

DMC runs on wine without any issues.

kinke:
> In fact, I am very indignant: duplication here is because someone was too lazy to write a more smart script, not because wildcards isn't supported.

Christ, as should be absolutely obvious after Mathias' explanation and looking at the CI results, it's not about laziness, but as Mathias clearly stated, the usage of DigitalMars make (and not GNU make) on Windows, at least for the CI boxes...

denizzzka:
Fixed in #3026

PR #2999, opened 2020-03-20
abort() msg parameter escapes

WalterBright:
and so it cannot be used to call `assert(0, msg)`.


WalterBright:
Blocking https://github.com/dlang/dmd/pull/10947

PR #3003, opened 2020-03-24
[WIP] [DONT MERGE] Enable demangling of C++ symbols in the stacktrace

ghost:
This allows you to view demangled and pretty printed C++ symbols
in the stacktrace.
Is especially useful considering that D is getting more and more support for C++.

Now the C++ symbols in the stacktrace will be printed like this:
```
        test.d:7 [C++] Test<int>::SomeName(int, long, int) [0x55711efee36f]
        test.d:18 [C++] testcpp() [0x557228e17978]
```

This need the PR #3002

### DIFF
--- a/changelog/cpptrace.dd
+++ b/changelog/cpptrace.dd
@@ -1,0 +1,51 @@
+Added experimental `C++` symbol demanling in the stacktrace
+
+This feature is availableby  passing the following switches to your executable (not to the compiler):
+- `--DRT-cpptrace=enable:y`: Enable demangling of C++ symbols
+- `--DRT-cpptrace=prefix:<your_prefix>`: Change the prefix preceding the demangled C++ name (by default would be `[C++]`)
+- `--DRT-cpptrace=noprefix:n`: Disable adding the prefix to C ++ demangle names
+
+To use this function in the Posix platform you will need to link your executable to the phobos shared library, compile the program by passing -defaultlib=libphobos2.so to DMD.
+
+Example:
+
+```
+module cpp_trace.d
+
+extern(C++) void f1()
+{
+    throw new Exception("exception");
+}
+
+void main()
+{
+    try
+    {
+        f1();
+    }
+    catch (Exception e)
+    {
+        import core.stdc.stdio;
+        auto str = e.toString();
+        printf("%.*s\n", cast(int)str.length, str.ptr);
+    }
+}
+```
+
+If you run the executable with the following switches `./<executable> --DRT-cpptrace=enable:y`, you would get this output:
+
+```
+object.Exception@cpp_trace.d(3): exception
+----------------
+src/cpp_trace.d:5 [C++] f1() [ADDR]
+src/cpp_trace.d:12 _Dmain [ADDR]
+```
+
+Instead, if you run the executable with the following switches `./<executable> --DRT-cpptrace=enable:y --DRT-cpptrace=prefix:CPP`, you would get this output:
+
+```
+object.Exception@cpp_trace.d(5): exception
+----------------
+src/cpp_trace.d:5 [CPP] f1() [ADDR]
+src/cpp_trace.d:12 _Dmain [ADDR]
+```
--- a/src/core/internal/cpptrace.d
+++ b/src/core/internal/cpptrace.d
@@ -1,0 +1,124 @@
+/**
+* This module provides the demangling of C++ symbols for stacktrace
+*
+* Copyright: Copyright © 2020, The D Language Foundation
+* License: $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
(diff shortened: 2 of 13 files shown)


Geod24:
FYI we now have https://github.com/dlang/dmd/pull/10964 in DMD.
We might be interested in having it here, too.

rainers:
FYI there was a similar PR: https://github.com/dlang/druntime/pull/2083

I think this one has better separation. For Windows, the VC mangling should be supported via UnDecorateSymbolName. Please also add a test case.

src/core/sys/windows/stacktrace.d
```diff
@@ -294,7 +294,19 @@ private:
             size_t decodeIndex = 0;
             tempSymName = decodeDmdString(tempSymName, decodeIndex);
         }
-        res ~= demangle(tempSymName, demangleBuf);
+        auto demangledName = demangle(tempSymName, demangleBuf);
+
+        version (Shared)
```
rainers: Why limit this to a shared build of the runtime?

ErnyTech: you're right, for Windows it is not necessary

rainers: Why is it necessary on other platforms. "Shared" is not a version for dlopen support, but whether the runtime itself is a shared library.

ErnyTech: druntime must be launched with -ldl, otherwise the linker will not find dlopen.
I don't know how i can check this in a better way

rainers: The compiler already adds -ldl to the linker command line (at least for linux):
https://github.com/dlang/dmd/blob/e8f984995598543279c07b9800d19a9e19321562/src/dmd/link.d#L756

ErnyTech: Yes, but only if phobos is linked

ErnyTech: Ok but if I create the object files with dmd and then link it manually? I'm not sure it's a good idea to depend on libdl for druntime

rainers: Ah, I see that `rt_loadLibrary` is only implemented if `version(Shared)`, so it's consistent to not use it here, too. That restriction seems a bit strange, though.

ErnyTech: We could do better if dmd supports weak symbols.
It would be enough to create a dlopen function that returns null marked as weak symbol, so if the user links libdl the real dlopen would be used

changelog/cpptrace.dd
```diff
@@ -0,0 +1,51 @@
+Added experimental `C++` symbol demanling in the stacktrace
+
+This feature is availableby  passing the following switches to your executable (not to the compiler):
```
Geod24, line 3: Ideally, the release note should start by describing the feature itself. You've been working on it for quite some time, so you know exactly what it entails, but not everyone does. Something like:
```
When a stack trace was generated by the runtime (e.g. when calling `Throwable.toString`), C++ names were not previously demangled. Hence the following code:
---
<INSERT CODE>
---
Would lead to the following stack trace:
---
<INSERT MANGLED STACK TRACE>
---

Starting from this release, C++ names will be demangled whenever possible, hence the previous stack trace will now show as:
---
<INSERT DEMANGLED STACK TRACE>
---

This feature is still experimental and must be opted-in by <BLAH BLAH>.
Demangling is only possible when <BLAH BLAH>.
```

Note that this is just an example, but it illustrate a few points:
- Always describe the feature first, what it brings to the user: imagine giving an elevator pitch about it;
- Try to show the difference between the old behavior and the new one: remember users might not be familiar with the old behavior, especially if someone looks at this release note in a few year;
- Include a code example whenever possible (which you did already);
- Put restrictions / limitations at the end;

src/core/internal/cpptrace.d
```diff
@@ -0,0 +1,119 @@
...
+    {
+        return copyResult(buf, dst);
+    }
+
+    return CPPTrace.instance.cppdemangle(buf, dst, !CPPTrace.config.noprefix, CPPTrace.config.prefix);
+}
+
+private struct CPPTrace
+{
+    __gshared CPPDemangle instance;
```
Geod24, line 43: Don't we already have a singleton in the other module?

src/core/internal/cpptrace.d
```diff
@@ -0,0 +1,119 @@
...
+    __gshared Config config;
+
+    static this()
```
Geod24, line 46: A `static this()` is run on each thread constructions. You want a `shared static this` here, no ?

Geod24 (commented):
Added a few comments about the second commit

RazvanN7:
@ErnyTech What's the status on this?

PR #3004, opened 2020-03-24
Rename `AppleARM` version identifier to `iOSDerived`

jacob-carlborg:
The name `AppleARM` was used to indicate any of the platforms: iOS, tvOS and watchOS. The name was misleading since the version identifier for these platforms would be enabled also when compiling for the simulators. The simulators are running x86(-64).


Geod24:
Bypassing buildkite because it's broken

PR #3007, opened 2020-03-26
add @system to posix interface

WalterBright:


CyberShadow:
Is this for DIP 1028? If so, would it make sense to wait until the discussion regarding what to do about `extern(C)` function finalizes?

Are library (binding) authors expected to follow suit?

WalterBright:
> Is this for DIP 1028?

Yes.

> If so, would it make sense to wait until the discussion regarding what to do about extern(C) function finalizes?

It should be done regardless.

> Are library (binding) authors expected to follow suit?

Yes.

WalterBright:
Hmm, this is a little strange:
```
```

WalterBright:
Dang, I can't repro it locally.

PR #3009, opened 2020-03-26
core.sys.posix.signal: add @system

WalterBright:


Geod24:
@WalterBright : Looks like there is a difference in the output of `stringof` for Linux32. The test relies on `stringof`, so I don't think it's a good test (we explicitly advise against it), but regardless the difference should be investigated.

pbackus:
I did some digging on the failing `.stringof` test for #3117 , and I think it's [issue 3796](https://issues.dlang.org/show_bug.cgi?id=3796). Here's a reduced test case:

```d
module testInference;

import core.demangle: demangle;

void foo8504()() {}

auto toDelegate8504a(F)(auto ref F fp) { return fp; }

extern(C) void testC8504() {}

void test8504()
{
    // commenting out the following line makes the test pass on 32-bit linux
    static assert(demangle(foo8504!().mangleof) == "pure nothrow @nogc @safe void testInference.foo8504!().foo8504()");

    auto fp1 = toDelegate8504a(&testC8504);
    static assert(typeof(fp1).stringof == "extern (C) void function()");
}
```

RazvanN7:
There doesn't seem to be any failures any more (for the tests that completed up until now). Is there anything blocking this PR?

RazvanN7 (approved):

RazvanN7:
This seems to break the dmd test suite:

```
```

dkorpel:
Probably blocked by https://issues.dlang.org/show_bug.cgi?id=18269, where because of an import of this module, the first `extern (C) void function()` that dmd sees now has an explicit `@system` attribute.

PR #3010, opened 2020-03-26
demangle.d: remove dead code

WalterBright:
`fakePureReprintReal()` is not used.


WalterBright:
guess it's used after all, under a different name.

PR #3011, opened 2020-03-30
Implement Github actions to test `core.stdcpp`

Geod24:
Building on https://github.com/dlang/dmd/pull/10964

This adds a Github action to test with multiple versions of clang on OSX and multiple versions of g++ on Linux. However it is not possible to test clang+Linux because DMD assumes we're targeting g++. Likewise it probably won't be possible to test Windows+clang or Windows+mingw (although I will give it a shot).

CC @TurkeyMan : This should allow us to test with the new ABI as well (by adding a matrix dimension).

### DIFF
--- a/.github/workflows/cxx.yml
+++ b/.github/workflows/cxx.yml
@@ -1,0 +1,247 @@
+# Github action to test for C++ interoperability
+#
+# This file is modeled after DMD's Github action definition (`runnable_cxx.yml`).
+# Its goal is to test interaction with the C++ standard library on multiple platforms.
+# Look up the DMD file for more information about reasoning, patterns, caveats, etc...
+#
+# TODO:
+# - Implement Windows + MSVC support
+# - Implement Windows + clang support
+# - Implement Linux + Clang 32 bits support (if possible)
+name: stdcpp
+
+on:
+  pull_request:
+    branches:
+      - master
+      - stable
+  push:
+    branches:
+      - master
+      - stable
+      # Use this branch name in your fork to test changes
+      - github-actions
+
+jobs:
+  main:
+    name: Run
+    strategy:
+      fail-fast: false
+      # Matches the matrix in DMD to support the same platforms
+      matrix:
+        os: [ macOS-10.15, ubuntu-16.04, windows-2019 ]
+        target: [
+          clang-9.0.0, clang-8.0.0, clang-7.0.0,
+          clang-6.0.0,  clang-5.0.2, clang-4.0.0, clang-3.9.0,
+          g++-9, g++-8, g++-7, g++-6, g++-5,
+          msvc-2019, msvc-2017, msvc-2015, msvc-2013
+        ]
+
+        exclude:
+          # Ubuntu supports clang and g++
+          - { os: ubuntu-16.04, target: msvc-2019 }
+          - { os: ubuntu-16.04, target: msvc-2017 }
+          - { os: ubuntu-16.04, target: msvc-2015 }
+          - { os: ubuntu-16.04, target: msvc-2013 }
+          # FIXME: Can't test clang+Linux because DMD assume g++
+          - { os: ubuntu-16.04, target: clang-9.0.0 }
+          - { os: ubuntu-16.04, target: clang-8.0.0 }
+          - { os: ubuntu-16.04, target: clang-7.0.0 }
+          - { os: ubuntu-16.04, target: clang-6.0.0 }
+          - { os: ubuntu-16.04, target: clang-5.0.2 }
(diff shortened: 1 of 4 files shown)


.github/workflows/cxx.yml
```diff
@@ -0,0 +1,247 @@
...
+    runs-on: ${{ matrix.os }}
+    steps:
+
+    ########################################
+    #    Setting up the host D compiler    #
+    ########################################
+    - name: Prepare compiler
+      uses: mihails-strasuns/setup-dlang@v0.5.0
+      with:
+          compiler: dmd-2.091.0
```
PetarKirov:
```suggestion
          compiler: ${{ matrix.host_dc }}
```

Make this a variable, by adding it as a matrix dimension, but with just a single item:

```yml
matrix:
  host_dc: [ dmd-2.091.0 ]
  // ...
```

.github/workflows/cxx.yml
```diff
@@ -0,0 +1,247 @@
...
+        persist-credentials: false
+
+
+    ########################################
+    #   Setting up the host C++ compiler   #
+    ########################################
+    - name: '[Posix] Setting up clang ${{ matrix.cxx-version }}'
+      if: matrix.compiler == 'clang' && runner.os != 'Windows'
+      run: |
+        wget --quiet --directory-prefix=${{ github.workspace }} https://releases.llvm.org/${{ matrix.cxx-version }}/clang+llvm-${{ matrix.cxx-version }}-${{ matrix.arch }}.tar.xz
```
PetarKirov: (I'm on the phone, so I can't check, but) it seems like this very similar or the same as script inside the dmd GitHub Actions config. What do you think about extracting it to dmd/ci/setup_cxx.bash?

AFAIR we're already doing something similar for CircleCI in phobos.

PetarKirov:
> However it is not possible to test clang+Linux because DMD assumes we're targeting g++

Would this work: https://askubuntu.com/a/1198094?

Geod24:
> Would this work: https://askubuntu.com/a/1198094?

No, the problem is that `CppRuntime_Gcc` is unconditionally defined.

.github/workflows/cxx.yml
Geod24: I considered doing this at first, but decided against it. One of the reason to do so is that it shows up in the matrix, thus leading to longer name (and the host compiler information is not that important). The other reason is that I didn't want to add a level of indirection or mislead people into believing more dimensions should be added.

Is there any reason to make it part of the matrix when we only want 1 entry ?

.github/workflows/cxx.yml
Geod24: We can't extract it because github pre-process those scripts, e.g. the `${{ github.workspace }}` is replaced to create a script.
Of course I could use the environment variables, but it looks like a step back in readability.
I think a better course of action would be to extract this into its own separate action, which I'll do once the Windows support is ready.

.github/workflows/cxx.yml
PetarKirov: The biggest reason for me is to make it easy change configuration by extracting things like that as variables in one central place. I don't know if using a matrix of the best way to achieve this. In most other CI scripts in the dlang repos, test inputs like the D compiler are parameterized.

.github/workflows/cxx.yml
PetarKirov: Yes, extracting it as an action (or a Docker image) that can be reused actually repos is u ultimately the best way to go. I didn't propose this as I thought it would require much more work from your side.

PetarKirov:
> No, the problem is that CppRuntime_Gcc is unconditionally defined.

That's ill designed to say the least. The way it should work, is the `Target` class in dmd should be aware of the full target triple, the C and C++ compiler(s) used, the linker (bfd vs gold vs lld on Linux, MSVC link vs lld vs Optlink on Windows, etc.) , the C++ standard, and the C++ library used (e.g. libstdc++ vs libc++ on Posix).
All this info should be accessible through `__traits(getTargetInfo, ...)`.

Geod24:
> The way it should work [...]

Yes 100%.

But who's going to implement it ? GCC has proper target triple support, and so does LDC. DMD is lacking behind and it's going to be an uphill battle and an hopeless bikeshedding over how we deal with the CLI arguments (what happens when one calls `dmd -m64 -mtriple=x86-pc-linux-gnu` ?).

And *even* if we implement it in DMD, spending dozens of contributors hours to get it right, it won't solve the real problem, because *we would need to distribute two runtimes* for people to be able to link with both libc++ (because unlike the C definitions, the C++ ones are not "definition only").

Really those bindings shouldn't be in druntime, but no one was able to convince W&A.

PR #3012, opened 2020-03-31
Fix Issue 20711 - object.update requires the "update" callback to was…

CyberShadow:
…tefully return a copy of the updated value

Also update the examples to illustrate the new, more efficient option.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2649,7 +2649,7 @@ private enum bool isSafeCopyable(T) = is(typeof(() @safe { union U { T x; } T *x
  *      update = The callable to apply on update.
  */
 void update(K, V, C, U)(ref V[K] aa, K key, scope C create, scope U update)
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -2678,14 +2683,14 @@ if (is(typeof(create()) : V) && is(typeof(update(aa[K.init])) : V))
     aa.update("k1", {
         return -1; // create (won't be executed
```
GilesBathgate: Missing closing paren in comment.

GilesBathgate:
If I understand the fix is to ensure that if the callback function returns void (i.e its a subroutine in other languages) then no copy of the result needs to be made.

In which case the change looks good to me.

src/object.d
CyberShadow: Thanks. https://github.com/dlang/druntime/pull/3014

CyberShadow:
Had to rebase as it seems the Buildkite Phobos test does not merge the branch into master before testing.

PR #3015, opened 2020-04-01
[SystemZ] Fix layout of struct sigaction_t

ibuclaw:
The ordering of some fields in  struct sigaction on s390x (64bit)
differs compared to s390 and other architectures.
This patch adjusts this order according to the definition of
glibc-src/sysdeps/unix/sysv/linux/s390/bits/sigaction.h

Without this fix e.g. the call
sigaction( suspendSignalNumber, &sigusr1, null ) in thread.d
leads to setting the sa_restorer field to 0xffffffffffffffff.
In case a signal, the signal handler returns to this address
and the process stops with a SIGILL.

This was observable in several execution testcases on s390x.

### DIFF
--- a/src/core/sys/posix/signal.d
+++ b/src/core/sys/posix/signal.d
@@ -576,24 +576,51 @@ else
 
 version (CRuntime_Glibc)
 {
-    struct sigaction_t
+    version (SystemZ)
     {
-        static if ( true /* __USE_POSIX199309 */ )
+        struct sigaction_t
         {
-            union
+            static if ( true /* __USE_POSIX199309 */ )
+            {
+                union
+                {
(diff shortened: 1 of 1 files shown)


Geod24:
I'll give it a look soon. Also shouldn't this target stable and dmd-cxx?

ibuclaw:
dmd-cxx yes, stable... who will use it?

Geod24:
Musl!

Geod24:
It's just such a simple change that including it in stable is a no-brainer.
Otherwise LGTM. I wish we didn't have to duplicate the struct definition like this every time and that we could just use in-line `version` condition though.

Geod24:
Well let's go with master then!

PR #3016, opened 2020-04-01
Turn on markdown for druntime doc build

schveiguy:
I want to use the auto tester to see what happens when we turn on markdown. ~~Don't merge, I may have to fix things first.~~

### DIFF
--- a/src/rt/dwarfeh.d
+++ b/src/rt/dwarfeh.d
@@ -746,9 +746,9 @@ LsdaResult scanLSDA(const(ubyte)* lsda, _Unwind_Ptr ip, _Unwind_Exception_Class 
  *      exceptionClass = which language threw the exception
  *      lsda = pointer to LSDA table
  * Returns:
- *      >=1 means the handler index of the classType
- *      0 means classType is not in the Action Table
- *      <0 means corrupt
+ *      - &gt;=1 means the handler index of the classType
+ *      - 0 means classType is not in the Action Table
+ *      - &lt;0 means corrupt
  */
 int actionTableLookup(_Unwind_Exception* exceptionObject, uint actionRecordPtr, const(ubyte)* pActionTable,
                       const(ubyte)* tt, ubyte TType, _Unwind_Exception_Class exceptionClass, const(ubyte)* lsda)
--- a/posix.mak
+++ b/posix.mak
@@ -65,7 +65,7 @@ LINKDL=$(if $(findstring $(OS),linux),-L-ldl,)
 
 MAKEFILE = $(firstword $(MAKEFILE_LIST))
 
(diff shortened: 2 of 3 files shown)


schveiguy:
OK I think as long as the docs come out OK, this is mergeable.

schveiguy:
Hm... the ddox changed, how is that generated? I don't see anything in the makefile. The change is slightly worse than the original, but the original sucked too.

schveiguy:
Even with the ddox change, I think this is good to merge.

PetarKirov:
> Hm... the ddox changed, how is that generated? I don't see anything in the makefile.

https://github.com/dlang/dlang.org/tree/36c76465537501a0b34f32792f5a76f1f9b3888d/dpl-docs

https://github.com/dlang/dlang.org/blob/36c76465537501a0b34f32792f5a76f1f9b3888d/posix.mak#L706

schveiguy:
@PetarKirov thanks!

Hm.. not sure I can do anything. It looks like json output doesn't spit out markdown results.

PetarKirov:
@schveiguy Actually the `-preview=markdown` flag shouldn't affect Ddox at all, since Ddox has its own parsing [based on](https://github.com/rejectedsoftware/ddox/tree/master/source/ddox/parsers) the json files output from dmd and and libdparse. Additionally it has its own macro engine for Ddoc: https://github.com/rejectedsoftware/ddox/blob/master/source/ddox/ddoc.d

src/rt/dwarfeh.d
```diff
@@ -746,9 +746,9 @@ LsdaResult scanLSDA(const(ubyte)* lsda, _Unwind_Ptr ip, _Unwind_Exception_Class
  *      exceptionClass = which language threw the exception
  *      lsda = pointer to LSDA table
  * Returns:
- *      >=1 means the handler index of the classType
- *      0 means classType is not in the Action Table
- *      <0 means corrupt
+ *      - &gt;=1 means the handler index of the classType
+ *      - 0 means classType is not in the Action Table
+ *      - &lt;0 means corrupt
```
PetarKirov, line 751: I wonder why was this necessary, as both Ddoc and Ddox are supposed to escape this Sunday automatically.

schveiguy: The > was interpreted as a block quote in markdown. I added the dashes to make it into a list as the glued together statements looked bad.

schveiguy: This is what it looked like before the change: http://dtest.dlang.io/artifact/website-36c76465537501a0b34f32792f5a76f1f9b3888d-01a9ae3ff87a8f9c5eb4438e29e16951/web/phobos-prerelease/rt_dwarfeh.html#.actionTableLookup

PetarKirov: Ah yes, thanks for checking

PR #3018, opened 2020-04-05
[stable] core.internal.dassert: Fix miniFormat for structs with non-copyable fields

kinke:
This is blocking dlang/dmd#11005.

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -156,10 +156,10 @@ private string miniFormat(V)(const ref V v)
         string msg = "[";
         foreach (k, ref val; v)
         {
-            if (i++ > 0)
+            if (i > 0)
                 msg ~= ", ";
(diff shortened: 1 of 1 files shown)


src/core/internal/dassert.d
```diff
@@ -156,10 +156,10 @@ private string miniFormat(V)(const ref V v)
         string msg = "[";
         foreach (k, ref val; v)
         {
-            if (i++ > 0)
```
MoonlightSentinel, line 159: Unrelated change?

kinke: Yeah, noticed this inconsistency (would have only shown first 29 items) and fixed it.

kinke:
> Could use a test

The whole module could, but I didn't want to start adding the first unittest either. ;) - It's tested by the DMD PR.

PR #3019, opened 2020-04-06
Explicitly call _d_assert_fail in test/exceptions/assert_fail.d

MoonlightSentinel:
This test should only check the druntime component of `-checkaction=context`,
i.e. that `_d_assert_fail` yields an appropriate assert message. Using `assert`
directly linked this test to the lowering details and made changes changes
quite difficult because one cannot make changes to dmd and druntime in a
single PR (see e.g.  dlang/dmd#11005).

CC @kinke, @wilzbach

### DIFF
--- a/test/exceptions/src/assert_fail.d
+++ b/test/exceptions/src/assert_fail.d
@@ -1,35 +1,19 @@
-import core.stdc.stdio : fprintf, printf, stderr;
+import core.stdc.stdio : fprintf, stderr;
+import core.internal.dassert : _d_assert_fail;
 
 void test(string comp = "==", A, B)(A a, B b, string msg, size_t line = __LINE__)
 {
-    test(assert(mixin("a " ~ comp ~ " b")), msg, line);
+    test(_d_assert_fail!(comp, A, B)(a, b), msg, line);
 }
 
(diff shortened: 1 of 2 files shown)


kinke:
Thanks, that will definitely help; at some point, these tests can be turned to regular unittests (for miniformat etc. too). I'd need it in stable though.

test/exceptions/src/assert_fail.d
```diff
@@ -1,35 +1,19 @@
...
-    }();
-    // don't use assert here for better debugging
-    if (ret != 0) {
-        import core.stdc.stdlib : exit;
-        exit(1);
+    import core.exception : AssertError;
+
+    if (actual != expected)
+    {
+        const msg = "Missmatch!\nExpected: <" ~ expected ~ ">\nActual:   <" ~ actual ~ '>';
```
kinke, line 15: Mi**s**match

MoonlightSentinel: Done

MoonlightSentinel:
Rebased onto stable

MoonlightSentinel:
BuildKite fails because it uses dmd-master ...

PR #3021, opened 2020-04-09
Fix UnDecorateSymbolNameFunc arguments

ErnyTech:
The arguments of UnDecorateSymbolNameFunc are different than the Windows documentation.
A DecorateSymbolName should only accept ASCII following what specified by its documentation, the Unicode version is UnDecorateSymbolNameW.

See https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-undecoratesymbolname


ErnyTech:
Wait! I just saw that UnDecorateSymbolNameFunc isn't even loaded from the library.

ErnyTech:
See https://github.com/dlang/druntime/pull/3022

PR #3022, opened 2020-04-09
Fix UnDecorateSymbolNameFunc from dbghelp

ErnyTech:
Fix UnDecorateSymbolNameFunc from dbghelp

UnDecorateSymbolName must be loaded from dbghelp, before it was always null.
The arguments of UnDecorateSymbolNameFunc were different than the Windows documentation.
The function DecorateSymbolName should only accept ASCII following what specified by its documentation, the Unicode version is UnDecorateSymbolNameW.

See https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-undecoratesymbolname


ErnyTech:
I need it for the stacktrace ;-)

PR #3023, opened 2020-04-09
Add CPU_ALLOC and CPU_FREE

kubo39:


kubo39:
This is not useful for most users, so I don't think I need to add it to the changelog.
I only use it to improve totalCPUs for the system which has processors >= 1024.

PR #3024, opened 2020-04-10
~1000 redundant lines removed from mak/WINDOWS

denizzzka:


thewilsonator (approved):
Looks great, ping me when its good.

Geod24:
Looks like the right direction, however I'd ping some people that are familiar with this build system before merging. Also can we not add more top-level file ? Put them in `mak` for example.

denizzzka:
>  Also can we not add more top-level file ? Put them in mak for example.

I suggest first to decide exactly where to move it. Because it looks like ./mak/ directory is for makefiles strictly.

wildtype.bat
```diff
@@ -0,0 +1,16 @@
+@echo off
+REM Generates makefile sequences for Digital Mars make.exe (it isn't supports wildcards by itself)
```
thewilsonator, line 2: (it _doesn_'t supports wildcards by itself)

denizzzka:
@thewilsonator ping :-)

thewilsonator:
Please put those `.bat` files in `mak`

denizzzka:
@thewilsonator done!

mak/wildtype.bat
```diff
@@ -0,0 +1,16 @@
...
+set PATH=%2
+set MASK=%3
+set DESTINATION=%~4
+set REMOVE_FIRST_NUM=%5
+set PREFIX=%~6
+
+set WORKING_DIR=%cd%
+set ROUTINE=%cd%\mak\wroutine.bat
+
+C:\Windows\System32\forfiles.exe /p %PATH% /m %MASK% /s /c "cmd /c %ROUTINE% @RELPATH"
```
rainers, line 14: Don't use the absolute path here. I would expect c:\Windows\System32 to be in PATH. Even if not, c:\Windows is not guaranteed to be the installation path, %windir% contains it.

Geod24:
@thewilsonator : I [*explicitly* asked](https://github.com/dlang/druntime/pull/3024#issuecomment-611974564) to ping people familiar with the build system before mindlessly merging this.

For example, doing a bit of `git blame` on those files would have pointed you to [this PR](https://github.com/dlang/druntime/pull/2500) which explicitly reverts some of the changes that were done here.

CC @rainers .

rainers:
If we use wildcard processing, why not just use xcopy?

Does anyone remember what's special about the modules in the core/sync folder? The generated di files still contain platform specific code (the ctors are not reduced to the declaration !?), so why are they not simply copied aswell?

mak/wildtype.bat
denizzzka: At first I tried it without full path, but it works only in local console, but not in .bat files. I don't know why. If you can please fix it.

denizzzka:
> If we use wildcard processing, why not just use xcopy?

> Does anyone remember what's special about the modules in the core/sync folder? The generated di files still contain platform specific code (the ctors are not reduced to the declaration !?), so why are they not simply copied aswell?

I just replaced lines of code by same but generated. I have no idea why it was implemented by this way.

rainers:
I don't like the verbosity of the makefiles, but I'm afraid that this PR adds too many subtle details and brittle build steps. Walter is rather fond of the explicit list of commands.
If we have a host compiler with phobos available, most of mak\WINDOWS can be replaced by
```
copy: generated\copyimports.exe $(COPY)
	generated\copyimports.exe $(COPY)

generated\copyimports.exe: mak\copyimports.d
	$(DMD) -of=$@ $**
```
with this tiny D tool:
```D
module copyimports;
import std.array, std.file, std.path, std.stdio;

void main(string[] args)
{
    string importPath = absolutePath("import");
    string srcPath = absolutePath("src");
    foreach(impfile; args[1..$])
    {
        impfile = absolutePath(impfile);
        string srcfile = buildPath(srcPath, asRelativePath(impfile, importPath).array);
        if (std.file.exists(impfile))
            if (timeLastModified(impfile) >= timeLastModified(srcfile))
                continue;
        writeln("copying ", srcfile, " -> ", impfile);
        mkdirRecurse(dirName(impfile));
        std.file.copy(srcfile, impfile);
    }
}
```

denizzzka:
> Walter is rather fond of the explicit list of commands.

Me too, but this is was duplecate of COPY list

Can anyone implement @rainers proposal?

12345swordy:
https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/

denizzzka:
@12345swordy I am personally currently not have problem with Windows VM. But stil, this is alien system which is not my "goal".

I look at it like this:

By some coincidence, current mak/ WINDOWS affects all systems (any addition/deletion of .d file requires changes to it). I fixed it by this PR. 

It would be fair to accept the patch as is and then, people interested in the Windows platform will abble fix it or do something else. Non-Windows people will be able to do things as they see fit without paying attention to mak/WINDOWS.

Otherwise, this situation will continue rot for years.

rainers:
> Can anyone from WIndows World implement @rainers proposal?

See https://github.com/dlang/druntime/pull/3026

denizzzka:
#3026 concluded as better solution

PR #3026, opened 2020-04-11
use D tool to copy imports instead of explicit list of rules

rainers:
as an alternative to https://github.com/dlang/druntime/pull/3024

This adds a host dmd as an additional dependency, but it's usually available as dmd needs one to build anyway. It might also pave the way for using D to replace other makefile functionality.

In addition, it builds the tool into generated/windows hoping to move other build artifacts there, too.


denizzzka:
I faced with this error when forgot to implement creation of ./import/ directory

rainers:
> I faced with this error when forgot to implement creation of ./import/ directory

The main problem is to find a host dmd, as the CI scripts are either in a different repo (for azure-pipelines) or virtually immutable (auto-tester).

denizzzka:
auto-tester stalled?

rainers:
> auto-tester stalled?

Not sure what happened, the test have passed successfully. I amended a couple of comments to restart the build.

mak/copyimports.d
```diff
@@ -0,0 +1,30 @@
...
+void main(string[] args)
+{
+    // DigitalMars make passes long command line through this environment variable _CMDLINE
+    if (auto p = getenv("_CMDLINE"))
+        args = split(to!string(p));
+    else
+        args = args[1..$];
+
+    string importPath = absolutePath("import");
+    string srcPath = absolutePath("src");
```
Geod24, line 13: This assumes that the script is run from the root project, right ?

Wouldn't the following at the module level be more robust:
```D
immutable RootPath    = __FILE_FULL_PATH__.dirName();
immutable ImportPath   = RootPath.buildPath("import");
immutable SourceTarget = RootPath.buildPath("src");
```
?

mak/copyimports.d
```diff
@@ -0,0 +1,30 @@
+import std.array, std.conv, std.file, std.getopt, std.path, std.stdio;
```
Geod24, line 1: Perhaps add a module documentation explaining what this script do and why it's necessary ?
```suggestion
/**
 * Helper script to copy source files to the `import` directory
 *
 * When building druntime, `src` are copied to the `import`, directory, excluding modules from the `rt` package.
 * On POSIX this is handled by a Makefile rule, however DigitalMars' `make` on Windows does not support defining
 * custom rules, originally leading to an unhealthy duplication between mak/COPY and mak/WINDOWS,
 * which this script removes.
 */
 module copyimports;
 
 import std.array, std.conv, std.file, std.getopt, std.path, std.stdio;
```

rainers:
> but why not go all the way and copy things using a recursive descent instead of relying on `mak/WINDOWS` ?

dependency is on mak/COPY and mak/IMPORTS.

There are currently a number of exceptions, namely files in core/sync, that are treated differently (i.e. are processed to di file). No idea why that is the case, though.

Walter has objected the wildcard approach to avoid processing stale or temporary files for dmd, so I didn't try to change it here.

mak/copyimports.d
rainers: Thanks, added the header with a few minor modifications.

mak/copyimports.d
rainers: My initial version allowed the import and src path to be configured via getopt, allowing it to be used in similar situations The current version assumes a bit more about files being passed on the command line, while your proposed change fixes it to the current single use.

It seems a bit strange to pass relative paths on the command line, but then assume other src/import paths internally. I'm not too opinionated, though.

Geod24:
> Walter has objected the wildcard approach to avoid processing stale or temporary files for dmd, so I didn't try to change it here.

Could we then move those files into the tool itself, like it has been done for `build.d` ?
If you think it is too much change, let me know, I'm fine gong ahead with the current PR.

rainers:
> Could we then move those files into the tool itself, like it has been done for `build.d` ?
> If you think it is too much change, let me know, I'm fine gong ahead with the current PR.

I think that belongs into a separate PR as it should happen for all platforms. Otherwise we have new duplications in the build tool and mak/*.

kinke:
Is there anyone actually using these 'elaborately' copied imports on Windows? The DMD 2.091 package contains the full druntime and Phobos src trees, the .7z archive and the installer package too AFAICT, so at least the official Windows packages seem not to need any of this.

mak/copyimports.d
```diff
@@ -0,0 +1,40 @@
+/**
+ * Helper script to copy source files to the `import` directory
+ *
+ * When building druntime, files in `src` are copied to the `import` directory, excluding modules from the `rt` package.
```
kinke, line 4: Nit: 3 things are excluded, `rt.*`, `gc.*` and `test_runner.d`. And the .d -> .di conversions in `core.sync.*` as you've mentioned (LDC doesn't do that).

rainers:
> **Edit**: Ah, I hadn't realized the druntime src dir contains an additional `import` dir with the copies.

Yeah, the distribution is a bit of a mess. I think it should look more like LDC, not distributing sources of the compiler and the runtime, only "public" imports. For example, adding `import test.dub_stdx_allocator;` to your code does not raise an error.

mak/copyimports.d
rainers: Slightly adapted the comment, but left out the di-conversion as I plan to copy these files, too.

rainers:
Does anyone have an idea why the auto-tester keeps reporting "Pass: 8 Pending: 2" while the details show they are all done? Happens on other PRs, too, e.g. https://github.com/dlang/druntime/pull/3005. @braddr ?

denizzzka:
buildkite/druntime also failed

rainers:
> buildkite/druntime also failed

Looks pretty unrelated, too, as this only changes the Windows build.

MartinNowak:
This of course broke the release builder. Weird cause I thought we had at least the Windows part covered by now.

- doesn't support tool being build with ldc and x86 target (see #3075)
- does not forward HOST_DC make variable (HOST_DC is still used by dmd, will set HOST_DMD as well https://github.com/MartinNowak/installer/commit/4886449681cb73b5dca0018ab405f9f5df69f5b6)

Btw, would be nice if the next build system changes didn't land just before the scheduled beta ;).
https://dlang.org/changelog/release-schedule.html

rainers:
> This of course broke the release builder.

Sorry for the breakage.

> Weird cause I thought we had at least the Windows part covered by now.

The installer CI doesn't automatically run for changes in other repos. Maybe we can make it part of the nightly builds?


> * doesn't support tool being build with ldc and x86 target (see #3075)

With the latest LDC, setting LDC_VSDIR_FORCE=1 avoids this trouble as it auto-detects the appropriate environment unconditionally.


> * does not forward HOST_DC make variable

AFAICT that variable is also deprecated for DMD, see https://github.com/dlang/dmd/blob/05ad63263925b7b2730660f90361b650a3ace8e6/src/posix.mak#L75

CyberShadow:
This also broke Digger.

CyberShadow:
Another problem with this change is that it reports success despite that this step fails. This causes an error much later, when trying to build Phobos, manifesting as not finding `core.stdc.stddef` etc.

PR #3027, opened 2020-04-11
Add CPU_ALLOC_SIZE

kubo39:


kubo39:
Add missing thing: 3023

PR #3031, opened 2020-04-13
add pragma printf and scanf annotations to stdio.d

WalterBright:
Now that the annotations are ignored per https://github.com/dlang/dmd/pull/11030 we can add them.

### DIFF
--- a/src/core/stdc/stdio.d
+++ b/src/core/stdc/stdio.d
@@ -1166,61 +1166,73 @@ version (MinGW)
     // Prefer the MinGW versions over the MSVC ones, as the latter don't handle
     // reals at all.
     ///
+    pragma(printf)
     int __mingw_fprintf(FILE* stream, scope const char* format, ...);
     ///
     alias __mingw_fprintf fprintf;
 
     ///
+    pragma(scanf)
     int __mingw_fscanf(FILE* stream, scope const char* format, ...);
     ///
     alias __mingw_fscanf fscanf;
 
     ///
+    pragma(printf)
     int __mingw_sprintf(scope char* s, scope const char* format, ...);
     ///
     alias __mingw_sprintf sprintf;
 
     ///
+    pragma(scanf)
     int __mingw_sscanf(scope const char* s, scope const char* format, ...);
     ///
     alias __mingw_sscanf sscanf;
 
     ///
+    pragma(printf)
     int __mingw_vfprintf(FILE* stream, scope const char* format, va_list arg);
     ///
     alias __mingw_vfprintf vfprintf;
 
     ///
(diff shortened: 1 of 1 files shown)


WalterBright:
@jacob-carlborg looks like dlp-test-application needs an update?

```
Running ./dlp-test-application
--
  | /var/lib/buildkite-agent/builds/ci-agent-68ef3a9c-5d12-453f-8e0a-78cd3f822dd9-1/dlang/druntime/build/distribution/imports/core/stdc/stdio.d(1243,5): Error: unrecognized `pragma(printf)`
```

src/core/stdc/stdio.d
```diff
@@ -1367,13 +1395,17 @@ else version (CRuntime_Microsoft)
...
     ///
+    pragma(printf)
     int  snprintf(scope char* s, size_t n, scope const char* format, ...);
 
     ///
+    pragma(printf)
     int _vsnprintf(scope char* s, size_t n, scope const char* format, va_list arg);
     ///
+    pragma(printf)
     int  vsnprintf(scope char* s, size_t n, scope const char* format, va_list arg);
```
Geod24, line 1409: @WalterBright : In this case, how will the compiler know to check the 3rd argument ?

WalterBright:
@CyberShadow can you please look into the DAutoTest failure, please?

src/core/stdc/stdio.d
WalterBright: The last 2 parameters are always the same for the "v" functions.

Geod24: True, but you still rely on a convention. Is it the last `char*` argument, or the argument named format, or some other rule ?

jacob-carlborg:
> jacob-carlborg looks like dlp-test-application needs an update?

I'll update it.

src/core/stdc/stdio.d
WalterBright: The last one is `va_arg`, the second to last is `const(char)*`.

WalterBright:
Blocking https://github.com/dlang/dmd/pull/11034

jacob-carlborg:
> jacob-carlborg looks like dlp-test-application needs an update?

I've pushed an update and triggered a new build.

WalterBright:
@jacob-carlborg thank you!

WalterBright:
All green. Someone please approve!

WalterBright:
> I'm pretty certain you're going to end up needing to give the argument index / name as GCC does...

We'll see. I did leave the design open to adding that if it becomes necessary. In the meantime, I'm biased towards doing things as simple as possible.

PR #3032, opened 2020-04-14
Prepare for stack-allocated array literals passed to scope slice params

kinke:


schveiguy:
Shouldn't templates be inferred for scope-ness?

kinke:
I didn't want to dig into why it was necessary to overcome DIP1000 errors when making the testcase in the linked DMD PR `@safe`:
```
test\runnable\test20734.d(17): Error: reference to local variable __arrayliteral_on_stack3 assigned to non-scope parameter a calling core.internal.destruction.__ArrayDtor!(S).__ArrayDtor
```

MoonlightSentinel:
Another case where DIP1000 doesn't like expression temporarys ...

schveiguy:
OK yeah, template inference sometimes breaks down, I've seen it with other things. I wish dip1000 was easier to figure out.

Geod24:
This fixed issue https://issues.dlang.org/show_bug.cgi?id=18872

PR #3034, opened 2020-04-16
`dup`, `idup`, `keys`, `values`: if a type is known to have a postblit do not emit code for the non-postblit path and vice versa

n8sh:


kinke:
Hmm, this doesn't seem to consider copy constructors at all. This should probably be refactored to use `core.lifetime.emplace` for copy-construction, which seems to correctly handle both cases:
```D
import core.lifetime, core.stdc.stdio;

struct WithPostBlit
{
    int a = 42;
    this(int a) { this.a = a; }
    this(this) { printf("postblit: %d\n", a); }
}

struct WithCopyCtor
{
    int a = 42;
    this(int a) { this.a = a; }
    this(const ref WithCopyCtor rhs) { printf("copy ctor: %d => %d\n", a, rhs.a); }
}

void main()
{
    static void test(T)()
    {
    	T src = T(123);
        T dst = void;
        emplace(&dst, src);
    }

    test!WithPostBlit();
    test!WithCopyCtor();
}
```

n8sh:
That sounds right to me although making it work with `keys` and `values` would require more extensive changes since the arrays are built by the opaque functions `rt.aaA._aaKeys` and `rt.aaA._aaValues` which are non-templated and instead use `TypeInfo`. Touching up the arrays afterwards works for a postblit but not for a copy constructor.

Possible solutions:

* Expose the structure of built-in AAs so the `object` module can traverse them directly.
* Add a `copyConstructor(void* src, void* dst)` function to `TypeInfo` and change `rt.aaA._aaKeys` and `rt.aaA._aaValues` to call the copy constructor on their side.
* Add variants of `rt.aaA._aaKeys` and `rt.aaA._aaValues` that accept a copy constructor parameter and call it on their side.

kinke:
Yeah I know that it's not directly related to this PR, it just stood out when looking at this code. There are surely more places where the copy ctor is ignored, incl. the compiler itself.

thewilsonator:
```
src/object.d:2535:    static if(__traits(hasPostblit, Value))
posix.mak:440: recipe for target 'style_lint' failed
```
```diff
-    static if(__traits(hasPostblit, Value))
+    static if (__traits(hasPostblit, Value))
```

PR #3035, opened 2020-04-16
[stable] Fix code bloat regressions wrt. core.internal.lifetime.emplaceInitializer

kinke:
This partially reverts #2883 (introduced in 2.091). The problem is that for every template instantiation, i.e., emplaced type, 2 new structs are generated, each of which comes with its TypeInfo etc. And that includes all types, incl. primitive types and zero-initialized ones.

So as before that PR, try to handle the trivial cases first, and only use a (simplified) union version if that's not possible. [The union version in #2883 is probably an improvement over the previously used static constant.]

And as Suleyman mentioned that it's code copied from `object.d`, I went the other way and use `emplaceInitializer` in object.d to get rid of needless code duplication.

### DIFF
--- a/src/core/internal/lifetime.d
+++ b/src/core/internal/lifetime.d
@@ -89,8 +89,8 @@ Emplaces T.init.
 In contrast to `emplaceRef(chunk)`, there are no checks for disabled default
 constructors etc.
 +/
-nothrow pure @trusted
-void emplaceInitializer(T)(scope ref T chunk)
+void emplaceInitializer(T)(scope ref T chunk) nothrow pure @trusted
+    if (!is(T == const) && !is(T == immutable) && !is(T == inout))
 {
     import core.internal.traits : hasElaborateAssign;
 
@@ -98,30 +98,70 @@ void emplaceInitializer(T)(scope ref T chunk)
     {
         chunk = T.init;
     }
+    else static if (__traits(isZeroInit, T))
+    {
+        static if (is(T U == shared U))
+            alias Unshared = U;
+        else
+            alias Unshared = T;
+
+        import core.stdc.string : memset;
+        memset(cast(Unshared*) &chunk, 0, T.sizeof);
+    }
     else
     {
-        static if (__traits(isZeroInit, T))
-        {
-            static if (is(T U == shared U))
-                alias Unshared = U;
-            else
-                alias Unshared = T;
+        // emplace T.init (an rvalue) without extra variable (and according destruction)
+        alias RawBytes = void[T.sizeof];
 
-            import core.stdc.string : memset;
-            memset(cast(Unshared*) &chunk, 0, T.sizeof);
+        static union U
+        {
+            T dummy = T.init; // U.init corresponds to T.init
+            RawBytes data;
         }
-        else
+
+        *cast(RawBytes*) &chunk = U.init.data;
+    }
+}
+
+@safe unittest
+{
+    static void testInitializer(T)()
+    {
+        // mutable T
         {
-            // emplace T.init (an rvalue) without extra variable (and according destruction)
-            alias RawBytes = void[T.sizeof];
+            T dst = void;
+            emplaceInitializer(dst);
+            assert(dst is T.init);
+        }
 
-            static union U
-            {
-                T dummy = T.init; // U.init corresponds to T.init
-                RawBytes data;
-            }
+        // shared T
+        {
+            shared T dst = void;
+            emplaceInitializer(dst);
+            assert(dst is shared(T).init);
+        }
 
-            *cast(RawBytes*) &chunk = U.init.data;
+        // const T
+        {
+            const T dst = void;
+            static assert(!__traits(compiles, emplaceInitializer(dst)));
         }
     }
+
+    static struct ElaborateAndZero
+    {
+        int a;
+        this(this) {}
+    }
+
+    static struct ElaborateAndNonZero
+    {
+        int a = 42;
+        this(this) {}
+    }
+
+    testInitializer!int();
+    testInitializer!double();
+    testInitializer!ElaborateAndZero();
+    testInitializer!ElaborateAndNonZero();
 }
--- a/src/core/internal/lifetime.d
+++ b/src/core/internal/lifetime.d
@@ -84,23 +84,44 @@ if (is(UT == core.internal.traits.Unqual!UT))
     emplaceRef!(UT, UT)(chunk, forward!args);
 }
 
-//emplace helper functions
-private nothrow pure @trusted
+/+
+Emplaces T.init.
+In contrast to `emplaceRef(chunk)`, there are no checks for disabled default
+constructors etc.
++/
+nothrow pure @trusted
 void emplaceInitializer(T)(scope ref T chunk)
 {
-    // Emplace T.init.
-    // Previously, an immutable static and memcpy were used to hold an initializer.
-    // With improved unions, this is no longer needed.
-    union UntypedInit
-    {
-        T dummy;
-    }
-    static struct UntypedStorage
-    {
-        align(T.alignof) void[T.sizeof] dummy;
-    }
+    import core.internal.traits : hasElaborateAssign;
 
-    () @trusted {
-        *cast(UntypedStorage*) &chunk = cast(UntypedStorage) UntypedInit.init;
-    } ();
+    static if (!hasElaborateAssign!T && __traits(compiles, chunk = T.init))
+    {
+        chunk = T.init;
+    }
+    else
+    {
+        static if (__traits(isZeroInit, T))
+        {
+            static if (is(T U == shared U))
+                alias Unshared = U;
+            else
+                alias Unshared = T;
+
+            import core.stdc.string : memset;
+            memset(cast(Unshared*) &chunk, 0, T.sizeof);
+        }
+        else
+        {
+            // emplace T.init (an rvalue) without extra variable (and according destruction)
+            alias RawBytes = void[T.sizeof];
+
+            static union U
+            {
+                T dummy = T.init; // U.init corresponds to T.init
+                RawBytes data;
+            }
+
+            *cast(RawBytes*) &chunk = U.init.data;
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


src/core/internal/lifetime.d
```diff
@@ -84,23 +84,87 @@ if (is(UT == core.internal.traits.Unqual!UT))
...
+
+    static if (!hasElaborateAssign!T && __traits(compiles, chunk = T.init))
     {
-        T dummy;
+        chunk = T.init;
     }
-    static struct UntypedStorage
+    else
+    {
+        static if (__traits(isZeroInit, T))
```
thewilsonator, line 103: `else static if (__traits(isZeroInit, T))`

src/core/internal/lifetime.d
```diff
@@ -84,23 +84,87 @@ if (is(UT == core.internal.traits.Unqual!UT))
...
+        }
+        else
+        {
+            // emplace T.init (an rvalue) without extra variable (and according destruction)
+            alias RawBytes = void[T.sizeof];
+
+            static union U
+            {
+                T dummy = T.init; // U.init corresponds to T.init
+                RawBytes data;
```
thewilsonator, line 121: I take it this solves the need for `align(T.alignof)`?

src/core/internal/lifetime.d
```diff
@@ -84,23 +84,87 @@ if (is(UT == core.internal.traits.Unqual!UT))
...
+    {
+        static if (__traits(isZeroInit, T))
+        {
+            static if (is(T U == shared U))
+                alias Unshared = U;
+            else
+                alias Unshared = T;
+
+            import core.stdc.string : memset;
+            memset(cast(Unshared*) &chunk, 0, T.sizeof);
```
MoonlightSentinel, line 111: Does this need to cast to the explicit type (as `memset` takes a `void*` anyway)?

```suggestion
            memset(cast(void*) &chunk, 0, T.sizeof);
```

kinke: It doesn't need to, but I consider it good practice. As soon as someone has the great idea to drop the template constraints, a cast to `void*` will allow non-mutable types too.

src/core/internal/lifetime.d
kinke: Yes, that's one of the advantages.

src/core/internal/lifetime.d
kinke: I'd rather leave as-is; e.g., in case a `memset` is slower than the union version for DMD, one could add an `enum preferMemset = ...`.

kinke:
Some more context - this:
```D
import core.lifetime;

void foo()
{
    static void testInitializer(T)()
    {
        T dst = void;
        emplace(&dst);
        assert(dst is T.init);
    }

    static struct ElaborateAndZero
    {
        int a;
        this(this) {}
    }

    static struct ElaborateAndNonZero
    {
        int a = 42;
        this(this) {}
    }

    testInitializer!int();
    testInitializer!double();
    testInitializer!ElaborateAndZero();
    testInitializer!ElaborateAndNonZero();
}
```

yields the following number of lines in the unoptimized textual LLVM IR file (`-output-ll`):
- LDC v1.20 (v2.090): 338
- LDC master (v2.091): 509 (+50%)
- LDC with this PR: 353

Codegen pessimization for trivial `double` - current master:
```LL
define weak_odr x86_vectorcallcc void @"\01_D4core8internal8lifetime__T18emplaceInitializerTdZQxFNaNbNiNeKdZv"(double* dereferenceable(8) %chunk) #0 comdat {
  %.frame = alloca %nest.emplaceInitializer.0, align 8 ; [#uses = 2, size/byte = 8]
  %chunk1 = getelementptr inbounds %nest.emplaceInitializer.0, %nest.emplaceInitializer.0* %.frame, i32 0, i32 0 ; [#uses = 1, type = double**]
  store double* %chunk, double** %chunk1, align 8
  %1 = bitcast %nest.emplaceInitializer.0* %.frame to i8* ; [#uses = 1]
  %2 = insertvalue { i8*, void (i8*)* } undef, i8* %1, 0 ; [#uses = 1]
  %.func = insertvalue { i8*, void (i8*)* } %2, void (i8*)* @"\01_D4core8internal8lifetime__T18emplaceInitializerTdZQxFNaNbNeKdZ9__lambda4MFNaNbNiNeZv", 1 ; [#uses = 2]
  %.funcptr = extractvalue { i8*, void (i8*)* } %.func, 1 ; [#uses = 1]
  %.ptr = extractvalue { i8*, void (i8*)* } %.func, 0 ; [#uses = 1]
  call x86_vectorcallcc void %.funcptr(i8* nonnull %.ptr)
  ret void
}

define weak_odr x86_vectorcallcc void @"\01_D4core8internal8lifetime__T18emplaceInitializerTdZQxFNaNbNeKdZ9__lambda4MFNaNbNiNeZv"(i8* nonnull %.nest_arg) #0 comdat {
  %.structliteral = alloca %"core.internal.lifetime.emplaceInitializer!double.emplaceInitializer.UntypedInit", align 8 ; [#uses = 2, size/byte = 8]
  %1 = bitcast i8* %.nest_arg to %nest.emplaceInitializer.0* ; [#uses = 1]
  %chunk = getelementptr inbounds %nest.emplaceInitializer.0, %nest.emplaceInitializer.0* %1, i32 0, i32 0 ; [#uses = 1, type = double**]
  %2 = load double*, double** %chunk, align 8     ; [#uses = 1]
  %3 = bitcast double* %2 to %"core.internal.lifetime.emplaceInitializer!double.emplaceInitializer.UntypedStorage"* ; [#uses = 1]
  %4 = getelementptr inbounds %"core.internal.lifetime.emplaceInitializer!double.emplaceInitializer.UntypedInit", %"core.internal.lifetime.emplaceInitializer!double.emplaceInitializer.UntypedInit"* %.structliteral, i32 0, i32 0 ; [#uses = 1, type = double*]
  store double 0x7FF8000000000000, double* %4
  %.structliteral.repaint = bitcast %"core.internal.lifetime.emplaceInitializer!double.emplaceInitializer.UntypedInit"* %.structliteral to %"core.internal.lifetime.emplaceInitializer!double.emplaceInitializer.UntypedStorage"* ; [#uses = 1]
  %5 = bitcast %"core.internal.lifetime.emplaceInitializer!double.emplaceInitializer.UntypedStorage"* %3 to i8* ; [#uses = 1]
  %6 = bitcast %"core.internal.lifetime.emplaceInitializer!double.emplaceInitializer.UntypedStorage"* %.structliteral.repaint to i8* ; [#uses = 1]
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* align 1 %5, i8* align 1 %6, i64 8, i1 false)
  ret void
}
```
vs. this now:
```LL
define weak_odr x86_vectorcallcc void @"\01_D4core8internal8lifetime__T18emplaceInitializerTdZQxFNaNbNiNeKdZv"(double* dereferenceable(8) %chunk) #0 comdat {
  store double 0x7FF8000000000000, double* %chunk
  ret void
}
```

src/core/internal/lifetime.d
thewilsonator: I meant to reduce nesting.

kinke: [I know.]

kinke:
Not allowing `const` also makes this terrible thing no longer compile (& pass):
```D
import core.lifetime;

struct S { int a; }

void main() @safe
{
    const s = S(42);
    emplace(&s); // needs -dip1000
    destroy(s);
    assert(s.a == 0);
}
```

n8sh:
Not allowing `const` is the right thing to do (plus anything that has a field that is `const` or a field that recursively meets this condition), but I am wary of breaking working code in a stable release even if that code arguably never should have worked.

kinke:
> even if that code arguably never should have worked

I don't have the faintest idea how someone could argue that that should have worked. It's all friggin `@safe` too.

thewilsonator:
```
git clone -b v2.091.1 --depth=1 https://github.com/dlang/druntime .generated/druntime-2.091.1
Cloning into '.generated/druntime-2.091.1'...
warning: Could not find remote branch v2.091.1 to clone.
fatal: Remote branch v2.091.1 not found in upstream origin
posix.mak:595: recipe for target '.generated/druntime-2.091.1' failed
```

Please force push to restart the doc tester.

n8sh:
Would it be a hardship to limit the PR targeting `stable` to just fixing the regression and put the safety-minded constraints in a PR targeting `master`?

>I don't have the faintest idea how someone could argue that that should have worked. It's all friggin @safe too.

The larger problem is that the documentation of `emplace` requires that the destination be uninitialized memory but that's not enforced by the compiler.

kinke:
I really don't want to argue about this; I guess we'll just have to agree to disagree on this. It doesn't matter anyway, as 2.091.1 has already been tagged, so this won't make it before v2.092 anyway. I'll probably cherry-pick this for LDC v1.21 final though, unless there are more people thinking that disallowing `emplace/destroy(immutable stuff)` might be a hassle of a breaking change.

n8sh:
>unless there are more people thinking that disallowing `emplace/destroy(immutable stuff)` might be a hassle of a breaking change

Accepting that there is no longer an argument (please feel no obligation to read further) I want to defend the idea that I'm not insane.

We have a reasonable-looking function...
```d
T[] allocateDefaultInitializedArray(T)(size_t length)
{
    import core.checkedint : mulu;
    import core.lifetime : emplace;
    import core.memory : GC;

    if (length == 0)
        return null;
    bool overflow;
    const allocSize = mulu(length, T.sizeof, overflow);
    if (overflow)
        onOutOfMemoryError;
    auto result = (() @trusted => (cast(T*) GC.malloc(allocSize))[0 .. length])();
    foreach (ref a; result)
        emplace(&a); // @safe if compiled with -dip1000
    return result;
}
```
and it gets called like this...
```d
void main()
{
    // The guy who wrote this comes from a functional programming background
    // and believes anything that can be immutable should be immutable.
    immutable array = allocateDefaultInitializedArray!(immutable float)(2);
    foreach (a; array)
        assert(a is float.init);
}
```

It is easy to rewrite `allocateDefaultInitializedArray` so this works again. Cast the pointer to `Unqual!T*` instead of `T*`.
Is it okay to break this code? I think so.
Is it okay to break it in a point release? That's now a moot point.

PR #3036, opened 2020-04-18
 Musl: add more stdlib definitions and enable HaveMemstream

Cogitri:


src/core/sys/posix/stdlib.d
```diff
@@ -575,9 +575,45 @@ else version (CRuntime_Bionic)
...
+    void   setkey(const scope char*);
+    char*  setstate(const scope char*);
+    void   srand48(c_long);
+    void   srandom(uint);
+    int    unlockpt(int);
+
+  static if ( __USE_LARGEFILE64 )
+  {
+    int    mkstemp64(char*);
+    alias  mkstemp64 mkstemp;
```
Geod24, line 611: This `alias` means that any call to `mkstemp` will link in `mkstemp64`. That's uncommon in the C bindings and rarely do what you want. Is that intended (e.g. `mkstemp` is a define) ?

src/core/sys/posix/stdio.d
```diff
@@ -332,6 +332,8 @@ else version (OpenBSD)                      // as of OpenBSD 5.4
     version = HaveMemstream;
 else version (CRuntime_UClibc)
     version = HaveMemstream;
+else version(CRuntime_Musl)
```
Geod24, line 335:
```suggestion
else version (CRuntime_Musl)
```

src/core/sys/posix/stdlib.d
Cogitri: Yup, it's a define:

```
#if defined(_LARGEFILE64_SOURCE) || defined(_GNU_SOURCE)
#define mkstemp64 mkstemp
#define mkostemp64 mkostemp
#if defined(_GNU_SOURCE) || defined(_BSD_SOURCE)
#define mkstemps64 mkstemps
#define mkostemps64 mkostemps
#endif
#endif
```

src/core/sys/posix/stdio.d
Cogitri: Ah thanks. Is there some standard formatter for the druntime? dfmt didn't like the alignment by variable name in this file.

Geod24: `make -f posix.mak style`

Cogitri: Will do that next time, thanks :)

src/core/sys/posix/stdio.d
```diff
@@ -278,6 +309,30 @@ else version (CRuntime_UClibc)
     off_t ftello(FILE*);
   }
 }
+else version(CRuntime_Musl)
```
Geod24, line 312:
```suggestion
else version (CRuntime_Musl)
```

That's why the CI is red BTW

src/core/sys/posix/stdio.d
```diff
@@ -332,6 +387,8 @@ else version (OpenBSD)                      // as of OpenBSD 5.4
     version = HaveMemstream;
 else version (CRuntime_UClibc)
     version = HaveMemstream;
+else version (CRuntime_Musl)
+    version = HaveMemstream;                // http://git.musl-libc.org/cgit/musl/commit/src/stdio/open_memstream.c?id=b158b32a44d56ef20407d4285b58180447ffff1f
```
Geod24, line 391:
```suggestion
// http://git.musl-libc.org/cgit/musl/commit/src/stdio/open_memstream.c?id=b158b32a44d56ef20407d4285b58180447ffff1f
else version (CRuntime_Musl)
    version = HaveMemstream;
```

The style checker also complains about lines > 120 chars

PR #3037, opened 2020-04-18
Add method TypeInfo.argTypes() for non-Apple AArch64 targets

kinke:
The AAPCS64 varargs implementation isn't trivial. As for the SysV x64 ABI, prepare for a `core.stdc.stdarg` implementation making use of `TypeInfo.argTypes()`.

[This is a tiny piece of ldc-developers/ldc#3393.]

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -39,6 +39,23 @@ alias dstring = immutable(dchar)[];
 version (D_ObjectiveC) public import core.attribute : selector;
 version (Posix) public import core.attribute : gnuAbiTag;
 
+// Some ABIs use a complex varargs implementation requiring TypeInfo.argTypes().
+version (X86_64)
+{
+    version (DigitalMars) version = WithArgTypes;
+    else version (Windows) { /* no need for Win64 ABI */ }
+    else version = WithArgTypes;
+}
+version (AArch64)
+{
+    // Apple uses a trivial varargs implementation
+    version (OSX) {}
+    else version (iOS) {}
+    else version (TVOS){}
+    else version (WatchOS) {}
+    else version = WithArgTypes;
+}
+
 /**
  * All D class objects inherit from Object.
  */
@@ -390,7 +407,7 @@ class TypeInfo
     /** Return internal info on arguments fitting into 8byte.
      * See X86-64 ABI 3.2.3
      */
-    version (X86_64) int argTypes(out TypeInfo arg1, out TypeInfo arg2) @safe nothrow
+    version (WithArgTypes) int argTypes(out TypeInfo arg1, out TypeInfo arg2) @safe nothrow
     {
         arg1 = this;
         return 0;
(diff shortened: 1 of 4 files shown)


kinke:
Pinging @ibuclaw, as this needs compiler adaptations when emitting the TypeInfos.

ibuclaw:
I try to pretend that this va_arg function doesn't exist, nor does the trait that it depends on (and as such, argTypes is still unimplented in gdc).

As you are broadening the addition of the argTypes tuple to more targets, I think that there should be two things done:

1. Add a Target hook in the compiler which returns true if argTypes is implemented for the target.

2. Add a new predefined version condition to export this to the user code. This would ensure that there is no disconnect between runtime and library.

kinke:
> I try to pretend that this va_arg function doesn't exist

Heh, no complaints from users? - The TypeInfo-based va_arg is horrible; I think it's a leftover from the D1 days. Getting rid of it would greatly simplify adding support for ABIs with non-trivial varargs.

ibuclaw:
> Heh, no complaints from users? - The TypeInfo-based va_arg is horrible; I think it's a leftover from the D1 days. Getting rid of it would greatly simplify adding support for ABIs with non-trivial varargs.

Yes, it is a leftover.  Back when std.format was va_list based.  Nowadays it's all been converted to variadic templates, and no one AFAIK no one as any justifiable use of __traits(argTypes) that they can't do using better alternatives. :-)

ibuclaw:
@thewilsonator I'll repeat what I said again - as this is the proper route needed to support this.

As you are broadening the addition of the argTypes tuple to more targets, I think that there should be two things done:

1. Add a Target hook in the compiler which returns true if argTypes is implemented for the target.

2. Add a new predefined version condition to export this to the user code. This would ensure that there is no disconnect between runtime and library.

kinke:
A `Target` hook would be elegant, I agree, but require a compiler extension first. LDC doesn't need it to check for a disconnect, as that's handled when emitting a `TypeInfo_Struct` (checks the number of fields, which is +2 if the `argTypes` are supported). As to exposing that to the user via predefined version, I don't think that's needed, because it only ever matters for `TypeInfo`, and in the extremely unlikely case the user needs that method, it can be trivially checked as shown here.

ibuclaw:
> LDC doesn't need it to check for a disconnect, as that's handled when emitting a `TypeInfo_Struct` (checks the number of fields, which is +2 if the `argTypes` are supported). 

I could argue that counting the number of fields is quite an arbitrary check. But meh.

Having the logic in druntime means the compiler has no say in the matter over which targets should or not have these extra fields.

kinke:
> Having the logic in druntime means the compiler has no say in the matter over which targets should or not have these extra fields.

Yep, but that's the unchanged status quo, just that non-DMD newly doesn't expose them for Win64 but newly does for non-Apple AArch64. If I had known GDC doesn't support them at the time I wrote this, I'd have prepended a `version (GNU) enum withArgTypes = false`.

PR #3038, opened 2020-04-18
Generalize meaning of 2nd bit of TypeInfo.flags to pass-in-SIMD-register(s)

kinke:
Instead of x86-only XMM. Designate floats, doubles and non-x87 reals as well as all vector types as being passed in SIMD registers, if available.

[This is a tiny piece of ldc-developers/ldc#3393.]

### DIFF
--- a/src/rt/typeinfo/ti_double.d
+++ b/src/rt/typeinfo/ti_double.d
@@ -67,12 +67,6 @@ class TypeInfo_d : TypeInfo
 
     override @property immutable(void)* rtInfo() nothrow pure const @safe { return rtinfoNoPointers; }
 
-    version (Windows)
-    {
-    }
-    else version (X86_64)
-    {
-        // 2 means arg to function is passed in XMM registers
-        override @property uint flags() const { return 2; }
-    }
+    // passed in SIMD register
+    override @property uint flags() const { return 2; }
 }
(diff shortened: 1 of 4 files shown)


src/rt/typeinfo/ti_real.d
```diff
@@ -66,4 +66,10 @@ class TypeInfo_e : TypeInfo
     }
 
     override @property immutable(void)* rtInfo() nothrow pure const @safe { return rtinfoNoPointers; }
+
+    static if (real.mant_dig != 64) // exclude 80-bit X87
```
thewilsonator, line 70: The comment doesn't seem to match the logic here.

n8sh: [80-bit extended precision floating point numbers have 64 bits of mantissa, 15 bits of exponent, and a sign bit.](https://en.wikipedia.org/wiki/Extended_precision#x86_extended_precision_format)

thewilsonator: Ah, derp I read `sizeof` for some reason.

src/rt/typeinfo/ti_double.d
```diff
@@ -67,12 +67,6 @@ class TypeInfo_d : TypeInfo
 
     override @property immutable(void)* rtInfo() nothrow pure const @safe { return rtinfoNoPointers; }
 
-    version (Windows)
-    {
-    }
```
n8sh, line 72: Why is the special case for Windows no longer necessary?

kinke: It never was, as floats and doubles are passed in a XMM on Win64 too. On 32-bit x86, XMMs aren't used at all.

PR #3040, opened 2020-04-18
[stable] Refactor core.stdc.stdarg & add support for ARM, PowerPC and MIPS (AArch64: Apple only)

kinke:


src/core/stdc/stdarg.d
```diff
@@ -202,12 +183,6 @@ else version (X86_64)
         __va_list va;
     }
 
-    /*
-     * Making it an array of 1 causes va_list to be passed as a pointer in
-     * function argument lists
-     */
-    alias va_list = __va_list*;
```
ibuclaw, line 209: I don't think this should be moved to internal.

src/core/stdc/stdarg.d
```diff
@@ -218,261 +193,6 @@ else version (X86_64)
         return a;
     }
 
-    ///
-    void va_arg(T)(va_list apx, ref T parmn)
-    {
-        __va_list* ap = cast(__va_list*) apx;
```
ibuclaw, line 224: Could this instead be:
```
///
void va_arg(T)(va_list apx, ref T parmn)
{
    __va_arg_internal!T(apx, parmn);
}
```
With main body of implementation moved as appropriate?

kinke: This is marked as WIP because it's not finished yet, there's a 2nd restructuring upcoming, just waiting for this first commit to turn green.

ibuclaw: It's never too early to get suggestions in. :-)

kinke: The first commit wasn't going to get green due to something mitigated in the 2nd, so pushed earlier. - I suppose GDC has a bunch of `va_arg` extensions similar to LDC; can you give me a link so that I can cross-check and try to upstream at least some more platforms?

kinke:
I've added LDC versions (slightly revised, but mostly untested) for ARM, PowerPC and MIPS, primarily to show where I'd like to go with this (full AArch64 support still requires an extra module to be added in another PR). Big-endian support might be lacking for some archs; I didn't look up the docs.

I can move the additions to another PR later if that helps review-ability.

src/core/stdc/stdarg.d
ibuclaw: It's all using built-ins, there's a version(GNU) block with all the signatures, but no bodies.

kinke: Oh wow, that's convenient.

kinke:
Iain, how much sense does the upstreaming make then if you don't need it? It's not an awful lot of code, but none of it is CI-tested (incl. LDC)...

src/core/stdc/stdarg.d
ibuclaw: > Oh wow, that's convenient.

- `__builtin_va_list` is constructed during language init from a gcc tree node made available by backend init (it is then inserted into the gcc.builtins module)
- va_copy and va_end are aliasing `__builtin_` symbols from gcc.builtins.
- va_start and va_arg are the only specially handled intrinsics, the calls to the template instantiations are matched and replaced with `__builtin_va_start` and `VA_ARG_EXPR` respectively.

The target takes care of providing the actual implementations, meaning that va_args just works on every platform and I do sod all in order to make that happen. :-)

The only omission would be the rtti based va_arg function, which relies on either the language parts doing something target specific (bad abstraction/separation of concerns), or implementing a language specific backend hook that duplicates/fleshes out the internal type -> mode inference done when backend lowers va_arg (duplication is a code smell at best, and fleshing out internals needs approval from each maintainer for each backend  that we need this, which may hit upon strong resistance).

When I spoke to the C frontend (and global gcc) maintainer about this, his immediate reaction was that I/we should be using libffi, and if it doesn't fit our case, then we can fix it so that it can be used for our purposes.

ibuclaw:
> Iain, how much sense does the upstreaming (of additional targets) make then if you don't need it? It's not an awful lot of code, but none of it is CI-tested (incl. LDC)...

I'm OK with this, as they are useful to have anyway.

It may even be a good excuse to upstream / align up my gdc local changes as well, so this module can truly be shared between all compilers. (https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libphobos/libdruntime/core/stdc/stdarg.d;h=586fe20d9913dda749b92a9eadd5d0de0787678e;hb=refs/heads/master#l22)

I'm pretty sure that there wouldn't be much to do, so long as the general skeleton is there, and available to all targets.

src/core/stdc/stdarg.d
```diff
@@ -34,93 +81,75 @@ unittest
...
+        void* __stack;
+        void* __gr_top;
+        void* __vr_top;
+        int __gr_offs;
+        int __vr_offs;
     }
+}
+else
+{
+    alias va_list = char*;
```
ibuclaw, line 119: FYI, the 'default' va_list type is `void*` for targets that don't have any specific abi.

src/core/stdc/stdarg.d
```diff
@@ -34,93 +81,75 @@ unittest
...
-     */
-    void va_arg()(ref va_list ap, TypeInfo ti, void* parmn)
+/**
+ * Initialize ap.
+ * parmn should be the last named parameter;
+ * for DMD and non-Windows x86_64 targets, it should be __va_argsave.
+ */
+version (X86)
+{
+    void va_start(T)(out va_list ap, ref T parmn)
```
ibuclaw, line 130: Can the version condition not be put inside the function?

src/core/stdc/stdarg.d
```diff
@@ -129,25 +158,107 @@ else version (Windows) // Win64
...
+    {
+        auto p = ap;
+        version (BigEndian)
+            static if (T.sizeof < size_t.sizeof)
+                p += size_t.sizeof - T.sizeof;
+        parmn = *cast(T*) p;
+        ap += T.sizeof.alignUp;
+    }
+    else
+        static assert(0, "Unsupported platform");
```
ibuclaw, line 241: FYI, the 'default' va_arg is that arguments are passed on the stack for any target that doesn't have its own specific abi.

src/core/stdc/stdarg.d
kinke: It's an intrinsic for DMD+x86_64 and LDC.

src/core/stdc/stdarg.d
kinke: Would you prefer such a 'dangerous' default implementation over forcing the guys to look up the ABI docs when porting a new arch?

src/core/stdc/stdarg.d
ibuclaw: If it is an intrinsic, that still means that the body _should_ be ignored.

src/core/stdc/stdarg.d
kinke: LLVM doesn't make it that easy, unfortunately.

- I think we can safely agree that the TypeInfo-based `va_arg` should be deprecated; there was a single failure in the whole testsuite as long as that was missing for LDC and AArch64 (and that was supposed to test something completely different IIRC), so it's pretty much untested anyway.
- As for the regular `va_arg(T)` template, I think DMD's `argTypes` solution for complex vararg implementations isn't bad, acting as a bridge between compiler and druntime (it's not a trait, but the special `is(T ArgTypes == __argTypes)` expression), in order to have a more legible implementation of `va_arg` in druntime rather than having to emit the equivalent in each compiler via intrinsic. Unless of course the intrinsic/builtin is already handled as in your lucky case :), then I wouldn't use the argTypes at all either.

ibuclaw: > LLVM doesn't make it that easy, unfortunately.
> 
>  **edit**: yeah well and D-style variadic functions

There's nothing inherently wrong with D style variadics, type inference using only RTTI is just a pain if you want to support structs.

kinke: I just imagine them to be a pain without according TypeInfo-based `va_arg`, even if just using primitive types. I might have written some of these variadics at some point for a test, but never had any real need for them.

src/core/stdc/stdarg.d
ibuclaw: It would appear that it is not an intrinsic on x86 dmd, that should be fixed, but it won't block this pr.

src/core/stdc/stdarg.d
ibuclaw: A big warning sign might suffice. Out of scope for this pr though, but the main primary targets should be trivial to add.

Note, you have no static assert for the declaration of va_list type.

ibuclaw:
Nothing major to object on the inclusion of this.  The only litmus test that should pass is that it should integrate with what gdc does too with no extra changes to this module.

https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libphobos/libdruntime/core/stdc/stdarg.d;h=586fe20d9913dda749b92a9eadd5d0de0787678e;hb=refs/heads/master#l22

va_copy and va_end can be made compiler recognised intrinsics, leaving only the declaration of va_list remaining, perhaps by adding version(GNU) as the first condition in the place where it is defined.

kinke:
@ibuclaw: I've pushed a commit attempting to add the GDC/LDC specifics (there's another Win64 special case for LDC that isn't included yet). I'm not sure if it's sufficient for GDC, i.e., whether it really ignores the bodies of functions you detect as intrinsics. - I think the legibility is still okay-ish.

ibuclaw:
> I'm not sure if it's sufficient for GDC, i.e., whether it really ignores the bodies of functions you detect as intrinsics. - I think the legibility is still okay-ish.

1. In `interpretFunction`, `evaluateIfBuiltin` is checked before the function body is interpreted.  In the case of va_args, it is not possible to do this at CTFE, so errors about not being able to evaluate all around.
2. Front-end inlining is stubbed out/disabled in gdc (`inlineCopy(e) => return e.copy()`).
3. In the code generation of CallExp, intrinsic detection and expansion is done.
4. When compiling the module, if there's a body, it will get emitted, even though all routes to call it (except through a pointer) are elided by the compiler.

kinke:
I.e., not working because of point 4, as I didn't want to import the internal x64 module with GDC.

src/core/stdc/stdarg.d
```diff
@@ -34,120 +85,183 @@ unittest
...
 
-version (X86)
+
+/**
+ * The argument pointer type.
+ */
+version (GNU)
+{
+    alias va_list = __gnuc_va_list;
+    alias __gnuc_va_list = __builtin_va_list;
```
ibuclaw, line 95: I know order of evaluation doesn't matter, but swap the two around?

kinke: It's for the ddoc comment (one for all version branches).

ibuclaw:
> I.e., not working because of point 4, as I didn't want to import the internal x64 module with GDC.

Actually, I forgot about the fact that semantic analysis is done on the body of these instantiated templates anyway.

This means that all those static asserts will be a new pain for GDC where support already exists everywhere.

src/core/stdc/stdarg.d
ibuclaw: Ah, ok then.

ibuclaw:
Though I now see that between me writing the last message and hitting send, you've already fixed that concern.

ibuclaw:
Another thought that comes to mind is the the `TypeInfo` overload for `va_arg` should instead live in `core.varargs`, as there is no equivalent "C" std macro.

kinke:
I agree, transplanted that to `core.vararg`.

kinke:
I've moved the `va_arg` implementation from the version taking the destination by ref to the version returning a copy, which should be a prerequisite for further `va_arg` fixes.

src/core/stdc/stdarg.d
```diff
@@ -34,466 +80,297 @@ unittest
...
-                            }
-                            p = ap.stack_args;
-                            ap.stack_args += 8;
-                        }
-                    }
-                    auto sz = ti.tsize - 8;
-                    (parmn + 8)[0..sz] = p[0..sz];
-                }
+                const arg = va_arg!T(_argptr);
+                //FIXME: assert(arg.value == i);
```
kinke, line 343: This remaining issue here is IMO due to a DMD codegen bug when emitting the varargs. It doesn't work for LDC either; to be honest, I didn't expect non-PODs to be allowed as varargs (only those with dtors are currently disallowed).

kinke:
The non-Apple AArch64 implementation is another 200 lines, so I'll save that for another PR.

src/core/stdc/stdarg.d
ibuclaw: And now postblits and copyctors are disallowed too.

Nested structs are also passed by invisible ref in gdc, but that doesn't affect va_arg, because as soon as the type symbol escapes, it's no longer "enclosed".

ibuclaw:
By the way, I've just noticed that this targets stable, shouldn't refactorings be on master instead?

src/core/internal/vararg/sysv_x64.d
```diff
@@ -0,0 +1,304 @@
...
+    {
+        static assert(false, "not a valid argument type for va_arg");
+    }
+}
+
+///
+void va_arg()(va_list ap, TypeInfo ti, void* parmn)
+{
+    TypeInfo arg1, arg2;
+    if (!ti.argTypes(arg1, arg2))
```
thewilsonator, line 198: Please invert the condition check and move the assert here s.t. the function has less indentation.

src/core/internal/vararg/sysv_x64.d
```diff
@@ -0,0 +1,304 @@
...
+    {
+        bool inXMMregister(TypeInfo arg) pure nothrow @safe
+        {
+            return (arg.flags & 2) != 0;
+        }
+
+        TypeInfo_Vector v1 = arg1 ? cast(TypeInfo_Vector) arg1 : null;
+        if (arg1 && (arg1.tsize <= 8 || v1))
```
thewilsonator, line 206: ditto here, the else branch would make a nice early return.

src/core/internal/vararg/sysv_x64.d
```diff
@@ -0,0 +1,304 @@
...
+                else
+                {
+                    p = ap.stack_args;
+                    ap.stack_args += 8;
+                    stack = true;
+                }
+            }
+            parmn[0..tsize] = p[0..tsize];
+
+            if (arg2)
```
thewilsonator, line 244: then this could become 
```
if (!arg2)
    return;
```

kinke: I wasn't trying to refactor/touch this x64 implementation at all. Especially not for these kind of aesthetics.

kinke:
> By the way, I've just noticed that this targets stable, shouldn't refactorings be on master instead?

I need/want this in 2.092, along with full AArch64 ABI support in LDC. As the master->stable merge has already happened a few days after 2.091.1, I take it I need to target stable to prevent this from taking 2+ months to land in a release.

kinke:
> You don't need the newly added unittests, I've made postblit and copyctors an error in the compiler - same as dtors.

Right, merged without any discussion. - Well I don't care, but getting LDC to support them would probably have been easy, and the move from ref-va_arg to return-va_arg fixed the copy ctor issues (was an assignment before that, not a construction).

kinke:
Unittests removed again.

ibuclaw:
> > You don't need the newly added unittests, I've made postblit and copyctors an error in the compiler - same as dtors.
> 
> Right, merged without any discussion. - Well I don't care, but getting LDC to support them would probably have been easy (=> leaving DMD's backend as only blocker), and the move from ref-va_arg to return-va_arg fixed the copy ctor issues (was an assignment before that, not a construction).

If there are clear fixes, then they should be opened back up, but that is otherwise for another pr.

src/core/internal/vararg/sysv_x64.d
ibuclaw: Yeah, don't break one of the Walter rules, changing code whilst at the same time moving it.

Improvements can go in a follow up pr

kinke:
[Merging now before beta1 is probably no big deal, but it might be as soon as the 2.092 `stable` cycle has started with a release.]

ibuclaw:
> Let's go ahead for now, any problems found using this with other compilers can go on master 

Also, all mentioned nits and refactorings to go on master in follow ups please.

ibuclaw:
@kinke just waiting for the WIP label to be removed. :-)

kinke:
> just waiting for the WIP label to be removed

Looks like that's not in my power (no dlang member).

thewilsonator:
Done.

PR #3042, opened 2020-04-19
Change LCG in garbage collector treap to use full 64 bits of state instead of 48

n8sh:
`rt.util.random.Rand48` is an LCG that holds its state in a 64-bit long but only makes use of the low 48 bits. Replacing it with an LCG that makes use of the full 64 bits is a strict improvement as it takes no additional space and is not slower.

This PR renames `Rand48` to "Rand" so the name will not be misleading. "Rand64" was not chosen because it could give the misimpression that the output is 64 bits.

### DIFF
--- a/src/rt/util/container/treap.d
+++ b/src/rt/util/container/treap.d
@@ -29,8 +29,8 @@ nothrow:
 
     void initialize(ulong randSeed)
     {
-        Rand48 _rand48 = { randSeed };
-        rand48 = _rand48;
+        Rand _rand = { randSeed };
+        rand = _rand;
     }
 
     void insert(E element) @nogc
@@ -109,13 +109,13 @@ nothrow:
 
 private:
     Node* root;
-    Rand48 rand48;
+    Rand rand;
 
     Node* allocNode(E element) @nogc
     {
         Node* node = cast(Node*)common.xmalloc(Node.sizeof);
         node.left = node.right = null;
-        node.priority = rand48();
+        node.priority = rand();
         node.element = element;
         return node;
     }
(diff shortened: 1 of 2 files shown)


n8sh:
>truncations is a common technique to improve the output of an LCG.

The replacement also uses truncation. Its exposed output is the top 32 bits of a 64 bit state; drand48 exposes as output the top 32 bits of a 48 bit state.

>Do you have proof that this improves the LCGs output?

I can prove it has a greater period ;), and every possible two-uint sequence appears in its output with equal frequency (that is, it is two-dimensionally equidistributed) unlike drand48 which will never output certain two-uint sequences.

MoonlightSentinel:
> I can prove it has a greater period ;), and every possible two-uint sequence appears in its output with equal frequency (that is, it is two-dimensionally equidistributed) unlike lrand48 which will never output certain two-uint sequences.

That sounds promising, do you have references for those values (esp. `c = 1`)? The closest I could find was [this](https://nuclear.llnl.gov/CNP/rng/rngman/node4.html) which uses `c = 3037000493`.

n8sh:
The multiplier 2862933555777941757 comes from [Tables of Linear Congruential Generators of Different Sizes and Good Lattice Structure (L'Ecuyer, 1999)](https://www.ams.org/journals/mcom/1999-68-225/S0025-5718-99-00996-5/S0025-5718-99-00996-5.pdf). As for the choice of `c`, I have no proof that any odd number is better than any other odd number, so I picked one with a small representation so on a wide variety of processors it can be used as an immediate value instead of loaded from a constant pool.

PR #3044, opened 2020-04-19
Fix Issue 20748 - Deprecation for assert using shared type and checka…

MoonlightSentinel:
…ction=context

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -75,7 +75,21 @@ private string miniFormat(V)(const ref V v)
     import core.internal.traits: isAggregateType;
     import core.stdc.stdio : sprintf;
     import core.stdc.string : strlen;
-    static if (is(V : bool))
+
+    static if (is(V == shared T, T))
+    {
+        // Use atomics to avoid race conditions whenever possible
+        static if (__traits(compiles, atomicLoad(v)))
+        {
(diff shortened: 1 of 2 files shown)


src/core/internal/dassert.d
```diff
@@ -75,7 +75,21 @@ private string miniFormat(V)(const ref V v)
     import core.internal.traits: isAggregateType;
     import core.stdc.stdio : sprintf;
     import core.stdc.string : strlen;
-    static if (is(V : bool))
+
+    static if (is(V == shared T, T))
+    {
+        // Use atomics to avoid race conditions whenever possible
+        static if (__traits(compiles, atomicLoad(*cast(T*) &v)))
```
thewilsonator, line 82: Wont this always succeed by using the [first](https://github.com/dlang/druntime/blob/master/src/core/atomic.d#L69) overload of AtomicLoad. It seems to me that we should be calling the second overload. and that this should be written as 

```d
 static if (__traits(compiles, atomicLoad(v)) // <<<
{
           T tmp = atomicLoad(*cast(T*) &v);
             return miniFormat(tmp);
}
else
{
     // Fall back to a simple cast - we're violating the type system anyways
             return miniFormat(*cast(T*) &v);
}
```

MoonlightSentinel: Using the second overload now

PR #3045, opened 2020-04-19
Fix Issue 20750 - checkaction=context segfaults for null references

MoonlightSentinel:


Geod24:
I can see this creating some bloat, though. But at the same time, trying to reduce the bloat will lead to issues with attributes.

MoonlightSentinel:
Should this be inlined instead of an additional function?

MoonlightSentinel:
Sqashed the inline changes into the first commit.

PR #3047, opened 2020-04-19
Extend _d_assert_fail for unary expressions

MoonlightSentinel:
First step towards `-checkaction=context` for unary expressions. Currently doesn't do any elaborate formatting as done for comparisons but suggestions are welcome.

### DIFF
--- a/test/exceptions/src/assert_fail.d
+++ b/test/exceptions/src/assert_fail.d
@@ -168,6 +168,7 @@ void testAttributes() @safe pure @nogc nothrow
 {
     int a;
     string s = _d_assert_fail!"=="(a, 0);
+    string s2 = _d_assert_fail!"!"(a);
 }
 
 // https://issues.dlang.org/show_bug.cgi?id=20066
@@ -208,6 +209,12 @@ void testEnum()
     test(_d_assert_fail!"=="(ctfe.data, data), "[1] != []");
 }
 
+void testUnary()
+{
+    test(_d_assert_fail!""(9), "9 != true");
+    test(_d_assert_fail!"!"([1, 2, 3]), "[1, 2, 3] == true");
+}
+
 void main()
 {
     testIntegers();
@@ -222,5 +229,6 @@ void main()
     testVoidArray();
     testTemporary();
     testEnum();
+    testUnary();
     fprintf(stderr, "success.\n");
 }
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -4,7 +4,15 @@ on assertion failures
 */
 module core.internal.dassert;
 
-/// Allows customized assert error messages
+/// Allows customized assert error messages for unary expressions
+string _d_assert_fail(string op, A)(auto ref const scope A a)
+{
+    string val = miniFormatFakeAttributes(a);
+    enum token = op == "!" ? "==" : "!=";
+    return combine(val, token, "true");
+}
+
+/// Allows customized assert error messages for binary expressions
 string _d_assert_fail(string comp, A, B)(auto ref const scope A a, auto ref const scope B b)
 {
     /*


test/exceptions/src/assert_fail.d
```diff
@@ -199,6 +199,11 @@ void testEnum()
     test(_d_assert_fail!"=="(ctfe.data, data), "[1] != []");
 }
 
+void testUnary()
+{
+    test(_d_assert_fail(9), "9");
```
schveiguy, line 204: Good, but why not a condition that actually is false?

schveiguy: Meh, I guess it's not important. the other ones aren't real asserts also.

schveiguy: I wonder, though, should there be a complete sentence here. e.g. `assert(0)` is going to print "0". What about "0 evaluates to `false`"

schveiguy: Or to be consistent, it could be "0 != true"

MoonlightSentinel: These tests just check the druntime side of `checkaction=context`, i.e. that `d_assert_fail` creates an appropriate string. The correct behaviour of `assert` is tested in the dmd test suite.

(This test used to rely on the correct lowering but that was just a PITA because one cannot make a PR against both dmd and druntime )

MoonlightSentinel: > Or to be consistent, it could be "0 != true"

Seems nice.

kinke: The lowering will probably need another template param, at least to account for negation, e.g.:
```D
string[] errors = ...;
assert(!errors);
```
=> `[ ... ] == true`.

schveiguy: Wouldn't the `!` come through? You could also check to see if the first character is a `!` and change the `==` to `!=`.

schveiguy: Oh, I see, you meant that's what it *should* do.

MoonlightSentinel: Agreed. Should this consider other unary operators besides negation?

schveiguy: I think you can probably handle all the other operators by pasting the operator on the front of the expression.

The only one that comes to mind is bitwise negation (`~`). I'm assuming `assert(~x) => _d_assert_fail!("~")("x")` ?

You kind of have to handle it in that case as otherwise the string won't make sense.

MoonlightSentinel: Not sure if that makes sense for all operaters, e.g. `*` would be weird.
```
assert(*ptr) => *<value of *ptr> // misleading
assert(*ptr) => *<value of ptr>  // probably useless
assert(*ptr) => <value of *ptr>  // probably right
```

schveiguy: Yes, I hadn't thought of `*`.  I'm not sure how the hook works from the compiler side. The value passed in, is that the data without the operator applied? In that case, it would be the second printout, and I agree it's useless.

Makes sense to handle that one specially.

Might as well address all unary operators:

* `!` => switch `!=` to `==` (as implemented)
* `*` => dereference value for printing
* `~` => paste onto front
* `++` / `--` => Not sure on this, it's a modifying operator. I'm assuming by the time this gets to the function, the value has already been modified?
* `+` / `-` => paste onto front
* `&` => apply to value for printing

I think that's all of them.

One further problem, but I don't know how this would be handled is combinations of unary operators. e.g. `!!` or `!*`?

MoonlightSentinel: We can examine the expression before semantic analysis (see dlang/dmd#11051).

> One further problem, but I don't know how this would be handled is combinations of unary operators. e.g. `!!` or `!*`

The current implementations folds multiple negations (because `!!` is kinda  common)

kinke: I think the other operators are uninteresting.

schveiguy: > I think the other operators are uninteresting.

I don't know what this means. I'm speaking from ignorance of how the hook works, so maybe it's fine, but the result should be consistent and not confusing. It would be odd for instance for this to be the case:

```D
struct S {...}

S s;
S* sptr;

assert(s); // prints nice message
assert(*sptr); // prints no message
```

kinke: Of course it would print a message; assuming `sptr = &s`, it would print the same thing as `assert(s)`. When handling the deref operator, the message would be something like `*<address> != true`, which is obviously way less useful.

schveiguy: @kinke OK that's what I would prefer too. Thanks.

MoonlightSentinel: (BTW the current lowering is available in dlang/dmd#11051)

MoonlightSentinel:
This handles negations now. Anything else that could be improved?

PR #2045, opened 2018-01-17
Fix Issue 18247 - core.stdc.math functions that never set errno should be pure

n8sh:
These include:
* ~~cos, sin, tan,~~ [atan](https://linux.die.net/man/3/atan);
* [tanh](https://linux.die.net/man/3/tanh), [asinh](https://linux.die.net/man/3/asinh);
* [frexp](https://linux.die.net/man/3/frexp), [modf](https://linux.die.net/man/3/modf);
* [cbrt](https://linux.die.net/man/3/cbrt);
* [erf](https://linux.die.net/man/3/erf);
* [ceil](https://linux.die.net/man/3/ceil), [floor](https://linux.die.net/man/3/floor), [trunc](https://linux.die.net/man/3/trunc), [round](https://linux.die.net/man/3/round), [rint](https://linux.die.net/man/3/rint), [nearbyint](https://linux.die.net/man/3/nearbyint);
* [copysign](https://linux.die.net/man/3/copysign), nan;
* [fdim](https://linux.die.net/man/3/fdim), [fmax](https://linux.die.net/man/3/fmax), [fmin](https://linux.die.net/man/3/fmin);
* [fabs](https://linux.die.net/man/3/fabs), abs, [fma](https://linux.die.net/man/3/fma);
* fpclassify, isfinite, isinf, isnan, isnormal, [signbit](https://linux.die.net/man/3/signbit);
* [isgreater](https://linux.die.net/man/3/isgreater), [isgreaterequal](https://linux.die.net/man/3/isgreaterequal), [isless](https://linux.die.net/man/3/isless), [islessequal](https://linux.die.net/man/3/islessequal), [islessgreater](https://linux.die.net/man/3/islessgreater), [isunordered](https://linux.die.net/man/3/isunordered).

PR #3050, opened 2020-04-21
[core.stdc.math] everything is pure

thewilsonator:
Everything in core.stdc.math so use an attribute at the top.

### DIFF
--- a/src/core/stdc/math.d
+++ b/src/core/stdc/math.d
@@ -43,6 +43,7 @@ version (X86_64)  version = X86_Any;
 extern (C):
 @trusted: // All functions here operate on floating point and integer values only.
 nothrow:
+pure:
 @nogc:
 
 ///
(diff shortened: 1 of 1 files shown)


src/core/stdc/math.d
```diff
@@ -671,7 +672,7 @@ else version (CRuntime_Musl)
         FP_FAST_FMAL = 0,
     }
 
-  pure {
+  {
```
Geod24, line 675: Error here

thewilsonator:
Thanks, fixed.

ibuclaw:
I don't think you can say that everything is pure when most alter errno.

Geod24:
For ref: https://en.cppreference.com/w/cpp/numeric/math/math_errhandling

ibuclaw:
For your reference:
```
echo "module gcc.builtins;" | gdc -fsyntax-only -x d -Hf /dev/stdout - | less
```
Functions that are not intrinsically pure (because they store into memory).
- frexp{f,l}
- gamma{f,l}
- gamma_r{f,l}
- lgamma{f,l}
- lgamma_r{f,l}
- modf{f,l}
- remquo{f,l}
- sincos{f,l}

n8sh:
I think https://github.com/dlang/druntime/pull/2045 doesn't leave anything in `core.stdc.math` that can be marked `pure` unless the definition of `pure` is changed to allow changes to `errno`. IIRC Walter Bright has toyed with that idea in the past.

Geod24:
> unless the definition of `pure` is changed to allow changes to `errno`

I really hope we don't go in that direction. Logical `pure` would be as useful as logical `const`.

ibuclaw:
> > unless the definition of `pure` is changed to allow changes to `errno`
> 
> I really hope we don't go in that direction. Logical `pure` would be as useful as logical `const`.

Well pure+errno is the main reason why we do things ourselves in std.math over just forwarding to libm.

RazvanN7:
Closing as per:

>I think #2045 doesn't leave anything in core.stdc.math that can be marked pure unless the definition of pure is changed to allow changes to errno. IIRC Walter Bright has toyed with that idea in the past.

PR #3051, opened 2020-04-21
[core.sys.darwin.execinfo] use an alternative backtrace

thewilsonator:
cc @kinke @ibuclaw 

I modelled this of the freebsd execinfo and LDC's downstream implementation in `core.runtime`. 

Is the versioning on this correct for GDC?

### DIFF
--- a/src/core/sys/darwin/execinfo.d
+++ b/src/core/sys/darwin/execinfo.d
@@ -21,6 +21,55 @@ extern (C):
 nothrow:
 @nogc:
 
-int backtrace(void** buffer, int size);
+version (GNU)
+version = BacktraceExternal;
+
 char** backtrace_symbols(const(void*)* buffer, int size);
 void backtrace_symbols_fd(const(void*)* buffer, int size, int fd);
+
+version (BacktraceExternal)
+{
+    int backtrace(void** buffer, int size);
+else
+{
+    enum _URC_NO_REASON = 0;
+    enum _URC_END_OF_STACK = 5;
+
+    alias _Unwind_Context_Ptr = void*;
+    alias _Unwind_Trace_Fn = int function(_Unwind_Context_Ptr, void*);
+    int _Unwind_Backtrace(_Unwind_Trace_Fn, void*);
+    ptrdiff_t _Unwind_GetIP(_Unwind_Context_Ptr context);
+
+    extern (D) int backtrace(void** buffer, int maxSize)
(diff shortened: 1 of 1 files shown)


src/core/sys/darwin/execinfo.d
```diff
@@ -21,6 +21,55 @@ extern (C):
...
+
+    extern (D) int backtrace(void** buffer, int maxSize)
+    {
+        if (maxSize < 0) return 0;
+
+        struct State
+        {
+            void** buffer;
+            int maxSize;
+            int entriesWritten = 0;
```
Geod24, line 51: Useless initializer

src/core/sys/darwin/execinfo.d
```diff
@@ -21,6 +21,55 @@ extern (C):
...
+    int backtrace(void** buffer, int size);
+else
+{
+    enum _URC_NO_REASON = 0;
+    enum _URC_END_OF_STACK = 5;
+
+    alias _Unwind_Context_Ptr = void*;
+    alias _Unwind_Trace_Fn = int function(_Unwind_Context_Ptr, void*);
+    int _Unwind_Backtrace(_Unwind_Trace_Fn, void*);
+    ptrdiff_t _Unwind_GetIP(_Unwind_Context_Ptr context);
```
Geod24, line 41: Should those be public ?

jacob-carlborg:
Why do you want to make this change?

ibuclaw:
I can't immediately see why GNU would be BacktraceExternal.  Nominally gcc's own libbacktrace is used, though that relies on dwarf, and I don't recall if machos format is supported either.

kinke:
> Why do you want to make this change?

Yeah, the most crucial part is missing, the original downstream comment explaining its goal:
> Use our own backtrce() based on _Unwind_Backtrace(), as the former (from
> execinfo) doesn't seem to handle missing frame pointers too well.

David added this 3 years ago in https://github.com/ldc-developers/druntime/commit/5d26c4bc1119d4d2b8db7de8d23961eb454530d7. So some respective test results would be great to judge whether this should be upstreamed or removed in LDC.

RazvanN7:
@thewilsonator What are your plans with this PR?

thewilsonator:
I do not remember the context of why I was doing this in the first place.

RazvanN7:
Ok to close?

PR #3052, opened 2020-04-21
[core.atomic] atomicStore for alias this

thewilsonator:
@kinke the corresponding LDC commit mentions some unit tests, but I presume that they are in the main LDC repo, but I can't fin the correspond PR/commit. Do you know where they are and should they be moved to druntime?

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -110,7 +110,7 @@ TailShared!T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)(ref shared const T 
  *  newval = The value to store.
  */
 void atomicStore(MemoryOrder ms = MemoryOrder.seq, T, V)(ref T val, V newval) pure nothrow @nogc @trusted
-    if (!is(T == shared) && !is(V == shared))
+    if (!is(T == shared) && !is(V == shared) && is(V : T))
 {
     import core.internal.traits : hasElaborateCopyConstructor;
     static assert (!hasElaborateCopyConstructor!T, "`T` may not have an elaborate copy: atomic operations override regular copying semantics.");
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -116,7 +116,17 @@ void atomicStore(MemoryOrder ms = MemoryOrder.seq, T, V)(ref T val, V newval) pu
     static assert (!hasElaborateCopyConstructor!T, "`T` may not have an elaborate copy: atomic operations override regular copying semantics.");
 
     // resolve implicit conversions
-    T arg = newval;
+    import core.internal.traits : Unqual;
+    static if (is(Unqual!T == Unqual!V))
+    {
+        alias arg = newval;
+    }
+    else
+    {
+        // don't construct directly from `newval`, assign instead (`alias this` etc.)
+        T arg;
+        arg = newval;
+    }
 
     static if (__traits(isFloating, T))
     {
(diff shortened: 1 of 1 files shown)


kinke:
I couldn't find the unittest after a quick search either, maybe it has been removed in the meantime. Anyway, a simple test that works for LDC but not DMD due to this would be:
```D
unittest
{
    static struct S
    {
        int a;
        alias a this;
    }

    import core.atomic;
    S s;
    atomicStore(s, 123);
}
```

thewilsonator:
Awesome. Thanks!

thewilsonator:
Tests added.

src/core/atomic.d
```diff
@@ -1204,4 +1204,17 @@ version (CoreUnittest)
...
+
+    unittest
+    {
+        static struct S
+        {
+            int a;
+            alias a this;
+        }
+
+        import core.atomic;
```
kinke, line 1216: no need ;)

thewilsonator: Derp.

Geod24:
Do we really need this implicit conversion magic though ? The argument is passed by `ref`, implicit conversions are *not* expected. I'd argue that anyone that wants to do this should specify `T` explicitly on the caller side.
If you really want implicit conversions, the signature should reflect it, e.g. T should be deduced to be the common acceptable type.

thewilsonator:
Git blames https://github.com/dlang/druntime/pull/2745 for the comment ` // resolve implicit conversions` 
cc'ing @TurkeyMan

TurkeyMan:
My refactor to simplify and untangle the implementation expects that all weird type wrangling is in the front-end wrappers. The `internal` functions deal only with one type.

"Implicit conversion" is for cases like `T == int` and `V == short`. Or `T == Base*` and `V == Derived*`.

The case in question with `alias this` feels weird to me. I'm not sure I call that an 'implicit conversion'... that's something else, and I don't know how deep that rabbit hole goes.
I think I'd rather make a patch to exclude this case (leading to compile error) rather than trying to handle it like this.

Should we add the constraint `is(V : T)` to the function?

RazvanN7:
@Geod24 Since alias this is a language feature I don't see why we should not support it in this case. I wouldn't call it implicit conversion since the user explicitly defined the alias this.

I agree that a better solution would be to stress out the condition in the template constraint so that in case of screw ups the compiler doesn't point to the innards of `atomicStore`. 

@thewilsonator Please update the function signature so that we can get this in.

thewilsonator:
> Please update the function signature so that we can get this in.

To what? `&& is(V : T)`?

RazvanN7:
> To what? && is(V : T)?

Yes.

thewilsonator:
Done..

RazvanN7 (approved):

RazvanN7:
Thanks!

thewilsonator:
Hmm, this probably needs a rebase

RazvanN7:
Nope, the error is legit. You probably need to switch the types of the `is` expression: `is(V : T)`

thewilsonator:
Sorry, seems I missed this. Suggestion applied. Lets see if this works.

RazvanN7:
This doesn't seem to be going anywhere. Also, pondering upon it, I agree with @Geod24 : "Do we really need this implicit conversion magic though ? The argument is passed by ref, implicit conversions are not expected.". Feel free to reopen if you still want to pursue this.

PR #3053, opened 2020-04-21
clean up test/thread/src/fiber_guard_page

thewilsonator:
Don't depend on absolute offsets for Fiber private members.

### DIFF
--- a/test/thread/src/fiber_guard_page.d
+++ b/test/thread/src/fiber_guard_page.d
@@ -1,6 +1,9 @@
 import core.thread;
 import core.sys.posix.sys.mman;
 
+version (LDC) import ldc.attributes;
+else struct optStrategy { string a; }
+
 // this should be true for most architectures
 // (taken from core.thread)
 version = StackGrowsDown;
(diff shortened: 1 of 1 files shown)


test/thread/src/fiber_guard_page.d
```diff
@@ -18,21 +22,34 @@ void main()
 {
     auto test_fiber = new Fiber(&stackMethod, stackSize);
 
+    auto getPrivateFiberField(string id)()
```
Geod24, line 25: I am a bit puzzled, why not use `__traits(getMember, id)` directly instead of this function ?

thewilsonator: 'cut thats what LDC did ;) Maybe because its private? Feel free to change it as Im about to go to bed.

Geod24: [`getMember` bypasses visibility checks](https://github.com/dlang/dmd/blob/c6ab9c400f54704a1287facc2de03d7a10c10e33/src/dmd/traits.d#L959-L978)

But yeah, I'll raise a PR in LDC.

PR #3057, opened 2020-04-21
Fix Issue 20757 - checkaction=context prints characters as integers

MoonlightSentinel:


schveiguy:
It should check for invalid single code units, and print those as "cast(char)255" e.g. Otherwise, the string is going to look weird when printed (and might cause a UTF exception, I'm not sure).

MoonlightSentinel:
Done

src/core/internal/dassert.d
```diff
@@ -95,10 +95,32 @@ private string miniFormat(V)(const scope ref V v)
...
-        char[20] val;
-        const len = sprintf(&val[0], printfFormat, v);
-        return val.idup[0 .. len];
+        static if (is(V == char))
+        {
+            // Avoid invalid code points
+            if (v < 0x7F)
+                return ['\'', v, '\''];
+
+            return "cast(char) " ~ miniFormat(*cast(ubyte*) &v);
```
schveiguy, line 104: I'm not in love with this reinterpret cast. It wouldn't be horrible to create a tmp as in the other branch.

PR #3058, opened 2020-04-22
Fix Issue 20760 - checkaction=context doesnt print floating point num…

MoonlightSentinel:
…bers correctly

This fixes most issues allthough casting from real to c_long_double might be
problematic for platforms where D's real uses 80 bits but C's long double is
only 64 bit wide (e.g. Win64).

Note that this might cause some deprecations because of https://issues.dlang.org/show_bug.cgi?id=20759

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -125,16 +125,29 @@ private string miniFormat(V)(const scope ref V v)
     }
     else static if (__traits(isFloating, V))
     {
+        import core.stdc.config : LD = c_long_double;
+
         char[60] val;
         int len;
-        static if (is(V == cfloat) || is(V == cdouble))
+        static if (is(V == float) || is(V == double))
+            len = sprintf(&val[0], "%g", v);
+        else static if (is(V == real))
+            len = sprintf(&val[0], "%Lg", cast(LD) v);
+        else static if (is(V == cfloat) || is(V == cdouble))
             len = sprintf(&val[0], "%g + %gi", v.re, v.im);
         else static if (is(V == creal))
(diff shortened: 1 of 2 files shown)


src/core/internal/dassert.d
```diff
@@ -102,16 +102,28 @@ private string miniFormat(V)(const scope ref V v)
...
-            len = sprintf(&val[0], "%Lg", v);
-        else
-            len = sprintf(&val[0], "%g", v);
+            len = sprintf(&val[0], "%Lg + %Lgi", cast(LD) v.re, cast(LD) v.im);
+        else static if (is(V == ifloat) || is(V == idouble))
+            len = sprintf(&val[0], "%gi", v);
+        else // ireal
+        {
+            static if (is(LD == real))
+                alias R = ireal;
```
schveiguy, line 122: Shouldn't `R` be a C type?  I didn't think that C supported imaginary D types.

MoonlightSentinel: Casting to a non-imaginary type yields 0, e.g.
```
ireal num = 2i;
writeln(cast(double) num);    // 0
writeln(cast(idouble) num);   // 2i
```

But ireal is stored the same way as real IIRC.

schveiguy: OK, I see what the point of this is then. I was thinking the issue was that C wasn't properly using the types because of the imaginariness.

One last thing: put a static assert that `V` is `ireal`. It shouldn't affect compilation time at all, and ensures some weird problem doesn't creep in here.

MoonlightSentinel: Done

PR #3061, opened 2020-04-23
Fix Issue 20763 - checkaction=context does not format pointers

MoonlightSentinel:
This adds hexadecimal printing for pointers. ~~Note that it still prefers `toString` for struct pointers if possible (because toString will probably be more useful than raw pointer values).~~

### DIFF
--- a/test/exceptions/src/assert_fail.d
+++ b/test/exceptions/src/assert_fail.d
@@ -60,6 +60,22 @@ void testFloatingPoint()()
     test(creal(1 + 2i), creal(-1 + 2i), "1 + 2i != -1 + 2i");
 }
 
+void testPointers()
+{
+    static struct S
+    {
+        string toString() const { return "S(...)"; }
+    }
+
+    static if ((void*).sizeof == 4)
+        enum ptr = "0x12345670";
+    else
+        enum ptr = "0x123456789ABCDEF0";
+
+    int* p = cast(int*) mixin(ptr);
+    test(cast(S*) p, p, ptr ~ " != " ~ ptr);
+}
+
 void testStrings()
 {
     test("foo", "bar", `"foo" != "bar"`);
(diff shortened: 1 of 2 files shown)


schveiguy:
> Note that it still prefers `toString` for struct pointers if possible

This doesn't make sense. If you are using struct pointers, you are comparing two pointers, not the contents of the structs. It doesn't matter what the structs contain. The pointer values should be printed instead.

For instance:

```D
struct S
{
   string val;
   string toString() { return "S(" ~ val ~ ")"; }
}

void main()
{
   auto s1 = S("hello");
   auto s2 = s1;
   assert(&s1 == &s2); // prints "Error: S(hello) != S(hello)" ?
}
```

MoonlightSentinel:
RIght, that would be weird.
It'll always print the pointer now.

test/exceptions/src/assert_fail.d
```diff
@@ -60,6 +60,29 @@ void testFloatingPoint()()
...
+
+    static if ((void*).sizeof == 4)
+        enum ptr = "12345678";
+    else
+        enum ptr = "123456789012345";
+
+    int* p = cast(int*) mixin("0x" ~ ptr);
+    string msg = _d_assert_fail!"=="(cast(S*) p, p);
+
+    // Some sprintf implementations prepend 0 or 0x
```
schveiguy, line 78: Ugh. Would it make sense to just print the equivalent `size_t` as hex so it's consistent? We don't have to be slaves to the vagaries of C implementations.

i.e. I would prefer:

```D
sprintf("0x%x", p); // use appropriate lx or llx if needed
```

schveiguy: Note this is especially important for implementations that don't print `0x`, as if they do not contain any letters (as your example doesn't) it's not clear that it's hex.

MoonlightSentinel: Done.

PR #3068, opened 2020-04-26
[stable] core.cpuid: Slightly adapt GDC-style inline asm

kinke:


src/core/cpuid.d
```diff
@@ -977,22 +977,19 @@ bool hasCPUID()
     else
     {
         uint flags;
-        version (GNU)
+        version (GNU_OR_LDC)
         {
             // http://wiki.osdev.org/CPUID#Checking_CPUID_availability
-            // ASM template supports both AT&T and Intel syntax.
```
ibuclaw, line 983: I don't think we should be dropping Intel syntax, this'll break Windows/Mingw targets.

ibuclaw: Not to forget, anyone who is daring enough to compile with `-masm=intel`.

kinke: So GDC requires that? It's the only Intel syntax I've seen so far; LDC doesn't need it for MinGW. [MinGW is unsupported anyway.]

kinke: > `-masm=intel`

I think that's bad, having a cmdline flag control the expected format of the asm templates. If there's really a need to support both, some kind of suffix for the string literal or so would be nicer, to have full control over it in the source. [LLVM would support the Intel dialect too, on a per-expression basis, but I don't plan to bother.]

ibuclaw: I've checked, and the default is ATT, so unless someone explicitly changes the build dflags otherwise, it won't be a problem.

PR #3074, opened 2020-04-28
Refactor: Replace usage of custom page size with core.memory

Geod24:


Geod24:
Restored the import to `unistd`.

src/rt/lifetime.d
```diff
@@ -316,7 +313,7 @@ bool __setArrayAllocLength(ref BlkInfo info, size_t newlength, bool isshared, co
             *typeInfo = cast() tinext;
         }
     }
-    else if (info.size < PAGESIZE)
+    else if (info.size < pageSize)
```
rainers, line 316: This should be the same size as the "large object" granularity in the GC (extending is expected to be supported for larger blocks), and must not exceed 64kB, as the size saved as the length of the allocated array has to fit into 2 bytes.

src/core/thread/fiber.d
```diff
@@ -929,9 +930,9 @@ private:
     }
     do
     {
-        // adjust alloc size to a multiple of PAGESIZE
-        sz += PAGESIZE - 1;
-        sz -= sz % PAGESIZE;
+        // adjust alloc size to a multiple of pageSize
+        sz += pageSize - 1;
+        sz -= sz % pageSize;
```
n8sh, line 935:
```suggestion
        sz -= sz & (pageSize -1);
```
This becomes a lot slower when `pageSize` isn't a compile time constant. Since further down you allow the assumption that `pageSize` is a power of 2 I suggest doing so here too.

src/core/thread/fiber.d
```diff
@@ -548,8 +549,8 @@ class Fiber
      * In:
      *  fn must not be null.
      */
-    this( void function() fn, size_t sz = PAGESIZE * defaultStackPages,
-          size_t guardPageSize = PAGESIZE ) nothrow
+    this( void function() fn, size_t sz = pageSize * defaultStackPages,
```
n8sh, line 552: Does using a fixed number of memory pages make more sense here than a fixed number of bytes?

Geod24: Perhaps, but that would be an API change.

n8sh: I meant the default values. Having thought more I believe the requested stack size should be some fixed number of bytes. Having 2MB memory pages shouldn't make fibers _default_ to having 8MB or 16MB of stack (those would be thick fibers indeed). Thanks to other code in this PR the allocation is rounded up to a full page anyway.

Making `guardPageSize` default to the actual page size is good, since the two most sensible values are the real page size or 0.

Geod24:
@rainers : I reverted the change to `rt/lifetime.d` and instead added a commit that rename `PAGESIZE` to `LARGEALLOCSIZE`. Lmk if there's any other improvements I can make to it.

src/rt/lifetime.d
```diff
@@ -437,17 +437,15 @@ BlkInfo __arrayAlloc(size_t arrsize, const TypeInfo ti, const TypeInfo tinext) n
...
-    auto padded_size = addu(arrsize, padsize, overflow);
+    immutable padsize = __arrayPad(arrsize, tinext);
+    immutable padded_size = addu(arrsize, padsize, overflow);
 
     if (overflow)
         return BlkInfo();
 
     uint attr = (!(tinext.flags & 1) ? BlkAttr.NO_SCAN : 0) | BlkAttr.APPENDABLE;
-    if (typeInfoSize)
+    if (structTypeInfoSize(tinext))
```
rainers, line 448: The call to __arrayPad was inlined before to avoid calling structTypeInfoSize twice. Not sure if optimizers inline it and detect the duplication.

test/thread/src/fiber_guard_page.d
```diff
@@ -8,25 +9,23 @@ else struct optStrategy { string a; }
...
 
-enum stackSize = 4096;
-
 // Simple method that causes a stack overflow
 @optStrategy("none")
 void stackMethod()
 {
     // Over the stack size, so it overflows the stack
-    int[stackSize/int.sizeof+100] x;
+    int[minimumPageSize/int.sizeof+100] x;
```
ibuclaw, line 17: I think this might regress a PPC Bugfix I did yesterday. Can you test with a pageSize of 64kb?

src/rt/lifetime.d
Geod24: I checked if `structTypeInfoSize` had any side effect, but it's only checking if the struct has destructors. I guess I would need to refactor this further, but it's not really worth it ATM, so reverting that change.

test/thread/src/fiber_guard_page.d
jacob-carlborg: Can `alloca` with the runtime value `pageSize` be used instead?

RazvanN7:
@Geod24 Do you have any time to take this to the finish line?

PR #3076, opened 2020-04-29
Add module core.sys.darwin.sys.attr with getattrlist, setattrlist, searchfs, and related definitions

n8sh:
Some of the stuff here is used by https://github.com/dlang/phobos/pull/7466.

### DIFF
--- a/src/core/sys/darwin/sys/attr.d
+++ b/src/core/sys/darwin/sys/attr.d
@@ -1,0 +1,316 @@
+/**
+ * D header file for Darwin
+ *
+ * $(LINK2 https://opensource.apple.com/source/xnu/xnu-2422.115.4/bsd/sys/attr.h.auto.html, Apple sys/attr.h)
+ */
+module core.sys.darwin.sys.attr;
+
+version (OSX)
+    version = Darwin;
+else version (iOS)
+    version = Darwin;
+else version (TVOS)
+    version = Darwin;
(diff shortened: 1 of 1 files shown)


src/core/sys/darwin/sys/attr.d
```diff
@@ -0,0 +1,305 @@
+/**
+ * D header file for Darwin
+ */
```
Geod24, line 3: Needs proper documentation, and changelog entry

n8sh: It has an associated issue https://issues.dlang.org/show_bug.cgi?id=20787 although the D bot hasn't linked it yet for whatever reason.

n8sh: Added a link to the macOS 10.9 sys/attr.h and put comments next to constants that were added since then. Since public documentation isn't generated for `core.sys.*` I think this is as much as is needed.

ibuclaw:
Not for now, but soon there should be clearer documentation of what should go in these bindings modules, a number of modules already do this, [core.sys.posix.dlfcn](https://github.com/dlang/druntime/blob/fa3eeed29b5d892da921c585e6526635475059a4/src/core/sys/posix/dlfcn.d#L51) is a good example to follow (I am thinking of ways to facilitate automatically generating these from C headers).

PR #3078, opened 2020-04-29
[dmd-cxx] core.sys.posix.sys.stat: Fix previous SPARC32/Solaris change.

ibuclaw:
(Backport)
The pull had removed the right field, but from the wrong struct


ibuclaw:
Patch on master was merged.  Also merging this too.

PR #3080, opened 2020-04-29
core.internal.convert: Handle ireal types on PPC

ibuclaw:
There is no toPrec intrinsic for imaginary types, so must use .im to convert to real as needed.


ibuclaw:
Ping @kinke

PR #3082, opened 2020-04-29
merge stable

MartinNowak:
- forward and use -m$(MODEL) when building tools with host dmd
- core.math: Remove static linkage from toPrec instantiation tests
- core.math: Use internal approxEqual to allow some inaccuracies on certain hardware
- core.math: Update toPrec comment to reflect how rounding is done.
- core.math: Re-enable disabled tests
- core.cpuid: Simplify GDC-style inline asm for 32-bit x86 hasCPUID()
- core.cpuid: Fix deprecated syntax for GDC-style inline asm
- Fix potential memory corruption
- Fix code bloat regressions wrt. core.internal.lifetime.emplaceInitializer

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -55,6 +55,9 @@ COPY=\
 	\
 	$(IMPDIR)\core\internal\util\array.d \
 	\
+	$(IMPDIR)\core\internal\vararg\aarch64.d \
+	$(IMPDIR)\core\internal\vararg\sysv_x64.d \
+	\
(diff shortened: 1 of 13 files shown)


MartinNowak:
LLVM's download cache (varnish) seems to be down.

ibuclaw:
Probably biased as it seems I'm either the author or main reviewer in almost all these changes, but I see nothing blocking here.

PR #3084, opened 2020-04-30
core.sys: Add SPARC/CRuntime_Glibc bindings

ibuclaw:
With this, both SPARC64 and its SPARC multilib are building phobos just fine.

No run-time test done just yet.

### DIFF
--- a/src/core/sys/posix/sys/stat.d
+++ b/src/core/sys/posix/sys/stat.d
@@ -30,6 +30,8 @@ else version (WatchOS)
 
 version (RISCV32) version = RISCV_Any;
 version (RISCV64) version = RISCV_Any;
+version (SPARC)   version = SPARC_Any;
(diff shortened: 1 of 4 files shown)


src/core/sys/posix/sys/stat.d
```diff
@@ -790,7 +799,15 @@ version (CRuntime_Glibc)
...
         }
-        static assert(stat_t.sizeof == 144);
+        static if (__USE_LARGEFILE64) alias stat_t stat64_t;
+
+        static if (__WORDSIZE == 64)
+            static assert(stat_t.sizeof == 144);
+        else static if (__USE_FILE_OFFSET64)
+            static assert(stat_t.sizeof == 104);
+        else
+            static assert(stat_t.sizeof == 108);
```
ibuclaw, line 809: That should have been an 88, #3085.

PR #3085, opened 2020-05-01
core.sys.posix.sys.stat: Fix typo from #3084

ibuclaw:
The non-64bit, non-largefile size on SPARC is 88.


ibuclaw:
(Ignoring the build failure preventing auto merging).

PR #3069, opened 2020-04-26
Add publictests test checker to druntime (reboot of #2087)

MoonlightSentinel:
Checking public examples is always worthwhile IMHO.

- [x] dlang/tools#402
- [x] dlang-community/libdparse#391
- [x] dlang-community/libdparse#398
- [x] dlang-community/libdparse#401
- [x] dlang-community/libdparse#402
- [x] dlang-community/libdparse#404 (optional)
- [x] dlang/tools#403
- [x] Update libdparse used by `test_extractor.d` to include latest bugfixes

PR #3087, opened 2020-05-01
Remove superflous semicolons

MoonlightSentinel:
They cause a lot of warnings when using DParse-based tools (DScanner/test_extractor/...) which needlessly clutter the log files (e.g. in #3069 )

### DIFF
--- a/src/core/stdc/errno.d
+++ b/src/core/stdc/errno.d
@@ -1757,7 +1757,7 @@ else version (Solaris)
     enum ECONNREFUSED =   146     /** Connection refused */;
     enum EHOSTDOWN =      147     /** Host is down */;
     enum EHOSTUNREACH =   148     /** No route to host */;
-    enum EWOULDBLOCK =    EAGAIN;      /** Resource temporarily unavailable     */;
+    enum EWOULDBLOCK =    EAGAIN  /** Resource temporarily unavailable */;
     enum EALREADY =       149     /** operation already in progress */;
     enum EINPROGRESS =    150     /** operation now in progress */;
     enum ESTALE =         151     /** Stale NFS file handle */;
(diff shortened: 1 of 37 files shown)


src/core/stdc/errno.d
```diff
@@ -1757,7 +1757,7 @@ else version (Solaris)
     enum ECONNREFUSED =   146     /** Connection refused */;
     enum EHOSTDOWN =      147     /** Host is down */;
     enum EHOSTUNREACH =   148     /** No route to host */;
-    enum EWOULDBLOCK =    EAGAIN;      /** Resource temporarily unavailable     */;
+    enum EWOULDBLOCK =    EAGAIN  /** Resource temporarily unavailable */;
```
Geod24, line 1760: That's a weird wait to document symbol though. 👍 for making it consistent, but does it even work ?

schveiguy: That stood out to me too, but I think fix in a separate PR.

MoonlightSentinel: No, those comments need to be after the semicolon but thats for another PR

schveiguy: I don't know if those DDocs are used, that's in `version(Solaris)`. Probably it was that way in the original.

MoonlightSentinel:
buildkite/dub failure is (obviously) unrelated

schveiguy:
All buildkite jobs are going to fail with this unrelated failure. Long story short, code.dlang.org starts forgetting versions after 100 tags are present. See for more details: https://github.com/dlang/dub-registry/pull/458

src/core/stdc/errno.d
MoonlightSentinel: See #3088

PR #3092, opened 2020-05-02
MMapRegion: Split between typed and untyped, handle error

Geod24:
Slight refactoring and documentation of MMapRegion in order to make it less awkward to use for consumer that have no need for a typed mapping.

### DIFF
--- a/src/core/internal/elf/io.d
+++ b/src/core/internal/elf/io.d
@@ -296,6 +296,8 @@ private struct TypedMMapRegion(T)
         const pageMapped = (mappedSize / pageSize) + !!(mappedSize % pageSize);
         this.region = MMapRegion(fd, pageOffset, pageMapped);
         this.data = cast(const(T)*) (this.region.data.ptr + offsetDiff);
+        if (this.region.data.ptr !is null)
+            this.data = cast(const(T)*) (this.region.data.ptr + offsetDiff);
     }
 
     private MMapRegion region;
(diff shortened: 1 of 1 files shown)


src/core/internal/elf/io.d
```diff
@@ -283,37 +284,65 @@ version (LinuxOrBSD)
...
-
-        auto ptr = cast(ubyte*) mmap(null, mappedSize, PROT_READ, MAP_PRIVATE, fd, pagedOffset);
+        const offset = pageOffset * pageSize;
+        const mappedSize = pageCount * pageSize;
 
-        mappedRegion = ptr[0 .. mappedSize];
-        data = cast(T*) (ptr + offsetDiff);
+        auto ptr = mmap(null, mappedSize, PROT_READ, MAP_PRIVATE, fd, offset);
+        if (ptr !is MAP_FAILED)
+            this.data = (cast(const(ubyte)*) ptr)[0 .. mappedSize];
```
PetarKirov, line 333: I think you missed one of my suggestions:

```D
        const ptr = cast(ubyte*) mmap(null, mappedSize, PROT_READ, MAP_PRIVATE, fd, offset);
        if (ptr !is MAP_FAILED)
            this.data = ptr[0 .. mappedSize];
 ```

Geod24: I just wasn't so found of it, but suggestion applied nonetheless :)

Geod24:
@PetarKirov : All comments should have been addressed (selective import and the different `cast`).

PetarKirov:
Thanks @Geod24, I've set the `auto-merge` label.
P.S. And sorry for the delay - please message me directly on Slack, as I have too many GitHub notifications and I don't always manage to catch up with all of them in a timely manner.

PR #3093, opened 2020-05-02
[stable] Fix regressions crept in with core.stdc.stdarg refactoring

kinke:


PetarKirov:
The druntime job was failing on BuildKite with OOM. After I restarted it is now green.

PR #3094, opened 2020-05-03
Improve `core.atomic.cas` documentation, formatting, and constraints

Geod24:
Trying to get a lazily initialized cache working in `rt.backtrace.dwarf`, of course things had to blow up somewhere else :)

The first commit (slightly) correct the definition of `hasUnsharedIndirections` to allow for `immutable` (which is used in the second commit). Ultimately, we really should clean up this code duplication.

The second commit contain a `unittest` with the code I was actually attempting to write, as well as the fix to make it pass. The last commit contains a slight rewrite of the documentation and formatting.

CC @TurkeyMan

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -277,7 +277,7 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
  *  true if the store occurred, false if not.
  */
 bool cas(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V1,V2)(T* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
-    if (!is(T == shared) && is(T : V1))
+    if (!is(T == shared) && is(V1 : T))
 in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 {
     // resolve implicit conversions
@@ -295,7 +295,7 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 
 /// Ditto
 bool cas(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V1,V2)(shared(T)* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
-    if (!is(T == class) && (is(T : V1) || is(shared T : V1)))
+    if (!is(T == class) && (is(V1 : T) || is(V1 : shared T)))
 in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 {
     static if (is (V1 == shared U1, U1))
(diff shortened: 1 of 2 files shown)


Geod24:
I'm going to try another approach (a prior refactoring)

src/core/atomic.d
```diff
@@ -1194,4 +1211,42 @@ version (CoreUnittest)
...
+     * Error: template core.atomic.cas cannot deduce function from argument types
+     * !()(shared(immutable(Image)*)*, typeof(null), immutable(Image*)), candidates are:
+     *
+     * cas(MemoryOrder succ = MemoryOrder.seq, MemoryOrder fail = MemoryOrder.seq, T, V1, V2)
+     * (T* here, V1 ifThis, V2 writeThis)
+     *
+     */
+    pure nothrow @safe unittest
+    {
+        // A structure that is not atomically initialization (e.g. open files)
```
thewilsonator, line 1227: Grammar.

src/core/atomic.d
```diff
@@ -264,20 +264,35 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
...
  * Returns:
- *  true if the store occurred, false if not.
+ *  Whether the swap occured or not.
+ *
+ * See_Also:
+ *  https://en.wikipedia.org/wiki/Compare-and-swap
  */
-bool cas(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V1,V2)(T* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
-    if (!is(T == shared) && is(T : V1))
+bool cas
```
TurkeyMan, line 292: weird life choices

src/core/atomic.d
```diff
@@ -264,20 +264,35 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
...
+    (MemoryOrder succ = MemoryOrder.seq, MemoryOrder fail = MemoryOrder.seq, T, V1, V2)
+    (T* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
+    if (!is(T == shared) && is(V1 : T))
```
TurkeyMan, line 295: hard to tell from the diff if you made functional changes or just reformat the existing code...
i'm having to audit this carefully, but i probably don't need to?
if this sort of refactor is fair game, i'll definitely just change this right back next time i touch it ;)

src/core/atomic.d
```diff
@@ -294,8 +309,10 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 }
 
 /// Ditto
-bool cas(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V1,V2)(shared(T)* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
-    if (!is(T == class) && (is(T : V1) || is(shared T : V1)))
+bool cas
+    (MemoryOrder succ = MemoryOrder.seq, MemoryOrder fail = MemoryOrder.seq, T, V1, V2)
+    (shared(T)* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
+    if (!is(T == class) && (is(V1 : T) || is(V1 : shared T)))
```
TurkeyMan, line 315: oh I see there is hidden changes!
what's the logic behind turning the type conversion around?
why is it interesting if `here` is implicitly convertible to `ifThis`? `here` is what's being mutated...

src/core/atomic.d
Geod24: Yeah I don't like it either, hence why I'm experimenting with: https://github.com/dlang/druntime/pull/3095

src/core/atomic.d
Geod24: You should review commit-by-commit, that will be much easier to see, and you will also see the rationale for it ;)

TurkeyMan: Oh I'm sorry! I somehow looked at the inverse of what you're doing.
This is bizarre. Is it my mistake? Weird bug! O_O

So, I wonder if it's actually necessary for `V1` to participate in the constraint?
It also seems odd for `V1` to participate in the constraint but not `V2`...
I wonder if this `is(V1 : T) || is(V1 : shared T)` could be something like `is(RemoveShared!V1 : T)` ?

Geod24:
Blocked by https://github.com/dlang/druntime/pull/3095

thewilsonator:
#3095 is merged

PR #3095, opened 2020-05-04
Refactor core.atomic : cas into a two-level template

Geod24:
```
This should reduce the amount of boilerplate needed in the implementation,
and hopefully make the code easier to read for everyone.
```

I want to see how it shows up in the doc though, before we merge this.

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -264,9 +264,16 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
 }
 
 /**
- * Stores 'writeThis' to the memory referenced by 'here' if the value
- * referenced by 'here' is equal to 'ifThis'.  This operation is both
- * lock-free and atomic.
+ * Performs either compare-and-set or compare-and-swap (or exchange).
+ *
+ * There are two categories of overloads in this template:
+ * the first one does a simple compare-and-set, and returns a boolean if the
+ * operation happened. The value this is written (`writeThis`) can be an rvalue.
+ * the second category does a compare-and-swap, or compare-and-exchange,
+ * and expects `writeThis` to be a pointer type, where the previous value
+ * of `here` will be written.
+ *
+ * This operation is both lock-free and atomic.
  *
  * Params:
  *  here      = The address of the destination variable.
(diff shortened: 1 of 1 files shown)


Geod24:
@TurkeyMan : Alright I have to admit when I started refactoring this I missed the point that this was essentially both compare-and-set and compare-and-exchange (as in, `ifThis` might get written to).

Combining this with templates leads to a very unintuitive API IMO. Things blew up for me when I tried to use compare-and-set with pointers (basically initializing a `shared immutable(T)*`). Thoughts ?

TurkeyMan:
This is terrifying. Does argument inference work with this arrangement?
I have no feel for what the semantics are with this change :/

src/core/atomic.d
```diff
@@ -276,112 +283,117 @@ in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
...
-        return atomicCompareExchangeStrongNoResult!(succ, fail)(here, arg1, arg2);
-}
-
-/// Ditto
-bool cas(MemoryOrder succ = MemoryOrder.seq,MemoryOrder fail = MemoryOrder.seq,T,V1,V2)(shared(T)* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
+    /// Compare-and-set for non-shared values
+    bool cas(T, V1, V2)(T* here, V1 ifThis, V2 writeThis) pure nothrow @nogc @trusted
+    if (!is(T == shared) && is(T : V1))
+    in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
+       {
```
TurkeyMan, line 292: Why is this double-indented?

Geod24: Fixed

Geod24:
> This is terrifying. Does argument inference work exactly the same with this arrangement?

As long as one does not explicitly specify the type of arguments, and relies on IFTI, everything works the same. This pattern is very helpful when one needs to have some template parameters part of the interface and others not.

> I don't have time right now to think this through, but I'd like the opportunity to understand this before it's merged.

Any chance you could give it a look now ?

Geod24:
Ping @TurkeyMan

n8sh:
The change to a two-level template will cause any existing fully-parameterized uses of `cas` to fail to compile!

Geod24:
Ping @TurkeyMan

n8sh:
I don't see documentation indicating to library users that `core.atomic.cas` is not supposed to be explicitly parameterized, but as [the parameterization was changed a year ago](https://github.com/dlang/druntime/pull/2745) without objection I can't on principle say this is a blocker.

Geod24:
Well it's been more than 72h so adding `auto-merge`.

Geod24:
I don't know what's up with FreeBSD... It doesn't seem spurious, however this was seen in other unrelated PR. Will investigate next week.

PR #3098, opened 2020-05-10
time.d: remove use of typesafe variadic functions

WalterBright:
Seeing how things work without TVFs.

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -1196,8 +1196,8 @@ public:
       +/
     template split(units...)
         if (allAreAcceptedUnits!("weeks", "days", "hours", "minutes", "seconds",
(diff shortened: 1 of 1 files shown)


jmdavis (approved):
I don't see any problems with the implementation here, but the buildkite tests are failing when building Phobos. I don't know if it's related to this or not, since the errors appear to be in std.array (which shouldn't be affected by this AFAIK), but the errors do appear to be related to scope. e.g.

    std/array.d(631): Error: reference to local variable `__arrayliteral_on_stack1540` assigned to non-scope parameter `a` calling core.internal.destruction.__ArrayDtor!(ThrowingElement).__ArrayDtor

MoonlightSentinel:
That's just another case where DIP1000 chokes on expression temporaries, e.g. `-checkaction=context` has the same problem.

PR #3099, opened 2020-05-10
switch_.d: remove use of typesafe variadic functions

WalterBright:
To see if TVF can be removed.


schveiguy:
Wait, doesn't this allocate when before it didn't?

MoonlightSentinel:
Not anymore, see dlang/dmd#11039

schveiguy:
Hm... is run.dlang.io nightly up to date? It still complains that `[1, 2, 3]` is not `@nogc`, even when passed to a scope array accepting function.

PR #3101, opened 2020-05-10
attribute.d: remove use of typesafe variadic functions

WalterBright:
Seeing how it goes.

### DIFF
--- a/src/core/attribute.d
+++ b/src/core/attribute.d
@@ -221,12 +221,12 @@ version (UdaOptional)
  * }
  * Tagged1_2 inst1;
  * // ---- foo.d
- * @gnuAbiTag("tag1", "tag2") struct Tagged1_2
+ * @gnuAbiTag(["tag1", "tag2"]) struct Tagged1_2
  * {
  *     // Notice the repetition
- *     @gnuAbiTag("tag1", "tag2", "tag3") struct Tagged3
+ *     @gnuAbiTag(["tag1", "tag2", "tag3"]) struct Tagged3
  *     {
- *         @gnuAbiTag("tag1", "tag2", "tag3", "tag4") int Tagged4 ();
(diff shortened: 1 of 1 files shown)


WalterBright:
Part of https://github.com/dlang/dmd/pull/11124

src/core/attribute.d
```diff
@@ -136,7 +136,7 @@ version (UdaGNUAbiTag) struct gnuAbiTag
 {
     string[] tags;
 
-    this(string[] tags...)
```
thewilsonator, line 139: Why are these mutually exclusive?

WalterBright: Not sure what you mean.

thewilsonator: Why does `this(scope string[] tags...)` not work?

WalterBright: Still not sure what you mean. `scope` is applied automatically to ... parameters. The replacement for `T[]...` is `scope T[]`.

src/core/attribute.d
```diff
@@ -136,8 +136,14 @@ version (UdaGNUAbiTag) struct gnuAbiTag
...
+    this(scope string[] tags)
     {
         this.tags = tags;
     }
+
+    /* Overloads for backwards compatibility
+     */
+    this(string s1)                       { this.tags = [s1]; }
```
ibuclaw, line 146: This is a rabbit hole. Maybe it's better moving the variadic part from the parameters to the template definition.

RazvanN7:
@WalterBright It seems that the dmd testsuire is failing:

```
>>> TARGET FAILED: fail_compilation/cpp_abi_tag2.d
FAILED targets:
- fail_compilation/cpp_abi_tag2.d
```

WalterBright:
Replaced by https://github.com/dlang/dmd/pull/14311

PR #3103, opened 2020-05-12
Fix ioctl request for TCSETS2

gizmomogwai:


n8sh:
Messed up, was trying to change the comment to reference https://issues.dlang.org/show_bug.cgi?id=20832.

PR #3108, opened 2020-05-18
Deprecate `object.selector`

jacob-carlborg:
The reasoning being it's expected that other Objective-C related UDAs will be added in the future. These will have too common names to be imported into the `object` module. To make it consistent, deprecate `object.selector`. Instead the user needs to explicitly import `selector` using `import core.attribute : selector;`.

The impact is expected to be small. It only needs to be added once per module. It's expected that bindings need to only be created once. Tools can be used to generate the bindings automatically, in which case the tools can add the import automatically.

Will follow up with PRs in other projects to match this.

### DIFF
--- a/changelog/depercate_object_selector.dd
+++ b/changelog/depercate_object_selector.dd
@@ -1,0 +1,26 @@
+Deprecated `object.selector`
+
+The public import of `selector` in the module `object` has been deprecated.
+Please explicitly import `selector` instead using: `import core.attribute : selector;`
+
+The following code has been deprecated:
+
+```
+extern (Objective-C)
+extern class NSObject
+{
+    static NSObject alloc() @selector("alloc");
+}
+```
+
+Replace with:
+
+```
+import core.attribute : selector;
+
+extern (Objective-C)
+extern class NSObject
+{
+    static NSObject alloc() @selector("alloc");
+}
+```
--- a/src/object.d
+++ b/src/object.d
@@ -36,7 +36,11 @@ alias string  = immutable(char)[];
 alias wstring = immutable(wchar)[];
 alias dstring = immutable(dchar)[];
 
-version (D_ObjectiveC) public import core.attribute : selector;
+version (D_ObjectiveC)
+{
+    deprecated("explicitly import `selector` instead using: `import core.attribute : selector;`")
+        public import core.attribute : selector;
+}
 version (Posix) public import core.attribute : gnuAbiTag;
 
 // Some ABIs use a complex varargs implementation requiring TypeInfo.argTypes().


Geod24:
Does the deprecation work ? @ibuclaw said it does not.

src/object.d
```diff
@@ -36,7 +36,11 @@ alias string  = immutable(char)[];
 alias wstring = immutable(wchar)[];
 alias dstring = immutable(dchar)[];
 
-version (D_ObjectiveC) public import core.attribute : selector;
+version (D_ObjectiveC)
+{
+    deprecated("explicitly import `selector` instead using: `import core.attribute : selector;`")
+        public import core.attribute : selector;
```
thewilsonator, line 42: if deprecation on the import does not work then 
```
private static import core.attribute : selector;
deprecated pubic alias selector = core.attribute.selector;
```
should.

changelog/depercate_object_selector.dd
```diff
@@ -0,0 +1,26 @@
+Deprecated `object.selector`
+
+The public import of `selector` in the module `object` has been deprecated.
+Please explicitly import `selector` instead using: `import core.attribute : selector;`
+
+The following code has been deprecated:
+
+```
```
kubo39, line 8: Shouldn't we use `---` with ddoc?

ibuclaw:
> Does the deprecation work ? @ibuclaw said it does not.

Templates and aliases to templates do not work.

There is no such thing as `deprecated import`, but maybe there should be...

changelog/depercate_object_selector.dd
jacob-carlborg: Possibly. Is markdown enabled for changelog entires.

jacob-carlborg:
> Does the deprecation work ?

Yes.

changelog/depercate_object_selector.dd
Geod24: [Yes](https://github.com/dlang/druntime/blob/40d5de0c52a85f1458c97e119bcde2846c6dd7f5/posix.mak#L68).

jacob-carlborg:
> There is no such thing as deprecated import, but maybe there should be...

I’m guessing it works because it’s a selective import which is lowered to an alias.

changelog/depercate_object_selector.dd
jacob-carlborg: Then this is fine.

jacob-carlborg:
As you can see on the failing tests it does work. I guess I should have started with the other repositories first. 

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=4051486&isPull=true

dnadlinger:
Wouldn't it be more appropriate to have something like `core.objc` then? `core.attribute` doesn't seem like a very cohesive unit of abstraction.

jacob-carlborg:
> Wouldn't it be more appropriate to have something like core.objc then? core.attribute doesn't seem like a very cohesive unit of abstraction. 

It’s a general place for attributes. Like compiler recognized UDAs. I can also imagine if Phobos had a serialization module `core.attribute` could contain UDAs indicating if a symbol should be serialized or not.

changelog/depercate_object_selector.dd
wilzbach: Yup - http://dtest.dlang.io/artifact/website-403dc89ebaba38a7ee42cfbb6c2b6c5f8657bffd-e9d3a6d3493d65a0e0bec935c528c07a/web/changelog/pending.html#depercate_object_selector

For future reference: you should even be able to use .md as an extension.

jacob-carlborg:
Everything is green now.

PR #3109, opened 2020-05-19
Fix Issue 20844 - Insufficient alignment check for 16-bytes CAS on x86_64

kinke:


src/core/atomic.d
```diff
@@ -621,36 +619,32 @@ else
...
-            static if (T.sizeof > size_t.sizeof)
+            version (X86)
+            {
+                // cmpxchg8b only requires 4-bytes alignment
                 return cast(size_t)ptr % size_t.sizeof == 0;
+            }
             else
+            {
+                // e.g., x86_64 cmpxchg16b requires 16-bytes alignment
                 return cast(size_t)ptr % T.sizeof == 0;
```
kinke, line 642: @ibuclaw: Is that a 'good' default logic for non-x86 targets? I think it's at least better than the current `return true` for non-DMD...

PR #3111, opened 2020-05-21
Upstream a few things from LDC to reduce the diff

kinke:


src/rt/sections_elf_shared.d
```diff
@@ -95,6 +95,7 @@ struct DSO
         return _moduleGroup;
     }
 
+    version (DigitalMars)
     @property immutable(FuncTable)[] ehTables() const nothrow @nogc
```
kinke, line 99: @ibuclaw: I assume gdc doesn't feature this method either.

src/rt/sections_darwin_64.d
```diff
@@ -176,3 +178,28 @@ const(section_64)[] sections(const segment_command_64* segment) pure nothrow @no
     const firstSection = cast(const(section_64)*)(cast(ubyte*) segment + size);
     return firstSection[0 .. segment.nsects];
 }
+
+/// Identifies a section via segment and section names.
+struct SectionRef
+{
+    immutable(char)* seg;
+    immutable(char)* sect;
+}
```
Geod24, line 187: Personally I'd rather have them as string, as previously. Why the change ?

src/rt/sections_darwin_64.d
```diff
@@ -176,3 +178,28 @@ const(section_64)[] sections(const segment_command_64* segment) pure nothrow @no
...
+
+/// The data sections of an image (executable or shared library).
+static immutable SectionRef[] dataSections = [{SEG_DATA, SECT_DATA},
```
Geod24, line 190: `static` here is redundant

src/rt/sections_darwin_64.d
kinke: Because the single usage - for `getSection()` below - requires null-terminated char* pointers.

kinke: I've refactored this & amended the last commit.

PR #3116, opened 2020-05-24
Followup to #3076: make core.sys.darwin.sys.attr.attrlist usable in betterC

n8sh:
Achieves this by removing the field initializers. This should be fine because `core.sys.darwin.sys.attr.attrlist` has not yet appeared in a release.


thewilsonator:
cc @jacob-carlborg

jacob-carlborg:
Looks fine to me. The fields in the header don't have any values.

PR #3118, opened 2020-05-24
Port more efficient Filter & staticMap from dlang/phobos#7490

n8sh:


wilzbach:
Ideally, after this PR the Phobos declarations could be changed to an alias to avoid this duplication, but I guess ddoc is blocking this?

thewilsonator:
I don't remember but https://github.com/dlang/phobos/pull/7148 fails CI.

PR #3120, opened 2020-06-03
Store memory-related GC options in bytes instead of megabytes.

denizzzka:
Rationale: this is need to fine-tune GC on low memory embedded devices.

User interface default units is MB, unchanged.

### DIFF
--- a/changelog/gc_suffixes.dd
+++ b/changelog/gc_suffixes.dd
@@ -1,0 +1,12 @@
+Memory-related GC options stored in bytes
+
+Memory-related GC options stored in bytes instead of megabytes to
+provide fine-tune GC on low memory devices.
+
+Added ability to use suffixes B, K or M:
+---
+extern(C) __gshared string[] rt_options =
+    [ "gcopt=minPoolSize:4K maxPoolSize:2M incPoolSize:8K" ];
+---
+
+Values without suffix treated as megabytes.
--- a/changelog/gc_suffixes.dd
+++ b/changelog/gc_suffixes.dd
@@ -1,4 +1,4 @@
-Memory-related GC options stored in bytes
+Memory releated GC options can now be specified with higher accuracy
 
 Memory-related GC options stored in bytes instead of megabytes to
 provide fine-tune GC on low memory devices.
--- a/changelog/gc_suffixes.dd
+++ b/changelog/gc_suffixes.dd
@@ -3,7 +3,7 @@ Memory-related GC options stored in bytes
 Memory-related GC options stored in bytes instead of megabytes to
 provide fine-tune GC on low memory devices.
 
-Added ability to use suffixes B, K or M:
+Added ability to use suffixes B, K, M or G:
 ---
 extern(C) __gshared string[] rt_options =
     [ "gcopt=minPoolSize:4K maxPoolSize:2M incPoolSize:8K" ];
--- a/src/core/gc/config.d
+++ b/src/core/gc/config.d
@@ -55,6 +55,8 @@ struct Config
     parallel:N     - number of additional threads for marking (%lld)
     heapSizeFactor:N - targeted heap size to used memory ratio (%g)
     cleanup:none|collect|finalize - how to treat live objects when terminating (collect)
+
+    Memory-related values can use B, K or M suffixes.
 ".ptr,
                cast(long)initReserve, cast(long)minPoolSize,
                cast(long)maxPoolSize, cast(long)incPoolSize,
--- a/src/core/gc/config.d
+++ b/src/core/gc/config.d
@@ -56,7 +56,7 @@ struct Config
     heapSizeFactor:N - targeted heap size to used memory ratio (%g)
     cleanup:none|collect|finalize - how to treat live objects when terminating (collect)
 
-    Memory-related values can use B, K or M suffixes.
+    Memory-related values can use B, K, M or G suffixes.
 ".ptr,
                cast(long)initReserve, cast(long)minPoolSize,
                cast(long)maxPoolSize, cast(long)incPoolSize,
--- a/src/core/gc/config.d
+++ b/src/core/gc/config.d
@@ -46,22 +46,82 @@ struct Config
             if (i) printf("|");
             printf("%.*s", cast(int) entry.name.length, entry.name.ptr);
         }
+        auto _initReserve = initReserve.bytes2prettyStruct;
+        auto _minPoolSize = minPoolSize.bytes2prettyStruct;
+        auto _maxPoolSize = maxPoolSize.bytes2prettyStruct;
+        auto _incPoolSize = incPoolSize.bytes2prettyStruct;
         printf(" - select gc implementation (default = conservative)
 
-    initReserve:N  - initial memory to reserve in MB (%lldB)
-    minPoolSize:N  - initial and minimum pool size in MB (%lldB)
-    maxPoolSize:N  - maximum pool size in MB (%lldB)
-    incPoolSize:N  - pool size increment MB (%lldB)
+    initReserve:N  - initial memory to reserve in MB (%lld%c)
+    minPoolSize:N  - initial and minimum pool size in MB (%lld%c)
+    maxPoolSize:N  - maximum pool size in MB (%lld%c)
+    incPoolSize:N  - pool size increment MB (%lld%c)
     parallel:N     - number of additional threads for marking (%lld)
     heapSizeFactor:N - targeted heap size to used memory ratio (%g)
     cleanup:none|collect|finalize - how to treat live objects when terminating (collect)
 
     Memory-related values can use B, K, M or G suffixes.
 ".ptr,
-               cast(long)initReserve, cast(long)minPoolSize,
-               cast(long)maxPoolSize, cast(long)incPoolSize,
+               _initReserve.v, _initReserve.u,
+               _minPoolSize.v, _minPoolSize.u,
+               _maxPoolSize.v, _maxPoolSize.u,
+               _incPoolSize.v, _incPoolSize.u,
                cast(long)parallel, heapSizeFactor);
     }
 
     string errorName() @nogc nothrow { return "GC"; }
+}
+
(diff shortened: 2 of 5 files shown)


src/core/internal/parseoptions.d
```diff
@@ -156,16 +173,64 @@ inout(char)[] find(alias pred)(inout(char)[] str)
     return null;
 }
 
-bool parse(T:size_t)(const(char)[] optname, ref inout(char)[] str, ref T res, const(char)[] errName)
+bool parse (T:size_t)(const(char)[] optname, ref inout(char)[] str, ref T res, const(char)[] errName)
```
MoonlightSentinel, line 176:
```suggestion
bool parse(T : size_t)(const(char)[] optname, ref inout(char)[] str, ref T res, const(char)[] errName)
```

src/core/internal/parseoptions.d
```diff
@@ -94,7 +94,24 @@ bool parseOptions(CFG)(ref CFG cfg, string opt)
...
+
+                        static if (is(fieldT == size_t))
+                        {
+                             // names of field what can contain B/K/M suffixes
+                            bool isSuffValue = (
+                                name == "initReserve" ||
+                                name == "minPoolSize" ||
+                                name == "maxPoolSize" ||
+                                name == "incPoolSize"
+                            );
```
MoonlightSentinel, line 107: This seems to be unused?

src/core/internal/parseoptions.d
```diff
@@ -156,16 +173,64 @@ inout(char)[] find(alias pred)(inout(char)[] str)
     return null;
 }
 
-bool parse(T:size_t)(const(char)[] optname, ref inout(char)[] str, ref T res, const(char)[] errName)
+bool parse (T:size_t)(const(char)[] optname, ref inout(char)[] str, ref T res, const(char)[] errName)
+{
+    return parseSuff(optname, str, res, errName, false);
+}
+
+bool parseSuff(T:size_t)(const(char)[] optname, ref inout(char)[] str, ref T res, const(char)[] errName, bool mayHaveSuffix)
```
MoonlightSentinel, line 181:
```suggestion
bool parseSuff(T : size_t)(const(char)[] optname, ref inout(char)[] str, ref T res, const(char)[] errName, bool mayHaveSuffix)
```

src/core/internal/parseoptions.d
rainers, line 107: Can you use an UDA instead of having these special cases creeping in here?

src/core/gc/config.d
```diff
@@ -18,10 +18,10 @@ struct Config
...
     string gc = "conservative"; // select gc implementation conservative|precise|manual
 
-    size_t initReserve;      // initial reserve (MB)
-    size_t minPoolSize = 1;  // initial and minimum pool size (MB)
-    size_t maxPoolSize = 64; // maximum pool size (MB)
-    size_t incPoolSize = 3;  // pool size increment (MB)
+    size_t initReserve;      // initial reserve (bytes)
+    size_t minPoolSize = 1  << 20;  // initial and minimum pool size (bytes)
+    size_t maxPoolSize = 64 << 20;  // maximum pool size (bytes)
+    size_t incPoolSize = 3  << 20;  // pool size increment (bytes)
```
rainers, line 24: The help output should be adapted, too.

src/core/internal/parseoptions.d
```diff
@@ -156,16 +173,64 @@ inout(char)[] find(alias pred)(inout(char)[] str)
...
+
+        if (isdigit(c))
+            v = 10 * v + c - '0';
+        else // non-digit
+        {
+            if (mayHaveSuffix && i == len-1) // suffix
+            {
+                switch (c)
+                {
+                    case 'M':
```
rainers, line 202: While you're at it, I would add 'G', too.

src/core/internal/parseoptions.d
denizzzka: Used while parsing integers into `Config` (`CFG`) structure

MoonlightSentinel: How? This is the only occurence of `isSuffValue`.

denizzzka: You mean mark some `Config` fields by special UDA and then find it?

denizzzka: std.traits.hasUDA is not available in druntime

denizzzka: Can you explain more?

MoonlightSentinel: isSuffValue is set but never used. I assume you want to pass it as a parameter for `parseSuff` instead of `true`?

denizzzka:
> With possibly now working with large numbers overflow checking might be nice.

Probably, no one operates with such huge pool sizes. If you do not agreed let it be a separate PR. This patch is already becomes too big for such small improvement.

src/core/internal/parseoptions.d
```diff
@@ -156,16 +173,68 @@ inout(char)[] find(alias pred)(inout(char)[] str)
     return null;
 }
 
-bool parse(T:size_t)(const(char)[] optname, ref inout(char)[] str, ref T res, const(char)[] errName)
+bool parse(T : size_t)(const(char)[] optname, ref inout(char)[] str, ref T res, const(char)[] errName)
+{
+    return parseSuff(optname, str, res, errName, false);
+}
+
+bool parseSuff(T : size_t)(const(char)[] optname, ref inout(char)[] str, ref T res, const(char)[] errName, bool mayHaveSuffix)
```
MoonlightSentinel, line 181: You could simply add `mayHaveSuffix` as a default parameter to avoid this intermediate function (which isn't named appropriatly anyway because `mayHaveSuffix`  determines whether a suffix is accepted or not).

```suggestion
bool parse(T : size_t)(const(char)[] optname, ref inout(char)[] str, ref T res, const(char)[] errName, bool mayHaveSuffix = false)
```

denizzzka: This leads to adding same parameter to all another parse() functions

MoonlightSentinel: Not really, you only pass `mayHaveSuffix` if the option type is `size_t` which will match this overload anyway.

denizzzka: Can you make PR for this PR?

src/core/gc/config.d
```diff
@@ -55,6 +55,8 @@ struct Config
     parallel:N     - number of additional threads for marking (%lld)
     heapSizeFactor:N - targeted heap size to used memory ratio (%g)
     cleanup:none|collect|finalize - how to treat live objects when terminating (collect)
+
+    Memory-related values can use B, K or M suffixes.
```
rainers, line 59: I'd rather meant that the displayed values don't match the shown unit. For bonus points, it would be nice to show the suffix if possible for exact values (e.g. 0x4000 could show 16K).

src/core/internal/parseoptions.d
rainers: You can use `__traits(getAttributes, fieldT)` directly.

src/core/internal/parseoptions.d
rainers: Thanks, but 'G' is now missing in the descriptions.

changelog/gc_suffixes.dd
```diff
@@ -0,0 +1,12 @@
+Memory-related GC options stored in bytes
```
rainers, line 1: Storage is not a user visible change, I'd rather describe it as "Memory releated GC options can now be specified with higher accuracy" or similar.

src/core/gc/config.d
denizzzka: Ah, totally forgot about this

> For bonus points, it would be nice to show the suffix if possible for exact values (e.g. 0x4000 could show 16K).

I don’t plan to code in the next couple of days, maybe implement this in a separate PR?

src/core/internal/parseoptions.d
denizzzka: > You can use __traits(getAttributes, fieldT) directly.

Uhh. Too many code duplication with std.traits.
Anybody knows why it is impossible to use std.traits from druntime?

MoonlightSentinel: E.g. `--DRT-gcopt=parallel:16G` should issue an error that the suffix is not supported.

src/core/gc/config.d
denizzzka: > For bonus points, it would be nice to show the suffix if possible for exact values (e.g. 0x4000 could show 16K)

Done

src/core/internal/parseoptions.d
denizzzka: Ah, yes, thanks! Forgot to replace after debugging. Now I write patch with using of UDA, so, this code will be fixed

denizzzka: Done

src/core/internal/parseoptions.d
denizzzka: @MoonlightSentinel got your point, parseSuff removed. Thanks!

MoonlightSentinel: Thanks, sorry if my initial suggestion was misleading.

src/core/gc/config.d
```diff
@@ -12,16 +12,19 @@ import core.internal.parseoptions;
 
 __gshared Config config;
 
+/// UDA for field treated as memory value
+struct MemVal {}
```
rainers, line 16: Thanks. `MemVal` should be declared in parseoptions.d, so there is no circular dependency and it is available for other options, too. The conversion methods could be static members of the struct.

denizzzka: > Thanks. MemVal should be declared in parseoptions.d, so there is no circular dependency and it is available for other options, too.

Done

> The conversion methods could be static members of the struct.

I don’t like that. In the future, maybe someone else uses these functions somewhere.

denizzzka:
auto-tester stall?

PR #3124, opened 2020-06-06
cannot inline setPointerBitmapSmall()

WalterBright:
This is blocking https://github.com/dlang/dmd/pull/11236

The pragma annotation is superfluous anyway because the code wasn't compiled with `-inline`.

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -3449,7 +3449,6 @@ struct Pool
         }
     }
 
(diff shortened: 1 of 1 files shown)


rainers:
> The pragma annotation is superfluous anyway because the code wasn't compiled with `-inline`

druntime is compiled with `-inline`, see for example https://github.com/dlang/druntime/blob/cc361274ed9049a3f33b5f2143422d31a65a5935/posix.mak#L88

Why did it work before?

PR #3125, opened 2020-06-06
Please, no memset and friends in core.lifetime,

radcapricorn:
(and please, no TypeInfo either)

Language spec does not address handling of otherwise well-known functions like memcpy and memset. Therefore, a D compiler is not required to replace them with appropriate codegen (which is exactly the case for dmd at the moment). That optimizing compilers approach this differently is incidental.

On the other hand, move(), emplace() et al are low-level constructs that deal with typed memory, providing the compiler with full awareness of data layout. Library code should favor expressing such constructs without relying on external dependencies. This does not preclude implementation-specific behavior, but removes ambiguous behavior (i.e. that which is not covered by the spec).

A runtime typeid() call within a function that is statically aware of the type is simply unnecessary. I'm not sure about the history here, it may have been required for nested types at some point, but with current language the initializer can be resolved without a runtime query.

### DIFF
--- a/src/core/internal/lifetime.d
+++ b/src/core/internal/lifetime.d
@@ -88,30 +88,33 @@ if (is(UT == core.internal.traits.Unqual!UT))
 Emplaces T.init.
 In contrast to `emplaceRef(chunk)`, there are no checks for disabled default
 constructors etc.
+The keepContext template argument instructs this function to keep a nested struct's
+context pointer intact (used by move et al.).
 +/
-void emplaceInitializer(T)(scope ref T chunk) nothrow pure @trusted
+void emplaceInitializer(T,bool keepContext = false)(scope ref T chunk) nothrow pure @trusted
     if (!is(T == const) && !is(T == immutable) && !is(T == inout))
 {
     import core.internal.traits : hasElaborateAssign;
 
-    static if (!hasElaborateAssign!T && __traits(compiles, chunk = T.init))
+    static if (keepContext) {
+        static assert(is(T == struct) && __traits(isNested, T),
+                      "keepContext should only be set to true when keeping a nested struct's context pointer is required");
+        enum sizeToBlit = T.sizeof - (void*).sizeof;
+    } else
+        enum sizeToBlit = T.sizeof;
+
+    static if (!keepContext && !hasElaborateAssign!T && __traits(compiles, chunk = T.init))
     {
         chunk = T.init;
     }
     else static if (__traits(isZeroInit, T))
     {
-        static if (is(T U == shared U))
-            alias Unshared = U;
-        else
-            alias Unshared = T;
-
-        import core.stdc.string : memset;
-        memset(cast(Unshared*) &chunk, 0, T.sizeof);
+        *cast(ubyte[sizeToBlit]*) &chunk = 0;
     }
     else
     {
         // emplace T.init (an rvalue) without extra variable (and according destruction)
-        alias RawBytes = void[T.sizeof];
+        alias RawBytes = void[sizeToBlit];
 
         static union U
         {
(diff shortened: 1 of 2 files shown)


n8sh:
Have you verified this doesn't cause a performance regression?

radcapricorn:
> Have you verified this doesn't cause a performance regression?

With dmd, it of course does.
Artificial benchmark (linearly moving objects of sizes in first 10 powers of 2 (so 2 - 1024)) shows huge (5x) increase in speed for small objects (up to 16 bytes), but a ~1.5x slowdown for objects of larger sizes.
The former shows the necessity of removing the call to memcpy, the latter - appalling performance of array copies, which I think should be filed against dmd.

With ldc, it does not.

ibuclaw:
`Sarray = 0` still gets lowered to a memset too. 

The difference you see is that the compiler knows what the implicit memset is and inlines the implementation if it knows how to. Sufficiently large enough objects may still leave the memset in place.

(gdc recognises functions in core.stdc.* as built-ins so you'll see no speed up either way).

radcapricorn:
I'm well aware of that, @ibuclaw :) I didn't make this patch expecting performance improvement in an optimizing compiler. I made it to remove explicit C and TypeInfo dependencies from fundamental functions (which really should be intrinsics, but that's not the issue at hand), which is what is stated in the patch comment.

However, improvement on small objects (as well as slowdown for larger object) I see for dmd is not that, it is the exact opposite. It's due to dmd *not* inserting a call to memcpy, which otherwise it does unconditionally. That it generates poor code should be addressed in the compiler itself.

denizzzka:
To understand what is happening: @radcapricorn can you provide a rationale for this PR or link to discussion?

MoonlightSentinel:
This PR contains two independent changes, they should be discussed in separate PRs

src/core/internal/lifetime.d
```diff
@@ -88,30 +88,33 @@ if (is(UT == core.internal.traits.Unqual!UT))
...
     else static if (__traits(isZeroInit, T))
     {
-        static if (is(T U == shared U))
-            alias Unshared = U;
-        else
-            alias Unshared = T;
-
-        import core.stdc.string : memset;
-        memset(cast(Unshared*) &chunk, 0, T.sizeof);
+        *cast(ubyte[sizeToBlit]*) &chunk = 0;
```
denizzzka, line 112: It is makes sense to cover this into pretty-looking template?

Then it will be possible to replace all `memset` entries in whole druntime.

radcapricorn: I'm not against such per se, but from the history of this particular function, the additional static if for isZeroInit above it was introduced to *reduce* the amount of additional types.

Geod24:
Please, provide a useful commit title / description. The current one doesn't tell people anything about why this change is useful.

radcapricorn:
> Please, provide a useful commit title / description. The current one doesn't tell people anything about why this change is useful.

Removing C and typeid calls from generic low-level functions is not self-explanatory?

ibuclaw:
> Removing C and typeid calls from generic low-level functions 

Well, it doesn't remove the C call, as there's still an implicit one done in its place.  I had got the impression from the PR title that you wanted to _remove_ libc and rtti dependencies, not hide them.

radcapricorn:
It removes the C calls, and associated dependencies, from the *library code*. There's no hiding. Whether *a particular compiler* inserts them regardless would be up to that compiler and whatever associated user control is provided for it.

As for rtti - that one is removed for good.

src/core/internal/lifetime.d
```diff
@@ -88,30 +88,33 @@ if (is(UT == core.internal.traits.Unqual!UT))
...
+void emplaceInitializer(T,bool keepContext = false)(scope ref T chunk) nothrow pure @trusted
     if (!is(T == const) && !is(T == immutable) && !is(T == inout))
 {
     import core.internal.traits : hasElaborateAssign;
 
-    static if (!hasElaborateAssign!T && __traits(compiles, chunk = T.init))
+    static if (keepContext) {
+        static assert(is(T == struct) && __traits(isNested, T),
+                      "keepContext should only be set to true when keeping a nested struct's context pointer is required");
+        enum sizeToBlit = T.sizeof - (void*).sizeof;
```
kinke, line 102: This wasn't and isn't fully correct and should probably be `enum sizeToBlit = T.tupleof[$-1].offsetof`:
```D
void foo()
{
    int x;
	struct S
	{
    	align(64) void* data;
        void bar() { x = 123; }
	}
    static assert(S.sizeof == 64);
    static assert(S.tupleof[$-1].offsetof == size_t.sizeof);
}
```

RazvanN7:
@radcapricorn Do you intend on taking this to the finish line?

RazvanN7:
Please reopen if you want to pick this up.

PR #3126, opened 2020-06-07
fix signed multiply overflow detection

WalterBright:
Dividing `0x8000_0000_000_0000 / -1L` causes an arithmetic exception in 64 bit code. When the `muls()` unittests were inlined, the compiler tried to constant fold that division, and crashed.

Don't need more tests, as the newer dmd should inline the existing unittests.


WalterBright:
Reference:

https://www.informit.com/articles/article.aspx?p=1959565&seqNum=13 (look at the signed division section)

WalterBright:
Blocking https://github.com/dlang/dmd/pull/11236

PR #3130, opened 2020-06-09
workaround Issue 20906 - unnecessary divide-by-zero errors when const…

WalterBright:
…ant folding short circuits

Doesn't look like this bug will be resolved anytime soon, see https://github.com/dlang/dmd/pull/11252

In the meantime, need to move forward with https://github.com/dlang/dmd/pull/11236

### DIFF
--- a/src/core/checkedint.d
+++ b/src/core/checkedint.d
@@ -567,7 +567,7 @@ unittest
 }
 
 /// ditto
-pragma(inline, true)
+pragma(inline, false) // false as workaround to http://issues.dlang.org/show_bug.cgi?id=20906
 long muls()(long x, long y, ref bool overflow)
 {
     immutable long r = cast(ulong)x * cast(ulong)y;


WalterBright:
Baffling log file errors that have seemingly no connection with this PR.

MoonlightSentinel:
The azure failure is an unrelated (it's a phobos bug that rarely manifests inside of `run.d`).

src/core/checkedint.d
```diff
@@ -567,7 +567,7 @@ unittest
 }
 
 /// ditto
-pragma(inline, true)
+pragma(inline, false) // false as workaround to http://issues.dlang.org/show_bug.cgi?id=20906
 long muls()(long x, long y, ref bool overflow)
```
UplinkCoder, line 571: Why is this function a template?

WalterBright:
> The azure failure is an unrelated (it's a phobos bug that rarely manifests inside of run.d).

Is there a bugzilla for it? Any fix?

WalterBright:
> which would mean accepting introducing a regression

How so?

src/core/checkedint.d
WalterBright: There was a reason, but I don't recall what it was at the moment, and it wasn't documented.

MoonlightSentinel:
> Is there a bugzilla for it? Any fix?

Working on it, gonna open a PR soon

Geod24:
> How so?

If you fix the bug, we can pull your DMD PR and this workaround becomes useless.
If you don't fix the bug, we shouldn't pull your DMD PR, and there's no point in pulling this workaround either.

src/core/checkedint.d
n8sh: [It was to make core.checkedint usable in betterC.](https://github.com/dlang/druntime/pull/2426)  After [#11236](https://github.com/dlang/dmd/pull/11236) is pulled, if DMD is taught how to inline functions like this, `pragma(inline, true)` would be a better solution than a dummy template.

WalterBright:
> If you don't fix the bug, we shouldn't pull your DMD PR

Why?

> If you fix the bug

I could use some help with that (finding the piece of code that causes it).

kinke:
In any case, such an ugly workaround should be restricted to DMD; no reason to pessimize LDC etc. because of a DMD limitation.

WalterBright:
This can be closed when https://github.com/dlang/dmd/pull/11252 passes.

@kinke this function has likely never been inlined, otherwise the failure would have showed up before.

kinke:
> this function has likely never been inlined

Just because it wasn't for DMD doesn't mean it wasn't/isn't with LDC - if it isn't clear by now, LDC does *not* use the DMD AST inliner at all.

WalterBright:
No longer needed.

PR #3134, opened 2020-06-11
document template rationale in checkedint.d

WalterBright:
introduced here https://github.com/dlang/druntime/pull/2426

### DIFF
--- a/src/core/checkedint.d
+++ b/src/core/checkedint.d
@@ -33,6 +33,10 @@ nothrow:
 @nogc:
 pure:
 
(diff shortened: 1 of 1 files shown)


Geod24:
Can you move this to the module documentation since it's relevant to the users and not just the maintainers ? Perhaps even a `BetterC:` section ?

RazvanN7 (approved):
@Geod24 I have addressed your review. I haven't created a `BetterC` section since this module those not have any sections. I think this should be fine as is.

PR #3131, opened 2020-06-09
Split Thread into Thread implementation and common ThreadBase

denizzzka:
(Work in progress)

PR #3135, opened 2020-06-11
Split osthread.d

denizzzka:
Rationale:

* This PR opens ways to write custom Thread implementations for support another architectures
* osthread.d contains incredible 4k lines

Split occurs along OS-specific code (`Thread` class, 'osthread.d', ~2.7k lines) and OS-independent code (`ThreadBase` abstract class, `threadbase.d`, ~1.4k lines). Two other small files is for: one is for supporting purposes and to avoid ctors/dtors cyclic dependencies and one fits `ThreadGroup` class what also moved from osthread.d.

PR was made, primarily, just by copying and pasting code and connecting it if it is need by changing attributes or passthrough from implementation to base class by `extern` symbols. Thus, code behaviour was not changed.

At the next stage it will be possible to move some `Thread` methods to `ThreadBase`. Currently it is impractical to do this in order to avoid complicate of analyzing of this PR.

This branch was created by squashing of draft https://github.com/dlang/druntime/pull/3131

### DIFF
--- a/src/core/thread/osthread.d
+++ b/src/core/thread/osthread.d
@@ -12,9 +12,15 @@
 
 module core.thread.osthread;
 
+import core.thread.threadbase;
 import core.thread.context;
+import core.thread.types;
+import core.atomic;
+import core.memory : GC;
 import core.time;
 import core.exception : onOutOfMemoryError;
+import core.internal.traits : externDFunc;
+
 
 ///////////////////////////////////////////////////////////////////////////////
 // Platform Detection and Memory Allocation
@@ -50,8 +56,6 @@ else version (D_InlineAsm_X86_64)
 
 version (Posix)
 {
-    import core.sys.posix.unistd;
-
     version (AsmX86_Windows)    {} else
     version (AsmX86_Posix)      {} else
     version (AsmX86_64_Windows) {} else
@@ -68,48 +72,40 @@ version (Posix)
     }
 }
 
-package(core.thread)
+version (Windows)
 {
-    static immutable size_t PAGESIZE;
-    version (Posix) static immutable size_t PTHREAD_STACK_MIN;
+    import core.stdc.stdint : uintptr_t; // for _beginthreadex decl below
+    import core.stdc.stdlib;             // for malloc, atexit
+    import core.sys.windows.basetsd /+: HANDLE+/;
+    import core.sys.windows.threadaux /+: getThreadStackBottom, impersonate_thread, OpenThreadHandle+/;
+    import core.sys.windows.winbase /+: CloseHandle, CREATE_SUSPENDED, DuplicateHandle, GetCurrentThread,
+        GetCurrentThreadId, GetCurrentProcess, GetExitCodeThread, GetSystemInfo, GetThreadContext,
+        GetThreadPriority, INFINITE, ResumeThread, SetThreadPriority, Sleep,  STILL_ACTIVE,
+        SuspendThread, SwitchToThread, SYSTEM_INFO, THREAD_PRIORITY_IDLE, THREAD_PRIORITY_NORMAL,
+        THREAD_PRIORITY_TIME_CRITICAL, WAIT_OBJECT_0, WaitForSingleObject+/;
+    import core.sys.windows.windef /+: TRUE+/;
+    import core.sys.windows.winnt /+: CONTEXT, CONTEXT_CONTROL, CONTEXT_INTEGER+/;
+
+    private extern (Windows) alias btex_fptr = uint function(void*);
+    private extern (C) uintptr_t _beginthreadex(void*, uint, btex_fptr, void*, uint, uint*) nothrow @nogc;
+}
+else version (Posix)
+{
+    import core.stdc.errno;
+    import core.sys.posix.semaphore;
+    import core.sys.posix.stdlib; // for malloc, valloc, free, atexit
+    import core.sys.posix.pthread;
+    import core.sys.posix.signal;
+    import core.sys.posix.time;
+
+    version (Darwin)
+    {
+        import core.sys.darwin.mach.thread_act;
+        import core.sys.darwin.pthread : pthread_mach_thread_np;
+    }
 }
 
-shared static this()
-{
-    version (Windows)
-    {
-        SYSTEM_INFO info;
-        GetSystemInfo(&info);
-
-        PAGESIZE = info.dwPageSize;
-        assert(PAGESIZE < int.max);
-    }
-    else version (Posix)
-    {
-        PAGESIZE = cast(size_t)sysconf(_SC_PAGESIZE);
-        PTHREAD_STACK_MIN = cast(size_t)sysconf(_SC_THREAD_STACK_MIN);
-    }
-    else
-    {
-        static assert(0, "unimplemented");
-    }
-}
-
-private
-{
-    // interface to rt.tlsgc
-    import core.internal.traits : externDFunc;
-
-    alias rt_tlsgc_init = externDFunc!("rt.tlsgc.init", void* function() nothrow @nogc);
-    alias rt_tlsgc_destroy = externDFunc!("rt.tlsgc.destroy", void function(void*) nothrow @nogc);
-
-    alias ScanDg = void delegate(void* pstart, void* pend) nothrow;
-    alias rt_tlsgc_scan =
-        externDFunc!("rt.tlsgc.scan", void function(void*, scope ScanDg) nothrow);
-
-    alias rt_tlsgc_processGCMarks =
-        externDFunc!("rt.tlsgc.processGCMarks", void function(void*, scope IsMarkedDg) nothrow);
-}
+package extern(C) immutable size_t threadSizeof = Thread.sizeof;
 
 version (Solaris)
 {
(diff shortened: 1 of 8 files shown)


denizzzka:
Done, ready for review! Hehe :-)

thewilsonator:
`core.thread` seems to use an odd style of including a space after opening parentheses and and a space before closing parentheses e.g. 
```
Thread t = new Thread( fn ).start();
                      ^  ^
```
Please remove them on all the affected lines in the diff

denizzzka:
Is that really a good idea? I expected that someone would look to diff of changes, and such spaces removal will highlights additional lines that actually haven't changed.

denizzzka:
@thewilsonator done!

src/core/thread/types.d
```diff
@@ -0,0 +1,79 @@
...
+    version (Windows)
+        void delegate() nothrow cbDllUnload;
+}
+
+version (GNU)
+{
+    import gcc.builtins;
+
+    version (GNU_StackGrowsDown)
+        enum isStackGrowsDown = true;
```
kinke, line 40: This was renamed, not for the better.

src/core/thread/threadbase.d
```diff
@@ -0,0 +1,1374 @@
...
+
+    enum mutexAlign = classInstanceAlignment!Mutex;
+    enum mutexClassInstanceSize = __traits(classInstanceSize, Mutex);
+
+    extern(C) void* swapContext(void* newContext) nothrow @nogc;
+
+    extern extern(C) immutable size_t threadSizeof;
+
+    extern(C) void* getStackBottom() nothrow @nogc;
+    extern(C) void* getStackTop() nothrow @nogc;
```
kinke, line 90: `extern(C)` isn't a proper way to call functions from another module without importing it. These C functions are mostly used for functions invoked by the compiler; druntime-internal, there are helpers like `externDFunc`. `extern(D)` provides type safety and makes sure the functions don't clash with user code (that's why the C ones usually start with `_d_`).

IIRC, there are other places where similar stuff is handled by conceptually making `ThreadBase` a template taking a platform-specific `NativeContext` as template parameter, which exposes functionality like `getStack{Bottom,Top}()` etc. In the end, there'd be something like
```D
version (Windows)
    alias Thread = ThreadBase!WindowsContext;
else version (Posix)
    alias Thread = ThreadBase!PthreadContext;
...
```

src/core/thread/types.d
denizzzka: I renamed it because its "type" changed: `bool enum` is used instead of `version` because it is need to pass it between modules.

kinke: Okay, but then please change it to `isStackGrowingDown` or so.

src/core/thread/threadbase.d
denizzzka: Ok, will substitute it by externDFunc.

About `NativeContext`: yes. But in this PR I tried to do everything gradually and make as few changes as possible, just copy and pasting.

src/core/thread/osthread.d
```diff
@@ -68,48 +72,40 @@ version (Posix)
...
-    alias rt_tlsgc_destroy = externDFunc!("rt.tlsgc.destroy", void function(void*) nothrow @nogc);
-
-    alias ScanDg = void delegate(void* pstart, void* pend) nothrow;
-    alias rt_tlsgc_scan =
-        externDFunc!("rt.tlsgc.scan", void function(void*, scope ScanDg) nothrow);
-
-    alias rt_tlsgc_processGCMarks =
-        externDFunc!("rt.tlsgc.processGCMarks", void function(void*, scope IsMarkedDg) nothrow);
-}
+package extern(C) immutable size_t threadSizeof = Thread.sizeof;
```
rainers: You don't need this variable, the size of a class reference is always the same, i.e. the size of a pointer.

src/core/thread/threadbase.d
```diff
@@ -0,0 +1,1374 @@
...
+        tlsGCdataInit();
+    }
+
+    package void dataStorageDestroy() nothrow @nogc
+    {
+        rt_tlsgc_destroy( m_tlsgcdata );
+        m_tlsgcdata = null;
+    }
+
+    package void dataStorageDestroyIfAvail() nothrow @nogc
```
rainers, line 152: The usual naming convention is to start with the verb, e.g. `destroyDataStorage`.

src/core/thread/osthread.d
```diff
@@ -1415,1204 +965,613 @@ class Thread
         else version (Posix)
             sched_yield();
     }
+}
 
+private Thread toThread(ThreadBase t) @safe nothrow @nogc pure
+{
+    return cast(Thread) t;
```
rainers, line 974: As you don't expect this to fail, going through dynamic casting should be avoided. `cast(Thread) cast(void*) t` should be a noop for optimized code.

src/core/thread/types.d
```diff
@@ -0,0 +1,79 @@
...
+struct ll_ThreadData
+{
+    ThreadID tid;
+    version (Windows)
+        void delegate() nothrow cbDllUnload;
+}
+
+version (GNU)
+{
+    import gcc.builtins;
```
ibuclaw, line 37: Import is unused, perhaps you meant to put it in another module you split out?

denizzzka: Already putted, but forgot to remove from here.
Fixed.

src/core/thread/threadbase.d
denizzzka: Done, except `scanWindowsOnly` and `swapContext`. For some reason Windows version isn't links if extern(D) used.
I investigate this problem.

src/core/thread/types.d
denizzzka: Actually we don't know `GNU_StackGrowsDown` is really defined anywhere or not.

This is why I changed `version = StackGrowsDown` to enum `bool isStackGrowingDown`.

denizzzka: > Actually we don't know GNU_StackGrowsDown is really defined anywhere or not.

On the other hand, that’s all the compiler can tell us, so this sentence isn't make sense. :-)

ibuclaw: > <s>Already putted, but forgot to remove from here.</s>
> 
> Copied and pasted from the original `osthread.d` source. Maybe, in previous, `GNU_StackGrowsDown` was defined inside of `gcc.builtins`.
> 

You can't import version conditions from another module.

ibuclaw: > Actually we don't know `GNU_StackGrowsDown` is really defined anywhere or not.
> 

Version conditions can only be defined or undefined.

denizzzka:
About extern(D) for scanWindowsOnly:

Something wrong with mangling. Binaries contains two slightly different declarations:

```D
writeln(demangle("_D4core6thread8osthread15scanWindowsOnlyFNbDFNbE4core6thread10threadbase8ScanTypePvQcZvCQBoQBmQBi10ThreadBaseZv"));

writeln(demangle("_D4core6thread8osthread15scanWindowsOnlyFNbDFNbEQBuQBs10threadbase8ScanTypePvQcZvCQDcQDaQBi10ThreadBaseZv")); 
```

Both of them demangled into equal text:
```
nothrow void core.thread.osthread.scanWindowsOnly(void delegate(core.thread.threadbase.ScanType, void*, void*) nothrow, core.thread.threadbase.ThreadBase)
nothrow void core.thread.osthread.scanWindowsOnly(void delegate(core.thread.threadbase.ScanType, void*, void*) nothrow, core.thread.threadbase.ThreadBase)
```

Reproduceable on Linux too.

denizzzka:
(I don't know what to do with this)

kinke:
This is likely a compiler bug and reminds me of https://github.com/ldc-developers/ldc/issues/3098#issuecomment-506483327.

rainers:
> ```d
> //Extern:
> writeln(demangle("_D4core6thread8osthread15scanWindowsOnlyFNbDFNbE4core6thread10threadbase8ScanTypePvQcZvCQBoQBmQBi10ThreadBaseZv"));
> ```

This looks like a bug in `externDFunc`: the identifiers `core` and `thread` are not supposed to appear again, but be back referenced. An (ugly) workaround could be to use pragma(mangle) directly.

denizzzka:
@kinke @rainers ok, pragma(mangle, ..) is used directly for scanWindowsOnly

denizzzka:
auto-tester already passed twice but now run again. It started automatically for each day?

thewilsonator:
It is rerun after a merge of any other PR

denizzzka:
Is there any problem with merging of this PR?

andre2007:
If possible also the block comments `///////////////...` could be removed in this module. It looks really bad in inline help of VS Code.

denizzzka:
>  It looks really bad in inline help of VS Code.

Heck! I carefully kept this comments because I thought that they are used in some kind of generator!

wilzbach:
> > It looks really bad in inline help of VS Code.
> 
> Heck! I carefully kept this comments because I thought that they are used in some kind of ddoc generator!

Nope, sorry. They are just very old.

For reference, when refactoring it's really okay to change things to the current guidelines: https://dlang.org/dstyle.html#phobos_documentation

ibuclaw:
This broke builds for GDC.  I told you that you were missing an `import gcc.builtins` somewhere.

denizzzka:
@ibuclaw sorry, I saw it but did not realize what it affects GDC

It would be nice to add build of other compilers as optional for CI pass

PR #3138, opened 2020-06-17
Fix TypeInfo.opEquals for some templates

rmanthorpe:
https://issues.dlang.org/show_bug.cgi?id=20579

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -337,13 +337,11 @@ class TypeInfo
     override bool opEquals(Object o)
     {
         /* TypeInfo instances are singletons, but duplicates can exist
-         * across DLL's. Therefore, comparing for a name match is
-         * sufficient.
+         * across DLL's. Comparing for a name match is not sufficient.
+         * as the name is not necessarily unique. Better to fail early
+         * than compare equal when they are not.
          */
-        if (this is o)
-            return true;
-        auto ti = cast(const TypeInfo)o;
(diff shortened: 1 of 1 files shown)


rmanthorpe:
I wish that were true, but `typeid` already doesn't work across DLL boundaries and that `opEquals` does nothing to help it. The reason being the cast to from `Object` to `TypeInfo` fails because it relies on `TypeInfo.opEquals` to do the cast. All the `opEquals` does right now is break `typeid` where there wasn't a problem before.

You're right the real solution is better `name` but your suggestion still isn't good enough, you need fully qualified names of all template parameters like `std.traits.fullyQualifiedName` does, or even better just directly comparing mangles.

src/object.d
```diff
@@ -422,19 +420,20 @@ class TypeInfo
     @property immutable(void)* rtInfo() nothrow pure const @safe @nogc { return rtinfoHasPointers; } // better safe than sorry
 }
 
+// Bugzilla 20579
```
andralex, line 423:
```suggestion
// Bugzilla 20579: https://issues.dlang.org/show_bug.cgi?id=20579
```

src/object.d
```diff
@@ -422,19 +420,20 @@ class TypeInfo
...
+@system unittest
```
andralex, line 424: Could this be `@safe`?

kinke:
This is obsolete after #3527 now.

PR #3139, opened 2020-06-17
Fix dup and idup property in the presence of copy constructors

RazvanN7:
This is a first step to fixing: https://issues.dlang.org/show_bug.cgi?id=20879 . There are a ton of other places that need fixing, but I want to tackle these in small incremental changes.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3105,11 +3105,20 @@ private U[] _dup(T, U)(T[] a) // pure nothrow depends on postblit
     import core.stdc.string : memcpy;
 
     void[] arr = _d_newarrayU(typeid(T[]), a.length);
-    memcpy(arr.ptr, cast(const(void)*)a.ptr, T.sizeof * a.length);
+    static if (__traits(hasCopyConstructor, T))
+    {
+        _doCopyCtor(a, *cast(U[]*)&arr);
+    }
+    else
+    {
+        memcpy(arr.ptr, cast(const(void)*)a.ptr, T.sizeof * a.length);
+    }
+
     auto res = *cast(U[]*)&arr;
 
     static if (__traits(hasPostblit, T))
         _doPostblit(res);
+
     return res;
 }
 
@@ -3340,6 +3349,14 @@ private void _doPostblit(T)(T[] arr)
     }
 }
 
+private void _doCopyCtor(T, U)(T[] src, U[] dst)
+{
+    foreach (i, ref elem; src)
+    {
+        dst[i].__ctor(elem);
+    }
+}
+
 @safe unittest
 {
     static struct S1 { int* p; }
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -3421,6 +3438,31 @@ private void _doPostblit(T)(T[] arr)
...
+    static struct ISunpure { this(ref const typeof(this)) immutable @safe nothrow {} }
+    static struct ISthrow { this(ref const typeof(this)) immutable @safe pure {} }
+    static struct ISunsafe { this(ref const typeof(this)) immutable @system pure nothrow {} }
+    static assert( __traits(compiles, ()         { [].idup!ISunpure; }));
+    static assert(!__traits(compiles, () pure    { [].idup!ISunpure; }));
+    static assert( __traits(compiles, ()         { [].idup!ISthrow; }));
+    static assert(!__traits(compiles, () nothrow { [].idup!ISthrow; }));
+    static assert( __traits(compiles, ()         { [].idup!ISunsafe; }));
+    static assert(!__traits(compiles, () @safe   { [].idup!ISunsafe; }));
+}
```
wilzbach, line 3464: Shouldn't the tests also test that the code actually works, i.e. the copy constructor of all array elements gets called?

RazvanN7: I've added one test

src/object.d
```diff
@@ -3340,6 +3349,14 @@ private void _doPostblit(T)(T[] arr)
     }
 }
 
+private void _doCopyCtor(T, U)(T[] src, U[] dst)
+{
+    foreach (i, ref elem; src)
+    {
+        dst[i].__ctor(elem);
```
radcapricorn, line 3356: What if it throws? Or rather, what if it throws, exception is caught, and later GC collects the array? Will it call any dtors? All of them? On a partially-initialized array?

src/object.d
radcapricorn: Also missing implementation, and a test for a struct that contains pointers and has a `this(ref const typeof(this))` copy ctor. Current dup/idup won't compile with those (they should). `__traits(hasCopyConstructor, T)` is really not sufficient.

src/object.d
andralex: Yah, this needs care:

```d
size_t i = 0;
scope (failure)
{
    for (size_t j = 0; j < i; ++j)
        dst[j].__dtor();
    delete dst;
}
for (; i < src.length; ++i)
    dst[i].__ctor(elem);
```

andralex: Well not `delete` :)

radcapricorn: `__xdtor`. If exists. And probably, before this call, not a `_d_newarrayU`, but a `_d_newarrayiT`. ;)

radcapricorn: Direct call to __ctor doesn't seem to write .init. Bad things may happen (c).

edi33416: This should probably call [_d_arrayctor](https://github.com/dlang/druntime/blob/master/src/core/internal/array/construction.d#L24). Otherwise, we'll need to replicate the logic there as well. It seems that `_d_arrayctor` already implements the logic necessary to handle exceptions.

Also, `_d_arrayctor` uses [postblitRecurse](https://github.com/dlang/druntime/blob/master/src/core/internal/postblit.d#L28) which knows about the new copyCtor.

There is, however, an issue with `_d_arrayctor`: it does a [memcpy](https://github.com/dlang/druntime/blob/master/src/core/internal/array/construction.d#L56) before calling `postblitRecurse`, which defies the purpose of the copyCtor, but this is an easy fix (I'll open a PR for it)

PR #3141, opened 2020-06-21
fix Issue 18734 - bitnum parameter of core.bitop.bt should be signed

WalterBright:
The BT instructions take signed offsets, not unsigned ones.

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -267,7 +267,7 @@ unittest
  * (No longer an intrisic - the compiler recognizes the patterns
  * in the body.)
  */
-int bt(const scope size_t* p, size_t bitnum) pure @system
+int bt(const scope size_t* p, ptrdiff_t bitnum) pure @system
 {
     static if (size_t.sizeof == 8)
         return ((p[bitnum >> 6] & (1L << (bitnum & 63)))) != 0;


WalterBright:
Looks like `dlang.druntime (Windows_OMF x86)` is looking at an outdated import of `core/bitop.d`:

Error 42: Symbol Undefined __D4core5bitop2btFNaNbNiMxPkkZi

rainers:
> Looks like `dlang.druntime (Windows_OMF x86)` is looking at an outdated import of `core/bitop.d`:

The problem is that the new phobos.lib is copied into the test directory, and optlink uses this one if d_do_test is compiled by the host compiler.

WalterBright:
> The problem is that the new phobos.lib is copied into the test directory, and optlink uses this one if d_do_test is compiled by the host compiler.

The import path and the library used should be the same. Otherwise, we can never update the library interface.

rainers:
> The import path and the library used should be the same.

Sure. This workaround https://github.com/dlang/dmd/blob/master/.azure-pipelines/windows.sh#L117-L119 probably was set up when the tools were still built with the new compiler.

Mixing builds with different compilers is not so easy with optlink because it picks up information from sc.ini in its containing folder. It probably needs to be copied into the `generated` folder and an appropriate sc.ini to be generated.

kinke:
I fail to see a good reason why the druntime API should depend on the 'signature' of a particular x86 instruction, and why a bit index should all of a sudden be a signed `ptrdiff_t`.

n8sh:
>I fail to see a good reason why the druntime API should depend on the 'signature' of a particular x86 instruction

Isn't the primary purpose of core.bitop to expose various instructions?

RazvanN7:
@WalterBright This change doesn't really seem to provide any benefit. Should we close this?

PR #3142, opened 2020-06-22
Revise core.internal.array.equality

kinke:
* Use memcmp for more types (incl. pointers and some static arrays), and don't require both element types to be equivalent (e.g., use it for comparing `wchar[]` and `ushort[]` too).
* Simplify the non-memcmp branch - the previous code seemed to try to do what the compiler already does.
* The previous code for structs without custom `opEquals` basically enforced `-preview=fieldwise` via `.tupleof` comparison. This breaking change with 2.078 almost certainly wasn't intended (and not mentioned in the changelog) and led to undesirable inconsistent behavior.
This reverts that change in semantics.
* Avoid superfluously nested templates and convert the remaining `at()` helper to a module-level template to reduce the number of redundant instantiations.

### DIFF
--- a/src/core/internal/array/equality.d
+++ b/src/core/internal/array/equality.d
@@ -1,7 +1,7 @@
 /**
- * This module contains compiler support determining equality of dynamic arrays.
+ * This module contains compiler support determining equality of arrays.
  *
- * Copyright: Copyright Digital Mars 2000 - 2019.
+ * Copyright: Copyright Digital Mars 2000 - 2020.
  * License: Distributed under the
  *      $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0).
  *    (See accompanying file LICENSE)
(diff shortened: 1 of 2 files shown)


kinke:
Azure CI issue for the OMF job:
```
D:/a/1/dmd/tools/dmd2/windows/bin/dmd.exe -m32  -g -lowmem -i -Itools -version=NoMain -odtest_results -oftest_results/d_do_test.exe tools/d_do_test.d

OPTLINK (R) for Win32  Release 8.00.17
Copyright (C) Digital Mars 1989-2013  All rights reserved.
http://www.digitalmars.com/ctg/optlink.html
test_results\d_do_test.obj(d_do_test) 
 Error 42: Symbol Undefined __D4core8internal5array8equality__T8__equalsTxkTxkZQqFNaNbNiNfAxkQdZb
```

AFAICT, that's the DMD *host* compiler failing to generate a linkable d_do_test tool. A druntime change shouldn't affect this at all.

The unittest failure in https://github.com/CyberShadow/ae/blob/ae2afb17cae70c8df739fb0dccd00b6eef212015/utils/regex.d#L345-L346 is somewhat worrying. I'll try to reproduce that later with LDC.

kinke:
Those 2 lines in the CyberShadow/ae repo nicely show the current problem with diverging behavior for `__equals` (doing a field-wise comparison of a struct's fields (if the struct has no custom opEquals), i.e., basically enforcing `-preview=fieldwise`). This wrongly passes with master:
```D
Transformation expected = { type : Transformation.Type.replace, replace : { search : "from", replacement : "to", flags : "" } };
auto actual = splitRETransformation("s/from/to/");

Transformation[] expectedSlice = [expected];
assert(actual == expectedSlice);  // equal (lowered to __equals); unfortunately, this is tested

Transformation[1] expectedSarray = [expected];
assert(actual != expectedSarray); // NOT equal (not lowered)

assert(actual[0] != expected);    // single element NOT equal (compiled without -preview=fieldwise)
```

rainers:
> AFAICT, that's the DMD _host_ compiler failing to generate a linkable d_do_test tool. A druntime change shouldn't affect this at all.

That's probably the same issue as this one: https://github.com/dlang/druntime/pull/3141#issuecomment-647313190

kinke:
The ae project has been adapted (no tagged release available yet though). See the linked PR above for more context wrt. that issue, boiling down to:
```D
union Transformation
{
    string search;
}

void main()
{
    static immutable from = "from", from2 = "from2";
    Transformation[] a = [{ search : from }];
    Transformation[] b = [{ search : from2[0..4] }];
    assert(a == b);
}
```
failing the assertion now with this PR, as it used to before v2.078, presumably when `__equals` (with diverging semantics) was introduced.

schveiguy:
oops, didn't know there was already something for this instead of #3151.

Is there anything in this that still needs merging? I think it would need to at least be rebased now that the other PR is merged.

kinke:
Yeah, a rebase is now necessary, but it's still blocked by the missing new tag for @CyberShadow's `ae` repo.

src/core/internal/array/equality.d
```diff
@@ -10,125 +10,48 @@
...
-        foreach (const u; 0 .. lhs.length)
-        {
-            if (at(lhs, u) != at(rhs, u))
-                return false;
-        }
-        return true;
-    }
-    else static if (__traits(isIntegral, U1))
+    // Use memcmp if applicable to both element types and the size matches.
+    static if (T1.sizeof == T2.sizeof && useMemcmp!T1 && useMemcmp!T2)
```
n8sh, line 26: Need to also check `__traits(isUnsigned, T1) == __traits(isUnsigned, T2)` or else:

```d
ubyte[] a = [cast(ubyte) 0xff];
byte[] b = [cast(byte) 0xff];
assert(a != b); // Currently passes but will fail after this PR.
```

src/core/internal/array/equality.d
```diff
@@ -203,3 +126,48 @@ bool __equals(T1, T2)(T1[] lhs, T2[] rhs)
...
     assert(a1 != a2);
 }
+
+
+private:
+
+template isVoid(T)
+{
+    import core.internal.traits : Unqual;
+    enum isVoid = is(Unqual!T == void);
```
n8sh, line 136: From the perspective of the compiler frontend there anything to recommend this over `is(immutable T == immutable void)`?

MoonlightSentinel: What about `shared`?

n8sh: It works since `immutable(shared T)` == `immutable T`.

src/core/internal/array/equality.d
kinke: Right, thx. That only affects integers < 32 bit though, due to int promotion.

src/core/internal/array/equality.d
kinke: > From the perspective of the compiler frontend there anything to recommend this over is(immutable T == immutable void)?

Absolutely not, thx for the suggestion.

n8sh:
>Those 2 lines in the CyberShadow/ae repo nicely show the current problem with diverging behavior for __equals (doing a field-wise comparison of a struct's fields (if the struct has no custom opEquals), i.e., basically enforcing -preview=fieldwise).

I think everyone would agree that a fix for something like this should appear in the changelog.

kinke:
The `ae` repo is no blocker anymore (new tag available).

kinke:
Thx @n8sh. I'm sure you'll update and re-evaluate #3152 (sorry, I haven't found the time to perform the promised tests yet).

JohanEngelen:
Because it removed all the casting, this PR fixed this bug: https://issues.dlang.org/show_bug.cgi?id=21094.

PR #3145, opened 2020-06-26
object: Don't compile in argTypes on all GNU targets

ibuclaw:
GDC supports C-style variadic arguments on all backends supported by GCC as intrinsics.  There is no desire to support the run-time based va_arg that uses TypeInfo.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -44,7 +44,11 @@ version (D_ObjectiveC)
 version (Posix) public import core.attribute : gnuAbiTag;
 
 // Some ABIs use a complex varargs implementation requiring TypeInfo.argTypes().
-version (X86_64)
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -44,7 +44,11 @@ version (D_ObjectiveC)
 version (Posix) public import core.attribute : gnuAbiTag;
 
 // Some ABIs use a complex varargs implementation requiring TypeInfo.argTypes().
-version (X86_64)
+version (GNU)
+{
+    // Varargs handled using intrinsics.
```
kinke, line 49:
```suggestion
    // No support for TypeInfo-based core.vararg.va_arg().
```

PR #3148, opened 2020-07-02
core.internal.hash.bytesHash: in 64-bit builds use a 64-bit-oriented algorithm

n8sh:
Motivation: doesn't discard high 32 bits of seed; produces 64 bits of output; and is a bit faster.


n8sh:
```console
Test runnable/testaa2.d failed: 
expected:
----
foo()
foo() 2
foo() 3
foo() 4
c["foo"] = 3
c["bar"] = 4
Success
----
actual:
----
foo()
foo() 2
foo() 3
foo() 4
c["bar"] = 4
c["foo"] = 3
Success
```

A DMD test is failing because it depends on a specific AA traversal order. I am inclined to consider that erroneous based on https://dlang.org/spec/hash-map.html:

>The built-in associative arrays do not preserve the order of the keys inserted into the array. **In particular, in a foreach loop the order in which the elements are iterated is typically unspecified.**

Geod24:
Indeed. This test should be fixed.

MoonlightSentinel:
Please add a reference for this hash algorithm in a comment

src/core/internal/hash.d
```diff
@@ -725,6 +728,84 @@ private size_t bytesHash(bool dataKnownToBeAligned)(scope const(ubyte)[] bytes,
...
+            auto k1 =
+                   ((cast(ulong) data[0]) << 56) |
+                   ((cast(ulong) data[1]) << 48) |
+                   ((cast(ulong) data[2]) << 40) |
+                   ((cast(ulong) data[3]) << 32) |
+                   ((cast(ulong) data[4]) << 24) |
+                   ((cast(ulong) data[5]) << 16) |
+                   ((cast(ulong) data[6]) << 8)  |
+                    (cast(ulong) data[7]);
+        }
```
MoonlightSentinel, line 759: Couldn't this use the pointer read optimization as done for the 32 bit hash above?

n8sh: Now it does.

RazvanN7 (approved):
@n8sh maybe a rebase will fix the current failures.

PR #3149, opened 2020-07-05
fix test for issue 11393 introduced by #2436

rainers:
As noted by @WalterBright the test can fail spuriously because the tested pointer might actually reference a valid allocation base address for `memsize == 16`.

Now make sure, the used pointers are inside the area to reallocate.


WalterBright:
Thanks for the quick action!

PR #3150, opened 2020-07-07
fix Issue 21026 - add core.bitop.bswap(ushort)

WalterBright:
Prepare for adding this to the compiler as an intrinsic. It should generate an XCHG instruction.

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -495,6 +495,33 @@ struct BitRange
 }
 
 /**
+ * Swaps bytes in a 2 byte ushort.
+ * Params:
+ *      x = value
+ * Returns:
+ *      `x` with bytes swapped
+ */
+pragma(inline, false)
+ushort byteswap(ushort x) pure
+{
+    /* Calling it bswap(ushort) would break existing code that calls bswap(uint).
+     *
+     * This pattern is meant to be recognized by the dmd code generator.
+     * Don't change it without checking that an XCH instruction is still
+     * used to implement it.
+     * Inlining may also throw it off.
+     */
+    return cast(ushort) (((x >> 8) & 0xFF) | ((x << 8) & 0xFF00u));
+}
+
+unittest
+{
+    assert(byteswap(cast(ushort)0xF234) == 0x34F2);
+    static ushort xx = 0xF234;
+    assert(byteswap(xx) == 0x34F2);
+}
+
+/**
  * Swaps bytes in a 4 byte uint end-to-end, i.e. byte 0 becomes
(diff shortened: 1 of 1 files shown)


dnadlinger:
Hmm, maybe `bswap16`, `bswapUshort`, `shortBswap` or something like that would have been better. Having `byteswap` and `bswap` with no obvious connection seems confusing.

(Not everyone knows x86's `bswap` instruction – to them, would reading a piece of code that uses both maybe suggest that `byteswap` swaps bytes and `bswap` swaps bits?)

wilzbach:
There was a time when new symbols weren't merged within ten hours and with actual consideration about their name (as they are basically set in stone once released)...

WalterBright:
I don't mind if you want to rename them to bswap16, bswap32, and bswap64. Those functions should have separate names because they don't overload nicely. (The original sin was when bswap(ulong) came along https://github.com/dlang/druntime/pull/1480)

> Not everyone knows x86's bswap instruction

Consider where the word "blitting" came from? It came from the PDP-10 instruction BLT which stood for "BLock Transfer". It gradually evolved into a word.

Anyhow, core.bitop has numerous functions named after their corresponding instructions https://dlang.org/phobos/core_bitop.html which is appropriate. It builds on a programmer's existing vocabulary.

schveiguy:
I'm fine with bswap as a name. I can never imagine it being considered "bit swap", as I don't even know what that means or what you would do with it.

I think they should be renamed bswap16, bswap32, and bswap64, and deprecate bswap(ulong). bswap(uint) is probably fine to leave in there, as it's likely the most common one.

dnadlinger:
> I'm fine with bswap as a name. I can never imagine it being considered "bit swap", as I don't even know what that means or what you would do with it.

I'm fine with `bswap` as well – it's perhaps a bit more obscure than necessary for people not that familiar with low-level stuff, but the name has been in `core.bitop` for a long time, so no need to have a discussion about it now.

What I do find potentially misleading is having both `byteswap` and `bswap`, the former suggesting by exclusion that `bswap` in fact doesn't swap bytes.

n8sh:
>I think they should be renamed bswap16, bswap32, and bswap64, and deprecate bswap(ulong). bswap(uint) is probably fine to leave in there, as it's likely the most common one.

In my opinion the annoyance of asymmetrical naming between `bswap(uint)` / `bswap(ulong)` / `bswap16(ushort)` isn't sufficient justification to impose a maintenance burden on every program that has used `bswap(ulong)` in the 5 years it has existed in core.bitop.

PR #3151, opened 2020-07-08
Remove __ArrayEq which the compiler no longer uses as of https://github.com/dlang/dmd/pull/11212

n8sh:


kinke:
[My druntime follow-up was/is #3142.]

PR #3152, opened 2020-07-08
Reduce template function instantiations related to array equality

n8sh:
This is achieved by splitting off the all-scalar case ~~and by moving the `at` and `trustedCast` helper functions out of the body of `core.internal.array.equality.__equals`~~.

### DIFF
--- a/src/core/internal/array/equality.d
+++ b/src/core/internal/array/equality.d
@@ -14,7 +14,41 @@ module core.internal.array.equality;
 // * dynamic arrays,
 // * (most) arrays of different (unqualified) element types, and
 // * arrays of structs with custom opEquals.
+
+ // The scalar-only overload takes advantage of known properties of scalars to
+ // reduce template instantiation. This is expected to be the most common case.
+bool __equals(T1, T2)(scope const T1[] lhs, scope const T2[] rhs)
+@nogc nothrow pure @trusted
+if (__traits(isScalar, T1) && __traits(isScalar, T2))
+{
+    if (lhs.length != rhs.length)
+        return false;
+
+    static if (T1.sizeof == T2.sizeof
+        // Signedness needs to match for types that promote to int.
+        // (Actually it would be okay to memcmp bool[] and byte[] but that is
+        // probably too uncommon to be worth checking for.)
+        && (T1.sizeof >= 4 || __traits(isUnsigned, T1) == __traits(isUnsigned, T2))
+        && !__traits(isFloating, T1) && !__traits(isFloating, T2))
+    {
+        if (!__ctfe)
+        {
+            // This would improperly allow equality of integers and pointers
+            // but the CTFE branch will stop this function from compiling then.
+            import core.stdc.string : memcmp;
+            return lhs.length == 0 ||
+                0 == memcmp(cast(const void*) lhs.ptr, cast(const void*) rhs.ptr, lhs.length * T1.sizeof);
+        }
+    }
+
+    foreach (const i; 0 .. lhs.length)
+        if (lhs.ptr[i] != rhs.ptr[i])
+            return false;
+    return true;
+}
+
 bool __equals(T1, T2)(scope T1[] lhs, scope T2[] rhs)
+if (!__traits(isScalar, T1) || !__traits(isScalar, T2))
 {
     if (lhs.length != rhs.length)
         return false;


n8sh:
Pushing to fix a comment.

src/core/internal/array/equality.d
```diff
@@ -23,20 +23,50 @@ bool __ArrayEq(T1, T2)(T1[] a, T2[] b)
...
+        }
+    }
+    foreach (const i; 0 .. lhs.length)
+        if (lhs.ptr[i] != rhs.ptr[i])
+            return false;
+    return true;
+}
+
+pragma(inline, true)
+private @trusted ref R at(R)(return scope R[] r, size_t i) { return r.ptr[i]; }
```
n8sh, line 54: If I change the argument of `at` to be `inout` and its return type to be inferred (but still a reference) I get an error in a DMD test. It seems like that shouldn't happen so I am noting this as a reminder to maybe look into the cause later.

```console
Test compilable/interpret3.d failed.  The logged output:
/home/braddr/sandbox/at-client/pull-4107266-Linux_32_64/dmd/generated/linux/release/32/dmd -conf= -m64 -Icompilable  -fPIC  -odgenerated/compilable -ofgenerated/compilable/interpret3_0.o  -c compilable/interpret3.d 
compilable/interpret3.d(2914): Deprecation: `case` variables have to be `const` or `immutable`
compilable/interpret3.d(6313): Deprecation: identity comparison of static arrays implicitly coerces them to slices, which are compared by reference
/home/braddr/sandbox/at-client/pull-4107266-Linux_32_64/dmd/test/../../druntime/import/core/internal/array/equality.d(54): Error: reinterpreting cast from `immutable(string[2])*` to `inout(char[][2])*` is not supported in CTFE
/home/braddr/sandbox/at-client/pull-4107266-Linux_32_64/dmd/test/../../druntime/import/core/internal/array/equality.d(92):        called from here: `at(lhs, u)`
```

src/core/internal/array/equality.d
```diff
@@ -10,20 +10,56 @@
...
+        // Signedness needs to match for types that promote to int.
+        && (T1.sizeof >= 4 || __traits(isUnsigned, T1) == __traits(isUnsigned, T2))
+        && !__traits(isFloating, T1) && !__traits(isFloating, T2))
+    {
+        if (!__ctfe)
+        {
+            // This would improperly allow equality of integers and pointers
+            // but the CTFE branch will stop this function from compiling then.
+            import core.stdc.string : memcmp;
+            return lhs.length == 0 || 0 == memcmp(lhs.ptr, rhs.ptr, lhs.length * T1.sizeof);
```
UplinkCoder, line 32: not sure you need the length == 0 case.

UplinkCoder: it's an optimisation, for an uncommon case, therefore a slowdown for the common case.

n8sh: It's there because I am under the impression that `memcmp` has undefined behavior for null pointers. If not it can be removed.

UplinkCoder:
@n8sh Have you verified that this helps?

src/core/internal/array/equality.d
UplinkCoder: @n8sh you are correct.
memcpy on null is undefined

n8sh:
@UplinkCoder Yes, I wrote a long program that was nothing but array comparisons, and with the change compiling it takes about 23% as much time and about 22% as much memory.

src/core/internal/array/equality.d
```diff
@@ -10,20 +10,57 @@
...
+            return false;
+    return true;
+}
+
+pragma(inline, true)
+private @trusted R trustedCast(R, S)(return scope S[] r) { return cast(R) r; }
+pragma(inline, true)
+private @trusted ref at(T)(return scope T[] r, size_t i)
+    // exclude opaque structs due to https://issues.dlang.org/show_bug.cgi?id=20959
+    if (!(is(T == struct) && !is(typeof(T.sizeof))))
```
n8sh, line 47: Copied from https://github.com/dlang/druntime/pull/3142

kinke:
I've rebased and updated #3142 in the meantime too. I don't really want to rebase yet another time after pieces of it are ripped out into yet another separate PR (and each with its own dlang issue? Is that tracker spamming (ab)used for changelog generation?).

n8sh:
>(and each with its own dlang issue? Is that tracker spamming (ab)used for changelog generation?)

I am under the impression that any non-trivial PR should have an associated issue.

n8sh:
Anyway, if this PR is merged before that one (which seems possible because this PR isn't waiting on anyone else's work), I'd be fine with rebasing #3142 for you.

kinke:
It'd be nice to have a comparison with your test program, but the non-identical semantics (also wrt. memcmp application) don't allow a direct comparison. Anyway, getting rid of superfluous instantiations was obviously my main goal after seeing the original code.

n8sh:
If I remove all of this PR except the addition of the scalar-specific `__equals` definition and the addition of a template constraint, that would minimize conflict while still doing something useful performance-wise.

kinke:
For the time being, maybe, but my version, AFAICT, doesn't need any of this, so I'd basically just revert this and apply mine instead. In my version, there's no `trustedCast()` helper template at all anymore. `at` was extracted to the module-level, just like you did (but does the void->ubyte promotion in my case, instead of `trustedCast` and the `void[]` special cases for `__equals`). I don't use a 2nd `__equals`, but a little nested `trustedMemcmp` helper (in a `static if` for memcmp-able arrays only).

n8sh:
>It'd be nice to have a comparison with your test program, but the non-identical semantics (also wrt. memcmp application) don't allow a direct comparison.

With those provisos, yours compiles in 75% as much time and using 60% as much memory as the baseline.

n8sh:
>I don't use a 2nd `__equals`

Right, but doing so improves compilation speed without interfering with any of the stuff you *are* doing.

kinke:
Your version doesn't use memcmp for static array element types etc. though, that's why runtime performance would be interesting as well. Please post your test code as gist or so somewhere, so that I can experiment as well.

n8sh:
>Your version doesn't use memcmp for static array element types etc. though

Right, and static arrays are handled by the non-scalar version (since `__isScalar` is never true for static arrays) with your memcmp check etc.

>Please post your test code as gist or so somewhere, so that I can experiment as well.

Sure. You'll see that the code does nothing but test whether there is a net benefit in compilation time for the happy path given the additional work done in template resolution. My starting assumption was that array equality checks are primarily performed between arrays of scalars so that is the case I was focused on improving. https://pastebin.com/d1A5f07s

kinke:
Thx a lot, I'll look into it another time - I hope(d) to have covered your happy path optimization in a not-unnecessarily-complex `useMemcmp` check; if it doesn't, that's definitely an improvement of yours.

PR #3155, opened 2020-07-10
core.simd: Fix simd templates to take parameters by reference

ibuclaw:
Seems like nobody is using any of these templates, and #1900 should probably be reverted.  It looks like they are all broken anyway unless changed to accept parameters by reference.

Should be enough to get CI for dlang/dmd#7065 green.


ibuclaw:
I have no time to look into why win64 doesn't work, I'll just revert the originating PR then. (#3157)

PR #3158, opened 2020-07-12
TypeInfo_Enum: Fix equals/compare signature

Geod24:
All other TypeInfo derivatives use in, but this used scope const.


12345swordy:
The reason why it use const scope,  was because "in" was suppose to be "const scope" instead of just "const" there was talks regarding deprecating "in" in the past.

Geod24:
Then the whole `TypeInfo` hierarchy should have been updated, just not one specific overload.

Geod24:
Buildkite seems broken ?

PR #3159, opened 2020-07-13
take advantage of bswap(ulong) being an intrinsic

WalterBright:
Will now work with betterC without all the template rigamarole.

Also adds missing unittests.


WalterBright:
Blocked by https://github.com/dlang/dmd/pull/11412

PR #3160, opened 2020-07-14
Bindings to linux perf_event header.

maxhaton:
I am working towards a uarch-level benchmarking and tracing library for phobos, so: these are bindings to perf_event.h . eBPF coming soon.

The system call perf_event_open is not included by glibc so I have included a function to call it that includes system call numbers for X86, amd64 and both varieties of ARM. The bitfields were done by including Phobos bitmanips manually - there could potentially be namespace collisions due to the extern(C) but I chose to keep it simple for now. 

A simple example (Converted almost verbatim from the manpage) works fine but I am looking for a more thorough test suite (or at least one that covers most of the functions without converting hacky "kernel C" into D by hand).

```D
import core.sys.linux.perf_event;
import core.stdc.stdio;
import core.stdc.string;
import core.stdc.stdlib;
import std.stdio : writeln;
import core.sys.posix.sys.ioctl;
import core.sys.posix.unistd;
import core.sys.linux.unistd;
extern(C)
int main(int argc, char** argv)
{
    perf_event_attr pe;
    long count;
    int fd;

    memset(&pe, 0, perf_event_attr.sizeof);
    pe.type = perf_type_id.PERF_TYPE_HARDWARE;
    pe.size = perf_event_attr.sizeof;
    pe.config = perf_hw_id.PERF_COUNT_HW_INSTRUCTIONS;
    pe.disabled = 1;
    pe.exclude_kernel = 1;
    pe.exclude_hv = 1;

    fd = cast(int) perf_event_open(&pe, 0, -1, -1, 0);
    if (fd == -1)
    {
        fprintf(stderr, "Error opening leader %llx\n", pe.config);
        exit(EXIT_FAILURE);
    }

    ioctl(fd, PERF_EVENT_IOC_RESET, 0);
    ioctl(fd, PERF_EVENT_IOC_ENABLE, 0);

    puts("Measuring instruction count for this puts");

    ioctl(fd, PERF_EVENT_IOC_DISABLE, 0);
    read(fd, &count, long.sizeof);
    
    printf("Used %lld instructions\n", count);

    close(fd);
    return 0;
}
```

### DIFF
--- a/src/core/sys/linux/perf_event.d
+++ b/src/core/sys/linux/perf_event.d
@@ -1,0 +1,2441 @@
+/**
+  * D header file for perf_event_open system call.
+  * 
+  * Converted/Bodged from linux userspace header, comments included. 
+  *
+  * Copyright: Max Haughton 2020
+  * License: <a href="http://www.boost.org/LICENSE_1_0.txt">Boost License 1.0</a>.
+  * Authors: Max Haughton
+  */
+module core.sys.linux.perf_event;
+version (linux) :
+extern (C):
+@nogc:
+nothrow:
+@system:
+
+import core.sys.posix.sys.ioctl;
+import core.sys.posix.unistd;
+
+version(X86_64) {
+    enum __NR_perf_event_open = 298;
+} else version(X86) {
+    enum __NR_perf_event_open = 336;
+} else version(ARM) {
+    enum __NR_perf_event_open = 364;
+} else version(ARM64) {
+    enum __NR_perf_event_open = 241;
+}
+extern(C) extern long syscall (long __sysno, ...);
+static long
+perf_event_open(perf_event_attr *hw_event, pid_t pid,
+                int cpu, int group_fd, ulong flags)
+{
+    return syscall(__NR_perf_event_open, hw_event, pid, cpu,
+                  group_fd, flags);
+}
+//a la 
+//#include <linux/perf_event.h>
+/*
+ * User-space ABI bits:
+ */
+
+/*
+ * attr.type
+ */
+enum perf_type_id
+{
+    PERF_TYPE_HARDWARE = 0,
+    PERF_TYPE_SOFTWARE = 1,
+    PERF_TYPE_TRACEPOINT = 2,
+    PERF_TYPE_HW_CACHE = 3,
+    PERF_TYPE_RAW = 4,
+    PERF_TYPE_BREAKPOINT = 5,
+
+    PERF_TYPE_MAX = 6 /* non-ABI */
+}
+/*
+ * Generalized performance event event_id types, used by the
+ * attr.event_id parameter of the sys_perf_event_open()
+ * syscall:
+ */
+enum perf_hw_id
+{
+    /*
+    	 * Common hardware events, generalized by the kernel:
+    	 */
+    PERF_COUNT_HW_CPU_CYCLES = 0,
+    PERF_COUNT_HW_INSTRUCTIONS = 1,
+    PERF_COUNT_HW_CACHE_REFERENCES = 2,
+    PERF_COUNT_HW_CACHE_MISSES = 3,
+    PERF_COUNT_HW_BRANCH_INSTRUCTIONS = 4,
+    PERF_COUNT_HW_BRANCH_MISSES = 5,
+    PERF_COUNT_HW_BUS_CYCLES = 6,
+    PERF_COUNT_HW_STALLED_CYCLES_FRONTEND = 7,
+    PERF_COUNT_HW_STALLED_CYCLES_BACKEND = 8,
+    PERF_COUNT_HW_REF_CPU_CYCLES = 9,
+
+    PERF_COUNT_HW_MAX = 10 /* non-ABI */
+}
+
+/*
+ * Generalized hardware cache events:
+ *
+ *       { L1-D, L1-I, LLC, ITLB, DTLB, BPU, NODE } x
+ *       { read, write, prefetch } x
+ *       { accesses, misses }
+ */
+enum perf_hw_cache_id
+{
+    PERF_COUNT_HW_CACHE_L1D = 0,
+    PERF_COUNT_HW_CACHE_L1I = 1,
+    PERF_COUNT_HW_CACHE_LL = 2,
+    PERF_COUNT_HW_CACHE_DTLB = 3,
+    PERF_COUNT_HW_CACHE_ITLB = 4,
+    PERF_COUNT_HW_CACHE_BPU = 5,
+    PERF_COUNT_HW_CACHE_NODE = 6,
+
+    PERF_COUNT_HW_CACHE_MAX = 7 /* non-ABI */
+}
+
+enum perf_hw_cache_op_id
+{
+    PERF_COUNT_HW_CACHE_OP_READ = 0,
+    PERF_COUNT_HW_CACHE_OP_WRITE = 1,
+    PERF_COUNT_HW_CACHE_OP_PREFETCH = 2,
+
+    PERF_COUNT_HW_CACHE_OP_MAX = 3 /* non-ABI */
+}
+
+enum perf_hw_cache_op_result_id
+{
+    PERF_COUNT_HW_CACHE_RESULT_ACCESS = 0,
+    PERF_COUNT_HW_CACHE_RESULT_MISS = 1,
+
+    PERF_COUNT_HW_CACHE_RESULT_MAX = 2 /* non-ABI */
+}
+
+/*
+ * Special "software" events provided by the kernel, even if the hardware
+ * does not support performance events. These events measure various
+ * physical and sw events of the kernel (and allow the profiling of them as
+ * well):
+ */
+enum perf_sw_ids
+{
+    PERF_COUNT_SW_CPU_CLOCK = 0,
+    PERF_COUNT_SW_TASK_CLOCK = 1,
+    PERF_COUNT_SW_PAGE_FAULTS = 2,
+    PERF_COUNT_SW_CONTEXT_SWITCHES = 3,
+    PERF_COUNT_SW_CPU_MIGRATIONS = 4,
+    PERF_COUNT_SW_PAGE_FAULTS_MIN = 5,
+    PERF_COUNT_SW_PAGE_FAULTS_MAJ = 6,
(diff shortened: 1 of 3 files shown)


WalterBright:
I'm seeing the same buildcite/druntime ldc-developers/ldc failure in other PRs. What broke it?

maxhaton:
@WalterBright as far as I can see it's a style issue in the comments from the C header. I'll push something to fix it soon - if I can't fix it easily I'll just strip them given that the ddoc for the sys.linux packages isn't generated anyway as far as I can see.

maxhaton:
I removed the comments, passes the checks on my machine now.

src/core/sys/linux/perf_event.d
```diff
@@ -0,0 +1,1873 @@
...
+}
+else version (ARM)
+{
+    enum __NR_perf_event_open = 364;
+}
+else version (ARM64)
+{
+    enum __NR_perf_event_open = 241;
+}
+extern (C) extern long syscall(long __sysno, ...);
```
Geod24, line 36: I'm pretty sure we have a binding for this

src/core/sys/linux/perf_event.d
```diff
@@ -0,0 +1,1873 @@
...
+    }
+
+    ulong sample_type;
+    ulong read_format;
+    private ulong _disabled_inherit_pinned_exclusive_exclude_user_exclude_kernel_exclude_hv_exclude_idle_mmap_comm_freq_inherit_stat_enable_on_exec_task_watermark_precise_ip_mmap_data_sample_id_all_exclude_host_exclude_guest_exclude_callchain_kernel_exclude_callchain_user_mmap2_comm_exec_use_clockid_context_switch_write_backward_namespaces___reserved_1;
+    @property ulong disabled() @safe pure nothrow @nogc const
+    {
+        auto result = (_disabled_inherit_pinned_exclusive_exclude_user_exclude_kernel_exclude_hv_exclude_idle_mmap_comm_freq_inherit_stat_enable_on_exec_task_watermark_precise_ip_mmap_data_sample_id_all_exclude_host_exclude_guest_exclude_callchain_kernel_exclude_callchain_user_mmap2_comm_exec_use_clockid_context_switch_write_backward_namespaces___reserved_1 & 1U) >> 0U;
+        return cast(ulong) result;
+    }
```
Geod24, line 282: What ?

src/core/sys/linux/perf_event.d
```diff
@@ -0,0 +1,1873 @@
...
+enum PERF_ATTR_SIZE_VER1 = 72;
+enum PERF_ATTR_SIZE_VER2 = 80;
+enum PERF_ATTR_SIZE_VER3 = 96;
+
+enum PERF_ATTR_SIZE_VER4 = 104;
+enum PERF_ATTR_SIZE_VER5 = 112;
+
+struct perf_event_attr
+{
+
```
Geod24, line 262: Extra whitespace

src/core/sys/linux/perf_event.d
```diff
@@ -0,0 +1,1873 @@
...
+        _mispred_predicted_in_tx_abort_cycles_type_reserved = cast(
+                typeof(_mispred_predicted_in_tx_abort_cycles_type_reserved))(
+                (_mispred_predicted_in_tx_abort_cycles_type_reserved & (
+                -1 - cast(typeof(_mispred_predicted_in_tx_abort_cycles_type_reserved)) 1U)) | (
+                (cast(typeof(_mispred_predicted_in_tx_abort_cycles_type_reserved)) v << 0U) & 1U));
+    }
+
+    enum ulong mispred_min = cast(ulong) 0U;
+    enum ulong mispred_max = cast(ulong) 1U;
+    @property ulong predicted() @safe pure nothrow @nogc const
```
Geod24, line 1757: Since the whole thing is already `nothrow @nogc` you don't need those

Geod24:
Removing the comment is not really a solution IMO, they add a lot of value. What was the failure ?

maxhaton:
@Geod24 it was a whitespace check in the CI. The comments removal was easier than manually going through several hundred lines of comments - the ABI of perf_event is completely stable at this point so the documentation is in the manpages/online.

src/core/sys/linux/perf_event.d
maxhaton: It's a mixin from the Phobos bitmanip template. It's not a public interface so I didn't bother renaming it.

src/core/sys/linux/perf_event.d
```diff
@@ -0,0 +1,1873 @@
+/*
+* D header file for perf_event_open system call.
+*
+* Converted/Bodged from linux userspace header
```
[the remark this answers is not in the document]
maxhaton: Ha. I'll get rid of that tomorrow. Although the docs aren't generated for this folder if I'm not mistaken.

wilzbach:
I'm sorry, but how is removing trailing whitespace going to take more than ten seconds?

atilaneves:
> @Geod24 it was a whitespace check in the CI. The comments removal was easier than manually going through several hundred lines of comments - the ABI of perf_event is completely stable at this point so the documentation is in the manpages/online.

Open in Emacs, `M-x delete-trailing-whitespace`, done.

src/core/sys/linux/perf_event.d
```diff
@@ -0,0 +1,1873 @@
...
+
+    union
+    {
+        ulong sample_period;
+        ulong sample_freq;
+    }
+
+    ulong sample_type;
+    ulong read_format;
+    private ulong _disabled_inherit_pinned_exclusive_exclude_user_exclude_kernel_exclude_hv_exclude_idle_mmap_comm_freq_inherit_stat_enable_on_exec_task_watermark_precise_ip_mmap_data_sample_id_all_exclude_host_exclude_guest_exclude_callchain_kernel_exclude_callchain_user_mmap2_comm_exec_use_clockid_context_switch_write_backward_namespaces___reserved_1;
```
NilsLankila, line 277: seriously ? It is private so the mangling does not matter. Reduce the size of this identifier, please.

NilsLankila: eventually you can keep the long stuff in a comment, as it is certainly a very very self documenting name.

maxhaton:
Fixed it. It wasn't trailing whitespace, it was tabs vs spaces but somehow only in the comments. I was under the impression that it was all spaces, but apparently not - it is now.

maxhaton:
@Geod24 everything passes now, feedback is incorporated, anything else?

src/core/sys/linux/perf_event.d
```diff
@@ -0,0 +1,2259 @@
...
+}
+else version (X86)
+{
+        enum __NR_perf_event_open = 336;
+}
+else version (ARM)
+{
+        enum __NR_perf_event_open = 364;
+}
+else version (ARM64)
```
n8sh, line 29:
```suggestion
else version (AArch64)
```

The [official naming](https://dlang.org/spec/version.html) is not symmetric with `version (ARM)`.

n8sh:
Except for in the definition of `perf_callchain_context` it looks like every `cast(ulong)` is unnecessary.

src/core/sys/linux/perf_event.d
maxhaton: Good spot. The [kernel](https://www.kernel.org/doc/html/latest/arm64/index.html) refers to it as ARM64.

src/core/sys/linux/perf_event.d
maxhaton: There is not. There are a few PRs but none of them made it in #2820

RazvanN7 (approved):
@maxhaton Please rebase and let's get this in.

maxhaton:
> @maxhaton Please rebase and let's get this in.

@RazvanN7 Blast from the past, but will do

Geod24:
Damn I thought this was merged a while ago.

src/core/sys/linux/perf_event.d
```diff
@@ -0,0 +1,2468 @@
...
+        enum __NR_perf_event_open = 364;
+}
+else version (AArch64)
+{
+        enum __NR_perf_event_open = 241;
+}
+else
+{
+        static assert(0, "Architecture not supported");
+}
```
ibuclaw, line 36: We should really be rejecting PRs that don't cover all primary supported linux platforms.  It's annoying to be running through some finishing line bootstrap tests, and I can't even build druntime anymore on PPC.

src/core/sys/linux/perf_event.d
```diff
@@ -0,0 +1,2468 @@
...
+        //     ulong, "mmap2", 1,
+        //     ulong, "comm_exec", 1,
+        //     ulong, "use_clockid", 1,
+        //     ulong, "context_switch", 1,
+        //     ulong, "write_backward", 1,
+        //     ulong, "namespaces", 1,
+        //     ulong, "__reserved_1", 35));
+        private ulong perf_event_attr_bitmanip;
+        ///
+        @property ulong disabled() @safe pure nothrow @nogc const
```
ibuclaw, line 452: There's functions here and no unittests?  How can we vet that this actually works?

src/core/sys/linux/perf_event.d
Geod24: Well that's why we need to extend the CI here to include you.

maxhaton:
If you can find a platform with perf_event turned on I can get a testsuite
going but I'm guessing paranoid is set to the NO_WAY_JOSE setting.

maxhaton:
I completely forgot I did that, my bad, do we have syscall tables I can
rely upon at build time?

ibuclaw:
> I completely forgot I did that, my bad, do we have syscall tables I can rely upon at build time?

In the `version (X86_64)` path, `D_X32` (`-mx32`) adds the magic number `0x40000000` to `298`.

For all others (HPPA, MIPS, PPC, RISCV, S390, SPARC), you can grab them from musl or glibc sources.

src/core/sys/linux/perf_event.d
```diff
@@ -0,0 +1,2468 @@
...
+{
+        ///
+        union perf_mem_data_src
+        {
+                ///
+                ulong val;
+
+                struct
+                {
+                        import std.bitmanip : bitfields;
```
ibuclaw, line 2029: $#^*% !!! ???

PR #3161, opened 2020-07-15
core.bitop: use recognized pattern for rotates

WalterBright:
https://github.com/dlang/druntime/pull/1495 says that DMD doesn't recognize rotate patterns. But it does, it recognizes the standard pattern. This #1495 uses an unusual pattern, and is not recognized by DMD.

This fixes it to use the usual pattern which DMD recognizes.


andralex:
I hope you took a look at the generated code to make sure!

WalterBright:
Yes, yes I did :-)

PR #3162, opened 2020-07-17
Issue 19524 revisited - Revert now-unnecessary templatization of core.checkedint

n8sh:
Now that pragma(inline, true) works without -inline it should no longer be necessary for these to be templates for them to work in betterC.

### DIFF
--- a/src/core/checkedint.d
+++ b/src/core/checkedint.d
@@ -47,7 +47,7 @@ pure:
  */
 
 pragma(inline, true)
-int adds()(int x, int y, ref bool overflow)
+int adds(int x, int y, ref bool overflow)
 {
     long r = cast(long)x + cast(long)y;
     if (r < int.min || r > int.max)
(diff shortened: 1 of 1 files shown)


MoonlightSentinel:
It would be nice to have those functions actually used inside of an `@betterC` unittest (given that D differs from C++ regarding inline)

ibuclaw:
Ping @kinke as signature change may break downstream compiler's ability to recognise intrinsic.

RazvanN7 (commented):
@n8sh Could you add some @betterC unittetests to make sure anything doesn't break?

RazvanN7:
Actually, I think it's better that we leave these functions as templates since we want to make druntime a header only library anyway. Until then, having them as templates makes druntime smaller.

RazvanN7:
I assumed that all foundation members, since one of our top priorities from last year was to lower the druntime hooks to templates and get rid of typeinfo.

kinke:
druntime isn't just hooks invoked by the compiler. Converting those to templates (where it makes sense - i.e., not abused for crappy betterC workarounds) is fine. But druntime also contains some state that at least currently needs to be central/shared for the whole process (e.g., list of DSOs and threads). We can discuss this tomorrow.

adamdruppe:
btw betterC is a cancer. It needs to be cut out before it spreads to more and more things.

ibuclaw:
> Actually, I think it's better that we leave these functions as templates

Boo.  These are intrinsics, they emit no code in the library anyway.  Templatizing them just makes matching them more arduous with no real benefit.

PR #3163, opened 2020-07-18
Fix issue 21055 - core.stdc.stdarg is not nogc

dkorpel:
This is targeting master since there is a chance certain template instantiations break when forced `@nogc`. I hope the test suite covers stdarg well.


dkorpel:
It's green now

PR #3164, opened 2020-07-24
Fix issue 21070: -profile=gc makes the program much slower

tim-dlang:
Replacing the call to GC.stats with a new function, which
only returns the needed information, makes the program fast again.

### DIFF
--- a/changelog/faster_profilegc.dd
+++ b/changelog/faster_profilegc.dd
@@ -1,0 +1,6 @@
+`core.memory.GC.allocatedInCurrentThread()` was added, which makes `-profile=gc` faster
+
+The new function `core.memory.GC.allocatedInCurrentThread()` returns
+the same as `core.memory.GC.stats().allocatedInCurrentThread`, but
+is faster, because it avoids calculating other statistics.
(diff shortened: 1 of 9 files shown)


tim-dlang:
Unfortunately this change breaks the ABI of druntime.

n8sh:
>Unfortunately this change breaks the ABI of druntime.

Shouldn't it be fine if you move `gc_allocatedInCurrentThread` to be the last declared function in the class/interface?

tim-dlang:
> Shouldn't it be fine if you move `gc_allocatedInCurrentThread` to be the last declared function in the class/interface?

It would still break applications with a custom implementation of the GC interface. But it is probably a good idea for applications, which only access the interface.

UplinkCoder:
This has no tests.

tim-dlang:
> This has no tests.

A test for -profile=gc already exists in test/profile/src/profilegc.d, but I will add a new test, which tests allocatedInCurrentThread directly.

tim-dlang:
I have now moved the new function to the end and added a test and a changelog entry.

PR #3165, opened 2020-07-24
No need for dstrcmp in object.d

andralex:
Just dogfood `__cmp`.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -328,14 +328,11 @@ class TypeInfo
 
     override int opCmp(Object o)
     {
-        import core.internal.string : dstrcmp;
-
-        if (this is o)
-            return 0;
+        assert(this !is o); // __cmp takes care of that
         TypeInfo ti = cast(TypeInfo)o;
         if (ti is null)
             return 1;
-        return dstrcmp(this.toString(), ti.toString());
+        return __cmp(this.toString(), ti.toString());
     }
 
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -328,14 +328,11 @@ class TypeInfo
 
     override int opCmp(Object o)
     {
-        import core.internal.string : dstrcmp;
-
-        if (this is o)
-            return 0;
+        assert(this !is o); // __cmp takes care of that
```
n8sh, line 331: I do not understand the reasoning here.

andralex: See [this line](https://github.com/dlang/druntime/blob/master/src/object.d#L4029). The call to `opCmp` comes from the lowering of `==`, which is `__cmp`.

n8sh: The lowering of `==` is `opEquals`. In any case isn't it valid to call `Object.opCmp` directly?

andralex: Sorry, I meant `a < b`. But maybe I am wrong?

rainers: `__cmp` takes care of the comparison, but you are disallowing `a < a` here before it is called.

andralex: @rainers OK OK I see what happens here. I assumed that https://github.com/dlang/dmd/pull/9629 and https://github.com/dlang/druntime/pull/2562 had been pulled.

PR #3167, opened 2020-07-25
EASY: Move radix to template parameter in unsignedToTempString

andralex:
This is dangerously close to meh, but it's the right thing. The radix is always known statically.

### DIFF
--- a/src/core/internal/string.d
+++ b/src/core/internal/string.d
@@ -34,18 +34,18 @@ if (radix >= 2 && radix <= 16)
     size_t i = buf.length;
     do
     {
+        uint x = void;
         if (value < radix)
         {
-            ubyte x = cast(ubyte)value;
-            buf[--i] = cast(char)((x < 10) ? x + '0' : x - 10 + 'a');
-            break;
+            x = cast(uint)value;
+            value = 0;
         }
         else
         {
-            ubyte x = cast(ubyte)(value % radix);
-            value = value / radix;
-            buf[--i] = cast(char)((x < 10) ? x + '0' : x - 10 + 'a');
+            x = cast(uint)(value % radix);
+            value /= radix;
         }
+        buf[--i] = cast(char)((radix <= 10 || x < 10) ? x + '0' : x - 10 + 'a');
     } while (value);
     return buf[i .. $];
 }
--- a/src/core/internal/string.d
+++ b/src/core/internal/string.d
@@ -28,7 +28,7 @@ Params:
 Returns:
     The unsigned integer value as a string of characters
 */
-char[] unsignedToTempString()(ulong value, return scope char[] buf, uint radix = 10) @safe
+char[] unsignedToTempString(uint radix = 10)(ulong value, return scope char[] buf) @safe
 {
     if (radix < 2)
         // not a valid radix, just return an empty string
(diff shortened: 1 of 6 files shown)


andralex:
One of this and https://github.com/dlang/phobos/pull/7566/files must be force-merged because both are necessary to make the build work.

src/core/internal/string.d
```diff
@@ -28,7 +28,7 @@ Params:
 Returns:
     The unsigned integer value as a string of characters
 */
-char[] unsignedToTempString()(ulong value, return scope char[] buf, uint radix = 10) @safe
+char[] unsignedToTempString(uint radix = 10)(ulong value, return scope char[] buf) @safe
 {
     if (radix < 2)
```
schveiguy, line 33: You need static if/else here, or else it will complain about unreachable code (in the case where the radix is less than 2, which should never happen)

Alternatively, just statically disallow it. In fact, we should probably disallow greater than base 36 as well.

andralex: done

andralex:
> Is there an expectation of what this will help with? Is this supposed to increase performance?

Mostly a matter of principle. It will improve performance, too, because although ldc [does inline the call](https://godbolt.org/z/GKGcWq), [dmd](https://godbolt.org/z/9GxTsh) and [gdc](https://godbolt.org/z/5eWxjj) do not. Division by a variable is real slow, and with the constant all compilers get rid of it.

I confess I thought it would've been a smaller change. Looks like the primitive is used all over the place, so that bodes well for the efficiency gains. 

> Also, if you are going to make this a compile-time value, base 10 or less should avoid the `x < 10` check. Should be doable by changing the line (I can't comment on that line because it's not a change) to:
> 
> ```d
> buf[--i] = cast(char)((radix <= 10 || x < 10) ? x + '0' : x - 10 + 'a');
> ```
> 
> It's possible ldc can already see this, but I'm not sure about dmd.

Done.

src/core/internal/string.d
```diff
@@ -28,27 +28,24 @@ Params:
...
     do
     {
+        uint x = void;
         if (value < radix)
         {
-            ubyte x = cast(ubyte)value;
-            buf[--i] = cast(char)((x < 10) ? x + '0' : x - 10 + 'a');
-            break;
+            x = cast(uint)value;
+            value = 0;
```
schveiguy, line 41: Nitpicking, but is this better than the original that repeated the code, but didn't test value again?

In fact, the while loop could be `while(true)` if there is a break in here, as a value of 0 will never result from the other branch.

ibuclaw:
> > Is there an expectation of what this will help with? Is this supposed to increase performance?
> 
> Mostly a matter of principle. It will improve performance, too, because although ldc [does inline the call](https://godbolt.org/z/GKGcWq), [dmd](https://godbolt.org/z/9GxTsh) and [gdc](https://godbolt.org/z/5eWxjj) do not.

Indeed, normal functions are not inlined across modules. Making it a template will allow inlining to at least be attempted.

src/core/internal/string.d
andralex: Good point. Judging by the size of code generated, the finite loop is marginally better: https://godbolt.org/z/G6z3Kj. Add to that the mild inconvenience in source code duplication and it's okay to stick with the current version.

andralex:
I force merged the two PRs, let's see the build recover.

wilzbach:
> let's see the build recover.

That's not how things are done here :/

Geod24:
@andralex : I really wish we had the same set of rules for everyone. We wouldn't let any other contributor break the build when there's a trivial solution not to. You also ignored feedback on the Phobos PR.

Not to mention the wider problem that this is done with little justification. There's no measurement on the improvement, and how impactful it is.

andralex:
No need to get all agitated, I don't do this with any frequency and don't plan to. I think it's the second time I've ever done it.

andralex:
> You also ignored feedback on the Phobos PR.

What was that?

Geod24:
> What was that?

[Additionally to "don't break the build", this piece](https://github.com/dlang/phobos/pull/7566#issuecomment-663918007):
> Also please provide a proper commit title / message, something that can be understood without fetching additional context.

It's interesting that during beerconf we were discussing how good PRs come with an explanation of the what / why, not just a link or expecting people to know the context.

> No need to get all agitated, I don't do this with any frequency and don't plan to. I think it's the second time I've ever done it.

I'm not agitated, I'm concerned when someone with administrative access bypasses all checks in place and end up breaking the build for everyone. Because "Merge and see what happens" only works for your benefit: Anyone else has to wait for you to fix the build until they can get their PR in. It creates a single bottleneck for the entire core ecosystem.
And, with all due respect, I cannot see any other reason than laziness to go with a force-merge, as the correct alternative was straightforward and obvious.

andralex:
> > What was that?
> 
> [Additionally to "don't break the build", this piece](https://github.com/dlang/phobos/pull/7566#issuecomment-663918007):
> 
> > Also please provide a proper commit title / message, something that can be understood without fetching additional context.
> 
> It's interesting that during beerconf we were discussing how good PRs come with an explanation of the what / why, not just a link or expecting people to know the context.

OK, missed that.

> > No need to get all agitated, I don't do this with any frequency and don't plan to. I think it's the second time I've ever done it.
> 
> I'm not agitated, I'm concerned when someone with administrative access bypasses all checks in place and end up breaking the build for everyone. Because "Merge and see what happens" only works for your benefit: Anyone else has to wait for you to fix the build until they can get their PR in. It creates a single bottleneck for the entire core ecosystem.
> And, with all due respect, I cannot see any other reason than laziness to go with a force-merge, as the correct alternative was straightforward and obvious.

Heh, point taken although if I were lazy I'd probably sleep in instead of breaking builds in OSS on a Sunday morning. I'd have done this for someone else because it's more expeditious. Of course, the fact that I botched it kinda serves as a good example on why one shouldn't do it.

schveiguy:
I get the blowback, but there really isn't a good way to test for dependent PRs. Without that, it's kind of hard to make this work. In this case, there are workaround (not pleasant, I don't like the idea of merging a "temporary" PR), but there are possibly other cases where it's not possible.

With all the CIs we depend on, it is pretty unlikely we could fix this, but it is annoying.

andralex:
I'd put this in the "would be so nice to make druntime header-only and integrate with phobos" bin.

schveiguy:
OK, we need a redo of this now.

PR #3168, opened 2020-07-26
Revert "EASY: Move radix to template parameter in unsignedToTempString"

Geod24:
Reverts dlang/druntime#3167

It was force pushed while red and broke master. We should *never* do that.
See companion PR: https://github.com/dlang/phobos/pull/7567

### DIFF
--- a/src/core/demangle.d
+++ b/src/core/demangle.d
@@ -279,7 +279,7 @@ pure @safe:
 
         UnsignedStringBuf buf = void;
 
-        auto s = unsignedToTempString!16(val, buf);
+        auto s = unsignedToTempString(val, buf, 16);
         int slen = cast(int)s.length;
         if (slen < width)
         {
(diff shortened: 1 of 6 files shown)


Geod24:
Force-pushed the Phobos revert, this should go green now.

Geod24:
Azure fails at the checkout step:
```console
Generating script.
Script contents:
git clone --branch revert-3167-unsignedToTempString https://github.com/dlang/dmd.git ../dmd
========================== Starting Command Output ===========================
"C:\windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\a\_temp\1ea1e34b-99f0-426c-91f0-c9f3787a0065.cmd""
Cloning into '../dmd'...
fatal: Remote branch revert-3167-unsignedToTempString not found in upstream origin
##[error]Cmd.exe exited with code '128'.
```

Looks like a CI bug, where it gets confused because I clicked the "revert" button. @wilzbach do you know what might cause it ?

andralex:
Wow that's quite some aggravation. Sorry folks. Should I force push this, or does it make it just worse?

wilzbach:
There's some logic for most CIs to check-out matching branches like master or stable. It looks like the azure try to do the same, but fail.
Solutions:
- force-merge revert
- open revert from user account branch
- fix azure scripts to soft-fallback to master as default

andralex:
OK I force merged...

PR #3169, opened 2020-07-26
(un)signedToTempString: Make radix a template parameter to avoid inte…

andralex:
…gral division

The following test prints:

```
500 ms, 638 μs, and 1 hnsec
163 ms, 153 μs, and 7 hnsecs
```

when built with `dmd -O -release -inline` 

```D
char[] unsignedToTempString(uint radix = 10)(ulong value, return scope char[] buf) @safe
if (radix >= 2 && radix <= 16)
{
    size_t i = buf.length;
    do
    {
        uint x = void;
        if (value < radix)
        {
            x = cast(uint)value;
            value = 0;
        }
        else
        {
            x = cast(uint)(value % radix);
            value /= radix;
        }
        buf[--i] = cast(char)((radix <= 10 || x < 10) ? x + '0' : x - 10 + 'a');
    } while (value);
    return buf[i .. $];
}

char[] unsignedToTempString()(ulong value, return scope char[] buf, uint radix) @safe
{
    if (radix < 2)
        // not a valid radix, just return an empty string
        return buf[$ .. $];

    size_t i = buf.length;
    do
    {
        if (value < radix)
        {
            ubyte x = cast(ubyte)value;
            buf[--i] = cast(char)((x < 10) ? x + '0' : x - 10 + 'a');
            break;
        }
        else
        {
            ubyte x = cast(ubyte)(value % radix);
            value = value / radix;
            buf[--i] = cast(char)((x < 10) ? x + '0' : x - 10 + 'a');
        }
    } while (value);
    return buf[i .. $];
}

void main(string[] args) {
    import std.stdio, std.datetime.stopwatch;
    //import core.internal.string : unsignedToTempString;

    ulong x = args.length * 92837434;
    bool runtimeRadix() {
        char[20] buffer;
        return unsignedToTempString(x, buffer[], 10) == args[0];
    }
    bool compiletimeRadix() {
        char[20] buffer;
        return unsignedToTempString!10(x, buffer[]) == args[0];
    }

    auto r = benchmark!(runtimeRadix, compiletimeRadix)(10_000_000);
    auto f0Result = r[0];
    auto f1Result = r[1];
    writeln(f0Result, "\n", f1Result);
}
```

### DIFF
--- a/src/core/demangle.d
+++ b/src/core/demangle.d
@@ -279,7 +279,7 @@ pure @safe:
 
         UnsignedStringBuf buf = void;
 
-        auto s = unsignedToTempString(val, buf, 16);
+        auto s = unsignedToTempString!16(val, buf);
         int slen = cast(int)s.length;
         if (slen < width)
         {
--- a/src/core/internal/abort.d
+++ b/src/core/internal/abort.d
@@ -48,6 +48,6 @@ void abort(scope string msg, scope string filename = __FILE__, size_t line = __L
     UnsignedStringBuf strbuff = void;
 
     // write an appropriate message, then abort the program
-    writeStr("Aborting from ", filename, "(", line.unsignedToTempString(strbuff, 10), ") ", msg);
+    writeStr("Aborting from ", filename, "(", line.unsignedToTempString(strbuff), ") ", msg);
     c_abort();
 }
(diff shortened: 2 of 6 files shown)


andralex:
This is a redo of the ill-fated https://github.com/dlang/druntime/pull/3167.

schveiguy:
I can fix the io usage.

PR #3170, opened 2020-07-26
Remove unnecessary trusted lambdas

n8sh:
If everything a function does is wrapped in a trusted lambda instead mark the function as trusted.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -842,14 +842,14 @@ class TypeInfo_Vector : TypeInfo
 
 class TypeInfo_Function : TypeInfo
 {
-    override string toString() const
+    override string toString() const @trusted
     {
         import core.demangle : demangleType;
 
         alias SafeDemangleFunctionType = char[] function (const(char)[] buf, char[] dst = null) @safe nothrow pure;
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -842,14 +842,14 @@ class TypeInfo_Vector : TypeInfo
 
 class TypeInfo_Function : TypeInfo
 {
-    override string toString() const
+    override string toString() const @trusted
```
Geod24, line 845: Doesn't that change the API ? Meaning all overrides need to be `@safe` (or `@trusted`) ?

n8sh: It doesn't because `TypeInfo.toString` is already `@safe`.

thewilsonator (approved):

12345swordy:
This hasn't been merge for some reason.

Geod24 (approved):

thewilsonator:
@Geod24 There seems to be a lack of communication between the autotester and GH. This is all green but still not merging. It will likely have to be force merged.

Geod24:
Auto-merge toggled on

Geod24:
Well, it was worth a try.

PR #3172, opened 2020-07-27
WIP: Replace TypeInfo_Struct with templated class `RTTypeid`

andralex:
This is the druntime companion of https://github.com/dlang/dmd/pull/11459.

Open issues:

* `hasPointers` is not yet implemented, but it's easy
* `isDynamicType` is  not implemented and it points to some ABSOLUTELY HORRENDOUS code in associative arrays. Hair-raising code that fakes a `TypeInfo` without one. That would need to be fixed before this gets committed. Simple evolutionary step: add static key and value type information to all associative array primitives, e.g. replace calls such as `_aaGetY(paa, ti, valsz, pkey)` with `_aaGetY!(K, V)(paa, ti, valsz, pkey)`. We can then proceed with ignoring the dynamic type information and use the static type information instead. Then we drop the parameters entirely.
* I don't know how to implement `rtInfo`.
* No idea what `m_arg1` and `m_arg2` do.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -1431,6 +1431,132 @@ class TypeInfo_Struct : TypeInfo
     assert(!typeid(S).equals(&s, &s));
 }
 
+class RTTypeid(T) : TypeInfo
+if (is(T == struct))
+{
+    override string toString() const { return T.stringof; }
+
+    override bool opEquals(Object rhs)
+    {
+        if (this is rhs)
+            return true;
+        auto s = cast(const RTTypeid!T) rhs;
+        return s !is null;
+    }
+
+    override size_t getHash(scope const void* p) @trusted pure nothrow const
+    {
+        assert(p);
+        // Canonical prototype
+        alias HashFun = extern(D) size_t delegate() const nothrow @safe;
+        enum bool hasHash = is(typeof(&(cast(S*) null).toHash) == HashFun);
+        static if (hasHash)
+            return (cast(T*) p).toHash();
+        else
+            return hashOf(p[0 .. T.sizeof]);
+    }
+
+    override bool equals(in void* p1, in void* p2) @trusted pure nothrow const
+    {
+        immutable result = int(p1 is null) + int(p2 is null) - 1;
+        if (result >= 0)
+            return cast(bool) result;
+        return *(cast(T*) p1) == *(cast(T*) p2);
+    }
+
+    override int compare(in void* p1, in void* p2) @trusted pure nothrow const
+    {
+        if (p1 == p2)
+            return 0;
+        // Regard null references as always being "less than"
+        immutable result = int(p2 is null) - int(p1 is null);
+        if (result)
+            return result;
+        auto lhs = cast(T*) p1, rhs = cast(T*) p2;
+        static if (is(typeof(lhs.opCmp(*rhs)) == int))
+        {
+            return lhs.opCmp(*rhs);
+        }
+        else
+        {
+            // TODO: compare member by member. For now mimic TypeInfo_Struct.
+            import core.stdc.string : memcmp;
+            return memcmp(p1, p2, T.sizeof);
+        }
+    }
+
+    override @property size_t tsize() nothrow pure const
+    {
+        return T.sizeof;
+    }
+
+    override const(void)[] initializer() nothrow pure const @trusted
+    {
+        static immutable T data;
+        return (cast(void*) &data)[0 .. T.sizeof];
+    }
+
+    override @property uint flags() nothrow pure const { return m_flags; }
+
+    override @property size_t talign() nothrow pure const { return T.alignof; }
+
+    final override void destroy(void* p) const
+    {
+        .destroy(*(cast(T*) p));
+    }
+
+    override void postblit(void* p) const
+    {
+        static if (is(typeof((cast(T*) p).__postblit())))
+            (cast(T*) p).__postblit();
+    }
+
+    @safe pure nothrow
+    {
+        enum StructFlags : uint
+        {
+            hasPointers = 0x1,
+            isDynamicType = 0x2, // built at runtime, needs type info in xdtor
+        }
+        StructFlags m_flags;
+    }
+    union
+    {
+        void function(void*)                xdtor;
+        void function(void*, const TypeInfo_Struct ti) xdtorti;
+    }
+
+    override @property immutable(void)* rtInfo() nothrow pure const @safe { return m_RTInfo; }
+
+    version (WithArgTypes)
+    {
+        override int argTypes(out TypeInfo arg1, out TypeInfo arg2)
+        {
+            arg1 = m_arg1;
+            arg2 = m_arg2;
+            return 0;
+        }
+        TypeInfo m_arg1;
+        TypeInfo m_arg2;
+    }
+    immutable(void)* m_RTInfo;                // data for precise GC
+}
+
+unittest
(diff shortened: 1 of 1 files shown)


rainers:
> it points to some ABSOLUTELY HORRENDOUS code in associative arrays

That's propably mostly my code, implementing type info (and RTInfo) for the (key,value) pair, needed for running destructors and the precise GC.
At the time I also tried adding the generation of the respective TypeInfo to the compiler as part of the TypeInfo_AssocArray, but that failed. IIRC it was due to the delayed lowering in the glue layer which doesn't allow running semantics again. There are quite a few places where TypeInfo has to be generated pessimistically in advance because you need to predict what the glue layer will need (by calling `semanticTypeInfo`). That's not always working correctly and RTInfo generation still failed for some types last time I checked a couple of years ago. Maybe the situation has improved since then.

src/object.d
```diff
@@ -1431,6 +1431,132 @@ class TypeInfo_Struct : TypeInfo
     assert(!typeid(S).equals(&s, &s));
 }
 
+class RTTypeid(T) : TypeInfo
+if (is(T == struct))
```
rainers, line 1435: The current compiler implemtation in https://github.com/dlang/dmd/pull/11459 does not allow selectively replacing TypeInfo, it's either all or none.

src/object.d
```diff
@@ -1431,6 +1431,132 @@ class TypeInfo_Struct : TypeInfo
...
+        auto s = cast(const RTTypeid!T) rhs;
+        return s !is null;
+    }
+
+    override size_t getHash(scope const void* p) @trusted pure nothrow const
+    {
+        assert(p);
+        // Canonical prototype
+        alias HashFun = extern(D) size_t delegate() const nothrow @safe;
+        enum bool hasHash = is(typeof(&(cast(S*) null).toHash) == HashFun);
```
MoonlightSentinel, line 1452: `nothrow` and `@safe` are recommended but not enforced:
https://run.dlang.io/is/ZpYMqt
```suggestion
        alias HashFun = extern(D) size_t delegate() const;
        enum bool hasHash = is(typeof(&(cast(S*) null).toHash) : HashFun);
```

and add a try-catch if necessary below.

src/object.d
```diff
@@ -1431,6 +1431,132 @@ class TypeInfo_Struct : TypeInfo
...
+        static if (hasHash)
+            return (cast(T*) p).toHash();
```
MoonlightSentinel, line 1454: This needs tests with differently qualified `toHash` implementations.

src/object.d
```diff
@@ -1431,6 +1431,132 @@ class TypeInfo_Struct : TypeInfo
...
+    override int compare(in void* p1, in void* p2) @trusted pure nothrow const
+    {
+        if (p1 == p2)
+            return 0;
+        // Regard null references as always being "less than"
+        immutable result = int(p2 is null) - int(p1 is null);
+        if (result)
+            return result;
+        auto lhs = cast(T*) p1, rhs = cast(T*) p2;
+        static if (is(typeof(lhs.opCmp(*rhs)) == int))
```
MoonlightSentinel, line 1476: opCmp can return other types than `int`.
https://run.dlang.io/is/ZpYMqt

```suggestion
        static if (is(typeof(lhs.opCmp(*rhs)) : int))
```

src/object.d
```diff
@@ -1431,6 +1431,132 @@ class TypeInfo_Struct : TypeInfo
...
+
+    override bool opEquals(Object rhs)
+    {
+        if (this is rhs)
+            return true;
+        auto s = cast(const RTTypeid!T) rhs;
+        return s !is null;
+    }
+
+    override size_t getHash(scope const void* p) @trusted pure nothrow const
```
MoonlightSentinel, line 1447: TypeInfo doesn't require `pure`. This might fail below because toHash can be impure (allthough it obviously should be)

```suggestion
    override size_t getHash(scope const void* p) @trusted nothrow const
```

src/object.d
```diff
@@ -1431,6 +1431,132 @@ class TypeInfo_Struct : TypeInfo
...
+        // Canonical prototype
+        alias HashFun = extern(D) size_t delegate() const nothrow @safe;
+        enum bool hasHash = is(typeof(&(cast(S*) null).toHash) == HashFun);
+        static if (hasHash)
+            return (cast(T*) p).toHash();
+        else
+            return hashOf(p[0 .. T.sizeof]);
+    }
+
+    override bool equals(in void* p1, in void* p2) @trusted pure nothrow const
```
MoonlightSentinel, line 1459: These qualifiers are inconsistent with `TypeInfo` and might collide with less qualified `opEquals` implementations.

```suggestion
    override bool equals(in void* p1, in void* p2)
```

Ditto below

src/object.d
```diff
@@ -1431,6 +1431,132 @@ class TypeInfo_Struct : TypeInfo
...
+        .destroy(*(cast(T*) p));
+    }
+
+    override void postblit(void* p) const
+    {
+        static if (is(typeof((cast(T*) p).__postblit())))
+            (cast(T*) p).__postblit();
+    }
+
+    @safe pure nothrow
```
MoonlightSentinel, line 1514: THis has no effect?

andralex:
> > it points to some ABSOLUTELY HORRENDOUS code in associative arrays
> 
> That's propably mostly my code, implementing type info (and RTInfo) for the (key,value) pair, needed for running destructors and the precise GC.
> At the time I also tried adding the generation of the respective TypeInfo to the compiler as part of the TypeInfo_AssocArray, but that failed. IIRC it was due to the delayed lowering in the glue layer which doesn't allow running semantics again. There are quite a few places where TypeInfo has to be generated pessimistically in advance because you need to predict what the glue layer will need (by calling `semanticTypeInfo`). That's not always working correctly and RTInfo generation still failed for some types last time I checked a couple of years ago. Maybe the situation has improved since then.

I hasten to add I'm not attempting judgment on the quality of implementation, which is a glorious hack given the constraints (as far as I understand you're synthesizing layout and type information at runtime for a hashtable entry). Given that you don't have static type information to do that any other way, there are not many ways to wiggle. I assume if static type information about key and value are present, things would be radically simpler and there'd be no more need for that complication. Please advise.

adamdruppe:
I've been thinking about doing something like this for a long time. My old minimal object.d used to actually generate a bunch of the typeinfos from templates with the existing compiler too, mixing them in. For classes there's that static data block dmd puts out, will be very interesting when we can get around that.

src/object.d
WalterBright: That's right. I thought a good start would be to have RTTypeid duplicate what the compiler does, get everything working, and then start improving what it does. I.e. start by implementing instantiations of the TypeInfoXXXX classes.

rainers:
> I assume if static type information about key and value are present, things would be radically simpler and there'd be no more need for that complication. Please advise.

The TypeInfo for the separate key and value types is available, but what's needed is the TypeInfo for the combined `struct KeyVal { KEY key; VAL val; }`. This should be easy if the RTTypeid template actually works: just save `typeid(KeyVal)` somewhere in the TypeInfo for the AA.

rainers:
> * I don't know how to implement `rtInfo`.

`rtinfo = RTInfo!(T)`

andralex:
Thanks for the advice! I'll get back to this. Will close now to start with the simplest - `typeid(int)`.

PR #3173, opened 2020-07-27
unsignedToTempString no longer needs to always size the buffer for radix 2

n8sh:
Thanks to https://github.com/dlang/druntime/pull/3167 the buffer no longer has to be greatly oversized for the most common cases.

### DIFF
--- a/src/core/internal/string.d
+++ b/src/core/internal/string.d
@@ -50,10 +50,9 @@ if (radix >= 2 && radix <= 16)
     return buf[i .. $];
 }
 
-private struct TempStringNoAlloc()
+private struct TempStringNoAlloc(ubyte N)
 {
-    // need to handle 65 bytes for radix of 2 with negative sign.
-    private char[65] _buf = void;
+    private char[N] _buf = void;
     private ubyte _len;
     inout(char)[] get() inout return
     {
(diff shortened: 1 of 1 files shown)


src/core/internal/string.d
```diff
@@ -76,7 +75,10 @@ Returns:
 */
 auto unsignedToTempString(uint radix = 10)(ulong value) @safe
 {
-    TempStringNoAlloc!() result = void;
+    // Need a buffer of 65 bytes for radix of 2 with room for
+    // signedToTempString to possibly add a negative sign.
+    enum bufferSize = radix >= 10 ? 20 : 65;
+    TempStringNoAlloc!bufferSize result = void;
```
schveiguy, line 81: Hm... technically, we could exactly calculate the sizes. But this is a reasonable cutoff.

n8sh: Yeah a balancing consideration in my mind was having fewer distinct instantiations of `TempStringNoAlloc`, especially if (somehow) the return type of this function could end up being used to instantiate a cascade of other templates.

schveiguy: Possibly another level for octal would be useful, as it would be 1/4 the size of the 65-byte thing, and it's a possible common use case. But it's theoretical at this point (I'm 99% sure we only ever do decimal and hexadecimal), if it turns out someone wants it they can add it.

andralex: I recall the number of digits required to represent a number with `n` bits in base `k` is `1 + uint(n / log2(k))`. We have a `log2` thing in `allocator` that should work during compilation.

schveiguy: Realistically, I think the only bases people will likely use are 2, 8, 10, and 16. If you allocate 23 bytes (I think), you cover base 8 or higher. If you truly wanted to optimize, 16 only requires 17 bytes. So I think the diminishing returns probably aren't worth the extra cost of having different struct instantiations.

But 23 or 20 instead of 65 is definitely a lot more efficient.

PR #3174, opened 2020-07-28
Templated TypeInfo for int

andralex:
Starting small. This duplicates the implementation in https://github.com/dlang/druntime/blob/master/src/rt/typeinfo/ti_int.d. The idea is, of course, that the templated implementation will replace implementations for all scalar types. For testing purposes I restricted it to int.

The thing works if `typeid(int)` lowers to `RTTypeid(int)`. The use of eponymous templates elegantly hides the type name.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -424,6 +424,93 @@ class TypeInfo
     @property immutable(void)* rtInfo() nothrow pure const @safe @nogc { return rtinfoHasPointers; } // better safe than sorry
 }
 
+/*
+Run-time type information for scalar types (int for now).
+*/
+template RTTypeid(T)
+if (is(T == int))
+{
+    class Impl : TypeInfo
+    {
+        const: nothrow: pure: @safe:
+
+        override bool opEquals(const Object rhs) @nogc
+        {
+            return this is rhs
+                // Instance may be duplicated across DLLs, but has the same type.
+                || cast(const Impl) rhs !is null;
+        }
+
+        override int opCmp(const Object rhs)
+        {
+            if (this is rhs)
+                return 0;
+            if (auto ti = cast(const TypeInfo) rhs)
+                return __cmp(this.toString, ti.toString);
+            return 1;
+        }
+
+        override string toString() const pure nothrow @safe { return T.stringof; }
+
+        override size_t getHash(scope const void* p) @nogc @trusted
+        {
+            return *cast(const T *)p;
+        }
+
+        override bool equals(in void* p1, in void* p2) @nogc @trusted
+        {
+            return *cast(const T *)p1 == *cast(const T *)p2;
+        }
+
+        override int compare(in void* p1, in void* p2) @nogc @trusted
+        {
+            auto lhs = *cast(const T*) p1, rhs = *cast(const T*) p2;
+            return int(lhs > rhs) - int(lhs < rhs);
+        }
+
+        override @property size_t tsize() @nogc
+        {
+            return T.sizeof;
+        }
+
+        override const(void)[] initializer() @trusted @nogc
+        {
+            static immutable T data;
+            return (cast(const void *)&data)[0 .. T.sizeof];
+        }
+
+        override void swap(void *p1, void *p2) @nogc @trusted
+        {
+            T t = *cast(T *)p1;
+            *cast(T *)p1 = *cast(T *)p2;
+            *cast(T *)p2 = t;
+        }
+
+        override @property immutable(void)* rtInfo() { return rtinfoNoPointers; }
+    }
+
+    // On-demand singleton object in static storage
+    immutable RTTypeid = new Impl;
+}
+
+unittest
+{
+    alias id = RTTypeid!int;
+    static assert(id == id && id <= id && id >= id);
+    static assert(id.toString == "int");
+    int a = 42, b = 42, c = 43;
+    assert(id.getHash(&a) == 42);
+    assert(id.equals(&a, &b));
+    assert(!id.equals(&a, &c));
+    assert(id.compare(&a, &b) == 0);
+    assert(id.compare(&a, &c) == -1);
+    assert(id.compare(&c, &a) == 1);
+    static assert(id.tsize == 4);
+    assert(cast(ubyte[]) id.initializer() == [ 0, 0, 0, 0 ]);
+    id.swap(&a, &c);
+    assert(a == 43 && c == 42);
+}
+
 class TypeInfo_Enum : TypeInfo
 {
     override string toString() const { return name; }


andralex:
I took upon painting the bikeshed by changing the name from `RTTypeid` to `__typeid`. Reasons:

* A value with a capitalized name breaks the convention used everywhere in druntime and phobos
* The whole thing is awkward to type - was that `RTTypeID` or `RTTypeId` or...?
* The moment someone sees `__typeid` they'll figure a connection with `typeid`. With `RTTypeid`, they'll have to run to the manual.

wilzbach:
> __typeid

__ is reserved for compiler internals like auto-generated variables. For runtime template hooks we typically use `_d_`.

n8sh:
> > __typeid
> 
> __ is reserved for compiler internals like auto-generated variables. For runtime template hooks we typically use `_d_`.

Wouldn't this be okay for the same reason as `__cmp`, `__move_post_blt`, `__switch_error`, etc?

andralex:
Added statically-typed versions of all functions, so if people write `typeid(T).func` in generic code, an efficient version is picked up.

UplinkCoder:
Mind the 32bit platforms.
hash_t is 32bit there.

posix.mak
```diff
@@ -276,11 +276,10 @@ endif
 .PHONY : unittest
 ifeq (1,$(BUILD_WAS_SPECIFIED))
 unittest : $(UT_MODULES) $(addsuffix /.run,$(ADDITIONAL_TESTS))
-	@echo done
 else
 unittest : unittest-debug unittest-release
 unittest-%: target
-	$(MAKE) -f $(MAKEFILE) unittest OS=$(OS) MODEL=$(MODEL) DMD=$(DMD) BUILD=$*
+	$(MAKE) -s -f $(MAKEFILE) unittest OS=$(OS) MODEL=$(MODEL) DMD=$(DMD) BUILD=$*
```
wilzbach, line 282: What's the motivation behind silencing make?

andralex: Trying to stem the output noise. druntime has very noisy unittest output (including entirely useless things like the gdb version number) that has me routinely scroll back to find wherever errors were reported.

andralex:
Per discussions with @WalterBright, it turns out it's not reasonably feasible to do an incremental replacement. The plan instead is to develop `__typeid` separately from the built-in `typeid`, and replace uses of the latter with the former. At the very end we adjust `typeid` to lower to `__typeid`.

Currently no support for `struct`, `class`, and `interface`.

src/object.d
```diff
@@ -424,6 +424,547 @@ class TypeInfo
...
+            {
+                return __typeid!U.getHash(cast(const U*) p);
+            }
+            else
+            {
+                // Knuth's multiplicative hash with the golden ratio of 2^32 or 2^64
+                static if (size_t.sizeof == 4)
+                    return *p * 2_654_435_761U;
+                else
+                    return *p * 11_400_714_819_323_198_485UL;
```
n8sh, line 565: This would fail the `TypeInfo.getHash`/`object.hashOf` equivalence tests introduced in https://github.com/dlang/druntime/pull/2243.

src/object.d
```diff
@@ -424,6 +424,547 @@ class TypeInfo
...
+        // TypeInfo objects are all "greater" than all other types.
+        // TODO: devise a principled way to define/justify this.
+        return 1;
+    }
+
+    override string toString() @nogc @safe { return T.stringof; }
+
+    final static size_t getHash(scope const T* p) @nogc
+    {
+        pragma(inline, true);
```
n8sh, line 537: How about implementing this function as
```d
static if (__traits(isIntegral, T) && is(T : size_t))
    return *p;
else
    return hashOf(*p);
```

andralex: Was trying to do "the right thing", which is that a slight change in the integral should affect all bits in the output. So I went with the Knuth hash, which is a compromise between speed and quality. If compatibility is paramount, we can always simplify.

n8sh: My thinking (which may be wrong-headed) is that the primary purpose of `TypeInfo.getHash` is catering to built-in associative arrays. The `calcHash` function used by built-in AAs is already [using MurmurHash2's bit avalanche function](https://github.com/dlang/druntime/blob/104ac712331e4d3573fc277084334a528b5dadb1/src/rt/aaA.d#L446-L455) to transform the output of `TypeInfo.getHash`.

andralex: Cool. At any rate, here's hoping that AAs will be templated and rid of TypeInfo.

schveiguy: If we should do something different with integers, shouldn't `hashOf` be changed instead, and have this forward to it?

The point of `hashOf` (which is used by `TypeInfo`, but does not require `TypeInfo`) is to have a consistent mechanism to hash all types. I think we should use the existing mechanism.

Alternately, we can eliminate `hashOf`, and have it forward to `__typeid!T.getHash`.

I don't disagree with the change, as I have no specific thoughts on hashing, I always use what others have provided. But there shouldn't be multiple ways that druntime hashes things.

andralex: @schveiguy yes, mos def. A unified hashing scheme that all code uses is best. Will look into it when I get out of the woods wrt functionality.

andralex: @n8sh @schveiguy I reworked the code to use `getHash`. Any work on that can be done later.

src/object.d
```diff
@@ -424,6 +424,560 @@ class TypeInfo
     @property immutable(void)* rtInfo() nothrow pure const @safe @nogc { return rtinfoHasPointers; } // better safe than sorry
 }
 
+/*
+Generic implementation of TypeInfo. T must be unqualified.
```
AndrejMitrovic, line 428: The docs say T must be unqualified, so why is there a need to call `Unqual!T` below?

andralex: Comment got out of sync. Will fix, thanks.

andralex:
Blocked by:

* https://issues.dlang.org/show_bug.cgi?id=21141
* https://issues.dlang.org/show_bug.cgi?id=21142
* https://issues.dlang.org/show_bug.cgi?id=21143
* https://issues.dlang.org/show_bug.cgi?id=21144
* https://issues.dlang.org/show_bug.cgi?id=21145
* https://issues.dlang.org/show_bug.cgi?id=21146

src/rt/util/typeinfo.d
```diff
@@ -20,25 +21,21 @@ if (is(T == float) || is(T == double) || is(T == real))
 
     int compare(T d1, T d2)
     {
-        if (d1 != d1 || d2 != d2) // if either are NaN
+        auto d = d1 - d2;
+        auto result = int(d > 0) - int(d < 0);
```
WalterBright, line 25: The int casts are redundant.

src/rt/util/typeinfo.d
```diff
@@ -266,3 +263,183 @@ unittest
...
+                sa2 = {f2};
+
+            ti = __typeid!(S[]);
+            assert(ti.getHash(&da1) == ti.getHash(&da2));
+
+            assert(sa1  == sa2);
+            assert(sa1[] !is sa2[]);
+
+            ti = __typeid!(S[3]);
+            assert(ti.getHash(&sa1) == ti.getHash(&sa2));
```
WalterBright, line 442: nice coverage tests

src/rt/util/typeinfo.d
```diff
@@ -20,25 +21,21 @@ if (is(T == float) || is(T == double) || is(T == real))
...
-            {
-                if (d2 != d2)
-                    return 0;
-                return -1;
-            }
-            return 1;
+            return int(d1 == d1) - int(d2 == d2);
         }
-        return (d1 == d2) ? 0 : ((d1 < d2) ? -1 : 1);
+        return result;
```
WalterBright, line 30: Is this refactoring objectively better? If it isn't, I find it significantly more difficult to understand.

src/object.d
```diff
@@ -424,6 +433,807 @@ class TypeInfo
...
+    static assert(__typeid!S3.flags == 1);
+    static assert(__typeid!S4.flags == 1);
+
+    // Test a static array with types with destructor.
+    static uint x = 0;
+    static struct A { ~this() { ++x; } }
+    A[42] a;
+    __typeid!(A[42]).destroy(&a);
+    assert(x == 42);
+}
```
WalterBright, line 1235: any chance of more tests to eliminate the (non assert) uncovered lines of code?

src/rt/util/typeinfo.d
andralex: noted, thanks

src/rt/util/typeinfo.d
andralex: it's supposed to be faster

src/object.d
andralex: will try if this pr is to move forward

PR #3175, opened 2020-07-29
sys.posix.signal: SIGRT{MIN,MAX} is platform specific, not runtime

Geod24:
```
As explained in the comment, those functions are part of the
LSB (Linux Standard Base) core specs, hence why all runtimes
implement exactly the same interface.
This will also make those simples available on Musl.
```

Was trying to get eventcore working with `getaddrinfo_a` under Alpine when found this.
CC @Cogitri


Cogitri:
Thanks for working on this :)

PR #3176, opened 2020-07-30
add Object.initializer()

WalterBright:
Would like to reduce the role of `typeid` by replacing its capabilities with templates. Although this doesn't actually do it, it does lay the framework for doing this with the `typeid.initializer` functionality.

Using templates transitions code from runtime to compile time, and the TypeInfo structs produce a lot of exe file bloat.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -288,6 +288,24 @@ void setSameMutex(shared Object ownee, shared Object owner)
 }
 
 /**
+ * Gets default initializer for type T as an untyped array of data.
+ * Params:
+ *      T = type to get initializer for
+ * Returns:
+ *      If the type should be initialized to all
+ *      zeros, an array with a null ptr and a length equal to the type size will
+ *      be returned.
+ *      For static arrays, this returns the default initializer for
+ *      a single element of the array.
+ *      For class objects, the data returned is for an instance
+ *      of the class object.
+ */
+const(void)[] initializer(T)() nothrow pure @safe @nogc
+{
+    return typeid(T).initializer();
+}
+
+/**
  * Information about an interface.
  * When an object is accessed via an interface, an Interface* appears as the
  * first entry in its vtbl.
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -128,7 +128,7 @@ private GC initialize()
     if (!p)
         onOutOfMemoryErrorNoGC();
 
-    auto init = typeid(ConservativeGC).initializer();
+    auto init = initializer!ConservativeGC();
     assert(init.length == __traits(classInstanceSize, ConservativeGC));
     auto instance = cast(ConservativeGC) memcpy(p, init.ptr, init.length);
     instance.__ctor();
--- a/src/gc/impl/manual/gc.d
+++ b/src/gc/impl/manual/gc.d
@@ -49,7 +49,7 @@ private GC initialize()
     if (!p)
         onOutOfMemoryError();
 
-    auto init = typeid(ManualGC).initializer();
+    auto init = initializer!ManualGC();
     assert(init.length == __traits(classInstanceSize, ManualGC));
     auto instance = cast(ManualGC) memcpy(p, init.ptr, init.length);
     instance.__ctor();
--- a/src/rt/aaA.d
+++ b/src/rt/aaA.d
@@ -285,7 +285,7 @@ TypeInfo_Struct fakeEntryTI(ref Impl aa, const TypeInfo keyti, const TypeInfo va
     void* p = GC.malloc(sizeti + (2 + rtisize) * (void*).sizeof);
     import core.stdc.string : memcpy;
 
-    memcpy(p, typeid(TypeInfo_Struct).initializer().ptr, sizeti);
+    memcpy(p, initializer!(TypeInfo_Struct)().ptr, sizeti);
 
     auto ti = cast(TypeInfo_Struct) p;
     auto extra = cast(TypeInfo*)(p + sizeti);
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -410,7 +410,7 @@ extern(C) void _d_dso_registry(CompilerDSOData* data)
         if (firstDSO) initLocks();
 
         DSO* pdso = cast(DSO*).calloc(1, DSO.sizeof);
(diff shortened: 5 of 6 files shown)


rainers:
I'm fine with the idea in general (omitting TypeInfo when only the initializer is needed). Some thoughts:

Do we have to be careful with adding common names to object.d? This can easily conflict with another `initializer` symbol from another import.

Can/Should this be done by detecting the pattern in the compiler instead?

I think there are two major cases where `TypeInfo.initializer()` is used instead of `T.init`:

- the type is not known at compile time. The template will not help here.
- `T.init` is not an lvalue and cannot be copied from. Maybe this could be changed instead?

Geod24:
> Do we have to be careful with adding common names to object.d? This can easily conflict with another initializer symbol from another import.

I second this. The name is quite common and might lead to issues.

Regarding the goal: Couldn't we just const-fold `typeid` call when possible ? All this code seems to do is `emplace` on classes (which is a usage that seem to have been missed in this PR).
Or, if you really want to get rid of `typeid` (but I don't think we can?), just add `__traits(classInitializer)` which covers all the usages in this PR.

src/object.d
```diff
@@ -287,6 +287,24 @@ void setSameMutex(shared Object ownee, shared Object owner)
...
+ *      zeros, an array with a null ptr and a length equal to the type size will
+ *      be returned.
+ *      For static arrays, this returns the default initializer for
+ *      a single element of the array.
+ *      For class objects, the data returned is for an instance
+ *      of the class object.
+ */
+const(void)[] initializer(T)() nothrow pure @safe @nogc
+{
+    return typeid(T).initializer();
```
andralex, line 305: To actually get rid of `typeid`, do this:

```d
static immutable T empty;
return (cast(const void*) &empty)[0 .. T,sizeof];
```

schveiguy: This doesn't work for classes.

Also, typeid(T).initializer should return a null pointer, but with a specific length for zero-initialized items.

andralex: > This doesn't work for classes.

Oh ok... I wonder how that could be done without magic.

> Also, typeid(T).initializer should return a null pointer, but with a specific length for zero-initialized items.

I saw some code to that effect, but how is it ever used? Wouldn't it just segfault?

schveiguy: > Oh ok... I wonder how that could be done without magic.

I think the TypeInfo.initializer *is* the magic. I wonder if this template really shouldn't just be a __traits thing.



> I saw some code to that effect, but how is it ever used? Wouldn't it just segfault?

```D
auto init = ti.initializer;
if(init.ptr is null)
   memset(block, 0, init.length);
else
   memcpy(block, init.ptr, init.length);
```

You have to know this in order to use `TypeInfo.initializer`.

The idea is to avoid consuming static data space with a long array of zeros.

I'm looking through druntime usages of initializer, and I'm not seeing much (many compiler hooks have different versions of the functions that zero initialize than ones that use the initializer). But I know I've dealt with it in the past.

WalterBright: > I'm looking through druntime usages of initializer

They're part of this PR.

WalterBright:
> Do we have to be careful with adding common names to object.d?

This is why D doesn't have a global namespace, so one needn't worry about it. The compiler will complain if there is hijacking, and it can be resolved with qualification.

> Can/Should this be done by detecting the pattern in the compiler instead?

The idea is not just to eliminate TypeInfo as much as possible, but to push that complexity to druntime. This was done with great success with RTInfo - now gc implementers can try out new schemes without needing the compiler to change.

> `T.init`

`initializer()` has the same behavior as TypeInfo.initializer(), not T.init.

> Couldn't we just const-fold typeid call when possible ?

That is an interesting idea, but has the same issue as the next one:

> just add __traits(classInitializer) which covers all the usages in this PR.

That leaves the complexity in the compiler, whereas I'd like to move it to druntime. Moving things to druntime simplifies the complexity of the compiler, and allows changing things without needing a new compiler build. TypeInfo has a *lot* of magic compiler behavior associated with it.

Geod24:
> This is why D doesn't have a global namespace, so one needn't worry about it. The compiler will complain if there is hijacking, and it can be resolved with qualification.

Except that `object` is as close as it gets to a global namespace. Things added there will be automatically accessible from most user code.

> That leaves the complexity in the compiler, whereas I'd like to move it to druntime.

I think it would help a great deal to actually see what you plan to do. This PR, on its own, is a net negative: It shuffles things around and adds a new identifier in `object`. As a reviewer, just looking at this PR, I am not enticed to merge it. Having a separate PR, which shows the bigger picture, then small PRs to do the change incrementally greatly helps with that.

rainers:
> The compiler will complain if there is hijacking, and it can be resolved with qualification.

That still means it breaks the build.

> This was done with great success with RTInfo

I wouldn't call it a "great success":
- it's still buggy, e.g. https://issues.dlang.org/show_bug.cgi?id=10442
- It only works reasonably well since the RTInfo template boils down to just use `__traits(pointerBitmap)`. Before it could easily crash the compiler or produce errors (due to forward references IIRC)
- it had a notable impact on compilation speed

src/object.d
```diff
@@ -287,6 +287,24 @@ void setSameMutex(shared Object ownee, shared Object owner)
...
+/**
+ * Gets default initializer for type T as an untyped array of data.
+ * Params:
+ *      T = type to get initializer for
+ * Returns:
+ *      If the type should be initialized to all
+ *      zeros, an array with a null ptr and a length equal to the type size will
+ *      be returned.
+ *      For static arrays, this returns the default initializer for
+ *      a single element of the array.
```
kinke, line 299: I think this static-array special case is unintuitive and error-prone. I think it's better to disallow static array types via `if` template clause.

12345swordy:
What is the current blocker here?

src/object.d
JohanEngelen: > To actually get rid of `typeid`, do this:
> 
> ```d
> static immutable T empty;
> return (cast(const void*) &empty)[0 .. T,sizeof];
> ```

This will not work nicely with very large T: it will bloat the executable needlessly. For all work, keep a >10 MB struct type with non-zero initializer in mind ;-)

RazvanN7:
@WalterBright @andralex What is the status on this?

thewilsonator:
there's a trait for this now anyway.

PR #3177, opened 2020-08-01
Make _d_arrayctor and _d_arraysetctor aware of copy constructor

edi33416:
This PR makes the copy construction of arrays aware of the copy constructor.
IMHO, in the long run, this should be used in #3139 and other similar fixes.

### DIFF
--- a/src/core/internal/array/construction.d
+++ b/src/core/internal/array/construction.d
@@ -9,6 +9,35 @@
 */
 module core.internal.array.construction;
 
+package void cpCtorRecurse(S1, S2)(ref S1 to, ref S2 from)
+    if (is(S1 == struct) && is (immutable S1 == immutable S2))
+{
+    static if (__traits(hasCopyConstructor, S1))
+    {
+        to.__ctor(from);
+    }
+}
+
+package void cpCtorRecurse(T, size_t n, U)(ref T[n] to, ref U[n] from)
+{
+    import core.internal.destruction: destructRecurse;
+
+    static if (__traits(hasCopyConstructor, T))
+    {
+        size_t i;
+        scope(failure)
+        {
+            for (; i != 0; --i)
+            {
+                destructRecurse(to[i - 1]); // Don't care if it throws, as throwing in dtor can lead to UB
+            }
+        }
+
+        for (i = 0; i < to.length; ++i)
+            cpCtorRecurse(to[i], from[i]);
+    }
+}
+
 /**
  * Does array initialization (not assignment) from another array of the same element type.
  * Params:
@@ -54,7 +83,7 @@ if (is (immutable T == immutable U))
         {
             static if (__traits(hasCopyConstructor, T))
             {
-                to[i].__ctor(from[i]);
+                cpCtorRecurse(to[i], from[i]);
             }
             else
             {
@@ -137,6 +166,17 @@ if (is (immutable T == immutable U))
     _d_arrayctor(arr3[], arr2[]);
 
     assert(arr3 == [S(1), S(2), S(3), S(4)]);
+
+    S[2][2] arr4;
+    S[2][2] arr5 = [[S(0), S(1)], [S(2), S(3)]];
+    _d_arrayctor(arr4[], arr5[]);
+    assert(counter == 8);
+    assert(arr4 == arr5);
+
+    immutable S[2][2] arr6;
+    _d_arrayctor(arr6[], arr5[]);
+
+    assert(arr6 == [[S(1), S(2)], [S(3), S(4)]]);
 }
 
 @safe nothrow unittest
@@ -228,6 +268,20 @@ if (is (immutable T == immutable U))
     assert(didThrow);
     assert(counter == 2);
 
+    didThrow = false;
+    counter = 0;
+    try
+    {
+        Throw[2][2] a;
+        Throw[2][2] b = [[Throw(1), Throw(2)], [Throw(3), Throw(4)]];
+        _d_arrayctor(a[], b[]);
+    }
+    catch (Exception)
+    {
+        didThrow = true;
+    }
+    assert(didThrow);
+    assert(counter == 2);
 
     // Test that `nothrow` works
     didThrow = false;
@@ -284,7 +338,7 @@ if (is (immutable T == immutable U))
         {
             static if (__traits(hasCopyConstructor, T))
             {
-                p[walker].__ctor(value);
+                cpCtorRecurse(p[walker], value);
             }
             else
             {
@@ -359,6 +413,18 @@ if (is (immutable T == immutable U))
     immutable S[4] arr2;
     _d_arraysetctor(arr2[], s);
     assert(arr2 == [S(1235), S(1235), S(1235), S(1235)]);
+
+    counter = 0;
+    S[2] s2 = [S(1234), S(1234)];
+    S[2][2] arr3;
+
+    _d_arraysetctor(arr3[], s2);
+    assert(counter == arr.length);
+    assert(arr3 == [[S(1234), S(1234)], [S(1234), S(1234)]]);
+
+    immutable S[2][2] arr4;
+    _d_arraysetctor(arr4[], s2);
+    assert(arr4 == [[S(1235), S(1235)], [S(1235), S(1235)]]);
 }
 
 @safe nothrow unittest
@@ -379,8 +445,8 @@ if (is (immutable T == immutable U))
     try
     {
         Throw[4] a;
-        Throw[4] b = [Throw(1), Throw(2), Throw(3), Throw(4)];
-        _d_arrayctor(a[], b[]);
+        Throw b = Throw(1);
+        _d_arraysetctor(a[], b);
     }
     catch (Exception)
     {
@@ -440,8 +506,8 @@ if (is (immutable T == immutable U))
     try
     {
         Throw[4] a;
-        Throw[4] b = [Throw(1), Throw(2), Throw(3), Throw(4)];
-        _d_arrayctor(a[], b[]);
+        Throw b = Throw(1);
+        _d_arraysetctor(a[], b);
     }
     catch (Exception)
     {
--- a/src/core/internal/array/construction.d
+++ b/src/core/internal/array/construction.d
@@ -51,8 +51,17 @@ Tarr _d_arrayctor(Tarr : T[], T)(return scope Tarr to, scope Tarr from) @trusted
         size_t i;
         try
         {
-            for (i = 0; i < to.length; i++)
-                copyEmplace(from[i], to[i]);
+            static if (__traits(hasCopyConstructor, T))
+            {
+                to[i].__ctor(from[i]);
+            }
+            else
+            {
+                auto elem = cast(Unqual!T*)&to[i];
+                // Copy construction is defined as bit copy followed by postblit.
+                memcpy(elem, &from[i], element_size);
+                postblitRecurse(*elem);
+            }
         }
         catch (Exception o)
         {
@@ -94,15 +103,18 @@ Tarr _d_arrayctor(Tarr : T[], T)(return scope Tarr to, scope Tarr from) @trusted
     assert(arr1 == arr2);
 }
 
-// copy constructor
 @safe unittest
 {
+    // Test that copy constructor works
     int counter;
     struct S
     {
         int val;
-        this(int val) { this.val = val; }
-        this(const scope ref S rhs)
+        this(int v)
+        {
+            val = v;
+        }
+        this(ref typeof(this) rhs)
         {
             val = rhs.val;
             counter++;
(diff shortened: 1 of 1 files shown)


src/core/internal/array/construction.d
```diff
@@ -90,6 +98,32 @@ Tarr _d_arrayctor(Tarr : T[], T)(return scope Tarr to, scope Tarr from) @trusted
...
+        this(ref typeof(this) rhs)
+        {
+            val = rhs.val;
+            counter++;
+        }
+    }
+
+    S[4] arr1;
+    S[4] arr2 = [S(0), S(1), S(2), S(3)];
+    _d_arrayctor(arr1[], arr2[]);
```
edi33416, line 121: I was expecting that if I wrote
`S[4] arr1 = arr2;` this would call `_d_arrayctor`, but it seems it's not the case.

Can someone shed some light on this, please? What am I missing?

src/core/internal/array/construction.d
```diff
@@ -162,17 +196,25 @@ void _d_arraysetctor(Tarr : T[], T)(scope Tarr p, scope ref T value) @trusted
...
     {
         foreach (i; 0 .. p.length)
         {
-            auto elem = cast(Unqual!T*)&p[walker];
-            // Copy construction is defined as bit copy followed by postblit.
-            memcpy(elem, &value, element_size);
-            postblitRecurse(*elem);
+            static if (__traits(hasCopyConstructor, T))
+            {
+                emplace(&p[walker], value);
```
edi33416, line 209: I have an observation and a question regarding this `emplace`
If the `struct` in the `unittest` isn't static, so
```
int counter;
struct S { ... }
```
Then this will segfault without an error message.

If I replace the `emplace` line with
```
p[walker] = T(value);
```
the this will give the following compilation error
```
 Error: cannot access frame pointer of i20365.__unittest_L29_C7.S
```

What are your thoughts on this? Expected behaviour? Bug? Change `emplace` with `p[walker] = T(value);` ?

src/core/internal/array/construction.d
```diff
@@ -51,10 +52,17 @@ Tarr _d_arrayctor(Tarr : T[], T)(return scope Tarr to, scope Tarr from) @trusted
     {
         for (i = 0; i < to.length; i++)
         {
-            auto elem = cast(Unqual!T*)&to[i];
-            // Copy construction is defined as bit copy followed by postblit.
-            memcpy(elem, &from[i], element_size);
-            postblitRecurse(*elem);
+            static if (__traits(hasCopyConstructor, T))
+            {
+                emplace(&to[i], from[i]);
```
RazvanN7, line 57: Why use emplace and not just call the copy constructor?

src/core/internal/array/construction.d
RazvanN7: Yeah, the problem is the compiler does not rewrite array assignments that have structs with copy constructors as elements to _d_arrayctor. Take a look at this code [1][2]. It checks if the array has struct elements with a postblit, but not sign of copy ctor. This is trivial to solve by just adding a check for cp ctor, however, I think the course of action here should be: merge this first, then merge the compiler change, then make a PR to update the test in this one.

[1] https://github.com/dlang/dmd/blob/master/src/dmd/e2ir.d#L2939
[2] https://github.com/dlang/dmd/blob/master/src/dmd/e2ir.d#L3239

src/core/internal/array/construction.d
edi33416: > Why use emplace and not just call the copy constructor?

I wasn't sure if `to[i] = T(from[i])` would create the object in-place at `to[i]` or it would create a temporary and a copy

> Isn't this eliding the call to the cpctor or is emplace taking care of this under the hood?

What do you mean? The cpctor gets called. The unittest checks for this.

src/core/internal/array/construction.d
edi33416: I had already looked at [1] and [2] before, but they are not called.
For the following code
```
@safe unittest
{
    static int counter;
    static struct S
    {
        int val;
        this(int v)
        {
            val = v;
        }
        this(this)
        {
            counter++;
            writefln("Here with %s\n", counter);
        }
    }

    S[4] arr1;
    S[4] arr2 = [S(0), S(1), S(2), S(3)];
```

Using the line
```
    arr1 = arr2;
```
won't call either [1] or [2].

Changing the code to
```
arr1[] = arr2;
```
will call [1] but it won't call `_d_arrayctor`

Changing the code to
```
    S[4] arr2 = [S(0), S(1), S(2), S(3)];
    S[4] arr1 = arr2;
```
will call [2] but again won't call `_d_arrayctor`

I must say that I am puzzled.

Also, I don't think that [this code](https://github.com/dlang/dmd/blob/master/src/dmd/e2ir.d#L2954) could call the [templated](https://github.com/dlang/druntime/blob/master/src/core/internal/array/construction.d#L24) version of `_d_arrayctor`. I was expecting it to call the non template version found [src.rt.arrayassign](https://github.com/dlang/druntime/blob/master/src/rt/arrayassign.d#L170) but it doesn't call this either.

src/core/internal/array/construction.d
RazvanN7: Ok, I was not sure what emplace does.

src/core/internal/array/construction.d
RazvanN7: My point was that [1] and [2] should add checks for the copy constructor also otherwise the correct branches will not be taken in your unittest. I don'y actually understand what you mean by "it will call [1]". Do you mean the branch is taken?

edi33416: > My point was that [1] and [2] should add checks for the copy constructor also otherwise the correct branches will not be taken in your unittest.

Yes, I saw this. The `unittest` in the comment above uses postblit, not cpctor.

> I don'y actually understand what you mean by "it will call [1]". Do you mean the branch is taken?

Yes, I mean the branch is taken

src/core/internal/array/construction.d
edi33416: Changed to call `__ctor`. This should probably be faster than calling `emplace`

src/core/internal/array/construction.d
edi33416: Changing the code to `p[walker].__ctor(value)` fixed this

src/core/internal/array/construction.d
edi33416: This was actually a big messup on my part. I din't realize I had to recompile `Phobos` to get the `druntime` changes visible into my local `dmd` build; the confusion was also amplified by the fact that I was "proof-checking" that I am linking with the correct `druntime` by having some `printf`s in druntime: the catch is that I had them in templated functions which, you got it, will get instantiated at compile time in user code

edi33416: So, to sum it up, the course of action going forward is to hook the call to `_d_arrayctor` in dmd, as @RazvanN7 pointed out.

Before doing so, we'll have to convert `_d_arrayassign`, `_d_arrayassign_l` and `_d_arrayassign_r` to into templates and then change dmd to call the templated versions of `_d_arrayassign` and the current templated version of `_d_arrayctor` and their `set` cousins `_d_arraysetctor` and `_d_arraysetassign`

src/core/internal/array/construction.d
```diff
@@ -51,10 +52,17 @@ Tarr _d_arrayctor(Tarr : T[], T)(return scope Tarr to, scope Tarr from) @trusted
...
-            postblitRecurse(*elem);
+            static if (__traits(hasCopyConstructor, T))
+            {
+                to[i].__ctor(from[i]);
+            }
+            else
+            {
+                auto elem = cast(Unqual!T*)&to[i];
+                // Copy construction is defined as bit copy followed by postblit.
+                memcpy(elem, &from[i], element_size);
```
kinke, line 63: Not related to this PR, but memcpying each element separately is not optimal, and doesn't exploit the postblit advantage over copy ctors. The loop can additionally be completely elided if the type has no postblit.

kinke: Btw, is this stuff actually used at all? The compiler still lowers to the non-templated [extern(C) hook](https://github.com/dlang/druntime/blob/master/src/rt/arrayassign.d#L170)...

edi33416: You are right. This doesn't get lowered to by the compiler.
As you've pointed out, the compiler lowers to the array function from [rt.arrayassign](https://github.com/dlang/druntime/blob/master/src/rt/arrayassign.d#L170).

The course of action going forward is to hook the call to _d_arrayctor in dmd, as @RazvanN7 pointed out.

Before doing so, we'll have to convert _d_arrayassign, _d_arrayassign_l and _d_arrayassign_r to into templates and then change dmd to call the templated versions of _d_arrayassign and the current templated version of _d_arrayctor and their set cousins _d_arraysetctor and _d_arraysetassign

edi33416: > Not related to this PR, but memcpying each element separately is not optimal, and doesn't exploit the postblit advantage over copy ctors. The loop can additionally be completely elided if the type has no postblit.

I believe the memcpying is done separately so it will be able to correctly destroy previously constructed elements in case a of a throw from the postblit of the current `i`th element

kinke: When performing a single memcpy, the 'unconstructed' elements > i will have to be reset to `T.init` in the exception case, yes.

src/core/internal/array/construction.d
```diff
@@ -51,10 +52,17 @@ Tarr _d_arrayctor(Tarr : T[], T)(return scope Tarr to, scope Tarr from) @trusted
     {
         for (i = 0; i < to.length; i++)
         {
-            auto elem = cast(Unqual!T*)&to[i];
-            // Copy construction is defined as bit copy followed by postblit.
-            memcpy(elem, &from[i], element_size);
-            postblitRecurse(*elem);
+            static if (__traits(hasCopyConstructor, T))
+            {
+                to[i].__ctor(from[i]);
```
kinke, line 57: This won't work for a T = static array of structs with copy ctors (the trait returns true).

edi33416: Can you drop a unittest example, please?

kinke: Slightly modifying one of your existing ones:
```D
S[2][2] arr1;
S[2][2] arr2 = [[S(0), S(1)], [S(2), S(3)]];
_d_arrayctor(arr1[], arr2[]);
```

edi33416: So you're suggesting that the code should be changed to
```
try
    {
        static if (__traits(hasPostblit, T))
        {
            memcpy(to, from, element_size * to.length);
            for (i = 0; i < to.length; i++)
            {
                auto elem = cast(Unqual!T*)&to[i];
                postblitRecurse(*elem);
            }
        }
        else static if (__traits(hasCopyConstructor, T))
        {
            for (i = 0; i < to.length; i++)
            {
                to[i].__ctor(from[i]);
            }
        }
        else
        {
            memcpy(to, from, element_size * to.length);
        }
    catch (Exception o)
    {
        /* Destroy, in reverse order, what we've constructed so far
        */
        static if (__traits(hasPostblit, T) || !__traits(hasCopyConstructor, T))
        {
            for (int j = i; j < to.length; j++)
            {
                auto elem = cast(Unqual!T*)&to[i];
                memcpy(elem, &T.init, element_size);
            }
        }
        while (i--)
        {
            auto elem = cast(Unqual!T*)&to[i];
            destroy(*elem);
        }
        throw o;
    }
```

src/core/internal/array/construction.d
edi33416: So you're suggesting that the code should be changed to
```
try
    {
        static if (__traits(hasPostblit, T))
        {
            memcpy(to, from, element_size * to.length);
            for (i = 0; i < to.length; i++)
            {
                auto elem = cast(Unqual!T*)&to[i];
                postblitRecurse(*elem);
            }
        }
        else static if (__traits(hasCopyConstructor, T))
        {
            for (i = 0; i < to.length; i++)
            {
                to[i].__ctor(from[i]);
            }
        }
        else
        {
            memcpy(to, from, element_size * to.length);
        }
    catch (Exception o)
    {
        /* Destroy, in reverse order, what we've constructed so far
        */
        static if (__traits(hasPostblit, T) || !__traits(hasCopyConstructor, T))
        {
            for (int j = i; j < to.length; j++)
            {
                auto elem = cast(Unqual!T*)&to[i];
                memcpy(elem, &T.init, element_size);
            }
        }
        while (i--)
        {
            auto elem = cast(Unqual!T*)&to[i];
            destroy(*elem);
        }
        throw o;
    }
```

kinke: Something similar to this, yes. Elements without postblit/cpctor don't need a try/catch, that's a plain memcpy. `T.init` is an rvalue and cannot be used for memcpy; there's some `emplace` version blitting the default initializer which can be used - in case the `T.init` reset is really necessary (the array wasn't successfully constructed, so is it going to be destructed at some point later? have the elements already been initialized with T.init in the first place?).

src/core/internal/array/construction.d
edi33416: Thanks for catching this. Now it's fixed

src/core/internal/array/construction.d
edi33416: I'd keep this as a separate PR in an attempt to keep the changes small

RazvanN7:
@edi33416 Fix the conflicts and let's merge this!

edi33416:
Looks like this has been implemented in #3239

PR #3180, opened 2020-08-02
Reduce trait template usage in core.internal.hash core.internal.convert

n8sh:
Verified in an artificial benchmark that evaluating `__traits(isFloating, T) && is(T : real)` takes 85% or less the time and 66% or less the space of evaluating `is(Unqual!T == float) || is(Unqual!T == double) || is(Unqual!T == real))` as a template constraint even when `T` is always `float`.

### DIFF
--- a/src/core/internal/convert.d
+++ b/src/core/internal/convert.d
@@ -8,13 +8,12 @@
  * Source: $(DRUNTIMESRC core/internal/_convert.d)
  */
 module core.internal.convert;
-import core.internal.traits : Unqual;
 
 /+
 A @nogc function can allocate memory during CTFE.
 +/
 @nogc nothrow pure @trusted
-private ubyte[] ctfe_alloc()(size_t n)
+private ubyte[] ctfe_alloc(size_t n)
 {
     if (!__ctfe)
     {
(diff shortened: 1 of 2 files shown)


src/core/internal/convert.d
```diff
@@ -183,14 +181,15 @@ const(ubyte)[] toUbyte(T)(const ref T val) if (is(Unqual!T == float) || is(Unqua
 }
 
 @safe pure nothrow @nogc
-private Float parse(bool is_denormalized = false, T)(T x) if (is(Unqual!T == ifloat) || is(Unqual!T == idouble) || is(Unqual!T == ireal))
+private Float parse(bool is_denormalized = false, T:ireal)(T x)
```
MoonlightSentinel, line 184: Needs a `__traits(isFloating)` check to keep the current semantics.

src/core/internal/hash.d
```diff
@@ -235,10 +235,8 @@ if (!is(T == enum) && !is(T : typeof(null)) && is(T S: S[]) && !__traits(isStati
 }
 
 // Indicates if F is a built-in complex number type.
-private enum bool isComplex(F) = is(Unqual!F == cfloat) || is(Unqual!F == cdouble) || is(Unqual!F == creal);
-
 private F coalesceFloat(F)(const F val)
-if (__traits(isFloating, val) && !is(F == __vector) && !isComplex!F)
+if (__traits(isFloating, val) && !is(F == __vector) && !is(F : creal))
```
MoonlightSentinel, line 239: ditto?

src/core/internal/convert.d
n8sh: Doesn't need it because `parse` is private and only called for 80 bit real / quadruple precision real.

PR #3181, opened 2020-08-02
Reduce trait template use in core.time

n8sh:
Verified evaluating `is(immutable T == immutable Duration)` takes at most 85% the time and at most 77% the memory of `is(Unqual!T == Duration)`. Those figures include fixed compiler overhead since the point was not to precisely measure the difference but to verify that improvement was occurring.

### DIFF
--- a/src/core/time.d
+++ b/src/core/time.d
@@ -68,7 +68,6 @@ module core.time;
 import core.exception;
 import core.stdc.time;
 import core.stdc.stdio;
-import core.internal.traits : _Unqual = Unqual;
 import core.internal.string;
 
 version (Windows)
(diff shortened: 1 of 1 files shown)


thewilsonator:
please rebase to clear errors.

schveiguy:
This is a cool trick. Why aren't we just fixing Unqual?

```D
alias Unqual(T : immutable(U), U) = U;

alias Unqual(T) = Unqual!(immutable(T));
```

It passes all the phobos Unqual unit tests.

n8sh:
It's not exactly the same. I tried that at some point in the past but it didn't pass some test, I believe regarding dynamic arrays. Here's an example of how they can differ:
```d
import std;

void main()
{
    alias A = const shared(char)[];
    alias B = char[];
    writeln(is(Unqual!A == Unqual!B)); // false
    writeln(is(immutable A == immutable B)); // true
}
```
Here the difference doesn't matter but elsewhere someone might be depending on `Unqual` acting the way it does.

schveiguy:
Yeah, I just tried it was going to make a PR, but it fails for arrays.

schveiguy:
I found a workable solution @n8sh, at least druntime passed unittests. We'll see what happens. #3184

PR #2847, opened 2019-11-02
[PoC - do not merge] demangle 'I' as `in`

rainers:
as a companion to https://github.com/dlang/dmd/pull/10526

PR #3183, opened 2020-08-04
Add demangling support for 'in' storage class

Geod24:
Implemented in dlang/dmd#11474
Inspired by #2847


ibuclaw:
@Geod24 but this won't demangle `in ref` parameters?

Geod24:
Good catch! https://github.com/dlang/druntime/pull/3188

PR #3184, opened 2020-08-04
Simplify Unqual to take advantage of immutable absorbing all qualifiers,

schveiguy:
and of template parameters automatically converting to tail qualifiers
for arrays and pointers.

I'm hoping the simplification for such an oft-used template will help with compile times/memory.

I want to see if this passes unittests, it passes druntime unittests on my system.

Inspired by #3181 and friends.

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -35,14 +35,17 @@ template Unconst(T)
 }
 
 /// taken from std.traits.Unqual (which now imports this definition)
-alias Unqual(T : T*) = T*;
-alias Unqual(T : T[N], size_t N) = Unqual!T[N];
-alias Unqual(T : T[]) = T[];
-template Unqual(T) {
+alias Unqual(T : T*) = T*; // pointers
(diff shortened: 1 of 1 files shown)


src/core/internal/traits.d
```diff
@@ -34,29 +34,15 @@ template Unconst(T)
...
-    }
+/// taken from std.traits.Unqual (which now imports this definition)
+alias Unqual(T : T*) = T*;
+alias Unqual(T : T[N], size_t N) = Unqual!T[N];
+alias Unqual(T : T[]) = T[];
+template Unqual(T) {
+    static if (is(immutable T : immutable(U), U))
+        alias Unqual = U;
+    else
+        static assert(0);
```
schveiguy, line 45: I don't know if I have to do this, but there isn't a way to really use an is expression without the static if that I know of.

schveiguy:
Have to abandon this. With all the implicit casting that can happen, it's getting too complicated for its own good.

PR #3185, opened 2020-08-05
Fix issue 21116 - Do not squirrel stack pointers away in AssertErrors.

schveiguy:


schveiguy:
Hm... I'm not sure why the FreeBSD box core-dumps. but it seems consistent.

MoonlightSentinel:
> Hm... I'm not sure why the FreeBSD box core-dumps. but it seems consistent.

Sadly an unrelated regression

WalterWaldron:
> Hm... I'm not sure why the FreeBSD box core-dumps. but it seems consistent.

A long time ago I was investigating a 32 bit x86 dmd wrong code bug and Freebsd was the only one to detect the illegal access.
It is possible that Freebsd is more strict w.r.t. memory surrounding the stack.

src/core/internal/array/casting.d
```diff
@@ -25,41 +25,38 @@ Params:
...
-        , toType
-        , "`"
-    ];
-
-    // convert discontiguous `msgComponents` to contiguous string on the stack
+    // convert discontiguous `msgComponents` to contiguous string on the C heap
     enum msgLength = 2048;
-    char[msgLength] msg;
+    // note: never freed!
+    char* msg = cast(char *)pureMalloc(msgLength);
```
WalterBright, line 33: I don't see any reason to use pureMalloc(). Just use malloc().

MoonlightSentinel: Not disallowing array casting in pure code? This function is used here https://github.com/dlang/druntime/blob/7d962321244ae1cb03a6c9f25ff85ece83f28277/src/core/internal/array/casting.d#L82

schveiguy: Yeah, it wouldn't compile otherwise.

schveiguy:
@WalterBright with malloc I get:

```
src/core/internal/array/casting.d(80): Error: pure function core.internal.array.casting.__ArrayCast!(byte, int).__ArrayCast cannot call impure function core.internal.array.casting.onArrayCastError!().onArrayCastError
...
```

schveiguy:
Not sure if anything happened, but now passing freebsd. Probably an intermittent issue.

WalterBright:
Is it possible this broke:
```
/home/braddr/sandbox/at-client/pull-4136533-Linux_64_32/dmd/generated/linux/release/64/dmd -conf= -m32  -g -lowmem -i -Itools -version=NoMain -odgenerated -ofgenerated/d_do_test tools/d_do_test.d
Makefile:213: recipe for target 'generated/d_do_test' failed
make[1]: *** [generated/d_do_test] Segmentation fault (core dumped)
```
and:
```
D:\a\1\s\generated\windows\release\64\dmd.exe -conf= -m64 -Irunnable -lowmem -Jrunnable -L/OPT:NOICF -O -inline -odD:\a\1\s\test\test_results\runnable -ofD:\a\1\s\test\test_results\runnable\testptrref_gc_3.exe  runnable\testptrref_gc.d 
==============================
Test runnable\testptrref_gc.d failed: expected rc == 0, exited with rc == 5
```

?

schveiguy:
Hm... I don't see how? All I'm doing is calling malloc instead of using stack data. The block shouldn't be freed. The GC shouldn't even be used.

A core dump was happening for this PR as well, but intermittently. Isn't that what your DMD PR fixed?

schveiguy:
Master is green on the autotester. Where are these errors coming from?

WalterWaldron:
> > Hm... I'm not sure why the FreeBSD box core-dumps. but it seems consistent.
> 
> A long time ago I was investigating a 32 bit x86 dmd wrong code bug and Freebsd was the only one to detect the illegal access.
> It is possible that Freebsd is more strict w.r.t. memory surrounding the stack.

Here is the bug I was refering to: https://issues.dlang.org/show_bug.cgi?id=11435
It still hasn't been fixed AFAIK, it was causing intermittent failures on 32 bit freebsd many years ago.

WalterBright:
@schveiguy can you please temporarily revert this and see if that clears up the Azure Pipeline Windows64 failure that is blocking everything? I tried reverting my last pulled PR, but the problem remains: https://github.com/dlang/dmd/pull/11530

schveiguy:
@WalterBright OK, we can try that. #3190 

Let me know what the result is. If this fixes the issue, I can't see how it is related to the code. We still will have safety violations.

schveiguy:
BTW, do you have a link to where these failures are occurring?

PR #3190, opened 2020-08-08
Revert "Fix issue 21116 - Do not squirrel stack pointers away in Asse…

schveiguy:
…rtErrors."

This reverts commit 660d911bbd3342c1f1c1478d12e3e943c6038da0.

Temporarily reverting this, per @WalterBright 's request.

Note, this reintroduces safety issues, so please let me know if this *doesn't* fix the issue, and we can reapply it. Otherwise, we need to find a better way.

If nothing else, we should just change this not to create a custom message.

### DIFF
--- a/src/core/internal/array/casting.d
+++ b/src/core/internal/array/casting.d
@@ -25,38 +25,41 @@ Params:
 private void onArrayCastError()(string fromType, size_t fromSize, string toType, size_t toSize) @trusted
 {
     import core.internal.string : unsignedToTempString;
-    import core.memory : pureMalloc;
 
-    // convert discontiguous `msgComponents` to contiguous string on the C heap
+    const(char)[][9] msgComponents =
+    [
+        "An array of size "
+        , unsignedToTempString(fromSize)
+        , " does not align on an array of size "
+        , unsignedToTempString(toSize)
+        , ", so `"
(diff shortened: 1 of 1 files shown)


MoonlightSentinel:
I highly doubt that this is responsible for those failures but let's wait for the CI.

1. Segmentation fault in `d_do_test` built by the generated compiler:
Doesn't happen anymore after some backend changes were reverted and `d_do_test` should never hit this code anyway.

2. Segmentation fault in `dmd` when compiling `runnable/testptrref_gc.d` with `-O -inline`.
  Apparently caused by a bug in the host compiler (2.084.1) as other Win64 machines pass. See dlang/dmd#11531

schveiguy:
@MoonlightSentinel's PR was merged updating the compiler version on the azure pipelines. I feel like maybe we should give this some time to see if the problem is corrected, and if not, you can try merging this.

schveiguy:
The seg fault is in dmd, right? not in actually running d_do_test?

MoonlightSentinel:
Yes

schveiguy:
Should we try this? The FreeBSD failure is really impending progress.

schveiguy:
Actually, nevermind, it's still happening in the auto tester, even with this PR.

PR #3192, opened 2020-08-10
fix issue 10826 -- make sure arrays obey 32-byte or greater alignment

schveiguy:
Large arrays (arrays larger than 1/2 PAGESIZE) store the array metadata at the beginning instead of the end of the block. The reason is simple -- they can be extended in-place, and so having to move the metadata would be painful, and subject to possible races (in the case of shared arrays).

Because of the block "used" size stored, and the typeinfo for struct destructors, it needs to be 16 bytes. But some types might need 32-byte alignment.

This PR adds an "alignment padding" to all calculations for block size and array start. The alignment padding will be 0 for anything of alignment 16 or less, but will be non-zero for things more than 16-byte alignment.

I don't know if this will be properly tested by the CI -- on my system, dmd does not support 32-byte vectors (which is the only type I know of that requires 32-byte alignment). Ideas welcome on how to make sure this test is proper.

Note that I had to change the hard-coded 16-byte alignment for vectors to be the size of the vector.

For more details, see this forum discussion: https://forum.dlang.org/post/rgionugyekzpxuetyslh@forum.dlang.org

### DIFF
--- a/src/gc/impl/conservative/gc.d
+++ b/src/gc/impl/conservative/gc.d
@@ -1630,7 +1630,12 @@ struct Gcx
 
     void* smallAlloc(size_t size, ref size_t alloc_size, uint bits, const TypeInfo ti) nothrow
     {
-        immutable bin = binTable[size];
+        immutable trybin = binTable[size];
+        // ensure alignment of intended typeinfo can be satisfied by the bin size.
+        immutable bin = ti is null
+            ? trybin
+            : cast(immutable(byte))(binsize[trybin] % ti.talign == 0 ? trybin : trybin + 1);
+        assert(bin < B_NUMSMALL);
         alloc_size = binsize[bin];
 
         void* p = bucket[bin];
--- a/src/object.d
+++ b/src/object.d
@@ -830,7 +830,7 @@ class TypeInfo_Vector : TypeInfo
         return base.initializer();
     }
 
-    override @property size_t talign() nothrow pure const { return 16; }
+    override @property size_t talign() nothrow pure const { return base.tsize; }
 
     version (WithArgTypes) override int argTypes(out TypeInfo arg1, out TypeInfo arg2)
     {
--- a/src/rt/lifetime.d
+++ b/src/rt/lifetime.d
@@ -349,7 +349,7 @@ bool __setArrayAllocLength(ref BlkInfo info, size_t newlength, bool isshared, co
     }
     else
     {
-        if (newlength + LARGEPAD > info.size)
+        if (newlength + LARGEPAD + __alignPad(tinext) > info.size)
             // new size does not fit inside block
             return false;
         auto length = cast(size_t *)(info.base);
@@ -395,12 +395,45 @@ size_t __arrayAllocLength(ref BlkInfo info, const TypeInfo tinext) pure nothrow
     return *cast(size_t *)(info.base);
 }
 
+pragma(inline, true)
+size_t __alignPad(const TypeInfo tinext) pure nothrow @nogc @safe
+{
+    immutable a = tinext.talign;
+    return a > LARGEPREFIX ? a - LARGEPREFIX : 0;
+}
+
+@safe unittest {
+    static struct S {
+        long x;
+        double y;
+        ubyte[11] bytes;
+    }
+
+    static struct S2 {
+        ubyte[32] bytes;
+    }
+
+    static struct S3 {
+        align(64) int x;
+    }
+
+    assert(__alignPad(typeid(S)) == 0);
+    assert(__alignPad(typeid(S2)) == 0);
+    assert(__alignPad(typeid(S3)) == 48);
+    static if (is(__vector(ubyte[32]))) // only on some platforms
+    {
+        alias V = __vector(ubyte[32]);
+        assert(__alignPad(typeid(V)) == 16);
+    }
+}
+
+
 /**
   get the start of the array for the given block
   */
-void *__arrayStart(return BlkInfo info) nothrow pure
+void *__arrayStart(return BlkInfo info, const TypeInfo tinext) nothrow pure
 {
-    return info.base + ((info.size & BIGLENGTHMASK) ? LARGEPREFIX : 0);
+    return info.base + ((info.size & BIGLENGTHMASK) ? LARGEPREFIX + __alignPad(tinext) : 0);
 }
 
 /**
@@ -410,7 +443,9 @@ void *__arrayStart(return BlkInfo info) nothrow pure
   */
 size_t __arrayPad(size_t size, const TypeInfo tinext) nothrow pure @trusted
 {
-    return size > MAXMEDSIZE ? LARGEPAD : ((size > MAXSMALLSIZE ? MEDPAD : SMALLPAD) + structTypeInfoSize(tinext));
+    return size > MAXMEDSIZE
+        ? LARGEPAD + __alignPad(tinext)
+        : ((size > MAXSMALLSIZE ? MEDPAD : SMALLPAD) + structTypeInfoSize(tinext));
 }
 
 /**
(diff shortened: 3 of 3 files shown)


rainers:
> Ideas welcome on how to make sure this test is proper.

You should be able to test this unconditionally with an explicit alignment, e.g.
```
struct SpinLock
{
    align(64) int x;
}
```

rainers:
> Note that I had to change the hard-coded 16-byte alignment for vectors to be the size of the vector.

This is very platform specific. For example, aligned SSE accesses are faster than unaligned accesses. That's why 16 byte alignment is often guaranteed for `float[4]`. Probably the information needs to be provided by the compiler.

rainers:
The additional padding needs to be communicated to the GC aswell, it is used here: https://github.com/dlang/druntime/blob/f2d3c6c575a99cc0f0cf87ac101bc35248d052af/src/gc/impl/conservative/gc.d#L2132-L2136

rainers:
>All blocks in the GC that are more than 16 bytes are aligned by 32 bytes.

This is no longer correct as we have introduced non-power-of-two sizes to reduce the memory overhead:
https://github.com/dlang/druntime/blob/f2d3c6c575a99cc0f0cf87ac101bc35248d052af/src/gc/impl/conservative/gc.d#L1130-L1146

schveiguy:
> Probably the information needs to be provided by the compiler.

I welcome this change, but for now, this isn't *bad*, right? Maybe it wastes a bit of space (an extra 16 bytes in a page allocation, when it could have used that space to store data), but it's not an invalid result?

schveiguy:
> The additional padding needs to be communicated to the GC aswell

I don't know if this is necessary, the extra 16 (or whatever) bytes being scanned is probably not a huge deal. However, I think I will update the special function that zeroes the padding to make sure this is zero as well, to avoid false pointers.

schveiguy:
> This is no longer correct as we have introduced non-power-of-two sizes to reduce the memory overhead

Hm.. I didn't know that. But that is somewhat orthogonal to this change, as this change only deals with large blocks. However, it does mean while this change is not invalid, there is potentially more work to be done. We will probably have to add an API (if there isn't one already) to ask the GC for specific aligned blocks, as I don't see a good reason to allocate e.g. a 48-byte block just to throw 16 bytes away. And I assume you can't ask for a 32-byte aligned 48-byte block? If only non-aligned ones are on the free list, you would spend some time going through them.

I would like to separate that concern into another PR, as it's likely more complicated than this one.

src/object.d
```diff
@@ -819,7 +819,11 @@ class TypeInfo_Vector : TypeInfo
     override size_t getHash(scope const void* p) const { return base.getHash(p); }
     override bool equals(in void* p1, in void* p2) const { return base.equals(p1, p2); }
     override int compare(in void* p1, in void* p2) const { return base.compare(p1, p2); }
-    override @property size_t tsize() nothrow pure const { return base.tsize; }
+    override @property size_t tsize() nothrow pure const
+    {
+        immutable sz = base.tsize;
+        return sz < 16 ? 16 : sz;
```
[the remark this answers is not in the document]
schveiguy: OK. I wanted to only change the case for currently failing situations, but I can change this to just return `base.size`.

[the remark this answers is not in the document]
schveiguy: aaaaand, I just realized I changed the wrong function. I should have changed talign to this (which is what I thought I did). Oops.

schveiguy:
> We will probably have to add an API (if there isn't one already) to ask the GC for specific aligned blocks

qalloc (used here) already sends in the typeinfo, so we can handle this without adding API. I can probably do this with anew lookup table for the bin size, and use that if the talign is greater than 16.

schveiguy:
I have a simple fix for the GC to ensure aligned allocations for small arrays. I need to design a test.

schveiguy:
OK, @rainers can you check that I did the GC change correctly? I think this should be valid.

src/rt/lifetime.d
```diff
@@ -2873,3 +2913,18 @@ unittest
...
+{
+    import core.memory;
+    static struct S {
+        align(64) int x;
+    }
+
+    auto arr = new S[1];
+    auto info = GC.query(arr.ptr);
+    assert(cast(size_t)arr.ptr % 64 == 0);
+    assert(info.size % 64 == 0);
```
schveiguy, line 2929: FYI, when I did not turn off this test for `debug(SENTINEL)`, arr.ptr was not aligned to 64 bytes, and info.size was 65!

rainers: Yes, even 16 byte alignment is not guaranteed with the SENTINEL code IIRC.

rainers:
> I don't know if this is necessary, the extra 16 (or whatever) bytes being scanned is probably not a huge deal.

The precise GC needs the exact offset to scan the actual pointers.

src/gc/impl/conservative/gc.d
```diff
@@ -1630,7 +1630,12 @@ struct Gcx
 
     void* smallAlloc(size_t size, ref size_t alloc_size, uint bits, const TypeInfo ti) nothrow
     {
-        immutable bin = binTable[size];
+        immutable trybin = binTable[size];
+        // ensure alignment of intended typeinfo can be satisfied by the bin size.
+        immutable bin = ti is null
+            ? trybin
+            : cast(immutable(byte))(binsize[trybin] % ti.talign == 0 ? trybin : trybin + 1);
```
rainers, line 1637: Looks ok, but as this only works for power of 2 alignments anyway, using `& (ti.align - 1)` should be more slightly efficient.

rainers:
What bothers me a bit is that this adds (sometimes multiple) virtual function calls to performance sensitive code to satisfy a rare use case.
Maybe the size and alignment information should just be plain data elements of the TypeInfo instead of virtual function calls. The initiative to make the TypeInfo a template defined in druntime might help with this transition.

src/gc/impl/conservative/gc.d
schveiguy: Yes, true. All alignments must be positive power of 2 (statically enforced).

schveiguy:
> The precise GC needs the exact offset to scan the actual pointers.

I forgot about the precise GC. I'll add the code to deal with it.

> What bothers me a bit is that this adds (sometimes multiple) virtual function calls

Yeah, it's not ideal. Like you said, the changes to TypeInfo can help by making this a non-virtual member. I could also store something in the block metadata to indicate there is an additional offset, and only check the typeinfo on that. But I think that means a penalty for something else.

I can for now reduce the number of calls to talign to one. I think technically, we can deal with only using talign on [re]allocation. But maybe in a future PR. The fact that this is passing, I don't want to rip it apart again.

schveiguy:
> The precise GC needs the exact offset to scan the actual pointers.

Hm... can someone answer this question? I can't rely on the typeinfo being in the block in the case of arrays of non-structs or non-classes. However, I don't know if there's anything OTHER THAN structs and classes which are both aligned to >16 bytes and have pointers. Is this something I can depend on?

My other option here is to change rtInfo so it contains the alignment information...

ibuclaw:
> > The precise GC needs the exact offset to scan the actual pointers.
> 
> Hm... can someone answer this question? I can't rely on the typeinfo being in the block in the case of arrays of non-structs or non-classes. However, I don't know if there's anything OTHER THAN structs and classes which are both aligned to >16 bytes and have pointers. Is this something I can depend on?
> 

Is the question, can __vectors appear anywhere that the GC won't have any information about?

schveiguy:
No, my question relates to the precise GC. So are there any types that AREN'T structs (I realized arrays of classes are already covered, they are all pointers), which would have an alignment greater than 16 (and so need alignment padding), and contain pointers.

And actually, I just realized, a struct's typeinfo is only put into the metadata if it has a dtor. Crap. So my first idea doesn't work. I'll have to use rtInfo...

schveiguy:
I'm going to close this for now. I'm having to rethink this whole thing, and might be better to start over. But right now, I don't have the time to put into this.

Final thoughts:

1. We can put the typeinfo into the block when it needs alignment information for the precise GC
2. The alignment padding should be included in the array content when storing the used length. This makes things simple and no-change for most of the code, you just have to deal with the alignment padding on allocation/reallocation.

ibuclaw:
> No, my question relates to the precise GC. So are there any types that AREN'T structs (I realized arrays of classes are already covered, they are all pointers), which would have an alignment greater than 16 (and so need alignment padding), and contain pointers.
> 

One thing does come to mind are closures. They are allocated only with a size (i.e `_d_allocmemory(16)`), no typeinfo at all.

schveiguy:
@ibuclaw it's only relevant for large array allocations where we are offsetting the actual data to put the metadata at the front, and small allocations that are not power-of-2 aligned.

rainers:
Some ideas:
- add alignment information to the padding block: there is enough space with 32-bit processes, and a couple of bits could be taken from the length for 64-bit processes.
- or if typeinfo is stored in the padding, check talign, otherwise the field could be reused to save the alignment. The two types of data could be distinguished by bit 0 (never set for typeinfo pointer), or just by assuming small values being alignment.
- or add a bit for each page of large object pools that indicates whether alignment is larger than 16. In that case there is enough room after the first 16 bytes.

PR #3196, opened 2020-08-16
remove deprecated usage of a class alocator in gc benchs

NilsLankila:
that blocked https://github.com/dlang/dmd/pull/11581.

### DIFF
--- a/benchmark/gcbench/vdparser.extra/vdc/ast/node.d
+++ b/benchmark/gcbench/vdparser.extra/vdc/ast/node.d
@@ -182,15 +182,6 @@ class Node
 
     version(COUNT) static __gshared int countNodes;
 
-    version(NODE_ALLOC)
-    new(size_t sz)
-    {
-        assert(sz < NodeAllocData.kSize / 2);
-            //return gc_malloc(sz, 1); // BlkAttr.FINALIZE
-        void* p = NodeAllocData.alloc(sz);
-        return p;
-    }
-
     this()
     {
         version(COUNT) InterlockedIncrement(&countNodes);


MoonlightSentinel:
This PR should not have been merged without an appropriate replacement for `version(NODE_ALLOC)` ...

NilsLankila:
the versioned block are not used in the test, what happened rather is that the parser still parsed them, even if not used.

MoonlightSentinel:
This breaks the configuration even if not used by the CI. 
Removing that configuration should be a conscious decision applied in standalone PR instead of "Well, its broken now..."

wilzbach:
> This breaks the configuration even if not used by the CI. 
> Removing that configuration should be a conscious decision applied in standalone PR instead of _"Well, its broken now..."_

No, this is fine.
The code in question has no functional purpose anymore except being a project that allows to benchmark GC performance and isn't used in any other configuration than the default one.

MoonlightSentinel:
Sorry for being pedantic here, but whether it's actually used by the CI is irrelevant here because it could be used by other tools.

If it's fine to remove this functionality then everything else included in `version(NODE_ALLOC)` should be removed as well instead of remaining as unusable code.

wilzbach:
> Sorry for being pedantic, but whether it's actually used by the CI is irrelevant because it could be used by external tools.

No it's not a tool. I think this is where the confusion is coming from. It's a (old) testsuite. We don't provide guarantees for a test in `dmd/test` either.

wilzbach:
> If it's fine to remove this functionality then everything else included in version(NODE_ALLOC) should be removed as well instead of remaining as unusable code.

There's a lot of dead code in this GC benchmark and I'm not even sure whether anyone has even used it in the last three years. We can't put general code cleanup burdens on such PRs as people will just give up then. It's already very hard to push deprecations along.

MoonlightSentinel:
> No it's not a tool. I think this is where the confusion is coming from. It's a (old) testsuite. We don't provide guarantees for a test in `dmd/test` either.

I see, then it's fine to remove this functionality.

> There's a lot of dead code in this GC benchmark and I'm not even sure whether anyone has even used it in the last three years. We can't put general code cleanup burdens on such PRs as people will just give up then. It's already very hard to push deprecations along.

Don't disagree with this in general, but leaving valid code in a broken state is bad irregardless. But as this PR was merged it's time to do cleanup in another PR.

rainers:
This is mostly a benchmark for the precise GC, inherited from visuald where it is still part of the repository, but has long been replaced by other implementations. I don't mind the disabled code being removed, but as @MoonlightSentinel mentioned it is probably better to remove other parts referring to the same version, too.

I'm not a big fan of deprecating code in the parser, though, because you cannot keep the code with conditionals based on the compiler version (only workaround is ugly mixins).

PR #3202, opened 2020-08-30
Investigate the autotester-win32 timeout

MoonlightSentinel:


rainers:
Using an optlink without the message box might be useful: https://github.com/DigitalMars/optlink/pull/34

I can make a prebuilt exe available somewhere if it can be incorporated into the build somehow.

MoonlightSentinel:
That should be possible after the Azure migration is done

MoonlightSentinel:
Not ready for rview, just checking if the the auto-tester will pick this up as a non-draft pr

PR #3208, opened 2020-09-07
Simplify typeinfo.d to take advantage of templated implementation

andralex:
Switching to templates allow us to eliminate `Floating` and `Array` entirely.


Geod24:
> As usual, please give a proper title to both PR and commits.
> Remember the [Contributing guidelines](https://github.com/dlang/dmd/blob/master/CONTRIBUTING.md#solving-bugs--submitting-pull-requests) include an article about [good commit messages](https://chris.beams.io/posts/git-commit/).

I should write a bot for this.

wilzbach:
> > As usual, please give a proper title to both PR and commits.
> > Remember the [Contributing guidelines](https://github.com/dlang/dmd/blob/master/CONTRIBUTING.md#solving-bugs--submitting-pull-requests) include an article about [good commit messages](https://chris.beams.io/posts/git-commit/).
> 
> I should write a bot for this.

Maybe we should add it to the dlang-bot's message?

Geod24:
> Maybe we should add it to the dlang-bot's message?

Experience shows that everyone just ignores it. Including me. I only look at it to see if the issue will properly auto-close (and since I know the syntax, I rarely do that). Note that an alternative for this would be to use PR templates.
I think it would only really work if the bot could give you real feedback on your PR instead of blindly pasting a message. But that's not easily done with commit messages.

andralex:
Changed the commit msg and the title.

andralex:
It bothers me that `TypeInfoXxx.compare` is just slightly different for floating point numbers than the built-in comparison (enshrined by `__cmp`). I wonder if we should just make all behave like `__cmp`. (I tried it, no unittests failed.)

andralex:
Added to https://issues.dlang.org/show_bug.cgi?id=13499

src/rt/util/typeinfo.d
```diff
@@ -1,107 +1,64 @@
...
-    }
-
-    public alias hashOf = core.internal.hash.hashOf;
+// Three-way compare for real fp types. NaN is smaller than all valid numbers.
+// Code is small and fast, see https://godbolt.org/z/fzb877
+pragma(inline, true)
+private int cmp3(T)(const T d1, const T d2)
+if (is(T == float) || is(T == double) || is(T == real))
+{
+    if (d2 != d2)
```
andralex, line 29: This codecov warning seems spurious - the unittest below exercise all paths in this function. cc @WalterBright

src/core/internal/array/comparison.d
```diff
@@ -60,26 +60,24 @@ int __cmp(T)(scope const T[] lhs, scope const T[] rhs) @trusted
         immutable len = lhs.length <= rhs.length ? lhs.length : rhs.length;
         foreach (const u; 0 .. len)
         {
-            static if (__traits(isFloating, T))
-            {
-                immutable a = lhs.ptr[u], b = rhs.ptr[u];
-                static if (is(T == cfloat) || is(T == cdouble)
+            auto a = lhs.ptr[u], b = rhs.ptr[u];
+            static if (__traits(isFloating, T) && is(T == cfloat) || is(T == cdouble)
```
n8sh, line 64: I suggest replacing the entire condition with the more succinct
```d
static if (is(T : creal))
````
 I have verified that for scalar `T` this has no performance penalty relative to
```d
static if (__traits(isFloating, T) && (is(T == cfloat) || is(T == cdouble) || is(T == creal)))
```
(in fact it is ever-so-slightly faster and ever-so-slightly more memory efficient).

andralex: OK

thewilsonator (approved):

thewilsonator:
Pease rebase

RazvanN7:
@andralex @thewilsonator I have rebased this and updated it to not use builtin complex numbers (which have been deprecated). Is this good to go?

thewilsonator:
Yes although this PR says This branch cannot be rebased due to conflicts.

RazvanN7 (approved):

andralex:
@RazvanN7 thanks for taking this over the finish line!

PR #3210, opened 2020-09-11
make cimag() and creal() inlineable

WalterBright:
For performance, and for some platforms (such as win32) they are header-only.

### DIFF
--- a/src/core/stdc/complex.d
+++ b/src/core/stdc/complex.d
@@ -149,14 +149,14 @@ creal   csqrtl(creal z);
  ///
  real   cargl(creal z);
 
- ///
- double cimag(cdouble z);
- ///
- float  cimagf(cfloat z);
- ///
- real   cimagl(creal z);
+///
+pragma(inline, true) double cimag(cdouble z) { return z.im; }
+///
+pragma(inline, true) float  cimagf(cfloat z) { return z.im; }
+///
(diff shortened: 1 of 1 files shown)


WalterBright:
Blocking https://github.com/dlang/dmd/pull/11710

thewilsonator (approved):

src/core/stdc/complex.d
```diff
@@ -149,14 +149,14 @@ creal   csqrtl(creal z);
...
  real   cargl(creal z);
 
- ///
- double cimag(cdouble z);
- ///
- float  cimagf(cfloat z);
- ///
- real   cimagl(creal z);
+///
+pragma(inline, true) double cimag(cdouble z) { return (cast(double*)&z)[1]; }
```
kinke, line 153: Why not `return z.im;`?

WalterBright: Because `z.im` would be an `idouble`, and casting that to `double` would result in `0`.

rainers: ?? typeof(z.im) == double!

rainers (changes requested):
Please use `z.im` instead.

wilzbach:
WHY WAS THIS MERGED???

It obviously led to a failure in mir-algorithm.

PR #3212, opened 2020-09-13
Document `rt.deh`, plus some cleanups

Geod24:
I've been looking into the exception-handling code on and off for a few months now, trying to make sense of all of it and see how to improve it. My original goal was to fix the code for OSX, but now I'm just yak shaving.

One of the great mystery was how stack frame are stored and generated. We have some [ugly hacks](https://github.com/dlang/druntime/blob/8baae6cb88813c0fc27e2f1ab5015206e64b3063/src/core/runtime.d#L794-L809) which I hope to get rid off, eventually. I'd also like to be able to provide a better user interface for the backtrace code so a user can cheaply store/display backtrace.

Anyway here's my findings. I don't think the fact that we have a GC-allocation in the middle of EH was expected / wanted, but it's the current state.

### DIFF
--- a/src/rt/deh.d
+++ b/src/rt/deh.d
@@ -1,14 +1,38 @@
 /**
- * Implementation of exception handling support routines.
+ * Entry point for exception handling support routines.
  *
- * Copyright: Copyright Digital Mars 1999 - 2013.
+ * There are three style of exception handling being supported by DMD:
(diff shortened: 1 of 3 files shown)


thewilsonator (approved):

Geod24:
Bypassing broken Buildkite: https://github.com/CyberShadow/ae/issues/46

PR #3216, opened 2020-09-25
[stable] core.internal.array.casting: Prevent superfluous slice copy checks

kinke:
Introduced by #3185. The compiler checks for matching lengths and no overlap (if bounds checks are enabled); for LDC, this is always delegated to a druntime function (DMD emits inline code in at least some circumstances), which means linking error regressions with `-betterC`.

### DIFF
--- a/src/core/internal/array/casting.d
+++ b/src/core/internal/array/casting.d
@@ -35,10 +35,13 @@ private void onArrayCastError()(string fromType, size_t fromSize, string toType,
     size_t index = 0;
     void add(const(char)[] m)
     {
+        import core.stdc.string : memcpy;
+
         auto N = msgLength - 1 - index;
         if (N > m.length)
             N = m.length;
-        msg[index .. index + N] = m[0 .. N];
+        // prevent superfluous and betterC-unfriendly checks via direct memcpy
+        memcpy(msg + index, m.ptr, N);
         index += N;
     }
 


src/core/internal/array/casting.d
```diff
@@ -35,10 +35,12 @@ private void onArrayCastError()(string fromType, size_t fromSize, string toType,
     size_t index = 0;
     void add(const(char)[] m)
     {
+        import core.stdc.string : memcpy;
+
         auto N = msgLength - 1 - index;
         if (N > m.length)
             N = m.length;
-        msg[index .. index + N] = m[0 .. N];
+        memcpy(msg + index, m.ptr, N);
```
thewilsonator, line 43: Please put a comment noting this so it doesn't get reverted.

thewilsonator (approved):
Otherwise feel free to merge.

src/core/internal/array/casting.d
kinke: Done.

thewilsonator:
Thanks.

MoonlightSentinel (changes requested):
Add a @betterC test

thewilsonator:
```
Cloning into '.generated/druntime-2.094.0'...
warning: Could not find remote branch v2.094.0 to clone.
fatal: Remote branch v2.094.0 not found in upstream origin
posix.mak:574: recipe for target '.generated/druntime-2.094.0' failed
```

Hmm this looks like this is required. Might be better to target master.

kinke:
> Add a `@betterC` test

The existing druntime betterC integration test showed this regression for LDC. DMD emits inline code in this case.

MoonlightSentinel:
> The existing druntime betterC integration test showed this regression for LDC. DMD emits inline code in this case.

I see, nevermind then.

kinke:
> Hmm this looks like this is required. Might be better to target master.

It's already in the corresponding LDC branch, so I don't really care. But stable's CI is broken for days and always will for a (usually brief) period where DMD's `VERSION` file is set to an inexistent tag, so this could use a proper fix.

wilzbach:
> But stable's CI is broken for days and always will for a (usually brief) period where DMD's VERSION file is set to an inexistent tag, so this could use a proper fix.

Yeah, this should be fixed...
I merged your PR for now ;-)

PR #3219, opened 2020-09-27
[stable] core.thread.threadbase: Fix memory leak regression

kinke:
Caught by an LDC address-sanitizer test. Introduced by abec054a065e82c3fc47e9ebabac9f5e3952b55d.

### DIFF
--- a/src/core/thread/threadbase.d
+++ b/src/core/thread/threadbase.d
@@ -416,10 +416,6 @@ class ThreadBase
         m_curr = &m_main;
     }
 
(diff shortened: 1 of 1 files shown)


thewilsonator:
Is it possible to add a test case for this?

kinke:
It's certainly possible, but I'm not going to write one. This threading code is ugly as hell and the refactoring wasn't for the better IMO (I think a follow-up was promised, but as many times, not realized).

thewilsonator (approved):

PR #3221, opened 2020-09-30
Add UDA for optional Objective-C interface methods

jacob-carlborg:
This is the druntime part of adding support for Objective-C protocols.

### DIFF
--- a/src/core/attribute.d
+++ b/src/core/attribute.d
@@ -16,7 +16,10 @@
 module core.attribute;
 
 version (D_ObjectiveC)
+{
+    version = UdaOptional;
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

PetarKirov (approved):
Looks clean and pretty well documented. Nice work!

PetarKirov:
I'm merging, as otherwise the dmd PR can't be properly tested. This public symbol is already versioned under `D_ObjectiveC` and the probably causing any harm is essentially zero.

jacob-carlborg:
Thanks.

PR #3222, opened 2020-10-01
Qualify some variables and params as constant

nordlow:


Geod24:
Why ?

nordlow:
Because they are not modified and they trigger warnings when using my new diagnostics. I thought it was desired to qualify such parameters as either const or in.

nordlow:
Got it. I'll change to const for non-template params then. And remove qualifier for template param.

nordlow:
I guess I should disable this diagnostics for templated types.

nordlow:
Done

Geod24:
> Got it. I'll change to const for non-template params then. And remove qualifier for template param.

Sorry if my side note was unclear. Using `in` for `size_t` is fine, but if you want to use `in` in more places, be aware that it shouldn't change the ABI.

src/core/atomic.d
```diff
@@ -162,7 +162,7 @@ void atomicStore(MemoryOrder ms = MemoryOrder.seq, T, V)(ref shared T val, share
  * Returns:
  *  The value held previously by `val`.
  */
-T atomicFetchAdd(MemoryOrder ms = MemoryOrder.seq, T)(ref T val, size_t mod) pure nothrow @nogc @trusted
+T atomicFetchAdd(MemoryOrder ms = MemoryOrder.seq, T)(ref T val, const size_t mod) pure nothrow @nogc @trusted
```
TurkeyMan, line 165: Why adding `const` to by-val arguments? Seems odd.

TurkeyMan:
Sorry, didn't read the thread before looking at the code.

> they trigger warnings when using my new diagnostics

Why would your diagnostics take issue with this? Your diagnostics are problematic if it calls this out as being a problem.

nordlow:
> Why would your diagnostics take issue with this? Your diagnostics are problematic if it calls this out as being a problem.

It's a question of where to draw the line, yes. I can always disable the diagnostics for value types. However, Walter has said in one of his recent lectures that he should have made variable declarations be `const` by default. I presume that includes function parameters.

TurkeyMan:
But that's not how D is. D is not immutable by default.
It's extremely unusual to qualify by-val params like this, especially primitive types. `const`/`in` only make sense if the argument may have indirections.
All the template arguments may have indirections. There's quite a few `ref const` args throughout this file which could be changed to `in`, but you haven't touched those O_o

nordlow:
> All the template arguments may have indirections. There's quite a few `ref const` args throughout this file which could be changed to `in`, but you haven't touched those O_o

I thought that rename should made when druntime defaults to compiling with `-preview=in`.

TurkeyMan:
Is that something that's going to happen any time soon?
That's aside from this PR though. I still don't understand this PR.

Geod24:
I'm going to go ahead an close this, as I don't think there's a path everyone agrees with ahead of us. Feel free to reopen if you think otherwise.

PR #3223, opened 2020-10-03
Convert rt/msvc*.c to D

kinke:
With `pragma(linkerDirective)`, we don't need these 2 modules to be in C anymore.


kinke:
Pinging @rainers. Unfortunately, the files have changed too much, so git doesn't detect them as renamed...

kinke:
```D
printf("&_set_output_format: %p\n", cast(void**) &_set_output_format);
printf("&_nullfunc:          %p\n", &_nullfunc);
printf("NOT equal:           %d\n", cast(void**) &_set_output_format != &_nullfunc ? 1 : 0);
```
=>
```
&_set_output_format: 00007FF6254B2C28
&_nullfunc:          00007FF6254B2C28
NOT equal:           1
```

And I've even excluded `-O -release`...

src/rt/msvc.d
```diff
@@ -0,0 +1,231 @@
...
+__gshared ubyte msvcUsesUCRT;
+
+version (X86)
+    enum cPrefix = "_";
+else
+    enum cPrefix = "";
+
+mixin template declareAlternateName(string name, string alternateName)
+{
+    mixin(`pragma(linkerDirective, "/alternatename:` ~ cPrefix~name ~ `=` ~ cPrefix~alternateName ~ `");`);
```
thewilsonator, line 52: mixin accepts a comma separated list of strings

rainers:
> Unfortunately, the files have changed too much, so git doesn't detect them as renamed...

IIRC it works better if you just rename it in a separate commit.

rainers (approved):

src/rt/msvc.d
kinke: I know, but it doesn't seem like an improvement to me:
```D
    mixin(`pragma(linkerDirective, "/alternatename:`, cPrefix, name, `=`, cPrefix, alternateName, `");`);

12345swordy:
It is all green here.

PR #3224, opened 2020-10-05
rt.monitor_: Remove user-level monitors

ibuclaw:
I don't see this field being set anywhere, and the `Monitor` struct is internal to this module.  So it would appear to be a dead feature to me.


ibuclaw:
Ah-ha, it works using magical indirection.

PR #3226, opened 2020-10-08
[macOS Big Sur] libc symbols for both ARM Mac and x86_64 Mac

p0nce:
Legacy symbols are still needed on macOS 10.16 "Big Sur" when targeting x86_64.
For AArch64 target, there is no need for such legacy symbols and the symbols are the same than on other Darwins.

This commit allows to build druntime and Phobos on Big Sur targetting either `arm64` or `x86_64`.

### DIFF
--- a/src/core/stdc/math.d
+++ b/src/core/stdc/math.d
@@ -1037,14 +1037,25 @@ else version (Darwin)
     // other Darwins
     version (OSX)
     {
-        pure int __fpclassify(real x);
-        pure int __isfinite(real x);
-        pure int __isinf(real x);
-        pure int __isnan(real x);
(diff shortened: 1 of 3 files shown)


thewilsonator (approved):

thewilsonator:
```
src/core/sys/posix/dirent.d:134:        version(AArch64)
src/core/sys/posix/dirent.d:424:    version(AArch64)
src/core/sys/posix/sys/stat.d:2213:        version(AArch64)
src/core/stdc/math.d:1040:        version(AArch64)
```

p0nce:
Updated whitespace and squashed.

PR #3227, opened 2020-10-10
core.stdcpp.utility was not being copied to imports

n8sh:
No bug issue for the changelog because as far as I can see the addition of `core.stdcpp.utility` wasn't mentioned in any release notes.


thewilsonator (approved):

Geod24:
Shouldn't this has gone to stable ?

n8sh:
> Shouldn't this has gone to stable ?

Yes it should have. I made a mistake because my local copy of stable was extremely out of date.

PR #3231, opened 2020-10-10
win**.mak: Use explicit path when running unittests

MoonlightSentinel:
Windows shells might not consider local files when not using an explicit path (`unittest`).

---

Maybe this will solve the reoccurring Azure issue.

### DIFF
--- a/win32.mak
+++ b/win32.mak
@@ -110,7 +110,7 @@ $(DRUNTIME): $(OBJS) $(SRCS) win$(MODEL).mak
 
 unittest : $(SRCS) $(DRUNTIME)
 	*$(DMD) $(UDFLAGS) -L/co $(UTFLAGS) -ofunittest.exe -main $(SRCS) $(DRUNTIME) -debuglib=$(DRUNTIME) -defaultlib=$(DRUNTIME)
-	unittest
+	.\unittest.exe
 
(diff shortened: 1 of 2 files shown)


kinke:
Looks like it does...

Geod24:
Indeed: https://github.com/dlang/dmd/pull/11850#issuecomment-706660571

Geod24 (approved):

rainers:
> Maybe this will solve the reoccurring Azure issue.

AFAICT this will at least report sensible error codes from unittest crashes. See https://github.com/dlang/dmd/pull/11850#issuecomment-706659713
druntime doesn't test the same azure configurations as dmd, namely also not the one using mingw based libraries.

rainers (approved):

WalterBright:
@rainers > AFAICT this will at least report sensible error codes from unittest crashes. See dlang/dmd#11850 (comment)
druntime doesn't test the same azure configurations as dmd, namely also not the one using mingw based libraries.

Azure is still failing the same way. https://github.com/dlang/dmd/pull/11846

Geod24:
@WalterBright : Not the same. Now we get an error code (the exe crash) instead of a cryptic "executable not found".
But yes, the underlying issue is still here and needs to be fixed.

PR #3232, opened 2020-10-12
Fix ci

thewilsonator:
Fingers crossed

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -137,14 +137,17 @@ private string miniFormat(V)(const scope ref V v)
 
         char[60] val;
         int len;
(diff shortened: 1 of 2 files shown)


MoonlightSentinel:
The deprecation are irrelevant. The compilations succeeds but running the executable fails.

thewilsonator:
Oh, Maybe i was misread the `-w` as `-W`.

MoonlightSentinel:
Unless there is an actuall issue with the `real` formatting which causes a crash in Mingw (the Azure job is the only one testing it IIRC)

thewilsonator:
I thought the command line that was set to error on deprecations.

PR #3233, opened 2020-10-12
Fix Azure CI regression

kinke:
I've introduced the regression with #3223, because the `core.stdc.stdio.FILE` definition is for the MSVC 2015+ runtime and incompatible with older versions, such as 2013 used with the MinGW-based libs.

As the druntime repo here hasn't featured any MinGW CI job, the regression was unnoticed here but later showed up for the DMD repo. So I've added a corresponding CI job to prevent stuff like this in the future.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -68,15 +68,20 @@ jobs:
     variables:
       VSINSTALLDIR: C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\
       HOST_DMD_VERSION: 2.093.1
+      D_COMPILER: ldc
+      VISUALD_VER: v0.49.0
+      LDC_VERSION: 1.20.0
     strategy:
       matrix:
         win32-ldc:
           OS: Win_32
-          MODEL: 32
+          MODEL: 32mscoff
           ARCH: x86
-          D_COMPILER: ldc
-          VISUALD_VER: v0.49.0
-          LDC_VERSION: 1.14.0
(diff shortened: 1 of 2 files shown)


thewilsonator (approved):

azure-pipelines.yml
```diff
@@ -68,15 +68,20 @@ jobs:
     variables:
       VSINSTALLDIR: C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\
       HOST_DMD_VERSION: 2.093.1
+      D_COMPILER: ldc
+      VISUALD_VER: v0.49.0
+      LDC_VERSION: 1.20.0
```
kinke, line 73: The DMD repo uses LDC v1.20, not v1.14, so bumped for consistency.

azure-pipelines.yml
```diff
@@ -68,15 +68,20 @@ jobs:
...
     strategy:
       matrix:
         win32-ldc:
           OS: Win_32
-          MODEL: 32
+          MODEL: 32mscoff
```
kinke, line 78: This is no real change, `MODEL` is overwritten in https://github.com/dlang/dmd/blob/3a55c54a8e5773c3f56e10365a98f4570f8c6ce6/.azure-pipelines/windows-msbuild.bat#L9, so this is just for clarity (testing MSCOFF, not OMF).

kinke:
The new Azure job should be made 'Required' to prevent problematic auto-merges.

PR #3234, opened 2020-10-13
Sync Azure CI script with DMD's

kinke:
For consistent job names etc.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,6 +2,7 @@
 
 variables:
   DMD_BRANCH: $[ coalesce(variables['System.PullRequest.TargetBranch'], variables['Build.SourceBranchName'], 'master') ]
+  VSINSTALLDIR: C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\
 
 jobs:
   - job: Windows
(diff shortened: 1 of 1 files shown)


kinke:
@Geod24: This could use the same umbrella branch requirement for Azure. And most likely needs a manual merge due to the Phobos regression on Win64.

Geod24:
> And most likely needs a manual merge due to the Phobos regression on Win64.

All green here, which regression are we talking about ? Did the same modifications here w.r.t. job names.

kinke:
The regression for DMD master wrt. x64 job with LDC as host compiler. - As it works fine with the DMD-host job, I strongly suspect the libcurl.dll copied from the host compiler for the Phobos unittests, which was updated with LDC v1.21 (used version is 1.20)...

kinke:
@Geod24: This one'e still missing in the Azure series.

kinke:
Ping.

thewilsonator (approved):

Geod24:
> @Geod24: This one'e still missing in the Azure series.

Just checked, settings seems correct. Perhaps the UI didn't refresh ?

kinke:
[I meant the PR wasn't merged yet, while the others in the other repos were.]

PR #3236, opened 2020-10-15
core.memory: Add moveToGC() to promote values to new GC allocations

kinke:
As hopefully idiomatic alternative to the `&[value][0]` 'pattern', which doesn't compile anymore since dlang/dmd#10124 (because the element of the array literal is an rvalue now).

### DIFF
--- a/src/rt/lifetime.d
+++ b/src/rt/lifetime.d
@@ -1120,7 +1120,7 @@ extern (C) void[] _d_newarraymiTX(const TypeInfo ti, size_t[] dims)
  * Allocate an uninitialized non-array item.
  * This is an optimization to avoid things needed for arrays like the __arrayPad(size).
  */
-extern (C) void* _d_newitemU(in TypeInfo _ti)
+extern (C) void* _d_newitemU(in TypeInfo _ti) pure nothrow
 {
     auto ti = unqualify(_ti);
     auto flags = !(ti.flags & 1) ? BlkAttr.NO_SCAN : 0;
@@ -1143,7 +1143,7 @@ extern (C) void* _d_newitemU(in TypeInfo _ti)
 }
 
 /// Same as above, zero initializes the item.
-extern (C) void* _d_newitemT(in TypeInfo _ti)
+extern (C) void* _d_newitemT(in TypeInfo _ti) pure nothrow
 {
     import core.stdc.string;
     auto p = _d_newitemU(_ti);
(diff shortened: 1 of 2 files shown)


src/rt/lifetime.d
```diff
@@ -1120,7 +1120,7 @@ extern (C) void[] _d_newarraymiTX(const TypeInfo ti, size_t[] dims)
  * Allocate an uninitialized non-array item.
  * This is an optimization to avoid things needed for arrays like the __arrayPad(size).
  */
-extern (C) void* _d_newitemU(in TypeInfo _ti)
+extern (C) void* _d_newitemU(in TypeInfo _ti) pure nothrow
```
kinke, line 1123: The corresponding `_d_newarray*` functions are `pure nothrow` already: https://github.com/dlang/druntime/blob/b7391cd28b5d1fff9ffa9d6e675f08c4b4b5385f/src/rt/lifetime.d#L936-L1044

ibuclaw (approved):
Seems reasonable.

Geod24:
This won't work with interior pointers, will it ?

kinke:
Nope, but interior pointers aren't valid in D structs anyway (a bit unclear in the spec, but it's §14.0.2 in https://dlang.org/spec/struct.html). At least as long as we don't have move ctors.

Geod24:
> Nope, but interior pointers aren't valid in D structs anyway

Pretty hard to interface with C++ without that. Plus, there are some other needs for controlling the amount of copy / blitting happening. Weka mentioned it as recently as today in Slack ([see discussion](https://dlang.slack.com/archives/C7FT6TPFG/p1602800499075500)).

In the use-case that sparked the discussion, the deserializer was accepting some C++ types, which could have interior pointers (not necessary to themselves). That's why so much care was put on avoiding copy and having NRVO working.

kinke:
It's not like that's something new, but nobody cared to finish the move ctors started by Suleyman. Interior pointers are *bad* and invalid for the time being, and attempts to prevent them don't work generally on all platforms, at least for LDC, I can guarantee that.

kinke:
Ping. The array literal pattern is valid code again, but this should be more appropriate in almost all cases.

thewilsonator (approved):

thewilsonator:
Ping, GH seems stuck on completed CI, please force push to restart.

12345swordy:
Another auto-merge that for some reason isn't merging.

PR #3239, opened 2020-10-18
core.lifetime: Add copyEmplace() to simulate copy-construction

kinke:
With a function signature analogous to `moveEmplace()` - source ref first, then target ref. Allowing non-mutable target types T is required to invoke the correct copy constructor, see the tests.

`std.variant` has a similar [implementation](https://github.com/dlang/phobos/blob/6224cb03a5e272ffc6f0b95b2a8fff498925cc36/std/variant.d#L665-L714) (but insufficient wrt. copy constructors), and there might be more potential usage sites in
druntime/Phobos/user code.

Destructing a partially copied static array has been adopted from copy-constructing a new dynamic array: https://github.com/dlang/druntime/blob/b7391cd28b5d1fff9ffa9d6e675f08c4b4b5385f/src/core/internal/array/construction.d#L49-L71


thewilsonator:
Looks fine to me. cc @RazvanN7

RazvanN7 (approved):

PR #3241, opened 2020-10-19
Fiber: malloc call fix

denizzzka:
Current code silently ignores malloc call even if it available - import is forgotten.

This PR fixes it and adds assert checks to avoid this for other whan Windows or Posix platforms.

### DIFF
--- a/src/core/thread/fiber.d
+++ b/src/core/thread/fiber.d
@@ -954,9 +954,10 @@ private:
             m_ctxt.tstack = pbase;
             m_size = sz;
         }
-        else
+        else version (Posix)
         {
-            version (Posix) import core.sys.posix.sys.mman; // mmap, MAP_ANON
+            import core.sys.posix.sys.mman; // mmap, MAP_ANON
+            import core.stdc.stdlib;
 
             static if ( __traits( compiles, mmap ) )
             {
@@ -981,9 +982,7 @@ private:
                 m_pmem = malloc( sz );
             }
             else
-            {
-                m_pmem = null;
-            }
+                static assert (false);
 
             if ( !m_pmem )
                 onOutOfMemoryError();
@@ -1042,23 +1041,23 @@ private:
         {
             VirtualFree( m_pmem, 0, MEM_RELEASE );
         }
-        else
+        else version (Posix)
         {
             import core.sys.posix.sys.mman; // munmap
+            import core.stdc.stdlib : free;
 
             static if ( __traits( compiles, mmap ) )
             {
                 munmap( m_pmem, m_size );
             }
-            else static if ( __traits( compiles, valloc ) )
-            {
-                free( m_pmem );
-            }
-            else static if ( __traits( compiles, malloc ) )
+            else
             {
                 free( m_pmem );
             }
         }
+        else
+            static assert(false);
+
         m_pmem = null;
         m_ctxt = null;
     }
--- a/src/core/thread/fiber.d
+++ b/src/core/thread/fiber.d
@@ -957,6 +957,7 @@ private:
         else version (Posix)
         {
             import core.sys.posix.sys.mman; // mmap, MAP_ANON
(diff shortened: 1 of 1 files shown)


src/core/thread/fiber.d
```diff
@@ -1042,23 +1041,23 @@ private:
...
-                free( m_pmem );
-            }
-            else static if ( __traits( compiles, malloc ) )
+            else
             {
                 free( m_pmem );
             }
         }
+        else
+            static assert(false);
```
Geod24, line 1059:
```suggestion
            static assert(false, "`Fiber` support is not available on this platform, please raise an issue");
```


Or something similar.

denizzzka: Not needed for now - there are no easy ways to get to versions other than windows or posix.

Geod24: It's not rare to have people stepping in and trying to port D to an architecture / OS we currently do not support.
Also, some people do funky stuff with druntime, building their own.
Those errors message ensure that they get *some kind of clue* of what is failing at a minimum cost to us.

denizzzka: If everything was as you say, then malloc already would work. :-)

Proposed text is a comment, and I consider what obvious comments like this are bad.

thewilsonator: On the contrary, the next poor schmuck that tries to build for a micro controller, e.g. this guy https://forum.dlang.org/thread/erwfgtigvcciohllvaos@forum.dlang.org , is going to hit that error and get 

```
src/core/thread/fiber.d(1018): Error: static assert:  *false* is false.
```
Rather than 
```
src/core/thread/fiber.d(1018): Error: static assert:  "*Fiber* support is not available on this platform, please raise an issue"
```

denizzzka: She uses Posix, as I understand

n8sh:
> Current code silently ignores malloc call even if it available - import is forgotten.
> 
> This PR fixes it

Good! If not removing valloc perhaps also import core.sys.posix.stdlib.

> and adds assert checks to avoid this for other whan Windows or Posix platforms.

Why? In `allocStack` & `freeStack` the only concern is the ability to allocate & deallocate memory. If malloc and free are defined in core.stdc.stdlib then they must work according to ANSI C. There are places where system-specific behavior is required and there is an error if the present system lacks support, like `initStack`. Let's leave the error messages for those places!

denizzzka:
> Good! If not removing valloc perhaps also import core.sys.posix.stdlib.

It is already imported above. I added redundand to clear things.

> If malloc and free are defined in core.stdc.stdlib then they must work according to ANSI C.

My target is to adding support for any platforms, including which not providing C API.

src/core/thread/fiber.d
```diff
@@ -954,9 +954,11 @@ private:
             m_ctxt.tstack = pbase;
             m_size = sz;
         }
-        else
+        else version (Posix)
         {
-            version (Posix) import core.sys.posix.sys.mman; // mmap, MAP_ANON
+            import core.sys.posix.sys.mman; // mmap, MAP_ANON
```
n8sh, line 959: This imported module will always define `mmap` or else compilation will fail with  `static assert(false, "Unsupported platform")`. So there seems to be no way the `malloc` branch can ever be reached.

https://github.com/dlang/druntime/blob/31c9e38d4b5d028a7cc42a09203734ba6609c959/src/core/sys/posix/sys/mman.d#L261-L326

denizzzka: Yes, I brought this fix by accident and did `git push -f`. :-(

I think we just need to get rid of these `__traits( compiles, mmap )` at whole file replacing them by `version(POSIX_MMAP)`, etc.

denizzzka:
This bug is still here, but I don't have time at the moment to deal with it. It will be great if someone fixes it. :-)

PR #3242, opened 2020-10-19
time unittest fix for fast boot devices

denizzzka:
If device is booted and performs druntime unittests, this test will fail if it occures in less than 1 second from device start.

This is funny!


Geod24 (approved):

thewilsonator:
@dlang-bot lazy bot!

PR #3243, opened 2020-10-19
sem_timedwait should use CLOCK_REALTIME

omerfirmak:


src/core/sync/config.d
```diff
@@ -28,7 +28,7 @@ version (Posix)
     {
         static if ( is (typeof ( pthread_condattr_setclock ) ) )
         {
-            clock_gettime( CLOCK_MONOTONIC, &t );
+            clock_gettime( CLOCK_REALTIME, &t );
```
thewilsonator, line 31: Can you add a comment so that it doesn't get changed back?

thewilsonator (commented):
Is it possible to add a test case for this?

src/core/sync/config.d
omerfirmak: I added the comment but cant think of a test case.

thewilsonator: OK neither can I.

thewilsonator (approved):

n8sh:
This PR would fix timed wait in Sempahore while breaking it in Condition and Event.

`mktspec` is used in three places in druntime: core.sync.condition.Condition.wait(Duration), core.sync.event.Event.wait(Duration), and core.sync.semaphore.Semaphore.wait(Duration). If `pthread_condattr_setclock` is available both Condition and Duration use it in their constructors to change their pthread_cond_t to use CLOCK_MONOTONIC:

https://github.com/dlang/druntime/blob/de974792e612f8dca20736db403069ce9d95b1c3/src/core/sync/condition.d#L119-L137
https://github.com/dlang/druntime/blob/de974792e612f8dca20736db403069ce9d95b1c3/src/core/sync/event.d#L108-L119

That's why there is `static if ( is (typeof ( pthread_condattr_setclock ) ) )` inside `mktspec` around the block that uses CLOCK_MONOTONIC. **However**, Semaphore uses semaphore_t which *doesn't* have a way to change `sem_timedwait` to use CLOCK_MONOTONIC.

In my view the correct solution is to leave `mktspec` unchanged and only change Semaphore.wait(Duration) internally from:


```d
        else version (Posix)
        {
            timespec t = void;
            mktspec( t, period );
            // etc...
        }
```
to
```d
        else version (Posix)
        {
            import core.sys.posix.time : clock_gettime, CLOCK_REALTIME;
            timespec t = void;
            clock_getime( CLOCK_REALTIME, &t );
            mvtspec( t, period) ;
            // etc...
        }
```

n8sh (dismissed):
See above comment. Take care not to break `core.sync.condition.Condition.wait(Duration)` and `core.sync.event.Event.wait(Duration)`.

omerfirmak:
> This PR would fix timed wait in Sempahore while breaking it in Condition and Event.
> 
> `mktspec` is used in three places in druntime: core.sync.condition.Condition.wait(Duration), core.sync.event.Event.wait(Duration), and core.sync.semaphore.Semaphore.wait(Duration). If `pthread_condattr_setclock` is available both Condition and Duration use it in their constructors to change their pthread_cond_t to use CLOCK_MONOTONIC:
> 
> https://github.com/dlang/druntime/blob/de974792e612f8dca20736db403069ce9d95b1c3/src/core/sync/condition.d#L119-L137
> 
> https://github.com/dlang/druntime/blob/de974792e612f8dca20736db403069ce9d95b1c3/src/core/sync/event.d#L108-L119
> 
> That's why there is `static if ( is (typeof ( pthread_condattr_setclock ) ) )` inside `mktspec` around the block that uses CLOCK_MONOTONIC. **However**, Semaphore uses semaphore_t which _doesn't_ have a way to change `sem_timedwait` to use CLOCK_MONOTONIC.
> 
> In my view the correct solution is to leave `mktspec` unchanged and only change Semaphore.wait(Duration) internally from:
> 
> ```d
>         else version (Posix)
>         {
>             timespec t = void;
>             mktspec( t, period );
>             // etc...
>         }
> ```
> 
> to
> 
> ```d
>         else version (Posix)
>         {
>             import core.sys.posix.time : clock_gettime, CLOCK_REALTIME;
>             timespec t = void;
>             clock_getime( CLOCK_REALTIME, &t );
>             mvtspec( t, period) ;
>             // etc...
>         }
> ```

Oh, I missed the constructors. If there is a specific reason event and condvar works with CLOCK_MONOTONIC, your solution looks like the right way to go.

n8sh (approved):

PR #3245, opened 2020-10-21
Fix Issue 14226 - Don't initialize/overwrite global trace handler with each thread

kinke:
Instead, initialize it once as part of `rt_init()`.

Keep it as a module ctor instead of manually initializing it in `rt_init()` because `defaultTraceHandler` might depend on other modules and their module ctors; that's the case for Windows, where generating a trace requires the `core.sys.windows.stacktrace` module ctor to be run first.

### DIFF
--- a/src/core/runtime.d
+++ b/src/core/runtime.d
@@ -105,7 +105,7 @@ private
 }
 
 
-static this()
+shared static this()
 {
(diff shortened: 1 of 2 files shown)


schveiguy:
LGTM

schveiguy (approved):

thewilsonator (approved):

PR #3247, opened 2020-10-22
Fiber: stack allocation issue fix

denizzzka:
If memory is over at call `new StackContext` it will lead to start collect routine, and then it calls call Fiber dtor and then freeStack().

It seems to me this is expensive solution but I do not know how to make it easier.

### DIFF
--- a/src/core/thread/fiber.d
+++ b/src/core/thread/fiber.d
@@ -906,7 +906,11 @@ private:
         //       room for this struct explicitly would be to mash it into the
         //       base of the stack being allocated below.  However, doing so
         //       requires too much special logic to be worthwhile.
+
+        import core.memory : GC;
+        GC.disable(); // to avoid destruction of this fiber if GC collection will be invoked
         m_ctxt = new StackContext;
+        GC.enable();
 
         version (Windows)
         {


denizzzka:
I'll see later, some kind of confusion.

denizzzka:
Stack trace how it looks without this PR:
https://gist.github.com/denizzzka/7187fa61e8ed8d9b0166c340d5989622

wilzbach:
Any chance we can add a test for this?

denizzzka:
It should be some kind of separate test with a tiny amount of GC memory available - every access to a GC allocation should invoke GC collection routine.

wilzbach:
The existing GC tests are here: https://github.com/dlang/druntime/tree/master/test/gc.
Would a simple `ulimit` help to recreate the issue?

denizzzka:
> Would a simple ulimit help to recreate the issue?

Maybe it will sufficient to use RT options like:
```D
rt_options = [ "gcopt=minPoolSize:16K maxPoolSize:5M incPoolSize:4K" ];
```

rainers:
Maybe I am missing something, but the pointer to the released object is on the stack as shown in your stack trace. So it should not have been collected to begin with.

Are you sure stacks are scanned at all by the GC? I suspect you are working on a version of druntime with major changes as line numbers in the stack trace don't match current druntime.

denizzzka:
> Are you sure stacks are scanned at all by the GC? I suspect you are working on a version of druntime with major changes as line numbers in the stack trace don't match current druntime.

Yes, I use fork of ldc version of druntime what is forked from ~master.

I noticed that another druntime unittest on my side does not work and it also causes an error related to the GC. I'll first get this test to work.

> Maybe I am missing something, but the pointer to the released object is on the stack as shown in your stack trace. So it should not have been collected to begin with.

If the behavior described in this issue is impossible, then this issue can be closed.

PR #3249, opened 2020-10-23
Last one fiber context removal contract fix

denizzzka:
If there is only one fiber context left and then `rt_term()` called `gc_term()` it tries to remove this context, but this contract prohibits to remove latest context because prev and next both is null in this case.


thewilsonator (approved):

denizzzka:
Not confirmed: wrong `thread_init()` implementation on my side

PR #3253, opened 2020-10-25
Fiber construction: GC test added

denizzzka:


test/gc/fiber_ctor.d
```diff
@@ -0,0 +1,14 @@
+// Check avoid of destruction of fiber if GC collection will be invoked during Fiber construction
```
thewilsonator, line 1: `//Check that the Fiber is not also collected when a GC collection is invoked during Fiber construction`

denizzzka:
This PR will not needed for production - problem found is on my side.

PR #3254, opened 2020-10-27
Fix - core.stdcpp.string.basic_string does not implement opEquals

n8sh:


thewilsonator (approved):

src/core/stdcpp/string.d
```diff
@@ -167,6 +167,24 @@ extern(D):
...
+
+    /// Performs lexicographical comparison.
+    bool opCmp(scope const ref basic_string rhs) const pure nothrow @safe   { return __cmp(as_array, rhs.as_array); }
+    /// ditto
+    bool opCmp(basic_string rhs) const                                      { return __cmp(as_array, rhs.as_array); }
+    /// ditto
+    bool opCmp(scope const T[] rhs) const pure nothrow @safe                { return __cmp(as_array, rhs); }
+
+    /// Hash to allow `basic_string`s to be used as keys for built-in associative arrays.
+    /// **The result will generally not be the same as C++ `std::hash<std::basic_string<T>>`.**
```
TurkeyMan, line 185: Maybe we should work this so that the result is the same as `std::hash`, although this is good for now.

src/core/stdcpp/string.d
```diff
@@ -167,6 +167,24 @@ extern(D):
     ///
     inout(T)[] opIndex() inout pure nothrow @safe @nogc                     { return as_array(); }
 
+    /// Two `basic_string`s are equal if they represent the same sequence of code units.
+    bool opEquals(scope const ref basic_string s) const pure nothrow @safe  { return as_array == s.as_array; }
+    /// ditto
+    bool opEquals(basic_string s) const                                     { return as_array == rhs.as_array; }
+    /// ditto
+    bool opEquals(scope const T[] s) const pure nothrow @safe               { return as_array == s; }
```
TurkeyMan, line 175: Do these work? This looks like it's comparing the slices? Ie, is it the same `ptr` and `length`, not is the string equal?

src/core/stdcpp/string.d
```diff
@@ -167,6 +167,24 @@ extern(D):
     ///
     inout(T)[] opIndex() inout pure nothrow @safe @nogc                     { return as_array(); }
 
+    /// Two `basic_string`s are equal if they represent the same sequence of code units.
+    bool opEquals(scope const ref basic_string s) const pure nothrow @safe  { return as_array == s.as_array; }
+    /// ditto
+    bool opEquals(basic_string s) const                                     { return as_array == rhs.as_array; }
```
TurkeyMan, line 173: What is this overload for? The function above is the same. Also, the operand on the right is `rhs.as_array`, but the parameter is `s`...?

TurkeyMan:
I actually had a good reason for these not existing originally... although I can't remember what it was.
I think I might have intended to do this:
`myString[] == myOtherString[]` for comparisons.
...why did I think that? I'm sure I had a good reason...

src/core/stdcpp/string.d
n8sh: `basic_string` has a destructor that zeroes its member so passing it by reference does less work and when passing by value one can't convert non-const to const.

src/core/stdcpp/string.d
n8sh: Yup.
```d
void main()
{
    auto s1 = "abc";
    auto s2 = "abc".idup;
    assert(s1 == s2); // Passes: equal.
    assert(s1.ptr !is s2.ptr); // Passes: pointers are not the same.
    assert(s1 !is s2); // Passes: the pointers are not the same.
}
```
`s1 == s2` gets lowered to `__equals(s1, s2)`.
https://github.com/dlang/druntime/blob/31c9e38d4b5d028a7cc42a09203734ba6609c959/src/core/internal/array/equality.d#L18-L48

src/core/stdcpp/string.d
n8sh: Also, oops fixed.

src/core/stdcpp/string.d
TurkeyMan: Ummm... wtaf?
Why would the slice comparison compare the contents element-wise? `s1[] == s2[]` should do that...

src/core/stdcpp/string.d
TurkeyMan: Why would you ever pass by value though? I think this by-val comparison shouldn't exist.

n8sh: You may be right. A pass-by-value overload is necessary to make things like `s == some_function_that_returns_a_string()` work when not compiling with `-preview=rvaluerefparam`, but when `-preview=rvaluerefparam` is enabled then an explicit pass-by-value overload is a pessimization.

TurkeyMan: `-preview=rvaluerefparam` has been in testing for a long time now; hopefully it becomes enabled by default in the near-ish future.

n8sh: Removed the pass-by-value overloads. This may cause existing code in the wild using stdcpp.string with default struct equality to stop compiling, but as that code was only producing the correct result for small strings this may be acceptable.

TurkeyMan: I'm fine with this. Thanks.

PR #3255, opened 2020-10-28
Fix - core.stdcpp.vector.vector does not implement opEquals

n8sh:


src/core/stdcpp/vector.d
```diff
@@ -114,6 +114,17 @@ extern(D):
     ///
     void append(T[] array)                                                  { insert(length, array); }
 
+    /// Performs elementwise equality check.
+    bool opEquals(this This, That)(auto ref That rhs)
```
n8sh, line 118: The complicated signature is because opEquals might or might not be callable with a const vector!T for any particular T.

src/core/stdcpp/vector.d
```diff
@@ -114,6 +114,17 @@ extern(D):
...
+    if (is(immutable That == immutable vector))                             { return as_array == rhs.as_array; }
+
+    /// Performs lexicographical comparison.
+    int opCmp(this This, That)(auto ref That rhs)
+    if (is(immutable That == immutable vector))                             { return __cmp(as_array, rhs.as_array); }
+
+    /// Hash to allow `vector`s to be used as keys for built-in associative arrays.
+    /// **The result will generally not be the same as C++ `std::hash<std::vector<T>>`.**
+    size_t toHash() const                                                   { return .hashOf(as_array); }
```
n8sh, line 127: `toHash` on the other hand needs to be a non-template const function or else it will be ignored by the compiler for the purpose of determining if a particular struct has a non-default hash code.

src/core/stdcpp/vector.d
```diff
@@ -114,6 +114,18 @@ extern(D):
     ///
     void append(T[] array)                                                  { insert(length, array); }
 
+    /// Performs elementwise equality check.
+    bool opEquals(this This, That)(auto ref That rhs)
+    if (is(immutable That == immutable vector))                             { return as_array == rhs.as_array; }
+
+    /// Performs lexicographical comparison.
+    static if (is(typeof((ref T a, ref T b) => a < b)))
```
n8sh, line 122: The `static if` is so `__traits(hasMember, vector!T, "opCmp")` will be false for vectors that do not support comparison.

12345swordy:
@TurkeyMan review this please.

src/core/stdcpp/vector.d
TurkeyMan: Please leave an obvious TODO that we should use std::hash in the future when it's written?

src/core/stdcpp/vector.d
TurkeyMan: Is this definitely the best way to express this comparison?

TurkeyMan (approved):
LGTM

src/core/stdcpp/vector.d
n8sh: Not definitely.

PR #3256, opened 2020-10-28
Implement some TODOs in core.stdcpp.vector

n8sh:


src/core/stdcpp/vector.d
```diff
@@ -452,8 +453,9 @@ extern(D):
 
         static void _Destroy(pointer _First, pointer _Last)
         {
-            for (; _First != _Last; ++_First)
-                destroy!false(*_First);
+            static if (hasElaborateDestructor!T)
+                for (; _First !is _Last; ++_First)
+                    destroy!false(*_First);
```
TurkeyMan, line 458: Why is this necessary?

src/core/stdcpp/vector.d
```diff
@@ -633,6 +635,7 @@ extern(D):
                 _First = _Dest;
                 _Last = _Dest;
             }
+            static if (hasElaborateDestructor!T)
```
TurkeyMan, line 638: C++ doesn't have static-if, `vector` always has a destructor.

src/core/stdcpp/vector.d
```diff
@@ -651,48 +654,105 @@ extern(D):
             pointer _First;
             pointer _Last;
         }
-        pointer _Utransfer(bool _move, bool _ifNothrow = false)(pointer _First, pointer _Last, pointer _Dest)
+        static pointer _Utransfer(bool _move, bool _ifNothrow = false)(pointer _First, pointer _Last, pointer _Dest)
```
TurkeyMan, line 657: It can be static, but I'm not happy to change the implementation from Microsoft's implementation. It should remain identical, at least for the time being.

TurkeyMan: To be clear, I did leave those TODO's here, because I think they're valid improvements... but I don't think we have unit tests that tests comprehensively enough to detect breakages that we might introduce by deviating from the reference implementation. I wanted to be confident our tests are very robust before making speculative optimisations that deviate from the C++ implementation.

src/core/stdcpp/vector.d
n8sh: The rationale was that not having a destructor (even if it's empty) could prevent unnecessary stack-unwinding code from being inserted.

>in the event that the linker chooses one or the other, it may mix & match.

Since every function is extern(D) the mangling will never match. Wasn't that why you made them extern(D) when you wrote this?

TurkeyMan: Oh yeah! That's exactly right... Haha, I haven't thought on this for like a year. I've forgotten all my working state >_<

So, I'm excited you're motivated to contribute to this. I think your work on the TODO's is cool, but like I say, I'm still not confident the test suite is comprehensive enough to deviate from the reference implementations. Are you interested in building out the test suite?

I think a good way to test this, would be to find OSS tests for the C++ library; maybe libc++ has the test suite source available, and adapt the tests to D.
The ultimate goal here, is to collapse the 3 distinct implementations into just one implementation.
The structs are different, and the allocation strategies are slightly different, so we'd keep versions of those functions, but the higher level functions, like the copy/move logic should be able to be centralised and not duplicated.

I wanted to do that last, since once you lose the reference code, it's very hard to refer back to the original C++ source and copy the functionality precisely. I'm not sure all of the internal functions are fleshed out yet, and the tests will help reveal what's missing.

src/core/stdcpp/vector.d
n8sh: That males sense.

RazvanN7:
@TurkeyMan @n8sh What is the status on this PR? How do we move forward?

PR #3258, opened 2020-10-29
Adds out of memory check

denizzzka:
Can be invoked if try to allocate big pool on small memory amount

### DIFF
--- a/src/gc/os.d
+++ b/src/gc/os.d
@@ -132,6 +132,8 @@ else static if (is(typeof(malloc))) // else version (GC_Use_Alloc_Malloc)
     void *os_mem_map(size_t nbytes) nothrow
     {   byte *p, q;
         p = cast(byte *) malloc(nbytes + PAGESIZE);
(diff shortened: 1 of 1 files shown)


src/gc/os.d
```diff
@@ -130,8 +130,13 @@ else static if (is(typeof(malloc))) // else version (GC_Use_Alloc_Malloc)
...
 
     void *os_mem_map(size_t nbytes) nothrow
-    {   byte *p, q;
+    {
+        import core.exception : onOutOfMemoryError;
+
+        byte *p, q;
         p = cast(byte *) malloc(nbytes + PAGESIZE);
+        if(!p)
+            onOutOfMemoryError();
```
rainers, line 139: I think you should just return null instead, the GC handles it correctly AFAICT.

denizzzka: Yes you are right. Fixed

rainers (approved):

PR #3259, opened 2020-10-29
MEMSTOMP debug is not compatible with this GC unittest

denizzzka:
Is that correct?


rainers:
> Is that correct?

You are right, but I think you could only exclude the explicit `destroy` call and the following assertion.

rainers (approved):

PR #3260, opened 2020-10-30
Moves ProtoGC.term() into gc.proxy.transfer()

denizzzka:
Reduces amount of magic: it was too hard to understand how roots and ranges moved to new GC.

### DIFF
--- a/src/gc/proxy.d
+++ b/src/gc/proxy.d
@@ -76,10 +76,28 @@ extern (C)
             }
             _instance = newInstance;
             // Transfer all ranges and roots to the real GC.
-            (cast(ProtoGC) protoInstance).term();
+            (cast(ProtoGC) protoInstance).transfer();
             isInstanceInit = true;
         }
         instanceLock.unlock();
+    }
+
+    // Call this function when initializing the real GC.
+    // This function should be called after the real GC is in place.
+    private void transfer(ProtoGC src)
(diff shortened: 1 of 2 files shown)


src/gc/proxy.d
```diff
@@ -76,12 +76,30 @@ extern (C)
             }
             _instance = newInstance;
             // Transfer all ranges and roots to the real GC.
-            (cast(ProtoGC) protoInstance).term();
+            (cast(ProtoGC) protoInstance).transfer();
```
rainers, line 79: Renaming seems reasonable, but the name could be even more descriptive making the comment obsolete, e.g. `transferRangesAndRoots`.

src/gc/proxy.d
```diff
@@ -76,12 +76,30 @@ extern (C)
...
             isInstanceInit = true;
         }
         instanceLock.unlock();
     }
 
+    // Call this function when initializing the real GC.
+    // This function should be called after the real GC is in place.
+    private void transfer(ProtoGC src)
```
rainers, line 87: Moving this function exposes internals of ProtoGC here. I don't think that's desirable.

rainers (changes requested):

denizzzka:
@rainers https://github.com/dlang/druntime/pull/3260

PR #3262, opened 2020-10-31
Demangle consumes megabytes of memory for each unsucessful demangle

denizzzka:
This PR shows this behaviour

### DIFF
--- a/src/core/demangle.d
+++ b/src/core/demangle.d
@@ -1991,6 +1991,8 @@ pure @safe:
                 auto b = 2 * dst.length;
                 auto newsz = a < b ? b : a;
                 debug(info) printf( "growing dst to %lu bytes\n", newsz );
+                version (CoreUnittest)
+                    assert(newsz <= 100 * 1024, `I want to use more than 100Kb for this paltry string`);
                 dst.length = newsz;
                 pos = len = brp = 0;
                 continue;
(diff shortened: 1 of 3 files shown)


denizzzka:
I can't figure out where is bug. This happens when RAM amount are not large and then at the one moment OverflowException handler calls out of memory exception while buf is grows and grows.

core.demangle looks as exception-driven and statefull code.. :- (

ibuclaw:
> I can't figure out where is bug. This happens when RAM amount are not large and then at the one moment OverflowException handler calls out of memory exception while buf is grows and grows.
> 
> core.demangle looks as exception-driven and statefull code.. :- ( 

Yeah. The backtrace printer for uncaught throwables uses try/catch IIRC for attempting to demangle found function names. Which doesn't bode well in some scenarios.

denizzzka:
> The backtrace printer for uncaught throwables uses try/catch IIRC for attempting to demangle found function names.

Yea, code must be changed to avoid exceptions. Maybe just replace exceptions by struct like { errStatus; value; }

rainers:
I just tried your test cases: there are only the regular 4k allocations once it starts unpacking, nothing unexpected, on run.dlang.io or Windows. You can avoid these allocations if you pass a destination buffer to `demangle`.

denizzzka:
> I just tried your test cases

This is first part of unittests from core.demangle what fails on small system at step i == 2. When this step occurs buf needs 1Mb * 2 in Overflow handler.

Maybe previous unittests also consume buf before to get into this condition

PR #3263, opened 2020-10-31
Simplify hasElaborateDestructor, hasElaborateCopyConstructor, isPointer

n8sh:


src/core/internal/traits.d
```diff
@@ -275,8 +275,7 @@ template hasElaborateDestructor(S)
     }
     else static if (is(S == struct))
     {
-        enum hasElaborateDestructor = __traits(hasMember, S, "__dtor")
-            || anySatisfy!(.hasElaborateDestructor, Fields!S);
+        enum hasElaborateDestructor = __traits(hasMember, S, "__xdtor");
```
n8sh, line 278: This doesn't work because it fails
```d
static struct S6 { S3[0] field; }
static assert(!hasElaborateDestructor!S6);
```

n8sh:
Closing this. It seems the right thing to do is first address this DMD-side to avoid unnecessary __xdtor / __xpostblit for zero-length static arrays and structs recursively containing them.

PR #3265, opened 2020-11-09
core.stdcpp.allocator: _Adjust_manually_vector_aligned checks for sentinel unconditionally (Windows only)

n8sh:
This bug resulted in assertion failures when deleting large blocks of memory using core.stdcpp.allocator on Windows. `_Allocate_manually_vector_aligned` only sets a sentinel when `version(_DEBUG)` so `_Adjust_manually_vector_aligned` should only check for this sentinel when `version(_DEBUG)`.

Additionally change the linkage of those functions from C++ to D to avoid possible linker confusion.

### DIFF
--- a/src/core/stdcpp/allocator.d
+++ b/src/core/stdcpp/allocator.d
@@ -283,8 +283,15 @@ version (CppRuntime_Microsoft)
         version = INTEL_ARCH;
 
     // HACK: should we guess _DEBUG for `debug` builds?
-    version (NDEBUG) {}
-    else debug version = _DEBUG;
+    version (_DEBUG)
+        enum _DEBUG = true;
+    else version (NDEBUG)
+        enum _DEBUG = false;
+    else
+    {
+        import core.stdcpp.xutility : __CXXLIB__;
+        enum _DEBUG = __CXXLIB__.length && 'd' == __CXXLIB__[$-1]; // libcmtd, msvcrtd
+    }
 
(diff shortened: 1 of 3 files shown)


TurkeyMan (approved):
Thanks for isolating this!

thewilsonator:
Hmm, the test seems to hang on windows

n8sh:
> Hmm, the test seems to hang on windows

My guess is that this is because the test is actually failing on Windows and an "abort or debug?" prompt is appearing. IIRC that happened recently on another PR. My guess for why a failure would be is happening is a difference in that, unlike xutility.d,  allocator.d does not infer _DEBUG based on mscrtlib . I'll see if changing it fixes this.

thewilsonator:
nope.

n8sh:
Should be working now.

src/core/stdcpp/allocator.d
```diff
@@ -283,8 +283,15 @@ version (CppRuntime_Microsoft)
...
-    version (NDEBUG) {}
-    else debug version = _DEBUG;
+    version (_DEBUG)
+        enum _DEBUG = true;
+    else version (NDEBUG)
+        enum _DEBUG = false;
+    else
+    {
+        import core.stdcpp.xutility : __CXXLIB__;
+        enum _DEBUG = __CXXLIB__.length && 'd' == __CXXLIB__[$-1]; // libcmtd, msvcrtd
```
TurkeyMan, line 293: I don't know if this is good. If the logic that drives the codegen doesn't match C++, then you'll get weird link errors that are hard for a user to diagnose.

PR #3266, opened 2020-11-09
core.internal.convert: Add support for FreeBSD x86 reals

ibuclaw:
The second change just makes it easier to slip in `version (GNU) { version (X86) { ... } }` into the FreeBSD block that sets FPU precision.

### DIFF
--- a/src/core/internal/convert.d
+++ b/src/core/internal/convert.d
@@ -37,7 +37,7 @@ const(ubyte)[] toUbyte(T)(const ref T val) if (__traits(isFloating, T) && (is(T 
 {
     if (__ctfe)
     {
-        static if (T.mant_dig == float.mant_dig || T.mant_dig == double.mant_dig)
+        static if (floatFormat!T == FloatFormat.Float || floatFormat!T == FloatFormat.Double)
         {
             static if (is(T : ireal)) // https://issues.dlang.org/show_bug.cgi?id=19932
                 const f = val.im;
@@ -624,7 +624,14 @@ template floatFormat(T) if (is(T:real) || is(T:ireal))
     static if (T.mant_dig == 24)
         enum floatFormat = FloatFormat.Float;
     else static if (T.mant_dig == 53)
-        enum floatFormat = FloatFormat.Double;
+    {
+        // Double precision, or real == double
+        static if (T.sizeof == double.sizeof)
+            enum floatFormat = FloatFormat.Double;
+        // 80-bit real with rounding precision set to 53 bits.
+        else static if (T.sizeof == real.sizeof)
+            enum floatFormat = FloatFormat.Real80;
+    }
     else static if (T.mant_dig == 64)
         enum floatFormat = FloatFormat.Real80;
     else static if (T.mant_dig == 106)
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -450,15 +450,8 @@ private extern (C) int _d_run_main2(char[][] args, size_t totalArgsLength, MainF
     version (CRuntime_Microsoft)
     {
         // enable full precision for reals
-        version (GNU)
+        version (D_InlineAsm_X86_64)
         {
-            size_t fpu_cw;
-            asm { "fstcw %0" : "=m" (fpu_cw); }
-            fpu_cw |= 0b11_00_111111;  // 11: use 64 bit extended-precision
-                                       // 111111: mask all FP exceptions
-            asm { "fldcw %0" : "=m" (fpu_cw); }
-        }
-        else version (Win64)
             asm
             {
                 push    RAX;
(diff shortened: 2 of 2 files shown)


thewilsonator (approved):
You might want to put a note in so this doesn't get accidentally changed

ibuclaw:
> You might want to put a note in so this doesn't get accidentally changed

Or if it's OK, I'll add in the gdc-iasm code?

thewilsonator:
Sure

thewilsonator:
I'll leave merging to you.

ibuclaw:
Done, @kinke does the `GNU_OR_LDC` change look OK?

src/core/internal/convert.d
```diff
@@ -624,7 +624,14 @@ template floatFormat(T) if (is(T:real) || is(T:ireal))
     static if (T.mant_dig == 24)
         enum floatFormat = FloatFormat.Float;
     else static if (T.mant_dig == 53)
-        enum floatFormat = FloatFormat.Double;
+    {
+        // Double precision, or real == double
+        static if (T.sizeof == double.sizeof)
+            enum floatFormat = FloatFormat.Double;
+        // 80-bit real with rounding precision set to 53 bits.
+        else static if (T.sizeof == real.sizeof)
```
n8sh, line 632: I wonder how much code I've written that assumes that anything with 53 bits of mantissa has the size of double or cdouble.

n8sh (approved):

kinke:
> does the `GNU_OR_LDC` change look OK?

It does, but it won't be used by LDC, because it also predefines `D_InlineAsm_*`, and those take precedence. It probably doesn't really matter, but ideally there should probably be a clobber for the FPU flags.

src/core/internal/convert.d
kinke: That's fine for LDC. - How do you select this format with GDC, Iain?

ibuclaw: Outside of phobos?  I doubt it's very much widespread.

src/core/internal/convert.d
```diff
@@ -37,7 +37,7 @@ const(ubyte)[] toUbyte(T)(const ref T val) if (__traits(isFloating, T) && (is(T
 {
     if (__ctfe)
     {
-        static if (T.mant_dig == float.mant_dig || T.mant_dig == double.mant_dig)
+        static if (floatFormat!T == FloatFormat.Float || floatFormat!T == FloatFormat.Double)
```
n8sh, line 40:
```suggestion
        static if (T.sizeof == float.sizeof || T.sizeof == double.sizeof)
```

src/core/internal/convert.d
kinke: I meant compiler cmdline flags or such to enable x87 real with `mant_dig == 53`.

src/core/internal/convert.d
ibuclaw: Does this really matter given that floatFormat is likely instantiated anyway, so we're just going to hit the template cache?

ibuclaw:
> but ideally there should probably be a clobber for the FPU flags.

Hmm, I just did a copy-paste of the original MinGW code, and assumed it was correct.  I looked again and it isn't, so 64-bit precision wasn't being set after all!  I'll have to re-run phobos unittests again from the top. :-(

src/core/internal/convert.d
ibuclaw: @kinke - sorry! I was responding to the first comment. :-)

`mant_dig == 53` is the default for i586-freebsd targets, there's no way to override this as far as I know ([see freebsd support headers](https://github.com/D-Programming-GDC/gcc/blob/cfc2ade5d0f5dc255b49d496c0d6d845e0cddcce/gcc/config/i386/freebsd.h#L120-L123)).

kinke: For C and C++, okay, but it should be trivial to use the regular 64 for GDC, given that you're setting it to 64 right here in `_d_run_main2` for all programs with D main.

src/core/internal/convert.d
n8sh: Even if it matters it probably doesn't matter.

src/core/internal/convert.d
ibuclaw: @kinke - it's hardcoded in the target headers, so there'll have to be some messing around and convincing maintainers that it's needed.

By and large, I don't think it's really worth caring about, supporting it is very much minimal, and we mostly only care about unittests.

And... I've just ran std.math unittests with working fpu setting, and tests are failing in more places than before.  Possibly because real literals are truncated to 53-bits by gcc backend.  So I'll just remove the FPU setting as it's an unsupported configuration.

ibuclaw:
Any outstanding comments here?

PR #3267, opened 2020-11-10
Fix Issue 21110 - OOB memory access, safety violation

RazvanN7:
It seems that this PR [1] forgot to add checks for OOB accesses. ~~I had to modify `_d_assert_fail` to be able to put a nice error message.~~

[1] https://github.com/dlang/druntime/pull/1891

### DIFF
--- a/src/core/internal/array/operations.d
+++ b/src/core/internal/array/operations.d
@@ -38,6 +38,14 @@ T[] arrayOp(T : T[], Args...)(T[] res, Filter!(isType, Args) args) @trusted @nog
     alias scalarizedExp = staticMap!(toElementType, Args);
     alias check = typeCheck!(true, T, scalarizedExp); // must support all scalar ops
 
+    foreach (argsIdx, arg; typeof(args))
+    {
+        static if (is(arg == U[], U))
+        {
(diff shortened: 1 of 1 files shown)


src/core/internal/array/operations.d
```diff
@@ -635,3 +647,28 @@ unittest
     result[] = -data[];
     assert(result[0] == -0.5);
 }
+
+// https://issues.dlang.org/show_bug.cgi?id=21110
+version(none) unittest
```
thewilsonator, line 652: Why the `version(none)`?

thewilsonator (commented):
Please target stable

src/core/internal/array/operations.d
RazvanN7: Forgot it in there.

thewilsonator (approved):

Geod24:
Test seems to be failing on all platforms

RazvanN7:
@Geod24 @thewilsonator This seems to be green now. Could you please take another look?

thewilsonator:
This still targets master.

RazvanN7:
It'a an old regression. Since 2.065.

RazvanN7:
Of course now we have spurious error messages :((

PR #3268, opened 2020-11-11
Adds out of memory check

denizzzka:


denizzzka:
Another one: https://github.com/dlang/druntime/blob/master/src/core/thread/osthread.d#L2552

thewilsonator (approved):

PR #3270, opened 2020-11-13
freebsd: Don't use versioned symbols with GDC

ibuclaw:
Support for `.symver` in the GNU toolchain only came in starting from [binutils 2.35](https://sourceware.org/binutils/docs-2.35/as/Symver.html#Symver), even then, calling such symbols directly is still not supported by gas.

See https://d.godbolt.org/z/PsdPGT

### DIFF
--- a/src/core/sys/freebsd/sys/event.d
+++ b/src/core/sys/freebsd/sys/event.d
@@ -161,13 +161,23 @@ enum
 }
 
 int kqueue();
-static if (__FreeBSD_version >= 1200000)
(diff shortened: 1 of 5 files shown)


kinke:
['GNU targets' is misleading - 'with GDC' would be more like it.]

Geod24:
Needs a rebase

ibuclaw:
@Geod24 rebased.

kinke (approved):

PR #3271, opened 2020-11-13
Issue 17596: Version bindings so that they work for both FreeBSD 11 and 12

ibuclaw:
Rebase of #2280, with `__FreeBSD_version` nits.

Haven't yet tested the second commit against gdc/freebsd.  But the first passes the testsuite and all unittests.

### DIFF
--- a/src/core/sys/freebsd/sys/event.d
+++ b/src/core/sys/freebsd/sys/event.d
@@ -39,7 +39,7 @@ enum
     EVFILT_SYSCOUNT =  11,
 }
 
-static if (__FreeBSD_version >= 1200000 && __FreeBSD_version < 1300000)
+static if (__FreeBSD_version >= 1200000)
 {
     struct kevent_t
     {
@@ -52,7 +52,7 @@ static if (__FreeBSD_version >= 1200000 && __FreeBSD_version < 1300000)
         ulong[4]  ext;
     }
 }
-else static if (__FreeBSD_version < 1200000)
+else
 {
     struct kevent_t
     {
@@ -64,8 +64,6 @@ else static if (__FreeBSD_version < 1200000)
         void        *udata; /* opaque user data identifier */
     }
 }
-else
-    static assert(0, "Unsupported version of FreeBSD");
 
 extern(D) void EV_SET(kevent_t* kevp, typeof(kevent_t.tupleof) args)
 {
--- a/src/core/sys/freebsd/sys/event.d
+++ b/src/core/sys/freebsd/sys/event.d
@@ -161,7 +161,13 @@ enum
 }
 
 int kqueue();
-pragma(mangle, "kevent@FBSD_1.0")
-int kevent(int kq, const kevent_t *changelist, int nchanges,
-           kevent_t *eventlist, int nevents,
-           const timespec *timeout);
+static if (__FreeBSD_version >= 1200000)
+    pragma(mangle, "kevent@@FBSD_1.5")
+    int kevent(int kq, const kevent_t *changelist, int nchanges,
+               kevent_t *eventlist, int nevents,
+               const timespec *timeout);
+else
+    pragma(mangle, "kevent@FBSD_1.0")
+    int kevent(int kq, const kevent_t *changelist, int nchanges,
+               kevent_t *eventlist, int nevents,
+               const timespec *timeout);
--- a/src/core/sys/freebsd/sys/event.d
+++ b/src/core/sys/freebsd/sys/event.d
@@ -18,6 +18,7 @@ extern (C):
 nothrow:
 @nogc:
 
+import core.sys.freebsd.config;
 import core.stdc.stdint;    // intptr_t, uintptr_t
 import core.sys.posix.time; // timespec
 
@@ -38,12 +39,7 @@ enum
     EVFILT_SYSCOUNT =  11,
 }
 
-extern(D) void EV_SET(kevent_t* kevp, typeof(kevent_t.tupleof) args)
-{
-    *kevp = kevent_t(args);
-}
-
-version (FreeBSD_12)
+static if (__FreeBSD_version >= 1200000 && __FreeBSD_version < 1300000)
 {
     struct kevent_t
     {
@@ -56,7 +52,7 @@ version (FreeBSD_12)
         ulong[4]  ext;
     }
 }
-else version (FreeBSD_11)
+else static if (__FreeBSD_version < 1200000)
 {
     struct kevent_t
     {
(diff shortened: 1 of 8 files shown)


src/core/sys/freebsd/sys/event.d
```diff
@@ -43,15 +44,33 @@ extern(D) void EV_SET(kevent_t* kevp, typeof(kevent_t.tupleof) args)
     *kevp = kevent_t(args);
 }
 
-struct kevent_t
+static if (__FreeBSD_version >= 1200000 && __FreeBSD_version < 1300000)
```
kinke, line 47: I'd be more optimistic and use `__FreeBSD_version >= 1200000` alone. If FreeBSD 13 wouldn't change any relevant struct, it'd just work without having to touch druntime. FreeBSD changes would have to be tracked anyway, as it could just as easily change a struct which hasn't been versioned already.

ibuclaw: Done in the third commit.

kinke:
I like `core.sys.freebsd.config`. :+1:

The existing symvers, i.e., custom mangle strings would need to be adapted too - either by simply getting rid of them (#3270, but not restricted to `version (GNU)`) or using the according versions for `version (FreeBSD_11)` etc., in the hopes that the latter makes programs compiled on a FBSD 11 host also work on FBSD 12+.

ibuclaw:
> I like `core.sys.freebsd.config`. :+1:
> 
> The existing symvers, i.e., custom mangle strings would need to be adapted too - either by simply getting rid of them (#3270, but not restricted to `version (GNU)`) or using the according versions for `version (FreeBSD_11)` etc. (non-GDC only), in the hopes that the latter makes programs compiled on a FBSD 11 host also run on FBSD 12+.

I guess that would be for later, when dmd catches up perhaps.

kinke:
Nope - as it is here, pre-inode64 functions expecting the old structs are used with LDC too, regardless of whether `FreeBSD_12` is defined.

ibuclaw:
I can deal with that in #3270, so as not to add conflicts between the two.  Though you may have so elaborate how you want it to look like, as I guess either the versioned symbol should be explicit for all versions of FreeBSD, or latest should not have any special mangling.

kinke:
> Though you may have so elaborate how you want it to look like

I'm not really suited, the last time I've shortly played with FreeBSD was something like a decade ago. ;) - In case the regular symbols are just aliases, e.g., `fstat` resolving to `fstat@FBSD_1.0` or so when linking on FBSD 11, we probably don't need any custom mangles. IIUC, that would only allow to compile a FBSD 11 version of druntime on FBSD 12 and link against it on FBSD 12 and so enable cross-compiling a FBSD 11 version of a program on more recent versions. Not sure if there's justified interest in something like that.

ibuclaw:
@kinke - I can't see any problem from here.
```
#include <sys/mount.h>
int main()
{
  struct statfs fs;
  statfs("42", &fs);
}
```
```
$  nm a.out  | grep statfs
         U statfs@@FBSD_1.5
```
@jmdavis I guess there is zero reason to have `pragma(mangle)` if this gets merged in then?

ibuclaw:
dlang/dmd#11982 got merged into master, so now the version 12 path this PR is now enabled on the FreeBSD 12.1 CI pipeline.

ibuclaw:
@kinke - I've just gone with explicit versions for now.  Waiting for CI pipeline results.

thewilsonator (approved):

thewilsonator:
Buildkie failure is unrelated.

src/core/sys/posix/dirent.d
```diff
@@ -155,19 +157,38 @@ else version (FreeBSD)
...
-        ubyte     d_type;
-        ubyte     d_namlen;
-        char[256] d_name = 0;
+        struct dirent
+        {
+            ino_t     d_fileno;
+            off_t     d_off;
+            ushort    d_reclen;
+            ubyte     d_type;
+            ushort    d_namlen;
```
jmdavis, line 168: There is a `__uint8_t  d_pad0;` before `d_namelen` in dirent.h on FreeBSD 12. So, there should be a `ubyte d_pad0;` above `d_namelen` here. If this was copied from what I did before, then I screwed it up.

https://github.com/freebsd/freebsd/blob/release/12.2.0/sys/sys/dirent.h#L71

src/core/sys/posix/sys/stat.d
```diff
@@ -1078,36 +1078,82 @@ else version (Darwin)
...
+            uid_t     st_uid;
+            gid_t     st_gid;
+            int st_padding1;
+            dev_t     st_rdev;
+
+            version (X86) int st_atim_ext;
 
-    struct stat_t
+            time_t      st_atime;
+            c_long      __st_atimensec;
```
jmdavis, line 1101: It seems to me that these really should just all be `timespec`s, since that's what they are in the C header file, but this is indeed what we have here already for some reason. So, I guess that that can be sorted out later.

src/core/sys/posix/sys/types.d
```diff
@@ -192,14 +192,28 @@ else version (Darwin)
...
+    import core.sys.freebsd.config;
+
     // https://github.com/freebsd/freebsd/blob/master/sys/sys/_types.h
     alias long      blkcnt_t;
-    alias uint      blksize_t;
-    alias uint      dev_t;
+
+    static if (__FreeBSD_version >= 1200000)
+    {
+        alias ulong blksize_t;
```
jmdavis, line 202: This is `__int32_t` on both FreeBSD 11 and FreeBSD 12. So, the `ulong` here is wrong, and the `uint` in the FreeBSD 11 version below is wrong. It should be `int` in both cases.

https://github.com/freebsd/freebsd/blob/release/12.2.0/sys/sys/_types.h#L40
https://github.com/freebsd/freebsd/blob/release/12.2.0/sys/sys/types.h#L80

https://github.com/freebsd/freebsd/blob/release/11.4.0/sys/sys/_types.h#L38
https://github.com/freebsd/freebsd/blob/release/11.4.0/sys/sys/types.h#L78

jmdavis (changes requested):
The mistakes here are probably taken from what I did before, but there are a few things that need fixing.

src/core/sys/posix/sys/types.d
ibuclaw: This was from your PR I think. :-)

src/core/sys/posix/dirent.d
ibuclaw: Yep, the content of the structs/aliases are all a copy-pasto from your PR.

ibuclaw:
@jmdavis addressed nits.

jmdavis (approved):

PR #3275, opened 2020-11-16
CRuntime_Musl: Support v1.2.0 for 32 bits

Geod24:
As explained in the comment, `time_t` on Musl is now always 64 bits,
but used to be 32 bits on 32 bits systems.

CC @Cogitri

### DIFF
--- a/changelog/musl-32bits.dd
+++ b/changelog/musl-32bits.dd
@@ -1,0 +1,11 @@
+Support time64 changes for `CRuntime_Musl`
+
+Up to v1.1.24, Musl used a 32 bits `time_t` on 32 bits architectures.
+Since v1.2.0, `time_t` is now always 64 bits.
+From this release, druntime will also default to a 64 bits `time_t`
+on 32 bits architecture, unless the `CRuntime_Musl_Pre_Time64w` is provided.
+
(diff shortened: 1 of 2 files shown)


thewilsonator (approved):

thewilsonator:
Issue number? (or changelog entry?)

Geod24:
Good point, added release notes

PetarKirov (approved):
LGTM, feel free to merge when the CI is green.

n8sh:
>This change should only affect packagers for Musl-based systems who support 32 bits architectures (64 bits architectures already use 64 bits time_t), who now need to define CRuntime_Musl_Pre_Time64 both when building druntime / Phobos and in the default configuration, if the linked Musl is < 1.2.0.

In my opinion, changes like this should target `master` instead of `stable` going forward.

Geod24:
The issue I was faced with was released 9 months ago, and the only user of Musl I know of is Alpine Linux, which updated to it 5 months ago. It's not great to change the default in `stable`, I agree, but we needed to fix this to avoid random memory corruption,, and adding a switch and reverting it the next minor release didn't seem like a great option either.

PR #3280, opened 2020-11-18
let everybody import core.sys.windows.stat

WalterBright:
because otherwise cross-dev tools cannot be written.

### DIFF
--- a/src/core/sys/windows/stat.d
+++ b/src/core/sys/windows/stat.d
@@ -3,7 +3,6 @@
 /// Author: Walter Bright
 
 module core.sys.windows.stat;
-version (Windows):
 
 extern (C) nothrow @nogc:
 @system:


Geod24:
Wait, what ? If you are cross compiling to Windows, why wouldn't `version (Windows)` be defined ?

WalterBright:
@Geod24 on further reflection, I think you're right, this is not needed for the cross-compiler.

CyberShadow:
> but I'm not sure I follow why we need this for stat ?

One use case for some of the declarations in this module are the `S_...` constants, such as `S_IRUSR`. POSIX-like octal file modes occur on non-POSIX systems in things such as Git (which uses a subset of the POSIX format for its file lists) or network filesystems.

PR #3281, opened 2020-11-21
Fix Issue 21410 - idup of an array of immutable elements should be a no-op

CyberShadow:
Not 100% sure about this, but let's see what breaks.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3086,6 +3086,9 @@ private size_t getArrayHash(const scope TypeInfo element, const scope void* ptr,
     static assert(is(T : immutable(T)), "Cannot implicitly convert type "~T.stringof~
                   " to immutable in idup.");
 
+    static if (is(T == immutable))
+        return a;
+    else
     // wrap unsafe _dup in @trusted to preserve @safe postblit
     static if (__traits(compiles, (T b) @safe { T a = b; }))
         return _trustedDup!(T, immutable(T))(a);
@@ -3106,6 +3109,13 @@ private size_t getArrayHash(const scope TypeInfo element, const scope void* ptr,
     string s = arr.idup;
     arr[0] = '.';
     assert(s == "abc");
+}
+
+// https://issues.dlang.org/show_bug.cgi?id=21410
+@safe unittest
+{
+    string s = "foo";
+    assert(s.idup.ptr == s.ptr);
 }
 
 private U[] _trustedDup(T, U)(T[] a) @trusted
@@ -3394,7 +3404,7 @@ private void _doPostblit(T)(T[] arr)
             S3[] m;
             immutable(S3)[] i;
             static assert(!is(typeof(m.dup)));
-            static assert(!is(typeof(i.idup)));
+            static assert( is(typeof(i.idup)));
         }
         {
             shared(S1)[] m;


src/object.d
```diff
@@ -3394,7 +3404,7 @@ private void _doPostblit(T)(T[] arr)
             S3[] m;
             immutable(S3)[] i;
             static assert(!is(typeof(m.dup)));
-            static assert(!is(typeof(i.idup)));
+            static assert( is(typeof(i.idup)));
```
CyberShadow, line 3407: :warning: Changing an existing test here. Arrays of non-copyable immutable objects are now "copyable" in the sense that `.idup` succeeds. In theory, this is perfectly OK because of the transitivity of D's constness.

CyberShadow:
Needs https://github.com/dlang/phobos/pull/7701

CyberShadow:
> but let's see what breaks.

Looks good! Everything so far:

- `object.d` unit test expected that `.idup` of an array of non-copyable but immutable objects does not compile
- `std.array` unit test expected that `.idup` always reallocates
- `libmir/mir-core` is broken on master

thewilsonator (approved):

n8sh:
>`std.array` unit test expected that `.idup` always reallocates

I have also used this in the past in tests expecting `idup` to allocate unconditionally.

CyberShadow:
> I have also used this in the past in tests expecting `idup` to allocate unconditionally.

Maybe you can try this on your codebase and see the impact?

If it's a dub package, `dub run digger -- run "master + druntime#3281" -- dub test --compiler=dmd` will do it.

schveiguy:
The only place I can think of that might be troublesome is when you want to ensure stack data gets put on the heap. However, I would imagine this to be a very rare thing, because the data would have to be immutable *on the stack* for it to apply here. Which means it's a static array (or a cast from a static array). It's not inconceivable, but also not likely.

Can we have a check to see if the data is on the stack? We have a way to scan the stack, so I would think it's possible to determine if the pointer is to the stack (probably the current stack, code that passes stack data to other threads is suspect anyway).

CyberShadow:
Hmm, you're right. I haven't thought of this situation. I thought pointers to immutable data are guaranteed to always remain valid, but obviously this isn't true for things on the stack.

Playing with this for a bit, I did notice the following. Given this function:

```d
immutable(int)[] fun()
{
	immutable int[3] a = [1, 2, 3];
	return a[].idup;
}
```

With this PR, the program does compile successfully with the default compiler options. However, with `-preview=dip1000`, the compiler identifies the problem:

```
test.d(4): Error: returning idup(a[]) escapes a reference to local variable a
```

I'm honestly surprised that DMD was able to track variable lifetime across function calls, that's pretty amazing.

So, it looks like this shouldn't be a concern combined with DIP1000. Though, merging this before DIP1000 would silently break code like this, and merging this after DIP1000 would be a new breaking change on top of the DIP1000 ones. Would it perhaps make sense to enable this optimization iff DIP1000 is enabled?

RazvanN7:
@CyberShadow I think it's fine to merge this before dip1000 becomes the default. Once that happens there will be tons of places that need fixing anyway; this case is just going to be a drop in the ocean.

Is there anything else (besides a rebase) blocking this PR?

dkorpel:
This really should not be merged as this is a breaking change that can silently introduce memory corruption into existing code. `idup` is specified to create a new dynamic array, so it's valid to input an array of limited lifetime that's stack-allocated / `malloc` allocated / memory mapped / whatever, and expect a GC allocated array with infinite lifetime as output. dip1000 is irrelevant here, it will turn some of the runtime breakage into compile time errors (which is good!), but doesn't change anything about the nature of the change.

Going back to issue 21410:
> The motivation is generic code such as the following:
> ```D
>int[string] dict;
>void save(C : char)(C[] key, int value)
>{
>	if (auto p = key in dict)
>		*p = value;
>	else
>		dict[key.idup] = value;
>}
>```

The solution here is to use something else than `idup`. Maybe `.text`, `.to!(immutable(C)[])` or a new symbol, but not `idup`.

RazvanN7:
@CyberShadow In light of @dkorpel 's comment I am tempted to close this (and the associated issue as WONTFIX). Is that ok with you?

CyberShadow:
Yes, let's close this. 

(Instead we could improve the documentation to make it extra clear that `idup` always allocates, even if the input data is immutable. However that change would be in the dlang.org repo.)

PR #3284, opened 2020-11-23
core.stdcpp.new_.cpp_delete unnecessarily requires destruction to be @nogc

n8sh:


src/core/stdcpp/new_.d
```diff
@@ -14,8 +14,6 @@ module core.stdcpp.new_;
 import core.stdcpp.xutility : __cpp_sized_deallocation, __cpp_aligned_new;
 import core.stdcpp.exception : exception;
 
-@nogc:
```
n8sh, line 17: Every function between here and where `@nogc:` is re-added is either a template or explicitly annotated `@nogc`.

TurkeyMan (approved):
Great catch!

TurkeyMan:
Is the parallel for construction also addressed here?
Also, are we confident that `emplace` and `destroy` will appropriately infer `@nogc` if the types they are interacting with have `@nogc` ctor/dtors?

TurkeyMan:
I'd like to see a unit test that demonstrates use with GC and also `@nogc` types so that we can prove inference is working properly.

n8sh:
>Great catch!

Thanks but this wasn't a "catch" so much as something stumbled across naturally while trying to mix C++ and D!

TurkeyMan:
Same thing. It's great that you're using this.
I'd love to get some experience reports from you when you've has more time with it.

test/stdcpp/src/new_test.d
```diff
@@ -32,3 +32,43 @@ unittest
     s.c = cast(MyStruct*)__cpp_new(MyStruct.sizeof);
     cpp_delete(s);
 }
+
+extern(C++) class MyClassNoGC
+{
+    __gshared ulong numDeleted;
+    int x = 3;
+    ~this() @nogc { ++numDeleted; }
```
TurkeyMan, line 40: I feel like this class should test @nogc constructors too.

test/stdcpp/src/new_test.d
```diff
@@ -32,3 +32,44 @@ unittest
...
+    assert(MyClassNoGC.numDeleted == 0);
+    cpp_delete(c1);
+    assert(MyClassNoGC.numDeleted == 1);
+}
+
+extern(C++) class MyClassGC
+{
+    __gshared ulong numDeleted;
+    int x = 5;
+    ~this() { if (++numDeleted < 0) throw new Exception("overflow"); }
```
TurkeyMan, line 60: No with-GC constructor test?

TurkeyMan: It should probably `new` something in the constructor...

test/stdcpp/src/new_test.d
```diff
@@ -32,3 +32,57 @@ unittest
...
+
+@nogc unittest
+{
+    // Test cpp_new and cpp_delete for a class infer @nogc.
+    import core.stdcpp.new_: cpp_new, cpp_delete;
+
+    MyClassNoGC c1 = cpp_new!MyClassNoGC(4);
+    assert(c1.x == 4);
+    assert(MyClassNoGC.numDeleted == 0);
+    cpp_delete(c1);
```
TurkeyMan, line 52: Seems this doesn't infer `@nogc`. I guess `destroy` is not inferring correctly?

test/stdcpp/src/new_test.d
```diff
@@ -32,3 +32,64 @@ unittest
...
+    this(int x) @nogc { this.x = x; }
+    ~this() @nogc { ++numDeleted; }
+}
+
+@nogc unittest
+{
+    // TEST NOT COMPILING: compiler can't tell dtor of MyClassNoGC is @nogc.
+    if (0)
+        MyClassNoGC.init.__xdtor();
+}
```
n8sh, line 49: The problem wasn't in `destroy` it's here:

```console
test\stdcpp\src\new_test.d(48): Error: @nogc function new_test.__unittest_L44_C7 cannot call non-@nogc destructor new_test.MyClassNoGC.~this
```

TurkeyMan: `__xdtor` is not the destructor; that's actually a function synthesised by the compiler.
Maybe there's a compiler bug where that function should be inferring some attributes but it's not?
It will also execute all the field destructors, and the super destructor, so if any of those are not `@nogc`, then that may be breaking the inference...?

n8sh:
Updated tests to test `@nogc` GC inference for struct instead of class. Making `cpp_delete` infer correct attributes for classes is left as an item for a future PR. As prior to this PR `cpp_delete` could not be used with classes at all there is no regression.

PR #3286, opened 2020-11-24
[dmd-cxx] Update bindings for FreeBSD 12 support

ibuclaw:


Geod24:
Looks like the failure is real ?

ibuclaw:
> Looks like the failure is real ?

Yeah, old windows make scripts used to work this way.

PR #3300, opened 2020-12-02
Add libint support

adelavais:
Hello! I was working on adding internationalization to a program and I observed that locale.h exists in core.stdc, but libintl.h doesn't. I think introducing this header would help D, as functions from locale.h and libintl.h are often used together, and it would bring internationalization into D. 
I am aware that this PR is technically a WIP, but I need help in continuing and feedback if this is something worth pursuing in D.

### DIFF
--- a/src/core/stdc/libintl.d
+++ b/src/core/stdc/libintl.d
@@ -1,0 +1,20 @@
+/**
+ * D header file for C99.
+ */
+
+
+module core.stdc.libint;
+
+char *gettext (scope const char *__msgid);
+char *dgettext (scope const char *__domainname, scope const char *__msgid);
+char *__dgettext (scope const char *__domainname, scope const char *__msgid);
+char *dcgettext (scope const char *__domainname, scope const char *__msgid, int __category);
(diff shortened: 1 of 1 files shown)


pbackus:
The `core.stdc` package is for standard C headers [1]. `locale.h` is a standard C header, but `libint.h` is not—it is part of GNU gettext. The appropriate place for this is a package on code.dlang.org, not druntime.

[1] http://port70.net/~nsz/c/c99/n1256.html#7.1.2

12345swordy:
@WalterBright what do you think?

Geod24 (changes requested):
I agree with @pbackus this should be a dub package. The only need that this could have in druntime was if we were using it. druntime itself is not a binding library.

Code wise, there's two issues:
1) Attributes are missing. This should be `@system nothrow @nogc`.
2) There's some `long int` here. For C long type, you should use `c_long`, as it's 4 bytes in 32 bits (unlike D's `long`).

adelavais:
Thank you so much for your feedback! I made a dub package instead (https://code.dlang.org/packages/libintl) and I will close this PR.

PR #3303, opened 2020-12-04
Fix documentation of cas

thaven:
I noticed a mistake in the documenttion on the `cas` function in `core.atomic`. It wrongly states that the second category of overloads expect the `writeThis` to be a pointer, this should be `ifThis`. I made some additional clarifications as well.

Specifically, I removed the phrase about the return value as it does not apply exclusively to the first category. It made no sense to add it somewhere else, as the return value is already documented correctly as well.


TurkeyMan (approved):
Thanks for the fix!

PR #3304, opened 2020-12-07
Separate string-formatting of backtrace into `Location.toString` and add documentation

Geod24:
The end goal is to add support for libunwind as a provider of backtrace.
In order to do so, quite a bit of refactoring is needed, as the (public) interfaces don't really allow this and are tied to `backtrace` at the moment.
Ideally, we could also expose `Location` to user code so that they can inspect it directly.


Geod24:
Ergh, the length of the backtrace depends on the address size, which is platform dependent... I'm inclined to just simplify this and not print the address, instead printing as much of the symbol as possible. Thoughts ?

thewilsonator (approved):

Geod24:
Fixed

Geod24:
Ping @thewilsonator

PR #3306, opened 2020-12-08
Cirrus, Github CI: Use HOST_DMD instead of HOST_DC

Geod24:
This currently trigger a (small) deprecation message.
The rationale is that HOST_DC should be any D compiler,
while HOST_DMD expects a DMD-like compiler.


Geod24:
Blocked by https://github.com/dlang/dmd/pull/12021

thewilsonator (approved):

PR #3307, opened 2020-12-08
Darwin: more fixes for ucontext_t implementation of core.thread.Fiber

ibuclaw:
It turns out I was wrong about the size, and there is a data field that brings the sizeof ucontext_t to some 1088 bytes, not 56. :-)

This isn't headers copied ad verbatim, rather I've collapsed all fields into static arrays where it made sense to do so.

On testing core.thread, I came across an issue specific to Darwin makecontext() implementation where the ucontext_t object [isn't set-up](https://github.com/Apple-FOSS-Mirror/Libc/blob/2ca2ae74647714acfc18674c3114b1a5d3325d7d/x86_64/gen/makecontext.c#L98-L107) if the stack size is smaller than MINSIGSTKSZ.

Although I don't see this mentioned anywhere in the ucontext documentation, I just went with applying this to all targets that use the fallback ucontext_t implementation, on the assumption that ucontext preserving the current signal mask is implied if you are relying on the fallback Fiber implementation.  (And FYI, Linux sets this to 2048, so we always easily exceed the minimum by default).

This brings darwin support for druntime between 10.6 and 10.12 to pass cleanly with gdc, and PPC is looking good too.  Though admittedly this is a stop-gap before asm implementations of Fiber are done properly.

### DIFF
--- a/src/core/sys/posix/ucontext.d
+++ b/src/core/sys/posix/ucontext.d
@@ -33,6 +33,8 @@ else version (TVOS)
 else version (WatchOS)
     version = Darwin;
 
+version (ARM)     version = ARM_Any;
+version (AArch64) version = ARM_Any;
 version (MIPS32)  version = MIPS_Any;
 version (MIPS64)  version = MIPS_Any;
 version (PPC)     version = PPC_Any;
@@ -921,16 +923,66 @@ else version (CRuntime_Musl)
 }
 else version (Darwin)
 {
-    alias mcontext_t = void;
+    private
+    {
+        version (X86_64)
+        {
+            struct __darwin_mcontext
+            {
+                ulong[89] __opaque;
+            }
+            static assert(__darwin_mcontext.sizeof == 712);
+        }
+        else version (X86)
+        {
+            struct __darwin_mcontext
+            {
+                uint[150] __opaque;
+            }
+            static assert(__darwin_mcontext.sizeof == 600);
+        }
+        else version (AArch64)
+        {
+            struct __darwin_mcontext
+            {
+                align(16) ulong[102] __opaque;
+            }
+            static assert(__darwin_mcontext.sizeof == 816);
+        }
+        else version (ARM)
+        {
+            struct __darwin_mcontext
+            {
+                uint[85] __opaque;
+            }
+            static assert(__darwin_mcontext.sizeof == 340);
+        }
+        else version (PPC_Any)
+        {
+            struct __darwin_mcontext
+            {
+                version (PPC64)
+                    ulong[129] __opaque;
+                else
+                    uint[258] __opaque;
+            }
+            static assert(__darwin_mcontext.sizeof == 1032);
+        }
+        else
+            static assert(false, "mcontext_t unimplemented for this platform.");
+    }
+
+    alias mcontext_t = __darwin_mcontext*;
 
     struct ucontext
     {
-        int         uc_onstack;
-        sigset_t    uc_sigmask;
-        stack_t     uc_stack;
-        ucontext*   uc_link;
-        size_t      uc_mcsize;
-        mcontext_t* uc_mcontext;
+        int                uc_onstack;
+        sigset_t           uc_sigmask;
+        stack_t            uc_stack;
+        ucontext*          uc_link;
+        size_t             uc_mcsize;
+        __darwin_mcontext* uc_mcontext;
+        __darwin_mcontext  __mcontext_data;
     }
 
     alias ucontext_t = ucontext;
--- a/src/core/thread/fiber.d
+++ b/src/core/thread/fiber.d
@@ -958,6 +958,13 @@ private:
         {
             version (Posix) import core.sys.posix.sys.mman; // mmap, MAP_ANON
 
+            static if ( __traits( compiles, ucontext_t ) )
+            {
+                // Stack size must be at least the minimum allowable by the OS.
+                if (sz < MINSIGSTKSZ)
+                    sz = MINSIGSTKSZ;
+            }
+
             static if ( __traits( compiles, mmap ) )
             {
                 // Allocate more for the memory guard


src/core/sys/posix/ucontext.d
```diff
@@ -921,19 +921,125 @@ else version (CRuntime_Musl)
...
+                __darwin_exception_state __es;
+                __darwin_thread_state __ss;
+                __darwin_float_state __fs;
+                __darwin_vector_state __vs;
+            }
+        }
     }
 
-    alias ucontext_t = ucontext;
+    static if (__traits(compiles, __darwin_mcontext))
```
ibuclaw, line 1026: This, because no other targets have ucontext.h other than x86 and PPC (it got deprecated before ARM was introduced).

src/core/thread/fiber.d
```diff
@@ -958,6 +958,13 @@ private:
         {
             version (Posix) import core.sys.posix.sys.mman; // mmap, MAP_ANON
 
+            static if ( __traits( compiles, ucontext_t ) )
+            {
+                // Stack size must be at least the minimum allowable by the OS.
+                if (sz < MINSIGSTKSZ)
+                    sz = MINSIGSTKSZ;
```
ibuclaw, line 965: I may come back to this if Darwin still craps out with a smaller stack using the asm Fibers.

src/core/sys/posix/ucontext.d
jacob-carlborg: As far as I can see, the functions are deprecated, but not the types [1].

[1] https://github.com/apple/darwin-xnu/blob/a449c6a3b8014d9406c2ddbdc81795da24aa7443/bsd/arm/_mcontext.h#L29

src/core/sys/posix/ucontext.d
```diff
@@ -921,19 +921,125 @@ else version (CRuntime_Musl)
...
+
+            struct __darwin_xmm_reg
+            {
+                char[16] __xmm_reg;
+            }
+
+            struct __darwin_exception_state
+            {
+                uint __trapno;
+                uint __err;
```
jacob-carlborg, line 942: You're missing a field here [1].

[1] https://github.com/apple/darwin-xnu/blob/0a798f6738bc1db01281fc08ae024145e84df927/osfmk/mach/i386/_structs.h#L1153

src/core/sys/posix/ucontext.d
```diff
@@ -921,19 +921,125 @@ else version (CRuntime_Musl)
...
+                else
+                    uint[16] __regs;
+            }
+
+            struct __darwin_float_state
+            {
+                int[2] __fpu_reserved;
+                ushort __fpu_fcw;
+                ushort __fpu_fsw;
+                uint[7] __fpu_regs;
```
jacob-carlborg, line 962: It's difficult to review this when you're not following the headers.

ibuclaw: Check the sizeof(ucontext_t) from a C program.

src/core/sys/posix/ucontext.d
ibuclaw: That doesn't exist on my system [1].

[1] https://github.com/apple/darwin-xnu/blob/b472f0612b8556cd1c6eb1c285ec1953de759e35/osfmk/mach/i386/_structs.h#L751-L753

ibuclaw: (Though I'll note that this is inconsequential as thankfully the size doesn't change between older and newer OS X versions).

src/core/sys/posix/ucontext.d
ibuclaw: They really don't make finding these definitions easy. I couldn't see it on opensource.apple.com :-)

src/core/sys/posix/ucontext.d
jacob-carlborg: I think you either should follow the headers or just do something like `ubyte[12] __hidden` to get the correct size.

src/core/sys/posix/ucontext.d
jacob-carlborg: Yeah, you need to know where to look. GitHub makes that easier.

You really need to have these files on your system and the latest version of the SDK. It will make it much easier. You can create a free Apple ID and download Xcode here: https://developer.apple.com/download/more.

src/core/sys/posix/ucontext.d
jacob-carlborg: Yeah, I did. Again, I don't understand why you're doing it like this. Either follow the headers exactly or make the fields opaque.

src/core/sys/posix/ucontext.d
ibuclaw: But that ignores the alignment. Packing repeat fields isn't all that much bad.

src/core/sys/posix/ucontext.d
ibuclaw: It would be 150 lines longer if I listed every single field.

src/core/thread/fiber.d
ibuclaw: On a related note, @Geod24 are you sorting out the fiber stack overflow test you have a PR for? The minimum stack size should be at least this to trigger overflow on ucontext targets.

ibuclaw: On a related note, @Geod24 are you sorting out the fiber stack overflow test you have a PR for? The minimum stack size should be at least this to trigger overflow on ucontext targets.

ibuclaw:
@jacob-carlborg all fine here?

src/core/sys/posix/ucontext.d
```diff
@@ -33,6 +33,8 @@ else version (TVOS)
 else version (WatchOS)
     version = Darwin;
 
+version (ARM)     version = ARM_Any;
+version (AArch64) version = ARM_Any;
```
jacob-carlborg, line 37: `ARM_Any ` seems not to be used anymore.

jacob-carlborg:
I don't have a system to verify the PPC code, but the rest looks good.

jacob-carlborg (approved):

PR #3308, opened 2020-12-09
Fix 21465: Allocate _tlsRanges in C heap

omerfirmak:
Similar to the issue fixed here https://github.com/dlang/druntime/pull/1655#issuecomment-279223238 ,
static version of the sections_elf_shared accesses TLS of a dead thread. The simplest fix is to allocate
_tlsRanges somewhere that would persist outside of the thread scope, for example the C heap.

This was encountered while porting ldc 1.24.0 to Alpine Linux which uses Musl.
https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/14364 But this issue should
be affecting other libc implementations as well.

Fixes https://issues.dlang.org/show_bug.cgi?id=21465


ibuclaw (approved):
Seems reasonable, did you encounter the same on gdc as well @Geod24 ?

src/rt/sections_elf_shared.d
```diff
@@ -369,7 +370,13 @@ else
      * Thread local array that contains TLS memory ranges for each
      * library initialized in this thread.
      */
-    @property ref Array!(void[]) _tlsRanges() @nogc nothrow { static Array!(void[]) x; return x; }
+    @property ref Array!(void[]) _tlsRanges() @nogc nothrow {
+        static Array!(void[])* x = null;
+        if (x is null)
+            x = cast(Array!(void[])*).calloc(1, Array!(void[]).sizeof);
+        assert(x);
```
kinke, line 377: I think this is probably too soon to throw Errors, and [safeAssert](https://github.com/dlang/druntime/blob/master/src/rt/util/utility.d#L21) a better fit.

omerfirmak: Only alternative is to crash or error in `initTLSRanges`. Making the change to `safeAssert`

PR #3310, opened 2020-12-11
Move rt.backtrace to core.internal.backtrace

Geod24:
This simplifies code and adopt the new pattern for internal modules.


src/core/internal/qsort.d
```diff
@@ -6,17 +6,11 @@
  * License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
  * Authors:   Walter Bright, Martin Nowak
  */
-
-/*          Copyright Digital Mars 2000 - 2010.
- * Distributed under the Boost Software License, Version 1.0.
- *    (See accompanying file LICENSE_1_0.txt or copy at
- *          http://www.boost.org/LICENSE_1_0.txt)
- */
```
Geod24, line 14: Deleted because the same copyright is present in the header

PetarKirov (approved):
Thanks! This will pave the way for further improvements in the future, like exposing a `Location` base backtrace interface. Looking forward to that!

Not to mention that it's nice finally having some `@nogc` containers that could be used for template-based runtime hook implementations as a side effect.

PR #3313, opened 2020-12-14
CRuntime_Musl: More fixes for time64

Geod24:
```
The original PR (#3275) missed quite a few spots and conversions,
which led to the build on Alpine Linux failing with Aithmetic Exception
on core.time module constructor.
Links to the two offending commits are included.
For further issues / investigation, search for 'time64' in the git repository.
```

CC @Cogitri , I'll rebuild my GDC and see if that solve LDC v1.24's issues.

### DIFF
--- a/src/core/sys/posix/signal.d
+++ b/src/core/sys/posix/signal.d
@@ -3427,7 +3427,28 @@ struct timespec
 }
 */
 
-version (linux)
+version (CRuntime_Musl)
+{
+    // Musl on 32 bits use 64 bits time_t (time64)
+    // See https://git.musl-libc.org/cgit/musl/commit/?id=9b2921bea1d5017832e1b45d1fd64220047a9802
+    struct timespec
+    {
+        time_t tv_sec;
+        // 32 bits of padding on 32 bits, or in C:
+        // int :8*(sizeof(time_t)-sizeof(long))*(__BYTE_ORDER==4321);
+        version (BigEndian)
+            static if (time_t.sizeof > c_long.sizeof)
+                int __padding;
+        c_long tv_nsec;
+        // Another 32 bits of padding on 32 bits:
+        // int :8*(sizeof(time_t)-sizeof(long))*(__BYTE_ORDER!=4321);
+        version (LittleEndian)
+            static if (time_t.sizeof > c_long.sizeof)
+                int __padding;
+    };
+
+}
+else version (linux)
 {
     struct timespec
     {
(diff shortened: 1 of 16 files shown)


PetarKirov:
Does tsunami on BuildKite treat deprecations as errors? Otherwise I can't tell why it fails from the logs.

Geod24:
> Exited with status -1 (process killed or agent lost; see the timeline tab for more information)

It's spurious. We're coming back to this and getting it tested, stay tuned.

src/core/sys/posix/signal.d
```diff
@@ -3427,7 +3427,28 @@ struct timespec
...
+        // int :8*(sizeof(time_t)-sizeof(long))*(__BYTE_ORDER==4321);
+        version (BigEndian)
+            static if (time_t.sizeof > c_long.sizeof)
+                int __padding;
+        c_long tv_nsec;
+        // Another 32 bits of padding on 32 bits:
+        // int :8*(sizeof(time_t)-sizeof(long))*(__BYTE_ORDER!=4321);
+        version (LittleEndian)
+            static if (time_t.sizeof > c_long.sizeof)
+                int __padding;
```
PetarKirov, line 3447: You can make the `__padding` as private for extra protection that is not available in C ;)

PetarKirov (approved):

src/core/sys/posix/sys/stat.d
```diff
@@ -1745,10 +1745,26 @@ else version (CRuntime_Musl)
...
+            // Time64 on 32 bits
+            // See https://git.musl-libc.org/cgit/musl/commit/?id=38143339646a4ccce8afe298c34467767c899f51
+            static if (time_t.sizeof > c_long.sizeof)
+            {
+                private struct old_timespec {
+                    long tv_sec;
+                    long tv_nsec;
+                }
+
+                old_timespec __st_atim32;
```
omerfirmak, line 1757: Shouldnt these be present for all 32-bit archs?

Also 
```
                timespec st_atim;
                timespec st_mtim;
                timespec st_ctim; 
```
These should not be in the else case, they are always needed.

Geod24: Yes

src/core/sys/posix/signal.d
Geod24: True, but no other place in druntime do this.

RazvanN7:
Is it ok to close this in favor of #3383 ?

PR #3314, opened 2020-12-16
Fix 21484: Infinite loop in core.memory : GC.{get,set,clr}Attr(const scope void*...)

Geod24:
Those functions never worked, and would cause an infinite recursion.
In a non-optimized build that would exhaust the stack and trigger a SEGV,
however in the optimized build we distribute, it results in an infinite
loop as tail call optimization is performed.

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -408,7 +408,7 @@ struct GC
      */
     static uint getAttr( const scope void* p ) nothrow
     {
-        return getAttr(cast()p);
+        return gc_getAttr(cast(void*) p);
     }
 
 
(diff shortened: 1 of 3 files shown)


PetarKirov (approved):

MoonlightSentinel:
Tests?

Geod24:
Usually I would say "it doesn't make sense to test for infinite loop here", but since the function were not covered at all, I added a small test.

MoonlightSentinel:
Thanks

Geod24:
So, anything holding this back ?

ibuclaw (approved):
Sgtm

thewilsonator (approved):

PR #3318, opened 2020-12-26
Fix 21472 (1/2) - Add tuple formatting to _d_assert_fail`

MoonlightSentinel:
This commits add another overload of `_d_assert_fail` that is capable
of formatting arbitrary value tuples.

This enables `-checkaction=context` to work with `tupleof` and other
tuples used inside an `assert`.

CC @Geod24

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -47,27 +47,50 @@ string _d_assert_fail(string comp, A, B)(auto ref const scope A a, auto ref cons
  * Returns:
  *   A string such as "$a != true" or "$a == true".
  */
+string _d_assert_fail2(A)(string op, auto ref const scope A a)
+{
+    string[2] vals = [ miniFormatFakeAttributes(a), "true" ];
+    immutable token = op == "!" ? "==" : "!=";
+    return combine(vals[0 .. 1], token, vals[1 .. $]);
+}
+
+/// Ditto
 string _d_assert_fail(A)(const scope string op, auto ref const scope A a)
 {
-    string val = miniFormatFakeAttributes(a);
+    string[2] vals = [ miniFormatFakeAttributes(a), "true" ];
     immutable token = op == "!" ? "==" : "!=";
-    return combine(val, token, "true");
+    return combine(vals[0 .. 1], token, vals[1 .. $]);
 }
 
 /**
  * Generates rich assert error messages for binary expressions
  *
  * The binary expression `assert(x == y)` will be turned into
- * `assert(x == y, _d_assert_fail("==", x, y))`.
+ * `assert(x == y, _d_assert_fail!(typeof(x))("==", x, y))`.
  *
  * Params:
  *   comp = Comparison operator that was used in the expression.
- *   a  = Left hand side operand.
- *   b  = Right hand side operand.
+ *   a  = Left hand side operand (can be a tuple).
+ *   b  = Right hand side operand (can be a tuple).
  *
  * Returns:
  *   A string such as "$a $comp $b".
  */
+template _d_assert_fail2(A...) {
+    string _d_assert_fail2(B...)(
+        string comp, auto ref const scope A a, auto ref const scope B b)
+    {
+        string[A.length + B.length] vals;
+        static foreach (idx; 0 .. A.length)
+            vals[idx] = miniFormatFakeAttributes(a[idx]);
+        static foreach (idx; 0 .. B.length)
+            vals[A.length + idx] = miniFormatFakeAttributes(b[idx]);
+        immutable token = invertCompToken(comp);
+        return combine(vals[0 .. A.length], token, vals[A.length .. $]);
+    }
+}
+
+/// Ditto
 string _d_assert_fail(A, B)(const scope string comp, auto ref const scope A a, auto ref const scope B b)
 {
     /*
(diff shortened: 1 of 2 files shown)


Geod24:
I was working on something similar in the past few days, and took a very similar approach.
Key differences:
- We should make the `comp` argument runtime, there's really no reason to have it be a template parameter;
- I used the exact two-level templates with arguments pack: however, I don't see the need to have two overloads: just name it `_d_assert_fail2`, merge it in druntime once you have confirmed it works with DMD, then switch DMD to it, and delete the old one, and voila;

MoonlightSentinel:
The operator is now a runtime argument and renamed the hook to `_d_assert_fail_tuple` (to make it more expressive).

Geod24:
Hum, for information this was my approach: https://github.com/Geod24/druntime/commit/ffde1a17c5eac3b9040fe843c102ac84a89d1739

From a quick look, it seems quite simpler to just change `combine` to support tuples.

Geod24:
Ping @MoonlightSentinel : Wouldn't something akin to https://github.com/Geod24/druntime/commit/ffde1a17c5eac3b9040fe843c102ac84a89d1739 make more sense, by only having one hook (and use explicit instantiation on the compiler side), and require less intrusive change ?

MoonlightSentinel:
Maybe, there were some problems w.r.t. to overload resolution when using only one hook. Not sure if that's still an issue, I'll give it another try at the weekend.

However, what bothers me about Geod24@ffde1a1 is that it completely omits the parentheses for tuples (but that can be easily changed using another parameter `bool printBraces`).

RazvanN7:
@MoonlightSentinel what is the status on this?

RazvanN7 (approved):

Geod24:
Ping @MoonlightSentinel , did you have a chance to take a look at it ? If not, I can give it a try on my end.

MoonlightSentinel:
Got it to work, see MoonlightSentinel/dmd@9ac88a12487202f3d1be24db57eb07cf332aa046 and MoonlightSentinel/druntime@7628dfc57d164496cd8e09b51f02c3666c0f3df8.

The new hook doesn't need special casing for tuples in dmd but might be more expensive for binary expressions without tuples (the common case).

Geod24:
> The new hook doesn't need special casing for tuples in dmd but might be more expensive for binary expressions without tuples (the common case).

Why do you think it'd be more expensive ?

> Got it to work, see MoonlightSentinel/dmd@9ac88a1 and MoonlightSentinel/druntime@7628dfc.

Great, but I do think we have to merge the `2` overloads first, otherwise the CI will be broken, won't it ?

Is this still a draft ?

MoonlightSentinel:
> Why do you think it'd be more expensive ?

Mostly because it's using `static foreach`, but this concern is probably unfounded.
 
> Great, but I do think we have to merge the `2` overloads first, otherwise the CI will be broken, won't it ?
> 
> Is this still a draft ?

Re-added the old hook and it still works with the old and new lowering.

MoonlightSentinel:
Ping @Geod24

src/core/internal/dassert.d
```diff
@@ -80,23 +85,48 @@ string _d_assert_fail(A, B)(const scope string comp, auto ref const scope A a, a
     string valA = miniFormatFakeAttributes(a);
     string valB = miniFormatFakeAttributes(b);
     immutable token = invertCompToken(comp);
-    return combine(valA, token, valB);
+    return combine([valA], token, [valB]);
```
Geod24, line 88:
```suggestion
    return combine(&valA[0 .. 1], token, &valB[0 .. 1]);
```

src/core/internal/dassert.d
```diff
@@ -80,23 +85,48 @@ string _d_assert_fail(A, B)(const scope string comp, auto ref const scope A a, a
...
 {
-    const totalLen = valA.length + token.length + valB.length + 2;
+    // Each separator is 2 chars (", "), plus the two spaces around the token.
+    size_t totalLen = (valA.length - 1) * 2 +
+        (valB.length - 1) * 2 + 2 + token.length;
+    foreach (v; valA) totalLen += v.length;
+    foreach (v; valB) totalLen += v.length;
+
+    // Include braces when printing tuples
+    const printBraces = 2 < (valA.length + valB.length);
```
Geod24, line 102:
```suggestion
    const printBraces = (valA.length + valB.length) > 2;
```

Subjective, but Yoda conditions aren't really needed in D.

src/core/internal/dassert.d
```diff
@@ -80,23 +85,48 @@ string _d_assert_fail(A, B)(const scope string comp, auto ref const scope A a, a
...
+    if (printBraces) totalLen += 4; // '(', ')' for both tuples
+
     char[] buffer = cast(char[]) pureAlloc(totalLen)[0 .. totalLen];
     // @nogc-concat of "<valA> <comp> <valB>"
-    auto n = valA.length;
-    buffer[0 .. n] = valA;
+    static void formatTuple (scope char[] buffer, ref size_t n, in string[] vals, const bool printBraces)
```
Geod24, line 107:
```suggestion
    static void formatTuple (scope char[] buffer, ref size_t n, in string[] vals, in bool printBraces)
```

If we use `in string[]`, might as well use `in bool`.

src/core/internal/dassert.d
```diff
@@ -371,7 +402,7 @@ private string invertCompToken(string comp) pure nothrow @nogc @safe
         case "!in":
             return "in";
         default:
-            assert(0, combine("Invalid comparison operator", "-", comp));
+            assert(0, combine(["Invalid comparison operator"], "-", [comp]));
```
Geod24, line 405:
```suggestion
            static immutable msg = [ "Invalid comparison operator" ];
            assert(0, combine(msg, "-", &comp[0 .. 1]));
```

src/core/internal/dassert.d
```diff
@@ -165,7 +195,8 @@ private string miniFormat(V)(const scope ref V v)
 
         // Format invalid enum values as their base type
         enum cast_ = "cast(" ~ V.stringof ~ ")";
-        return combine(cast_, "", miniFormat(*(cast(BaseType*) &v)));
+        const val = miniFormat(*(cast(BaseType*) &v));
+        return combine([ cast_ ], "", [ val ]);
```
Geod24, line 199:
```suggestion
        return combine(&cast_[0 .. 1], "", &val[0 .. 1]);
```

Geod24 (approved):
LGTM, feel free to merge after addressing suggestions

src/core/internal/dassert.d
MoonlightSentinel: Why? 
The current form is already `@nogc` due to the `scope` parameter.

Geod24: Ah, nvm then

PR #3319, opened 2020-12-27
Pass operator as a runtime parameter to _d_assert_fail

MoonlightSentinel:
The operator is not required at compile time and can hence be passed as a runtime parameter to create less unique template instations.

CC @Geod24

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -19,11 +19,23 @@
  */
 module core.internal.dassert;
 
+// Legacy hooks currently used by dmd, remove once dmd uses
+// the new runtime versions defined below
+string _d_assert_fail(string op, A)(auto ref const scope A a)
+{
+    return _d_assert_fail!(A)(op, a);
+}
+
+string _d_assert_fail(string comp, A, B)(auto ref const scope A a, auto ref const scope B b)
+{
+    return _d_assert_fail!(A, B)(comp, a, b);
+}
+
 /**
  * Generates rich assert error messages for unary expressions
  *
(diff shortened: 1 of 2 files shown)


src/core/internal/dassert.d
```diff
@@ -35,18 +47,18 @@ module core.internal.dassert;
  * Returns:
  *   A string such as "$a != true" or "$a == true".
  */
-string _d_assert_fail(string op, A)(auto ref const scope A a)
+string _d_assert_fail(A)(const scope string op, auto ref const scope A a)
 {
     string val = miniFormatFakeAttributes(a);
-    enum token = op == "!" ? "==" : "!=";
+    scope token = op == "!" ? "==" : "!=";
```
Geod24, line 53:
```suggestion
    immutable token = op == "!" ? "==" : "!=";
```

src/core/internal/dassert.d
```diff
@@ -319,9 +331,9 @@ private string miniFormat(V)(const scope ref V v)
 import core.atomic : atomicLoad;
 
 // Inverts a comparison token for use in _d_assert_fail
-private string invertCompToken(string comp)
+private string invertCompToken(string comp) pure nothrow @nogc @safe
 {
-    switch (comp)
+    final switch (comp)
```
Geod24, line 336: I don't see how it can be `final` though ?

MoonlightSentinel: This was initially done to avoid the `@gc` concatenation in the `assert` while still providing a better error.
Now using `combine`  to format the error message.

Geod24 (approved):

PR #3321, opened 2021-01-02
Add experimental support for libunwind-based stack trace printing

Geod24:
Each individual commit gradually refactor things, the second-to-last commit adds simple bindings, and the last commit adds actual support. I had a hard time getting this to a point where there is absolutely no breaking change, but after months of work, it's finally there.

The code left in `dwarf.d` is not pretty, but is surely much better than before. The main issue is that when we run the state machine, we need to iterate on all `Location`s to fill them, but when the user iterate, they iterate one `Location` at a time.
This forces us to keep an array, as re-running the state machine for every iteration would lead to insufferable performance issues (not actually benchmarked).

A proper solution would probably be to extend the debug infos emitted so we can do lookups more efficiently, but that's an even bigger project.

Note that ultimately I'd like to give users the ability to inspect their stack trace. Libunwind makes this very easy, but we're currently blocked by the fact that adding method to the `Throwable.TraceInfo` interface is a breaking change.

Marking this as draft as I still need to handle the segv handler, but I think that's a minor / side concern and the rest is properly documented and finished, and can be reviewed. I'm also going to submit the refactorings in a separate PR which will link to this so reviewers can see the point of such a refactoring.

Another step I'd like to take is to pass down the number of frames to skip, as currently it's hardcoded to druntime's usage, and using the handler directly will skip 3-4 relevant frames.

Depends on: https://github.com/dlang/druntime/pull/3322

### DIFF
--- a/src/core/internal/backtrace/handler.d
+++ b/src/core/internal/backtrace/handler.d
@@ -1,0 +1,117 @@
+/**
+ * Libunwind-based implementation of `TraceInfo`
+ *
+ * This module exposes an handler that uses libunwind to print stack traces.
+ * It is used when druntime is packaged with `DRuntime_Use_Libunwind` or when
+ * the user uses the following in `main`:
+ * ---
+ * import core.runtime;
+ * import core.internal.backtrace.handler;
+ * Runtime.traceHandler = &libunwindDefaultTraceHandler;
+ * ---
+ *
+ * Authors: Mathias 'Geod24' Lang
+ * Copyright: D Language Foundation - 2020
+ * See_Also: https://www.nongnu.org/libunwind/man/libunwind(3).html
+ */
+module core.internal.backtrace.handler;
+
+version (DRuntime_Use_Libunwind):
+
+import core.internal.backtrace.dwarf;
+import core.internal.backtrace.libunwind;
+
+/// Ditto
+class LibunwindHandler : Throwable.TraceInfo
+{
+    private static struct FrameInfo
+    {
+        char[1024] buff = void;
+
+        const(char)[] name;
+        const(void)* address;
+    }
+
+    size_t numframes;
+    enum MAXFRAMES = 128;
+    FrameInfo[MAXFRAMES] callstack = void;
+
+    /**
+     * Create a new instance of this trace handler saving the current context
+     *
+     * Params:
+     *   frames_to_skip = The number of frames leading to this one.
+     *                    Defaults to 5. Should normally default to 1,
+     *                    but since this information is not currently
+     *                    propagated in druntime, the default is right
+     *                    for druntime
+     */
+    public this (size_t frames_to_skip = 5) nothrow @nogc
+    {
+        import core.stdc.string : strlen;
+
+        static assert(typeof(FrameInfo.address).sizeof == unw_word_t.sizeof,
+                      "Mismatch in type size for call to unw_get_proc_name");
+
+        unw_context_t context;
+        unw_cursor_t cursor;
+        unw_getcontext(&context);
+        unw_init_local(&cursor, &context);
+
+        while (frames_to_skip > 0 && unw_step(&cursor) > 0)
+            --frames_to_skip;
+
+        unw_proc_info_t pip = void;
+        foreach (idx, ref frame; this.callstack)
+        {
+            if (int r = unw_get_proc_name(
+                    &cursor, frame.buff.ptr, frame.buff.length,
+                    cast(unw_word_t*) &frame.address))
+                frame.name = "<ERROR: Unable to retrieve function name>";
+            else
+                frame.name = frame.buff[0 .. strlen(frame.buff.ptr)];
+
+            if (unw_get_proc_info(&cursor, &pip) == 0)
+                frame.address += pip.start_ip;
+
+            this.numframes++;
+            if (unw_step(&cursor) <= 0)
+                break;
+        }
+    }
+
+    ///
+    override int opApply (scope int delegate(ref const(char[])) dg) const
+    {
+        return this.opApply((ref size_t, ref const(char[]) buf) => dg(buf));
+    }
+
+    ///
+    override int opApply (scope int delegate(ref size_t, ref const(char[])) dg) const
+    {
+        return traceHandlerOpApplyImpl2(this.callstack[0 .. this.numframes], dg);
+    }
+
+    ///
+    override string toString () const
+    {
+        string buf;
+        foreach ( i, line; this )
+            buf ~= i ? "\n" ~ line : line;
+        return buf;
+    }
+}
+
+/**
+ * Convenience function for power users wishing to test this module
+ * See `core.runtime.defaultTraceHandler` for full documentation.
+ */
+Throwable.TraceInfo defaultTraceHandler (void* ptr = null)
+{
+    // avoid recursive GC calls in finalizer, trace handlers should be made @nogc instead
+    import core.memory : gc_inFinalizer;
+    if (gc_inFinalizer)
+        return null;
+
+    return new LibunwindHandler();
+}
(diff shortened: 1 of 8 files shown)


Geod24:
Note: The second commit is not mandatory anymore, as it was there to allow adding `opApply(Location)`, an idea I discarded as it was a breaking change.

src/core/internal/backtrace/libunwind.d
```diff
@@ -0,0 +1,167 @@
...
+else version (RISCV64) // 32 is not supported
+{
+    enum _LIBUNWIND_CONTEXT_SIZE = 64;
+    enum _LIBUNWIND_CURSOR_SIZE = 76;
+    enum _LIBUNWIND_HIGHEST_DWARF_REGISTER = _LIBUNWIND_HIGHEST_DWARF_REGISTER_RISCV;
+}
+else
+    /*
+     * Libunwind also support OpenRISC 1000 (or1k), hexagon, and MIPS
+     * at the time of writing (December 2021).
```
jacob-carlborg, line 98: Is this PR from the future 😉.

src/core/internal/backtrace/handler.d
```diff
@@ -0,0 +1,116 @@
...
+ * See_Also: https://www.nongnu.org/libunwind/man/libunwind(3).html
+ */
+module core.internal.backtrace.handler;
+
+import core.internal.backtrace.dwarf;
+import core.internal.backtrace.libunwind;
+import core.internal.backtrace.primitives;
+
+/// Ditto
+public class LibunwindHandler : TraceInfo
```
jacob-carlborg, line 24: Do you want all this to be public in the `internal` package?

jacob-carlborg:
I know there are multiple libraries called "libunwind". Do they all provide the same API?

src/core/internal/backtrace/handler.d
Geod24: Ah, I have a habit of putting protection qualifier before the symbol, even when `public`.
Originally yes, but then I thought of adding `libunwindDefaultTraceHandler`, so this could be `package(core)`.

src/core/internal/backtrace/handler.d
```diff
@@ -0,0 +1,116 @@
...
+ * the user uses the following in `main`:
+ * ---
+ * import core.runtime;
+ * import core.internal.backtrace.handler;
+ * Runtime.traceHandler = &libunwindDefaultTraceHandler;
+ * ---
+ *
+ * Authors: Mathias 'Geod24' Lang
+ * Copyright: D Language Foundation - 2020
+ * See_Also: https://www.nongnu.org/libunwind/man/libunwind(3).html
```
jacob-carlborg, line 15: You're referring to LLVM libunwind in `core.internal.backtrace.libunwind`. Are they the same? Do they provide the same API.

Geod24:
> I know there are multiple libraries called "libunwind". Do they all provide the same API?

Similar ones, and I believe (but will verify) that the subset used is equivalent.
However, the cursor sizes might change. I'll test on Alpine Linux in the near future. So far I've developed this on Mac, so using LLVM's was just convenient.

There was a discussion about it on [the mailing list](http://lists.llvm.org/pipermail/cfe-dev/2016-September/050649.html).

Geod24:
@jacob-carlborg : I did more research and came up with bindings for both libraries. They have minor differences which makes them source compatible from C, but obviously not from D.

Geod24:
So this seems to work :|
Time to give it a spin on Alpine.

Geod24:
So, this is fun: The names exposed in the `.so` are actually now `unw_blah` but have a platform prefix.
Which makes sense, I guess, if you consider remote unwinding. Still, it will complicate the file even more.

Geod24:
@kinke : Does this make sense to you ? You did a pretty extensive rework of `execinfo` a while ago so you're probably the most qualified reviewer. Also, would LDC be able to use this by default provided it's been tested properly ?

Geod24:
>  druntime's EH code on Posix already depends on libunwind; the respective bindings are in rt.unwind.

Indeed. After some more research, this part of `libunwind` is in `libgcc_s.so`.
So I guess I should be using the high-level function [`_Unwind_Backtrace` which is also implemented by LLVM](https://github.com/llvm/llvm-project/blob/993d8ac5cb935b78fb136c25a7e4bae18852f429/libunwind/include/unwind.h#L298-L303), and no additional linking would be required.

Geod24:
> So I guess I should be using the high-level function `_Unwind_Backtrace` which is also implemented by LLVM, and no additional linking would be required.

Except there seems to be no way to get the procedure name directly from the `_Unwind_` functions, so I'll need to implement the lookup. Anyway, I have completely removed support for `libunwind`, and only left support for LLVM's libunwind, which is enough to get started.

Geod24:
Note that I already penabled it by default in DMD v2.095.0 on Alpine Linux](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/16405).
The failing GDB tests might be unrelated, but the eh.d test was [down to one issue](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/16498) which, due to its nature, I hope might vanish if we implement things using `_Unwind_*` functions.

AndrejMitrovic (approved):
LGTM. Does it need anyone else's approval?

Geod24:
Since this code is not enabled by default, I think we're pretty safe.

ibuclaw:
This PR caused a build regression on many platforms,

i.e: mips-linux-gnu
```
../../../../libphobos/libdruntime/core/internal/backtrace/libunwind.d:49:36: error: undefined identifier ‘_LIBUNWIND_CONTEXT_SIZE’
   49 |     ulong[_LIBUNWIND_CONTEXT_SIZE] data = void;
      |                                    ^
../../../../libphobos/libdruntime/core/internal/backtrace/libunwind.d:55:35: error: undefined identifier ‘_LIBUNWIND_CURSOR_SIZE’
   55 |     ulong[_LIBUNWIND_CURSOR_SIZE] data = void;
      |   
```

Geod24:
Hum, it doesn't show the `static assert` that fails.
You can just stick `version (DRuntime_Use_Libunwind):` at the top of the module doing the bindings.

src/core/internal/backtrace/libunwind.d
```diff
@@ -0,0 +1,153 @@
...
+ * and lower-level (`unw_*`) functions.
+ * Many distributions have a `libunwind` package as well, that provides
+ * the `unw_*` functions, but since it also supports remote unwinding,
+ * the function names are actually platform dependent and binding them
+ * is a pain as many things rely on `#define`.
+ *
+ * In the future, we would like to implement backtrace using only the
+ * higher-level functions (`_Unwind_*`), which will allow us to not
+ * use `backtrace` and friends directly, and only retrieve the functions
+ * names when needed (currently we need to eagerly get the functions names).
```
ibuclaw, line 20: Why aren't you using `_Unwind_Backtrace` from the start?

Geod24: Because I learned on the job, mostly. The original implementation aimed to be compatible with the old API (HP libunwind), which doesn't have those functions IIRC. Later I realized that just going with the GCC / LLVM libunwind was much better and ditched compatibility with HP libunwind. Note that this is only (or should only) be used on Alpine Linux at the moment.

ibuclaw: Original meaning 40+ year old API?  If I'm not mistaken the backtrace function has been around since 2000s.

ibuclaw: Something like this should suffice (though I didn't write it, and not sure I've ever tested it either given that libbacktrace is always present in gcc's tree).
https://github.com/gcc-mirror/gcc/blob/b2cd32b743ba440e75505ce30c6b5c592ed144ea/libphobos/libdruntime/gcc/backtrace.d#L428-L444

PR #3334, opened 2021-01-18
Use __pthread_get_minstack to adjust stack size for TLS in Glibc

kubo39:


ibuclaw (changes requested):
If druntime has a need to get a lower-level view of the TLS runtime then we should work together with upstream glibc to formalize what's required and provide a low-level TLS ABI.

Note that is also the case for Rust which uses `__pthread_get_minstack@@GLIBC_PRIVATE`, and this has prevented glibc maintainers from doing valid and meaningful clean-up of the libpthread internals to remove this function.

The related glibc bug is here:  https://sourceware.org/bugzilla/show_bug.cgi?id=16291

This is probably a bad idea to use an internal glibc symbol outside of the glibc as it can be removed - or worse, modified - at anytime, and in-fact this function seems to be effectively internally deprecated in glibc anyway.  So I'd vote no on adding this.

RazvanN7:
@kubo39 , @ibuclaw has a point. Is it ok if we close this?

PR #3336, opened 2021-01-21
Fix 21544 - -checkaction=context formats enum members as their base type

MoonlightSentinel:
Generate code that detects the correct enum member (or defaults to the base type in case of an invalid enum value).

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -152,6 +152,21 @@ private string miniFormat(V)(const scope ref V v)
             return miniFormat(*cast(T*) &v);
         }
     }
+    // Format enum members using their name
+    else static if (is(V BaseType == enum))
+    {
+        // Always generate repeated if's instead of switch to skip the detection
+        // of non-integral enums. This method doesn't need to be fast.
+        static foreach (mem; __traits(allMembers, V))
+        {
+            if (v == __traits(getMember, V, mem))
+                return mem;
+        }
+
+        // Format invalid enum values as their base type
+        enum cast_ = "cast(" ~ V.stringof ~ ")";
+        return combine(cast_, "", miniFormat(*(cast(BaseType*) &v)));
+    }
     else static if (is(V == bool))
     {
         return v ? "true" : "false";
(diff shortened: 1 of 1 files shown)


src/core/internal/dassert.d
```diff
@@ -152,6 +152,20 @@ private string miniFormat(V)(const scope ref V v)
...
+        // Always generate repeated if's instead of switch to skip the detection
+        // of non-integral enums. This method doesn't need to be fast.
+        static foreach (mem; __traits(allMembers, V))
+        {
+            if (v == __traits(getMember, V, mem))
+                return mem;
+        }
+
+        // Format invalid enum values as their base type
+        return "cast(" ~ V.stringof ~ ") " ~ miniFormat(*(cast(BaseType*) &v));
```
Geod24, line 167: Doesn't this allocate ? Also `miniFormat!BaseType(v)` should work, no ?

MoonlightSentinel: > Doesn't this allocate ?

Yes, but string concatenation is used in several places. But I could also re-use `combine` here...

>  Also `miniFormat!BaseType(v)` should work, no ?

No because `miniFormat`'s accepts it's parameter by `ref`

Geod24: > Yes, but string concatenation is used in several places.

If we are to make this the default, `assert(a != b)` should not allocate.
Note that it currently does anyway, because of `throw new AssertError`, but it really shouldn't, because that makes it unusable in destructors.

MoonlightSentinel: Guess this could use `combine` (or some other utility method) s.t. we can change the allocation strategy in the future.

But that will always be a problem for a user-defined `toString()`s that are not `@nogc` (unless those are skipped while running destructors)

schveiguy: Assert errors shouldn't really be caught. Can we just malloc it?

schveiguy: Another solution -- set up a "for asserts" allocation that uses malloc, and then frees everything just before the program exits.

schveiguy: > But that will always be a problem for a user-defined `toString()`s that are not `@nogc` (unless those are skipped while running destructors)

that's going to be a really tough problem to solve.

schveiguy: > Can we just malloc it?

hah, just looked at what combine does, and it does just that ;) That's the solution right there.

MoonlightSentinel: Done. Maybe we should rework `miniFormat` to use an `OutputRange`-like buffer s.t. it doesn't need to reallocate to build the final string.

src/core/internal/dassert.d
```diff
@@ -383,3 +398,33 @@ private auto pureAlloc(size_t t)
...
+    // Normal enum values
+    enum E { A, BCDE }
+    E e = E.A;
+    assert(miniFormat(e) == "A");
+    e = E.BCDE;
+    assert(miniFormat(e) == "BCDE");
+
+    // Invalid enum value is printed as their implicit base type (int)
+    e = cast(E) 3;
+    assert(miniFormat(e) == "cast(E)  3");
```
Geod24, line 414: Why the two spaces ?

Geod24 (approved):
Just a question, LGTM

src/core/internal/dassert.d
MoonlightSentinel: `combine` inserts a space before and after the token (here `""`)

RazvanN7 (approved):

PR #3341, opened 2021-01-23
Fix Issue 14790: rt: cover: coverage merge should detect changed source code

ljmf00:
Merging different versions of the same source file could lead to mismatched
covered lines if the source code actually changes. Changing the .lst file
format will open a huge breaking change and rely on timestamps is impractical.

As an alternative to this, we can deeply compare the source code with generated
.lst which also has a copy of the original source code.

### DIFF
--- a/src/rt/cover.d
+++ b/src/rt/cover.d
@@ -177,6 +177,42 @@ uint parseNum(const(char)[] s)
     return res;
 }
 
+const(char)[] parseContent(const(char)[] s)
+{
+    while (s.length && s[0] != '|')
+        s = s[1 .. $];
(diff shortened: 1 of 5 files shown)


ljmf00:
buildkite is heisenberging ¯\\\_(ツ)\_/¯

12345swordy:
cc @RazvanN7 can you help with this?

thewilsonator (approved):

TungstenHeart:
This causes a regression. Sometimes that corrupts the input lst. see https://issues.dlang.org/show_bug.cgi?id=21712.

PR #3345, opened 2021-01-25
Fix issue 21309 along with other dead links on dlang.org

ljmf00:

---
This PR regenerates `mak/DOCS` files, fix some wrong documentation on `Params` and `Source` sections.

Also, if someone knows a better approach for generating `$(DOCDIR)/%.html : src/%.d $(DMD)` makefile targets in a more generic way, would be cool. I don't know much about Makefile and I think that's kinda dirty, although this approach was already there.


ljmf00:
Is documentation issues part of `regression` severity?

PR #3346, opened 2021-01-25
Remove duplicated definitions for .min() and .max() functions

ljmf00:

### DIFF
--- a/src/core/internal/util/math.d
+++ b/src/core/internal/util/math.d
@@ -1,0 +1,53 @@
+// Written in the D programming language
+
+/**
+ * Internal math utilities.
+ *
+ * Copyright: The D Language Foundation 2021.
+ * License: $(HTTP boost.org/LICENSE_1_0.txt, Boost License 1.0).
+ * Authors: Luís Ferreira
+ * Source: $(DRUNTIMESRC core/internal/util/_math.d)
+ */
+module core.internal.util.math;
+
+/**
+ * Calculates the maximum of the passed arguments
+ * Params:
+ *   a = first value to select the maximum from
+ *   b = second value to select the maximum from
+ * Returns: The maximum of the passed-in values.
+ */
+T max(T)(T a, T b) pure nothrow @nogc @safe
+{
+    return b > a ? b : a;
+}
+
+/**
+ * Calculates the minimum of the passed arguments
+ * Params:
+ *   a = first value to select the minimum from
+ *   b = second value to select the minimum from
+ * Returns: The minimum of the passed-in values.
+ */
+T min(T)(T a, T b) pure nothrow @nogc @safe
+{
+    return b < a ? b : a;
+}
+
+///
+@safe pure @nogc nothrow
+unittest
+{
+    assert(max(1,3) == 3);
+    assert(max(3,1) == 3);
+    assert(max(1,1) == 1);
+}
+
+///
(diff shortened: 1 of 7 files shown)


src/core/internal/util/math.d
```diff
@@ -0,0 +1,53 @@
+// Written in the D programming language
+
+/**
+ * Internal math utilities.
+ *
+ * Copyright: Luís Ferreira 2021
```
ljmf00, line 6: I'm not really sure if this should be my or D Programming Language Foundation copyrights. Can someone please clarify it to me?

wilzbach: Should be DLF. Old code only has it for historical reasons.

ljmf00: Force pushed :)

src/core/internal/util/math.d
```diff
@@ -0,0 +1,53 @@
...
+ *   b = second value to select the minimum from
+ * Returns: The minimum of the passed-in values.
+ */
+T min(T)(T a, T b) pure nothrow @nogc @safe
+{
+    return b < a ? b : a;
+}
+
+///
+@safe pure
```
RazvanN7, line 38: `@nogc nothrow`

src/core/internal/util/math.d
```diff
@@ -0,0 +1,53 @@
...
+unittest
+{
+    assert(max(1,3) == 3);
+    assert(max(3,1) == 3);
+    assert(max(1,1) == 1);
+}
+
+///
+@safe pure
```
RazvanN7, line 47: ditto

RazvanN7 (commented):
Creating a new module just for those functions seems like an overkill to me, however, the duplication is also annoying. @schveiguy you have more experience with druntime, what do you think about this?

ljmf00:
> Creating a new module just for those functions seems like an overkill to me, however, the duplication is also annoying. @schveiguy you have more experience with druntime, what do you think about this?

Yes, I understand, although I thought it would be fine because it will:
- remove some duplication and possibly other future math duplications out there
- not introduce anything new to the public API nor breaking changes.

maxhaton:
> Creating a new module just for those functions seems like an overkill to me, however, the duplication is also annoying. @schveiguy you have more experience with druntime, what do you think about this?

Maybe the name could be improved (something more general to avoid future modules?) but it seems likely that this situation will happen again in the future so both setting the standard early of avoiding duplication and avoiding encouraging putting generic stuff in one file is a good thing.

src/core/internal/util/math.d
ljmf00: Done

src/core/internal/util/math.d
ljmf00: Done

RazvanN7 (approved):

src/core/internal/util/math.d
```diff
@@ -0,0 +1,53 @@
+// Written in the D programming language
```
dkorpel, line 1: What's the point of this line?

ljmf00: I'm not really sure honestly, but I just followed the criteria around some files in `druntime`/`phobos`.

e.g.:
https://github.com/dlang/druntime/blob/f94e816c1dd229cbb133085a6fc41d1bd6d45594/src/core/math.d#L1

PR #3351, opened 2021-01-28
core.internal.convert: Deprecate complex toUbyte

ibuclaw:
This function isn't used anywhere in druntime, and as it's internal, probably doesn't even need a deprecation cycle.


thewilsonator (approved):

wilzbach:
Yeah internal functionality doesn't need to go through the deprecation cycle.

ibuclaw:
> Yeah internal functionality doesn't need to go through the deprecation cycle.

I'll just remove it after the next release then. :-)

PR #3352, opened 2021-01-29
core.lifetime.emplace(): Properly forward r/lvalue-ness of arguments

kinke:
Some `auto ref` args weren't forwarded.

### DIFF
--- a/src/core/lifetime.d
+++ b/src/core/lifetime.d
@@ -121,11 +121,11 @@ T emplace(T, Args...)(T chunk, auto ref Args args)
     else alias args1 = args;
 
     // Call the ctor if any
-    static if (is(typeof(chunk.__ctor(args1))))
+    static if (is(typeof(chunk.__ctor(forward!args1))))
     {
         // T defines a genuine constructor accepting args
         // Go the classic route: write .init first, then call ctor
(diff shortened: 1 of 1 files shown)


src/core/lifetime.d
```diff
@@ -931,51 +945,6 @@ version (CoreUnittest)
     }
 }
 
-/+ these tests can't be performed in druntime, but a mirror still exists in phobos...
```
kinke, line 934: These tests are still in `std.conv`. I don't think this comment and the 2 disabled tests provide any value whatsoever here in druntime and see it rather as confusing clutter.

RazvanN7 (approved):

thewilsonator:
cc @TurkeyMan does this fix https://issues.dlang.org/show_bug.cgi?id=19904 ?

TurkeyMan:
Hard to know. It looks like it's the right path to fixing it... whether every missing call to `forward` has been caught though, I'd need to audit. I have no working project right now to test on.

kinke:
There was this and 2 Phobos PRs: https://github.com/dlang/phobos/pull/7745 and https://github.com/dlang/phobos/pull/7776

PR #3353, opened 2021-01-30
[stable] rt.sections_elf_shared: Upstream MIPS/RISC-V fix wrt. .so dependency names from GDC

kinke:
See https://github.com/gcc-mirror/gcc/blob/25f303e9a27a0d799f9979446d9f11ed62b8d648/libphobos/libdruntime/gcc/sections/elf_shared.d#L760-L772. It's also relevant for LDC.

### DIFF
--- a/src/rt/sections_elf_shared.d
+++ b/src/rt/sections_elf_shared.d
@@ -19,6 +19,11 @@ else version (CRuntime_UClibc) enum SharedELF = true;
 else enum SharedELF = false;
 static if (SharedELF):
 
+version (MIPS32)  version = MIPS_Any;
+version (MIPS64)  version = MIPS_Any;
+version (RISCV32) version = RISCV_Any;
+version (RISCV64) version = RISCV_Any;
+
 // debug = PRINTF;
 import core.internal.elf.dl;
 import core.memory;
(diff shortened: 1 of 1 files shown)


src/rt/sections_elf_shared.d
```diff
@@ -725,17 +730,29 @@ version (Shared)
             if (dyn.d_tag == DT_STRTAB)
             {
                 version (CRuntime_Musl)
-                    strtab = cast(const(char)*)(object.baseAddress + dyn.d_un.d_ptr); // relocate
+                    enum relocate = true;
                 else version (linux)
-                    strtab = cast(const(char)*)dyn.d_un.d_ptr;
+                {
```
Geod24, line 735: Shouldn't this be `CRuntime_Glibc` then, instead of "linux" ?

kinke: I'm not keen on looking up `CRuntime_UClibc` etc.

kinke: At least the MIPS part seems to apply to uclibc too: https://cgit.uclibc-ng.org/cgi/cgit/uclibc-ng.git/tree/ldso/include/ldsodefs.h#n12

@ibuclaw: Do you know if there'd be a more robust way to detect this? Like checking the `PF_W` flag of the `PT_DYNAMIC` program header?

ibuclaw: This is an internal detail, so no I don't think so.  It did come up when the RISCV developers implemented this in gdc because it seems to have been a mistake to copy MIPS, I guess we'll just have to address that hurdle when we get to it.

RazvanN7 (approved):

ibuclaw:
Forgot that I raised a bug report about it years ago.
https://issues.dlang.org/show_bug.cgi?id=19622

PR #3357, opened 2021-02-01
Add platform-specific sysctl declarations

ibuclaw:
There might be a few enum definitions still missing, but at least the usage in phobos std.file and std.parallelism is covered.


ibuclaw:
There's darwin to sort out as well...

RazvanN7 (approved):

PR #3358, opened 2021-02-01
core.stdc.stdio: Fix "read-modify-write operations are not allowed for shared variables" on OpenBSD

ibuclaw:


maxhaton (approved):

src/core/stdc/stdio.d
```diff
@@ -1567,7 +1567,7 @@ else version (OpenBSD)
     {
         void __sclearerr()(FILE* p)
         {
-            p._flags &= ~(__SERR|__SEOF);
+            p._flags = p._flags & ~(__SERR|__SEOF);
```
RazvanN7, line 1570: how does this fix anything? Isn't the generated code going to look the same?

ibuclaw: See PR title.

RazvanN7: I have read the title. Since read-modify-write operations are not allowed on shared variables on OpenBSD, how does this change fix that? Isn't `a &= b`  going to be lowered to the exact same code as `a = a & b`? Isn't the code you are adding still a RMW?

ibuclaw: Raise a bug against dmd if you disagree.  Let me make a slight punctuation change to the title.

Fix "read-modify-write operations are not allowed for shared variables" on OpenBSD

RazvanN7: I see, ok, thanks for the explanation. I didn't expect that `a &= b` yields different results from `a = a & b` in generated code, hence my confusion with regards to the fix.

RazvanN7 (approved):

src/core/stdc/stdio.d
ibuclaw: [CRuntime_DigitalMars](https://github.com/dlang/druntime/blob/d018691bad95520bd8d0b79510c9aad67ceb3621/src/core/stdc/stdio.d#L1339) will also fail if the compiler starts recognising the `foo = foo + y` pattern.

PetarKirov:
@ibuclaw Isn't the correct fix to use `core.atomic.atomicOp`?

ibuclaw:
> @ibuclaw Isn't the correct fix to use `core.atomic.atomicOp`?

I think that's overkill, as it's only ever called in single-threaded mode.

PR #3363, opened 2021-02-03
[dmd-cxx] Synchronize bindings with upstream druntime master

ibuclaw:
Includes locale_t fixes for Solaris, OpenBSD additions and bindings to supplant extern(C) declarations in phobos.


ibuclaw:
Was going to wait for `brk`/`sbrk`, but right now I think it can wait a while for the decision on what to do in upstream first.

PR #3364, opened 2021-02-03
core.sys.posix.unistd: Add XSI legacy bindings

ibuclaw:
In the unistd module, there are a number of legacy XSI functions, not all are presently available on all supported platforms.

[std.experimental](https://github.com/dlang/phobos/blob/1c77cdad29161f8bdb84475ffa52c22e7745a883/std/experimental/allocator/building_blocks/region.d#L724-L743) also declares a couple of these functions locally (they shouldn't be doing this).

Some thoughts:
1. Should these be labelled as deprecated?
2. Should they be moved to `core.sys.<platform-specific>.unistd`?  Existing functions in `core.sys.posix.unistd` will then be labelled as `deprecated`.

### DIFF
--- a/src/core/sys/posix/unistd.d
+++ b/src/core/sys/posix/unistd.d
@@ -2408,7 +2408,6 @@ int        fchdir(int);
 c_long     gethostid();
 pid_t      getpgid(pid_t);
 pid_t      getsid(pid_t);
-char*      getwd(char*); // LEGACY
 int        lchown(const scope char*, uid_t, gid_t);
 int        lockf(int, int, off_t);
 int        nice(int);
@@ -2420,9 +2419,6 @@ int        setreuid(uid_t, uid_t);
 void       swab(const scope void*, void*, ssize_t);
 void       sync();
 int        truncate(const scope char*, off_t);
-useconds_t ualarm(useconds_t, useconds_t);
-int        usleep(useconds_t);
-pid_t      vfork();
 */
 
 version (CRuntime_Glibc)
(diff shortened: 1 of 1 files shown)


maxhaton:
What does not supported mean in this case - completely missing or a runtime error? If we are silently assuming they are there when they aren't, deprecation seems the way (then they should be moved to a platform specific alternative).

Also, what platforms are effected specifically (I'm not all that familiar with XSI beyond the very basics)?

ibuclaw:
> What does not supported mean in this case - completely missing or a runtime error?

Depends on the platform.  For Musl, calling `sbrk()` only sets errno to `ENOMEM` and nothing else, whereas on DragonFly, it is completely absent.

> If we are silently assuming they are there when they aren't, deprecation seems the way (then they should be moved to a platform specific alternative).
> 
> Also, what platforms are effected specifically (I'm not all that familiar with XSI beyond the very basics)?

All posix-compliant platforms should implement XSI, but when the baseline supported revision is adjusted - and most probably support version 7 by now - then these functions are `#ifdef`'d out of the headers.

RazvanN7:
I think that the simplest way would be to just deprecate them. @ibuclaw care to do that in this PR?

RazvanN7:
ping @ibuclaw

RazvanN7:
@ibuclaw How do we move forward here?

RazvanN7:
@ibuclaw Should we close this?

Geod24:

PR #3372, opened 2021-02-10
macOS: add core.sys.darwin.fcntl

kubo39:
This is the preparation for https://issues.dlang.org/show_bug.cgi?id=21627.

### DIFF
--- a/src/core/sys/darwin/fcntl.d
+++ b/src/core/sys/darwin/fcntl.d
@@ -1,0 +1,20 @@
+module core.sys.darwin.fcntl;
+
+public import core.sys.posix.fcntl;
+
+version (OSX)
+    version = Darwin;
+else version (iOS)
+    version = Darwin;
+else version (TVOS)
+    version = Darwin;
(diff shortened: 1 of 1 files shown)


src/core/sys/darwin/fcntl.d
```diff
@@ -0,0 +1,20 @@
...
+else version (WatchOS)
+    version = Darwin;
+
+version (Darwin):
+extern (C):
+nothrow:
+@nogc:
+@system:
+
+enum F_FULLSYNC = 51;
```
jacob-carlborg, line 20: It's named `F_FULLFSYNC`.

jacob-carlborg (changes requested):

src/core/sys/darwin/fcntl.d
kubo39: Ahh, sorry.

jacob-carlborg (approved):

RazvanN7 (approved):

PR #3373, opened 2021-02-12
core.atomic.cas: add 'const' attribute to local variable 'arg1'

thaven:
Avoid possible unnecessary const to mutable conversion, which is not implicit for pointers. See [Issue 21631](https://issues.dlang.org/show_bug.cgi?id=21631) for details.

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -292,7 +292,7 @@ template cas(MemoryOrder succ = MemoryOrder.seq, MemoryOrder fail = MemoryOrder.
     in (atomicPtrIsProperlyAligned(here), "Argument `here` is not properly aligned")
     {
         // resolve implicit conversions
-        T arg1 = ifThis;
+        const T arg1 = ifThis;
         T arg2 = writeThis;
 
         static if (__traits(isFloating, T))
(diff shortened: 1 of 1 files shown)


RazvanN7:
This seems to break both the dmd's and phobos's test suites.

Geod24:
That looks like an invalid issue to me. You should not be able to store a `const` pointer in a non-`const` pointer.
The code is essentially trying to do the following:
```D
const(int)* ptr;
int* ptr = ptr;
```

Which is not sound.

thaven:
@Geod24  not really. The case is more like:

```D
int* a;
const(int)* b;
int* c;

if (a is b)
    a = c;
```

So it is just comparing pointer to const against pointer to mutable.

Geod24:
Oh, I see. Indeed. However using `const` is not going to cut it. Those "implicit conversions" statements IMO are wrong and shouldn't be there. But you'd have to dig the git history to see why they are there in the first place.

thaven:
Those statements seem to originate from this commit: https://github.com/dlang/druntime/commit/1b139035a32f6aeedcdb8d1589af45e9deb73f88 from PR https://github.com/dlang/druntime/pull/2745

RazvanN7:
ping @thaven . How do we proceed here?

RazvanN7:
This seems to be passing all tests now.

thewilsonator (approved):

PR #3376, opened 2021-02-17
[REG 2.084] hashOf will fail to compile for some structs/unions that recursively contain shared enums

n8sh:


test/hash/src/test_hash.d
```diff
@@ -232,6 +233,19 @@ void issue20034()
...
+void issue21642() @safe nothrow pure
+{
+    enum C : char { _ = 1, }
+    union U { C c; void[0] _; }
+    shared union V { U u; }
+    cast(void) hashOf(V.init);
+    // Also test the underlying reason the above was failing.
+    import core.internal.convert : toUbyte;
+    shared C c;
+    cast(void) toUbyte(c);
```
PetarKirov, line 246: Instead of casting to void here and above, it's better to check if the return value is the expected one.

n8sh: Added a check here. Above there is no particular value the hash code is supposed to be.

PetarKirov: My motivation is the question: why would an optimizing compiler not remove a call to a pure function, if it's result is not used?

Technically such optimization happens after semantic analysis (where the bug manifests), in the backend, but in theory the frontend could also discard such function calls, so it's better to make the code more robust.

RazvanN7 (approved):

test/hash/src/test_hash.d
n8sh: Is there a change you would suggest that wouldn't make the test brittle?

RazvanN7: @PetarKirov Pure functions that may throw exceptions should not be optimized away. This pattern is prevalent in druntime: a function that does some checks and throws if the input is not the expected one.

RazvanN7:
@n8sh This seems to fail the tester pipeline. Any update on that?

src/core/internal/convert.d
```diff
@@ -769,7 +769,7 @@ const(ubyte)[] toUbyte(T)(const ref T val) if (is(T == enum))
     if (__ctfe)
     {
         static if (is(T V == enum)){}
-        return toUbyte(cast(const V) val);
+        return toUbyte(*cast(const V*) &val);
```
RazvanN7, line 772: This change is reported to not be covered by any tests.

n8sh: The coverage report doesn't take into account the `tests` directory. It is covered by the new test added in this PR.

n8sh:
Accidentally rebased on master instead of stable. Fixing.

PR #3377, opened 2021-02-20
core.stdc.config: Add `c_complex_float`, `c_complex_double`, and `c_complex_real` aliases

ibuclaw:
Can now be added to druntime as dlang/dmd#12167 is merged.

Start using them in `rt.memset`, which does not result in any changes to the generated assembly.

### DIFF
--- a/src/core/stdc/config.d
+++ b/src/core/stdc/config.d
@@ -31,6 +31,7 @@ version (StdDdoc)
             alias ddoc_long = int;
             alias ddoc_ulong = uint;
         }
+        struct ddoc_complex(T) {};
     }
 
     /***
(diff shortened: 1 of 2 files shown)


Geod24 (approved):

src/core/stdc/config.d
```diff
@@ -208,4 +227,28 @@ else
...
+    align(float.alignof)  enum __c_complex_float : _Complex!float;
+    align(double.alignof) enum __c_complex_double : _Complex!double;
+    align(real.alignof)   enum __c_complex_real : _Complex!real;
+}
+else
+{
+    align(float.sizeof * 2)  enum __c_complex_float : _Complex!float;
+    align(double.sizeof * 2) enum __c_complex_double : _Complex!double;
+    align(real.alignof)      enum __c_complex_real : _Complex!real;
+}
```
kinke, line 249: Where do the non-Posix aliases come from? - MSVC doesn't have these complex types built-in, it has `_{F,D,L}complex` typedefs (https://docs.microsoft.com/en-us/cpp/c-runtime-library/complex-math-support?view=msvc-160). `_Lcomplex` for `__c_complex_real` uses its 64-bit `long double`. The alignments are regular (4/8/8 bytes).

So is this based on the DMC runtime for 32-bit Windows?

ibuclaw: > So is this based on the DMC runtime for 32-bit Windows?

Or DMD 32 and 64-bit Windows.

If there's a discrepancy, the bug is in the compiler and needs fixing there.

https://github.com/dlang/dmd/blob/master/src/dmd/target.d#L214-L236

kinke: Yeah that looks quite awful; I'll just put this behind `version(DigitalMars)` in an upcoming PR, streamlining the other definitions.

PR #3379, opened 2021-02-21
rt.util.typeinfo: Deprecate typeinfo for built-in complex and imaginary types

ibuclaw:
Implements phase 3 of https://issues.dlang.org/show_bug.cgi?id=14488

To get around the deprecation errors when building druntime, we can use the new aliases for the "magic" complex types to build to the built-in TypeInfo classes.  These have been marked as deprecated with the provisional removal date of after version 2.105.

Druntime can now be built with `-transition=complex`.

### DIFF
--- a/src/rt/util/typeinfo.d
+++ b/src/rt/util/typeinfo.d
@@ -7,6 +7,7 @@
  * Source: $(DRUNTIMESRC rt/util/_typeinfo.d)
  */
 module rt.util.typeinfo;
+import core.stdc.config;
 static import core.internal.hash;
 
 template Floating(T)
@@ -36,14 +37,15 @@ if (is(T == float) || is(T == double) || is(T == real))
 
     public alias hashOf = core.internal.hash.hashOf;
 }
+
 template Floating(T)
-if (is(T == cfloat) || is(T == cdouble) || is(T == creal))
+if (is(T == c_complex_float) || is(T == c_complex_double) || is(T == c_complex_real))
 {
   pure nothrow @safe:
 
     bool equals(T f1, T f2)
     {
-        return f1 == f2;
+        return f1.re == f2.re && f1.im == f2.im;
     }
 
     int compare(T f1, T f2)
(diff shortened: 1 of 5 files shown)


src/rt/util/typeinfo.d
```diff
@@ -36,14 +37,15 @@ if (is(T == float) || is(T == double) || is(T == real))
...
 template Floating(T)
-if (is(T == cfloat) || is(T == cdouble) || is(T == creal))
+if (is(T == c_complex_float) || is(T == c_complex_double) || is(T == c_complex_real))
 {
   pure nothrow @safe:
 
     bool equals(T f1, T f2)
     {
-        return f1 == f2;
+        return f1.re == f2.re && f1.im == f2.im;
```
RazvanN7, line 48: Is this change really necessary? The compiler is smart enough to generate an appropriate `opEquals` for `struct Complex(T) {T re; T im;}`

src/rt/util/typeinfo.d
```diff
@@ -105,6 +109,52 @@ if (is(T ==  float) || is(T ==  double) || is(T ==  real) ||
     public alias hashOf = core.internal.hash.hashOf;
 }
 
+template Array(T)
```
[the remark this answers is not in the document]
ibuclaw: Makes it easy to remove later, and doesn't hinder the ability to use git-blame with useless whitespace changes.

src/rt/util/typeinfo.d
ibuclaw: This is an enum type, not a struct, so there is no `opEquals`, and there is no harm in inlining it explicitly.

src/rt/util/typeinfo.d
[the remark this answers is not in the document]
RazvanN7: Fair enough.

RazvanN7 (approved):

src/rt/util/typeinfo.d
ibuclaw: Actually, I'll clarify that this is only a struct type when `version=StdDdoc`.  Hence why I had to add fields to `ddoc_complex` in order to satisfy the documentation CI pipeline. :-)

PR #3381, opened 2021-02-23
DIP1034 add noreturn to object, abort, exit, _Exit

WalterBright:
Time to give it a spin.

https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1034.md

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -15,6 +15,7 @@ alias size_t = typeof(int.sizeof);
 alias ptrdiff_t = typeof(cast(void*)0 - cast(void*)0);
 
 alias sizediff_t = ptrdiff_t; // For backwards compatibility only.
+alias noreturn = typeof(*null);  /// bottom type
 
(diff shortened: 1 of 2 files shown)


src/object.d
```diff
@@ -15,6 +15,7 @@ alias size_t = typeof(int.sizeof);
 alias ptrdiff_t = typeof(cast(void*)0 - cast(void*)0);
 
 alias sizediff_t = ptrdiff_t; // For backwards compatibility only.
+alias noreturn = typeof(*null);  /// bottom type
```
Geod24, line 18: Do we want it in `object.d` ? It's a rare enough case that I think we can live with `typeof(*null)` in the signature.

WalterBright: We call it noreturn in the documentation, code, and error messages. Because of the way module scopes work, it shouldn't interfere with user use of the name, just like it doesn't for string.

Geod24 (approved):

PR #3322, opened 2021-01-02
Refactor some backtrace code to make it possible to use libunwind

Geod24:
See https://github.com/dlang/druntime/pull/3321 for the full context.
Technically, the first commit could be discarded since `processCallback` is reused by `traceInfoOpApplyImpl2`, but it just makes the code much easier to reason about and opens up the possibility to fold it into the `opApply` directly in the future.

PR #3382, opened 2021-02-23
core.internal.backtrace.dwarf: Fix segfault regression in case executable cannot be opened

kinke:
Introduced by #3322. `image.baseAddress()` must not be called if `!image.isValid()`, otherwise a segfault occurs if the executable file/image could not be opened (e.g., because of the process changing the initial working dir and https://issues.dlang.org/show_bug.cgi?id=21656).

### DIFF
--- a/src/core/internal/backtrace/dwarf.d
+++ b/src/core/internal/backtrace/dwarf.d
@@ -187,7 +187,7 @@ int traceHandlerOpApplyImpl(size_t numFrames,
 
 
     if (!image.isValid())
-        return locations[startIdx .. $].processCallstack(null, image.baseAddress, dg);
+        return locations[startIdx .. $].processCallstack(null, 0, dg);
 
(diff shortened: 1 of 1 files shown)


PetarKirov (approved):
Thanks for the fix, LGTM.

---

Some extra detail for other reviewers:

Given that the `baseAddress` parameter of `processCallstack` [is only used](https://github.com/dlang/druntime/blob/b2b7732ab1c9cdf5e1dda85b239f43a92558c535/src/core/internal/backtrace/dwarf.d#L247) when `debugLineSectionData` parameter is not null, it may seem like passing `0` for it here is of no consequence. However `Image.baseAddress` is actually a member function property which triggers the bug that this PR fixes. I must have overlooked this detail here: https://github.com/dlang/druntime/pull/3322/files#diff-7c0ebe7e385f5fadd65ebb53ce4ceaad90d419af813405a5efb828fb99deea79R181-R184

PR #3383, opened 2021-02-26
CRuntime_Musl: fixes for time64

omerfirmak:
```
The original PR (#3275) missed quite a few spots and conversions,
which led to the build on Alpine Linux failing with Aithmetic Exception
on core.time module constructor.
Links to the two offending commits are included.
For further issues / investigation, search for 'time64' in the git repository.
```

### DIFF
--- a/changelog/musl-32bits.dd
+++ b/changelog/musl-32bits.dd
@@ -1,0 +1,11 @@
+Support time64 changes for `CRuntime_Musl`
+
+Up to v1.1.24, Musl used a 32 bits `time_t` on 32 bits architectures.
+Since v1.2.0, `time_t` is now always 64 bits.
+From this release, druntime will also default to a 64 bits `time_t`
+on 32 bits architecture, unless the `CRuntime_Musl_Pre_Time64` is provided.
+
+This change should only affect packagers for Musl-based systems who support
+32 bits architectures (64 bits architectures already use 64 bits `time_t`),
+who now need to define `CRuntime_Musl_Pre_Time64` both when building
+druntime / Phobos and in the default configuration, if the linked Musl is < 1.2.0.
--- a/src/core/stdc/time.d
+++ b/src/core/stdc/time.d
@@ -16,9 +16,17 @@
 module core.stdc.time;
 
 version (Posix)
+{
     public import core.sys.posix.stdc.time;
+    import core.sys.posix.sys.types : CRuntime_Musl_Needs_Time64_Compat_Layer;
+}
 else version (Windows)
+{
     public import core.sys.windows.stdc.time;
+    // This enum is defined only for Posix, this file is the only one
+    // needing it in `core.stdc`.
+    private enum CRuntime_Musl_Needs_Time64_Compat_Layer = false;
+}
 else
     static assert(0, "unsupported system");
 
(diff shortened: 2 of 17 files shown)


changelog/musl-32bits.dd
```diff
@@ -0,0 +1,11 @@
+Support time64 changes for `CRuntime_Musl`
+
+Up to v1.1.24, Musl used a 32 bits `time_t` on 32 bits architectures.
+Since v1.2.0, `time_t` is now always 64 bits.
+From this release, druntime will also default to a 64 bits `time_t`
+on 32 bits architecture, unless the `CRuntime_Musl_Pre_Time64w` is provided.
```
ibuclaw, line 6: Typo in version condition.

changelog/musl-32bits.dd
```diff
@@ -0,0 +1,11 @@
...
+
+This change should only affect packagers for Musl-based systems who support
+32 bits architectures (64 bits architectures already use 64 bits `time_t`),
+who now need to define `CRuntime_Musl_Pre_Time64` both when building
+druntime / Phobos and in the default configuration, if the linked Musl is < 1.2.0.
```
ibuclaw, line 11: Is there a configure test that could be used by gdc for its build system, so that the burden isn't shouldered to the users/client code?

src/core/stdc/time.d
```diff
@@ -16,9 +16,17 @@
...
+{
     public import core.sys.posix.stdc.time;
+    import core.sys.posix.sys.types : CRuntime_Musl_Needs_Time64_Compat_Layer;
+}
 else version (Windows)
+{
     public import core.sys.windows.stdc.time;
+    // This enum is defined only for Posix, this file is the only one
+    // needing it in `core.stdc`.
+    private enum CRuntime_Musl_Needs_Time64_Compat_Layer = false;
```
ibuclaw, line 28: Perhaps all function declarations present here should be moved to `core.sys.<platform>.stdc.time` then.

src/core/sys/linux/sys/socket.d
```diff
@@ -13,6 +13,22 @@ extern(C):
...
+    enum SO_TIMESTAMP = 29;
+    enum SO_TIMESTAMPNS = 35;
+    enum SO_TIMESTAMPING = 37;
+
+}
+else
+{
+    enum SO_TIMESTAMP = 63;
+    enum SO_TIMESTAMPNS = 64;
+    enum SO_TIMESTAMPING = 65;
```
ibuclaw, line 29: Wait, so now there is backwards compatibility with Musl, but not older Linux kernels?

src/core/stdc/time.d
```diff
@@ -29,20 +37,43 @@ extern (C):
...
 ///
 @system char*   asctime(const scope tm* timeptr); // @system: MT-Unsafe race:asctime locale
-///
-@system char*   ctime(const scope time_t* timer); // @system: MT-Unsafe race:tmbuf race:asctime env locale
-///
-@system tm*     gmtime(const scope time_t* timer); // @system: MT-Unsafe race:tmbuf env locale
-///
-@system tm*     localtime(const scope time_t* timer); // @system: MT-Unsafe race:tmbuf env locale
-///
-@system size_t  strftime(scope char* s, size_t maxsize, const scope char* format, const scope tm* timeptr); // @system: MT-Safe env locale
```
n8sh, line 48: Is it deliberate that `strftime` is removed entirely?

Geod24: Nope that's a mistake

src/core/sys/linux/sys/socket.d
Geod24: We probably should just ditch backward compatibility for Musl. And that's a mistake, too.

src/core/stdc/time.d
omerfirmak: Updated

changelog/musl-32bits.dd
ibuclaw: In particular, when running the testsuite.

RazvanN7:
@Geod24 What is your stance on this?

RazvanN7:
@ibuclaw Is it ok to merge this?

ibuclaw (changes requested):
We should avoid using `static if` inside core.stdc.time, otherwise what was the point of splitting the module into core.sys.posix.stdc.time and windows.stdc.time in the first place.

There's no apparent backwards compatibility for older Linux versions (people are still occasionally running gdc testsuite on centos with 2.6.xx), and yet musl is given that luxury.

A single core.sys.posix.config option might be the best way to go, rather than the mixed version/enum approach.

Geod24:
@RazvanN7 : This needs a bit more love, I just didn't have much time for it recently. Will give it a shot this WE.

RazvanN7:
@Geod24 any progress on this?

nmeum:
Hello, Alpine Linux contributor here 👋 We just upgraded our Alpine GCC package from GCC 11 to GCC 12.1 but due to the Aithmetic Exception described in this PR we were not able to get GDC to work on our 32-bit architectures (armhf, armv7, x86) and hence [had to disable it](https://github.com/alpinelinux/aports/commit/4b3952f03f53f2f60ee68a5b83a58c65d0c79343).

I would appreciate it if someone finds the time to finish work on this PR as this would allow us to enable GDC again for these architectures. If this PR gets merged it would also be nice if it could be backported to the GCC 12.1 release branch (`releases/gcc-12.1.0`), if possible.

thewilsonator:
@nmeum this PR needs to be retargeted to dlang/dmd as this repo is effectively archive only (it was merged into `dlang/dmd` to reduce hassle of multiple git modules). It can then be picked up by @ibuclaw (ping) to be merged into GDC.

thewilsonator:
Oh I just realised this PR is targeted to the dmd-cxx branch. 
I'm not sure how that factors into the merging of the druntime repo in to the compiler repo.

ibuclaw:
> Hello, Alpine Linux contributor here wave We just upgraded our Alpine GCC package from GCC 11 to GCC 12.1 but due to the Aithmetic Exception described in this PR we were not able to get GDC to work on our 32-bit architectures (armhf, armv7, x86) and hence [had to disable it](https://github.com/alpinelinux/aports/commit/4b3952f03f53f2f60ee68a5b83a58c65d0c79343).
> 

You have a huge list of custom patches to gcc, you can simply include this as another.

https://gitlab.alpinelinux.org/alpine/aports/-/tree/master/main/gcc

> I would appreciate it if someone finds the time to finish work on this PR as this would allow us to enable GDC again for these architectures. If this PR gets merged it would also be nice if it could be backported to the GCC 12.1 release branch (`releases/gcc-12.1.0`), if possible.

GCC release tags are immutable, you have a week to sort it out until 12.2 is cut, otherwise wait until 12.3.

nmeum:
> You have a huge list of custom patches to gcc, you can simply include this as another.

I am well aware of the size of our downstream GCC patchset. I don't understand how this relates to fixing a known bug upstream. We are actively trying to get musl compatibility issues fixed upstream to reduce the size of our downstream patchset, which is why I pinged this again.

On a side note, I would also like to point out that most of our patches are quite small. This particular patches changes ~400 lines across 17 files and would thus be by far our biggest patch if we end up backporting it eventually (ideally after the outstanding comments have addressed and this patch has been merged upstream).

PR #3387, opened 2021-03-05
benchmark: add `__MANGLED_FUNCTION__` special keyword to the lexer and parser

ljmf00:

---

CC: @maxhaton 
Related to https://github.com/dlang/dmd/pull/12252.


RazvanN7:
@ljmf00 This seems to be stalled on dlang/dmd#12252 which is also stalled. How do we proceed here?

ljmf00:
> @ljmf00 This seems to be stalled on [dlang/dmd#12252](https://github.com/dlang/dmd/pull/12252) which is also stalled. How do we proceed here?

I think this is not to be implemented in favour of the trait described during the discussion of the referred PR. I will try to revive the `trait` implementation from the ghost user.

PR #3394, opened 2021-03-14
Finish (mostly) OpenBSD support

ibara:
Hello --

This PR, along with the previously posted dmd PR, creates a dmd compiler that can reproduce itself on OpenBSD/amd64.
No changes were needed to Phobos.

There are some small bugs left I hope to fix.

Thanks!

### DIFF
--- a/src/core/internal/elf/dl.d
+++ b/src/core/internal/elf/dl.d
@@ -26,6 +26,11 @@ else version (DragonFlyBSD)
     import core.sys.dragonflybsd.sys.link_elf;
     version = LinuxOrBSD;
 }
+else version (OpenBSD)
+{
+    import core.sys.openbsd.sys.link_elf;
+    version = LinuxOrBSD;
+}
 
 version (LinuxOrBSD):
 
@@ -74,9 +79,10 @@ struct SharedObject
      */
     static bool findForAddress(const scope void* address, out SharedObject result)
     {
-        version (linux)       enum IterateManually = true;
-        else version (NetBSD) enum IterateManually = true;
-        else                  enum IterateManually = false;
+        version (linux)        enum IterateManually = true;
+        else version (NetBSD)  enum IterateManually = true;
+        else version (OpenBSD) enum IterateManually = true;
+        else                   enum IterateManually = false;
 
         static if (IterateManually)
         {
--- a/src/rt/sections.d
+++ b/src/rt/sections.d
@@ -27,6 +27,15 @@ else version (FreeBSD)
     public import rt.sections_elf_shared;
 else version (NetBSD)
     public import rt.sections_elf_shared;
+else version (OpenBSD)
+{
+    /**
+     * OpenBSD is missing support needed for elf_shared.
+     * See the top of sections_solaris.d for more info.
+     */
+
+    public import rt.sections_solaris;
+}
 else version (DragonFlyBSD)
     public import rt.sections_elf_shared;
 else version (Solaris)
--- a/src/rt/sections_solaris.d
+++ b/src/rt/sections_solaris.d
@@ -2,6 +2,10 @@
  * Written in the D programming language.
  * This module provides Solaris-specific support for sections.
  *
+ * OpenBSD (as of 6.9) lacks dlinfo, RTLD_NOLOAD, and dlpt_tls_modid.
+ * Faking support caused problems.
+ * That is why we are not using elf_shared on OpenBSD.
+ *
  * Copyright: Copyright Martin Nowak 2012-2013.
  * License:   $(HTTP www.boost.org/LICENSE_1_0.txt, Boost License 1.0).
  * Authors:   Martin Nowak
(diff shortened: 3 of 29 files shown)


src/core/internal/elf/dl.d
```diff
@@ -76,6 +81,7 @@ struct SharedObject
     {
         version (linux)       enum IterateManually = true;
         else version (NetBSD) enum IterateManually = true;
+        else version (OpenBSD) enum IterateManually = true;
         else                  enum IterateManually = false;
```
thewilsonator, line 85: Please keep these aligned

src/rt/sections.d
```diff
@@ -27,6 +27,8 @@ else version (FreeBSD)
     public import rt.sections_elf_shared;
 else version (NetBSD)
     public import rt.sections_elf_shared;
+else version (OpenBSD)
+    public import rt.sections_solaris;
```
thewilsonator, line 31: silly question: why `solaris` and not `elf_shared`?

src/core/internal/elf/dl.d
ibara: Done in latest (forced update, to keep commit squashed)

src/rt/sections.d
ibara: OpenBSD doesn't support everything you're looking for in elf_shared (e.g., no dlinfo, no RTLD_NOLOAD, no dlpi_tls_modid). Faking it caused problems. The Solaris approach worked.

thewilsonator: In which case can you add a comment here and to the top of ...

src/rt/sections_solaris.d
```diff
@@ -10,7 +10,16 @@
 
 module rt.sections_solaris;
```
thewilsonator, line 11: the module documentation at the top of this file

src/rt/sections.d
ibara: Sure. Do you mean inline or at the top of this whole block? (I am new to D and its coding style, did my best to follow the conventions I saw.)

src/rt/sections_solaris.d
ibara: Sure.

ibara: Done.

PetarKirov (approved):
Looks good to me, nice work!

src/rt/sections.d
ibuclaw: elf_shared works on Solaris though.  On Solaris 11.0, the tls_modid hidden, but it is attached to the RTLD_DI_LINKMAP (you then have to internally define the private Rt_map and cast the pointer to it).  Solaris 11.4 and later the tls_modid has since been made public.

ibara: I noticed that with the druntime library you have in GDC. I still am unsure how to operationalize that information though.

kinke:
What about backtraces? Without touching https://github.com/dlang/druntime/blob/master/src/core/internal/execinfo.d or having the compiler predefine `DRuntime_Use_Libunwind`, there shouldn't be any...

ibara:
> What about backtraces? Without touching https://github.com/dlang/druntime/blob/master/src/core/internal/execinfo.d or having the compiler predefine `DRuntime_Use_Libunwind`, there shouldn't be any...

I see. Ideally we would use Unwind. But on OpenBSD the Unwind symbols are in libc++abi, so I would have to look into this a little more.

kinke:
Don't let that stop you; I've just mentioned it because backtraces are quite helpful when ironing out the last issues. :)

On FreeBSD, LDC's druntime currently depends on libexecinfo and adds the lib dependency as pragma supported by the lld linker: https://github.com/ldc-developers/druntime/blob/aba5e0f2d5729c354ceb95e1cf30d533afc185de/src/core/internal/execinfo.d#L50-L51

llvm-libunwind supported is brand-new: https://dlang.org/changelog/2.096.0.html#libunwind

ibara:
> Don't let that stop you / this PR; I've just mentioned it because backtraces are quite helpful when ironing out the last issues. :)

Agreed.

> On FreeBSD, LDC's druntime currently depends on libexecinfo and adds the lib dependency as pragma supported by the lld linker: https://github.com/ldc-developers/druntime/blob/aba5e0f2d5729c354ceb95e1cf30d533afc185de/src/core/internal/execinfo.d#L50-L51

This wouldn't work out so well for us I don't think. It would be very preferable for us to use Unwind. Even if that means backporting, that's better than using libexecinfo.

RazvanN7:
Please re-push this to retrigger the stdcpp test.

RazvanN7 (approved):

ibara:
It appears that the auto-tester CI got stuck. I don't imagine it takes 10 days for completion. Let me know what I can do.

ibuclaw:
@ibara you didn't rebase against master, as you can see in the new pipeline failures.

ibara:
I see. Will fix.

ibuclaw:
@ibara any progress?  Time is becoming sensitive now as we gear up for the double release at the beginning of May (DMD 2.097.0 and GDC 11)

ibara:
See #3443

PR #3395, opened 2021-03-14
fix issue 21712 - [REG 2.096.0] sometimes coverage *.lst files are corrupted

TungstenHeart:
dont use the same buffer to read+split the LST and read+split the D. (reg introduced in #3341)

### DIFF
--- a/src/rt/cover.d
+++ b/src/rt/cover.d
@@ -250,9 +250,10 @@ shared static ~this()
         foreach (i; 0 .. minLineLength)
             lines[i] = expandTabs(lines[i]);
 
-        if (config.merge && readFile(flst, buf))
+        auto buf2 = new char[NUMCHARS];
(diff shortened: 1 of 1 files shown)


thewilsonator:
Is there a test case for this regression?

TungstenHeart:
Sorry but no. So far the bug is only observed in relatively big projects (>=10K) and each time I tried to isolate an affected module so problem went away. For example core/internal/execinfo.d is affected but if I copy the file, add a main to activate dmd_merge_coverage and run it twice, then the lst looks ok.

However note that the codecov upload for this PR does not contain any corrupted lst and also that the coverage for the all project increases by 2%, which is quie a sign that the fix is good (just like the decrease by 3% causes by #3341 was a sign that something was wrong).

thewilsonator (approved):

thewilsonator:
please force push to restart CI

PR #3397, opened 2021-03-15
Fix Issue 21015 - aa.byKeyValue, byKey provide mutable reference acce…

wolframw:
…ss to keys

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2789,7 +2789,7 @@ auto byKey(T : V[K], K, V)(T aa) pure nothrow @nogc @safe
         @property bool empty()  @safe { return _aaRangeEmpty(r); }
         @property ref front() @trusted
         {
-            return *cast(substInout!K*) _aaRangeFrontKey(r);
+            return *cast(substInout!(const(K*))) _aaRangeFrontKey(r);
         }
         void popFront() @safe { _aaRangePopFront(r); }
         @property Result save() { return this; }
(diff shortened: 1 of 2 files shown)


thewilsonator (approved):

RazvanN7:
This change appears to be breaking phobos.

wolframw:
> This change appears to be breaking phobos.

The way I see it, it's not really Phobos that breaks. Phobos does what it should: `array` takes a range with const elements and returns an array of const elements, and `sort` errors because it can't sort a range whose elements are neither swappable nor assignable.

For code relying on `byKey` returning non-const keys (like dub), the solution would be to use `keys` instead, which returns a mutable copy of the AA's keys. I can make a PR taking care of such occurences in dub (I already made the changes locally to see if it compiles anyway). However, considering the breakage, there's probably not a good chance this PR will make the cut, right?

n8sh:
All that is logically necessary is that the keys' *hash codes* do not change. Requiring references to keys to be `const` in the D sense or even in the less strict C++ sense is nice because the compiler can check it automatically but it also outlaws a great deal of correct code.

wolframw:
> All that is logically necessary is that the keys' _hash codes_ do not change.

Perhaps I misunderstand, but that is not enough because lookup fails if keys diverge from their hash code:
```d
void main() {
    import std.stdio;

    int[string] aa = ["one": 1,
                      "two": 2];

    foreach (kv; aa.byKeyValue)
        kv.key = "three";

    writeln(aa["one"]);    // throws RangeError
    writeln(aa["two"]);    // throws RangeError
    writeln(aa["three"]);  // throws RangeError
}
```

n8sh:
>lookup fails if keys diverge from their hash code

That is what I meant to communicate: the key must not change in a way that would alter the result of calculating its hash code.

RazvanN7:
@wolframw any progress on this?

PR #3399, opened 2021-03-15
core.lifetime: Tag -betterC compatible unittests

MoonlightSentinel:
To avoid future regressions w.r.t. to -betterC.

CC @kinke

### DIFF
--- a/src/core/lifetime.d
+++ b/src/core/lifetime.d
@@ -1,5 +1,7 @@
 module core.lifetime;
 
+import core.internal.attributes : betterC;
+
 // emplace
 /**
 Given a pointer `chunk` to uninitialized memory (but already typed
@@ -17,6 +19,7 @@ T* emplace(T)(T* chunk) @safe pure nothrow
 }
 
 ///
+@betterC
 @system unittest
 {
     static struct S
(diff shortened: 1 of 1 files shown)


src/core/lifetime.d
```diff
@@ -2161,3 +2195,31 @@ pure nothrow @nogc @system unittest
...
+//         A a;
+//         this(int i)
+//         {
+//             this.a = A(i);
+//         }
+//     }
+
+//     B b = void;
+//     emplace(b, 1);
+// }
```
PetarKirov, line 2225: If you want to add this unittest, but it doesn't work yet, I suggest placing the part of it that fails under `version (none)` (that way the compiler will do at least a minimal checking of the syntax) and a FIXME comment.

PetarKirov (commented):
Minor comment, otherwise LGTM.

src/core/lifetime.d
MoonlightSentinel: Removed.

PetarKirov (approved):

MoonlightSentinel:
Rebased because the CI got stuck...

PR #3400, opened 2021-03-19
Platform-independant definitions out of version blocks in core.stdc.stdint

dukc:


thewilsonator:
```
src/core/stdc/stdint.d:81: 
posix.mak:480: recipe for target 'checkwhitespace' failed
```

src/core/stdc/stdint.d
```diff
@@ -45,23 +61,6 @@ static if (is(ucent))
...
 version (Windows)
 {
-    alias int8_t   = byte;   ///
-    alias int16_t  = short;  ///
-    alias uint8_t  = ubyte;  ///
-    alias uint16_t = ushort; ///
-    version (CRuntime_DigitalMars)
-    {
-        alias int32_t  = cpp_long;  ///
-        alias uint32_t = cpp_ulong; ///
```
thewilsonator, line 55: `cpp_long` and friends have special mangling so these will have to remain I think

src/core/stdc/stdint.d
```diff
@@ -80,22 +79,11 @@ version (Windows)
...
 else version (Darwin)
 {
-    alias int8_t   = byte;          ///
-    alias int16_t  = short;         ///
-    alias uint8_t  = ubyte;         ///
-    alias uint16_t = ushort;        ///
-    alias int32_t  = int;           ///
-    alias uint32_t = uint;          ///
-    alias int64_t  = cpp_longlong;  ///
-    alias uint64_t = cpp_ulonglong; ///
```
thewilsonator, line 97: ditto

thewilsonator (commented):
but `[u]int8_t` and 16 should be fine

src/core/stdc/stdint.d
dukc: Ouch, totally forgot that. I was in the C world forgetting C++ considerations.

thewilsonator (approved):

kinke (commented):
Sorry, doesn't look like an improvement to me.

dukc:
> Sorry, doesn't look like an improvement to me.

@kinke this PR should in principle be an improvement in three separate ways:

1. What the title says - no need to define (u)int8_t nor (u)int16_t for each platform.
2. Comments added on why aliases to the larger types need to be platform-specific. At least for me, that was not immediately obvious.
3. The file should now keep working if c++ mangling of D `long` starts to obey the spec - see my last comment.

Which one(s) of the above are not an improvement, and why?

kinke:
These aliases are all tested to work correctly - with an actual C++ mangling test for the compiler using this druntime file: https://github.com/dlang/dmd/blob/master/test/runnable_cxx/stdint.d. So defining them as aliases to other aliases defined in another file is definitely no improvement to me. I'm not sure what part of the spec you are referring to, but the C++ mangling of D long will most likely not change - it has shortly for DMD 2.079-2.081 IIRC, but broke size_t interop across C++ and D on macOS etc., and was a pretty big disaster. See https://dlang.org/changelog/2.082.0.html#osx64_long_cppmangling.

Extracting the 4 really independent aliases is IMO hardly worth it but might be an improvement.

dukc:
I'm referring to https://dlang.org/spec/cpp_interface.html#data-type-compatibility . D `long` should mangle to C++ `long long`, but on 64-bit POSIX (judging by `core.config`) it currently mangles to C++ `long`. Bit length is correct but mangling still wrong. Thus updating the compiler to follow the spec would break this file if it aliases POSIX stuff to `long`. Well, I did replace the Windows `long` aliases too, which is technically needless but I decided to be consistent.

kinke:
> Thus updating the compiler to follow the spec would break this file if it aliases POSIX stuff to long.

This is extremely unlikely to happen, and the spec should be fixed instead.

dukc:
> This is extremely unlikely to happen, and the spec should be fixed instead.

Perhaps, but as long as it reads the current way I think DRuntime should not rely on compiler violating the spec. After the spec is updated, this file can be changed back if it feels better that way.

kinke:
We'll just have to disagree. ;)

dukc:
So let someone else decide.

To anyone: Change 64-bit definitions back to `long` or no?

RazvanN7:
@dukc since nobody else entered this discussion, we can go with @kinke 's assessment since he is the biggest authority in this context.

dukc:
I was hoping that my spec PR would get merged or rejected, but I have waited for too long. So yeah, let's do what Kinke suggested.

RazvanN7:
OK to close this?

RazvanN7 (approved):

ibuclaw:
@RazvanN7 what have you done?

dukc:
@ibuclaw He woke me up and told me how to proceed with the stalemate this PR was in. And then approved and merged the pr after I did what he suggested. How so?

RazvanN7:
@ibuclaw This PR simply extracts the 4 independent aliases as to reduce code duplication and documents the behavior. Are there any issues with that?

ibuclaw:
This PR caused a regression. https://issues.dlang.org/show_bug.cgi?id=22834

dukc:
Sorry, can't follow - how? Solaris is a Posix system, `int8_t` was `byte` for it both before and after this PR, or..?

PR #3401, opened 2021-03-20
Fix Issue 21014 - aa.byKeyValue, byKey, byValue very under-documented

wolframw:
This PR aims to improve the documentation and the examples of the AA-specific functions `byKeyValue`, `byKey`, `byValue`, `keys`, and `values`.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2633,7 +2633,7 @@ void clear(Value, Key)(Value[Key] aa)
     _aaClear(*cast(AA *) &aa);
 }
 
-/* ditto */
+/** ditto */
 void clear(Value, Key)(Value[Key]* aa)
 {
     _aaClear(*cast(AA *) aa);
@@ -2680,21 +2680,21 @@ T rehash(T : Value[Key], Value, Key)(T aa)
     return aa;
 }
 
-/* ditto */
+/** ditto */
 T rehash(T : Value[Key], Value, Key)(T* aa)
 {
     _aaRehash(cast(AA*)aa, typeid(Value[Key]));
     return *aa;
 }
 
-/* ditto */
+/** ditto */
 T rehash(T : shared Value[Key], Value, Key)(T aa)
 {
     _aaRehash(cast(AA*)&aa, typeid(Value[Key]));
     return aa;
 }
 
-/* ditto */
+/** ditto */
 T rehash(T : shared Value[Key], Value, Key)(T* aa)
 {
     _aaRehash(cast(AA*)aa, typeid(Value[Key]));
@@ -2702,8 +2702,8 @@ T rehash(T : shared Value[Key], Value, Key)(T* aa)
 }
 
 /***********************************
- * Create a new associative array of the same size and copy the contents of the
- * associative array into it.
+ * Creates a new associative array of the same size and copies the contents of
+ * the associative array into it.
  * Params:
  *      aa =     The associative array.
  */
@@ -2744,7 +2744,7 @@ V[K] dup(T : V[K], K, V)(T aa)
     return result;
 }
 
-/* ditto */
+/** ditto */
 V[K] dup(T : V[K], K, V)(T* aa)
 {
     return (*aa).dup;
@@ -2771,11 +2771,27 @@ private AARange _aaToRange(T: V[K], K, V)(ref T aa) pure nothrow @nogc @safe
 }
 
 /***********************************
- * Returns a forward range over the keys of the associative array.
+ * Returns a $(REF_ALTTEXT forward range, isForwardRange, std,range,primitives)
+ * which will iterate over the keys of the associative array. The keys are
+ * returned by reference.
+ *
+ * If structural changes are made to the array (removing or adding keys), all
+ * ranges previously obtained through this function are invalidated. The
+ * following example program will dereference a null pointer:
+ *
+ *---
+ * import std.stdio : writeln;
+ *
+ * auto dict = ["k1": 1, "k2": 2];
+ * auto keyRange = dict.byKey;
+ * dict.clear;
+ * writeln(keyRange.front);
+ *---
+ *
  * Params:
  *      aa =     The associative array.
  * Returns:
- *      A forward range.
+ *      A forward range over the keys of the associative array.
  */
 auto byKey(T : V[K], K, V)(T aa) pure nothrow @nogc @safe
 {
@@ -2798,7 +2814,7 @@ auto byKey(T : V[K], K, V)(T aa) pure nothrow @nogc @safe
     return Result(_aaToRange(aa));
 }
 
-/* ditto */
+/** ditto */
 auto byKey(T : V[K], K, V)(T* aa) pure nothrow @nogc
 {
     return (*aa).byKey();
@@ -2807,7 +2823,7 @@ auto byKey(T : V[K], K, V)(T* aa) pure nothrow @nogc
 ///
 @safe unittest
 {
-    auto dict = [1: 0, 2: 0];
+    auto dict = [1: "v1", 2: "v2"];
     int sum;
     foreach (v; dict.byKey)
         sum += v;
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -2771,11 +2771,28 @@ private AARange _aaToRange(T: V[K], K, V)(ref T aa) pure nothrow @nogc @safe
...
 
 /***********************************
- * Returns a forward range over the keys of the associative array.
+ * Returns a $(REF_ALTTEXT forward range, isForwardRange, std,range,primitives)
+ * which will iterate over the keys of the associative array. The keys are
+ * returned by reference.
+ *
+ * If changes are made to the associative array, all ranges previously obtained
+ * through this function are invalidated. The following example program will
+ * crash:
```
schveiguy, line 2780: Not exactly true. If you change just a value, the iteration will be fine.

Maybe say "If structural changes are made to the associative array (removing or adding keys), all ranges..."

This applies to all the other locations where you have this message.

I would also change "crash" to "dereference a null pointer" to clarify the behavior is not unsafe.

src/object.d
```diff
@@ -2807,20 +2824,37 @@ auto byKey(T : V[K], K, V)(T* aa) pure nothrow @nogc
 ///
 @safe unittest
 {
-    auto dict = [1: 0, 2: 0];
-    int sum;
+    auto dict = [1: "v1", 2: "v2"];
+    int sum = 0;
```
schveiguy, line 2828: Unclear why this change is made.

src/object.d
```diff
@@ -2914,21 +2973,25 @@ auto byKeyValue(T : V[K], K, V)(T* aa) pure nothrow @nogc
...
 @safe unittest
 {
+    import std.stdio : writeln;
+
     auto dict = ["k1": 1, "k2": 2];
-    int sum;
+    int sum = 0;
     foreach (e; dict.byKeyValue)
+    {
+        writeln(e.key, ": ", e.value);
```
schveiguy, line 2982: No writeln in unittests. Use assert. It will replace the call with writeln in the docs.

Maybe `assert(e.key[1] == e.value + '0')`

src/object.d
```diff
@@ -2914,21 +2973,25 @@ auto byKeyValue(T : V[K], K, V)(T* aa) pure nothrow @nogc
...
         sum += e.value;
+    }
 
     assert(sum == 3);
 }
 
 /***********************************
- * Returns a dynamic array, the elements of which are the keys in the
- * associative array.
+ * Copies the keys of the associative array into a dynamic array and returns it.
```
schveiguy, line 2990: A little awkward wording. Maybe:

"returns a newly allocated dynamic array containing a copy of the keys from the associative array"

src/object.d
```diff
@@ -2975,12 +3038,12 @@ Key[] keys(T : Value[Key], Value, Key)(T *aa) @property
 }
 
 /***********************************
- * Returns a dynamic array, the elements of which are the values in the
- * associative array.
+ * Copies the values of the associative array into a dynamic array and returns
+ * it.
```
schveiguy, line 3042: Similar treatment.

schveiguy (changes requested):
Overall looks pretty good. Thanks for tackling this!

schveiguy:
FYI the reason the CI is failing is due to the import of std.stdio in the unittests.

src/object.d
wolframw: In the other examples, the keys and values are of different types and it's the values being added up. When I looked at this example, it honestly took me a bit of time to figure out that it's the keys being added up, so I thought having different types for keys and values might be an improvement.

About the initialization of `sum`: I was under the impression that using uninitialized variables was a bad practice, even in D land. I was wrong.

src/object.d
```diff
@@ -2771,11 +2771,27 @@ private AARange _aaToRange(T: V[K], K, V)(ref T aa) pure nothrow @nogc @safe
...
+ * ranges previously obtained through this function are invalidated. The
+ * following example program will dereference a null pointer:
+ *
+ *---
+ * import std.stdio : writeln;
+ *
+ * auto dict = ["k1": 1, "k2": 2];
+ * auto keyRange = dict.byKey;
+ * dict.clear;
+ * writeln(keyRange.front);
```
RazvanN7, line 2788: Maybe add a comment to this line that says "Segmentation fault"

src/object.d
```diff
@@ -2771,11 +2771,27 @@ private AARange _aaToRange(T: V[K], K, V)(ref T aa) pure nothrow @nogc @safe
...
+ *---
+ *
  * Params:
  *      aa =     The associative array.
  * Returns:
- *      A forward range.
+ *      A forward range over the keys of the associative array.
```
RazvanN7, line 2794:
```suggestion
 *      A forward range containing the keys of the associative array.
```

src/object.d
```diff
@@ -2816,11 +2832,27 @@ auto byKey(T : V[K], K, V)(T* aa) pure nothrow @nogc
...
+ * ranges previously obtained through this function are invalidated. The
+ * following example program will dereference a null pointer:
+ *
+ *---
+ * import std.stdio : writeln;
+ *
+ * auto dict = ["k1": 1, "k2": 2];
+ * auto valueRange = dict.byValue;
+ * dict.clear;
+ * writeln(valueRange.front);
```
RazvanN7, line 2849: ditto

src/object.d
```diff
@@ -2816,11 +2832,27 @@ auto byKey(T : V[K], K, V)(T* aa) pure nothrow @nogc
...
+ *---
+ *
  * Params:
  *      aa =     The associative array.
  * Returns:
- *      A forward range.
+ *      A forward range over the values of the associative array.
```
RazvanN7, line 2855: Same as above

src/object.d
```diff
@@ -2861,11 +2893,35 @@ auto byValue(T : V[K], K, V)(T* aa) pure nothrow @nogc
...
+ * ranges previously obtained through this function are invalidated. The
+ * following example program will dereference a null pointer:
+ *
+ *---
+ * import std.stdio : writeln;
+ *
+ * auto dict = ["k1": 1, "k2": 2];
+ * auto kvRange = dict.byKeyValue;
+ * dict.clear;
+ * writeln(kvRange.front.key, ": ", kvRange.front.value);
```
RazvanN7, line 2912: I think that this is a bit misleading. The segfault will occur because of `front`. Does it really matter if we access `key` or `value`?

src/object.d
```diff
@@ -2861,11 +2893,35 @@ auto byValue(T : V[K], K, V)(T* aa) pure nothrow @nogc
...
+ * array and is not compatible withth the
+ * $(LINK2 $(ROOT_DIR)phobos/std_typecons.html#.Tuple,`Tuple`) type in Phobos.
+ * For compatibility with `Tuple`, use
+ * $(LINK2 $(ROOT_DIR)phobos/std_array.html#.byPair,std.array.byPair) instead.
+ *
  * Params:
  *      aa =     The associative array.
  * Returns:
- *      A forward range.
+ *      A forward range over the pairs of the associative array.
```
RazvanN7, line 2924:
```suggestion
 *      A forward range containing the pairs of the associative array.
```

RazvanN7: Or maybe it would be better to say "A forward range referencing ..."

RazvanN7: This goes for all of the above and below similar changes

RazvanN7:
@schveiguy Are we good with this PR?

RazvanN7:
ping @wolframw

RazvanN7:
@wolframw are you still pursuing this PR?

RazvanN7:
@wolframw Please apply the requested changes so that we can merge this PR

RazvanN7:
Closing due to inactivity. @wolframw feel free to reopen and apply the changes.

wolframw:
> Closing due to inactivity. @wolframw feel free to reopen and apply the changes.

Sorry for the inactivity. I'd like to continue with this (in fact, it's already committed and pushed to my fork) but I don't have the rights to reopen this PR. Could you do that for me, please?

src/object.d
wolframw: `kvRange.front` will return `Pair(null, null)`, so I explicitly access `key` and `value` to make it blow up.

thewilsonator:
> Could you do that for me, please?

No, it says  "the `fix_21014`branch was force-pushed or recreated". 

Just open another PR and link to this one.

PR #3402, opened 2021-03-22
Add pledge and unveil to core/sys/posix/unistd.d for OpenBSD

ibara:
Hello --

It was pointed out to me on the dlang forums that this is the correct location to add bindings for pledge and unveil for OpenBSD. Both are found in `<unistd.h>`.

These two facilities are expected to always be present on an OpenBSD system. They are parts of OpenBSD's sandboxing strategy.

### DIFF
--- a/src/core/sys/posix/unistd.d
+++ b/src/core/sys/posix/unistd.d
@@ -2566,6 +2566,7 @@ else version (OpenBSD)
     int        lchown(const scope char*, uid_t, gid_t);
     int        lockf(int, int, off_t) @trusted;
     int        nice(int) @trusted;
+    int        pledge(const scope char*, const scope char*);
     ssize_t    pread(int, void*, size_t, off_t);
     ssize_t    pwrite(int, const scope void*, size_t, off_t);
     int        setpgrp(pid_t, pid_t) @trusted;
@@ -2575,6 +2576,7 @@ else version (OpenBSD)
     void       sync() @trusted;
     int        truncate(const scope char*, off_t);
     useconds_t ualarm(useconds_t, useconds_t) @trusted;
+    int        unveil(const scope char*, const scope char*);
     int        usleep(useconds_t) @trusted;
     pid_t      vfork();
 }


thewilsonator (approved):

PetarKirov:
@ibara thanks for the addition! I have one suggestion, which I think @thewilsonator did not consider with his review as historically we haven't been very strict about it (and there are probably many exceptions that we have to fix): 

**TL;DR** Platform specific (non-portable) bindings should be placed in `core.sys.{platform}.*`, rather than `core.stdc.*`, `core.stdcpp.*` or `core.sys.posix.*` (as you did with this PR).  In this case, as far as I understand, `pledge` and `unveil` are APIs unique to OpenBSD, so they should be placed in `core.sys.openbsd.unistd`.

---

The idea being that if a program imports:
1. only `std.*` (Phobos) and/or `core.*` (Druntime), but not `core.stdc.*`, `core.stdcpp.*`, or `core.sys.*`
2. only `core.stdc` (lib C bindings) -> it should on all platforms with libc for which D has complete (and correct) declarations 
(*BSD, Linux (including various libc implementations), Windows, macOS)
3. only `core.stdcpp` (C++ `std::*` bindings) and `core.stdc`
4. only `core.sys.posix`

It should be portable (not requiring explicit use of conditional compilation on the programmer's side) to various extents:

#### Option 1.
A program importing only Phobos and Druntime does not need to to use any form of conditional compilation e.g. `version (OpenBSD) { ... } else version (linux) { ... } else version (Windows) { ... } ...` in order to be portable across all fully supported by D platforms, as conditional compilation is already handled for the user by these libraries.

#### Option 2.
If a program does not wish to link Druntime and Phobos (or they're not yet supported for the target platform), the next best option is to import solely `core.stdc.*`. Ideally importing `core.stdc.*` will only make visible symbols that are the common libc subset of all supported platforms by the D compiler - e.g. *BSD, Windows, Linux (glibc, musl libc, etc.).

#### Option 3.
In the future, when D's `extern (C++)` support and corresponding bindings improve sufficiently, importing only `core.stdcpp` could also be a good alternative to `core.stdc.*`. Though, in theory, this would limit number of platforms a program can target, as presumably not all targets with libc support also have have standards compliant C++ support).

#### Option 4.
If a program does not link druntime and only needs to target Posix systems, then importing solely `core.sys.posix` should be portable for all Posix-complaint OSes - Linux, *BSD, Solaris, macOS. Of course each of those OSes support a different subset of Posix (and its various editions), and so `core.sys.posix` should only make visible the lowest common denominator.

ibara:
Sure. Whatever makes the most sense to the D community makes sense to me.
It wasn't @thewilsonator who provided the location suggestion, but someone named Johan on the forums. In any event, no resistance from me to move things.

PR #3404, opened 2021-03-22
core.sys.posix.unistd: Add non-standard environ variable

ibuclaw:
This variable is declared in unistd.h in glibc and musl.  Other platforms document it, but usually require you to declare it explicitly in the user program, or have it as a different name/macro.

[Phobos](https://github.com/dlang/phobos/blob/d8cde81d0a73cce66c57e81a9dbe8089dee21250/std/process.d#L157-L173) has some extern declarations for it, it should instead be importing druntime to get the symbol.

### DIFF
--- a/src/core/sys/posix/unistd.d
+++ b/src/core/sys/posix/unistd.d
@@ -2724,3 +2724,51 @@ else version (CRuntime_UClibc)
     int        truncate(const scope char*, off_t);
   }
 }
+
+// Non-standard definition to access user process environment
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):
Changelog?

PetarKirov (approved):

PetarKirov:
> Changelog?

We don't require a changelog entry for sys bindings, unless the change is substantial, so I think there's no need in this case.

PR #3405, opened 2021-03-22
core.sys.windows.malloc: New bindings module for malloc.h

ibuclaw:
Only added malloc-related functions and not any of the heapwalk stuff - someone can add it later if they require it.

This covers extern(C) functions that are currently declared in [Phobos](https://github.com/dlang/phobos/blob/1c77cdad29161f8bdb84475ffa52c22e7745a883/std/experimental/allocator/mallocator.d#L199-L206).

@adamdruppe `export` should be OK here?
```
0000000000000000 T _aligned_recalloc
0000000000000000 I __imp__aligned_recalloc
0000000000000000 T _aligned_realloc
0000000000000000 I __imp__aligned_realloc
0000000000000000 T _aligned_offset_recalloc
0000000000000000 I __imp__aligned_offset_recalloc
0000000000000000 T _aligned_offset_realloc
0000000000000000 I __imp__aligned_offset_realloc
0000000000000000 T _aligned_offset_malloc
0000000000000000 I __imp__aligned_offset_malloc
0000000000000000 T _aligned_msize
0000000000000000 I __imp__aligned_msize
0000000000000000 T _aligned_malloc
0000000000000000 I __imp__aligned_malloc
0000000000000000 T _aligned_free
0000000000000000 I __imp__aligned_free
```

### DIFF
--- a/mak/COPY
+++ b/mak/COPY
@@ -479,6 +479,7 @@ COPY=\
 	$(IMPDIR)\core\sys\windows\sspi.d \
 	$(IMPDIR)\core\sys\windows\stacktrace.d \
 	$(IMPDIR)\core\sys\windows\stat.d \
+	$(IMPDIR)\core\sys\windows\stdc\malloc.d \
 	$(IMPDIR)\core\sys\windows\stdc\time.d \
 	$(IMPDIR)\core\sys\windows\subauth.d \
 	$(IMPDIR)\core\sys\windows\threadaux.d \
--- a/mak/DOCS
+++ b/mak/DOCS
@@ -501,6 +501,7 @@ DOCS=\
 	$(DOCDIR)\core_sys_windows_wtsapi32.html \
 	$(DOCDIR)\core_sys_windows_wtypes.html \
 	\
+	$(DOCDIR)\core_sys_windows_stdc_malloc.html \
 	$(DOCDIR)\core_sys_windows_stdc_time.html \
 	\
 	$(DOCDIR)\core_thread_context.html \
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -480,6 +480,7 @@ SRCS=\
 	src\core\sys\windows\sspi.d \
 	src\core\sys\windows\stacktrace.d \
 	src\core\sys\windows\stat.d \
+	src\core\sys\windows\stdc\malloc.d \
 	src\core\sys\windows\stdc\time.d \
 	src\core\sys\windows\subauth.d \
 	src\core\sys\windows\threadaux.d \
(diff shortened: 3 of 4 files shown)


PetarKirov (approved):
I compared with copy of `include/corecrt_malloc.h` that I found online and everything looks good to me.

CyberShadow:
Wait. Is this actually part of the Windows API? Isn't it just a binding to Microsoft's C runtime? In which case, it should be in another package?

PetarKirov:
@CyberShadow `src/core/sys/windows/stdc/malloc.d` seems good to me. What do you think?

CyberShadow:
Yes, I see that package already existed, I don't see why not :)

ibuclaw:
> @CyberShadow `src/core/sys/windows/stdc/malloc.d` seems good to me. What do you think?
> 
> Alternatively, it could be `ucrt`, instead of `stdc`.

Mingw declares this in both malloc.h and stdlib.h, so the latter location would have been the next logical place for it.

ibuclaw:
> Wait. Is this actually part of the Windows API? Isn't it just a binding to Microsoft's C runtime? In which case, it should be in another package?
> 
> If it's not something that would be available when targeting other C runtimes, such as the DigitalMars C runtime (i.e. the default on Windows), then I think it probably should be in `core.stdc.malloc` (and versioned accordingly).

By that logic, many declarations in core.sys.linux and others are probably in the wrong location too.

CyberShadow:
Hmm. Well, I understand that on Linux, the OS (i.e. kernel) does not provide any API directly to applications, unless we count special filesystems like `/dev` / `/proc` / `/sys`. All access to the kernel's features goes through the libC, so, effectively, the libC is the OS API. That's not how things are on Windows, where the OS API and the libC are very, very separate things. For a long time, Windows didn't even ship a libC, and even today I believe relying on the libC included with Windows (i.e. the Microsoft C runtime) is not recommended.

rainers:
The MS C runtime is distributed with the Windows SDK these days, not with the compiler (or as a separate install package). Its called "Universal C Runtime" (UCRT), so core.sys.windows.ucrt might be the most appropriate package name, but without other implementations that are bundled as tightly, core.sys.windows.stdc is close enough IMO.

ibuclaw:
> The MS C runtime is distributed with the Windows SDK these days, not with the compiler (or as a separate install package). Its called "Universal C Runtime" (UCRT), so core.sys.windows.ucrt might be the most appropriate package name, but without other implementations that are bundled as tightly, core.sys.windows.stdc is close enough IMO.

I've been getting a little acquainted with UCRT, only to find out that MinGW doesn't provide first-class support for it.  And even if it did, GCC automatically links to the MSVCRT library instead.

There'll be a few more Windows/MinGW-related bug fixes coming your way (mostly related to some I/O wrappers in core.stdc.stdio).

PR #3409, opened 2021-03-24
windows: Move rt_loadLibrary and rt_unloadLibrary to rt.sections

ibuclaw:
GDC defines its own version of these functions for shared library integration in `gcc.sections` (based off elf_shared), because phobos as a shared library should just work out of the box on Win32 and Win64.

As rt_loadLibrary and rt_unloadLibrary is define here for Posix targets, moving here as well for Windows to avoid symbol conflicts.

### DIFF
--- a/src/rt/dmain2.d
+++ b/src/rt/dmain2.d
@@ -26,7 +26,7 @@ version (Windows)
     import core.sys.windows.basetsd : HANDLE;
     import core.sys.windows.shellapi : CommandLineToArgvW;
     import core.sys.windows.winbase : FreeLibrary, GetCommandLineW, GetProcAddress,
-        IsDebuggerPresent, LoadLibraryA, LoadLibraryW, LocalFree, WriteFile;
+        IsDebuggerPresent, LoadLibraryW, LocalFree, WriteFile;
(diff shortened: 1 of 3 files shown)


rainers:
Would it make sense to introduce a sections_win_shared.d file similar to sections_elf_shared.d to avoid the code duplication?
On a side note: Please note that the sections_win32/sections_win64 module names are misleading: one is for the DigitalMars runtime/toolchain, the other for the MS runtime.

ibuclaw:
> Would it make sense to introduce a sections_win_shared.d file similar to sections_elf_shared.d to avoid the code duplication?

On my end, I've defined these functions in `gcc.sections.pecoff`.  It doesn't matter where these functions are defined, so long as it defined consistently.

kinke:
I've simply disabled these here for `version(Shared)`; the Posix implementations in `rt.section_elf_shared` are guarded by `version(Shared)` and can be easily extended to Windows:
https://github.com/ldc-developers/druntime/pull/197/files#diff-d80ffb0ff5350734a54a14c3907c1b8052a9e730a2b38c6d4a12a45943440297

RazvanN7 (approved):

PR #3411, opened 2021-03-25
Fix 21764 - checkaction=context doesn't work for empty tuples

MoonlightSentinel:
The template constraint was intended to avoid collisions with the
unary hook but also prevented instances for zero elements.

`combine` was adapted to print empty arrays as `()`.

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -60,7 +60,7 @@ template _d_assert_fail(A...)
 {
     string _d_assert_fail(B...)(
         const scope string comp, auto ref const scope A a, auto ref const scope B b)
-    if (B.length > 0)
+    if (B.length != 0 || A.length != 1) // Resolve ambiguity with unary overload
     {
         string[A.length + B.length] vals;
         static foreach (idx; 0 .. A.length)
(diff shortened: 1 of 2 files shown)


src/core/internal/dassert.d
```diff
@@ -60,7 +60,7 @@ template _d_assert_fail(A...)
 {
     string _d_assert_fail(B...)(
         const scope string comp, auto ref const scope A a, auto ref const scope B b)
-    if (B.length > 0)
+    if (B.length > 0 || !A.length)
```
Geod24, line 63: Doesn't that mean `assert(NonEmptyTuple == EmptyTuple)` will not work ?

MoonlightSentinel: Fixed. But is this even possible to compare tuples with different lengths?

Geod24: > onlineapp.d(6): Error: mismatched tuple lengths, 3 and 1

Guess not :D

thewilsonator (approved):

ibuclaw:
This is causing all pipelines to fail in dmd master.

MoonlightSentinel:
That weird. Seems like the CI's used an outdated ref?

PR #3414, opened 2021-03-26
Begin OpenBSD unistd.d extension module with plegde and unveil.

ibara:
From @PetarKirov's comment in https://github.com/dlang/druntime/pull/3402, create a new module `core/sys/openbsd/unistd.d` and put pledge and unveil in there.


PetarKirov (approved):
Thanks @ibara for following through! Looks good to me.

PR #3417, opened 2021-03-28
insure against flow analysis changes in DMD PR 12311

WalterBright:
This will work regardless of how dmd detects unreachable code. Blocking https://github.com/dlang/dmd/pull/12311

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2168,14 +2168,22 @@ const:
      */
     @property string name() nothrow pure @nogc
     {
-        if (true || flags & MIname) // always available for now
+        import core.stdc.string : strlen;
+        static if (1) // always available for now
         {
-            import core.stdc.string : strlen;
-
             auto p = cast(immutable char*)addrOf(MIname);
             return p[0 .. strlen(p)];
         }
-        // return null;
+        else
+        {
+            if (flags & MIname)
+            {
+                auto p = cast(immutable char*)addrOf(MIname);
+                return p[0 .. strlen(p)];
+            }
+            else
+                return null;
+        }
     }
 
     static int opApply(scope int delegate(ModuleInfo*) dg)


src/object.d
```diff
@@ -2168,14 +2168,22 @@ const:
      */
     @property string name() nothrow pure @nogc
     {
-        if (true || flags & MIname) // always available for now
+        import core.stdc.string : strlen;
+        static if (1) // always available for now
```
ibuclaw, line 2172: And why can't we just remove the `if (1)`?  The assumption that this information is always available is naive at best.

WalterBright: Because this PR isn't about changing the behavior of the code, it's clarifying what it does and making it compatible with the compiler's flow analysis.

Changing the behavior is an entirely different thing.

Geod24: Removing an always-true `if` does not change the behavior of the code.

Geod24: Ping @WalterBright: Can you address the reviews so we can merge this ?

WalterBright: You're asking me to assume the `true` is not needed, that `flags` is reliably set. That is beyond the scope of this PR.

Geod24: That assumption is *already* there. The code generated before and after removing the `if` will have *exactly the same behavior*. I don't see why you insist on this particular form which just adds dead code.

WalterBright: You want to change `if(true || flags)` to `if (flags)`. This is *different*.

Geod24: @WalterBright : My suggestion is:
```D
    @property string name() nothrow pure @nogc
    {
        auto p = cast(immutable char*)addrOf(MIname);
        return p[0 .. strlen(p)];
    }
```

Which is what the code is currently doing.

RazvanN7: @Geod24 @WalterBright I think that there was a confusion here. @ibuclaw requested that the `true` part of the `if` is removed whereas @Geod24 requested that the `if` is discarded completely. Maybe the resolution is to discard the `if` but add a comment detailing in what circumstances should the `flag` condition be added.

RazvanN7:
ping @WalterBright

src/object.d
ibuclaw: > @Geod24 @WalterBright I think that there was a confusion here. @ibuclaw requested that the `true` part of the `if` is removed whereas @Geod24 requested that the `if` is discarded completely. Maybe the resolution is to discard the `if` but add a comment detailing in what circumstances should the `flag` condition be added.

Or raise a new PR making the if explicit, because it allows this to be a PAYG feature, the same as all other getter methods for minfo, and avoids segfault when the compiler/user decides to omit giving a module a name at runtime.

RazvanN7:
@WalterBright what is your final decision on this? Should we merge as is or are you willing to apply the suggestions? We should be moving forward with this as to unblock https://github.com/dlang/dmd/pull/12311

Geod24:
Superseded by #3452

PR #3421, opened 2021-04-01
Fix Issue 21789: Let umask prune permissions of coverage lst files

omerfirmak:


thewilsonator:
Please update the commit message to say "Fix issue 21789: ..."

omerfirmak:
> Please update the commit message to say "Fix issue 21789: ..."

Done

thewilsonator:
Its hard to understand from the issue what the problem is . Also this should have a test case and target `master`

omerfirmak:
> Its hard to understand from the issue what the problem is . 

Which part exactly? I think the problem is pretty self-explanatory. 

> Also this should have a test case 

I can't think of a sane way to test this.

> and target `master`

sigh

src/rt/cover.d
```diff
@@ -459,7 +459,8 @@ FILE* openOrCreateFile(string name)
     version (Windows)
         immutable fd = _wopen(toUTF16z(name), _O_RDWR | _O_CREAT | _O_BINARY, _S_IREAD | _S_IWRITE);
     else
-        immutable fd = open((name ~ '\0').ptr, O_RDWR | O_CREAT, S_IRUSR | S_IWUSR);
+        immutable fd = open((name ~ '\0').ptr, O_RDWR | O_CREAT, S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP |
```
Geod24, line 462: I would say, just remove the 3rd argument ?

omerfirmak: `O_CREAT` requires them.

ibuclaw:
> > Its hard to understand from the issue what the problem is . 
> 
> Which part exactly? I think the problem is pretty self-explanatory. 
> 
> > Also this should have a test case 
> 
> I can't think of a sane way to test this.
> 

Get the umask in effect and check the output file matches what is expected? It can be done in a shell script. I'm sure there are examples of shell tests in the druntime testsuite.

omerfirmak:
> > > Its hard to understand from the issue what the problem is .
> > 
> > 
> > Which part exactly? I think the problem is pretty self-explanatory.
> > > Also this should have a test case
> > 
> > 
> > I can't think of a sane way to test this.
> 
> Get the umask in effect and check the output file matches what is expected? It can be done in a shell script. I'm sure there are examples of shell tests in the druntime testsuite.

So you want me to test the `open` syscall?

Geod24:
To add a bit on this issue:

1) Why are the coverage at mode 600 to begin with ?
The only thing I can think of is to "protect" the source code from accidental leaks, but for it to be a useful protection, the attacker would need a level of access which implies a lot of other, easier attack vectors.
For example for coverage to work, the source code needs to be accessible, so if one can access the directory in which the `.lst` files are located, one is also likely to have access to the source files, which are unlikely to be 600.

2) Why was this a problem ?
Our program runs its integration tests inside of Docker, and creates those files as root, which means that later on the coverage upload fails (we use Github CI). There are workarounds for this, but it would propagate everywhere, for a problem which shouldn't be there in the first place. One of the workaround is to pass arguments to Docker (`--user`), but that falls over as soon as you build a more complex pipeline (we spawn a full network using `docker-compose`, not calling `docker` directly). The workaround we came up with was to `sudo chmod` those files, and we actually have to do that in a loop because the list of arguments is so long it reaches the shell's limit. And again, why are we not following `umask` to begin with ?

Geod24 (approved):

PR #3424, opened 2021-04-02
[dmd-cxx] Add noreturn alias and more printf format fixes

ibuclaw:
Had this locally, but forgot to check-in.


ibuclaw:
Merging as obvious.

PR #3426, opened 2021-04-03
-checkaction=context: Mark context pointer of structs

MoonlightSentinel:
Alternative to omission implemented in #2720

### DIFF
--- a/test/exceptions/src/assert_fail.d
+++ b/test/exceptions/src/assert_fail.d
@@ -338,6 +338,22 @@ void testStructEquals6()
     test(t1, t3, "T(false, 0, 0, \"\", [], `null`) != T(false, 0, 0, \"\", [], C())");
 }
 
+void testContextPointer()
+{
+    int i;
+    struct T
+    {
+        int j;
+        int get()
+        {
+            return i * j;
+        }
+    }
+    T t = T(1);
+    t.tupleof[$-1] = cast(void*) 0xABCD; // Deterministic context pointer
+    test(t, t, `T(1, <context>: 0xABCD) != T(1, <context>: 0xABCD)`);
+}
+
 void main()
 {
     testIntegers();
@@ -361,5 +377,7 @@ void main()
     testStructEquals4();
     testStructEquals5();
     testStructEquals6();
+    testContextPointer();
+
     fprintf(stderr, "success.\n");
 }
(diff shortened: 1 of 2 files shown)


PetarKirov (approved):

test/exceptions/src/assert_fail.d
```diff
@@ -338,6 +338,26 @@ void testStructEquals6()
     test(t1, t3, "T(false, 0, 0, \"\", [], `null`) != T(false, 0, 0, \"\", [], C())");
 }
 
+void testStructOpEquals()
+{
+    struct T1 {
```
PetarKirov, line 343:
```suggestion
    struct T1
    {
```

test/exceptions/src/assert_fail.d
```diff
@@ -338,6 +338,26 @@ void testStructEquals6()
...
+        bool opEquals(T1) { return false; }
+    }
+
+    T1 t1;
+    test(t1, t1, "T1() != T1()");
+}
+
+void testStructOpEquals2()
+{
+    struct T1 {
```
PetarKirov, line 353:
```suggestion
    struct T1
    {
```

test/exceptions/src/assert_fail.d
PetarKirov, line 343:
```suggestion
    struct T1
    {
```

test/exceptions/src/assert_fail.d
PetarKirov, line 353:
```suggestion
    struct T1
    {
```

Geod24:
Can it happen that the struct equality fails only because of different ctx pointer ? In which case the error message would be quite confusing ?

MoonlightSentinel:
Actually yes, so it must not be omitted. 
Maybe this should add a marker for  the context ptr, e.g.
```d
S(2, 0x20714600010 <context>)
S(2, context: 0x20714600010)
???
```

Geod24:
The former sounds better as a struct could have a `context` member.
Alternatively, detecting the difference in stack pointer and printing it if it exists:
```
S(2) != S(3) // Different content, same ctx
Context pointer differ: S(2, <context>: 0x20714600010) != S(3, <context>: 0x20714600010) // Different or same content
```

MoonlightSentinel:
Additional messages would only work well for structs at the top level. Adding a prefix also works for embedded structs (e.g. `S[4]`).

Geod24 (approved):
LGTM

PR #3428, opened 2021-04-04
core.stdc.config: Fix magic __c_complex_* enums for Windows

kinke:
MSVC has `_{F,D,L}complex` typedefs in complex.h (see https://docs.microsoft.com/en-us/cpp/c-runtime-library/complex-math-support?view=msvc-160). `_Lcomplex` (for `__c_complex_real`) uses its 64-bit `long double`. And the alignments are regular (4/8/8 bytes).

### DIFF
--- a/src/core/stdc/config.d
+++ b/src/core/stdc/config.d
@@ -235,18 +235,9 @@ private struct _Complex(T)
     T im;
 }
 
-version (Posix)
-{
-    align(float.alignof)  enum __c_complex_float : _Complex!float;
-    align(double.alignof) enum __c_complex_double : _Complex!double;
-    align(real.alignof)   enum __c_complex_real : _Complex!real;
-}
-else
-{
-    align(float.sizeof * 2)  enum __c_complex_float : _Complex!float;
-    align(double.sizeof * 2) enum __c_complex_double : _Complex!double;
-    align(real.alignof)      enum __c_complex_real : _Complex!real;
-}
+enum __c_complex_float  : _Complex!float;
+enum __c_complex_double : _Complex!double;
+enum __c_complex_real   : _Complex!c_long_double;
 
 alias c_complex_float = __c_complex_float;
 alias c_complex_double = __c_complex_double;
--- a/src/core/stdc/config.d
+++ b/src/core/stdc/config.d
@@ -164,7 +164,18 @@ else version (Posix)
   }
 }
 
-version (CRuntime_Microsoft)
+version (GNU)
+    alias c_long_double = real;
+else version (LDC)
+    alias c_long_double = real; // 64-bit real for MSVC targets
+else version (SDC)
+{
+    version (X86)
+        alias c_long_double = real;
+    else version (X86_64)
+        alias c_long_double = real;
+}
+else version (CRuntime_Microsoft)
 {
     /* long double is 64 bits, not 80 bits, but is mangled differently
      * than double. To distinguish double from long double, create a wrapper to represent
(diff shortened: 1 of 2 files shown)


ibuclaw:
Is there a missing compiler fix?

kinke:
I was hoping you could tell me why on earth a test using `creal` etc., *not* any of these magic enums, fails after such a change.

kinke:
Ah, I missed https://github.com/dlang/druntime/pull/3379. So you assume `__c_complex_real` matches `creal`? DMD with its 80-bit `real` for MSVC targets breaks that assumption badly.

kinke:
I'm wondering if it's not too late yet to rename `c_complex_real` to `c_complex_long_double`...

src/rt/memset.d
```diff
@@ -100,10 +100,21 @@ real *_memset80(real *p, real value, size_t count)
...
+else
+{
+    private struct d_complex_real
+    {
+        real re;
+        real im;
+    }
+}
+
+d_complex_real *_memset160(d_complex_real *p, d_complex_real value, size_t count)
```
ibuclaw, line 114: Why add all the boilerplate for what is essentially just:
```suggestion
_Complex!real *_memset160(_Complex!real *p, _Complex!real value, size_t count)
```

kinke: `_Complex(T)` is private.

src/core/stdc/config.d
```diff
@@ -164,7 +164,9 @@ else version (Posix)
   }
 }
 
-version (CRuntime_Microsoft)
+version (LDC)
+    alias c_long_double = real; // 64-bit real for MSVC targets
+else version (CRuntime_Microsoft)
```
ibuclaw, line 169: Could you move all compiler version conditions above CRuntime_Microsoft?

src/rt/memset.d
ibuclaw: A `package(rt)` declaration of it could be copied to `rt.util.utility` with aliases for d_cfloat, d_cdouble, and d_creal.

kinke: A one-liner `private struct d_complex_real { real re, im; }` would be another alternative. I really don't want to extract this struct or template anywhere just to reference it in some DMD-specific file and save some lines.

src/core/stdc/config.d
kinke: I take it GDC will never have MSVC support, just MinGW on Windows?

src/rt/memset.d
ibuclaw: It'd be used by rt.typeinfo, so no, it won't be restricted to a dmd-specific file.

ibuclaw: You'll need to go through the enum interface anyway, otherwise it may not have the correct ABI.

kinke: > It'd be used by rt.typeinfo, so no, it won't be restricted to a dmd-specific file.

For DMD only, and only for about ~20 months until we can remove this crap after the deprecation is over.

> You'll need to go through the enum interface anyway, otherwise it may not have the correct ABI.

[DMD only again.] Ouch, I assumed it's all good because all tests still pass. - My motivation just sank to 0.

ibuclaw: >  My motivation just sank to 0.

I'll own this then, since I introduced this "regression".  Will leave you just to deal with the core/stdc side.

ibuclaw: Done, please rebase @kinke.

src/core/stdc/config.d
kinke: Done.

ibuclaw (approved):

PR #3418, opened 2021-03-29
Fix Issue 21784: Allow detached threads to be joined

omerfirmak:
This is useful when a thread terminates before the
caller is able to join(). With a reset `m_addr`, caller would segfault.

PR #3429, opened 2021-04-04
[stable] Allow detached threads to be joined on Windows too

kinke:
PR #3418 allowed them on Posix; extend it to Windows too. The added `test/thread/src/join_detach.d` test now passes on Windows (tested by LDC).


Geod24 (approved):
Sweet, thanks!

PR #3432, opened 2021-04-06
rt.util: Define D complex types, use them in rt instead of the C complex types

ibuclaw:
The `creal` interfaces in `rt.memset` and `rt.util.typeinfo` are not compatible with C complex reals, so define our own type internally distinct from `core.stdc.config`.

Based on @kinke's original work, and unblocks #3428.


kinke:
Thanks Iain.

PR #3435, opened 2021-04-14
Move __ctfeWrite to core.builtins

MoonlightSentinel:
Makes it easier to recognize `__ctfeWrite` as a builtin function and removes some clutter from `object.d` (especially if further builtins are added in the future)

Proposed in dlang/dmd#12412

### DIFF
--- a/src/core/builtins.d
+++ b/src/core/builtins.d
@@ -1,0 +1,13 @@
+/**********************************************
+ * This module implements common builtins for the D frontend.
+ *
+ * Copyright: Copyright © 2019, The D Language Foundation
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   Walter Bright
+ * Source:    $(DRUNTIMESRC core/builtins.d)
+ */
+
+module core.builtins;
+
+/// Writes `s` to `stderr` during CTFE (does nothing at runtime).
+void __ctfeWrite(scope const(char)[] s) @nogc @safe pure nothrow {}
--- a/src/object.d
+++ b/src/object.d
@@ -4506,4 +4506,5 @@ template _arrayOp(Args...)
     alias _arrayOp = arrayOp!Args;
 }
 
-void __ctfeWrite(scope const(char)[] s) @nogc @safe pure nothrow {}
+/// Legacy import
+public import core.builtins : __ctfeWrite;
--- a/mak/COPY
+++ b/mak/COPY
@@ -8,6 +8,7 @@ COPY=\
 	$(IMPDIR)\core\atomic.d \
 	$(IMPDIR)\core\attribute.d \
 	$(IMPDIR)\core\bitop.d \
+	$(IMPDIR)\core\builtins.d \
 	$(IMPDIR)\core\checkedint.d \
 	$(IMPDIR)\core\cpuid.d \
 	$(IMPDIR)\core\demangle.d \


src/core/builtins.d
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * Copyright: Copyright © 2019, The D Language Foundation
+ * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Authors:   Walter Bright
+ * Source:    $(DRUNTIMESRC core/builtins.d)
+ */
+
+module core.builtins;
+
+/// Writes `s` to `stdout` during CTFE (does nothing at runtime).
```
PetarKirov, line 12:
```suggestion
/// Writes `s` to `stderr` during CTFE (does nothing at runtime).
```

`pragma(msg)` writes to stderr, so I guess `__ctfeWrite` would so the same.

PetarKirov (approved):

src/core/builtins.d
PetarKirov: That said, it may actually be a good idea to separate things, so that `pragma(msg)` and `__ctfeWrite` would use `stdio`, while all other compiler output (error, warning and deprecation messages) would still go to `stderr`.

ibuclaw: You can't ever assume that (stdout is used to emit compiled code for `-pipe` in gcc)

MoonlightSentinel: The currently proposed implementation actually writes to `stdout`. I'll fix it.

src/object.d
```diff
@@ -4506,4 +4506,5 @@ template _arrayOp(Args...)
     alias _arrayOp = arrayOp!Args;
 }
 
-void __ctfeWrite(scope const(char)[] s) @nogc @safe pure nothrow {}
+/// Legacy import
+public import core.builtins : __ctfeWrite;
```
ibuclaw, line 4510: Should this be deprecated if legacy? Or I guess there is no need to punish users in this case.

src/core/builtins.d
ibuclaw: DMD is free to do what it wants so long as it's done in the non-common modules (builtins, compiler, ctfloat, errors, target, ...). At one point I introduced `stdmsg` to the compiler for just this reason (dmd and gdc emitted verbose and pragma messages to different streams).

src/object.d
RazvanN7: I think we should be fine without a deprecation in this case.

MoonlightSentinel: This should stay IMHO, It's nice to have `__ctfeWrite` available without having to search for the right module to import.

src/core/builtins.d
MoonlightSentinel: Nevertheless, it is better if the compilers are consistent in this regard (e.g. if tools want to use the output generated from `__ctfeWrite`).

MoonlightSentinel:
Auto-merge toggled on

src/object.d
PetarKirov: The import is a good idea, but the comment "legacy import" is not, as it implies that the import exist for backwards compatibility reasons (which is not true, since this wasn't implemented until now). Removing it (while keeping the import) will reduce confusion.

MoonlightSentinel: Fair enough. I wasn't sure whether this import would be deprecated or not when I wrote this comment.

=> See #3437

PR #3436, opened 2021-04-15
core.{attribute,builtins}: Publicly import GDC and LDC builtins and attributes.

ibuclaw:
Exposes a common module to import these compiler-specific modules - used occasionally throughout druntime and phobos - without the need to repeat `version (GNU)`/`version (LDC)` everywhere.

Ping @kinke for review/thoughts/approval/rejection of the idea.

### DIFF
--- a/src/core/attribute.d
+++ b/src/core/attribute.d
@@ -15,6 +15,12 @@
  */
 module core.attribute;
 
+version (GNU)
+    public import gcc.attributes;
+
+version (LDC)
+    public import ldc.attributes;
+
 version (D_ObjectiveC)
 {
     version = UdaOptional;
--- a/src/core/builtins.d
+++ b/src/core/builtins.d
@@ -9,5 +9,11 @@
 
 module core.builtins;
 
+version (GNU)
+    public import gcc.builtins;
+
+version (LDC)
+    public import ldc.intrinsics;
+
 /// Writes `s` to `stderr` during CTFE (does nothing at runtime).
 void __ctfeWrite(scope const(char)[] s) @nogc @safe pure nothrow {}


Geod24:
Looks good, waiting for @kinke 's opinion.

Geod24 (approved):

kinke:
This alone won't bring any convenience IMO, the interfaces would need to be checked for similarities and then streamlined towards identical naming and almost-identical semantics, otherwise we just end up with
```D
version (GNU)
    import core.attributes : myAttrib = gccname;
else version (LDC)
    import core.attributes : myAttrib = ldcname;
```

Similar attributes are probably way simpler than similar intrinsics.

ibuclaw:
> Similar attributes are probably way simpler than similar intrinsics.

Attributes are [already covered](https://github.com/gcc-mirror/gcc/commit/8e84a142913a1f3084d415462024964f97356bee).

maxhaton:
Would it be worth (say) implementing a generic unreachable for example (i.e. on gcc I have a built-in, on LLVM I use IR but they do the same thing).

ibuclaw:
> Would it be worth (say) implementing a generic unreachable for example (i.e. on gcc I have a built-in, on LLVM I use IR but they do the same thing).

You can do that in the language today.
```
noreturn unreachable() { }
```

maxhaton:
> > Would it be worth (say) implementing a generic unreachable for example (i.e. on gcc I have a built-in, on LLVM I use IR but they do the same thing).
> 
> You can do that in the language today without any compiler-specific intrinsics.
> 
> ```
> noreturn unreachable() { }
> ```
> 
> Saying that, there may be some warnings that need suppressing.

I was just using unreachable as an example of something which could be unified (it however trivial to do unlike other things which potentially could be unified)

RazvanN7:
@kinke What is your position with regards to the comments? Do you oppose this PR?

kinke:
I think the `core.attribute` change is good, given that GDC provides the same names in there as LDC.

I'm less convinced about `core.builtins` (probably not a single shared name for GDC and LDC), but it doesn't hurt either.

RazvanN7 (approved):

PR #3439, opened 2021-04-18
core.stdc.stdlib: MinGW is a sub-target of CRuntime_Microsoft

ibuclaw:
MinGW targets always links to one of `libcrtdll`, `libmsvcrt` (default for gdc), `libcmt`, or `libucrt` (default for ldc?).  Rather than having to duplicate all of `CRuntime_Microsoft` to `MinGW` where there's missing definitions, MinGW should instead be a sub-target, where deviations (such as proper 80-bit real support) are handled within.

Doing it this way means that GDC can support building druntime and phobos on x86_64-w64-mingw32.

Ping @kinke to check this.

### DIFF
--- a/src/core/stdc/stdlib.d
+++ b/src/core/stdc/stdlib.d
@@ -121,19 +121,22 @@ ulong   strtoull(scope inout(char)* nptr, scope inout(char)** endptr, int base);
 
 version (CRuntime_Microsoft)
 {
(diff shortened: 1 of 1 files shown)


kinke (approved):

kinke:
> MinGW should instead be a sub-target

That's what I thought too: https://github.com/dlang/druntime/pull/2505#discussion_r261888080

PR #3442, opened 2021-04-18
core.sys.posix.stdio: Add flockfile and other related POSIX declarations for more targets

ibuclaw:
Phobos `std.stdio` unconditionally declares `flockfile` and `funlockfile` in the `GENERIC_IO` path, so they must all exist on these platforms - and indeed, looking up system headers, I can see that they are all present.

Adding it to druntime means we can remove the extern declaration from Phobos.

### DIFF
--- a/src/core/sys/posix/stdio.d
+++ b/src/core/sys/posix/stdio.d
@@ -516,7 +516,57 @@ version (CRuntime_Glibc)
     int    putc_unlocked(int, FILE*);
     int    putchar_unlocked(int);
 }
+else version (CRuntime_Musl)
+{
+    void   flockfile(FILE*);
+    int    ftrylockfile(FILE*);
+    void   funlockfile(FILE*);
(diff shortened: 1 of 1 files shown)


kinke (approved):

src/core/sys/posix/stdio.d
```diff
@@ -516,6 +516,46 @@ version (CRuntime_Glibc)
...
+{
+    void   flockfile(FILE*);
+    int    ftrylockfile(FILE*);
+    void   funlockfile(FILE*);
+    int    getc_unlocked(FILE*);
+    int    getchar_unlocked();
+    int    putc_unlocked(int, FILE*);
+    int    putchar_unlocked(int);
+}
+else version (Darwin)
```
thewilsonator, line 529: `version(OSX)`

thewilsonator (changes requested):

PR #3444, opened 2021-04-18
core.stdc.stdio: Add 'under-the-hood' C I/O functions for DigitalMars, Microsoft, and Glibc runtimes

ibuclaw:
~The functions `_fputc_nolock` and `_fgetc_nolock` are macros on MinGW, so they've been added in as templated functions.~

Edit: moved them to a gdc-internal msvc support module.

### DIFF
--- a/src/core/stdc/stdio.d
+++ b/src/core/stdc/stdio.d
@@ -1347,7 +1347,7 @@ version (CRuntime_DigitalMars)
     ///
     pure int  fileno()(FILE* stream)   { return stream._file; }
   }
-  ///
+    ///
     pragma(printf)
     int   _snprintf(scope char* s, size_t n, scope const char* fmt, scope const ...);
     ///
@@ -1358,6 +1358,26 @@ version (CRuntime_DigitalMars)
     int   _vsnprintf(scope char* s, size_t n, scope const char* format, va_list arg);
     ///
     alias _vsnprintf vsnprintf;
+
+    //
+    // Digital Mars under-the-hood C I/O functions. Uses _iobuf* for the
+    // unshared version of FILE*, usable when the FILE is locked.
+    //
+
+    ///
+    int _fputc_nlock(int c, _iobuf* fp);
+    ///
+    int _fputwc_nlock(int c, _iobuf* fp);
+    ///
+    int _fgetc_nlock(_iobuf* fp);
+    ///
+    int _fgetwc_nlock(_iobuf* fp);
+    ///
+    int __fp_lock(FILE* fp);
+    ///
+    void __fp_unlock(FILE* fp);
+    ///
+    int setmode(int fd, int mode);
 }
 else version (CRuntime_Microsoft)
 {
@@ -1410,16 +1430,31 @@ else version (CRuntime_Microsoft)
     int  vsnprintf(scope char* s, size_t n, scope const char* format, va_list arg);
   }
 
-    ///
-    int _fputc_nolock(int c, FILE *fp);
-    ///
-    int _fgetc_nolock(FILE *fp);
+    //
+    // Microsoft under-the-hood C I/O functions. Uses _iobuf* for the unshared
+    // version of FILE*, usable when the FILE is locked.
+    //
+    import core.stdc.stddef : wchar_t;
+    import core.stdc.wchar_ : wint_t;
 
     ///
-    int _lock_file(FILE *fp);
+    int _fputc_nolock(int c, _iobuf* fp);
     ///
-    int _unlock_file(FILE *fp);
-
+    int _fgetc_nolock(_iobuf* fp);
+    ///
+    wint_t _fputwc_nolock(wchar_t c, _iobuf* fp);
+    ///
+    wint_t _fgetwc_nolock(_iobuf* fp);
+    ///
+    void _lock_file(FILE* fp);
+    ///
+    void _unlock_file(FILE* fp);
+    ///
+    int _setmode(int fd, int mode);
+    ///
+    int _fseeki64(FILE* stream, long offset, int origin);
+    ///
+    long _ftelli64(FILE* stream);
     ///
     intptr_t _get_osfhandle(int fd);
     ///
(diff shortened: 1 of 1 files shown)


src/core/stdc/stdio.d
```diff
@@ -1410,16 +1446,85 @@ else version (CRuntime_Microsoft)
...
+                import core.stdc.wchar_ : fputwc;
+                return fputwc(c, cast(shared) fp);
+            }
+            ///
+            wint_t _fgetwc_nolock()(_iobuf* fp)
+            {
+                pragma(inline, true);
+                import core.stdc.wchar_ : fgetwc;
+                return fgetwc(cast(shared) fp);
+            }
```
ibuclaw, line 1502: I'm not bothered if instead we move this to be instead handled in phobos.

ibuclaw: I convinced myself to remove it.  Will put these helper functions in Phobos std.stdio.

src/core/stdc/stdio.d
```diff
@@ -391,7 +391,21 @@ else version (CRuntime_Microsoft)
...
+            char* _base;
+            int _flag;
+            int _file;
+            int _charbuf;
+            int _bufsiz;
+            char* _tmpfname;
+        }
+        else
+        {
+            void* undefined;
```
ibuclaw, line 407: Is this really unknown in Microsoft headers?

src/core/stdc/stdio.d
```diff
@@ -1410,16 +1446,71 @@ else version (CRuntime_Microsoft)
...
+                pragma(inline, true);
+                fp._cnt--;
+                if (fp._cnt >= 0)
+                {
+                    immutable c = *fp._ptr++;
+                    return c & 0xFF;
+                }
+                else
+                    return _filbuf(fp);
+            }
```
kinke, line 1488: We actually already have fallbacks for these (were macros for MSVC 2013 and older, symbols since MSVC 2015/UCRT): https://github.com/dlang/druntime/blob/master/src/rt/msvc.d#L145-L169 - not sure if MinGW supports the embedded linker directives though.

Anyway, why make them a template, why not simply `extern(D) pragma(inline, true) int ...`?

src/core/stdc/stdio.d
```diff
@@ -391,7 +391,21 @@ else version (CRuntime_Microsoft)
...
+        version (MinGW)
+        {
+            char* _ptr;
+            int _cnt;
+            char* _base;
+            int _flag;
+            int _file;
+            int _charbuf;
+            int _bufsiz;
+            char* _tmpfname;
```
kinke, line 403: Does MinGW really have its own `_iobuf`, not inheriting the MSVC one? - It's one of the structs which has seen a major change with the introduction of the UCRT with MSVC 2015 IIRC.

src/core/stdc/stdio.d
```diff
@@ -1373,7 +1407,9 @@ else version (CRuntime_Microsoft)
     ///
     pure int  ferror(FILE* stream);
     ///
-    pure int  fileno(FILE* stream);
+    pure int  _fileno(FILE* stream);
+    ///
+    alias fileno = _fileno;
```
kinke, line 1412: This is also already taken care of by https://github.com/dlang/druntime/blob/27e0bb8b957675ef6af91a86a1af7f9a3f449b9c/src/rt/msvc.d#L201-L204 for older MSVC runtimes.

kinke:
May some more context: Microsoft's Universal CRT introduced with VS 2015 was a big step towards C99 compliance, making some MS-specific names with `_` prefix obsolete, adding missing math functions and slightly changing semantics of some functions (to conform to the other C runtimes, e.g., (v)snprintf IIRC). We try to support both older and newer MSVCRT versions here in druntime, using linker tricks in that `rt.msvc` module etc. - mostly because DMD ships with VS 2013 libs IIRC (simpler to redistribute compared to the UCRT, which is a Windows component), while LDC requires VS 2015 and newer (so the UCRT). The MinGW-**based** libs shipped with LDC also require the UCRT and some MSVC++ runtime >= 2015.

src/core/stdc/stdio.d
ibuclaw: I don't see it in my locally installed msys2 environment, but it appears upstream mingw have gotten round to adding in UCRT `_iobuf`.
https://github.com/mirror/mingw-w64/blob/1c773877f4a13c8bd7bfb8da80e1e8761a889f51/mingw-w64-headers/crt/stdio.h#L24-L37

However, given your previous comment, I guess I could instead put the `_fputc_unlocked` definitions inside an MSVC support module.

src/core/stdc/stdio.d
ibuclaw: GCC doesn't support these linker directives.  I can get around it in another way though (knowing that gcc always links in libmsvcrt.a) and just leave druntime assuming that these always exist (even though the wide char variants are UCRT functions).

src/core/stdc/stdio.d
ibuclaw: The alias is more for the benefit of `std.stdio` than anything else.
https://github.com/dlang/phobos/blob/e9014c43428ac3f01d02fe28b2845074e7c80421/std/stdio.d#L207-L222

ibuclaw:
However, phobos currently ignores the support that druntime has, and assumes that functions only found in UCRT must be present for CRuntime_Microsoft platforms.  So as far an I'm concerned there's really only two variants that are worth supporting - libmsvcrt and libucrt.

src/core/stdc/stdio.d
ibuclaw: I've implemented this function in an msvc support module in gdc.

src/core/stdc/stdio.d
ibuclaw: I've implemented `_fgetc_unlocked` and `_fputc_unlocked` in an msvc support module, these definitions are no longer required to be exported.

src/core/stdc/stdio.d
kinke: Yeah but resorting to the MS-specific name wasn't needed before either; newer MS runtimes have that symbol which was previously just declared, and in case the MS runtime doesn't have it, there's the fallback in the other module (and could be added to your gdc module as well). So Phobos could just use `core.stdc.stdio.fileno` for `CRuntime_Microsoft` today.

src/core/stdc/stdio.d
```diff
@@ -1410,16 +1432,33 @@ else version (CRuntime_Microsoft)
...
-
+    wint_t _fgetwc_nolock(_iobuf* fp);
+    ///
+    void _lock_file(FILE* fp);
+    ///
+    void _unlock_file(FILE* fp);
+    ///
+    int _setmode(int fd, int mode);
+    ///
+    alias setmode = _setmode;
```
kinke, line 1457: This doesn't seem to be a standard Posix name, so I'd prefer not to expose it as alias without underscore. [The Phobos module uses both names, but mainly the one with underscore...]

ibuclaw: I'll try renaming the function on import then in std.stdio (to keep the shared digitalmars/microsoft runtime code aligned).

kinke (approved):

PR #3447, opened 2021-04-19
DIP1034: add noreturn to more functions

n8sh:
Followup to https://github.com/dlang/druntime/pull/3381. This may allow removal of some of the `void* pretend_sideffect = null` parameters but for the purpose of easy rollback I'm not doing that yet.

### DIFF
--- a/src/core/exception.d
+++ b/src/core/exception.d
@@ -10,7 +10,7 @@
 module core.exception;
 
 // Compiler lowers final switch default case to this (which is a runtime error)
-void __switch_errorT()(string file = __FILE__, size_t line = __LINE__) @trusted
+noreturn __switch_errorT()(string file = __FILE__, size_t line = __LINE__) @trusted
 {
     // Consider making this a compile time check.
     version (D_Exceptions)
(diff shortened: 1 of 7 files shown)


n8sh:
This PR doesn't add `noreturn` to some functions like
https://github.com/dlang/druntime/blob/d91653c3240f3ead0ad21ff55e77e761e7956147/src/core/exception.d#L444-L449
since the definition of `_assertHandler` is
```d
private __gshared AssertHandler _assertHandler = null; 
// ...
alias AssertHandler = void function(string file, size_t line, string msg) nothrow; 
```
which has no compiler-enforced or documented requirement of terminating the program. If handlers that can return are not intended to be allowed then functions like `onAssertErrorMsg` could be re-written
```d
extern (C) noreturn onAssertErrorMsg( string file, size_t line, string msg ) nothrow
{
    if ( _assertHandler )
        _assertHandler( file, line, msg );
    throw new AssertError( msg, file, line );
}
```

Geod24:
Could you keep the error message at a reasonable length ? The recommended size is something like 72 characters IIRC, and while we don't enforce that, you don't have to list all affected functions for people to get a feel of what this commit do.

Besides that, it looks like a few projects need fixing before this can be pulled.

n8sh:
> Besides that, it looks like a few projects need fixing before this can be pulled.

Is it related to this PR? If so then this PR and/or DMD's behavior w.r.t. `noreturn` is what needs a fix.

Geod24:
> Is it related to this PR?

Well a few tests are failing that shouldn't, but looking at buildkite:
> ./src/ocean/util/container/cache/model/containers/ArrayPool.d(121): Warning: statement is not reachable

https://github.com/sociomantic-tsunami/ocean/blob/ba566c473249341633d723904a0b82ba1201a55b/src/ocean/util/container/cache/model/containers/ArrayPool.d#L120-L121

In this case it's only a matter of switching the two statements. There were other failures in Buildkite which I assumed were related. I'll fix the Ocean one.

PetarKirov (approved):
I checked the implementation / API doc of every function and the changes look good! Thanks for taking the initiative!

As far the commit message is concerned, a better suggest is to keep the message title terse and up to say 72 characters, but still include the detail in the message body.

RazvanN7:
`@noreturn` does not have a complete implementation and I think we should refrain from using it internally.

PR #3451, opened 2021-04-25
Temporarily disable join_detach test

MoonlightSentinel:
It sometime fails due to race conditions in the thread cleanup code.

CC @thewilsonator


thewilsonator (approved):

thewilsonator:
I think you need to force push to restart the Doc tester

kinke:
Thx Florian; I've seen quite frequent failures with LDC, only with shared libs IIRC, and thought it'd be some LDC-specific failure I'll need to look into at some point...

MoonlightSentinel:
> only with shared libs IIRC

The failure is a race condition causing both threads to enter the TLS cleanup code which contains some `version (Shared)` blocks. That could explain the higher probability for the shared libs.

PR #3454, opened 2021-04-27
Fix 19783 - Fail to emplace struct with betterC

MoonlightSentinel:
Defining the helper function as a template ensures that it is available when not linking druntime.

(`pragma(inline, true)` didn't work)

### DIFF
--- a/src/core/lifetime.d
+++ b/src/core/lifetime.d
@@ -82,13 +82,6 @@ T* emplace(T, Args...)(T* chunk, auto ref Args args)
     assert(i == 42);
 }
 
-private @nogc pure nothrow @safe
-void testEmplaceChunk(void[] chunk, size_t typeSize, size_t typeAlignment)
-{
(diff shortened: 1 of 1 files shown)


kinke:
Templatizing it isn't actually enough to guarantee that it's emitted (template culling, `needsCodegen()`), but yeah, it works in most cases. I'm absolutely no friend of these dummy templates; I don't want to end up with a templates-only druntime/Phobos (STL) just to allow people to use druntime/Phobos stuff with `-betterC`. People can e.g. still link druntime manually with `-betterC` - that'll most likely result in some bloat, unused TypeInfos and ModuleInfos etc., but still, missing symbols are resolved, and unless the called functions require a properly initialized druntime (module ctors etc.), it actually works.

MoonlightSentinel:
Fair enough. I've inlined those asserts because the helper function was only used in two places and this save some unnecessary template instantiations.

n8sh (approved):

PR #3457, opened 2021-05-01
Fix Issue 21363 - [REG2.094] Implementation of core.bitop.ror(x,0) is…

Chigusa0w0:
… using UB

This is a quick fix to Issue 21363. When `count` is set to `0`, `T.sizeof * 8 - count` will always be `T.sizeof * 8` therefore fails compiler check here:
https://github.com/dlang/dmd/blob/81f9f57257da8733cf741355524bb8a3ec9c45ce/src/dmd/dinterpret.d#L3018-L3024

### DIFF
--- a/src/core/bitop.d
+++ b/src/core/bitop.d
@@ -951,6 +951,9 @@ pure T rol(T)(const T value, const uint count)
     if (__traits(isIntegral, T) && __traits(isUnsigned, T))
 {
     assert(count < 8 * T.sizeof);
+    if (count == 0)
+        return cast(T) value;
+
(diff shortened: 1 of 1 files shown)


thewilsonator:
please add the test case from the bug report

Chigusa0w0:
@thewilsonator Test cases added

thewilsonator:
finally can you retarget this PR to stable (follow the bot's instructions) and this should be good to go.

thewilsonator (approved):

Chigusa0w0:
@thewilsonator rebase done

PR #3459, opened 2021-05-04
Simplify anySatisfy, allSatisfy, maxAlignment

andralex:
No measured difference in compile times or memory consumed.

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -208,54 +208,40 @@ template dtorIsNothrow(T)
 }
 
 // taken from std.meta.allSatisfy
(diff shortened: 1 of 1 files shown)


RazvanN7 (approved):

nordlow:
```d
enum allSatisfy(alias pred, items...) =
{
    static foreach (item; items)
        static if (!pred!item)
            if (__ctfe) return false;
    return true;
}();

enum anySatisfy(alias pred, items...) =
{
    static foreach (item; items)
        static if (pred!item)
            if (__ctfe) return true;
    return false;
}();
```

is even shorter.

PR #3463, opened 2021-05-07
Simplify anySatisfy and allSatisfy further

nordlow:
Follow up to https://github.com/dlang/druntime/pull/3459.

I didn't find a way to simplify

```d
template maxAlignment(Ts...)
if (Ts.length > 0)
{
    enum maxAlignment =
    {
        size_t result = 0;
        static foreach (T; Ts)
            if (T.alignof > result) result = T.alignof;
        return result;
    }();
}
```

in the same manner because of the template restriction.

It can be done if the restriction `if (Ts.length > 0)` is replaced by a

`static assert(Ts.length > 0)`

inside of `maxAlignment` though. But that's Phobos style, afaict.

### DIFF
--- a/src/core/internal/traits.d
+++ b/src/core/internal/traits.d
@@ -195,28 +195,22 @@ template dtorIsNothrow(T)
 }
 
 // taken from std.meta.allSatisfy
-template allSatisfy(alias pred, items...)
+enum allSatisfy(alias pred, items...) =
 {
-    enum allSatisfy =
-    {
-        static foreach (item; items)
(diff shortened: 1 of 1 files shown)


andralex:
Great , thanks!

I think @WalterBright mentioned that it's a mistake `if` constraints are not supported on aliases. You may want to submit a bug report to make sure it's not forgotten.

andralex (approved):

RazvanN7 (approved):

PR #3467, opened 2021-05-10
Fix Issue 21097 - prevent stack allocation when destroying large aggregates

JohanEngelen:
This fixes `destroy` and `core.internal.lifetime.emplaceInitializer` for large aggregates by preventing stack allocation for LDC and DMD.
For DMD, it also fixes a codegen issue for large static array of structs where `destroy`/`emplaceInitializer` would generate a string of `mov` instructions proportional to the size of the array.

~~Intentionally did not add unittests for this issue here, but instead added a runnable test case to the testsuite, to prevent a very large unittest binary. Test is added in separate DMD PR: https://github.com/dlang/dmd/pull/12509.~~
Tests have been added to this PR.

### DIFF
--- a/src/core/internal/lifetime.d
+++ b/src/core/internal/lifetime.d
@@ -89,37 +89,39 @@ Emplaces T.init.
 In contrast to `emplaceRef(chunk)`, there are no checks for disabled default
 constructors etc.
 +/
-void emplaceInitializer(T)(scope ref T chunk) nothrow pure @trusted
-    if (!is(T == const) && !is(T == immutable) && !is(T == inout))
+template emplaceInitializer(T)
+if (!is(T == const) && !is(T == immutable) && !is(T == inout))
 {
     import core.internal.traits : hasElaborateAssign;
 
-    static if (!hasElaborateAssign!T && __traits(compiles, chunk = T.init))
-    {
-        chunk = T.init;
-    }
-    else static if (__traits(isZeroInit, T))
-    {
-        static if (is(T U == shared U))
-            alias Unshared = U;
-        else
-            alias Unshared = T;
+    // Avoid stack allocation by hacking to get to the init symbol.
+    pragma(mangle, "_D" ~ T.mangleof[1..$] ~ "6__initZ")
+    __gshared extern immutable typeof(T.init) initializer;
 
-        import core.stdc.string : memset;
-        memset(cast(Unshared*) &chunk, 0, T.sizeof);
-    }
-    else
+    void emplaceInitializer(scope ref T chunk) nothrow pure @trusted
     {
-        // emplace T.init (an rvalue) without extra variable (and according destruction)
-        alias RawBytes = void[T.sizeof];
-
-        static union U
+        static if (__traits(isZeroInit, T))
         {
-            T dummy = T.init; // U.init corresponds to T.init
-            RawBytes data;
+            import core.stdc.string : memset;
+            memset(cast(void*) &chunk, 0, T.sizeof);
         }
-
-        *cast(RawBytes*) &chunk = U.init.data;
+        else static if (T.sizeof <= 16 && !hasElaborateAssign!T && __traits(compiles, (){ T chunk; chunk = T.init; }))
+        {
+            chunk = T.init;
+        }
+        else static if (is(T U : U[N], size_t N)) // if isStaticArray
+        {
+            // For static arrays there is no initializer symbol created. Instead, we emplace elements one-by-one.
+            foreach (i; 0..N)
+            {
+                emplaceInitializer(chunk[i]);
+            }
+        }
+        else
+        {
+            import core.stdc.string : memcpy;
+            memcpy(cast(void*)&chunk, &initializer, T.sizeof);
+        }
     }
 }
 
--- a/posix.mak
+++ b/posix.mak
@@ -385,7 +385,7 @@ UT_MODULES:=$(patsubst src/%.d,$(ROOT)/unittest/%,$(SRCS))
 HAS_ADDITIONAL_TESTS:=$(shell test -d test && echo 1)
 ifeq ($(HAS_ADDITIONAL_TESTS),1)
 	ADDITIONAL_TESTS:=test/init_fini test/exceptions test/coverage test/profile test/cycles test/allocations test/typeinfo \
-	    test/aa test/cpuid test/gc test/hash \
+	    test/aa test/cpuid test/gc test/hash test/lifetime \
 	    test/thread test/unittest test/imports test/betterc test/stdcpp test/config
 	ADDITIONAL_TESTS+=$(if $(SHARED),test/shared,)
 endif
--- a/test/lifetime/Makefile
+++ b/test/lifetime/Makefile
@@ -1,0 +1,17 @@
+include ../common.mak
+
+TESTS:=large_aggregate_destroy_21097
+
+.PHONY: all clean
+all: $(addprefix $(ROOT)/,$(addsuffix .done,$(TESTS)))
+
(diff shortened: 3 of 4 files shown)


src/core/internal/lifetime.d
```diff
@@ -89,37 +89,92 @@ Emplaces T.init.
...
+                    foreach (i; 0..N)
+                    {
+                        emplaceInitializer(chunk[i]);
+                    }
+                }
+            }
+            else
+            {
+                // Avoid stack allocation by hacking to get to the init symbol.
+                pragma(mangle, "_D" ~ T.mangleof[1..$] ~ "6__initZ")
```
JohanEngelen, line 141: @ibuclaw Can we also generate GDC's `.init` symbol names this way or with a slight tweak?

ibuclaw: Yes, initializers follow the same naming convention.

src/core/internal/lifetime.d
```diff
@@ -89,37 +89,90 @@ Emplaces T.init.
...
+                    memcpy(cast(void*)&chunk, &initializer, T.sizeof);
+                }
+            }
+        }
+        else
+        {
+            void emplaceInitializerImpl(T)(scope ref T chunk) nothrow pure @trusted @nogc
+            {
+                // Avoid stack allocation, at the cost of a virtual call to get the initializer.
+                const initializer = cast(ubyte[])typeid(T).initializer();
```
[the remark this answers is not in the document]
ibuclaw: This is also in a `version(none)` condition, so it's debatable whether this should be kept around anyway.

src/core/internal/lifetime.d
```diff
@@ -89,37 +89,39 @@ Emplaces T.init.
...
+            memset(cast(void*) &chunk, 0, T.sizeof);
+        }
+        else static if (T.sizeof <= 16 && !hasElaborateAssign!T && __traits(compiles, (){ T chunk; chunk = T.init; }))
+        {
+            chunk = T.init;
+        }
+        else static if (is(T U : U[N], size_t N)) // if isStaticArray
+        {
+            // For static arrays there is no initializer symbol created. Instead, we emplace elements one-by-one.
+            foreach (i; 0..N)
```
nordlow, line 115: `const i` or `immutable i`?

12345swordy (approved):

src/core/internal/lifetime.d
```diff
@@ -89,37 +89,39 @@ Emplaces T.init.
...
-    }
-    else static if (__traits(isZeroInit, T))
-    {
-        static if (is(T U == shared U))
-            alias Unshared = U;
-        else
-            alias Unshared = T;
+    // Avoid stack allocation by hacking to get to the init symbol.
+    pragma(mangle, "_D" ~ T.mangleof[1..$] ~ "6__initZ")
+    __gshared extern immutable typeof(T.init) initializer;
```
MoonlightSentinel, line 99: This is only used in the `else` block

JohanEngelen: Indeed. 
I guess your comment is to static-if this away, but I would have to replicate the full static-if chain, and it does not harm to have it here.

src/core/internal/lifetime.d
JohanEngelen: i don't see the benefit here

src/core/internal/lifetime.d
MoonlightSentinel: You could move it into the `else` block?

JohanEngelen: it needs to be a global variable  (and what i remember is that making it static did not work)

src/core/internal/lifetime.d
nordlow: I thought one should always prefer non-mutable for the sake of the optimizer. You certainly know that better than I do, though. :)

JohanEngelen: definitely won't help for optimization here

nordlow: Ok cool.

src/core/internal/lifetime.d
JohanEngelen: It's just a fwd decl, so no harm done if unused.

JohanEngelen:
ping

src/core/internal/lifetime.d
MoonlightSentinel: That's weird but should not block this PR.

MoonlightSentinel:
Your test is not executed by `make`, you'll need to add the directory to [this list](https://github.com/dlang/druntime/blob/f9246f413c5a256e1a74988782960aaa2fbc31e6/posix.mak#L387-L389).

JohanEngelen:
Thanks @MoonlightSentinel . Added it to the makefile (really, this test framework we have here is so horrible...), and squashed all fixup commits.

JohanEngelen:
And rebased to target stable branch.

MoonlightSentinel (approved):

JohanEngelen:
OK to rebase onto stable and merge?

MoonlightSentinel:
> Rebased back to master, because stable branch testing is broken.

What was the failure?

> OK to rebase onto stable and merge?

Yes as this fixes a regression.

src/core/internal/lifetime.d
```diff
@@ -89,37 +89,39 @@ Emplaces T.init.
...
-
-        static union U
+        static if (__traits(isZeroInit, T))
         {
-            T dummy = T.init; // U.init corresponds to T.init
-            RawBytes data;
+            import core.stdc.string : memset;
+            memset(cast(void*) &chunk, 0, T.sizeof);
+        }
+        else static if (T.sizeof <= 16 && !hasElaborateAssign!T && __traits(compiles, (){ T chunk; chunk = T.init; }))
```
MoonlightSentinel, line 108: Out of curiousity, why `16` instead of a larger value?

JohanEngelen: this value is indeed pretty arbitrary. It's to limit the stack space used.

JohanEngelen:
Rebased onto stable, PR targets stable now.

thewilsonator:
Build fails because CI checks out phobos master instead of phobos stable. I'm going to tentatively force merge this.

src/core/internal/lifetime.d
kinke: > It's just a fwd decl, so no harm done if unused.

Not for LDC:
```
FAILED: runtime/objects-unittest-debug/std/experimental/allocator/building_blocks/affix_allocator.o 
cd /home/runner/work/ldc/ldc/runtime/phobos && /home/runner/work/ldc/ldc/bin/ldc2 -c --output-o -conf= -w -de -dip1000 -preview=dtorfields -g -link-defaultlib-debug -d-debug -d-version=StdUnittest -unittest -linkonce-templates -relocation-model=pic -I/home/runner/work/ldc/ldc/runtime/druntime/src -I/home/runner/work/ldc/ldc/runtime/phobos -of=/home/runner/work/ldc/ldc/runtime/objects-unittest-debug/std/experimental/allocator/building_blocks/affix_allocator.o std/experimental/allocator/building_blocks/affix_allocator.d
/home/runner/work/ldc/ldc/runtime/druntime/src/core/internal/lifetime.d(99): Error: Global variable type does not match previous declaration with same mangled name: `_D6__initZ`
/home/runner/work/ldc/ldc/runtime/druntime/src/core/internal/lifetime.d(99):        Previous IR type: i64, const, non-thread-local
/home/runner/work/ldc/ldc/runtime/druntime/src/core/internal/lifetime.d(99):        New IR type:      i32, const, non-thread-local
```

PR #3472, opened 2021-05-14
Add scope to invertCompToken

dkorpel:
It gets called from `_d_assert_fail` with a `scope` string. The reason it currently compiles is because it's pure, so [Issue 20150](https://issues.dlang.org/show_bug.cgi?id=20150) applies. Blocker for https://github.com/dlang/dmd/pull/12010.

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -377,7 +377,7 @@ private string miniFormat(V)(const scope ref V v)
 import core.atomic : atomicLoad;
 
 // Inverts a comparison token for use in _d_assert_fail
-private string invertCompToken(string comp) pure nothrow @nogc @safe
+private string invertCompToken(scope string comp) pure nothrow @nogc @safe
 {
     switch (comp)
     {
@@ -402,7 +402,7 @@ private string invertCompToken(string comp) pure nothrow @nogc @safe
         case "!in":
             return "in";
         default:
-            assert(0, combine(["Invalid comparison operator"], "-", [comp]));
+            assert(0, combine(["Invalid comparison operator '"], comp, ["'"]));
     }
 }
 


Geod24 (approved):

src/core/internal/dassert.d
```diff
@@ -377,7 +377,7 @@ private string miniFormat(V)(const scope ref V v)
 import core.atomic : atomicLoad;
 
 // Inverts a comparison token for use in _d_assert_fail
-private string invertCompToken(string comp) pure nothrow @nogc @safe
+private string invertCompToken(scope string comp) pure nothrow @nogc @safe
```
aG0aep6G, line 380: I'm not sure if this should compile. Later in the function, we have `[comp]`. I.e., `comp` is copied to the heap. Usually, that's not allowed for `scope` variables.

MoonlightSentinel: I think  it works here because the array is assigned to a `scope` parameter and hence is promoted to the stack.

dkorpel: The `[comp]` is passed to parameter `const scope string[] valB` of `combine`, so the literal is allocated on the stack

aG0aep6G: > I assume it works here because the array is assigned to a `scope` parameter and hence promoted to the stack.

That might be DMD's reasoning. But it's unsound. It actually doesn't matter if the array is on the heap or on the stack. A `scope` is being lost.

`comp.ptr` is a `scope` pointer. `valB.ptr` is also a `scope` pointer. `valB[0].ptr` is not a `scope` pointer, because `scope` is not transitive. So `valB = [comp];` cannot be allowed. It's assigning a `scope` pointer to a non-`scope` pointer.

Reduced code demonstrating memory corruption:

```d
string invertCompToken(scope string comp) @safe
{
    return combine([comp]); /* Should not compile. */
}
string combine(const scope string[] valB) @safe
{
    return valB[0];
}
string f() @safe
{
    immutable char[3] eq = "foo";
    return invertCompToken(eq);
}
void main() @safe
{
    import std.stdio;
    string s = f();
    writeln(s); /* prints garbage */
}
```

MoonlightSentinel: Fair enough, that's just another oddity of `scope` not being transitive.
But there's no corruption here because `combine` doesn't escape the string. So it should be fine to wrap the call in a `@trusted` lambda to avoid future problems.

dkorpel: I put the operator in the middle argument of `combine`, side-stepping the issue.

MoonlightSentinel (approved):

src/core/internal/dassert.d
dkorpel: Btw nice catch @aG0aep6G! I think the relevant bugzilla issue is this:
https://issues.dlang.org/show_bug.cgi?id=20505

PR #3474, opened 2021-05-17
Rewrite `object.dup` to work with `scope` and copy constructors

dkorpel:
`.dup` and `.idup` used to be built-in properties of arrays along with `.sort` and `.reverse`.
In 2014 [they got implemented as druntime functions](https://github.com/dlang/druntime/pull/760), but being 7 years old, they now have the following shortcomings:
- `scope` inference of the input array fails, making it a blocker for [issue 20150 - dip1000 defeated by pure](https://issues.dlang.org/show_bug.cgi?id=20150)
- it does not account for copy-constructors, only postblits
- since 2018, the implicit conversion to `immutable` of e.g. `char[].dup` [relies on a compiler hack](https://github.com/dlang/dmd//commit/ab5a18635adaa3e2271e0f4b569500a89ac74235#diff-88cc711d774668de0af3461742be476454e8374545722dfd77457a3264aecf32R804).

This PR addresses the first two points, the third point depends on a dmd change.

The copy-constructor test are taken from https://github.com/dlang/druntime/pull/3139.
The post-blit / copy-constructor logic uses `copyEmplace`.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3480,31 +3480,47 @@ private size_t getArrayHash(const scope TypeInfo element, const scope void* ptr,
     assert(s == "abc");
 }
 
-private U[] _dup(T, U)(T[] a) // pure nothrow depends on postblit
+private U[] _dup(T, U)(scope T[] a) pure nothrow @trusted if (__traits(isPOD, T))
 {
     if (__ctfe)
-    {
-        static if (is(T : void))
-            assert(0, "Cannot dup a void[] array at compile time.");
-        else
-        {
-            U[] res;
-            foreach (ref e; a)
-                res ~= e;
-            return res;
-        }
-    }
+        return _dupCtfe!(T, U)(a);
 
     import core.stdc.string : memcpy;
+    auto arr = _d_newarrayU(typeid(T[]), a.length);
+    memcpy(arr.ptr, cast(const(void)*) a.ptr, T.sizeof * a.length);
+    return *cast(U[]*) &arr;
+}
 
+private U[] _dupCtfe(T, U)(T[] a)
+{
+    static if (is(T : void))
+        assert(0, "Cannot dup a void[] array at compile time.");
+    else
+    {
+        U[] res;
+        foreach (ref e; a)
+            res ~= e;
+        return res;
+    }
+}
+
+private U[] _dup(T, U)(T[] a) if (!__traits(isPOD, T))
+{
+    // note: copyEmplace is `@system` inside a `@trusted` block, so the __ctfe branch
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -3810,6 +3810,60 @@ private void _doPostblit(T)(T[] arr)
...
+    // that the struct defines a copy constructor that creates immutable objects
+    static struct ISunpure { this(ref const typeof(this)) immutable @safe nothrow {} }
+    static struct ISthrow { this(ref const typeof(this)) immutable @safe pure {} }
+    static struct ISunsafe { this(ref const typeof(this)) immutable @system pure nothrow {} }
+    static assert( __traits(compiles, ()         { [].idup!ISunpure; }));
+    static assert(!__traits(compiles, () pure    { [].idup!ISunpure; }));
+    static assert( __traits(compiles, ()         { [].idup!ISthrow; }));
+    static assert(!__traits(compiles, () nothrow { [].idup!ISthrow; }));
+    static assert( __traits(compiles, ()         { [].idup!ISunsafe; }));
+    static assert(!__traits(compiles, () @safe   { [].idup!ISunsafe; }));
```
MoonlightSentinel, line 3864: Remove the useless `static assert( __traits(compiles, ...)` and move them into a `if (false) { ... }` if they shouldn't be executed.
Also move the instantiation out of the `static assert(!__traits(compiles, ...)` and only check that the instance cannot be called in a `@safe` function. Otherwise the test can pass due to other failures.

dkorpel: Good call,  I now just call the valid cases directly and only use the trait to test the function attributes. 
If they don't compile because of other reasons, the valid cases should fail first with a decent error message.

dkorpel:
By the way, I noticed that `dup` seems to be [a CTFE builtin](https://github.com/dlang/dmd/blob/81f9f57257da8733cf741355524bb8a3ec9c45ce/src/dmd/dinterpret.d#L7286), while `idup` isn't? I'm not sure if the code in the `__ctfe` branch is actually used.

dkorpel:
I don't know what's happening with the auto-tester... One moment it's reporting "Pending: 8", another moment it's all green.

aG0aep6G:
> I don't know what's happening with the auto-tester... One moment it's reporting "Pending: 8", another moment it's all green.

"Pending: 8" is correct. The "Details" page is all green, but those results are old. Tests that haven't even started yet are not visible there.

You can also find your PR here: <https://auto-tester.puremagic.com/pulls.ghtml?projectid=1>. There you see that only two results are up to date (bright green). The other eight are not up to date but they passed previously (muted green).

RazvanN7 (approved):
Awesome! Thanks for taking care of this.

PR #3477, opened 2021-05-18
rt/dwarfeh: Improve error message on uncaught exception

Geod24:
I encountered this message while working on a large project
and was puzzled as to why I couldn't find the string 'uncaught exception'
anywhere in my code or dependencies'.


thewilsonator (approved):
test case?

Geod24:
It's just changing a message that should never trigger, do we really want a test case for this ?

thewilsonator:
I was just wondering if we had one. Theres no real need , just making sure it was intentional.

PR #3480, opened 2021-05-20
Annotate pure functions with return/scope

dkorpel:
This PR is similar to https://github.com/dlang/phobos/pull/8076, but for druntime and for https://github.com/dlang/dmd/pull/12010 instead of https://github.com/dlang/dmd/pull/10924
It's probably too big to merge in one go, this is a proof of concept that can later be split up into smaller chunks for easier review.

The offending functions were found with something I cobbled together in dmd: https://github.com/dkorpel/dmd/commit/7d1f0c5f68ff9127d06c8d54e391be71c15dc4a8
It outputs this when running druntime unittests:
```
src/core/internal/dassert.d(221): parameter `a` of `idup` is cheating `scope`
src/object.d(3410): parameter `value` of `getElement` is cheating `scope`
src/core/internal/dassert.d(164): parameter `val` of `atomicLoad` is cheating `scope`
src/core/atomic.d(1161): parameter `val` of `atomicFetchAdd` is cheating `scope`
src/core/atomic.d(1208): parameter `val` of `atomicFetchSub` is cheating `scope`
src/core/lifetime.d(1881): parameter `source` of `moveImpl` is cheating `scope`
src/core/memory.d(737): parameter `p` of `gc_query` is cheating `scope`
src/core/memory.d(1453): parameter `p` of `realloc` is cheating `scope`
src/core/internal/convert.d(479): parameter `a` of `dup` is cheating `scope`
src/core/internal/array/appending.d-mixin-42(50): parameter `px` of `_d_arrayappendcTX` is cheating `scope`
src/core/internal/array/casting.d(101): parameter `from` of `__ArrayCast` is cheating `scope`
src/core/internal/array/concatenation.d(41): parameter `arrs` of `_d_arraycatnTX` is cheating `scope`
src/core/internal/backtrace/dwarf.d(153): parameter `dst` of `demangle` is cheating `scope`
src/core/internal/hash.d(213): parameter `arr` of `toUbyte` is cheating `scope`
src/rt/aaA.d(802): parameter `cti` of `unqualify` is cheating `scope`
```

One of the harder ones is:
```
src/core/lifetime.d(1881): parameter `source` of `moveImpl` is cheating `scope`
```
It's really tricky, since it carefully uses `@trusted` blocks to nudge the `@safe` inference in the right direction, which now breaks. Maybe someone can help out with that.

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -66,7 +66,7 @@ enum MemoryOrder
  * Returns:
  *  The value of 'val'.
  */
-T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)(ref const T val) pure nothrow @nogc @trusted
+T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)(return ref const T val) pure nothrow @nogc @trusted
     if (!is(T == shared U, U) && !is(T == shared inout U, U) && !is(T == shared const U, U))
 {
     static if (__traits(isFloating, T))
@@ -80,7 +80,7 @@ T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)(ref const T val) pure nothrow 
 }
 
 /// Ditto
-T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)(ref shared const T val) pure nothrow @nogc @trusted
+T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)(return ref shared const T val) pure nothrow @nogc @trusted
     if (!hasUnsharedIndirections!T)
 {
     import core.internal.traits : hasUnsharedIndirections;
@@ -162,7 +162,7 @@ void atomicStore(MemoryOrder ms = MemoryOrder.seq, T, V)(ref shared T val, share
  * Returns:
  *  The value held previously by `val`.
  */
-T atomicFetchAdd(MemoryOrder ms = MemoryOrder.seq, T)(ref T val, size_t mod) pure nothrow @nogc @trusted
+T atomicFetchAdd(MemoryOrder ms = MemoryOrder.seq, T)(return ref T val, size_t mod) pure nothrow @nogc @trusted
     if ((__traits(isIntegral, T) || is(T == U*, U)) && !is(T == shared))
 in (atomicValueIsProperlyAligned(val))
 {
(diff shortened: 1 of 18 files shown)


aG0aep6G:
> The offending functions were found with something I cobbled together in dmd: [dkorpel/dmd@7d1f0c5](https://github.com/dkorpel/dmd/commit/7d1f0c5f68ff9127d06c8d54e391be71c15dc4a8)

Awesome!

> One of the harder ones is:
> 
> ```
> src/core/lifetime.d(1881): parameter `source` of `moveImpl` is cheating `scope`
> ```

I think that boils down to `moveEmplaceImpl` assigning its first parameter to its second parameter. For DIP 1000 it must be other way around. No, really, it's in the spec.

1. <https://dlang.org/spec/function.html#return-ref-parameters>: "If the function returns void, and the **first parameter** is ref or out, then all subsequent return ref parameters are considered as being assigned to the first parameter for lifetime checking.".
2. <https://dlang.org/spec/function.html#param-storage>: "[A `return` parameter] may be returned or copied to the **first parameter**, but otherwise does not escape from the function."

For internal functions like `moveEmplaceImpl` we can just reverse the parameters. But `move(source, target)` might simply not be compatible with DIP 1000.

RazvanN7 (approved):
Looks good to me. Why is this still a draft?

dkorpel:
The remaining errors are with `dup` and `move`. `dup` should be fixed in https://github.com/dlang/druntime/pull/3474, `move` is still a blocker since the `@safe` unittests `move` rely on the `pure` bug. I've made a [forum post](https://forum.dlang.org/post/zzovywgswjmwneqwbdnm@forum.dlang.org) about it.

This could be merged in the meanwhile, with a follow up PR for `move` eventually.

src/core/atomic.d
```diff
@@ -90,7 +90,7 @@ T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)(ref shared const T val) pure n
 }
 
 /// Ditto
-TailShared!T atomicLoad(MemoryOrder ms = MemoryOrder.seq, T)(ref shared const T val) pure nothrow @nogc @trusted
```
dkorpel, line 93: Wait a second, this is not right. `ref` should not be removed.

dkorpel: I remember `ref return` not compiling, I think it should be `ref return scope`. I removed this change, will look at it later today (or in a follow-up PR).

PR #3481, opened 2021-05-27
Fix issue 21550

adamdruppe:
See https://issues.dlang.org/show_bug.cgi?id=21550 for details.

### DIFF
--- a/src/core/memory.d
+++ b/src/core/memory.d
@@ -1229,12 +1229,12 @@ void __delete(T)(ref T x) @system
               is(T == class) ||
               is(T == U2*, U2))
     {
-        GC.free(cast(void*) x);
+        GC.free(GC.addrOf(cast(void*) x));
         x = null;
     }
     else static if (is(T : E2[], E2))
     {
-        GC.free(cast(void*) x.ptr);
+        GC.free(GC.addrOf(cast(void*) x.ptr));
         x = null;
     }
 }


schveiguy:
ugh, it's a shame this does 2 lookups in the GC for the same block. Which is, of course, a waste for small blocks.

In order to do this with one lookup, we need a mechanism to do the free based on an interior pointer. And actually, I think this mechanism isn't perfect either -- if the thing isn't an array, but just a slice into a block of something else (like a struct or class), then it's going to be surprising.

A better mechanism (one consistent with the original intent) is to:

1. Lookup the block info.
2. Subtract any prefix size (i.e., subtract 0 for blocks which are not arrays or are arrays but array info is stored at the end, subtract size_t.sizeof for blocks that are arrays and are large enough that the block info would be stored at the front), and pass that to GC.free

RazvanN7 (approved):

schveiguy:
This should not have been pulled. It changes the expectation for freeing an array. Did nobody read my comment?

schveiguy:
I guess there is no definition of what should happen when freeing an internal pointer, or pointer to a member of an object. For instance:

```d
class C
{
   int d;
}

auto c = new C;
__delete(&c.d); // now frees the class, but does not call the dtor
```

I wouldn't call this expected behavior, and it is eminently possible to properly work this out (there are various flags stored that tell what kind of object is stored there).

This does fix the behavior of freeing an interface, which it seems like was also not properly done.

adamdruppe:
The `delete` keyword refuses to allow `&c.d` because it is an lvalue... but putting it in a temp var it apparently does nothing; no segfault use after delete, no dtor called. hmmm so indeed not a match.

but i also wonder if it matters. the whole point of __delete is "you're on your own". the change here at least makes it actually work when you delete an array. deleting a slice matches the old keyword, for better or for worse, it is just this case that is different.

tbh i think the delete deprecation should just be undone.


but maybe your idea with the BlkInfo could fix it up better. i don't know, it'd take a while since i'm not that familiar with it. i have other things to do today though but we can come back to it with a supplemental PR over the weekend or something.

schveiguy:
Yeah, the `&c.d` wouldn't work, but it's not necessarily too far-fetched to think that a pointer is passed to some function that expects `__delete` to work on it, which currently does nothing, but with this will actually free the block it's in. The documentation is unclear on whether this is intended or expected, it says that passing something other than a GC pointer is undefined behavior, but this is a GC pointer *and* the documentation doesn't define what should happen with interior pointers.

The thing I don't like about this solution is that it goes against the spirit of what delete should actually do -- it frees something which it doesn't necessarily have reference to. I would prefer it either does nothing, or gives you an error (maybe optionally?)

Another way to fix this is clarify the documentation. It currently says
```
If `x` does not refer to memory previously allocated with `new` (or the lower-level equivalents in the GC API), the behavior is undefined.
```

It could say:
```
If `x` is not equivalent to a pointer previously returned from `new` (or the lower-level equivalents in the GC API), the behavior is undefined.
```

Which means you are free to assume the pointer must have been returned from `new` or equivalent.

rainers:
> 2\. Subtract any prefix size (i.e., subtract 0 for blocks which are not arrays or are arrays but array info is stored at the end, subtract size_t.sizeof for blocks that are arrays and are large enough that the block info would be stored at the front), and pass that to GC.free

The implementation of `delete` actually uses the base pointer of the allocation for any slice, but checks the APPENDABLE flag, see https://github.com/dlang/druntime/blob/master/src/rt/lifetime.d#L1172. So I'd say the version of this PR is closer to `delete` than the previous one.


> The `delete` keyword refuses to allow `&c.d` because it is an lvalue... but putting it in a temp var it apparently does nothing; no segfault use after delete, no dtor called. hmmm so indeed not a match.

You could get some bad free when using this on a pointer to the first element of an `new`ed struct. This skips the struct destructor, but still deallocates the memory. I guess `delete` would be fooled by this too, though.

Calling `GC.addrOf` on a class pointer should be unnecessary as it is identical to the base pointer, but using it on an interface is an improvement. Using `cast(Object)` might be more efficient, but will not work for C++ interfaces.

> tbh i think the delete deprecation should just be undone.

I agree, the rational was pretty weak.  AFAICT it's the only unsafe feature of the language being removed for not being safe.

schveiguy:
> The implementation of `delete` actually uses the base pointer of the allocation for any slice, but checks the APPENDABLE flag

Oh yeah, so this actually is close to the `delete` keyword. So I have to withdraw that comment, because if `__delete` is supposed to replace `delete`, then this seems to do the trick.



> Calling `GC.addrOf` on a class pointer should be unnecessary as it is identical to the base pointer, but using it on an interface is an improvement. Using `cast(Object)` might be more efficient, but will not work for C++ interfaces.

I'm pretty sure we can write an `isCPPInterface` trait quite easily, because it should be extern(C++) or derived from a known C++ interface (e.g. `IUnknown`). But I think it's reasonable to use the code as checked in (of course, deleting a class object can avoid the dual-lookup).



> the rational was pretty weak.

Wasn't it "we are wasting a keyword on this"? I doubt we ever get `delete` as an identifier. But I don't really care about the deprecation that much, as long as there's a way to do what `delete` did.

adamdruppe:
On Fri, May 28, 2021 at 02:27:16PM -0700, Steven Schveighoffer wrote:
> Oh yeah, so this actually is close to the `delete` keyword. So I have to withdraw that comment, because if `__delete` is supposed to replace `delete`, then this seems to do the trick.

Yeah, I actually found the baseOf thing by looking at the implementation
of the old delete.

Though that was many months ago so I probably did forget some details.
They don't exactly match as noted in the previous comment... just it at
last actually frees classes and arrays so that's better than nothing in
my eyes.

> Wasn't it "we are wasting a keyword on this"? I doubt we ever get `delete` as an identifier. But I don't really care about the deprecation that much, as long as there's a way to do what `delete` did.

This happened along with the deprecation of `scope` with classes as part
of a crusade against all memory unsafe features in the language. The
idea is use-after-free should be the library not in the lang itself.

This crusade didn't get far. Thankfully. But it did do some damage, some
of which has been undone.

PR #3483, opened 2021-05-28
Fix 21983 -  Initialize remaining elements in a partially dup'ed array if postblit/copy ctor throws

MoonlightSentinel:
This ensures deterministic behaviour in case of an exception and
avoids dtor calls on uninitialized elements.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3514,7 +3514,15 @@ private U[] _dup(T, U)(T[] a) if (!__traits(isPOD, T))
     import core.lifetime: copyEmplace;
     U[] res = () @trusted {
         auto arr = cast(U*) _d_newarrayU(typeid(T[]), a.length);
-        foreach (i; 0..a.length)
+        size_t i;
+        scope (failure)
+        {
+            import core.internal.lifetime: emplaceInitializer;
+            // Initialize all remaining elements to not destruct garbage
+            foreach (j; i .. a.length)
+                emplaceInitializer(cast() arr[j]);
+        }
+        for (; i < a.length; i++)
         {
             copyEmplace(a.ptr[i], arr[i]);
         }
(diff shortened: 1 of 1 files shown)


src/object.d
```diff
@@ -3498,6 +3498,13 @@ private U[] _dupCtfe(T, U)(T[] a)
     else
     {
         U[] res;
+        scope (failure)
+        {
+            // Destroy all already constructed elements
+            foreach_reverse (ref el; res)
+                destroy!false(el);
```
RazvanN7, line 3505: It seems like this call breaks the purity _dupCTFE.

MoonlightSentinel:
Explicit destruction is not feasible due to the breakage caused by less-qualified constructors.
This patch will now initialize the remaining elements s.t. there won't be dtors run on uninitialised arrays (effectively restoring the behaviour before #3474)

src/object.d
```diff
@@ -3514,7 +3514,15 @@ private U[] _dup(T, U)(T[] a) if (!__traits(isPOD, T))
...
     U[] res = () @trusted {
         auto arr = cast(U*) _d_newarrayU(typeid(T[]), a.length);
-        foreach (i; 0..a.length)
+        size_t i;
+        scope (failure)
+        {
+            import core.internal.lifetime: emplaceInitializer;
+            // Initialize all remaining elements to not destruct garbage
+            foreach (j; i .. a.length)
+                emplaceInitializer(cast() arr[j]);
```
kinke, line 3523: So you're resetting the element whose postblit/cctor threw - should we?

MoonlightSentinel: Yes, I think it's necessary because the GC will call the destructor for that element.

thewilsonator (approved):

src/object.d
kinke: My concern was that the postblit/cpctor might not clean up itself on an exception and depend on the dtor being run on the partially constructed instance. But that's hopefully handled by `-preview=dtorfields`, and resetting to `T.init` correct then. Maybe slightly adjust the comment to make that clear, to prevent someone from thinking it's a one-off bug as I did. ;)

John-Colvin: Isn't the correct thing instead for `copyEmplace` to reset to `T.init` on failure, then this loop would be from `i + 1`?

Otherwise this fix needs to be done in a bunch of other places because it would mean `copyEmplace` always needs a `try`/`catch` or `scope(failure)` to use safely if it might throw.

kinke: Yes, that seems to make a lot of sense. After all, regular instances failing their construction aren't destructed, so postblit/cpctor cannot rely on the dtor being run in case they throw. - Maybe just using a `try / catch(Exception)` to prevent the overhead for `nothrow` postblits and cpctors.

kinke: => https://issues.dlang.org/show_bug.cgi?id=22177

PR #3484, opened 2021-05-29
[stable] core.internal.lifetime: Fix recent regressions for emplaceInitializer()

kinke:
The cases where it's needed are restricted to struct types; don't create unused and conflicting declarations for primitive types etc.

### DIFF
--- a/src/core/internal/lifetime.d
+++ b/src/core/internal/lifetime.d
@@ -92,12 +92,12 @@ constructors etc.
 template emplaceInitializer(T)
 if (!is(T == const) && !is(T == immutable) && !is(T == inout))
 {
-    import core.internal.traits : hasElaborateAssign;
+    import core.internal.traits : hasElaborateAssign, Unqual;
 
     // Avoid stack allocation by hacking to get to the struct/union init symbol.
     static if (is(T == struct) || is(T == union))
     {
-        pragma(mangle, "_D" ~ T.mangleof[1..$] ~ "6__initZ")
+        pragma(mangle, "_D" ~ Unqual!T.mangleof[1..$] ~ "6__initZ")
         __gshared extern immutable T initializer;
     }
 
--- a/src/core/internal/lifetime.d
+++ b/src/core/internal/lifetime.d
@@ -94,9 +94,12 @@ if (!is(T == const) && !is(T == immutable) && !is(T == inout))
 {
     import core.internal.traits : hasElaborateAssign;
 
-    // Avoid stack allocation by hacking to get to the init symbol.
-    pragma(mangle, "_D" ~ T.mangleof[1..$] ~ "6__initZ")
-    __gshared extern immutable typeof(T.init) initializer;
+    // Avoid stack allocation by hacking to get to the struct/union init symbol.
+    static if (is(T == struct) || is(T == union))
+    {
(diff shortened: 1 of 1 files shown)


kinke:
Wrt. https://github.com/dlang/druntime/pull/3467#discussion_r641753797.

n8sh:
```d
    void emplaceInitializer(scope ref T chunk) nothrow pure @trusted
    {
        static if (__traits(isZeroInit, T))
        {
            import core.stdc.string : memset;
            memset(cast(void*) &chunk, 0, T.sizeof);
        }
        else static if (T.sizeof <= 16 && !hasElaborateAssign!T && __traits(compiles, (){ T chunk; chunk = T.init; }))
        {
            chunk = T.init;
        }
        else static if (is(T U : U[N], size_t N)) // if isStaticArray
        {
            // For static arrays there is no initializer symbol created. Instead, we emplace elements one-by-one.
            foreach (i; 0..N)
            {
                emplaceInitializer(chunk[i]);
            }
        }
        else
        {
            import core.stdc.string : memcpy;
            memcpy(cast(void*)&chunk, &initializer, T.sizeof);
        }
    }
```

It seems to me there are two non-`struct` cases that could fall to the `memcpy` branch and need `initializer`:

* Non-zero-init `union`s of size greater than 16
* `double4` and other non-zero-init SIMD vectors of size greater than 16

kinke:
Unions are structs. Vectors don't have init symbols, only structs and classes. `emplaceInitializer` for class references simply puts a `null`.

kinke:
The vector case is interesting though and probably fails with current master/stable.

n8sh:
>Unions are structs.

In the compiler yes but not for the purpose of `is(X == struct)`:
https://run.dlang.io/is/7Syt7p
```d
void main()
{
    union U {}
    static assert(is(U == struct)); // fails
}
```

>Vectors don't have init symbols, only structs and classes.

Then I guess the above-mentioned code would have failed anyway for vectors,  so this PR doesn't need to address that.

n8sh (approved):

kinke:
Argh, the new test seems to expose a compiler mangling bug...

kinke:
Nope, the new test exposed another regression (besides the large vector one) for `emplaceInitializer`, not working for qualified structs/unions (with extra mangle prefix); fixed now.

kinke:
Vector case fixed too. These regressions are fresh (#3467, merged 12 days ago), thus no bugzilla filings.

src/core/internal/lifetime.d
```diff
@@ -105,7 +108,8 @@ if (!is(T == const) && !is(T == immutable) && !is(T == inout))
             import core.stdc.string : memset;
             memset(cast(void*) &chunk, 0, T.sizeof);
         }
-        else static if (T.sizeof <= 16 && !hasElaborateAssign!T && __traits(compiles, (){ T chunk; chunk = T.init; }))
+        else static if (is(T == __vector) ||
```
n8sh, line 111: Doesn't work for enum types: https://run.dlang.io/is/aUNF83
```d
import core.simd : float4;
enum E : float { _ }
static assert(is(float4 == __vector));
static assert(is(E == __vector)); // fails
```

kinke: You don't run out of ideas. :] - I'm done with this PR though, still enough other 2.097 issues to overcome for LDC.

PR #3485, opened 2021-05-30
Rename raw to rlx

rymrg:
Raw is a bad name. It should be called relaxed / rlx like in C/C++11. Raw access means non atomic access.

### DIFF
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -29,9 +29,12 @@ enum MemoryOrder
      */
     rlx = 0,
     /**
-     * ditto
+     * Deprecated.
+     * Not sequenced.
+     * Corresponds to $(LINK2 https://llvm.org/docs/Atomics.html#monotonic, LLVM AtomicOrdering.Monotonic)
+     * and C++11/C11 `memory_order_relaxed`.
      */
-    deprecated raw = 0,
+    raw = 0,
     /**
      * Hoist-load + hoist-store barrier.
      * Corresponds to $(LINK2 https://llvm.org/docs/Atomics.html#acquire, LLVM AtomicOrdering.Acquire)
--- a/src/core/atomic.d
+++ b/src/core/atomic.d
@@ -27,7 +27,11 @@ enum MemoryOrder
      * Corresponds to $(LINK2 https://llvm.org/docs/Atomics.html#monotonic, LLVM AtomicOrdering.Monotonic)
      * and C++11/C11 `memory_order_relaxed`.
      */
-    raw = 0,
+    rlx = 0,
+    /**
+     * ditto
+     */
+    deprecated raw = 0,
     /**
      * Hoist-load + hoist-store barrier.
      * Corresponds to $(LINK2 https://llvm.org/docs/Atomics.html#acquire, LLVM AtomicOrdering.Acquire)


dnadlinger:
It's actually "monotonic" in LLVM.

rymrg:
The rationale is available here along with an atomic struct I would like to push into phobos.

https://github.com/rymrg/drm/blob/main/atomic_rationale.md

Geod24:
> It's actually "monotonic" in LLVM. I don't think this is a good change; the behaviour is clearly documented.

It's a bit odd, the `enum` links to cppreference (which uses relaxed) while individual entries link to LLVM which uses a different naming. So the change makes it consistent with cppreference at least (IDK where the "raw" cam from).

src/core/atomic.d
```diff
@@ -27,6 +27,13 @@ enum MemoryOrder
      * Corresponds to $(LINK2 https://llvm.org/docs/Atomics.html#monotonic, LLVM AtomicOrdering.Monotonic)
      * and C++11/C11 `memory_order_relaxed`.
      */
+    rlx = 0,
```
MoonlightSentinel, line 30: Using `relaxed` instead of this abbreviation would be more readable IMO

src/core/atomic.d
```diff
@@ -27,6 +27,13 @@ enum MemoryOrder
...
+    /**
+     * Deprecated.
+     * Not sequenced.
+     * Corresponds to $(LINK2 https://llvm.org/docs/Atomics.html#monotonic, LLVM AtomicOrdering.Monotonic)
+     * and C++11/C11 `memory_order_relaxed`.
+     */
     raw = 0,
```
MoonlightSentinel, line 37: Doesn't need to be deprecated, there's little benefit to be gained here.

```suggestion
    /// ditto
    raw = 0,
```

src/core/atomic.d
rymrg: I don't object to `relaxed`. But right now we use `acq` and `rel` instead of `acquire` and `release`. So it'd be a bit odd if `relaxed` is much longer all of a sudden.
You could argue that it should require more typing since it is easier to misuse and get synchronization that is too weak. But the answer to that would be don't use explicit access unless you know what you do. `SC` should be default and easy to use.

src/core/atomic.d
rymrg: We talked about this a bit during beerconf, but didn't reach a clear cut conclusion on how should we tackle this.
The following options were raised.
1. Deprecating the symbol - Resulting in clear code at the cost of deprecation. It will also allow easier reflection over the type.
2. Just adding the new enum member as documented - This results in two different names to the same memory access that can and will be used interchangeably. Nothing will nudge the programmer to use one over the other.
3. Keep `raw` available but undocumented - This will keep old code compiling and nudge users to the new symbol. The cost is that if someone does encounter the old symbol, they'll have to check the source for the meaning.

As a first step, the PR just introduced the new symbol to make sure nothing breaks as a result of this. Deprecating the symbol cause CI failure since it treats warnings as errors.

MoonlightSentinel: (Should've mentioned that these issues arised at BeerConf. I just wanted to let other people have a word in this discussion as well as https://github.com/dlang/druntime/pull/3485#discussion_r642149544)

n8sh:
> > It's actually "monotonic" in LLVM. I don't think this is a good change; the behaviour is clearly documented.
> 
> It's a bit odd, the `enum` links to cppreference (which uses relaxed) while individual entries link to LLVM which uses a different naming. So the change makes it consistent with cppreference at least (IDK where the "raw" cam from).

[The documentation for MemoryOrder didn't initially refer to either cppreference or the LLVM docs.  I added those links four years ago because I initially was unsure precisely what "hoist-load", "sink-store", etc. meant and after finding an answer added it to the documentation as potentially helpful to others.](https://forum.dlang.org/thread/vzzitsnvsxljscrzbcnm@forum.dlang.org)

n8sh:
I'm against anything that contributes to the treadmill of valid D code from 3 years ago not compiling with the most recent D compiler + runtime. (Since by policy of the D Language Foundation whenever something is deprecated it is scheduled for removal.) I'm **adamantly** against breaking code just on the belief that the name of a constant could be improved. Maybe next year someone will decide that three-letter abbreviations weren't the ideal tradeoff between clarity and economy of style ("I have autocomplete and a wide monitor") and decide every name should be fully spelled out. Time to break everyone's code again?

rymrg:
> > > It's actually "monotonic" in LLVM. I don't think this is a good change; the behaviour is clearly documented.
> > 
> > 
> > It's a bit odd, the `enum` links to cppreference (which uses relaxed) while individual entries link to LLVM which uses a different naming. So the change makes it consistent with cppreference at least (IDK where the "raw" cam from).
> 
> [The documentation for MemoryOrder didn't initially refer to either cppreference or the LLVM docs. I added those links four years ago because I initially was unsure precisely what "hoist-load", "sink-store", etc. meant and after finding an answer added it to the documentation as potentially helpful to others.](https://forum.dlang.org/thread/vzzitsnvsxljscrzbcnm@forum.dlang.org)

That is bad. Because it means that prior to you asking about it, the memory model wasn't even defined. 
It also makes me wonder, what is the definition of release sequence. It changed from C++11 to C++20 to allow more behaviors (i.e., break assumptions of programs about synchronization) since the deprecated part from C11 were hard to reason about.

rymrg:
> I'm against anything that contributes to the treadmill of valid D code from 3 years ago not compiling with the most recent D compiler + runtime. (Since by policy of the D Language Foundation whenever something is deprecated it is scheduled for removal.) I'm **adamantly** against breaking code just on the belief that the name of a constant could be improved. Maybe next year someone will decide that three-letter abbreviations weren't the ideal tradeoff between clarity and economy of style ("I have autocomplete and a wide monitor") and decide every name should be fully spelled out. Time to break everyone's code again?

I agree with wanting to avoid code breakage and changing symbol names in general. This specific case uses incorrect terminology. Hence one of the options was to simply hide the old symbol to prevent new usage.

The problem with raw is that it means a non-atomic access. This is a completely different type of access allowing different type of optimizations. Non-atomic access can be moved around the code (code reordering) much more freely while relaxed my not. This is because relaxed accompanied by fence release / acquire may trigger a release sequence or optimization. In addition, atomic relaxed access offers a read-modify-write operation while a non atomic does not.

Regarding the spelling of `rlx` versus `relaxed` it's a question of clarity versus consistency. I can point to papers that refer to relaxed access as `rlx`, so its not a new usage.

TurkeyMan:
> I'm adamantly against breaking code just on the belief that the name of a constant could be improved.

I also had this same conversation with myself last year or whenever. "Raw" is obviously a stupid name, but it's there now. Maybe a future major refactor or renovation might be an opportunity to change it, but not just at random because 'reasons'.

rymrg:
> > I'm adamantly against breaking code just on the belief that the name of a constant could be improved.
> 
> I also had this same conversation with myself last year or whenever. "Raw" is obviously a stupid name, but it's there now. Maybe a future major refactor or renovation might be an opportunity to change it, but not just at random because 'reasons'.

At least now this should be documented as reference point. Maybe we can rename it in Phobos v2 if no major refactor or renovation happens before that.

RazvanN7:
> As for the current status, what are the downsides for introducing an additional symbol (rlx or relaxed) to the enum?

It adds a maintenance cost with no obvious benefit.

I suggest closing this PR.

PR #3486, opened 2021-05-31
[stable] core.internal.dassert: Add support for vectors in miniFormat()

kinke:
As newly required for https://github.com/dlang/druntime/blob/c163fe678c58dc3a268cfa5deb93b8c8dedac2f1/src/core/internal/lifetime.d#L181 with LDC, which allows comparing vectors for equality/identity (unlike DMD).

Previously, `float4` would fail to compile (expected `ireal`), and `int4` would print some scalar integer.

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -191,6 +191,19 @@ private string miniFormat(V)(const scope ref V v)
     {
         return v ? "true" : "false";
     }
+    else static if (is(V == __vector(ET[N]), ET, size_t N))
+    {
+        string msg = "[";
+        foreach (i; 0 .. N)
+        {
+            if (i > 0)
+                msg ~= ", ";
+
+            msg ~= miniFormat(v[i]);
+        }
+        msg ~= "]";
+        return msg;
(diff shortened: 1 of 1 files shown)


src/core/internal/dassert.d
```diff
@@ -191,6 +191,19 @@ private string miniFormat(V)(const scope ref V v)
...
+        string msg = "[";
+        foreach (i; 0 .. N)
+        {
+            if (i > 0)
+                msg ~= ", ";
+
+            msg ~= miniFormat(v[i]);
+        }
+        msg ~= "]";
+        return msg;
```
MoonlightSentinel, line 205: Maybe forward the underlying array instead of copy+pasting the implementation:

```suggestion
        ET[] tmp = v.array;
        return miniFormat(tmp);
```

kinke: I wanted to avoid the cut-off after 30 elements (a bit unfortunate for `ubyte32` etc.).

MoonlightSentinel (approved):

PR #3489, opened 2021-06-03
add stdbool.d header inside core.stdc

aminya:
This adds the missing stdbool based on the C99 standard:

image
http://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf Page 265 of the PDF

There are three headers that are currently missing:
- stdbool: this PR adds it
- setjmp
- iso646

image

### DIFF
--- a/src/core/stdc/stdbool.d
+++ b/src/core/stdc/stdbool.d
@@ -11,7 +11,6 @@ extern (C):
 nothrow:
 @nogc:
 
-alias _Bool = bool;
 
 // bool is already defined in D
 // true is already defined in D
--- a/src/core/stdc/stdbool.d
+++ b/src/core/stdc/stdbool.d
@@ -1,0 +1,20 @@
+/**
+ * D header file for C99.
+ *
+ * Source:    $(DRUNTIMESRC core/stdc/stdbool.d)
+ * Standards: ISO/IEC 9899:1999 (E)
+ */
+module core.stdc.stdbool;
+
+extern (C):
+@trusted: // Types and constants only.
+nothrow:
+@nogc:
+
+alias _Bool = bool;
+
+// bool is already defined in D
+// true is already defined in D
+// false is already defined in D
+
+enum __bool_true_false_are_defined = 1;
--- a/mak/COPY
+++ b/mak/COPY
@@ -94,6 +94,7 @@ COPY=\
 	$(IMPDIR)\core\stdc\math.d \
 	$(IMPDIR)\core\stdc\signal.d \
 	$(IMPDIR)\core\stdc\stdarg.d \
+	$(IMPDIR)\core\stdc\stdbool.d \
 	$(IMPDIR)\core\stdc\stddef.d \
 	$(IMPDIR)\core\stdc\stdint.d \
 	$(IMPDIR)\core\stdc\stdio.d \
--- a/mak/DOCS
+++ b/mak/DOCS
@@ -89,6 +89,7 @@ DOCS=\
 	$(DOCDIR)\core_stdc_limits.html \
 	$(DOCDIR)\core_stdc_math.html \
 	$(DOCDIR)\core_stdc_stdarg.html \
+	$(DOCDIR)\core_stdc_stdbool.html \
 	$(DOCDIR)\core_stdc_stdint.html \
 	$(DOCDIR)\core_stdc_stdlib.html \
 	$(DOCDIR)\core_stdc_tgmath.html \
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -93,6 +93,7 @@ SRCS=\
 	src\core\stdc\math.d \
 	src\core\stdc\signal.d \
 	src\core\stdc\stdarg.d \
+	src\core\stdc\stdbool.d \
 	src\core\stdc\stddef.d \
 	src\core\stdc\stdint.d \
 	src\core\stdc\stdio.d \


ibuclaw:
Hi, thanks for your contribution. What is the rationale for this? Do we really need a module with a lone alias in?

aminya:
This allows safely replacing the `stdbool.h` with `stdbool.d` (e.g. in Dpp).

Another reason could be completeness and compliance with the standard. You can take a look at `stddef.d`, which similarly only has two aliases inside it.
https://github.com/aminya/druntime/blob/stdbool/src/core/stdc/stddef.d

dkorpel (commented):
> This allows for safely replacing stdbool.h with stdbool.d (e.g. in Dpp).

I'm afraid the equivalence just isn't there. In D, `bool` is the keyword that doesn't require imports, while in C `_Bool` is the keyword, and the type is not equivalent with D's bool. Imagine translating this C code:
```C
_Bool b; // 1

#ifndef __bool_true_false_are_defined 
const int true = 1;
#endif
```
D:
```D
_Bool b; // undefined identifier _Bool 
static if (!__bool_true_false_are_defined ) 
{
    const int true = 1; // parse error
}
```

Now take this C code:
```C
#include<stdbool.h>
bool x = (bool) 0.5; // = 1
int y = sizeof bool; // 1 or larger
```
D:
```D
import core.stdc.stdbool;
bool x = cast(bool) 0.5; // = 0
int y = bool.sizeof; // always 1
```
The import accomplishes nothing, and the behavior is different. Since dpp needs to have special logic to deal with C _Bool/bool anyways, I don't see how this module helps with "safely replacing stdbool.h".

aminya:
Using this PR, this code works:

```d
import core.stdc.stdbool;
import core.stdc.stdio: printf;

int main() {
  printf("%d", __bool_true_false_are_defined);
  return 0;
}
```

Equivalent to:

```c
#include <stdio.h>
#include <stdbool.h>

int main() {
  printf("%d", __bool_true_false_are_defined);
}
```

dkorpel:
> Using this PR, this code works:

I can't imagine you opened this PR because you encountered *that* kind of code 😛. Is there a dpp failure case of an actual C project that you are trying to solve?

aminya:
Well, I didn't argue that this PR revolutionizes anything special. 😄 It is just adding a header that is questionably missing from `core.stdc`. I want to make replacing the headers streamlined and bring the risks to 0.

Anyways, I don't want this PR to be judged based on my usage of it in Dpp. Please look at this from the perspective of the C99 standard.

ibuclaw:
(FYI, `_Bool` can be an `int`).

dkorpel:
> I want to make replacing the headers streamlined and bring the risks to 0.

That's not possible unfortunately, since C's `bool` is different from D's `bool`, and D's `bool` cannot be redefined. Since stdbool.d conflates C's `_Bool` with D's `bool`, it only adds risks: a programmer might think that replacing `bool` in C with `_Bool` is a good idea prior to translating to D, but then it turns out `core.stdc.stdbool._Bool` actually does not ensure the correct size, ABI and casting behavior.

src/core/stdc/stdbool.d
```diff
@@ -0,0 +1,20 @@
...
+ * Standards: ISO/IEC 9899:1999 (E)
+ */
+module core.stdc.stdbool;
+
+extern (C):
+@trusted: // Types and constants only.
+nothrow:
+@nogc:
+
+alias _Bool = bool;
```
aminya, line 14:
```suggestion
```

ibuclaw:
I think it's better if stdbool.h and iso646.h are just ignored for druntime.  The setjmp.h header could probably be added though, there's already Posix definitions, but not Windows/Mingw.

aminya:
Why do you want to ignore it when we have a working PR? I don't understand why it hurts to add another file to `core.stdc`.

ibuclaw:
> Why do you want to ignore it when we have a working PR? I don't understand why it hurts to add another file to `core.stdc`.

There should be a very good rationale for adding an empty module.

aminya:
> There should be a very good rationale for adding an empty module.

What's the downside?

RazvanN7:
> What's the downside?

Complex projects have a huge maintenance cost. It is crucial that each addition has a proven purpose, otherwise it is simply a maintenance burden. You may argue that this addition is small, however, if we start accepting every tiny drop, at some point the glass is going to be full.

aminya:
> Complex projects have a huge maintenance cost. It is crucial that each addition has a proven purpose, otherwise it is simply a maintenance burden. You may argue that this addition is small, however, if we start accepting every tiny drop, at some point the glass is going to be full.

Typical answer. 
https://twitter.com/ManishEarth/status/1358451655402901505

dkorpel:
> Typical answer.
> https://twitter.com/ManishEarth/status/1358451655402901505

I think that's a bit unfair, it's not like you're solving an issue or adding a feature that's being dismissed because of maintainability despite its merits. You're adding a near-empty module and the only reasons you gave so far are:
- to be able to translate `#include<stdbool.h>` to `import core.stdc.stdbool;`
- why not

I agree it *would* be nice if every C header had an equivalent D header, but that does not work [like I mentioned](https://github.com/dlang/druntime/pull/3489#pullrequestreview-675121398), so I think the false sense of compatibility will only confuse any programmer encountering the module instead of help them.

PR #3493, opened 2021-06-06
Fix 21996 - Skip message formatting for checkaction=context in finalizer

MoonlightSentinel:
The message formatting isn't `@nogc` yet and hence caused an `InvalidMemoryOperationError` when called from a finalizer.

CC @Geod24

### DIFF
--- a/src/core/internal/dassert.d
+++ b/src/core/internal/dassert.d
@@ -37,6 +37,10 @@ module core.internal.dassert;
  */
 string _d_assert_fail(A)(const scope string op, auto ref const scope A a)
 {
+    // Prevent InvalidMemoryOperationError when triggered from a finalizer
+    if (inFinalizer())
+        return "Assertion failed (rich formatting is disabled in finalizers)";
+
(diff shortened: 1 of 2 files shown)


Geod24 (approved):
Thanks! Should there be a special message to let the user know why the message is different from the expectations ?
E.g. `"Assertion failed (rich formatting disabled when in destructor)"`
Otherwise LGTM.

MoonlightSentinel:
Good idea. I've opted for a slightly different message because it doesn't apply to all destructors - only when called by the GC

thewilsonator (approved):

Geod24:
@MoonlightSentinel : Why is `inFinalizer` a template ?

MoonlightSentinel:
Not sure actually, the currrent state doesn't require it to be a template. Probably a remnant from testing / a previous iteration.

ibuclaw:
This PR caused a regression / exposed an issue in the compiler.

https://issues.dlang.org/show_bug.cgi?id=22700

PR #3494, opened 2021-06-06
Issue 21998 - Support more types during CTFE with checkaction=context

MoonlightSentinel:
Some small patches that allow `-checkaction=context` to work at compile time:
- use druntime's internal implementation instead of LibC
- avoid reinterpreting casts
- skip floats and `void[]`, return a warning instead

---

Best reviewed per commit.

~~Depends on dlang/dmd#12514~~

### DIFF
--- a/test/exceptions/src/assert_fail.d
+++ b/test/exceptions/src/assert_fail.d
@@ -63,7 +63,7 @@ void testPointers()
     static if ((void*).sizeof == 4)
         enum ptr = "0x12345670";
     else
-        enum ptr = "0x123456789ABCDEF0";
+        enum ptr = "0x123456789abcdef0";
 
     int* p = cast(int*) mixin(ptr);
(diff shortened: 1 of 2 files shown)


test/exceptions/src/assert_fail.d
```diff
@@ -29,6 +29,7 @@ void testIntegers()()
     test(uint.min, uint.max, "0 != 4294967295");
     test(long.min, long.max, "-9223372036854775808 != 9223372036854775807");
     test(ulong.min, ulong.max, "0 != 18446744073709551615");
+    if (!__ctfe)
     test(shared(ulong).min, shared(ulong).max, "0 != 18446744073709551615");
```
MoonlightSentinel, line 33: Disabled this test during CTFE because `shared` requires additional changes.

thewilsonator (approved):

Geod24:
This pull request introduced a regression: https://issues.dlang.org/show_bug.cgi?id=22300

PR #3497, opened 2021-06-14
[stable] hashOf does not work on enum types whose base type is a SIMD vector

n8sh:
Also incidentally do not pass vectors by ref to `toHash`.

### DIFF
--- a/src/core/internal/hash.d
+++ b/src/core/internal/hash.d
@@ -130,16 +130,10 @@ private template canBitwiseHash(T)
 
 //enum hash. CTFE depends on base type
 size_t hashOf(T)(auto ref T val, size_t seed = 0)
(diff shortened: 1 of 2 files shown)


src/core/internal/hash.d
```diff
@@ -130,16 +130,10 @@ private template canBitwiseHash(T)
 
 //enum hash. CTFE depends on base type
 size_t hashOf(T)(auto ref T val, size_t seed = 0)
-if (is(T EType == enum) && (!__traits(isScalar, T) || is(T == __vector)))
```
n8sh, line 133: The original code didn't behave as intended because `is(T == __vector)` is always false if `is(T == enum)`.

RazvanN7 (approved):

PR #3498, opened 2021-06-14
dwarf.d: remove spurious trailing comma

WalterBright:
Bad style exploiting what appears to be a bug in the D front end.

### DIFF
--- a/src/core/internal/backtrace/dwarf.d
+++ b/src/core/internal/backtrace/dwarf.d
@@ -1059,10 +1059,7 @@ LineNumberProgram readLineNumberProgram(ref const(ubyte)[] data) @nogc nothrow
         data.readULEB128(); // last mod
         data.readULEB128(); // file len
 
-        return SourceFile(
-            file,
-            dirIndex,
-        );
+        return SourceFile(file, dirIndex);
     }
     lp.sourceFiles = readSequence!readFileNameEntry(data);
 


WalterBright:
```
std/stdio.d(299): Error: undefined identifier `flockfile` in module `core.sys.posix.stdio`
```
What the heck? That has nothing to do with this trivial PR.

wilzbach:
It's not a bug, it's a feature.

WalterBright:
If it's used exactly once, it's not much of a feature.

CyberShadow:
More uses here: https://github.com/dlang/dmd/pull/12684#issuecomment-860988350

thewilsonator:
This might be bad style for just two arguments, but as noted this is a feature not a bug. 
It has the same use as trailing comma for enum declarations.

thewilsonator:
>What the heck? That has nothing to do with this trivial PR.

How many commits between `trailingComma~1` and `dlang/master`? Maybe try rebasing?

thewilsonator (approved):

Geod24 (changes requested):
Not an oversight, very much on purpose, as there is another usage of it in the commit that introduced it:
https://github.com/dlang/druntime/commit/f6534412ac65fb1f343d5a71ec8779faf1c69662#diff-1b94c982760aac40515b2b539895bf36e83f45159711f983f09290277a996587R91-R96

And we established it's not a bug, but a feature introduced 10 years ago as a fix to issue 2477.

DoctorNoobingstoneIPresume:
FWMOIW (For Whatever My Opinion Is Worth), this feature makes it easier to write some code-generating (`mixin`) code.

Plus, if code is indented like this:

```
f
(
    very-long-arg1,
    veey-long-arg2,
    very-long-arg3,
);
```
that trailing comma ensures we can add `very-long-arg4` without affecting the history of the `very-long-arg3` line.

Geod24:
The DMD PR has been closed, so closing this too.

PR #3501, opened 2021-06-23
[stable] hashOf(S) can segfault if S.toHash is forwarded via 'alias this' to a receiver which may be null

n8sh:


thewilsonator (approved):
Does this have a bugzilla number?

PR #3189, opened 2020-08-06
Update `LineNumberProgram` up to DWARF v5

Luhrel:
Needed for https://github.com/dlang/dmd/pull/11509

PR #3502, opened 2021-06-24
[stable] Fix Issue 22081 - Utterly broken DWARF v5 support

kinke:
This has actually been tested to work with LDC, fixing the `assert(0)` and **numerous** terrible issues of #3189 after looking at the [DWARF v5 spec](http://dwarfstd.org/doc/DWARF5.pdf) (section 6.2.4).

LLVM apparently puts all directory/file name strings into the new `.debug_line_str` section (`DW_FORM_line_strp` form code, one of various encoding options); reading the strings from there would require mmapping that section too, not just the `.debug_line` one.

So at least with LLVM, *all* source file paths (of according DWARF v5 line-number programs) in the exception backtrace string currently show up as `<unknown dir>/<unknown file>`.

### DIFF
--- a/src/core/internal/backtrace/dwarf.d
+++ b/src/core/internal/backtrace/dwarf.d
@@ -317,7 +317,9 @@ void resolveAddresses(const(ubyte)[] debugLineSectionData, Location[] locations,
                     // Can be called with either `locInfo` or `lastLoc`
                     void update(const ref LocationInfo match)
                     {
-                        const sourceFile = lp.sourceFiles[match.file - 1];
+                        // File indices are 1-based for DWARF < 5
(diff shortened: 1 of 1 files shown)


thewilsonator:
Tests?

kinke:
LOL! No f*ckin test for #3189, for something that should never have been merged in the first place (the previously - completely wrongly - parsed stuff wasn't used at all and should have been noticed during review).

Test requires DWARF v5 support - I don't dare looking at what DMD emits with `-gdwarf=5`, it's almost certainly totally wrong as it comes from the same author. So we'd have to rely on the C(++) compiler to produce DWARF v5 for us. Not sure since when it's considered to be stable, as gcc has only switched since v11, and clang apparently defaults to v4 still.

thewilsonator (approved):
Looks like a spurious failure

PR #3507, opened 2021-07-05
Add `return` to moduleinfo methods returning `this`

dkorpel:
Blocking https://github.com/dlang/dmd/pull/12812

`ModuleInfo.addrOf` is a return-ref method returning a pointer to `this` with an offset. Functions that return `this.addrOf` should therefore also be ref-return.

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -2180,7 +2180,7 @@ const:
      * Returns:
      *  array of pointers to the ModuleInfo's of modules imported by this one
      */
-    @property immutable(ModuleInfo*)[] importedModules() nothrow pure @nogc
+    @property immutable(ModuleInfo*)[] importedModules() return nothrow pure @nogc
     {
         if (flags & MIimportedModules)
         {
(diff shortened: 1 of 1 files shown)


dkorpel:
Note that while the functions are not `@safe`, they are still partially checked by dip1000 because they have 'default safety'. An alternative fix would be to explicitly mark the methods `@system`. 

It seems that `return` is not needed for functions of this form:
```D
return flags & MItlsctor ? *cast(typeof(return)*)addrOf(MItlsctor) : null;
```
I'm not sure if that's because it's `@system` or because there's a hole in dip1000 + the ternary operator, I will take a look.

MoonlightSentinel:
> Note that while the functions are not `@safe`, they are still partially checked by dip1000 because they have 'default safety'. An alternative fix would be to explicitly mark the methods `@system`.

The real fix is to make DMD adhere to the spec and not do these checks in `@system`. There's no such thing as `default` safety - it's a mere implementation detail.

MoonlightSentinel (approved):
`return` is appropriate nevertheless

Geod24:
> default safety

[There's a bugzilla entry for this](https://issues.dlang.org/show_bug.cgi?id=19873).

dkorpel:
The stack overflow test of [issue 20855](https://issues.dlang.org/show_bug.cgi?id=20855) fails on the auto-tester, nice.

RazvanN7 (approved):

PR #3509, opened 2021-07-06
Fix Issue 22107 - [scope][dip1000] Can't .dup an array of structs wit…

CyberShadow:
…h impure copy constructor

### DIFF
--- a/src/object.d
+++ b/src/object.d
@@ -3491,7 +3491,7 @@ private U[] _dup(T, U)(scope T[] a) pure nothrow @trusted if (__traits(isPOD, T)
     return *cast(U[]*) &arr;
 }
 
-private U[] _dupCtfe(T, U)(T[] a)
+private U[] _dupCtfe(T, U)(scope T[] a)
 {
     static if (is(T : void))
         assert(0, "Cannot dup a void[] array at compile time.");
@@ -3522,6 +3522,21 @@ private U[] _dup(T, U)(T[] a) if (!__traits(isPOD, T))
     } ();
 
     return res;
+}
+
+// https://issues.dlang.org/show_bug.cgi?id=22107
+@safe unittest
+{
+    static int i;
+    @safe struct S
+    {
+        this(this) { i++; }
+    }
+
+    void fun(scope S[] values...) @safe
+    {
+        values.dup;
+    }
 }
 
 // HACK:  This is a lie.  `_d_arraysetcapacity` is neither `nothrow` nor `pure`, but this lie is


CyberShadow:
CC @dkorpel

dkorpel:
I thought `scope` would be inferred now [issue 21209](https://issues.dlang.org/show_bug.cgi?id=21209) is fixed.

Is `dup` still `@safe` on structs that have a `@safe` but a non-`scope` copy-constructor with this change?

CyberShadow:
> I thought `scope` would be inferred now [issue 21209](https://issues.dlang.org/show_bug.cgi?id=21209) is fixed.

I didn't see that ... but, I can reproduce the issue even with DMD `master`, if it helps.

> Is `dup` still `@safe` on structs that have a `@safe` but a non-`scope` copy-constructor with this change?

Sorry, not sure how to test this.

dkorpel:
> Sorry, not sure how to test this.

Actually, since the array is `scope`, its elements are never `scope`, so it doesn't matter. The only tricky case I can think of is:
```D
@safe:
struct S
{
    int x;
    int* y;

    this(ref S old)
    {
        y = &old.x;
    }
}

void main()
{   
    S[] oldArr = new S[1]; 
    S[] newArr = oldArr[].dup;
    assert(newArr[0].y == &oldArr[0].x); // allowed since it's on the heap

    S[1] oldArr1; // on the stack
    S[] newArr1 = oldArr1[].dup;  // now it's not allowed
}
```
This is horrible code, but the first case is technically valid.

CyberShadow:
Well, I don't know what's going on here, then. Especially considering that the "fix" is in the CTFE helper, and none of this pertains CTFE.

CyberShadow:
> ```d
>     S[] newArr1 = oldArr1[].dup;  // now it's not allowed
> ```

And now it is allowed with the changes in this PR, which I understand is incorrect. But then why wasn't this flagged as an error somewhere else?

dkorpel:
> Especially considering that the "fix" is in the CTFE helper, and none of this pertains CTFE.

The `if (__ctfe)` branch is still taken as far as escape analysis is concerned

dkorpel:
> And now it is allowed with the changes in this PR, which I understand is incorrect. But then why wasn't this flagged as an error somewhere else?

I think the fact that it didn't compile is completely accidental: it didn't compile for the same reason the correct case in this PR's unittest didn't compile. My copy constructor example is actually invalid, you're not allowed to assign a pointer to a `ref` to a member. I think that's a manifestation of [issue 20245](https://issues.dlang.org/show_bug.cgi?id=20245).

dkorpel (approved):

dkorpel:
I just noticed that this is also needed to make druntime unittests compile with https://github.com/dlang/dmd/pull/12010.

thewilsonator (approved):

